PR #25553, opened 2023-05-19
datapath: support Hybrid-DSR mode for Geneve dispatch

julianwiedmann:
We've initially had DSR with info dispatch via IP option/extension. Here we support "full" DSR, and also a hybrid mode where TCP connections use DSR, but UDP traffic continues to use NAT and its replies pass back through the LB node.

With https://github.com/cilium/cilium/pull/23890 we've added info dispatch via Geneve Options. Here DSR traffic gets Geneve-encapsulated and has DSR info added as needed. But this currently only supports "full" DSR.

Adding support for hybrid mode is easy enough - it's just a few changes in the agent's option processing, we don't even need to touch the BPF datapath.

```release-note
Add support for Hybrid mode when using DSR with Geneve dispatch.
```

### DIFF
--- a/pkg/datapath/linux/config/config.go
+++ b/pkg/datapath/linux/config/config.go
@@ -143,7 +143,7 @@ func (h *HeaderfileWriter) WriteNodeConfig(w io.Writer, cfg *datapath.LocalNodeC
 	encapProto := option.Config.TunnelProtocol
 	if !option.Config.TunnelingEnabled() &&
 		option.Config.EnableNodePort &&
-		option.Config.NodePortMode == option.NodePortModeDSR &&
+		option.Config.NodePortMode != option.NodePortModeSNAT &&
 		option.Config.LoadBalancerDSRDispatch == option.DSRDispatchGeneve {
 		encapProto = option.TunnelGeneve
 	}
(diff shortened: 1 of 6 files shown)


julianwiedmann:
@ysksuzuki I don't think we discussed hybrid mode at all during https://github.com/cilium/cilium/pull/23890. Was there a reason we didn't enable it right away? :thinking:

ysksuzuki:
> Was there a reason we didn't enable it right away?

I think it was just an oversight🙈

pkg/datapath/linux/config/config.go
```diff
@@ -143,7 +143,7 @@ func (h *HeaderfileWriter) WriteNodeConfig(w io.Writer, cfg *datapath.LocalNodeC
 	encapProto := option.Config.TunnelProtocol
 	if !option.Config.TunnelingEnabled() &&
 		option.Config.EnableNodePort &&
-		option.Config.NodePortMode == option.NodePortModeDSR &&
+		option.Config.NodePortMode != option.NodePortModeSNAT &&
```
lmb, line 146: This is more concise than `== DSR || == Hybrid` but maybe less future proof?

lmb (commented):
sig-datapath files LGTM.

pkg/datapath/linux/config/config.go
julianwiedmann: Ack, not a big fan of it either - for now, I believe this is in sync with how the remaining code does it. But I can follow up with some `NodePortUsesDSR()` helper.

ysksuzuki (approved):
LGTM

squeed (approved):

viktor-kurchenko (approved):
LGTM.

dylandreimerink (approved):
LGTM

pkg/datapath/linux/config/config.go
dylandreimerink: +1. If we are going to make setting more logical, I also vote for not using the name NodePort to refer to everything related to forwarding traffic

PR #25049, opened 2023-04-21
clustermesh-apiserver: rework identities, endpoints and nodes synchronization to improve performance

giorio94:
This PR modifies the clustermesh-apiserver implementation to use the Resource[T] abstraction to retrieve CiliumIdentities, CiliumNodes and CiliumEndpoints from Kubernetes, rather than a vanilla informer, and synchronize them through the WorkqueueSyncStore. This ensures consistent behavior across all resource types (i.e., the retry logic is implemented by the SyncStore, rather than through the Resource[T] underlying workqueue).

This approach has several advantages, including:
* Enables retrying kvstore operations in case of temporary failures;
* Performs coalescencing in case of multiple updates for the same key;
* Simplifies the overall logic, since conversions and sanity checks are already performed by the Resource[T] abstraction.

Please, refer to the commit descriptions for more details about the individual changes.

Related: #25562

```release-note
clustermesh-apiserver: rework identities, endpoints and nodes synchronization to improve performance
```

PR #25554, opened 2023-05-19
clustermesh-apiserver: extract kvstore client initialization and heartbeat logic in separate cells

giorio94:
This PR extracts the kvstore client initialization and heartbeat logic from the clustermesh-apiserver in two separate cells, to better integrate in the hive paradigm and allow them to be reused by other components. In this phase, the clustermesh-apiserver still relies on a global variable for the kvstore client, which is planned to be removed with subsequent refactoring (in particular, it involves much less changes once https://github.com/cilium/cilium/pull/25049 is in).

```release-note
clustermesh-apiserver: extract kvstore client initialization and heartbeat logic in separate cells
```

### DIFF
--- a/clustermesh-apiserver/main.go
+++ b/clustermesh-apiserver/main.go
@@ -56,6 +56,7 @@ import (
 	nodeTypes "github.com/cilium/cilium/pkg/node/types"
 	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/pprof"
+	"github.com/cilium/cilium/pkg/promise"
 )
 
 type configuration struct {
@@ -108,6 +109,11 @@ var (
 	ciliumNodeStore *store.SharedStore
 
 	identityStore = cache.NewStore(cache.DeletionHandlingMetaNamespaceKeyFunc)
+
+	// Holds the backend retrieved from the promise. This global variable is used
+	// temporarily while the refactoring of the clustermesh-apiserver is in progress
+	// to reduce the amount of modifications required in this first step.
+	backend kvstore.BackendOperations
 )
 
 func init() {
@@ -121,6 +127,10 @@ func init() {
 		gops.Cell(defaults.GopsPortApiserver),
 		k8sClient.Cell,
 		k8s.SharedResourcesCell,
+
+		kvstore.Cell(kvstore.EtcdBackendName),
+		cell.Provide(func() *kvstore.ExtraOptions { return nil }),
+
 		healthAPIServerCell,
 		cmmetrics.Cell,
 		usersManagementCell,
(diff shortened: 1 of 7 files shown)


pkg/option/config.go
```diff
@@ -1356,9 +1356,16 @@ func LogRegisteredOptions(vp *viper.Viper, entry *logrus.Entry) {
 	keys := vp.AllKeys()
 	sort.Strings(keys)
 	for _, k := range keys {
-		v := vp.GetStringSlice(k)
-		if len(v) > 0 {
-			entry.Infof("  --%s='%s'", k, strings.Join(v, ","))
+		ss := vp.GetStringSlice(k)
```
aanm: Should we move this commit into a separate PR so that it can be backported to all branches?

aanm (approved):
just one small nit.

pkg/option/config.go
giorio94: It actually seems that `flags.StringToString` should not be used (https://github.com/cilium/cilium/blob/main/contrib/scripts/check-viper.sh#L20-L23). Let me try to figure out how to make `option.NewNamedMapOptions` work for  hive flags. In that way, this commit should no longer be necessary.

thorn3r (approved):
excellent work! :+1: 
looks like at least one of your CI failures is due to: https://github.com/cilium/cilium/issues/25483

YutaroHayakawa (approved):
LGTM!

pkg/option/config.go
giorio94: I've opened https://github.com/cilium/cilium/pull/25643 which includes the logic to correctly parse StringMap flags through the hive machinery, and the commit to fix the logging of that type of flags. Backporting is not required, since StringMap flags are not currently used (an intermediate wrapper is used to support both json and KV encoding for legacy flags).

asauber (approved):
LGTM. I appreciate that the temporary global has a good comment.

giorio94:
Rebased onto main to include the changes introduced in https://github.com/cilium/cilium/pull/25643

giorio94:
Rebased onto main to fix a conflict. I've additionally dropped a leftover kvstore initialization, which caused the spurious creation of an additional etcd client:

```diff
--- a/clustermesh-apiserver/main.go
+++ b/clustermesh-apiserver/main.go
@@ -576,21 +588,19 @@ func startServer(startCtx hive.HookContext, clientset k8sClient.Clientset, servi
 	}
 
 	var err error
-	if err = kvstore.Setup(context.Background(), "etcd", option.Config.KVStoreOpt, nil); err != nil {
-		log.WithError(err).Fatal("Unable to connect to etcd")
-	}
 
 	config := cmtypes.CiliumClusterConfig{
 		ID: cfg.clusterID,
 	}
@@ -602,6 +612,7 @@ func startServer(startCtx hive.HookContext, clientset k8sClient.Clientset, servi
 	}
 
 	ciliumNodeStore, err = store.JoinSharedStore(store.Configuration{
+		Backend:    backend,
 		Prefix:     nodeStore.NodeStorePrefix,
 		KeyCreator: nodeStore.KeyCreator,
 	})
```

PR #25555, opened 2023-05-19
metrics: Add k8s client rate limiter latency metric

ysksuzuki:
The rate limiter metrics visualize the extent of delays caused by the k8s client-side rate limit, providing valuable insights for making decisions on configuration adjustments.

~~Also, this commit fixes a conflict with controller-runtime on the metrics registration.~~

      GitHub issue.

### DIFF
--- a/pkg/k8s/watchers/watcher.go
+++ b/pkg/k8s/watchers/watcher.go
@@ -96,9 +96,9 @@ func init() {
 	k8s_metrics.Register(k8s_metrics.RegisterOpts{
 		ClientCertExpiry:      nil,
 		ClientCertRotationAge: nil,
-		RequestLatency:        &k8sMetrics{},
-		RateLimiterLatency:    nil,
-		RequestResult:         &k8sMetrics{},
+		RequestLatency:        &requestLatencyAdapter{},
+		RateLimiterLatency:    &rateLimiterLatencyAdapter{},
+		RequestResult:         &resultAdapter{},
 	})
 }
 
@@ -320,15 +320,24 @@ func NewK8sWatcher(
 	}
 }
 
-// k8sMetrics implements the LatencyMetric and ResultMetric interface from
-// k8s client-go package
-type k8sMetrics struct{}
+// requestLatencyAdapter implements the LatencyMetric interface from k8s client-go package
+type requestLatencyAdapter struct{}
 
-func (*k8sMetrics) Observe(_ context.Context, verb string, u url.URL, latency time.Duration) {
+func (*requestLatencyAdapter) Observe(_ context.Context, verb string, u url.URL, latency time.Duration) {
 	metrics.KubernetesAPIInteractions.WithLabelValues(u.Path, verb).Observe(latency.Seconds())
 }
 
-func (*k8sMetrics) Increment(_ context.Context, code string, method string, host string) {
+// rateLimiterLatencyAdapter implements the LatencyMetric interface from k8s client-go package
+type rateLimiterLatencyAdapter struct{}
+
+func (c *rateLimiterLatencyAdapter) Observe(_ context.Context, verb string, u url.URL, latency time.Duration) {
+	metrics.KubernetesAPIRateLimiterLatency.WithLabelValues(u.Path, verb).Observe(latency.Seconds())
+}
+
+// resultAdapter implements the ResultMetric interface from k8s client-go package
+type resultAdapter struct{}
+
+func (r *resultAdapter) Increment(_ context.Context, code, method, host string) {
 	metrics.KubernetesAPICallsTotal.WithLabelValues(host, method, code).Inc()
 	// The 'code' is set to '<error>' in case an error is returned from k8s
 	// more info:
--- a/Documentation/observability/metrics.rst
+++ b/Documentation/observability/metrics.rst
@@ -450,6 +450,17 @@ Name                                        Labels                              
 ``k8s_terminating_endpoints_events_total``                                                     Enabled    Number of terminating endpoint events received from Kubernetes
 =========================================== ================================================== ========== ========================================================
 
+Kubernetes Rest Client
+~~~~~~~~~~~~~~~~~~~~~~
+
+============================================= ============================================= ========== ===========================================================
+Name                                          Labels                                        Default    Description
+============================================= ============================================= ========== ===========================================================
+``k8s_client_api_latency_time_seconds``       ``path``, ``method``                          Enabled    Duration of processed API calls labeled by path and method
+``k8s_client_rate_limiter_duration_seconds``  ``path``, ``method``                          Enabled    Kubernetes client rate limiter latency in seconds. Broken down by path and method
+``k8s_client_api_calls_total``                ``host``, ``method``, ``return_code``         Enabled    Number of API calls made to kube-apiserver labeled by host, method and return code
+============================================= ============================================= ========== ===========================================================
+
 IPAM
 ~~~~
 
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -444,6 +444,9 @@ var (
 	// to the kube-apiserver
 	KubernetesAPIInteractions = NoOpObserverVec
 
+	// KubernetesAPIRateLimiterLatency is the client side rate limiter latency metric
+	KubernetesAPIRateLimiterLatency = NoOpObserverVec
+
 	// KubernetesAPICallsTotal is the counter for all API calls made to
 	// kube-apiserver.
 	KubernetesAPICallsTotal = NoOpCounterVec
@@ -613,6 +616,7 @@ type LegacyMetrics struct {
 	KubernetesEventProcessed         metric.Vec[metric.Counter]
 	KubernetesEventReceived          metric.Vec[metric.Counter]
 	KubernetesAPIInteractions        metric.Vec[metric.Observer]
+	KubernetesAPIRateLimiterLatency  metric.Vec[metric.Observer]
 	KubernetesAPICallsTotal          metric.Vec[metric.Counter]
 	KubernetesCNPStatusCompletion    metric.Vec[metric.Observer]
 	TerminatingEndpointsEvents       metric.Counter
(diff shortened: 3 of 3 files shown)


ysksuzuki:
Hit https://github.com/cilium/cilium/issues/25524

aojea:
watchers latency is not really important, since they are long lived sessions 

The metrics for client-go when not used for watch can be loaded automatically just importing
```
	_ "k8s.io/component-base/metrics/prometheus/clientgo" // for client metric registration
```

ysksuzuki:
Hi @aojea , Thank you for your comment.

> watchers latency is not really important, since they are long lived sessions

This PR adds the rate limiter latency metric, and I think it's important because we have had a problem before with write operations being delayed by rate limits in our environment. I wanted to know if it's necessary to adjust the QPS configs because the default config in client-go is a bit low.

client-go QPS default is 5
https://github.com/kubernetes/client-go/blob/b46677097d03b964eab2d67ffbb022403996f4d4/rest/config.go#L44

controller-runtime QPS default is 20
https://github.com/kubernetes-sigs/controller-runtime/blob/f6f37e6cc1ec7b7d18a266a6614f86df211b1a0a/pkg/client/config/config.go#L102 

> The metrics for client-go when not used for watch can be loaded automatically just importing

As far as I looked into this, we can't do this because controller-runtime also calls the client-go metrics registration [here](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/metrics/client_go_adapter.go#L52) and subsequent calls will be ignored. [Register](https://github.com/kubernetes/client-go/blob/master/tools/metrics/metrics.go#L124) can only be called once. 

This also calls `register`
https://github.com/kubernetes/component-base/blob/master/metrics/prometheus/restclient/metrics.go#L188

Aso, Cilium has already exported the client-go metrics under the names `cilium_k8s_client_api_latency_time_seconds` and `cilium_k8s_client_api_calls_total` [here](https://github.com/cilium/cilium/blob/main/pkg/metrics/metrics.go#L1184-L1204). We need to announce the breaking change if we use `k8s.io/component-base/metrics/prometheus/clientgo`.

aojea:
> This PR adds the rate limiter latency metric, and I think it's important because we have had a problem before with write operations being delayed by rate limits in our environment. I wanted to know if it's necessary to adjust the QPS configs because the default config in client-go is a bit low.

my point is that this file seems to be used only for watchers, and watchers are not affected by the rate limiter,

https://github.com/kubernetes/kubernetes/blob/b2522655b3ab70de0a5cc1af0d9ca71e41b25c97/staging/src/k8s.io/client-go/rest/request.go#L702-L705

 but I'm not familiar enough with cilium codebase to know if this client is used for more operations

> As far as I looked into this, we can't do this because controller-runtime also calls the client-go metrics registration [here](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/metrics/client_go_adapter.go#L52) and subsequent calls will be ignored. [Register](https://github.com/kubernetes/client-go/blob/master/tools/metrics/metrics.go#L124) can only be called once.

Showing my ignorance here about cilium codebase :) my point is about encouraging all projects that import client-go to use all the existing client-go metrics, it's an area I was working on the past to improve the "client-side network visibility" and I'm happy to get feedback and new ideas for improvements, as you can see there are some few interesting metrics that can be useful to troubleshoot client networking problems

https://github.com/kubernetes/kubernetes/pull/117295
https://github.com/kubernetes/kubernetes/pull/108296

ysksuzuki:
> my point is that this file seems to be used only for watchers, and watchers are not affected by the rate limiter,

Thank you for clarifying it. This client is used for the write operations too.

cilium_k8s_client_api_latency_time_seconds_bucket

> my point is about encouraging all projects that import client-go to use all the existing client-go metrics

Yeah, it makes sense to import `k8s.io/component-base/metrics/prometheus/clientgo` and use all the existing client-go metrics. Cilium v1.13 started using controller-runtime and seems to have a conflict with the metrics registration. I think we should create an issue on the controller-runtime repo and encourage them to use all the existing client-go metrics instead of registering `RequestResult` only.

https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/metrics/client_go_adapter.go#L52

ysksuzuki:
> I think we should create an issue on the controller-runtime repo and encourage them to use all the existing client-go metrics instead of registering RequestResult only.

contoller-runtime introduced `rest_client_request_duration_seconds`, `rest_client_request_size_bytes` and `rest_client_response_size_bytes`, but reverted because the cardinality of those metrics were too high when CR is used to communicate with a lot of different clusters. Considering this, we should not register all the client-go metrics by default.

https://github.com/kubernetes-sigs/controller-runtime/pull/2298

aojea:
> Considering this, we should not register all the client-go metrics by default.

wow, I didn't expect that, client-go use to talk with apiserver only.
It seems they are reusing the same client for multiple cluster that of course is not something scales well

ysksuzuki:
They seem to intentionally override the host label to keep the cardinality low but in fact, it was too high, then reverted. I'm not sure why it turned out that way.

https://github.com/kubernetes-sigs/controller-runtime/pull/2217
> I think the cardinality should be fine as the host label intentionally only contains the "template url".

tommyp1ckles (approved):

ysksuzuki:
Resolved the conflict and rebased

ysksuzuki:
Hi @tommyp1ckles , I resolved a conflict and rebased my branch. Please take a look

joestringer:
I don't see any docs updates for this, does it make sense to add some line to `Documentation/observability/metrics.rst` to help users to find these new metrics?

I've added the release-note/minor label to the PR because there are user-facing changes here. I also tried to kick the ginkgo CI into gear (the only remaining "required" test job). If it doesn't run & provide results in the next 20-30 minutes then it might just need the PR to be rebased.

ysksuzuki:
> I don't see any docs updates for this, does it make sense to add some line to Documentation/observability/metrics.rst to help users to find these new metrics?

Yeah, it makes sense. As far as I looked into it, there's no description for the  k8s rest client metrics. So I will add a section for it.

qmonnet (approved):
Doc change looks good :+1:

joestringer:
> Also, this commit fixes a conflict with controller-runtime on the metrics registration.

Typically we would suggest to propose a specific PR only for the fix, so that we can backport that fix to all releases that are affected. As I understand from our discussion during the community APAC call today, part of this change fixes a metrics bug in 1.13. Would you consider splitting that out and proposing as a dedicated PR so that we can ensure that gets fixed in the relevant releases?

joestringer (approved):
In principle I think the @cilium/sig-policy codeowner request is only added because of the metrics docs page. I can approve on behalf of that codeowner.

I didn't realize that there was a fix in this PR since it was also adding a new metrics feature. I'd suggest submitting fixes in separate PRs from features, so that we can more easily identify them and review in context of the fix, and then we can also backport those fixes to affected releases.

ysksuzuki:
> Would you consider splitting that out and proposing as a dedicated PR so that we can ensure that gets fixed in the relevant releases?

Sure, I will do that. I should have done that in the first place😅

ysksuzuki:
> Also, this commit fixes a conflict with controller-runtime on the metrics registration.

I tested this change with the latest main and found that the conflict issue doesn't persist. That's because the order in which the init function is called has changed. I couldn't identify which change caused the dependency graph to change, though.

@joestringer It seems that the conflict issue with controller-runtime is not the case anymore when it comes to the latest main. (It still persists with v1.13.4 though)  I think I need to modify this PR. I'm sorry for causing confusion.


latest main: 
`github.com/cilium/cilium/pkg/k8s/watchers` init -> `github.com/cilium/cilium/operator/metrics` -> `sigs.k8s.io/controller-runtime/pkg/metrics` init

v1.13.4:
`github.com/cilium/cilium/operator/metrics` -> `sigs.k8s.io/controller-runtime/pkg/metrics` init -> `github.com/cilium/cilium/pkg/k8s/watchers` init -

ysksuzuki:
I can see the rate limiter metrics
rate-limiter

joestringer:
> It seems that the conflict issue with controller-runtime is not the case anymore when it comes to the latest main. (It still persists with v1.13.4 though) 

OK I see. If it's only a regression in v1.13 now, then one option could be to propose the fix only for the v1.13 branch. That way we can ensure that users can upgrade to v1.13.x releases without breaking those metrics.

ysksuzuki:
>  If it's only a regression in v1.13 now, then one option could be to propose the fix only for the v1.13 branch. That way we can ensure that users can upgrade to v1.13.x releases without breaking those metrics.

I will raise a PR for the v1.13 branch

ysksuzuki:
I have created a fix for v1.13
https://github.com/cilium/cilium/pull/26412

pkg/k8s/watchers/watcher.go
```diff
@@ -93,13 +93,27 @@ func init() {
...
+	// Without calling Register function here, controller-runtime will have
+	// a chance to override our metrics registration.
+	registerOps := k8s_metrics.RegisterOpts{
 		ClientCertExpiry:      nil,
 		ClientCertRotationAge: nil,
-		RequestLatency:        &k8sMetrics{},
-		RateLimiterLatency:    nil,
-		RequestResult:         &k8sMetrics{},
-	})
+		RequestLatency:        &requestLatencyAdapter{},
```
joestringer, line 99: Does it make sense to separate this commit out into its own commit?

I'm still a little bit confused about how this change relates to the new metric, it seems like separate refactoring to address another issue.

But maybe the @cilium/sig-k8s reviewers can chime in here for closer inspection.

ysksuzuki: OK, ~~I will separate this change into its own commit.~~ Dropped the refactoring 

Although we can export the rate limiter metric without this change of the init function, there is a possibility that the dependency graph may change due to future changes, and the client-go metrics, including the rate limiter metric, may be missing, and it cannot be detected through testing. So I decided to include this change.

ysksuzuki: >  it seems like separate refactoring to address another issue.

But, yeah, you are right. I think I should tackle the conflict issue(with controller-runtime) on another issue. Let's drop this change and focus on adding new metric here.

ysksuzuki:
@joestringer Can we move this PR forward?

joestringer (approved):

PR #25563, opened 2023-05-19
hubble: Add GetNamespaces to observer API

### LINKED ISSUE #25266: Add GetNamespaces support to Hubble
We currently support querying flows, and nodes via hubble, but what would also be helpful would be the ability to query namespaces. This could be used by hubble UI and other tools (maybe even the CLI) to provide a list of namespaces that flows can be queried from.

chancez:
Fixes: #25266

```release-note
hubble: Add GetNamespaces to observer API
```

I also have a branch adding support to Hubble CLI ready once this is merged. https://github.com/cilium/hubble/tree/pr/chancez/get_namespaces


api/v1/observer/observer.proto
```diff
@@ -28,6 +28,9 @@ service Observer {
     // GetNodes returns information about nodes in a cluster.
     rpc GetNodes(GetNodesRequest) returns (GetNodesResponse) {}
 
+    // GetNamespaces returns information about namespaces in a cluster.
```
kaworu, line 31: I think the "memory period" of 1h and the order should be documented here as well.

daemon/cmd/hubble.go
```diff
@@ -172,7 +173,19 @@ func (d *Daemon) launchHubble() {
...
 
-	d.hubbleObserver, err = observer.NewLocalServer(payloadParser, logger,
+	namespaceManager := observer.NewNamespaceManager()
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		namespaceManager.Run(d.ctx)
+		wg.Done()
+	}()
+	defer wg.Wait()
```
kaworu: What happen on Hubble initialization failure? As far as I understand, we'll wait indefinitely on this `wg.Wait()` since `d.ctx` is not cancelled.

To me, the usage of `sync.WaitGroup` here indicate that we need something to finish cleanly from `namespaceManager.Run` before returning, but it doesn't seems to be the case. Instead, we want to shutdown the `namespaceManager` when we return. If correct, I would like to suggest

```suggestion
	ctx, cancel := context.WithCancel(d.ctx)
	defer cancel()
	go namespaceManager.Run(ctx)
```

pkg/hubble/observer/local_observer.go
```diff
@@ -681,3 +710,65 @@ func newRingReader(ring *container.Ring, req genericRequest, whitelist, blacklis
...
+}
+
+func NewNamespaceManager() *namespaceManager {
+	return &namespaceManager{
+		namespaces: make(map[string]namespaceRecord),
+	}
+}
+
+func (m *namespaceManager) Run(ctx context.Context) {
+	ticker := time.NewTicker(5 * time.Minute)
```
kaworu: please constantize this timeout

pkg/hubble/observer/local_observer.go
```diff
@@ -142,6 +153,19 @@ nextEvent:
...
+					Namespace: srcNs,
+					Cluster:   nodeTypes.GetClusterName(),
+				})
+			}
+			if dstNs := flow.GetDestination().GetNamespace(); dstNs != "" {
+				s.namespaceManager.AddNamespace(&observerpb.Namespace{
+					Namespace: dstNs,
+					Cluster:   nodeTypes.GetClusterName(),
+				})
+			}
```
kaworu: nit: this code would be cleaner in the `namespaceManager` itself, could we add a helper `func` taking a `*flowpb.Flow` and handling this logic?

pkg/hubble/observer/local_observer.go
```diff
@@ -681,3 +710,65 @@ func newRingReader(ring *container.Ring, req genericRequest, whitelist, blacklis
...
+		}
+	}
+}
+
+func (m *namespaceManager) GetNamespaces() []*observerpb.Namespace {
+	m.mu.RLock()
+	namespaces := make([]*observerpb.Namespace, 0, len(m.namespaces))
+	for _, ns := range m.namespaces {
+		namespaces = append(namespaces, ns.namespace)
+	}
```
kaworu: nit
```suggestion
	namespaces := make([]*observerpb.Namespace, len(m.namespaces))
	copy(namespaces, m.namespaces)
```

pkg/hubble/observer/local_observer.go
kaworu, line 1: This file is already long enough, can we extract the `namespaceManager` into its own file? I think it would also help wrt testing (e.g. timeout of 1h).

pkg/hubble/observer/local_observer.go
```diff
@@ -681,3 +710,65 @@ func newRingReader(ring *container.Ring, req genericRequest, whitelist, blacklis
...
+		case t := <-ticker.C:
+			// periodically remove any namespaces which haven't been seen in flows
+			// for the last hour
+			m.mu.Lock()
+			for key, record := range m.namespaces {
+				if record.added.Add(time.Hour).After(t) {
+					delete(m.namespaces, key)
+				}
+			}
+			m.mu.Unlock()
```
kaworu: I think using the ticker's time is cute, but a package-level `var NowFunc = time.Now` and `now := NowFunc()` before iterating would be effectively the same while easier to test.

I'm sure we'll be asked to make it configurable, but in the meantime let's add a constant as well for this 1h "timeout".

Finally, extracting this into a `namespaceManager` func would help wrt testing

```suggestion
			m.Cleanup(NowFunc())
```

````go
// package-level
var NowFunc = time.Now
func (m *namespaceManager) Cleanup(now time.Time) {
	m.mu.Lock()
	defer m.mu.Unlock()
	for key, record := range m.namespaces {
		if record.added.Add(time.Hour).After(now) {
			delete(m.namespaces, key)
		}
	}
}

File: pkg/hubble/observer/local_observer_test.go
Comment (kaworu) [inline pkg/hubble/observer/local_observer_test.go:1]: Can we add testing around the 1h "memory period" please?

File: pkg/hubble/observer/local_observer.go [hunk 7]
```diff
@@ -681,3 +710,65 @@ func newRingReader(ring *container.Ring, req genericRequest, whitelist, blacklis
...
+		}
+		return a.Namespace < b.Namespace
+	})
+	return namespaces
+}
+
+func (m *namespaceManager) AddNamespace(ns *observerpb.Namespace) {
+	m.mu.Lock()
+	key := ns.GetCluster() + "/" + ns.GetNamespace()
+	m.namespaces[key] = namespaceRecord{namespace: ns, added: time.Now()}
```
kaworu: We could add `now` as param for testing here too, and callers would use `NowFunc`

```suggestion
func (m *namespaceManager) AddNamespace(now time.Time, ns *observerpb.Namespace) {
	m.mu.Lock()
	key := ns.GetCluster() + "/" + ns.GetNamespace()
	m.namespaces[key] = namespaceRecord{namespace: ns, added: now}
```

pkg/hubble/relay/observer/server.go
```diff
@@ -231,6 +232,103 @@ func (s *Server) GetNodes(ctx context.Context, req *observerpb.GetNodesRequest)
 	return &observerpb.GetNodesResponse{Nodes: nodes}, nil
 }
 
+// GetNamespaces implements observerpb.ObserverClient.GetNamespaces.
+func (s *Server) GetNamespaces(ctx context.Context, req *observerpb.GetNamespacesRequest) (*observerpb.GetNamespacesResponse, error) {
```
kaworu, line 236: We should re-use the `namespaceManager` here in order to reduce the logic duplication. Something like:

1. create a `NewNamespaceManager()`
2. for each peer, add each ns to the manager (`(*namespaceManager).AddNamespace()` will handle the dedup and is already concurrency safe)
3. return the manager's `GetNamespaces` result (will handle the sorting).

What do you think?

kaworu (changes requested):
Thanks @chancez! Overall LGTM, I would like more testing around the `namespaceManager` and also the Relay part to be reworked as I think we can clean it up.

qmonnet (approved):
Doc update looks good. I only glanced quickly at the rest of the PR (and have no particular concern).

If anything, I'd like to have a bit more context and motivation in the commit description. The description used for https://github.com/cilium/cilium/issues/25266 would do just fine, for example.

daemon/cmd/hubble.go
chancez: Good point on initialization failure, and the approach to do cancellation seems like the right way to do it.

The waitGroup is just best practice from my perspective. In my opinion you should always ensure every go routine returns (even on shutdown). The waitGroup here is to ensure that it does. We just aren't doing a very (IMO) good job of it in the rest of this function currently (and I don't want to rewrite the whole function right now either).

This is to just avoid potential memory leaks. It's hard to know how code might be used in the future. For example, if we changed agent to restart hubble automatically on failure (right now it does not) then if we don't check the go routine returns before returning, then we might have an infinite number of go routines get started,.

pkg/hubble/observer/local_observer.go
chancez: Good point, I'll see how I can refactor this.

pkg/hubble/observer/local_observer.go
chancez: Yeah I was considering how to do the `Now()` check. I wish we had a `clock` interface, because I really dislike the globals, but I'll take that approach for now.

pkg/hubble/observer/local_observer.go
chancez: I think having `now` as a param is a bit of a leaky abstraction, instead we can make it a field on namespaceManager for testing. We might not use time in a different implementation of a NamespaceManager, so I think this would be  more flexible.

pkg/hubble/relay/observer/server.go
chancez: Hmm, I'm not sure I completely agree. The namespaceManager is keeping state over time, whereas the relay isn't, it's constructing everything within the RPC. We don't need expiration logic, etc, from the namespaceManager. I agree though that this could be refactored to share some more code between the two servers. The way they handle concurrency is a bit different too. The namespaceManager uses rwmutex because there's many readers, one writer, but in relay, we have many writers, zero readers (we read _after_ the results get aggregated). I don't think the same approach makes sense for both.

I think what would make more sense is to introduce a `NamespaceSet` or something which could reduce duplication some of the logic for managing the map, and then turning it into a deduplicated list.

pkg/hubble/observer/local_observer.go
chancez: Unfortunately, these aren't the same types, so we can't use copy.

pkg/hubble/relay/observer/server.go
chancez: Okay, with some refactoring I think it can be reused, and the RWMutex is semantically the same as the channel based approach, and the overhead of adding the time tracking for each namespace is probably negligible.

Documentation/internals/hubble.rst
```diff
@@ -49,11 +49,13 @@ as well as being exposed as a Kubernetes Service when enabled via TCP.
...
-``GetFlows``, ``GetNodes`` and ``ServerStatus``.  While ``ServerStatus`` and
-``GetNodes`` endpoints are pretty straightforward (they provides metrics and
-other information related to the running instance(s)), ``GetFlows`` is far more
-sophisticated and the more important one.
+The Observer service is the principal service. It provides four RPC endpoints:
+``GetFlows``, ``GetNodes``, ``GetNamespaces``  and ``ServerStatus``.
+
+* ``GetNodes`` returns a list of metrics and other information related for each Hubble instance
+* ``ServerStatus`` returns a summary the information in ``GetNodes``
+* ``GetNamespaces`` returns a list of namespaces that had network flows within the last 1 hour
```
kaworu: nit
```suggestion
* ``GetNamespaces`` returns a list of namespaces that had network flows within the last one hour
```

api/v1/observer/observer.proto
kaworu: thanks for the update, still missing the bits about ordering.

daemon/cmd/hubble.go
kaworu: Taking another look, before this patch, `launchHubble` is [executed into its own goroutine](https://github.com/cilium/cilium/blob/3e805306fc4125fca3398397acdf11a13f1a9600/daemon/cmd/daemon_main.go#L1871). In the happy path the "launch" goroutine setup a goroutine for the "local server" and a corresponding cleanup goroutine, another goroutine for the "remote server" and a corresponding cleanup goroutine, and then returns. So we start the "launch" goroutine, itself starting 4 other goroutines, and then the "launch" goroutine terminates.

My suggestion then doesn't work because when launchHubble returns then the context is cancelled, and the `namespaceManager` shutdown.

In the current version of the patch, the "launch" goroutine will be stuck on the deferred `wg.Wait()` until `d.ctx` is cancelled, cancelling the local `ctx` and shutting down the `namespaceManager` (which will eventually call `wg.Done()` and allow the "launch" goroutine to return from `launchHubble` and terminate).

If correct, although I agree with you wrt best practice on principle, practically I don't see the point of keeping the waitGroup. Technically it only make it so the "launch" goroutine stick around, which have any purpose anymore. As I see it it is one more goroutine that is sleeping or that we might leak.

pkg/hubble/relay/observer/server.go
kaworu: feels much cleaner thank you. I don't think there is anything wrong to re-use the `namespaceManager` here even though we do not care about time tracking.

pkg/hubble/observer/local_observer.go
kaworu: Ah yes my bad.

pkg/hubble/observer/namespace_manager_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// namespaces that we added should be returned, sorted
+	expected = []*observerpb.Namespace{
+		{Namespace: "ns-1"},
+		{Namespace: "ns-2"},
+	}
+	assert.Equal(t, expected, mgr.GetNamespaces())
+
+	// advance the clock by 30 minutes and verify our namespaces are still known (TTL is 1 hour)
+	advanceTime(30 * time.Minute)
```
kaworu: nit: could use `namespaceTTL / 2` here instead, so that updating the TTL won't break the test.

pkg/hubble/observer/namespace_manager_test.go
```diff
@@ -0,0 +1,87 @@
...
+	expected = []*observerpb.Namespace{
+		{Namespace: "ns-1"},
+		{Namespace: "ns-2"},
+		{Namespace: "ns-3"},
+		{Namespace: "ns-4"},
+	}
+	assert.Equal(t, expected, mgr.GetNamespaces())
+
+	// advance the clock another 31 minutes to put us past the hour
+	advanceTime(31 * time.Minute)
```
kaworu: ditto about computing the drift time based on `namespaceTTL`

pkg/hubble/observer/namespace_manager_test.go
```diff
@@ -0,0 +1,87 @@
...
+	// when we added ns-3 and ns-4 30 minutes ago
+	expected = []*observerpb.Namespace{
+		{Namespace: "ns-1"},
+		{Namespace: "ns-3"},
+		{Namespace: "ns-4"},
+	}
+	assert.Equal(t, expected, mgr.GetNamespaces())
+
+	// advance the clock another 31 minutes to put us past the hour again
+	advanceTime(31 * time.Minute)
```
kaworu: ditto about computing the drift time based on `namespaceTTL`

kaworu (changes requested):
Thanks for the update @chancez, looking good! A few more comments, and also:
1. [**golangci-lint** is failing](https://github.com/cilium/cilium/actions/runs/5050968499/jobs/9062298994?pr=25563):

    ```
    pkg/hubble/observer/namespace_manager_test.go:10: File is not `goimports`-ed with -local github.com/cilium/cilium (goimports)
  	    observerpb "github.com/cilium/cilium/api/v1/observer"
    ```
2. [**generate-api** CI is failing](https://github.com/cilium/cilium/actions/runs/5050968499/jobs/9062297955?pr=25563):

    ```
    Please run 'make generate-api generate-health-api generate-hubble-api generate-operator-api' and submit your changes
    ```

daemon/cmd/hubble.go
```diff
@@ -88,6 +89,8 @@ func (d *Daemon) getHubbleStatus(ctx context.Context) *models.HubbleStatus {
 }
 
 func (d *Daemon) launchHubble() {
+	ctx, cancel := context.WithCancel(d.ctx)
+	defer cancel()
```
pippolo84: `launchHubble` is not blocking, so If we use this new derived context and defer its cancellation, it will be cancelled as soon as `launchHubble` returns.
I don't think that is what we want, looking at how `ctx` is used in `namespaceManager.Run()` and in the two `select` blocks at lines 284 and 311.

daemon/cmd/hubble.go
```diff
@@ -171,8 +174,19 @@ func (d *Daemon) launchHubble() {
...
 		}
 	}
+	namespaceManager := observer.NewNamespaceManager()
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		namespaceManager.Run(ctx)
+		wg.Done()
+	}()
+	defer wg.Wait()
```
pippolo84: To manage background goroutines while remaining "sensitive" to external context cancellation (and ensuring proper cleanup) I suggest to look into pkg/controller. Alternatively, you can consider the new pkg/hive/job (I think the OneShot is what you're looking for here), even if it is targeted to be integrated with the hive/cell framework.

pippolo84 (changes requested):

api/v1/observer/observer.proto
chancez: Do we want to document ordering, or should we lave that unspecified in case we decide we don't want to sort anymore? Glad to do it either way.

daemon/cmd/hubble.go
chancez: Interesting, I rebuilt and tested locally with this change but didn't actually notice any issues. I guess that means the namespace manager just wasn't going to expire resources, so I didn't catch it. I'll take another look at this.

chancez:
Okay, I believe I got everything addressed. I opted to just remove the waitGroup from `launchHubble` and start the go routine without any guards about it returning, like the other go routines in the function. I'd rather spend the effort on refactoring all of `launchHubble`, perhaps as a Hive/Cell module, which presumably is something we'll need to do soon anyways.

pippolo84 (approved):
LGTM for agent related changes :100:

api/v1/observer/observer.proto
kaworu: I think users will rely on the order regardless of whether it is documented or not, so I would say let's document it. It's nice to output a consistent order.

kaworu (approved):
Thanks @chancez!

chancez:
/test-1.26-net-next

api/v1/observer/observer.proto
```diff
@@ -28,6 +28,11 @@ service Observer {
...
+    // The namespaces returned are namespaces which have had network flows in
+    // the last hour. The namespaces returned will be returned in sorted order.
```
rolinh:
```suggestion
    // the last hour. The namespaces are returned sorted by cluster name and namespace in ascending order.
```

Documentation/internals/hubble.rst
```diff
@@ -49,11 +49,13 @@ as well as being exposed as a Kubernetes Service when enabled via TCP.
...
 
-The Observer service is the principal service. It provides three RPC endpoints:
-``GetFlows``, ``GetNodes`` and ``ServerStatus``.  While ``ServerStatus`` and
-``GetNodes`` endpoints are pretty straightforward (they provides metrics and
-other information related to the running instance(s)), ``GetFlows`` is far more
-sophisticated and the more important one.
+The Observer service is the principal service. It provides four RPC endpoints:
+``GetFlows``, ``GetNodes``, ``GetNamespaces``  and ``ServerStatus``.
+
+* ``GetNodes`` returns a list of metrics and other information related for each Hubble instance
```
rolinh:
```suggestion
* ``GetNodes`` returns a list of metrics and other information related to each Hubble instance
```

pkg/hubble/observer/namespace_manager.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	m.mu.Unlock()
+}
+
+func (m *namespaceManager) GetNamespaces() []*observerpb.Namespace {
+	m.mu.RLock()
+	namespaces := make([]*observerpb.Namespace, 0, len(m.namespaces))
+	for _, ns := range m.namespaces {
+		namespaces = append(namespaces, ns.namespace)
+	}
```
rolinh, line 76: nit: no need for a loop here
```suggestion
	namespaces := make([]*observerpb.Namespace, len(m.namespaces))
	copy(namespaces, m.namespaces)
```

pkg/hubble/relay/observer/server.go
```diff
@@ -231,6 +232,52 @@ func (s *Server) GetNodes(ctx context.Context, req *observerpb.GetNodesRequest)
...
+	if md, ok := metadata.FromIncomingContext(ctx); ok {
+		ctx = metadata.NewOutgoingContext(ctx, md)
+	}
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
rolinh: Is this needed? It looks like the new context is only passed to the goroutines started with `g.Go` and we do wait for all of them to return with `g.Wait()` before the function returns.
```suggestion
```

rolinh (changes requested):
Thanks! lgtm overall but I left a few comments to address.

pkg/hubble/observer/namespace_manager.go
chancez: Alex made the same comment, this isn't a slice we're coping from, but a map, so we cannot use `copy()`.

pkg/hubble/relay/observer/server.go
chancez: I was just following the already established pattern. I believe the intent here is to ensure the context is cancelled when we return from the function. However, since we're waiting on the errgroup to return for all running routines, your right, we do not need this.

pkg/hubble/observer/namespace_manager.go
rolinh: Oops, my bad, sorry.

rolinh (approved):

pkg/hubble/observer/namespace_manager.go
```diff
@@ -0,0 +1,95 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			// periodically remove any namespaces which haven't been seen in flows
+			// for the last hour
+			m.cleanupNamespaces()
+		}
+	}
+}
```
joestringer, line 59: @chancez Did you consider using `pkg/controller` in order to run this logic? The benefits are that it's easier to get visibility into when this periodic logic runs, whether there are errors, and so on. pkg/controller will automatically hook the logic into metrics as well as registering the status each time it runs with the `cilium status` reporter, so that users can understand whether the logic ran, when it ran, and whether it's stuck.

chancez: No, but Hubble also is a bit of a snowflake in general, and hasn't had nearly as much focus as the rest of Cilium, including building better reusable primitives.

PR #25564, opened 2023-05-19
Updates egressgateway pkg to use netip.Addr

### LINKED ISSUE #24246: Convert net.IP to netip.Addr
Meta issue to track the conversion of net.IP to netip.Addr in our code base

See the [steps to contribute](https://github.com/cilium/cilium/issues/24246#issuecomment-1500445741) to this issue below.

danehans:
Updates the egressgateway pkg to use `netip.Addr` type instead of `net.IP`.

__Note:__ The `net` pkg can be removed from `pkg/egressgateway/manager_privileged_test.go` after node/type and netlink pkgs support `netip.Addr`.

Partially Fixes: https://github.com/cilium/cilium/issues/24246

### DIFF
--- a/pkg/egressgateway/endpoint.go
+++ b/pkg/egressgateway/endpoint.go
@@ -5,7 +5,7 @@ package egressgateway
 
 import (
 	"fmt"
-	"net"
+	"net/netip"
 
 	"k8s.io/apimachinery/pkg/types"
 
@@ -21,14 +21,14 @@ type endpointMetadata struct {
 	// Endpoint ID
 	id endpointID
 	// ips are endpoint's unique IPs
-	ips []net.IP
+	ips []netip.Addr
 }
 
 // endpointID includes endpoint name and namespace
 type endpointID = types.NamespacedName
 
 func getEndpointMetadata(endpoint *k8sTypes.CiliumEndpoint, identityLabels labels.Labels) (*endpointMetadata, error) {
-	var ipv4s []net.IP
+	var ipv4s []netip.Addr
 	id := types.NamespacedName{
 		Name:      endpoint.GetName(),
 		Namespace: endpoint.GetNamespace(),
@@ -40,7 +40,9 @@ func getEndpointMetadata(endpoint *k8sTypes.CiliumEndpoint, identityLabels label
 
 	for _, pair := range endpoint.Networking.Addressing {
 		if pair.IPV4 != "" {
-			ipv4s = append(ipv4s, net.ParseIP(pair.IPV4).To4())
+			if ip := netip.MustParseAddr(pair.IPV4); ip.Is4() {
+				ipv4s = append(ipv4s, ip)
+			}
 		}
 	}
 
--- a/pkg/egressgateway/manager.go
+++ b/pkg/egressgateway/manager.go
@@ -6,12 +6,13 @@ package egressgateway
 import (
 	"context"
 	"fmt"
-	"net"
+	"net/netip"
 	"sort"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/pflag"
 	"github.com/vishvananda/netlink"
+	"go4.org/netipx"
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/cilium/cilium/pkg/datapath/linux/probes"
@@ -34,10 +35,10 @@ var (
 	log = logging.DefaultLogger.WithField(logfields.LogSubsys, "egressgateway")
 	// GatewayNotFoundIPv4 is a special IP value used as gatewayIP in the BPF policy
 	// map to indicate no gateway was found for the given policy
-	GatewayNotFoundIPv4 = net.ParseIP("0.0.0.0")
+	GatewayNotFoundIPv4 = netip.MustParseAddr("0.0.0.0")
 	// ExcludedCIDRIPv4 is a special IP value used as gatewayIP in the BPF policy map
 	// to indicate the entry is for an excluded CIDR and should skip egress gateway
-	ExcludedCIDRIPv4 = net.ParseIP("0.0.0.1")
+	ExcludedCIDRIPv4 = netip.MustParseAddr("0.0.0.1")
 )
 
 // Cell provides a [Manager] for consumption with hive.
@@ -344,15 +345,15 @@ func (manager *Manager) updatePoliciesBySourceIP() {
 // - the given endpoint
 // - the destination CIDR
 // - a boolean value indicating if the CIDR belongs to the excluded ones
-// - the gatewayConfig of the  policy
+// - the gatewayConfig of the policy
 //
 // This method returns true whenever the f callback matches one of the endpoint
 // and CIDR tuples (i.e. whenever one callback invocation returns true)
-func (manager *Manager) policyMatches(sourceIP net.IP, f func(net.IP, *net.IPNet, bool, *gatewayConfig) bool) bool {
+func (manager *Manager) policyMatches(sourceIP netip.Addr, f func(netip.Addr, netip.Prefix, bool, *gatewayConfig) bool) bool {
 	for _, policy := range manager.policyConfigsBySourceIP[sourceIP.String()] {
 		for _, ep := range policy.matchedEndpoints {
 			for _, endpointIP := range ep.ips {
-				if !endpointIP.Equal(sourceIP) {
+				if endpointIP.Compare(sourceIP) != 0 {
 					continue
 				}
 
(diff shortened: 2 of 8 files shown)


danehans:
@jschwinger233 @nathanjsweet PTAL when you have a moment.

pkg/ip/cidr.go
```diff
@@ -71,6 +71,28 @@ func PrefixToIPNet(prefix netip.Prefix) *net.IPNet {
 	}
 }
 
+// IPNetsToPrefixes is a convenience helper for migrating from the older 'net.IPNet'
+// standard library type to the newer 'netip.Prefix' type. Use this to plug the old
+// type in existing code into the newer type.
+func IPNetsToPrefixes(ipNets []*net.IPNet) []*netip.Prefix {
```
jschwinger233: I feel reluctant to add these helpers back as they were removed by 929089ecb540a03a4ace8f15984ce5639caf7da8, but I see the reason to do so, we need them until pkg/ip gets rid of `net.IP` too,

jschwinger233 (approved):

danehans:
@nathanjsweet when you have a moment, PTAL.

danehans:
@christarazi when you have a moment, do you mind helping me with a review or assigning another maintainer to take a look?

pkg/ip/cidr.go
gandro: Can we use `netipx.FromStdIPNet` here? It's already a dependency we use in other packages.

pkg/egressgateway/manager.go
```diff
@@ -528,33 +529,42 @@ nextIpRule:
 }
 
 func (manager *Manager) addMissingEgressRules() {
+	logger := log.WithFields(logrus.Fields{})
+
```
christarazi: No need for this if `logger` is not different from `log.`

pkg/egressgateway/manager.go
```diff
@@ -528,33 +529,42 @@ nextIpRule:
...
 
-	addEgressRule := func(endpointIP net.IP, dstCIDR *net.IPNet, excludedCIDR bool, gwc *gatewayConfig) {
-		policyKey := egressmap.NewEgressPolicyKey4(endpointIP, dstCIDR.IP, dstCIDR.Mask)
+	addEgressRule := func(endpointIP netip.Addr, dstPrefix *netip.Prefix, excludedCIDR bool, gwc *gatewayConfig) {
+		dstNet := ip.PrefixToIPNet(*dstPrefix)
+		if dstNet == nil {
+			logger.Infof("invalid cidr %q", dstPrefix)
+			return
+		}
+		policyKey := egressmap.NewEgressPolicyKey4(endpointIP.AsSlice(), dstPrefix.Addr().AsSlice(), dstNet.Mask)
```
christarazi: I think this sort of pattern is causing unnecessary code churn. The root culprit here is that the `egressmap` package exposes `net.*` types in its API so we have to do conversions like the highlighted code above in order to interface with it. This makes the call site unnecessarily complicated because of these conversions. If we update `egressmap` pkg's API then it removes complicated call sites. Could you update that pkg as well? It should be fairly small.

pkg/egressgateway/manager_privileged_test.go
```diff
@@ -541,10 +543,18 @@ func assertIPRules(c *C, rules []ipRule) {
 nextRule:
 	for _, rule := range parsedRules {
 		for _, installedRule := range installedRules {
-			if rule.sourceIP.Equal(installedRule.Src.IP) && rule.destCIDR.String() == installedRule.Dst.String() &&
+			ruleSrcIP, ok := ip.AddrFromIP(installedRule.Src.IP)
+			if !ok {
+				continue nextRule
+			}
```
christarazi: I guess this is happening because of the `route` pkg...unfortunate, but probably best left for another PR. Could you do the conversion of `installedRules` before the for-loop so as to remove the extraneous error handling in the loop?

christarazi (changes requested):

pkg/egressgateway/manager.go
danehans: OK. I was following [this pattern](https://github.com/cilium/cilium/blob/main/pkg/egressgateway/manager.go#L468). I'll remove it.

pkg/egressgateway/manager.go
```diff
@@ -34,10 +35,10 @@ var (
 	log = logging.DefaultLogger.WithField(logfields.LogSubsys, "egressgateway")
 	// GatewayNotFoundIPv4 is a special IP value used as gatewayIP in the BPF policy
 	// map to indicate no gateway was found for the given policy
-	GatewayNotFoundIPv4 = net.ParseIP("0.0.0.0")
+	GatewayNotFoundIPv4, _ = netip.ParseAddr("0.0.0.0")
```
christarazi: We can use `MustParseAddr` here which omits the 2nd return value

pkg/egressgateway/endpoint.go
```diff
@@ -40,7 +40,9 @@ func getEndpointMetadata(endpoint *k8sTypes.CiliumEndpoint, identityLabels label
 
 	for _, pair := range endpoint.Networking.Addressing {
 		if pair.IPV4 != "" {
-			ipv4s = append(ipv4s, net.ParseIP(pair.IPV4).To4())
+			if ip, err := netip.ParseAddr(pair.IPV4); err == nil && ip.Is4() {
```
christarazi: We can use `MustParseAddr` here as well as the previous code assumed that the IP was always a valid IPv4 address

pkg/egressgateway/manager_privileged_test.go
```diff
@@ -538,13 +540,32 @@ func assertIPRules(c *C, rules []ipRule) {
...
+		var src, dst *netip.Addr
+		if srcAddr, ok := netipx.FromStdIP(installedRule.Src.IP); ok {
+			src = &srcAddr
+		}
+		if dstAddr, ok := netipx.FromStdIP(installedRule.Dst.IP); ok {
+			dst = &dstAddr
+		}
+		if src != nil && dst != nil {
+			localRules = append(localRules, localRule{table: installedRule.Table, src: *src, dst: *dst})
+		}
```
christarazi: No need for `src` and `dst` to be pointers if they just get dereferenced in the end during the `append`, right?

danehans:
@christarazi thanks for the review. Commit 2345363 resolves your feedback, PTAL.

danehans:
Marked as a draft while I resolve the `Runtime Test (privileged)` job failure.

pkg/egressgateway/policy.go
```diff
@@ -177,20 +177,34 @@ func (gwc *gatewayConfig) deriveFromPolicyGatewayConfig(gc *policyGatewayConfig)
...
 	for _, dstCIDR := range config.dstCIDRs {
-		dstCIDRMinusExcludedCIDRs := []*net.IPNet{dstCIDR}
+		dstCIDRMinusExcludedCIDRs := []netip.Prefix{dstCIDR}
 		for _, excludedCIDR := range config.excludedCIDRs {
-			newDstCIDRMinuxExcludedCIDRs := []*net.IPNet{}
+			newDstCIDRMinuxExcludedCIDRs := []netip.Prefix{}
 			for _, cidr := range dstCIDRMinusExcludedCIDRs {
-				r, _, l := ip.PartitionCIDR(*cidr, *excludedCIDR)
-				newDstCIDRMinuxExcludedCIDRs = append(newDstCIDRMinuxExcludedCIDRs, append(r, l...)...)
+				rightNets, _, leftNets := ip.PartitionCIDR(*netipx.PrefixIPNet(cidr), *netipx.PrefixIPNet(excludedCIDR))
```
christarazi: Seems like a lot of unnecessary conversions here if we had a `PartitionCIDR` function that natively worked with `netip` types. Could you take a crack at implementing one?

pkg/egressgateway/manager_privileged_test.go
danehans: Since this block now uses the netipx lib, I dropped the `localRule` custom type. Let me know if this looks good to you.

pkg/egressgateway/policy.go
danehans: @christarazi I took a quick look at trying to replicate `PartitionCIDR()` for the netip pkg. It seems to be a fair bit of work. Since this PR is already quite large, are you OK if I create an issue to track this need and work on it separately?

christarazi: To be honest in that case, I don't really see the benefit of merging this PR in its current form because it will just introduce unnecessary churn into the tree by having a bunch of calls to convert `net` types back and forth, which will eat away at CPU cycles. What seems like we are uncovering here is that `pkg/egressgateway` has a lot more interdependence on `net.*` types so, maybe it makes more sense to revisit it when more of the surrounding code has been converted over to `netip`. Does that sound reasonable?

danehans: Yes, I think that sounds reasonable. Should I close this PR for the time being?

christarazi: You can convert it into draft mode also. Up to you.

danehans:
Converting to draft due to https://github.com/cilium/cilium/pull/25564#discussion_r1234547366.

PR #24677, opened 2023-03-31
datapath: Devices table and controller

joamaki:
This PR adds the following new tables to pkg/datapath/tables:
```
// Device is a local network device along with addresses associated with it.
type Device struct {
	Index        int              // positive integer that starts at one, zero is never used
	MTU          int              // maximum transmission unit
	Name         string           // e.g., "en0", "lo0", "eth0.100"
	HardwareAddr net.HardwareAddr // IEEE MAC-48, EUI-48 and EUI-64 form
	Flags        net.Flags        // e.g. net.FlagUp, net.eFlagLoopback, net.FlagMulticast

	Selected    bool            // If true this device can be used by Cilium
	Addrs       []DeviceAddress // Addresses assigned to the device
	RawFlags    uint32          // Raw interface flags
	Type        string          // Device type, e.g. "veth" etc.
	MasterIndex int             // Index of the master device (e.g. bridge or bonding device)
}

var deviceTableSchema = &memdb.TableSchema{
	Name: "devices",
	Indexes: map[string]*memdb.IndexSchema{
		string(statedb.IDIndex): {
			Name:         string(statedb.IDIndex),
			AllowMissing: false,
			Unique:       true,
			Indexer:      &memdb.IntFieldIndex{Field: "Index"},
		},
		string(DeviceNameIndex): {
			Name: string(DeviceNameIndex),
			// Name can be temporarily missing if we create the device
			// from an address update.
			AllowMissing: true,
			Unique:       true,
			Indexer:      &memdb.StringFieldIndex{Field: "Name"},
		},
	},
}

type Route struct {
	Table     int
	LinkIndex int

	Scope uint8
	Dst   netip.Prefix
	Src   netip.Addr
	Gw    netip.Addr
}

var (
	routeTableSchema = &memdb.TableSchema{
		Name: "routes",
		Indexes: map[string]*memdb.IndexSchema{
			"id": {
				Name:         "id",
				AllowMissing: false,
				Unique:       true,
				Indexer: &memdb.CompoundIndex{
					Indexes: []memdb.Indexer{
						&memdb.IntFieldIndex{Field: "Table"},
						&memdb.IntFieldIndex{Field: "LinkIndex"},
						&statedb.NetIPPrefixFieldIndex{Field: "Dst"},
					},
				},
			},
			string(linkIndexIndex): {
				Name:         string(linkIndexIndex),
				AllowMissing: false,
				Unique:       false,
[...]

PR #25565, opened 2023-05-19
ipsec: Attach to new ENI interfaces as they are added

pchaigno:
This pull request ensures that we rerun the detection and attachment of bpf_network to new ENI interfaces whenever they are added. See commits for details.

```release-note
Fix bug where Cilium wouldn't attach its IPsec BPF program to new ENI interfaces, resulting in connectivity loss between pods on remote nodes.  
```

### DIFF
--- a/pkg/datapath/loader/base.go
+++ b/pkg/datapath/loader/base.go
@@ -27,7 +27,11 @@ import (
 	datapath "github.com/cilium/cilium/pkg/datapath/types"
 	"github.com/cilium/cilium/pkg/defaults"
 	"github.com/cilium/cilium/pkg/identity"
+	"github.com/cilium/cilium/pkg/ipam"
 	ipamOption "github.com/cilium/cilium/pkg/ipam/option"
+	"github.com/cilium/cilium/pkg/k8s"
+	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
+	"github.com/cilium/cilium/pkg/lock"
 	"github.com/cilium/cilium/pkg/logging/logfields"
 	"github.com/cilium/cilium/pkg/mac"
 	"github.com/cilium/cilium/pkg/node"
@@ -507,5 +511,65 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner, 
 		}
 	}
 
+	return nil
+}
+
+// OnUpdateCiliumNode is only called if running with IPAM ENI and IPsec. In
+// that case, we need to rerun reinitializeIPSec whenever new ENIs are added to
+// the local node, to attach our BPF program for IPsec decryption to the new
+// interfaces.
+func (l *Loader) OnUpdateCiliumNode(oldNode, newNode *ciliumv2.CiliumNode, _ *lock.StoppableWaitGroup) error {
+	if !k8s.IsLocalCiliumNode(newNode) {
+		return nil
+	}
+
+	log.Debug("Checking for new ENIs on local node")
+
+	newENIs := make([]string, 0)
+	for newENI := range newNode.Status.ENI.ENIs {
+		existed := false
+		for oldENI := range oldNode.Status.ENI.ENIs {
+			if newENI == oldENI {
+				existed = true
+				break
+			}
+		}
+		if !existed {
+			newENIs = append(newENIs, newENI)
+		}
+	}
+
+	if len(newENIs) == 0 {
+		return nil
+	}
+
+	// ... and we can then trigger loading of BPF programs to those interfaces.
+	log.WithField(logfields.NewENIs, newENIs).Info("New ENIs detected")
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// If we have any new ENIs, we first need to wait for the corresponding
+	// interfaces to show up...
+	missingENIByMac, err := ipam.WaitForENIInterfaces(oldNode, newNode)
+	if err != nil {
+		log.WithError(err).WithFields(logrus.Fields{
+			logfields.MissingENIs: missingENIByMac,
+		}).Error("Failed while waiting for ENIs to be attached")
+		return err
+	}
+
+	return l.reinitializeIPSec(ctx)
+}
+
+// OnAddCiliumNode doesn't need to perform anything contrary to
+// OnUpdateCiliumNode because we are not interested in new remote nodes.
+func (l *Loader) OnAddCiliumNode(node *ciliumv2.CiliumNode, swg *lock.StoppableWaitGroup) error {
+	return nil
+}
+
+// OnDeleteCiliumNode doesn't need to perform anything contrary to
+// OnUpdateCiliumNode because we are not interested in remote nodes.
+func (l *Loader) OnDeleteCiliumNode(node *ciliumv2.CiliumNode, swg *lock.StoppableWaitGroup) error {
 	return nil
 }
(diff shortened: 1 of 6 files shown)


pkg/datapath/loader/base.go
```diff
@@ -499,3 +500,50 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner,
...
+		if !existed {
+			newENIs = append(newENIs, newENI)
+		}
+	}
+
+	// If we have any new ENIs, we need to call trigger loading of BPF programs
+	// to them.
+	if len(newENIs) > 0 {
+		log.WithField(logfields.NewENIs, newENIs).Info("New ENIs detected")
+		return l.reinitializeIPSec(context.TODO())
```
pchaigno: For reviewers: How can I fix this to have a proper context?

pkg/datapath/loader/base.go
```diff
@@ -499,3 +500,50 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner,
 
 	return nil
 }
+
+// OnUpdateCiliumNode is only called if running with IPAM ENI and IPsec. In
+// that case, we need to rerun reinitializeIPSec whenever new ENIs are added to
+// the local node, to attach our BPF program for IPsec decryption to the new
+// interfaces.
+func (l *Loader) OnUpdateCiliumNode(oldNode, newNode *ciliumv2.CiliumNode, _ *lock.StoppableWaitGroup) error {
```
joamaki, line 521: I'd like to solve this in a cleaner way (and without k8s dependencies in pkg/datapath) by having loader observe the devices table (coming in https://github.com/cilium/cilium/pull/24677). I'm fine with this workaround until the proper ground work is done. Is there any pitfalls to just watching for the local network devices and addresses versus the CiliumNode object?

gandro (commented):
A high-level comment: We do have somewhat very similar code in https://github.com/cilium/cilium/blob/main/pkg/ipam/crd_eni.go 

One main difference seems to be that in `crd_eni` we wait for the netlink devices to be attached. This was needed, because sometimes the operator (which populates `.Status.ENI.ENIs` based on the state of the EC2 API) would know about ENIs before the Linux guest on the actual node saw it. I think similar logic is needed here, as having an ENI in `Status.ENI.ENIs` just means that AWS has attached the interface to the virtual machine, but not that the local Linux kernel has actually already discovered it and exposed it as a netlink interface.

As a side-note: This PR intended to be backported, so I think it's fine to have this fix architected as is for now. But long-term, I think we should probably introduce an "ENI device" manager (Jussi has suggested that in the past too) that merges the two pieces of code, since both seem to concern themselves with setting up the netlink interface.

pkg/datapath/loader/base.go
```diff
@@ -499,3 +500,50 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner,
...
+	for newENI := range newNode.Status.ENI.ENIs {
+		existed := false
+		for oldENI := range oldNode.Status.ENI.ENIs {
+			if newENI == oldENI {
+				existed = true
+				break
+			}
+		}
+		if !existed {
+			newENIs = append(newENIs, newENI)
```
gandro, line 538: What interfaces do we actually want to attach the IPSec program to?

In IPAM, we actually do allow the user to exclude certain interfaces for pod IP allocation. This happens if the user wants to use those interfaces for other purposes than pod IPs, e.g. for egressgw or if they are running non-pod workloads on the node (see `spec.eni.exclude-interface-tags` and `spec.eni.first-interface-index` in https://docs.cilium.io/en/v1.13/network/concepts/ipam/eni/#ipam-eni)

Do we need to have the same logic here, or can the IPSec program be attached to any interface even if it doesn't have anything to do with Cilium?

Checking if an interface is excluded is luckily very easy:
https://github.com/cilium/cilium/blob/3e805306fc4125fca3398397acdf11a13f1a9600/pkg/ipam/crd_eni.go#L50-L52

pchaigno: We can attach to any interface currently. That might change once we support IPsec+KPR, but right now, it's fine. We already attach to e.g. the `nodelocaldns` interface :shrug:  Unless those interface receive ESP traffic, the BPF program won't actually do anything.

pkg/datapath/loader/base.go
pchaigno: > Is there any pitfalls to just watching for the local network devices and addresses versus the CiliumNode object?

I think that would work. Given Sebastian's comment below, it would probably even be easier once we have the new controller in place.

pkg/datapath/loader/base.go
gandro: In that case I think it's also fine to just attach to those interfaces too. It's hard to judge how users treat the excluded interface, but we also never made any explicit guarantees, except that we're not using them for pod IP allocation.

pkg/logging/logfields/logfields.go
```diff
@@ -617,6 +617,9 @@ const (
 	// ExpectedENIs are the ENIs which are expected to be available
 	ExpectedENIs = "expectedENIs"
 
+	// NewENIs are ENIs that have been added to the CiliumNode.
+	NewENIs = "newENIs"
```
schlosna, line 618: `NewENIs` does not appear to be used

pkg/ipam/crd_eni.go
```diff
@@ -104,6 +106,13 @@ func setupENIDevices(eniConfigByMac configMap) {
 				Error("Failed to configure ENI device")
 		}
 	}
+
+	// Attach relevant BPF program to them.
+	if len(eniLinkByMac) > 0 {
+		if err := loader.ReinitializeIPsec(context.TODO()); err != nil {
```
gandro: There is one thing I just remembered which might be relevant: Because `setupENIDevices` runs in a go routine in the background, there might be multiple concurrent invocations of it.

So `ReinitializeIPsec` must be able to deal with the fact that it might be called concurrently. Which looking at the implementation, I don't think it is. Especially the synchronized write to `option.Config.EncryptInterface` looks like a potential data race to me.

(Another reason why we should refactor this whole logic here :disappointed: )


Also, feel free to introduce a context at the top of `setupENIDevices` like this:
```
func setupENIDevices(eniConfigByMac configMap) {
	ctx, cancel = context.WithCancel(context.Background())
	defer cancel()
```

Not sure why `setupENIDevices` did not use a context for e.g. the timeout, probably because it was implemented rather quickly to address a bug fix.

pkg/datapath/loader/base.go
```diff
@@ -511,3 +513,63 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner,
...
+
+	// ... and we can then trigger loading of BPF programs to those interfaces.
+	log.WithField(logfields.NewENIs, newENIs).Info("New ENIs detected")
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// If we have any new ENIs, we first need to wait for the corresponding
+	// interfaces to show up...
+	missingENIByMac, err := ipam.WaitForENIInterfaces(oldNode, newNode)
```
gandro, line 554: This will block any CiliumNode events from being processed in the agent. We should move this into a go routine, a [job group](https://docs.cilium.io/en/latest/contributing/development/hive/#job-groups) or good old fashioned one-shot [controller](https://pkg.go.dev/github.com/cilium/cilium@v1.13.2/pkg/controller)

pkg/datapath/loader/base.go
```diff
@@ -511,3 +513,63 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner,
...
+	if err != nil {
+		log.WithError(err).WithFields(logrus.Fields{
+			logfields.MissingENIs: missingENIByMac,
+		}).Error("Failed while waiting for ENIs to be attached")
+		return err
+	}
+
+	return l.reinitializeIPSec(ctx)
```
gandro, line 562: Doesn't this need to be called mutually exclusive with `l.Reinitialize`? Otherwise both will write to shared global state without synchronization.

pchaigno:
Closing in favor of the approach Jussi is taking.

PR #25571, opened 2023-05-21
test: Switch target FQDN

pchaigno:
First commit refactors a bit to have the target FQDN in a single place. Second commit changes it to stop using jenkins.cilium.io since it's down.


nbusseneau (approved):

pchaigno:
The affected tests passed. Marking ready to merge.

joestringer (approved):

PR #25051, opened 2023-04-21
nat: fix missing L3 pseudo-header csum diff applied into L4

sahid:
In context of an ICMP need to frag received, the NAT happening in the embbeded packet is missing applying the L3 pseudo-header csum diff into the TCP/UDP header.

PR #25576, opened 2023-05-22
nat: fix usage in nat.h of csum.h module

sahid:
csum.h has been initiated to be used for csum update based on protocol that has pseudo header.  In some places in the code of nat.h it has been "hacked" to be used even for other protocol like ICMP that does not have pseudo headers.

### DIFF
--- a/bpf/lib/nat.h
+++ b/bpf/lib/nat.h
@@ -470,7 +470,6 @@ snat_v4_icmp_rewrite_egress_embedded(struct __ctx_buff *ctx,
 	int ret;
 	int flags = BPF_F_PSEUDO_HDR;
 	struct csum_offset csum = {};
-	__be32 sum_l4 = 0;
 	__be32 sum;
 
 	if (state->to_saddr == tuple->saddr &&
@@ -494,18 +493,16 @@ snat_v4_icmp_rewrite_egress_embedded(struct __ctx_buff *ctx,
 			return DROP_CSUM_L4;
 #endif  /* ENABLE_SCTP */
 		case IPPROTO_ICMP: {
-			__be32 from, to;
-
 			if (ctx_store_bytes(ctx, inner_l4_off +
 						offsetof(struct icmphdr, un.echo.id),
 						&state->to_sport,
 						sizeof(state->to_sport), 0) < 0)
 				return DROP_WRITE_ERROR;
-			from = tuple->sport;
-			to = state->to_sport;
-			flags = 0; /* ICMPv4 has no pseudo-header */
-			sum_l4 = csum_diff(&from, 4, &to, 4, 0);
-			csum.offset = offsetof(struct icmphdr, checksum);
+			if (l4_csum_replace(ctx, inner_l4_off + offsetof(struct icmphdr, checksum),
+					    tuple->sport,
+					    state->to_sport,
+					    sizeof(tuple->sport)) < 0)
+				return DROP_CSUM_L4;
 			break;
 		}}
 	}
@@ -514,8 +511,6 @@ snat_v4_icmp_rewrite_egress_embedded(struct __ctx_buff *ctx,
 		return DROP_WRITE_ERROR;
 	if (ipv4_csum_update_by_diff(ctx, l4_off + sizeof(struct icmphdr), sum) < 0)
 		return DROP_CSUM_L3;
-	if (tuple->nexthdr == IPPROTO_ICMP)
-		sum = sum_l4;
 	if (csum.offset &&
 	    csum_l4_replace(ctx, inner_l4_off, &csum, 0, sum, flags) < 0)
 		return DROP_CSUM_L4;
@@ -529,7 +524,7 @@ static __always_inline int snat_v4_rewrite_egress(struct __ctx_buff *ctx,
 {
 	int ret, flags = BPF_F_PSEUDO_HDR;
 	struct csum_offset csum = {};
-	__be32 sum_l4 = 0, sum;
+	__be32 sum;
 
 	if (state->to_saddr == tuple->saddr &&
 	    state->to_sport == tuple->sport)
@@ -554,18 +549,16 @@ static __always_inline int snat_v4_rewrite_egress(struct __ctx_buff *ctx,
 				return DROP_CSUM_L4;
 #endif  /* ENABLE_SCTP */
 			case IPPROTO_ICMP: {
-				__be32 from, to;
-
 				if (ctx_store_bytes(ctx, off +
 						    offsetof(struct icmphdr, un.echo.id),
 						    &state->to_sport,
 						    sizeof(state->to_sport), 0) < 0)
 					return DROP_WRITE_ERROR;
-				from = tuple->sport;
-				to = state->to_sport;
-				flags = 0; /* ICMPv4 has no pseudo-header */
-				sum_l4 = csum_diff(&from, 4, &to, 4, 0);
-				csum.offset = offsetof(struct icmphdr, checksum);
+				if (l4_csum_replace(ctx, off + offsetof(struct icmphdr, checksum),
+						    tuple->sport,
+						    state->to_sport,
+						    sizeof(tuple->sport)) < 0)
+					return DROP_CSUM_L4;
 				break;
 			}}
 		}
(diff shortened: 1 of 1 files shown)


bpf/lib/nat.h
```diff
@@ -602,18 +602,15 @@ static __always_inline int snat_v4_rewrite_egress(struct __ctx_buff *ctx,
...
 						    sizeof(state->to_sport), 0) < 0)
 					return DROP_WRITE_ERROR;
-				from = tuple->sport;
-				to = state->to_sport;
-				flags = 0; /* ICMPv4 has no pseudo-header */
-				sum_l4 = csum_diff(&from, 4, &to, 4, 0);
-				csum.offset = offsetof(struct icmphdr, checksum);
+				if (l4_csum_replace(ctx, off + offsetof(struct icmphdr, checksum),
+						    tuple->sport,
+						    state->to_sport, 0) < 0)
```
jschwinger233: Do we need to specify the number of bytes (2) on the lowest four bits of _flags_(the last parameter in bpf_l4_csum_replace)?

sahid: Yes Interesting point I'm wondering whether we do right in other places as-well. Let me see...

sahid: Thank you for this review I fixed it.

sahid:
Once this is accepted I will update #25051 to make everything coherent.

bpf/lib/nat.h
```diff
@@ -1642,17 +1639,17 @@ static __always_inline int snat_v6_rewrite_egress(struct __ctx_buff *ctx,
...
 			if (ctx_store_bytes(ctx, off +
 					    offsetof(struct icmp6hdr,
 						     icmp6_dataun.u_echo.identifier),
 					    &state->to_sport,
 					    sizeof(state->to_sport), 0) < 0)
 				return DROP_WRITE_ERROR;
-			from = tuple->sport;
-			to = state->to_sport;
-			sum = csum_diff(&from, 4, &to, 4, sum);
+			if (l4_csum_replace(ctx, off + offsetof(struct icmphdr, checksum),
```
jschwinger233: `ctx` here holds an ICMPv6 packet, maybe we should use `offsetof(struct icmp6hdr, icmp6_cksum)`?

bpf/lib/nat.h
```diff
@@ -1707,9 +1703,11 @@ static __always_inline int snat_v6_rewrite_ingress(struct __ctx_buff *ctx,
 						    &state->to_dport,
 						    sizeof(state->to_dport), 0) < 0)
 					return DROP_WRITE_ERROR;
-				from = tuple->dport;
-				to = state->to_dport;
-				sum = csum_diff(&from, 4, &to, 4, sum);
+				if (l4_csum_replace(ctx, off + offsetof(struct icmphdr, checksum),
```
jschwinger233: Same as above.

bpf/lib/nat.h
```diff
@@ -1642,17 +1639,17 @@ static __always_inline int snat_v6_rewrite_egress(struct __ctx_buff *ctx,
...
 			if (ctx_store_bytes(ctx, off +
 					    offsetof(struct icmp6hdr,
 						     icmp6_dataun.u_echo.identifier),
 					    &state->to_sport,
 					    sizeof(state->to_sport), 0) < 0)
 				return DROP_WRITE_ERROR;
-			from = tuple->sport;
-			to = state->to_sport;
-			sum = csum_diff(&from, 4, &to, 4, sum);
+			if (l4_csum_replace(ctx, off + offsetof(struct icmp6hdr, icmp6_cksum),
```
sahid: Actually I should not do it for icmp6

bpf/lib/nat.h
sahid: I should not do this at all for icmp6

jschwinger233:
Do you think your change leaves these code dead, as we no longer update the `sum_l4` and `csum.offset` in the ICMP branch?
https://github.com/cilium/cilium/blob/5c94a0dc975bdc067cd8b8886ca60ed694471e1e/bpf/lib/nat.h#L563-L564

jschwinger233 (approved):
Let's move forward and don't block your following PRs, we can clean some dead code later if there is any.

jschwinger233:
No worries, I should have pointed out the potential issues in the first place.

bpf/lib/nat.h
```diff
@@ -542,18 +542,16 @@ snat_v4_icmp_rewrite_egress_embedded(struct __ctx_buff *ctx,
...
-			from = tuple->sport;
-			to = state->to_sport;
-			flags = 0; /* ICMPv4 has no pseudo-header */
-			sum_l4 = csum_diff(&from, 4, &to, 4, 0);
-			csum.offset = offsetof(struct icmphdr, checksum);
+			if (l4_csum_replace(ctx, inner_l4_off + offsetof(struct icmphdr, checksum),
+					    tuple->sport,
+					    state->to_sport,
+					    sizeof(tuple->sport)) < 0)
+				return DROP_WRITE_ERROR;
```
jschwinger233: I'm thinking about returning `DROP_CSUM_L4` rather than `DROP_WRITE_ERROR`, as we did before at https://github.com/cilium/cilium/blob/b93af2b484cc16f05c281a2be227066783322b15/bpf/lib/nat.h#L569

jschwinger233: Maybe we can change this returned error in your following PR.

jschwinger233 (approved):

jschwinger233:
CI failed due to BPF complexity issue lol

bpf/lib/nat.h
sahid: Yes right It's probably what I should have used.

sahid:
> CI failed due to BPF complexity issue lol

Are you sure, I see all the k8s-*kernel tests green, no?

In all cases if I have to respine for the complexity issue I will use that to fix all the point you have mentioned previously.

jschwinger233:
It's L4LB XDP test: https://github.com/cilium/cilium/actions/runs/5172796260/jobs/9317480797#step:11:2196

USERS.md
```diff
@@ -509,6 +509,12 @@ Users (Alphabetically)
       U: CNI (instead of amazon-vpc-cni-k8s), DefaultDeny(Zero Trust), Hubble, CiliumNetworkPolicy, CiliumClusterwideNetworkPolicy, EKS
       L: https://schubergphilis.com/en
       Q: @stimmerman @shoekstra @mbaumann
+      
+    * N: SIGHUP
```
ti-mo: Why this change? Seems to be copied from the block below. :thinking:

ti-mo (changes requested):

USERS.md
sahid: Hum I made a mistake somewhere during my rebasing... I will fix that

bpf/lib/nat.h
joestringer: @jschwinger233 @sahid what's the tradeoff, any particular reason to defer that to a subsequent PR? I'm not sure I understand why we don't just address the feedback in the PR that's adding the code. :thinking:

jschwinger233: @joestringer That was because Sahid said he had another PR based on this, so I thought we can defer the minor changes to that PR, so as to make things happen fast. But that was before this PR hit the CI failures, since this PR has been here for weeks, I think we should address the issues at this moment. Thanks for pointing out.

sahid: @joestringer It's my mistake, I should have fixed both of the suggestions. I wanted to rebase the code and see if by any chance the complexity issue was gone... Sorry @jschwinger233 to have put you in that situation.

All is fixed now.

ti-mo (approved):

sahid:
The previous PR has been closed du to a lack of activity I created a new one #26453

borkmann:
Heya :wave: Why is this change better than before? Imo, it makes things worse. Now we need an additional helper call whereas before it was not needed and the csum_diff() was actually all inlined.

sahid:
> Heya wave Why is this change better than before? Imo, it makes things worse. Now we need an additional helper call whereas before it was not needed and the csum_diff() was actually all inlined.

:wave:, previously the code was forcing some fields in https://github.com/cilium/cilium/blob/main/bpf/lib/csum.h#L60 and later the block to retrieve `sum` from `sum_l4` a bit obscure (I considered that a bit hacky). I thought I was walking in the good direction changing this part :disappointed_relieved:

Issue #21142, opened 2022-08-30
Rework ipcache to handle metadata from multiple sources via a dedicated worker goroutine


joestringer:
This meta issue tracks a variety of enhancements being made to `pkg/ipcache` in order to make it more resilient and able to consistently merge information gathered about IPs from various sources (kvstore, k8s resources, local node, etc.).

Broadly speaking, these changes define a new "desired state" structure plus a resolution loop in order to realize that desired state. The desired state is synchronously updated by callers to define information such as a list of labels that one resource wants to associate with an IP, or an encryption key to associate with it. Then the upsert/remove call (typically from a k8s watcher) would trigger a separate goroutine to then realize the state internally. The realization of the state will combine all of the information from different sources together, potentially allocate identities to associate with the IP, trigger policy calculation for any new identities, and ensure that the datapath is consistently updated with any new policy state as well as any new bpf ipcache map state. For more details on the high level API, see #19765.

Tasks:

1. Remove static package-level variables from `pkg/ipcache`
   - [x] https://github.com/cilium/cilium/pull/19073
2. Create new asynchronous API for handling insertion of labels / metadata /auxiliary data into the IPCache
   - [x] https://github.com/cilium/cilium/pull/19765
3. Convert the existing callers of IPCache to the new API
   - CIDR policies
     - [x] https://github.com/cilium/cilium/pull/20116
     - [x] ~(optional) Optimization: Use sentinel type to represent CIDR labels to be expanded~
       - https://github.com/cilium/cilium/pull/30897
   - NodeManager
     - [x] Inject encryption state into the new API
     - [x] https://github.com/cilium/cilium/pull/21667
     - [x] ~~https://github.com/cilium/cilium/pull/20117~~
     - [x] https://github.com/cilium/cilium/pull/23208
       - [x] Ensure ingress IPs are excluded from CIDR identity resolution similar to health IPs
       - [x] Ensure the node manager uses a unique resource ID depending on the source of information (k8s or kvstore; k8s node or CiliumNode) 
   - KVStore (pkg/ipcache/kvstore.go)
     Note: KVStore path likely does not need to allocate identities.
     - [ ] pkg/ipcache/kvstore.go
       -  https://github.com/cilium/cilium/pull/31409
     - [ ] pkg/node/store/store.go
     - [ ] Define resource IDs for kvstore paths
   - Local NodeDiscovery
     - [x] Convert Daemon insert host IPs -> UpsertMetadata
[...]

PR #25581, opened 2023-05-22
[WIP] policy: allocate identity for CIDR selectors on first use

squeed:
This moves allocation of local identities for CIDRs from policy ingestion to selector parsing.

See individual commits for more detailed commentary.

### TODO:
- [ ] Get soft consensus that this is a good approach
- [ ] Fix existing unit tests
- [ ] Write more tests

### DIFF
--- a/pkg/ipcache/ipcache.go
+++ b/pkg/ipcache/ipcache.go
@@ -473,16 +473,42 @@ func (ipc *IPCache) RemoveMetadata(prefix netip.Prefix, resource ipcacheTypes.Re
 // labels with these prefixes, thereby making these prefixes selectable in
 // policy via local ("CIDR") identities.
 //
+// It will return a list of identities corresponding the supplied prefixes.
+// The identity for a given prefix may be nil, as allocation may fail. Callers
+// should handle this accordingly.
+//
 // This will trigger asynchronous calculation of any datapath updates necessary
 // to implement the logic associated with the new CIDR labels.
-func (ipc *IPCache) UpsertPrefixes(prefixes []netip.Prefix, src source.Source, resource ipcacheTypes.ResourceID) {
+func (ipc *IPCache) UpsertPrefixes(prefixes []netip.Prefix, src source.Source, resource ipcacheTypes.ResourceID) []*identity.NumericIdentity {
+	out := make([]*identity.NumericIdentity, 0, len(prefixes))
+	updates := make(prefixToIDs, len(prefixes))
 	ipc.metadata.Lock()
 	for _, p := range prefixes {
 		ipc.metadata.upsertLocked(p, src, resource, cidr.GetCIDRLabels(p))
-		ipc.metadata.enqueuePrefixUpdates(p)
+
+		// Allocate and ID for this cidr synchronously.
+		// We will need to release this ID when injectLabels finishes.
+		prefixInfo := ipc.metadata.getLocked(p)
+		// it is safe to use context.Background here, since we are only dealing with CIDRs here, which have
+		// local identity and thus don't need a timeout for allocation.
+		pid, _, err := ipc.resolveIdentity(context.Background(), p, prefixInfo, identity.InvalidIdentity)
+		if err != nil {
+			// An error here is OK; it means that, for whatever reason, we failed to
+			// allocate an ID for the prefix. We will try again when InjectLabels is (eventually)
+			// called.
+			log.WithError(err).WithField(logfields.IPAddr, p).Warn("failed to allocate identity for prefix in UpsertPrefixes (will retry)")
+			ipc.metadata.enqueuePrefixUpdates(p)
+			out = append(out, nil)
+		} else {
+			out = append(out, &pid.ID)
+			// enqueue prefix for upsert and ID for release (to keep the bookkeeping straight)
+			updates[p] = []identity.NumericIdentity{pid.ID}
+		}
 	}
 	ipc.metadata.Unlock()
+	ipc.metadata.enqueuePrefixIdUpdates(updates)
 	ipc.TriggerLabelInjection()
+	return out
 }
 
 // RemovePrefixes removes the association between the prefixes and the CIDR
--- a/pkg/ipcache/metadata.go
+++ b/pkg/ipcache/metadata.go
@@ -76,35 +76,54 @@ type metadata struct {
 	// 'queuedPrefixes'. Each time label injection is triggered, it will
 	// process the metadata changes for these prefixes and potentially
 	// generate updates into the ipcache, policy engine and datapath.
+	// Prefixes may optionally have a list of numeric identities that
+	// have already been allocated synchronously, and should be released
+	// upon successful update (to ensure the reference count is correct).
+	// There could be more than one identity for a given CIDR, since multiple
+	// queued calls could have changed the labels of a cidr and, thus,
+	// allocated multiple identities.
 	queuedChangesMU lock.Mutex
-	queuedPrefixes  map[netip.Prefix]struct{}
+	queuedPrefixes  prefixToIDs
 }
+
+type prefixToIDs map[netip.Prefix][]identity.NumericIdentity
 
 func newMetadata() *metadata {
 	return &metadata{
 		m:              make(map[netip.Prefix]PrefixInfo),
-		queuedPrefixes: make(map[netip.Prefix]struct{}),
+		queuedPrefixes: make(map[netip.Prefix][]identity.NumericIdentity),
 	}
 }
 
-func (m *metadata) dequeuePrefixUpdates() (modifiedPrefixes []netip.Prefix) {
+func (m *metadata) dequeuePrefixUpdates() (modifiedPrefixes prefixToIDs) {
 	m.queuedChangesMU.Lock()
-	modifiedPrefixes = make([]netip.Prefix, 0, len(m.queuedPrefixes))
-	for p := range m.queuedPrefixes {
-		modifiedPrefixes = append(modifiedPrefixes, p)
-	}
-	m.queuedPrefixes = make(map[netip.Prefix]struct{})
+	modifiedPrefixes = m.queuedPrefixes
+	m.queuedPrefixes = make(prefixToIDs)
 	m.queuedChangesMU.Unlock()
 
 	return
 }
 
+// enqueuePrefixUpdates inserts one or more prefixes in the queue to be updated,
+// without any identities to clean up.
 func (m *metadata) enqueuePrefixUpdates(prefixes ...netip.Prefix) {
 	m.queuedChangesMU.Lock()
 	defer m.queuedChangesMU.Unlock()
 
 	for _, prefix := range prefixes {
-		m.queuedPrefixes[prefix] = struct{}{}
+		// leave unchanged if extant, insert nil if not
+		m.queuedPrefixes[prefix] = m.queuedPrefixes[prefix]
+	}
+}
+
+// enqueuePrefixIdUpdates inserts prefixes in the queue to inject (and additional ids to release)
+// We need to track identities to release on a per-prefix basis, since injection for each prefix can
+// theoretically fail.
+func (m *metadata) enqueuePrefixIdUpdates(prefixToIDs map[netip.Prefix][]identity.NumericIdentity) {
+	m.queuedChangesMU.Lock()
+	defer m.queuedChangesMU.Unlock()
+	for prefix, ids := range prefixToIDs {
+		m.queuedPrefixes[prefix] = append(m.queuedPrefixes[prefix], ids...)
 	}
 }
 
@@ -166,7 +185,7 @@ func (m *metadata) getLocked(prefix netip.Prefix) PrefixInfo {
 // Returns the CIDRs that were not yet processed, for example due to an
 // unexpected error while processing the identity updates for those CIDRs
 // The caller should attempt to retry injecting labels for those CIDRs.
-func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.Prefix) (remainingPrefixes []netip.Prefix, err error) {
+func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes prefixToIDs) (remainingPrefixes prefixToIDs, err error) {
 	if ipc.IdentityAllocator == nil {
 		return modifiedPrefixes, ErrLocalIdentityAllocatorUninitialized
 	}
@@ -185,8 +204,14 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
 
 	var (
 		// previouslyAllocatedIdentities maps IP Prefix -> Identity for
-		// old identities where the prefix will now map to a new identity
+		// old identities where the prefix will now map to a new identity.
+		// Only used to silence an inocuous warning message.
 		previouslyAllocatedIdentities = make(map[netip.Prefix]Identity)
+
+		// idsToRelease is the list of identities for which we have taken a reference
+		// which we need to release.
+		idsToRelease = make([]identity.NumericIdentity, 0, len(modifiedPrefixes))
+
 		// idsToAdd stores the identities that must be updated via the
 		// selector cache.
 		idsToAdd    = make(map[identity.NumericIdentity]labels.LabelArray)
@@ -198,11 +223,16 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
 
 	ipc.metadata.RLock()
 
-	for i, prefix := range modifiedPrefixes {
+	for prefix, extraIDs := range modifiedPrefixes {
 		pstr := prefix.String()
 		oldID, entryExists := ipc.LookupByIP(pstr)
 		oldTunnelIP, oldEncryptionKey := ipc.GetHostIPCache(pstr)
 		prefixInfo := ipc.metadata.getLocked(prefix)
+		log.WithError(err).WithFields(logrus.Fields{
+			logfields.IPAddr:   prefix,
+			logfields.Identity: oldID,
+			logfields.Labels:   prefixInfo.ToLabels(),
+		}).Debug("cdc - ipcache upsert")
 		if prefixInfo == nil {
 			if !entryExists {
 				// Already deleted, no new metadata to associate
@@ -231,7 +261,7 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
 				}).Warning(
 					"Failed to allocate new identity while handling change in labels associated with a prefix.",
 				)
-				remainingPrefixes = modifiedPrefixes[i:]
+				remainingPrefixes[prefix] = extraIDs
 				err = fmt.Errorf("failed to allocate new identity during label injection: %w", err)
 				break
 			}
@@ -274,6 +304,8 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
 			// releasing the previous reference.
 			previouslyAllocatedIdentities[prefix] = oldID
 
+			idsToRelease = append(idsToRelease, oldID.ID)
+
 			// If all associated metadata for this prefix has been removed,
 			// and the existing IPCache entry was never touched by any other
 			// subsystem using the old Upsert API, then we can safely remove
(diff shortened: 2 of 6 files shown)


pkg/policy/selectorcache.go
```diff
@@ -901,6 +929,15 @@ func (sc *SelectorCache) AddIdentitySelector(user CachedSelectionUser, selector
 		newIDSel.namespaces = namespaces
 	}
 
+	// Trigger any asynchronous allocation of any CIDR labels we may be selecting
+	// This will cause any endpoints waiting on this to be regenerated twice in
+	// quick succession, but will reduce load on the ipcache.
+	prefixes := newIDSel.selectedCIDRs()
+	if len(prefixes) > 0 {
+		log.WithField("prefixes", prefixes).Debug("inserting CIDR selector prefixes ")
+		sc.ipcache.UpsertPrefixes(prefixes, source.Generated, sc.ipcacheResource(newIDSel.String()))
```
christarazi: This call will eventually attempt to lock the SelectorCache from inside the ipcache, while we are currently holding the SelectorCache lock at this point. 

```
TriggerLabelInjection -> InjectLabels -> UpdatePolicyMaps -> UpdateIdentities
```

Worst case, I believe doing this can cause deadlocks between SelectorCache and ipcache. Otherwise, it may cause double endpoint regeneration. If it doesn't cause double endpoint regeneration, then we'd likely be in the case where Cilium has completed "policy recalculation" for any given endpoint, but all the state has not been synced, aka race condition.

christarazi: > we'd likely be in the case where Cilium has completed "policy recalculation" for any given endpoint, but all the state has not been synced, aka race condition.

To elaborate on the above, when a policy containing CIDR rules are applied and a new endpoint arrives on the node, the endpoint goes through initial regeneration. It will trigger the above code path to `UpsertPrefixes` and continues on. This trigger queues the ipcache operation to be done **later**. When the endpoint regeneration completes, it will **not** have actually implemented the CIDR policy, as that comes when the second regeneration completes (what was queued just before, aka "double regeneration"). 

This is problematic because endpoint regeneration includes saying "endpoint now implements policy revision X", but this double regeneration is breaks that. Anything downstream / depending on policy revision will actually see temporary drops briefly until the second regeneration completes. In other words, the endpoint only implements policy X when the second regeneration completes. 

One example of something downstream of the policy revision is how CI system for policy tests, both in the Cilium CLI and in the Ginkgo suite.


To fix this, I think we need synchronization mechanism for ipcache upsert, which is something that has been discussed in context of the [ipcache refactor](https://github.com/cilium/cilium/issues/21142), specifically for the FQDN piece.

christarazi: Let's say we implement the above sync mechanism, the next concern I have is what assumptions might we be breaking by having the selectorcache call into the ipcache, i.e. deadlocks, etc.

squeed:
@christarazi I've updated this to fix your concerns around and endpoint's PolicyRevision not being quite correct. The fix is to synchronously allocate Identities as part of inserting a prefix in to the ipcache metadata.

As per a synchronous discussion: "@joestringer might not hate this. might."

This required a bit of a rethink, along with some silly bookkeeping to ensure we didn't leak identities, but it seems to work. CIDR identities are allocated as needed, and de-allocated when no longer selected.

There are some small TODOs, mostly around patching up existing unit tests and writing new ones, but I don't want to get started on that until we agree that this is a good approach.

pkg/ipcache/ipcache.go
```diff
@@ -473,16 +473,42 @@ func (ipc *IPCache) RemoveMetadata(prefix netip.Prefix, resource ipcacheTypes.Re
...
 // policy via local ("CIDR") identities.
 //
+// It will return a list of identities corresponding the supplied prefixes.
+// The identity for a given prefix may be nil, as allocation may fail. Callers
+// should handle this accordingly.
+//
 // This will trigger asynchronous calculation of any datapath updates necessary
 // to implement the logic associated with the new CIDR labels.
-func (ipc *IPCache) UpsertPrefixes(prefixes []netip.Prefix, src source.Source, resource ipcacheTypes.ResourceID) {
+func (ipc *IPCache) UpsertPrefixes(prefixes []netip.Prefix, src source.Source, resource ipcacheTypes.ResourceID) []*identity.NumericIdentity {
```
joestringer, line 482: As a whole, the cilium-agent is moving more and more into the direction of self-contained modules that inform other modules about desired state, then those modules take the desired state and realize it in some form, then emit events to subsequent modules, which can then take the output and realize it further down the stack.

I think that the existing design of the `UpsertPrefixes()` logic is in line with that general direction, even if it's not using the fancy new `pkg/stream` and `Cell` infrastructure (yet). What concerns me about this proposal is that it is attempting to revert back partially to a previous model where calls across modules are synchronous. In earlier versions of Cilium code, this path made the code prone to race conditions, resource leaks and locking issues. While I don't assume that this proposal introduces such issues, I think that the framing of those code weaknesses is useful context to inform what the correct path for this proposal should be.

Here's an example scenario with this approach:
1. Subsystem A attempts to assign labels and allocate an identity for IP X, gets 123
2. Subsystem B immediately afterwards attempts to assign (different) labels and allocate an identity for IP X, gets 456
3. Subsystem A starts to plumb through id 123 while subsystem B plumbs through id 456
4. Earlier steps queued up the ipcache internals which will eventually trigger and asynchronously evaluate the identity for the labels and determine that it should be 456
5. Now, in order to correct subsystem A, it must be informed about id 456 in order to plumb whatever state is needed. This is the responsibility of the ipcache module.
6. (sidenote, of course that implies that subsystem A must also be informed that id 123 is stale and must be cleaned up, which means plumbing both 123 and 456 through the ipcache and handling this condition then informing other subsystems about the deallocation of 123. In particular subsystem A needs to know. But actually at that point subsystem B doesn't care about 123, so we must also keep track of ordering of events or otherwise each subsystem must track their own used identities)
7. Finally ipcache can safely transition the underlying BPF datapath ipcache map to start handling traffic with id 456, since all subsystems are in sync.

In the above steps we can see that steps (1) and (5) are both attempting to configure subsystem A. The impact on the codebase therefore is that not only does the ipcache internal code need to handle both sync and async paths (and expose that through its interfaces per this PR), but subsystem A _also_ needs to be taught about both the sync and async paths in order to be eventually consistent. Step (6) touches on some of the nuances here but I suspect it's not yet comprehensive.

pkg/ipcache/ipcache.go
```diff
@@ -473,16 +473,42 @@ func (ipc *IPCache) RemoveMetadata(prefix netip.Prefix, resource ipcacheTypes.Re
...
+		// We will need to release this ID when injectLabels finishes.
+		prefixInfo := ipc.metadata.getLocked(p)
+		// it is safe to use context.Background here, since we are only dealing with CIDRs here, which have
+		// local identity and thus don't need a timeout for allocation.
+		pid, _, err := ipc.resolveIdentity(context.Background(), p, prefixInfo, identity.InvalidIdentity)
+		if err != nil {
+			// An error here is OK; it means that, for whatever reason, we failed to
+			// allocate an ID for the prefix. We will try again when InjectLabels is (eventually)
+			// called.
+			log.WithError(err).WithField(logfields.IPAddr, p).Warn("failed to allocate identity for prefix in UpsertPrefixes (will retry)")
```
joestringer, line 499: I think there's actually not too many error conditions here, but one of them is a bit tricky: If we're up against the total number of identities in the allocation space, then we may just run out of numbers to pick for the identities. The current design is intended to attempt to make progress: The user could always potentially delete a policy, so `InjectLabels()` should theoretically free up enough identity numbers to unblock new identity allocations. That said, it's a bit hard to make strong guarantees in this sort of scenario. We should probably be trying to put some guard rails up far before Cilium reaches that state, to avoid the user reaching such a scenario where Cilium can't implement the policy due to identity overallocation in the local (CIDR) identity space.

As a sidenote I'd love to have more comprehensive testing to confirm that if Cilium gets in that situation, it can also dig itself back out without requiring the user to restart the agent, but I think the code is not quite modular enough to provide such guarantees today. With the latest iteration of `InjectLabels()` it might not be far off... but I digress.

pkg/ipcache/ipcache.go
```diff
@@ -473,16 +473,42 @@ func (ipc *IPCache) RemoveMetadata(prefix netip.Prefix, resource ipcacheTypes.Re
 // labels with these prefixes, thereby making these prefixes selectable in
 // policy via local ("CIDR") identities.
 //
+// It will return a list of identities corresponding the supplied prefixes.
```
joestringer, line 476: I recognize the need for signalling the callers with the allocated prefixes. While many ipcache users do not need this ability, I think there's a couple of users that do need something like this.

I've briefly toyed with the idea of allowing the caller to supply a channel/ signalling mechanism/opaque subsystem-specific identifier down through a new special `aux IPMetadata` type in order to support this use case. The idea is that `InjectLabels()` could then see the auxiliary notifier data and use that to notify the source subsystem for the corresponding `UpsertPrefixes()`/`UpsertMetadata()`. I haven't gone so far as to determine a concrete form of this, but it would also need to be done with care for the ipcache locking patterns as well as synchronization on the corresponding ipcache notifications to (a) other subsystems and (b) for the resulting datapath ipcache map update.

joestringer: Thought experiment, trying to explore the implications of this suggestion for `pkg/endpoint` along the lines of Chris' discussion thread above:

If a new endpoint is deployed on the node, then it will of course go through policy generation, and with this new overall feature, trigger selectorache update -> allocate identities -> selectorcache update -> etc. In the earlier iteration, this path will trigger the async logic for implementing the policy, then return back and assume that the policy is now implemented at revision X. As Chris mentioned, this is a concern because some portion of the policy implementation is now asynchronously deferred. This is bad for two reasons: (a) Falsely reporting the policy revision is implemented, and even (b) Potentially allowing the CNI ADD to return with "endpoint is ready to run the process and handle traffic", but actually the realized policy cannot successfully handle the traffic.

However, what if instead, the endpoint package passes down a waitgroup that gets closed when the ipcache has been fully plumbed for the corresponding prefixes? Or endpoint passes down a token that `pkg/ipcache` can then hand back to `pkg/endpointmanager` -> `pkg/endpoint` to inform the endpoint that the policy is truly realized? Then, rather than reporting that the policy is implemented at revision X, the newly deployed endpoint can potentially report revision 0 or maybe even block the CNI add completion until the ipcache reports that the operation is complete, then it can trigger the policy revision bump and CNI return from agent API?

Little bit handwavy, maybe some iterations on the idea could clean it up a bit more.

pkg/ipcache/metadata.go
```diff
@@ -231,7 +261,7 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
 				}).Warning(
 					"Failed to allocate new identity while handling change in labels associated with a prefix.",
 				)
-				remainingPrefixes = modifiedPrefixes[i:]
+				remainingPrefixes[prefix] = extraIDs
 				err = fmt.Errorf("failed to allocate new identity during label injection: %w", err)
 				break
```
joestringer, line 266: I won't point out every bug yet since there's higher level points to align on, but this change looks like it will incorrectly discard the remainder of the modifiedPrefixes slice.

(sorry, lack of testing here that fails to catch this is partially my fault)

pkg/ipcache/ipcache.go
joestringer, line 1: Related to my first comment, but I couldn't see a good line to raise the discussion point so I'll raise it here:

Thinking long term, I think that there's a reasonable chance that the overall daemon will boil down to a bunch of Cell structures that embed `pkg/stream` objects. Each Cell that provides input to the ipcache or receives the output from ipcache would specify the stream for the appropriate resources in their initialization parameters, and the Hive would plumb it together.

In such a scenario, the inputs to ipcache would be entirely isolated from the outputs, and the main question related to this PR is how we would ensure that only the relevant destination modules/objects are informed about the relevant changes based on the policy relationship with the Endpoints.

Even if there's a variation on this where we end up making something like `pkg/statedb` as the core of much of the state transfer, we'll still need some sort of signalling mechanism on top in order for Cells to coordinate.

I raise this partly because I find it a bit hard to reason about how this proposal (in particular with the UpsertPrefixes change) will fit into that sort of a future. I'd be interested if you or anyone has thoughts on that. While I don't think that this design is realizable in the short or mid-term, it would be nice to at least start nudging this code towards that sort of a future if we can, rather than further complicating the process to get to that desired state.

joestringer: One more point on this: For the selectorcache, if we need to update it depending on newly allocated identities... ideally it should be able to just sign up with one interface to receive updates from the ipcache (or even the identity allocator). As I understand it there's already two mechanisms to receive updates asynchronously (+1 for initialization), and this proposal adds yet another. Some of the recent PR proposals building on `pkg/stream` seem like they're able to streamline similar sorts of event streams down into a single handling path, which seems like it has the potential to greatly simplify these codepaths.

joestringer (commented):
I'm recognizing that there's a lack of design docs and state sharing around a lot of the logic here, which I'll take partial responsibility for... sorry about that :-) Unfortunately this means that it may take a bit more work on each of our parts to close the gap on future vision for this code and what we think are reasonable tradeoffs to achieve the changes that you're wanting to achieve.

Although I feel that I have a fair understanding of where this proposal is coming from, I admit that I found it difficult to follow the full reasoning about the approach. I think that statements like `it is useful for the generated identities to be available synchronously` are doing a lot of heavy lifting. In a vacuum I can agree that it's "useful", but there's a lot more background to unpack here. The background ranges from code structure implications, how we isolate the responsibilities of different modules, how we mitigate concurrent code issues, and how any proposed alternatives compare to this one. The changes being proposed here feel like they touch a lot more on the general architecture and direction of how we're designing this code going forward. This puts the proposal into a different category than if the changes represented a "business logic" change that fits the current structure.

I've got some broad code structure concerns about the proposal, both in terms of the code complexity and the way that the ipcache module interfaces with other modules. I've discussed some scenarios below, and raised some brainstorming threads on possible alternatives. I do have some vested interest/opinions in this area, but at the same time I've tried to focus my feedback on technical discussion topics that we can hopefully iterate and explore further together.

pkg/ipcache/ipcache.go
squeed: You are quite correct that there is an asynchronous-update problem ( prefix A has identities 123 and 456 allocated, but only 456 is "committed"). To some extent, we have this problem now as well.

As it stands right now in `main`, when we ingest a network policy, the steps are, basically,
1. Allocate CIDR identities
2. Insert policy in the Repository
3. Trigger regeneration for endpoints (this is asynchronous, but may proceed immediately)
4. Upsert CIDR metadata, trigger InjectLabels, which does...
5. UpdatePolicyMaps, thus recalculating policy to use the new identity
6. Update ipcache BPF

Just as you've outlined, the identity for a CIDR can also change anywhere in this series of steps. So, it seems that *some* eventual consistency is inevitable.

pkg/ipcache/ipcache.go
squeed: I like the idea of preserving the existing ipcache behavior and waiting for the updates to make it *back* to the selector cache.

We could implement this one of several ways; my first thought is to block in policy distillation until every CachedSelector (which performed allocation) has been updated with the identities for those allocations.

This would also work well with the fqdn proxy; rather than having the dns layer be responsible for allocation and the selector cache be responsible for deallocation, we could change it so the dns layer only thinks about IPs, and the selectorcache concerns itself with identities.

Since policy distillation necessarily requires the endpoint lock, it means that the ipcache must **never require endpoint locks** to make progress. Right now, this is not true, but not for any fundamental reasons. There are just some read-only fields behind a mutex that can easily be switched to atomic pointers.

Thoughts?

pkg/ipcache/ipcache.go
squeed: Based on what you've described, its a bit as if the individual components themselves are transitioning to a "level-driven" architecture. However, streams are not enough for that; one also needs a way to wait for the system to reach a given state. Typically, this requires the stream's consumers to maintain a local store. That, then requires some kind of initial listing. Oh wait, what do we have: an Informer :-).

To make this concrete, what does it mean for a given Endpoint to have a given policyRevision? That is a single serial number that encompasses several states: policy ingestion, endpoint identity, and peer identity just to name a few.

So, if we have a consumer that relies on the assertion "endpoint X has policyRevision 42". That is actually the assertion "endpoint X has policyRevision 42, and all the dependent resources (read: identities) are also extant". But that can be decomposed to "endpoint X has policyRevision 42, which also means "CachedSelector `1.1.1.1/32` has a valid identity". The question, then, is who waits for all the state to coalesce? I suspect that's a local decision - sometimes the waiting is local to the leaf resource (preserving synchronicity), and sometimes a promise / waitgroup is returned, and the caller waits for consistency at the end of the "transaction".

Incidentally, as it stands right now, the policy engine (Repository and SelectorCache), IPCache, and IDAllocator are all part of a single Cell. I'm curious what a unified interface to them should look like (in the context of requests and Streams). In a magical distant world, can we coalesce all of these in to a single, simple interface? What are the inputs and outputs? And, based on that, does it make sense for their datastores to be unified as well?

pkg/ipcache/ipcache.go
squeed: This plays well with the k8s notion of level-triggered-ness. Every component has a local store and a stream of changes. If those components require "downstream" state, they check their store to see if any changes are needed, "propose" that change, and watch the change stream until the desired state is achieved.

It makes debugging inter-related issues very hard of course :-), but it does mean that deadlocks are less likely.

squeed:
> I'm recognizing that there's a lack of design docs and state sharing around a lot of the logic here, which I'll take partial responsibility for... sorry about that :-)

No worries at all. In fact, I feel I've a fairly good idea of where this all is going; and I've been trying to help work towards that goal. Especially in leaving the door open for the fqdn identity dance to be simplified.


> Although I feel that I have a fair understanding of where this proposal is coming from, I admit that I found it difficult to follow the full reasoning about the approach.

Apologies for not writing more. This started out as an experiment, and most of the design discussions have been too in-person and "backroom".

> I've got some broad code structure concerns about the proposal, both in terms of the code complexity and the way that the ipcache module interfaces with other modules. I've discussed some scenarios below, and raised some brainstorming threads on possible alternatives. I do have some vested interest/opinions in this area, but at the same time I've tried to focus my feedback on technical discussion topics that we can hopefully iterate and explore further together.

I've commented below; I think you're broadly right and this PR doesn't take the correct approach. I'll try and play around with an asynchronous implementation.

pkg/ipcache/ipcache.go
joestringer: Agree we have an eventual consistency model, though with a bit less complexity in the ipcache interface I think?

Step (1) allocate CIDR identities only occurs async as part of step (4) from the `InjectLabels()` call in main today but yeah I agree it's a bit odd even today because the policy calculation is triggered separately from the identity allocation. So the endpoint regeneration -> policy calculation step may calculate policy without taking into account the new identities for the new prefixes, then those new identities/policy impacts will only be applied at step (5).

pkg/ipcache/ipcache.go
joestringer: Agree yes, I think conceptually the policyRevision would ideally be represented as a tree of dependent other revisions - maybe things like datapath revision, selectorcache revision, etc., but none of the latter revisions are formalized today. I'm not sure what the answer to the qusetion of "who waits" is, and whether we can generally set some expectation that applies tree-wide or whether it ends up being up to each "level" / "subsystem".

Re: Cell split of policy, ipcache, identity.. I think that ipcache and identity likely should be tied together with any interfacing being done through something more like the ipcache interfaces (input: "prefix X has properties Y", eg labels/auxiliary data and output: "prefix X has identity+aux Z"). As for the policy engine, I think longer term there's probably a "policy ingestion" cell and some sort of "identity ingestion" cell (the latter including selectorcache). I suspect that the 'output' parameters that this PR is proposing as part of selectorcache are convenient in selectorcache today because the selectorcache can efficiently map to specific, more detailed policy objects that are tied into endpoints.. so if we can tease out that efficient mapping into a "policy realization" cell (may even be part of the datapath interface/cell), that may serve to cleanly separate concerns.

squeed:
Closing this in favor of a new approach.

PR #25583, opened 2023-05-22
daemon: Separate endpoint API handlers from daemon and add tests

joamaki:
This moves the handlers for `/endpoint` to `daemon/restapi/endpoint.go` and thus decouples them from `type Daemon struct` which reduces their dependencies and allows to easily test them.

While most of the handlers now only require access to `EndpointManager` API, the endpoint creation still need to go through `Daemon` as not everything it depends on has been modularized (it is close though!). For creation `Daemon` implements `EndpointModifier` API which the endpoint handlers can depend on.

For testing the handlers this PR also introduces the `server.TestServer` which runs the API server using `httptest`.

### DIFF
--- a/daemon/restapi/endpoint_test.go
+++ b/daemon/restapi/endpoint_test.go
@@ -1,0 +1,394 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package restapi
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/go-openapi/runtime"
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/exp/maps"
+
+	"github.com/cilium/cilium/api/v1/models"
+	"github.com/cilium/cilium/api/v1/server"
+	"github.com/cilium/cilium/pkg/client"
+	"github.com/cilium/cilium/pkg/endpoint"
+	endpointid "github.com/cilium/cilium/pkg/endpoint/id"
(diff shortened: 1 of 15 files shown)


daemon/restapi/endpoints_test.go
```diff
@@ -0,0 +1,395 @@
+// SPDX-License-Identifier: Apache-2.0
```
joamaki: TODO remember to rename this file as well

aanm (approved):

daemon/restapi/endpoint_test.go
```diff
@@ -0,0 +1,394 @@
...
+	})
+	t.Run("200", func(t *testing.T) {
+		for _, epNumID := range testEndpointIDs {
+			epID := endpointid.NewCiliumID(int64(epNumID))
+			ok, err := ctx.client.EndpointGet(epID)
+			assert.NoError(t, err)
+			if assert.NotNil(t, ok) {
+				assert.EqualValues(t, epNumID, ok.ID)
+			}
+			// TODO what else to assert?
```
bimmlerd, line 189: I'd vote in favour of having golden tests here; manually asserting seems more error-prone

bimmlerd (approved):

joestringer (changes requested):
LGTM from the perspective of improving code cleanliness, code coverage in testing. There's a scattering of TODOs that look like they should be grappled with prior to merging. Once those are addressed, feel free to dismiss this review and go ahead with merging.

PR #25589, opened 2023-05-22
helm: Improve spire template 

### LINKED ISSUE #24552: Ensure mutual auth and Policy interact in safe ways
Spun out from https://github.com/cilium/design-cfps/pull/5#discussion_r1142674630

This issue covers ensuring that we ensure that Cilium can't be used to block its own mutual auth, or that mutual auth policy can't be used to block things that are required for mutual auth to work.

From @joestringer:
> For instance (different options):
> 
> * We have some structure internally that tracks "exceptions to auth", and those peers are _always_ excluded from the auth requirement, transparently
>   
>   * Given the requirement for `toEndpoints`, kube-apiserver might be transparently protected since it's _probably_ not running as a Pod in the cluster so can't be targeted with such policy statements. Does the same apply to SPIRE?
> * We expose some CLI config flag that excludes certain labels, IPs or tuples from auth, then set some sort of default there, and the user can configure it to exclude other things from auth if they have specific constraints in their environment.
> * We provide documentation to protect against footgunnery
> 
> A couple of other corner cases come to mind:
> 
> * Can these statements be used for host firewall?
> * Etcd configuration also comes to mind here for Cilium<->etcd in etcd configuration mode
> * Cilium <-> Cilium auth connections must be exempt

An issue for exceptions to host firewall for mutual auth authentication traffic: #24542.

sayboras:
Please refer to individual commit for more details.

Fixes: https://github.com/cilium/cilium/issues/24552

### DIFF
--- a/Documentation/helm-values.rst
+++ b/Documentation/helm-values.rst
@@ -139,7 +139,7 @@
    * - auth.mTLS.spire.serverAddress
      - SPIRE server address
      - string
-     - ``"spire-server.cilium-spire.svc.cluster.local:8081"``
+     - ``"spire-server.cilium-spire.svc:8081"``
    * - auth.mTLS.spire.trustDomain
      - SPIFFE trust domain to use for fetching certificates
      - string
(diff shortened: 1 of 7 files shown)


zacharysarah (approved):
LGTM from a docs perspective

kaworu (approved):

sayboras:
/test-1.26-net-next

sayboras:
/test-1.26-net-next

joamaki (approved):

PR #25590, opened 2023-05-22
crd: Refactor RegisterCRDsCell to be extensible

pippolo84:
Current implementation of RegisterCRDsCell is strictly tied to the registration of the CRDs belonging to the "cilium.io" group.

This PR refactors the cell to make it easily extendable if other groups needs to register their CRDs at operator startup.

### DIFF
--- a/pkg/k8s/apis/cell.go
+++ b/pkg/k8s/apis/cell.go
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Authors of Cilium
 
-package client
+package apis
 
 import (
 	"fmt"
@@ -11,6 +11,7 @@ import (
 
 	"github.com/cilium/cilium/pkg/hive"
 	"github.com/cilium/cilium/pkg/hive/cell"
+	"github.com/cilium/cilium/pkg/k8s/apis/cilium.io/client"
 	k8sClient "github.com/cilium/cilium/pkg/k8s/client"
 )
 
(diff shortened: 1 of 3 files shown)


joamaki (approved):
LGTM. At some point we could also consider making the CRDs we want to register a group value.

pippolo84:
> LGTM. At some point we could also consider making the CRDs we want to register a group value.

Actually, I wasn't completely satisfied by the package level variable and tried the group approach.
I think the group is better, so I'm gonna push-force the new commit.

brlbil (approved):
LGTM

pkg/k8s/apis/cell.go
```diff
@@ -61,13 +70,28 @@ func createCRDs(p params) {
...
-				return fmt.Errorf("unable to create CRDs: %w", err)
+			for _, f := range p.RegisterCRDsFuncs {
+				if err := f(p.Clientset); err != nil {
+					return fmt.Errorf("unable to create CRDs: %w", err)
+				}
 			}
+
 			return nil
 		},
 		OnStop: func(_ hive.HookContext) error {
```
joamaki: No need to provide an implementation for the stop hook. `nil` OnStop will be skipped.

joamaki (approved):

pippolo84:
External workloads flake tracked in https://github.com/cilium/cilium/issues/25736, going to rerun.

PR #25593, opened 2023-05-22
Removes IngressBandwidth Annotation

### LINKED ISSUE #25532: CFP: Remove kubernetes.io/ingress-bandwidth Annotation
## Cilium Feature Proposal
        
The [kubernetes.io/ingress-bandwidth](https://pkg.go.dev/github.com/cilium/cilium/pkg/bandwidth) annotation exists but is not implemented. If used, the pod controller simply logs a message stating the annotation is not supported.

To avoid the possibility of confusion, remove the annotation and user documentation that references this annotation, [for example](https://pkg.go.dev/github.com/cilium/cilium/pkg/bandwidth).

See [this](https://cilium.slack.com/archives/C2B917YHE/p1684336714340109) Slack thread for additional background. Feel free to assign me if others agree with this proposed feature.

danehans:
The IngressBandwidth is unimplemented but is represented as an exportable constant. This PR removes the IngressBandwidth annotation to avoid potential confusion and misuse.

Fixes: #25532

### DIFF
--- a/daemon/cmd/endpoint.go
+++ b/daemon/cmd/endpoint.go
@@ -42,6 +42,9 @@ import (
 	"github.com/cilium/cilium/pkg/proxy"
 )
 
+// ingressBandwidth represents the K8s ingress bandwidth Pod annotation (currently unsupported).
+const ingressBandwidth = "kubernetes.io/ingress-bandwidth"
+
 var errEndpointNotFound = errors.New("endpoint not found")
 
 type getEndpoint struct {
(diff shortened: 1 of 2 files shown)


danehans:
cc: @pchaigno

jschwinger233 (commented):
Could you please explain why we intend to remove this annotation from the package constant?  I understand the reasons mentioned in the Slack thread, but it would be preferable to have them documented in the commit message.

daemon/cmd/endpoint.go
```diff
@@ -430,12 +430,13 @@ func (d *Daemon) createEndpoint(ctx context.Context, owner regeneration.Owner, e
 			}
 			addLabels.MergeLabels(identityLabels)
 			infoLabels.MergeLabels(info)
-			if _, ok := annotations[bandwidth.IngressBandwidth]; ok {
+			ingBandwidth := "kubernetes.io/ingress-bandwidth"
```
borkmann: How is this any better? I thought from your explanation that bandwidth.IngressBandwidth is unused and should be removed. Yet it is used in the code. So your change makes it worse.

danehans: @borkmann thanks for the review. IMHO this is better b/c the annotation can no longer be imported by another pkg ([xref](https://go.dev/ref/spec#Exported_identifiers)). @pchaigno suggested in the associated Slack thread to remove the constant. Based on your feedback, I will update the PR to remove the associated log message.

pchaigno: Did you consider simply not exporting the constant outside of the package where it's defined, eventually by moving it to another package?

danehans: @pchaigno ^ is an option too. Does the team prefer the removal of the constant and associated log message or move the constant to the cmd pkg, make it internal, and keep the log message?

danehans:
@jschwinger233 Thanks for the review. I updated the PR description and commit message based on your feedback, PTAL.

danehans:
@qmonnet @pchaigno @borkmann @jschwinger233 I resolved feedback on this PR and it's passing CI. Is anything else needed to get approvals?

brb (changes requested):
Thanks for the PR!

However, we tend to keep all related consts in the same place, and [our docs](https://docs.cilium.io/en/stable/network/kubernetes/bandwidth-manager/) already should protect from a potential misuse.

PR #25598, opened 2023-05-22
probes: use veth device for probing managed neighbor support

ti-mo:
Some distributions don't ship with dummy device support/module, leading to the following error when probing for managed neighbors:

```
level=fatal msg="failed to probe managed neighbor support"
error="failed to add dummy link: operation not supported" subsys=probes
```

This patch uses a veth device instead, which should be available on nodes running Cilium, if it's not already loaded.

```release-note
Use veth device for probing managed neighbor support
```

### DIFF
--- a/pkg/datapath/linux/probes/managed_neighbors.go
+++ b/pkg/datapath/linux/probes/managed_neighbors.go
@@ -70,17 +70,22 @@ func haveManagedNeighbors() (outer error) {
 		runtime.UnlockOSThread()
 	}()
 
-	la := netlink.NewLinkAttrs()
-	la.Name = "cilium-dummy"
+	// Use a veth device instead of a dummy to avoid the kernel having to modprobe
+	// the dummy kmod, which could potentially be compiled out. veth is currently
+	// a hard dependency for Cilium, so safe to assume the module is available if
+	// not already loaded.
+	veth := &netlink.Veth{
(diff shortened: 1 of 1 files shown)


rgo3 (approved):

ti-mo:
/test-1.26-net-next

pkg/datapath/linux/probes/managed_neighbors.go
```diff
@@ -70,17 +70,22 @@ func haveManagedNeighbors() (outer error) {
...
 	}()
 
-	la := netlink.NewLinkAttrs()
-	la.Name = "cilium-dummy"
+	// Use a veth device instead of a dummy to avoid the kernel having to modprobe
+	// the dummy kmod, which could potentially be compiled out. veth is currently
+	// a hard dependency for Cilium, so safe to assume the module is available if
+	// not already loaded.
+	veth := &netlink.Veth{
+		LinkAttrs: netlink.LinkAttrs{Name: "veth0"},
```
squeed, line 78: are you worried about name collisions? what about `cilium-dummy-veth0`?

ti-mo: This runs in a new netns, so no.

ti-mo:
/test-1.26-net-next

PR #25600, opened 2023-05-22
pkg/ipam: Update histogram buckets for trigger metrics

hemanthmalla:
Currently, trigger related histogram metrics in `pgk/ipam` use the default prometheus histogram buckets. Resync operation in cloud providers like Azure tend to take a long time and the current buckets are inadequate to track changes in behavior. This commit extends the buckets to allow for measuring longer durations.

Currently, some metrics plateau at 10 secs
image

Reusing buckets [defined in Kubernetes API server to measure request duration](https://github.com/kubernetes/kubernetes/blob/ce05a4f7fcc234dd3132a4e40e6afdf0b2ebc94e/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go#L96-L100 )

### DIFF
--- a/pkg/ipam/metrics/metrics.go
+++ b/pkg/ipam/metrics/metrics.go
@@ -290,12 +290,16 @@ func NewTriggerMetrics(namespace, name string) *triggerMetrics {
 			Subsystem: ipamSubsystem,
 			Name:      name + "_duration_seconds",
 			Help:      "Duration of trigger runs",
+			Buckets: []float64{0.005, 0.025, 0.05, 0.1, 0.2, 0.4, 0.6, 0.8, 1.0, 1.25, 1.5, 2, 3,
+				4, 5, 6, 8, 10, 15, 20, 30, 45, 60},
 		}),
 		latency: prometheus.NewHistogram(prometheus.HistogramOpts{
 			Namespace: namespace,
 			Subsystem: ipamSubsystem,
 			Name:      name + "_latency_seconds",
 			Help:      "Latency between queue and trigger run",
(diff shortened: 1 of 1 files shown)


hemanthmalla:
Reusing buckets defined in Kubernetes API server to measure request duration

https://github.com/kubernetes/kubernetes/blob/ce05a4f7fcc234dd3132a4e40e6afdf0b2ebc94e/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go#L96-L100

gandro (approved):
The change overall looks fine by me. But I don't know much about metric best practises. Is there any overlap with the ongoing work in https://github.com/cilium/cilium/pull/25256?

hemanthmalla:
I don't think its related since we're only updating the histogram buckets. https://github.com/cilium/cilium/pull/25256 doesn't seem to touch bucket values.

hemanthmalla:
@gandro net-next and test-runtime seems to be flaky now ? I don't see the corresponding Jenkins jobs for failures. Are we cleaning up jenkins jobs in less than a day now ?

gandro:
> @gandro net-next and test-runtime seems to be flaky now ? I don't see the corresponding Jenkins jobs for failures. Are we cleaning up jenkins jobs in less than a day now ?

We had a Jenkins outage last week and had to re-provision all Jenkins instances. During that process, we accidentally fell back to a Jenkins config to only retained the last 30 jobs. That has now been fixed to retain job logs up to 15 days. If you rerun those pipelines, you should see be able to access the failure now.

Issue #25243, opened 2023-05-03
CFP: High-Scale IPcache


pchaigno:
Actual CFP: https://github.com/cilium/design-cfps/pull/7.

PR #25601, opened 2023-05-22
High-scale IPcache: Chapter 5

pchaigno:

This pull request adds support for endpoint routes in high-scale ipcache mode. See commits for details.

```release-note
Support `enable-endpoint-routes` with `enable-high-scale-ipcache`.
```

### DIFF
--- a/bpf/bpf_host.c
+++ b/bpf/bpf_host.c
@@ -1209,6 +1209,8 @@ handle_srv6(struct __ctx_buff *ctx)
 __section("from-netdev")
 int cil_from_netdev(struct __ctx_buff *ctx)
 {
+	__u32 __maybe_unused src_id = 0;
+
 #ifdef ENABLE_NODEPORT_ACCELERATION
 	__u32 flags = ctx_get_xfer(ctx, XFER_FLAGS);
 #ifdef HAVE_ENCAP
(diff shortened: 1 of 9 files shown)


christarazi (approved):

bpf/bpf_host.c
```diff
@@ -1288,6 +1290,15 @@ int cil_from_netdev(struct __ctx_buff *ctx)
 #endif
 #endif
 
+#ifdef ENABLE_HIGH_SCALE_IPCACHE
+	ret = decapsulate_overlay(ctx, &src_id);
+	if (IS_ERR(ret))
+		send_drop_notify_error(ctx, src_id, ret, CTX_ACT_DROP,
```
jschwinger233: Missing `return`?

pchaigno: Nice catch! Thanks!

squeed (approved):

bpf/bpf_lxc.c
```diff
@@ -2194,12 +2193,42 @@ int cil_to_container(struct __ctx_buff *ctx)
...
+# ifdef ENABLE_HIGH_SCALE_IPCACHE
+	if (identity == WORLD_ID) {
+		struct endpoint_info *ep;
+		void *data, *data_end;
+		struct ipv6hdr *ip6;
+
+		if (!revalidate_data(ctx, &data, &data_end, &ip6))
+			return DROP_INVALID;
+
+		ep = __lookup_ip6_endpoint((union v6addr *)&ip6->saddr);
```
jschwinger233, line 2205: Could be a stupid question: should it be `ip6->daddr`?

If the traffic is from remote, local endpoint map doesn't have an entry for the saddr, right? So this lookup can only be useful for local-to-local traffic. In that case, can we perform policy check at `from-container`'s `handle_ipv4_from_lxc`, where we can get the `dst_sec_id` from local endpoint map if ipcache doesn't return an entry: https://github.com/cilium/cilium/blob/4535f132b70854b9deb9c987cec71a33cff31d00/bpf/bpf_lxc.c#L832-L834

pchaigno:
@squeed Better not trust those MLH labeling blindly. There was an open question from Gray above.

squeed:
@pchaigno argh, good catch, sorry.

pchaigno:
No big deal. Everything here is behind a feature flag so we won't break anything else.
I just thought I'd warn given MLH is often sneaky like that.

PR #25282, opened 2023-05-05
k8s: Use slim Node in LocalNode Resource and K8s watchers

joamaki:
We were performing unnecessary conversions from corev1.Node to slim_corev1.Node.
Only blocker for using slim_corev1.Node was the missing ProviderID.

PR #25604, opened 2023-05-22
Updates informer pkg to use TransformFunc()

### LINKED ISSUE #24923: Make use of the TransFunc introduced in k8s 1.27
Kubernetes has introduced in https://github.com/kubernetes/kubernetes/pull/116623, and backported to all branches, the TransFunc that allows to mutate certain resources in K8s Caches. This will allow us to remove our own TransFunc implementation functions in https://github.com/cilium/cilium/blob/264dd3f350022c7626438f2dab0401dcb41ec616/pkg/k8s/informer/informer.go#L101-L106 by using the DeltaFIFO directly from the library.

Depending how non-invasive the patch will be, we might consider backporting it to our stable branches as well.

cc @odinuge

danehans:
- Removes `ConvertFunc` in favor of [TransformFunc](https://pkg.go.dev/k8s.io/client-go@v0.26.0/tools/cache#TransformFunc) for object transformation. Informers are now created with `TransformFunc` instead of `ConvertFunc` to mutate certain resources in K8s Caches.
- Renames ConvertXXX functions to TransformXXX.
- Adds tests to `pkg/k8s/init_test.go` to improve test coverage of transformer functions.

Fixes: #24923

### DIFF
--- a/operator/cmd/root.go
+++ b/operator/cmd/root.go
@@ -520,14 +520,24 @@ func (legacy *legacyOnLeader) onStart(_ hive.HookContext) error {
 							// k8s service for etcd. As soon the k8s caches are
 							// synced, this hijack will stop happening.
 							sc := k8s.NewServiceCache(nil)
-							slimSvcObj := k8s.ConvertToK8sService(k8sSvc)
-							slimSvc := k8s.ObjToV1Services(slimSvcObj)
-							if slimSvc == nil {
-								// This will never happen but still log it
-								scopedLog.Warnf("BUG: invalid k8s service: %s", slimSvcObj)
+							slimSvcObj, err := k8s.TransformToK8sService(k8sSvc)
+							if err != nil {
+								scopedLog.WithFields(logrus.Fields{
+									logfields.ServiceName:      k8sSvc.Name,
+									logfields.ServiceNamespace: k8sSvc.Namespace,
+								}).Error("Failed to transform k8s service")
+							} else {
+								slimSvc := k8s.ObjToV1Services(slimSvcObj)
+								if slimSvc == nil {
+									// This will never happen but still log it
+									scopedLog.WithFields(logrus.Fields{
+										logfields.ServiceName:      k8sSvc.Name,
+										logfields.ServiceNamespace: k8sSvc.Namespace,
+									}).Warn("BUG: invalid k8s service")
+								}
+								sc.UpdateService(slimSvc, nil)
+								svcGetter = operatorWatchers.NewServiceGetter(sc)
 							}
-							sc.UpdateService(slimSvc, nil)
-							svcGetter = operatorWatchers.NewServiceGetter(sc)
 						case k8sErrors.IsNotFound(err):
 							scopedLog.Error("Service not found in k8s")
 						default:
--- a/operator/watchers/cilium_endpoint.go
+++ b/operator/watchers/cilium_endpoint.go
@@ -108,7 +108,7 @@ func CiliumEndpointsInit(ctx context.Context, wg *sync.WaitGroup, clientset k8sC
 			&cilium_api_v2.CiliumEndpoint{},
 			0,
 			cacheResourceHandler,
-			convertToCiliumEndpoint,
+			transformToCiliumEndpoint,
 			CiliumEndpointStore,
 		)
 
@@ -123,11 +123,12 @@ func CiliumEndpointsInit(ctx context.Context, wg *sync.WaitGroup, clientset k8sC
 	})
 }
 
-// convertToCiliumEndpoint converts a CiliumEndpoint to a minimal CiliumEndpoint
+// transformToCiliumEndpoint transforms a CiliumEndpoint to a minimal CiliumEndpoint
 // containing only a minimal set of entities used to identity a CiliumEndpoint
 // Warning: The CiliumEndpoints created by the converter are not intended to be
-// used for Update operations in k8s.
-func convertToCiliumEndpoint(obj interface{}) interface{} {
+// used for Update operations in k8s. If the given obj can't be cast into either
+// CiliumEndpoint nor DeletedFinalStateUnknown, an error is returned.
+func transformToCiliumEndpoint(obj interface{}) (interface{}, error) {
 	switch concreteObj := obj.(type) {
 	case *cilium_api_v2.CiliumEndpoint:
 		p := &cilium_api_v2.CiliumEndpoint{
@@ -147,11 +148,11 @@ func convertToCiliumEndpoint(obj interface{}) interface{} {
 			},
 		}
 		*concreteObj = cilium_api_v2.CiliumEndpoint{}
-		return p
+		return p, nil
 	case cache.DeletedFinalStateUnknown:
 		ciliumEndpoint, ok := concreteObj.Obj.(*cilium_api_v2.CiliumEndpoint)
 		if !ok {
-			return obj
+			return nil, fmt.Errorf("unknown object type %T", concreteObj.Obj)
 		}
 		dfsu := cache.DeletedFinalStateUnknown{
 			Key: concreteObj.Key,
@@ -174,9 +175,9 @@ func convertToCiliumEndpoint(obj interface{}) interface{} {
 		}
 		// Small GC optimization
 		*ciliumEndpoint = cilium_api_v2.CiliumEndpoint{}
-		return dfsu
+		return dfsu, nil
 	default:
-		return obj
+		return nil, fmt.Errorf("unknown object type %T", concreteObj)
 	}
 }
 
--- a/pkg/k8s/informer/informer.go
+++ b/pkg/k8s/informer/informer.go
@@ -36,47 +36,42 @@ func init() {
 	)
 }
 
-type ConvertFunc func(obj interface{}) interface{}
-
-// NewInformer is a copy of k8s.io/client-go/tools/cache/NewInformer with a new
-// argument which converts an object into another object that can be stored in
-// the local cache.
+// NewInformer is a copy of k8s.io/client-go/tools/cache/NewInformer includes the default cache MutationDetector.
 func NewInformer(
 	lw cache.ListerWatcher,
 	objType k8sRuntime.Object,
 	resyncPeriod time.Duration,
 	h cache.ResourceEventHandler,
-	convertFunc ConvertFunc,
+	transformer cache.TransformFunc,
 ) (cache.Store, cache.Controller) {
 	// This will hold the client state, as we know it.
 	clientState := cache.NewStore(cache.DeletionHandlingMetaNamespaceKeyFunc)
 
-	return clientState, NewInformerWithStore(lw, objType, resyncPeriod, h, convertFunc, clientState)
+	return clientState, NewInformerWithStore(lw, objType, resyncPeriod, h, transformer, clientState)
 }
 
-// NewIndexerInformer is a copy of k8s.io/client-go/tools/cache/NewIndexerInformer with a new
-// argument which converts an object into another object that can be stored in
-// the local cache.
+// NewIndexerInformer is a copy of k8s.io/client-go/tools/cache/NewIndexerInformer but includes the
+// default cache MutationDetector.
 func NewIndexerInformer(
 	lw cache.ListerWatcher,
 	objType k8sRuntime.Object,
 	resyncPeriod time.Duration,
 	h cache.ResourceEventHandler,
-	convertFunc ConvertFunc,
+	transformer cache.TransformFunc,
 	indexers cache.Indexers,
 ) (cache.Indexer, cache.Controller) {
 	clientState := cache.NewIndexer(cache.DeletionHandlingMetaNamespaceKeyFunc, indexers)
-	return clientState, NewInformerWithStore(lw, objType, resyncPeriod, h, convertFunc, clientState)
+	return clientState, NewInformerWithStore(lw, objType, resyncPeriod, h, transformer, clientState)
 }
 
-// NewInformerWithStore uses the same arguments as NewInformer for which a
-// caller can also set a cache.Store.
+// NewInformerWithStore uses the same arguments as NewInformer for which a caller can also set a
+// cache.Store and includes the default cache MutationDetector.
 func NewInformerWithStore(
 	lw cache.ListerWatcher,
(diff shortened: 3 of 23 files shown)


operator/watchers/cilium_endpoint.go
```diff
@@ -147,11 +148,11 @@ func convertToCiliumEndpoint(obj interface{}) interface{} {
 			},
 		}
 		*concreteObj = cilium_api_v2.CiliumEndpoint{}
-		return p
+		return p, nil
 	case cache.DeletedFinalStateUnknown:
 		ciliumEndpoint, ok := concreteObj.Obj.(*cilium_api_v2.CiliumEndpoint)
 		if !ok {
-			return obj
+			return nil, fmt.Errorf("unknown object type %v", concreteObj.Obj)
```
christarazi:
```suggestion
			return nil, fmt.Errorf("unknown object type %T", concreteObj.Obj)
```

Please apply this to all other instances

operator/cmd/root.go
```diff
@@ -512,14 +512,18 @@ func (legacy *legacyOnLeader) onStart(_ hive.HookContext) error {
...
 							// synced, this hijack will stop happening.
 							sc := k8s.NewServiceCache(nil)
-							slimSvcObj := k8s.ConvertToK8sService(k8sSvc)
-							slimSvc := k8s.ObjToV1Services(slimSvcObj)
-							if slimSvc == nil {
-								// This will never happen but still log it
-								scopedLog.Warnf("BUG: invalid k8s service: %s", slimSvcObj)
+							slimSvcObj, err := k8s.TransformToK8sService(k8sSvc)
+							if err != nil {
+								scopedLog.Errorf("Failed to transform k8s service %s/%s", k8sSvc.Namespace, k8sSvc.Name)
```
christarazi: There should be some relevant logfields in `pkg/logging/logfields` that we can use here without making the log msg dynamic. This enables the use of structured logging.

```suggestion
								scopedLog.WithError(err).WithFields({
								  ...
								}.)Error("Failed to transform k8s service")
```

operator/cmd/root.go
```diff
@@ -512,14 +512,18 @@ func (legacy *legacyOnLeader) onStart(_ hive.HookContext) error {
...
+							} else {
+								slimSvc := k8s.ObjToV1Services(slimSvcObj)
+								if slimSvc == nil {
+									// This will never happen but still log it
```
christarazi, line 532: Same here regarding structured logging

christarazi (changes requested):
Thanks for the PR! A few comments to address below.

pkg/k8s/informer/informer.go
```diff
@@ -36,37 +36,31 @@ func init() {
...
 }
 
-type ConvertFunc func(obj interface{}) interface{}
-
-// NewInformer is a copy of k8s.io/client-go/tools/cache/NewInformer with a new
-// argument which converts an object into another object that can be stored in
-// the local cache.
-func NewInformer(
+// NewTransformingInformer is a copy of k8s.io/client-go/tools/cache/NewTransformingInformer.
+func NewTransformingInformer(
```
giorio94: Given that this function is now a vanilla copy of the upstream one, I would suggest to update the callers to use the upstream one instead (either `NewInformer` or `NewTransformingInformer` depending on whether the transformer function is nil).

joamaki: Looks like `pkg/k8s/informer` does two things: it allows using a custom store and it injects a `CacheMutationDetector`. The custom store we can't do directly with the upstream informer, but the latter can be done with a transforming informer. Since we can't get rid of the custom store use yet, it's probably easier to leave the callers as is for now.

EDIT: Ah, I don't think we actually enable the mutation detection anywhere (`git grep KUBE_CACHE_MUTATION_DETECTOR` )

giorio94: Yeah, I was referring to `NewTransformingInformer` only, which doesn't take a custom store as parameter. But I missed that there is currently the custom `CacheMutatorDetector` part. Given that, I agree that it is probably better to not modify the callers for the moment (I would suggest to update the comment though, since it is currently a bit misleading).

pkg/nodediscovery/nodediscovery.go
```diff
@@ -440,15 +439,13 @@ func (n *NodeDiscovery) mutateNodeResource(nodeResource *ciliumv2.CiliumNode) er
...
+		UID:        slimNode.UID,
 	}}
-	providerID = k8sNode.Spec.ProviderID
+	providerID = slimNode.Spec.ProviderID
 
 	// Get the addresses from k8s node and add them as part of Cilium Node.
 	// Cilium Node should contain all addresses from k8s.
-	nodeInterface := k8s.ConvertToNode(k8sNode)
-	typesNode := nodeInterface.(*k8sTypes.Node)
-	k8sNodeParsed := k8s.ParseNode(typesNode, source.Unspec)
```
danehans, line 451: __Note:__ This transformation was removed since https://github.com/cilium/cilium/pull/25282 updated `GetK8sNode()` to return a slim node instead of a k8s node.

pkg/k8s/informer/informer.go
```diff
@@ -36,47 +36,46 @@ func init() {
...
 	clientState := cache.NewIndexer(cache.DeletionHandlingMetaNamespaceKeyFunc, indexers)
-	return clientState, NewInformerWithStore(lw, objType, resyncPeriod, h, convertFunc, clientState)
+	return clientState, NewInformerWithStore(lw, objType, resyncPeriod, h, transformer, clientState)
 }
 
-// NewInformerWithStore uses the same arguments as NewInformer for which a
-// caller can also set a cache.Store.
+// NewInformerWithStore uses the same arguments as NewInformer for which a caller can also set a
+// cache.Store that includes the default cache MutationDetector. The MutationDetector provides a
+// way to detect if a cached object has been mutated.
```
giorio94: Nit. The MutationDetector is actually unrelated from the Store.

pkg/k8s/informer/informer.go
```diff
@@ -36,47 +36,46 @@ func init() {
...
-type ConvertFunc func(obj interface{}) interface{}
-
-// NewInformer is a copy of k8s.io/client-go/tools/cache/NewInformer with a new
-// argument which converts an object into another object that can be stored in
-// the local cache.
-func NewInformer(
+// NewTransformingInformer is a copy of k8s.io/client-go/tools/cache/NewTransformingInformer but creates
+// a cache store that includes the default cache MutationDetector. The MutationDetector provides a way
+// to detect if a cached object has been mutated.
+func NewTransformingInformer(
```
giorio94: Nit. I wonder if it could be worth to revert the renaming of this function (and of the one below) to reduce the code churn. I don't have a strong opinion though.

giorio94 (approved):
Thanks! Changes LGTM

Just a couple of nits inline. Could you please also extend the commit message with the context around this change?

danehans:
@giorio94 thanks for the /approval. I updated the PR description and commit 44598e0 resolves your nits.

giorio94:
> @giorio94 thanks for the /approval. I updated the PR description and commit [44598e0](https://github.com/cilium/cilium/commit/44598e0a48fd040eeafa5871a14d58fd58130681) resolves your nits.

Thanks! Could you please also extend the commit message with a bit of description?

mhofstetter (commented):
LGTM, thanks!

mhofstetter (approved):

pkg/nodediscovery/nodediscovery.go
```diff
@@ -440,15 +439,13 @@ func (n *NodeDiscovery) mutateNodeResource(nodeResource *ciliumv2.CiliumNode) er
...
 		Name:       nodeTypes.GetName(),
-		UID:        k8sNode.UID,
+		UID:        slimNode.UID,
 	}}
-	providerID = k8sNode.Spec.ProviderID
+	providerID = slimNode.Spec.ProviderID
 
 	// Get the addresses from k8s node and add them as part of Cilium Node.
 	// Cilium Node should contain all addresses from k8s.
-	nodeInterface := k8s.ConvertToNode(k8sNode)
```
gandro, line 449: This seems to have been the last use of `k8s.ConvertToNode` function outside of test code. We should probably remove it (doesn't have to be done in this PR though)

gandro (approved):
Thanks!

pkg/nodediscovery/nodediscovery.go
danehans: @gandro thanks for catching this. I created https://github.com/cilium/cilium/issues/25680 to track the removal of the func.

danehans:
@christarazi you still have requested changes that I resolved from your review. PTAL when you have a moment.

christarazi (approved):

joamaki (approved):

pkg/k8s/informer/informer.go
```diff
@@ -99,8 +94,11 @@ func NewInformerWithStore(
 			for _, d := range obj.(cache.Deltas) {
 
 				var obj interface{}
-				if convertFunc != nil {
-					obj = convertFunc(d.Object)
+				if transformer != nil {
+					var err error
+					if obj, err = transformer(d.Object); err != nil {
+						return err
```
joamaki, line 101: What's the behavior of the informer when we return an error here?

danehans: The informer will not process the object. For example, the clustermesh-apiserver creates an informer that transforms k8s nodes to cilium nodes. If `TransformToCiliumNode()` returns an error, the provided object is not created/updated. We could set `RetryOnError` to re-enqueue the object. Thoughts?

danehans:
@christarazi and others, is there anything else needed to merge this PR?

danehans:
Commit c2078e9 rebases due to https://github.com/cilium/cilium/pull/25049.

danehans:
@christarazi thanks for kicking off the tests. The `test-1.26-net-next` job is failing due to:

```sh
/home/jenkins/workspace/Cilium-PR-K8s-1.26-kernel-net-next/src/github.com/cilium/cilium/test/ginkgo-ext/scopes.go:515
Unexpected missed tail call
Expected
    <int>: 1
to be ==
    <int>: 0
/home/jenkins/workspace/Cilium-PR-K8s-1.26-kernel-net-next/src/github.com/cilium/cilium/test/k8s/updates.go:129
```

It appears the upgrade was unsuccessful (^ calls `assertUpgradeSuccessful()`), possibly due to one of the agents failing readiness probes:
```sh
22:18:19 STEP: Cilium is not ready yet: connectivity health is failing: Cluster connectivity is unhealthy on 'cilium-rwcn6': Exitcode: 1 
Err: exit status 1
Stdout:
 	 
Stderr:
	 Error: Cannot get status/probe: Put "[http://%2Fvar%2Frun%2Fcilium%2Fhealth.sock/v1beta/status/probe":](http://%2Fvar%2Frun%2Fcilium%2Fhealth.sock/v1beta/status/probe) dial unix /var/run/cilium/health.sock: connect: no such file or directory
	 
	 command terminated with exit code 1
```

I don't believe this is related to the PR so I wanted to get your input on how to proceed.

christarazi:
It's flake https://github.com/cilium/cilium/issues/24514.

christarazi:
Let's see if the next test run can pass without flakes.

danehans:
@christarazi when you have a moment, can you please advise me on how to move this PR forward?

joestringer:
@danehans once the reviews are all approved from relevant code owner areas and the required CI tests are all passing, the maintainer's little helper marks the PR ready-to-merge like above. Someone then comes through and does last sanity checks and merges the PR. Thanks for the contribution!

odinuge:
cc @joestringer @danehans @christarazi ^

odinuge:
Also, feel free to ping me on PRs around this part of client-go, given I re(wrote) some of it a few months ago to fix some policy leaks we saw in cilium 😄

danehans:
@odinuge thanks for the feedback. As you can see [here](https://github.com/cilium/cilium/pull/25604/files#diff-fec30a88ffb515619228eb30191a763c3df4d6131791e74a6a2fae2b4265b69dL558-L599), this PR did not change the logic of `TransformToCCNP()`. Prior to this PR, the code you're referencing looked like:
```go
case cache.DeletedFinalStateUnknown:
		ccnp, ok := concreteObj.Obj.(*cilium_v2.CiliumClusterwideNetworkPolicy)
		if !ok {
			return obj
		}
```

and now the same codeblock returns nil and an error instead of the provided object:
```go
	case cache.DeletedFinalStateUnknown:
		ccnp, ok := concreteObj.Obj.(*cilium_v2.CiliumClusterwideNetworkPolicy)
		if !ok {
			return nil, fmt.Errorf("unknown object type %T", concreteObj.Obj)
		}
```

`TransformToCCNP()` before and after this PR expected `obj` to be a `CiliumClusterwideNetworkPolicy` (embedded in the case of `DeletedFinalStateUnknown`) and return a `SlimCNP` (embedded in `DeletedFinalStateUnknown`). This pattern is consistent with other TransformXXX funcs in the pkg. The associated tests were updated accordingly and currently pass when I run them locally. It appears that `TransformToCCNP()` is being called somewhere in the code when it shouldn't be, e.g. no need to transform an object that is already a `SlimXXX`.

odinuge:
Thanks for looking into this @danehans!

Yeah, the old godocs were pretty ambiguous, and the k8s behavior in `client-go` that the `informer` mimics, was pretty inconsistent previously. It still has some legacy it has to deal with, and i think thats the reason for the current function . I have also spent a lot of time wrapping my head around this. So I very much agree this is a bit confusing 😅 

I think it all boils down to this; https://github.com/cilium/cilium/blob/a82f0fc303c5ff511b2550b775849abec30dbb3e/pkg/k8s/informer/informer.go#L99-L102

There, we always call the transform funcs, just like `client-go`. So even if the object is an `DeletedFinalStateUnknown`, we run the transformer.

I should also have been explicit that my comment applies equally to all those transformer functions that return a different type than they receive; not only the CCNP one. 😄 

Should we maybe open a new issue to track this? What do you think @danehans? If you to implement a fix here, I'm happy to review! 😄  (tho. I'm heading out of the office for the next two weeks 🌴)

danehans:
I assigned myself https://github.com/cilium/cilium/issues/26240 and will push a PR to fix it.

PR #25612, opened 2023-05-23
add a verdict before use the pointer variable

pengbinbin1:

      GitHub issue.

Fixes: #issue-number

```release-note
<!-- I have deploy the cilium on my cluster,and the cilium-agent ds is crashed. I have checked the err log and discovered the function _getIPv4NativeRouting_   maybe return nil, so add the judge before use the pointer variable, i think return err when the pointer is nil is more gracefuly than crashed -->
```

### DIFF
--- a/pkg/cidr/cidr.go
+++ b/pkg/cidr/cidr.go
@@ -89,6 +89,14 @@ func (n *CIDR) Equal(o *CIDR) bool {
 	return Equal(n.IPNet, o.IPNet)
 }
 
+// String returns the string about the CIDR like "192.0.2.0/24"
+func (in *CIDR) String() string {
+	if in == nil {
+		return "<nil>"
+	}
+	return in.IPNet.String()
+}
+
 // Equal returns true if the n and o net.IPNet CIDRs arr Equal.
 func Equal(n, o *net.IPNet) bool {
 	if n == nil || o == nil {


jschwinger233 (commented):
Do you think we can add a `String()` method for `*cidr.CIDR` and check nil pointer inside there? My concern is there are other places with the same nil pointer risks, such as `kHost.GetCIDR().String()` at https://github.com/cilium/cilium/blob/847e8ad660abb07c4a99482d7b77a647ab269cca/pkg/maps/lbmap/source_range.go#L53.

Rather than return an error, we could also keep the same logic as `*net.IPNet`: return `<nil>` for nil pointer.

pengbinbin1:
> Do you think we can add a `String()` method for `*cidr.CIDR` and check nil pointer inside there? My concern is there are other places with the same nil pointer risks, such as `kHost.GetCIDR().String()` at
> 
> https://github.com/cilium/cilium/blob/847e8ad660abb07c4a99482d7b77a647ab269cca/pkg/maps/lbmap/source_range.go#L53
> 
> .
> Rather than return an error, we could also keep the same logic as `*net.IPNet`: return `<nil>` for nil pointer.

Yes，maybe add a check inside there is not appropriate. But i can not understand what you mean  about _**also keep the same logic as `*net.IPNet`: return `<nil>` for nil pointe**r_ , return nil for nil pointer  can cause to crash too.

pkg/datapath/iptables/iptables.go
```diff
@@ -1564,7 +1564,11 @@ func (m *IptablesManager) installRules(ifName string) error {
 	}
 
 	if skipPodTrafficConntrack(false) {
-		podsCIDR := option.Config.GetIPv4NativeRoutingCIDR().String()
+		nativeRoutingCIDR := option.Config.GetIPv4NativeRoutingCIDR()
+		if nativeRoutingCIDR == nil {
+			return fmt.Errorf("the IPv4NativeRoutingCIDR is nil")
+		}
```
jschwinger233: Let's discuss in the thread.

Could we check the nil pointer inside CIDR.String()? For example:

```go
func (cidr *CIDR) String() string {
    if cidr == nil {
        return "<nil>"
    }
    return cidr.IPNet.String()
}
```
So we won't be bothered by nil pointer of *CIDR at other places.

pengbinbin1: I got it and the code is added

pkg/datapath/iptables/iptables.go
```diff
@@ -1565,7 +1565,6 @@ func (m *IptablesManager) installRules(ifName string) error {
 
 	if skipPodTrafficConntrack(false) {
 		podsCIDR := option.Config.GetIPv4NativeRoutingCIDR().String()
-
```
squeed: nit: can you remove this delta from the change?

squeed:
Can you paste the traceback you saw on your cluster, so other users can find this? Is there an issue?

Separately, might you know in which version the bug may have appeared, so that we can backport the change as necessary?

pengbinbin1:
> Can you paste the traceback you saw on your cluster, so other users can find this? Is there an issue?
> 
> Separately, might you know in which version the bug may have appeared, so that we can backport the change as necessary?
> 
> If this fixes a user-facing bug, then we need to write a release note and, possibly, backport the patches accordingly.

cilium-crash
cilium version 1.13.0

pkg/datapath/iptables/iptables.go
pengbinbin1: done

jschwinger233:
From my point of view, your patch prevents the panic due to nil pointer, but cilium still runs into error because it can't install an iptables rule with `-s <nil>`. It's better for us to find the real root cause: why `option.Config.GetIPv4NativeRoutingCIDR()` returns a nil pointer rather a legit CIDR, Even if it's a misconfiguration issue, we'd better fail fast during validating configuration, right?

pengbinbin1:
> From my point of view, your patch prevents the panic due to nil pointer, but cilium still runs into error because it can't install an iptables rule with `-s <nil>`. It's better for us to find the real root cause: why `option.Config.GetIPv4NativeRoutingCIDR()` returns a nil pointer rather a legit CIDR, Even if it's a misconfiguration issue, we'd better fail fast during validating configuration, right?

Yes,it it better to find the bug duiring validating configuration, but i think the check about nil pointer  is also needed even though all the configuration is validated

jschwinger233:
To be honest, I feel reluctant to approve this PR, because the patch merely converts a panic into an error without any helpful information to tell user what's wrong; from the user's point of view, the error log is something like "failed to install iptables rules due to nil", which is confusing and hiding the real problems; from developer's point of view, this patch doesn't turn the probably wrong configured cilium cluster to normal, it's still broken.

@squeed What's your opinion?

ldelossa:
@pengbinbin1 please rebase this PR to fix broken tests.

Issue #25073, opened 2023-04-24
ipam/crd: Improve cloud API synchronization during node IP pool maintenance


jaffcheng:
In AWS/Alibabacloud ipam modes, every time the IP allocation/release happens, the cilium operator triggers a resync and fetches all the related data from cloud API, including VPCs, subnets, SGs, and ENIs. In a large (we have 10k ENIs and more to come) and frequently changing cluster, a full resync could take several tens of seconds and imposes a lot of pressure on the cloud API.
We could consider whether there is unnecessary data and see if full synchronization could be avoided.

jaffcheng:
Hi @ti-mo, I think the resync triggered by a pool maintenance event should not ferch all the ENIs, but only the ENIs related to the node:
https://github.com/cilium/cilium/blob/7335aa38a9c6a148b12b267ff6625a2290bb7d2a/pkg/ipam/node.go#L962

I plan to change the resync above to a single-node resync, and put the two types of resync into an eventqueue to avoid data races. With this change, I'd expect the IP allocation to be much faster when the periodic full resync is not holding the lock.

What do you think?

PR #25619, opened 2023-05-23
ipam,alibabacloud: Improve event driven instance resync

jaffcheng:
Currently in AWS/Alibabacloud ipam modes, every time the IP allocation/release event happens, the cilium operator triggers a resync and fetches all the ENIs from instance API. In a large and frequently changing cluster, the full sync might take several tens of seconds. This slows down the IP allocation process severely and also imposes a lot of pressure on the instance API. The full sync here should be unnecessary since we only need to update the ENIs of the instance that triggered the event.

This patch introduces a new `Node.instanceSync` trigger to replace `NodeManager.resyncTrigger`. Whenever an instance is mutated due to IP pool maintenance, the trigger attempts to incrementally resynchronize the corresponding instance to the local cache. This is achieved through the newly introduced `InstanceSync` method of the `AllocationImplementation` interface.
While this feature is implemented for Alibabacloud, AWS and Azure still fall back to full resynchronization.

Here are some time cost data from an Alibabacloud cluster during different periods:

                                  full sync    InstanceSync
    time cost with ~8000  ENIs    ~20s         ~2s
    time cost with ~13000 ENIs    ~35s         ~2s

Related: #25073

```release-note
ipam,alibabacloud: Improve event driven instance resync
```

### DIFF
--- a/pkg/alibabacloud/api/api.go
+++ b/pkg/alibabacloud/api/api.go
@@ -62,6 +62,30 @@ func NewClient(vpcClient *vpc.Client, client *ecs.Client, metrics MetricsAPI, ra
 	}
 }
 
+// GetInstance returns the instance including its ENIs by the given instanceID
+func (c *Client) GetInstance(ctx context.Context, vpcs ipamTypes.VirtualNetworkMap, subnets ipamTypes.SubnetMap, instanceID string) (*ipamTypes.Instance, error) {
+	instance := ipamTypes.Instance{}
+	instance.Interfaces = map[string]ipamTypes.InterfaceRevision{}
+
+	networkInterfaceSets, err := c.describeNetworkInterfacesByInstance(ctx, instanceID)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, iface := range networkInterfaceSets {
+		ifId := iface.NetworkInterfaceId
+		_, eni, err := parseENI(&iface, vpcs, subnets)
+		if err != nil {
+			return nil, err
+		}
+
+		instance.Interfaces[ifId] = ipamTypes.InterfaceRevision{
+			Resource: eni,
+		}
+	}
+	return &instance, nil
+}
+
 // GetInstances returns the list of all instances including their ENIs as instanceMap
 func (c *Client) GetInstances(ctx context.Context, vpcs ipamTypes.VirtualNetworkMap, subnets ipamTypes.SubnetMap) (*ipamTypes.InstanceMap, error) {
 	instances := ipamTypes.NewInstanceMap()
@@ -396,6 +420,35 @@ func (c *Client) describeNetworkInterfaces(ctx context.Context) ([]ecs.NetworkIn
 		} else {
 			req.NextToken = resp.NextToken
 		}
+	}
+
+	return result, nil
+}
+
+func (c *Client) describeNetworkInterfacesByInstance(ctx context.Context, instanceID string) ([]ecs.NetworkInterfaceSet, error) {
+	var result []ecs.NetworkInterfaceSet
+
+	for i := 1; ; {
+		req := ecs.CreateDescribeNetworkInterfacesRequest()
+		req.PageNumber = requests.NewInteger(i)
+		req.PageSize = requests.NewInteger(1000)
+		req.InstanceId = instanceID
+		c.limiter.Limit(ctx, "DescribeNetworkInterfaces")
+		resp, err := c.ecsClient.DescribeNetworkInterfaces(req)
+		if err != nil {
+			return nil, err
+		}
+		if len(resp.NetworkInterfaceSets.NetworkInterfaceSet) == 0 {
+			break
+		}
+
+		for _, v := range resp.NetworkInterfaceSets.NetworkInterfaceSet {
+			result = append(result, v)
+		}
+		if resp.TotalCount < resp.PageNumber*resp.PageSize {
+			break
+		}
+		i++
 	}
 
 	return result, nil
--- a/pkg/alibabacloud/eni/instances.go
+++ b/pkg/alibabacloud/eni/instances.go
@@ -19,6 +19,7 @@ import (
 
 // AlibabaCloudAPI is the API surface used of the ECS API
 type AlibabaCloudAPI interface {
+	GetInstance(ctx context.Context, vpcs ipamTypes.VirtualNetworkMap, subnets ipamTypes.SubnetMap, instanceID string) (*ipamTypes.Instance, error)
 	GetInstances(ctx context.Context, vpcs ipamTypes.VirtualNetworkMap, subnets ipamTypes.SubnetMap) (*ipamTypes.InstanceMap, error)
 	GetVSwitches(ctx context.Context) (ipamTypes.SubnetMap, error)
 	GetVPC(ctx context.Context, vpcID string) (*ipamTypes.VirtualNetwork, error)
@@ -36,6 +37,7 @@ type AlibabaCloudAPI interface {
 // by calling resync() regularly.
 type InstancesManager struct {
 	mutex          lock.RWMutex
+	resyncLock     lock.RWMutex
 	instances      *ipamTypes.InstanceMap
 	vSwitches      ipamTypes.SubnetMap
 	vpcs           ipamTypes.VirtualNetworkMap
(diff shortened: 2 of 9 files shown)


pkg/ipam/node_manager.go
```diff
@@ -370,6 +395,25 @@ func (n *NodeManager) Update(resource *v2.CiliumNode) (nodeSynced bool) {
 			return false
 		}
 
+		if n.instancesAPI.SupportInstanceSync() {
+			instanceSync, err := trigger.NewTrigger(trigger.Parameters{
+				Name:        fmt.Sprintf("ipam-node-instance-sync-%s", resource.Name),
+				MinInterval: 10 * time.Millisecond,
+				TriggerFunc: func(reasons []string) {
+					if syncTime, ok := node.instanceIncrementalSync(context.TODO(), resource.InstanceID()); ok {
```
tommyp1ckles: Should ensure that we constrain the timeout ourselves with the context?

pkg/alibabacloud/eni/instances.go
```diff
@@ -122,6 +123,55 @@ func (m *InstancesManager) Resync(ctx context.Context) time.Time {
 	return resyncStart
 }
 
+func (m *InstancesManager) SupportInstanceSync() bool {
+	return true
+}
+
+func (m *InstancesManager) InstanceSync(ctx context.Context, instanceID string) time.Time {
```
tommyp1ckles, line 94: opinion: using a time object to represent if the call succeeded seems a bit unintuitive to me.

pkg/alibabacloud/api/api.go
```diff
@@ -84,6 +84,27 @@ func (c *Client) GetInstances(ctx context.Context, vpcs ipamTypes.VirtualNetwork
 	return instances, nil
 }
 
+// UpdateInstance fetches the latest ENIs of the given instanceID,
+// updates the local instanceMap and return
+func (c *Client) UpdateInstance(ctx context.Context, vpcs ipamTypes.VirtualNetworkMap, subnets ipamTypes.SubnetMap, instanceID string, instances *ipamTypes.InstanceMap) (*ipamTypes.InstanceMap, error) {
```
tommyp1ckles: This function seems to be very similar to GetInstances, except rather than creating a new map it populates one that's passed in. Could we simplify this somehow?

pkg/alibabacloud/eni/instances.go
```diff
@@ -122,6 +123,55 @@ func (m *InstancesManager) Resync(ctx context.Context) time.Time {
...
+	}
+
+	securityGroups, err := m.api.GetSecurityGroups(ctx)
+	if err != nil {
+		log.WithError(err).Warning("Unable to retrieve ECS security group list")
+		return time.Time{}
+	}
+
+	oldInstances := m.instances.DeepCopy()
+	instances, err := m.api.UpdateInstance(ctx, vpcs, vSwitches, instanceID, oldInstances)
```
tommyp1ckles: Two questions here:

* What is the implication of merging maps, could this result in leaking of old instances that may have been removed via the GetInstances call?

* Similar to my other question - this function looks almost identical to Resync, except for using UpdateInstance instead of GetInstance. I'm wondering if we could reduce the duplication a bit?

pkg/alibabacloud/api/api.go
tommyp1ckles: (imo: neither of these really fit into the Alibaba API interface very well which seems like it should just be an actions interface around the base alibaba API client.  Processing/population maps should maybe just be in a helper function?)

pkg/alibabacloud/eni/instances.go
jaffcheng: Thanks for the comments! I'll try to address them and refactor the code.

> What is the implication of merging maps, could this result in leaking of old instances that may have been removed via the GetInstances call?

Thanks for pointing this out. One issue I can imagine is that after this change, if an ENI on the instance is somehow detached while maintaining the IP pool, it will not be deleted from the map until the next background full resync, I think this could be fixed, will do that in the next amending.  
Other than that, I think ENIs on different nodes would not overlap and the introduced `instancesAPIResyncLock` would avoid data races.

christarazi:
Moving to draft while comments are being addressed. Feel free to mark ready to merge when are ready for reviewers to take another look.

jaffcheng:
@tommyp1ckles  Refactored, please take another look, thank you!

pkg/alibabacloud/eni/instances.go
jaffcheng: I agree but it looks like the `Resync` interface involves many places, I'm not sure if we should change it in this PR.

pkg/alibabacloud/eni/instances.go
```diff
@@ -99,21 +115,43 @@ func (m *InstancesManager) Resync(ctx context.Context) time.Time {
...
-		"numSecurityGroups": len(securityGroups),
-	}).Info("Synchronized ENI information")
+		log.WithFields(logrus.Fields{
+			"numInstances":      instances.NumInstances(),
+			"numVPCs":           len(vpcs),
+			"numVSwitches":      len(vSwitches),
+			"numSecurityGroups": len(securityGroups),
+		}).Info("Synchronized ENI information")
+
+		m.mutex.Lock()
```
tommyp1ckles, line 140: nit: it may be better to defer the unlock here, makes inadvertently introducing lock bugs less likely.

pkg/ipam/node_manager.go
```diff
@@ -218,7 +227,17 @@ func NewNodeManager(instancesAPI AllocationImplementation, k8sAPI CiliumNodeGett
 }
 
 func (n *NodeManager) instancesAPIResync(ctx context.Context) (time.Time, bool) {
+	if n.instancesAPI.SupportInstanceSync() {
+		// Full API resync should block all the instance incremental sync from other nodes
+		n.instancesAPIResyncLock.Lock()
```
tommyp1ckles: I think this in this case this is fine, but this strikes me as a bit risky, theoretically (perhaps unlikely?) there's nothing to say that any future implementation of the instancesAPI couldn't alternate between returning true here in which case you could have a dangling lock.

Or to put this another way, this puts the onus on the implementer of the cloud API implementation to know the semantics of the ipam node manager - otherwise they risk introducing bugs.

pkg/ipam/node_manager.go
```diff
@@ -119,6 +119,14 @@ type AllocationImplementation interface {
...
 	Resync(ctx context.Context) time.Time
 
+	// SupportInstanceSync returns whether the instance incremental sync is
+	// supported. If not, full Resync is called instead.
+	SupportInstanceSync() bool
+
+	// InstanceSync is called to sync the state of the specified instance with
+	// external APIs or systems.
+	InstanceSync(ctx context.Context, instanceID string) time.Time
+
```
tommyp1ckles: Looking at this a bit more, it feels a bit cumbersome for the node manager to have to branch based on what the underlying instance API implementation does.

At the end of the day, the IPAM manager just wants to request a resync of the instance state, having a function that returns whether the implementation supports a feature feels like we're leaking implementation details a bit.  It also makes the node handler code more complex.

I wonder if this optimization couldn't be done more opaquely by the underlying implementations?

jaffcheng:
Thanks for the comments, please take another look

jaffcheng:
After previous force push, a linting error appeared that I don't have locally, looks like l have to rebase.

tommyp1ckles (approved):
Thanks for the changes 🥳 I think this all makes sense to me.

I'm not as familiar with the alibaba implementation, so I'm approving for Azure/general-IPAM

@christarazi do you mind taking a look as well 🙏

tommyp1ckles:
@jaffcheng One more thought, do you have any data from testing this showing the reduction in latency.  If so, might be good to add here/commit message for future reference.

christarazi (approved):
LGTM thanks for the PR!

My only nit is please remove the `Fixes` from the commit msg because https://github.com/cilium/cilium/issues/25073 will not be fully resolved for all cloud-based IPAM modes which suffer from the full sync. AFAIU, this PR only resolves it for Alibaba. You can used `Related` in the commit msg instead.

jaffcheng:
@tommyp1ckles @christarazi Thanks for the comments, I have updated the commit message, please take another look.
Looks like the linting error is from the main branch.

tommyp1ckles:
> @tommyp1ckles @christarazi Thanks for the comments, I have updated the commit message, please take another look. Looks like the linting error is from the main branch.

awesome 🙏

tommyp1ckles:
@gandro you mind taking a look as well?

gandro (approved):
Nice work, the new iteration looks much cleaner! Thanks

jaffcheng:
Thanks again for the informative guidance! If a rebase for retest is needed, please let me know.

gandro:
> Thanks again for the informative guidance! If a rebase for retest is needed, please let me know.

The remaining failures look unrelated (i.e. `too many arguments in call to newGlobalServiceCache`) - but could you do a rebase nontheless so we can re-test?

jaffcheng:
> The remaining failures look unrelated (i.e. `too many arguments in call to newGlobalServiceCache`) - but could you do a rebase nontheless so we can re-test?

Sure, rebased

gandro:
CI-ginko failed with ` 	 The Service "echo-b" is invalid: spec.ports[0].nodePort: Invalid value: 31414: provided port is already allocated` which is a known flake (e.g. https://github.com/cilium/cilium/issues/13071). Let's see if I can restart it.

gandro:
CI is green. This would be ready to merge if it wasn't for the feature freeze

ldelossa:
@jaffcheng - can you rebase this PR on `main` and push up? That will fix the issue where the gateway-api tests are stuck.

jaffcheng:
@ldelossa Sure, rebased

ldelossa:
All required tests pass. Merging.

PR #25624, opened 2023-05-23
cilium, hf, 1.10: test setting prio

borkmann:
the prio can be pushed into Pod's net_cls cgroup: `echo 123 > net_cls.classid`


borkmann:
/test-backport-1.10

PR #25627, opened 2023-05-23
Unify feature probing packages

### LINKED ISSUE #25621: Merge `pkg/probe` and `pkg/datapath/linux/probes`
Currently there are two go packages for feature probing APIs, `pkg/probe` and `pkg/datapath/linux/probes`. Going forward we should migrate any APIs in `pkg/probe` to `pkg/datapath/linux/probes` to only have one `pkg` for feature probing.

`pkg/probe` currently contains a feature probe called `HaveFullLPM`, since cilium now requires 4.19 as a minimum kernel version we might be able to remove that probe entirely.

rgo3:
With this PR we remove the `HaveFullLPM()` feature probe as it is no longer needed since cilium bumped its minimal supported kernel version to 4.19.

We also migrate any leftover APIs from `pkg/probe` to `pkg/datapath/linux/probes` to only have one go `pkg` for feature probing APIs.

Fixes: #25621


pkg/datapath/prefilter/prefilter.go
```diff
@@ -250,13 +243,9 @@ func (p *PreFilter) init() (*PreFilter, error) {
 
 // NewPreFilter returns prefilter handle
 func NewPreFilter() (*PreFilter, error) {
-	haveLPM := probe.HaveFullLPM()
-	if !haveLPM {
-		log.Warning("Kernel too old for full LPM map support. Needs kernel 4.16 or higher. Only enabling /32 and /128 prefixes for prefilter.")
-	}
 	c := preFilterConfig{
```
ti-mo: It looks like `preFilterConfig` can only ever have all its flags set to `true`, so we should probably remove it in a subsequent commit.

pkg/datapath/linux/probes/probes.go
```diff
@@ -504,6 +504,18 @@ func HaveOuterSourceIPSupport() (err error) {
 	return nil
 }
 
+// HaveIPv6Support tests whether kernel can open an IPv6 socket. This will
+// also implicitly auto-load IPv6 kernel module if available and not yet
+// loaded.
+func HaveIPv6Support() error {
```
ti-mo, line 510: Could we add a test that expects this to succeed unconditionally?

ti-mo (changes requested):
Thanks! We can remove moar code!

squeed (approved):

mhofstetter (approved):

ti-mo (approved):

jschwinger233 (approved):

ti-mo:
/test-1.16-4.19

PR #25633, opened 2023-05-24
hive/jobs: fix enqueueing of multiple jobs via variadic func

mhofstetter:
Currently, when registering multiple jobs to the hive job queue at once via the variadic function, only the first job gets enqueued.

This PR fixes this by adding all jobs to the queue of the job group.


dylandreimerink (approved):
Nice catch, thx!

joamaki (approved):

meyskens (approved):
LGTM

Issue #25213, opened 2023-05-01
Garbage Collection for auth map


mhofstetter:
This issue covers work to ensure that we do garbace collect the BPF map `auth` which contains information about authenticated relationships between workloads. 

Auth Map Layout

```
local identity | remote identity | remote node id | auth type | expiration
```

The following events should trigger a cleanup of the related auth map entries

* identity gets deleted -> all entries referencing the deleted identity as local or remote identity should be deleted
* node gets deleted -> all entries referencing the deleted node as remote node should be deleted
* entry expires -> all expired entries should be deleted

mhofstetter:
PR #25754 opened

PR #25634, opened 2023-05-24
auth: auth map cache

mhofstetter:
In preparation for the auth map garbage collection (https://github.com/cilium/cilium/issues/25213), this PR adds a caching layer for the auth map.

**auth: rename authenticator to internal authmap abstraction**

With the replacement of the contrack map with the auth map, the authenticator interface is no longer necessary.

Therefore, this commit reuses and refactores the interface for an internal abstraction of the auth map itself - which will be reused in the cache implementation.

The new interface and its types are moved from the manager to its own file `authmap.go`.

**auth: auth map cache implementation**

This commit introduces a caching layer for the auth map.

The intention of the cache is to be used as replacement for the auth writer itself and acts as an write/read-through cache.

At startup, the cache gets restored with entries from the BPF map.


meyskens (approved):
Thanks for adressing the export comment :)

LGTM for me

PR #24976, opened 2023-04-19
CODEOWNERS: Add sig-foundations

joamaki:
Assign pkg/{hive,statedb,promise,stream} and daemon/cmd/cells.go to sig-foundations.

PR #25643, opened 2023-05-24
hive: add support for map[string]string flags

giorio94:
Reporting the description of the main commit for convenience:

```
Introduce a decoding hook to convert from a string to the corresponding
map[string]string representation when the target configuration entry is
of the corresponding type. It reuses the same logic already adopted
today outside of hive, assuming initially that the string is json
encoded, and falling back to the KV format in case of errors. This
allows to have a config struct contain map[string]string options,
registering the corresponding flag through flags.StringToString.
```

```release-note
hive: add support for map[string]string flags
```

### DIFF
--- a/Documentation/contributing/development/hive.rst
+++ b/Documentation/contributing/development/hive.rst
@@ -403,12 +403,25 @@ returns a cell that "provides" the parsed configuration to the application:
 
     type MyConfig struct {
         MyOption string
+
+        SliceOption []string
+        MapOption map[string]string
     }
 
     func (def MyConfig) Flags(flags *pflag.FlagSet) {
         // Register the "my-option" flag. This matched against the MyOption field
         // by removing any dashes and doing case insensitive comparison.
         flags.String("my-option", def.MyOption, "My config option")
+
+        // Flags are supported for representing complex types such as slices and maps.
+        // * Slices are obtained splitting the input string on commas.
+        // * Maps support different formats based on how they are provided:
+        //   - CLI: key=value format, separated by commas; the flag can be
+        //     repeated multiple times.
+        //   - Environment variable or configuration file: either JSON encoded
+        //     or comma-separated key=value format.
+        flags.StringSlice("slice-option", def.SliceOption, "My slice config option")
+        flags.StringToString("map-option", def.MapOption, "My map config option")
     }
 
     var defaultMyConfig = MyConfig{
--- a/pkg/command/map_string.go
+++ b/pkg/command/map_string.go
@@ -32,10 +32,17 @@ func GetStringMapString(vp *viper.Viper, key string) map[string]string {
 
 // GetStringMapStringE is same as GetStringMapString, but with error
 func GetStringMapStringE(vp *viper.Viper, key string) (map[string]string, error) {
-	data := vp.Get(key)
+	return ToStringMapStringE(vp.Get(key))
+}
+
+// ToStringMapStringE casts an interface to a map[string]string type. The underlying
+// interface type might be a map or string. In the latter case, it is attempted to be
+// json decoded, falling back to the k1=v2,k2=v2 format in case it doesn't look like json.
+func ToStringMapStringE(data interface{}) (map[string]string, error) {
 	if data == nil {
 		return map[string]string{}, nil
 	}
+
 	v, err := cast.ToStringMapStringE(data)
 	if err != nil {
 		var syntaxErr *json.SyntaxError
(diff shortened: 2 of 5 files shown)


pkg/command/map_string.go
```diff
@@ -36,6 +36,18 @@ func GetStringMapStringE(vp *viper.Viper, key string) (map[string]string, error)
 	if data == nil {
```
mhofstetter: duplicated nil-check and returning empty map can be removed because this is already done in the invoked `ToStringMapStringE` function

-> only `return ToStringMapStringE(vp.Get(key))` in body

pkg/command/map_string.go
```diff
@@ -36,6 +36,18 @@ func GetStringMapStringE(vp *viper.Viper, key string) (map[string]string, error)
...
 		return map[string]string{}, nil
 	}
+
+	return ToStringMapStringE(vp.Get(key))
+}
+
+// ToStringMapStringE casts an interface to a map[string]string type. The underlying
```
mhofstetter, line 38: would be great to have this change reflected in the cilium docs -> https://docs.cilium.io/en/latest/contributing/development/hive/#config

especially because of the special encoding (json & kv fallback)

pkg/hive/cell/config.go
```diff
@@ -138,3 +142,18 @@ func (c *config[Cfg]) Info(cont container) (info Info) {
 	})
 	return
 }
+
+// stringToMapHookFunc returns a DecodeHookFunc that converts string
+// to map[string]string supporting both json and KV formats.
+func stringToMapHookFunc() mapstructure.DecodeHookFunc {
+	return func(
```
mhofstetter: nit: IMO unnecessary line-breaks between parameters

mhofstetter (commented):
thanks! just left some small optimizations inline.

due to an upcoming change in the ownership of the hive packages i propose an additional pair of eyes from sig-foundations / @joamaki (https://github.com/cilium/cilium/pull/24976)

pkg/command/map_string.go
giorio94: Sure, that was indeed my intention, but I forgot to delete those lines.

pkg/hive/cell/config.go
giorio94: Removed, I adopted the same style as the other HookFuncs in the upstream library, but I agree that it is easier to read with all parameters inline.

joamaki (approved):
Looks good, nothing to add on top of @mhofstetter's comments.

pkg/command/map_string.go
giorio94: I've added an [additional commit](https://github.com/cilium/cilium/pull/25643/commits/372d3f295c2480819dad2b8d6fc3274e43c2abc6) which extends the example and comments on how parsing is performed. PTAL

mhofstetter (approved):
LGTM - thanks for the changes!

giorio94:
Upon further testing, I noticed that I mistakenly used `viper.Set` rather than `flags.Set` in the newly introduced test. The latter is the one actually executed when setting a flag through the CLI, and it does not support decoding maps from JSON format. I've updated the tests and the documentation accordingly. @mhofstetter PTAL

bimmlerd (approved):
LGTM - just out of curiosity, what do we need this for?

giorio94:
> LGTM - just out of curiosity, what do we need this for?

It will be initially used to register the `kvstore-opt` flag in https://github.com/cilium/cilium/pull/25554. I've extracted these commits in a separated PR since they have a more general scope.

mhofstetter (approved):

Documentation/contributing/development/hive.rst
```diff
@@ -403,12 +403,25 @@ returns a cell that "provides" the parsed configuration to the application:
...
+        SliceOption []string
+        MapOption map[string]string
     }
 
     func (def MyConfig) Flags(flags *pflag.FlagSet) {
         // Register the "my-option" flag. This matched against the MyOption field
         // by removing any dashes and doing case insensitive comparison.
         flags.String("my-option", def.MyOption, "My config option")
+
+        // Flags representing complex types such as slices and maps are supported.
```
zacharysarah: If we can't fully use active voice, I'd like less distance between the subject and its most important qualifiers.
```suggestion
        // Flags are supported for representing complex types such as slices and maps.
```

zacharysarah (approved):
@giorio94 Good work! ✨ One nit, otherwise LGTM. Approving with the understanding that changes are required prior to merge.

giorio94:
Rebased onto main to pick the last changes and allow the Cilium Runtime checks to run successfully.

Issue #23111, opened 2023-01-16
CFP: Reduce Hubble traffic


AwesomePatrol:
## Cilium Feature Proposal
        
Reduce traffic passing through Hubble relay in a typical use case

**Is your feature request related to a problem?**

Yes, partially.

When using `hubble observe -f` on Hubble relay (i.e. trying to get metadata on all traffic in a cluster), you can quickly reach a network throughput limit. Currently, a single flow is approximately 500 bytes in size (depending on the number of labels used), so at around 236k flows/s you saturate a typical gigabit connection. Given that only one node sends data locally and that the information needs to be retransmitted without change, it results in 118k flows/s of theoretical performance limit. In practice message size also adds to CPU consumption. It is usually fully occupied by flow decoding and reencoding before reaching the theoretical limit (around 71k flows/s on `n2d` machine).

**Describe the feature you'd like**

Add an option to reduce the size of transmitted messages and/or optimize the protocol to avoid transmitting the same information multiple times.

**(Optional) Describe your proposed solution**

When writing proof of concept code I focused on Hubble observer <-> Hubble relay communication. Some changes are optional for implementation by Hubble CLI, some may break forward compatibility.

1. Add an option to choose fields in Flow proto that are transmitted. By default `hubble-cli` user sees only: `time`, `verdict`, `{source,destination}.{ID,namespace,pod_name}`, `ip`, `l4.{source_port,destination_port}`, `trace_observation_point`, `summary` (10 unique fields). Message size can be significantly reduced especially when `{source,destination}.labels` is excluded.
2. If [Endpoint](https://github.com/cilium/cilium/tree/master/api/v1/flow#flow-Endpoint) is guaranteed to not change unless `ID` changes, we can transmit `ID`. A map on the client side will be used to rebuild the information.
3. `NodeName` is transmitted with every flow. If it can be assumed that it doesn’t change during the connection lifecycle, we can transmit it once and have the client rebuild this part of a `flow`.
4. `Time` and `NodeName` fields were moved between `flow` and `event`, so for backwards-compatibility they are duplicated. Breaking it will yield another ~6% smaller messages.

Implementation: https://github.com/AwesomePatrol/cilium/tree/reduce-hubble-traffic

Pros:
* Higher throughput
* Lower cpu requirements
```
name              old time/op    new time/op    delta
RelayGetFlows-12    14.1µs ± 7%     8.8µs ± 8%  -37.46%  (p=0.000 n=11+20)

[...]

PR #25647, opened 2023-05-24
Add new experiments to reduce Hubble traffic

AwesomePatrol:
Part of https://github.com/cilium/cilium/issues/23111 (more info there)

```
name \ time/op    base         field_mask   endpoint_and_nodename_dedup  endpoint_and_nodename_dedup_with_field_mask
RelayGetFlows-24  20.0µs ±10%  12.7µs ± 7%                  13.7µs ± 5%                                  11.8µs ± 4%

name \ alloc/op   base         field_mask   endpoint_and_nodename_dedup  endpoint_and_nodename_dedup_with_field_mask
RelayGetFlows-24  13.1kB ± 0%  6.7kB ± 0%                   7.6kB ± 1%                                   6.1kB ± 0%

name \ allocs/op  base         field_mask   endpoint_and_nodename_dedup  endpoint_and_nodename_dedup_with_field_mask
RelayGetFlows-24  192 ± 0%     140 ± 0%                     133 ± 0%                                     121 ± 1%
```

I am not sure about `TransmitNodenameOnce` as it gives the least benefit compared to introduced complexity.
Benchmarks were refactored (they are also tests as it is easier to check the functionality there than in `TestLocalObserverServer_GetFlows`).

```release-note
Reduce Hubble observe traffic by transmitting identity once (experimental)
```


pkg/hubble/observer/local_observer.go
```diff
@@ -286,6 +286,10 @@ func (s *LocalObserverServer) GetFlows(
 		flow = new(flowpb.Flow)
 		mask.alloc(flow.ProtoReflect())
 	}
+	var seen map[uint32]*flowpb.Endpoint
+	if req.Experimental.GetTransmitEndpointOnce() {
+		seen = make(map[uint32]*flowpb.Endpoint)
+	}
```
kaworu, line 292: Couple of issues here:

First, an endpoint ID can be reused at some point, which is not handled by this codepath. To that end we could [subscribe to the `endpointManager`](https://github.com/cilium/cilium/blob/2a966f4f140c47a4d90b3358de7c0ab9f2984546/pkg/endpointmanager/subscribe.go#L15-L18) (specifically `EndpointDeleted` would remove the corresponding entry in `seen`).

Secondly (unrelated to the above code), since endpoint IDs are per-node we should assume that duplicate values are possible. This, in combination with endpoint ID re-use (and potentially flowmasking `node_name`) could make it non-trivial to track from the client's perspective.

cc @joestringer

kaworu (changes requested):
Thanks for the patch @AwesomePatrol!

> I am not sure about `TransmitNodenameOnce` as it gives the least benefit compared to introduced complexity.

Generally speaking, we try our best to avoid introducing logic into Hubble Relay itself. Specifically, mutating either the request or the response should be considered carefully, and in this instance it is not worth the performance gain in my view. cc @rolinh @gandro

See my inline comment about `TransmitEndpointOnce`.

Additionally, I'm requesting changes here because I feel the PR is missing the logic of how clients should "reconstruct full flows" when both (or either) of theses flags are given in the request (a POC patch of https://github.com/cilium/hubble would be a great example). Although I'm hesitant to introduce state tracking into the Hubble server, the server-side complexity seems reasonable. I fear that the client code might be trickier though, hence the request.

AwesomePatrol:
Sounds fair enough. Now that I read the code allocating Endpoint IDs, I understand that there are no guarantees to when it will be reused. Hubble-relay would either need to get this information in GetFlowResponses (for example as a Timestamp when ID was allocated or a notification on changes to endpoints) or it would need to fetch this data locally. In both cases I agree that it unnecessarily complicates Hubble-relay and most benefit is already part of field masking.

Thanks for the response. I will close this PR until I have better idea

AwesomePatrol:
Maybe I was to quick to discard this PR. There two more ideas:
1. Use Identity ID - as it is correlated to labels, does it have better guarantees to avoid collisions and be consistent across the nodes?
2. Add UUIDs to Flows and send only data that changes (needs to keep a cache of active flows on client side which can be a lot, but at least it can avoid changes to Hubble-relay)

kaworu:
> Use Identity ID - as it is correlated to labels, does it have better guarantees to avoid collisions and be consistent across the nodes?

Assuming you're referring to [Cilium security identity](https://docs.cilium.io/en/stable/internals/security-identities/#security-identities), I think we'll run into a different kind of collision as distinct endpoints (e.g. `ep0`, `ep1` colocated on the same node)  can share the same security identity (`secid`) when they have the same labels. Here `secid` is not enough information disambiguate whether the endpoint was `ep0` or `ep1`.

> Add UUIDs to Flows and send only data that changes (needs to keep a cache of active flows on client side which can be a lot, but at least it can avoid changes to Hubble-relay)

This is an interesting idea! Note that [flows already have an `uuid` field](https://github.com/cilium/cilium/blob/0991c6d992a2b6f0ad6ef13f4c1df900b09cbced/api/v1/flow/flow.proto#L16-L17).

We could have a signal that a given flow should be cached to be used as base for following "diff" flows, e.g. `experimental.parent="true"` for the first flow in a TCP session, then "diff" flows would reference their parent (`experimental.parent="$parentUUID"`). Cache GC could be TTL based on the base flow time, maybe also LRU limited on the server side? Also need to think about how multiple clients requesting caching will be handled server-side. Since the Hubble server is part of the Cilium agent, we should be very careful about memory consumption.

Once cache management is solved, then we need a solution to clear fields (e.g. base flow has `traffic_direction=EGRESS` and diff flow has `traffic_direction=TRAFFIC_DIRECTION_UNKNOWN` which is the zero value) but maybe that could be done throuh a flowmask of the updated fileds (`experimental.fieldMask` would have `{paths: ["traffic_direction", …]}`).

Finally, the CPU/memory overhead should be measured, especially server side. I would be less worried about the introduced complexity as I'm expecting this facility to be easier to cover with unit tests than the patch proposed by this PR. Although most of the gain will probably come from caching the source/destination endpoints, it's generic and has huge potential going forward in my opinion.

PR #23936, opened 2023-02-22
Modularization of metrics

dylandreimerink:
This PR restructures the way we work with metrics within Cilium. The goal of this PR is to address a few pain points with the current metrics system. We get three big things out of this work:
* The ability to define metrics in a modular way, close to the location where it is actually going to be used.
* The ability to define richer metadata like label descriptions, known values and their meaning etc.
* The ability to automatically generate metrics references from all defined metrics

To accomplish we did the following things:
* Created custom metric types on top of the existing Prometheus metrics, which are API compatible, but carry additional information and setting used for other features.
* Metrics are now always registered with the repository, but they can be enabled or disabled. Instead of having a centralized switch statement to determine which metrics will be enabled or disabled, the metric definitions now have a field to say if they are enabled by default or not. The main advantage here is that we are able to list all metrics even those who are disabled.
* A new cell was introduced to wire metric definitions into Hive. Metrics are always contained within a struct so this type can be used to inject the references into other cells that need to use them. The cell also makes the individual metrics available so the registry can collect all metrics defined everywhere.
* All metric definitions in `pkg/metrics/metrics.go` are not converted to a single large cell called `LegacyMetrics` and (for now) are written to the same global variables as were used before.
* The registry has been modularized and uses Hive to collect all metrics
* A tool was introduced in `tools/metricdocgen` which can collect all metrics and render the `Documentation/observability/metrics.rst` file with agent metric references generated from the metrics.

Future PRs will focus on removing app-wide dependency on global variables and start moving metric definitions next to the code that uses it, this will likely go in steps as we modularize the rest of Cillium step by step.

PR #25651, opened 2023-05-24
metrics: Metrics initial modularization

dylandreimerink:
This PR is the first step towards full modularization of metrics. It is a smaller version of #23936 to make review easier. This PR makes the registry into a cell, yet still registers this cell as global variable so the global functions stay working. So in essence we are in a sort of hybrid mode so we do not have to modify as much metric call sites.

This version adds the `cell.Metric` cell-type which allows developers to define metrics wherever they want. This type works just like a `cell.Provide`, you give it a constructor which returns a struct containing metrics. The main difference is that both the struct and its fields are provided to the DAG so the registry can collect the individual metrics while normal cells can use the struct as param and have named metrics.

To provide support for disabled-by-default metrics, we added wrapper types around existing prometheus types, defined in `pkg/metrics/metric`. They are a super-set of their normal counterparts so they don't break the existing call sites. Having this config inside the metrics options at creation time eliminates the need for the "is enabled" feedback to the central config. So the metrics are not fully standalone without dependency on the Daemon it init or the Daemon config.

All existing metrics are converted to these new metrics and are made available to Hive via a new `LegacyMetrics` struct. They are also still registered at the old global variables so the call sites do not have to be updated in this PR. Follow-up PRs should incrementally remove usage of these global variables in favor of taking LegacyMetrics as parameter or moving metrics of interest into a dedicated metrics struct closer to the call site.

Follow-up PRs should also enrich the metrics type to allow for self documenting functionality like the original PR and all dynamic registration/unregistration should be refactored to use the registry as parameter or to make the collectors smarter so we can move away from global functions to modify the registry.

### DIFF
--- a/Documentation/cmdref/cilium-agent.md
+++ b/Documentation/cmdref/cilium-agent.md
@@ -243,7 +243,7 @@ cilium-agent [flags]
       --mesh-auth-rotated-identities-queue-size int             The size of the queue for signaling rotated identities. (default 1024)
       --mesh-auth-spiffe-trust-domain string                    The trust domain for the SPIFFE identity. (default "spiffe.cilium")
       --mesh-auth-spire-admin-socket string                     The path for the SPIRE admin agent Unix socket.
-      --metrics strings                                         Metrics that should be enabled or disabled from the default metric list. The list is expected to be separated by a space. (+metric_foo to enable metric_foo , -metric_bar to disable metric_bar)
+      --metrics strings                                         Metrics that should be enabled or disabled from the default metric list. (+metric_foo to enable metric_foo, -metric_bar to disable metric_bar)
       --monitor-aggregation string                              Level of monitor aggregation for traces from the datapath (default "None")
       --monitor-aggregation-flags strings                       TCP flags that trigger monitor reports when monitor aggregation is enabled (default [syn,fin,rst])
       --monitor-aggregation-interval duration                   Monitor report interval when monitor aggregation is enabled (default 5s)
--- a/clustermesh-apiserver/metrics/metrics.go
+++ b/clustermesh-apiserver/metrics/metrics.go
@@ -64,6 +64,9 @@ func (mm *metricsManager) Start(hive.HookContext) error {
 
 	mm.registry.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
 	mm.registry.MustRegister(collectors.NewGoCollector())
+	// Constructing the legacy metrics and register them at the metrics global variable.
+	// This is a hack until we can unify this metrics manager with the metrics.Registry.
+	metrics.NewLegacyMetrics()
 	mm.registry.MustRegister(
 		metrics.KVStoreOperationsDuration,
 		metrics.KVStoreEventsQueueDuration,
--- a/daemon/cmd/cells.go
+++ b/daemon/cmd/cells.go
@@ -22,6 +22,7 @@ import (
 	ipamMetadata "github.com/cilium/cilium/pkg/ipam/metadata"
 	"github.com/cilium/cilium/pkg/k8s"
 	k8sClient "github.com/cilium/cilium/pkg/k8s/client"
+	"github.com/cilium/cilium/pkg/metrics"
 	"github.com/cilium/cilium/pkg/node"
 	nodeManager "github.com/cilium/cilium/pkg/node/manager"
 	"github.com/cilium/cilium/pkg/option"
@@ -60,6 +61,9 @@ var (
 		k8sClient.Cell,
 
 		cni.Cell,
+
+		// Provide the modular metrics registry, metric HTTP server and legacy metrics cell.
+		metrics.Cell,
 
 		// Provide option.Config via hive so cells can depend on the agent config.
 		cell.Provide(func() *option.DaemonConfig { return option.Config }),
--- a/daemon/cmd/daemon_test.go
+++ b/daemon/cmd/daemon_test.go
@@ -11,9 +11,6 @@ import (
 	"testing"
 	"time"
 
-	. "github.com/cilium/checkmate"
-	"github.com/prometheus/client_golang/prometheus"
-
 	"github.com/cilium/cilium/api/v1/models"
 	cnicell "github.com/cilium/cilium/daemon/cmd/cni"
 	fakecni "github.com/cilium/cilium/daemon/cmd/cni/fake"
@@ -44,6 +41,8 @@ import (
 	"github.com/cilium/cilium/pkg/proxy"
 	"github.com/cilium/cilium/pkg/testutils"
 	"github.com/cilium/cilium/pkg/types"
+
+	. "github.com/cilium/checkmate"
 )
 
 type DaemonSuite struct {
@@ -64,9 +63,6 @@ type DaemonSuite struct {
 	OnGetCompilationLock   func() *lock.RWMutex
 	OnSendNotification     func(typ monitorAPI.AgentNotifyMessage) error
 	OnGetCIDRPrefixLengths func() ([]int, []int)
-
-	// Metrics
-	collectors []prometheus.Collector
 }
 
 func setupTestDirectories() {
@@ -138,17 +134,6 @@ func (epSync *dummyEpSyncher) DeleteK8sCiliumEndpointSync(e *endpoint.Endpoint) 
 
 func (ds *DaemonSuite) SetUpSuite(c *C) {
 	testutils.IntegrationCheck(c)
-
-	// Register metrics once before running the suite
-	_, ds.collectors = metrics.CreateConfiguration([]string{"cilium_endpoint_state"})
-	metrics.MustRegister(ds.collectors...)
-}
-
-func (ds *DaemonSuite) TearDownSuite(c *C) {
-	// Unregister the metrics after the suite has finished
-	for _, c := range ds.collectors {
-		metrics.Unregister(c)
-	}
 }
 
 func (ds *DaemonSuite) SetUpTest(c *C) {
--- a/pkg/bpf/bpf_linux.go
+++ b/pkg/bpf/bpf_linux.go
@@ -17,7 +17,6 @@ import (
 
 	"github.com/cilium/cilium/pkg/logging/logfields"
 	"github.com/cilium/cilium/pkg/metrics"
-	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/spanstat"
 )
 
@@ -28,13 +27,13 @@ func createMap(spec *ebpf.MapSpec, opts *ebpf.MapOptions) (*ebpf.Map, error) {
 	}
 
 	var duration *spanstat.SpanStat
-	if option.Config.MetricsConfig.BPFSyscallDurationEnabled {
+	if metrics.BPFSyscallDuration.IsEnabled() {
 		duration = spanstat.Start()
 	}
 
 	m, err := ebpf.NewMapWithOptions(spec, *opts)
 
-	if option.Config.MetricsConfig.BPFSyscallDurationEnabled {
+	if metrics.BPFSyscallDuration.IsEnabled() {
 		metrics.BPFSyscallDuration.WithLabelValues(metricOpCreate, metrics.Error2Outcome(err)).Observe(duration.End(err == nil).Total().Seconds())
 	}
 
--- a/pkg/bpf/map_linux.go
+++ b/pkg/bpf/map_linux.go
@@ -279,7 +279,7 @@ func (m *Map) updatePressureMetric() {
 
 	// Do a lazy check of MetricsConfig as it is not available at map static
 	// initialization.
-	if !option.Config.MetricsConfig.BPFMapPressure {
+	if !metrics.BPFMapPressure {
 		if !m.withValueCache {
 			m.cache = nil
 		}
@@ -566,13 +566,13 @@ func (m *Map) Close() error {
 
 func (m *Map) NextKey(key, nextKeyOut interface{}) error {
 	var duration *spanstat.SpanStat
-	if option.Config.MetricsConfig.BPFSyscallDurationEnabled {
+	if metrics.BPFSyscallDuration.IsEnabled() {
 		duration = spanstat.Start()
 	}
 
 	err := m.m.NextKey(key, nextKeyOut)
 
-	if option.Config.MetricsConfig.BPFSyscallDurationEnabled {
+	if metrics.BPFSyscallDuration.IsEnabled() {
 		metrics.BPFSyscallDuration.WithLabelValues(metricOpGetNextKey, metrics.Error2Outcome(err)).Observe(duration.End(err == nil).Total().Seconds())
 	}
 
@@ -782,14 +782,14 @@ func (m *Map) Lookup(key MapKey) (MapValue, error) {
 	defer m.lock.RUnlock()
 
 	var duration *spanstat.SpanStat
-	if option.Config.MetricsConfig.BPFSyscallDurationEnabled {
+	if metrics.BPFSyscallDuration.IsEnabled() {
 		duration = spanstat.Start()
 	}
 
 	value := key.NewValue()
 	err := m.m.Lookup(key.GetKeyPtr(), value.GetValuePtr())
 
-	if option.Config.MetricsConfig.BPFSyscallDurationEnabled {
+	if metrics.BPFSyscallDuration.IsEnabled() {
 		metrics.BPFSyscallDuration.WithLabelValues(metricOpLookup, metrics.Error2Outcome(err)).Observe(duration.End(err == nil).Total().Seconds())
 	}
 
@@ -846,7 +846,7 @@ func (m *Map) Update(key MapKey, value MapValue) error {
 
 	err = m.m.Update(key.GetKeyPtr(), value.GetValuePtr(), ebpf.UpdateAny)
 
-	if option.Config.MetricsConfig.BPFMapOps {
+	if metrics.BPFMapOps.IsEnabled() {
 		metrics.BPFMapOps.WithLabelValues(m.commonName(), metricOpUpdate, metrics.Error2Outcome(err)).Inc()
 	}
 
@@ -920,13 +920,13 @@ func (m *Map) delete(key MapKey, ignoreMissing bool) (_ bool, err error) {
 	}
 
 	var duration *spanstat.SpanStat
-	if option.Config.MetricsConfig.BPFSyscallDurationEnabled {
+	if metrics.BPFSyscallDuration.IsEnabled() {
 		duration = spanstat.Start()
 	}
 
 	err = m.m.Delete(key.GetKeyPtr())
 
-	if option.Config.MetricsConfig.BPFSyscallDurationEnabled {
+	if metrics.BPFSyscallDuration.IsEnabled() {
 		metrics.BPFSyscallDuration.WithLabelValues(metricOpDelete, metrics.Error2Outcome(err)).Observe(duration.End(err == nil).Total().Seconds())
 	}
 
@@ -939,7 +939,7 @@ func (m *Map) delete(key MapKey, ignoreMissing bool) (_ bool, err error) {
 		return false, nil
 	}
 
-	if option.Config.MetricsConfig.BPFMapOps {
+	if metrics.BPFMapOps.IsEnabled() {
 		// err can be nil or any error other than ebpf.ErrKeyNotExist.
 		metrics.BPFMapOps.WithLabelValues(m.commonName(), metricOpDelete, metrics.Error2Outcome(err)).Inc()
 	}
@@ -1122,7 +1122,7 @@ func (m *Map) resolveErrors(ctx context.Context) error {
 		case Insert:
 			// Call into ebpf-go's Map.Update() directly, don't go through the cache.
 			err := m.m.Update(e.Key.GetKeyPtr(), e.Value.GetValuePtr(), ebpf.UpdateAny)
-			if option.Config.MetricsConfig.BPFMapOps {
+			if metrics.BPFMapOps.IsEnabled() {
 				metrics.BPFMapOps.WithLabelValues(m.commonName(), metricOpUpdate, metrics.Error2Outcome(err)).Inc()
 			}
 			if err == nil {
@@ -1139,7 +1139,7 @@ func (m *Map) resolveErrors(ctx context.Context) error {
 		case Delete:
 			// Holding lock, issue direct delete on map.
 			err := m.m.Delete(e.Key.GetKeyPtr())
-			if option.Config.MetricsConfig.BPFMapOps {
+			if metrics.BPFMapOps.IsEnabled() {
 				metrics.BPFMapOps.WithLabelValues(m.commonName(), metricOpDelete, metrics.Error2Outcome(err)).Inc()
 			}
 			if err == nil || errors.Is(err, ebpf.ErrKeyNotExist) {
--- a/pkg/endpoint/endpoint_test.go
+++ b/pkg/endpoint/endpoint_test.go
@@ -13,7 +13,6 @@ import (
 	"time"
 
 	. "github.com/cilium/checkmate"
-	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/cilium/cilium/pkg/datapath/fake"
 	datapath "github.com/cilium/cilium/pkg/datapath/types"
@@ -56,9 +55,6 @@ type EndpointSuite struct {
 	OnRemoveFromEndpointQueue func(epID uint64)
 	OnGetCompilationLock      func() *lock.RWMutex
 	OnSendNotification        func(msg monitorAPI.AgentNotifyMessage) error
-
-	// Metrics
-	collectors []prometheus.Collector
 }
 
 // suite can be used by testing.T benchmarks or tests as a mock regeneration.Owner
@@ -77,15 +73,12 @@ func (s *EndpointSuite) SetUpSuite(c *C) {
 	}
 
 	// Register metrics once before running the suite
-	_, s.collectors = metrics.CreateConfiguration([]string{"cilium_endpoint_state"})
-	metrics.MustRegister(s.collectors...)
+	metrics.NewLegacyMetrics().EndpointStateCount.SetEnabled(true)
 }
 
 func (s *EndpointSuite) TearDownSuite(c *C) {
 	// Unregister the metrics after the suite has finished
-	for _, c := range s.collectors {
-		metrics.Unregister(c)
-	}
+	metrics.EndpointStateCount.SetEnabled(false)
 }
 
 func (s *EndpointSuite) GetPolicyRepository() *policy.Repository {
--- a/pkg/endpointmanager/manager.go
+++ b/pkg/endpointmanager/manager.go
@@ -11,7 +11,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 
 	"github.com/cilium/cilium/pkg/completion"
@@ -28,6 +27,7 @@ import (
 	"github.com/cilium/cilium/pkg/logging/logfields"
 	"github.com/cilium/cilium/pkg/mcastmanager"
 	"github.com/cilium/cilium/pkg/metrics"
+	"github.com/cilium/cilium/pkg/metrics/metric"
 	"github.com/cilium/cilium/pkg/node"
 	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/policy"
@@ -192,7 +192,7 @@ func (mgr *endpointManager) InitMetrics() {
 		// would result in negative counts.
 		// It must be thread-safe.
 
-		metrics.Endpoint = prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+		metrics.Endpoint = metric.NewGaugeFunc(metric.GaugeOpts{
 			Namespace: metrics.Namespace,
 			Name:      "endpoint",
 			Help:      "Number of endpoints managed by this agent",
--- a/pkg/hive/cell/metric.go
+++ b/pkg/hive/cell/metric.go
@@ -1,0 +1,138 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package cell
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/dig"
+
+	"github.com/cilium/cilium/pkg/hive/internal"
+	pkgmetric "github.com/cilium/cilium/pkg/metrics/metric"
+)
+
+var (
+	withMeta  pkgmetric.WithMetadata
+	collector prometheus.Collector
+)
+
+// Metric constructs a new metric cell.
+//
+// This cell type provides `S` to the hive as returned by `ctor`, it also makes each individual field
+// value available via the `hive-metrics` value group. Infrastructure components such as a registry,
+// inspection tool, or documentation generator can collect all metrics in the hive via this value group.
+//
+// The `ctor` constructor must return a struct or pointer to a struct of type `S`. The returned struct
+// must only contain public fields. All field types should implement the
+// `github.com/cilium/cilium/pkg/metrics/metric.WithMetadata`
+// and `github.com/prometheus/client_golang/prometheus.Collector` interfaces.
+func Metric[S any](ctor func() S) Cell {
+	var nilOut S
+	outTyp := reflect.TypeOf(nilOut)
+	if outTyp.Kind() == reflect.Ptr {
+		outTyp = outTyp.Elem()
+	}
+
+	if outTyp.Kind() != reflect.Struct {
+		panic(fmt.Errorf(
+			"cell.Metric must be invoked with a constructor function that returns a struct or pointer to a struct, "+
+				"a constructor which returns a %s was supplied",
+			outTyp.Kind(),
+		))
+	}
+
+	// Let's be strict for now, could lift this in the future if we ever need to
+	if outTyp.NumField() == 0 {
+		panic(fmt.Errorf(
+			"cell.Metric must be invoked with a constructor function that returns exactly a struct with at least 1 " +
+				"metric, a constructor which returns a struct with zero fields was supplied",
+		))
+	}
+
+	withMetaTyp := reflect.TypeOf(&withMeta).Elem()
+	collectorTyp := reflect.TypeOf(&collector).Elem()
+	for i := 0; i < outTyp.NumField(); i++ {
+		field := outTyp.Field(i)
+		if !field.IsExported() {
+			panic(fmt.Errorf(
+				"The struct returned by the constructor passed to cell.Metric has a private field '%s', which "+
+					"is not allowed. All fields on the returning struct must be exported",
+				field.Name,
+			))
+		}
+
+		if !field.Type.Implements(withMetaTyp) {
+			panic(fmt.Errorf(
+				"The struct returned by the constructor passed to cell.Metric has a field '%s', which is not metric.WithMetadata.",
+				field.Name,
+			))
+		}
+
+		if !field.Type.Implements(collectorTyp) {
+			panic(fmt.Errorf(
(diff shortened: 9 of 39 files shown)


pkg/hive/example/metrics.go
```diff
@@ -0,0 +1,28 @@
...
+// Copyright Authors of Cilium
+
+package main
+
+import (
+	"github.com/cilium/cilium/pkg/hive/cell"
+	"github.com/cilium/cilium/pkg/metrics/metric"
+)
+
+var exampleMetricsCell = cell.Metric(newExampleMetrics)
```
bimmlerd, line 11: I believe the the expected use case of metrics cells to be part of a module; maybe the example could be extended slightly to show that pattern?

pkg/hive/example/metrics.go
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: Apache-2.0
```
bimmlerd, line 1: [commenting here for lack of a better place]
Could we add a section to the "Guide to the hive" docs which explains these modular metrics in this first commit?

pkg/metrics/metric/metric.go
```diff
@@ -0,0 +1,93 @@
...
+// additional ability to remove labels without re-creating the metric.
+type DeletableVec[T any] interface {
+	Vec[T]
+	Delete(labels prometheus.Labels) bool
+	DeleteLabelValues(lvs ...string) bool
+	DeletePartialMatch(labels prometheus.Labels) int
+	Reset()
+}
+
+// Opts are a modified and extended version of the prometheus.Opts
```
bimmlerd, line 156: I can only see extensions (`ConfigName`, `EnabledByDefault` are added). If that's the case; would it not be better to embed a prometheus.Opts?

pkg/metrics/metric/histogram.go
```diff
@@ -0,0 +1,331 @@
...
+		if histVec, ok := cv.ObserverVec.(*prometheus.HistogramVec); ok {
+			histVec.Reset()
+		}
+	}
+
+	cv.metric.SetEnabled(e)
+}
+
+// HistogramOpts are a modified and expanded version of the prometheus.HistogramOpts.
+// https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#HistogramOpts
```
bimmlerd, line 170: Similar question here, why not embed - I only see extensions.

Documentation/cmdref/cilium-agent.md
```diff
@@ -242,7 +242,7 @@ cilium-agent [flags]
       --mesh-auth-rotated-identities-queue-size int             The size of the queue for signaling rotated identities. (default 1024)
       --mesh-auth-spiffe-trust-domain string                    The trust domain for the SPIFFE identity. (default "spiffe.cilium")
       --mesh-auth-spire-admin-socket string                     The path for the SPIRE admin agent Unix socket.
-      --metrics strings                                         Metrics that should be enabled or disabled from the default metric list. The list is expected to be separated by a space. (+metric_foo to enable metric_foo , -metric_bar to disable metric_bar)
+      --metrics strings                                         Metrics that should be enabled or disabled from the default metric list. (+metric_foo to enable metric_foo , -metric_bar to disable metric_bar)
```
bimmlerd: nit: is the space after `metric_foo` intentional? (unchanged from existing, just curious)

daemon/cmd/daemon_test.go
```diff
@@ -43,6 +40,8 @@ import (
 	"github.com/cilium/cilium/pkg/proxy"
 	"github.com/cilium/cilium/pkg/testutils"
 	"github.com/cilium/cilium/pkg/types"
+
+	. "github.com/cilium/checkmate"
```
bimmlerd, line 45: nit: is this a goimports thing?

Documentation/cmdref/cilium-agent.md
```diff
@@ -242,7 +242,7 @@ cilium-agent [flags]
       --mesh-auth-rotated-identities-queue-size int             The size of the queue for signaling rotated identities. (default 1024)
```
bimmlerd, line 243: 2nd commit msg nit: "independency" -> "independently"

daemon/cmd/endpoint_test.go
```diff
@@ -146,6 +147,7 @@ func assertOnMetric(c *C, state string, expected int64) {
 		return obtained == expected
 	}, 10*time.Second)
 	if err != nil {
+		spew.Dump(metrics.EndpointStateCount, state)
```
bimmlerd: nit: leftover from debugging or intentional?

pkg/metrics/registry.go
```diff
@@ -0,0 +1,242 @@
...
+)
+
+var defaultRegistryConfig = RegistryConfig{
+	PrometheusServeAddr: ":9962",
+}
+
+type RegistryConfig struct {
+	// PrometheusServeAddr IP:Port on which to serve prometheus metrics (pass ":Port" to bind on all interfaces, "" is off)
+	PrometheusServeAddr string
+	//
```
bimmlerd: nit: empty comment

pkg/metrics/registry.go
```diff
@@ -0,0 +1,242 @@
...
+			BPFMapPressure = false
+			continue
+		}
+
+		metric := metrics[metricFlag[1:]]
+		if metric == nil {
+			continue
+		}
+
+		switch metricFlag[0] {
```
bimmlerd, line 159: I think we should add a default case with a warning to this switch, to prevent silently losing some metrics - I _think_ the following would lose the middle metric disable (due to the space before it), as well as silently ignoring the final metric (because there's no plus/minus):

`--metrics "+first_metric, -middle_metric,final_metric"`

I traced `viper.StringSlice` to `encoding/csv` [which doesn't trim whitespace](https://go.dev/play/p/fok-8A8Sdg9).
Given this has been the case in the existing code, probably error/fatal would be breaking and can't be done.

pkg/metrics/registry.go
```diff
@@ -0,0 +1,242 @@
...
+// models.Metrics structure.If metrics cannot be retrieved, returns an error
+func (r *Registry) DumpMetrics() ([]*models.Metric, error) {
+	result := []*models.Metric{}
+	currentMetrics, err := r.inner.Gather()
+	if err != nil {
+		return result, err
+	}
+
+	for _, val := range currentMetrics {
+
```
bimmlerd: nit: spurious newline

pkg/endpoint/endpoint_test.go
```diff
@@ -77,15 +73,12 @@ func (s *EndpointSuite) SetUpSuite(c *C) {
...
-	metrics.MustRegister(s.collectors...)
+	metrics.NewLegacyMetrics().EndpointStateCount.SetEnabled(true)
 }
 
 func (s *EndpointSuite) TearDownSuite(c *C) {
 	// Unregister the metrics after the suite has finished
-	for _, c := range s.collectors {
-		metrics.Unregister(c)
-	}
+	metrics.EndpointStateCount = metrics.NoOpGaugeVec
```
bimmlerd: How does this differ from `SetEnabled(false)`? Just wondering about the asymmetry with above

pkg/endpointmanager/manager.go
```diff
@@ -192,10 +192,11 @@ func (mgr *endpointManager) InitMetrics() {
...
 
-		metrics.Endpoint = prometheus.NewGaugeFunc(prometheus.GaugeOpts{
-			Namespace: metrics.Namespace,
-			Name:      "endpoint",
-			Help:      "Number of endpoints managed by this agent",
+		metrics.Endpoint = metric.NewGaugeFunc(metric.GaugeOpts{
+			Namespace:        metrics.Namespace,
+			Name:             "endpoint",
+			Help:             "Number of endpoints managed by this agent",
+			EnabledByDefault: true,
```
bimmlerd: Should we, by Go principles of making the empty value useful, instead be using `Disabled`, defaulting to `true`?

pkg/hive/example/main.go
```diff
@@ -15,10 +15,11 @@ import (
...
 	Hive = hive.New(
-		serverCell,        // An HTTP server, depends on HTTPHandler's
-		eventsCell,        // Example event source (ExampleEvents)
-		helloHandlerCell,  // Handler for /hello
-		eventsHandlerCell, // Handler for /events
+		serverCell,         // An HTTP server, depends on HTTPHandler's
+		eventsCell,         // Example event source (ExampleEvents)
+		helloHandlerCell,   // Handler for /hello
+		eventsHandlerCell,  // Handler for /events
+		exampleMetricsCell, // Register the metrics
```
bimmlerd, line 22: I think this topically belongs to the first commit

pkg/metrics/registry.go
```diff
@@ -0,0 +1,242 @@
...
+type RegistryConfig struct {
+	// PrometheusServeAddr IP:Port on which to serve prometheus metrics (pass ":Port" to bind on all interfaces, "" is off)
+	PrometheusServeAddr string
+	//
+	Metrics []string
+}
+
+func (rc RegistryConfig) Flags(flags *pflag.FlagSet) {
+	flags.String("prometheus-serve-addr", rc.PrometheusServeAddr, "IP:Port on which to serve prometheus metrics (pass \":Port\" to bind on all interfaces, \"\" is off)")
+	flags.StringSlice("metrics", rc.Metrics, "Metrics that should be enabled or disabled from the default metric list. (+metric_foo to enable metric_foo , -metric_bar to disable metric_bar)")
```
bimmlerd: github doesn't let me do the suggestion thing, but the spurious space comes from here and can go :)

bimmlerd (commented):
Thanks a bunch for making this easier to review! Have a couple of comments, but nothing major, awesome work!

pkg/hive/example/metrics.go
dylandreimerink: > I believe the the expected use case of metrics cells to be part of a module

That is not necessary... what gave you the impression?

pkg/hive/example/metrics.go
dylandreimerink: Ah, I always forget the docs. Will do!

pkg/metrics/metric/metric.go
dylandreimerink: There are two reasons for not doing so. First is that I expect to change the types of `Subsystem` in a later PR which will make embedding impossible. Secondly, making literals with embedded objects is a pain:

```
opt := NoEmbeds{
  Name: "",
  Namespace: "",
  EnabledByDefault: true,
}

opt2 := WithEmbeds{
  opts: prometheus.Opts{
    Name: "",
    Namespace: "",
  },
  EnableByDefault: true,
}
```

pkg/metrics/metric/histogram.go
dylandreimerink: See answer above

Documentation/cmdref/cilium-agent.md
dylandreimerink: These files are auto generated, no idea why its different.

daemon/cmd/daemon_test.go
dylandreimerink: Yes it is, for some reason goimports wants it below once the prometheus is gone

pkg/endpoint/endpoint_test.go
dylandreimerink: Originally, the code would create just the `EndpointStateCount` metric, and register it at the prometheus registry, initializing its global variable in the process. Then they unregister the metric from the registry, in case a second tests want to register the same metric, which would not be allowed.

But in this whole exchange, the registry is never actually used, just the metric. So what I do instead is only create the metric, don't bother with the registry, and at the end reset the global variable back to the value before starting this test.

pkg/endpointmanager/manager.go
dylandreimerink: That is a good question. I made it like this because the original pattern was disabled by default unless specifically enabled. I can go either way.

pkg/hive/example/main.go
dylandreimerink: It was supposed to, must have made a mistake somewhere.

pkg/endpoint/endpoint_test.go
dylandreimerink: I think `SetEnabled(true)` would have also worked here btw, I can change that.

pkg/hive/example/metrics.go
bimmlerd: I guess I expect metrics to be associated with cells, typically. Not as a requirement, just as the usual way of doing things - if that's not the case, feel free to resolve this.

pkg/metrics/metric/metric.go
bimmlerd: Ack re the first reason. The reason why I asked is because we'd get future changes in prometheus.Opts for free if we embed; as is the maintenance is potentially increased (?).

pkg/endpoint/endpoint_test.go
bimmlerd: Either way works, my OCD just likes symmetry :sweat_smile:

pkg/endpointmanager/manager.go
bimmlerd: [I mean to write "defaulting to `false`" in the first comment]

I don't have a strong opinion. If it's not a large change, I think it's a little nicer.

bimmlerd (approved):
LGTM with the understanding that docs will be added. Cool stuff!

pkg/endpointmanager/manager.go
dylandreimerink: Agreed, but if the field is "Disabled" and you don't specify it, then the metric is enabled by default. Which is the opposite of what it was before. I personally like your suggestion better since we have more metrics enabled by default anyways. But perhaps someone from @cilium/metrics should way in

chancez: In practice, most metrics should be enabled by default, so I would lean towards `Disabled`, as that would promote metrics being enabled by default. It's okay if some metrics are just `Disabled: true` by default, especially if they're for specific use-cases or have high cardinality.

clustermesh-apiserver/metrics/metrics.go
```diff
@@ -36,7 +35,7 @@ type MetricsConfig struct {
 }
 
 func (def MetricsConfig) Flags(flags *pflag.FlagSet) {
-	flags.String(option.PrometheusServeAddr, def.PrometheusServeAddr, "Address to serve Prometheus metrics")
+	flags.String("prometheus-serve-addr", def.PrometheusServeAddr, "Address to serve Prometheus metrics")
```
nathanjsweet: Can we keep the option string a constant?

pkg/bpf/bpf_linux.go
```diff
@@ -58,7 +57,7 @@ type bpfAttrMapOpElem struct {
 // UpdateElementFromPointers updates the map in fd with the given value in the given key.
 func UpdateElementFromPointers(fd int, mapName string, structPtr unsafe.Pointer, sizeOfStruct uintptr) error {
 	var duration *spanstat.SpanStat
-	if option.Config.MetricsConfig.BPFSyscallDurationEnabled {
+	if metrics.BPFSyscallDuration.IsEnabled() {
```
tommyp1ckles: q: Do we even need to check here, if its a no-op if its not enabled?

pkg/hive/cell/metric.go
```diff
@@ -0,0 +1,139 @@
...
+	// Lets be strict for now, could lift this in the future if we ever need to
+	if outTyp.NumField() == 0 {
+		panic(fmt.Sprintf(
+			"cell.Metric must be invoked with a constructor function that returns exactly a struct with at least 1 "+
+				"metric, %s supplied a constructor which returns a struct with zero fields",
+			caller,
+		))
+	}
+
+	var withMeta pkgmetric.WithMetadata
```
tommyp1ckles: nit: could we do this as a package scoped variable once instead?

pkg/hive/cell/metric.go
```diff
@@ -0,0 +1,139 @@
...
+	withMetaTyp := reflect.TypeOf(&withMeta).Elem()
+
+	var collector prometheus.Collector
+	collectorTyp := reflect.TypeOf(&collector).Elem()
+	for i := 0; i < outTyp.NumField(); i++ {
+		field := outTyp.Field(i)
+		if !field.IsExported() {
```
tommyp1ckles, line 59: Why do we need this constraint on metrics, I don't think other cells enforce this?

pkg/hive/cell/metric.go
```diff
@@ -0,0 +1,139 @@
...
+	"runtime"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/dig"
+
+	"github.com/cilium/cilium/pkg/hive/internal"
+	pkgmetric "github.com/cilium/cilium/pkg/metrics/metric"
+)
+
+func Metric[S any](ctor func() S) Cell {
```
tommyp1ckles, line 32: Given all the runtime constraints on the passed constructor, might be worth documenting this in the function comment.

pkg/bpf/bpf_linux.go
dylandreimerink: We technically don't need the check, if not enabled functions like `Inc`, `Add`, `Observe` are no-op. This was the case before as well so I assume this is an optimization to avoid the cycles used to collect the measurement if we are not going to use it anyway.  No clue if that is warranted or not.

pkg/hive/cell/metric.go
dylandreimerink: If a field containing a metric isn't exported, we can't make it separately available in Hive. That would result in the odd behavior where the package defining the metric struct can still access, modify and use the metric but it will never be exported. That would happen silently, so by enforcing this I hope to avoid users making this hard to find bug and easy to overlook type of bug.

pkg/hive/cell/metric.go
```diff
@@ -0,0 +1,152 @@
...
+package cell
+
+import (
+	"fmt"
+	"reflect"
+	"runtime"
+
+	"go.uber.org/dig"
+
+	"github.com/prometheus/client_golang/prometheus"
```
giorio94: Very minor nit. These two imports should be likely merged in a single block.

pkg/hive/cell/metric.go
```diff
@@ -0,0 +1,152 @@
...
+	var nilOut S
+	outTyp := reflect.TypeOf(nilOut)
+	if outTyp.Kind() == reflect.Ptr {
+		outTyp = outTyp.Elem()
+	}
+
+	if outTyp.Kind() != reflect.Struct {
+		panic(fmt.Sprintf(
+			"cell.Metric must be invoked with a constructor function that returns a struct or pointer to a struct, "+
+				"%s supplied a constructor which returns a %d",
```
giorio94: Nit. I guess `%d` should be `%s`.

pkg/metrics/metric/metric.go
```diff
@@ -0,0 +1,93 @@
...
+	// server, or by one specific metric (e.g. a build_info or a
+	// machine_role metric). See also
+	// https://prometheus.io/docs/instrumenting/writing_exporters/#target-labels-not-static-scraped-labels
+	ConstLabels prometheus.Labels
+
+	// The name used to enable/disable this metric via the config/flags
+	ConfigName string
+
+	// If true, the metric has to be explicitly enabled via config or flags
+	Disabled bool
```
giorio94, line 190: Nit. The commit message still mentions `EnabledByDefault` rather than `Disabled`.

pkg/metrics/metric/metric.go
```diff
@@ -0,0 +1,93 @@
...
+// our version of a prometheus.CounterVec.
+type Vec[T any] interface {
+	prometheus.Collector
+	WithMetadata
+
+	CurryWith(labels prometheus.Labels) (Vec[T], error)
+	GetMetricWith(labels prometheus.Labels) (T, error)
+	GetMetricWithLabelValues(lvs ...string) (T, error)
+	With(labels prometheus.Labels) T
+	WithLabelValues(lvs ...string) T
```
giorio94: I wonder whether it could make sense to propagate the comment of these methods, for reference. Same for the ones below.

pkg/metrics/metric/counter.go
```diff
@@ -0,0 +1,179 @@
...
+			opts:    Opts(opts),
+		},
+	}
+}
+
+type Counter interface {
+	prometheus.Counter
+	WithMetadata
+
+	Get() float64
```
giorio94, line 25: Out of curiosity, is `Get` provided for test purposes only, or is it there another reason? In that case, is it there any reason to prefer it over the `"prometheus/testutil"` package?

pkg/metrics/metric/counter.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (c *counter) Collect(metricChan chan<- prometheus.Metric) {
+	if c.enabled {
+		c.Counter.Collect(metricChan)
+	}
+}
+
+func (c *counter) Opts() Opts {
+	return c.opts
+}
```
giorio94: Nit. It might be moved to the `metric` common implementation.

pkg/metrics/metric/gauge.go
```diff
@@ -0,0 +1,237 @@
...
+		gv.Reset()
+	}
+
+	gv.metric.SetEnabled(e)
+}
+
+type GaugeFunc interface {
+	prometheus.GaugeFunc
+	WithMetadata
+}
```
giorio94, line 194: `Collect()` doesn't seem to be aliased for this metric type.

pkg/metrics/metric/gauge.go
```diff
@@ -0,0 +1,237 @@
...
+
+type gauge struct {
+	prometheus.Gauge
+	metric
+}
+
+func (g *gauge) Describe(descChan chan<- *prometheus.Desc) {
+	g.Gauge.Describe(descChan)
+}
```
giorio94: Nit. `Describe()` is not strictly required to be aliased given that `prometheus.Gauge` is embedded (same for the other metric types).

pkg/hive/example/hello_handler.go
```diff
@@ -16,11 +16,12 @@ var helloHandlerCell = cell.Module(
 	cell.Provide(newHelloHandler),
 )
 
-func newHelloHandler() HTTPHandlerOut {
+func newHelloHandler(ExampleMetrics exampleMetrics) HTTPHandlerOut {
```
giorio94, line 19: I guess these changes should be part of the previous commit.

pkg/metrics/middleware_test.go
```diff
@@ -35,8 +36,10 @@ func (s *MetricsSuite) TestAPIEventsTSHelperMiddleware(c *C) {
 	} {
 		req, err := http.NewRequest(http.MethodGet, test.url, nil)
 		c.Assert(err, Equals, nil)
-		gauge := prometheus.NewGaugeVec(prometheus.GaugeOpts{}, []string{LabelEventSource, LabelScope, LabelAction})
-		hist := prometheus.NewHistogramVec(prometheus.HistogramOpts{Name: "test_api_hist"}, []string{LabelEventSource, LabelScope, LabelAction})
+		gauge := metric.NewGaugeVec(metric.GaugeOpts{}, []string{LabelEventSource, LabelScope, LabelAction})
+		gauge.SetEnabled(true)
```
giorio94: Nit `gauge.SetEnabled` is superfluous as enabled by default.

pkg/metrics/metric/metric.go
```diff
@@ -8,10 +8,8 @@ import (
 )
 
 // WithMetadata is the interface implemented by any metric defined in this package. These typically embed existing
-// prometheus metric types and add additional metadata to the definitions which can be used to generate proper metric
-// documentation from de definitions. In addition, these metrics have the concept of being enabled or disabled which
-// is used in place of conditional registration so all metric types can always be registered for the purposes of
-// documenting them.
+// prometheus metric types and add additional metadata. In addition, these metrics have the concept of being enabled
+// or disabled which is used in place of conditional registration so all metric types can always be registered.
```
giorio94, line 12: Nit. I guess that this change should belong to the previous commit.

pkg/metrics/registry.go
```diff
@@ -0,0 +1,249 @@
...
+			r.params.Logger.Warning(
+				"--metrics flag contains value which does not start with + or -, '%s', ignoring",
+				metricFlag,
+			)
+		}
+	}
+
+	for _, m := range metrics {
+		if c, ok := m.(prometheus.Collector); ok {
+			r.MustRegister(c)
```
giorio94, line 174: Previously, the `Describe()` function of NoOp metrics was a no-op, while AFAICT it now propagates the description regardless of whether the metric is enabled or not. I assume that this makes prometheus report the metric, even though with zero value. Is this change intentional?

clustermesh-apiserver/metrics/metrics.go
giorio94: clustermesh-apiserver metrics are no longer exported, because they are no longer initialized: https://github.com/cilium/cilium/blob/a7a483ac3782574e56d63e7adbd5d6fb61686743/clustermesh-apiserver/metrics/metrics.go#L67-L73 I think it should be enough to make the metrics cell depend on the `LegacyMetrics` struct.

giorio94 (commented):
Overall this PR looks awesome to me! I've left only a bunch of mostly minor comments and nits inline.

pkg/hive/cell/metric.go
dylandreimerink: I really need to look into the goimports rules, I am not making this shit up myself, just the result of running ` goimports -w -local github.com/cilium/cilium <file>` :man_shrugging:

pkg/metrics/metric/counter.go
dylandreimerink: Its meant to replace `metrics.GetCounterValue` and `metrics.GetGaugeValue` in future PRs. These are used both in tests and the metricsmap which is a bit of an oddball. 

I never knew `prometheus/testutil` existed, just thought I would replace the above with a nice methods since we are wrapping anyway. I have no personal preference.

pkg/hive/cell/metric.go
giorio94: IIRC it does preserve empty lines, which is kinda weird. [gci](https://github.com/daixiang0/gci) should be stricter from that point of view.

pkg/metrics/metric/counter.go
giorio94: I don't have a strong preference either. My only concern is that if being part of the interface they might encourage a bit more incorrect usages outside of testing.

pkg/hive/cell/metric.go
dylandreimerink: Anyway, its fixed now

pkg/metrics/metric/metric.go
dylandreimerink: I added the comments to the interface and modified it to fit with the generic interfaces. I think this will end up neater in code completion and generated docs.

pkg/metrics/registry.go
dylandreimerink: Did a dive into the code to check how this works and a test to confirm. The `Describe()` is only used to check the options of the metric at registration time (and perhaps does some additional runtime checks for consistency). But if the `Collect()` does not return any metrics, no entries will appear on `/metrics` not even zero values. So this does what we expect it to do. Plus it adds some additional collision detection even for Disabled by default metrics.

clustermesh-apiserver/metrics/metrics.go
dylandreimerink: Making `LegacyMetrics` a dependency requires that `metrics.Cell` is included in the clustermesh-apiserver hive, but that would cause a collision since they both register this same flag.

For now I solved it by just calling the constructor without hive. In the future, once these KVStore metrics have their own metric cell we can make the clustermesh-apiserver "just" re-use the `metrics.Registry` without the legacy metrics but include the KVStore metric cell in the hive. That way we have 1 copy of the registry logic but different hives include different metrics. The same goes for the operator metrics

tommyp1ckles (approved):

giorio94 (approved):

daemon/cmd/daemon_test.go
```diff
@@ -138,18 +134,9 @@ func (epSync *dummyEpSyncher) DeleteK8sCiliumEndpointSync(e *endpoint.Endpoint)
...
-	metrics.MustRegister(ds.collectors...)
 }
 
-func (ds *DaemonSuite) TearDownSuite(c *C) {
-	// Unregister the metrics after the suite has finished
-	for _, c := range ds.collectors {
-		metrics.Unregister(c)
-	}
-}
+func (ds *DaemonSuite) TearDownSuite(c *C) {}
```
lmb: Nit: why not just get rid of this?

pkg/bpf/map_linux.go
```diff
@@ -279,7 +279,7 @@ func (m *Map) updatePressureMetric() {
 
 	// Do a lazy check of MetricsConfig as it is not available at map static
 	// initialization.
-	if !option.Config.MetricsConfig.BPFMapPressure {
+	if !metrics.BPFMapPressure {
```
lmb, line 282: Why is this not `.IsEnabled()`?

pkg/hive/cell/metric.go
```diff
@@ -0,0 +1,151 @@
...
+// inspection tool, or documentation generator can collect all metrics in the hive via this value group.
+//
+// The `ctor` constructor must return a struct or pointer to a struct of type `S`. The returned struct
+// must only contain public fields. All field types should implement the
+// `github.com/cilium/cilium/pkg/metrics/metric.WithMetadata`
+// and `github.com/prometheus/client_golang/prometheus.Collector` interfaces.
+func Metric[S any](ctor func() S) Cell {
+	caller := "caller"
+	if _, file, no, ok := runtime.Caller(1); ok {
+		caller = fmt.Sprintf("%s:%d", file, no)
```
lmb: Why manually annotate with the caller if the function panics anyways?

pkg/hive/cell/metric.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	var nilOut S
+	outTyp := reflect.TypeOf(nilOut)
+	if outTyp.Kind() == reflect.Ptr {
+		outTyp = outTyp.Elem()
+	}
+
+	if outTyp.Kind() != reflect.Struct {
```
lmb, line 39: Meta comment that isn't specific to this PR, more of a discussion. Any time we design an API where the constraints can only be enforced at runtime we lose developer productivity / velocity.

pkg/hive/cell/metric.go
```diff
@@ -0,0 +1,151 @@
...
+
+	Metrics []pkgmetric.WithMetadata `group:"hive-metrics,flatten"`
+}
+
+func (m *metric[S]) provideMetrics(metricSet S) metricOut {
+	var metrics []pkgmetric.WithMetadata
+
+	value := reflect.ValueOf(metricSet)
+	typ := value.Type()
+	if typ.Kind() == reflect.Pointer {
```
lmb, line 102: You can use https://go.dev/pkg/reflect/?m=old#Indirect to avoid this

pkg/hive/cell/metric.go
```diff
@@ -0,0 +1,151 @@
...
+	collector prometheus.Collector
+)
+
+// Metric constructs a new metric cell.
+//
+// This cell type provides `S` to the hive as returned by `ctor`, it also makes each individual field
+// value available via the `hive-metrics` value group. Infrastructure components such as a registry,
+// inspection tool, or documentation generator can collect all metrics in the hive via this value group.
+//
+// The `ctor` constructor must return a struct or pointer to a struct of type `S`. The returned struct
```
lmb, line 28: Does it make sense to return non-pointer structs at all? I'm guessing that metrics structs will be on the larger side.

daemon/cmd/daemon_test.go
dylandreimerink: Honestly, I am not that familiar with the whole suite tests, I had assumed you can't just remove functions without consequences, will try that :smile:

pkg/bpf/map_linux.go
dylandreimerink: Well because the `.IsEnabled()` is a feature of the new `pkg/metrics/metric` types, and the map pressure metrics are implemented with their own custom type `metrics.GaugeWithThreshold`. To keep this PR manageable, I decided to not refactor it, will do that in a followup PR.

lmb (approved):

daemon/cmd/cells.go
```diff
@@ -60,6 +61,8 @@ var (
 
 		cni.Cell,
```
bimmlerd, line 64: nit: comment to explain the purpose of this cell?

bimmlerd (approved):
contributing part lgtm too

pkg/hive/cell/metric.go
dylandreimerink: Yea, that makes sense

pkg/hive/cell/metric.go
dylandreimerink: I tries, but that only works on a `reflect.Value` not on a `reflect.Type`.

pkg/hive/cell/metric.go
dylandreimerink: It depends. The fields of these structs will all be pointers, so its not that big of a deal. Besides, they are mostly transferred during construction, after that they will likely be part of the field of a cell somewhere. I don't think we should restrict it any further if we can help it.

pkg/hive/cell/metric.go
dylandreimerink: Very good point. I try to use types wherever I can to make it clear to users what can and can't be used at compile time. But at this moment we just do not have that capability to the depth I would like.

In this case, we could also have gone for a `cell.Config` like model where the generic type must implement a given interface and have the interface return a slice of metrics. The very first iteration worked like this, but it requires the user to write a lot of extra boilerplate which I disliked.

At the moment I feel like I have to pick between two evils, this being the lesser from my point of view. But I am open to discussing this further, perhaps on slack somewhere.

pkg/hive/cell/metric.go
```diff
@@ -0,0 +1,140 @@
...
+
+	if outTyp.Kind() != reflect.Struct {
+		panic(fmt.Errorf(
+			"cell.Metric must be invoked with a constructor function that returns a struct or pointer to a struct, "+
+				"a constructor which returns a %s was supplied",
+			outTyp.Kind(),
+		))
+	}
+
+	// Lets be strict for now, could lift this in the future if we ever need to
```
qmonnet:
```suggestion
	// Let's be strict for now, could lift this in the future if we ever need to
```

pkg/metrics/metric/counter.go
```diff
@@ -0,0 +1,167 @@
...
+// Inc increments the counter by 1. Use Add to increment it by arbitrary
+// non-negative values.
+func (c *counter) Inc() {
+	if c.enabled {
+		c.Counter.Inc()
+	}
+}
+
+// Add adds the given value to the counter. It panics if the value is <
+// 0.
```
qmonnet:
```suggestion
// Add adds the given value to the counter. It panics if the value is < 0.
```

pkg/hive/cell/metric.go
```diff
@@ -0,0 +1,140 @@
...
+				"The struct returned by the constructor passed to cell.Metric has a private field '%s', which "+
+					"is not allowed. All fields on the returning struct must be exported",
+				field.Name,
+			))
+		}
+
+		if !field.Type.Implements(withMetaTyp) {
+			panic(fmt.Errorf(
+				"The struct returned by the constructor passed to cell.Metric has a field '%s', which is not "+
+					"metric.WithMetadata.",
```
qmonnet: Let's not break the string? Same comment for the block below.

qmonnet (approved):
Doc changes look good, thanks!

The few nits below are not blocking.

nathanjsweet (approved):
LGTM for cli

dylandreimerink:
What is with?

> [nathanjsweet](https://github.com/nathanjsweet) requested a review from https://github.com/orgs/cilium/teams/sig-policy

bimmlerd:
I think in the last rebase you included one commit more than intentional, hence the tophat requirement.

dylandreimerink:
> I think in the last rebase you included one commit more than intended, hence the tophat requirement.

I excellently did a `git rebase upstream/master` instread of `git rebase upstream/main` which included that extra commit, thx for the heads up

PR #25653, opened 2023-05-24
backporting: Fix pattern to handle commit subjects that begin with a space

gentoo-root:
Commit da04e683fa (" bpf/nat: introduce snat_v*_handle_mapping() nat and rev_nat") has a subject that starts with a space, which results in this warning when trying to backport it:

     |  Warning: No commit correlation found!    via b968e9db2a58aaa9ad5e68a470d52bc2357a4f15 ("bpf/nat: introduce snat_v*_handle_mapping() nat and rev_nat")

Fix parsing so that it doesn't trim leading and trailing spaces from the commit subject.

### DIFF
--- a/contrib/backporting/check-stable
+++ b/contrib/backporting/check-stable
@@ -151,10 +151,9 @@ generate_commit_list_for_pr () {
   echo "                 ----------                          -------------------"
   echo "     v (start)"
   while read entry; do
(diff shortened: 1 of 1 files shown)


bimmlerd (approved):
gotta love the shell - thanks for taking a look!

PR #25325, opened 2023-05-08
Updates k8sTest pkg to use netip.Addr

danehans:
Updates the k8sTest pkg to use `netip.Addr()` instead of `net.IP`.

Related: #24246

PR #25659, opened 2023-05-24
Skip ConformanceGatewayAPI and ConformanceIngress test when not needed

christarazi:
- .github/workflows: Skip ConformanceGatewayAPI test when not needed
- .github/workflows: Skip ConformanceIngress test when not needed

### DIFF
--- a/.github/workflows/conformance-gateway-api.yaml
+++ b/.github/workflows/conformance-gateway-api.yaml
@@ -14,7 +14,15 @@ on:
       - 'Documentation/**'
       - 'test/**'
 
-permissions: read-all
+# By specifying the access of one of the scopes, all of those that are not
+# specified are set to 'none'.
+permissions:
+  # To be able to access the repository with actions/checkout
+  contents: read
+  # To allow retrieving information from the PR API
+  pull-requests: read
+  # So that Sibz/github-status-action can write into the status API
+  statuses: write
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.after }}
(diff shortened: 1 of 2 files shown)


.github/workflows/conformance-gateway-api.yaml
```diff
@@ -31,8 +39,92 @@ env:
...
+        sha: ${{ steps.vars.outputs.sha }}
+        context: ${{ github.workflow }}
+        description: GatewayAPI connectivity test in progress...
+        state: pending
+        target_url: ${{ env.check_url }}
+
+  skip-test-run:
+    # If the modified files are not relevant for this test then we can skip
+    # this test and mark it as successful.
+    if: github.event.comment.body == '/test' && needs.check_changes.outputs.tested == 'false'
```
christarazi: Hm, maybe this if condition is wrong because this workflow is not triggered on `/test`?

christarazi: :heavy_check_mark:

.github/workflows/conformance-gateway-api.yaml
```diff
@@ -31,8 +39,92 @@ env:
   timeout: 5m
 
 jobs:
+  check_changes:
+    name: Deduce required tests from code changes
+    if: |
+      github.event_name == 'schedule' ||
```
brlbil, line 45: `github.event_name == 'schedule'` is checked here but neither workflows are triggered with a schedule.
There are two triggers for the workflows `pull_request` and `push`, both of them already have path filters defined.

The question is what is the expected behavior?

brlbil (changes requested):

.github/workflows/conformance-gateway-api.yaml
christarazi: I copy-pasted from `.github/workflows/conformance-clustermesh.yaml`. I'm not skilled at GH Actions so it's possible I'm doing something completely wrong. 

The desired behavior is for ConformanceGatewayAPI and ConformanceIngress to be skipped if there are changes only to `test/` directory. However, they get stuck as required jobs as seen in https://github.com/cilium/cilium/pull/25325.

brlbil: I think the referenced issue is about not properly reporting the status. I will discuss this with a couple of stakeholders and we can move forward with a better approach, either changing the triggers for the workflows or handling status reporting properly, and will follow up the issue to a resolution. But I think in this form this PR is not very useful. if no objection let's close this one.

PR #25660, opened 2023-05-24
bgpv1: graceful restart capability

### LINKED ISSUE #24256: CFP:  support bgp peer graceful restart feature in control-plane
## Cilium Feature Proposal
        
env：
k8s version: 1.24.11
cilium version: 1.13.0
kernel: 5.10 

install with --enable-bgp-control-plane=true

Yes

If so, please describe the problem

the bgp peer will reconnection when cilium-agent restart or rebuild，and this will make the ping down for about 10s+.this is Unacceptable in online.

image

I known the bird support graceful restart feature，this may very friendly for this case

control-plane can support graceful restart，it will not ping down when restart or re-create cilium-agent.

harsimran-pabla:
#### Summary
Introducing BGP graceful restart configuration option per peer along with restart time parameter. This change includes setting this configuration via BGP CRDs and fetching the configuration via BGP CLI.

#### Behaviour change
Previously, BGP CP removes all peer configuration at shutdown. This part of the code is removed, because we do not want to send PEER_DECONFIGURED notification which results peer to tear down BGP session even if graceful restart was negotiated. 

#### RFC 4724 
This change supports only restarting speaker as described [here](https://github.com/osrg/gobgp/blob/master/docs/sources/graceful-restart.md#restarting-speaker). 

#### Example CRD
```
---
apiVersion: "cilium.io/v2alpha1"
kind: CiliumBGPPeeringPolicy
metadata:
  name: rack1
spec:
  nodeSelector:
    matchLabels:
      rack: rack1
  virtualRouters:
  - localASN: 65011
    exportPodCIDR: true
    neighbors:
    - peerAddress: "10.0.0.2/32"
      peerASN: 65011
      gracefulRestart:
        enabled: true
        restartTime: "20s"
```

#### CLI output 
```
cilium# cilium bgp peers -o json
[
  {
    "applied-hold-time-seconds": 9,
    "applied-keep-alive-time-seconds": 3,
    "configured-hold-time-seconds": 90,
    "configured-keep-alive-time-seconds": 30,
    "connect-retry-time-seconds": 120,
    "families": [
      {
        "accepted": 11,
        "advertised": 2,
        "afi": "ipv4",
        "received": 11,
        "safi": "unicast"
      },
      {
        "advertised": 1,
        "afi": "ipv6",
        "safi": "unicast"
      }
    ],
    "graceful-restart": {
      "enabled": true,
      "restart-time-seconds": 20
    },
    "local-asn": 65011,
    "peer-address": "10.0.0.2",
    "peer-asn": 65011,
    "session-state": "established",
    "uptime-nanoseconds": 46313329048
  }
]
```

#### PCAP

Graceful Restart Open Message

#### Raw PCAP

[graceful-restart.pcap.txt](https://github.com/cilium/cilium/files/11621908/graceful-restart.pcap.txt)

Fixes: https://github.com/cilium/cilium/issues/24256

```release-note
Add support for BGP graceful restart configuration via CiliumBGPPeeringPolicy CRD
```


pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumbgppeeringpolicies.yaml
```diff
@@ -132,6 +132,18 @@ spec:
...
+                          gracefulRestartEnable:
+                            description: GracefulRestartEnable flag, when set enables
+                              graceful restart capability.
+                            type: boolean
+                          gracefulRestartRestartTime:
+                            description: "RestartTime is the estimated time it will
+                              take for the BGP session to be re-established with peer
+                              after a restart. After this period, peer will remove
+                              stale routes. This is described RFC 4724 section 4.2.
+                              \n Default is 120s if empty or zero."
```
rastislavs: I wonder why our default for `RestartTime` is `120s`:
 - RFC 4724, section 4 says "A suggested default for the Restart Time is a value less than or equal to the HOLDTIME carried in the OPEN."
- We default HoldTime to `90s`
- GoBGP defaults `RestartTime` to the [HoldTime value](https://github.com/osrg/gobgp/blob/f8c6d24e32d71a7ea0273a51b574fd63559e2943/internal/pkg/config/default.go#L224)

pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumbgppeeringpolicies.yaml
```diff
@@ -132,6 +132,18 @@ spec:
                               whole second.
                             format: duration
                             type: string
+                          gracefulRestartEnable:
```
rastislavs: This is just my personal preference (feel free to ignore), but I would prefer using `gracefulRestartEnabled`, as CRD describes the desired state, not an action

api/v1/openapi.yaml
```diff
@@ -3501,6 +3501,14 @@ definitions:
       uptime-nanoseconds:
         description: BGP peer connection uptime in nano seconds.
         type: integer
+      graceful-restart:
```
rastislavs, line 3540: Just a personal preference, I would prefer `graceful-restart-enabled` more, also to be consistent with the CRD API.

pkg/bgpv1/gobgp/state.go
```diff
@@ -92,6 +92,11 @@ func (g *GoBGPServer) GetPeerState(ctx context.Context) (types.GetPeerStateRespo
 			}
 		}
 
+		if peer.GracefulRestart != nil {
+			peerState.GracefulRestart = peer.GracefulRestart.Enabled
+			peerState.GracefulRestartRestartTimeSeconds = int64(peer.GracefulRestart.RestartTime)
```
rastislavs: Maybe it would also make sense to expose the [PeerRestartTime](https://github.com/osrg/gobgp/blob/f8c6d24e32d71a7ea0273a51b574fd63559e2943/api/gobgp.pb.go#L7560) in the state API?

pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumbgppeeringpolicies.yaml
harsimran-pabla: There was some discussion around this here : https://isovalent.slack.com/archives/C036D1MNKNG/p1682352752459779 

Conclusion was 120s default is in FRR and BIRD, so keeping things consistent with them.

rastislavs: Got it, no major objections about that then.

pkg/bgpv1/gobgp/state.go
harsimran-pabla: hmm, gobgp does not set PeerRestartTime [here](https://github.com/osrg/gobgp/blob/master/internal/pkg/config/util.go#L550-L558) 

I will create upstream PR to fix it. In the meantime, we will just not display peer restart time - since it will be 0

rastislavs (approved):
LGTM now

pkg/bgpv1/gobgp/state.go
harsimran-pabla: https://github.com/osrg/gobgp/pull/2660

pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumbgppeeringpolicies.yaml
```diff
@@ -132,6 +132,24 @@ spec:
                               whole second.
                             format: duration
                             type: string
+                          gracefulRestart:
+                            description: GracefulRestart defined graceful restart
+                              parameters which are negotiated with this neighbor.
```
qmonnet:
```suggestion
                            description: GracefulRestart defines graceful restart
                              parameters which are negotiated with this neighbor.
```

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -100,6 +117,11 @@ type CiliumBGPNeighbor struct {
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:validation:Format=duration
 	KeepAliveTime metav1.Duration `json:"keepAliveTime,omitempty"`
+	// GracefulRestart defined graceful restart parameters which are negotiated
+	// with this neighbor.
```
qmonnet:
```suggestion
	// GracefulRestart defines graceful restart parameters which are negotiated
	// with this neighbor.
```

qmonnet (approved):
I looked mostly at the doc changes, which are minimal and look good.

I spotted one typo in comments/description, please see below.

dylandreimerink (approved):

nathanjsweet (approved):
LGTM for K8s and API

youngnick (approved):

api/v1/openapi.yaml
```diff
@@ -3550,3 +3553,24 @@ definitions:
...
+      restart-time-seconds:
+        description: |-
+          This is the time advertised to peer for the BGP session to be re-established 
+          after a restart. After this period, peer will remove stale routes. 
+          (RFC 4724 section 4.2)
+        type: integer
+      peer-restart-time-seconds:
+        description: |-
+          This is the restart time advertised by the peer.
+        type: integer
```
YutaroHayakawa: Are we willing to help peer's restart? Since we don't import the route from peers, maybe it's not so useful to do it. Then, probably it doesn't make sense to show this to users?

pkg/bgpv1/agent/controller.go
```diff
@@ -443,6 +438,11 @@ func (c *Controller) applyPolicyDefaults(policy *v2alpha1api.CiliumBGPPeeringPol
 			if n.KeepAliveTime.Duration == 0 {
 				n.KeepAliveTime.Duration = n.HoldTime.Duration / 3
 			}
+
+			// apply graceful restart defaults
+			if n.GracefulRestart.Enabled && n.GracefulRestart.RestartTime.Duration == 0 {
+				n.GracefulRestart.RestartTime.Duration = types.DefaultGRRestartTime
```
YutaroHayakawa, line 444: RFC4724 Section 4 says 

```
A suggested default for the Restart Time is a value less than or equal to the HOLDTIME carried in the OPEN.
```

So, Restart Time should be dynamic, so we could delete the default constant, and it should be at least `n.GracefulRestart.RestartTime.Duration = n.HoldTime.Duration`?

pkg/bgpv1/gobgp/state.go
```diff
@@ -92,6 +92,13 @@ func (g *GoBGPServer) GetPeerState(ctx context.Context) (types.GetPeerStateRespo
 			}
 		}
 
+		peerState.GracefulRestart = &models.BgpGracefulRestart{}
+		if peer.GracefulRestart != nil {
+			peerState.GracefulRestart.Enabled = peer.GracefulRestart.Enabled
+			peerState.GracefulRestart.RestartTimeSeconds = int64(peer.GracefulRestart.RestartTime)
+			peerState.GracefulRestart.PeerRestartTimeSeconds = int64(peer.GracefulRestart.PeerRestartTime)
```
YutaroHayakawa: We can delete this if we agree that peer restart time is not needed.

pkg/bgpv1/types/bgp.go
```diff
@@ -17,6 +17,9 @@ const (
 	DefaultBGPConnectRetryTime = 120 * time.Second
 	// DefaultBGPHoldTime defines the default initial value for the BGP HoldTimer (RFC 4271, Section 4.2).
 	DefaultBGPHoldTime = 90 * time.Second
+
+	// DefaultGRRestartTime defines default Restart Time for graceful restart (RFC 4724, section 4.2)
+	DefaultGRRestartTime = 120 * time.Second
```
YutaroHayakawa, line 22: We can delete this as I suggested in the above comment.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -62,6 +62,23 @@ type CiliumBGPPeeringPolicySpec struct {
...
+	// Enabled flag, when set enables graceful restart capability.
+	//
+	// +kubebuilder:validation:Optional
+	Enabled bool `json:"enabled"`
+	// RestartTime is the estimated time it will take for the BGP
+	// session to be re-established with peer after a restart.
+	// After this period, peer will remove stale routes. This is
+	// described RFC 4724 section 4.2.
+	//
+	// Default is 120s if empty or zero.
```
YutaroHayakawa, line 75: We should mention that this only have second precision since we are rounding it up internally. Also, don't forget to update this default value when you agree with my suggestion and change the default.

pkg/bgpv1/manager/reconcile_test.go
```diff
@@ -167,7 +177,8 @@ func TestNeighborReconciler(t *testing.T) {
 				{PeerASN: 64124, PeerAddress: "192.168.0.1/32"},
 				{PeerASN: 64124, PeerAddress: "192.168.0.2/32"},
 			},
-			err: nil,
+			checks: checkTimers{},
```
YutaroHayakawa: We can just omit this since we're embedding the struct. Same for below.

pkg/bgpv1/manager/reconcile_test.go
```diff
@@ -1087,3 +1137,13 @@ func TestReconcileAfterServerReinit(t *testing.T) {
...
 }
+
+// hostPrefixLen returns addr/32 for ipv4 address and addr/128 for ipv6 address
+func toHostPrefix(addr string) string {
+	addrNet := netip.MustParseAddr(addr)
+	bits := 32
+	if addrNet.Is6() {
+		bits = 128
+	}
+	return netip.MustParsePrefix(fmt.Sprintf("%s/%d", addr, bits)).String()
```
YutaroHayakawa: You can use https://pkg.go.dev/net/netip#PrefixFrom here.

pkg/bgpv1/manager/reconcile_test.go
```diff
@@ -270,45 +322,43 @@ func TestNeighborReconciler(t *testing.T) {
...
-				for _, n := range tt.newNeighbors {
-					addr := netip.MustParsePrefix(n.PeerAddress)
-					if paddr == addr.Addr() {
-						seen = true
-						if n.ConnectRetryTime.Duration != 0 && int64(n.ConnectRetryTime.Seconds()) != p.ConnectRetryTimeSeconds {
-							t.Fatalf("ConnectRetryTime does not match: wanted: %d, got: %d", int64(n.ConnectRetryTime.Seconds()), p.ConnectRetryTimeSeconds)
-						}
+				if tt.checks.keepaliveTimer {
+					toCiliumPeer.KeepAliveTime = metav1.Duration{
+						Duration: time.Duration(peer.ConfiguredKeepAliveTimeSeconds) * time.Second,
```
YutaroHayakawa, line 342: It's a bit hard to follow from commit message what is the new test. I think this commit introduces 1) generic testing facilities for timers 2) graceful restart test using 1). So, please consider separating those two into different commits or describe them in detail in the commit message.

pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumbgppeeringpolicies.yaml
```diff
@@ -133,7 +133,7 @@ spec:
                             format: duration
                             type: string
                           gracefulRestart:
-                            description: GracefulRestart defined graceful restart
+                            description: GracefulRestart defines graceful restart
```
YutaroHayakawa, line 136: Please consolidate this fix to the original commit that introduced this description. Same for `bgpp_types.go`.

pkg/bgpv1/test/fixtures.go
```diff
@@ -218,6 +218,8 @@ func setup(ctx context.Context, peerConfigs []gobgpConfig, fixConfig fixtureConf
 			peer.stopGoBGP()
 		}
 
+		f.bgp.BGPMgr.Stop()
```
YutaroHayakawa, line 221: Can't we hook this `Stop()` into the Hive's Stop lifecycle callback?

YutaroHayakawa (changes requested):
Thanks for working for this! I made some comments for existing code. Also, I think we should elaborate more about which RFC/Draft we are referring in the commit message. At least I'm aware of the three GR-related standards that GoBGP implements; RFC4724 (The original GR), RFC8538 (NOTIFICATION support), draft-ietf-idr-long-lived-gr (LLGR). Also, it's hard to say we are compliant to the RFC without more evidence (e.g. PCAP of the exchanged BGP messages). Could you collect that?

api/v1/openapi.yaml
harsimran-pabla: That's right, we do not really program any routes in FIB so we are not helper in this case. I will remove this part.

pkg/bgpv1/agent/controller.go
harsimran-pabla: I was debating what should be the default. [FRR](https://docs.frrouting.org/en/latest/bgp.html#clicmd-bgp-graceful-restart-restart-time-0-4095) uses 120s regardless of hold time. 

I think risk of making it equal to hold time is that cilium agent might not boot up within user defined hold time.

Default hold time is 90s, which is reasonable for restart time as well. 

But if user sets hold time too aggressive - like 3s, restart time of 3s is bit too short imo.

pkg/bgpv1/manager/reconcile_test.go
harsimran-pabla: This is awesome!

pkg/bgpv1/test/fixtures.go
harsimran-pabla: Including Stop in hive would mean that in prod code hive would call gobgp STOP while shutting down. This terminates all neighbors by sending PEER_DECONFIGURED, which results in peer not going into helper mode of graceful restart. 

So, calling this in test cleanup code - so we free up the BGP listen port.

pkg/bgpv1/manager/reconcile_test.go
harsimran-pabla: This refactor was to use require.ElementsMatch to compare two sets. I will split this into two commits

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
harsimran-pabla: I will circle back to this once we conclude on 120s GR default time.

pkg/bgpv1/agent/controller.go
YutaroHayakawa: Ah, indeed 💡 Then, no objection to the current value.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
YutaroHayakawa: No need to change the default value, but still, the precision should be shown.

YutaroHayakawa:
I uploaded one more PCAP file. I checked that if the restarting procedure is initiated correctly according to the RFC4724 Sec4.1. https://datatracker.ietf.org/doc/html/rfc4724#section-4.1

This is a GR capability message sent from Cilium after the restart.

```
            Capability: Graceful Restart capability
                Type: Graceful Restart capability (64)
                Length: 10
                Restart Timers: 0x001e
                    0... .... .... .... = Restart state: No
                    .0.. .... .... .... = Graceful notification: No
                    .... 0000 0001 1110 = Time: 30
                AFI: IPv4 (1)
                SAFI: Unicast (1)
                Flag: 0x00
                    0... .... = Preserve forwarding state: No
                AFI: IPv6 (2)
                SAFI: Unicast (1)
                Flag: 0x00
                    0... .... = Preserve forwarding state: No
```

I could notice that `Restart state` is not set. RFC says `To re-establish the session with its peer, the Restarting Speaker MUST set the "Restart State" bit in the Graceful Restart Capability`. So, I think we must set this.

YutaroHayakawa:
This is an update message sent right after the restart. This contains all of the routes added by reconcilers.

```
Border Gateway Protocol - UPDATE Message
    Marker: ffffffffffffffffffffffffffffffff
    Length: 52
    Type: UPDATE Message (2)
    Withdrawn Routes Length: 0
    Total Path Attribute Length: 20
    Path attributes
        Path Attribute - ORIGIN: IGP
            Flags: 0x40, Transitive, Well-known, Complete
                0... .... = Optional: Not set
                .1.. .... = Transitive: Set
                ..0. .... = Partial: Not set
                ...0 .... = Extended-Length: Not set
                .... 0000 = Unused: 0x0
            Type Code: ORIGIN (1)
            Length: 1
            Origin: IGP (0)
        Path Attribute - AS_PATH: 65002
            Flags: 0x40, Transitive, Well-known, Complete
                0... .... = Optional: Not set
                .1.. .... = Transitive: Set
                ..0. .... = Partial: Not set
                ...0 .... = Extended-Length: Not set
                .... 0000 = Unused: 0x0
            Type Code: AS_PATH (2)
            Length: 6
            AS Path segment: 65002
                Segment type: AS_SEQUENCE (2)
                Segment length (number of ASN): 1
                AS4: 65002
        Path Attribute - NEXT_HOP: 169.254.2.2
            Flags: 0x40, Transitive, Well-known, Complete
                0... .... = Optional: Not set
                .1.. .... = Transitive: Set
                ..0. .... = Partial: Not set
                ...0 .... = Extended-Length: Not set
                .... 0000 = Unused: 0x0
            Type Code: NEXT_HOP (3)
            Length: 4
            Next hop: 169.254.2.2
    Network Layer Reachability Information (NLRI)
        10.255.1.0/24
            NLRI prefix length: 24
            NLRI prefix: 10.255.1.0
        10.254.0.71/32
            NLRI prefix length: 32
            NLRI prefix: 10.254.0.71
```

End-of-Rib follows correctly.

```
Border Gateway Protocol - UPDATE Message
    Marker: ffffffffffffffffffffffffffffffff
    Length: 23
    Type: UPDATE Message (2)
    Withdrawn Routes Length: 0
    Total Path Attribute Length: 0
```

Looks perfect, but I don't understand why this works. RFC says,

```
Once the session between the Restarting Speaker and the Receiving
   Speaker is re-established, the Restarting Speaker will receive and
   process BGP messages from its peers.  However, it MUST defer route
   selection for an address family until it either (a) receives the
   End-of-RIB marker from all its peers (excluding the ones with the
   "Restart State" bit set in the received capability and excluding the
   ones that do not advertise the graceful restart capability) or (b)
   the Selection_Deferral_Timer referred to below has expired.  It is
   noted that prior to route selection, the speaker has no routes to
   advertise to its peers and no routes to update the forwarding state.
```

harsimran-pabla:
@YutaroHayakawa  

> Looks perfect, but I don't understand why this works. RFC says,

I think this is mostly for transit router case, where restarting router must wait to receive all routes before adding them into its RIB and advertising out the best paths. 

I do agree that we can have race, where neighbors are configured first and we send out EOR before receiving all pod cidr adverts. I adjusted the neighbor reconciler priority to be higher than pod/service announcements.

YutaroHayakawa:
Ok, that makes sense to me. Thanks for your clarification. Reordering the reconcilers addressed my last concern. Now looks good to me! Thanks for all of your effort including developing the testing framework!

YutaroHayakawa (approved):
Ok, that makes sense to me. Thanks for your clarification. Reordering the reconcilers addressed my last concern. Now looks good to me! Thanks for all of your effort to make this PR land including developing the testing framework!

YutaroHayakawa (approved):
Ok, that makes sense to me. Thanks for your clarification. Reordering the reconcilers addressed my last concern. Now looks good to me! Thanks for all of your effort to make this PR land including developing the testing framework!

harsimran-pabla:
/test-1.25-4.19

Founderbn2014:
> I uploaded one more PCAP file that shows the full restarting procedure. And I checked that the restarting procedure is initiated correctly according to the RFC4724 Sec4.1. https://datatracker.ietf.org/doc/html/rfc4724#section-4.1
> 
> This is a GR capability message sent from Cilium after the restart.
> 
> ```
>             Capability: Graceful Restart capability
>                 Type: Graceful Restart capability (64)
>                 Length: 10
>                 Restart Timers: 0x001e
>                     0... .... .... .... = Restart state: No
>                     .0.. .... .... .... = Graceful notification: No
>                     .... 0000 0001 1110 = Time: 30
>                 AFI: IPv4 (1)
>                 SAFI: Unicast (1)
>                 Flag: 0x00
>                     0... .... = Preserve forwarding state: No
>                 AFI: IPv6 (2)
>                 SAFI: Unicast (1)
>                 Flag: 0x00
>                     0... .... = Preserve forwarding state: No
> ```
> 
> I could notice that `Restart state` is not set. RFC says `To re-establish the session with its peer, the Restarting Speaker MUST set the "Restart State" bit in the Graceful Restart Capability`. So, I think we must set this.

how to change the FLag from number 0 to 1 ?

YutaroHayakawa:
It's set by default in the latest version: https://github.com/cilium/cilium/pull/28293

Founderbn2014:
> It's set by default in the latest version: #28293

i  want to set F bit for 1，the latest version also setted F bit = 1 in default？

YutaroHayakawa:
Yes

Founderbn2014:
> Yes

thanks a lot。

PR #25662, opened 2023-05-25
Module Health: Add Health Provider/Reporter

tommyp1ckles:
# Modular Health Checking

This PR adds an initial implementation for a health reporter that provides Hive Module health statuses.

The first commit implements the base health provider, which can create health reporters that can be provided as part of cell.Module creation such that all cells in the module can access the reporter and begin reporting their status.

Following commits add the HealthProvider as a provided default dependency that can be used by all cells in the Hive.

This will be the first of three PRs that add a "beta" feature of modular health checking, which will ideally be released in v1.14.

1. Add module health checker implementation.
2. Add Agent API / CLI functionality to dump current agent runtime module health status.
3. Implement health checking for the "Node Manager" (i.e. `pkg/node/manager/`).  
4. (If possible) Implement checking for another module.

### DIFF
--- a/pkg/hive/cell/health.go
+++ b/pkg/hive/cell/health.go
@@ -1,0 +1,240 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package cell
+
+import (
+	"context"
+	"fmt"
+	"sync/atomic"
+	"time"
+
+	"github.com/cilium/cilium/pkg/lock"
+
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
+)
+
+// Level denotes what kind an update is.
+type Level string
+
+const (
+	// StatusUnknown is the default status of a Module, prior to it reporting
+	// any status.
+	// All created
+	StatusUnknown Level = "Unknown"
+
+	// StatusStopped is the status of a Module that has completed, further updates
+	// will not be processed.
+	StatusStopped Level = "Stopped"
+
+	// StatusDegraded is the status of a Module that has entered a degraded state.
+	StatusDegraded Level = "Degraded"
+
+	// StatusOK is the status of a Module that has achieved a desired state.
+	StatusOK Level = "OK"
+)
+
+// HealthReporter provides a method of declaring a Modules health status.
+type HealthReporter interface {
+	// OK declares that a Module has achieved a desired state and has not entered
+	// any unexpected or incorrect states.
+	// Modules should only declare themselves as 'OK' once they have stabilized,
+	// rather than during their initial state. This should be left to be reported
+	// as the default "unknown" to denote that the module has not reached a "ready"
+	// health state.
+	OK(status string)
+
+	// Stopped reports that a module has completed, and will no longer report any
+	// health status.
+	Stopped(reason string)
+
+	// Degraded declares that a module has entered a degraded state.
+	// This means that it may have failed to provide it's intended services, or
+	// to perform it's desired task.
+	Degraded(reason string, err error)
+}
+
+// Health provides exported functions for accessing health status data.
+// As well, provides unexported functions for use during module apply.
+type Health interface {
+	// All returns a copy of all module statuses.
+	// This includes unknown status for modules that have not reported a status yet.
+	All() []Status
+
+	// Get returns a copy of a modules status, by module ID.
+	// This includes unknown status for modules that have not reported a status yet.
+	Get(string) *Status
+
+	// Stop stops the health provider from processing updates.
+	Stop(context.Context) error
+
+	// forModule creates a moduleID scoped reporter handle.
+	forModule(string) HealthReporter
+
+	// processed returns the number of updates processed.
+	processed() uint64
+}
+
+// Update is an event that denotes the change of a modules health state.
+type Update struct {
+	Level
+	ModuleID string
+	Message  string
+	Err      error
+}
+
+// Status is a modules last health state, including the last update.
+type Status struct {
+	// Update is the last reported update for a module.
+	Update
+	// Stopped is true when a module has been completed, thus it contains
+	// its last reporter status. New updates will not be processed.
+	Stopped bool
+	// Final is the stopped message, if the module has been stopped.
+	Final string
+	// LastOK is the time of the last OK status update.
+	LastOK time.Time
+	// LastUpdated is the time of the last status update.
+	LastUpdated time.Time
+}
+
+// String returns a string representation of a Status, implements fmt.Stringer.
+func (s *Status) String() string {
+	var sinceLast string
+	if s.LastUpdated.IsZero() {
+		sinceLast = "never"
+	} else {
+		sinceLast = time.Since(s.LastUpdated).String() + " ago"
+	}
+	return fmt.Sprintf("Status{ModuleID: %s, Level: %s, Since: %s, Message: %s, Err: %v}",
+		s.ModuleID, s.Level, sinceLast, s.Message, s.Err)
+}
+
+// NewHealthProvider starts and returns a health status which processes
+// health status updates.
+func NewHealthProvider() Health {
+	p := &healthProvider{
+		moduleStatuses: make(map[string]Status),
+		running:        true,
+	}
+	return p
+}
+
+func (p *healthProvider) processed() uint64 {
+	return p.numProcessed.Load()
+}
+
+func (p *healthProvider) process(u Update) {
+	prev := func() Status {
+		p.mu.Lock()
+		defer p.mu.Unlock()
+
+		t := time.Now()
+		prev := p.moduleStatuses[u.ModuleID]
+
+		if !p.running {
+			return prev
+		}
+
+		ns := Status{
+			Update:      u,
+			LastUpdated: t,
+		}
+		switch u.Level {
+		case StatusOK:
+			ns.LastOK = t
+		case StatusStopped:
+			// If Stopped, set that module was stopped and preserve last known status.
+			ns = prev
+			ns.Stopped = true
+			ns.Final = u.Message
+		}
+		p.moduleStatuses[u.ModuleID] = ns
+		log.WithField("status", ns.String()).Debug("Processed new health status")
+		return prev
+	}()
+	p.numProcessed.Add(1)
+	if prev.Stopped {
+		log.Warnf("module %q reported health status after being Stopped", u.ModuleID)
+	}
+}
+
+// Finish stops the status provider, and waits for all updates to be processed or
+// returns an error if the context is cancelled first.
+func (p *healthProvider) Stop(ctx context.Context) error {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.running = false // following this, no new reporters will send.
+	return nil
+}
+
+// forModule returns a module scoped status reporter handle for emitting status updates.
+// This is used to automatically provide declared modules with a status reported.
+func (p *healthProvider) forModule(moduleID string) HealthReporter {
+	p.mu.Lock()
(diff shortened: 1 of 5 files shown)


pkg/hive/hive.go
```diff
@@ -134,16 +136,22 @@ type defaults struct {
...
 	Flags       *pflag.FlagSet
 	Lifecycle   Lifecycle
+	Health      cell.Health
 	Logger      logrus.FieldLogger
 	Shutdowner  Shutdowner
 	InvokerList cell.InvokerList
 }
 
 func (h *Hive) provideDefaults() error {
+	h.lifecycle.Append(Hook{OnStop: func(ctx HookContext) error {
```
joamaki: I'd instantiate `cell.Health` locally here since `h.healthStatus` doesn't need to be used by `Hive` directly. And perhaps for clarity we could have `NewHealthProvider` take `Lifecycle` as argument and thus we can just call `h.container.Provide(cell.NewHealthProvider)`.

pkg/hive/cell/health.go
```diff
@@ -0,0 +1,249 @@
...
+		moduleStatuses: make(map[string]Status),
+	}
+	p.running.Store(true)
+
+	p.obs, p.emit, p.complete = stream.Multicast[Update]()
+
+	// Listen for updates, use buffered channel to avoid blocking.
+	go func() {
+		for u := range p.updates {
+			prev := func() Status {
```
joamaki: Why are declaring a closure and then calling it immediately?

pkg/hive/cell/health.go
```diff
@@ -0,0 +1,249 @@
...
+	return &reporter{
+		moduleID: moduleID,
+		emit: func(u Update) {
+			if !p.running.Load() {
+				return
+			}
+			p.wg.Add(1)
+			defer p.wg.Done()
+			p.queued.Add(1)
+			p.updates <- u
```
joamaki: I don't see the benefit over using a buffered channel and separate worker over just taking a write lock here and updating the module's status directly.

pkg/hive/cell/health.go
```diff
@@ -0,0 +1,249 @@
...
+
+	wg        sync.WaitGroup
+	running   atomic.Bool
+	processed atomic.Uint64
+	queued    atomic.Uint64
+
+	moduleStatuses map[string]Status
+
+	obs      stream.Observable[Update]
+	emit     func(Update)
```
joamaki: `emit` doesn't seem to be used, e.g. we can't get streaming updates.

joamaki (changes requested):

pkg/hive/cell/health.go
```diff
@@ -0,0 +1,249 @@
...
+	"github.com/cilium/cilium/pkg/stream"
+
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
+)
+
+// How big the queue buffer for health status updates should be.
+// Insufficient will result in reporter status updates blocking while
+// waiting for the queue to be processed.
+const updatesBufferSize = 2048
```
derailed: This seems dubious and yield allocations. Why 2048? Was this measured?

pkg/hive/cell/health.go
```diff
@@ -0,0 +1,249 @@
...
+
+// Level denotes what kind an update is.
+type Level string
+
+const (
+	StatusUnknown  Level = "Unknown"
```
derailed: nit: comments

pkg/hive/cell/health.go
```diff
@@ -0,0 +1,249 @@
...
+	StatusStopped  Level = "Stopped"
+	StatusDegraded Level = "Degraded"
+	StatusOK       Level = "OK"
+)
+
+// StatusReporter provides a method of declaring a Modules health status.
+type StatusReporter interface {
+	// OK declares that a Module has achieved a desired state and has not entered
```
derailed, line 40: Wondering if we should have a single accessor here is Status(Level, string). If we need to add more levels.
Also should this be call HealthReporter as status is a bit vague?

derailed: Also wondering if health should be in its own package ie cell/health?

pkg/hive/cell/health.go
```diff
@@ -0,0 +1,249 @@
...
+	// Get returns a copy of a modules status, by module ID.
+	// This includes unknown status for modules that have not reported a status yet.
+	Get(string) *Status
+
+	// forModule creates a moduleID scoped reporter handle.
+	forModule(string) StatusReporter
+}
+
+// Update is an event that denotes the change of a modules health state.
+type Update struct {
```
derailed, line 80: Is there value for this to be exported?

pkg/hive/cell/health.go
```diff
@@ -0,0 +1,249 @@
...
+	Level
+	ModuleID string
+	Message  string
+	Err      error
+}
+
+// Status is a modules last health state, including the last update.
+type Status struct {
+	Update
+	Stopped     bool
```
derailed: nit: comments

pkg/hive/cell/health.go
```diff
@@ -0,0 +1,249 @@
...
+	LastOK      time.Time
+	LastUpdated time.Time
+}
+
+func (s *Status) String() string {
```
derailed, line 103: nit: comments

pkg/hive/cell/health.go
derailed: Not to mention the potential for gor leak if Stop never hits. Better to fire this off on start with a context to ensure we clean up?

pkg/hive/cell/health.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	p.running.Store(true)
+
+	p.obs, p.emit, p.complete = stream.Multicast[Update]()
+
+	// Listen for updates, use buffered channel to avoid blocking.
+	go func() {
+		for u := range p.updates {
+			prev := func() Status {
+				p.mu.Lock()
```
derailed: Seems we are holding the lock a bit longer than necessary. Better to scope them if we can

pkg/hive/hive.go
```diff
@@ -55,6 +55,7 @@ type Hive struct {
 	flags                     *pflag.FlagSet
 	viper                     *viper.Viper
 	lifecycle                 *DefaultLifecycle
+	healthStatus              *cell.HealthProvider
```
derailed: nit: naming is this really a health status or the provider

pkg/hive/hive_test.go
```diff
@@ -134,6 +135,50 @@ func TestProvideInvoke(t *testing.T) {
...
+			lc.Append(hive.Hook{
+				OnStop: func(hive.HookContext) error {
+					all = hr.All()
+					return nil
+				}})
+		}),
+		shutdownOnStartCell,
+	)
+	assert.NoError(t, h.Run(), "expected Run to succeed")
+	assert.Equal(t, all[0].Level, cell.StatusOK)
```
derailed: Should probably check that we have actually 2?

pkg/hive/cell/health.go
```diff
@@ -0,0 +1,249 @@
...
+	// Wait for all updates to be processed, ensuring a consistent state.
+done:
+	for {
+		select {
+		case <-inctimer.After(providerDrainPollingInterval):
+			processed := p.processed.Load()
+			queued := p.queued.Load()
+			// Wait for existing reporters to finish.
+			if processed >= queued {
+				break done
```
derailed: why the label? Could we just return at this point?

pkg/hive/cell/health_test.go
```diff
@@ -0,0 +1,91 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package cell
```
derailed, line 4: Better to test the external apis at this stage.

pkg/hive/cell/health.go
tommyp1ckles: I think you're right, a bit of premature optimization here.

derailed (changes requested):
@tommyp1ckles Looks good as a whole... Thank you! 
Just a few Qs/nits.

pkg/hive/cell/health.go
tommyp1ckles: I use this to capture the critical section with a `defer mu.Unlock()` (i.e. makes it less likely someone will accidentally introduce some logic that leaves the lock stuck), rather than having an explicit unlock.

pkg/hive/cell/health.go
tommyp1ckles: Going to remove the buffered channel, I don't think it's actually giving us much yet.

pkg/hive/hive.go
tommyp1ckles: Going to remove this from this struct, we can just provide it to the container directly since Hive doesn't actually need a reference to it.

pkg/hive/cell/health.go
tommyp1ckles: this logic isnt necessary anymore following removing buffered channel.

pkg/hive/cell/health.go
tommyp1ckles: Makes the composed Status Update data available.

pkg/hive/cell/health.go
tommyp1ckles: > Also should this be call HealthReporter as status is a bit vague?

👍 

> Wondering if we should have a single accessor here is Status(Level, string).

I think we should just keep it just the predefined status' for now - I think we want to avoid a proliferation of update types, having OK, Degraded etc is more readable/cleaner imo. If it gets out of hand we can reevaluate.

> Also wondering if health should be in its own package ie cell/health?

I considered this, I like that we can keep some functionality unexported to the Hive (such as forModule).

pkg/hive/cell/health.go
tommyp1ckles: we need the prev state in order to process the current one.

pkg/hive/cell/health_test.go
tommyp1ckles: Makes sense, I added testing the Get function as well.

pkg/hive/cell/health.go
derailed: This seems odd. I would favor defaulting to unexporting as much as we can given the api is in flux. 
Also it seems this is only used within the package? otherwise provide a constructor if needed... 
If the tests used an external package we would gain a better appreciation of the apis we're trying to expose.

pkg/hive/cell/health.go
```diff
@@ -0,0 +1,240 @@
...
+	all := maps.Values(p.moduleStatuses)
+	p.mu.RUnlock()
+	slices.SortFunc(all, func(a, b Status) bool {
+		return a.ModuleID < b.ModuleID
+	})
+	return all
+}
+
+// Get returns the latest status for a module, by module ID.
+func (p *healthProvider) Get(moduleID string) *Status {
```
joamaki, line 201: Could perhaps just return by value so this can't be misused to mutate the status?

joamaki (approved):
LGTM!

lmb (approved):
Approving since Jussi already took a look.

derailed (approved):
@tommyp1ckles Looks good. Just a comment on how much we want/need to export on this api...

pkg/hive/cell/health.go
tommyp1ckles: The map stores values, so this should return a pointer to a copied value.

(I think I want to avoid having to deal with a bunch of `_, ok := hr.Get(...)`)

tommyp1ckles:
force-push/conflict resolution was just import/goimports change, going to set ready-to-merge following linting passing.

PR #25665, opened 2023-05-25
docs: document missing entity 'ingress'

mhofstetter:
The entity `ingress` is missing from the list of pre-defined entities which are available when defining policies which `fromEntities` and `toEntities`.

This PR documents the missing entity.

:exclamation:  backport info 1.12: note that the ingress entity only applies to cluster external traffic

### DIFF
--- a/Documentation/security/policy/language.rst
+++ b/Documentation/security/policy/language.rst
@@ -329,6 +329,10 @@ kube-apiserver
     The kube-apiserver entity represents the kube-apiserver in a Kubernetes
     cluster. This entity represents both deployments of the kube-apiserver:
     within the cluster and outside of the cluster.
+ingress
+    The ingress entity represents the Cilium Envoy instance that handles ingress
+    L7 traffic. Be aware that this also applies for pod-to-pod traffic within
+    the same cluster when using ingress endpoints (also known as *hairpinning*).
 cluster
     Cluster is the logical group of all network endpoints inside of the local
     cluster. This includes all Cilium-managed endpoints of the local cluster,
--- a/Documentation/spelling_wordlist.txt
+++ b/Documentation/spelling_wordlist.txt
@@ -506,6 +506,7 @@ grafana
 graphviz
 grep
 hairpinned
+hairpinning
 hardcode
 hardcoded
 hardcoding


jrajahalme (approved):
Also needs to be backported to v1.12, but noting that the `ingress` entity only applies to cluster external traffic in v1.12.

Documentation/security/policy/language.rst
```diff
@@ -329,6 +329,10 @@ kube-apiserver
     The kube-apiserver entity represents the kube-apiserver in a Kubernetes
     cluster. This entity represents both deployments of the kube-apiserver:
     within the cluster and outside of the cluster.
+ingress
+    The ingress entity represents the Cilium Envoy instance which handles ingress
+    L7 traffic. Be aware that this also applies for pod-to-pod traffic within
+    the same cluster when using ingress endpoints (hairpinning).
```
zacharysarah: Deep into the nitty weeds: s/which/that, some small clarification, italics when introducing a new term
```suggestion
    The ingress entity represents the Cilium Envoy instance that handles ingress
    L7 traffic. Be aware that this also applies for pod-to-pod traffic within
    the same cluster when using ingress endpoints (also known as *hairpinning*).
```

zacharysarah: Handy guidance for ["which" vs. "that"](https://prowritingaid.com/art/438/-Which--or--That-%3A-Know-When-to-Use-Each.aspx#:~:text=That%3A%20Quiz-,The%20Trick%20for%20Remembering%20Which%20vs.,clause%2C%20use%20%E2%80%9Cthat.%E2%80%9D):

> If your sentence has a clause but does not need it, use “which”; if the sentence does need the clause, use “that.”

zacharysarah (approved):
@mhofstetter Good work. Some nits to fix, otherwise LGTM. I'm approving with the understanding that changes are required prior to merge.

Documentation/security/policy/language.rst
mhofstetter: @zacharysarah thanks for your review - i adapted your recommendations. special thanks for the which/that guidance - definitely an important one for me. will try to remember next time! :brain:

jrajahalme (approved):

mhofstetter:
change in the documentation doesn't require running the full ci test suite -> adding label `ready-to-merge`

PR #25671, opened 2023-05-25
clustermesh: allow waiting for the CiliumClusterConfig to appear when required

giorio94:
Reporting the description of the first commit for convenience:

```
857060bbcf07 ("clustermesh: Introduce CiliumClusterConfig") introduced
the CiliumClusterConfig, which the clustermesh-apiserver stores into the
kvstore during startup, and the remote cilium agents read when connecting
to new clusters. Cilium agents tolerate the possible lack of the cluster
configuration, to account for backward compatibility.

This process currently suffers from the possibility of a race condition,
because the agents might connect to the kvstore before that the cluster
configuration has been actually written. Although as of today it doesn't
raise particular concerns given that it is only used for sanity checks,
the issue will be more serious when depending on the information therein
contained to tune the behavior. One such case concerns the upcoming
kvstoremesh approach (caching the information of remote clusters in
the local kvstore), which leverages different prefixes for separation
(hence requiring the cluster config to be present for disambiguation)
and increases the race condition window (since the cluster configuration
of new clusters would be written at runtime rather than on startup).

This commit introduces a new well-known key (`cilium/.has-cilium-config`)
that, when present, conveys that the cluster configuration will be
eventually created, and remote clusters shall wait until it is present
rather than falling back to the legacy approach. This key will be
created by the clustermesh-apiserver init container, ensuring that it is
always present when etcd is exposed to the remote agents.

Alternative approaches that have been considered, but discarded because
they would solve the problem for vanilla clustermesh but not for
kvstoremesh include:
* Setting the CiliumClusterConfig in the clustermesh-apiserver init
  container rather than during the initialization phase.
* Leveraging a readiness probe to set the clustermesh-apiserver as ready
  only once the initialization phase completed, and the configuration
  has been written.
```

Another possibility might involve being strict and making the cluster configuration mandatory starting from v1.14. This looks a no-go to me though, because:
* It would prevent a cluster Cilium v1.14 to mesh with another running Cilium v1.12 (I couldn't find any official documentation about the maximum supported skew in the clustermesh scenario to say whether this is acceptable or not).
* AFAIK. we are not currently setting the cluster configuration when the clustermesh-apiserver is not used (i.e., Cilium is run with the kvstore backend). In this scenario, even a cluster running Cilium v1.13 would be incompatible (unless the cluster config is created manually before upgrading the other clusters).  

/cc @YutaroHayakawa, @marseel, @aanm, @joestringer for opinions.

```release-note
clustermesh: allow waiting for the CiliumClusterConfig to appear when required
```

### DIFF
--- a/pkg/clustermesh/clustermesh.go
+++ b/pkg/clustermesh/clustermesh.go
@@ -66,7 +66,7 @@ type Configuration struct {
 	ClusterSizeDependantInterval kvstore.ClusterSizeDependantIntervalFunc `optional:"true"`
 }
 
-func SetClusterConfig(clusterName string, config *cmtypes.CiliumClusterConfig, backend kvstore.BackendOperations) error {
+func SetClusterConfig(ctx context.Context, clusterName string, config *cmtypes.CiliumClusterConfig, backend kvstore.BackendOperations) error {
 	key := path.Join(kvstore.ClusterConfigPrefix, clusterName)
 
 	val, err := json.Marshal(config)
@@ -74,7 +74,7 @@ func SetClusterConfig(clusterName string, config *cmtypes.CiliumClusterConfig, b
 		return err
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	ctx, cancel := context.WithTimeout(ctx, time.Minute)
 	defer cancel()
 
 	_, err = backend.UpdateIfDifferent(ctx, key, val, true)
@@ -85,10 +85,10 @@ func SetClusterConfig(clusterName string, config *cmtypes.CiliumClusterConfig, b
 	return nil
 }
 
-func GetClusterConfig(clusterName string, backend kvstore.BackendOperations) (*cmtypes.CiliumClusterConfig, error) {
+func GetClusterConfig(ctx context.Context, clusterName string, backend kvstore.BackendOperations) (*cmtypes.CiliumClusterConfig, error) {
 	var config cmtypes.CiliumClusterConfig
 
-	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	ctx, cancel := context.WithTimeout(ctx, time.Minute)
 	defer cancel()
 
 	val, err := backend.Get(ctx, path.Join(kvstore.ClusterConfigPrefix, clusterName))
--- a/pkg/clustermesh/clustermesh.go
+++ b/pkg/clustermesh/clustermesh.go
@@ -108,6 +108,16 @@ func GetClusterConfig(clusterName string, backend kvstore.BackendOperations) (*c
 	return &config, nil
 }
 
+// IsClusterConfigRequired returns whether the remote kvstore guarantees that the
+// cilium cluster config will be eventually created.
+func IsClusterConfigRequired(ctx context.Context, backend kvstore.BackendOperations) (bool, error) {
+	ctx, cancel := context.WithTimeout(ctx, time.Minute)
+	defer cancel()
+
+	val, err := backend.Get(ctx, kvstore.HasClusterConfigPath)
+	return val != nil, err
+}
+
 // RemoteIdentityWatcher is any type which provides identities that have been
 // allocated on a remote cluster.
 type RemoteIdentityWatcher interface {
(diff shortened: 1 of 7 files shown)


pkg/clustermesh/clustermesh.go
```diff
@@ -110,6 +110,16 @@ func GetClusterConfig(clusterName string, backend kvstore.BackendOperations) (*c
 	return &config, nil
 }
 
+// IsClusterConfigRequired returns whether the remote kvstore guarantees that the
+// cilium cluster config will be eventually created.
+func IsClusterConfigRequired(backend kvstore.BackendOperations) (bool, error) {
```
YutaroHayakawa: Why can't we propagate `DoFunc` context to here?

pkg/clustermesh/remote_cluster.go
```diff
@@ -275,6 +277,53 @@ func (rc *remoteCluster) restartRemoteConnection(allocator RemoteIdentityWatcher
...
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, clusterConfigPollTimeout)
+	defer cancel()
+
+	// We retry here rather than returning an simply returning an error and relying on the
+	// controller backoff period to avoid recreating every time a new connection to the remote
+	// kvstore, which would introduce an unnecessary overhead. Still, we do return in case of
+	// consecutive failures, to ensure that we do not retry forever is something strange happened.
+	err = wait.PollImmediateInfiniteWithContext(ctx, clusterConfigPollInterval, func(ctx context.Context) (done bool, err error) {
```
YutaroHayakawa: We don't need the name for return value? At least, `done` is not used.

pkg/clustermesh/clustermesh.go
giorio94: Updated. While I was there, I've also added the context propagation to the Get/SetClusterConfig helpers.

pkg/clustermesh/remote_cluster.go
giorio94: Yeah, I agree that it is not strictly required, but IMHO it helps a bit clarifying the meaning of the boolean parameter.

giorio94:
Force-pushed to rebased onto main and fix a conflict.

sayboras (approved):
The explanation in the commit message is just awesome :100:, the overall changes make sense to me :heavy_check_mark:

YutaroHayakawa (approved):
Now looks good to me. Thanks!

pkg/clustermesh/remote_cluster.go
```diff
@@ -274,6 +276,53 @@ func (rc *remoteCluster) restartRemoteConnection(allocator RemoteIdentityWatcher
...
```
aanm: Any reason why we don't use [our controllers](https://github.com/cilium/cilium/blob/main/pkg/controller/manager.go) for this? They have the benefit of showing up in `cilium status` if something fails. With ` wait.PollImmediateInfiniteWithContext` the failures will get hidden AFAICT.

aanm (approved):
Approving because my changes are not blocking but would be nice to understand the reason behind the choice of `wait.PollImmediateInfiniteWithContext`

pkg/clustermesh/remote_cluster.go
giorio94: Essentially, the remote cluster logic is already wrapped by a controller. The idea here was to have a retry mechanism which doesn't cause that controller to restart (by simply returning the error there), because it would also cause the restart of the etcd connection, introducing an unnecessary overhead. But yeah, it makes sense to use another controller also for that rather than `wait.PollImmediateInfiniteWithContext`.

giorio94:
Removing the ready-to-merge label while addressing the above suggestion.

pkg/clustermesh/remote_cluster.go
giorio94: I've updated the PR to use the controller. @aanm PTAL

giorio94:
Rebased onto main to pick the last changes and allow the Cilium Runtime checks to run successfully.

aanm (approved):

giorio94:
/test-1.16-4.19

PR #25679, opened 2023-05-25
[v1.11] Reinitialize IPsec on device changes

joamaki:
A custom fix for v1.11 to reload IPsec when the network devices change. This is required when IPsec is enabled
and used with IPAM ENI mode where new network devices may be added at runtime.

Since v1.11 does not contain the "runtime-device-detection" feature, this fixes the issue by subscribing to
link updates directly in loader.

### DIFF
--- a/pkg/datapath/loader/base.go
+++ b/pkg/datapath/loader/base.go
@@ -173,6 +173,10 @@ func (l *Loader) reinitializeIPSec(ctx context.Context) error {
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


joamaki:
/test-backport-1.11

joamaki:
This fix was tested to be working on a EKS cluster with IPsec and ENI mode. The new `ensX` interface added at runtime was verified to have the `bpf_network` program loaded onto it.

joamaki:
Closing in favor of backporting to v1.11 from https://github.com/cilium/cilium/pull/25744

PR #23977, opened 2023-02-23
resource: Add Resource[Endpoints] and adapt existing watchers

joamaki:
Adds `Resource[*Endpoints]` which moves the resolving of the Endpoint or EndpointSlice into the constructor for the resource. K8sWatcher refactored to use the `Resource[*Endpoints]` and code cleaned up to work with just `*Endpoints` rather than the 3 different kinds. Some messiness around collapsing them into one type and hiding the underlying resource kind. Please review those changes carefully (CachesSynced, metrics in endpoints watcher, ipcache.ResourceKindEndpoint). I've tried to keep existing behavior as much as I could, but could've potentially missed something.

See individual commits and #23734 for motivation.

PR #25687, opened 2023-05-25
Fix bug with `toServices` policy where service backend churn left stale CIDR identities

### LINKED ISSUE #20477: Deleting service selected by ToServices policy rules can cause policy denies until agent restart
When two services have the same backends and a same label selected in `toServices` rule in a network policy, the internal storage of the policy structure in the form of "generated CIDRs" are removed from the datapath when one of the services is deleted. This can happen even though the other service should still be allowed by policy. As a result, traffic can be dropped for the remaining service.

Likely affects versions v1.8.x onwards.

christarazi:
- k8s: Plumb old Endpoints object through handlers
- k8s: Fix toServices rule translation cleanup
- k8s: Add test for RuleTranslator.Translate()
- k8s: Remove unnecessary AllocatePrefixes from RuleTranslator

Fixes: #20477

---

Main commit "k8s: Fix toServices rule translation cleanup" will be pasted below for ease of review:
> Previously, toServices-based rules did not properly cleanup CIDR
identities. When service backends were removed or changed, the deletion
logic acted on the new object rather than on the old object, thus
the entries that were supposed to be deleted were simply added back in
generateToCidrFromEndpoint().
>
> Fix this by passing the deletion logic the old endpoint object state
and performing a diff between the old and new states.

### DIFF
--- a/pkg/k8s/service_cache.go
+++ b/pkg/k8s/service_cache.go
@@ -57,11 +57,14 @@ type ServiceEvent struct {
 	// Service is the service structure
 	Service *Service
 
-	// OldService is the service structure
+	// OldService is the old service structure
 	OldService *Service
 
 	// Endpoints is the endpoints structured correlated with the service
 	Endpoints *Endpoints
+
+	// OldEndpoints is old endpoints structure.
+	OldEndpoints *Endpoints
 
 	// SWG provides a mechanism to detect if a service was synchronized with
 	// the datapath.
@@ -213,12 +216,13 @@ func (s *ServiceCache) UpdateService(k8sSvc *slim_corev1.Service, swg *lock.Stop
 	if serviceReady {
 		swg.Add()
 		s.Events <- ServiceEvent{
-			Action:     UpdateService,
-			ID:         svcID,
-			Service:    newService,
-			OldService: oldService,
-			Endpoints:  endpoints,
-			SWG:        swg,
+			Action:       UpdateService,
+			ID:           svcID,
+			Service:      newService,
+			OldService:   oldService,
+			Endpoints:    endpoints,
+			OldEndpoints: endpoints,
+			SWG:          swg,
 		}
 	}
 
@@ -232,12 +236,13 @@ func (s *ServiceCache) EnsureService(svcID ServiceID, swg *lock.StoppableWaitGro
 		if endpoints, serviceReady := s.correlateEndpoints(svcID); serviceReady {
 			swg.Add()
 			s.Events <- ServiceEvent{
-				Action:     UpdateService,
-				ID:         svcID,
-				Service:    svc,
-				OldService: svc,
-				Endpoints:  endpoints,
-				SWG:        swg,
+				Action:       UpdateService,
+				ID:           svcID,
+				Service:      svc,
+				OldService:   svc,
+				Endpoints:    endpoints,
+				OldEndpoints: endpoints,
+				SWG:          swg,
 			}
 			return true
 		}
@@ -279,9 +284,11 @@ func (s *ServiceCache) UpdateEndpoints(newEndpoints *Endpoints, swg *lock.Stoppa
 
 	esID := newEndpoints.EndpointSliceID
 
+	var oldEPs *Endpoints
 	eps, ok := s.endpoints[esID.ServiceID]
 	if ok {
-		if eps.epSlices[esID.EndpointSliceName].DeepEqual(newEndpoints) {
+		oldEPs = eps.epSlices[esID.EndpointSliceName]
+		if oldEPs.DeepEqual(newEndpoints) {
 			return esID.ServiceID, newEndpoints
 		}
 	} else {
@@ -297,11 +304,12 @@ func (s *ServiceCache) UpdateEndpoints(newEndpoints *Endpoints, swg *lock.Stoppa
 	if ok && serviceReady {
 		swg.Add()
 		s.Events <- ServiceEvent{
-			Action:    UpdateService,
-			ID:        esID.ServiceID,
-			Service:   svc,
-			Endpoints: endpoints,
-			SWG:       swg,
+			Action:       UpdateService,
+			ID:           esID.ServiceID,
+			Service:      svc,
+			Endpoints:    endpoints,
+			OldEndpoints: oldEPs,
+			SWG:          swg,
 		}
 	}
 
@@ -314,21 +322,27 @@ func (s *ServiceCache) DeleteEndpoints(svcID EndpointSliceID, swg *lock.Stoppabl
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
 
+	var oldEPs *Endpoints
 	svc, serviceOK := s.services[svcID.ServiceID]
-	isEmpty := s.endpoints[svcID.ServiceID].Delete(svcID.EndpointSliceName)
-	if isEmpty {
-		delete(s.endpoints, svcID.ServiceID)
+	eps, ok := s.endpoints[svcID.ServiceID]
+	if ok {
+		oldEPs = eps.epSlices[svcID.EndpointSliceName].DeepCopy() // copy for passing to ServiceEvent
+		isEmpty := eps.Delete(svcID.EndpointSliceName)
+		if isEmpty {
+			delete(s.endpoints, svcID.ServiceID)
+		}
 	}
 	endpoints, _ := s.correlateEndpoints(svcID.ServiceID)
 
 	if serviceOK {
 		swg.Add()
 		event := ServiceEvent{
-			Action:    UpdateService,
-			ID:        svcID.ServiceID,
-			Service:   svc,
-			Endpoints: endpoints,
-			SWG:       swg,
+			Action:       UpdateService,
+			ID:           svcID.ServiceID,
+			Service:      svc,
+			Endpoints:    endpoints,
+			OldEndpoints: oldEPs,
+			SWG:          swg,
 		}
 
 		s.Events <- event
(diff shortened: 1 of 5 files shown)


christarazi:
cc @squeed @joestringer for opening https://github.com/cilium/cilium/issues/20477

operator/watchers/k8s_service_sync.go
```diff
@@ -229,7 +229,7 @@ func endpointsInit(slimClient slimclientset.Interface, swgEps *lock.StoppableWai
 						if oldk8sEP.DeepEqual(newk8sEP) {
 							return
 						}
-						K8sSvcCache.UpdateEndpoints(newk8sEP, swgEps)
+						K8sSvcCache.UpdateEndpoints(oldk8sEP, newk8sEP, swgEps)
```
giorio94: Directly passing the old object will become problematic once starting to use `Resource[Endpoints]` (https://github.com/cilium/cilium/pull/23977), as no longer available as part of the event. Would it be possible to retrieve the old endpoints from the already existing `ServiceCache.endpoints[<service-name>]` map (https://github.com/cilium/cilium/blob/pr%2Fchristarazi%2Fto-services-fixups/pkg/k8s/service_cache.go#L88)?

joamaki: Yep, once https://github.com/cilium/cilium/pull/23977/ is merged you don't have the old object anymore so this approach won't work. The right approach is for `ServiceCache` to keep the previous endpoint. I'd wait until #23977 is merged before refactoring as it changes a lot of the code in this PR (all the different endpoint variants will be gone)

joamaki (changes requested):

operator/watchers/k8s_service_sync.go
christarazi: Thanks. PR should be updated now!

pkg/k8s/service_cache.go
```diff
@@ -291,17 +295,25 @@ func (s *ServiceCache) UpdateEndpoints(newEndpoints *Endpoints, swg *lock.Stoppa
 
 	eps.Upsert(esID.EndpointSliceName, newEndpoints)
 
+	oldEPs, _ := cache[key]
+	// cache is optional as it's only needed for Agents (not for Operator). If
+	// it's nil, then we don't expect to use oldEPs anyway.
+	if cache != nil {
+		cache[key] = newEndpoints
+	}
```
giorio94: Why is this external cache necessary? I think it should be possible to retrieve the oldEPs from `eps.epSlices[esID.EndpointSliceName]`, which is already kept in sync.

christarazi: Good point, removed!

joamaki (approved):

pkg/k8s/watchers/watcher.go
```diff
@@ -637,7 +642,11 @@ func (k *K8sWatcher) k8sServiceHandler() {
 				return
 			}
 
-			translator := k8s.NewK8sTranslator(event.ID, *event.Endpoints, true, svc.Labels, true)
+			var oldEP k8s.Endpoints
+			if event.OldEndpoints != nil {
+				oldEP = *event.OldEndpoints
+			}
```
giorio94: It seems to me that `OldEndpoints` is never set when the event type is `DeleteService`. IIUC the logic here should be adapted to use `*event.Endpoints` as the list of old endpoints.

pkg/k8s/service_cache.go
```diff
@@ -297,11 +302,12 @@ func (s *ServiceCache) UpdateEndpoints(newEndpoints *Endpoints, swg *lock.Stoppa
...
-			Action:    UpdateService,
-			ID:        esID.ServiceID,
-			Service:   svc,
-			Endpoints: endpoints,
-			SWG:       swg,
+			Action:       UpdateService,
+			ID:           esID.ServiceID,
+			Service:      svc,
+			Endpoints:    endpoints,
+			OldEndpoints: oldEPs,
```
giorio94, line 311: Should `OldEndpoints` be set also in the other cases in which an `UpdateService` event is emitted? IMO it is mainly necessary in the `DeleteEndpoints` method below.

giorio94 (commented):
A couple of comments/questions about the OldEndpoints propagation, while the rest of the changes looks good to me.

pkg/k8s/watchers/watcher.go
christarazi: There's no notion of "old" in other event types like "delete", there's only the "current" object. Only the "update" event type has "old" and "new", so we don't need to set `OldEndpoints` for any other event type besides `UpdateService`.

pkg/k8s/service_cache.go
christarazi: Given https://github.com/cilium/cilium/pull/25687#discussion_r1223262168, can you elaborate why it's necessary?

pkg/k8s/watchers/watcher.go
giorio94: I mean, I agree that in case of deletes all the endpoints that are part of the event are the old ones (and it is not necessary to to set `OldEndpoints`). What seems strange to me is that they are passed as `newEPs` to `k8s.NewK8sTranslator()`, and that AFAIU that wil result in `result.PrefixesToRelease` being always empty.

pkg/k8s/service_cache.go
giorio94: I'm referring in particular to the event generated [here](https://github.com/cilium/cilium/blob/pr%2Fchristarazi%2Fto-services-fixups/pkg/k8s/service_cache.go#L331-L340). Essentially, that happens when one endpointslice gets deleted, triggering an update event with the remaining endpoints for that service. Since `OldEndpoints` is not propagated, the corresponding `PrefixesToRelease` rules will not be eventually generated by the `K8sTranslator`.

giorio94: There's also another aspect that I didn't realize earlier, `OldEndpoints` will contain the previous list of endpoints from a given endpointslice, while `Endpoints` the ones merged from all endpointslices. It seems to me that it should not create problems though.

pkg/k8s/watchers/watcher.go
christarazi: Ah now I understand what you're referring to. 

So notice the 4th parameter to `NewK8sTranslator()` -- `revert bool`. This means it will take the "new endpoints" and do the inverse operation, as in it will release them instead of add them.

Given that, I should simply remove the highlighted code here to avoid confusion and just pass an empty `Endpoints` object so it's clearer. Good catch. I probably just added this snippet of code to be consistent with the `UpdateService` case, but then later realized that the delete path does not need the "old" object and forgot to cleanup this code.

pkg/k8s/service_cache.go
christarazi: In the end, I've set the `OldEndpoints` for all `UpdateService` events. Please take another look. Thanks for double checking!

pkg/k8s/watchers/watcher.go
giorio94: Right, I completely missed the revert parameter. Thanks for the clarification.

I still think there's an issue though (due to the subsequent modifications). The only place where revert seems to be used is [here](https://github.com/cilium/cilium/blob/pr%2Fchristarazi%2Fto-services-fixups/pkg/k8s/rule_translate.go#L48-L53), where it guards the population of the egress rules (which is obviously correct). But with the new changes `depopulateEgress` [operates on the `OldEndpoint` object](https://github.com/cilium/cilium/blob/pr%2Fchristarazi%2Fto-services-fixups/pkg/k8s/rule_translate.go#L83). And given that it is always empty nothing will happen. I feel that the revert parameter could be dropped, swapping the parameters in case of deletions:

```go
translator := k8s.NewK8sTranslator(event.ID, *event.Endpoints, k8s.Endpoints{}, svc.Labels)
```

pkg/k8s/service_cache.go
giorio94: Makes sense to me, thanks! Another possibility might be to just use `correlateEndpoints` to obtain the full list of old endpoints before updating the current one, but as mentioned above it shouldn't matter much.

giorio94: 🤔 I see that `OldEndpoints` and `Endpoints` are always the same for updates from remote service backends. Shouldn't they be different? (I'm not 100% sure about how this feature should play with global services)

christarazi: `toServices` (currently) only supports backends that are outside of the cluster (world), and not pod-based backends.

pkg/k8s/watchers/watcher.go
christarazi: Thanks! 

I opted to keep the revert boolean for now (followup on removing `revert` tracked [here](https://github.com/cilium/cilium/issues/26094)) and just go for the quickest change, which is that the old and new object are set to the same object with `revert=true`.

pkg/k8s/service_cache.go
giorio94: The same logic is also used in the external workloads case, and IIUC in that case the backends processed through that function might be outside of the cluster. Still that is most likely an extremely rare condition.

giorio94 (approved):
/lgtm

youngnick (approved):

pkg/k8s/service_cache.go
christarazi: @giorio94 Could you file a followup issue so that we don't forget this?

christarazi:
Thanks for the very thorough review @giorio94!

pkg/k8s/service_cache.go
giorio94: https://github.com/cilium/cilium/issues/26176

PR #25690, opened 2023-05-25
gh/workflows: Pin LVH VMs to v0.3

### LINKED ISSUE #25682: CI: Pin LVH VM version in GHA workflows
Tag LVH VM images, and then use that tag in the GHA workflows (ci-e2e and verifier tests). Otherwise, we risk to run again into situations when new LVH VM images stops working.

brb:
This commit pins LVH VMs version to the v0.3 to avoid situations when a
new VM version is not working (-main) or the temp tag gets deleted
(-20230420.212204).

For bpf-next stick to -main, as we want to test any potential
regressions of new bpf-next kernels.

Fix #25682

### DIFF
--- a/.github/workflows/conformance-e2e-v1.13.yaml
+++ b/.github/workflows/conformance-e2e-v1.13.yaml
@@ -179,26 +179,26 @@ jobs:
         include:
           # See https://github.com/cilium/cilium/issues/20606 for configuration table
           - name: '1'
-            kernel: '4.19-20230420.212204'
+            kernel: '4.19-v0.3'
             kube-proxy: 'iptables'
             kpr: 'disabled'
             tunnel: 'vxlan'
 
           - name: '2'
-            kernel: '5.4-20230420.212204'
+            kernel: '5.4-v0.3'
             kube-proxy: 'iptables'
             kpr: 'disabled'
             tunnel: 'disabled'
 
           - name: '3'
-            kernel: '5.10-20230420.212204'
+            kernel: '5.10-v0.3'
             kube-proxy: 'iptables'
             kpr: 'disabled'
             tunnel: 'disabled'
             endpoint-routes: 'true'
 
           - name: '4'
-            kernel: '5.10-20230420.212204'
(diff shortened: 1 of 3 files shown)


gentoo-root (approved):

aanm (changes requested):
Discussed offline. We should keep the specific tag of the bpf-next so that we don't have regressions of a specific bpf-next polluting the historical runs with flakes. If we have a specific tag then we know which tag has introduced regressions / flakes. This can be updated automatically with renovate.

.github/workflows/conformance-e2e-v1.13.yaml
```diff
@@ -236,14 +236,14 @@ jobs:
             lb-acceleration: 'testing-only'
 
           - name: '8'
-            kernel: 'bpf-next-20230420.212204'
+            kernel: 'bpf-next-main'
```
aanm, line 239: Briefly discussed with @tklauser

```suggestion
            kernel: 'bpf-next-main' # renovate: datasource=docker depName=quay.io/lvh-images/kind sha=sha256:bd6a4dbf830a73c787cc2e4e98db3f8928f494d7e5f22c82f8bb211b78c907b6
```

and then we need something like https://github.com/cilium/cilium-cli/pull/1667/files#diff-20ab1190d08a53a3012bc191ed56205c8f0ffb8fa1715e9d36ecfd1d2ea8a790R159-R168


let's sync next week.

aanm (changes requested):

viktor-kurchenko (approved):

.github/workflows/conformance-e2e-v1.13.yaml
brb: @aanm Bump, any action needed from my side?

aanm: @brb yes, we need something similar as what I suggested

aanm:
follow up done in https://github.com/cilium/cilium/pull/27328. cc @ti-mo @brb

PR #24968, opened 2023-04-19
Add configuration docs for API restrictions

joestringer:
Requires: ~#24967~, ~#25009~

This is the documentation update corresponding to #25009.

The following commits are new:
- api: Add 403 forbidden error codes
- api: Update autogenerated API code for 403 errors
- tools: Add cilium-api-flaggen
- docs: Add configuration docs for API restrictions

This PR is quite big, but primarily due to the auto-generated second commit which is primarily included in order to update this page: https://docs.cilium.io/en/v1.13/api/#api-ref .

PR #25696, opened 2023-05-26
Add missing LB IPAM description in the operator document

YutaroHayakawa:
LB IPAM description was missing from the operator document. Also, the list of CRDs was out-of-date. 

```release-note
Add missing LB IPAM description in the operator document
```

### DIFF
--- a/Documentation/internals/cilium_operator.rst
+++ b/Documentation/internals/cilium_operator.rst
@@ -50,6 +50,14 @@ Cilium. The following custom resources are registered by the Cilium Operator:
 -  CiliumExternalWorkload
 -  CiliumIdentity
 -  CiliumLocalRedirectPolicy
+-  CiliumEgressGatewayPolicy
+-  CiliumEndpointSlice
+-  CiliumClusterwideEnvoyConfig
+-  CiliumEnvoyConfig
+-  CiliumBGPPeeringPolicy
+-  CiliumLoadBalancerIPPool
+-  CiliumNodeConfig
+-  CiliumCIDRGroup
 
 IPAM
 ~~~~
--- a/Documentation/internals/cilium_operator.rst
+++ b/Documentation/internals/cilium_operator.rst
@@ -85,6 +85,12 @@ mode:
 
 For more information on IPAM visit :ref:`address_management`.
 
+Load Balancer IP Address Management
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When :ref:`lb_ipam` is used, Cilium Operator manages IP address
+for ``type: LoadBalancer`` services.
+
 KVStore operations
 ~~~~~~~~~~~~~~~~~~
 


Documentation/internals/cilium_operator.rst
```diff
@@ -50,6 +50,14 @@ Cilium. The following custom resources are registered by the Cilium Operator:
...
 -  CiliumIdentity
 -  CiliumLocalRedirectPolicy
+-  CiliumEgressGatewayPolicy
+-  CiliumEndpointSlice
+-  CiliumClusterwideEnvoyConfig
+-  CiliumClusterwideEnvoyConfig
+-  CiliumBGPPeeringPolicy
+-  CiliumLoadBalancerIPPool
+-  CiliumNodeConfig
+-  CiliumCIDRGroup
```
YutaroHayakawa, line 60: BTW, can't we say, "Operator registers all Cilium-related CRDs" instead of listing all CRDs here? So that we don't have to take care of updating this list. Otherwise, should we automate the update?

qmonnet: I was having the same reaction, I was about to ask if there was any kind of test that we should add to make sure we don't miss any in the future.

Would be nice to have the update automated I think. For example we could:

- Move this to a dedicated RST fragment, that we include from `cilium_operator.rst`
- Generate this fragment from a dedicated `Documentation/Makefile` target

Documentation/internals/cilium_operator.rst
```diff
@@ -50,6 +50,14 @@ Cilium. The following custom resources are registered by the Cilium Operator:
 -  CiliumExternalWorkload
 -  CiliumIdentity
 -  CiliumLocalRedirectPolicy
+-  CiliumEgressGatewayPolicy
+-  CiliumEndpointSlice
+-  CiliumClusterwideEnvoyConfig
+-  CiliumClusterwideEnvoyConfig
```
joestringer: Duplicate

Documentation/internals/cilium_operator.rst
```diff
@@ -77,6 +85,12 @@ mode:
 
 For more information on IPAM visit :ref:`address_management`.
 
+LB IPAM
+~~~~~~~
```
joestringer: Readability nit:

```suggestion
Load Balancer IP Address Management
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Documentation/internals/cilium_operator.rst
joestringer: Automation would be nice, and should be a pretty trivial change to submit along the lines of @qmonnet's suggestion above. That's a fairly common pattern. PR https://github.com/cilium/cilium/pull/24968 is doing something a bit like this, but tracking CRDs wouldn't need to be quite so elaborate.

joestringer (changes requested):

Documentation/internals/cilium_operator.rst
YutaroHayakawa: Actually, it should be `CiliumEnvoyConfig`. Fixed in the latest push.

Documentation/internals/cilium_operator.rst
YutaroHayakawa: Thanks for your suggestion! Fixed in the latest push.

Documentation/internals/cilium_operator.rst
YutaroHayakawa: Sounds like a plan. I created a new issue for this and assigned myself (https://github.com/cilium/cilium/issues/25750).

qmonnet (approved):
With the understanding that automation comes in a follow-up PR, I'm all good with the changes here. Thanks!

joestringer (approved):

PR #25697, opened 2023-05-26
ipcache: Handle error during label injection

christarazi:
There was an error case that was not handled, namely when
resolveIdentity() fails. Found by code inspection. Unclear whether this
was having further ramifications.


christarazi:
Opened by mistake.

Issue #20707, opened 2022-07-29
CI: Create control plane tests for IPsec


brb:
To have a better coverage of the IPSec integration, we could implement privileged control plane tests in a similar manner as we have done for testing the neighbor entries management - https://github.com/cilium/cilium/blob/v1.12.0/pkg/datapath/linux/node_linux_test.go#L1067. 

The test would start multiple node manager instances running in different network namespaces which would configure IPSec among the netns. Then, it would do basic connectivity checks.

pchaigno:
Do we have any examples of past bugs that would have been caught by such tests, but not by BPF unit tests or end-to-end tests?

jschwinger233:
> Do we have any examples of past bugs that would have been caught by such tests

If I remember correctly, @brb caught a bug by control plane test, but I couldn't find the record.

brb:
This was a unit test - https://github.com/cilium/cilium/issues/27280. Anyway, considering that I didn't come up with any scenario, we can close this issue.

PR #24181, opened 2023-03-06
Add BPF test facility to test skb->cb

YutaroHayakawa:
Please see commit message for more details. I piggybacked one small bug fix (node_config.h include guard).

```release-note
Add BPF test facility to test skb->cb
```

PR #25699, opened 2023-05-26
Add BPF unit tests for IPsec

jschwinger233:
A typical datapath for pod to pod ipsec connection is:

local host:
1\. ingress on lxc, where skb->mark is for encryption;
2. stack, xfrm encrypts skb data;
3. egress on cilium_host, where tunnel key is set and skb is redirected to cilium_vxlan;
4. egress on cilium_vxlan
5. egress on eth0

remote host:
6. ingress on eth0
7. ingress on cilium_vxlan (1st), where skb->mark is set for decryption;
8. stack: xfrm decrypts skb data
9. ingress on cilium_vxlan (2nd), skb is redirected to lxc
10. egress on lxc

This PR checks step 1 (`from-container`), step 3 (`from-host`), step 7 (1st `from-overlay`), and step 9 (2nd `from-overlay`). All the tests are done with `TUNNEL_MODE` enabled, both IPv6 and IPv4 are covered.

Step 2 and step 8 will be covered by control plane test mentioned by https://github.com/cilium/cilium/issues/20707

Step 4, 5, 6, 10 have nothing special for IPSec, so I'll leave them aside.

Corner cases and failure cases are to be tested in a separate PR.

### DIFF
--- a/bpf/tests/ipsec_from_host_test.c
+++ b/bpf/tests/ipsec_from_host_test.c
@@ -1,0 +1,317 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/* Copyright Authors of Cilium */
+
+#include "common.h"
+#include <bpf/ctx/skb.h>
+#include "pktgen.h"
+#define ROUTER_IP
+#include "config_replacement.h"
+#undef ROUTER_IP
+
+#define NODE_ID 2333
+#define ENCRYPT_KEY 3
+#define ENABLE_IPV4
+#define ENABLE_IPV6
+#define ENABLE_IPSEC
+#define TUNNEL_MODE
+#define HAVE_ENCAP
+#define ENCAP_IFINDEX 4
+#define SECCTX_FROM_IPCACHE 1
+
+#define skb_set_tunnel_key mock_skb_set_tunnel_key
+#define ctx_redirect mock_ctx_redirect
+
+int mock_skb_set_tunnel_key(__maybe_unused struct __sk_buff *skb,
+			    const struct bpf_tunnel_key *from,
+			    __maybe_unused __u32 size,
+			    __maybe_unused __u32 flags)
+{
+	/* 0xfffff is the default SECLABEL */
+	if (from->tunnel_id != 0xfffff)
+		return -1;
+	if (from->local_ipv4 != 0)
+		return -2;
+	if (from->remote_ipv4 != bpf_htonl(v4_node_two))
+		return -3;
+	return 0;
+}
+
+int mock_ctx_redirect(const struct __sk_buff *ctx __maybe_unused, int ifindex, __u32 flags)
+{
+	if (ifindex != ENCAP_IFINDEX)
+		return -1;
+	if (flags != 0)
+		return -2;
+	return CTX_ACT_REDIRECT;
+}
+
+#include "bpf_host.c"
+
+#define FROM_HOST 0
+#define ESP_SEQUENCE 69865
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PROG_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(max_entries, 1);
+	__array(values, int());
+} entry_call_map __section(".maps") = {
+	.values = {
+		[FROM_HOST] = &cil_from_host,
+	},
+};
+
+PKTGEN("tc", "ipv4_ipsec_from_host")
+int ipv4_ipsec_from_host_pktgen(struct __ctx_buff *ctx)
+{
+	struct pktgen builder;
+	struct ethhdr *l2;
+	struct iphdr *l3;
+	struct ip_esp_hdr *l4;
+	void *data;
+
+	pktgen__init(&builder, ctx);
+
+	l2 = pktgen__push_ethhdr(&builder);
+	if (!l2)
+		return TEST_ERROR;
+	ethhdr__set_macs(l2, (__u8 *)mac_one, (__u8 *)mac_two);
+
+	l3 = pktgen__push_default_iphdr(&builder);
+	if (!l3)
+		return TEST_ERROR;
+	l3->saddr = v4_pod_one;
+	l3->daddr = v4_pod_two;
+
+	l4 = pktgen__push_default_esphdr(&builder);
+	if (!l4)
+		return TEST_ERROR;
+	l4->spi = ENCRYPT_KEY;
+	l4->seq_no = ESP_SEQUENCE;
+
+	data = pktgen__push_data(&builder, default_data, sizeof(default_data));
+	if (!data)
+		return TEST_ERROR;
+
+	pktgen__finish(&builder);
+	return 0;
+}
+
+SETUP("tc", "ipv4_ipsec_from_host")
+int ipv4_ipsec_from_host_setup(struct __ctx_buff *ctx)
+{
+	struct ipcache_key cache_key = {};
+	struct remote_endpoint_info cache_value = {};
+
+	cache_key.lpm_key.prefixlen = IPCACHE_PREFIX_LEN(32);
+	cache_key.family = ENDPOINT_KEY_IPV4;
+	cache_key.ip4 = v4_pod_two;
+	cache_value.sec_identity = 233;
+	cache_value.tunnel_endpoint = v4_node_two;
+	cache_value.node_id = NODE_ID;
+	cache_value.key = ENCRYPT_KEY;
+	map_update_elem(&IPCACHE_MAP, &cache_key, &cache_value, BPF_ANY);
+
+	set_encrypt_key_mark(ctx, ENCRYPT_KEY, NODE_ID);
+	set_identity_meta(ctx, SECLABEL);
+	tail_call_static(ctx, &entry_call_map, FROM_HOST);
+	return TEST_ERROR;
+}
+
+CHECK("tc", "ipv4_ipsec_from_host")
+int ipv4_ipsec_from_host_check(__maybe_unused const struct __ctx_buff *ctx)
+{
+	void *data;
+	void *data_end;
+	__u32 *status_code;
+	struct ethhdr *l2;
+	struct iphdr *l3;
+	struct ip_esp_hdr *l4;
+	__u8 *payload;
+
+	test_init();
+
+	data = (void *)(long)ctx->data;
+	data_end = (void *)(long)ctx->data_end;
+
+	if (data + sizeof(*status_code) > data_end)
+		test_fatal("status code out of bounds");
+
+	status_code = data;
+	assert(*status_code == CTX_ACT_REDIRECT);
+
+	assert(ctx->mark == 0);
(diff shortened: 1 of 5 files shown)


jschwinger233:
BPF coverage test is killing me :sob:

jschwinger233:
/test-1.26-net-next

jschwinger233:
/test-1.26-net-next

dylandreimerink (commented):
Thank you! Improvements of the test runner and pktgen look great! :rocket:

test/bpf_tests/bpf_test.go
```diff
@@ -541,3 +554,21 @@ func (l *Log) FmtString() string {
 
 	return sb.String()
 }
+
+func runBpfProgram(prog *ebpf.Program, data, ctx []byte) (statusCode uint32, dataOut, ctxOut []byte, err error) {
```
jschwinger233, line 558: @YutaroHayakawa Do you think this `runBpfProgram` can help test skb->cb with reference to your PR  https://github.com/cilium/cilium/pull/24181? The parameter `ctx` will be used as skb metadata, like `memcpy(skb, ctx)`, so we can get cb by `ctx[offsetof(skb, cb)]`. In this PR we pass the returned `ctx` from the last BPF test program to the next BPF test program, so whatever changes made by the previous BPF program can be seen afterwards.

rgo3:
Removing myself as a reviewer as cilium/loader was removed again and I miss a lot of context to meaningfully review this PR.

test/bpf_tests/bpf_test.go
YutaroHayakawa: Yeah, this helps, but I intentionally didn't use the `ctx_in` parameter of `BPF_PROG_TEST_RUN` because it's only available after kernel `5.2` and our minimal possible kernel is `4.19`. However, if we can make sure we only run these tests on kernel `5.2` or later, then it's fine. Have you discussed this with someone?

YutaroHayakawa (approved):
LGTM with a little concern around kernel versioning 👍

bpf/tests/ipsec_from_host_test.c
```diff
@@ -0,0 +1,321 @@
...
+	return 0;
+}
+
+SETUP("tc", "ipv4_ipsec_from_host")
+int ipv4_ipsec_from_host_setup(struct __ctx_buff *ctx)
+{
+	struct ipcache_key cache_key = {};
+	struct remote_endpoint_info cache_value = {};
+
+	cache_key.lpm_key.prefixlen = 32;
```
julianwiedmann:
```suggestion
	cache_key.lpm_key.prefixlen = IPCACHE_PREFIX_LEN(32);
```

bpf/tests/ipsec_from_host_test.c
```diff
@@ -0,0 +1,321 @@
...
+	cache_key.family = ENDPOINT_KEY_IPV4;
+	cache_key.ip4 = v4_pod_two;
+	cache_value.sec_identity = 233;
+	cache_value.tunnel_endpoint = v4_node_two;
+	cache_value.node_id = NODE_ID;
+	cache_value.key = ENCRYPT_KEY;
+	map_update_elem(&IPCACHE_MAP, &cache_key, &cache_value, BPF_ANY);
+
+	ctx->mark = 0x6d3e00;
+	ctx->cb[CB_ENCRYPT_IDENTITY] = SECLABEL;
```
julianwiedmann:
```suggestion
	set_identity_meta(ctx, SECLABEL);
```

bpf/tests/ipsec_from_host_test.c
```diff
@@ -0,0 +1,321 @@
...
+	return 0;
+}
+
+SETUP("tc", "ipv6_ipsec_from_host")
+int ipv6_ipsec_from_host_setup(struct __ctx_buff *ctx)
+{
+	struct ipcache_key cache_key = {};
+	struct remote_endpoint_info cache_value = {};
+
+	cache_key.lpm_key.prefixlen = 32;
```
julianwiedmann:
```suggestion
	cache_key.lpm_key.prefixlen = IPCACHE_PREFIX_LEN(128);
```

bpf/tests/ipsec_from_host_test.c
```diff
@@ -0,0 +1,321 @@
...
+	if (!l3)
+		return TEST_ERROR;
+	l3->saddr = v4_pod_one;
+	l3->daddr = v4_pod_two;
+
+	l4 = pktgen__push_esphdr(&builder);
+	if (!l4)
+		return TEST_ERROR;
+	if ((void *)l4 + sizeof(struct ip_esp_hdr) > ctx_data_end(builder.ctx))
+		return 0;
```
julianwiedmann: I would have expected that `pktgen__push_esphdr()` handles this sort of check :thinking:

bpf/tests/ipsec_from_host_test.c
```diff
@@ -0,0 +1,321 @@
...
+	cache_key.lpm_key.prefixlen = 32;
+	cache_key.family = ENDPOINT_KEY_IPV4;
+	cache_key.ip4 = v4_pod_two;
+	cache_value.sec_identity = 233;
+	cache_value.tunnel_endpoint = v4_node_two;
+	cache_value.node_id = NODE_ID;
+	cache_value.key = ENCRYPT_KEY;
+	map_update_elem(&IPCACHE_MAP, &cache_key, &cache_value, BPF_ANY);
+
+	ctx->mark = 0x6d3e00;
```
julianwiedmann: Should we spell out this sort of value more clearly - it's essentially the same calculation as in `set_encrypt_key_mark()`, right?

bpf/tests/ipsec_from_overlay_test.c
```diff
@@ -0,0 +1,562 @@
...
+	memcpy(&ep_value.node_mac, (__u8 *)DEST_NODE_MAC, ETH_ALEN);
+
+	struct endpoint_key ep_key = {
+		.family = ENDPOINT_KEY_IPV4,
+		.ip4 = v4_pod_two,
+	};
+
+	map_update_elem(&ENDPOINTS_MAP, &ep_key, &ep_value, BPF_ANY);
+
+	ctx->mark = 0xd00;
```
julianwiedmann:
```suggestion
	ctx->mark = MARK_MAGIC_DECRYPT;
```

bpf/tests/ipsec_from_overlay_test.c
```diff
@@ -0,0 +1,562 @@
...
+
+	data = (void *)(long)ctx->data;
+	data_end = (void *)(long)ctx->data_end;
+
+	if (data + sizeof(*status_code) > data_end)
+		test_fatal("status code out of bounds");
+
+	status_code = data;
+	assert(*status_code == CTX_ACT_OK);
+	assert(ctx->mark == 0xd00);
```
julianwiedmann: can we also test `skb->pkt_type` here? (potentially other places too?)

julianwiedmann (commented):
Looks great! Just a few smaller comments, that probably apply to more places than just the ones I picked.

And some more general questions for my understanding

- It seems like you've decided to not use the encryption macros / helpers. Could you explain a bit more what the advantage is - I would think it's nice to use the same patterns as in the datapath, so it's easy to find all users?

- Do we need the same set of tests for native routing (without `TUNNEL_MODE`) ?

- are there any other datapath configs (ie. other CONFIG_* combinations) that we should cover, because they interact with IPSec? Something like per-EP routes for instance. Not saying we need to handle all combinations here, but just so that we are aware what's missing.

bpf/tests/ipsec_from_overlay_test.c
jschwinger233: Julian, your comment made me find this very interesting issue: the context bytes passed to BPF_PROG_TEST_RUN aren't used as skb metadata unconditionally; before I said "briefly we can expect `memcpy(skb, ctx)` happened in kernel", but that's not true.

I did some probe and found only a few fields are copied, briefly speaking it's implemented like:

```c
struct sk_buff *skb, struct __sk_buff *__skb;
memcpy(__skb, ctx);
skb->mark = __skb->mark;
skb->priority = __skb->priority;
skb->skb_iif = __skb->ingress_ifindex;
skb->tstamp = __skb->tstamp;
memcpy(&cb->data, __skb->cb, QDISC_CB_PRIV_LEN);
...
```

skb->pkt_type won't be copied, so I can't check the skb->pkt_type in our bpt unittest.

The source code for your reference: https://github.com/torvalds/linux/blob/5f63595ebd82f56a2dd36ca013dd7f5ff2e2416a/net/bpf/test_run.c#L1079-L1083, that `convert___skb_to_skb` function is called by `bpf_prog_test_run_skb`.

jschwinger233: For this PR, I'll mock the `ctx_change_type` to make sure the test cases run as expected. Thank you!

jschwinger233: This `convert_skb_to___skb` is even more clear to tell what fields are copied: https://github.com/torvalds/linux/blob/5f63595ebd82f56a2dd36ca013dd7f5ff2e2416a/net/bpf/test_run.c#L1103

```c
	struct qdisc_skb_cb *cb = (struct qdisc_skb_cb *)skb->cb;

	if (!__skb)
		return;

	__skb->mark = skb->mark;
	__skb->priority = skb->priority;
	__skb->ingress_ifindex = skb->skb_iif;
	__skb->ifindex = skb->dev->ifindex;
	__skb->tstamp = skb->tstamp;
	memcpy(__skb->cb, &cb->data, QDISC_CB_PRIV_LEN);
	__skb->wire_len = cb->pkt_len;
	__skb->gso_segs = skb_shinfo(skb)->gso_segs;
	__skb->hwtstamp = skb_shinfo(skb)->hwtstamps.hwtstamp;
```

This is also called inside `bpf_prog_test_run_skb`: https://github.com/torvalds/linux/blob/5f63595ebd82f56a2dd36ca013dd7f5ff2e2416a/net/bpf/test_run.c#L1243

jschwinger233: image
According to the blame info and the commit which introduced these conversion functions, I don't see reasons to forbid us copying skb->pkt_gen in bpf_prog_test_run. I'll see what I can do in future, but now let it be.

julianwiedmann: Thanks for digging, that's perfectly fine then :+1:

julianwiedmann (approved):
Looks good, thank you!

If the issue with `skb_change_type_called` isn't easily resolvable, I would just punt the testing of `pkt_type` to a follow-on PR. No need to hold this one :).

pchaigno:
I'm marking for backport up to v1.12 because it seems likely to be easy to backport and can give us feedback on IPsec backports. If it gets too complicated (i.e., depends on other PRs being backported), I'll drop the backports.

pchaigno:
Removing from v1.12 backports as the BPF unit tests are too different there for this to be easily backportable.

PR #25703, opened 2023-05-26
Makefile: remove -test.v from GOTEST_BASE

ti-mo:
This makes `go test` rather verbose in CI, since it will echo all stdio generated during all tests. The default behaviour of 'go test' (without -v) is to buffer all stdio and only display it when a test fails. This should be sufficient for privileged and integration tests.


ti-mo:
Turns out -v is ignored when combined with -json, so it doesn't really change anything. Created a https://github.com/mfridman/tparse/pull/93 PR to finetune the verbosity a little. What actually needs changing is setting tparse's `-follow` flag when V=1. V=0 is too quiet for CI.

Moving this into draft until tparse has been updated.

ti-mo:
tparse gained the `-progress` flag in https://github.com/mfridman/tparse/pull/95. Updated this PR to use it and pull in tparse at the latest commit from main.

borkmann (approved):

nebril (approved):

joestringer (approved):

PR #25516, opened 2023-05-17
Transfer Runtime tests to GitHub actions

aanm:
### Description

The current CI setup runs Runtime and Privileged Tests on Jenkins. However, we can achieve the same functionality using GitHub Actions, providing faster feedback and a better user experience. Jenkins, which can sometimes take an hour to allocate nodes just to run the tests, can be replaced with GitHub Actions to obtain quicker results.

### Testing

The jobs are taking less than 15 minutes to complete. The GitHub Actions also offers operations such as re-run a specific focus group which is helpful in case of flakes since it's not necessary to re-run the entire test suite.

- GitHub Action run: https://github.com/cilium/cilium/actions/runs/5003625978

### Notes

- @nbusseneau can we remove `jenkinsfiles/ginkgo-runtime-kernel.Jenkinsfile` or we need to leave it here because of stable branches? If we leave it can we stop triggering it once this PR is merged?

- In the  GitHub Action example run you will see a summary for each Job (for example [here](https://github.com/cilium/cilium/actions/runs/5003625978#summary-13551098106)). This summary is **not** part of the changes on this PR but they will be added in a follow up PR.

- There are some optimizations made specifically for the LVH as part of the GH worfklow. Those optimizations will be removed from this workflow in a follow up PR once they are added as part of the LVH action itself.

PR #25713, opened 2023-05-26
Run all ginkgo tests on GitHub actions

aanm:
### Description

The current CI setup runs all ginkgo tests on Jenkins. Some of the runtime tests were already transferred into Github actions https://github.com/cilium/cilium/pull/25516 and this PR is moving the missing pieces.

### Testing

Example of some runs:
- https://github.com/cilium/cilium/actions/runs/5090956330
- https://github.com/cilium/cilium/actions/runs/5133753485

#### Pros :tada: 

- Instead of hours to wait for test feedback, the entire CI is executed in **less than 20 minutes**.
- If a particular test is flaky, the developer can re-run the specific job for that test instead of re-running the entire suite. A good example just happened with https://github.com/cilium/cilium/actions/runs/5090956330 where attempt [#1](https://github.com/cilium/cilium/actions/runs/5090956330/attempts/1) failed with Kafka (also flaky on Jenkins) and the attempt [#2](https://github.com/cilium/cilium/actions/runs/5090956330/attempts/2) only had this Kafka test re-triggered while the remaining others were marked as successful.

#### Cons :disappointed: 

- It might introduce some new flakes. This can be reassured by re-running the tests 10 times before this PR is merged.
- Kind and GitHub don't have the same capabilities has a baremetal node and some changes had to be done:
  - Tests that need be disabled because Kind doesn't mount `/proc/sys/net/core/default_qdisc` from the host:
    - K8sDatapathBandwidthTest Checks Bandwidth Rate-Limiting Checks Pod to Pod bandwidth, direct routing
    - K8sDatapathBandwidthTest Checks Bandwidth Rate-Limiting Checks Pod to Pod bandwidth, geneve tunneling
    - K8sDatapathBandwidthTest Checks Bandwidth Rate-Limiting Checks Pod to Pod bandwidth, vxlan tunneling
  - Tests had to be modified to take into account that GitHub doesn't support IPv6 connectivity:
    - K8sDatapathConfig Encapsulation Check iptables masquerading with random-fully
- The oldest K8s version supported by kind is 1.19.x. We will need to bump our minimal tested version from 1.16 to 1.19.

#### Re-introduce the `f-XX` prefix?
  The https://github.com/cilium/cilium/pull/25516 had a prefix (`f-XX`) for "focus" matrix dimension but it was removed 
  because it didn't make sense. However, it might make sense to introduce them in this PR because it's hard to navigate 
  through the web UI between jobs that have the same. For example: `datapath-egress-tun-disabled-endpoint-routes- 
  disabled`, `datapath-egress-tun-disabled-endpoint-routes-enabled`, `datapath-egress-tun-vxlan-endpoint-routes- 
  disabled` and `datapath-egress-tun-vxlan-endpoint-routes-enabled`.

### DIFF
--- a/.github/workflows/conformance-ginkgo.yaml
+++ b/.github/workflows/conformance-ginkgo.yaml
@@ -1,0 +1,759 @@
+name: Conformance ginkgo (ci-ginkgo)
+
+# Any change in triggers needs to be reflected in the concurrency group.
+on:
+  issue_comment:
+    types:
+      - created
+  # Run every 6 hours
+  schedule:
+    - cron:  '0 1/6 * * *'
+  ### FOR TESTING PURPOSES
+  # This workflow runs in the context of `main`, and ignores changes to
+  # workflow files in PRs. For testing changes to this workflow from a PR:
+  # - Make sure the PR uses a branch from the base repository (requires write
+  #   privileges). It will not work with a branch from a fork (missing secrets).
+  # - Uncomment the `pull_request` event below, commit separately with a `DO
+  #   NOT MERGE` message, and push to the PR. As long as the commit is present,
+  #   any push to the PR will trigger this workflow.
+  # - Don't forget to remove the `DO NOT MERGE` commit once satisfied. The run
+  #   will disappear from the PR checks: please provide a direct link to the
+  #   successful workflow run (can be found from Actions tab) in a comment.
+  #
+  # pull_request: {}
+  ###
+
+# By specifying the access of one of the scopes, all of those that are not
+# specified are set to 'none'.
+permissions:
+  # To be able to access the repository with actions/checkout
+  contents: read
+  # To allow retrieving information from the PR API
+  pull-requests: read
+  # So that Sibz/github-status-action can write into the status API
+  statuses: write
+
+concurrency:
+  # Structure:
+  # - Workflow name
+  # - Event type
+  # - A unique identifier depending on event type:
+  #   - schedule: SHA
+  #   - issue_comment: PR number
+  #   - pull_request: PR number
+  #
+  # This structure ensures a unique concurrency group name is generated for each
+  # type of testing:
+  # - schedule: {name} schedule {SHA}
+  # - issue_comment: {name} issue_comment {PR number}
+  # - pull_request: {name} pull_request {PR number}
+  #
+  # Note: for `issue_comment` triggers, we additionally need to filter out based
+  # on comment content, otherwise any comment will interrupt workflow runs.
+  group: |
+    ${{ github.workflow }}
+    ${{ github.event_name }}
+    ${{
+      (github.event_name == 'schedule' && github.sha) ||
+      (github.event_name == 'issue_comment' && (
+        github.event.comment.body == '/ci-ginkgo' ||
+        github.event.comment.body == '/test'
+      ) && github.event.issue.number) ||
+      (github.event_name == 'pull_request' && github.event.pull_request.number)
+    }}
+  cancel-in-progress: true
+
+env:
+  check_url: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
+
+jobs:
+  # Pre-build the ginkgo binary so that we don't have to build it for all
+  # runners.
+  build-ginkgo-binary:
+    runs-on: ubuntu-latest
+    name: Build Ginkgo E2E
+    timeout-minutes: 30
+    if: |
+      (github.event_name == 'issue_comment' && (
+        github.event.comment.body == '/ci-ginkgo' ||
+        (github.event.comment.body == '/test' && needs.check_changes.outputs.tested == 'true')
+      )) ||
+      github.event_name == 'schedule' ||
+      github.event_name == 'pull_request'
+    steps:
+      - name: Set up job variables
+        id: vars
+        run: |
+          if [ ${{ github.event.issue.pull_request || github.event.pull_request }} ]; then
+            PR_API_JSON=$(curl \
+              -H "Accept: application/vnd.github.v3+json" \
+              -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+              ${{ github.event.issue.pull_request.url || github.event.pull_request.url }})
+            SHA=$(echo "$PR_API_JSON" | jq -r ".head.sha")
+          else
+            SHA=${{ github.sha }}
+          fi
+
+          echo "sha=${SHA}" >> $GITHUB_OUTPUT
+
+      - name: Set commit status to pending
+        uses: Sibz/github-status-action@650dd1a882a76dbbbc4576fb5974b8d22f29847f # v1.1.6
+        with:
+          authToken: ${{ secrets.GITHUB_TOKEN }}
+          sha: ${{ steps.vars.outputs.sha }}
+          context: ${{ github.workflow }}
+          description: Ginkgo test in progress...
+          state: pending
+          target_url: ${{ env.check_url }}
+
+      - name: Checkout code
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+        with:
+          ref: ${{ steps.vars.outputs.sha }}
+          persist-credentials: false
+
+      - name: Set Environment Variables
+        uses: ./.github/actions/set-env-variables
+
+      # Load Ginkgo build from GitHub
+      - name: Load ginkgo E2E from GH cache
+        uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
+        id: cache
+        with:
+          path: /tmp/.ginkgo-build/
+          key: ${{ runner.os }}-ginkgo-e2e-${{ hashFiles('**/*.go') }}
+
+      - name: Install Go
+        if: ${{ steps.cache.outputs.cache-hit != 'true' }}
+        uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
+        with:
+          # renovate: datasource=golang-version depName=go
+          go-version: 1.20.4
+
+      - name: Build Ginkgo
+        if: ${{ steps.cache.outputs.cache-hit != 'true' }}
+        shell: bash
+        run: |
+          go install github.com/onsi/ginkgo/ginkgo@v1.16.5
+          mkdir -p /tmp/.ginkgo-build
+
+      - name: Build Test
+        if: ${{ steps.cache.outputs.cache-hit != 'true' }}
+        shell: bash
+        run: |
+          cd test
+          /home/runner/go/bin/ginkgo build
+          strip test.test
+          tar -cz test.test -f test.tgz
+
+      - name: Store Ginkgo Test in GitHub cache path
+        if: ${{ steps.cache.outputs.cache-hit != 'true' }}
+        shell: bash
+        run: |
+          mkdir -p /tmp/.ginkgo-build/
+          if [ -f test/test.tgz ]; then
+            cp test/test.tgz /tmp/.ginkgo-build/
+            echo "file copied"
+          fi
+
+      - name: Wait for images to be available
+        timeout-minutes: 10
+        shell: bash
+        run: |
+          for image in cilium-ci operator-generic-ci hubble-relay-ci ; do
+            until docker manifest inspect quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/$image:${{ steps.vars.outputs.sha }} &> /dev/null; do sleep 45s; done
+          done
+
+  setup-and-test:
+    needs: build-ginkgo-binary
+    runs-on:
+      group: ginkgo-runners
+    timeout-minutes: 35
+    name: "E2E Test (${{ matrix.k8s-version }}, ${{matrix.focus}})"
+    env:
+      job_name: "E2E Test (${{ matrix.k8s-version }}, ${{matrix.focus}})"
+    strategy:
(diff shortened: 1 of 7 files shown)


test/helpers/kubectl.go
```diff
@@ -1139,6 +1139,35 @@ func (kub *Kubectl) GetNodeIPByLabel(label string, external bool) (string, error
...
+		return "", fmt.Errorf("cannot retrieve node to read IP: %s", res.CombineOutput())
+	}
+
+	out := strings.Trim(res.Stdout(), "\n")
+	if len(out) == 0 {
+		return "", fmt.Errorf("no matching node to read IP with label '%v'", label)
+	}
+
+	for _, ipStr := range strings.Fields(out) {
+		if ip := net.ParseIP(ipStr); ip.To4() == nil {
```
gentoo-root: Is it possible to check something like `ip.To6() != nil`? I think, that would be more straightforward.

test/helpers/kubectl.go
```diff
@@ -1139,6 +1139,35 @@ func (kub *Kubectl) GetNodeIPByLabel(label string, external bool) (string, error
...
+			return ipStr, nil
+		}
+	}
+
+	return "", fmt.Errorf("found %s ip addrs, but they do not belong to the v4 family",
```
gentoo-root: v4 -> v6?

test/helpers/kubectl.go
```diff
@@ -1139,6 +1139,35 @@ func (kub *Kubectl) GetNodeIPByLabel(label string, external bool) (string, error
 		out)
 }
 
+// GetNodeIPv6ByLabel returns the IPv6 of the node with cilium.io/ci-node=label.
+// An error is returned if a node cannot be found.
+func (kub *Kubectl) GetNodeIPv6ByLabel(label string, external bool) (string, error) {
```
gentoo-root, line 1161: The whole function duplicates GetNodeIPByLabel with really minor differences. Can they be combined into a single one?

test/k8s/manifests/demo-customcalls.yaml
```diff
@@ -101,9 +101,9 @@ spec:
   - name: cilium-builder
     image: quay.io/cilium/cilium-builder:b6f1bba2566617226bce2cf0ab10430dd1cdcb69@sha256:2d9ea3ed195fd352903cccc2aea450c27a04f96ccb6ef6f424f5a4fcae7a2aa3
     workingDir: /cilium
-    command: ["sleep"]
+    command: ["/bin/sh", "-c"]
     args:
-      - "1000h"
+      - "git config --global --add safe.directory /cilium; sleep 1000h"
```
gentoo-root: `sleep infinity` would be more appropriate here?

test/helpers/kubectl.go
aanm: Unfortunately, `ip.To16()` doesn't return nil even if the IP address is an IPv4 address.

gentoo-root (approved):

bimmlerd (approved):
only looked at the contributing change, which is kind.sh - printing the version is fine :)

viktor-kurchenko (approved):
LGTM.

.github/workflows/conformance-ginkgo.yaml
```diff
@@ -0,0 +1,926 @@
...
+      - name: Upload artifacts
+        if: ${{ !success() }}
+        uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
+        with:
+          name: cilium-sysdumps
+          path: |
+            cilium-sysdump-*.zip
+            bugtool-*.tar.gz
+            test_results-*.tar.gz
+          retention-days: 5
```
pchaigno, line 702: Where are these uploaded? I can't see any in the case of the failing test: https://github.com/cilium/cilium/actions/runs/5090956330/jobs/9150414890. Did I miss some page?

aanm: They are uploaded to the summary page of the job run here: https://github.com/cilium/cilium/actions/runs/5090956330

Special note that previous attempts, which the link that you have posted is pointing to attempt [#1](https://github.com/cilium/cilium/actions/runs/5090956330/attempts/1), will have their attachments removed once a new attempt is triggered. Also if we re-trigger a job failure, which is what I did with the Kafka test, the attachments of the previous attempts for successful jobs will be "transferred" to the new attempt.

pchaigno: Uh, wait. You don't collect full ginkgo artifacts anymore, just sysdumps? That's not quite the same. In sysdumps, we don't have journalctl, dmesg, or test-output.log.

I can't believe I missed that for the Runtime tests :grimacing:

aanm: Which ginkgo artifacts are you referring to? You mean the `test_results-*.tar.gz`? That's part of the `cilium-sysdumps.zip` artifact.

.github/workflows/conformance-ginkgo.yaml
```diff
@@ -0,0 +1,926 @@
...
+  # K8sDatapathCustomCalls Basic test with byte-counter Loads byte-counter and gets consistent values, with per-endpoint routes
+  # K8sDatapathLRPTests Checks local redirect policy LRP connectivity
+  # K8sDatapathLRPTests Checks local redirect policy LRP restores service when removed
+  ###
+  update: "K8sUpdates"
+  # K8sUpdates Tests upgrade and downgrade from a Cilium stable image to master
+  ###
+  kafka: "K8sKafkaPolicyTest"
+  # K8sKafkaPolicyTest Kafka Policy Tests KafkaPolicies
+  ###
```
nbusseneau: I think the comments with test names will be a pain to maintain, do we really need them?

.github/workflows/conformance-ginkgo.yaml
```diff
@@ -0,0 +1,926 @@
...
+            SHA=${{ github.sha }}
+          fi
+
+          echo "sha=${SHA}" >> $GITHUB_OUTPUT
+
+          ipfamily="dual"
+          case ${{ matrix.k8s-version }} in
+              1.27)
+                  kube_image="kindest/node:v1.27.1@sha256:b7d12ed662b873bd8510879c1846e87c7e676a79fefc93e17b2a52989d3ff42b"
+                  kernel="bpf-next-20230526.105339@sha256:4133d4e09b1e86ac175df8d899873180281bb4220dc43e2566c47b0241637411"
```
nbusseneau: Is there a specific reason for handling kind and kernel versions like this instead of with `matrix` env variables? I think the less bash processing the better.

.github/workflows/conformance-ginkgo.yaml
```diff
@@ -0,0 +1,926 @@
...
+                  ipfamily="ipv4"
+                  ;;
+              *)
+                  echo "Invalid k8s version"
+                  ;;
+          esac
+
+          case ${{ matrix.focus }} in
+              agent-chaos)
+                  focus="${{ env.agent-chaos }}"
```
nbusseneau: In a similar fashion, instead of `matrix` + `env` glued together with this bash processing, wouldn't it be more convenient to have `matrix.focus` handle this?

.github/workflows/conformance-ginkgo.yaml
```diff
@@ -0,0 +1,926 @@
...
+          sudo cp /tmp/.ubuntu-pkgs/*.deb /var/cache/apt/archives/
+          sudo apt-get -y --no-install-recommends install /tmp/.ubuntu-pkgs/*.deb
+          sudo kvm-ok
+
+      - name: Install cilium-cli
+        shell: bash
+        run: |
+          cid=$(docker create quay.io/cilium/cilium-cli-ci:latest ls)
+          docker cp $cid:/usr/local/bin/cilium ./cilium-cli
+          docker rm $cid
```
nbusseneau, line 562: Love this :D

Any reason to use `latest`? We typically pin the CLI version in other workflows, and it's then handled by Renovate.

nbusseneau (approved):
LGTM overall, some questions below.

.github/workflows/conformance-ginkgo.yaml
aanm: Yes, it is really useful to get an quick idea oh what tests are being covered by the focus regex.

.github/workflows/conformance-ginkgo.yaml
aanm: Yes. Having the kernel and kube_image as part of the matrix it so would be considered a two new dimensions in the matrix and increase the number of combinations to test.

.github/workflows/conformance-ginkgo.yaml
aanm: This bash hack is simply to give a better UX in terms of maintaining the page of focus that we run and visualizing the page of test results. I've tried that in the beginning and they looked like this https://github.com/cilium/cilium/actions/runs/4495666151

.github/workflows/conformance-ginkgo.yaml
aanm: This is not on me, I'm simply copy pasting what other workflows are doing :sweat_smile:

.github/workflows/conformance-ginkgo.yaml
nbusseneau: But can this not be handled via `include`, such that the matrix does not have 2 dimensions but the proper kernel env variables are inserted for the proper K8s version? 🤔

.github/workflows/conformance-ginkgo.yaml
nbusseneau: I agree, but I think they will become obsolete very quickly as people will forget to add them when creating new tests, even though we're not supposed to it still happens xD

.github/workflows/conformance-ginkgo.yaml
nbusseneau: Makes sense, then I would take same approach as suggested in the discussion above with `include`.

.github/workflows/conformance-ginkgo.yaml
nbusseneau: Fair 😄

.github/workflows/conformance-ginkgo.yaml
aanm: Ah, also one of my first approaches :D After working with it for a couple days it was really frustrating to maintain the workflow https://github.com/cilium/cilium/actions/runs/4573014881/workflow#L175-L497

aanm: Discussed offline with @nbusseneau about this. I've tested out @nbusseneau proposal [here](https://github.com/cilium/cilium/actions/runs/5158740999/jobs/9292789247) and it works really well. 

Unfortunately, GitHub workflow files lack a GitHub context that allows access to the job name during their execution. This presents a significant challenge when generating the JUnit filename because it is generated based on the job name and matrix name. On datastudio, we retrieve the job name for the GitHub API, which typically appears as something like `"name": "E2E Test (1.26, datapath-bgp-custom-lrp)"`. However, the JUnit filename incorporates all matrix dimensions and looks like this: `E2E Test (1.26, datapath-bgp-custom-lrp, 5.4-20230526.105339@sha256:523ff0c81ed9be73a2d0d02a64c72655225efdd32e35220f530ca3eff3eada3c).xml`. This discrepancy makes it exceedingly difficult to establish a clear correlation between the JUnit filename and the job name. We need to know which JUnit file was generated by which GitHub job so that we can store it properly on datastudio.

nbusseneau: This is because we set up the filenames for sysdump and JUnit with `${{ join(matrix.*, '-') }}`:
```
./cilium-cli sysdump --output-filename cilium-sysdump-${{ join(matrix.*, '-') }}-final || true
...
junit_filename="${{ env.job_name }} (${{ join(matrix.*, ', ') }}).xml"
```
To get back to the same behaviour as initially proposed in the PR, we can use `${{ matrix.k8s-version }}-${{ matrix.focus }}`:
```
./cilium-cli sysdump --output-filename cilium-sysdump-${{ matrix.k8s-version }}-${{ matrix.focus }}-final || true
...
junit_filename="${{ env.job_name }} (${{ matrix.k8s-version }}-${{ matrix.focus }}).xml"
```

joestringer:
Nice work! I think the overall approach here where we roll out the new logic, get at least a week's worth of feedback, then disable the old approach is a reasonable way to mitigate the risks associated with migrating this infrastructure.

I glanced through the PR at a high level and it made sense to me, but I didn't go into any depth on specific changes.

I note that during the discussion during the community meeting, we raised that Jenkins will continue to run on older branches. This is fine for now. It may be worth exploring the backport to 1.12 and _maybe_ 1.11, but probably not 1.10. Given the upcoming 1.14 release, that could provide a path to migrate entirely off Jenkins definitely this year if not in a couple of months.

As for the missing datapath tests, will you file issues to follow up on the reduction of coverage?

> Re-introduce the f-XX prefix?

I think that having focus groups would be helpful to be able to easily narrow down groups of tests. I glanced at the four datapath groups that you listed in the PR description, and I couldn't tell the difference at first look. I think that failing tests may result in similar assumptions amongst readers unless we add some sort of numeric identifier for the groups. I'd be in favour of reintroducing that.

I'll note my approval here though I'll skip the GitHub codeowners review since I didn't look closely at the implementation.

.github/workflows/conformance-ginkgo.yaml
pchaigno: Nevermind. I don't have a lot of time to check and compare the exact list, so let's forget about it and I'll send fixes when/if I need to debug a test.

jibi (approved):

PR #25714, opened 2023-05-26
Surface agent modules health from cilium status command

derailed:

      GitHub issue.

- Add hive module health injection to support for module health reporting (pulled from @joamaki prototype)
- Add swagger health apis
- Surface agent modules health via agent CLI

> There are dependencies here on these 2 PRs: https://github.com/cilium/cilium/pull/25725, https://github.com/cilium/cilium/pull/25662 

```release-note
Augments `cilium status` CLI to report on agent modules health status.
```


pkg/node/manager/manager.go
```diff
@@ -294,11 +299,13 @@ func (m *manager) backgroundSync(ctx context.Context) error {
 			m.mutex.RUnlock()
 			m.Iter(func(nh datapath.NodeHandler) {
 				nh.NodeValidateImplementation(entry.node)
+
 			})
 			entry.mutex.Unlock()
 
 			m.metricDatapathValidations.Inc()
 		}
+		m.healthReporter.OK("All good here! Nothing to see")
```
tommyp1ckles: We can drop this code from this PR, I trust that this is all working and will pick up any modules reporting a status in future PRs.

pkg/datapath/linux/probes/managed_neighbors.go
```diff
@@ -83,6 +83,9 @@ func haveManagedNeighbors() (outer error) {
 		return fmt.Errorf("failed to add dummy veth: %w", err)
 	}
 
+	if err := netlink.LinkAdd(veth); err != nil {
```
tommyp1ckles: Test code - should probably remove?

api/v1/openapi.yaml
```diff
@@ -1,33 +1,33 @@
 ---
-swagger: '2.0'
+swagger: "2.0"
```
tommyp1ckles: Did your editor making these changes? we should avoid committing auto-linting changes, especially if they're not codified in the codebase.

pkg/node/manager/manager.go
derailed: Indeed... Thanks Tom!

api/v1/openapi.yaml
derailed: Rats! it did - will update

api/v1/openapi.yaml
```diff
@@ -1202,6 +1221,33 @@ parameters:
     type: string
 
 definitions:
+  ModulesHealth:
+    description: Reports agent's modules health status
```
zacharysarah: Clarity nit
```suggestion
    description: Reports health status of agent's modules
```

zacharysarah (approved):
One nit to make it easier to differentiate ModulesHealth from ModuleHealth, otherwise LGTM from a docs perspective. Approving with the understanding that the change is required prior to merge.

pkg/node/manager/manager.go
```diff
@@ -294,6 +299,7 @@ func (m *manager) backgroundSync(ctx context.Context) error {
 			m.mutex.RUnlock()
 			m.Iter(func(nh datapath.NodeHandler) {
 				nh.NodeValidateImplementation(entry.node)
+
```
thorn3r: tiny nit: whitespace change here

pkg/datapath/linux/probes/managed_neighbors.go
```diff
@@ -83,9 +83,6 @@ func haveManagedNeighbors() (outer error) {
 		return fmt.Errorf("failed to add dummy veth: %w", err)
 	}
 
-	if err := netlink.LinkAdd(veth); err != nil {
-		return fmt.Errorf("failed to add dummy veth: %w", err)
-	}
```
thorn3r: Since this was for testing can we drop the change from both commits? I imagine this could cause problems during a bisect

api/v1/openapi.yaml
derailed: @zacharysarah Good catch. Thank you!

pkg/datapath/linux/probes/managed_neighbors.go
derailed: @thorn3r Thank you for the review! I'll axe it

rolinh (approved):
Vendor changes lgtm

thorn3r (approved):
lgtm! :+1:

qmonnet (approved):
Looks good from my side

tommyp1ckles:
Looks like ci-ginkgo is hitting [#25964](https://github.com/cilium/cilium/issues/25964)

tommyp1ckles:
@derailed FYI For the tests that run as Github Actions, you can just click on their link and do "retry all tests" most of the time.

For Jenkins tests, you can comment: slash + test-1.26-net-next to run a specific test suite to retry.  Its generally a bit unlikely that an entire test run won't have at least one flake 😆

daemon/cmd/daemon.go
```diff
@@ -216,7 +217,9 @@ type Daemon struct {
 	// just used to tie together some status reporting
 	cniConfigManager cni.CNIConfigManager
 
-	l2announcer *l2announcer.L2Announcer
+	l2announcer    *l2announcer.L2Announcer
+	healthProvider cell.Health
```
christarazi: Let's add a newline between these to represent the logical separation of these two fields

daemon/cmd/vitals.go
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package cmd
+
+import (
+	"time"
+
+	"github.com/go-openapi/runtime/middleware"
+	"k8s.io/apimachinery/pkg/util/duration"
```
christarazi, line 10: Can you squash the vendor commit into this commit since it's being used here and it's fairly small? This makes bisectability better as the new dependency is added with the implementation.

daemon/cmd/vitals.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/cilium/cilium/api/v1/models"
+	. "github.com/cilium/cilium/api/v1/server/restapi/daemon"
+	"github.com/cilium/cilium/pkg/hive/cell"
+)
+
+const naValue = "n/a"
```
christarazi: Nit: No need for this const if it's just used once

cilium/cmd/status.go
```diff
@@ -107,7 +107,7 @@ func statusDaemon() {
 			healthPkg.GetAndFormatHealthStatus(w, true, allHealth, healthLines)
 			healthPkg.GetAndFormatModulesHealth(w, client.Daemon, allHealth)
 		} else {
-			fmt.Fprint(w, "Cluster health:\t\tProbe disabled\n")
+			fmt.Fprint(w, "Cluster Health:\t\tProbe disabled\n")
```
christarazi: While I appreciate the intent of these changes, I don't particularly think they should be part of this PR and merged late into the release cycle. My short term concern is that we might have scripts somewhere relying on this. Longer term, without a style guide to point to, these changes are arbitrary and "subjective". Someone later down the line might come along and propose the effective inverse of these changes, :shrug:.

My proposal is to drop this commit for now.

Happy to discuss and pull in the right folks if we think a style guide for our CLI outputs to be more consistent / following a particular pattern is something we should pursue.

daemon/cmd/daemon.go
```diff
@@ -216,7 +217,9 @@ type Daemon struct {
...
+	healthReporter cell.HealthReporter
```
christarazi, line 225: Speaking of logical separations, what do you think of combining the reporter into the provider? Do we really need two separate cells? At a surface level, it seems reasonable that a health provider has a "reporter" component.

christarazi (changes requested):
Nice work!

I have a few comments regarding the PR. My other comment is that I'd like to see some motivation written in each commit msg describing the "why", rather than just the "what". This allows us to preserve the original intent in the history for better bisect and context sharing in general for code readers.

daemon/cmd/daemon.go
derailed: I think SCP usage here make sense as these are managing separate concerns.

cilium/cmd/status.go
derailed: Makes sense. I'll axe it. We have a bit of a `Frankenstein` theme going at the moment regarding the output with a strong majority to capitalize, hence this effort to normalize the output for consistency. 
That said, If scripts are scraping human readable console format while we have options to use structured data via yaml or json, we have other issues.

daemon/cmd/vitals.go
derailed: I am a bit confused with the heuristics to track dependencies. On previous commit I was ask to specifically keep dependencies update in separate commit. But here since it's small it's ok to roll it in?

daemon/cmd/daemon.go
derailed: @christarazi Thanks for the review. Makes sense!

daemon/cmd/vitals.go
christarazi: Ah I see. I'll share more on why I made this comment: as I reviewed commit by commit, it was difficult to tell why the "Update dependencies" commit was necessary because there was no context in the msg for why it's being included. Later, I noticed the new package was included in used in the "Add modules health report" commit. 

Because the diff of the new dependency was so small, I probably wouldn't have reviewed in if it was squashed into the latter commit. The package was already included indirectly as a dependency of Cilium and the main diff is that it's now a direct dependency (that's why only the `vendor/` changed). Given this, I'd say it's a very minor change that doesn't require a separate review (by being in its own commit). If we were including a new module (requiring changes to `go.mod`) then I'd say a separate commit makes sense. 

However, if the new module is small enough anyway, then you can get away with it being bundled. I'd say the main heuristic is what makes the most sense to reviewers. Should they review the new dependency? If so, make it separate with context in the commit msg. If the dependency is not controversial*, probably just include it with the implementation commit.

*: I'm hand waving a bit here for expediency purposes, but I realize we probably need to document better what are our expectations around vendor changes and commits.

api/v1/openapi.yaml
nathanjsweet: This isn't resolved. There are still a lot of spurious changes; please revert all of these redundant changes.

nathanjsweet (changes requested):

daemon/cmd/vitals.go
derailed: Thanks for the clarifications and explanations Chris!
 I thought since part of the review council are folks specifically in charge of checking for dependencies updates, it made sense for it to be in a separate commit so it can ease there review no matter how small. Tho no go.mod was changed /vendor did. One the other hand and as you pointed at, I makes sense for it to be part of the commit that warrants it.

api/v1/openapi.yaml
derailed: Rats! good catch. Tx Nate! tabs and valuable EOL spaces now restored to their former glory.

api/v1/client/daemon/daemon_client.go
```diff
@@ -211,6 +213,46 @@ func (a *Client) GetDebuginfo(params *GetDebuginfoParams, opts ...ClientOption)
...
+	if err != nil {
+		return nil, err
+	}
+	success, ok := result.(*GetHealthOK)
+	if ok {
+		return success, nil
+	}
+	// unexpected success response
+	// safeguard: normally, absent a default response, unknown success responses return an error above: so this is a codegen issue
+	msg := fmt.Sprintf("unexpected success response for GetHealth: API contract not enforced by server. Client expected to get an error, but got: %T", result)
```
asauber, line 252: Using `%#v` rather than `%T` here would be useful, as it would include both the type and value rather than just the type

api/v1/client/daemon/get_health_parameters.go
```diff
@@ -0,0 +1,159 @@
...
+// NewGetHealthParamsWithHTTPClient creates a new GetHealthParams object
+// with the ability to set a custom HTTPClient for a request.
+func NewGetHealthParamsWithHTTPClient(client *http.Client) *GetHealthParams {
+	return &GetHealthParams{
+		HTTPClient: client,
+	}
+}
+
+/*
+GetHealthParams contains all the parameters to send to the API endpoint
```
asauber, line 60: Comment formatting here looks like it was mangled.

api/v1/client/daemon/get_health_parameters.go
```diff
@@ -0,0 +1,159 @@
...
+
+	Typically these are written to a http.Request.
+*/
+type GetHealthParams struct {
+
+	/* Brief.
+
+	   Brief will return a brief representation of the Cilium status.
+
+	*/
```
asauber:
```suggestion
	// Brief is a brief representation of the Cilium status.
```

I'm aware that this is autogenerated, maybe there's a way to hint that this be formatted better.

api/v1/client/daemon/get_health_parameters.go
```diff
@@ -0,0 +1,159 @@
...
+// WriteToRequest writes these params to a swagger request
+func (o *GetHealthParams) WriteToRequest(r runtime.ClientRequest, reg strfmt.Registry) error {
+
+	if err := r.SetTimeout(o.timeout); err != nil {
+		return err
+	}
+	var res []error
+
+	if o.Brief != nil {
+
```
asauber, line 148: nit: extra newline

aanm (commented):
It's not clear to me why we can't extend the current `GetHealthz` API endpoint for this purpose given that `GetHealth` seems to have the same parameters as `GetHealthz`

api/v1/models/modules_health.go
```diff
@@ -0,0 +1,119 @@
...
+		if m.Modules[i] != nil {
+			if err := m.Modules[i].Validate(formats); err != nil {
+				if ve, ok := err.(*errors.Validation); ok {
+					return ve.ValidateName("modules" + "." + strconv.Itoa(i))
+				} else if ce, ok := err.(*errors.CompositeError); ok {
+					return ce.ValidateName("modules" + "." + strconv.Itoa(i))
+				}
+				return err
+			}
+		}
```
asauber, line 62:
```suggestion
		if m.Modules[i] == nil {
			continue
		}
		
		if err := m.Modules[i].Validate(formats); err != nil {
			if ve, ok := err.(*errors.Validation); ok {
				return ve.ValidateName("modules" + "." + strconv.Itoa(i))
			} else if ce, ok := err.(*errors.CompositeError); ok {
				return ce.ValidateName("modules" + "." + strconv.Itoa(i))
			}
			return err
		}
```

api/v1/models/modules_health.go
```diff
@@ -0,0 +1,119 @@
...
+
+	}
```
asauber, line 98:
```suggestion
		if m.Modules[i] == nil {
			continue
		}
		
		if err := m.Modules[i].ContextValidate(formats); err != nil {
			if ve, ok := err.(*errors.Validation); ok {
				return ve.ValidateName("modules" + "." + strconv.Itoa(i))
			} else if ce, ok := err.(*errors.CompositeError); ok {
				return ce.ValidateName("modules" + "." + strconv.Itoa(i))
			}
			return err
		}
```

api/v1/server/restapi/daemon/get_health_parameters.go
```diff
@@ -0,0 +1,82 @@
...
+// GetHealthParams contains all the bound params for the get health operation
+// typically these are obtained from a http.Request
+//
+// swagger:parameters GetHealth
+type GetHealthParams struct {
+
+	// HTTP Request Object
+	HTTPRequest *http.Request `json:"-"`
+
+	/*Brief will return a brief representation of the Cilium status.
```
asauber:
```suggestion
	/*Brief is a brief representation of the Cilium status.
```

api/v1/server/restapi/daemon/get_health_parameters.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// bindBrief binds and validates parameter Brief from header.
+func (o *GetHealthParams) bindBrief(rawData []string, hasKey bool, formats strfmt.Registry) error {
+	var raw string
+	if len(rawData) > 0 {
+		raw = rawData[len(rawData)-1]
+	}
+
+	// Required: false
```
asauber, line 69: I'm aware that this is autogenerated, do you know what this comment is in reference to?

api/v1/models/modules_health.go
asauber: I'm aware that this is auto-generated. Kind of a shame that the code is unnecessarily nested.

pkg/health/client/modules_test.go
```diff
@@ -0,0 +1,92 @@
...
+		u := uu[k]
+		t.Run(k, func(t *testing.T) {
+			w := bytes.NewBufferString("")
+			client.GetAndFormatModulesHealth(w, u.h, u.v)
+			assert.Equal(t, u.e, strings.TrimSpace(w.String()))
+		})
+	}
+}
+
+// Helpers...
```
asauber: The `...` makes me nervous 😅

asauber (approved):
Overall looks good, some nits and formatting questions.

tommyp1ckles:
@aanm 
> It's not clear to me why we can't extend the current GetHealthz API endpoint for this purpose given that GetHealth seems to have the same parameters as GetHealthz

It made sense to me to separate these, the default health endpoint already has a pretty large/ambiguous payload-size/scope - seems rational to avoid adding to that.

As well, this functionality is still in a "beta" state and thus is likely to be unstable, I see a possibility of: changing/deprecating/removing this endpoint in the future as the modular health feature evolves, which instability we don't want to have in the actual healthz endpoint.

api/v1/server/restapi/daemon/get_health_parameters.go
tommyp1ckles: I'm guessing it's this: https://github.com/cilium/cilium/pull/25714/files#diff-508a09ed0e66f45b2f1c78e3650eedf4de551b63150c6dac457298af46f9152bR67 ?

api/v1/client/daemon/daemon_client.go
derailed: Not sure if this would be that useful here given we are expecting a specific type that somehow did not come across the wire. Imho this is a clearer error that dumping the type and entire resp content which would not be that actionable?

pkg/health/client/modules_test.go
derailed: Not sure I get the reference??

api/v1/client/daemon/get_health_parameters.go
derailed: Good catch!

api/v1/models/modules_health.go
derailed: Right! short of PRing their repo not sure if there is much we can do here??

asauber: Yep

api/v1/client/daemon/daemon_client.go
asauber: You're right, I agree that %T is better.

api/v1/client/daemon/get_health_parameters.go
derailed: This is generated code

api/v1/server/restapi/daemon/get_health_parameters.go
derailed: Tx Andrew!

joestringer:
@tommyp1ckles , @christarazi and I had a bit of a chat out-of-band about this PR regarding the release-blocker status and usage of the PR.

If I understand correctly, right now this PR by itself will introduce some extra lines in every `cilium status` output saying something like:

```
....
Modules Health:
    no health payload detected
```

Then some extra lines that report that there's no modules registered, none healthy, none degraded, etc. By itself, that would confuse users. For this PR to make sense, I think we need at least one user of this functionality on top.

The current proposal is https://github.com/cilium/cilium/pull/25994, but unfortunately that interacts with ipsec which is being actively worked on at the moment, hence delaying that effort. Given this, I think it's more reasonable to defer this PR until after 1.14 is branched, and target this to land early during 1.15. We can continue to review & iterate on this up to that point so it's ready to merge as soon as feature freeze is lifted.

If we have a strong argument for why this PR + a related user could provide significant visibility/debuggability improvements, _maybe_ there's an argument to include it in 1.14 under the [backporting criteria](https://docs.cilium.io/en/latest/contributing/release/backports/#backport-criteria-for-current-minor-release) "Debug tool improvements", but my gut feeling is that this new feature is just not quite ready at the right time to ship in 1.14.

Either way, thanks for your work on this area and let's work on getting this into the first v1.15 snapshot so we can start to try out this new feature and get some feedback on it :)

aanm (approved):

tommyp1ckles:
@derailed test failures look legitimate

christarazi (approved):

christarazi:
@nathanjsweet Could you take another look?

nathanjsweet (approved):

christarazi:
@derailed I see that this PR is getting rebased quite often. It's not necessary to rebase unless there's a merge conflict. AFAIK, GitHub will signal if that's the case.

PR #25719, opened 2023-05-26
egressgw: switch to Cilium CLI connectivity tests

### LINKED ISSUE #24151: Move K8sDatapathEgressGatewayTest tests to cilium-cli's connectivity tests running on ci-datapath
This issue is about moving the egress gateway e2e tests from ginkgo to the new cilium-cli's datapath test suite.

The entire `K8sDatapathEgressGatewayTest` test suite lives in https://github.com/cilium/cilium/blob/master/test/k8s/egress.go.

This suite runs the same set of tests for each combination of the following configurations:
* tunneling: disabled/vxlan
* endpoint routes: disabled/enabled
* XDP: disabled/enabled

This might partially overlap with the different configurations that the `ci-datapath` workflow is running on: https://github.com/cilium/cilium/blob/f9d290c20ee9349da28bf8164ab8e7eb178079b0/.github/workflows/conformance-datapath.yaml#L121-L195 

Then, for each configuration, we run 3 group of tests:

* `no egress gw policy`: we run a generic connectivity test, both from the worker and the gateway node, with egress gateway enabled and without any policy in place
* `egress gw policy`: we run the same generic connectivity test together with a specific egress gateway test,  both from the worker and the gateway node, with egress gateway enabled and a few policies in place
* `egress gw policy with exclusion CIDRs`: similar to the previous one, but we specifically test the `excludedCIDRs` property

The generic connectivity test mentioned before: https://github.com/cilium/cilium/blob/666d6991d9d0a9759507737eaf0281e47671b802/test/k8s/egress.go#L177-L267

runs the following checks:
* TCP/UDP pod to pod connectivity works
* reply traffic for a connection initiated outside the cluster does not go through egress gateway (both nodeport services and pods)

The other egress gateway specific test mentioned before: https://github.com/cilium/cilium/blob/666d6991d9d0a9759507737eaf0281e47671b802/test/k8s/egress.go#L138-L175

tests that a connection from a pod matching an egress gateway policy is (or isn't if we are testing excluded CIDRs) SNATed through egress gateway

jibi:
Bump Cilium CLI in e2e conformance tests to v0.14.6 in order to pick the new egressgw connectivity tests and delete the ginkgo testsuite

Fixes: #24151


jibi:
green run https://github.com/cilium/cilium/actions/runs/5146050794/jobs/9264811851?pr=25719

viktor-kurchenko (approved):
LGTM.

brb (approved):

PR #24773, opened 2023-04-05
ipsec: Clean up stale XFRM policies and states

pchaigno:
First commit is a bit of refactoring. Second implements a proper function to update XFRM states. Third adds a catch-all default-drop policy to avoid leaking plain-text traffic during the cleanup. Fourth commit implements the cleanup itself.
See commit descriptions for details.

Fixes: https://github.com/cilium/cilium/pull/24030.
Fixes: https://github.com/cilium/cilium/issues/24780.
Updates: https://github.com/cilium/cilium/issues/24010.

```
Fix IPsec bug that causes connectivity issues across nodes when upgrading from a previous version of Cilium.
```

PR #24030, opened 2023-02-26
IPsec: Remove `IP_POOLS` logic

pchaigno:
This pull request removes the `IP_POOLS` logic in an attempt to make our IPsec logic as independent from IPAM modes as possible. These differences between IPAM modes have caused a lot of bugs in IPsec in the past.

The `IP_POOLS` logic is specific to AKS and EKS' IPAM modes, where we don't have one pod CIDR per node. Therefore, with those IPAM modes, we can't simply match on the destination pod CIDR in the kernel's IPsec configuration to know what outer destination IP address to use for the IPsec encapsulation. Instead, we rely on the ipcache to tell us what IP address to use. We tell the kernel to encapsulate with IP 0.0.0.0 and then rewrite the destination IP address to whatever the ipcache gave us.

This `IP_POOLS` logic is easy to implement, but fairly hackish and it means that the IPsec kernel configuration differs significantly between IPAM modes. This pull request replaces it with a new implementation.

In the new implementation, we rely on the recently introduced node IDs (cf. https://github.com/cilium/cilium/pull/23202). We install the same IPsec kernel configuration for all IPAM modes and use the packet mark to determine the destination node. The node ID is read from the ipcache before encryption and written into the packet mark, such that we will match the appropriate IPsec kernel config. and encapsulate with the right IP address.

Commit summary below. See commits for details.

1. Match on node ID from packet mark in XFRM. Doesn't handle AKS & EKS yet.
0. Document the limitation to 64k nodes for IPsec due to the use of the node IDs.
0. Free `skb->cb[4]` as we need it in a subsequent commit.
0. Remove `IP_POOLS` special case for AKS & EKS. This includes rewrite of the outer IP header after encryption.
0. Free the packet mark by using `skb->cb[4]` instead to pass the source security ID.
0. Write the node ID in the packet mark, to be matched against the XFRM state at encryption time.
0. Free packet mark on bpf_overlay recirculation, as we need it for the node IDs.
0. Rework the XFRM states and policies for AKS & EKS to have one XFRM state per remote nodes. This is the complement of commit 4.

PR #25724, opened 2023-05-26
ipsec: Fix `XfrmInNoStates` drops on ESK & AKS upgrades

pchaigno:
This pull request fixes IPsec packet drops of type `XfrmInNoStates` that would happen on EKS and AKS for the whole duration of the upgrade (until the last node is upgraded). The root of the issue is that we switched from using CiliumInternalIP to using NodeInternalIP for the IPsec encapsulation and that change isn't properly handled on the upgrade.

To avoid the `XfrmInNoStates` drops, this pull request introduces a 2-steps upgrade process. First, after the usual upgrade, both CiliumInternalIPs and NodeInternalIPs encapsulations will be accepted on ingress, but NodeInternalIPs will still be used for the encapsulation on egress. Then, in a second time, users can switch to `use-cilium-internal-ip-for-ipsec=true` (defaults false) to start using CiliumInternalIPs.

:warning: Note this pull request doesn't fix other drops of type `XfrmOutPolBlock` also happening during the IPsec + EKS/AKS upgrade. Those will be addressed in a separate PR.

### How was this tested?

I performed two types of tests: first with a normal upgrade process and second with a split cluster:
1. The usual upgrade path was tested on 2 clusters of 20 nodes each, with EKS and GKE. The upgrade was performed from Cilium v1.13.0 to this patchset, with the drop-sensitive `migrate-svc` application running during the upgrade. The two-steps upgrade was carried out and, after each step, I checked that no `XfrmInNoStates` drops were reported. The `migrate-svc` application did have drops during the first step because of the aforementioned `XfrmOutPolBlock` drops. It did not have any additional drops afterward. Our connectivity tests suite also passed after each step.
2. I deployed a 3-nodes EKS cluster with a mix of Cilium versions to simulate the impact of a prolonged upgrade. Two nodes ran Cilium v1.13.0, while the third ran this patchset backported on the v1.13 branch. I ran the connectivity tests and confirmed that no drops, neither `XfrmInNoStates` nor `XfrmOutPolBlock`, were reported.

Fixes: https://github.com/cilium/cilium/pull/24773.
```release-note
Fix a bug that would cause connectivity drops of type XfrmInNoStates on upgrade when IPsec is enabled with ENI or Azure IPAM mode.
```

### DIFF
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -1052,11 +1052,15 @@ func (n *linuxNodeHandler) enableIPsecIPv4(newNode *nodeTypes.Node, zeroMark boo
 		remoteNodeID := n.allocateIDForNode(newNode)
 
 		if n.subnetEncryption() {
-			localIP, err = getV4LinkLocalIP()
-			if err != nil {
-				log.WithError(err).Error("Failed to get local IPv4 for IPsec configuration")
+			// Check if we should use the NodeInternalIPs instead of the
+			// CiliumInternalIPs for the IPsec encapsulation.
+			if !option.Config.UseCiliumInternalIPForIPsec {
+				localIP, err = getV4LinkLocalIP()
+				if err != nil {
+					log.WithError(err).Error("Failed to get local IPv4 for IPsec configuration")
+				}
+				remoteIP = newNode.GetNodeIP(false)
 			}
-			remoteIP = newNode.GetNodeIP(false)
 
 			for _, cidr := range n.nodeConfig.IPv4PodSubnets {
 				spi, err = ipsec.UpsertIPsecEndpoint(wildcardCIDR, cidr, localIP, remoteIP, remoteNodeID, ipsec.IPSecDirOut, zeroMark)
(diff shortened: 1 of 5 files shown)


pkg/datapath/linux/node.go
```diff
@@ -962,101 +993,164 @@ func (n *linuxNodeHandler) enableIPsec(newNode *nodeTypes.Node) {
...
+			localNodeInternalIP, err := getV6LinkLocalIP()
+			if err != nil {
+				log.WithError(err).Error("Failed to get local IPv6 for IPsec configuration")
+			}
+
+			for _, cidr := range n.nodeConfig.IPv6PodSubnets {
+				spi, err := ipsec.UpsertIPsecEndpoint(wildcardCIDR, cidr, localIP, wildcardIP, 0, ipsec.IPSecDirIn, zeroMark)
+				upsertIPsecLog(err, "in CiliumInternalIPv6", wildcardCIDR, cidr, spi)
+
+				spi, err = ipsec.UpsertIPsecEndpoint(wildcardCIDR, cidr, localNodeInternalIP, wildcardIP, 0, ipsec.IPSecDirIn, zeroMark)
```
jschwinger233, line 1116: I was wondering if we can skip generating ingress policy for native node IP when `UseCiliumInternalIPForIPsec` is true, but It looks unsafe during upgrade, so be it.

jschwinger233 (approved):

pkg/datapath/linux/node.go
pchaigno: Yep, we need this for the upgrade. We can clean it up later, in v1.15, once we're sure everyone has upgraded from the old IPs.

mhofstetter (approved):
config changes (agent & CLI) look good to me!

PR #25728, opened 2023-05-26
Add support of storing the caBundle in a secret

farcaller:
      GitHub issue.

This allows to e.g. pull it from vault via external-secrets (as ES cannot write to a ConfigMap).

```release-note
Allow to use a Secret for the caBundle
```

### DIFF
--- a/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
+++ b/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
@@ -845,7 +845,7 @@ spec:
               - key: ca.crt
                 path: common-etcd-client-ca.crt
           {{- else }}
-          - configMap:
+          - {{ .Values.tls.caBundle.useSecret | ternary "secret" "configMap" }}:
               name: {{ .Values.tls.caBundle.name }}
               optional: true
               items:
@@ -904,7 +904,7 @@ spec:
               - key: ca.crt
                 path: client-ca.crt
           {{- else }}
-          - configMap:
+          - {{ .Values.tls.caBundle.useSecret | ternary "secret" "configMap" }}:
               name: {{ .Values.tls.caBundle.name }}
               optional: true
               items:
--- a/install/kubernetes/cilium/values.yaml.tmpl
+++ b/install/kubernetes/cilium/values.yaml.tmpl
@@ -2061,6 +2061,9 @@ tls:
     # -- Entry of the ConfigMap containing the CA trust bundle.
     key: ca.crt
 
+    # -- Use a Secret instead of a ConfigMap.
+    useSecret: false
+
     # If uncommented, creates the ConfigMap and fills it with the specified content.
     # Otherwise, the ConfigMap is assumed to be already present in .Release.Namespace.
     #
--- a/Documentation/helm-values.rst
+++ b/Documentation/helm-values.rst
@@ -2631,7 +2631,7 @@
    * - tls
      - Configure TLS configuration in the agent.
      - object
-     - ``{"ca":{"cert":"","certValidityDuration":1095,"key":""},"caBundle":{"enabled":false,"key":"ca.crt","name":"cilium-root-ca.crt"},"secretsBackend":"local"}``
+     - ``{"ca":{"cert":"","certValidityDuration":1095,"key":""},"caBundle":{"enabled":false,"key":"ca.crt","name":"cilium-root-ca.crt","useSecret":false},"secretsBackend":"local"}``
    * - tls.ca
      - Base64 encoded PEM values for the CA certificate and private key. This can be used as common CA to generate certificates used by hubble and clustermesh components. It is neither required nor used when cert-manager is used to generate the certificates.
      - object
@@ -2651,7 +2651,7 @@
    * - tls.caBundle
      - Configure the CA trust bundle used for the validation of the certificates leveraged by hubble and clustermesh. When enabled, it overrides the content of the 'ca.crt' field of the respective certificates, allowing for CA rotation with no down-time.
      - object
-     - ``{"enabled":false,"key":"ca.crt","name":"cilium-root-ca.crt"}``
+     - ``{"enabled":false,"key":"ca.crt","name":"cilium-root-ca.crt","useSecret":false}``
    * - tls.caBundle.enabled
      - Enable the use of the CA trust bundle.
      - bool
@@ -2664,6 +2664,10 @@
      - Name of the ConfigMap containing the CA trust bundle.
      - string
      - ``"cilium-root-ca.crt"``
+   * - tls.caBundle.useSecret
+     - Use a Secret instead of a ConfigMap.
+     - bool
+     - ``false``
    * - tls.secretsBackend
      - This configures how the Cilium agent loads the secrets used TLS-aware CiliumNetworkPolicies (namely the secrets referenced by terminatingTLS and originatingTLS). Possible values:   - local   - k8s
      - string
--- a/Documentation/spelling_wordlist.txt
+++ b/Documentation/spelling_wordlist.txt
@@ -1127,6 +1127,7 @@ uptime
 uretprobes
 url
 useAPIServer
+useSecret
 username
 userspace
 userspaceFallback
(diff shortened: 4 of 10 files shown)


sayboras:
:wave: friendly ping if you have a chance to perform the above. Also, would you mind to rebase with main branch for latest CI configuration.

farcaller:
@sayboras thanks for a ping! I got a bit distracted by my cilium setup breaking down in new and exciting ways but I'm back to this and I have a working solution that I just need to rewrite this nix into helm yamls: 😄 

```nix
{
      patchCM = object:
        let
          volumes = object.spec.template.spec.volumes;
          projected = builtins.filter (v: builtins.hasAttr "projected" v) volumes;
          nonProjected = builtins.filter (v: builtins.hasAttr "projected" v == false) volumes;
          reprojected = builtins.map
            (p: lib.recursiveUpdate p (
              let
                caSources = builtins.filter (s: (s.configMap or { name = ""; }).name == "cilium-ca") p.projected.sources;
                nonCaSources = builtins.filter (s: (s.configMap or { name = ""; }).name != "cilium-ca") p.projected.sources;
                patchedSources = builtins.map
                  (s: {
                    secret = {
                      items = [{ key = "certificate"; path = (builtins.head s.configMap.items).path; }];
                      name = "cilium-ca";
                    };
                  })
                  caSources;
              in
              {
                projected.sources = nonCaSources ++ patchedSources;
              }
            ))
            projected;
        in
        (lib.recursiveUpdate object {
          spec.template.spec.volumes = nonProjected ++ reprojected;
        });

      patchOn = cond: object:
        if (cond object) then patchCM object
        else object;

      patchCiliumDS = patchOn (object: (object.kind == "DaemonSet" && object.metadata.name == "cilium"));
      patchApiServerDeployment = patchOn (object: (object.kind == "Deployment" && object.metadata.name == "clustermesh-apiserver"));
      patchHubbleRelayDeployment = patchOn (object: (object.kind == "Deployment" && object.metadata.name == "hubble-relay"));
      patchHubbleUIDeployment = patchOn (object: (object.kind == "Deployment" && object.metadata.name == "hubble-ui"));
}
```

joamaki:
Heading on vacation so unassigned myself and picked @tommyp1ckles from sig-k8s by lottery (`sort -R sig-k8s | head -1`).

joestringer:
cc @cilium/sig-hubble I think some of you folks have spent more time with Cilium's secret management previously and could potentially help advise on the solution here.

gandro:
Also cc @giorio94 who added the CA bundle feature.

To me, the overall motivation "this allows to e.g. pull it from vault via external-secrets (as ES cannot write to a ConfigMap)." sounds very reasonable. Since we have not shipped the CA bundle feature in a release yet, I'd even say let's always make it a secret if that makes the CA bundle easier to manage (and I'm not a big fan of having a resource either be a configmap or a secret based on a Helm value).

It was also the case that before we added the CA bundle, the CA was already part of a secret, so that should also not be a problem. But I think it would be interesting to know from @giorio94 if there is a particular reason why the CA bundle is a ConfigMap, or if the intent just was that "it's a public key, so it doesn't need to be a secret".

I'm marking this as a release blocker for now, since I think this is something that we can quite easily resolve now, where it will become harder to resolve once we release the CA bundle feature.

giorio94:
> It was also the case that before we added the CA bundle, the CA was already part of a secret, so that should also not be a problem. But I think it would be interesting to know from @giorio94 if there is a particular reason why the CA bundle is a ConfigMap, or if the intent just was that "it's a public key, so it doesn't need to be a secret".

No specific reason beside the fact that it is not a "secret" value as you mentioned, and to mimic the kubernetes behavior (specifically the `kube-root-ca.crt` ConfigMap, indeed also the default name was picket to look like that). But I don't personally have any objection to making that a secret if it simplifies the CABundle management process.

install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
```diff
@@ -843,7 +843,7 @@ spec:
               - key: ca.crt
                 path: common-etcd-client-ca.crt
           {{- else }}
-          - configMap:
+          - {{ if .Values.tls.caBundle.useSecret }}secret{{ else }}configMap{{ end }}:
```
gandro: So there are multiple uses of `.Values.tls.caBundle.enabled` in our Helm charts. I think you will have to change all of them to use this pattern.

gandro (changes requested):
Please run `make -C install/kubernetes`, it generates the `install/kubernetes/cilium/README.md` which needs to be updated if you add or change Helm values.

gandro:
Thanks @giorio94 for the quick response! 

I did a quick look at trust-manager, and it seems to only support ConfigMaps unfortunately https://cert-manager.io/docs/projects/trust-manager/api-reference/#bundlespectarget 

So maybe we do need to use flag that switches between ConfigMap and Secrets :disappointed: 

In that case I'll remove the release-blocker label again, since the change with `useSecret` can be done in a backwards compatible manner and I think this PR still needs a bit of iteration.

giorio94:
> I did a quick look at trust-manager, and it seems to only support ConfigMaps unfortunately https://cert-manager.io/docs/projects/trust-manager/api-reference/#bundlespectarget

Oh right, I actually forgot to double-check it. Thanks!

farcaller:
(sorry for a sudden pull of all the people. I rebased from a shadow clone 😞 )

gandro:
Another CI failure:

```
 Validating documentation (syntax, spelling)...

Please fix the following spelling mistakes:
* Documentation/helm-values.rst:2667: (useSecret)  tls.caBundle.useSecret
* Documentation/helm-values.rst:2667: (useSecret)  tls.caBundle.useSecret

If the words are not misspelled, run:
Documentation/update-spelling_wordlist.sh useSecret 
```

chancez (commented):
LGTM in terms of helm changes.

sayboras (approved):
The changes lgtm ✅

install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
```diff
@@ -845,7 +845,7 @@ spec:
...
```
kaworu: nitpicking:
```suggestion
          - {{ .Values.tls.caBundle.useSecret | ternary "secret" "configMap" }}:
```

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -2061,6 +2061,9 @@ tls:
     # -- Entry of the ConfigMap containing the CA trust bundle.
     key: ca.crt
 
+    # -- Use a Secret instead of a ConfigMap
```
kaworu: nitpicking

```suggestion
    # -- Use a Secret instead of a ConfigMap.
```

kaworu (approved):
Thanks for the patch @farcaller!

Couple of cosmetic comments, but other than that LGTM.

install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
```diff
@@ -845,7 +845,7 @@ spec:
               - key: ca.crt
                 path: common-etcd-client-ca.crt
           {{- else }}
-          - configMap:
+          - {{ .Values.tls.caBundle.useSecret | ternary "secret" "configMap" }}::
```
kaworu: I think there is an extra column here
```suggestion
          - {{ .Values.tls.caBundle.useSecret | ternary "secret" "configMap" }}:
```

kaworu (changes requested):
Please squash all the PR's commits into one on the way!

install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
farcaller: Derp, thanks for catching it!

kaworu (approved):

ti-mo:
One more: https://github.com/cilium/cilium/actions/runs/5354804806/jobs/9712313069?pr=25728
```
-     - Use a Secret instead of a ConfigMap
+     - Use a Secret instead of a ConfigMap.
```

farcaller:
> One more

Done (thought that one was supposed to be re-generated from `make -C install/kubernetes`)

gandro (approved):

ldelossa:
All required tests passing. Merging.

michi-covalent (approved):

qmonnet (approved):
“docs” change looks good. Thanks!

PR #25735, opened 2023-05-28
ipsec: Fix `XfrmOutPolBlock` drops on upgrades

pchaigno:
This pull request fixes IPsec drops of type `XfrmOutPolBlock` that would happen with all IPAM modes during upgrades. The root of the issue is that we would delete the old XFRM OUT policies and/or states instead of leaving them around for the duration of the upgrade.

This pull request makes various changes to ensure that we smoothly transition from the old IPsec implementation to the new. After this pull request, we shouldn't have any packet drops on upgrades when using ENI or Azure IPAM modes. Otherwise (for e.g. cluster-pool IPAM mode), it is theoretically possible to have a few drops of type `XfrmOutNoStates`; those however haven't been noticed in any of the upgrade tests. There is more information on those `XfrmOutNoStates` drops in the last commit of this pull request.

### How was this tested?

The usual upgrade path from v1.13.0 to this patchset was tested on two clusters of 20 nodes each, on EKS and GKE. Before the upgrade, our drop-sensitive migrate-svc application was deployed with 50 clients and 30 backends. That application sends a continuous stream of messages back and forth and fails whenever a message is dropped.

After the upgrade to this patchset, I verified that:
- All IPsec error counters were at zero.
- None of the pods from the migrate-svc application had crashed.
- The connectivity tests passed on the upgraded cluster.
- No related errors or warnings were reported in the agent logs.

Fixes: https://github.com/cilium/cilium/pull/24773.
```release-note
Fix a bug that would cause connectivity drops of type XfrmOutPolBlock on upgrade when IPsec is enabled.
```

### DIFF
--- a/daemon/cmd/state.go
+++ b/daemon/cmd/state.go
@@ -18,6 +18,7 @@ import (
 	"github.com/cilium/cilium/pkg/controller"
 	"github.com/cilium/cilium/pkg/endpoint"
 	"github.com/cilium/cilium/pkg/ipam"
+	ipamOption "github.com/cilium/cilium/pkg/ipam/option"
 	slim_corev1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/core/v1"
 	"github.com/cilium/cilium/pkg/k8s/watchers/resources"
 	"github.com/cilium/cilium/pkg/labels"
@@ -304,7 +305,39 @@ func (d *Daemon) regenerateRestoredEndpoints(state *endpointRestoreState) (resto
 		}
 	}
 
+	if option.Config.EnableIPSec &&
+		(option.Config.IPAM == ipamOption.IPAMENI || option.Config.IPAM == ipamOption.IPAMAzure) {
+		// If IPsec is enabled on EKS or AKS, we need to restore the host
+		// endpoint before any other endpoint, to ensure a dropless upgrade.
+		// This code can be removed in v1.15.
+		// This is necessary because we changed how the IPsec encapsulation is
+		// done. In older version, bpf_lxc would pass the outer destination IP
+		// via skb->cb to bpf_host which would write it to the outer header.
+		// In newer versions, the header is written by the kernel XFRM
+		// subsystem and bpf_host must therefore not write it. To allow for a
+		// smooth upgrade, bpf_host has been updated to handle both cases. But
+		// for that to succeed, it must be reloaded first, before the bpf_lxc
+		// programs stop writing the IP into skb->cb.
+		for _, ep := range state.restored {
+			if ep.IsHost() {
+				log.WithField(logfields.EndpointID, ep.ID).Info("Successfully restored endpoint. Scheduling regeneration")
+				if err := ep.RegenerateAfterRestore(); err != nil {
+					log.WithField(logfields.EndpointID, ep.ID).WithError(err).Debug("error regenerating restored host endpoint")
+					epRegenerated <- false
+				} else {
+					epRegenerated <- true
+				}
+				break
+			}
+		}
+	}
+
 	for _, ep := range state.restored {
+		if ep.IsHost() && option.Config.EnableIPSec &&
+			(option.Config.IPAM == ipamOption.IPAMENI || option.Config.IPAM == ipamOption.IPAMAzure) {
+			// The host endpoint was handled above.
+			continue
+		}
 		log.WithField(logfields.EndpointID, ep.ID).Info("Successfully restored endpoint. Scheduling regeneration")
 		go func(ep *endpoint.Endpoint, epRegenerated chan<- bool) {
 			if err := ep.RegenerateAfterRestore(); err != nil {
--- a/pkg/datapath/linux/ipsec/ipsec_linux.go
+++ b/pkg/datapath/linux/ipsec/ipsec_linux.go
@@ -61,7 +61,8 @@ const (
 	// encoded in a XfrmMark
 	ipSecXfrmMarkSPIShift = 12
 
-	defaultDropPriority = 100
+	defaultDropPriority      = 100
+	oldXFRMOutPolicyPriority = 50
 )
 
 type ipSecKey struct {
@@ -380,7 +381,7 @@ func IpSecReplacePolicyFwd(dst *net.IPNet, tmplDst net.IP) error {
 //
 // We do need to match on the mark because there is also traffic flowing
 // through XFRM that we don't want to encrypt (e.g., hostns traffic).
-func IPsecDefaultDropPolicy(ipv6 bool) (err error) {
+func IPsecDefaultDropPolicy(ipv6 bool) error {
 	defaultDropPolicy := defaultDropPolicyIPv4
 	family := netlink.FAMILY_V4
 	if ipv6 {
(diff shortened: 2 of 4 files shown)


jschwinger233 (approved):
The way to make xfrm states coexist is marvelous! 
Just a newbee question: we coexist xfrm states, but update xfrm policies directly, and upgrade test still passes, so where do we use xfrm policy?

pchaigno:
> Just a newbee question: we coexist xfrm states, but update xfrm policies directly, and upgrade test still passes, so where do we use xfrm policy?

Actually, we don't update the XFRM policies directly. In the second commit, we deprioritize the old XFRM policies such that they can coexist with the new ones added as part of the upgrade. I've updated the commit descriptions to include example of how the XFRM configs will change.

If a packet comes into the XFRM subsystem, it will first be matched against the new XFRM policies with marks of the form 0xXXXXYe00/0xffffff00. If that doesn't match because the packet is still using the old mark format (because bpf_lxc hasn't been reloaded yet), then it will match against the deprioritized XFRM policies with marks of the form 0xYe00/0xff00.

Once we've matched against a policy, it will search for the matching XFRM state using the IP addresses from the policy's template. For example, with the below policies, it will look for an XFRM state with `src 10.24.1.77 dst 10.24.2.207` that also matches the packet mark.

```
     src 10.24.1.0/24 dst 10.24.2.0/24
          dir out priority 50
          mark 0x3e00/0xff00
          tmpl src 10.24.1.77 dst 10.24.2.207
              proto esp spi 0x00000003 reqid 1 mode tunnel
```

pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -380,88 +381,44 @@ func IpSecReplacePolicyFwd(dst *net.IPNet, tmplDst net.IP) error {
...
 	})
+
+	return err
 }
 
-func removeStalePolicies(family int) {
+// Lowers the priority of the old XFRM OUT policy. We rely on the mark mask to
+// identify it. By lowering the priority, we will allow the new XFRM OUT
+// policies to take precedence. We cannot simply remove and replace the old
+// XFRM OUT configs because that would cause traffic interruptions on upgrades.
```
tommyp1ckles, line 445: Question: are these policies just left like this forever - could we not clean them up after some time?

I guess we'd have to know that all nodes have had their host endpoints reloaded before doing that...

(I'm wondering if leaving old policies around might provide other places for issues during future version upgrades

daemon/cmd/state.go
```diff
@@ -304,7 +305,39 @@ func (d *Daemon) regenerateRestoredEndpoints(state *endpointRestoreState) (resto
 		}
 	}
 
+	if option.Config.EnableIPSec &&
+		(option.Config.IPAM == ipamOption.IPAMENI || option.Config.IPAM == ipamOption.IPAMAzure) {
+		// If IPsec is enabled on EKS or AKS, we need to restore the host
+		// endpoint before any other endpoint, to ensure a dropless upgrade.
+		// This code can be removed in v1.15.
```
tommyp1ckles, line 312: Is there a specific PR(s) that allows this? Probably worth specifying those.

daemon/cmd/state.go
```diff
@@ -304,7 +305,39 @@ func (d *Daemon) regenerateRestoredEndpoints(state *endpointRestoreState) (resto
...
+		// In newer versions, the header is written by the kernel XFRM
+		// subsystem and bpf_host must therefore not write it. To allow for a
+		// smooth upgrade, bpf_host has been updated to handle both cases. But
+		// for that to succeed, it must be reloaded first, before the bpf_lxc
+		// programs stop writing the IP into skb->cb.
+		for _, ep := range state.restored {
+			if ep.IsHost() {
+				log.WithField(logfields.EndpointID, ep.ID).Info("Successfully restored endpoint. Scheduling regeneration")
+				if err := ep.RegenerateAfterRestore(); err != nil {
+					log.WithField(logfields.EndpointID, ep.ID).WithError(err).Debug("error regenerating restored host endpoint")
```
tommyp1ckles, line 325: nit: Given that the downstream endpoints rely on this being in place, this seems to warrant at least an error log?

pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -197,16 +203,47 @@ func xfrmStateReplace(new *netlink.XfrmState) error {
...
+		//
+		// Thus, we temporarily remove the old, conflicting XFRM state and
+		// re-add it in a defer. In between the removal of the old state and
+		// the addition of the new, we can have a packet drops due to the
+		// missing state. These drops should be limited to the specific node
+		// pair we are handling here and the window during which they can
+		// happen should be really small. This is also specific to the upgrade
+		// and can be removed in v1.15. Finally, this shouldn't happen with ENI
+		// and Azure IPAM modes because they don't have such conflicting states.
+		if xfrmIPEqual(s.Src, new.Src) && xfrmIPEqual(s.Dst, new.Dst) &&
```
tommyp1ckles, line 224: maybe a silly question - do we not need to verify that the state dir is out in this case, or are we just going by the src/dst?

tommyp1ckles:
Question: are the prioritized policies just left like this forever - could we not clean them up after some time?

I guess we'd have to know that all nodes have had their host endpoints reloaded before doing that...

(I'm wondering if leaving old policies around might provide other places for issues during future version upgrades

tommyp1ckles:
Agent endpoint changes seem fine to me, just a few questions, I'll take another look tomorrow morning.

pkg/datapath/linux/ipsec/ipsec_linux.go
pchaigno: > Question: are these policies just left like this forever - could we not clean them up after some time?
> I guess we'd have to know that all nodes have had their host endpoints reloaded before doing that...

I'm planning to clean them for v1.15. That way we will be sure noone needs them to upgrade from a version of Cilium using the old XFRM OUT configs.

> (I'm wondering if leaving old policies around might provide other places for issues during future version upgrades

I don't think so because they are deprioritized and newer policies will therefore match first. If we don't change the XFRM configs again before v1.15, then upgrades should have no impact on this. If we do change the XFRM configs, then we'll have to take this into account; it's also why I left some space between priorities 0 (new), 50 (old), and 100 (catch-all).

daemon/cmd/state.go
pchaigno: The code removal isn't being enabled by any specific PR. We can remove in v1.15 because by then we will be sure that nobody needs to up/downgrade from/to a version of Cilium with the old XFRM configs. The old XFRM configs only exist in v1.13 and prior versions.

pkg/datapath/linux/ipsec/ipsec_linux.go
pchaigno: States don't have an explicit `dir` field. We know they are OUT states from the mark (includes the encrypt bit, 0xe00) and from the IPs (the source IP is the local IP). So comparing IP addresses and marks is enough here.

daemon/cmd/state.go
pchaigno: It's correct that the host endpoint regeneration is a critical part of the agent startup, but that was already the case before this PR. The reason why the existing code also had only a debug log message is simply because the actual errors are logged elsewhere, in the called function (see `Endpoint.Regenerate()`).

jibi (approved):

pkg/datapath/linux/ipsec/ipsec_linux.go
tommyp1ckles: ah right, missed that it was state, not policy.

tommyp1ckles (approved):

PR #25737, opened 2023-05-29
CODEOWNERS: Assign pkg/slices to sig-foundations

pippolo84:
Assign pkg/slices to sig-foundations


bimmlerd (approved):
yay more reviews ;)

PR #19753, opened 2022-05-09
bpf/nat: fix current behavior that is silently ignoring errors in a revSNAT context

sahid:
The PR is fixing current behavior of nodeport in context of a revSNAT that is silently ignoring errors to process them to the main path anyway.

- In the first patch we remove usage of DROP_NAT_UNSUPP_PROTO in favor of NAT_PUNT_TO_STACK considering that we want to be conservative in such situation.
- In the second patch we we have determined errors that can happen from a revSNAT
point of view to only accept those that are legitimate.
- The last one clarify comment.

PR #25740, opened 2023-05-29
bpf: nat: tolerate unhandled protocol types in revSNAT path

julianwiedmann:
In the past we would just pass up packets with unknown protocols, and let the next layer (or the stack) deal with them.

https://github.com/cilium/cilium/pull/19753 recently added error handling for the revSNAT paths, so that we can catch errors while eg. rewriting the packet headers.

But we want to preserve the old behaviour of tolerating unknown protocol types. So return NAT_PUNT_TO_STACK from a few additional ICMP paths.

### DIFF
--- a/bpf/lib/nat.h
+++ b/bpf/lib/nat.h
@@ -1152,7 +1152,7 @@ snat_v4_rev_nat_handle_icmp_frag_needed(struct __ctx_buff *ctx, __u64 off)
 		tuple.dport = identifier;
 		break;
 	default:
-		return DROP_UNKNOWN_L4;
+		return NAT_PUNT_TO_STACK;
 	}
 	state = snat_v4_lookup(&tuple);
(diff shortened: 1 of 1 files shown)


gentoo-root (approved):

julianwiedmann:
https://jenkins.cilium.io/job/Cilium-PR-K8s-1.25-kernel-4.19/208/ failed with https://github.com/cilium/cilium/issues/24961.

julianwiedmann:
/test-1.25-4.19

julianwiedmann:
> @julianwiedmann Too late to the party, but do you plan to do the same for `snat_v4_nat()`?

We should make the same change, but not sure whether I will get to it. It's also not quite as trivial (as we should at least consider whether all types of traffic are ok to leave the node untranslated).

PR #25340, opened 2023-05-09
bpf: Handle fragments in SNAT flows

gentoo-root:
The code that extracts ports in snat_v4_nat and snat_v4_rev_nat doesn't take into account that the packet may be a second or further fragment. In this case, garbage will be read instead of the real port numbers. Fix this by using the existing ipv4_ct_extract_l4_ports that takes into account fragmentation.

This change makes the behavior of the switch inside the aforementioned functions closer to ct_extract_ports4.

Consider fragments in snat_v4_rewrite_ingress, avoid rewriting ports if it's not the first fragment.

Fixes: 14a653ad4aac ("bpf/nat: introduce snat_v4_nat() and snat_v4_rev_nat functions")

This should be applied on top of #25112.

```release-note
Handle IPv4 fragments in SNAT flows correctly.
```

PR #25741, opened 2023-05-29
bpf: nat: consistently use has_l4_header in IPv4 SNAT path

julianwiedmann:
Clean up a redundant IPv4 header validation, and make the IPv4 SNAT code a bit more consistent.

Note that this won't help to actually fix IPv4 fragmentation vs SNAT. That's going to land via https://github.com/cilium/cilium/pull/25340.

### DIFF
--- a/bpf/lib/nat.h
+++ b/bpf/lib/nat.h
@@ -938,7 +938,8 @@ skip_egress_gateway:
 }
 
 static __always_inline __maybe_unused int
-snat_v4_nat_handle_icmp_frag_needed(struct __ctx_buff *ctx, __u64 off)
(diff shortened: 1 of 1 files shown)


bpf/lib/nat.h
```diff
@@ -1044,6 +1040,7 @@ snat_v4_nat(struct __ctx_buff *ctx, const struct ipv4_nat_target *target, __s8 *
 		return DROP_INVALID;
 
 	snat_v4_init_tuple(ip4, NAT_DIR_EGRESS, &tuple);
+	has_l4_header = ipv4_has_l4_header(ip4);
```
gentoo-root, line 1043: This will be replaced by ipv4_ct_extract_l4_ports in #25340.

gentoo-root (approved):
Nice cleanup!

PR #25742, opened 2023-05-29
bpf: dsr: fix typo in tail_nodeport_dsr_ingress_ipv4()

julianwiedmann:
This really should be a semicolon ...


gentoo-root (approved):
Ouch! Due to that typo, has_l4_header would always be false :scream:

PR #25744, opened 2023-05-29
ipsec: Reinitialize IPsec for new devices in ENI mode

joamaki:
In ENI mode new network devices are added as more IPs are requested. For IPsec support we
require the bpf_network program to be loaded on any device that receives the IPsec traffic
to mark the packets for decryption. 

Since Cilium uses the IP of the `cilium_host` device for the IPsec traffic the ENI device hosting
the IP requires bpf_network to be loaded onto it and since the `cilium_host` IP can change, this
PR watches for new network devices and reinitializes IPsec to make sure the `bpf_network` program
is loaded onto all used non-virtual devices.

```release-note
Fix bug affecting EKS installations with IPsec encryption enabled, where Cilium wouldn't attach its IPsec BPF program to new ENI interfaces, resulting in connectivity loss between pods on remote nodes.  
```

### DIFF
--- a/pkg/datapath/loader/netlink.go
+++ b/pkg/datapath/loader/netlink.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"time"
 
 	"github.com/vishvananda/netlink"
 	"golang.org/x/sys/unix"
@@ -17,6 +18,8 @@ import (
 
 	"github.com/cilium/cilium/pkg/bpf"
 	"github.com/cilium/cilium/pkg/defaults"
+	"github.com/cilium/cilium/pkg/inctimer"
+	ipamOption "github.com/cilium/cilium/pkg/ipam/option"
 	"github.com/cilium/cilium/pkg/mac"
 	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/sysctl"
@@ -328,4 +331,75 @@ func SetupBaseDevice(mtu int) (netlink.Link, netlink.Link, error) {
 	}
 
 	return linkHost, linkNet, nil
+}
+
+// reloadIPSecOnLinkChanges subscribes to link changes to detect newly added devices
+// and reinitializes IPsec on changes. Only in effect for ENI mode in which we expect
+// new devices at runtime.
+func (l *Loader) reloadIPSecOnLinkChanges() {
+	// settleDuration is the amount of time to wait for further link updates
+	// before proceeding with reinitialization. This avoids back-to-back
+	// reinitialization when multiple link changes are made at once.
+	const settleDuration = 1 * time.Second
+
+	if !option.Config.EnableIPSec || option.Config.IPAM != ipamOption.IPAMENI {
+		return
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	updates := make(chan netlink.LinkUpdate)
+
+	if err := netlink.LinkSubscribe(updates, ctx.Done()); err != nil {
+		log.WithError(err).Fatal("Failed to subscribe for link changes")
+	}
+
+	go func() {
+		defer cancel()
+
+		timer, stop := inctimer.New()
+		defer stop()
+
+		// If updates arrive during settle duration a single element
+		// is sent to this channel and we reinitialize right away
+		// without waiting for further updates.
+		trigger := make(chan struct{}, 1)
+
+		for {
+			// Wait for first update or trigger before reinitializing.
+			select {
+			case _, ok := <-updates:
+				if !ok {
+					return
+				}
+			case <-trigger:
+			}
+
+			log.Info("Reinitializing IPsec due to link changes")
+			err := l.reinitializeIPSec(ctx)
+			if err != nil {
+				// We may fail if links have been removed during the reload. In this case
+				// the updates channel will have queued updates which will retrigger the
+				// reinitialization.
+				log.WithError(err).Warn("Failed to reinitialize IPsec after device change")
+			}
+
+			// Avoid reinitializing repeatedly in short period of time
+			// by draining further updates for 'settleDuration'.
+			settled := timer.After(settleDuration)
+		settleLoop:
+			for {
+				select {
+				case <-settled:
+					break settleLoop
+				case <-updates:
+					select {
+					case trigger <- struct{}{}:
+					default:
+					}
+					break settleLoop
+				}
+
+			}
+		}
+	}()
 }
--- a/pkg/datapath/loader/base.go
+++ b/pkg/datapath/loader/base.go
@@ -198,18 +198,17 @@ func (l *Loader) reinitializeIPSec(ctx context.Context) error {
 		// IPAMENI mode supports multiple network facing interfaces that
 		// will all need Encrypt logic applied in order to decrypt any
 		// received encrypted packets. This logic will attach to all
-		// !veth devices. Only use if user has not configured interfaces.
-		if len(interfaces) == 0 {
-			if links, err := netlink.LinkList(); err == nil {
-				for _, link := range links {
-					isVirtual, err := ethtool.IsVirtualDriver(link.Attrs().Name)
-					if err == nil && !isVirtual {
-						interfaces = append(interfaces, link.Attrs().Name)
-					}
+		// !veth devices.
+		interfaces = nil
+		if links, err := netlink.LinkList(); err == nil {
+			for _, link := range links {
+				isVirtual, err := ethtool.IsVirtualDriver(link.Attrs().Name)
+				if err == nil && !isVirtual {
+					interfaces = append(interfaces, link.Attrs().Name)
 				}
 			}
-			option.Config.EncryptInterface = interfaces
 		}
+		option.Config.EncryptInterface = interfaces
 	}
 
 	// No interfaces is valid in tunnel disabled case
--- a/pkg/datapath/loader/base.go
+++ b/pkg/datapath/loader/base.go
@@ -193,6 +193,9 @@ func (l *Loader) reinitializeIPSec(ctx context.Context) error {
 		return nil
 	}
 
+	l.ipsecMu.Lock()
+	defer l.ipsecMu.Unlock()
+
 	interfaces := option.Config.EncryptInterface
 	if option.Config.IPAM == ipamOption.IPAMENI {
 		// IPAMENI mode supports multiple network facing interfaces that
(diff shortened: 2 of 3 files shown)


joamaki:
This fix was tested to be working on a EKS cluster with IPsec and ENI mode. The new `ensX` interface added at runtime was verified to have the `bpf_network` program loaded onto it.

```
ens4(27) clsact/ingress bpf_network.o:[from-network] id 2576
```

pkg/datapath/loader/netlink.go
```diff
@@ -329,3 +332,57 @@ func SetupBaseDevice(mtu int) (netlink.Link, netlink.Link, error) {
...
 }
+
+// reloadIPSecOnLinkChanges subscribes to link changes to detect newly added devices
+// and reinitializes IPsec on changes. Only in effect for ENI mode in which we expect
+// new devices at runtime.
+func (l *Loader) reloadIPSecOnLinkChanges() {
+	// settleDuration is the amount of time to wait for further link updates
+	// before proceeding with reinitialization. This avoids reinitializing
+	// twice when seeing NEWLINK followed by SETLINK.
+	const settleDuration = 5 * time.Second
```
pchaigno: That seems like a fairly long time. Couldn't we instead filter out events we don't want to see (maybe `u.Header.Type == unix.RTM_NEWLINK`)?

pkg/datapath/loader/netlink.go
```diff
@@ -329,3 +332,57 @@ func SetupBaseDevice(mtu int) (netlink.Link, netlink.Link, error) {
...
+
+	if !option.Config.EnableIPSec || option.Config.IPAM != ipamOption.IPAMENI {
+		return
+	}
+
+	updates := make(chan netlink.LinkUpdate, 16)
```
pchaigno: Where is the 16 coming from?

pchaigno (commented):
I like how this is fairly independent from the rest of the codebase. Much easier to backport indeed!

I'm wondering if we could avoid having any wait time, but other than that looks good to me.

pchaigno:
> Fix bug affecting EKS installations with IPsec encryption enabled, where Cilium wouldn't attach its IPsec BPF program to new ENI interfaces, resulting in connectivity loss between pods on remote nodes.  

Regarding above release note, small nit: this is not specific to EKS and can happen on EC2 as well, as long as you're using ENI mode.

pkg/datapath/loader/netlink.go
joamaki: Since this is also triggering compilation of the bpf_network program I wanted to err on the side of caution and not end up doing repeated work in a tight loop if the system has some weird flapping interfaces. It's a pretty rare occasion that new interfaces are added. We do need some delay to "coalesce" related updates.

Did also think about just reacting to NEWLINK, but then thought it'd be safer to also handle SETLINK (wasn't 100% sure if BPF prog loading requires interface to be up).

Doing just 1 second of "settling" would probably be fine so can drop it down to that.

pkg/datapath/loader/netlink.go
joamaki: The buffering isn't strictly necessary so I'll just drop that rather than have a random amount buffering.

pkg/datapath/loader/netlink.go
```diff
@@ -329,3 +332,58 @@ func SetupBaseDevice(mtu int) (netlink.Link, netlink.Link, error) {
...
+// reloadIPSecOnLinkChanges subscribes to link changes to detect newly added devices
+// and reinitializes IPsec on changes. Only in effect for ENI mode in which we expect
+// new devices at runtime.
+func (l *Loader) reloadIPSecOnLinkChanges() {
+	// settleDuration is the amount of time to wait for further link updates
+	// before proceeding with reinitialization. This avoids back-to-back
+	// reinitialization when multiple link changes are made at once.
+	const settleDuration = 1 * time.Second
+
+	if !option.Config.EnableIPSec || option.Config.IPAM != ipamOption.IPAMENI {
```
gandro, line 345: side-note: We don't support AWS CNI chaining with IPSec at the moment, so this is fine for now. But long-term, it might make sense to enable this feature independent of the IPAM mode, since I could imagine that there might be other setups (besides ENI mode) where pod egress/ingress interfaces may be added at run-time.

I wouldn't be surprised if alibabacloud or azure modes suffer from similar issues, but I'm not familiar enough with them to judge this.

pkg/datapath/loader/netlink.go
```diff
@@ -329,3 +332,58 @@ func SetupBaseDevice(mtu int) (netlink.Link, netlink.Link, error) {
...
+		return
+	}
+
+	ctx := context.Background()
+	updates := make(chan netlink.LinkUpdate)
+
+	if err := netlink.LinkSubscribe(updates, ctx.Done()); err != nil {
```
gandro, line 352: I guess this is only started once, so the fact that `ctx.Done()` will never be closed is somewhat fine.

But it might be cleaner if we actually cancelled `ctx` if the spawned go routine exits, to avoid leaking subscriptions e.g. if `reloadIPSecOnLinkChanges` is ever called from a unit test.

pkg/datapath/loader/netlink.go
```diff
@@ -329,3 +332,58 @@ func SetupBaseDevice(mtu int) (netlink.Link, netlink.Link, error) {
...
 }
+
+// reloadIPSecOnLinkChanges subscribes to link changes to detect newly added devices
+// and reinitializes IPsec on changes. Only in effect for ENI mode in which we expect
+// new devices at runtime.
+func (l *Loader) reloadIPSecOnLinkChanges() {
+	// settleDuration is the amount of time to wait for further link updates
+	// before proceeding with reinitialization. This avoids back-to-back
+	// reinitialization when multiple link changes are made at once.
+	const settleDuration = 1 * time.Second
```
gandro, line 343: The way I read this comment initially, I assumed this was done for every time we observe a burst of link changes. But is seems we only coalesce events for the initial round. Is that intentional?

gandro (commented):
Thanks for taking care of this! Looks good to me overall! Left some minor feedback

pkg/datapath/loader/netlink.go
pchaigno: This bug was initially reported to us as a race on agent startup, where the CiliumInternalIP is allocated to a new ENI interface, but we haven't attached our BPF program to it. We also have indications that it happens on node creation, node reboot, and on Cilium upgrades. If at any time we have CiliumInternalIP allocated from an interface where we didn't attach the BPF program, then that leads to IPsec ingress drops.

That's why I'd really like to get this to a minimum. We can't get to 0 because the loading of the BPF program itself takes time, but I think we can avoid waiting. If we are worried about the CPU utilization, then we could easily update `reinitializeIPSec` to only reload BPF programs when new interfaces actually showed up. If we can filter on NEWLINK, we also don't need to worry about flapping interfaces, right?

If we want to make this even faster, we could also avoid recompiling `bpf_network.c` every time we call `reinitializeIPSec`, but that's going to be more work.

> Did also think about just reacting to NEWLINK, but then thought it'd be safer to also handle SETLINK (wasn't 100% sure if BPF prog loading requires interface to be up).

That I think we can easily confirm, no?

pkg/datapath/loader/netlink.go
joamaki: We do coalesce for every burst, e.g. `for range updates` receives the first event, then we coalesce, reinit and then get back to the loop to wait for the first event of the burst.

pkg/datapath/loader/netlink.go
joamaki: Yeah, the context lives forever as loader is never stopped. I'll add the `defer cancel()` into the spawned goroutine even though the plan is to kill this code as soon as we have cleaner fix (https://github.com/cilium/cilium/pull/24677 and follow-ups to always do runtime detection of devices).

pkg/datapath/loader/netlink.go
pchaigno: Longer-term I want to get rid of the bpf_network. bpf_host can already do the same job.

> I wouldn't be surprised if alibabacloud or azure modes suffer from similar issues, but I'm not familiar enough with them to judge this.

Does Azure have any sort of interfaces like ENIs that can be added at runtime?

pkg/datapath/loader/netlink.go
gandro: If that's the intent, then the code is currently broken. `timer, stop := inctimer.New()` only fires once and it's never reset

joamaki: looking at `After` it seems like it stops the old timer and creates a new one. Did I misunderstand the API?
`inctimer.New` says:

```
 // New creates a new IncTimer and a done function.
 // IncTimer only uses one timer and correctly stops
 // the timer, clears the channel, and resets it every
 // time the `After` function is called.
```

pkg/datapath/loader/netlink.go
joamaki: Changed this now so that the delay/coalescing is after the first reinitialization so there won't be any delay. I've also moved the call to recompile `bpf_network.c` into `Reinitialize` so it won't be performed every time.

pkg/datapath/loader/netlink.go
joamaki: Though what was a bit unintended was resetting the timer after every update. I've now refactored the code a bit. PTAL.

pkg/datapath/loader/netlink.go
pchaigno: > I've also moved the call to recompile bpf_network.c into Reinitialize so it won't be performed every time.

I hadn't considered that. Nice!

> Changed this now so that the delay/coalescing is after the first reinitialization so there won't be any delay.

Did you check that loading the BPF program right after the NEWLINK and without waiting for SETLINK works fine? I'm guessing if we don't see any errors/warnings when testing this code, it means it's okay, no?

pchaigno (approved):

pkg/datapath/loader/netlink.go
gandro: Oh, sorry, you are of course right. I forgot that `inctimer` is started when we call `After`, not when we call `inctimer.New()` :facepalm: 

Sorry for the noise.

pkg/datapath/loader/netlink.go
gandro: I'm not 100% sure, but looking at the `NodeOperations` of AlibabaCloud and Azure, it looks like AlibabaCloud has a method to add new interfaces, but Azure does not:

https://github.com/cilium/cilium/blob/7e5bf218d963d2800ab6351373b13540faba394d/pkg/azure/ipam/node.go#L133-L137

gandro (approved):
Thanks!

rgo3 (approved):

PR #25745, opened 2023-05-29
High-scale IPCache: Nodeport LB support Part 1

julianwiedmann:
This PR is in the context of the high-scale ipcache feature described at https://github.com/cilium/design-cfps/pull/7.

It enables the nodeport LB to handle an unencapsulated service request, and forward the request to the backend using GENEVE-DSR. It also adds handling for reply traffic.

In detail:

- the request enters the LB in `from-netdev` (either XDP or TC), and `nodeport_lb4()` selects a backend.
- the DNATed packet goes down the DSR egress code path, and has GENEVE encapsulation added (+ DSR option as needed). In the context of hs-ipcache, we use the backend's IP address as OuterDstIP (same as if it was a pod-to-pod connection). If the load-balancing is done in XDP, we punt up to TC for adding the tunnel headers.
 - the packet is sent to the backend node
 - at the backend node, the `from-netdev` program strips off the encapsulation and redirects it to `from-overlay`. We manually transfer the DSR info across this redirect. The `from-overlay` program processes the DSR info and creates a corresponding SNAT entry.
 - replies are revDNATed in `to-overlay` (when they go to a destination inside the Clustermesh), or `to-netdev` (when the client belongs to one of the configured WorldCIDRs).

```release-note
Add support for load-balancing unencapsulated requests in a configuration with high-scale ipcache.
```

### DIFF
--- a/bpf/bpf_overlay.c
+++ b/bpf/bpf_overlay.c
@@ -617,6 +617,30 @@ int cil_from_overlay(struct __ctx_buff *ctx)
 
 	case bpf_htons(ETH_P_IP):
 #ifdef ENABLE_IPV4
+# ifdef ENABLE_HIGH_SCALE_IPCACHE
+#  if defined(ENABLE_DSR) && DSR_ENCAP_MODE == DSR_ENCAP_GENEVE
+		if (ctx_load_meta(ctx, CB_HSIPC_ADDR_V4)) {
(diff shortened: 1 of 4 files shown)


bleggett (approved):
LGTM

ldelossa (approved):
Changes look good to me. :+1:

Issue #23887, opened 2023-02-20
Hubble: add option to sanitize sensitive L7 data from flows


rolinh:
Hubble has the capability of providing visibility on L7 protocols such as HTTP or Kafka. This [layer 7 protocol visibility](https://docs.cilium.io/en/stable/observability/visibility/#proxy-visibility) feature is opt-in and requires users to either create a L7 policy or to add explicit pod annotations to be enabled. [Layer 7 Hubble flows](https://github.com/cilium/cilium/tree/master/api/v1/flow#flow-Layer7), however, may contain sensitive information, for instance as part of some HTTP headers or in a URL itself.

Hubble should provide an option for users to decide which potentially sensitive L7 data to keep in Hubble flows and it should be finely configurable.

- [x] HTTP URL: sanitize user info part (ie the optional username and password details for a URL), on/off: #28848
- [x] HTTP URL: remove query and fragment part of a URL, on/off: #25746
- [x] HTTP Headers: use a list of allowed HTTP header fields. A default list (e.g. tracing related fields) could be provided. Any header not in the list will have it's value sanitized (the key should be preserved). #26724
- [x] Kafka: redact API key, on/off: #25844 
- [ ] DNS: probably nothing to do.

ioandr:
Hi @rolinh,

we would like to take a stab on this with https://github.com/ChrsMark/, maybe we could split the items you listed. 

Looking at the codebase we understand that we need to extend Hubble's [Layer 7 parser](https://github.com/cilium/cilium/tree/main/pkg/hubble/parser/seven) and implement the described logic there. This is similar to how Hubble's Layer 7 parser already redacts the user's password to not include it in the network flow. Does this hold?

Thanks!

rolinh:
Hi @ioandr,

> we would like to take a stab on this with @ChrsMark, maybe we could split the items you listed.

Nice! I'm happy to create sub-issues if that helps but we can otherwise simply add your github handle next to each of the task.

> Looking at the codebase we understand that we need to extend Hubble's [Layer 7 parser](https://github.com/cilium/cilium/tree/main/pkg/hubble/parser/seven) and implement the described logic there. This is similar to how Hubble's Layer 7 parser already redacts the user's password to not include it in the network flow. Does this hold?

Yes, that's the way to do it indeed.

Thanks for your contributions!

ioandr:
Hi @rolinh 

[...]

PR #25746, opened 2023-05-29
Hubble: Add option to remove query from HTTP flows

ChrsMark:
This PR adds `--hubble-redact` flag for cilium agent with only accepted value the `http-url-query` for now. This is to provide the option to remove the Query and Fragment part from the URL in http flows.

Part of: https://github.com/cilium/cilium/issues/23887

```release-note
Add option to remove query from HTTP flows
```

#### TODOS
- [x] Add unit tests
- [x] Allign naming with https://github.com/cilium/cilium/pull/25844

### How to test this manually

1. In the root directory of the project edit https://github.com/cilium/cilium/blob/23bac97b3aa7a34e9d03f7d901bd77a124378446/contrib/testing/kind-values.yaml to add the following lines:
```yaml
extraArgs: ["--hubble-redact=rand1,http-url-query,rand2"]
```
2. Deploy cilium from source with `make kind-image && make kind-install-cilium`
3. Enable hubble with `cilium hubble enable --chart-directory=/home/chrismark/go/src/github.com/cilium/cilium/install/kubernetes/cilium`
4. Apply a L7 example endpoint with `kubectl apply -f https://raw.githubusercontent.com/cilium/cilium/1.13.3/examples/minikube/sw_l3_l4_l7_policy.yaml`
5. Port forward hubble relay svc with `kubectl port-forward -n kube-system svc/hubble-relay --address 0.0.0.0 --address :: 4245:80`
6. Deploy the sample app with `kubectl create -f https://raw.githubusercontent.com/cilium/cilium/1.13.3/examples/minikube/http-sw-app.yaml`. From https://docs.cilium.io/en/stable/gettingstarted/demo/#starwars-demo
7. Observe the flows with `hubble observe --pod deathstar --protocol http -f`
8. Execute a curl request with `kubectl exec tiefighter -- curl -s -XPOST "http://deathstar.default.svc.cluster.local/v1/request-landing#v2?foo=bar"`
9. Verify that sensitive parts are removed from the URL in the flows' observe output. 

### Screenshot
cilium-working

### DIFF
--- a/Documentation/observability/visibility.rst
+++ b/Documentation/observability/visibility.rst
@@ -66,6 +66,26 @@ In order for Cilium to populate the ``INGRESS ENFORCEMENT``, ``EGRESS ENFORCEMEN
 and ``VISIBILITY POLICY`` fields, it must run with ``--endpoint-status=policy``
 to make field values visible.
 
+Security Implications
+---------------------
+
+Monitoring Layer 7 traffic involves security considerations for handling
+potentially sensitive information, such as usernames, passwords, query
+parameters, API keys, and others.
+
+.. warning::
+
+   By default, Hubble does not redact potentially sensitive information
+   present in `Layer 7 Hubble Flows <https://github.com/cilium/cilium/tree/master/api/v1/flow#flow-Layer7>`_.
+
+To harden security, Cilium provides the ``--hubble-redact`` option to control
+how Hubble handles sensitive information present in Layer 7 flows. More
+specifically, it offers the following features for supported Layer 7 protocols:
+
+* For HTTP: redacting URL query (GET) parameters
+
+For more information on configuring Cilium, see :ref:`Cilium Configuration <configuration>`.
+
 Troubleshooting
 ---------------
 
--- a/install/kubernetes/cilium/templates/cilium-configmap.yaml
+++ b/install/kubernetes/cilium/templates/cilium-configmap.yaml
@@ -829,6 +829,12 @@ data:
 {{- end }}
   enable-hubble-open-metrics: {{ .Values.hubble.metrics.enableOpenMetrics | quote }}
 {{- end }}
+{{- if .Values.hubble.redact }}
+  # A space separated list of redact options for Hubble.
+  hubble-redact: {{- range .Values.hubble.redact }}
+    {{.}}
+{{- end }}
+{{- end }}
 {{- if hasKey .Values.hubble "listenAddress" }}
   # An additional address for Hubble server to listen to (e.g. ":4244").
   hubble-listen-address: {{ .Values.hubble.listenAddress | quote }}
--- a/pkg/hubble/parser/options/options.go
+++ b/pkg/hubble/parser/options/options.go
@@ -3,17 +3,48 @@
 
 package options
 
+import (
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	HttpUrlQuery = "http-url-query"
+)
+
 // Option is used to configure parsers
 type Option func(*Options)
 
 // Options contains all parser options
 type Options struct {
-	CacheSize int
+	CacheSize       int
+	RedactHTTPQuery bool
 }
 
 // CacheSize configures the amount of L7 requests cached for latency calculation
 func CacheSize(size int) Option {
 	return func(opt *Options) {
 		opt.CacheSize = size
+	}
+}
+
+// Redact configures which data Hubble will redact.
+func Redact(logger logrus.FieldLogger, hubbleRedactOptions []string) Option {
+	return func(opt *Options) {
+		validOpts := []string{}
+		for _, option := range hubbleRedactOptions {
+			switch option {
+			case HttpUrlQuery:
+				opt.RedactHTTPQuery = true
+			default:
+				if logger != nil {
+					logger.WithField("option", option).Warn("ignoring unknown Hubble redact option")
+				}
+				continue
+			}
+			validOpts = append(validOpts, option)
+		}
+		if logger != nil {
+			logger.WithField("options", validOpts).Info("configured Hubble with redact options")
+		}
 	}
 }
--- a/pkg/hubble/parser/seven/http_test.go
+++ b/pkg/hubble/parser/seven/http_test.go
@@ -15,6 +15,7 @@ import (
 
 	flowpb "github.com/cilium/cilium/api/v1/flow"
 	v1 "github.com/cilium/cilium/pkg/hubble/api/v1"
+	"github.com/cilium/cilium/pkg/hubble/parser/options"
 	"github.com/cilium/cilium/pkg/hubble/testutils"
 	"github.com/cilium/cilium/pkg/ipcache"
 	"github.com/cilium/cilium/pkg/proxy/accesslog"
@@ -138,6 +139,54 @@ func TestDecodeL7HTTPRequest(t *testing.T) {
 		},
 	}, f.GetL7().GetHttp())
 	assert.Equal(t, "4bf92f3577b34da6a3ce929d0e0e4736", f.GetTraceContext().GetParent().GetTraceId())
+}
+
+func TestDecodeL7HTTPRequestRemoveUrlQuery(t *testing.T) {
+	requestPath, err := url.Parse("http://myhost/some/path?foo=bar")
+	require.NoError(t, err)
+	lr := &accesslog.LogRecord{
+		Type:                accesslog.TypeRequest,
+		Timestamp:           fakeTimestamp,
+		NodeAddressInfo:     fakeNodeInfo,
+		ObservationPoint:    accesslog.Ingress,
+		SourceEndpoint:      fakeSourceEndpoint,
+		DestinationEndpoint: fakeDestinationEndpoint,
+		IPVersion:           accesslog.VersionIPv4,
+		Verdict:             accesslog.VerdictForwarded,
+		TransportProtocol:   accesslog.TransportProtocol(u8proto.TCP),
+		ServiceInfo:         nil,
+		DropReason:          nil,
+		HTTP: &accesslog.LogRecordHTTP{
+			Code:     0,
+			Method:   "POST",
+			URL:      requestPath,
+			Protocol: "HTTP/1.1",
+			Headers: http.Header{
+				"Host":        {"myhost"},
+				"Traceparent": {"00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01"},
+			},
+		},
+	}
+	lr.SourceEndpoint.Port = 56789
+	lr.DestinationEndpoint.Port = 80
+
+	opts := []options.Option{options.Redact(nil, []string{"http-url-query"})}
+	parser, err := New(log, nil, nil, nil, nil, opts...)
+	require.NoError(t, err)
+
+	f := &flowpb.Flow{}
+	err = parser.Decode(lr, f)
+	require.NoError(t, err)
+	assert.Equal(t, &flowpb.HTTP{
+		Code:     0,
+		Method:   "POST",
+		Url:      "http://myhost/some/path",
+		Protocol: "HTTP/1.1",
+		Headers: []*flowpb.HTTPHeader{
+			{Key: "Host", Value: "myhost"},
+			{Key: "Traceparent", Value: "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01"},
+		},
+	}, f.GetL7().GetHttp())
 }
 
 func TestDecodeL7HTTPRecordResponse(t *testing.T) {
--- a/pkg/hubble/parser/seven/parser.go
+++ b/pkg/hubble/parser/seven/parser.go
@@ -33,6 +33,7 @@ type Parser struct {
 	ipGetter          getters.IPGetter
 	serviceGetter     getters.ServiceGetter
 	endpointGetter    getters.EndpointGetter
+	opts              *options.Options
 }
 
 // New returns a new L7 parser
@@ -45,7 +46,8 @@ func New(
 	opts ...options.Option,
 ) (*Parser, error) {
 	args := &options.Options{
-		CacheSize: 10000,
+		CacheSize:       10000,
+		RedactHTTPQuery: false,
 	}
 
 	for _, opt := range opts {
@@ -70,6 +72,7 @@ func New(
 		ipGetter:          ipGetter,
 		serviceGetter:     serviceGetter,
 		endpointGetter:    endpointGetter,
+		opts:              args,
 	}, nil
 }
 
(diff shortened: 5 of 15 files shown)


kaworu:
Thanks for the PR @ChrsMark! I put this one in draft, feel free to mark it ready for review once the TODOs are implemented.

ChrsMark:
I added a unit test and aligned naming with https://github.com/cilium/cilium/pull/25844. We can review this now or wait for https://github.com/cilium/cilium/pull/25844 to have everything addressed and then debase on top of that (or vise-versa :) ). cc:@ioandr

ChrsMark:
Updated as per https://github.com/cilium/cilium/pull/25844#pullrequestreview-1471633042.

Documentation/observability/visibility.rst
```diff
@@ -66,6 +66,26 @@ In order for Cilium to populate the ``INGRESS ENFORCEMENT``, ``EGRESS ENFORCEMEN
 and ``VISIBILITY POLICY`` fields, it must run with ``--endpoint-status=policy``
 to make field values visible.
 
+Security Implications
+---------------------
+
+Monitoring Layer 7 traffic comes with some serious security considerations in
+terms of handling potentially sensitive information, such as usernames,
```
kaworu: nit
```suggestion
terms of handling potentially sensitive information such as usernames,
```

pkg/hubble/parser/options/options.go
```diff
@@ -3,12 +3,21 @@
...
+)
+
 // Option is used to configure parsers
 type Option func(*Options)
 
 // Options contains all parser options
 type Options struct {
-	CacheSize int
+	CacheSize       int
+	HTTPQueryRemove bool
```
kaworu: nit
```suggestion
	RedactHTTPQuery bool
```

pkg/hubble/parser/options/options.go
```diff
@@ -17,3 +26,25 @@ func CacheSize(size int) Option {
 		opt.CacheSize = size
 	}
 }
+
+// Redact configures which L7 data Hubble will redact
```
kaworu: nit
```suggestion
// Redact configures which data Hubble will redact.
```

pkg/hubble/parser/options/options.go
```diff
@@ -17,3 +26,25 @@ func CacheSize(size int) Option {
...
+func Redact(logger logrus.FieldLogger, hubbleRedactOptions []string) Option {
+	return func(opt *Options) {
+		validOpts := []string{}
+		for _, option := range hubbleRedactOptions {
+			switch option {
+			case HttpUrlQuery:
+				opt.HTTPQueryRemove = true
+			default:
+				if logger != nil {
+					logger.Warn("ignoring unknown Hubble redact option: ", option)
```
kaworu: nit
```suggestion
					logger.WithField("option", option).Warn("ignoring unknown Hubble redact option")
```

pkg/hubble/parser/seven/parser.go
```diff
@@ -134,7 +137,7 @@ func (p *Parser) Decode(r *accesslog.LogRecord, decoded *flowpb.Flow) error {
 	decoded.Type = flowpb.FlowType_L7
 	decoded.SourceNames = sourceNames
 	decoded.DestinationNames = destinationNames
-	decoded.L7 = decodeLayer7(r)
+	decoded.L7 = decodeLayer7(r, *p.opts)
```
kaworu: this reads a tad weird IMHO, maybe the pointer dance would be clearer if both `p.opts` and `args` (line 48) were not pointers.

pkg/option/config.go
```diff
@@ -218,6 +218,9 @@ const (
 	// K8sSyncTimeout is the timeout since last event was received to synchronize all resources with k8s.
 	K8sSyncTimeoutName = "k8s-sync-timeout"
 
+	// HubbleL7ParserHttpRemoveUrlQuery
+	HubbleL7ParserHttpRemoveUrlQuery = "hubble-l7-parser-http-remove-url-query"
```
kaworu: This constant is a left-over of a previous version of the patch correct?

pkg/option/config.go
```diff
@@ -1572,6 +1579,9 @@ type DaemonConfig struct {
 	// HTTPRetryTimeout is the time in seconds before an uncompleted request is retried.
 	HTTPRetryTimeout int
 
+	// HubbleL7ParserHttpRemoveUrlQuery
+	HubbleL7ParserHttpRemoveUrlQuery bool
+
```
kaworu: ditto

pkg/option/config.go
```diff
@@ -2998,6 +3012,7 @@ func (c *DaemonConfig) Populate(vp *viper.Viper) {
 	c.IPv6ServiceRange = vp.GetString(IPv6ServiceRange)
 	c.JoinCluster = vp.GetBool(JoinClusterName)
 	c.K8sEnableK8sEndpointSlice = vp.GetBool(K8sEnableEndpointSlice)
+	c.HubbleL7ParserHttpRemoveUrlQuery = vp.GetBool(HubbleL7ParserHttpRemoveUrlQuery)
```
kaworu: ditto

pkg/hubble/parser/seven/http_test.go
```diff
@@ -140,6 +141,126 @@ func TestDecodeL7HTTPRequest(t *testing.T) {
...
+				return &testutils.FakeEndpointInfo{
+					ID: fakeDestinationEndpoint.ID,
+				}, true
+			}
+			return nil, false
+		},
+	}
+
+	opts := []options.Option{options.Redact(nil, []string{"http-url-query"})}
+	parser, err := New(log, dnsGetter, IPGetter, serviceGetter, endpointGetter, opts...)
```
kaworu: It doesn't feel right to duplicate all the unrelated checks here. I think we could pass `nil` for `dnsGetter`, `IPGetter`, `serviceGetter`, and `endpointGetter` and check only the HTTP related fields in the decoded flow.

kaworu (changes requested):
Thanks for the update @ChrsMark! I have some comments and nits but overall the patch LGTM. Could you please cleanup the commit history on the way? Ideally three commits
1. refactor `decodeHTTP` to avoid mutating the accesslog when we redact the URL's password
2. Add support `--hubble-react=http-url-query` (helm integration can be in the same commit)
3. "Security Implications" section in "Layer 7 Protocol Visibility" doc.

Documentation/observability/visibility.rst
```diff
@@ -66,6 +66,26 @@ In order for Cilium to populate the ``INGRESS ENFORCEMENT``, ``EGRESS ENFORCEMEN
...
+passwords, query parameters, API keys, etc.
+
+.. Warning::
```
qmonnet:
```suggestion
.. warning::
```

You also need to indent the body of your warning, or Sphinx won't know what's part of it. This is why the Netlify preview build currently fails with `ERROR: Content block expected for the "Warning" directive; none found.`.

qmonnet (changes requested):

pkg/option/config.go
ChrsMark: Yeap, sorry for these 😞

pkg/hubble/parser/seven/parser.go
ChrsMark: I think it's not a good idea to to not have them as pointers, cause we also pass them by reference when evaluating the `opts` at https://github.com/cilium/cilium/blob/e383fb572e0a3900887fac54c09f78b164083527/pkg/hubble/parser/seven/parser.go#L54.
It would be better if we pass the `p.opts` by reference here too like `decoded.L7 = decodeLayer7(r, p.opts)`. Looks more consistent and less ugly :). But let me know what you think.

ChrsMark:
Thanks @kaworu and @qmonnet for the reviews!

I have updated the PR's changes and re-organised the commits as requested. Feel free to have another look when you have the time.

Note1: We will need to also add the `--hubble-redact=http-basic-auth-password` option but better to do it in a separate follow-up PR.
Note2: When you folks want to merge this, please make sure that the commits (or the merge commit) will include @ioandr as co-author cause we paired in this after some point :)

qmonnet:
> Note2: When you folks want to merge this, please make sure that the commits (or the merge commit) will include @ioandr as co-author cause we paired in this after some point :)

We're not touching your commit description (we're not squashing the commits, we'd ask you to do it if it was necessary) so your `Co-authored-by:` will of course be preserved.

Thanks! I'll take another look, although not sure if I'll have time today.

qmonnet:
I note that the Netlify preview is still not building, so maybe something you want to investigate. I'll look at it later otherwise.

install/kubernetes/cilium/templates/cilium-configmap.yaml
```diff
@@ -819,6 +819,11 @@ data:
   # https://github.com/cilium/hubble/blob/master/Documentation/metrics.md
   hubble-metrics: {{- range .Values.hubble.metrics.enabled }}
     {{.}}
+{{- end }}
```
qmonnet, line 829: I think the CI errors are because we're missing a closing `{{- end}}` for the [`range`](https://helm.sh/docs/chart_template_guide/control_structures/#looping-with-the-range-action). This `{{- end}}` closes the `{{- range [...] }}`, not the `{{- if .Values.hubble.metrix.enabled }}`. You need to close both, then add your `if` and `range`, and close both again.

qmonnet (changes requested):

qmonnet:
Please also shorten your commit title, we're trying to keep them under 75 characters.

image

pkg/hubble/parser/seven/parser.go
kaworu: @ChrsMark yes I fully agree, to clarify I wasn't suggesting to change the functions themselves, but declare the variables (`p.opts` and `args`) as value as pass them as references (`&p.opts`, `&args`) when needed instead.

Documentation/observability/visibility.rst
```diff
@@ -66,6 +66,26 @@ In order for Cilium to populate the ``INGRESS ENFORCEMENT``, ``EGRESS ENFORCEMEN
...
+Security Implications
+---------------------
+
+Monitoring Layer 7 traffic comes with some serious security considerations in
+terms of handling potentially sensitive information such as usernames,
+passwords, query parameters, API keys, etc.
+
+.. warning::
+
+    By default, Hubble does **not** protect potentially sensitive information
```
kaworu: nit: "protect" is vague here, maybe redact or edit or hide would be more appropriate?

pkg/hubble/parser/seven/parser.go
kaworu: current patch is great, resolving this conversation.

kaworu (approved):
Awesome work @chrsMark and @ioandr :tada:, the patch LGTM!

youngnick (approved):

Documentation/observability/visibility.rst
```diff
@@ -66,6 +66,25 @@ In order for Cilium to populate the ``INGRESS ENFORCEMENT``, ``EGRESS ENFORCEMEN
...
+
+.. warning::
+
+   By default, Hubble does not redact potentially sensitive information
+   present in `Layer 7 Hubble Flows <https://github.com/cilium/cilium/tree/master/api/v1/flow#flow-Layer7>`_.
+
+To harden security, Cilium provides options to configure how Hubble handles
+sensitive information present in Layer 7 flows. More specifically:
+
+* HTTP: redact URL query parameters
```
qmonnet: “More specifically:”... What is `HTTP` in this context?

If this is a keyword (I don't think so?), how can I use it?

If not, should it be something like “For HTTP: Hubble can redact URL query parameters” (and in that case, what's the point of having a list if we only have a single item in it)?

(Nice work otherwise, thanks a lot!)

ChrsMark: TBH this was "cherry-picked" from https://github.com/cilium/cilium/pull/25844#discussion_r1224102713 after we decided to move on with this PR first. Based on this, for now we only add the `HTTP: redact URL query parameters` note and we will extend it accordingly with https://github.com/cilium/cilium/pull/25844 and https://github.com/cilium/cilium/issues/23887 in general.

cc @ioandr since he mostly worked on this specific content.

pippolo84 (approved):
LGTM for agent-related changes, thanks! :100:

Documentation/observability/visibility.rst
ioandr: @qmonnet Indeed, we are planning to extend this section once we add support to Hubble to conditionally redact Kafka API keys in L7 network flows.

Also see https://github.com/cilium/cilium/pull/25844#issuecomment-1588897180.

qmonnet: OK so let's keep the list, but please rephrase to make it easier to understand what it's about. Maybe something along:


```rst
... Layer 7 flows. More specifically, it supports the following features:

* For HTTP: redacting URL query (GET) parameters
```

Does that make sense?

qmonnet: My concern is that a user discovering this lists may not understand what it is. Is “HTTP” a keyword I should use in a CRD somewhere? In a command line? Is this some instruction I _should_ follow? Is this a list of supported features? (Answer: the latter, but let's make it explicit)

ioandr: @qmonnet ACK, I just pushed a FIXUP for this.

Let us know if you're OK with it, else feel free to propose an alternative.

qmonnet: Looks good, thank you!

qmonnet (approved):
Looks good, thanks for this work!

ChrsMark:
Hey @kaworu @qmonnet , anything else missing here? I see some tests are still pending maybe due to one test failure at https://app.travis-ci.com/github/cilium/cilium/builds/263803637#L22985? I'm not sure if this failure could be related to this patch. Any ideas on this?

qmonnet:
Hi @ChrsMark, the tests status is reset each time you update the branch. Some of them are triggered with specific comments (but only by members of the Cilium org).

What we need to merge this is 1) tests passing and 2) approvals from all code owners. We're missing the tests results at the moment, and two reviews (cc @asauber @aditighag :pray:).

Note that you don't have to rebase if the branch goes out of date with `main` (unless there are conflicts to solve). If tests are still passing then the ball is in our (reviewers') court

asauber (approved):
CLI-related changes LGTM

aditighag (approved):

ioandr:
> Looking at the results, I think there are flakes/temporary failures. I opened GitHub issues to track them, and re-triggered the tests. We'll see if the issues persist.
This is our understanding as well, let's see. It seems that tests pass now:

   ```
   58 successful and 4 skipped checks
   ```

> There is [some (new!) documentation here for the `Conformance ginkgo` workflow](https://docs.cilium.io/en/latest/contributing/testing/e2e/#running-locally-ginkgo-tests-based-on-ginkgo-s-github-workflow), I'm not sure whether it applies to all workflows; and I don't think all of them are documented (the CI has changed quite a lot recently). BPF unit tests, or runtime tests, should be documented at least. We usually run on GitHub runners, save for some platform-specific tests. You're best chance at figuring the details is probably to look at the workflow description files, but we're also happy to help if you have questions on a specific test - here or on Slack.
That's good to know, we'll definitely take a look, thanks!

PR #25757, opened 2023-05-30
k8s: Split SharedResources into binary specific cells

### LINKED ISSUE #24070: Modularization: split SharedResources into OperatorResources and DaemonResources
SharedResourcesCell provides a set of shared handles to Kubernetes resources.
At the moment, the same cell is used in the daemon, the operator and the clustermesh-apiserver, despite each binary does not use all the resources.

To avoid having a single cell with too many unrelated resources, we should split the SharedResourcesCell into separate ones specific for each binary. This will help to understand which resources each binary is actually using.

**Success Criteria**

- The SharedResourcesCell is split in more binary-specific cell

Related: #23425

pippolo84:
Split the SharedResources cell into multiple cells, one for each binary (agent, operator and clustermesh-apiserver).
This improves the dependencies management, since each binary will depend only on the k8s resources that it actually uses.

Also, the functional options pattern is introduced to give more flexibility when customizing a resource events stream.

_Notes for the reviewers_: review commit by commit might be simpler. The last commit is the one responsible for the high number of files modified, but it is just a rename (`sharedResources` -> `resources`) after the removal of the single SharedResources cell.

Fixes https://github.com/cilium/cilium/issues/24070


pkg/k8s/resource_ctors.go
```diff
@@ -0,0 +1,120 @@
...
+		return nil, nil
+	}
+	lw := utils.ListerWatcherWithModifiers(
+		utils.ListerWatcherFromTyped[*cilium_api_v2.CiliumIdentityList](cs.CiliumV2().CiliumIdentities()),
+		opts...,
+	)
+	return resource.New[*cilium_api_v2.CiliumIdentity](lc, lw), nil
+}
+
+func CiliumNetworkPolicy(lc hive.Lifecycle, cs client.Clientset, opts ...func(*metav1.ListOptions)) (resource.Resource[*cilium_api_v2.CiliumNetworkPolicy], error) {
```
joamaki: I wonder if it'd make sense to namespace these by package, e.g. `pkg/k8s/resources` and then we'd have `resources.CiliumNetworkPolicy`, or by the suffix (`NamespaceResource`). I think either is probably fine and I do like having `k8s` in the package name. If you go the suffix route, then let's add it to these `Cilium*` functions here.

pippolo84: Added the `Resource` suffix where missing.

lmb (approved):

nebril (approved):

gandro (approved):
My codeowners look good to me.

It might be a naive question, but why do we even need the per-binary packages? Can't there be something like `cell.LazyProvide` that only executes the constructor if it's actually requested by a dependency?

pippolo84:
> It might be a naive question, but why do we even need the per-binary packages? Can't there be something like `cell.LazyProvide` that only executes the constructor if it's actually requested by a dependency?

I think it'd be a reasonable alternative for the current usage of the resources, even if keeping a single cell would not allow to customize each resource events stream when consumed by multiple binaries.
As an example, we might want to use different selectors (or different [TransformFunc](https://pkg.go.dev/k8s.io/client-go@v0.26.0/tools/cache#TransformFunc) when we'll add support for that in resource.Resource[T]).

christarazi (approved):

clustermesh-apiserver/k8s/resources.go
```diff
@@ -0,0 +1,25 @@
...
+	// clustermesh-apiserver. Each of the resources share a client-go informer and backing store so we only
+	// have one watch API call for each resource kind and that we maintain only one copy of each object.
+	//
+	// See pkg/k8s/resource/resource.go for documentation on the Resource[T] type.
+	ResourcesCell = cell.Module(
+		"k8s-resources",
+		"Clustermesh-apiserver Kubernetes resources",
+
+		cell.Provide(
+			k8s.ServiceResource,
```
YutaroHayakawa, line 22: I think we need CiliumNode, Endpoints, and CiliumEndpoint resources, but they are currently using the legacy `clientset`. Approving with the understanding that they will be followed up at some point.

YutaroHayakawa (approved):
Looks good to me from ClusterMesh's perspective.

clustermesh-apiserver/k8s/resources.go
pippolo84: Yep, there is already work in progress to port other resources to the new `resource.Resource[T]` framework.

pippolo84:
Force-pushed to solve a conflict in `clustermesh-apiserver/main.go`

ldelossa (approved):
LGTM

Issue #24867, opened 2023-04-13
Confusing name for CNP authentication type


lizrice:
CiliumNetworkPolicy can now define an authentication type thanks to #24263. However, the name of the supported type `mtls-spiffe` is confusing: the "mtls" part implies that the resulting connection will be encrypted (as in TLS), but at the moment this field is only used to specify that the _handshake_ that performs authentication must take place. The intention is that the resulting identities will be passed to the kernel where transparent encryption will take place, but it's possible that the encryption part will be controlled separately. With the name as it is, we risk disappointing users who will assume that they have a CNP that defines a requirement for encryption as well as authentication. 

Also, if possible, it would be nice to decouple the authentication _type_ specified in the CNP from the authentication _method_ so that a CNP could specify that it requires mutual authentication, but doesn't care whether the identities are provided by SPIFFE, X.509 certs, or whatever. 

My suggestion is that instead of `auth: mtls-spiffe` we use `auth: mutual` if it can be independent of the authentication method, or `auth: mutual-spiffe` if it has to be specifically SPIFFE.  

It would be great if the encryption requirement could also be specified in the CNP too, like this:

```
auth: mutual
encrypt: true
```

youngnick:
I've spent some time looking into this and thinking today, here's my summary:

I agree that we should use a non-`mtls` name for the SPIFFE auth method.

From what I can see, neither Wireguard nor IPSec transparent encryption modes in Cilium have a way to exclude specific flows, so I don't know if adding an `encrypt` field would be useful - the encryption properties are currently set at the cluster or Cilium install level (as is the mutual auth method currently), so setting `encrypt: false` for only a single NetworkPolicy isn't possible _yet_.

The one thing I can think of if we're going to be changing things is that since `auth` will _always_ imply `mutual`, maybe we should move the word `mutual` into the type field?

So,
```yaml
  auth:
    mutual-type: spiffe
```
or similar?

youngnick:
oh, that is a good point, I _think_ we can still change that since these fields are not useful until 1.14 is released.

youngnick:
Okay, coming back to this to try and close this out, I have two proposals:
- [ ] rename the value in NetworkPolicy to `spiffe`. Just removing the `tls` part removes any possibility of confusion here, and we'll address that `spiffe` uses TLS in other parts of the docs. Something like this is basically a must-do.
- [ ] rename the current `mTLS` stanza in Helm to `mutual` instead. Something like this is another must-do.
[...]

Issue #25291, opened 2023-05-05
CI: runtime-net-next: Privileged unit tests: pkg/datapath/linux: panic: test timed out after 10m0s


joestringer:
## CI failure

https://jenkins.cilium.io/job/Cilium-PR-Runtime-net-next/6472/testReport/junit/(root)/Suite-runtime/RuntimeDatapathPrivilegedUnitTests_Run_Tests/

```

	 [02:56:17] 	github.com/cilium/cilium/pkg/datapath/linux	coverage: 38.7% of statements
	 [02:56:17] panic: test timed out after 10m0s
	 [02:56:17] running tests:
	 [02:56:17] 	Test (10m0s)
	 [02:56:17] 
	 [02:56:17] goroutine 26 [running]:
	 [02:56:17] testing.(*M).startAlarm.func1()
	 [02:56:17] 	/usr/local/go/src/testing/testing.go:2241 +0x3b9
	 [02:56:17] created by time.goFunc
	 [02:56:17] 	/usr/local/go/src/time/sleep.go:176 +0x32
	 [02:56:17] 
	 [02:56:17] goroutine 1 [chan receive, 10 minutes]:
	 [02:56:17] testing.(*T).Run(0xc00017b040, {0x280bb90?, 0x56f0a5?}, 0x29414c8)
	 [02:56:17] 	/usr/local/go/src/testing/testing.go:1630 +0x405
	 [02:56:17] testing.runTests.func1(0x410fc00?)
	 [02:56:17] 	/usr/local/go/src/testing/testing.go:2036 +0x45
	 [02:56:17] testing.tRunner(0xc00017b040, 0xc0004ffc80)
	 [02:56:17] 	/usr/local/go/src/testing/testing.go:1576 +0x10b
	 [02:56:17] testing.runTests(0xc000141360?, {0x40cc140, 0x1, 0x1}, {0x80?, 0x1000000000040?, 0x410e480?})
	 [02:56:17] 	/usr/local/go/src/testing/testing.go:2034 +0x489
	 [02:56:17] testing.(*M).Run(0xc000141360)
	 [02:56:17] 	/usr/local/go/src/testing/testing.go:1906 +0x63a
	 [02:56:17] main.main()
	 [02:56:17] 	_testmain.go:80 +0x1c5
	 [02:56:17] 
	 [02:56:17] goroutine 55 [chan receive]:
[...]

PR #25761, opened 2023-05-30
mutual-auth: Avoid confusion on mTLS wording

sayboras:
### Description

### Tasks

_Copied from https://github.com/cilium/cilium/issues/24867_

- [x] rename the field in NetworkPolicy to `authentication` (from `auth`)
- [x] rename the value in NetworkPolicy to `spiffe` from `mtls-spiffe`. Just removing the `tls` part removes any possibility of confusion here, and we'll address that `spiffe` uses TLS in other parts of the docs. Something like this is basically a must-do.
- [x] rename the existing `auth` stanza in Helm to `authentication` instead.
- [x] rename the current `mTLS` stanza in Helm to `mutual` instead. This means that the complete path will be `authentication.mutual`. To me, this leaves the `authentication` level with the possibility to add other types of authentication for Cilium to configure. I don't know if this is viable, or if we should collapse what's currently `mTLS` into `authentication` instead.
- [x] Change all command-line flags and config file options for `cilium` to use `mutual`, `mutual-auth`, or `MutualAuth`  instead of `mtls`, `mtls-auth`, or `MTLS` respectively.
- [x] Refactor code to remove mTLS references (lowest priority).

### Testing

Testing was done locally with cilium cli build from https://github.com/cilium/cilium-cli/pull/1673

```
$ cilium connectivity test --test auth
...
[=] Test [echo-ingress-auth-always-fail]
........
[=] Test [echo-ingress-mutual-auth-spiffe]
........

[=] Skipping Test [pod-to-ingress-service]
...
[=] Skipping Test [to-fqdns]

✅ All 2 tests (16 actions) successful, 53 tests skipped, 0 scenarios skipped.
```


operator/auth/cell.go
```diff
@@ -11,23 +11,23 @@ import (
 )
 
 const (
-	MTLSEnabled = "mesh-auth-mtls-enabled"
+	mutualAuthEnabled = "mesh-mutual-auth-enabled"
```
meyskens: This became magically unexported, do we export these in other packages too?

sayboras: yeah, I noticed that, the compilation didn't fail, so I think it's not imported in other package right now.

meyskens (approved):
One nit but LGTM

operator/auth/cell.go
meyskens: in that case it probably is fine

Documentation/cmdref/cilium-agent.md
```diff
@@ -237,11 +237,11 @@ cilium-agent [flags]
...
       --log-opt map                                             Log driver options for cilium-agent, configmap example for syslog driver: {"syslog.level":"info","syslog.facility":"local5","syslog.tag":"cilium-agent"}
       --log-system-load                                         Enable periodic logging of system load
-      --mesh-auth-mtls-listener-port int                        Port on which the Cilium Agent will perfom mTLS handshakes between other Agents
       --mesh-auth-queue-size int                                Queue size for the auth manager (default 1024)
       --mesh-auth-rotated-identities-queue-size int             The size of the queue for signaling rotated identities. (default 1024)
       --mesh-auth-spiffe-trust-domain string                    The trust domain for the SPIFFE identity. (default "spiffe.cilium")
       --mesh-auth-spire-admin-socket string                     The path for the SPIRE admin agent Unix socket.
+      --mesh-mutual-auth-listener-port int                      Port on which the Cilium Agent will perform mutual authentication handshakes between other Agents
```
mhofstetter: this rename kind of breaks the unwritten rule that all config values are starting with `mesh-auth-*` :thinking: 

do we want to rename it `mesh-auth-mutual-listener-port`? or `mesh-auth-handshake-listener-port`?

or do we want to introduce `mutual` to other config properties.

having a look at https://github.com/cilium/cilium/pull/25761/files#diff-83e9b8b2ccacddb2fbdb28fbcfa148473d05056c0b4c955fc971709b469d7331 kind of unveils the differences in the naming handling between agent config properties & helm values.

Documentation/cmdref/cilium-operator.md
```diff
@@ -78,11 +78,11 @@ cilium-operator [flags]
       --limit-ipam-api-qps float                             Queries per second limit when accessing external IPAM APIs (default 4)
       --log-driver strings                                   Logging endpoints to use for example syslog
       --log-opt map                                          Log driver options for cilium-operator, configmap example for syslog driver: {"syslog.level":"info","syslog.facility":"local4"}
-      --mesh-auth-mtls-enabled                               The flag to enable mTLS for the SPIRE server.
       --mesh-auth-spiffe-trust-domain string                 The trust domain for the SPIFFE identity. (default "spiffe.cilium")
       --mesh-auth-spire-agent-socket string                  The path for the SPIRE admin agent Unix socket. (default "/run/spire/sockets/agent/agent.sock")
       --mesh-auth-spire-server-address string                SPIRE server endpoint. (default "spire-server.spire.svc.cluster.local:8081")
       --mesh-auth-spire-server-connection-timeout duration   SPIRE server endpoint. (default 10s)
+      --mesh-mutual-auth-enabled                             The flag to enable mutual authentication for the SPIRE server.
```
mhofstetter: same here regarding the naming differences with the prefix `mesh-auth-*`

mhofstetter (commented):
thanks for the renames @sayboras! 

only one small concern regarding naming inline.

Documentation/cmdref/cilium-agent.md
youngnick: I think I'm +1 on `mesh-auth-mutual-*` for this type of flag.

operator/auth/cell.go
```diff
@@ -11,23 +11,23 @@ import (
...
+	"Cilium Mutual Authentication Identity management",
 	spire.Cell,
 	cell.Config(Config{}),
 	cell.Invoke(registerIdentityWatcher),
 )
 
 // Config contains the configuration for the identity-gc.
 type Config struct {
-	Enabled bool `mapstructure:"mesh-auth-mtls-enabled"`
+	Enabled bool `mapstructure:"mesh-mutual-auth-enabled"`
```
youngnick: Ditto other comments on `mesh-auth-mutual` rather than the other way around.

Documentation/cmdref/cilium-agent.md
sayboras: ah good point, I was kind of wondering between two options as well, let me make the update.

youngnick (approved):
LGTM with that one change about using `mesh-auth-mutual` for command line flags instead of `mesh-mutual-auth` (just so the flags are all grouped together).

Documentation/cmdref/cilium-operator.md
sayboras: thanks :100:

mhofstetter (approved):

api/v1/flow/flow.proto
```diff
@@ -127,7 +127,7 @@ enum FlowType {
 enum AuthType {
     NONE = 0;
     NULL = 1;
-    MTLS_SPIFFE = 2;
+    SPIFFE = 2;
```
rolinh: FYI, this is technically a breaking change for the API. However, I believe we haven't tagged any release with this enum yet so we're OK. If we had, then we would have to use an alias (see [here](https://protobuf.dev/programming-guides/proto3/#enum)).

rolinh (approved):
LGTM

api/v1/flow/flow.proto
sayboras: yeah, we don't have any tagged release yet.

sayboras:
runtime hits https://github.com/cilium/cilium/issues/25291, re-trigger with the hope the button will become green.

michi-covalent (approved):

lmb (approved):

kaworu (approved):
Helm changes LGTM

youngnick (approved):

pkg/policy/l4.go
```diff
@@ -220,30 +220,27 @@ func (a *PerSelectorPolicy) Equal(b *PerSelectorPolicy) bool {
...
-	case "null":
-		return AuthTypeNull
-	case "mtls-spiffe":
-		return AuthTypeMTLSSpiffe
-	case "always-fail":
+	if a == nil || a.Authentication == nil {
+		return AuthTypeDisabled
+	}
+	switch a.Authentication.Mode {
+	case "required":
```
mhofstetter: we should use the constants from `AuthType`

pkg/policy/l4.go
```diff
@@ -255,14 +252,12 @@ func (a AuthType) Uint8() uint8 {
...
-	case AuthTypeNone:
-		return ""
-	case AuthTypeNull:
-		return "null"
-	case AuthTypeMTLSSpiffe:
-		return "mtls-spiffe"
+	case AuthTypeDisabled:
+		return "disabled"
+	case AuthTypeAuthSpire:
+		return "required"
```
mhofstetter: string representation of `AuthTypeAuthSpire` should be `spire` and not `required`?

pkg/policy/l4.go
```diff
@@ -220,30 +220,27 @@ func (a *PerSelectorPolicy) Equal(b *PerSelectorPolicy) bool {
...
-	// AuthTypeNone means no authentication required
-	AuthTypeNone AuthType = iota
-	// AuthTypeNull is a simple auth type that always succeeds
-	AuthTypeNull
-	// AuthTypeMTLSSpiffe is a mTLS auth type that uses SPIFFE identities with a SPIRE server
-	AuthTypeMTLSSpiffe
+	// AuthTypeDisabled means no authentication required
+	AuthTypeDisabled AuthType = iota
+	// AuthTypeAuthSpire is a mutual auth type that uses SPIFFE identities with a SPIRE server
+	AuthTypeAuthSpire
```
mhofstetter: why having `Auth` twice in the const name? 

-> `AuthTypeSpire`?

pkg/auth/cell.go
```diff
@@ -33,13 +33,13 @@ var Cell = cell.Module(
 		newSignalRegistration,
 		// Null auth handler provides support for auth type "null" - which always succeeds.
```
mhofstetter, line 34: it shouldn't be possible to use the null auth handler (always pass) -> should be removed from the cell

pkg/auth/disabled_authhandler.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"time"
+
+	"github.com/cilium/cilium/pkg/auth/certs"
+	"github.com/cilium/cilium/pkg/policy"
+)
+
+// disabledAuthHandler implements an authHandler for auth type disable by just authenticate every request.
+type disabledAuthHandler struct {
```
mhofstetter: there isn't an authhandler for type `disabled` -> these requests should never be passed to the auth manager (and therefore an auth handler)

-> instead of renaming from `nullAuthHandler` to `disabledAuthHandler` -> i would recommend to rename this one to `alwaysPassAuthHandler` to be aligned with the `always-fail` authhandler.

please ntoe that in comparison to the "always-fail", the "always-pass" shouldn't be available via configuration and only be used for unit-testing

pkg/auth/disabled_authhandler.go
```diff
@@ -0,0 +1,32 @@
...
+	// Authentication trivially done
+	log.Debugf("auth: Successfully authenticated request")
+
+	return &authResponse{
+		expirationTime: time.Now().Add(1 * time.Minute),
+	}, nil
+}
+
+func (r *disabledAuthHandler) authType() policy.AuthType {
+	return policy.AuthTypeDisabled
```
mhofstetter: we won't be able to use an auth type from the constant. we have to return something like `policy.AuthType(100)` :thinking:

pkg/auth/mutual_authhandler.go
```diff
@@ -140,35 +141,35 @@ func (m *mtlsAuthHandler) authenticate(ar *authRequest) (*authResponse, error) {
 	}, nil
 }
 
-func (m *mtlsAuthHandler) authType() policy.AuthType {
-	return policy.AuthTypeMTLSSpiffe
+func (m *mutualAuthHandler) authType() policy.AuthType {
+	return policy.AuthTypeAuthSpire
```
mhofstetter: having the SPIRE/SPIFFE agnostic "mutual authhandler" returning type "spire" is ok for the time being. we should refactor this once we provide support for multiple CertificateProviders.

mhofstetter (changes requested):

pkg/policy/l4.go
sayboras: good catch :100:

pkg/auth/cell.go
sayboras: good point :100:

joamaki (approved):

pkg/auth/mutual_authhandler.go
sayboras: yeah, we might need to revisit this one later :(

pkg/auth/always_pass_authhandler.go
```diff
@@ -0,0 +1,33 @@
...
+package auth
+
+import (
+	"time"
+
+	"github.com/cilium/cilium/pkg/auth/certs"
+	"github.com/cilium/cilium/pkg/policy"
+)
+
+// alwaysPassAuthHandler implements an authHandler for auth type disable by just authenticate every request.
```
mhofstetter: nit: comment is wrong regarding auth type `disable` -> maybe document that this one is for testing and lets every auth request pass without any check.

mhofstetter (approved):
LGTM - thanks for the additional changes :1st_place_medal: 

only one small non-blocking nit regarding the comment of `alwaysPassAuthHandler`

bimmlerd (changes requested):
GH is somehow not letting me comment on the file, but `tools/maptool/maptool` is a binary, I'm guessing it was accidentially commited?

sayboras:
> GH is somehow not letting me comment on the file, but `tools/maptool/maptool` is a binary, I'm guessing it was accidentially commited?

ah thanks for pointing it out, it the left over file in old releases :( (probably due to my backport duty)

bimmlerd (approved):
contributing review requirement is gone now, so I can :disappear:

Documentation/operations/upgrade.rst
```diff
@@ -329,6 +329,10 @@ Annotations:
   being rerouted towards an Egress Gateway.
 * If Gateway API feature is enabled, please upgrade related CRDs to v0.6.x. This is
   mainly for ReferenceGrant resource version change (i.e. from v1alpha2 to v1beta1).
+* The attribute ``auth.type`` is renamed to ``authentication.mode`` in both Ingress and
+  Egress rules in CiliumNetworkPolicy CRD. The old attribute name is no longer supported,
```
qmonnet, line 333: Just checking: No deprecation period required?

qmonnet (approved):
Looks all good from my side, thanks!

Documentation/operations/upgrade.rst
sayboras: We have auth.type attribute added in 1.13, but it's basically just to print a log line to indicate that mutual auth is required (i.e. no actual mutual auth happens). 

I can say that this existing attribute was having very less impact, and it's unlikely to have any users using it for actual workload, hence we decided to skip deprecation cycle and went directly with rename.

aanm (approved):

aanm:
The majority of the codeowneres reviewed the PR. Merging since it's blocking a lot of other PRs and this PR is simply a name refactoring.

Documentation/operations/upgrade.rst
qmonnet: Sounds good to me, thank you!

PR #25511, opened 2023-05-17
multi-pool: Determine IP pool based on `ipam.cilium.io/ip-pool` annotation

gandro:
This adds a mechanism which allows workloads to specify what IP pool their IP should be allocated from in multi-pool IPAM mode (see https://github.com/cilium/cilium/pull/22762 for details).

It will do that by first checking if the pod has a `ipam.cilium.io/ip-pool`  annotation, and if present, will return that pool to be used for IP allocation. If the pod does not have such an annotation, it will then check for the same annotation on the pod's namespace. If that is also not present, we fall back on the `default` pool.

:bulb: **Review per commit**

I have manually tested this on kind as follows:

```console
$ helm upgrade --install cilium ./install/kubernetes/cilium --namespace kube-system \
    --set debug.enabled=true \
    --set image.override=localhost:5000/cilium/cilium-dev:local \
    --set image.pullPolicy=IfNotPresent \
    --set operator.image.override=localhost:5000/cilium/operator-generic:local \
    --set operator.image.pullPolicy=IfNotPresent \
    --set operator.replicas=1 \
    --set hubble.relay.enabled=true \
    --set ipam.mode=multi-pool \
    --set "ipam.operator.multiPoolMap.default.ipv4CIDRs[0]=10.10.0.0/16" \
    --set "ipam.operator.multiPoolMap.default.ipv4MaskSize=27" \
    --set "ipam.operator.multiPoolMap.mars.ipv4CIDRs[0]=10.20.0.0/16" \
    --set "ipam.operator.multiPoolMap.mars.ipv4MaskSize=24" \
    --set "extraConfig.multi-pool-node-pre-alloc=default=64\,mars=64" \
    --set tunnel=disabled \
    --set autoDirectNodeRoutes=true \
    --set ipv4NativeRoutingCIDR=10.0.0.0/8 \
    --set endpointRoutes.enabled=true \
    --set-string extraConfig.enable-local-node-route=false \
    --set kubeProxyReplacement=strict \
    --set bpf.masquerade=true
$ k create ns cilium-test
$ k annotate ns cilium-test ipam.cilium.io/ip-pool=mars
$ cilium connectivity test
...
✅ All 41 tests (291 actions) successful, 8 tests skipped, 0 scenarios skipped.
$ k -n default get pods -o wide
NAME                         READY   STATUS    RESTARTS   AGE   IP            NODE           NOMINATED NODE   READINESS GATES
[...]

PR #22762, opened 2022-12-15
Add support for allocating PodCIDRs from multiple IPAM pools

gandro:
This adds support for IPAM pools. More specifically, this adds a new IPAM mode (`multi-pool`) which allows nodes to allocate PodCIDRs from different pools, depending on properties of the workload as defined by the user. An example where this will be useful is multi-homing (see https://github.com/cilium/cilium/issues/20129). 

Notably, this PR only adds the IPAM infrastructure which allows consumers inside Cilium-Agent to allocate PodCIDRs from non-default pools, it does not yet add any consumers which actually make use of this feature. This means that this PR is not going to be very useful on its own. At the moment, we require user to statically configure which IPAM pools a node should allocate from via the `multi-pool-node-pre-alloc` flag. This way of configuring a node's IPAM pools will likely be superseded by a custom network configuration CRD.

Overview of the current mechanism:

Pooled IPAM CiliumNode CRD

As with other IPAM modes, cilium-operator is responsible for global allocation decisions, i.e. allocating PodCIDRs to nodes. Each cilium-agent indicates its own demand in the CiliumNode CRD. Once it has received a PodCIDR for a given pool, it will then hand out individual IPs based on pool specified in the IP allocation request (typically received from the CNI plugin).

*Known Limitations*:
- Encryption and tunnel mode are not supported (blocked on https://github.com/cilium/cilium/issues/20170)
- In the current form, this will not set up local node routes, since the setup of routes for alternative networks will be handled by the multi-homing feature instead. For now, we requires endpoint routes being enabled.
- Cilium's built-in legacy masquerading does not support multiple pod CIDRs (c.f. https://github.com/cilium/cilium/issues/22273). To work around this, use either BPF masquerading if there is only one single native-routing-cidr. Otherwise, use ip-masq-agent (iptables or BPF work both)
- This version of the PR requires that pools are statically defined in the operator config, using the newly added `ipam.operator.multiPoolMap` Helm value. Support for managing via CRDs instead will be implemented in a follow-up PR (c.f. [`ippool-crd`](https://github.com/tklauser/cilium/tree/pr/tklauser/ippool-crd) branch here)

Joint work with @tklauser

*Review per commit*

Meta-issue: #24764

PR #25765, opened 2023-05-30
multi-pool: Support allocating from new IPAM pools on demand

gandro:
This PR adds support for dynamically allocating IPs from a new pool. Previously, we required the user to specify in advance what pool the agent would allocate from. This however is very limiting, since it's often not known in advance what workload a particular node will run.

Therefore, this PR adds a mechanism where upon observing a allocation attempt for a new pool, we update the local CiliumNode CRD to request additional IPs for that pool. This then allows the allocation request to be successful upon retry. This retry mechanism is similar to how Cilium IPAM also works e.g. in ENI mode (where if the local node is out of IPs, we fail the CNI ADD request and expect kubelet to re-try again later).

This PR also implicitly adds support for allocating from pools without a "pre-allocate" value. This allows users to configure very small pools for specific workloads and assign those few IPs only if they are actually used.

:information_source: **Review per commit**. The first few commits are mainly cleanup, the main logic is contained in the last two commits

For further context, see:

- https://github.com/cilium/cilium/pull/22762
- https://github.com/cilium/cilium/pull/25511
- https://github.com/cilium/cilium/issues/25470

CI and documentation will be added in a separate PR. See meta issue https://github.com/cilium/cilium/issues/25470

### DIFF
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -289,10 +289,11 @@ func initializeFlags() {
 	flags.StringSlice(option.IPv6PodSubnets, []string{}, "List of IPv6 pod subnets to preconfigure for encryption")
 	option.BindEnv(Vp, option.IPv6PodSubnets)
 
-	flags.Var(option.NewNamedMapOptions(option.IPAMMultiPoolNodePreAlloc, &option.Config.IPAMMultiPoolNodePreAlloc, nil),
-		option.IPAMMultiPoolNodePreAlloc, "List of IP pools which should be pre-allocated on this node")
-	flags.MarkHidden(option.IPAMMultiPoolNodePreAlloc)
-	option.BindEnv(Vp, option.IPAMMultiPoolNodePreAlloc)
+	flags.Var(option.NewNamedMapOptions(option.IPAMMultiPoolPreAllocation, &option.Config.IPAMMultiPoolPreAllocation, nil),
+		option.IPAMMultiPoolPreAllocation,
+		fmt.Sprintf("Defines how the minimum number of IPs a node should pre-allocate from each pool (default %s)", defaults.IPAMMultiPoolPreAllocation))
+	Vp.SetDefault(option.IPAMMultiPoolPreAllocation, defaults.IPAMMultiPoolPreAllocation)
+	option.BindEnv(Vp, option.IPAMMultiPoolPreAllocation)
 
 	flags.StringSlice(option.ExcludeLocalAddress, []string{}, "Exclude CIDR from being recognized as local address")
 	option.BindEnv(Vp, option.ExcludeLocalAddress)
--- a/pkg/ipam/multipool.go
+++ b/pkg/ipam/multipool.go
@@ -189,6 +189,7 @@ type multiPoolManager struct {
 	owner Owner
 
 	preallocatedIPsPerPool preAllocatePerPool
+	pendingIPsPerPool      *pendingAllocationsPerPool
 
 	pools        map[Pool]*poolPair
 	poolsUpdated chan struct{}
@@ -227,6 +228,7 @@ func newMultiPoolManager(conf Configuration, nodeWatcher nodeWatcher, owner Owne
 		owner:                  owner,
 		conf:                   conf,
 		preallocatedIPsPerPool: preallocMap,
+		pendingIPsPerPool:      newPendingAllocationsPerPool(),
 		pools:                  map[Pool]*poolPair{},
 		poolsUpdated:           make(chan struct{}, 1),
 		node:                   nil,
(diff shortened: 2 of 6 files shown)


daemon/cmd/daemon_main.go
```diff
@@ -288,10 +288,11 @@ func initializeFlags() {
 	flags.StringSlice(option.IPv6PodSubnets, []string{}, "List of IPv6 pod subnets to preconfigure for encryption")
 	option.BindEnv(Vp, option.IPv6PodSubnets)
 
-	flags.Var(option.NewNamedMapOptions(option.IPAMMultiPoolNodePreAlloc, &option.Config.IPAMMultiPoolNodePreAlloc, nil),
-		option.IPAMMultiPoolNodePreAlloc, "List of IP pools which should be pre-allocated on this node")
-	flags.MarkHidden(option.IPAMMultiPoolNodePreAlloc)
-	option.BindEnv(Vp, option.IPAMMultiPoolNodePreAlloc)
+	flags.Var(option.NewNamedMapOptions(option.IPAMMultiPoolPreAllocation, &option.Config.IPAMMultiPoolPreAllocation, nil),
+		option.IPAMMultiPoolPreAllocation,
+		fmt.Sprintf("Defines how the minimum number of IPs a node should pre-allocate from each pool (default %s)", defaults.IPAMMultiPoolPreAllocation))
```
meyskens, line 294: Given this is a string might `%q` not be more cleat to the user? `(default default=8)` vs `(default "default=8")`

meyskens (approved):
Code looks good to me, one tiny nit :) 
Functionality I am not the best person for to ask so if a better expert has time to also review :smile:

pkg/ipam/multipool.go
```diff
@@ -543,43 +596,59 @@ func (m *multiPoolManager) poolByFamilyLocked(poolName Pool, family Family) *pod
 	return nil
 }
 
-func (m *multiPoolManager) allocateNext(owner string, poolName Pool, family Family, syncUpstream bool) (*AllocationResult, error) {
+func (m *multiPoolManager) allocateNext(owner string, poolName Pool, family Family, syncUpstream bool) (result *AllocationResult, err error) {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
 
+	defer func() {
```
squeed, line 603: arguably, this defer is a bit more confusing then just pasting `upsertPendingAllocations` twice :-)

squeed (approved):

squeed:
One thought occurs to me: is the owner a pod or the CNI container ID?

If it's a container ID, then that is unique for every CNI ADD (even retries), so we will have a churn issue. If it's a pod, then we're OK.

gandro:
> One thought occurs to me: is the owner a pod or the CNI container ID?

Good point. It's the pod: https://github.com/cilium/cilium/blob/2588462e8fa6c7f077b07c547bb7f0ed4e01fb03/plugins/cilium-cni/main.go#L121

daemon/cmd/daemon_main.go
gandro: Thanks! Yes, I noticed this too, and I am torn on this. It's not really a string, it's a map. But `option.NewNamedMapOptions` does not allow me to define a default map, which is why I'm setting a default "string" value (which is later parsed as a map).

I've went back and forth between `%s` and `%q`. I ended up with `%s`, because other non-string types (like e.g. integers, durations and booleans) don't use quotes either: https://github.com/cilium/cilium/blob/d137d831a26ba72fa85db4972f32737d28068a9e/Documentation/cmdref/cilium-agent.md?plain=1#L17

tklauser (approved):

pkg/ipam/multipool.go
gandro: Yeah the tricky thing is that `m.k8sUpdater.TriggerWithReason` now needs to be called also in the error case. But I guess I could keep the upstreamSync logic in defer and have `upsertPendingAllocations` called inline with the code.

Might change it once all reviews are in.

daemon/cmd/daemon_main.go
christarazi: Another note here is that you don't need to explicitly mention the default here because `flags` does it for you, AFAIR.

pkg/ipam/multipool.go
```diff
@@ -347,61 +349,112 @@ func neededIPCeil(numIP int, preAlloc int) int {
...
-	m.mutex.Lock()
-	newNode := m.node.DeepCopy()
-	requested := []types.IPAMPoolRequest{}
-	allocated := []types.IPAMPoolAllocation{}
+// computeNeededIPsPerPoolLocked computes how many IPs we want to request from
+// the operator for each pool. The formula we use for each pool is basically
+//
+//	  neededIPs = roundUp(inUseIPs + pendingIPs + preAllocIPs, preAlloc)
+//
+//		  inUseIPs 		Number of IPs that are currently actively in use
```
christarazi: Spacing is messed up here

christarazi (approved):

daemon/cmd/daemon_main.go
gandro: Unfortunately, for some reason it doesn't here. See the generated docs, there is no duplicate `default`.

pkg/ipam/multipool.go
gandro: Thanks for pointing it out, seems like I mixed tabs and spaces. Fixed!

pkg/ipam/multipool.go
gandro: Made it hopefully less confusing!

PR #25766, opened 2023-05-30
helm: Avoid error in IDE due to .range keyword

sayboras:
### Description 

As range is a keyword in jinja template, using the expression with usual attribute name (e.g. .Values.nodePort.range) will cause the error in IDE. This issue makes it hard for deployment helm for normal operation like checking where the flag is used.

This commit is to use alternative way to get around the issue with get function.

Testing was done as per below

```
$ cat temp_values.yaml
nodePort:
  range: "30000,32767"
$ helm template cilium install/kubernetes/cilium -f temp_values.yaml | grep node-port-range
  node-port-range: "30000,32767"
  enable-auto-protect-node-port-range: "true"
```

image


kaworu (approved):
:ninja:

michi-covalent (approved):

PR #21333, opened 2022-09-16
Remove references to node encryption

pchaigno:
This pull request hides the node encryption flag and removes a missed reference to the feature in the docs.

PR #19401, opened 2022-04-11
Add WireGuard host2host and LB encryption

brb:
This PR adds support for node-to-node encryption to WireGuard. To achieve this, we've completely changed the WireGuard integration in the datapath. Previously, WireGuard support was implemented by marking packets to be encrypted in `"from-container"` and redirecting it to the WireGuard tunnel via a hostns IP rule. This worked fine for traffic originating in pods - but for node-to-node traffic, we need to redirect the packets on the outgoing network interface. Thus, the new implementation attaches `bpf_host` to the outgoing device and redirects packets to the WireGuard tunnel from there. See commit descriptions for more details.

On the agent side, there are also changes to the implementation. Previously, the datapath assumed that any IPCache entry with an associated tunnel endpoint would need encryption. To determine if we need to encryption traffic to a remote endpoint, we now rely on the `encrypt_key` field instead. This allows us to more precisely track if traffic to a particular destination needs to be encrypted, and allows certain nodes to opt out of encryption (see below). The agent code has been updated to populate the CiliumEndpoint and CiliumNode CRDs with a static non-zero `EncryptKey` value if encryption for those resources is enabled.

Additional points worth noting:
 - This new implementation now supports L7 proxy with WireGuard (due to it not requiring skb marks pre-encryption)
 - The agent opts out of node-to-node encryption if it is running on a kube-apiserver control plane node. This is done to
ensure worker nodes are always able to communicate with the kube-apiserver node, in order to be able to manage their own encryption key. See docs for more details.
 - CI is handled in the new CI-Datapath suite via Cilium-CLI. The run with the new testsuite (c.f. https://github.com/cilium/cilium-cli/pull/1308) was successful: 
   :heavy_check_mark:  https://github.com/cilium/cilium/actions/runs/3884806887/jobs/6627856819

:information_source:  **Please see commit messages for many more details**

Joint work between @gandro and @brb

Follow-ups, to be done in separate PRs:
- [ ] (follow-up) Ability to store private key persistently (as an alternative to opting out of node-to-node encryption for control-plane nodes)

PR #25770, opened 2023-05-30
helm: `nodeEncryption` is only supported with WireGuard

gandro:
This commit fixes an outdated comment in our Helm `values.yaml` file. Originally, node-to-node encryption was a beta feature only supported with IPSec. However, since then, we have removed the IPSec support (#21333), but have added support in WireGuard instead (#19401).

Fixes: 5e980372f3d6 ("cmd: Unhide node-encryption flag")


pchaigno (approved):

sayboras (approved):
Thanks

PR #25772, opened 2023-05-30
.github: run scruffy for cilium/cilium only

aanm:
This script is only intent to run on Cilium organization. We need to check for the github repository to avoid running it on forks and accidentally DDoS quay.io.

Also, update the docker image to the v0.0.3 since the previous docker images were deleted to prevent the execution of the github action by the forks.

### DIFF
--- a/.github/workflows/ci-images-garbage-collect.yaml
+++ b/.github/workflows/ci-images-garbage-collect.yaml
@@ -9,13 +9,14 @@ permissions: read-all
 
 jobs:
   scruffy:
+    if: github.repository_owner == 'cilium'
     name: scruffy
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
         with:
           fetch-depth: 0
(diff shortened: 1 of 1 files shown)


ldelossa (approved):

joestringer (approved):

michi-covalent (approved):
😇

.github/workflows/ci-images-garbage-collect.yaml
```diff
@@ -9,6 +9,7 @@ permissions: read-all
 
 jobs:
   scruffy:
+    if: github.repository == 'cilium/cilium'
```
bleggett:
```suggestion
    if: github.repository_owner == 'cilium'
```

nit: this probably could be just a check against the repo owner:

```
github.repository_owner == 'cilium'
```

(it would be good to do this check as a matter of course in more jobs, so we could [drop the doc guidance](https://docs.cilium.io/en/stable/contributing/development/contributing_guide/#clone-and-provision-environment) that suggests people disable Actions in their forks to avoid junk runs, this is probably not the only culprit)

viktor-kurchenko (approved):

PR #25784, opened 2023-05-31
ipsec: Fix leak of XFRM OUT policies

pchaigno:
This pull request updates our cleanup logic for XFRM config when remote node are deleted, to fix a leak in Azure and ENI IPAM modes.

```release-note
Fix a bug due to which we would leak Linux XFRM policies, potentially leading to increased CPU consumption, when IPsec is enabled with Azure or ENI IPAM.
```


borkmann (approved):
lgtm, thanks Paul!

jschwinger233 (approved):

PR #25792, opened 2023-05-31
bpf: remove MapInfo, DumpParser and MapKey/Value DeepCopy

ti-mo:
Please review per commit.

High-level overview of this change:

- Replace MapInfo with MapSpec and accessors on Map
- Reduce MapKey and MapValue interfaces to just fmt.Stringer and New(), making it significantly easier to implement new key/value types
- Make OpenMap more usable by taking MapKey and MapValue arguments
- Remove k8s deepcopy machinery from all MapKey and MapValue types

```release-note
bpf: remove MapInfo, DumpParser and MapKey/Value DeepCopy
```


pkg/bpf/map.go
```diff
@@ -10,33 +10,8 @@ import (
 	"github.com/cilium/cilium/pkg/controller"
 )
 
-// MapType is an enumeration for valid BPF map types
-type MapType int
-
 // This enumeration must be in sync with enum bpf_map_type in <linux/bpf.h>
```
lmb: Nit: out of date.

pkg/bpf/bpf.go
```diff
@@ -38,12 +40,15 @@ func DisableMapPreAllocation() {
...
+func GetPreAllocateMapFlags(t ebpf.MapType) uint32 {
+	switch t {
+	// LPM Tries don't support preallocation.
+	case ebpf.LPMTrie:
 		return BPF_F_NO_PREALLOC
-	case t.requiresPreallocation():
-		return 0
+	// Support disabling preallocation for these map types.
+	case ebpf.Hash, ebpf.PerCPUHash, ebpf.HashOfMaps:
+		return atomic.LoadUint32(&preAllocateMapSetting)
```
lmb, line 50: It's kind of weird that this returns preAllocateMapSetting directly. This only works because BPF_F_NO_PREALLOC is 1 like the constant in DisableMapPreAllocation.

pkg/bpf/map_linux.go
```diff
@@ -81,9 +67,11 @@ type cacheEntry struct {
 }
 
 type Map struct {
-	MapInfo
-	m         *ebpf.Map
-	innerSpec *ebpf.MapSpec
+	m    *ebpf.Map
+	spec *ebpf.MapSpec
```
lmb, line 60: Nit: `// Nil after the map has been created`?

pkg/bpf/map_linux.go
```diff
@@ -1142,17 +1166,17 @@ func (m *Map) resolveErrors(ctx context.Context) error {
 // match, deletes the map.
 //
 // Returns true if the map was upgraded.
-func (m *Map) CheckAndUpgrade(desired *MapInfo) bool {
-	desired.Flags |= GetPreAllocateMapFlags(desired.MapType)
+func (m *Map) CheckAndUpgrade(desired *Map) bool {
```
lmb, line 1120: Would it make sense to pass `*ebpf.MapSpec` here?

pkg/bpf/map_linux.go
```diff
@@ -570,33 +560,27 @@ type MapValidator func(path string) (bool, error)
...
-		// TODO(tb): Remove DumpParser altogether if it's only ever used with
-		// bpf.ConvertKeyValue. It does binary.Read and Iterate().Next() already
-		// does that ootb.
-		var err error
-		mk, mv, err = m.DumpParser(key, value, mk, mv)
-		if err != nil {
-			return err
-		}
+	for i.Next(mk, mv) {
+		cb(mk, mv)
```
lmb, line 574: The function doc makes it sound like `cb` is responsible for copying mk and mv if it wants to hold onto them, but here were creating new copies for each iteration. Is the doc incorrect or should this not create copies?

pkg/bpf/map_linux.go
```diff
@@ -627,14 +611,23 @@ func (m *Map) DumpWithCallbackIfExists(cb DumpCallback) error {
...
+	if cb == nil {
+		return errors.New("empty callback")
+	}
+
+	if stats == nil {
+		return errors.New("stats is nil")
+	}
+
 	var (
-		prevKey    = make([]byte, m.KeySize())
```
lmb: Using `[]byte` for prevKey / nextKey is a lot cheaper even when using the library, since it avoids marshaling. See the gymnastics that `ebpf.MapIterator` does.

pkg/bpf/map_linux.go
```diff
@@ -980,44 +962,26 @@ func (m *Map) DeleteAll() error {
...
-		err := m.m.Delete(key)
-
-		mk, _, err2 := m.DumpParser(key, []byte{}, mk, mv)
-		if err2 == nil {
-			m.deleteCacheEntry(mk, err)
-		} else {
-			log.WithError(err2).Warningf("Unable to correlate iteration key %v with cache entry. Inconsistent cache.", key)
-		}
+	for i.Next(mk, mv) {
+		err := m.m.Delete(mk)
```
lmb, line 961: I guess this will hit the "concurrent delete from hash map resets iteration" case, no? Is there a test that checks we actually delete all elements?

pkg/maps/ctmap/ctmap.go
```diff
@@ -709,7 +709,7 @@ func DeleteIfUpgradeNeeded(e CtEndpoint) {
 		}
 		scopedLog := log.WithField(logfields.Path, path)
 
-		oldMap, err := bpf.OpenMap(path)
+		oldMap, err := bpf.OpenMap(path, nil, nil)
```
lmb, line 714: Why is `nil, nil` okay here? Maybe worth documenting on OpenMap.

pkg/bpf/bpf.go
tommyp1ckles: nit: seems like a good use for atomic.Bool?

pkg/maps/bwmap/bwmap.go
```diff
@@ -43,8 +41,7 @@ type EdtInfo struct {
 	Pad             [4]uint64 `align:"pad"`
 }
 
-func (v *EdtInfo) GetValuePtr() unsafe.Pointer { return unsafe.Pointer(v) }
-func (v *EdtInfo) String() string              { return fmt.Sprintf("%d", int(v.Bps)) }
+func (v *EdtInfo) String() string { return fmt.Sprintf("%d", int(v.Bps)) }
```
tommyp1ckles: I know this is inherited code, but do we need to cast down to a signed int, seems like a recipe for trouble.

pkg/bpf/bpf.go
```diff
@@ -38,12 +40,15 @@ func DisableMapPreAllocation() {
...
-	case !t.allowsPreallocation():
+func GetPreAllocateMapFlags(t ebpf.MapType) uint32 {
+	switch t {
+	// LPM Tries don't support preallocation.
+	case ebpf.LPMTrie:
 		return BPF_F_NO_PREALLOC
-	case t.requiresPreallocation():
-		return 0
+	// Support disabling preallocation for these map types.
+	case ebpf.Hash, ebpf.PerCPUHash, ebpf.HashOfMaps:
```
tommyp1ckles, line 49: question - we don't support this for trie type anymore?

pkg/bpf/map_linux.go
```diff
@@ -570,33 +560,27 @@ type MapValidator func(path string) (bool, error)
...
 	}
 
 	m.lock.RLock()
 	defer m.lock.RUnlock()
 
-	key := make([]byte, m.KeySize())
-	value := make([]byte, m.ValueSize())
-
+	// Don't need deep copies here, only fresh pointers.
 	mk := m.MapKey.DeepCopyMapKey()
```
tommyp1ckles: why don't we just call `New()` here?

pkg/bpf/map_linux.go
```diff
@@ -570,33 +560,27 @@ type MapValidator func(path string) (bool, error)
...
-		if err != nil {
-			return err
-		}
+	for i.Next(mk, mv) {
+		cb(mk, mv)
 
-		if cb != nil {
-			cb(mk, mv)
-		}
+		mk = m.MapKey.DeepCopyMapKey()
```
tommyp1ckles: why do we need to make copies here (following the callback invocation)?

ti-mo: Good question! Handing out copies to callbacks is what the existing implementation does (even though docstrings claim the opposite), and we don't want to change that behaviour in this PR.

In the next cycle, we'll remove the `Dump*` concept altogether in favor of MapIterator, which has semantics closer to those of Go's `for` (so, unmarshal the next key and value into a caller-provided pointer argument). This will remove the ambiguity (and improve performance!), but requires going through the existing codebase to audit and potentially fix up all spots where callback arguments are stuffed directly into slices/maps etc. without copying.

pkg/bpf/map_linux.go
ti-mo: That's changed in a subsequent commit. :) Essentially, DeepCopyMap* was repurposed to mean New().

pkg/bpf/bpf.go
ti-mo: Could you clarify the question? Tries don't support preallocation.

pkg/maps/bwmap/bwmap.go
ti-mo: I'm trying to avoid altering behaviours like these in this PR. Agree overall, this was likely copypasta'd, but wouldn't cause anything worse than a cosmetical bug.

pkg/bpf/bpf.go
ti-mo: The fact that this is a global to begin with is a red flag. We'll need to sort this out when map creation in `pkg/bpf` takes MapSpec. Currently, the global is only manipulated from tests.

With MapSpec in the picture, we'd only set the default map flags based on option.Config or leave it untouched if the caller populates the flags explicitly in the spec (such as from tests, etc.).

pkg/bpf/map_linux.go
ti-mo: MapSpec is no longer available after a Map had OpenOrCreate() called on it, or when it was opened from a pin. CheckAndUpgrade will be removed in 1.15. Calling Recreate() should be the responsibility of individual subsystems, as only they possess the knowledge of whether or not recreating a map causes data loss, and if so, it needs to take care of any migration, logging, etc. This is was one responsibility too many that was delegated to pkg/bpf.

pkg/bpf/map_linux.go
ti-mo: I'll fix the doc, we're changing this (and removing Dump* altogether) in the next cycle.

pkg/bpf/bpf.go
tommyp1ckles: Think I was referencing this deleted code: https://github.com/cilium/cilium/blob/main/pkg/bpf/map.go#L106

pkg/bpf/map_linux_test.go
```diff
@@ -774,3 +776,39 @@ func (s *BPFPrivilegedTestSuite) TestCreateUnpinned(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(got, checker.DeepEquals, v)
 }
+
+func BenchmarkMapLookup(b *testing.B) {
```
tommyp1ckles, line 721: Q: What's the context for this benchmark, are we proving that the ebpf-go has better overhead over the old Cilium bpf.Map?

pkg/bpf/map_linux.go
```diff
@@ -124,54 +112,99 @@ type Map struct {
 	events *eventsBuffer
 }
 
-// NewMap creates a new Map instance - object representing a BPF map
-func NewMap(name string, mapType MapType, mapKey MapKey, keySize int,
-	mapValue MapValue, valueSize, maxEntries int, flags uint32, dumpParser DumpParser) *Map {
+func (m *Map) Type() ebpf.MapType {
+	if m.m != nil {
+		return m.m.Type()
```
tommyp1ckles, line 103: question: it will follow this code path if the map has been created, or use the spec struct otherwise?

pkg/bpf/map_linux.go
```diff
@@ -286,7 +319,7 @@ func (m *Map) updatePressureMetric() {
 		return
 	}
 
-	pvalue := float64(len(m.cache)) / float64(m.MaxEntries)
+	pvalue := float64(len(m.cache)) / float64(m.MaxEntries())
```
tommyp1ckles, line 306: It looks like the `MaxEntries()` func can return 0 in some cases, not sure what happens here in that situation.

pkg/bpf/map_linux.go
```diff
@@ -124,54 +112,99 @@ type Map struct {
...
+		return m.spec.Flags
 	}
+	return 0
+}
+
+// NewMap creates a new Map instance - object representing a BPF map
+func NewMap(name string, mapType ebpf.MapType, mapKey MapKey, mapValue MapValue,
+	maxEntries int, flags uint32, dumpParser DumpParser) *Map {
+
+	keySize := reflect.TypeOf(mapKey).Elem().Size()
```
tommyp1ckles, line 155: note: this panics if mapKey is not a pointer

pkg/bpf/map_linux_test.go
ti-mo: Yes, there is a rather large performance regression due to getting rid of unsafe.Pointer in favor of reflect-based unmarshaling in most of the maps. We're currently addressing this in ebpf-go by adding a bytes.Buffer pool on the key marshaling path in the short term and investigating in which cases we can pass unsafe.Pointers to the kernel internally. The latter will require a few more cycles and won't make it in time for the 1.14 freeze.

ti-mo: ebpf-go PR for this is up: https://github.com/cilium/ebpf/pull/1053

asauber (approved):
CLI-related changes LGTM

tommyp1ckles (approved):

pkg/bpf/map_linux.go
ti-mo: True, but in this case, that's intended. Each GetNext call will return the first key, which we then immediately delete, until all records are gone. The process continues until GetNext is called on a map that has no keys, which could never happen if the insertion rate is high enough. This method is not advertised to deal with the termination problem, though, and we're not going to change it in this series.

ti-mo: Looking at the code again, I noticed I put unnecessary `m.{key,value}.New()` etc. in the loop, I've removed those.

pkg/bpf/map_linux.go
ti-mo: Since we need to pass all keys and values to cb() in an unmarshaled state, what would this gain us?

pkg/maps/ctmap/ctmap.go
ti-mo: I've added a comment to the call site, doesn't make much sense to call this out in the method doc IMO.

pkg/bpf/map_linux.go
ti-mo: This could occur before this change as well, I don't think we ever validated this piece of input. Seeing as `updatePressureMetric()` is only ever invoked after calls to `Map.open()`, which is guaranteed to fail with maxentries 0, I don't think this is an issue.

pkg/bpf/map_linux.go
ti-mo: Yep, that's the idea!

pkg/bpf/map_linux.go
ti-mo: You've flagged this in a previous PR as well. Currently, all code assumes that MapKey and MapValue are implemented on pointer receivers. Panicking isn't a friendly way of enforcing that, but this also isn't too high up on the list of things to tackle. We can make this friendlier when NewMap() is changed to take a MapSpec, but eventually, bpf.Map should no longer hold MapKey and MapValue, so this would be somewhat wasted effort.

pkg/bpf/map_linux_test.go
ti-mo: Upstream PR merged and pulled into vendor in this PR.

pkg/bpf/map_linux.go
lmb: Just wanted to raise this as an FYI, all good if you've done the due diligence.

lmb (approved):

rolinh (approved):

jrajahalme (approved):
LGTM for my "CODEOWNERS"

PR #25795, opened 2023-05-31
bpf,datapath: read jiffies from /proc/schedstat

### LINKED ISSUE #25463: BPF jiffies never enabled due to missing `/proc/timer_list`
https://github.com/cilium/cilium/pull/11434 introduced a new flag `enable-bpf-clock-probe` to use jiffies instead of `bpf_ktime_get_sec` which can be expensive. This feature however always fails to enable due to the missing `/proc/timer_list` file. I think that's because we don't mount that file in the agent container.

cc @borkmann

ti-mo:
The Cilium agent has been throwing the 'Auto-disabling "enable-bpf-clock-probe" feature since kernel doesn't expose /proc/timer_list' warning for a while now.

Since /proc/timer_list is not available under the default k8s SecurityContext and Docker also masks the file by default, read it from /proc/schedstat instead.

Reuse the existing parsing code from probes.KernelHZ to obtain the value.

Fixes: https://github.com/cilium/cilium/issues/25463

Backporting to 1.13 only since 1.12 doesn't have the new /proc/schedstat reader infra yet.

```release-note
bpf,datapath: read jiffies from /proc/schedstat
```

### DIFF
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -2044,11 +2044,11 @@ func initClockSourceOption() {
 
 	if option.Config.EnableBPFClockProbe {
 		if probes.HaveProgramHelper(ebpf.XDP, asm.FnJiffies64) == nil {
-			t, err := bpf.GetJtime()
+			t, err := probes.Jiffies()
 			if err == nil && t > 0 {
 				option.Config.ClockSource = option.ClockSourceJiffies
 			} else {
-				log.WithError(err).Warningf("Auto-disabling %q feature since kernel doesn't expose %q.", option.EnableBPFClockProbe, bpf.TimerInfoFilepath)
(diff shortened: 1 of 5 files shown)


rgo3 (approved):
Probably should have asked that back when we introduced the probe that relies on `/proc/schedstat`: Should we add `CONFIG_SCHEDSTATS=y` as a system requirement (for certain features) somewhere? I just realized that it's not set on my kernel and I'm missing `/proc/schedstat`.

borkmann (approved):

pchaigno:
> Probably should have asked that back when we introduced the probe that relies on `/proc/schedstat`: Should we add `CONFIG_SCHEDSTATS=y` as a system requirement (for certain features) somewhere? I just realized that it's not set on my kernel and I'm missing `/proc/schedstat`.

I think we should document it at https://docs.cilium.io/en/stable/operations/system_requirements/#linux-kernel.

PR #21052, opened 2022-08-23
docs: Update ToServices docs section

joestringer:
Update this section to highlight the limitation, improve wording, and
remove future-looking statements.

Issue #20067, opened 2022-06-02
Policy with `k8sService` on port 443 allows traffic on port 443 to `world`


arjantop-cai:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Policy defined:
```
apiVersion: cilium.io/v2
kind: CiliumNetworkPolicy
metadata:
  name: my-policy
spec:
  endpointSelector: {}
  ingress:
    - {}
  egress:
    - toServices:
        - k8sService:
            namespace: ns
            serviceName: svc
      toPorts:
        - ports:
            - port: "443"
              protocol: TCP
```

Expected: Only traffic to service svc.ns on port 443 is allowed
Result: All traffic to "world" on port 443 is allowed

Observed in local k3d install and EKS

### Cilium Version

1.10.x and 1.11.x

### Kernel Version

```
Linux hostname 5.4.188-104.359.amzn2.x86_64 #1 SMP Thu Apr 14 20:53:13 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

### Kubernetes Version

```
v1.21.5-eks-9017834
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

arjantop-cai:
Reproduction steps using sw-app example:
```
kubectl create -ntest -f https://github.com/cilium/cilium/blob/master/examples/minikube/http-sw-app.yaml
```

Requests to world:80 and world:443 work:
```
kubectl -ntest exec -it tiefighter -- curl "http://google.com"
kubectl -ntest exec -it tiefighter -- curl "https://google.com"
```

Apply the policy:
```
apiVersion: "cilium.io/v2"
kind: CiliumNetworkPolicy
metadata:
  name: "rule1"
spec:
  endpointSelector:
    matchLabels:
      org: empire
      class: tiefighter
  ingress:
    - {}
  egress:
    - toServices:
        - k8sService:
            namespace: test
            serviceName: deathstar
      toPorts:
        - ports:
            - port: "80"
              protocol: TCP
    - toEndpoints:
        - matchLabels:
            io.kubernetes.pod.namespace: kube-system
            k8s-app: kube-dns
      toPorts:
        - ports:
            - port: "53"
              protocol: UDP
          rules:
            dns:
              - matchPattern: "*"

```

Request to world:80 works (the one specified in the policy `toPorts`) and traffic to world:443 is dropped (not specified in the policy)
```
[...]

PR #25796, opened 2023-05-31
Change wording on toServices limitations (see #20067)

atykhyy:
#21052 updated Cilium documentation to say that, in network policy rules, `toServices` statements _cannot_ be combined with `toPorts` statements. I believe it would be more informative for users of Cilium to say (following RFC 2119) that `toServices` _must not_ be combined with `toPorts`, as technically Cilium accepts such a configuration as valid but handles it in the unexpected and unuseful manner described in #20067.

### DIFF
--- a/Documentation/security/policy/language.rst
+++ b/Documentation/security/policy/language.rst
@@ -304,8 +304,9 @@ have ``external:yes`` set as the label.
 Limitations
 ~~~~~~~~~~~
 
-``toServices`` statements cannot be combined with ``toPorts`` statements in the
-same rule.
+``toServices`` statements must not be combined with ``toPorts`` statements in the
+same rule. In the presence of a ``toPorts`` statement, ``toServices`` does nothing
+and as a result all egress traffic to port(s) specified by ``toPorts`` is allowed.
 
 .. _Entities based:
 
--- a/Documentation/security/policy/language.rst
+++ b/Documentation/security/policy/language.rst
@@ -305,8 +305,7 @@ Limitations
 ~~~~~~~~~~~
 
 ``toServices`` statements must not be combined with ``toPorts`` statements in the
-same rule. In the presence of a ``toPorts`` statement, ``toServices`` does nothing
-and as a result all egress traffic to port(s) specified by ``toPorts`` is allowed.
+same rule. If a rule combines both these statements, the policy is rejected.
 
 .. _Entities based:
 
--- a/pkg/policy/api/rule_validation.go
+++ b/pkg/policy/api/rule_validation.go
@@ -190,7 +190,7 @@ func (e *EgressRule) sanitize() error {
 		"ToCIDRSet":   true,
 		"ToEndpoints": true,
 		"ToEntities":  true,
-		"ToServices":  true,
+		"ToServices":  false, // see https://github.com/cilium/cilium/issues/20067
 		"ToFQDNs":     true,
 		"ToGroups":    true,
 	}
--- a/pkg/policy/api/rule_validation_test.go
+++ b/pkg/policy/api/rule_validation_test.go
@@ -627,7 +627,7 @@ func (s *PolicyAPITestSuite) TestToServicesSanitize(c *C) {
 	}
 
 	err := toServicesL3L4.Sanitize()
-	c.Assert(err, IsNil)
+	c.Assert(err, NotNil)
 
 }
 


qmonnet (approved):
Looks good, thank you!

christarazi (commented):
Thanks for the PR!

Checkpatch (checks your commits) is failing with 

```
   Warning: WARNING:TYPO_SPELLING: 'unuseful' may be misspelled - perhaps 'useless'?
  #11: 
  as valid but handles it in the unexpected and unuseful manner described
                                                ^^^^^^^^
```

christarazi (approved):

joestringer:
I agree about the "must not" RFC 2119 wording. Personally I'd avoid making any statements about how the limitation behaves, since it's a limitation - we don't want users to rely on this, this is not API, it is subject to change without warning. Ideally someone will pick this up and fix the limitation, and at that point I'd hope that users aren't assuming that the limitation always has these semantics.

But :shrug: I don't feel super strongly about this. Just want to say, we reserve the right to break this statement at any time.

atykhyy:
I agree that documenting incorrect behavior is not a very good idea, even if it's explicitly stated that the behavior is incorrect. On the other hand a simple statement of limitation feels insufficiently scary: the rule apparently works (makes the service accessible), but has non-obvious and potentially dangerous additional effects. Perhaps it's best to simply make rules with both `toServices` and `toPorts` fail validation? A couple lines in `pkg\policy\api\rule_validation.go:sanitize(*EgressRule)` would suffice.

nathanjsweet (approved):

ti-mo:
@atykhyy Please rebase to re-run Travis and to pick up some fixes for flakes etc.

ti-mo:
@atykhyy one test failing:
```
--- FAIL: Test/PolicyAPITestSuite#01/TestToServicesSanitize (0.00s)

    rule_validation_test.go:630: 

        ... value *errors.errorString = &errors.errorString{s:"Combining ToServices and ToPorts is not supported yet"} ("Combining ToServices and ToPorts is not supported yet")
```

atykhyy:
@ti-mo Right. It's the test which checks for toServices+toPorts policy which fails validation per this PR. I have reversed the test result.

ti-mo:
Sorry for the trouble, looks like image builds are failing, something probably changed on `main`. Could you give this another rebase? :pray:

fy-openai:
@atykhyy 
we just ran into this during the upgrade from 1.13.4 to 1.15.6. am wondering what does `fail the validation` mean here.

What I observed is cilium went from allowing all traffic destined for that one port to allowing all traffic because now it simply rejects that CNP and the default behavior is Allow All.

atykhyy:
That's exactly what 'fail validation' means: Cilium rejects policy objects set up that way. Since Cilium does not, to my knowledge, provide a Kubernetes admission controller, one can create all sorts of `CiliumNetworkPolicy` objects that are invalid. Cilium operator marks them as such in their `status` section.

fy-openai:
Nothing shows up in CNP's `status` section in my case though.. I only accidentally found out this when looking at logs (level=warning, not even errors).

To me, "silently" refusing to apply any policy and ending up allowing all is worse than allowing all for ports in ToPorts. The former is definitely more permissive and still not straightforward to be caught by developers as `kubectl apply` simply succeed.
I think an invalid spec should either be rejected upfront at creation time or marked explicitly as so like you mentioned through status on CNPs.

atykhyy:
Hm. What happens when you try something like this?
```
$ <<EOF kubectl apply -f -
apiVersion: cilium.io/v2
kind: CiliumNetworkPolicy
metadata:
  name: test2
  namespace: default
specs:
- egress:
  - toPorts:
    - ports:
      - port: "53"
        protocol: ANY
    toServices:
    - k8sService:
        namespace: kube-system
        serviceName: kube-dns
  endpointSelector:
    matchLabels:
      k8s:io.kubernetes.pod.namespace: default
EOF
ciliumnetworkpolicy.cilium.io/test2 created
$ sleep 5
$ kubectl get ciliumnetworkpolicy test2 -o yaml
... # snip
status:
  conditions:
  - lastTransitionTime: "2024-07-25T05:49:57Z"
    message: Combining ToServices and ToPorts is not supported yet
    status: "False"
    type: Valid
```

fy-openai:
```
➜  lab14 git:(master) ✗ kubectl logs ds/cilium -n cilium  | grep warning | tail
Found 10 pods, using pod/cilium-jhm9s
time="2024-07-25T08:55:49Z" level=warning msg="Unable to add CiliumNetworkPolicy" ciliumNetworkPolicyName=test2 error="Invalid CiliumNetworkPolicy specs: Combining ToServices and ToPorts is not supported yet" k8sApiVersion= k8sNamespace=default subsys=k8s-watcher
time="2024-07-25T08:55:49Z" level=warning msg="Unable to add CiliumNetworkPolicy" ciliumNetworkPolicyName=test2 error="Invalid CiliumNetworkPolicy specs: Combining ToServices and ToPorts is not supported yet" k8sApiVersion= k8sNamespace=default subsys=k8s-watcher
time="2024-07-25T08:55:50Z" level=warning msg="Unable to add CiliumNetworkPolicy" ciliumNetworkPolicyName=test2 error="Invalid CiliumNetworkPolicy specs: Combining ToServices and ToPorts is not supported yet" k8sApiVersion= k8sNamespace=default subsys=k8s-watcher
time="2024-07-25T08:55:50Z" level=warning msg="Unable to add CiliumNetworkPolicy" ciliumNetworkPolicyName=test2 error="Invalid CiliumNetworkPolicy specs: Combining ToServices and ToPorts is not supported yet" k8sApiVersion= k8sNamespace=default subsys=k8s-watcher
time="2024-07-25T08:55:52Z" level=warning msg="Unable to add CiliumNetworkPolicy" ciliumNetworkPolicyName=test2 error="Invalid CiliumNetworkPolicy specs: Combining ToServices and ToPorts is not supported yet" k8sApiVersion= k8sNamespace=default subsys=k8s-watcher
time="2024-07-25T08:55:54Z" level=warning msg="Unable to add CiliumNetworkPolicy" ciliumNetworkPolicyName=test2 error="Invalid CiliumNetworkPolicy specs: Combining ToServices and ToPorts is not supported yet" k8sApiVersion= k8sNamespace=default subsys=k8s-watcher
time="2024-07-25T08:55:59Z" level=warning msg="Unable to add CiliumNetworkPolicy" ciliumNetworkPolicyName=test2 error="Invalid CiliumNetworkPolicy specs: Combining ToServices and ToPorts is not supported yet" k8sApiVersion= k8sNamespace=default subsys=k8s-watcher
time="2024-07-25T08:56:10Z" level=warning msg="Unable to add CiliumNetworkPolicy" ciliumNetworkPolicyName=test2 error="Invalid CiliumNetworkPolicy specs: Combining ToServices and ToPorts is not supported yet" k8sApiVersion= k8sNamespace=default subsys=k8s-watcher
time="2024-07-25T08:56:30Z" level=warning msg="Unable to add CiliumNetworkPolicy" ciliumNetworkPolicyName=test2 error="Invalid CiliumNetworkPolicy specs: Combining ToServices and ToPorts is not supported yet" k8sApiVersion= k8sNamespace=default subsys=k8s-watcher
time="2024-07-25T08:57:11Z" level=warning msg="Unable to add CiliumNetworkPolicy" ciliumNetworkPolicyName=test2 error="Invalid CiliumNetworkPolicy specs: Combining ToServices and ToPorts is not supported yet" k8sApiVersion= k8sNamespace=default subsys=k8s-watcher
```

fy-openai:
weird. does this depend on some versions? like kube?

fy-openai:
> I don't know enough to say. But on the off chance it's simple, do you have `cilium-operator` running, and is it the new version? If I understand correctly, `cilium-operator` pod is the one responsible for marking network policy objects with that status, not `cilium-agent`.

Yes I do. They're on the same version as cilium-agent and run just fine.

PR #25797, opened 2023-05-31
Add improvements in Conformance Runtime

aanm:
Add some improvements in the Conformance runtime tests to make it easier to understand which test failed.

An example of such output can be found [here](https://github.com/cilium/cilium/actions/runs/5133502410)

image

### DIFF
--- a/.github/workflows/conformance-runtime.yaml
+++ b/.github/workflows/conformance-runtime.yaml
@@ -358,3 +358,9 @@ jobs:
           path: |
             cilium-junits/*.xml
           retention-days: 2
+
+      - name: Publish Test Results As GitHub Summary
+        if: ${{ always() }}
+        uses: aanm/junit2md@86789108ea0fd9be947a4232dd560e2fa3b9467a # v0.0.1
+        with:
+          junit-directory: "cilium-junits"
--- a/.github/workflows/conformance-runtime.yaml
+++ b/.github/workflows/conformance-runtime.yaml
@@ -321,6 +321,18 @@ jobs:
             export GOTEST_FORMATTER="/root/go/bin/go-junit-report -set-exit-code -iocopy -out test/runtime.xml"
             make tests-privileged NO_COLOR=1
 
+      - name: Debug failure on VM
+        # Only debug the failure on the LVH that have Cilium running as a service,
+        # which is 'agent' and 'datapath' focus.
+        if:  ${{ !success() && (matrix.focus == 'agent' || matrix.focus == 'datapath') }}
+        timeout-minutes: 10
+        uses: cilium/little-vm-helper@0b7d7157dae56a44dc531c852e7756dc671071af # v0.0.3
+        with:
+          provision: 'false'
+          cmd: |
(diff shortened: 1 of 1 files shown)


.github/workflows/conformance-runtime.yaml
```diff
@@ -322,6 +321,16 @@ jobs:
             export GOTEST_FORMATTER="/root/go/bin/go-junit-report -set-exit-code -iocopy -out test/runtime.xml"
             make tests-privileged NO_COLOR=1
 
+      - name: Debug failure on VM
+        if:  ${{ !success() && (matrix.focus == 'agent' || matrix.focus == 'datapath') }}
```
pchaigno, line 327: Could you explain the condition here?

.github/workflows/conformance-runtime.yaml
```diff
@@ -358,3 +367,9 @@ jobs:
           path: |
             cilium-junits/*.xml
           retention-days: 2
+
+      - name: Publish Test Results As GitHub Summary
+        if: ${{ always() }}
+        uses: aanm/junit2md@86789108ea0fd9be947a4232dd560e2fa3b9467a # v0.0.1
```
pchaigno, line 375: Should we move that to the Cilium organization?

pchaigno (approved):

.github/workflows/conformance-runtime.yaml
aanm: Sure, the `${{ matrix.focus == 'privileged' }}` is not running Cilium service, it's only running unit tests.

.github/workflows/conformance-runtime.yaml
aanm: Yes, once we address https://github.com/cilium/community/issues/27

.github/workflows/conformance-runtime.yaml
pchaigno: Ah, I see. Should the condition be inverted to make that clear?

aanm: I'm not following, you mean `if:  ${{ !success() && matrix.focus != 'privileged' }}` would make it more clear than `(matrix.focus == 'agent' || matrix.focus == 'datapath')`?

pchaigno: Yes, because it makes it clear this is specific to the privileged test. It's not so much that we only want to run this for agent and datapath, it's more that we don't want to run it for the privileged test. Said some other way, if we add a suite for hubble, we would likely want to run it there.

That said, nothing beats a small comment to clarify code.

aanm: > Yes, because it makes it clear this is specific to the privileged test. It's not so much that we only want to run this for agent and datapath, it's more that we don't want to run it for the privileged test. Said some other way, if we add a suite for hubble, we would likely want to run it there.

Sure, but the same argument can be used the other way around. What if we add a suite for hubble but we don't want it to run it there?

I've added the comment to make it clear.

PR #25798, opened 2023-05-31
operator: fix crash that occurs when cep CRD is disabled and CES enabled with kvstore

### LINKED ISSUE #24396: Cilium Operator crashes when CiliumEndpoint CRD is disabled

### What happened?

Operator crashes when endpoint CRD is disabled: `disable-endpoint-crd: "true"`

```
188 2023-03-15T16:37:15.327129738Z Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer der    eference)
189 2023-03-15T16:37:15.328109280Z goroutine 461 [running]:
190 2023-03-15T16:37:15.328116488Z k8s.io/apimachinery/pkg/util/runtime.logPanic({0x1f5f440?, 0x3e79d70})
191 2023-03-15T16:37:15.328118655Z  /go/src/github.com/cilium/cilium/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:75 +0x84
192 2023-03-15T16:37:15.328123571Z k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0x0?})
193 2023-03-15T16:37:15.328124738Z  /go/src/github.com/cilium/cilium/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:49 +0x80
194 2023-03-15T16:37:15.328126321Z panic({0x1f5f440, 0x3e79d70})
195 2023-03-15T16:37:15.328131071Z  /usr/local/go/src/runtime/panic.go:884 +0x20c
196 2023-03-15T16:37:15.328137988Z github.com/cilium/cilium/operator/watchers.endpointUpdated(0x4000886ea0)
197 2023-03-15T16:37:15.328139196Z  /go/src/github.com/cilium/cilium/operator/watchers/cilium_endpoint.go:216 +0x4c
```

### Cilium Version

v1.13.0

This doesn't appear to happen in master, it happens on both v1.13.0 as well as (current) v1.13.2 release branch.

### Kernel Version

5.15.0-60-generic

### Kubernetes Version

```
Server Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.3", GitCommit:"434bfd82814af038ad94d62ebe59b133fcb50506", GitTreeState:"clean", BuildDate:"2022-10-25T19:38:29Z", GoVersion:"go1.19.2", Compiler:"gc", Platform:"linux/arm64"}
```

### Sysdump

[operator-panic.zip](https://github.com/cilium/cilium/files/10984255/operator-panic.zip)

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

doniacld:
## Description
Make the operator crash early if CiliumEndpointSlice is enabled and CiliumEndpoint CRDs is disabled since it does not make sense to have CES without cep running. 

Add a check at helm installation level.

Fixes #24396 

### Test at runtime

Adding the following configuration in kind-values.yaml and deploy a kind cluster
```yaml
enableCiliumEndpointSlice: true
disableEndpointCRD: true
identityAllocationMode: "kvstore"
```
NB: Tested with and without `identityAllocationMode: "kvstore"`

Check the logs from the operator
```
$ ks get pod                                                                                    
NAME                                         READY   STATUS             RESTARTS     AGE
cilium-24plg                                 0/1     Init:3/6           0            3s
cilium-fblnk                                 0/1     Init:3/6           0            3s
cilium-operator-65fb77cf85-jv8xh             0/1     CrashLoopBackOff   1 (2s ago)   3s
```

```
$ ks logs cilium-operator-65fb77cf85-jv8xh -f                                                                  
level=debug msg="Skipped reading configuration file" error="Config File \"cilium-operators\" Not Found in \"[/]\"" subsys=config
level=fatal msg="Running Cilium with enable-cilium-endpoint-slice=true requires disable-endpoint-crd set to false to enable CRDs." subsys=config
```

### Test during deployment

The error is properly raised.
```
$ helm install  --namespace kube-system    -f ./contrib/testing/kind-values.yaml cilium ./install/kubernetes/cilium   
Error: INSTALLATION FAILED: execution error at (cilium/templates/validate.yaml:76:7): Cilium operator requires .Values.enableCiliumEndpointSlice=true and .Values.disableEndpointCRD=false
```

```release-note
Change enableEndpointCRD helm option type from string to boolean
Fix operator panic that occurs when Endpoint CRD is disabled and CiliumEndpointSlice is enabled
```

### DIFF
--- a/install/kubernetes/cilium/templates/cilium-configmap.yaml
+++ b/install/kubernetes/cilium/templates/cilium-configmap.yaml
@@ -131,9 +131,9 @@ data:
   skip-cnp-status-startup-clean: "{{ .Values.operator.skipCNPStatusStartupClean }}"
 {{- end }}
 
-{{- if hasKey .Values "disableEndpointCRD" }}
+{{- if eq .Values.disableEndpointCRD true }}
   # Disable the usage of CiliumEndpoint CRD
-  disable-endpoint-crd: "{{ .Values.disableEndpointCRD }}"
+  disable-endpoint-crd: "true"
 {{- end }}
 
 {{- if .Values.identityChangeGracePeriod }}
--- a/install/kubernetes/cilium/templates/validate.yaml
+++ b/install/kubernetes/cilium/templates/validate.yaml
@@ -69,3 +69,10 @@
     {{ fail "SPIRE installation requires .Values.auth.mTLS.enabled=true and .Values.auth.mTLS.spire.enabled=true" }}
   {{- end }}
 {{- end }}
+
+{{/* validate Cilium operator */}}
+{{- if eq .Values.enableCiliumEndpointSlice true }}
+  {{- if eq .Values.disableEndpointCRD true }}
+    {{ fail "if Cilium Endpoint Slice is enabled (.Values.enableCiliumEndpointSlice=true), it requires .Values.disableEndpointCRD=false" }}
+  {{- end }}
+{{- end }}
--- a/Documentation/helm-values.rst
+++ b/Documentation/helm-values.rst
@@ -690,8 +690,8 @@
      - ``nil``
    * - disableEndpointCRD
      - Disable the usage of CiliumEndpoint CRD.
-     - string
-     - ``"false"``
+     - bool
+     - ``false``
    * - dnsPolicy
      - DNS policy for Cilium agent pods. Ref: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy
      - string
--- a/Documentation/operations/upgrade.rst
+++ b/Documentation/operations/upgrade.rst
@@ -434,6 +434,9 @@ Helm Options
   will be removed in v1.15.
 * Values ``proxy.prometheus.enabled`` and ``proxy.prometheus.port`` are deprecated in favor of
   their ``envoy.prometheus.*`` counterparts.
+* Value ``disableEndpointCRD`` is now a boolean type instead of a string. Instead of using "true"
+  or "false" as values, you should remove the quotes. For example in helm command, instead of
+  ``--set-string disableEndpointCRD="true"``, it should be replaced by ``--set disableEndpointCRD=true``.
 
 .. _earlier_upgrade_notes:
 
--- a/install/kubernetes/cilium/README.md
+++ b/install/kubernetes/cilium/README.md
@@ -222,7 +222,7 @@ contributors across the globe, there is almost always someone available to help.
 | dashboards | object | `{"annotations":{},"enabled":false,"label":"grafana_dashboard","labelValue":"1","namespace":null}` | Grafana dashboards for cilium-agent grafana can import dashboards based on the label and value ref: https://github.com/grafana/helm-charts/tree/main/charts/grafana#sidecar-for-dashboards |
 | debug.enabled | bool | `false` | Enable debug logging |
 | debug.verbose | string | `nil` | Configure verbosity levels for debug logging This option is used to enable debug messages for operations related to such sub-system such as (e.g. kvstore, envoy, datapath or policy), and flow is for enabling debug messages emitted per request, message and connection.  Applicable values: - flow - kvstore - envoy - datapath - policy |
-| disableEndpointCRD | string | `"false"` | Disable the usage of CiliumEndpoint CRD. |
+| disableEndpointCRD | bool | `false` | Disable the usage of CiliumEndpoint CRD. |
 | dnsPolicy | string | `""` | DNS policy for Cilium agent pods. Ref: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy |
 | dnsProxy.dnsRejectResponseCode | string | `"refused"` | DNS response code for rejecting DNS requests, available options are '[nameError refused]'. |
 | dnsProxy.enableDnsCompression | bool | `true` | Allow the DNS proxy to compress responses to endpoints that are larger than 512 Bytes or the EDNS0 option, if present. |
--- a/install/kubernetes/cilium/templates/cilium-operator/clusterrole.yaml
+++ b/install/kubernetes/cilium/templates/cilium-operator/clusterrole.yaml
@@ -15,7 +15,7 @@ rules:
   - list
   - watch
 {{- if hasKey .Values "disableEndpointCRD" }}
-{{- if (eq (.Values.disableEndpointCRD | quote ) ( "false" | quote )) }}
+{{- if not .Values.disableEndpointCRD }}
 {{- if (and .Values.operator.unmanagedPodWatcher.restart (ne (.Values.operator.unmanagedPodWatcher.intervalSeconds | int64) 0 ) ) }}
   # to automatically delete [core|kube]dns pods so that are starting to being
   # managed by Cilium
--- a/install/kubernetes/cilium/values.yaml
+++ b/install/kubernetes/cilium/values.yaml
@@ -2039,7 +2039,7 @@ tunnelPort: 0
 MTU: 0
 
 # -- Disable the usage of CiliumEndpoint CRD.
-disableEndpointCRD: "false"
+disableEndpointCRD: false
 
 wellKnownIdentities:
   # -- Enable the use of well-known identities.
--- a/install/kubernetes/cilium/values.yaml.tmpl
+++ b/install/kubernetes/cilium/values.yaml.tmpl
@@ -2036,7 +2036,7 @@ tunnelPort: 0
 MTU: 0
 
 # -- Disable the usage of CiliumEndpoint CRD.
-disableEndpointCRD: "false"
+disableEndpointCRD: false
 
 wellKnownIdentities:
   # -- Enable the use of well-known identities.
--- a/operator/cmd/root.go
+++ b/operator/cmd/root.go
@@ -397,7 +397,8 @@ func (legacy *legacyOnLeader) onStart(_ hive.HookContext) error {
 	isLeader.Store(true)
 
 	// If CiliumEndpointSlice feature is enabled, create CESController, start CEP watcher and run controller.
-	if legacy.clientset.IsEnabled() && !option.Config.DisableCiliumEndpointCRD && option.Config.EnableCiliumEndpointSlice {
+	// Knowing CES are enabled only if CiliumEndpoint CRD are enabled too.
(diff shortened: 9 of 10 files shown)


mhofstetter (commented):
hey @doniacld 

thanks for the PR - looks good so far! only some proposals regarding the failing checks and missing labels from my side.

**Checks**

looks like the check `BPF checks / checkpatch` is failing due to the long commit message subject.

`Error: ERROR:CUSTOM: Please avoid long commit subjects (max: 75, found: 85)`

Not sure what's going on with the other failing `CIlium Runtime` checks that have been introduced lately. Might be worth to rebase to `main` to fetch the latest version while renaming the commit message.

**Labels**

you should also add a relevant release note label to this PR (https://docs.cilium.io/en/latest/contributing/development/contributing_guide/#submitting-a-pull-request - 11.) - IMO `release-note/misc` should be appropriate in this case (which doesn't require an explicit release note).

it's also helpful to add some labels to bring the PR into context - e.g. `area/operator`, `kind/bug`

doniacld:
Thanks @mhofstetter for your feedback! I added the labels you mentioned and reduced the char size of my commit subject.

mhofstetter (commented):
@doniacld thanks for adding the labels and fixing the commit subject.

Check in the Go-Code when starting the operator looks good to me.

But it might be worth to add this check to the Helm Chart too - to prevent an installation with these incompatible values in an earlier stage. This would also cover an installation via Cilium CLI (with helm mode).

The two config values are configured in the configmap ([disableEndpointCRD](https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/templates/cilium-configmap.yaml#L131-L133) & [enableCiliumEndpointSlice](https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/templates/cilium-configmap.yaml#L985-L987)). 

I'd recommend to place a check based on the two helm values `.Values.enableCiliumEndpointSlice` & `.Values.disableEndpointCRD` in the helm validation file (https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/templates/validate.yaml)

doniacld:
Thanks for your comment regarding the helm check, we had this conversation with @gandro who gave me some history regarding the `validate.yaml` strategy (@gandro being the first author of this idea). Our first thought was that we do not want really to maintain this file and keep it but since it is here, I do think it will not hurt anyone to have a supplementary check for the moment.

joamaki (approved):

install/kubernetes/cilium/templates/validate.yaml
```diff
@@ -69,3 +69,10 @@
     {{ fail "SPIRE installation requires .Values.auth.mTLS.enabled=true and .Values.auth.mTLS.spire.enabled=true" }}
   {{- end }}
 {{- end }}
+
+{{/* validate Cilium operator */}}
+{{- if .Values.enableCiliumEndpointSlice }}
+  {{- if .Values.disableEndpointCRD }}
+    {{ fail "if enableCiliumEndpointSlice is true, that then disableEndpointCRD needs to be false" }}
```
kaworu: nitpicking

```suggestion
    {{ fail "Cilium Endpoint Slice (.Values.enableCiliumEndpointSlice=true) requires .Values.disableEndpointCRD=false" }}
```

kaworu (approved):
Thanks @doniacld patch LGTM!

install/kubernetes/cilium/templates/validate.yaml
doniacld: Updated.

mhofstetter (approved):
@doniacld thanks for adding the helm validation!

looks like the docker build failed due to some infra issues (and therefore all dependent checks).

i'd recommend to rebase to `main` to re-trigger all basic checks. In addition you can trigger the e2e-tests by commenting `/test` (in a single comment)

@tommyp1ckles your review became mandatory (with sig-k8s). please chime in :)

doniacld:
I added a last commit to change `disableEndpointCRD` type from string to boolean. The default value was a string "false", with a boolean type we avoid transformation in helm from string to boolean. Note that in Go code, `disableCiliumEndpointCRD` is a boolean. 

Can you please review the last commit? @mhofstetter

qmonnet (approved):
Looks good, thank you!

mhofstetter:
> I added a new commit to change disableEndpointCRD type from string to boolean. The default value was a string "false", with a boolean type we avoid transformation in helm from string to boolean. Note that in Go code, disableCiliumEndpointCRD is a boolean.

@doniacld thanks for the heads-up.

in general i prefer having explicit types. it's also nice that helm automatically converts the strings to booleans.

it's just that setting `disableEndpointCRD="false"` (explicitly set as string in the customers `values.yaml` or via `--set-string` - which might be the case at customer side) will result in `disable-endpoint-crd: "true"` in the configmap :exclamation: this is kind of how the Go template `if` works.

```
helm template ./install/kubernetes/cilium --set-string disableEndpointCRD="false" | grep "disable-endpoint-crd"
  disable-endpoint-crd: "true"
```

i don't know whether we want that and whether the note in the upgrade guide is enough :man_shrugging: maybe a question for more seasoned teammembers? @qmonnet @kaworu

qmonnet:
> it's just that setting `disableEndpointCRD="false"` (explicitly set as string in the customers `values.yaml` or via `--set-string` - which might be the case at customer side (because it was a string until now)) will result in `disable-endpoint-crd: "true"` in the configmap

Right, I'm no Helm expert but this sounds like the ideal thing to introduce bugs, good catch. Is there a way to add some validation on the type (or value), so that users get an error if they pass `"false"`?

install/kubernetes/cilium/templates/cilium-configmap.yaml
```diff
@@ -131,9 +131,9 @@ data:
   skip-cnp-status-startup-clean: "{{ .Values.operator.skipCNPStatusStartupClean }}"
 {{- end }}
 
-{{- if hasKey .Values "disableEndpointCRD" }}
+{{- if .Values.disableEndpointCRD }}
```
mhofstetter: to enforce an explicit type check - we have to change this to `{{- if eq .Values.disableEndpointCRD true }}`

this will result in an error if trying to pass an incompatible type.

```
error calling eq: incompatible types for comparison
```

qmonnet:
I was thinking more of something along what we have in `install/kubernetes/cilium/templates/validate.yaml`. A simple check could be to error out is the value is neither `true` or `false` (or empty). Maybe a bit hacky, but could work?

qmonnet:
Ah, just seeing your suggestion now on `eq .Value... true`. This looks better.

install/kubernetes/cilium/templates/cilium-configmap.yaml
doniacld: Updated and tested.

### Configuration with the wrong type
```
enableCiliumEndpointSlice: false
disableEndpointCRD: "true"
```
Error is as expected
```
Error: template: cilium/templates/cilium-configmap.yaml:134:7: executing "cilium/templates/cilium-configmap.yaml" at <eq .Values.disableEndpointCRD true>: error calling eq: incompatible types for comparison
```

### Configuration with the wrong type BUT enableCiliumEndpointSlice: true
```
enableCiliumEndpointSlice: true
disableEndpointCRD: "true"
```
The error type is not thrown since I guess validate.yaml is the first test to be executed.
```
Error: execution error at (cilium/templates/validate.yaml:76:7): if Cilium Endpoint Slice is enabled (.Values.enableCiliumEndpointSlice=true), it requires .Values.disableEndpointCRD=false
```

mhofstetter: @doniacld great

we're missing one case:

```
enableCiliumEndpointSlice: true
disableEndpointCRD: "false"
```

the validation error gets thrown even though the endpointCRD isn't actually disabled 

```
❯ helm template ./install/kubernetes/cilium -f ./contrib/testing/kind-values.yaml
Error: execution error at (cilium/templates/validate.yaml:76:7): if Cilium Endpoint Slice is enabled (.Values.enableCiliumEndpointSlice=true), it requires .Values.disableEndpointCRD=false
```

-> we should change the validation to enforce type checks too

```
{{- if and (eq .Values.enableCiliumEndpointSlice true ) (eq .Values.disableEndpointCRD true ) }}
  {{ fail "if Cilium Endpoint Slice is enabled (.Values.enableCiliumEndpointSlice=true), it requires .Values.disableEndpointCRD=false" }}
{{- end }}
```

doniacld: Little update:

### Config enableCiliumEndpointSlice:false & disableEndpointCRD: "true"
```yaml
enableCiliumEndpointSlice: false
disableEndpointCRD: "true"
```

Error
```yaml
Error: template: cilium/templates/cilium-configmap.yaml:134:7: executing "cilium/templates/cilium-configmap.yaml" at <eq .Values.disableEndpointCRD true>: error calling eq: incompatible types for comparison
```

### Config enableCiliumEndpointSlice:true & disableEndpointCRD: "true"
```yaml
enableCiliumEndpointSlice: true
disableEndpointCRD: "true"
```
Error
```
Error: template: cilium/templates/validate.yaml:75:9: executing "cilium/templates/validate.yaml" at <eq .Values.disableEndpointCRD true>: error calling eq: incompatible types for comparison
```

### Helm command enableCiliumEndpointSlice=true & disableEndpointCRD="true"
```
 --set enableCiliumEndpointSlice=true --set-string disableEndpointCRD="true"
```
Error
```
Error: template: cilium/templates/validate.yaml:75:9: executing "cilium/templates/validate.yaml" at <eq .Values.disableEndpointCRD true>: error calling eq: incompatible types for comparison
```

Now we should be good!

mhofstetter: thanks @doniacld

mhofstetter (approved):

doniacld:
> i don't know whether we want that and whether the note in the upgrade guide is enough 🤷‍♂️ maybe a question for more seasoned cilium teammembers? @qmonnet @kaworu
I tried to be more explicit in the documentation, please let me know if that's okay.

tommyp1ckles (approved):

tommyp1ckles:
Nice work!

The helm validation makes me wonder if we should somehow unify checking/exiting config incompatibilities and doing helm based validation one day (failing to install is a lot better than crashing).

PR #25803, opened 2023-05-31
daemon: Reject BPF Host Routing without KPR=strict

pchaigno:
It's currently possible to enable BPF Host Routing with KPR=partial if masquerading is disabled. If masquerading is enabled, then we will require it to be BPF masquerading, which itself requires KPR. But if masquerading is disabled, then we currently don't have a check that prevents KPR=partial from being enabled at the same time as BPF Host Routing.

```release-note
Reject incorrect configuration enable-host-legacy-routing=false kube-proxy-replacement=partial.
```


julianwiedmann (approved):
Thanks!

The patch description didn't parse well for me. So for posterity:

If kube-proxy is needed for any kind of service handling (KPR != strict), we can't bypass netfilter. Thus we mustn't use BPF Host Routing in such a config. The existing config validations already cover *some* scenarios where KPR != strict, but it's not complete. So add an explicit validation.

PR #25805, opened 2023-05-31
Increase the max wait time for EndpointCreate API requests

### LINKED ISSUE #24361: CI: flakes due to putEndpointIdTooManyRequests
## CI failure

I saw a [ci failure](https://github.com/cilium/cilium/actions/runs/4406427789/jobs/7737432405) where some server pods failed to come up. The error message as returned by the CNI plugin was

```
Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "64d0ede7d7465a1f77351e8f1eb3b724ab1ea35488e2a02a53cbf4f7fbb50bdd": plugin type="cilium-cni" name="cilium" failed (add): Unable to create endpoint: [PUT /endpoint/{id}][429] putEndpointIdTooManyRequests
```

Now, this is interesting because the CNI ADD should have been retried, but given that this node didn't have very many pods running on it. Nevertheless, we should consider raising the rate limit, and maybe just add a backoff + retry internal to the client. That way we can avoid surfacing these sorts of errors to the end-user.

squeed:
In the event that a node has a burst of PodSanboxCreate requests, API requests will pile up. However, we should wait longer for the queue to clear before short-circuting and returing failure.

This is because the kubelet has a relatively relaxed timeout for PodSandbox creation -- 4 minutes. Furthermore, if we return a failure here, it is propagated all the way back through containerd to kubelet, which will tear down the entire PodSandbox and try again, which can be expensive.

So, increase the maximum queue duration time to 1 minute. That should hopefully give enough time for the queue to clear.

Fixes: #24361

```release-note
Cilium now waits longer before returning a failure in the event of a pod creation burst.
```

### DIFF
--- a/daemon/cmd/api_limits.go
+++ b/daemon/cmd/api_limits.go
@@ -25,8 +25,9 @@ var apiRateLimitDefaults = map[string]rate.APILimiterParameters{
 		RateLimit:                   0.5,
 		RateBurst:                   4,
 		ParallelRequests:            4,
+		MinParallelRequests:         2,
 		SkipInitial:                 4,
-		MaxWaitDuration:             15 * time.Second,
+		MaxWaitDuration:             60 * time.Second, // Kubelet has a PodSandbox creation timeout of 4 minutes, in total.
 		Log:                         false,
 	},
 	// DELETE /endpoint/{id}


thorn3r (approved):
nice investigation on this one :+1:

squeed:
With regards to feature freeze, I argue this is fixing a known bug (and it tweaks a few configs, rather than introducing any sort of new behavior).

PR #25807, opened 2023-05-31
renovate: do not update 'github.com/mdlayher/arp'

aanm:
We are still using an old version of this dependency in github.com/cilium/metallb so we should avoid updating it in the meanwhile.

It will fix PR https://github.com/cilium/cilium/pull/25539

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -125,6 +125,8 @@
         // want to update them automatically.
         "go.universe.tf/metallb",
         "github.com/cilium/metallb",
+        // metallb is still using an old version of "github.com/mdlayher/arp"
+        "github.com/mdlayher/arp",
         "github.com/miekg/dns",
         "github.com/cilium/dns",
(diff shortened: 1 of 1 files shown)


michi-covalent (approved):
magical

.github/renovate.json5
```diff
@@ -125,6 +125,8 @@
         // want to update them automatically.
         "go.universe.tf/metallb",
         "github.com/cilium/metallb",
+        // metallb is still using an old version of "github.com/mdlayher/arp"
+        "github.com/mdlayher/arp",
```
bimmlerd, line 129: Do we also track the work to get back to being able to update this?

bimmlerd (approved):

.github/renovate.json5
aanm: Good point, created https://github.com/cilium/cilium/issues/25814

PR #25810, opened 2023-05-31
Makefile: use CLI options to set local images for kind-install-cilium-clustermesh

thorn3r:
Makefile: use CLI options to set local images for kind-install-cilium-clustermesh

`make-kind-clustermesh-images` uses `KIND_ENV` to set image repositories for local images to be pushed to kind clusters, however these are not actually used when installing Cilium on the kind clusters in `kind-install-cilium-clustermesh`.

The hardcoded image overrides have been removed from the helm values for these clusters and replaced with CLI flags to specify using the local images for Agent and Operator.

Fixes: #issue-number

```release-note
Makefile: use CLI options to set local images for kind-install-cilium-clustermesh
```


bimmlerd (approved):
noice.

borkmann (approved):

PR #25817, opened 2023-06-01
etcd: extend rate limiting to consider the number of inflight requests

giorio94:
Reporting the description of the last commit for convenience:

```
Currently, the etcd rate limiter operates based on a token bucket of a
fixed size. Yet, this is problematic when etcd is overloaded, since we
may continue to issue new requests even if it cannot keep up. Hence,
let's start using the custom APILimiter, which also takes into account
the number of currently inflight requests. The maximum amount of
inflight requests can configured through the etcd.maxInflight parameter,
and defaults to the same value as etcd.qps if unset.
```

```release-note
etcd: extend rate limiting to consider the number of inflight requests
```

### DIFF
--- a/pkg/kvstore/etcd.go
+++ b/pkg/kvstore/etcd.go
@@ -837,16 +837,17 @@ func (e *etcdClient) LockPath(ctx context.Context, path string) (KVLocker, error
 }
 
 func (e *etcdClient) DeletePrefix(ctx context.Context, path string) (err error) {
-	defer func() { Trace("DeletePrefix", err, logrus.Fields{fieldPrefix: path}) }()
+	defer func(duration *spanstat.SpanStat) {
+		Trace("DeletePrefix", err, logrus.Fields{fieldPrefix: path})
+		increaseMetric(path, metricDelete, "DeletePrefix", duration.EndError(err).Total(), err)
+	}(spanstat.Start())
 
-	duration := spanstat.Start()
 	e.limiter.Wait(ctx)
 	_, err = e.client.Delete(ctx, path, client.WithPrefix())
 	if err == nil {
 		e.leaseManager.ReleasePrefix(path)
 	}
 
-	increaseMetric(path, metricDelete, "DeletePrefix", duration.EndError(err).Total(), err)
 	return Hint(err)
 }
 
@@ -1167,9 +1168,11 @@ func (e *etcdClient) Status() (string, error) {
 
 // GetIfLocked returns value of key if the client is still holding the given lock.
 func (e *etcdClient) GetIfLocked(ctx context.Context, key string, lock KVLocker) (bv []byte, err error) {
-	defer func() { Trace("GetIfLocked", err, logrus.Fields{fieldKey: key, fieldValue: string(bv)}) }()
+	defer func(duration *spanstat.SpanStat) {
+		Trace("GetIfLocked", err, logrus.Fields{fieldKey: key, fieldValue: string(bv)})
+		increaseMetric(key, metricRead, "GetLocked", duration.EndError(err).Total(), err)
+	}(spanstat.Start())
 
-	duration := spanstat.Start()
 	e.limiter.Wait(ctx)
 	opGet := client.OpGet(key)
 	cmp := lock.Comparator().(client.Cmp)
@@ -1177,7 +1180,7 @@ func (e *etcdClient) GetIfLocked(ctx context.Context, key string, lock KVLocker)
 	if err == nil && !txnReply.Succeeded {
 		err = ErrLockLeaseExpired
 	}
-	increaseMetric(key, metricRead, "GetLocked", duration.EndError(err).Total(), err)
+
 	if err != nil {
 		return nil, Hint(err)
 	}
@@ -1192,12 +1195,13 @@ func (e *etcdClient) GetIfLocked(ctx context.Context, key string, lock KVLocker)
 
 // Get returns value of key
 func (e *etcdClient) Get(ctx context.Context, key string) (bv []byte, err error) {
-	defer func() { Trace("Get", err, logrus.Fields{fieldKey: key, fieldValue: string(bv)}) }()
+	defer func(duration *spanstat.SpanStat) {
+		Trace("Get", err, logrus.Fields{fieldKey: key, fieldValue: string(bv)})
+		increaseMetric(key, metricRead, "Get", duration.EndError(err).Total(), err)
+	}(spanstat.Start())
 
-	duration := spanstat.Start()
 	e.limiter.Wait(ctx)
 	getR, err := e.client.Get(ctx, key)
-	increaseMetric(key, metricRead, "Get", duration.EndError(err).Total(), err)
 	if err != nil {
 		return nil, Hint(err)
 	}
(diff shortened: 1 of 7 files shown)


marseel:
Related Etcd issue: https://github.com/etcd-io/etcd/issues/15993

pkg/kvstore/etcd.go
```diff
@@ -1055,9 +1094,22 @@ reList:
...
-		e.limiter.Wait(ctx)
+		rl, err := e.limiter.Wait(ctx)
+		if err != nil {
+			select {
+			case <-e.client.Ctx().Done():
+				return
+			case <-ctx.Done():
+				return
+			default:
+				goto recreateWatcher
```
marseel, line 1008: We are missing `rl.Done()` here, right?

giorio94: We don't need to (and cannot) call `rl.Done()` when `Wait()` returns an error, since `rl` is nil in that case.

marseel: ah yea, I don't know why I thought error was coming from different place...

pkg/kvstore/etcd.go
```diff
@@ -1162,11 +1214,16 @@ func (e *etcdClient) determineEndpointStatus(ctx context.Context, endpointAddres
 
 	e.getLogger().Debugf("Checking status to etcd endpoint %s", endpointAddress)
 
-	e.limiter.Wait(ctxTimeout)
+	rl, err := e.limiter.Wait(ctxTimeout)
```
oblazek: just a small nit: is there a reason you are mixing lr/rl ? e.g. `rl, err := e.limiter.Wait(ctxTimeout)` and `lr, err := e.limiter.Wait(ctx)`

giorio94: Thanks for noticing, they should be consistent now.

YutaroHayakawa (approved):

pkg/kvstore/etcd.go
```diff
@@ -1162,7 +1214,6 @@ func (e *etcdClient) determineEndpointStatus(ctx context.Context, endpointAddres
 
 	e.getLogger().Debugf("Checking status to etcd endpoint %s", endpointAddress)
 
-	e.limiter.Wait(ctxTimeout)
```
giorio94, line 1066: I've dropped the rate limiter wait here, to prevent that the status check fails due to rate limiting, as that would put a high pressure on etcd in case the connection gets restarted.

pkg/kvstore/etcd.go
```diff
@@ -1002,8 +1037,13 @@ reList:
 		default:
 		}
 
-		e.limiter.Wait(ctx)
+		lr, err := e.limiter.Wait(ctx)
+		if err != nil {
+			continue
+		}
 		kvs, revision, err := e.paginatedList(ctx, scopedLog, w.Prefix)
+		// Using lr.Error for convenience, as it matches lr.Done() when err is nil
```
marseel: err is always nil here though, right?

pkg/kvstore/etcd.go
```diff
@@ -1270,7 +1322,12 @@ func (e *etcdClient) GetIfLocked(ctx context.Context, key string, lock KVLocker)
 	defer func() { Trace("GetIfLocked", err, logrus.Fields{fieldKey: key, fieldValue: string(bv)}) }()
 
 	duration := spanstat.Start()
-	e.limiter.Wait(ctx)
+	lr, err := e.limiter.Wait(ctx)
+	if err != nil {
+		increaseMetric(key, metricRead, "GetLocked", duration.EndError(err).Total(), err)
```
marseel: Can we do `defer increaseMetric(key, metricRead, "GetLocked", duration.EndError(err).Total(), err)` two lines above (not inside if) and remove another increaseMetric below?

pkg/kvstore/etcd.go
```diff
@@ -1295,12 +1354,19 @@ func (e *etcdClient) Get(ctx context.Context, key string) (bv []byte, err error)
 	defer func() { Trace("Get", err, logrus.Fields{fieldKey: key, fieldValue: string(bv)}) }()
 
 	duration := spanstat.Start()
-	e.limiter.Wait(ctx)
+	lr, err := e.limiter.Wait(ctx)
+	if err != nil {
+		increaseMetric(key, metricRead, "Get", duration.EndError(err).Total(), err)
```
marseel: Similar defer here.

pkg/kvstore/etcd.go
```diff
@@ -1315,7 +1381,12 @@ func (e *etcdClient) GetPrefixIfLocked(ctx context.Context, prefix string, lock
 	}()
 
 	duration := spanstat.Start()
-	e.limiter.Wait(ctx)
+	lr, err := e.limiter.Wait(ctx)
+	if err != nil {
+		increaseMetric(prefix, metricRead, "GetPrefixLocked", duration.EndError(err).Total(), err)
```
marseel: Similar defer here.

pkg/kvstore/etcd.go
```diff
@@ -1341,12 +1415,19 @@ func (e *etcdClient) GetPrefix(ctx context.Context, prefix string) (k string, bv
 	}()
 
 	duration := spanstat.Start()
-	e.limiter.Wait(ctx)
+	lr, err := e.limiter.Wait(ctx)
+	if err != nil {
+		increaseMetric(prefix, metricRead, "GetPrefix", duration.EndError(err).Total(), err)
```
marseel: Similar defer here.

pkg/kvstore/etcd.go
```diff
@@ -1359,7 +1440,14 @@ func (e *etcdClient) Set(ctx context.Context, key string, value []byte) (err err
 	defer func() { Trace("Set", err, logrus.Fields{fieldKey: key, fieldValue: string(value)}) }()
 
 	duration := spanstat.Start()
-	e.limiter.Wait(ctx)
+	lr, err := e.limiter.Wait(ctx)
+	if err != nil {
+		increaseMetric(key, metricSet, "Set", duration.EndError(err).Total(), err)
```
marseel: Similar defer here and bunch of below :)

marseel (approved):
lgtm except for small nits.

pkg/kvstore/etcd.go
giorio94: We cannot with the current `increaseMetric` implementation, since `duration.EndError(err).Total()` would be executed immediately, hence not counting correctly the duration. It should work modifying `increaseMetric` to take the span as input (rather than the pre-computed duration), but the change tends to be on the large side. WDYT?

pkg/kvstore/etcd.go
giorio94: Not necessarily, `e.paginatedList` might return an error if one of the Get operations failed.

pkg/kvstore/etcd.go
marseel: ah yea, we could do that in separate PR.

giorio94: I'm giving it a try. If it ends up being too large I'll open a separate PR

chancez (approved):

pkg/kvstore/etcd.go
giorio94: I've updated the PR adding one more commit which does the `increaseMetrics` deferring, to avoid having to repeat that in case of errors. @marseel PTAL

tklauser (approved):

michi-covalent (approved):

Issue #24764, opened 2023-04-05
CFP: Pools for PodCIDR IPAM


tklauser:
## Cilium Feature Proposal

Allow Cilium to allocate IP addresses to Kubernetes workloads from different pools depending on properties of the workload as defined by the user. An example where this will be useful is multi-homing (see #20129):
* Multi-homed pods for workloads that require additional network interfaces which are to be kept separate from the pod network (e.g. database replication traffic, legacy workloads expecting multiple interfaces)  
* Multi-homed pods might want IP addresses from a pool different than the standard podCIDR
* Not all pods are multi-homed and not all nodes will run multi-homed pods. We want to avoid wasting IPs by not assigning a pool to a node where no multi-homed pods are expected to be run

See https://docs.google.com/document/d/1FIyfhVegljG9anXpyA25gEYzuUGpfHo_MycZ-Wk0VeE for further details.

PR #25824, opened 2023-06-01
Support defining IPAM pools using CiliumPodIPPool CRD

tklauser:
This PR introduces the `CiliumPodIPPool` CRD and adds support to use it to define IP pools in Cilium's multi-pool IPAM mode (ref. #22762).

An example `CiliumPodIPPool` CRD definition of the pool named `jupiter` with the IPv4 CIDRs `10.10.0.0/16` and `10.20.0.0/16`, an IPv4 mask size of `27` bits, the IPv6 CIDR `fd00:100::/80` and an IPv6 mask size of `96` bits looks as follows:

```yaml
kind: CiliumPodIPPool
metadata:
  name: jupiter
spec:
  ipv4:
    cidrs:
      - 10.10.0.0/16
      - 10.20.0.0/16
    maskSize: 27
  ipv6:
    cidrs:
      - fd00:100::/80
    maskSize: 96
```

These CRDs are watched by cilium-operator. Deletion and update of existing pools are supported, however updates changing mask sizes on existing pools are rejected.

Please see individual commits for details.

Manually tested on kind as follows:

```console
$ WORKERS=2 make kind
$ make kind-image
$ helm upgrade --install cilium ./install/kubernetes/cilium --namespace kube-system \
    --set debug.enabled=true \
    --set image.override=localhost:5000/cilium/cilium-dev:local \
    --set image.pullPolicy=IfNotPresent \
    --set operator.image.override=localhost:5000/cilium/operator-generic:local \
    --set operator.image.pullPolicy=IfNotPresent \
    --set operator.replicas=1 \
    --set hubble.relay.enabled=true \
    --set ipam.mode=multi-pool \
    --set "ipam.operator.multiPoolMap.default.ipv4CIDRs[0]=10.10.0.0/16" \
    --set "ipam.operator.multiPoolMap.default.ipv4MaskSize=27" \
    --set tunnel=disabled \
    --set autoDirectNodeRoutes=true \
    --set ipv4NativeRoutingCIDR=10.0.0.0/8 \
    --set endpointRoutes.enabled=true \
    --set-string extraConfig.enable-local-node-route=false \
    --set bpf.masquerade=true
$ cat <<EOF | k apply -f -
kind: CiliumPodIPPool
metadata:
  name: mars
spec:
  ipv4:
    cidrs:
      - 10.20.0.0/16
    maskSize: 24
EOF
$ k get ciliumpodippools
NAME      AGE
mars      10s
$ # note that the default pool was configured using the operator flag and doesn't show up here (yet). This will be addressed in a follow-up PR.
$ k create -f https://raw.githubusercontent.com/cilium/cilium/1.13.3/examples/minikube/http-sw-app.yaml
$ k -n default get pods -o wide
NAME                         READY   STATUS    RESTARTS   AGE   IP           NODE           NOMINATED NODE   READINESS GATES
deathstar-8464cdd4d9-9bc24   1/1     Running   0          12s   10.10.0.42   kind-worker2   <none>           <none>
deathstar-8464cdd4d9-s5nfm   1/1     Running   0          12s   10.10.0.29   kind-worker    <none>           <none>
tiefighter                   1/1     Running   0          12s   10.10.0.20   kind-worker    <none>           <none>
xwing                        1/1     Running   0          12s   10.10.0.3    kind-worker    <none>           <none>
$ # note that all pods have IPs from the default pool's CIDRs
$ k create ns cilium-test
$ k annotate ns cilium-test ipam.cilium.io/ip-pool=mars
$ cilium connectivity test
...
✅ All 42 tests (295 actions) successful, 13 tests skipped, 0 scenarios skipped.
$ k -n cilium-test get pods -o wide
NAME                                  READY   STATUS    RESTARTS   AGE     IP            NODE                 NOMINATED NODE   READINESS GATES
client-6f6788d7cc-7fw9w               1/1     Running   0          8m56s   10.20.0.238   kind-worker          <none>           <none>
client2-bc59f56d5-hsv2g               1/1     Running   0          8m56s   10.20.0.193   kind-worker          <none>           <none>
echo-external-node-787c859b66-zrblr   0/1     Pending   0          8m55s   <none>        <none>               <none>           <none>
echo-other-node-646976b7dd-5zlr4      2/2     Running   0          8m56s   10.20.1.145   kind-worker2         <none>           <none>
echo-same-node-58f99d79f4-4k5v4       2/2     Running   0          8m56s   10.20.0.202   kind-worker          <none>           <none>
...
$ # note the different IP range (from pool mars) for the cilium-test pods
```

The following will be added in a separate follow-up PR, see meta issue #25470:

- Creation of `CiliumPodIPPool`s from operator `ipam.operator.multiPoolMap` helm flags
- CI coverage
- Documentation

Extends #24764


pkg/k8s/watchers/watcher.go
```diff
@@ -424,6 +424,7 @@ var ciliumResourceToGroupMapping = map[string]watcherInfo{
 	synced.CRDResourceName(v2alpha1.LBIPPoolName): {skip, ""}, // Handled in LB IPAM
 	synced.CRDResourceName(v2alpha1.CNCName):      {skip, ""}, // Handled by init directly
 	synced.CRDResourceName(v2alpha1.CCGName):      {start, k8sAPIGroupCiliumCIDRGroupV2Alpha1},
+	synced.CRDResourceName(v2alpha1.CIPName):      {skip, ""}, // Handled by cluster-pool-v2beta2 IPAM allocator
```
gandro: nit: Old IPAM mode name used here

pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumippools.yaml
```diff
@@ -0,0 +1,98 @@
...
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.7.0
+  creationTimestamp: null
+  name: ciliumippools.cilium.io
+spec:
+  group: cilium.io
+  names:
+    categories:
+    - cilium
+    kind: CiliumIPPool
```
gandro: I do wonder if we should call it `CiliumPodIPPool`, to clearly differentiate it from the `CiliumLoadBalancerIPPool`, though `cip` is a nicer short name than `cpip` :sweat_smile:

pkg/ipam/allocator/multipool/allocator.go
```diff
@@ -104,3 +105,38 @@ func (a *Allocator) Init(ctx context.Context) (err error) {
...
+	ipv4CIDRs := make([]string, len(pool.Spec.IPv4.CIDRs))
+	for i, cidr := range pool.Spec.IPv4.CIDRs {
+		ipv4CIDRs[i] = string(cidr)
+	}
+	ipv6CIDRs := make([]string, len(pool.Spec.IPv6.CIDRs))
+	for i, cidr := range pool.Spec.IPv6.CIDRs {
+		ipv6CIDRs[i] = string(cidr)
+	}
+
+	return a.poolAlloc.UpsertPool(
```
gandro, line 127: I think we need to take a mutex here, since `poolAlloc` is not thread-safe.

At the moment, `NewNodeHandler` takes ownership of `a.poolAlloc`. So from that perspective, it would make sense if we exposed `UpsertPool` on the NodeHandler itself. Unfortunately, there seems to be a chicken and egg problem where we want to be able to call Upsert/DeletePool before the allocator is started.

Not sure how to best address that. Maybe we need to introduce a mutex to `PoolAllocator` itself?

pkg/ipam/allocator/multipool/pool_allocator.go
```diff
@@ -101,28 +102,170 @@ func (p *PoolAllocator) RestoreFinished() {
...
+			v4 = slices.Delete(v4, i, i+1)
+		}
+	}
 	if err != nil {
 		return err
 	}
+	v4 = append(v4, newv4...)
 
-	v6, err := cidralloc.NewCIDRSets(true, ipv6CIDRs, ipv6MaskSize)
+	// allocate new CIDR set for each IPv6 CIDR not yet in the pool
```
gandro: nit: How hard would it be to move the "detect new CIDRs and add them and remove old ones" logic into a separate function that we invoke once for IPv4 once for IPv6?

pkg/ipam/allocator/multipool/pool_allocator.go
```diff
@@ -101,28 +102,170 @@ func (p *PoolAllocator) RestoreFinished() {
...
+		return fmt.Errorf("pool %q requested for deletion doesn't exist", poolName)
+	}
+
+	for node, pools := range p.nodes {
+		for pool := range pools {
+			if pool == poolName {
+				log.WithFields(logrus.Fields{
+					"pool": poolName,
+					"node": node,
+				}).Debug("pool still in use by node")
```
gandro: I think we should make this a warning.

pkg/ipam/allocator/multipool/pool_allocator.go
```diff
@@ -101,28 +102,170 @@ func (p *PoolAllocator) RestoreFinished() {
...
+				err = errors.Join(err, fmt.Errorf("invalid IPv4 CIDR: %w", parseErr))
+			}
+			if oldv4cidr.IsClusterCIDR(cidr) {
+				exists = true
+				break
+			}
+		}
+		if !exists {
+			v4[i] = nil
+			v4 = slices.Delete(v4, i, i+1)
```
gandro: This doesn't check `p.nodes` if the deleted CIDR is in use. I think we should try to do that same as in `DeletePool`.

gandro (commented):
Awesome work! Left a bit of early feedback

pkg/k8s/watchers/watcher.go
tklauser: There are even a few more :sweat_smile:

```
% git grep cluster-pool-v2beta2
install/kubernetes/cilium/Chart.yaml:138:        CiliumIPPool defines an IP pool that can be used for pooled IPAM (i.e. the cluster-pool-v2beta2 IPAM mode).
pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumippools.yaml:27:          (i.e. the cluster-pool-v2beta2 IPAM mode).
pkg/k8s/apis/cilium.io/v2alpha1/ippool_types.go:15:// CiliumIPPool defines an IP pool that can be used for pooled IPAM (i.e. the cluster-pool-v2beta2
```

I'm going to change these too.

pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumippools.yaml
tklauser: Yeah, good point. The short name is definitely nicer but I agree that `CiliumPodIPPool` is more precise and clearly differentiates from `CiliumLoadBalancerIPPool`. Let's change it.

pkg/ipam/allocator/multipool/allocator.go
```diff
@@ -86,11 +87,11 @@ func (a *Allocator) Init(ctx context.Context) (err error) {
...
-			"ipv4-masksize": pool.ipv4MaskSize,
-			"ipv6-cidrs":    pool.ipv6CIDRs,
-			"ipv6-masksize": pool.ipv6MaskSize,
+			"pool-name":      poolName,
+			"ipv4-cidrs":     pool.ipv4CIDRs,
+			"ipv4-mask-size": pool.ipv4MaskSize,
+			"ipv6-cidrs":     pool.ipv6CIDRs,
+			"ipv6-mask-size": pool.ipv6MaskSize,
 		}).Debug("adding pool")
 		if addErr := a.poolAlloc.AddPool(poolName, pool.ipv4CIDRs, pool.ipv4MaskSize, pool.ipv6CIDRs, pool.ipv6MaskSize); addErr != nil {
```
gandro, line 96: Should we actually remove the `operatorOption.Config.IPAMMultiPoolMap` as part of this PR too?

pkg/ipam/allocator/multipool/allocator.go
tklauser: > Not sure how to best address that. Maybe we need to introduce a mutex to `PoolAllocator` itself?

Yes, looks like the cleanest solution currently is to introduce a mutex to `PoolAllocator`. For now I've added them to `PoolAllocator`'s exported API at the top level, i.e. it protects all of its fields (`pools`, `nodes`, `ready`). I'd assume this is fine for now since these operations aren't expected to be called frequently. If needed, we could refine the scope of the mutex as a follow-up.

pkg/ipam/allocator/multipool/allocator.go
tklauser: Good point. I've added a separate commit removing the option and related parsing code.

pkg/ipam/allocator/multipool/pool_allocator.go
tklauser: Not too hard. Refactored into `(*PoolAllocator).updateCIDRSets`, thanks for the suggestion!

pkg/ipam/allocator/multipool/pool_allocator.go
tklauser: Folded into `(*PoolAllocator).updateCIDRSets`, thanks.

pkg/ipam/allocator/multipool/pool_allocator.go
```diff
@@ -105,20 +106,127 @@ func (p *PoolAllocator) RestoreFinished() {
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// delete CIDR set for CIDRs not present in the new CIDRs
+	for i, oldCIDR := range cidrSets {
+		exists := false
+		for _, cidrString := range newCIDRs {
+			_, cidr, parseErr := net.ParseCIDR(cidrString)
```
gandro, line 170: nit: It seems we're parsing the CIDRs in `newCIDRs` twice. Once here, and once for producing `alloc` above. Would it make sense to just introduce a new slice containing all the parsed CIDRs?

Maybe even in the call site.

pkg/ipam/allocator/multipool/pool_allocator.go
```diff
@@ -105,20 +106,127 @@ func (p *PoolAllocator) RestoreFinished() {
...
+		if !exists {
+			cidrSets[i] = nil
+			cidrSets = slices.Delete(cidrSets, i, i+1)
+
+			prefix := oldCIDR.Prefix()
+
+			for node, pools := range p.nodes {
+				for pool, allocatedCIDRSets := range pools {
+					if isV6 {
+						if _, ok := allocatedCIDRSets.v6[prefix]; ok {
```
gandro, line 189: If we delete prefixes from nodes, then we assume that they are still in use. We should emit a warning here

gandro (commented):
Awesome! One missing warning and a relatively minor nit.

pkg/ipam/allocator/multipool/pool_allocator.go
tklauser: Good point. Unfortunately, we'll need it once as a `netip.Prefix` (to call `hasCIDR`) and once as `net.IPNet` (to call `oldCIDR.IsClusterCIDR`). I think the entire code could benefit from a `net.IP{,Net}` -> `netip.{Addr,Prefix}` refactoring anyway, so I'd address this as part of that in a follow-up PR.

operator/watchers/cilium_podippool.go
```diff
@@ -0,0 +1,58 @@
...
+
+// PooledAllocatorProvider defines the functions of IPAM provider front-end which additionally allow
+// definition of IP pools at runtime.
+// This is implemented by e.g. pkg/ipam/allocator/multipool
+type PooledAllocatorProvider interface {
+	UpsertPool(ctx context.Context, pool *cilium_v2alpha1.CiliumPodIPPool) error
+	DeletePool(ctx context.Context, pool *cilium_v2alpha1.CiliumPodIPPool) error
+}
+
+func StartIPPoolAllocator(
```
joamaki, line 22: How hard would it be to implement this as a hive module instead? I suppose we'd need to pull out `PooledAllocatorProvider` first from the legacy code plus make `alloc.Start` depend on `IPPoolAllocator` to make sure it's started first?

joamaki (approved):
LGTM. A nice to have would be to start modularizing operator's IPAM related code as we're touching these parts.

operator/watchers/cilium_podippool.go
tklauser: I initially thought about implementing this as a hive cell. Unfortunately, it would require lifting quite a bit of IPAM code around. For example, I think we cannot make `alloc.Start` depend on `IPPoolAllocator` unconditionally, since only once specific implementation (namely `ipam/allocator/multipool.Allocator`) requires the pooled allocator, while all the others don't. Initially, I wanted to include `IPPoolAllocator` into `ipam/allocator.AllocatorProvider`, but IIRC that lead to nasty cyclic dependencies. Also, given this PR is already quite large and should make it into the tree for 1.14, I didn't want to include additional refactoring.

Admittedly it's not ideal to introduce new code in a non-modular fashion, but based on the above I'd defer addressing modularization of this code together with the other IPAM bits to a follow-up PR. I'll discuss the specifics with @gandro and @cilium/sig-foundations in a Slack thread or GH issue. Hope that's OK?

gandro: Yeah, given the feature freeze coming up in a week, I don't think it's realistic to tackle a larger refactor at the moment unfortunately :crying_cat_face: 

The upside is that this is very little and simple code, so it should not make any future refactor more complicated

gandro (approved):
Awesome work!

gandro:
Merge button is green! :ship:

PR #25826, opened 2023-06-01
bpf: nodeport: reduce CT lookup scope

julianwiedmann:
The standard behaviour for a CT lookup is that we first check whether the packet is a **reply** for some tracked connection. If that doesn't return a match, we do an additional lookup for the packet's actual direction (which might either find a match, or return `CT_NEW`). The exception is when we look for entries of type `CT_SERVICE` - here we already avoid the second lookup.

But we have some additional code paths that would benefit from excluding the second lookup
- nodeport RevDNAT only wants to handle a `CT_REPLY` lookup result (as it only  wants to reverse-translate the replies for nodeport connections). This PR converts those users.
- the inbound nodeport code only cares about the forward direction, but currently needs to tolerate results from the reverse direction (treating them as stale CT entries that need to be re-created). All of this can go away if we switch to `SCOPE_FORWARD`.
- the SNAT and RevSNAT paths should only care about their respective direction (I'll convert those in a follow-on, they need a bit more cleanup before).
 
Make the needed changes so that we can avoid the second lookup. This reduces code size / complexity, and offers more robust behaviour (as we no longer have to consider "impossible" CT results).

### DIFF
--- a/bpf/lib/conntrack.h
+++ b/bpf/lib/conntrack.h
@@ -23,6 +23,12 @@ enum {
 	ACTION_CLOSE,
 };
 
+enum ct_scope {
+	SCOPE_FORWARD,
+	SCOPE_REVERSE,
+	SCOPE_BIDIR,
+};
+
 #ifdef ENABLE_IPV4
 struct ct_buffer4 {
 	struct ipv4_ct_tuple tuple;
@@ -365,8 +371,8 @@ ipv6_ct_tuple_swap_ports(struct ipv6_ct_tuple *tuple)
 
 static __always_inline int
 __ct_lookup6(const void *map, struct ipv6_ct_tuple *tuple, struct __ctx_buff *ctx,
-	     int l4_off, int action, enum ct_dir dir, struct ct_state *ct_state,
-	     __u32 *monitor)
+	     int l4_off, int action, enum ct_dir dir, enum ct_scope scope,
+	     struct ct_state *ct_state, __u32 *monitor)
 {
 	bool is_tcp = tuple->nexthdr == IPPROTO_TCP;
 	union tcp_flags tcp_flags = { .value = 0 };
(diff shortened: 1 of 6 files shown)


julianwiedmann:
@gentoo-root ha, I was hoping it would hit you :smile: . No particular rush, let's give this a good thought.

But the complexity results look pretty nice....

julianwiedmann:
(back to draft for a bit, we can do even better ...)

bpf/lib/conntrack.h
```diff
@@ -380,32 +386,33 @@ __ct_lookup6(const void *map, struct ipv6_ct_tuple *tuple, struct __ctx_buff *ct
...
-				ret = CT_REPLY;
+
+	switch (scope) {
+	case SCOPE_REVERSE:
+	case SCOPE_BIDIR:
+		/* Lookup in the reverse direction first: */
+		ret = __ct_lookup(map, ctx, tuple, action, dir, ct_state, is_tcp,
+				  tcp_flags, monitor);
+		if (ret != CT_NEW || scope == SCOPE_REVERSE) {
+			if (likely(ret == CT_ESTABLISHED || ret == CT_REOPENED)) {
```
gentoo-root: These two lines look a bit confusing now.

The behavior stays the same, because if ret == CT_NEW, but scope == SCOPE_REVERSE, the `if (likely(...` condition will be false, and we will just bail out, just like before we did for CT_SERVICE. However, it's sort of counter-intuitive that we suppose `scope == SCOPE_REVERSE` is a sufficient condition to consider checking `if (likely(ret == ...`, and it's also sort of wrong that the `likely` clause is actually impossible in this case.

Could you split the first check and just do `if (scope == SCOPE_REVERSE) goto out;`?

bpf/lib/conntrack.h
```diff
@@ -380,32 +386,33 @@ __ct_lookup6(const void *map, struct ipv6_ct_tuple *tuple, struct __ctx_buff *ct
...
+			goto out;
 		}
-		goto out;
-	}
 
-	/* Lookup entry in forward direction */
-	if (dir != CT_SERVICE) {
+		/* now lookup in forward direction: */
+	case SCOPE_FORWARD:
 		ipv6_ct_tuple_reverse(tuple);
```
gentoo-root, line 415: A further improvement could be to make the callers pass the right tuple from the beginning, without the need to reverse it here if scope is SCOPE_FORWARD.

gentoo-root (commented):
Really nice cleanup! Conntrack lookups are becoming easier and easier to understand.

I'll take another look after you move it out from draft.

bpf/lib/conntrack.h
julianwiedmann: Ack, that didn't feel pretty to me either. Let's make it obvious then :+1:.

bpf/lib/conntrack.h
julianwiedmann: Oh yes, nice idea. Definitely follow-on material - we'd need to check how far back we can push this in the call chain without having subtle changes in behaviour. But would make a lot of sense to me.

julianwiedmann:
> Really nice cleanup! Conntrack lookups are becoming easier and easier to understand.
> 
> I'll take another look after you move it out from draft.

Glad that you feel it becomes easier to understand!. That's worth even more than the complexity reductions :blush:.

FWIW, I'm in no rush to squeeze this in just before feature freeze. Thinking it might make more sense to merge it early for v1.15, add the SNAT changes, give it some time to mature, and take a thorough pass over all the debug statements etc in the CT code to make sure that everything still fits.

bpf/lib/conntrack.h
```diff
@@ -380,32 +386,37 @@ __ct_lookup6(const void *map, struct ipv6_ct_tuple *tuple, struct __ctx_buff *ct
...
+	cilium_dbg3(ctx, DBG_CT_LOOKUP6_2, (tuple->nexthdr << 8) | tuple->flags,
+		    dir, scope);
+
+	switch (scope) {
+	case SCOPE_REVERSE:
+	case SCOPE_BIDIR:
+		/* Lookup in the reverse direction first: */
+		ret = __ct_lookup(map, ctx, tuple, action, dir, ct_state, is_tcp,
+				  tcp_flags, monitor);
+		if (ret != CT_NEW) {
```
gentoo-root, line 400: Nice! Much clearer now.

bpf/lib/dbg.h
```diff
@@ -60,7 +60,7 @@ enum {
 				 */
 	DBG_CT_LOOKUP4_2,       /* arg1: (nexthdr << 8) | flags
 				 * arg2: direction
```
gentoo-root, line 62: Hmm, it seems it used to be 0, but documented as direction.

gentoo-root (approved):

PR #25827, opened 2023-06-01
bgpv1: Reset BGP session in UpdateNeighbor if necessary

rastislavs:
As Cilium's user-facing API based on CRDs is intent-based, this PR ensures that updated BGP peer configuration is always applied also on existing BGP sessions immediately, even if it requires a session reset. If the underlying BGP implementation (GoBGP at the moment) does not reset the session for particular configuration items automatically, we enforce the session reset explicitly.

For example, changing `HoldTime` for a peer will cause the following `NOTIFICATION` message:

image

Followed by a session re-establishment and the new timer interval exchanged in the `OPEN` message:

image

### DIFF
--- a/pkg/bgpv1/gobgp/server.go
+++ b/pkg/bgpv1/gobgp/server.go
@@ -102,7 +102,7 @@ func NewGoBGPServerWithConfig(ctx context.Context, log *logrus.Entry, params typ
 // AddNeighbor will add the CiliumBGPNeighbor to the gobgp.BgpServer, creating
 // a BGP peering connection.
 func (g *GoBGPServer) AddNeighbor(ctx context.Context, n types.NeighborRequest) error {
-	peer, err := g.getPeerConfig(ctx, n.Neighbor, false)
+	peer, _, err := g.getPeerConfig(ctx, n.Neighbor, false)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


pkg/bgpv1/gobgp/server.go
```diff
@@ -117,39 +117,58 @@ func (g *GoBGPServer) AddNeighbor(ctx context.Context, n types.NeighborRequest)
...
 	}
-	if _, err = g.server.UpdatePeer(ctx, peerReq); err != nil {
+	updateRes, err := g.server.UpdatePeer(ctx, peerReq)
+	if err != nil {
 		return fmt.Errorf("failed while updating peer %v %v: %w", n.Neighbor.PeerAddress, n.Neighbor.PeerASN, err)
 	}
+
+	// perform full / soft peer reset if necessary
+	if needsFullReset || updateRes.NeedsSoftResetIn {
+		g.logger.Infof("Reseting peer %s (ASN %d) due to a config change", peer.Conf.NeighborAddress, peer.Conf.PeerAsn)
```
harsimran-pabla: nit: reseting -> resetting

pkg/bgpv1/gobgp/server.go
```diff
@@ -117,39 +117,58 @@ func (g *GoBGPServer) AddNeighbor(ctx context.Context, n types.NeighborRequest)
...
+		g.logger.Infof("Resetting peer %s (ASN %d) due to a config change", peer.Conf.NeighborAddress, peer.Conf.PeerAsn)
+		resetReq := &gobgp.ResetPeerRequest{
+			Address:       peer.Conf.NeighborAddress,
+			Communication: "Peer configuration changed",
+		}
+		if !needsFullReset {
+			resetReq.Soft = true
+			resetReq.Direction = gobgp.ResetPeerRequest_IN
+		}
+		if err = g.server.ResetPeer(ctx, resetReq); err != nil {
```
YutaroHayakawa, line 145: Do we have any situation that needs a soft reset now? IIUC, a soft reset is needed when we update the route filter or something that needs a peer to resend the route. In our current use case, it will never happen, right?

pkg/bgpv1/gobgp/server.go
```diff
@@ -117,39 +117,58 @@ func (g *GoBGPServer) AddNeighbor(ctx context.Context, n types.NeighborRequest)
 
 // UpdateNeighbor will update the existing CiliumBGPNeighbor in the gobgp.BgpServer.
 func (g *GoBGPServer) UpdateNeighbor(ctx context.Context, n types.NeighborRequest) error {
-	peer, err := g.getPeerConfig(ctx, n.Neighbor, true)
+	peer, needsFullReset, err := g.getPeerConfig(ctx, n.Neighbor, true)
```
YutaroHayakawa: Nit: I'd prefer calling this "hard reset" in contrast to the "soft reset".

pkg/bgpv1/gobgp/server.go
rastislavs: Right, at the moment we do not have a case for for it, I put it here mostly to have it covered correctly for the future.

YutaroHayakawa: Ok. Then, I don't have any objection to keeping it. BTW, TIL GoBGP doesn't use route refresh messages for inbound soft reset.

YutaroHayakawa (approved):
LGTM 👍 Thanks!

ldelossa (approved):
LGTM.

youngnick (approved):

rastislavs:
All checks passing and reviews are in. Ready to merge.

PR #24422, opened 2023-03-17
add native tunnel encapsulation support for the XDP Loadbalancer

julianwiedmann:
https://github.com/cilium/cilium/pull/21364 added support for combining XDP with tunnel-routing. This was done through a fallback mechanism, where we would (1) punt packets from XDP up to the `from-netdev` TC program, and (2) redirect them to `cilium_vxlan` / `cilium_geneve` as usual.

Besides the performance implications of the additional hop through the TC path, this also meant that https://github.com/cilium/cilium/pull/23890 later couldn't add support for IPv6 traffic with DSR-GENEVE using XDP (as there was no easy way of transporting the IPv6 DSR information from XDP all the way to the TC encap path).

This PR introduces native XDP support for adding VXLAN / GENEVE tunnel headers to a packet, and thus avoids the detour through the TC path.

open aspects
- ~~select a proper per-flow OuterSrcPort. It's not a hard blocker to get all the core infrastructure merged, so I plan to tackle this in a follow-on PR.~~ https://github.com/cilium/cilium/pull/26177
- ~~the `send_trace_notify()` in `__encap_with_nodeid()` happens too late for the XDP path - we want it to describe the inner packet, but it already has the tunnel headers. Easy enough to sort out in a follow-on PR.~~ https://github.com/cilium/cilium/pull/25828

PR #25828, opened 2023-06-01
bpf: encap: send TO_OVERLAY trace before adding encapsulation

julianwiedmann:
When used from TC context, ctx_set_encap_info() merely adds the tunnel metadata without touching the packet content. So the send_trace_notify() still describes the content of the inner packet.

But when called from XDP context, the ctx_set_encap_info() invocation pushes the actual tunnel headers. So the send_trace_notify() will report the *outer* packet.

Move the trace statement slightly up, so that it consistently describes the inner packet. Note that we now get a trace entry when the subsequent ctx_set_encap_info() fails - so we would need to match any resulting drop notification against the preceding TRACE_TO_OVERLAY.


julianwiedmann:
This addresses one of the follow-ons from https://github.com/cilium/cilium/pull/24422.

gentoo-root (approved):

PR #25829, opened 2023-06-01
tests: replace more incorrect DeepEquals uses

markpash:
The current script used to locate the wrong DeepEquals is incapable of finding all instances. Fix the remaining.

I used the following rule with `go-ruleguard` to find these instances. But I'm not entirely confident in introducing it into the CI yet.
```go
func checkAssertDeepEquals(m dsl.Matcher) {
	m.Match(
		`c.Assert($*_, check.DeepEquals, $*_)`,
		`c.Assert($*_, DeepEquals, $*_)`,
	).Report(`Found tests which use DeepEquals checker imported from github.com/cilium/checkmate.
		Cilium implements its own DeepEquals checker which can be imported from:
		github.com/cilium/cilium/pkg/checker
		It gives a more detailed trace when the assertion fails.`)
}
```

### DIFF
--- a/pkg/endpoint/policy_test.go
+++ b/pkg/endpoint/policy_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"k8s.io/apimachinery/pkg/util/sets"
 
+	"github.com/cilium/cilium/pkg/checker"
 	"github.com/cilium/cilium/pkg/identity"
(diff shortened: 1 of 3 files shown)


nebril (approved):
LGTM!

jrajahalme (approved):

tommyp1ckles (approved):

PR #25837, opened 2023-06-01
bpf: always inline .bss in absence of a .data section

ti-mo:
Due to an early vars == nil check in inlineGlobalData, ELFs without a .data section would not get their .bss references rewritten. This would cause verifier errors on pre-5.x kernels like 'unrecognized bpf_ld_imm64 insn', since those don't support pseudo-register load insns yet.

Delay the check and add a regression test.

```release-note
bpf: always inline .bss in absence of a .data section
```

### DIFF
--- a/pkg/bpf/collection_test.go
+++ b/pkg/bpf/collection_test.go
@@ -120,3 +120,36 @@ func TestInlineGlobalData(t *testing.T) {
 		t.Errorf("unexpected Instruction constant: want: 0x%x, got: 0x%x", want, got)
 	}
 }
+
+func TestInlineBSS(t *testing.T) {
+	// Rewrite references to .bss in the absence of any maps.
+	spec := &ebpf.CollectionSpec{
+		ByteOrder: binary.LittleEndian,
+		Programs: map[string]*ebpf.ProgramSpec{
+			"prog1": {
+				Instructions: asm.Instructions{
+					asm.LoadMapValue(asm.R8, 0, 0).WithReference(".bss"),
+					asm.Return(),
+				},
+			},
+		},
+	}
+
+	if err := inlineGlobalData(spec); err != nil {
+		t.Fatal(err)
+	}
+
+	insns := spec.Programs["prog1"].Instructions
+
+	if want, got := 0, int(insns[0].Constant); want != got {
+		t.Errorf("unexpected Instruction constant: want: 0x%x, got: 0x%x", want, got)
+	}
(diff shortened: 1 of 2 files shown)


rgo3 (approved):

qmonnet (approved):
Thanks!

pkg/bpf/collection_test.go
```diff
@@ -120,3 +120,36 @@ func TestInlineGlobalData(t *testing.T) {
...
+	if want, got := 0, int(insns[0].Constant); want != got {
+		t.Errorf("unexpected Instruction constant: want: 0x%x, got: 0x%x", want, got)
+	}
+
+	if want, got := asm.LoadImmOp(asm.DWord), insns[0].OpCode; want != got {
+		t.Errorf("unexpected Instruction OpCode: want: %s, got: %s", want, got)
+	}
+
+	if want, got := asm.R8, insns[0].Dst; want != got {
+		t.Errorf("unexpected Instruction OpCode: want: %s, got: %s", want, got)
```
gentoo-root, line 153: OpCode -> Dst

gentoo-root (changes requested):

qmonnet (changes requested):
The change breaks some of the unit tests, it introduces patterns such as:

```
             	150: (18) r2 = 0x0
            	152: (69) r2 = *(u16 *)(r2 +0)
            	R2 invalid mem access 'inv'
```

Observable [here](https://github.com/cilium/cilium/actions/runs/5179475666/jobs/9332408508?pr=25909) for [this test PR](https://github.com/cilium/cilium/pull/25909/files). Unit tests were skipped on your PR because it doesn't touch `bpf/` files.

dylandreimerink:
It looks like this change breaks global data which we use in the tests. For example in `bpf/tests/xdp_nodeport_lb4_nat_lb.c` we have a variable `static __be16 nat_source_port;`. It is zero initialized so it ends up in `.bss`, this change inlines it which should not happen. Giving the variables an initial value does not work, relocation entry moves to `.data` and then it also gets inlined.

I guess the tests always worked because we were hitting the edge case this PR is removing of not having any global data in `.data` so our global `.bss` variables would behave as expected.

I don't really know how to fix this one since we use global data for mocking and to confirm that mocked functions were invoked. Ideally we would mark our oddball Cilium variables in such a way that we can selectively inline them and not dynamic global variables.

PR #25844, opened 2023-06-01
Hubble: improve security by adding an option to redact API key in Kafka requests (L7)

ioandr:
Add option to redact API key in Kafka requests.

### TODOS

- [X] Accept option to redact Kafka API key
- [X] Add business logic to L7 parser
- [X] Add unit tests
- [X] Update Helm charts
- [X] Update docs 

Refs: https://github.com/cilium/cilium/issues/23887

Depends on: https://github.com/cilium/cilium/pull/25746

### DIFF
--- a/Documentation/observability/visibility.rst
+++ b/Documentation/observability/visibility.rst
@@ -78,11 +78,13 @@ parameters, API keys, and others.
    By default, Hubble does not redact potentially sensitive information
    present in `Layer 7 Hubble Flows <https://github.com/cilium/cilium/tree/master/api/v1/flow#flow-Layer7>`_.
 
-To harden security, Cilium provides the ``--hubble-redact`` option to control
-how Hubble handles sensitive information present in Layer 7 flows. More
-specifically, it offers the following features for supported Layer 7 protocols:
+To harden security, Cilium provides the ``--hubble-redact`` option which
+accepts a comma-separated list of values that control how Hubble handles
+sensitive information present in Layer 7 flows. More specifically, it offers
+the following features and values for supported Layer 7 protocols:
 
-* For HTTP: redacting URL query (GET) parameters
+* For HTTP: redacting URL query (GET) parameters (``--hubble-redact=http-url-query``)
+* For Kafka: redacting API key (``--hubble-redact=kafka-api-key``)
 
 For more information on configuring Cilium, see :ref:`Cilium Configuration <configuration>`.
 
--- a/install/kubernetes/cilium/values.yaml.tmpl
+++ b/install/kubernetes/cilium/values.yaml.tmpl
@@ -1010,10 +1010,11 @@ hubble:
   #
   #   redact:
   #   - http-url-query
+  #   - kafka-api-key
   #
   # You can specify the list of options from the helm CLI:
   #
-  #   --set hubble.redact="{http-url-query}"
+  #   --set hubble.redact="{http-url-query,kafka-api-key}"
   #
   redact: ~
 
--- a/pkg/hubble/parser/seven/kafka.go
+++ b/pkg/hubble/parser/seven/kafka.go
@@ -7,15 +7,26 @@ import (
 	"fmt"
 
 	flowpb "github.com/cilium/cilium/api/v1/flow"
+	"github.com/cilium/cilium/pkg/hubble/defaults"
+	"github.com/cilium/cilium/pkg/hubble/parser/options"
 	"github.com/cilium/cilium/pkg/proxy/accesslog"
 )
 
-func decodeKafka(flowType accesslog.FlowType, kafka *accesslog.LogRecordKafka) *flowpb.Layer7_Kafka {
+func decodeKafka(flowType accesslog.FlowType, kafka *accesslog.LogRecordKafka, opts *options.Options) *flowpb.Layer7_Kafka {
+	// Conditionally exclude the API key from the flow.
+	var apiKey string
+	if opts.RedactKafkaAPIKey {
+		apiKey = defaults.SensitiveValueRedacted
+	} else {
+		apiKey = kafka.APIKey
+	}
+
 	if flowType == accesslog.TypeRequest {
+		// Set only fields that are relevant for requests.
 		return &flowpb.Layer7_Kafka{
 			Kafka: &flowpb.Kafka{
 				ApiVersion:    int32(kafka.APIVersion),
-				ApiKey:        kafka.APIKey,
+				ApiKey:        apiKey,
 				CorrelationId: kafka.CorrelationID,
 				Topic:         kafka.Topic.Topic,
 			},
@@ -25,7 +36,7 @@ func decodeKafka(flowType accesslog.FlowType, kafka *accesslog.LogRecordKafka) *
 		Kafka: &flowpb.Kafka{
 			ErrorCode:     int32(kafka.ErrorCode),
 			ApiVersion:    int32(kafka.APIVersion),
-			ApiKey:        kafka.APIKey,
+			ApiKey:        apiKey,
 			CorrelationId: kafka.CorrelationID,
 			Topic:         kafka.Topic.Topic,
 		},
--- a/Documentation/helm-values.rst
+++ b/Documentation/helm-values.rst
@@ -1401,7 +1401,7 @@
      - bool
      - ``false``
    * - hubble.redact
-     - Configures the list of redact options for Hubble. Example:    redact:   - http-url-query  You can specify the list of options from the helm CLI:    --set hubble.redact="{http-url-query}"
+     - Configures the list of redact options for Hubble. Example:    redact:   - http-url-query   - kafka-api-key  You can specify the list of options from the helm CLI:    --set hubble.redact="{http-url-query,kafka-api-key}"
      - string
      - ``nil``
    * - hubble.relay.affinity
--- a/install/kubernetes/cilium/README.md
+++ b/install/kubernetes/cilium/README.md
@@ -400,7 +400,7 @@ contributors across the globe, there is almost always someone available to help.
 | hubble.peerService.clusterDomain | string | `"cluster.local"` | The cluster domain to use to query the Hubble Peer service. It should be the local cluster. |
 | hubble.peerService.targetPort | int | `4244` | Target Port for the Peer service, must match the hubble.listenAddress' port. |
 | hubble.preferIpv6 | bool | `false` | Whether Hubble should prefer to announce IPv6 or IPv4 addresses if both are available. |
-| hubble.redact | string | `nil` | Configures the list of redact options for Hubble. Example:    redact:   - http-url-query  You can specify the list of options from the helm CLI:    --set hubble.redact="{http-url-query}"  |
+| hubble.redact | string | `nil` | Configures the list of redact options for Hubble. Example:    redact:   - http-url-query   - kafka-api-key  You can specify the list of options from the helm CLI:    --set hubble.redact="{http-url-query,kafka-api-key}"  |
 | hubble.relay.affinity | object | `{"podAffinity":{"requiredDuringSchedulingIgnoredDuringExecution":[{"labelSelector":{"matchLabels":{"k8s-app":"cilium"}},"topologyKey":"kubernetes.io/hostname"}]}}` | Affinity for hubble-replay |
 | hubble.relay.dialTimeout | string | `nil` | Dial timeout to connect to the local hubble instance to receive peer information (e.g. "30s"). |
 | hubble.relay.enabled | bool | `false` | Enable Hubble Relay (requires hubble.enabled=true) |
--- a/install/kubernetes/cilium/templates/cilium-configmap.yaml
+++ b/install/kubernetes/cilium/templates/cilium-configmap.yaml
@@ -806,7 +806,7 @@ data:
 
 {{- if .Values.hubble.enabled }}
   # Enable Hubble gRPC service.
-  enable-hubble: {{ .Values.hubble.enabled  | quote }}
+  enable-hubble: {{ .Values.hubble.enabled | quote }}
   # UNIX domain socket for Hubble server to listen to.
   hubble-socket-path: {{ .Values.hubble.socketPath | quote }}
 {{- if hasKey .Values.hubble "eventQueueSize" }}
--- a/install/kubernetes/cilium/values.yaml
+++ b/install/kubernetes/cilium/values.yaml
@@ -1013,10 +1013,11 @@ hubble:
   #
   #   redact:
   #   - http-url-query
+  #   - kafka-api-key
   #
   # You can specify the list of options from the helm CLI:
   #
-  #   --set hubble.redact="{http-url-query}"
+  #   --set hubble.redact="{http-url-query,kafka-api-key}"
   #
   redact: ~
 
--- a/pkg/hubble/defaults/defaults.go
+++ b/pkg/hubble/defaults/defaults.go
@@ -17,4 +17,8 @@ const (
 	// DomainName specifies the domain name to use when constructing the server
 	// name for peer change notifications.
 	DomainName = "cilium.io"
+
+	// SensitiveValueRedacted is the string constant that is used to redact
+	// sensitive information.
+	SensitiveValueRedacted = "HUBBLE_REDACTED"
 )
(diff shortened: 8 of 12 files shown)


kaworu (changes requested):
Thanks for the PR @ioandr, patch LGTM.

I would like to see a new test ensuring that `KafkaRedactAPIKey` is honored. Also although a breaking change I think it should be `true` by default cc @rolinh.

pkg/hubble/parser/seven/kafka.go
```diff
@@ -7,11 +7,18 @@ import (
...
 	flowpb "github.com/cilium/cilium/api/v1/flow"
+	"github.com/cilium/cilium/pkg/hubble/defaults"
 	"github.com/cilium/cilium/pkg/proxy/accesslog"
 )
 
-func decodeKafka(flowType accesslog.FlowType, kafka *accesslog.LogRecordKafka) *flowpb.Layer7_Kafka {
+func decodeKafka(flowType accesslog.FlowType, kafka *accesslog.LogRecordKafka, redact bool) *flowpb.Layer7_Kafka {
+	// Don't include the API key in the flow.
+	if redact {
+		kafka.APIKey = defaults.SensitiveValueRedacted
```
rolinh: Unfortunately, we can only allow read access to the LogRecord structure and thus we should only change the value that is assigned to the flow's `Kafka.ApiKey` (see [here](https://github.com/cilium/cilium/blob/a899cf05a4b220568f142f6147951edea23ce10d/pkg/hubble/parser/seven/parser.go#L77-L80) for more context).

rolinh (changes requested):
Thanks for the patch! Unfortunately, we can't mutate LogRecord structures so the patch needs to be updated (see comment below).

> Also although a breaking change I think it should be true by default cc @rolinh.

Mmh, I'm not sure about that. If L7 visibility was an opt-out feature, then I'd clearly say yes, Hubble should redact Kafka's API keys by default. However, L7 visibility is opt-in and its (security) implications should be well understood when enabling it (note: we should probably update the documentation to make that clear). My point is that even if we try our best to redact sensitive data, there's no guarantee that Hubble will always correctly redact all sensitive data from L7 payloads. Therefore, L7 flows should be treated as sensitive data overall. I also think that if we redact (known) sensitive data by default, then that means we should do it overall which would imply that we also redact all HTTP headers by default and I'm not sure we would want that.

pkg/hubble/parser/seven/kafka.go
ioandr: Hi @rolinh, thanks for your feedback and for guiding the overall implementation.

>  [...] L7 visibility is opt-in and its (security) implications should be well understood when enabling it (note: we should probably update the documentation to make that clear)

Totally agree, it definitely makes sense to include this in the docs. Is the following guide the right place to add this piece of information?

https://docs.cilium.io/en/stable/observability/visibility/

Maybe we could add a new subsection named "Security" above "Troubleshooting" and "Limitations". WDYT?

> I also think that if we redact (known) sensitive data by default, then [...] and I'm not sure we would want that.

Based on this approach I understand that we should **not** redact sensitive info present in L7 flows by default. That is, we should set the default value of the newly introduced `KafkaRedactAPIKey` option to `false`.

kaworu:
@rolinh generally agree with all that you wrote

> I also think that if we redact (known) sensitive data by default, then that means we should do it overall which would imply that we also redact all HTTP headers by default and I'm not sure we would want that.

HTTP headers may or may not be sensitive data, while I think we can agree basic auth password and Kafka API key always are. We're already redacting the former unconditionally, so in my opinion doing it by default for the latter would be consistent. Not feeling too strongly about this though.

pkg/hubble/parser/seven/kafka.go
kaworu: A "Security Implications" section at https://docs.cilium.io/en/stable/observability/visibility/ sounds good to me, please document on the way that we're redacting HTTP basic auth password :pray:

kaworu (approved):
Thanks @ioandr for the quick update! The patch LGTM besides the missing doc update :rocket:

pkg/hubble/parser/seven/kafka.go
ChrsMark: Kind of related: isn't the mutation of the HTTP LogRecord at https://github.com/cilium/cilium/blob/e727175569f99c988d6474b9ca457fd082ec51dd/pkg/hubble/parser/seven/http.go#L33 
violating this "read-only" rule? Came across this while at https://github.com/cilium/cilium/pull/25746/files#diff-1df644554f9d2598342e7c4d8f9f3ae5af5fd29b52ed0dc88e01ea986f218a36R37 :)

ioandr:
> Thanks @ioandr for the quick update! The patch LGTM besides the missing doc

Sure, I can prepare a short doc as discussed above.

However the following is not 100% clear to me:

How are we planning to expose this option to end-users? That is, do we want to support allowing end-users to control this option to via the `cilium` CLI? For instance:

```
$ cilium config set hubble-l7-parser-kafka-redact-api-key true
```

Looking at the codebase, this requires the following:

1. Extend Cilium https://github.com/cilium/cilium/blob/main/pkg/option/config.go with an option to control whether Hubble's L7 parser redacts Kafka API keys - we already declare `Hubble*` options there
2. Set a sane default value for the newly added Hubble L7 option under https://github.com/cilium/cilium/blob/main/pkg/defaults/defaults.go
3. Pass context to the L7 payload parser upon creation: https://github.com/cilium/cilium/blob/main/daemon/cmd/hubble.go#L141
4. Possibly update Cilium's Helm chart as needed: https://docs.cilium.io/en/stable/helm-reference/

This is what @ChrsMark has done in his PR that tackles the redaction of HTTP query params, so maybe we want a uniform approach in both. PTAL: https://github.com/cilium/cilium/pull/25746/files

CC @kaworu @rolinh

Documentation/observability/visibility.rst
```diff
@@ -66,6 +66,32 @@ In order for Cilium to populate the ``INGRESS ENFORCEMENT``, ``EGRESS ENFORCEMEN
...
+sensitive information present in Layer 7 flows. More specifically:
+
+* **Kafka**: redact API key
+* **HTTP**: redact URL query parameters
+
+In future versions, Cilium will also support the following:
+
+* **HTTP**: sanitize specified headers
+* **HTTP**: sanitize user info in URL
+
```
kaworu: Without a set plan for theses feature (i.e. deadline or event release target) I would prefer to avoid mentioning them in the documentation.

```suggestion
```

Documentation/observability/visibility.rst
```diff
@@ -66,6 +66,32 @@ In order for Cilium to populate the ``INGRESS ENFORCEMENT``, ``EGRESS ENFORCEMEN
...
+    By default, Hubble does **not** protect potentially sensitive information
+    present in `Layer 7 Hubble Flows
+    <https://github.com/cilium/cilium/tree/master/api/v1/flow#flow-Layer7>`_.
+
+To harden security, Cilium provides options to configure how Hubble handles
+sensitive information present in Layer 7 flows. More specifically:
+
+* **Kafka**: redact API key
+* **HTTP**: redact URL query parameters
```
kaworu: IIUC this is part of https://github.com/cilium/cilium/pull/25746 so let's wait for it to be merged before merging this PR.

kaworu (changes requested):
Thanks @ioandr and @ChrsMark!

I think we should merge https://github.com/cilium/cilium/pull/25746 first, as it touches on HTTP for which we already redact the basic auth password, so it'll be easier to bring consistency starting from there (e.g. avoid mutating accesslog https://github.com/cilium/cilium/pull/25844#discussion_r1220380104).

About enabling/disabling, I think one cilium cli flag should be added instead of one per sensitive data redaction, otherwise we'll bloat the cli flag list. We can take inspiration of the current [Hubble metric configuration](https://github.com/cilium/cilium/blob/3f78ab717d855f4a376bf34606430280a902b9c8/install/kubernetes/cilium/templates/cilium-configmap.yaml#L813-L822) (see also [here](https://github.com/cilium/cilium/blob/3f78ab717d855f4a376bf34606430280a902b9c8/daemon/cmd/daemon_main.go#L935-L936)). Something along the line of:

```
--hubble-redact=kafa-api-key
--hubble-redact=http-basic-auth-password
--hubble-redact=http-url-query
```

What do you think? cc @rolinh

Documentation/observability/visibility.rst
ioandr: @kaworu this makes sense, we will remove those from the doc.

ioandr:
> Thanks @ioandr and @ChrsMark!
> 
> I think we should merge #25746 first, as it touches on HTTP for which we already redact the basic auth password, so it'll be easier to bring consistency starting from there (e.g. avoid mutating accesslog [#25844 (comment)](https://github.com/cilium/cilium/pull/25844#discussion_r1220380104)).
> 
> About enabling/disabling, I think one cilium cli flag should be added instead of one per sensitive data redaction, otherwise we'll bloat the cli flag list. We can take inspiration of the current [Hubble metric configuration](https://github.com/cilium/cilium/blob/3f78ab717d855f4a376bf34606430280a902b9c8/install/kubernetes/cilium/templates/cilium-configmap.yaml#L813-L822) (see also [here](https://github.com/cilium/cilium/blob/3f78ab717d855f4a376bf34606430280a902b9c8/daemon/cmd/daemon_main.go#L935-L936)). Something along the line of:
> 
> ```
> --hubble-redact=kafa-api-key
> --hubble-redact=http-basic-auth-password
> --hubble-redact=http-url-query
> ```
> 
> What do you think? cc @rolinh

Hi @kaworu @rolinh 

we did some work with @ChrsMark based on your feedback, PTAL https://github.com/cilium/cilium/pull/25746.

Let's move the review/discussion there and rebase the current (Kafka) PR once #25746 is merged.

Documentation/observability/visibility.rst
```diff
@@ -66,6 +66,27 @@ In order for Cilium to populate the ``INGRESS ENFORCEMENT``, ``EGRESS ENFORCEMEN
 and ``VISIBILITY POLICY`` fields, it must run with ``--endpoint-status=policy``
 to make field values visible.
 
+Security Implications
+---------------------
+
+Monitoring Layer 7 traffic comes with some serious security considerations in
+terms of handling potentially sensitive information, such as usernames,
+passwords, query parameters, API keys, etc.
```
zacharysarah: Edits for clarity, no Latin abbreviations
```suggestion
Monitoring Layer 7 traffic involves security considerations for handling potentially sensitive information, such as usernames, passwords, query parameters, API keys, and others.
```

Documentation/observability/visibility.rst
```diff
@@ -66,6 +66,27 @@ In order for Cilium to populate the ``INGRESS ENFORCEMENT``, ``EGRESS ENFORCEMEN
...
+
+.. Warning::
+
+By default, Hubble does **not** protect potentially sensitive information
+present in `Layer 7 Hubble Flows
+<https://github.com/cilium/cilium/tree/master/api/v1/flow#flow-Layer7>`_.
```
zacharysarah: Indent, omit bold
```suggestion
    By default, Hubble does not protect potentially sensitive information
    present in `Layer 7 Hubble Flows <https://github.com/cilium/cilium/tree/master/api/v1/flow#flow-Layer7>`_.
```

Documentation/observability/visibility.rst
```diff
@@ -66,6 +66,27 @@ In order for Cilium to populate the ``INGRESS ENFORCEMENT``, ``EGRESS ENFORCEMEN
...
+
+To harden security, Cilium provides options to configure how Hubble handles
+sensitive information present in Layer 7 flows. More specifically:
+
+* **HTTP**: redact URL query parameters
+* **Kafka**: redact API key
```
zacharysarah: Omit bold
```suggestion
* HTTP: redact URL query parameters
* Kafka: redact API key
```

Documentation/observability/visibility.rst
```diff
@@ -66,6 +66,27 @@ In order for Cilium to populate the ``INGRESS ENFORCEMENT``, ``EGRESS ENFORCEMEN
...
+
+For more information see :ref:`Cilium Configuration <configuration>`.
```
zacharysarah: Specific is better than general, comma nit
```suggestion
For more information on configuring Cilium, see :ref:`Cilium Configuration <configuration>`.
```

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -960,6 +960,18 @@ hubble:
   # -- Unix domain socket path to listen to when Hubble is enabled.
   socketPath: /var/run/cilium/hubble.sock
 
+  # -- Configures the list of redact options for Hubble.
+  # Example:
+  #
+  #   redact:
+  #   - http-url-query
+  #
+  # You can specify the list of options from the helm CLI:
```
zacharysarah, line 1015: Good use of present tense, active voice, and addressing the user as "you" ✨

zacharysarah (changes requested):
@ioandr Good start! Some small changes, otherwise LGTM for docs

ioandr:
Hi @zacharysarah thanks for your comments, they are spot on!

Note that in this PR we are adding to the "Security Implications" section of the `visibility.rst` document that https://github.com/cilium/cilium/pull/25746 introduces.

IMO it makes sense to apply suggested changes there, rebase the current PR on top of it and finally stick to your guidelines for adding Kafka-specific content.

CC @ChrsMark

ioandr:
Since https://github.com/cilium/cilium/pull/25746 got merged into `main` I rebased the current PR and resolved conflicts so that we follows the same conventions and guidelines (naming, documentation, CLI option) for both HTTP and Kafka.

LMK there are any actions needed from our side for the review.

CC @rolinh @kaworu @zacharysarah @qmonnet

qmonnet (approved):

rolinh (approved):
LGTM, thanks!

ChrsMark:
Hey @pippolo84 @nathanjsweet @asauber @aditighag any chances having a look into this one? It should be a simple one 🙏🏽. Despite being labeled with `dont-merge/wait-until-release` it would be great if can have it reviewed and ready for merge :).

zacharysarah (approved):
LGTM

pippolo84 (approved):
Thanks! :100:

jrajahalme (approved):
LGTM

kaworu (approved):
Thanks for the update @ioandr patch LGMT!

> @kaworu Looks like your concern was addressed?

Yes, although I still find it inconsistent to redact the HTTP password unconditionally while the Kafka API key redaction is opt-in, but that's arguably out of the scope of this PR.

ldelossa:
@ioandr, Can you please rebase this PR on `main` and push? Should fix some issues with the tests.

ioandr:
@ldelossa the only test that is failing is the mergeability one - I suppose due to the `dont-merge/wait-until-release` label that prevents merge.

LMK if you need any further action from my side.

asauber (approved):
LGTM, my main feedback here would be that we should list the "allowed" redacted values more clearly in the docs; make it clear that right now there are exactly two options and what they are.

ioandr:
> LGTM, my main feedback here would be that we should list the "allowed" redacted values more clearly in the docs; make it clear that right now there are exactly two options and what they are.

Hi @asauber, thank you for your feedback on this. How about we rephrase the doc as follows:

```
To harden security, Cilium provides the ``--hubble-redact`` option which
accepts a comma-separated list of values that control how Hubble handles
sensitive information present in Layer 7 flows. More specifically, it offers
the following features and values for supported Layer 7 protocols:

* For HTTP: redacting URL query (GET) parameters (``--hubble-redact=http-url-query``)
* For Kafka: redacting API key (``--hubble-redact=kafka-api-key``)
```

nathanjsweet (approved):

asauber:
@ioandr That phrasing looks good to me

ioandr:
@ldelossa I see that you removed the `dont-merge/wait-until-release` label so I suppose there is no blocker towards merging this.

I see that [one test](https://github.com/cilium/cilium/actions/runs/5413874539/jobs/9840682841) is failing with:

```
connectivity test failed: timeout reached waiting lookup for localhost from pod cilium-test/client-6965d549d5-mjxk9 to server on pod cilium-test/echo-other-node-545c9b778b-m6d6t to succeed (last error: context deadline exceeded)
```

I believe this is something temporary/flaky, that is, the test failure is not caused by the changes of this PR. Maybe we can re-trigger tests?

PR #25854, opened 2023-06-02
High-scale IPCache: Nodeport LB support Part 2

julianwiedmann:
Continuing from https://github.com/cilium/cilium/pull/25745, this PR is in the context of the high-scale ipcache feature described at https://github.com/cilium/design-cfps/pull/7.

It enables the nodeport LB to handle an GENEVE-encapsulated service request (from inside the clustermesh), and forward the request with the client's original security identity to the backend using GENEVE-DSR. It also adds full XDP passthrough support for this feature.

As we want to support in-clustermesh traffic to dedicated LB nodes (where the service translation happens in the LB, and not at the source), reduce the MTU sufficiently to tolerate the GENEVE DSR option that gets added by the LB.

In detail without XDP:

- encapsulated packet enters `from-netdev`. Gets manually decapsulated, and redirected to `cilium_geneve`.
- in `from-overlay`, `nodeport_lb4()` load-balances the inner packet. Tail-calls to the DSR egress path, and propagates the client's security identity along (that we previously obtained from the tunnel encapsulation).
- The DSR-GENEVE egress code adds fresh encapsulation, and inserts the Security identity here (instead of WORLD_ID).
- the load-balanced request is forwarded to the backend node.

And with XDP:

- as the packet enters the XDP program, we detect that it's GENEVE-encapsulated and locate the inner IPv4 header.
- `nodeport_lb4()` now operates on the inner packet (service lookup, DNAT etc).
- The DSR-GENEVE egress code learns how to update the existing encapsulation headers, and punches the DSR GENEVE option into the right location.
- the load-balanced request is forwarded to the backend node from inside XDP.

```release-note
Add support for load-balancing encapsulated requests in a configuration with high-scale ipcache.
```


julianwiedmann:
(the `coccicheck` warning is a false-positive. We merely raise the `DROP_MISSED_TAIL_CALL` drop notification a bit later in the code).

bpf/bpf_xdp.c
```diff
@@ -101,18 +101,101 @@ int tail_lb_ipv4(struct __ctx_buff *ctx)
...
-		} else if (IS_ERR(ret)) {
-			return send_drop_notify_error_ext(ctx, 0, ret, ext_err,
-							  CTX_ACT_DROP, METRIC_INGRESS);
+			ret = DROP_MISSED_TAIL_CALL;
 		}
 	}
 
+out:
+	if (IS_ERR(ret))
+		return send_drop_notify_error_ext(ctx, 0, ret, ext_err,
```
lmb, line 196: Nit: this will now send `_ext` err if the tail call failed. Does that matter?

bpf/lib/common.h
```diff
@@ -194,16 +205,17 @@ ____revalidate_data_pull(struct __ctx_buff *ctx, void **data_, void **data_end_,
...
 
-	*l3 = data + eth_hlen;
+	*l3 = data + l3_off;
 	return true;
 }
 
 static __always_inline __maybe_unused bool
 __revalidate_data_pull(struct __ctx_buff *ctx, void **data, void **data_end,
-		       void **l3, const __u32 l3_len, const bool pull)
+		       void **l3, const __u32 l3_off, const __u32 l3_len,
```
lmb, line 214: Nit: feels more conventional if `l3` and `l3_len` arent split in args.

bpf/lib/overloadable_xdp.h
```diff
@@ -239,6 +239,32 @@ ctx_set_encap_info(struct xdp_md *ctx, __u32 src_ip, __be16 src_port,
...
 }
+
+static __always_inline __maybe_unused int
+ctx_set_tunnel_opt(struct xdp_md *ctx, void *opt, __u32 opt_len)
+{
+	const __u32 geneve_off = ETH_HLEN + sizeof(struct iphdr) + sizeof(struct udphdr);
+	struct genevehdr geneve;
+
+	/* add free space after GENEVE header: */
+	if (ctx_adjust_hroom(ctx, opt_len, BPF_ADJ_ROOM_MAC, ctx_adjust_hroom_flags()) < 0)
```
lmb, line 250: Noob question: this doesn't take `geneve_off` so how does the function know where to add space?

bpf/lib/overloadable_xdp.h
```diff
@@ -239,6 +239,32 @@ ctx_set_encap_info(struct xdp_md *ctx, __u32 src_ip, __be16 src_port,
...
+
+	/* write the options */
+	if (ctx_store_bytes(ctx, geneve_off + sizeof(geneve), opt, opt_len, 0) < 0)
+		return DROP_WRITE_ERROR;
+
+	/* update the options length in the GENEVE header: */
+	if (ctx_load_bytes(ctx, geneve_off, &geneve, sizeof(geneve)) < 0)
+		return DROP_INVALID;
+
+	geneve.opt_len += (__u8)(opt_len >> 2);
```
lmb, line 261: I guess opt_len must be a multiple of 4 or something. Can we `build_bug_on` or similar? Or document that somewhere.

pkg/mtu/mtu.go
```diff
@@ -101,7 +105,7 @@ type Configuration struct {
 // specified, otherwise it will be automatically detected. if encapEnabled is
 // true, the MTU is adjusted to account for encapsulation overhead for all
 // routes involved in node to node communication.
-func NewConfiguration(authKeySize int, encryptEnabled bool, encapEnabled bool, wireguardEnabled bool, mtu int, mtuDetectIP net.IP) Configuration {
+func NewConfiguration(authKeySize int, encryptEnabled bool, encapEnabled bool, wireguardEnabled bool, hsIpcacheDSRenabled bool, mtu int, mtuDetectIP net.IP) Configuration {
```
lmb, line 108: Drive by comment: this really ought to take a struct. All of the bools makes this unreadable at callsites.

Nit: capitalization of `hsIpcacheDSRenabled` seems wonky.

lmb (approved):
Total noob as to sig-datapath changes, I'll try and find someone else to review. Other owned changes LGTM.

bpf/bpf_xdp.c
julianwiedmann: That's fine. In the end this is for fine-granular debugging, so we would know whether `ext_err` is even relevant for the reported DROP reason.

bpf/lib/common.h
julianwiedmann: :shrug: either way feels fine to me

bpf/lib/overloadable_xdp.h
julianwiedmann: See the corresponding change in `xdp.h`'s `ctx_adjust_hroom()`. It's completely custom and fragile - we just "know" how to behave for specific `hroom` values.

bpf/lib/overloadable_xdp.h
julianwiedmann: To a certain extent I would hope that callers know what data format they need to follow for such an option (as it also needs to work for the TC variant of `set_tunnel_opt()`), and get to keep the pieces if they don't.

But yep, a build_bug() should be doable - will queue that up as follow-on, thanks!

pkg/mtu/mtu.go
julianwiedmann: No disagreement from my side, this has grown way out of proportion :grimacing: .

julianwiedmann:
> Total noob as to sig-datapath changes, I'll try and find someone else to review. Other owned changes LGTM.

Thank you! Hopefully @ldelossa and/or @borkmann can have a look :).

gandro (approved):
My codeowners looks good to me. Seconding the proposal to refactor the MTU constructor at some point.

pkg/mtu/mtu.go
```diff
@@ -32,6 +32,10 @@ const (
 	//    Total extra bytes:  50B
 	TunnelOverhead = 50
 
+	// DsrTunnelOverhead is about the GENEVE DSR option that gets inserted
+	// by the LB, when addressing a Service in hs-ipcache mode
+	DsrTunnelOverhead = 12
```
ldelossa, line 37: Why this name? 

If the comments immediately point out that this value is used for "GENEVE DSR" then why not make the value also contain this detail? Seems pretty specific to Geneve.

pkg/mtu/mtu.go
ldelossa, line 108: Just a nit but:

s/hsIpcacheDSRenabled/hsIPCacheDSREnabled

bpf/lib/common.h
```diff
@@ -150,12 +150,13 @@ union v6addr {
 	__u8 addr[16];
 } __packed;
 
-static __always_inline bool validate_ethertype(struct __ctx_buff *ctx,
-					       __u16 *proto)
+static __always_inline bool validate_ethertype_l2_off(struct __ctx_buff *ctx,
```
ldelossa, line 153: What's your thoughts on taking and optionally filling a pointer to an `ethhdr` here, akin to how we fill in layer 3 headers in `revalidate_data`. 

Could be handy if we quickly need to inspect l2 attributes?

ldelossa (approved):
All nits, LGTM.

pkg/mtu/mtu.go
julianwiedmann: :camel:

pkg/mtu/mtu.go
julianwiedmann: :+1: probably just leftovers from the initial prototype, and missed while polishing.

bpf/lib/common.h
julianwiedmann: Do we have any paths that need more than just the ethertype today?

julianwiedmann:
Reviews are in, tests are green (with one exception for https://github.com/cilium/cilium/pull/25854#issuecomment-1573709083) -> ready to merge.

@dylandreimerink could you please do the honors? :pray:

dylandreimerink:
What is the deal with the `file ./bpf_xdp.c: DROP_MISSED_TAIL_CALL missing after tail call on line 189` error thrown by coccicheck?

pchaigno:
@dylandreimerink @julianwiedmann Why was this merged when it is known to break Coccicheck? Coccicheck is a Required CI job.

julianwiedmann:
> @dylandreimerink @julianwiedmann Why was this merged when it is known to break Coccicheck? Coccicheck is a Required CI job.

It was a false-positive for this change, and we already had plenty of existing code paths that follow the same pattern. Why is coccicheck now complaining about this specific part, but the 20 others are fine? :angry:

Either way, https://github.com/cilium/cilium/pull/25924.

pchaigno:
I understand all that, but it's still a CI breakage.

julianwiedmann:
> I understand all that, but it's still a CI breakage.

ack, sorry. The expectation was that it would only check the actual patch diff :disappointed:

pchaigno:
Ah, I see :+1: Only checkpatch works like that today, which is part of the reason why checkpatch is not marked Required.

PR #25862, opened 2023-06-02
CI: Verifier tests: Keep generated object files and logs on test failure

qmonnet:
The CI workflow for testing Cilium's programs against the verifier is supposed to upload its log files and the generated files on failures. However, we lose some of this data, because:

- Multiple jobs (for different kernel versions) upload their log and object files to the same artifact, overwriting previous uploads in the process, so that we only keep the files uploaded by the last job to complete. We can address that by using different artifact names for the different jobs.

- The test itself runs `make -C bpf/ clean` before testing the programs of each source file, thus removing the object files previously generated. Instead, we can clean up the directory just once before running the tests, not between two tests.

- At last, each source file is compiled multiple times with different sets of options, resulting in only the last version of this object file being kept. The solution implemented here consists in renaming these object files (on test failure) to make sure we keep them around.

### DIFF
--- a/test/verifier/verifier_test.go
+++ b/test/verifier/verifier_test.go
@@ -87,6 +87,12 @@ func TestVerifier(t *testing.T) {
 	kernelVersion, source := getCIKernelVersion(t)
 	t.Logf("CI kernel version: %s (%s)", kernelVersion, source)
 
+	cmd := exec.Command("make", "-C", "bpf/", "clean")
+	cmd.Dir = *ciliumBasePath
+	if out, err := cmd.CombinedOutput(); err != nil {
+		t.Fatalf("Failed to clean bpf objects: %v\ncommand output: %s", err, out)
+	}
+
 	for _, bpfProgram := range []struct {
 		name      string
 		macroName string
@@ -124,12 +130,6 @@ func TestVerifier(t *testing.T) {
 
 			name := fmt.Sprintf("%s_%d", bpfProgram.name, i)
 			t.Run(name, func(t *testing.T) {
-				cmd := exec.Command("make", "-C", "bpf/", "clean")
-				cmd.Dir = *ciliumBasePath
-				if out, err := cmd.CombinedOutput(); err != nil {
-					t.Fatalf("Failed to clean bpf objects: %v\ncommand output: %s", err, out)
-				}
-
 				cmd = exec.Command("make", "-C", "bpf", fmt.Sprintf("%s.o", bpfProgram.name))
 				cmd.Dir = *ciliumBasePath
 				cmd.Env = append(cmd.Env,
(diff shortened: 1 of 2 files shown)


test/verifier/verifier_test.go
```diff
@@ -179,6 +181,10 @@ func TestVerifier(t *testing.T) {
 				})
 				var ve *ebpf.VerifierError
 				if errors.As(err, &ve) {
+					// Rename object file to avoid subsequent runs to overwrite it,
+					// so we can keep it for CI's artifact upload.
+					os.Rename(objFile, path.Join(*ciliumBasePath, "bpf", fmt.Sprintf("%s_%d.o", bpfProgram.name, i)))
```
tklauser: nit: this could reuse `name` here which is defined here: https://github.com/cilium/cilium/pull/25862/files#diff-941fe2ebb6a483a4cc321ef3fff0f95f576b45f810bce9416eaefb35b46587b7R133

```suggestion
					os.Rename(objFile, path.Join(*ciliumBasePath, "bpf", name + ".o"))
```

tklauser (approved):
Nice, thanks! One nit inline in case a respin is needed.

test/verifier/verifier_test.go
qmonnet: Makes sense, I'll update. Thanks!

ti-mo:
Let's give all .o's a unique name instead and include it as part of the name of the subtest. I think renaming is confusing, and it'd be nice to have all objs in the zip for debugging.

test/verifier/verifier_test.go
```diff
@@ -179,6 +181,10 @@ func TestVerifier(t *testing.T) {
 				})
 				var ve *ebpf.VerifierError
 				if errors.As(err, &ve) {
+					// Rename object file to avoid subsequent runs to overwrite it,
+					// so we can keep it for CI's artifact upload.
+					os.Rename(objFile, path.Join(*ciliumBasePath, "bpf", name+".o"))
```
viktor-kurchenko: Ignored error, maybe log it at least for better visibility?

viktor-kurchenko (approved):
LGTM.

julianwiedmann (approved):
lgtm

qmonnet:
Updated to rename all object files regardless of exit status. I checked (locally) that all object files are preserved. @ti-mo does this correspond to what you had in mind?

ti-mo:
> Updated to rename all object files regardless of exit status. I checked (locally) that all object files are preserved. @ti-mo does this correspond to what you had in mind?

Yes, thanks! The make invocation is indeed what deterred me from doing the renaming at the time. We can start using Cilium's Go clang pipeline here at some point for more flexibility. In the meantime, this will do!

ti-mo:
Not sure why `ci-verifier` is not starting since the last push was 18 hours ago.

ti-mo (approved):

qmonnet:
Runtime tests on Jenkins and GitHub are currently broken due to https://github.com/cilium/cilium/issues/25968, and net-next on Jenkins hit https://github.com/cilium/cilium/issues/25958 - both unrelated to the current PR.

None of these workflows are affected by the change in this PR. I'm marking as ready to merge.

PR #25864, opened 2023-06-02
k8s: Update comment about rule preprocessing

odinuge:
There is a difference between headless services, and services without a selector (sometimes referred to as external services in cilium). Those services without a selector can be headless or have a ClusterIP; but the important thing, is that this function care about services without a selector, independent on the ClusterIP. See commit messages for more context.

### DIFF
--- a/pkg/k8s/rule_translate.go
+++ b/pkg/k8s/rule_translate.go
@@ -221,7 +221,7 @@ func (k RuleTranslator) deleteToCidrFromEndpoint(
 	return toReleasePrefixes, nil
 }
 
-// PreprocessRules translates rules that apply to headless services
+// PreprocessRules translates egress rules that apply to external services (ToServices)
 func PreprocessRules(r api.Rules, cache *ServiceCache) error {
 
 	cache.mutex.Lock()
--- a/Documentation/security/policy/language.rst
+++ b/Documentation/security/policy/language.rst
@@ -284,8 +284,8 @@ namespace ``default``.
         .. literalinclude:: ../../../examples/policies/l3/service/service.json
 
 This example shows how to allow all endpoints with the label ``id=app2``
-to talk to all endpoints of all kubernetes headless services which
-have ``head:none`` set as the label.
+to talk to all endpoints of all kubernetes services without selectors which
+have ``external:yes`` set as the label.
 
 .. only:: html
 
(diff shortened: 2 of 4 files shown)


pkg/k8s/rule_translate.go
```diff
@@ -221,7 +221,7 @@ func (k RuleTranslator) deleteToCidrFromEndpoint(
 	return toReleasePrefixes, nil
 }
 
-// PreprocessRules translates rules that apply to headless services
+// PreprocessRules translates rules that apply to services without selectors
```
qmonnet: Thanks! I'm not familiar with the terminology here, so I may be wrong. But looking at the code and history, I see that this function would initially apply rules on headless services indeed. It changed with commit a270d52ef158371fa2be06db53df39866af0080c. Based on the code, should the comment be:

```suggestion
// PreprocessRules translates egress rules that apply to external services (ToServices)
```

Looking at the description for `ToServices`, the fact that these K8s services have no selectors seems more like the current support status and could possibly change one day?

Since 09b3df13a5f3e7232075bff6b33fd3b885726301:

```
	// ToServices is a list of services to which the endpoint subject
	// to the rule is allowed to initiate connections.
	// Currently Cilium only supports toServices for K8s services without
	// selectors.
```

odinuge: > Thanks! I'm not familiar with the terminology here, so I may be wrong. But looking at the code and history, I see that this function would initially apply rules on headless services indeed. It changed with commit https://github.com/cilium/cilium/commit/a270d52ef158371fa2be06db53df39866af0080c. Based on the code, should the comment be:

Sweet. I guess the egress part is implicit, since there is no way to have ingress rules service. I'll update the comment based on your suggestion, thanks!

qmonnet (approved):
Thank you

joamaki:
@joestringer could you take a quick look to see if this is correct from the policy perspective? Do we potentially have more cases where this might be wrong?

joestringer (approved):

odinuge:
Are we good to merge this now? Want to get this merged to avoid conflicts with other pending work

qmonnet:
Ping @nathanjsweet

nathanjsweet (approved):

PR #25866, opened 2023-06-02
[WIP] bpf: modularize lib/fib.h

ldelossa:
This commit has further modularized fib.h, making it possible to perform a fib lookup independently from a redirect. As a result of this refactor, two new functions have been created: fib_lookup_v4 and fib_lookup_v6.

This is a WIP PR which will not merge. 

Opening to get an initial view on the split refactor. 

This work ties into an SRv6 related fix and will be ultimately incorporated into a PR which includes SRv6 fix.

### DIFF
--- a/bpf/lib/fib.h
+++ b/bpf/lib/fib.h
@@ -40,50 +40,63 @@ static __always_inline bool fib_ok(int ret)
 	return likely(ret == CTX_ACT_TX || ret == CTX_ACT_REDIRECT);
 }
 
-/* fib_redirect() is common helper code which performs fib lookup, populates
- * the corresponding hardware addresses and pushes the packet to a target
- * device for the next hop. Calling fib_redirect_v{4,6} is preferred unless
- * due to NAT46x64 struct bpf_fib_lookup_padded needs to be prepared at the
- * callsite. oif must be 0 if otherwise not passed in from the BPF CT. The
- * needs_l2_check must be true if the packet could transition between L2->L3
- * or L3->L2 device.
- */
+ /* fib_do_redirect will redirect the ctx to a particular output interface.
+  *
+  * the redirect can occur with or without a previous call to fib_lookup.
+  *
+  * if a previous fib_lookup was performed, this function will attempt to redirect
+  * to the output interface in the provided 'fib_params', as long as 'fib_ret'
+  * is set to 'BPF_FIB_LKUP_RET_SUCCESS'
+  *
+  * if a previous fib_lookup was performed and the return was 'BPF_FIB_LKUP_NO_NEIGH'
+  * this function will then attempt to copy the af_family and destination address
+  * out of 'fib_params' and into 'redir_neigh' struct then perform a
+  * 'redirect_neigh'.
+  *
+  * if no previous fib_lookup was performed, and the desire is to simply use
+  * 'redirect_neigh' then set 'fib_params' to nil and 'fib_ret' to
+  * 'BPF_FIB_LKUP_RET_NO_NEIGH'.
+  * in this case, the 'oif' value will be used for the 'redirect_neigh' call.
+  *
+  * in a special case, if a previous fib_lookup was performed, and the return
+  * was 'BPF_FIB_LKUP_RET_NO_NEIGH', and we are on a kernel version where
+  * the target interface for the fib lookup is not returned
+  * (due to ARP failing, see Kernel commit d1c362e1dd68) the provided 'oif'
+  * will be used as output interface for redirect.
+  */
 static __always_inline int
-fib_redirect(struct __ctx_buff *ctx, const bool needs_l2_check,
-	     struct bpf_fib_lookup_padded *fib_params, __s8 *fib_err, int *oif)
+fib_do_redirect(struct __ctx_buff *ctx, const bool needs_l2_check,
+		const struct bpf_fib_lookup_padded *fib_params, __s8 *fib_ret,
+		int *oif)
 {
 	struct bpf_redir_neigh nh_params;
 	struct bpf_redir_neigh *nh = NULL;
-	bool no_neigh = is_defined(ENABLE_SKIP_FIB);
 	int ret;
 
-#ifndef ENABLE_SKIP_FIB
-	ret = fib_lookup(ctx, &fib_params->l, sizeof(fib_params->l), 0);
-	if (ret != BPF_FIB_LKUP_RET_SUCCESS) {
-		*fib_err = (__s8)ret;
-		if (likely(ret == BPF_FIB_LKUP_RET_NO_NEIGH)) {
-			nh_params.nh_family = fib_params->l.family;
-			__bpf_memcpy_builtin(&nh_params.ipv6_nh,
-					     &fib_params->l.ipv6_dst,
-					     sizeof(nh_params.ipv6_nh));
-			nh = &nh_params;
-			no_neigh = true;
+	/* sanity check, we only enter this function with these two fib lookup
+	 * return codes.
+	 */
+	if (*fib_ret && (*fib_ret != BPF_FIB_LKUP_RET_NO_NEIGH))
+		return DROP_NO_FIB;
+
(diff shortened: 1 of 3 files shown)


bpf/lib/fib.h
```diff
@@ -93,94 +106,172 @@ fib_redirect(struct __ctx_buff *ctx, const bool needs_l2_check,
...
+		__s8 *fib_err __maybe_unused, int iif, int *oif)
+{
+	struct bpf_fib_lookup_padded fib_params __maybe_unused = {0};
+	void *data_end;
+	void *data;
+	__s8 skip_fib __maybe_unused = BPF_FIB_LKUP_RET_NO_NEIGH;
+	int ret;
+
+	/* TODO: Fix this*/
+	(void)iif;
```
ldelossa, line 216: This is here since all invocation of `fib_redirect_*` get the iif from the ctx, which occurs now in fib_lookup.

Moving forward, if there's no reason to pass `iif` into the function, it can be removed from the signature, but deferring that to a later commit to update call sights.

ldelossa:
Verifier is complaining about misaligned stack access when we get to [here](https://github.com/cilium/cilium/pull/25866/files#diff-ad84666674cdaf193b58f325d0953ac4df4cda8b37ed73b17a1116a4d7e9ef62R113). That's where we re-write the l2 addresses from the fib lookup params into the sk_buff. 

:thinking: 

```
             	; fib_params->l.ipv4_dst	= ip4->daddr;
            	3093: (63) *(u32 *)(r10 -40) = r2
            	3094: (b7) r2 = 14099
            	; case  6: __it_mob(d, s, 16); __it_mob(d, s, 32);
            	3095: (6b) *(u16 *)(r10 -10) = r2
            	3096: (18) r2 = 0xefbeadde
            	; case  6: __it_mob(d, s, 16); __it_mob(d, s, 32);
            	3098: (63) *(u32 *)(r10 -20) = r2
            	3099: (b7) r2 = 924006163
            	; case  6: __it_mob(d, s, 16); __it_mob(d, s, 32);
            	3100: (63) *(u32 *)(r10 -14) = r2
            	misaligned stack access off (0x0; 0x0)+0+-14 size 4
            	processed 695 insns (limit 1000000) max_states_per_insn 0 total_states 44 peak_states 44 mark_read 21
```

ldelossa:
Oh wait I was wrong. 

The verifier was upset because in two `mock_fib_lookup` `memcpy` was being use directly, and assuming `fib_params` were 16 byte aligned. 

Once I swapped these to build to:

```
long mock_fib_lookup(__maybe_unused void *ctx, struct bpf_fib_lookup *params,
		     __maybe_unused int plen, __maybe_unused __u32 flags)
{
	__bpf_memcpy_builtin(params->smac, fib_smac, ETH_ALEN);
	__bpf_memcpy_builtin(params->dmac, fib_dmac, ETH_ALEN);
	return 0;
}
```

unit tests began to pass. pushing change now.

ldelossa:
Oh wait I was wrong. 

The verifier was upset because in two `mock_fib_lookup` `memcpy` was being use directly, and assuming `fib_params` were 16 byte aligned. 

Once I swapped these to build to:

```
long mock_fib_lookup(__maybe_unused void *ctx, struct bpf_fib_lookup *params,
		     __maybe_unused int plen, __maybe_unused __u32 flags)
{
	__bpf_memcpy_builtin(params->smac, fib_smac, ETH_ALEN);
	__bpf_memcpy_builtin(params->dmac, fib_dmac, ETH_ALEN);
	return 0;
}
```

unit tests began to pass. pushing change now.

ldelossa:
lol, unit test failed on this:

```
-- FAIL: TestBPF (34.19s)
    --- FAIL: TestBPF/tc_nodeport_l3_dev.o (1.86s)
        bpf_test.go:231: Failed to map addr=192.168.0.2 to identity
        bpf_test.go:231: Attempting local delivery for container id 233 from seclabel 2
        bpf_test.go:231: Failed to map addr.p4=[::0:2] to identity
        bpf_test.go:231: Attempting local delivery for container id 233 from seclabel 2
        bpf_test.go:268: error while interpolating using source files: cparser parse file: new parser: lex next: /home/runner/work/cilium/cilium/bpf/lib/fib.h:48:45: invalid input text "`fib_params`, as..."
```

Which is part of a comment string ? O_o. 

Just made checkpatch happy and re-pushed, maybe I had some odd formatting.

ldelossa:
Hmm tests look mostly good, but I dont like that nat64 tests in "Cilium L4LB XDP" are failing. 

The nat64 code calls the plain "fib_redirect" which is altered.

bpf/lib/fib.h
```diff
@@ -112,75 +145,135 @@ fib_redirect(struct __ctx_buff *ctx, const bool needs_l2_check,
...
 
 	ret = ipv4_l3(ctx, l3_off, NULL, NULL, ip4);
 	if (unlikely(ret != CTX_ACT_OK))
 		return ret;
 
-	return fib_redirect(ctx, needs_l2_check, &fib_params, fib_err, oif);
+	/* we touched ttl above, revalidate for any header reads/writes. */
+	if (!revalidate_data(ctx, &data, &data_end, &ip4))
+		return DROP_INVALID;
+
```
julianwiedmann, line 267: Would be great if we could avoid this hidden revalidation.

How about the two following two call patterns instead:

``` 
fib_fill_params_v4(fib_params, ip4);
ret = fib_lookup(params);
```

and

``` 
int fib_redirect_v4()
{
    fib_fill_params_v4(fib_params, ip4);
    ret = ipv4_l3(...);
    ...
    ret = fib_lookup(...);
    return fib_do_redirect(...);
}
```

bpf/lib/fib.h
```diff
@@ -40,50 +40,63 @@ static __always_inline bool fib_ok(int ret)
...
-			__bpf_memcpy_builtin(&nh_params.ipv6_nh,
-					     &fib_params->l.ipv6_dst,
-					     sizeof(nh_params.ipv6_nh));
-			nh = &nh_params;
-			no_neigh = true;
+	/* sanity check, we only enter this function with these two fib lookup
+	 * return codes.
+	 */
+	if (*fib_ret && (*fib_ret != BPF_FIB_LKUP_RET_NO_NEIGH))
+		return DROP_NO_FIB;
```
julianwiedmann, line 80: These games with `fib_ret` are a bit difficult to follow :/. Can't we do this error check in the caller already, and pass just a `l2_info_avail` flag instead?

julianwiedmann (commented):
Some comments, mostly laying things out a bit differently to better understand the impact. Right now the diff for `fib_do_redirect()` is a bit :headache:.

But I get the overall idea :+1:

bpf/lib/fib.h
ldelossa: Hmm, 

Calling `fib_lookup` directly inside `fib_redirect_v4` kinda defeats the purpose of this PR, which is to expose a function which performs `fib_lookup` independently. 

Eventually, the DP will use `fib_lookup_v4` on its own, without a redirect involved. 

I'm not sure I understand what you mean by "hidden". 
We need to re-validate since `ipv4_l3` updates the ttl, and this writes to the sk_buff.

bpf/lib/fib.h
ldelossa: The desire here was allow the calling code to perform a `fib_lookup` and then immediately call `fib_do_redirect` with the results, simplifying the caller's usage of lib/fib.h.

This "feels" better to me, as we are asking the caller to known less about these fib functions. 

What I think you're asking is basically replace `fib_ret` with a bool, something like `redirect_neigh`, taking true or false. Which I think is sane, but then the caller is responsible for knowing to take the results from a fib_lookup and turn it into the bool, which I don't love because the caller then bloats the call site with (or something similar). 

```
bool redirect_neigh = false
if (fib_ret && fib_ret == BPF_FIB_LKUP_RET_NO_NEIGH) {
  redirect_neigh = true 
} else {
  // handle fib error
}
```

PR #25870, opened 2023-06-02
Allow users to ignore links when enabling XDP

### LINKED ISSUE #24768: XDP: Failing to attach XDP program to a tagged VLAN interface

### What happened?

With Cilium 1.13.0, we could successfully enable XDP when a node has a single native VLAN. However, after we added a tagged VLAN (namely, vlan3 with the vlan id 3) on the same node along with the native VLAN, and then restarted its cilium agent, we got the following error.
```
Failed to compile XDP program" error="program cil_xdp_entry: attaching XDP program to interface vlan3: operation not supported" subsys=datapath-loader
```
Here is our installation parameter set. Basically we use BPG to advertise both POD CIRD and Service CIDR with xvlan tunneling for Pod-to-pod communication.
```
helm upgrade --namespace kube-system cilium cilium/cilium --version 1.13.0 --set bgpControlPlane.enabled=true,bpf.masquerade=true,cluster.id=0,cluster.name=k8s-cluster-01,encryption.nodeEncryption=false,ipMasqAgent.enabled=true,ipam.mode=kubernetes,ipam.operator.clusterPoolIPv4MaskSize=26,ipam.operator.clusterPoolIPv4PodCIDR=172.16.0.0/12,kubeProxyReplacement=strict,loadBalancer.acceleration=native,operator.replicas=1,serviceAccounts.cilium.name=cilium,serviceAccounts.operator.name=cilium-operator,tunnel=vxlan
Release "cilium" has been upgraded. Happy Helming!
NAME: cilium
LAST DEPLOYED: Tue Apr  4 15:47:21 2023
NAMESPACE: kube-system
STATUS: deployed
REVISION: 9
TEST SUITE: None
NOTES:
You have successfully installed Cilium with Hubble.

Your release version is 1.13.0.
```
Just in case, here is an example of the success case when there is only a single VLAN as a good reference. I think such a result can prove our Mellanox driver supports XDP.
But unfortunately I could not see such a good result after adding a tagged VLAN on the native VLAN interface.
```
# bpftool net
xdp:
ens785np0(2) driver id 1641

tc:
ens785np0(2) clsact/ingress cilium-ens785np0 id 1754
ens785np0(2) clsact/egress cilium-ens785np0 id 1756
cilium_net(4) clsact/ingress cilium-cilium_net id 1737
cilium_host(5) clsact/ingress cilium-cilium_host id 1718
cilium_host(5) clsact/egress cilium-cilium_host id 1724
cilium_vxlan(6) clsact/ingress bpf_overlay.o:[from-overlay] id 1576
cilium_vxlan(6) clsact/egress bpf_overlay.o:[to-overlay] id 1587
lxc_health(11) clsact/ingress cilium-lxc_health id 1674
```

### Cilium Version

1.13.0

### Kernel Version

Linux worknode-1 5.15.0-67-generic #74~20.04.1-Ubuntu SMP Wed Feb 22 14:52:34 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

### Kubernetes Version

v1.24.10+rke2r1

### Sysdump

Somehow my sysdump file is too big (189MB), I cannot upload it here because its size exceeds 25 MB.

### Relevant log output

```shell
$ cilium status
...
...
Errors:           cilium             cilium-jmgtm    unable to retrieve cilium status: container cilium-agent is in CrashLoopBackOff, exited with code 1: level=fatal msg="Failed to compile XDP program" error="program cil_xdp_entry: attaching XDP program to interface vlan3: operation not supported" subsys=datapath-loader
...
```

### Anything else?

# lspci | grep -i ether
01:00.0 Ethernet controller: Intel Corporation Ethernet Controller 10G X550T (rev 01)
65:00.0 Ethernet controller: Mellanox Technologies MT28908 Family [ConnectX-6]

### Code of Conduct

poblahblahblah:

- [?] All code is covered by unit and/or runtime tests where feasible.
      GitHub issue.
- [?] Are you a user of Cilium? Please add yourself to the [Users doc](https://github.com/cilium/cilium/blob/main/USERS.md)

This allows a user to specify a regex that will cause a link to be ignored when enabling XDP. We need this ability since we are running physical hosts where the primary NICs do support XDP, but have additional `vlan` devices that _do not_ have XDP support in the driver.

More background can be found in #24768, but to summarize we have 1 physical device (ens785np0) per host and due to network design we need an additional `vlan` type interface (ens785np0.3). Regrettably the `vlan` driver in the kernel does not support XDP. When enabling XDP without this change Cilium enters a crashloop with this error:

```
Failed to compile XDP program" error="program cil_xdp_entry: attaching XDP program to interface ens785np0.3: operation not supported" subsys=datapath-loader
```

With this change applied and by specifying `xdp-ignore-device-name-regex` in the Cilium config we now get XDP enabled on the device and `tc` for the vlan device:

```
$ sudo bpftool net
xdp:
ens785np0(2) driver id 22899

tc:
ens785np0(2) clsact/ingress cil_from_netdev-ens785np0 id 23063
ens785np0(2) clsact/egress cil_to_netdev-ens785np0 id 23077
cilium_net(4) clsact/ingress cil_to_host-cilium_net id 23052
cilium_host(5) clsact/ingress cil_to_host-cilium_host id 23005
cilium_host(5) clsact/egress cil_from_host-cilium_host id 23051
cilium_vxlan(6) clsact/ingress cil_from_overlay-cilium_vxlan id 22907
cilium_vxlan(6) clsact/egress cil_to_overlay-cilium_vxlan id 22903
lxc4ecafafd804a(10) clsact/ingress cil_from_container-lxc4ecafafd804a id 23022
lxc9e4006f5d109(12) clsact/ingress cil_from_container-lxc9e4006f5d109 id 22950
lxcb6797ab24b31(14) clsact/ingress cil_from_container-lxcb6797ab24b31 id 22914
lxc22d432e999bb(16) clsact/ingress cil_from_container-lxc22d432e999bb id 22913
lxcecd9c72cbc66(18) clsact/ingress cil_from_container-lxcecd9c72cbc66 id 22983
lxc123aaba50ec5(20) clsact/ingress cil_from_container-lxc123aaba50ec5 id 22937
lxc602295b67e63(22) clsact/ingress cil_from_container-lxc602295b67e63 id 23028
lxcd5f791798257(24) clsact/ingress cil_from_container-lxcd5f791798257 id 22917
lxc3bd4bca11864(26) clsact/ingress cil_from_container-lxc3bd4bca11864 id 23017
lxc1ecd38cdf625(28) clsact/ingress cil_from_container-lxc1ecd38cdf625 id 23039
lxcb36787663030(30) clsact/ingress cil_from_container-lxcb36787663030 id 22922
ens785np0.3(57) clsact/ingress cil_from_netdev-ens785np0.3 id 23090
ens785np0.3(57) clsact/egress cil_to_netdev-ens785np0.3 id 23103
lxc_health(79) clsact/ingress cil_from_container-lxc_health id 23008

flow_dissector:
```

@borkmann indicated in https://github.com/cilium/cilium/issues/24768#issuecomment-1568942302 that a regex to filter out a device name would be acceptable.

Happy to make any adjustments and I am looking forward to getting this merged!

Note: I have not yet added support for this option to the Helm chart. Happy to do so in this PR or in a follow up if desired. I am holding off on it just to ensure that the variable names I chose are acceptable.

Fixes: #24768

```release-note
Adds ability to filter out device names via regex when enabling XDP
```

### DIFF
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -1543,6 +1543,13 @@ func initEnv(vp *viper.Viper) {
 			)
 		}
 	}
+
+	// Ensure the xdp-ignore-device-regex compiles
+	if len(option.Config.XDPIgnoreDeviceNameRegex) > 0 {
+		if _, err := regexp.Compile(option.Config.XDPIgnoreDeviceNameRegex); err != nil {
+			log.WithError(err).Fatal("Invalid xdp-ignore-device-regex regular expression")
+		}
+	}
 }
 
 func (d *Daemon) initKVStore() {
--- a/pkg/datapath/loader/base.go
+++ b/pkg/datapath/loader/base.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/vishvananda/netlink"
@@ -239,6 +240,9 @@ func (l *Loader) reinitializeXDPLocked(ctx context.Context, extraCArgs []string)
 	if option.Config.XDPMode == option.XDPModeDisabled {
 		return nil
 	}
+
+	rgxIgnoredXDPDevices := regexp.MustCompile(option.Config.XDPIgnoreDeviceNameRegex)
+
 	for _, dev := range option.Config.GetDevices() {
 		// When WG & encrypt-node are on, the devices include cilium_wg0 to attach bpf_host
 		// so that NodePort's rev-{S,D}NAT translations happens for a reply from the remote node.
@@ -247,6 +251,16 @@ func (l *Loader) reinitializeXDPLocked(ctx context.Context, extraCArgs []string)
 		if dev == wgTypes.IfaceName {
 			continue
 		}
+
+		// Don't enable XDP on links that match the option.Config.XDPIgnoreDeviceNameRegex regex
+		// This is useful for when you want XDP but a required device driver on the system does
+		// not yet have support for XDP in the kernel.
+		if len(option.Config.XDPIgnoreDeviceNameRegex) > 0 {
+			if rgxIgnoredXDPDevices.MatchString(dev) {
+				continue
+			}
+		}
+
 		if err := compileAndLoadXDPProg(ctx, dev, option.Config.XDPMode, extraCArgs); err != nil {
 			return err
 		}
--- a/install/kubernetes/cilium/templates/cilium-configmap.yaml
+++ b/install/kubernetes/cilium/templates/cilium-configmap.yaml
@@ -732,6 +732,9 @@ data:
 {{- if hasKey .Values.loadBalancer "serviceTopology" }}
   enable-service-topology: {{ .Values.loadBalancer.serviceTopology | quote }}
 {{- end }}
+{{- if hasKey .Values.loadBalancer "xdpIgnoreDeviceRegex" }}
+  xdp-ignore-device-regex: {{ .Values.loadBalancer.xdpIgnoreDeviceRegex | quote }}
+{{- end }}
 
 {{- end }}
 {{- if hasKey .Values.maglev "tableSize" }}
--- a/install/kubernetes/cilium/values.yaml
+++ b/install/kubernetes/cilium/values.yaml
@@ -1737,6 +1737,11 @@ loadBalancer:
   # e.g. native, disabled
   # acceleration: disabled
 
+  # -- xdpIgnoreDeviceRegex is the option to selectively ignore devices that would
+  # otherwise have xdp enabled on them. Useful if you have devices that do not yet
+  # have driver support for XDP
+  # xdpIgnoreDeviceRegex:
+
   # -- dsrDispatch configures whether IP option or IPIP encapsulation is
   # used to pass a service IP and port to remote backend
   # dsrDispatch: opt
--- a/install/kubernetes/cilium/values.yaml.tmpl
+++ b/install/kubernetes/cilium/values.yaml.tmpl
@@ -1734,6 +1734,11 @@ loadBalancer:
   # e.g. native, disabled
   # acceleration: disabled
 
+  # -- xdpIgnoreDeviceRegex is the option to selectively ignore devices that would
+  # otherwise have xdp enabled on them. Useful if you have devices that do not yet
+  # have driver support for XDP
+  # xdpIgnoreDeviceRegex:
+
   # -- dsrDispatch configures whether IP option or IPIP encapsulation is
   # used to pass a service IP and port to remote backend
   # dsrDispatch: opt
--- a/pkg/datapath/loader/netlink.go
+++ b/pkg/datapath/loader/netlink.go
@@ -191,7 +191,7 @@ func attachProgram(link netlink.Link, prog *ebpf.Program, progName string, qdisc
 	}
 
 	if err := netlink.FilterReplace(filter); err != nil {
-		return fmt.Errorf("replacing tc filter: %w", err)
+		return fmt.Errorf("replacing tc filter for interface %s: %w", link.Attrs().Name, err)
 	}
 
 	return nil
--- a/pkg/option/config.go
+++ b/pkg/option/config.go
@@ -154,6 +154,9 @@ const (
 	// EncryptNode enables node IP encryption
 	EncryptNode = "encrypt-node"
 
+	// XDPIgnoreDeviceNameRegex allows devices to be skipped when XDP is enabled
+	XDPIgnoreDeviceNameRegex = "xdp-ignore-device-regex"
+
 	// EnvoyLog sets the path to a separate Envoy log file, if any
 	EnvoyLog = "envoy-log"
 
@@ -1438,6 +1441,9 @@ type DaemonConfig struct {
 	HostV6Addr          net.IP       // Host v6 address of the snooping device
 	EncryptInterface    []string     // Set of network facing interface to encrypt over
 	EncryptNode         bool         // Set to true for encrypting node IP traffic
+
+	// Set to ignore devices when enabling XDP
+	XDPIgnoreDeviceNameRegex string
 
 	// If set to true the daemon will detect new and deleted datapath devices
 	// at runtime and reconfigure the datapath to load programs onto the new
@@ -3092,6 +3098,7 @@ func (c *DaemonConfig) Populate(vp *viper.Viper) {
 	c.EnableLocalRedirectPolicy = vp.GetBool(EnableLocalRedirectPolicy)
 	c.EncryptInterface = vp.GetStringSlice(EncryptInterface)
 	c.EncryptNode = vp.GetBool(EncryptNode)
+	c.XDPIgnoreDeviceNameRegex = vp.GetString(XDPIgnoreDeviceNameRegex)
 	c.EnvoyLogPath = vp.GetString(EnvoyLog)
 	c.HTTPNormalizePath = vp.GetBool(HTTPNormalizePath)
 	c.HTTPIdleTimeout = vp.GetInt(HTTPIdleTimeout)


squeed:
Out of curiosity, should the vlan devices be completely ignored, or just for the sake of XDP?

pkg/datapath/loader/base.go
```diff
@@ -262,7 +263,22 @@ func (l *Loader) reinitializeXDPLocked(ctx context.Context, extraCArgs []string)
...
 	}
+
 	for _, dev := range option.Config.GetDevices() {
+
+		// Don't enable XDP on links that match the option.Config.XDPIgnoreDeviceNameRegex regex
+		// This is useful for when you want XDP but a required device driver on the system does
+		// not yet have support for XDP in the kernel for XDP in the kernel.
+		if len(option.Config.XDPIgnoreDeviceNameRegex) > 0 {
+
+			ignored_interface_re := regexp.MustCompile(option.Config.XDPIgnoreDeviceNameRegex)
```
squeed: Minor nit: can you compile this in `initEnv` in daemon_main.go? No need to persist the result, just validate the regex there.

squeed:
Question for you, @borkmann  -- given the usual XDP use-case, would an allowing regex make more sense (i.e. use XDP for devices matching the regex)? Or should we stick with an excluding regex?

Separately, are there any other specifiers that might make sense? Only interfaces with a given driver?

pkg/datapath/loader/base.go
```diff
@@ -262,7 +263,22 @@ func (l *Loader) reinitializeXDPLocked(ctx context.Context, extraCArgs []string)
...
+
+			if ignored_interface_re.MatchString(dev) {
+				return nil
```
rgo3:
```suggestion
				continue
```
As I understand we just want to skip devices that match the regex and this terminates the function at the first device that we find.

pkg/datapath/loader/base.go
poblahblahblah: Done in 13085b2

pkg/datapath/loader/base.go
poblahblahblah: Good catch! Fixed.

poblahblahblah:
> Out of curiosity, should the vlan devices be completely ignored, or just for the sake of XDP?

I'm not positive!

I think regardless of whether it's beneficial to use tc on a `vlan` type device the change would be useful for other groups who want to have XDP but have mixed NICs where one supports XDP and the other does not.

pkg/datapath/loader/base.go
```diff
@@ -262,7 +263,22 @@ func (l *Loader) reinitializeXDPLocked(ctx context.Context, extraCArgs []string)
 	if option.Config.XDPMode == option.XDPModeDisabled {
 		return nil
 	}
+
 	for _, dev := range option.Config.GetDevices() {
+
+		// Don't enable XDP on links that match the option.Config.XDPIgnoreDeviceNameRegex regex
+		// This is useful for when you want XDP but a required device driver on the system does
+		// not yet have support for XDP in the kernel for XDP in the kernel.
```
thorn3r:
```suggestion
		// not yet have support for XDP in the kernel.
```

thorn3r (approved):
changes lgtm, pending the question @squeed raised about inverting the option. Mind squashing the commits?

poblahblahblah:
Would you all like me to update the Helm chart and docs while I am in here?

poblahblahblah:
Hello, is there anything else you would like to see for this change?

poblahblahblah:
Hello!

Wondering if there's anything I can do to help move this change along.

Let me know!

daemon/cmd/daemon_main.go
```diff
@@ -1540,6 +1540,15 @@ func initEnv() {
 			)
 		}
 	}
+
+	// Ensure the xdp-ignore-device-name-regex compiles
+	if len(option.Config.XDPIgnoreDeviceNameRegex) > 0 {
+		_, err := regexp.Compile(option.Config.XDPIgnoreDeviceNameRegex)
+		if err != nil {
+			log.WithError(err).Fatal("Invalid xdp-ignore-device-name-regex regular expression")
+			return
```
rgo3: Nit: the Fatal will already return.

pkg/datapath/loader/base.go
```diff
@@ -262,7 +263,22 @@ func (l *Loader) reinitializeXDPLocked(ctx context.Context, extraCArgs []string)
...
 	}
+
 	for _, dev := range option.Config.GetDevices() {
+
+		// Don't enable XDP on links that match the option.Config.XDPIgnoreDeviceNameRegex regex
+		// This is useful for when you want XDP but a required device driver on the system does
+		// not yet have support for XDP in the kernel.
+		if len(option.Config.XDPIgnoreDeviceNameRegex) > 0 {
+
+			ignored_interface_re, _ := regexp.Compile(option.Config.XDPIgnoreDeviceNameRegex)
```
rgo3: Nit to have Golang style variables:
```suggestion
			ignored, _ := regexp.Compile(option.Config.XDPIgnoreDeviceNameRegex)
```
Also could you remove some of the blank lines in this `if` block?

rgo3 (approved):
Apologies for the wait @poblahblahblah. We should still clarify some of the questions by @squeed above, but a release is coming up so a lot of folks are stretched a bit thin. Apart from the two code style nits, this LGTM so I'll approve.

poblahblahblah:
@ldelossa rebased against main and resolved all conflicts!

ldelossa:
@poblahblahblah looks like these changes have caused the following issue:

```
Error: Unable to install Cilium: execution error at (cilium/templates/cilium-configmap.yaml:58:5): kubeproxyReplacement must be explicitly set to a valid value (disabled, partial, or strict) to continue.
```

Seen in these tests: https://github.com/cilium/cilium/actions/runs/5393251788/jobs/9794130572

Its possible the changes to configuration has caused a conflict, making Cilium bail out while parsing the config.

poblahblahblah:
@ldelossa I think that issue was introduced by https://github.com/cilium/cilium/commit/9029ce09ade11992f7a9f6c0dc7116e0d00d3e62 and fixed by https://github.com/cilium/cilium/commit/d2b63e33c8c5099b62cd45508a99aab03b1be9da

I will rebase and push to verify.

poblahblahblah:
@ldelossa 

I was on PTO last week so I didn't have a chance to check up on this. It looks like there was a single test that failed - is there a way I can trigger these locally?

poblahblahblah:
Hey @dylandreimerink,

No problem - I understand how busy we all are.

I have pushed up a branch that is now up to date with main.

🤞🤞🤞

ti-mo:
I kicked the IntegrationTests workflow due to a flake. This looks good, thanks!

poblahblahblah:
Cool! Looks like all tests are passing now!

Is there anything else I can do for this PR?

squeed (approved):
looks good! Sorry for missing the update.

(note that it will need a rebase)

poblahblahblah:
Rebased!

pkg/datapath/loader/base.go
```diff
@@ -239,14 +240,29 @@ func (l *Loader) reinitializeXDPLocked(ctx context.Context, extraCArgs []string)
...
 		if dev == wgTypes.IfaceName {
 			continue
 		}
+
+		// Don't enable XDP on links that match the option.Config.XDPIgnoreDeviceNameRegex regex
+		// This is useful for when you want XDP but a required device driver on the system does
+		// not yet have support for XDP in the kernel.
+		if len(option.Config.XDPIgnoreDeviceNameRegex) > 0 {
+
+			ignored, _ := regexp.Compile(option.Config.XDPIgnoreDeviceNameRegex)
```
ti-mo: Nit: call `regexp.MustCompile` and assign to an (unexported) package-scoped variable. Compiling regular expressions is expensive, no need to do that repeatedly.

pkg/datapath/loader/base.go
```diff
@@ -239,14 +240,29 @@ func (l *Loader) reinitializeXDPLocked(ctx context.Context, extraCArgs []string)
 	if option.Config.XDPMode == option.XDPModeDisabled {
 		return nil
 	}
+
```
ti-mo, line 243: Nit: No need for these whitespace changes.

daemon/cmd/daemon_main.go
```diff
@@ -1540,6 +1540,14 @@ func initEnv(vp *viper.Viper) {
 			)
 		}
 	}
+
+	// Ensure the xdp-ignore-device-name-regex compiles
+	if len(option.Config.XDPIgnoreDeviceNameRegex) > 0 {
+		_, err := regexp.Compile(option.Config.XDPIgnoreDeviceNameRegex)
+		if err != nil {
+			log.WithError(err).Fatal("Invalid xdp-ignore-device-name-regex regular expression")
```
ti-mo: Note: if the goal is to point the user to the corresponding Helm flag, this should say `xdp-ignore-device-regex`.

ti-mo (changes requested):
Thanks, left a few nits.

poblahblahblah:
@ti-mo updated per your recommendations

pkg/datapath/loader/base.go
```diff
@@ -239,14 +240,28 @@ func (l *Loader) reinitializeXDPLocked(ctx context.Context, extraCArgs []string)
...
+	ignored_xdp_device_regex := regexp.MustCompile(option.Config.XDPIgnoreDeviceNameRegex)
```
ti-mo: My initial comment below read:
> call `regexp.MustCompile` and assign to an (unexported) package-scoped variable.

`ignored_xdp_device_regex` is a local variable, not a package-scoped variable. `MustCompile` will still be called on every call to `reinitializeXDPLocked`. What I mean is `var rgxIgnoredXDPDevices = regexp.MustCompile(option.Config.XDPIgnoreDeviceNameRegex)` at the top of the file.

pkg/datapath/loader/base.go
```diff
@@ -239,14 +240,28 @@ func (l *Loader) reinitializeXDPLocked(ctx context.Context, extraCArgs []string)
...
+
 	for _, dev := range option.Config.GetDevices() {
+
```
ti-mo: Remove this.

ti-mo (changes requested):

pkg/datapath/loader/base.go
poblahblahblah: This didn't end up working correctly. When the regex is compiled as part of assigning it to a var it does not look like `option.Config.XDPIgnoreDeviceNameRegex` has been initialized yet, so the compiled regex is effectively an empty regex. The consequence of this is that all devices match the regex and none get configured for XDP.

I'm not sure what a path forward would be since I have much less experience in the code base than you all. Any guidance here would be _very_ appreciated!

ti-mo: Fair enough, this function isn't called often enough for the alternatives to be worth it (e.g. putting it behind a sync.Once). Please do adjust the variable name, though. I recommend giving the document at https://go.dev/doc/effective_go#mixed-caps a read just in case!

daemon/cmd/daemon_main.go
```diff
@@ -1540,6 +1540,14 @@ func initEnv(vp *viper.Viper) {
 			)
 		}
 	}
+
+	// Ensure the xdp-ignore-device-regex compiles
+	if len(option.Config.XDPIgnoreDeviceNameRegex) > 0 {
+		_, err := regexp.Compile(option.Config.XDPIgnoreDeviceNameRegex)
+		if err != nil {
```
ti-mo: Tip: this can be folded like `if _, err := regexp.Compile(option.Config.XDPIgnoreDeviceNameRegex); err != nil { ... }`

pkg/datapath/loader/base.go
poblahblahblah: Will do - appreciate the doc link :)

daemon/cmd/daemon_main.go
poblahblahblah: TIL - thank you!

ti-mo (approved):
Looks good! Thanks for addressing all of my feedback. :pray:

daemon/cmd/daemon_main.go
```diff
@@ -1543,6 +1543,13 @@ func initEnv(vp *viper.Viper) {
 			)
 		}
 	}
+
+	// Ensure the xdp-ignore-device-regex compiles
+	if len(option.Config.XDPIgnoreDeviceNameRegex) > 0 {
```
joestringer, line 1548: Was this question from @squeed ever discussed? Scanning through I could see a few folks saying there should be a discussion but I couldn't see a clear resolution to the question:

> Question for you, @borkmann -- given the usual XDP use-case, would an allowing regex make more sense (i.e. use XDP for devices matching the regex)? Or should we stick with an excluding regex?
> 
> Separately, are there any other specifiers that might make sense? Only interfaces with a given driver?

poblahblahblah:
I would love to see this MR merged so we can enable XDP in our environment!

Is there anything I can do to help move this along further?

daemon/cmd/daemon_main.go
borkmann: Sorry for the huge delay on my side. First release, and then out for long PTO, it slipped my radar.

Just thinking some more on this topic, @poblahblahblah / @joestringer / @joamaki, what if instead of regex we just had a simple bool option where we say "if XDP attach fails on this device, then just fallback to tc BPF". Would this be more generic and easier to use rather than specifying device regex? Something like `--bpf-lb-acceleration-fallback=true    [ In case BPF load balancing acceleration via native XDP is not supported for some devices, fall back to tc BPF attachment ]`. Other option could be to have a new mode `--bpf-lb-acceleration=best-effort` ?

ti-mo: Makes sense to me, no need for the config flag even IMO. If the device is expected to be managed, we should make it so.

borkmann: You mean just to fallback in case of failure with the `native` option? I think this perhaps could lead to false conclusions, meaning people will try to active, then some ethtool driver channel setting is not correct and it'll fail and people just assume XDP would be there now but actually we run in tc BPF mode. So I would bail out by default and let users make a conscious  decision to fallback upon failure. Also what would help as well in this context would be to let `cilium status --verbose` dump the devices which have Cilium progs on XDP so that this is easily visible for troubleshooting / and sysdump.

poblahblahblah: So you are proposing the following:

* Keep the default behavior of bailing out in place
* Implement `--bpf-lb-acceleration=best-effort` which if set would have devices fall back to TC

If I updated this PR to do the above, would the PR be considered mergeable?

borkmann: > If I updated this PR to do the above, would the PR be considered mergeable?

Yes, definitely - if code looks good then I consider it mergable.

ti-mo:
@lmb Small note: we can convert PRs to drafts if they need a course change.

@poblahblahblah Please mark ready for review if this is ready for another round.

daemon/cmd/daemon_main.go
poblahblahblah: @borkmann Sorry for the delay. I made a first pass at this change here: https://github.com/cilium/cilium/pull/28666

poblahblahblah:
Closing in favor of https://github.com/cilium/cilium/pull/28666

PR #25873, opened 2023-06-02
Makefile: Fix kind deployment in quiet mode

joestringer:
The "$(QUIET)" directive is a Makefile syntax converting to something
like "@" when env V=0, so as a Makefile directive it must be at the
beginning of the line, or "make kind" will fail. Fix it.

Fixes: cb2b1124e57d ("Set SED binary from Makefile in kind.sh")


joestringer:
Heads up @chancez just FYI

chancez (approved):

borkmann (approved):

PR #25881, opened 2023-06-05
bpf: minor HostFW cleanups

julianwiedmann:
Some innocent cleanups from poking around in the HostFW paths.

### DIFF
--- a/bpf/bpf_host.c
+++ b/bpf/bpf_host.c
@@ -139,15 +139,15 @@ handle_ipv6(struct __ctx_buff *ctx, __u32 secctx __maybe_unused,
 	    const bool from_host __maybe_unused,
 	    __s8 *ext_err __maybe_unused)
 {
-	struct ct_buffer6 __maybe_unused ct_buffer = {};
+#ifdef ENABLE_HOST_FIREWALL
+	struct ct_buffer6 ct_buffer = {};
+	bool need_hostfw = false;
+#endif /* ENABLE_HOST_FIREWALL */
 	void *data, *data_end;
 	struct ipv6hdr *ip6;
 	int __maybe_unused ret;
 	int hdrlen;
 	__u8 nexthdr;
-#ifdef ENABLE_HOST_FIREWALL
-	bool need_hostfw = false;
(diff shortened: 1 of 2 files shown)


bpf/bpf_host.c
```diff
@@ -236,16 +233,15 @@ handle_ipv6_cont(struct __ctx_buff *ctx, __u32 secctx, const bool from_host,
 	struct endpoint_info *ep;
 	int ret;
 
-#ifdef ENABLE_HOST_FIREWALL
-	from_host_raw = ctx_load_meta(ctx, CB_FROM_HOST);
-	ctx_store_meta(ctx, CB_FROM_HOST, 0);
-#endif /* ENABLE_HOST_FIREWALL */
-
```
gentoo-root, line 243: I believe we want this block to be above any potential early return, so that we don't leak any non-zero metadata.

bpf/bpf_host.c
```diff
@@ -621,23 +617,21 @@ handle_ipv4_cont(struct __ctx_buff *ctx, __u32 secctx, const bool from_host,
...
 	struct iphdr *ip4;
 	struct remote_endpoint_info *info;
 	struct endpoint_info *ep;
 	int ret;
 
-#ifdef ENABLE_HOST_FIREWALL
-	from_host_raw = ctx_load_meta(ctx, CB_FROM_HOST);
-	ctx_store_meta(ctx, CB_FROM_HOST, 0);
-#endif /* ENABLE_HOST_FIREWALL */
-
```
gentoo-root, line 635: Same here.

gentoo-root (changes requested):

bpf/bpf_host.c
julianwiedmann: But the only preceding return is a drop?

gentoo-root: Well, yes, I guess it should be safe then.

I was under impression we try to clear metadata the first thing in the function for extra robustness, but it doesn't look problematic in this case.

gentoo-root (approved):

PR #24716, opened 2023-04-03
Pass error codes from map_update_elem (ct_create) to drop notifications

gentoo-root:
```release-note
Report the kernel error code in case of packet drops due to failures to create conntrack map entries.
```

PR #25883, opened 2023-06-05
bpf: nat: report detailed error from map_update_elem() for NAT entries

julianwiedmann:
Continuing from https://github.com/cilium/cilium/pull/24716, also report an `ext_err` when inserting entries into the NAT map fails.

```release-note
Report the kernel error code in case of packet drops due to failures to create NAT map entries.
```

### DIFF
--- a/bpf/lib/nat.h
+++ b/bpf/lib/nat.h
@@ -678,7 +678,7 @@ snat_v4_rev_nat_can_skip(const struct ipv4_nat_target *target, const struct ipv4
  */
 static __always_inline __maybe_unused int
 snat_v4_create_dsr(const struct ipv4_ct_tuple *tuple,
-		   __be32 to_saddr, __be16 to_sport)
+		   __be32 to_saddr, __be16 to_sport, __s8 *ext_err)
 {
 	struct ipv4_ct_tuple tmp = *tuple;
 	struct ipv4_nat_entry state = {};
@@ -695,8 +695,10 @@ snat_v4_create_dsr(const struct ipv4_ct_tuple *tuple,
 	state.to_sport = to_sport;
 
 	ret = map_update_elem(&SNAT_MAPPING_IPV4, &tmp, &state, 0);
-	if (ret)
-		return ret;
+	if (ret) {
+		*ext_err = (__s8)ret;
+		return DROP_NAT_NO_MAPPING;
+	}
 
 	return CTX_ACT_OK;
 }
@@ -1648,7 +1650,7 @@ snat_v6_rev_nat_can_skip(const struct ipv6_nat_target *target, const struct ipv6
 
 static __always_inline __maybe_unused int
 snat_v6_create_dsr(const struct ipv6_ct_tuple *tuple, union v6addr *to_saddr,
-		   __be16 to_sport)
+		   __be16 to_sport, __s8 *ext_err)
 {
 	struct ipv6_ct_tuple tmp = *tuple;
 	struct ipv6_nat_entry state = {};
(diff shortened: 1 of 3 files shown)


bpf/lib/nat.h
```diff
@@ -250,7 +250,7 @@ static __always_inline int snat_v4_new_mapping(struct __ctx_buff *ctx,
 					       const struct ipv4_nat_target *target,
 					       bool needs_ct)
 {
-	int ret = DROP_NAT_NO_MAPPING, retries;
```
gentoo-root, line 202: I believe this change is wrong.

> If the very first lookup in the loop succeeds, we never update `ret` and
> it stays initialized to the initial value (!= 0).

If it succeeds, we set `ret = __snat_update(map, otuple, ostate, &rtuple, &rstate);` and break only on `ret == 0`, therefore we return 0, as we should.

On the other hand, if we exceed the number of retries, we should return DROP_NAT_NO_MAPPING, but with your change we would return 0.

gentoo-root (changes requested):

bpf/lib/nat.h
julianwiedmann: On success, `__snat_lookup()` returns the map entry. So `!__snat_lookup()` is false, and we don't step into the branch that touches `ret`.

gentoo-root: Ah right, sorry, I got confused a bit. But I still don't get what the issue is that you are fixing.

If __snat_lookup returns non-NULL, on the first iteration, we change dport and try again. If it returns NULL on the second iteration, we set ret. If it doesn't return NULL on any iteration, we should return DROP_NAT_NO_MAPPING, not 0.

bpf/lib/nat.h
```diff
@@ -248,7 +248,14 @@ static __always_inline int snat_v4_new_mapping(struct __ctx_buff *ctx,
 
 	if (retries > SNAT_SIGNAL_THRES)
 		send_signal_nat_fill_up(ctx, SIGNAL_PROTO_V4);
-	return !ret ? 0 : DROP_NAT_NO_MAPPING;
+
+	if (ret < 0) {
+		if (ext_err)
+			*ext_err = (__s8)ret;
```
gentoo-root, line 254: ret could still be DROP_NAT_NO_MAPPING at this point, which doesn't make much sense as ext_err. On the other hand, ext_err < -128 is currently replaced by 0 in __DROP_REASON_EXT, so maybe it's not worth changing anything here.

gentoo-root (approved):

bpf/lib/nat.h
julianwiedmann: > If __snat_lookup returns non-NULL, on the first iteration, we change dport and try again. If it returns NULL on the second iteration, we set ret. If it doesn't return NULL on any iteration, we should return DROP_NAT_NO_MAPPING, not 0.

Yep, agreed - this wouldn't fix anything. I must have convinced myself that we would bail out when finding an entry, but logic-wise that wouldn't make any sense either. So let's drop this part for now, I'll try to circle back during the 1.15 cycle and clean this up a bit. Thanks for keeping me honest, Maxim!

And sorry for not getting back to this earlier :(.

bpf/lib/nat.h
julianwiedmann: > ret could still be DROP_NAT_NO_MAPPING at this point, which doesn't make much sense as ext_err. On the other hand, ext_err < -128 is currently replaced by 0 in __DROP_REASON_EXT, so maybe it's not worth changing anything here.

Yep, it's a bit of an annoyance. But imho the `ext_err` always needs closer inspection/interpretation, based on where it comes from. So I think that's ok for now, and we can clean it up a bit more at a later time.

Thank you for the review, as always!

julianwiedmann:
The test-runtime fail was https://github.com/cilium/cilium/issues/25939, but it's no longer required :tada:.

PR #25887, opened 2023-06-05
node: Replace node managers dynamic metrics with modular metrics

dylandreimerink:
The node manager does dynamic metrics registration and cleanup. This PR replaces it with modular metrics to remove the usage of the global registry functions.

```release-note
converted node manager dynamic metrics into modular metrics
```


joamaki (approved):
LGTM and confirmed this fixes the slow test/controlplane (back down to 8s versus 85s)

tklauser (approved):

PR #25904, opened 2023-06-05
build: Avoid cross compilation issue on Windows

sayboras:
Happens in cilium-cli PR when trying to import the latest cilium/cilium main hash.

```
Error: vendor/github.com/cilium/cilium/pkg/hive/hive.go:202:44: undefined: unix.SIGINT
Error: vendor/github.com/cilium/cilium/pkg/hive/hive.go:202:57: undefined: unix.SIGTERM
```

### DIFF
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -10,6 +10,7 @@ import (
 	"os/signal"
 	"reflect"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/sirupsen/logrus"
@@ -17,7 +18,6 @@ import (
 	"github.com/spf13/viper"
 	"go.uber.org/dig"
 	"go.uber.org/multierr"
-	"golang.org/x/sys/unix"
 
 	"github.com/cilium/cilium/pkg/hive/cell"
 	"github.com/cilium/cilium/pkg/logging"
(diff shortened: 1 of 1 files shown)


pkg/hive/hive.go
```diff
@@ -213,7 +213,8 @@ func (h *Hive) Run() error {
 func (h *Hive) waitForSignalOrShutdown() error {
 	signals := make(chan os.Signal, 1)
 	defer signal.Stop(signals)
-	signal.Notify(signals, os.Interrupt, unix.SIGINT, unix.SIGTERM)
+	// equivalent to unix.SIGINT and unix.SIGTERM, but works on cross compilation for Windows
+	signal.Notify(signals, os.Interrupt, syscall.Signal(0x2), syscall.Signal(0xf))
```
joamaki: The `os.Interrupt` is the same as `unix.SIGINT` so `SIGINT` can be removed. Let's not use magic constants here, but use instead `syscall.SIGTERM`. It's also defined for Windows: https://github.com/golang/go/blob/master/src/syscall/types_windows.go#LL51C9-L51C9, but likely does nothing.

joamaki (changes requested):

pkg/hive/hive.go
sayboras: Thanks for your review, updated as per the suggestion :heavy_check_mark:

rgo3 (approved):

joamaki (approved):

PR #25910, opened 2023-06-05
CRD List Generation

### LINKED ISSUE #25750: docs: Automate the listing of CRDs
In the Cilium Operator document, we list all CRDs registered by the Operator. Currently, we need to update this list manually and it's very easy to miss the update when we add a new CRD. We should automate the process.

Background: https://github.com/cilium/cilium/pull/25696#discussion_r1206112326

Task list
- [ ] Move list of CRDs to a dedicated RST fragment, that we include from `cilium_operator.rst`
- [ ] Generate the fragment from a dedicated Documentation/Makefile target
- [ ] If we have any other document that using the list of CRDs, update it

Assigning myself for the moment to guarantee someone eventually takes care of it, but feel free to take if you want to work for this.

dhawton:

      GitHub issue.

Replaces a manually edited list of CRDs with an automatically generated and included RST.

Running `make generate-crd-docs` in the root project or `make update-crdlist` in Documentation will run the crdlistgen tool. This tool rebuilds the Documentation/crdlist.rst file by looking at the pkg/k8s/apis/cilium.io/client package's CRD List. It will also scan the rst files of Documentation to see if the CRD is being used as a header anywhere to generate a link to it.

Fixes: #25750

### DIFF
--- a/Documentation/Makefile
+++ b/Documentation/Makefile
@@ -56,6 +56,11 @@ update-cmdref: builder-image cilium-build ## Update the command reference docume
 	-$(QUIET)rm -rf cmdref/cilium*.md
 	$(QUIET)$(DOCKER_RUN) ./update-cmdref.sh
 
+.PHONY: update-crdlist
+update-crdlist:
+	@$(ECHO_GEN)crdlist
+	make -C ../ generate-crd-docs
+
 codeowners.rst: $(ROOT_DIR)/CODEOWNERS
 	@$(ECHO_GEN)$@
 	$(QUIET)$(DOCKER_RUN) ./update-codeowners.sh
@@ -63,7 +68,7 @@ codeowners.rst: $(ROOT_DIR)/CODEOWNERS
 .PHONY: update-codeowners
 update-codeowners: codeowners.rst
 
-check: builder-image api-flaggen update-cmdref update-helm-values update-codeowners ## Validate command and Helm references, as well as policy examples.
+check: builder-image api-flaggen update-cmdref update-crdlist update-helm-values update-codeowners ## Validate command and Helm references, as well as policy examples.
 	@$(ECHO_CHECK) cmdref
 	$(QUIET) ./check-cmdref.sh
 	@$(ECHO_CHECK) $(HELM_VALUES)
@@ -74,6 +79,8 @@ check: builder-image api-flaggen update-cmdref update-helm-values update-codeown
 	$(QUIET) ./check-codeowners.sh
 	@$(ECHO_CHECK) configuration/api-restrictions-table.rst
 	$(QUIET) ./check-flaggen.sh
+	@$(ECHO_CHECK) crdlist.rst
+	$(QUIET) ./check-crdlist.sh
 
 ifeq ($(V),0)
 SPHINX_OPTS += -q
--- a/pkg/k8s/apis/cilium.io/client/register.go
+++ b/pkg/k8s/apis/cilium.io/client/register.go
@@ -93,40 +93,100 @@ var (
 	comparableCRDSchemaVersion = versioncheck.MustVersion(k8sconst.CustomResourceDefinitionSchemaVersion)
 )
 
-type crdCreationFn func(clientset apiextensionsclient.Interface) error
+type CRDList struct {
+	Name     string
+	FullName string
+}
+
+// Returns a map of CRDs
+func CustomResourceDefinitionList() map[string]*CRDList {
+	return map[string]*CRDList{
+		synced.CRDResourceName(k8sconstv2.CNPName): {
+			Name:     CNPCRDName,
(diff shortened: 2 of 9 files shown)


Documentation/Makefile
```diff
@@ -56,14 +56,19 @@ update-cmdref: builder-image cilium-build ## Update the command reference docume
...
+
 codeowners.rst: $(ROOT_DIR)/CODEOWNERS
 	@$(ECHO_GEN)$@
 	$(QUIET)$(DOCKER_RUN) ./update-codeowners.sh
 
 .PHONY: update-codeowners
 update-codeowners: codeowners.rst
 
-check: builder-image api-flaggen update-cmdref update-helm-values update-codeowners ## Validate command and Helm references, as well as policy examples.
+check: builder-image api-flaggen update-cmdref update-crdlist update-helm-values update-codeowners ## Validate command and Helm references, as well as policy examples.
```
qmonnet, line 71: Here we're adding the update as a dependency for target `check`, but we would also need to add a test below to verify in CI that the update was performed. Would you mind adding something similar to the other `check-*` in the recipe below, please?

qmonnet (changes requested):
Preview at https://deploy-preview-25910--docs-cilium-io.netlify.app/internals/cilium_operator.html#crd-registration: it works great, thanks for this!

There's the CI check missing (see below), but looks good otherwise.

Documentation/Makefile
dhawton: Done

youngnick (approved):

pkg/k8s/apis/cilium.io/client/register.go
```diff
@@ -90,39 +90,121 @@ var (
...
-type crdCreationFn func(clientset apiextensionsclient.Interface) error
+type CRDList struct {
+	ResourceName string
+	Name         string
+	FullName     string
+}
+
+// Returns a map of CRDs
+func CustomResourceDefinitionList() []*CRDList {
+	return []*CRDList{
```
tklauser: Couldn't we keep the CRDs in a map (i.e. `resourceToCreateFnMapping`) instead of converting it to a slice and use that in `CreateCustomResourceDefinitions`? Otherwise lookup in that func becomes `O(n^2)` instead of `O(n*log(n))`, i.e. for each entry of `synced.AllCiliumCRDResourceNames()` we'd search this list linearly.

For use in `tools/crdlistgen` we could instead lift out `resourceToCreateFnMapping` of `CreateCustomResourceDefinitions` and then let this func convert the map to a slice and sort it?

tklauser (changes requested):

tklauser:
PR #25824 will introduce another CRD (`CiliumPodIPPool`), so we'll need to coordinate in which order the two PRs are merged.

pkg/k8s/apis/cilium.io/client/register.go
dhawton: I was trying to not create another slice somewhere that had to be maintained, and using the resourceToCreateFnMapping indexes would result in a fully lowercase list of CRDs so wouldn't be doc friendly.

tklauser: I see. How about introducing `CustomResourceDefinitionMap` that would keep this info in a map and that. `CreateCustomResourceDefinitions` would use and then let`CustomResourceDefinitionList` (which AFAICT is only used in `tools/crdlistgen`) call `CustomResourceDefinitionMap`, convert it to a slice and sort it?

dhawton: I have converted it back to a map. The tool doesn't need it as a slice, I just need access to the `Name` field for the docs which contains the "pretty" format of it.

qmonnet (commented):
Thanks for adding the check!

I'm thinking of additional details now, apologies for missing them on the first pass.

- Can we mark `Documentation/crdlist.rst` as a generated file to GitHub, so that changes are folded by default in the diffs? We need a new entry for the file in `.gitattributes` for that, just like for `helm-values.rst` for example.

- Can we also remove ownership for this generated file in `CODEOWNERS`, to avoid team docs-structure to be pulled for review each time the list changes? Again, see the example of `helm-values.rst` in `CODEOWNERS`. Given that CI should catch missing updates, we shouldn't need manual validation for this file.

qmonnet:
Please, also rebase your PR on `main` to address the merge conflict with `pkg/k8s/apis/cilium.io/client/register.go`.

qmonnet (approved):
I just had the time to observe that the CI check works as expected :) Thanks a lot!

tklauser (approved):
Thanks!

qmonnet:
/test-1.25-4.19

qmonnet:
/test-1.25-4.19

qmonnet (approved):

PR #25912, opened 2023-06-05
Module health node manager

tommyp1ckles:
This PR seeks to provide modular health status checking on the "node manager" component of the Agent. Specifically the goals are:

### Bubble-up all errors in Linux node handler code

Node Manager related code relies heavily on logging (or just ignoring) errors far down the stack, instead of bubbling them up.

The result of this is:
* Logged errors have little context on the root event causing them to be invoked.
* It's hard to trace failure from a particular event relating to all reconcile procedures that need to run.
* It's hard to definitively to say if reconciliation for a Node Event was successfully.

The changes in this PR seek to address this by handling and bubbling up all relevant errors that may be encountered while doing Linux Node update reconciliation logic.

As well, this PR seeks to **avoid** any changes to the control-flow logic of the reconciliation by _not_ immediately returning at places where errors may occur.
Rather, this follows a pattern of combining lists of errors into relevant/scoped multi errors (using the multierr library). All code will execute as before, with the difference being that errors are handled up the stack.

### Use Hive's Modular Health Reporter to report status of node manager.

This is a first attempt at trying to manage health status for a module. 

#### Follow up work:
* Investigate Reconciliation: Break Linux node logic into atomic and idempotent tasks that can be executed and retried (+ reported on more easily).

### DIFF
--- a/pkg/datapath/linux/ipsec/ipsec_linux.go
+++ b/pkg/datapath/linux/ipsec/ipsec_linux.go
@@ -9,6 +9,7 @@ import (
 	"bufio"
 	"context"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -517,7 +518,8 @@ func ipSecReplacePolicyOut(src, dst *net.IPNet, tmplSrc, tmplDst net.IP, nodeID 
 	return netlink.XfrmPolicyUpdate(policy)
 }
 
-func ipsecDeleteXfrmState(nodeID uint16) {
+func ipsecDeleteXfrmState(nodeID uint16) error {
+	var acc error
 	scopedLog := log.WithFields(logrus.Fields{
 		logfields.NodeID: nodeID,
 	})
@@ -525,18 +527,20 @@ func ipsecDeleteXfrmState(nodeID uint16) {
 	xfrmStateList, err := netlink.XfrmStateList(netlink.FAMILY_ALL)
 	if err != nil {
 		scopedLog.WithError(err).Warning("Failed to list XFRM states for deletion")
-		return
+		return acc
 	}
 	for _, s := range xfrmStateList {
 		if getNodeIDFromXfrmMark(s.Mark) == nodeID {
 			if err := netlink.XfrmStateDel(&s); err != nil {
 				scopedLog.WithError(err).Warning("Failed to delete XFRM state")
+				acc = errors.Join(acc, fmt.Errorf("delete xfrm state (%s): %w", s.String(), err))
 			}
 		}
 	}
+	return acc
 }
 
-func ipsecDeleteXfrmPolicy(nodeID uint16) {
+func ipsecDeleteXfrmPolicy(nodeID uint16) error {
 	scopedLog := log.WithFields(logrus.Fields{
 		logfields.NodeID: nodeID,
 	})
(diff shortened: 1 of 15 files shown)


pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -517,15 +519,16 @@ func ipSecReplacePolicyOut(src, dst *net.IPNet, tmplSrc, tmplDst net.IP, nodeID
 	return netlink.XfrmPolicyUpdate(policy)
 }
 
-func ipsecDeleteXfrmState(nodeID uint16) {
+func ipsecDeleteXfrmState(nodeID uint16) error {
+	var acc error
```
ti-mo, line 522: This variable is never assigned to?

pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -21,10 +21,12 @@ import (
 	"github.com/fsnotify/fsnotify"
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
+	"go.uber.org/multierr"
```
ti-mo: Go 1.20 introduced `errors.Join()`, maybe use that instead?

pkg/errs/__debug_bin
ti-mo, line 1: What is this file for?

ti-mo (changes requested):
Right off the bat I have some concerns around the pervasive use of multierr and the introduction of `pkg/errs`, so I'm flagging this early before reviewing the whole thing. Perhaps the changes to errors can be isolated from the modularity work? That will make the PR more manageable to review.

If we're going to deviate from the canonical Go error handling style, I'd like this to be discussed in the open, e.g. through a CFP, and make it part of a larger plan to convert the whole codebase. We need strong reasons for doing so, and it's not a decision we should take lightly, as it makes code look very different due to how often a function like `errs.Into` appears. It directly affects the readablilty and familiarity of the code, so personally I'd stick to the idiomatic Go style.

I think errors.Join() could be used in the majority of cases here, which is not as big of a departure as multierr or a new package.

Another observation: I'm seeing a few occurrences of `acc` (what's in a name, btw?) declared and assigned to, but never returned. IMO the pattern of declaring a separate `acc` (instead of an `err` that is `errors.Join()`ed into) is a code smell, as it splinters the source of truth. `errors.Join` doesn't really have this problem; it supports nil values and the resulting error value will remain nil if the inputs are all nil.

pkg/hubble/peer/handler.go
```diff
@@ -37,6 +37,10 @@ func newHandler(withoutTLSInfo bool, addressPref serviceoption.AddressFamilyPref
 	}
 }
 
+func (h *handler) Name() string {
+	return "hubble-peers"
```
chancez: Nit: should this be singular? Package is singular, and we refer to it as the "hubble peer service" elsewhere.

tommyp1ckles:
@ti-mo 

> Right off the bat I have some concerns around the pervasive use of multierr and the introduction of pkg/errs, so I'm flagging this early before reviewing the whole thing.

Fair enough,  the motivation for using the multierr was that errors.Join wraps the error, whereas multierr acts more like a list of errors. But the decision to use the uber multierr I think was a bit hasty.

> Perhaps the changes to errors can be isolated from the modularity work? That will make the PR more manageable to review.

Can I propose that for the purposes of this change, I use errors.Join (or if necessary a list of errors) instead of the multierr.

I agree with bringing the error handling out into an open discussion, I've been compiling my thoughts on the matter which I was planning to make into a proposal - this can be part of it as well.

pkg/hubble/peer/handler.go
chancez: 🤔 did this change?

tommyp1ckles:
@ti-mo 

Updated to use errors.Join instead of multierr. If we need something with more functionality I'll start a discussion/proposal for that.

tommyp1ckles:
I also simplified the error status handling logic, dropping the `pkg/errs` changes altogether. . I think its better we start with a simplified implementation and iterate on that. Tracking backgroundSync seems sufficient as it happens every minute and will effectively retry updates on nodes in the manager.

I changed deleteNode to just log for now. I don't think the nodeDelete is quite ready to be incorporated into the status check. I'll follow up with a better approach to this in a separate PR.

tommyp1ckles:
Tried to close and reopen to fix issues with seeing my force-push but now I can't reopen. PR is now over at [25994](https://github.com/cilium/cilium/pull/25994)

Issue #25156, opened 2023-04-27
migrate main branch workflows to use cilium-cli with helm installation mode


michi-covalent:
we can start migrating main branch workflows to use cilium-cli with helm installation mode. these workflows need to be migrated:

- [x] workflows/conformance-aks.yaml https://github.com/cilium/cilium/pull/26382
- [x] workflows/conformance-aws-cni.yaml https://github.com/cilium/cilium/pull/25820
- [x] workflows/conformance-clustermesh.yaml https://github.com/cilium/cilium/pull/25834
- [x] workflows/conformance-eks.yaml https://github.com/cilium/cilium/pull/25820
- [x] workflows/conformance-externalworkloads.yaml https://github.com/cilium/cilium/pull/26382
- [x] workflows/conformance-gateway-api.yaml @sayboras https://github.com/cilium/cilium/pull/25608
- [x] workflows/conformance-gke.yaml https://github.com/cilium/cilium/pull/25820
- [x] workflows/conformance-ingress.yaml https://github.com/cilium/cilium/pull/25609
- [x] workflows/conformance-k8s-kind.yaml https://github.com/cilium/cilium/pull/25916
- [x] [workflows/conformance-k8s-network-policies.yaml](https://github.com/cilium/cilium/blob/main/.github/workflows/conformance-k8s-network-policies.yaml) it uses `helm install` to install cilium, so nothing to update.
- [x] [workflows/conformance-k8s-kind-network-policies.yam](https://github.com/cilium/cilium/blob/main/.github/workflows/conformance-k8s-kind-network-policies.yaml) https://github.com/cilium/cilium/pull/26692
- [x] workflows/conformance-kind-proxy-daemonset.yaml https://github.com/cilium/cilium/pull/25081
- [x] [workflows/tests-l4lb.yaml](https://github.com/cilium/cilium/blob/main/.github/workflows/tests-l4lb.yaml) it doesn't use cilium-cli to install cilium.
- [x] workflows/tests-smoke-ipv6.yaml: https://github.com/cilium/cilium/pull/25493
- [x] workflows/tests-smoke.yaml: https://github.com/cilium/cilium/pull/25493

helm installation mode should be largely compatible with the classic mode, but we have removed some obsolete flags, so we might need to make minor adjustments to these workflow files. 

ref: https://github.com/cilium/cilium-cli/releases/tag/v0.14.0

tklauser:
[...]

PR #25916, opened 2023-06-06
conformance-k8s-kind: Use Helm mode cilium-cli

michi-covalent:
Ref: https://github.com/cilium/cilium-cli#experimental-helm-installation-mode
Ref: #25156

### DIFF
--- a/.github/workflows/conformance-k8s-kind.yaml
+++ b/.github/workflows/conformance-k8s-kind.yaml
@@ -26,6 +26,7 @@ env:
   # renovate: datasource=github-releases depName=cilium/cilium-cli
   cilium_cli_version: v0.14.5
   cilium_cli_ci_version:
+  CILIUM_CLI_MODE: helm
   k8s_version: v1.27.1
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/conformance-k8s-kind.yaml
```diff
@@ -147,14 +148,8 @@ jobs:
             --helm-set=cni.chainingMode=portmap \
             --helm-set-string=kubeProxyReplacement=strict \
             --helm-set=sessionAffinity=true \
-            --rollback=false \
-            --config monitor-aggregation=none \
-            --version="
-          HUBBLE_ENABLE_DEFAULTS="--chart-directory=install/kubernetes/cilium \
-            --relay-image=quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/hubble-relay-ci:${SHA} \
-            --relay-version=${SHA}"
+            --helm-set=bpf.monitorAggregation=none"
```
sayboras, line 151: you also need to set hubble relay enabled flag as well.

.github/workflows/conformance-k8s-kind.yaml
```diff
@@ -26,6 +26,7 @@ env:
   # renovate: datasource=github-releases depName=cilium/cilium-cli
   cilium_cli_version: v0.14.5
   cilium_cli_ci_version:
+  CILIUM_CLI_MODE: helm
```
sayboras, line 29: I actually try with global env like this, but i faced a couple of issue like flag abc is not supported for cilium cli command (other than cilium install), not sure if it works here, and for all cases.

michi-covalent: yeah let's see if i'm lucky enough.

.github/workflows/conformance-k8s-kind.yaml
michi-covalent: this workflow doesn't use hubble relay right? 🤔 https://github.com/cilium/cilium/blob/main/.github/workflows/conformance-k8s-kind.yaml

sayboras: you are right, seems like the hubble_enable_defaults env is not used :face_exhaling:

sayboras (approved):
:ship:

aanm (approved):

PR #25917, opened 2023-06-06
bpf: nat: reduce CT lookup scope

julianwiedmann:
Continuing from https://github.com/cilium/cilium/pull/25826, this PR applies the concept of "single-scope" CT lookup for the SNAT / RevSNAT paths. Helps to make the code paths simpler & more robust, while reducing complexity.


bpf/lib/nat.h
```diff
@@ -385,12 +347,33 @@ snat_v4_nat_handle_mapping(struct __ctx_buff *ctx,
 
 	*state = __snat_lookup(map, tuple);
 	needs_ct = *state ? (*state)->common.needs_ct : snat_v4_needs_ct(tuple, target);
+
 	if (needs_ct) {
-		ret = snat_v4_track_connection(ctx, tuple, has_l4_header, ct_action,
-					       NAT_DIR_EGRESS, off, ext_err);
+		struct ipv4_ct_tuple tuple_snat = *tuple;
```
gentoo-root: There was some discussion that memcpy generates optimized code while regular struct copy moves data byte-by-byte. Is it still relevant? If so, I suggest keeping memcpy.

(Same for the other occurrences.)

bpf/lib/nat.h
```diff
@@ -404,29 +387,37 @@ snat_v4_rev_nat_handle_mapping(struct __ctx_buff *ctx,
...
-		memcpy(&tuple_revsnat, tuple, sizeof(tuple_revsnat));
 		tuple_revsnat.daddr = (*state)->to_daddr;
 		tuple_revsnat.dport = (*state)->to_dport;
 
-		ret = snat_v4_track_connection(ctx, &tuple_revsnat, has_l4_header, ct_action,
-					       NAT_DIR_INGRESS, off, ext_err);
+		/* CT expects a tuple with the source and destination ports reversed,
+		 * while NAT uses normal tuples that match packet headers.
+		 */
+		ipv4_ct_tuple_swap_ports(&tuple_revsnat);
```
gentoo-root, line 415: Nice saving on one tuple copy! It was bugging me a bit back when I worked on my revSNAT fix.

Might be also worth dropping ipv4_ct_tuple_swap_ports and just going with assigning sport and dport to the right places, what do you think?

gentoo-root (changes requested):

bpf/lib/nat.h
julianwiedmann: Ah, wasn't fully aware of that - thanks! Yes, let's go with the minimal change then :+1:.

bpf/lib/nat.h
julianwiedmann: I bet we can simplify this quite a bit more (especially in combination with your suggestion to pull the tuple swap out of the CT lookup). But let's do that in a follow-on PR, I still have nightmares from too much CT tuple wrangling :grimacing: .

gentoo-root (approved):

PR #25921, opened 2023-06-06
clustermesh: split the generic logic from the specific part

giorio94:
This PR refactors the clustermesh subsystem, splitting it into a generic part (`pkg/clustermesh`), which is responsible for watching the configuration directory and establishing the connections to remote etcd endpoints, and a clustermesh specific part (`pkg/clustermesh/clustermesh`), which handles the actual synchronization of the required information). This is a preparatory change to allow reusing the generic part for the upcoming kvstoremesh implementation, reducing code duplication and maintenance burden.

### DIFF
--- a/pkg/clustermesh/clustermesh.go
+++ b/pkg/clustermesh/clustermesh.go
@@ -5,43 +5,34 @@ package clustermesh
 
 import (
 	"context"
-	"encoding/json"
-	"fmt"
-	"path"
-	"time"
-
-	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/cilium/cilium/api/v1/models"
 	"github.com/cilium/cilium/pkg/allocator"
+	"github.com/cilium/cilium/pkg/clustermesh/internal"
 	"github.com/cilium/cilium/pkg/clustermesh/types"
 	cmtypes "github.com/cilium/cilium/pkg/clustermesh/types"
-	"github.com/cilium/cilium/pkg/controller"
 	"github.com/cilium/cilium/pkg/hive"
 	"github.com/cilium/cilium/pkg/hive/cell"
 	"github.com/cilium/cilium/pkg/ipcache"
 	"github.com/cilium/cilium/pkg/kvstore"
 	"github.com/cilium/cilium/pkg/kvstore/store"
 	"github.com/cilium/cilium/pkg/lock"
-	"github.com/cilium/cilium/pkg/metrics"
+	"github.com/cilium/cilium/pkg/logging"
+	"github.com/cilium/cilium/pkg/logging/logfields"
 	nodeTypes "github.com/cilium/cilium/pkg/node/types"
 	serviceStore "github.com/cilium/cilium/pkg/service/store"
 )
 
-const (
-	// configNotificationsChannelSize is the size of the channel used to
-	// notify a clustermesh of configuration changes
-	configNotificationsChannelSize = 512
+const subsystem = "clustermesh"
 
-	subsystem = "clustermesh"
-)
+var log = logging.DefaultLogger.WithField(logfields.LogSubsys, subsystem)
 
 // Configuration is the configuration that must be provided to
 // NewClusterMesh()
 type Configuration struct {
 	cell.In
 
-	Config
+	internal.Config
 
 	// ClusterIDName is the id/name of the local cluster. This is used for logging and metrics
 	types.ClusterIDName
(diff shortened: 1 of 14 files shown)


pkg/clustermesh/clustermesh/clustermesh.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/cilium/cilium/pkg/lock"
+	"github.com/cilium/cilium/pkg/logging"
+	"github.com/cilium/cilium/pkg/logging/logfields"
+	nodeTypes "github.com/cilium/cilium/pkg/node/types"
+	serviceStore "github.com/cilium/cilium/pkg/service/store"
+)
+
+const subsystem = "clustermesh"
+
+var log = logging.DefaultLogger.WithField(logfields.LogSubsys, subsystem)
```
joamaki: `cell.Module` provides `logrus.FieldLogger` with subsys scoped to module id. Using it will make sure logs match module identifiers and it'll make it easier in the future to use potentially different or augmented logger etc. when it's injected. But since we're mostly moving things around here I don't think it makes sense to change that in the same PR, but perhaps something to keep in mind.

pkg/clustermesh/clustermesh/clustermesh.go
```diff
@@ -0,0 +1,207 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package clustermesh
```
joamaki: `pkg/clustermesh/clustermesh/clustermesh.go` isn't great. Would it be possible to make this `internal` package with public APIs it implements exposed from `pkg/clustermesh` as interface types for example? Or wait, is it the other way around?

giorio94: Yeah, I also personally find `pkg/clustermesh/clustermesh` to be not that great. I'm not sure about a better one though. One possibility could be to just keep everything in the base clustermesh package, losing a bit of separation (and have the kvstoremesh stuff in `pkg/clustermesh/kvstoremesh`). Otherwise I could do the other way round and move the generic bits in `pkg/clustermesh/internal` or `pkg/clustermesh/generic` (is this what you were proposing?). WDYT?

pkg/clustermesh/clustermesh.go
```diff
@@ -119,113 +112,30 @@ func IsClusterConfigRequired(ctx context.Context, backend kvstore.BackendOperati
...
+func NewGenericClusterMesh(c Configuration) GenericClusterMesh {
+	return GenericClusterMesh{
+		conf:     c,
+		clusters: map[string]*remoteCluster{},
 	}
+}
 
-	if c.ClusterMeshConfig == "" {
+func (cm *GenericClusterMesh) Start(hive.HookContext) error {
+	if cm.conf.ClusterMeshConfig == "" {
```
YutaroHayakawa: What's the situation that `ClusterMeshConfig == ""`? Can't we move this check to `NewGenericClusterMesh`? Same for `Stop`.

pkg/clustermesh/clustermesh.go
```diff
@@ -356,64 +246,15 @@ func (cm *ClusterMesh) NumReadyClusters() int {
 	return nready
 }
 
-func (cm *ClusterMesh) canConnect(name string, config *cmtypes.CiliumClusterConfig) error {
+func (cm *GenericClusterMesh) Iterator(fn func(RemoteCluster) error) error {
```
YutaroHayakawa: Nit: For me, `Iterator` is the name of the function that returns an iterator object. I'd name this kind of function with `ForEachRemoteCluster` or something. Just a matter of the preference though 😓

YutaroHayakawa (approved):
LGTM with nits 👍

pkg/clustermesh/clustermesh.go
giorio94: Yeah, sounds good to me. I've renamed this method to `ForEachRemoteCluster`.

pkg/clustermesh/clustermesh.go
giorio94: Essentially that happens when the flag for the path to the clustermesh config is unset. In that case, the clustermesh subsystem should not be started. I've moved back the check directly to `NewClusterMesh` to restore the pre-existing behavior.

pkg/clustermesh/clustermesh/clustermesh.go
giorio94: Makes definitely sense to me. I would leave this as a followup (as you also mentioned) to avoid increasing the scope of this PR.

pkg/clustermesh/clustermesh/clustermesh.go
joamaki: `internal` or `shared` or `generic` package seems like it could work nicely. Hard to say what's best though, so I'll leave it to you.

joamaki (approved):

pkg/clustermesh/clustermesh/clustermesh.go
giorio94: I've moved the generic part to the `pkg/clustermesh/internal` package, and restored back the business logic in `pkg/clustermesh`, adapting variables names accordingly. I've additionally extracted a few clusterconfig-related utilities in a `pkg/clustermesh/utils` package, as they are also used by the clustermesh-apiserver.

giorio94:
@YutaroHayakawa @joamaki PTAL

giorio94:
Force pushed to rebase onto main.

pkg/clustermesh/internal/remote_cluster.go
```diff
@@ -0,0 +1,377 @@
...
+	backend := rc.backend
+	rc.backend = nil
+	rc.mutex.Unlock()
+
+	// Release resources asynchronously in the background. Many of these
+	// operations may time out if the connection was closed due to an error
+	// condition.
+	go func() {
+		if backend != nil {
+			backend.Close(context.TODO())
```
nebril: Nit: could we use `context.Background()` here? It seems to be appropriate for async resource release.

pkg/clustermesh/internal/remote_cluster.go
```diff
@@ -0,0 +1,377 @@
...
+				return
+			}
+		}
+	}()
+
+	go func() {
+		for {
+			select {
+			// terminate routine when remote cluster is removed
+			case _, ok := <-rc.changed:
```
nebril: Please correct me if I am wrong, but I think that this channel receive will race with the receive from line 273. If this goroutine happens to get in `select` while the goroutine above is not executing and there is a message in `rc.changed` channel buffer, it will get the message from channel, check `ok` value to be `true` (since the channel is not closed) and this goroutine will return while `restartRemoteConnection` would not be called above.

To fix that behaviour, I would suggest merging both goroutines into one so that there is only one select.

nebril (changes requested):
Looks good overall, I think I might have found one bug around restarting remote connections.

pkg/clustermesh/internal/remote_cluster.go
giorio94: Yes, you are right. It seems to me that this logic is also possibly prone to other race conditions, as I've seen connections being restarted multiple times in a row following a single failure. I'll refactor it make it more robust.

YutaroHayakawa (approved):
🚀

pkg/clustermesh/internal/remote_cluster.go
giorio94: ~~9668440e8b98e32cd4a5281e8cd25e689a07bafc~~ faebb6f5368610a78c56bd59626dc37e088cc864 addresses this issue refactoring the watchdog logic. I've still maintained the existing `changed` channel logic when an update is detected to avoid introducing too many changes in this PR which is already large due to moving things around.

giorio94: 🤔 There's an issue with the new logic. Let me fix it

giorio94: Error cases should be correctly handled now.

giorio94:
/test-1.16-4.19

giorio94:
Rebased onto main to fix conflicts

giorio94:
Conformance ginkgo hit https://github.com/cilium/cilium/issues/22749 and https://github.com/cilium/cilium/issues/25964

nebril (approved):
LGTM!

joamaki (approved):
Structure looks good!

PR #25926, opened 2023-06-06
.github: add missing job to check for code changes

aanm:
Without this job, the workflow was assuming that code was never modified and therefore was never running the ginkgo tests on comment trigger.


brlbil (approved):
LGTM

PR #25927, opened 2023-06-06
auth: implement re-authentication in case of rotated certificates

### LINKED ISSUE #25475: Mutual Auth: Re-Authentication based on CertificateRotation Events
Since https://github.com/cilium/cilium/pull/24300, the `AuthHandler` interface provides the possibility to get notified about `CertificateRotation` events.

The AuthManager should react on these events by re-authenticating the relation between the two corresponding workloads.

mhofstetter:
This PR introduces mutual auth re-authentication. Whenever an
authhandler is emiting CertificateRotatedEvents, an authentication will
be triggered.

In addition, the existing authentication flow has been refactored to use the hive job framework.

Fixes https://github.com/cilium/cilium/issues/25475

### DIFF
--- a/pkg/auth/manager.go
+++ b/pkg/auth/manager.go
@@ -29,8 +29,9 @@ type authManager struct {
 	authHandlers map[policy.AuthType]authHandler
 	authmap      authMap
 
-	mutex   lock.Mutex
-	pending map[authKey]struct{}
+	mutex                    lock.Mutex
+	pending                  map[authKey]struct{}
+	handleAuthenticationFunc func(a *authManager, k authKey, reAuth bool)
 }
 
 // ipCache is the set of interactions the auth manager performs with the IPCache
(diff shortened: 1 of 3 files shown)


pkg/auth/manager.go
```diff
@@ -74,45 +70,67 @@ func newAuthManager(
...
+			a.handleAuthenticationFunc(a, k, true)
+		}
+	}
+
+	return nil
+}
+
+func handleAuthentication(a *authManager, k authKey, reAuth bool) {
+	if a.markPendingAuth(k) {
+		go func(key authKey) {
```
mhofstetter, line 116: Kind of wonder whether we really want to keep a dedicated go-routine for every (re-)authentication request.

I tend to an implementation with a worker-pool which is working on the auth requests. This would be easier to test and prevent the manager from overload the authhandlers (e.g. SPIRE).

But keeping it as is for the context of this PR.

meyskens (approved):
LGTM looks like a clean approach :) 

A worker pool would be a good improvement for drop-in deployments of spire where it would immediately create spire IDs at the same time so they all get the same renewal schedule. However it could also be a good follow up pr imo.

mhofstetter:
/test-1.26-net-next

PR #25933, opened 2023-06-06
Add gARP capability to L2 announcer feature

dylandreimerink:
When failover occurs a different node will start responding to ARP.
But other nodes on the network will not know this has happened and thus
keep using the ARP entry from their local cache.

This PR makes it so we broadcast a gARP reply with the new MAC
address of the IP any time we reconcile a IP+ifidx tuple that was not
yet present in the L2 responder map.

Nodes that honor gARP replies will update their ARP cache to minimize
downtime. Those who do not honer these replies will have more downtime
until their ARP cache entry expires and they perform a ARP request
themselves.

```release-note
Added gARP capability to L2 announcer feature
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -60,6 +60,7 @@ require (
 	github.com/kr/pretty v0.3.1
 	github.com/mattn/go-shellwords v1.0.12
 	github.com/mdlayher/arp v0.0.0-20220221190821-c37aaafac7f9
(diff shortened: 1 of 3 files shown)


rolinh (approved):
vendor changes lgtm

ldelossa (approved):
LGTM.

markpash (approved):
LGTM

dylandreimerink:
/test-1.26-net-next

PR #25936, opened 2023-06-06
ipsec: Fix IPsec reinitialization

joamaki:
Fixes issues in previous PR (https://github.com/cilium/cilium/pull/25744) on IPsec ENI BPF program loading:

```
loader: In IPsec reload ignore veth devices & fix settle wait
    
reloadIPSecOnLinkChanges() did not ignore veth device updates causing
reload to be triggered when new endpoints were created. Ignore any
updates with "veth" as device type.

The draining of updates during settle wait was broken due to unintentional
breaking out of the loop. Removed the break.
```

```
loader: Do not fatal on IPsec reinitialization

Now that the code is reloading the bpf_network program at runtime we
should not fatal if we fail to reload the program since this may be caused
by ongoing interface changes (e.g. interface was being removed). Change
the log.Fatal into log.Error and keep loading to other interfaces.
```

Validated these fixes in AWS ENI environment with IPsec enabled by creating and
destroying dummy devices:
```
$ for i in $(seq 1 100); do sleep 0.1; ip link add dummy$i type dummy; ip link del dummy$i; done
$ kubectl get logs ...
2023-06-06T13:55:03.474507212Z level=info msg="Reinitializing IPsec due to device changes" subsys=datapath-loader
2023-06-06T13:55:03.480640541Z level=info msg="Encryption network program (re)loaded" interface=ens3 subsys=datapath-loader
2023-06-06T13:55:03.480648286Z level=error msg="Load encryption network failed" error="getting interface dummy1 by name: Link not found" interface=dummy1 subsys=datapath-loader
2023-06-06T13:55:03.480656250Z level=warning msg="Failed to reinitialize IPsec after device change" error="failed to load encryption program: getting interface dummy1 by name: Link not found" subsys=datapath-loader
2023-06-06T13:55:04.479609463Z level=info msg="Reinitializing IPsec due to device changes" subsys=datapath-loader
2023-06-06T13:55:04.482395149Z level=info msg="Encryption network program (re)loaded" interface=ens3 subsys=datapath-loader
2023-06-06T13:55:05.482543443Z level=info msg="Reinitializing IPsec due to device changes" subsys=datapath-loader
```

The time between reinitializations is now at least 1 second and we do not crash when loading fails unexpectedly.

```release-note
Fix three issues in the bug fix to attach IPsec BPF programs to ENI interfaces: do not fatal if loading unexpectedly fails (which may happen if the device is suddenly deleted), ignore veth device changes in order not to reinitialize when new endpoints appear and wait 1 second for further device state changes between reinitializations.
```

### DIFF
--- a/pkg/datapath/loader/loader.go
+++ b/pkg/datapath/loader/loader.go
@@ -346,19 +346,21 @@ func (l *Loader) reloadDatapath(ctx context.Context, ep datapath.Endpoint, dirs 
 	return nil
 }
 
-func (l *Loader) replaceNetworkDatapath(ctx context.Context, interfaces []string) error {
+func (l *Loader) replaceNetworkDatapath(ctx context.Context, interfaces []string) (err error) {
 	progs := []progDefinition{{progName: symbolFromNetwork, direction: dirIngress}}
 	for _, iface := range option.Config.EncryptInterface {
-		finalize, err := replaceDatapath(ctx, iface, networkObj, progs, "")
-		if err != nil {
-			log.WithField(logfields.Interface, iface).WithError(err).Fatal("Load encryption network failed")
+		finalize, replaceErr := replaceDatapath(ctx, iface, networkObj, progs, "")
+		if replaceErr != nil {
+			log.WithField(logfields.Interface, iface).WithError(replaceErr).Error("Load encryption network failed")
+			// Return the error to the caller, but keep trying replacing other interfaces.
+			err = replaceErr
+		} else {
+			log.WithField(logfields.Interface, iface).Info("Encryption network program (re)loaded")
+			// Defer map removal until all interfaces' progs have been replaced.
+			defer finalize()
 		}
-		log.WithField(logfields.Interface, iface).Info("Encryption network program (re)loaded")
-
-		// Defer map removal until all interfaces' progs have been replaced.
-		defer finalize()
 	}
-	return nil
+	return
 }
 
 func (l *Loader) replaceOverlayDatapath(ctx context.Context, cArgs []string, iface string) error {
--- a/pkg/datapath/loader/netlink.go
+++ b/pkg/datapath/loader/netlink.go
@@ -366,15 +366,20 @@ func (l *Loader) reloadIPSecOnLinkChanges() {
 
 		for {
 			// Wait for first update or trigger before reinitializing.
+		getUpdate:
 			select {
-			case _, ok := <-updates:
+			case u, ok := <-updates:
 				if !ok {
 					return
+				}
+				// Ignore veth devices
+				if u.Type() == "veth" {
+					goto getUpdate
 				}
 			case <-trigger:
 			}
 
-			log.Info("Reinitializing IPsec due to link changes")
+			log.Info("Reinitializing IPsec due to device changes")
 			err := l.reinitializeIPSec(ctx)
 			if err != nil {
 				// We may fail if links have been removed during the reload. In this case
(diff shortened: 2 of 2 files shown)


pkg/datapath/loader/loader.go
```diff
@@ -346,19 +346,21 @@ func (l *Loader) reloadDatapath(ctx context.Context, ep datapath.Endpoint, dirs
...
 	progs := []progDefinition{{progName: symbolFromNetwork, direction: dirIngress}}
 	for _, iface := range option.Config.EncryptInterface {
-		finalize, err := replaceDatapath(ctx, iface, networkObj, progs, "")
-		if err != nil {
-			log.WithField(logfields.Interface, iface).WithError(err).Fatal("Load encryption network failed")
+		finalize, replaceErr := replaceDatapath(ctx, iface, networkObj, progs, "")
+		if replaceErr != nil {
+			log.WithField(logfields.Interface, iface).WithError(replaceErr).Error("Load encryption network failed")
+			// Return the error to the caller, but keep trying replacing other interfaces.
+			err = replaceErr
```
pchaigno, line 356: Could such an error ever happen during [the first IPsec initialization](https://github.com/cilium/cilium/blob/8582a62760c3cb1cb496bceb319b4fc679fa9b1b/pkg/datapath/loader/base.go#L507)? If it does, then we will skip a bunch of things. Is that okay? Will Reinitialize retry? Even if it retries, it does mean the watcher for netlink interfaces will never be started.

pkg/datapath/loader/loader.go
```diff
@@ -346,19 +346,21 @@ func (l *Loader) reloadDatapath(ctx context.Context, ep datapath.Endpoint, dirs
...
+		} else {
```
rgo3, line 357: Nit: instead of adding the `else` here we could just add a `continue` inside the `if`?

pkg/datapath/loader/netlink.go
```diff
@@ -391,12 +396,18 @@ func (l *Loader) reloadIPSecOnLinkChanges() {
 				select {
 				case <-settled:
 					break settleLoop
-				case <-updates:
+				case u := <-updates:
```
rgo3, line 399: I find the settleLoop block hard to read. If I understand it correctly I think this would do the equivalent:
```golang
startTime := time.Now()
for time.Since(startTime) < settleDuration {
	select {
	case u := <-updates:
		// Ignore veth devices
		if u.Type() == "veth" {
			continue
		}

		// Trigger reinit immediately after
		// settle duration has passed.
		select {
		case trigger <- struct{}{}:
		default:
		}
	default:
	}
}
```
Probably opinionated, but without the `settleLoop` label I find it easier to understand.

rgo3 (approved):
Just left non-blocking comments, as they don't address anything functional.

pkg/datapath/loader/loader.go
joamaki: The error gets returned to `Reinitialize` which will bubble it up all the way to daemon initialization which will then abort the start and thus leading to cilium-agent restart.

pkg/datapath/loader/netlink.go
joamaki: Yeah this looks nicer. I'll refactor it to something akin to this.

joamaki: Ah the semantics are not the same. Doing a for loop will get stuck if there's no updates, so we must do a select on `After` channel. Since I've already tested this code and the reason for this PR was last minute changes, I'm going to skip changing this :)

pchaigno:
@dylandreimerink Please don't trust MLH blindly. I had yet to re-review here.

pchaigno (commented):
LGTM.

PR #25946, opened 2023-06-06
[WIP] Test upgrade

jschwinger233:
Fixes: #issue-number

```release-note
```

### DIFF
--- a/bpf/init.sh
+++ b/bpf/init.sh
@@ -185,6 +185,7 @@ function encap_fail()
 	# If the host does not have an IPv6 address assigned, assign our generated host
 	# IP to make the host accessible to endpoints
 	if [ "$IP6_HOST" != "<nil>" ]; then
+                echo ">>>>>>>>>> $STALE_IP6_HOST"
                 if [ "$STALE_IP6_HOST" != "<nil>" ]; then
                     ip -6 addr del $STALE_IP6_HOST dev $HOST_DEV1
                 fi
--- a/bpf/init.sh
+++ b/bpf/init.sh
@@ -184,6 +184,7 @@ function encap_fail()
 	# If the host does not have an IPv6 address assigned, assign our generated host
 	# IP to make the host accessible to endpoints
 	if [ "$IP6_HOST" != "<nil>" ]; then
+                ip -6 addr flush dev $HOST_DEV1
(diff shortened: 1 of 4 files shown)


bpf/init.sh
```diff
@@ -184,6 +185,10 @@ function encap_fail()
 	# If the host does not have an IPv6 address assigned, assign our generated host
 	# IP to make the host accessible to endpoints
 	if [ "$IP6_HOST" != "<nil>" ]; then
+                echo ">>>>>>>>>> $STALE_IP6_HOST"
+                if [ "$STALE_IP6_HOST" != "<nil>" ]; then
+                    ip -6 addr del $STALE_IP6_HOST dev $HOST_DEV1
+                fi
```
jschwinger233: @brb Martynas, we can either delete the stale ipv6 here at init.sh or (see the other comment below)

pkg/datapath/loader/base.go
```diff
@@ -372,8 +374,24 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner,
...
+			return err
+		}
+		args[initArgStaleIPv6NodeIP] = "<nil>"
+		for _, addr := range addrs {
+			fmt.Printf(">>>>> matching: %s\n", addr.IPNet.IP.String())
+			if localIPv6CIDR.Contains(addr.IPNet.IP) {
+				args[initArgStaleIPv6NodeIP] = addr.IPNet.IP.String()
+				break
+			}
+		}
```
jschwinger233, line 388: Or we delete the stale ipv6 here, when we generate the new ipv6 for cilium_host.

Which one do you prefer? For me, deleting stale ipv6 at init.sh seems to be more cohesive, as that will be the only place we change the cilium_host's ipv6, but we have to pass an additional shell argument.

(There are other options, like deleting when installing ipv6 routes, but that's way too in-cohesive, so I ditched those ideas.)

brb: As we are migrating from `bpf/init.sh`, I'd suggest to delete it here.

jschwinger233:
Cilium Conformance E2E turns green, it appears our patch from https://github.com/cilium/cilium/pull/25962 is working.

Issue #25948, opened 2023-06-06
Deprecate and remove `cilium endpoint regenerate <id>`


christarazi:
As discussed in the [May 31st, 2023 Cilium community meeting](https://docs.google.com/document/d/1Y_4chDk4rznD6UgXPlPvn3Dc7l-ZutGajUv1eF0VDwQ/edit#heading=h.4oalh7ht365v), let's deprecate `cilium endpoint regenerate <id>` in v1.14 and remove in v1.15.

PR #25949, opened 2023-06-06
cilium/cmd: Deprecate `cilium endpoint regenerate` command

christarazi:
As mentioned in GH-25948, let's deprecate this command and queue it for
removal in v1.15.


qmonnet (approved):
Looks good, thanks

asauber (approved):

PR #25950, opened 2023-06-06
Fix and improve Conformance Ginkgo UX

aanm:
In order for needs.*.outputs.* work, the job needs to depend on the job that is being checked.

Fixes: 79e1c31260c3 (".github: add missing job to check for code changes")


brlbil (approved):
LGTM

PR #25953, opened 2023-06-06
ipsec: Change XFRM FWD policy to simplest wildcard

pchaigno:
This fixes the XFRM FWD policies and the IPv6 catch-all default-drop policy. See commits.

```release-note
Fix leak of IPsec XFRM FWD policies in IPAM modes `cluster-pool`, `kubernetes`, and `crd` when nodes are deleted.
Fix incorrect catch-all default-drop XFRM OUT policy for IPsec IPv6 traffic that could lead to leaking plain-text IPv6 traffic if combined with some other bug.
```

### DIFF
--- a/pkg/datapath/linux/ipsec/ipsec_linux.go
+++ b/pkg/datapath/linux/ipsec/ipsec_linux.go
@@ -97,7 +97,7 @@ var (
 	wildcardIPv6   = net.ParseIP("0::0")
 	wildcardCIDRv6 = &net.IPNet{
 		IP:   wildcardIPv6,
-		Mask: net.CIDRMask(128, 128),
+		Mask: net.CIDRMask(0, 128),
 	}
 
 	defaultDropMark = &netlink.XfrmMark{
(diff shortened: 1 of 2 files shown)


jschwinger233 (approved):

pchaigno:
I needed to repush to address two issues. One in the first commit (see :point_down:) and another one fixed in the second commit.

> There's a small implementation hurdle here. Because we used to install FWD policies of the form "src 0.0.0.0/0 dst 10.0.1.0/24", the kernel was able to deduce which IP family we are matching against and would adapt the 0.0.0.0/0 source CIDR to ::/0 as needed. Now that we are matching on 0/0 for both CIDRs, it cannot deduce this anymore. So instead, we must detect the IP family ourself and use the proper CIDRs.

@jschwinger233 I've therefore re-requested a review from you just to be on the safe side :slightly_smiling_face:

jschwinger233 (approved):
Looks good! I actually noticed that mask for wildcard CIDR before, but I was thinking "yeah since it's been working as expected for long time, no reason there is an obvious bug unknown..." :sob:

PR #24208, opened 2023-03-07
Change cilium_host IPv6 address

jschwinger233:
Originally the IPv6 of cilium_host is set to a native address. Instead, this commit uses an IPv6 allocated from IPAM for
it, which is the same way we set IPv4 address for cilium_host.

Fixes: #23445
Fixes: #21954 
Fixes: #23461

```release-note
Change cilium_host IPv6 address, use node router IPv6 instead of native node IPv6, and fixed several relative IPv6 issues.
```

Signed-off-by: Zhichuan Liang <gray.liang@isovalent.com>

PR #25962, opened 2023-06-07
Delete stale v6 address from cilium_host

### LINKED ISSUE #25938: datapath: When doing downgrade from v1.14 to v1.13 cilium_host IPv6 addr is not changed
In v1.14 we changed the `cilium_host` IPv6 addr derivation (TL;DR - it's allocated from Pod CIDRs IPAM instead of deriving from `eth0`). When doing downgrade to v1.13, both IPv6 addrs are set which breaks installing PodCIDR via `cilium_host` routes.

jschwinger233:
Prior to 1.14, Cilium set the cilium_host IPv6 addr to the same one as the native iface, but #24208 replaces the native IPv6 with the one allocated from IPAM. That change breaks the downgrade path due to failures on installing CIDR routes.

To fix the downgrade path, the ideal way is to delete the stale IPv6 on cilium_host, as long as the IPv6 is from IPAM; but in practical, we don't have a perfect approach to tell if an IPv6 is from IPAM due to the complicated situations for multi-pool IPAM, ENI IPAM, and so on.

Therefore, this PR deletes global scope IPv6 on cilium_host as long as the address is not the one we want. We believe this is so far the most robust way to make sure stale addresses are gone.

Fixes: #25938

```release-note
Fix downgrade path from 1.14 to 1.13 due to stale IPAM-allocated IPv6 on cilium_host
```

### DIFF
--- a/pkg/datapath/loader/base.go
+++ b/pkg/datapath/loader/base.go
@@ -297,17 +297,6 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner, 
 		args[initArgIPv4NodeIP] = "<nil>"
 	}
 
-	if option.Config.EnableIPv6 {
-		args[initArgIPv6NodeIP] = node.GetIPv6().String()
-		// Docker <17.05 has an issue which causes IPv6 to be disabled in the initns for all
-		// interface (https://github.com/docker/libnetwork/issues/1720)
-		// Enable IPv6 for now
-		sysSettings = append(sysSettings,
-			sysctl.Setting{Name: "net.ipv6.conf.all.disable_ipv6", Val: "0", IgnoreErr: false})
-	} else {
-		args[initArgIPv6NodeIP] = "<nil>"
-	}
-
 	args[initArgMTU] = fmt.Sprintf("%d", deviceMTU)
 
 	if option.Config.EnableSocketLB {
@@ -390,8 +379,6 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner, 
 		}
 	}
 
-	sysctl.ApplySettings(sysSettings)
-
 	// Datapath initialization
 	hostDev1, hostDev2, err := SetupBaseDevice(deviceMTU)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


jschwinger233:
The patch has been [tested and confirmed](https://github.com/cilium/cilium/pull/25946) to fix the issue.

brb:
Thanks! I think we should open this PR for the `v1.13`, but not the `main` branch?

brb:
@jschwinger233 Probably you want to base your work from the v1.13 branch too to avoid those 2k commits :sweat_smile:

pkg/datapath/loader/base.go
```diff
@@ -400,6 +389,32 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner,
 	args[initArgHostDev1] = hostDev1.Attrs().Name
 	args[initArgHostDev2] = hostDev2.Attrs().Name
 
+	if option.Config.EnableIPv6 {
+		args[initArgIPv6NodeIP] = node.GetIPv6().String()
+		// Docker <17.05 has an issue which causes IPv6 to be disabled in the initns for all
+		// interface (https://github.com/docker/libnetwork/issues/1720)
+		// Enable IPv6 for now
+		sysSettings = append(sysSettings,
```
brb, line 395: The warning bellow looks legit, as `sysctl.ApplySettings(sysSettings)` is called above.

pkg/datapath/loader/base.go
```diff
@@ -400,6 +387,34 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner,
...
+		localIPv6CIDR := node.GetIPv6AllocRange()
+
+		ciliumHostV6Addrs, err := netlink.AddrList(hostDev1, netlink.FAMILY_V6)
+		if err != nil {
+			return err
+		}
+		for _, addr := range ciliumHostV6Addrs {
+			// Delete cilium_host v6 addresses allocated from IPAM to make sure downgrade from 1.14 to 1.13 works:
+			// https://github.com/cilium/cilium/issues/25938
+			if localIPv6CIDR.Contains(addr.IPNet.IP) {
```
brb: One more question - is there any other way to tell that IP was allocated from the IPAM? Asking, as with the multi-pool IPAM, ENI IPAM, etc. we cannot assume that the IPs will be allocated from a single CIDR.

jschwinger233: Maybe we can call this function to check all the CIDRs: 
https://github.com/cilium/cilium/blob/592777da560bea7838b99223386e943c08d5d052/pkg/node/types/node.go#L578

pkg/datapath/loader/base.go
```diff
@@ -400,6 +387,33 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner,
...
+			sysctl.Setting{Name: "net.ipv6.conf.all.disable_ipv6", Val: "0", IgnoreErr: false})
+
+		ciliumHostV6Addrs, err := netlink.AddrList(hostDev1, netlink.FAMILY_V6)
+		if err != nil {
+			return err
+		}
+		for _, addr := range ciliumHostV6Addrs {
+			// Delete stale cilium_host IPv6 to make sure downgrade from 1.14 to 1.13 works:
+			// https://github.com/cilium/cilium/issues/25938
+			if addr.Scope == int(netlink.SCOPE_UNIVERSE) && addr.IP.String() == args[initArgIPv6NodeIP] {
```
brb: Last nit - let's add a info msg that we are removing that IP addr. This would give a visibility to users which manually added IPv6 addrs.

jschwinger233: Actually, I implemented the logic totally wrong :rofl: it should be `addr.IP.String() != args[initArgIPv6NodeIP] ` but I wrote `==`.
And of course, info log :+1:

brb (approved):
Thanks!

jschwinger233:
Opened an empty PR based on v1.13 to see which jobs have been broken: https://github.com/cilium/cilium/pull/25981

jschwinger233:
/test-backport-1.13

brb:
Tested with the downgrade CI - it works! https://github.com/cilium/cilium/actions/runs/5198987404/jobs/9375802753 (please ignore the red builds, as they are suffering from other problems).

pkg/datapath/loader/base.go
```diff
@@ -400,6 +387,41 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner,
...
 
+	if option.Config.EnableIPv6 {
+		args[initArgIPv6NodeIP] = node.GetIPv6().String()
+		// Docker <17.05 has an issue which causes IPv6 to be disabled in the initns for all
+		// interface (https://github.com/docker/libnetwork/issues/1720)
+		// Enable IPv6 for now
+		sysSettings = append(sysSettings,
+			sysctl.Setting{Name: "net.ipv6.conf.all.disable_ipv6", Val: "0", IgnoreErr: false})
+
+		ciliumHostV6Addrs, err := netlink.AddrList(hostDev1, netlink.FAMILY_V6)
```
lmb, line 398: Is hostDev1 one half of a pod veth pair or is this the real device connecting us to the outside world? Deleting addresses on the latter sounds kind of scary!

P.S: `hostDev1, hostDev2, err := SetupBaseDevice` makes me think it's the veth.

lmb (approved):

pkg/datapath/loader/base.go
jschwinger233: Yeah this is a pair of veth :+1:  they are cilium_host and cilum_net.

jschwinger233:
/test-1.19-4.19

jschwinger233:
All the `required` CI jobs have passed, the red ones are triggered by `/test` and they are supposed to fail for 1.13 according to the probe PR https://github.com/cilium/cilium/pull/25981.

PR #25963, opened 2023-06-07
ipam/allocator: remove unused allocator types

tklauser:
The PoolGroupAllocator and NoOpAllocator allocator types are unused since commit 24cb0618756b ("azure: Calculate available addresses based on subnet resource"). These are the only types implementing the corresponding Allocator interface.

The PoolAllocator type (not to be confused with
ipam/allocator/multipool.PoolAllocator) was only used by PoolGroupAllocator allocator, so remove it as well.


gandro (approved):
Nice catch! :recycle:

PR #25966, opened 2023-06-07
kvstore: limit keys attached to single lease, and react to expiration

giorio94:
Reporting the description of the first commit for convenience:

***
Currently, each etcd client is associated with a single session, and the
corresponding lease is attached to all upserted keys (if the lease
parameter is set). This approach, though, suffers from performance
issues, because put requests in etcd take linear apply time depending
on the number of keys already attached to the lease (https://github.com/etcd-io/etcd/issues/15993).
This performance penalty is planned to be fixed in etcd (at least for the
common case in which the user which performs the request has root role).

In the meanwhile, let's make sure that we attach a limited number of
keys to a single lease. In particular, this commit introduces the etcd
lease manager, which is responsible for managing the lease acquisitions,
tracking the keys that are attached to each of them. Once the number of
keys per lease exceeds the configured threshold, a new lease gets
automatically acquired. The lease usage counter is decremented when a
given key gets deleted. Finally, in case one of the leases fails to be
renewed, the manager allows to emit a notification event for all the keys
that were attached to it.
***

Additionally, the sync store implementation is modified to register for the lease expiration events, appropriately triggering the re-creation of the given keys. This allows to mitigate the possibility that keys get deleted because the associated lease failed to be renewed.

### DIFF
--- a/pkg/kvstore/etcd.go
+++ b/pkg/kvstore/etcd.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/blang/semver/v4"
@@ -335,6 +336,8 @@ type etcdClient struct {
 	listBatchSize int
 
 	lastHeartbeat time.Time
+
+	leaseExpiredObservers sync.Map
 }
 
 func (e *etcdClient) getLogger() *logrus.Entry {
@@ -640,7 +643,7 @@ func connectEtcdClient(ctx context.Context, config *client.Config, cfgPath strin
 		statusCheckErrors:    make(chan error, 128),
 	}
 
-	ec.leaseManager = newEtcdLeaseManager(c, option.Config.KVstoreLeaseTTL, etcdMaxKeysPerLease, nil, ec.getLogger())
+	ec.leaseManager = newEtcdLeaseManager(c, option.Config.KVstoreLeaseTTL, etcdMaxKeysPerLease, ec.expiredLeaseObserver, ec.getLogger())
 
 	// create session in parallel as this is a blocking operation
 	go func() {
@@ -1696,6 +1699,26 @@ func (e *etcdClient) ListAndWatch(ctx context.Context, name, prefix string, chan
 	go e.Watch(ctx, w)
 
 	return w
+}
+
+// RegisterLeaseExpiredObserver registers a function which is executed when
+// the lease associated with a key having the given prefix is detected as expired.
+// If the function is nil, the previous observer (if any) is unregistered.
+func (e *etcdClient) RegisterLeaseExpiredObserver(prefix string, fn func(key string)) {
+	if fn == nil {
+		e.leaseExpiredObservers.Delete(prefix)
+	} else {
+		e.leaseExpiredObservers.Store(prefix, fn)
+	}
+}
+
+func (e *etcdClient) expiredLeaseObserver(key string) {
+	e.leaseExpiredObservers.Range(func(prefix, fn any) bool {
+		if strings.HasPrefix(key, prefix.(string)) {
+			fn.(func(string))(key)
+		}
+		return true
+	})
 }
 
 // UserEnforcePresence creates a user in etcd if not already present, and grants the specified roles.
--- a/pkg/kvstore/etcd.go
+++ b/pkg/kvstore/etcd.go
@@ -57,8 +57,10 @@ const (
 
 	minRequiredVersionStr = ">=3.1.0"
 
-	etcdSessionRenewNamePrefix     = "kvstore-etcd-session-renew"
 	etcdLockSessionRenewNamePrefix = "kvstore-etcd-lock-session-renew"
+
+	// etcdMaxKeysPerLease is the maximum number of keys that can be attached to a lease
+	etcdMaxKeysPerLease = 1000
 )
 
 var (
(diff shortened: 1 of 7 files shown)


pkg/kvstore/etcd_lease.go
```diff
@@ -0,0 +1,270 @@
...
+		elm.mu.Lock()
+		if info := elm.leases[leaseID]; info != nil {
+			info.cancel()
+		}
+		elm.mu.Unlock()
+	}
+}
+
+// Leases returns the number of managed leases.
+func (elm *etcdLeaseManager) Leases() uint32 {
```
marseel: Can we rename it to `TotalLeases` or sth like that to indicate that it's a number and not list of leases?

pkg/kvstore/etcd_lease.go
```diff
@@ -0,0 +1,270 @@
...
+
+// KeyHasLease returns whether the given key is associated with the specified lease.
+func (elm *etcdLeaseManager) KeyHasLease(key string, leaseID client.LeaseID) bool {
+	elm.mu.RLock()
+	defer elm.mu.RUnlock()
+
+	return elm.keys[key] == leaseID
+}
+
+// CheckExpired verifies whether the error reports that the given lease has
```
marseel: Can we rename it to `CancelIfExpired` or something similar? `CheckExpired` suggests that it's a non-mutating method.

pkg/kvstore/store/syncstore.go
```diff
@@ -298,16 +309,46 @@ func (wss *wqSyncStore) handleSync(ctx context.Context) error {
...
+		}
 	}
 
 	return nil
 }
 
+// handleExpiredLease gets executed when the lease attached to a given key expired,
+// and is responsible for enqueuing the given key to recreate it.
+func (wss *wqSyncStore) handleExpiredLease(key string) {
+	defer wss.queuedMetric.Set(float64(wss.workqueue.Len()))
```
marseel, line 324: Can we add error log here with information that lease expired?

pkg/kvstore/etcd_lease.go
```diff
@@ -0,0 +1,270 @@
...
+
+	select {
+	case <-elm.client.Ctx().Done():
+		// The context of the etcd client was closed
+		return
+	case <-ctx.Done():
+	default:
+	}
+
+	elm.log.WithField("LeaseID", leaseID).Info("Lease expired")
```
marseel: I think we can change it to at least warning.

pkg/kvstore/store/syncstore.go
marseel: Actually nvm, onle one log will be fine when lease expires.

pkg/kvstore/etcd_lease.go
giorio94: Changed, `CancelIfExpired` is much clearer.

pkg/kvstore/etcd.go
```diff
@@ -1416,11 +1324,15 @@ func (e *etcdClient) UpdateIfLocked(ctx context.Context, key string, value []byt
 	duration := spanstat.Start()
 	e.limiter.Wait(ctx)
 	if lease {
-		leaseID := e.GetSessionLeaseID()
+		leaseID, err := e.leaseManager.GetLeaseID(ctx, key)
+		if err != nil {
+			return Hint(err)
```
marseel, line 1330: Probably we need to add `increaseMetric` here to include this error.

pkg/kvstore/etcd.go
```diff
@@ -1443,10 +1355,14 @@ func (e *etcdClient) Update(ctx context.Context, key string, value []byte, lease
 
 	if lease {
 		duration := spanstat.Start()
-		leaseID := e.GetSessionLeaseID()
+		leaseID, err := e.leaseManager.GetLeaseID(ctx, key)
+		if err != nil {
+			return Hint(err)
```
marseel, line 1362: Metrics here too.

pkg/kvstore/etcd.go
```diff
@@ -1602,7 +1515,10 @@ func (e *etcdClient) CreateOnly(ctx context.Context, key string, value []byte, l
 	duration := spanstat.Start()
 	var leaseID client.LeaseID
 	if lease {
-		leaseID = e.GetSessionLeaseID()
+		leaseID, err = e.leaseManager.GetLeaseID(ctx, key)
+		if err != nil {
+			return false, Hint(err)
```
marseel, line 1524: Metrics here too.

pkg/kvstore/etcd.go
```diff
@@ -1544,7 +1454,10 @@ func (e *etcdClient) CreateOnlyIfLocked(ctx context.Context, key string, value [
...
```
marseel, line 1462: Metrics here too.

pkg/kvstore/etcd.go
```diff
@@ -1627,7 +1543,10 @@ func (e *etcdClient) CreateIfExists(ctx context.Context, condKey, key string, va
 	duration := spanstat.Start()
 	var leaseID client.LeaseID
 	if lease {
-		leaseID = e.GetSessionLeaseID()
+		leaseID, err = e.leaseManager.GetLeaseID(ctx, key)
+		if err != nil {
+			return Hint(err)
```
marseel, line 1553: Metrics here too.

marseel (approved):
lgtm, except for one minor thing with metrics.

pkg/kvstore/etcd.go
giorio94: Yeah, you are right. I tend to always forget about metrics. They should be present now.

giorio94:
Rebased onto main to pick the CI fixes.

tklauser (approved):

PR #25979, opened 2023-06-07
policy: allocate identity for CIDR selectors on first use (v2)

squeed:
This is a rewrite #25581 after @joestringer had some suggestions. I'm leaving that PR as-is to preserve the discussion.

**TODO:**

- [ ] Get consensus that this is the right direction
- [ ] Write some unit tests
- [ ] Consider rewriting existing unit tests

This PR consists of two basic changes:
1: Make policy not rely on the endpoint lock
2: Allocate CIDR identities inside the SelectorCache

### Part 1: The endpoint lock
The endpoint lock is just used to protect access to a number of fields against concurrent writers. It is not used to serialize disparate operations - for that, there are other locks (e.g. the bpf lock). To that end, there are a number of small commits that either remove unnecessary locking (because a field is actually immutable after creation) or switch to atomic pointers. 

An additional commit changes `regeneratePolicy()` to not hold the endpoint lock until it is time to commit. This is necessary to prepare for part 2.

### Part 2: CIDR allocation
Before this change, we would allocate CIDR identities as soon as a policy was ingested - on every node. This meant that nodes would have entries in the ipcache and potentially unnecessary identities, because this was done even if no pods relied on those selectors. This change moves allocation in to the SelectorCache, which is similar to how FQDN policies work.

The interesting bit is that allocation is asynchronous, so we need to wait for the ipcache to make progress and update back in to the SelectorCache that the requested identities exist. As part of genernal paranoia, I added a timeout. This is ultimately safe since we will "eventually" coalesce on the desired state. I tried to find a way to reflect this waiting back to the caller, but the tree was very deep, so I couldn't find a good way.

```release-note
TODO
```

### DIFF
--- a/pkg/endpoint/bpf.go
+++ b/pkg/endpoint/bpf.go
@@ -266,16 +266,13 @@ func (e *Endpoint) addNewRedirectsFromDesiredPolicy(ingress bool, desiredRedirec
 				finalizeList.Append(finalizeFunc)
 				revertStack.Push(revertFunc)
 
-				if e.realizedRedirects == nil {
-					e.realizedRedirects = make(map[string]uint16)
-				}
-				if _, found := e.realizedRedirects[proxyID]; !found {
+				if _, found := e.realizedRedirects.Load(proxyID); !found {
 					revertStack.Push(func() error {
-						delete(e.realizedRedirects, proxyID)
+						e.realizedRedirects.Delete(proxyID)
 						return nil
 					})
 				}
-				e.realizedRedirects[proxyID] = redirectPort
+				e.realizedRedirects.Store(proxyID, redirectPort)
 
 				desiredRedirects[proxyID] = true
 
@@ -392,16 +389,13 @@ func (e *Endpoint) addVisibilityRedirects(ingress bool, desiredRedirects map[str
 		finalizeList.Append(finalizeFunc)
 		revertStack.Push(revertFunc)
 
-		if e.realizedRedirects == nil {
-			e.realizedRedirects = make(map[string]uint16)
-		}
-		if _, found := e.realizedRedirects[proxyID]; !found {
+		if _, found := e.realizedRedirects.Load(proxyID); !found {
 			revertStack.Push(func() error {
-				delete(e.realizedRedirects, proxyID)
+				e.realizedRedirects.Delete(proxyID)
 				return nil
 			})
 		}
-		e.realizedRedirects[proxyID] = redirectPort
+		e.realizedRedirects.Store(proxyID, redirectPort)
 
 		desiredRedirects[proxyID] = true
 
@@ -497,24 +491,27 @@ func (e *Endpoint) removeOldRedirects(desiredRedirects map[string]bool, proxyWai
 
 	var finalizeList revert.FinalizeList
 	var revertStack revert.RevertStack
-	removedRedirects := make(map[string]uint16, len(e.realizedRedirects))
-	updatedStats := make(map[uint16]*models.ProxyStatistics, len(e.realizedRedirects))
+	removedRedirects := make(map[string]uint16)
+	updatedStats := make(map[uint16]*models.ProxyStatistics)
 
-	for id, redirectPort := range e.realizedRedirects {
+	//for id, redirectPort := range e.realizedRedirects {
+	e.realizedRedirects.Range(func(k, v any) bool {
+		id := k.(string)
+		redirectPort := v.(uint16)
 		// Remove only the redirects that are not required.
 		if desiredRedirects[id] {
-			continue
+			return true
 		}
 
 		err, finalizeFunc, revertFunc := e.proxy.RemoveRedirect(id, proxyWaitGroup)
 		if err != nil {
 			e.getLogger().WithError(err).WithField(logfields.L4PolicyID, id).Warn("Error while removing proxy redirect")
-			continue
+			return true
 		}
 		finalizeList.Append(finalizeFunc)
 		revertStack.Push(revertFunc)
 
-		delete(e.realizedRedirects, id)
+		e.realizedRedirects.Delete(id)
 		removedRedirects[id] = redirectPort
 
 		// Update the endpoint API model to report that no redirect is
@@ -529,7 +526,8 @@ func (e *Endpoint) removeOldRedirects(desiredRedirects map[string]bool, proxyWai
 			e.getLogger().WithField(logfields.L4PolicyID, id).Warn("Proxy stats not found")
 		}
 		e.proxyStatisticsMutex.Unlock()
-	}
+		return true
+	})
 
 	return finalizeList.Finalize,
 		func() error {
@@ -543,7 +541,7 @@ func (e *Endpoint) removeOldRedirects(desiredRedirects map[string]bool, proxyWai
 			e.proxyStatisticsMutex.Unlock()
 
 			for id, redirectPort := range removedRedirects {
-				e.realizedRedirects[id] = redirectPort
+				e.realizedRedirects.Store(id, redirectPort)
 			}
 
 			err := revertStack.Revert()
@@ -1193,14 +1191,14 @@ func (e *Endpoint) applyPolicyMapChanges() error {
 	if e.visibilityPolicy != nil {
 		for _, visMeta := range e.visibilityPolicy.Ingress {
 			proxyID := policy.ProxyID(e.ID, visMeta.Ingress, visMeta.Proto.String(), visMeta.Port)
-			if redirectPort, exists := e.realizedRedirects[proxyID]; exists && redirectPort != 0 {
-				e.desiredPolicy.PolicyMapState.AddVisibilityKeys(e, redirectPort, visMeta, adds, nil)
+			if redirectPort, exists := e.realizedRedirects.Load(proxyID); exists && redirectPort.(uint16) != 0 {
+				e.desiredPolicy.PolicyMapState.AddVisibilityKeys(e, redirectPort.(uint16), visMeta, adds, nil)
 			}
 		}
 		for _, visMeta := range e.visibilityPolicy.Egress {
 			proxyID := policy.ProxyID(e.ID, visMeta.Ingress, visMeta.Proto.String(), visMeta.Port)
-			if redirectPort, exists := e.realizedRedirects[proxyID]; exists && redirectPort != 0 {
-				e.desiredPolicy.PolicyMapState.AddVisibilityKeys(e, redirectPort, visMeta, adds, nil)
+			if redirectPort, exists := e.realizedRedirects.Load(proxyID); exists && redirectPort.(uint16) != 0 {
+				e.desiredPolicy.PolicyMapState.AddVisibilityKeys(e, redirectPort.(uint16), visMeta, adds, nil)
 			}
 		}
 	}
@@ -1224,7 +1222,12 @@ func (e *Endpoint) applyPolicyMapChanges() error {
 		// due to the fact that proxies may not yet have bound to a specific port when a proxy
 		// policy is first instantiated.
 		if entry.IsRedirectEntry() {
-			entry.ProxyPort = e.realizedRedirects[policy.ProxyIDFromKey(e.ID, keyToAdd)]
+			pp, exists := e.realizedRedirects.Load(policy.ProxyIDFromKey(e.ID, keyToAdd))
+			if !exists {
+				entry.ProxyPort = 0
+			} else {
+				entry.ProxyPort = pp.(uint16)
+			}
 		}
 		if !e.addPolicyKey(keyToAdd, entry, true) {
 			errors++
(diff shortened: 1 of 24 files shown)


squeed:
Neat, caught a bug in go-staticcheck. It thinks a variable can't be nil, but it can.

pkg/policy/selectorcache.go
```diff
@@ -921,6 +978,28 @@ func (sc *SelectorCache) AddIdentitySelector(user CachedSelectionUser, selector
...
+		// If the ipcache is somehow not making progress, we don't want
+		// to block progress. This is safe, because the updated selections
+		// will be pushed back down in to the policy maps anyways
+		select {
+		case <-newIDSel.hasSelectionsCh:
+			log.WithField("prefixes", prefixes).Debug("Wait time for CIDR prefix allocation: ", time.Since(start))
+		case <-time.After(5 * time.Second):
+			log.WithField("prefixes", prefixes).Warn("Timed out waiting for CachedSelector to be populated! Policy may be temporarily inconsistent")
+		}
+	}
```
christarazi: I wonder if it's just much simpler to make a variant of `UpsertPrefixes` that's synchronous, now that the endpoint locks are no longer a concern in terms of deadlocks (`UpsertPrefixes` can take the endpoint lock for endpoint regeneration).

See https://github.com/christarazi/cilium/commit/df396ef4d545017acc1e852fc5f41671b63027db.

christarazi: To expand, I don't really see the benefits of having this complex logic at the call site when it should be safe to do synchronously. We will need to use the same approach for the FQDN selectors in this very same file anyway, see https://github.com/cilium/cilium/issues/21142.

squeed: Both of those proposed efforts are asynchronous -- the question is: who is doing the waiting. Is it the ipcache itself, or the consumer of the ipcache?

The advantage of moving the waiting in to the SelectorCache is that we can short-circuit waiting if we're already at the desired state. Say, for example, that the CIDR in question already exists. Then there's no need to wait for the ipcache to make progress; we can immediately continue. Likewise, if the ipcache makes partial progress but doesn't complete our unit of work, we could falsely assume that it is ready.

The disadvantage is that it is somewhat more complex; the waiting is now "far away" from the ipcache, code-wise. This means that debugging is more difficult, since the systems are loosely coupled.

pkg/policy/selectorcache.go
```diff
@@ -921,6 +978,28 @@ func (sc *SelectorCache) AddIdentitySelector(user CachedSelectionUser, selector
...
+	start := time.Now()
+	if newIDSel.waitingForSelections > 0 {
+		// Give ourselves a 5-second timeout, just out of paranoia.
+		// If the ipcache is somehow not making progress, we don't want
+		// to block progress. This is safe, because the updated selections
+		// will be pushed back down in to the policy maps anyways
+		select {
+		case <-newIDSel.hasSelectionsCh:
+			log.WithField("prefixes", prefixes).Debug("Wait time for CIDR prefix allocation: ", time.Since(start))
+		case <-time.After(5 * time.Second):
```
joestringer: I have a concern here that if there are many selectors that are forced to wait for a long time here, policy calculation will effectively stall.

Could we push the channel up a few layers of the stack so that it happens once for the overall policy calculation (l4filter level?) rather than for each selector within this l4 filter?

We may also want to revisit what a reasonable time to wait is here. This could potentially block the adding of new policies because pkg/endpoint is holding the policy repository RLock across this call AFAIK.

squeed: > Could we push the channel up a few layers of the stack so that it happens once for the overall policy calculation (l4filter level?) rather than for each selector within this l4 filter?

That's a fine idea -- I'll do some exploring to find the most reasonable place to put this. I'll probably refactor it to use a waitgroup, fwiw.

The call stack from `regeneratePolicy()` is about 10 calls deep :-).

squeed: It looks like we can actually do this easily enough all the way up in `regeneratePolicy()` by adding a method on the `CachedSelector` interface.

pkg/endpoint/bpf.go
```diff
@@ -497,24 +491,27 @@ func (e *Endpoint) removeOldRedirects(desiredRedirects map[string]bool, proxyWai
...
 		if err != nil {
 			e.getLogger().WithError(err).WithField(logfields.L4PolicyID, id).Warn("Error while removing proxy redirect")
-			continue
+			return true
 		}
 		finalizeList.Append(finalizeFunc)
 		revertStack.Push(revertFunc)
 
-		delete(e.realizedRedirects, id)
+		e.realizedRedirects.Delete(id)
```
christarazi, line 514: The old code did this too, but I'm wondering if it's safe to delete something while ranging over it?

pkg/endpoint/endpoint.go
```diff
@@ -302,7 +303,7 @@ type Endpoint struct {
 	// successfully added into a proxy for this endpoint, to the redirect's
 	// proxy port number.
 	// You must hold Endpoint.mutex to read or write it.
-	realizedRedirects map[string]uint16
+	realizedRedirects sync.Map
```
christarazi, line 308: The comment about the mutex is no longer accurate anymore, right?

pkg/endpoint/endpoint.go
```diff
@@ -128,9 +128,11 @@ type Endpoint struct {
 	mutex lock.RWMutex
 
 	// containerName is the name given to the endpoint by the container runtime
+	// mutable, must be read with the endpoint lock!
```
christarazi, line 131: Nit: It would be easier to read with a period after the first sentence here. Also applies below with the other changes you've made in this file as well.

pkg/policy/selectorcache.go
christarazi: After looking into this more deeply, the current implementation introduces a window for transient policy drops. Let me elaborate below:

We have a general constraint that for policy updates to occur "correctly" (no transient packet drops), BPF policymap updates must occur strictly before BPF ipcachemap updates for a particular IP <-> Identity mapping.

We ensure this today in IPCache via `InjectLabels()` -> `UpdatePolicyMaps()`, which blocks the ipcache bpf map sync until the endpoint policymaps are populated with the identities allocated by the ipcache.

That same logic from `UpdatePolicyMaps()` reaches through the selectorcache to this function here. When this line does `close()` of the channel, it asynchronously triggers the policymap sync, but does not guarantee that it completes before returning back up to the ipcache. This violates the constraint at the top of this message.

Some ideas to resolve this come to mind like adding two channels (ipcache signals that identity is allocated, selectorcache signals that the policymaps are plumbed), but we're getting more and more into complex locking / synchronous territory. We may want to look outside the box a bit more.

christarazi: Additionally, the synchronous variant I suggested above and the current implementation could completely halt Cilium in the case that the ipcache cannot make progress for some reason. By halt, I mean that the ipcache will be stuck retrying to "inject labels" for some prefixes, which will cause the caller (the selectorcache here) to wait for ipcache, and the caller of the selectorcache (policy repo) will then also get backed up. Since the policy repo lock is taken while all of this is occurring, no policy updates can be processed--not even deletes.

One instance where the ipcache cannot make progress is if it cannot allocate an identity for a prefix. In the case of a local CIDR identity, it can happen if the local identity space is exhausted. The mitigation would be to tell the user to delete some policies to free up some identities. But that mitigation would be rendered impossible due to the above.

christarazi: I think we can avoid this by having the ipcache expose an API that allows us to associate a channel to some prefixes. The ipcache then closes the channel associated with a set of prefixes to signal that they have been "processed". This enables the caller to know when the ipcache work is done. With this async API, the caller can spin off ipcache and continue doing whatever work necessary, and then before returning, wait for ipcache to finish (with some timeout) so it's not fully blocking.

One approach that comes to mind is to modify the `UpsertPrefixes` call to taken in a channel and then plumbing that channel all the way through such that it's closed **only when all** the prefixes passed in from `UpsertPrefixes` are "done".

christarazi: With `UpsertPrefixes` now able to signal when a set of prefixes have been plumbed, we can leverage this during policy calculation. The commits to remove the locking requirement for policy calculation during endpoint regeneration means we can further decouple policy calculation from endpoint regeneration. 

In other words, since policy calculation ultimately is based on identities, rather than endpoints, we can logically extract out the policy calculation from endpoint regeneration into a separate function. This function will eventually call down to `UpsertPrefixes`. To mitigate the backpressure problem when ipcache can't make progress, we can avoid holding the policy repo lock (potentially only need to hold it until we call down to `UpsertPrefixes`?).

To solve the original problem that this PR is meant to do which is dynamically allocate CIDR identities, we can call the new function on policy-update and on CNI-add / CNI-delete. This should cover 

1. new policy has new CIDR rules
2. new endpoint is newly selected by existing policy with CIDR rules
3. last endpoint with identity X referencing CIDR rules is deleted

which I believe to be all the cases for ensuring only active CIDR identities.

christarazi: There's some obvious hand waving here and details to work out, but architecturally I think this is getting towards a more robust model that better separates and defines contracts between endpoints, identities, and policy. With a cleaner architecture, it will help avoid footguns like potentially locking up Cilium in case of errors and transient drops. 

My biggest concern the whole time with the selectorcache approach was the added complex the selectorcache layer which really had no business allocating new identities. It took a while before realizing what an alternative approach can look like, but going forward, I'm hoping the new model also lends itself a bit better to leverage `pkg/stream` objects.

christarazi (commented):
Continuing my review from prior. I've left some non-trivial comments that dive pretty deep into the policy calculation architecture, as well as some minor comments. 

My suggestion at this point is to make this PR just for the endpoint lock decoupling from regeneration and drop the dynamic CIDR identity commit for now. Regardless of the approach for the latter, we can separately evaluate the endpoint lock commits, so might as well separate them out. I'm happy to take the discussion of the dynamic CIDR identity to another PR, but I'll leave my comments as-is for now.

pkg/policy/selectorcache.go
christarazi: As Joe suggested, if we are able to do the policy calculation at the l4filter later, then it lends itself much better to what's suggested [here](https://github.com/cilium/cilium/pull/25979#discussion_r1227564883).

pkg/endpoint/bpf.go
squeed: It's certainly safe for a sync map.

pkg/endpoint/endpoint.go
squeed: agreed; I was somehow matching style but I don't know why.

sayboras (approved):
I only reviewed the files owned by cilium/proxy team (e.g. envoy/server.go, proxy/logger/*)

LGTM :heavy_check_mark:

pkg/policy/selectorcache.go
squeed: > That same logic from UpdatePolicyMaps() reaches through the selectorcache to this function here. When this line does close() of the channel, it asynchronously triggers the policymap sync, but does not guarantee that it completes before returning back up to the ipcache.

I don't think this is an issue.

Closing the channel means that the policy calculation is allowed to proceed. In other words, it now knows that the cached selector in question is up-to-date. This is actually irrelevant to policymap syncs, because we're not consuming deltas, we're generating the policy from scratch. In other words, the policy map for the newly-calculated policy will be correct.

The critical observation is that `ipcache.UpdatePolicyMaps` does two things:

1. Update the SelectorCache. This is `UpdateIdentities()`
2. Push deltas in to every `endpoint.desiredPolicy`. This is `endpointmanager.UpdatePolicyMaps()`

Step 1 will close the channel and allow policy distillation to proceed. But, again, I think this is safe because distillation only consumes the SelectorCache, not the bpf ipcache. In other words, while we have added a timing point, we haven't actually reordered any operations.

squeed:
I've been testing this with highscale ipcache and it does work as intended. The only downside is that it presently triggers N unnecessary regenerations, where N is the number of CIDR identities allocated while and endpoint is being created. I'm curious if this is an existing bug, or if this is expected behavior w.r.t. the identity allocator and its consumers.

In other words, right now, allocating a new identity will trigger regeneration of all endpoints, all the time. Is that desired?

pkg/policy/selectorcache.go
christarazi: Ah, so the reason why it's not an issue is because I assumed that `ipcache.UpdatePolicyMaps` went through the full policy calculation path which is part of endpoint regeneration, but it's actually the incremental path which directly updates the policymaps via `applyPolicyMapChanges()` instead.

squeed:
We've decided to punt the dynamic CIDR allocation bits for now, and just focus on lockless policy generation. I'll close this PR and file a follow-up one.

squeed:
Closing in favor of #26242, which just shuffles the endpoint lock but doesn't touch CIDR allocation.

PR #25983, opened 2023-06-07
Temporarily disable part of the conformance-kind test

aspsk:
Patch two tests so that CI is green again; because currently it fails of complexity on new kernels (including all github runners). Namely:
  * conformance-kind: temporarily disable ipsec
  * smoke-test: set hostServices=true to reduce complexity

```release-note
Temporarily disable part of the conformance-kind test.
```

### DIFF
--- a/.github/workflows/conformance-kind.yaml
+++ b/.github/workflows/conformance-kind.yaml
@@ -117,22 +117,31 @@ jobs:
           pkill -f "cilium.*hubble.*port-forward|kubectl.*port-forward.*hubble-relay"
           cilium uninstall --chart-directory=install/kubernetes/cilium --wait
 
+      # The recent [stable] linux kernels contain a commit which increases
+      # complexity a lot. In particular, this breaks the connectivity test when
+      # ipsec is enabled. Temporarily disable this test until a better solution
+      # is found.
+
       - name: Install Cilium with encryption
+        if: ${{ false }} # temporary, see the comment above
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/conformance-kind.yaml
```diff
@@ -117,25 +117,6 @@ jobs:
           pkill -f "cilium.*hubble.*port-forward|kubectl.*port-forward.*hubble-relay"
           cilium uninstall --chart-directory=install/kubernetes/cilium --wait
 
-      - name: Install Cilium with encryption
```
brlbil, line 120: If the intention is to disable tests temporarily, then it makes more sense to add the line below
```
if: ${{ false }}
``` 
And please add a comment if there is an issue that tracks this, or anything relevant.

.github/workflows/tests-smoke.yaml
```diff
@@ -123,7 +123,7 @@ jobs:
              --namespace kube-system \
              --set nodeinit.enabled=true \
              --set kubeProxyReplacement=partial \
-             --set hostServices.enabled=false \
+             --set hostServices.enabled=true \
```
brlbil, line 129: Same here is it is a temporary change then a comment would be most helpful.

brlbil (changes requested):

aspsk:
No more tests required, as `ConformanceKind1.19` and `Smoke test` passed.

brlbil (approved):
Thanks, 👍

YutaroHayakawa (approved):
🙏

PR #22587, opened 2022-12-06
Validate "ownership" of hostPort service being deleted

yasz24:
This addresses a bug where deleting a completed pod deletes the service map entry for the corresponding running pod on the same node. This is due to the hostPort mapping keys being the 
same for both old and new pods.

Ideally we should validate whether the (completed) pod being deleted  
"owns" the host port service - to prevent breaking host port 
connectivity for any running pods with the same service as frontend.

This PR adds such a validation.

Testing-
Automated (Control Plane):
This fix is captured by a control plane test case that does the
following:
1. Create a hostport pod, and terminate it's running containers to mark
   it as "Completed".
2. Create another hostport pod using the same port as the "Completed" pod.
3. Delete the "Completed" pod, and verify that the hostport service has
   not been deleted in the Datapath.

Manual Testing -
1. Add the GracefulNodeShutdown in the kubelet config on all nodes by
   modifying the configuration in `/var/lib/kubelet/config.yaml`
   ```
    featureGates:
      GracefulNodeShutdown: true
    shutdownGracePeriod: 30s
    shutdownGracePeriodCriticalPods: 10s
   ```
2. Run `sudo systemctl restart kubelet` on each node to apply the kubelet
   config change
3. Deploy an nginx web server with hostPort set, as well as a
   nodeSelector, so pods get scheduled on the same node after node
   restarts.
   ```
   apiVersion: apps/v1
   kind: Deployment
   metadata:
     name: nginx-deployment
   spec:
     selector:
       matchLabels:
         app: nginx
    replicas: 1
    template:
      metadata:
        labels:
          app: nginx
      spec:
        nodeSelector:
	  kubernetes.io/hostname: <node-name>
	containers:
	- name: nginx
          image: nginx:1.14.2
          ports:
          - containerPort: 80
            hostPort: 8081
   ```
4. Run `systemctl reboot` on the worker node to restart the machine.
5. After reboot spot the old pod in `Completed state`, while the new pod
   is `Running`.
   ```
   $ kubectl get pods
   NAME                                READY   STATUS      RESTARTS    AGE
   nginx-deployment-645797c867-8p2hp   0/1     Completed   0           13m
   nginx-deployment-645797c867-dx2m8   1/1     Running     0           4m2s
   ```
6. `curl nodeIP:hostPort` successfully get the result.
7. Manually deleted the old pod which is in Completed state.
   ```
   $ kubectl delete pod/nginx-deployment-645797c867-8p2hp
   ```
[...]

Issue #19571, opened 2022-04-26
Review all usages of `fmt.Sprintf()` or similar `fmt` function inside hot paths


christarazi:
`fmt.Sprintf()` is well-known for not being opimal in terms of CPU and memory allocations. If these functions are called in hot paths such as `pkg/policy/api` or even parts of `pkg/policy`, it can be quite impactful, especially in environments that experience high or bursty churn. 

Writing a benchmark and replacing the current code with `strings.Builder` or `bytes.NewBuffer` can show us if we can achieve some quick wins.

Some examples: 

* https://github.com/cilium/cilium/pull/19570
* https://github.com/cilium/cilium/pull/19423

MikeLing:
Hi @christarazi ,I'm new to cilium and I think this issue can be my starting point about cilium contribution. Can you please assign this issue to me. Thanks

MikeLing:
I just change some code in `decision.go` and the benchmark output like:
<img width="622" alt="截屏2022-04-30 下午4 01 03" src="https://user-images.githubusercontent.com/2166227/166097322-11100868-3a96-4123-a7bf-47c0778f3653.png">

Does it make sense to you to optimize? I will commit PR if so.
Thank you

christarazi:
Splitting makes sense to me.

MikeLing:
@christarazi Does this commit make sense to you? https://github.com/MikeLing/cilium/commit/b86b803fdd1b08411a4a4c48d808fe13c0439baa

christarazi:
Yes please submit a PR.

MikeLing:
> Yes please submit a PR.

Done ;) 

https://github.com/cilium/cilium/pull/19822

MikeLing:
Hi @christarazi , these 3 PRs (https://github.com/cilium/cilium/pull/19822, https://github.com/cilium/cilium/pull/19704, https://github.com/cilium/cilium/pull/19661) have been approved. Could we merge them although some CI tests failed due to container timeout? 

Just want to avoid too many PR pending parallel, so I can head to other part of code. Thank you

christarazi:
Please rebase all PRs and ping me so I can retrigger the tests. Sorry, CI has not been very stable lately.

MikeLing:
> Please rebase all PRs and ping me so I can retrigger the tests. Sorry, CI has not been very stable lately.

No problem :D, I just rebase and push all of them.

MikeLing:
Hmmm, the CI failed again :(

[...]

PR #25986, opened 2023-06-07
Modular daemon and operator

pippolo84:
Current implementation of both the agent and the operator uses global instances of both cobra.Command and hive.
This makes the agent and the operator less flexible, especially when composing the application externally (e.g: for testing purposes, like in the controlplane test framework).

This PR removes the usage of globals and allows building the agent and operator command with constructors, passing a reference to an external hive.

The removal of the global Viper instance has consequences in other part of the codebase that are also addressed in this PR:

- refactoring of the agent debuginfo API handler that list all daemon flags and their values
- refactoring of the controlplane test framework to separate agent and operator configuration for the tests
- refactoring of the provider-specific flags for the operator variants

Notes for the reviewers: review commit by commit might be simpler.

### DIFF
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -405,11 +405,6 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 
 	bootstrapStats.daemonInit.Start()
 
-	// Validate the daemon-specific global options.
-	if err := option.Config.Validate(Vp); err != nil {
-		return nil, nil, fmt.Errorf("invalid daemon configuration: %s", err)
-	}
-
 	// Validate configuration options that depend on other cells.
 	if option.Config.IdentityAllocationMode == option.IdentityAllocationModeCRD && !params.Clientset.IsEnabled() &&
 		option.Config.DatapathMode != datapathOption.DatapathModeLBOnly {
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -219,6 +219,9 @@ type Daemon struct {
 
 	// statedb for implementing /statedb/dump
 	db statedb.DB
+
+	// read-only map of all the hive settings
+	settings cellSettings
 }
 
 func (d *Daemon) initDNSProxyContext(size int) {
@@ -541,6 +544,7 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 		l2announcer:          params.L2Announcer,
 		l7Proxy:              params.L7Proxy,
 		db:                   params.DB,
+		settings:             params.Settings,
 	}
 
 	d.configModifyQueue = eventqueue.NewEventQueueBuffered("config-modify-queue", ConfigModifyQueueSize)
--- a/daemon/cmd/debuginfo.go
+++ b/daemon/cmd/debuginfo.go
@@ -38,11 +38,14 @@ func getDebugInfoHandler(d *Daemon, params restapi.GetDebuginfoParams) middlewar
 	dr.CiliumMemoryMap = memoryMap(os.Getpid())
 
 	dr.EnvironmentVariables = []string{}
-	for _, k := range Vp.AllKeys() {
-		// Assuming we are only getting strings
-		v := fmt.Sprintf("%s:%s", k, Vp.GetString(k))
-		dr.EnvironmentVariables = append(dr.EnvironmentVariables, v)
-	}
+	// The debuginfo handler relies on the global Viper instance to list
+	// the agent settings and their values. Thus, this is temporarily commented
+	// out and will be fixed in a subsequent commit.
+	// for _, k := range Vp.AllKeys() {
+	// 	// Assuming we are only getting strings
+	// 	v := fmt.Sprintf("%s:%s", k, Vp.GetString(k))
+	// 	dr.EnvironmentVariables = append(dr.EnvironmentVariables, v)
+	// }
 
 	dr.ServiceList = getServiceList(d.svc)
 
--- a/daemon/cmd/debuginfo.go
+++ b/daemon/cmd/debuginfo.go
@@ -8,11 +8,13 @@ import (
 	"os"
 
 	"github.com/go-openapi/runtime/middleware"
+	"github.com/spf13/cast"
 
 	"github.com/cilium/cilium/api/v1/models"
 	restapi "github.com/cilium/cilium/api/v1/server/restapi/daemon"
 	"github.com/cilium/cilium/api/v1/server/restapi/endpoint"
 	"github.com/cilium/cilium/pkg/debug"
+	"github.com/cilium/cilium/pkg/hive/cell"
 	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/version"
 )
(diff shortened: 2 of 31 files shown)


test/controlplane/suite/testcase.go
```diff
@@ -123,7 +122,7 @@ func (cpt *ControlPlaneTest) SetupEnvironment(modConfig func(*agentOption.Daemon
 	agentOption.Config.EnableHealthCheckNodePort = false
 	agentOption.Config.Debug = true
 
-	operatorOption.Config.Populate(operatorCmd.Vp)
+	// operatorOption.Config.Populate(operatorCmd.Vp)
```
christarazi: Remove?

test/controlplane/suite/testcase.go
```diff
@@ -164,7 +163,7 @@ func (cpt *ControlPlaneTest) StartOperator(modCellConfig func(vp *viper.Viper))
 			cell.Provide(func() k8sClient.Clientset {
 				return cpt.clients
 			}),
-			operatorCmd.OperatorCell,
+			// operatorCmd.OperatorCell,
```
christarazi: Remove?

daemon/cmd/debuginfo.go
```diff
@@ -48,11 +48,12 @@ func (h *getDebugInfo) Handle(params restapi.GetDebuginfoParams) middleware.Resp
 	dr.CiliumMemoryMap = memoryMap(os.Getpid())
 
 	dr.EnvironmentVariables = []string{}
-	for _, k := range Vp.AllKeys() {
-		// Assuming we are only getting strings
-		v := fmt.Sprintf("%s:%s", k, Vp.GetString(k))
-		dr.EnvironmentVariables = append(dr.EnvironmentVariables, v)
-	}
+	// FIXME
```
christarazi: :slightly_smiling_face:

test/controlplane/suite/testcase.go
```diff
@@ -103,7 +102,7 @@ func (cpt *ControlPlaneTest) SetupEnvironment(modConfig func(*agentOption.Daemon
 	// Configure k8s and perform capability detection with the fake client.
 	version.Update(cpt.clients, true)
 
-	agentOption.Config.Populate(agentCmd.Vp)
+	//agentOption.Config.Populate(agentCmd.Vp)
```
christarazi: Remove?

daemon/cmd/debuginfo.go
```diff
@@ -48,12 +50,11 @@ func (h *getDebugInfo) Handle(params restapi.GetDebuginfoParams) middleware.Resp
...
-	// for _, k := range Vp.AllKeys() {
-	// 	// Assuming we are only getting strings
-	// 	v := fmt.Sprintf("%s:%s", k, Vp.GetString(k))
-	// 	dr.EnvironmentVariables = append(dr.EnvironmentVariables, v)
-	// }
+
+	for k, v := range d.settings {
+		envValue := fmt.Sprintf("%s:%s", k, v)
+		dr.EnvironmentVariables = append(dr.EnvironmentVariables, envValue)
+	}
```
christarazi: I see you've resolved it here, but can we just remove the extra diff from the previous commit?

daemon/cmd/debuginfo.go
```diff
@@ -48,12 +50,11 @@ func (h *getDebugInfo) Handle(params restapi.GetDebuginfoParams) middleware.Resp
...
 	dr.EnvironmentVariables = []string{}
-	// FIXME
-	// for _, k := range Vp.AllKeys() {
-	// 	// Assuming we are only getting strings
-	// 	v := fmt.Sprintf("%s:%s", k, Vp.GetString(k))
-	// 	dr.EnvironmentVariables = append(dr.EnvironmentVariables, v)
-	// }
+
+	for k, v := range d.settings {
+		envValue := fmt.Sprintf("%s:%s", k, v)
```
christarazi: Let's avoid `fmt.Sprintf` here due to #19571

daemon/cmd/daemon.go
```diff
@@ -217,6 +217,8 @@ type Daemon struct {
 	cniConfigManager cni.CNIConfigManager
 
 	l2announcer *l2announcer.L2Announcer
+
+	settings cellSettings
```
christarazi: Given the name `settings` is quite generic, it's not obvious what is represents. Can you add a godoc?

test/controlplane/suite/testcase.go
pippolo84: This can actually be changed to `operatorCmd.ControlPlane` in the first commit, I'm gonna fix :+1:

daemon/cmd/debuginfo.go
pippolo84: Unfortunately we can't do that because of the removal of the global `Vp` variable (wouldn't compile).
What do you think about adding a more extended comment (for both the controlplane tests and the debuginfo handler) instead of a bare `// FIXME`, reporting what I already pointed out in the commit description?

The alternative would be to squash all the commits, but that would produce something too large, IMHO.

daemon/cmd/debuginfo.go
```diff
@@ -48,14 +51,14 @@ func (h *getDebugInfo) Handle(params restapi.GetDebuginfoParams) middleware.Resp
...
-	// 	v := fmt.Sprintf("%s:%s", k, Vp.GetString(k))
-	// 	dr.EnvironmentVariables = append(dr.EnvironmentVariables, v)
-	// }
+	var sb strings.Builder
+	for k, v := range d.settings {
+		sb.Reset()
+		sb.WriteString(k)
+		sb.WriteRune(':')
+		sb.WriteString(v)
+		dr.EnvironmentVariables = append(dr.EnvironmentVariables, sb.String())
```
christarazi: Why not

```
dr.EnvironmentVariables = append(dr.EnvironmentVariables, k+':'+v)
```

christarazi (approved):

daemon/cmd/debuginfo.go
pippolo84: Usually I go for the `strings.Builder` to avoid the extra allocation for each `+`. But I guess for this simple case `+` might be faster (and is also more readable). I'm gonna change it. :+1:

pippolo84:
@nathanjsweet rebased, could you PTAL? :pray: Thanks!

nathanjsweet (approved):

pippolo84:
ci-clustermesh hits a [known flake](https://github.com/cilium/cilium/pull/26662)

julianwiedmann:
@pippolo84 heads-up, looks like this might have introduced CI breakage in main: https://github.com/cilium/cilium/actions/runs/5483476340

julianwiedmann:
> @pippolo84 heads-up, looks like this might have introduced CI breakage in main: https://github.com/cilium/cilium/actions/runs/5483476340

ah nevermind, Daniel already spotted that it's https://github.com/cilium/cilium/pull/22587 (presumably not rebased on top of this PR?).

pippolo84:
> ah nevermind, Daniel already spotted that it's #22587 (presumably not rebased on top of this PR?).

I think so, the reason is that this PR changed the controlplane tests API (specifically, the commit [here](https://github.com/cilium/cilium/pull/25986/commits/9c00b80a9bed4824a508b8986a969631f4161a7e)) to account for the removal of the globals.
Let me figure out if I can suggest the fix to the author of https://github.com/cilium/cilium/pull/22587

PR #25991, opened 2023-06-07
ipam: Add ability to automatically create `CiliumPodIPPool` resources in multi-pool IPAM mode

gandro:
Before this PR, we populated the multi-pool allocator from pre-defined pools based on the `multi-pool-map` flag. Since we added a custom `CiliumPodIPPool` resource however (c.f. #25824), there was a mismatch between those pre-defined pools (which are static and not backed up by a resource) and pools defined via CRD.

This PR addresses this mismatch by instead creating a `CiliumPodIPPool` resource for every pre-defined pool present in `auto-create-cilium-pod-ip-pools`. This means that those pre-defined pools are now regular pools that are automatically created upon operator startup. The name of the new flag is taken from the `auto-create-cilium-node-resource` flag which also creates resources upon startup.

Having the option to pre-define pools at installation time is important, because cilium-agent requires pod CIDRs at least from the default pool before it becomes ready. By providing an option to have cilium-operator auto-create some pre-defined pools, we solve this bootstrapping problem in an easy way. The alternative would require users to manually create
default pools after running `helm install`, which increases operational complexity.

Note that if the pre-defined pools already exist, the operator will not modify re-create or update them. This behavior is intentional, to avoid overwriting potential changes made by the user.

Meta: #25470 

:information_source: **Review per commit**

Example values.yaml:
```yaml
ipam:
    mode: multi-pool
    operator:
        autoCreateCiliumPodIPPools:
            default:
              ipv4:
                cidrs:
                  - 10.10.0.0/16
                maskSize: 24
            mars:
              ipv4:
                cidrs:
                  - 10.20.0.0/16
                maskSize: 24
            other:
              ipv6:
                cidrs:
                  - fd00:100::/80
                maskSize: 96

debug:
    enabled: true

# Kind values
image:
    override: localhost:5000/cilium/cilium-dev:local
    pullPolicy: IfNotPresent
operator:
    replicas: 1
    image:
        override: localhost:5000/cilium/operator-generic:local
        pullPolicy: IfNotPresent

# Datapath requirements for multi-pool
tunnel: disabled
autoDirectNodeRoutes: true
ipv4NativeRoutingCIDR: 10.10.0.0/8
endpointRoutes:
    enabled: true
extraConfig:
    enable-local-node-route: "false"
kubeProxyReplacement: "strict"
bpf:
    masquerade: true
```

Will create three IP pools (default, mars and other), as can be seen with:

```console
$ kubectl get cpip -o yaml   # abbreviated output
apiVersion: v1
items:
- apiVersion: cilium.io/v2alpha1
  kind: CiliumPodIPPool
  metadata:
    name: default
  spec:
    ipv4:
      cidrs:
      - 10.10.0.0/16
      maskSize: 24
- apiVersion: cilium.io/v2alpha1
  kind: CiliumPodIPPool
  metadata:
    name: mars
  spec:
    ipv4:
      cidrs:
      - 10.20.0.0/16
      maskSize: 24
- apiVersion: cilium.io/v2alpha1
  kind: CiliumPodIPPool
  metadata:
    name: other
  spec:
    ipv6:
      cidrs:
      - fd00:100::/80
      maskSize: 96
kind: List
metadata:
  resourceVersion: ""

```

Proper CI and documentation will be added in a follow-up PR.


pkg/ipam/allocator/multipool/allocator.go
```diff
@@ -5,131 +5,66 @@ package multipool
...
-	poolKeyIPv6CIDRs    = "ipv6-cidrs"
-	poolKeyIPv6MaskSize = "ipv6-mask-size"
-)
-
-// parsePoolSpec parses a pool spec string in the form
-// "ipv4-cidrs:172.16.0.0/16,172.17.0.0/16;ipv4-mask-size:24".
-func parsePoolSpec(poolString string) (poolSpec, error) {
-	fields := strings.Split(strings.ReplaceAll(poolString, " ", ""), ";")
-
-	var pool poolSpec
```
qmonnet, line 45: GitHub says (on the diff for the last commit):

image

qmonnet (approved):
Looks good as far as I'm concerned, there's one report to address below

pkg/ipam/allocator/multipool/allocator.go
gandro: Already fixed, thanks!

operator/watchers/cilium_podippool.go
```diff
@@ -19,6 +28,107 @@ type PooledAllocatorProvider interface {
...
+				Name: poolName,
+			},
+			Spec: poolSpec,
+		}
+
+		_, err = clientset.CiliumV2alpha1().CiliumPodIPPools().Create(ctx, pool, metav1.CreateOptions{})
+		if err != nil {
+			if k8sErrors.IsAlreadyExists(err) {
+				// Nothing to do, we will not try to update an existing resource
+				log.WithField("poolName", poolName).Debug("CiliumPodIPPool resource already exists")
```
tklauser: I wonder whether this should be `Info` or even `Warning` such that it is more visible to users?

operator/watchers/cilium_podippool.go
```diff
@@ -27,6 +137,10 @@ func StartIPPoolAllocator(
 ) {
 	log.Info("Starting CiliumPodIPPool allocator watcher")
 
+	if len(operatorOption.Config.IPAMAutoCreateCiliumPodIPPools) > 0 {
```
tklauser: I think this check is redundant because `multiPoolAutoCreatePools` will range over the map at the top level, i.e. it's a no-op in case of an empty map. Might still be good to keep an explicit check in case we change ever `multiPoolAutoCreatePools` to perform other operations.

pkg/ipam/allocator/multipool/allocator.go
```diff
@@ -5,131 +5,66 @@ package multipool
...
-			continue
+func (a *Allocator) UpsertPool(ctx context.Context, pool *cilium_v2alpha1.CiliumPodIPPool) error {
+	var ipv4CIDRs, ipv6CIDRs []string
+	var ipv4MaskSize, ipv6MaskSize int
+
+	if pool.Spec.IPv4 != nil {
+		ipv4MaskSize = int(pool.Spec.IPv4.MaskSize)
+		ipv4CIDRs = make([]string, len(pool.Spec.IPv4.CIDRs))
+		for i, cidr := range pool.Spec.IPv4.CIDRs {
+			ipv4CIDRs[i] = string(cidr)
```
tklauser, line 42: Maybe as a follow-up: Given that we pass `ipv4CIDRs` to `a.poolAlloc.UpsertPool` only and that method then converts its entries to `netip.Prefix` and `net.IPNet` later on (also ref. https://github.com/cilium/cilium/pull/25824#discussion_r1218257180.). I wonder whether we should just directly create a `[]netip.Prefix` here and pass that. This would also have the benefit that we could report CIDR format errors earlier. 

Same applies for `ipv6CIDRs`.

tklauser (approved):
Nice work!

Very minor comments and a suggestion for a follow-up change inline.

install/kubernetes/cilium/templates/cilium-configmap.yaml
```diff
@@ -875,19 +875,19 @@ data:
...
-    {{- $attrs = append $attrs (printf "ipv6-cidrs:%s" (join "," $cidrs.ipv6CIDRs)) }}
-    {{- $attrs = append $attrs (printf "ipv6-mask-size:%s" (toString $cidrs.ipv6MaskSize)) }}
+    {{- if hasKey $spec "ipv6" }}
+    {{- $attrs = append $attrs (printf "ipv6-cidrs:%s" (join "," $spec.ipv6.cidrs)) }}
+    {{- $attrs = append $attrs (printf "ipv6-mask-size:%s" (toString $spec.ipv6.maskSize)) }}
     {{- end }}
     {{- $pools = append $pools (printf "%s=%s" $pool (join ";" $attrs)) }}
   {{- end }}
-  multi-pool-map: {{ join "," $pools | quote }}
+  auto-create-cilium-pod-ip-pools: {{ join "," $pools | quote }}
```
tommyp1ckles, line 890: Maybe we should have some upgrade notes on this, in case someone is doing something like managing this themselves (not sure if thats a reasonable expectation?)

operator/watchers/cilium_podippool.go
```diff
@@ -19,6 +28,107 @@ type PooledAllocatorProvider interface {
...
+	poolKeyIPv4CIDRs    = "ipv4-cidrs"
+	poolKeyIPv4MaskSize = "ipv4-mask-size"
+	poolKeyIPv6CIDRs    = "ipv6-cidrs"
+	poolKeyIPv6MaskSize = "ipv6-mask-size"
+)
+
+// parsePoolSpec parses a pool spec string in the form
+// "ipv4-cidrs:172.16.0.0/16,172.17.0.0/16;ipv4-mask-size:24".
+func parsePoolSpec(poolString string) (cilium_v2alpha1.IPPoolSpec, error) {
+	fields := strings.Split(strings.ReplaceAll(poolString, " ", ""), ";")
```
tommyp1ckles: strings.Split doesn't do well in [edge cases](https://go.dev/play/p/I4VLrKjCQ8M) in cases like this, perhaps it'd be better to use strings.FieldsFunc to make this more robust?:

```go
fields := strings.FieldsFunc(strings.ReplaceAll(poolString, " ", ""), func(c rune) {
    return c == ';'
})
```

operator/watchers/cilium_podippool.go
```diff
@@ -19,6 +28,107 @@ type PooledAllocatorProvider interface {
...
+		key, value, ok := strings.Cut(field, ":")
+		if !ok {
+			return cilium_v2alpha1.IPPoolSpec{}, fmt.Errorf("invalid number of key delimiters in pool spec %s", poolString)
+		}
+		switch key {
+		case poolKeyIPv4CIDRs:
+			for _, cidr := range strings.Split(value, ",") {
+				_, err := netip.ParsePrefix(cidr)
+				if err != nil {
+					return cilium_v2alpha1.IPPoolSpec{}, fmt.Errorf("invalid value for key %q: %w", poolKeyIPv4CIDRs, err)
```
tommyp1ckles, line 58: For errors such as this one (and similar below), it might be helpful for someone reading logs if you specify what kind of validation failed.

operator/watchers/cilium_podippool.go
```diff
@@ -19,6 +28,107 @@ type PooledAllocatorProvider interface {
...
+				}
+				ipv4CIDRs = append(ipv4CIDRs, cilium_v2alpha1.PoolCIDR(cidr))
+			}
+		case poolKeyIPv4MaskSize:
+			mask, err := strconv.ParseUint(value, 10, 64)
```
tommyp1ckles: not a real point, but its a bit odd we use uint64 for these field.

christarazi (approved):

install/kubernetes/cilium/templates/cilium-configmap.yaml
gandro: Ah. I should have mentioned that in the commit message. This flag was never part of a stable release, so no upgrade notes are needed.

operator/watchers/cilium_podippool.go
gandro: I don't think it should be a warning. This is expected to happen if the operator restarts. Even a `info` is a bit noisy, which is why I made it a debug log statement. But I'll change it to Info again, to avoid any confusion as to why objects are not touched once they have been created.

operator/watchers/cilium_podippool.go
gandro: Great suggestion, will add!

operator/watchers/cilium_podippool.go
gandro: Removed the redundant if.

operator/watchers/cilium_podippool.go
gandro: Yeah true. I'll change it to a smaller int type before it's too late.

operator/watchers/cilium_podippool.go
gandro: Not sure what you mean by that. It currently prints the following error if this line is hit, what do you think is missing?

```
level=fatal msg="Failed to parse IP pool spec in \"auto-create-cilium-pod-ip-pools\" flag" error="invalid value for key \"ipv4-cidrs\": netip.ParsePrefix(\"172.300.0.0/16\"): ParseAddr(\"172.300.0.0\"): IPv4 field has value >255" poolName=broken poolSpec="ipv4-cidrs:172.300.0.0/16,172.17.0.0/16;ipv4-mask-size:24" subsys=watchers
```

operator/watchers/cilium_podippool.go
gandro: Changed to a uint8.

pkg/ipam/allocator/multipool/allocator.go
gandro: Good idea. Doing it properly requires seems to require a bit more changes to `cidralloc`, so I'll do that in a follow-up PR.

gandro:
I believe I addressed all nits. Rebased and restarting CI

pkg/ipam/allocator/multipool/allocator.go
tklauser: I have a semi-ready PR changing this (and some other cleanups). I'll rebase on top of this PR's branch and open the PR once this has been merged.

operator/watchers/cilium_podippool.go
tklauser: Good point. I didn't consider that this will be hit frequently during restart. Changing to Info or leaving as Debug both SGTM.

gandro:
- test-1.26-net-next had a VM provisioning error (failed to fetch docker image) https://jenkins.cilium.io/job/Cilium-PR-K8s-1.26-kernel-net-next/708/console
- ci-ginkgo failed with `invalid value: 31414: provided port is already allocated` (e.g. https://github.com/cilium/cilium/issues/13071) (https://github.com/cilium/cilium/actions/runs/5245376421/jobs/9472783968)
- ci-ginko also failed with with a provisioning error (https://github.com/cilium/cilium/actions/runs/5245376421/jobs/9472786521)
- ci-ginko also failed with https://github.com/cilium/cilium/issues/25964 (https://github.com/cilium/cilium/actions/runs/5245376421/jobs/9472788454)
- ci-aks failed with an connectivity infra flake between cilium-cli and kube-apiserver `error dialing backend: write unix @->/tunnel-uds/proxysocket: write: broken pipe`, this is the same root cause as [#26032](https://github.com/cilium/cilium/issues/26032), [#26034](https://github.com/cilium/cilium/issues/26034), [#26014](https://github.com/cilium/cilium/issues/26014)

Restarting all failed tests

gandro:
/test-1.26-net-next

gandro:
CI-net-next hit #25958

gandro:
/test-1.26-net-next

sayboras (approved):
the multiPoolMap attribute in helm is added as part of the 1.14 dev cycle, so we can skip the deprecation step for renaming. 

Helm changes lgtm :heavy_check_mark:

pkg/ipam/allocator/multipool/allocator.go
tklauser: Done in https://github.com/cilium/cilium/pull/26138

PR #25994, opened 2023-06-07
Module Health: Node Manager: First Iteration

tommyp1ckles:
This PR seeks to provide modular health status checking on the "node manager" component of the Agent. Specifically the goals are:

### Bubble-up all errors in Linux node handler code

Node Manager related code relies heavily on logging (or just ignoring) errors far down the stack, instead of bubbling them up.

The result of this is:
* Logged errors have little context on the root event causing them to be invoked.
* It's hard to trace failure from a particular event relating to all reconcile procedures that need to run.
* It's hard to definitively to say if reconciliation for a Node Event was successfully.

The changes in this PR seek to address this by handling and bubbling up all relevant errors that may be encountered while doing Linux Node update reconciliation logic.

As well, this PR seeks to **avoid** any changes to the control-flow logic of the reconciliation by _not_ immediately returning at places where errors may occur.
Rather, this follows a pattern of combining lists of errors into relevant/scoped multi errors (using the multierr library). All code will execute as before, with the difference being that errors are handled up the stack.

### Use Hive's Modular Health Reporter to report status of node manager.

This is a first attempt at trying to manage health status for a module. 

#### Follow up work:
* Investigate Reconciliation: Break Linux node logic into atomic and idempotent tasks that can be executed and retried (+ reported on more easily).

### DIFF
--- a/pkg/datapath/linux/ipsec/ipsec_linux.go
+++ b/pkg/datapath/linux/ipsec/ipsec_linux.go
@@ -9,6 +9,7 @@ import (
 	"bufio"
 	"context"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -536,7 +537,8 @@ func matchesOnNodeID(mark *netlink.XfrmMark) bool {
 		mark.Mask&linux_defaults.IPsecMarkMaskNodeID == linux_defaults.IPsecMarkMaskNodeID
 }
 
-func ipsecDeleteXfrmState(nodeID uint16) {
+func ipsecDeleteXfrmState(nodeID uint16) error {
+	var errs error
 	scopedLog := log.WithFields(logrus.Fields{
 		logfields.NodeID: nodeID,
 	})
@@ -544,18 +546,20 @@ func ipsecDeleteXfrmState(nodeID uint16) {
 	xfrmStateList, err := netlink.XfrmStateList(netlink.FAMILY_ALL)
 	if err != nil {
 		scopedLog.WithError(err).Warning("Failed to list XFRM states for deletion")
-		return
+		return err
 	}
 	for _, s := range xfrmStateList {
 		if matchesOnNodeID(s.Mark) && getNodeIDFromXfrmMark(s.Mark) == nodeID {
 			if err := netlink.XfrmStateDel(&s); err != nil {
 				scopedLog.WithError(err).Warning("Failed to delete XFRM state")
+				errs = errors.Join(errs, fmt.Errorf("failed to delete xfrm state (%s): %w", s.String(), err))
 			}
 		}
 	}
+	return errs
 }
 
-func ipsecDeleteXfrmPolicy(nodeID uint16) {
+func ipsecDeleteXfrmPolicy(nodeID uint16) error {
 	scopedLog := log.WithFields(logrus.Fields{
 		logfields.NodeID: nodeID,
 	})
@@ -563,15 +567,18 @@ func ipsecDeleteXfrmPolicy(nodeID uint16) {
 	xfrmPolicyList, err := netlink.XfrmPolicyList(netlink.FAMILY_ALL)
 	if err != nil {
 		scopedLog.WithError(err).Warning("Failed to list XFRM policies for deletion")
-		return
+		return fmt.Errorf("failed to list xfrm policies: %w", err)
 	}
+	var errs error
 	for _, p := range xfrmPolicyList {
 		if matchesOnNodeID(p.Mark) && getNodeIDFromXfrmMark(p.Mark) == nodeID {
 			if err := netlink.XfrmPolicyDel(&p); err != nil {
 				scopedLog.WithError(err).Warning("Failed to delete XFRM policy")
+				errs = errors.Join(errs, fmt.Errorf("failed to delete xfrm policy (%s): %w", p.String(), err))
 			}
 		}
 	}
+	return errs
 }
 
 /* UpsertIPsecEndpoint updates the IPSec context for a new endpoint inserted in
@@ -670,9 +677,8 @@ func UpsertIPsecEndpointPolicy(local, remote *net.IPNet, localTmpl, remoteTmpl n
 }
 
 // DeleteIPsecEndpoint deletes a endpoint associated with the remote IP address
-func DeleteIPsecEndpoint(nodeID uint16) {
-	ipsecDeleteXfrmState(nodeID)
-	ipsecDeleteXfrmPolicy(nodeID)
+func DeleteIPsecEndpoint(nodeID uint16) error {
+	return errors.Join(ipsecDeleteXfrmState(nodeID), ipsecDeleteXfrmPolicy(nodeID))
 }
 
 func isXfrmPolicyCilium(policy netlink.XfrmPolicy) bool {
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -1232,6 +1232,14 @@ func (n *linuxNodeHandler) nodeUpdate(oldNode, newNode *nodeTypes.Node, firstAdd
 		// assumptions:
 		// 1. newNode is accessed only by reads.
 		// 2. It is safe to invoke insertNeighbor for the same node.
+		//
+		// Because neighbor inserts is not synced, we do not currently
+		// collect/ bubble up errors.
+		// Instead we just rely on logging errors as they come up in the
+		// neighbor update procedure.
+		//
+		// In v1.15, we will have the neighbor sync component report its own
+		// health via stored errors. 
 		go n.insertNeighbor(context.Background(), newNode, false)
 	}
 
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -1207,8 +1207,11 @@ func (n *linuxNodeHandler) nodeUpdate(oldNode, newNode *nodeTypes.Node, firstAdd
 		newIP6                                   = newNode.GetNodeIP(true)
 		oldKey, newKey                           uint8
 		isLocalNode                              = false
-		remoteNodeID                             = n.allocateIDForNode(newNode)
 	)
+	remoteNodeID, err := n.allocateIDForNode(newNode)
+	if err != nil {
+		errs = errors.Join(errs, fmt.Errorf("failed to allocate ID for node %s: %w", newNode.Name, err))
+	}
 
 	if oldNode != nil {
 		oldIP4Cidr = oldNode.IPv4AllocCIDR
@@ -1239,7 +1242,7 @@ func (n *linuxNodeHandler) nodeUpdate(oldNode, newNode *nodeTypes.Node, firstAdd
 		// neighbor update procedure.
 		//
 		// In v1.15, we will have the neighbor sync component report its own
-		// health via stored errors. 
+		// health via stored errors.
 		go n.insertNeighbor(context.Background(), newNode, false)
 	}
 
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -107,6 +107,10 @@ func NewNodeHandler(datapathConfig DatapathConfiguration, nodeAddressing datapat
 	}
 }
 
+func (l *linuxNodeHandler) Name() string {
+	return "linux-node-datapath"
+}
+
 // updateTunnelMapping is called when a node update is received while running
 // with encapsulation mode enabled. The CIDR and IP of both the old and new
 // node are provided as context. The caller expects the tunnel mapping in the
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -1400,6 +1400,7 @@ func (n *linuxNodeHandler) NodeDelete(oldNode nodeTypes.Node) error {
 
 // Must be called with linuxNodeHandler.mutex held.
 func (n *linuxNodeHandler) nodeDelete(oldNode *nodeTypes.Node) error {
+	var errs error
 	if oldNode.IsLocal() {
 		return nil
 	}
@@ -1408,19 +1409,31 @@ func (n *linuxNodeHandler) nodeDelete(oldNode *nodeTypes.Node) error {
 	oldIP6 := oldNode.GetNodeIP(true)
 
 	if n.nodeConfig.EnableAutoDirectRouting {
-		n.deleteDirectRoute(oldNode.IPv4AllocCIDR, oldIP4)
-		n.deleteDirectRoute(oldNode.IPv6AllocCIDR, oldIP6)
+		if err := n.deleteDirectRoute(oldNode.IPv4AllocCIDR, oldIP4); err != nil {
+			errs = errors.Join(errs, fmt.Errorf("failed to remove old direct routing: deleting old routes %w", err))
+		}
+		if err := n.deleteDirectRoute(oldNode.IPv6AllocCIDR, oldIP6); err != nil {
+			errs = errors.Join(errs, fmt.Errorf("failed to remove old direct routing: deleting old routes %w", err))
+		}
 	}
 
 	if n.nodeConfig.EnableEncapsulation {
 		oldPrefix4 := cmtypes.PrefixClusterFromCIDR(oldNode.IPv4AllocCIDR, oldNode.ClusterID)
 		oldPrefix6 := cmtypes.PrefixClusterFromCIDR(oldNode.IPv6AllocCIDR, oldNode.ClusterID)
-		deleteTunnelMapping(oldPrefix4, false)
-		deleteTunnelMapping(oldPrefix6, false)
+		if err := deleteTunnelMapping(oldPrefix4, false); err != nil {
+			errs = errors.Join(errs, fmt.Errorf("failed to remove old encapsulation config: deleting tunnel mapping for ipv4: %w", err))
+		}
+		if err := deleteTunnelMapping(oldPrefix6, false); err != nil {
+			errs = errors.Join(errs, fmt.Errorf("failed to remove old encapsulation config: deleting tunnel mapping for ipv6: %w", err))
+		}
 
 		if !n.nodeConfig.UseSingleClusterRoute {
-			n.deleteNodeRoute(oldNode.IPv4AllocCIDR, false)
-			n.deleteNodeRoute(oldNode.IPv6AllocCIDR, false)
+			if err := n.deleteNodeRoute(oldNode.IPv4AllocCIDR, false); err != nil {
+				errs = errors.Join(errs, fmt.Errorf("failed to remove old encapsulation config: deleting old single cluster node route for ipv4: %w", err))
+			}
+			if err := n.deleteNodeRoute(oldNode.IPv6AllocCIDR, false); err != nil {
+				errs = errors.Join(errs, fmt.Errorf("failed to remove old encapsulation config: deleting old single cluster node route for ipv6: %w", err))
+			}
 		}
 	}
 
@@ -1429,12 +1442,16 @@ func (n *linuxNodeHandler) nodeDelete(oldNode *nodeTypes.Node) error {
 	}
 
 	if n.nodeConfig.EnableIPSec {
-		n.deleteIPsec(oldNode)
+		if err := n.deleteIPsec(oldNode); err != nil {
+			errs = errors.Join(errs, fmt.Errorf("failed to delete old ipsec config: %w", err))
+		}
 	}
 
-	n.deallocateIDForNode(oldNode)
+	if err := n.deallocateIDForNode(oldNode); err != nil {
+		errs = errors.Join(errs, fmt.Errorf("failed to deallocate old node ID: %w", err))
+	}
 
-	return nil
+	return errs
 }
 
 func (n *linuxNodeHandler) updateOrRemoveClusterRoute(addressing datapath.NodeAddressingFamily, addressFamilyEnabled bool) {
@@ -1625,41 +1642,47 @@ func (n *linuxNodeHandler) replaceNodeExternalIPSecOutRoute(ip *net.IPNet) error
 }
 
 // The caller must ensure that the CIDR passed in must be non-nil.
-func (n *linuxNodeHandler) deleteNodeIPSecOutRoute(ip *net.IPNet) {
+func (n *linuxNodeHandler) deleteNodeIPSecOutRoute(ip *net.IPNet) error {
 	if ip.IP.To4() != nil {
 		if !n.nodeConfig.EnableIPv4 {
-			return
+			return nil
 		}
 	} else {
 		if !n.nodeConfig.EnableIPv6 {
-			return
+			return nil
 		}
 	}
 
 	if err := route.Delete(n.createNodeIPSecOutRoute(ip)); err != nil {
 		log.WithError(err).WithField(logfields.CIDR, ip).Error("Unable to delete the IPsec route OUT from the host routing table")
+		return fmt.Errorf("failed to delete ipsec host route out: %w", err)
 	}
+	return nil
 }
 
 // The caller must ensure that the CIDR passed in must be non-nil.
-func (n *linuxNodeHandler) deleteNodeExternalIPSecOutRoute(ip *net.IPNet) {
+func (n *linuxNodeHandler) deleteNodeExternalIPSecOutRoute(ip *net.IPNet) error {
+	var errs error
 	if ip.IP.To4() != nil {
 		if !n.nodeConfig.EnableIPv4 {
-			return
+			return nil
 		}
 	} else {
 		if !n.nodeConfig.EnableIPv6 {
-			return
+			return nil
 		}
 	}
 
 	if err := route.Delete(n.createNodeExternalIPSecOutRoute(ip, true)); err != nil {
 		log.WithError(err).WithField(logfields.CIDR, ip).Error("Unable to delete the IPsec route External OUT from the ipsec routing table")
+		errs = errors.Join(errs, fmt.Errorf("failed to delete ipsec route out: %w", err))
 	}
 
 	if err := route.Delete(n.createNodeExternalIPSecOutRoute(ip, false)); err != nil {
 		log.WithError(err).WithField(logfields.CIDR, ip).Error("Unable to delete the IPsec route External OUT from the host routing table")
+		errs = errors.Join(errs, fmt.Errorf("failed to delete ipsec host route out: %w", err))
 	}
+	return errs
 }
 
 // replaceNodeIPSecoInRoute replace the in IPSec routes in the host routing
@@ -1683,7 +1706,8 @@ func (n *linuxNodeHandler) replaceNodeIPSecInRoute(ip *net.IPNet) error {
 	return nil
 }
 
-func (n *linuxNodeHandler) deleteIPsec(oldNode *nodeTypes.Node) {
+func (n *linuxNodeHandler) deleteIPsec(oldNode *nodeTypes.Node) error {
+	var errs error
 	scopedLog := log.WithField(logfields.NodeName, oldNode.Name)
 	scopedLog.Debugf("Removing IPsec configuration for node")
 
@@ -1691,7 +1715,7 @@ func (n *linuxNodeHandler) deleteIPsec(oldNode *nodeTypes.Node) {
 	if nodeID == 0 {
 		scopedLog.Warning("No node ID found for node.")
 	} else {
-		ipsec.DeleteIPsecEndpoint(nodeID)
+		errs = errors.Join(errs, ipsec.DeleteIPsecEndpoint(nodeID))
 	}
 
 	if n.nodeConfig.EnableIPv4 && oldNode.IPv4AllocCIDR != nil {
@@ -1699,13 +1723,13 @@ func (n *linuxNodeHandler) deleteIPsec(oldNode *nodeTypes.Node) {
 		// This is only needed in IPAM modes where we install one route per
 		// remote pod CIDR.
 		if !n.subnetEncryption() {
-			n.deleteNodeIPSecOutRoute(old4RouteNet)
+			errs = errors.Join(errs, n.deleteNodeIPSecOutRoute(old4RouteNet))
 		}
 		if n.nodeConfig.EncryptNode {
 			if remoteIPv4 := oldNode.GetNodeIP(false); remoteIPv4 != nil {
 				exactMask := net.IPv4Mask(255, 255, 255, 255)
 				ipsecRemote := &net.IPNet{IP: remoteIPv4, Mask: exactMask}
-				n.deleteNodeExternalIPSecOutRoute(ipsecRemote)
+				errs = errors.Join(errs, n.deleteNodeExternalIPSecOutRoute(ipsecRemote))
 			}
 		}
 	}
@@ -1720,10 +1744,11 @@ func (n *linuxNodeHandler) deleteIPsec(oldNode *nodeTypes.Node) {
 			if remoteIPv6 := oldNode.GetNodeIP(true); remoteIPv6 != nil {
 				exactMask := net.CIDRMask(128, 128)
 				ipsecRemote := &net.IPNet{IP: remoteIPv6, Mask: exactMask}
-				n.deleteNodeExternalIPSecOutRoute(ipsecRemote)
+				errs = errors.Join(errs, n.deleteNodeExternalIPSecOutRoute(ipsecRemote))
 			}
 		}
 	}
+	return errs
 }
 
 // NodeConfigurationChanged is called when the LocalNodeConfiguration has changed
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -116,16 +116,20 @@ func (l *linuxNodeHandler) Name() string {
 // node are provided as context. The caller expects the tunnel mapping in the
 // datapath to be updated.
 func updateTunnelMapping(oldCIDR, newCIDR cmtypes.PrefixCluster, oldIP, newIP net.IP,
-	firstAddition, encapEnabled bool, oldEncryptKey, newEncryptKey uint8) {
+	firstAddition, encapEnabled bool, oldEncryptKey, newEncryptKey uint8) error {
+	var errs error
 	if !encapEnabled {
 		// When the protocol family is disabled, the initial node addition will
 		// trigger a deletion to clean up leftover entries. The deletion happens
 		// in quiet mode as we don't know whether it exists or not
 		if newCIDR.IsValid() && firstAddition {
-			deleteTunnelMapping(newCIDR, true)
+			if err := deleteTunnelMapping(newCIDR, true); err != nil {
+				errs = errors.Join(errs,
+					fmt.Errorf("failed to delete tunnel mapping %q: %w", newCIDR, err))
+			}
 		}
 
-		return
+		return errs
 	}
 
 	if cidrNodeMappingUpdateRequired(oldCIDR, newCIDR, oldIP, newIP, oldEncryptKey, newEncryptKey) {
@@ -138,6 +142,8 @@ func updateTunnelMapping(oldCIDR, newCIDR cmtypes.PrefixCluster, oldIP, newIP ne
 			log.WithError(err).WithFields(logrus.Fields{
 				"allocCIDR": newCIDR,
 			}).Error("bpf: Unable to update in tunnel endpoint map")
+			errs = errors.Join(errs,
+				fmt.Errorf("failed to update tunnel endpoint map (prefix: %s, nodeIP: %s): %w", newCIDR.AddrCluster(), newIP, err))
 		}
 	}
 
@@ -150,8 +156,12 @@ func updateTunnelMapping(oldCIDR, newCIDR cmtypes.PrefixCluster, oldIP, newIP ne
 		fallthrough
 	// Node allocation CIDR has changed
 	case oldCIDR.IsValid() && newCIDR.IsValid() && !oldCIDR.Equal(newCIDR):
-		deleteTunnelMapping(oldCIDR, false)
+		if err := deleteTunnelMapping(oldCIDR, false); err != nil {
+			errs = errors.Join(errs,
+				fmt.Errorf("failed to delete tunnel mapping (oldCIDR: %s, newIP: %s): %w", oldCIDR, newIP, err))
+		}
 	}
+	return errs
 }
 
 // cidrNodeMappingUpdateRequired returns true if the change from an old node
@@ -180,9 +190,9 @@ func cidrNodeMappingUpdateRequired(oldCIDR, newCIDR cmtypes.PrefixCluster, oldIP
 	return !oldCIDR.Equal(newCIDR)
 }
 
-func deleteTunnelMapping(oldCIDR cmtypes.PrefixCluster, quietMode bool) {
+func deleteTunnelMapping(oldCIDR cmtypes.PrefixCluster, quietMode bool) error {
 	if !oldCIDR.IsValid() {
-		return
+		return nil
 	}
 
 	log.WithFields(logrus.Fields{
@@ -197,10 +207,12 @@ func deleteTunnelMapping(oldCIDR cmtypes.PrefixCluster, quietMode bool) {
 			log.WithError(err).WithFields(logrus.Fields{
 				"allocPrefixCluster": oldCIDR.String(),
 			}).Error("Unable to delete in tunnel endpoint map")
+			return fmt.Errorf("failed to delete tunnel endpoint map: %w", err)
 		}
 	} else {
-		_ = tunnel.TunnelMap().SilentDeleteTunnelEndpoint(addrCluster)
+		return tunnel.TunnelMap().SilentDeleteTunnelEndpoint(addrCluster)
 	}
+	return nil
 }
 
 func createDirectRouteSpec(CIDR *cidr.CIDR, nodeIP net.IP) (routeSpec *netlink.Route, err error) {
@@ -276,12 +288,13 @@ func installDirectRoute(CIDR *cidr.CIDR, nodeIP net.IP) (routeSpec *netlink.Rout
 }
 
 func (n *linuxNodeHandler) updateDirectRoutes(oldCIDRs, newCIDRs []*cidr.CIDR, oldIP, newIP net.IP, firstAddition, directRouteEnabled bool) error {
+
 	if !directRouteEnabled {
 		// When the protocol family is disabled, the initial node addition will
 		// trigger a deletion to clean up leftover entries. The deletion happens
 		// in quiet mode as we don't know whether it exists or not
 		if firstAddition {
-			n.deleteAllDirectRoutes(newCIDRs, newIP)
+			return n.deleteAllDirectRoutes(newCIDRs, newIP)
 		}
 		return nil
 	}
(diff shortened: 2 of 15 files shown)


tommyp1ckles:
@timoreimann @chancez 

I was running into some issues with seeing my force-pushed changes on the last PR, when I closed it it no longer allowed me to reopen that one so I had to fully recreate it 😵‍💫 

Please refer to some of my previous comments on the last PR: https://github.com/cilium/cilium/pull/25912

chancez (approved):
Hubble side looks good.

tommyp1ckles:
/test-1.26-net-next

pkg/datapath/linux/node.go
```diff
@@ -500,7 +500,7 @@ func (n *linuxNodeHandler) NodeUpdate(oldNode, newNode nodeTypes.Node) error {
 	return nil
 }
 
-func upsertIPsecLog(err error, spec string, loc, rem *net.IPNet, spi uint8) {
+func upsertIPsecLog(err error, spec string, loc, rem *net.IPNet, spi uint8, accs ...*error) {
```
pippolo84: I think it'd be better to add a new separate helper to wrap errors with additional scope, instead of piggyback that on `upsertIPsecLog`. That might also be a chance to add a comment and explain why we are doing that.

pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -517,42 +518,48 @@ func ipSecReplacePolicyOut(src, dst *net.IPNet, tmplSrc, tmplDst net.IP, nodeID
...
+	var acc error
 	scopedLog := log.WithFields(logrus.Fields{
 		logfields.NodeID: nodeID,
 	})
 
 	xfrmStateList, err := netlink.XfrmStateList(netlink.FAMILY_ALL)
 	if err != nil {
 		scopedLog.WithError(err).Warning("Failed to list XFRM states for deletion")
-		return
+		return acc
```
ti-mo: This will return `nil` if `netlink.XfrmStateList` returns error. Looks like `ipsecDeleteXfrmPolicy` below gets it (mostly) right.

pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -517,42 +518,48 @@ func ipSecReplacePolicyOut(src, dst *net.IPNet, tmplSrc, tmplDst net.IP, nodeID
...
+	var acc error
 	for _, p := range xfrmPolicyList {
 		if getNodeIDFromXfrmMark(p.Mark) == nodeID {
 			if err := netlink.XfrmPolicyDel(&p); err != nil {
 				scopedLog.WithError(err).Warning("Failed to delete XFRM policy")
+				acc = errors.Join(acc, fmt.Errorf("delete xfrm policy (%s): %w", p.String(), err))
 			}
 		}
 	}
+	return err
```
ti-mo:
```suggestion
	return acc
```
?

pkg/datapath/linux/node.go
```diff
@@ -544,7 +552,8 @@ func (n *linuxNodeHandler) enableSubnetIPsec(v4CIDR, v6CIDR []*net.IPNet) {
 
 // encryptNode handles setting the IPsec state for node encryption (subnet
 // encryption = disabled).
-func (n *linuxNodeHandler) encryptNode(newNode *nodeTypes.Node) {
+func (n *linuxNodeHandler) encryptNode(newNode *nodeTypes.Node) error {
+	var acc error
```
pippolo84: Nit: I see that `acc` is short for "accumulator" but I still find it a bit odd. Maybe `errs` might sound more idiomatic and familiar for the casual reader?

pkg/hive/hivetest/health.go
```diff
@@ -0,0 +1,52 @@
...
+	return m.m
+}
+
+func (m *MockHealthReporter) Err() error {
+	m.Lock()
+	defer m.Unlock()
+	return m.err
+}
+
+func (m *MockHealthReporter) IsStopped() bool {
```
ti-mo: This is personal preference, but IMO this reads a little more naturally and makes for a shorter method name:
```
if m.Stopped() {
  do something
}
```

```suggestion
func (m *MockHealthReporter) Stopped() bool {
```

pkg/hive/hivetest/health.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (m *MockHealthReporter) OK(msg string) {
+	m.Lock()
+	defer m.Unlock()
+	m.m = msg
+	m.err = nil
+}
+
+func (m *MockHealthReporter) Degraded(msg string, err error) {
```
ti-mo, line 45: Like above:

```suggestion
func (m *MockHealthReporter) Degrade(msg string, err error) {
```

pkg/hive/hivetest/health.go
```diff
@@ -0,0 +1,52 @@
...
+	m.Lock()
+	defer m.Unlock()
+	m.m = msg
+	m.err = err
+}
+
+func (m *MockHealthReporter) Stopped(msg string) {
```
ti-mo, line 56:
```suggestion
func (m *MockHealthReporter) Stop(msg string) {
```

pkg/node/manager/manager.go
```diff
@@ -133,7 +141,9 @@ func (m *manager) Subscribe(nh datapath.NodeHandler) {
 	m.mutex.RLock()
 	for _, v := range m.nodes {
 		v.mutex.Lock()
-		nh.NodeAdd(v.node)
+		if err := nh.NodeAdd(v.node); err != nil {
+			log.WithError(err).Errorf("initial subscribe initial node handler %s failed on node %s", nh.Name(), v.node.Name)
```
ti-mo: This stutters a bit, `initial subscribe to node handler`?

pkg/node/manager/manager.go
```diff
@@ -272,10 +283,15 @@ func (m *manager) backgroundSyncInterval() time.Duration {
...
 	defer syncTimerDone()
+	// Allows observers to be notified of sync update.
+	obs, tick, complete := stream.Multicast[any]()
+	m.syncObs = obs
+	defer complete(nil)
 	for {
 		syncInterval := m.backgroundSyncInterval()
 		log.WithField("syncInterval", syncInterval.String()).Debug("Performing regular background work")
 
+		var errs error
```
ti-mo, line 291: Why use `errs` here but `acc` everywhere else? We should converge on a single convention. Personally, I prefer `errs`.

pkg/node/manager/manager_test.go
```diff
@@ -829,3 +840,51 @@ func (s *managerTestSuite) TestNode(c *check.C) {
...
+	}] = &nodeEntry{node: nodeTypes.Node{Name: "node1", Cluster: "c1"}}
+	m.nodeHandlers = make(map[datapath.NodeHandler]struct{})
+	nh1 := newSignalNodeHandler()
+	nh1.EnableNodeValidateImplementationEvent = true
+	nh1.NodeValidateImplementationEventError = fmt.Errorf("test error")
+	m.nodeHandlers[nh1] = struct{}{}
+
+	// Start the manager
+	assert.NoError(m.Start(context.Background()))
+	time.Sleep(time.Millisecond * 100)
```
ti-mo: time.Sleep() is unreliable, can this be done without? Consider adding a readiness notification to the manager to both speed up the test and make it less flaky.

pkg/datapath/linux/node.go
```diff
@@ -176,9 +190,10 @@ func cidrNodeMappingUpdateRequired(oldCIDR, newCIDR cmtypes.PrefixCluster, oldIP
 	return !oldCIDR.Equal(newCIDR)
 }
 
-func deleteTunnelMapping(oldCIDR cmtypes.PrefixCluster, quietMode bool) {
+func deleteTunnelMapping(oldCIDR cmtypes.PrefixCluster, quietMode bool) error {
+	var acc error
 	if !oldCIDR.IsValid() {
-		return
+		return acc
```
ti-mo: I'd explicitly `return nil` here and declare `acc` closer to where it's used.

pkg/datapath/linux/node.go
```diff
@@ -176,9 +190,10 @@ func cidrNodeMappingUpdateRequired(oldCIDR, newCIDR cmtypes.PrefixCluster, oldIP
 	return !oldCIDR.Equal(newCIDR)
 }
 
-func deleteTunnelMapping(oldCIDR cmtypes.PrefixCluster, quietMode bool) {
+func deleteTunnelMapping(oldCIDR cmtypes.PrefixCluster, quietMode bool) error {
+	var acc error
```
ti-mo: I don't think there can be multiple errors in this function.

pkg/datapath/linux/node.go
```diff
@@ -272,14 +289,16 @@ func installDirectRoute(CIDR *cidr.CIDR, nodeIP net.IP) (routeSpec *netlink.Rout
...
 func (n *linuxNodeHandler) updateDirectRoutes(oldCIDRs, newCIDRs []*cidr.CIDR, oldIP, newIP net.IP, firstAddition, directRouteEnabled bool) error {
+	var acc error
+
 	if !directRouteEnabled {
 		// When the protocol family is disabled, the initial node addition will
 		// trigger a deletion to clean up leftover entries. The deletion happens
 		// in quiet mode as we don't know whether it exists or not
 		if firstAddition {
-			n.deleteAllDirectRoutes(newCIDRs, newIP)
+			acc = errors.Join(acc, n.deleteAllDirectRoutes(newCIDRs, newIP))
```
ti-mo: I'd probably do `return n.deleteAll...` here.

pkg/datapath/linux/node.go
```diff
@@ -272,14 +289,16 @@ func installDirectRoute(CIDR *cidr.CIDR, nodeIP net.IP) (routeSpec *netlink.Rout
...
 		}
-		return nil
+		return acc
```
ti-mo: .. and `return nil` here.

pkg/datapath/linux/node.go
```diff
@@ -272,14 +289,16 @@ func installDirectRoute(CIDR *cidr.CIDR, nodeIP net.IP) (routeSpec *netlink.Rout
 }
 
 func (n *linuxNodeHandler) updateDirectRoutes(oldCIDRs, newCIDRs []*cidr.CIDR, oldIP, newIP net.IP, firstAddition, directRouteEnabled bool) error {
+	var acc error
```
ti-mo: Doesn't look like there can be multiple errors in this function in the current incarnation.

pkg/datapath/linux/node.go
```diff
@@ -302,28 +321,46 @@ func (n *linuxNodeHandler) updateDirectRoutes(oldCIDRs, newCIDRs []*cidr.CIDR, o
...
+		if err := n.deleteDirectRoute(cidr, nodeIP); err != nil {
+			acc = errors.Join(acc, err)
+		}
 	}
+	return acc
 }
 
-func (n *linuxNodeHandler) deleteDirectRoute(CIDR *cidr.CIDR, nodeIP net.IP) {
+func (n *linuxNodeHandler) deleteDirectRoute(CIDR *cidr.CIDR, nodeIP net.IP) error {
+	var acc error
```
ti-mo: This can be declared later.

pkg/datapath/linux/node.go
```diff
@@ -335,14 +372,17 @@ func (n *linuxNodeHandler) deleteDirectRoute(CIDR *cidr.CIDR, nodeIP net.IP) {
 	routes, err := netlink.RouteListFiltered(family, filter, netlink.RT_FILTER_DST|netlink.RT_FILTER_GW)
 	if err != nil {
 		log.WithError(err).Error("Unable to list direct routes")
-		return
+		acc = errors.Join(acc, fmt.Errorf("list direct routes %s: %w", familyStr, err))
```
ti-mo: `return fmt.Errorf(...`

pkg/datapath/linux/node.go
```diff
@@ -335,14 +372,17 @@ func (n *linuxNodeHandler) deleteDirectRoute(CIDR *cidr.CIDR, nodeIP net.IP) {
...
+		return acc
 	}
```
ti-mo, line 374: Declare `acc` here.

pkg/datapath/linux/route/route_linux.go
```diff
@@ -267,7 +267,11 @@ func Upsert(route Route) error {
 
 	if err != nil {
 		if nexthopRouteCreated {
-			deleteNexthopRoute(route, link, routerNet)
+			if err2 := deleteNexthopRoute(route, link, routerNet); err2 != nil {
+				// TODO: If this fails, we may want to add some retry logic.
+				log.WithError(err2).
+					Errorf("unable to clean up nexthop route following failure to replace route")
+			}
```
pippolo84, line 274: Nit: since we don't need to reference the outer `err`, we can simply use again the `err` name here and shadow the outer one.

pkg/datapath/linux/node.go
```diff
@@ -505,6 +551,14 @@ func upsertIPsecLog(err error, spec string, loc, rem *net.IPNet, spi uint8) {
 	})
 	if err != nil {
 		scopedLog.WithError(err).Error("IPsec enable failed")
+
+		for _, acc := range accs {
```
ti-mo: This is a little questionable, this function now takes two errs.

Since you're updating many call sites already, why not wrap `err` and return it? Join it into `acc`/`errs` at the call site.

pkg/datapath/linux/node.go
```diff
@@ -996,36 +1070,38 @@ func (n *linuxNodeHandler) enableIPsec(newNode *nodeTypes.Node) {
 	// the mark fields. This uses XFRM_OUTPUT_MARK added in 4.14 kernels.
 	zeroMark := option.Config.EnableEndpointRoutes
 
-	n.enableIPsecIPv4(newNode, zeroMark)
-	n.enableIPsecIPv6(newNode, zeroMark)
+	acc = errors.Join(acc, n.enableIPsecIPv4(newNode, zeroMark), n.enableIPsecIPv6(newNode, zeroMark))
```
ti-mo: `return errors.Join(...`

pkg/datapath/linux/node.go
```diff
@@ -1035,24 +1111,27 @@ func (n *linuxNodeHandler) enableIPsecIPv4(newNode *nodeTypes.Node, zeroMark boo
...
 			localCIDR := n.nodeAddressing.IPv4().AllocationCIDR().IPNet
 			spi, err = ipsec.UpsertIPsecEndpoint(localCIDR, wildcardCIDR, localIP, wildcardIP, 0, ipsec.IPSecDirIn, false)
-			upsertIPsecLog(err, "in IPv4", localCIDR, wildcardCIDR, spi)
+			upsertIPsecLog(err, "in IPv4", localCIDR, wildcardCIDR, spi, &acc)
 		}
 	} else {
 		remoteIP := newNode.GetCiliumInternalIP(false)
 		if remoteIP == nil {
-			return
+			return acc
```
ti-mo: This is a subtle change in behaviour. If `ipsec.IPsecDefaultDropPolicy` above returns an error, this statement will now cause the function to also return non-nil, which could affect its callers. Not sure if that's intended.

pkg/datapath/linux/node.go
```diff
@@ -509,6 +509,14 @@ func upsertIPsecLog(err error, spec string, loc, rem *net.IPNet, spi uint8) {
 	})
 	if err != nil {
 		scopedLog.WithError(err).Error("IPsec enable failed")
+
+		for _, acc := range accs {
+			*acc = errors.Join(*acc, fmt.Errorf("create endpoint %s (local_ip=%s, rem=%s, spi=%d): %w",
```
pippolo84: Just a heads up that `errors.Join` [uses a newline](https://pkg.go.dev/errors#Join) to format the wrapped errors.
If this should not be convenient for our logging, an alternative is to simply use fmt.Errorf(...) with a `%w` specifier for each error to be wrapped (and then define whatever custom separator in the format string).

pkg/datapath/linux/node.go
```diff
@@ -1196,34 +1296,60 @@ func (n *linuxNodeHandler) nodeUpdate(oldNode, newNode *nodeTypes.Node, firstAdd
...
 		}
-		return nil
+		return acc
 	}
 
 	if n.nodeConfig.EnableAutoDirectRouting {
-		n.updateDirectRoutes(oldAllIP4AllocCidrs, newAllIP4AllocCidrs, oldIP4, newIP4, firstAddition, n.nodeConfig.EnableIPv4)
-		n.updateDirectRoutes(oldAllIP6AllocCidrs, newAllIP6AllocCidrs, oldIP6, newIP6, firstAddition, n.nodeConfig.EnableIPv6)
-		return nil
+		if err := func() error {
```
ti-mo: Instead of new acc and closure: `return errors.Join(n.updateDirectRoutes(), n.updateDirectRoutes())`...

pkg/datapath/linux/node.go
```diff
@@ -1196,34 +1296,60 @@ func (n *linuxNodeHandler) nodeUpdate(oldNode, newNode *nodeTypes.Node, firstAdd
...
+			acc = errors.Join(acc, n.updateDirectRoutes(oldAllIP6AllocCidrs, newAllIP6AllocCidrs, oldIP6, newIP6, firstAddition, n.nodeConfig.EnableIPv6))
+			return acc
+		}(); err != nil {
+			acc = errors.Join(acc, fmt.Errorf("enable direct routing: %w", err))
+		}
+		return acc
 	}
 
+	// For use in aggregating encapsulation code errors.
+	var encapAcc error
```
ti-mo: Is this really necessary? This just makes things less clear, and it's joined back into `acc` later.. I'd probably add extra wrapping to `updateTunnelMapping` calls instead.

pkg/datapath/linux/node.go
```diff
@@ -1301,33 +1440,45 @@ func (n *linuxNodeHandler) nodeDelete(oldNode *nodeTypes.Node) error {
...
-		n.deleteDirectRoute(oldNode.IPv6AllocCIDR, oldIP6)
+		if err := n.deleteDirectRoute(oldNode.IPv4AllocCIDR, oldIP4); err != nil {
+			acc = errors.Join(acc, fmt.Errorf("direct routing: %w", err))
+		}
+		if err := n.deleteDirectRoute(oldNode.IPv6AllocCIDR, oldIP6); err != nil {
+			acc = errors.Join(acc, fmt.Errorf("direct routing: %w", err))
+		}
 	}
 
+	var encapAcc error
```
ti-mo: Same as above, this is a little confusing.

pkg/datapath/linux/node.go
```diff
@@ -1196,34 +1296,60 @@ func (n *linuxNodeHandler) nodeUpdate(oldNode, newNode *nodeTypes.Node, firstAdd
...
-		return nil
+		if err := func() error {
+			var acc error
+			acc = errors.Join(acc, n.updateDirectRoutes(oldAllIP4AllocCidrs, newAllIP4AllocCidrs, oldIP4, newIP4, firstAddition, n.nodeConfig.EnableIPv4))
+			acc = errors.Join(acc, n.updateDirectRoutes(oldAllIP6AllocCidrs, newAllIP6AllocCidrs, oldIP6, newIP6, firstAddition, n.nodeConfig.EnableIPv6))
+			return acc
+		}(); err != nil {
+			acc = errors.Join(acc, fmt.Errorf("enable direct routing: %w", err))
+		}
+		return acc
```
pippolo84: `errors.Join` is variadic, we can rewrite this in a (slightly) simpler way:

```suggestion
if err := func() error {
			return errors.Join(
				n.updateDirectRoutes(oldAllIP4AllocCidrs, newAllIP4AllocCidrs, oldIP4, newIP4, firstAddition, n.nodeConfig.EnableIPv4),
				n.updateDirectRoutes(oldAllIP6AllocCidrs, newAllIP6AllocCidrs, oldIP6, newIP6, firstAddition, n.nodeConfig.EnableIPv6),
			)
		}(); err != nil {
			return errors.Join(acc, fmt.Errorf("enable direct routing: %w", err))
		}
		return acc
```

pkg/datapath/linux/node.go
```diff
@@ -1301,33 +1440,45 @@ func (n *linuxNodeHandler) nodeDelete(oldNode *nodeTypes.Node) error {
...
+	}
 
 	if n.enableNeighDiscovery {
 		go n.deleteNeighbor(oldNode)
 	}
 
 	if n.nodeConfig.EnableIPSec {
-		n.deleteIPsec(oldNode)
+		if err := n.deleteIPsec(oldNode); err != nil {
+			acc = errors.Join(acc, fmt.Errorf("ipsec: %w", err))
```
ti-mo:
```suggestion
			acc = errors.Join(acc, fmt.Errorf("delete ipsec: %w", err))
```

pkg/datapath/linux/node.go
```diff
@@ -1244,35 +1370,47 @@ func (n *linuxNodeHandler) nodeUpdate(oldNode, newNode *nodeTypes.Node, firstAdd
 			newPrefixCluster6 = cmtypes.PrefixClusterFromCIDR(newNode.IPv6AllocCIDR, 0)
 		}
 
-		nodeID := n.allocateIDForNode(newNode)
+		nodeID, err := n.allocateIDForNode(newNode)
+		encapAcc = errors.Join(encapAcc, err)
```
pippolo84: It is possible to declare the encapsulation accumulator here directly:

```
nodeId, encapAcc := n.allocateIDForNode(newNode)
```

and get rid of the declaration at line 1352.

Also, I'd use something like `encapErrs` as a name for that (see other comment about "accumulator" naming)

ti-mo (dismissed):
Looks good overall, thanks! I think we should hold off on merging this until 1.15, this might introduce subtle changes in control flow despite all the care that was taken.

pkg/datapath/linux/node.go
```diff
@@ -1430,33 +1431,45 @@ func (n *linuxNodeHandler) nodeDelete(oldNode *nodeTypes.Node) error {
...
-		n.deleteDirectRoute(oldNode.IPv6AllocCIDR, oldIP6)
+		if err := n.deleteDirectRoute(oldNode.IPv4AllocCIDR, oldIP4); err != nil {
+			acc = errors.Join(acc, fmt.Errorf("direct routing: %w", err))
+		}
+		if err := n.deleteDirectRoute(oldNode.IPv6AllocCIDR, oldIP6); err != nil {
+			acc = errors.Join(acc, fmt.Errorf("direct routing: %w", err))
+		}
 	}
 
+	var encapAcc error
```
pippolo84: Same here

pkg/datapath/linux/node.go
```diff
@@ -158,7 +158,7 @@ func updateTunnelMapping(oldCIDR, newCIDR cmtypes.PrefixCluster, oldIP, newIP ne
 	case oldCIDR.IsValid() && newCIDR.IsValid() && !oldCIDR.Equal(newCIDR):
 		if err := deleteTunnelMapping(oldCIDR, false); err != nil {
 			acc = errors.Join(acc,
-				fmt.Errorf("delete tunnel mapping (oldCIDR: %s): %w", oldCIDR, newIP))
+				fmt.Errorf("delete tunnel mapping (oldCIDR: %s, newIP: %s): %w", oldCIDR, newIP, err))
```
pippolo84: This fix here should be squashed in previous commit 98c694630497c0189eb47b6287b8acfaeb68d4f8.

pkg/datapath/linux/node.go
```diff
@@ -227,7 +227,7 @@ func createDirectRouteSpec(CIDR *cidr.CIDR, nodeIP net.IP) (routeSpec *netlink.R
 
 	routes, err = netlink.RouteGet(nodeIP)
 	if err != nil {
-		err = fmt.Errorf("unable to lookup route for node %s: %s", nodeIP, err)
+		err = fmt.Errorf("unable to lookup route for node %s: %w", nodeIP, err)
```
pippolo84, line 229: This change should be squashed in commit c076ddcfd45ca763f28914caab68c7596f34bc77

pkg/node/manager/manager_test.go
```diff
@@ -833,3 +840,51 @@ func (s *managerTestSuite) TestNode(c *check.C) {
...
```
pippolo84: Why we need this `time.Sleep` here? I am afraid that this might increase test flakyness when running in CI.

pippolo84 (changes requested):
Thanks for the PR! I've left some suggestions inline, mainly around naming and style.

pkg/node/manager/manager_test.go
tommyp1ckles: agreed, think this is an artifact at my first attempt where I used sleeping instead of syncing up with the sync loop.

pkg/datapath/linux/node.go
tommyp1ckles: I think this is an easy case to reason about, prior to this PR enableIPsecIPv4/enableIPsecIPv6 did not return errors, so in this case I think we're safe from the problem of a change in behaviour of nil error causes some divergent behaviour down the line

pkg/datapath/linux/node.go
tommyp1ckles: The reasoning here, as well as the separate errors, is to prevent duplication when a series of function calls are all to be similarly "scoped".

For example

```go
errs = errors.Join(errs, fmt.Errorf("tried some thing: %w", foo1()))
//...
errs = errors.Join(errs, fmt.Errorf("tried some thing: %w", foo2()))
//...
errs = errors.Join(errs, fmt.Errorf("tried some thing: %w", foo3()))
// ...
```

Perhaps the closure here is a bit out of place, I didn't want to go and start refactoring sections of this into separate functions so I was trying to settle between the three approaches you see (somewhat inconstantly):

* create separate error accumulator.
* use a closure, wrap one error
* duplicate the same string around.

I'll try to make this more consistent for now,  I think we can go with "good enough" with approach three for now.

pkg/datapath/linux/node.go
tommyp1ckles: See: Comment about the closure above for motivation, ill take another look to see if I can make things more consistent 😄

pkg/node/manager/manager.go
```diff
@@ -292,13 +308,25 @@ func (m *manager) backgroundSync(ctx context.Context) error {
...
+				}
 			})
 			entry.mutex.Unlock()
 
 			m.metrics.DatapathValidations.Inc()
 		}
 
+		if errs != nil {
+			log.WithError(errs).Debug("Failed to do node validation")
+			m.healthReporter.Degraded("Failed to do node validation", errs)
```
christarazi: Nit: `Failed to perform node validation`

pkg/node/manager/manager.go
```diff
@@ -292,13 +308,25 @@ func (m *manager) backgroundSync(ctx context.Context) error {
 			entry.mutex.Lock()
 			m.mutex.RUnlock()
 			m.Iter(func(nh datapath.NodeHandler) {
-				nh.NodeValidateImplementation(entry.node)
+				if err := nh.NodeValidateImplementation(entry.node); err != nil {
+					log.WithError(err).Errorf("Failed to validate handler %s on node %s", nh.Name(), entry.node.Name)
```
christarazi: Let's use structured logging (`WithFields`) and avoid dynamic log formatting. See `pkg/logging/logfields` as well for existing logfields and add any that are needed.

pkg/node/manager/manager.go
```diff
@@ -122,6 +126,10 @@ type manager struct {
 	// controllerManager manages the controllers that are launched within the
 	// Manager.
 	controllerManager *controller.Manager
+
+	healthReporter cell.HealthReporter
+
+	syncObs stream.Observable[any]
```
christarazi: Let's stick to the convention in this struct to have godocs on each field. Please also elaborate on `syncObs`.

pkg/datapath/linux/node.go
tommyp1ckles: yeah, I think you and @ti-mo are right 😄 , it seemed like a good choice at the time but I think errs is what most people would expect and thus will be more readable

pkg/node/manager/manager_test.go
tommyp1ckles: oops, leftover from old iteration, removed!

pkg/datapath/linux/node.go
tommyp1ckles: I wanted to piggyback on this just to reduce the number of changes for now, I agree with you and @ti-mo that the accs pointer approach isn't very idiomatic so ill change this to behave more similarly to errors.Join.  This should also make it more clear in line that the error is being pushed on to the errs list.

The error context being added isn't always that great atm, so I think we should defer this for now and add more helpful context more thoughtfully in the future.

pkg/datapath/linux/route/route_linux.go
tommyp1ckles: err gets returned, so that would be changing the potential behaviour here.

pkg/datapath/linux/node.go
tommyp1ckles: yup, much cleaner, updated!

pkg/datapath/linux/node.go
tommyp1ckles: Good point, the goal here was to consume as part of module health status, but in the meantime I have added extra logging and the node manager level so this could produce some ugly logs (I believe the newlines would be escaped by logrus though?). 

Wrapping doesn't really make sense in this context since the %w directive is supposed to add more context on top of an existing error, whereas the goal here is to provide a list of errors to indicate where reconciliation procedure calls have failed.

Ultimately, we do want to avoid this accumulation logic and just have well defined procedures that return when they encounter an error, thus reducing the potential number of states massively.  But I want to do more discussion/proposal work for that first.

pkg/datapath/linux/node.go
tommyp1ckles: I refactored this as part of timos request

pkg/node/manager/manager.go
```diff
@@ -133,7 +141,9 @@ func (m *manager) Subscribe(nh datapath.NodeHandler) {
 	m.mutex.RLock()
 	for _, v := range m.nodes {
 		v.mutex.Lock()
-		nh.NodeAdd(v.node)
+		if err := nh.NodeAdd(v.node); err != nil {
+			log.WithError(err).Errorf("handler initial subscribe: %s failed on node %s", nh.Name(), v.node.Name)
```
christarazi: For error and warning level logs, we need to be more thoughtful about the log msgs. The msgs need to be useful for the end user by telling them what are the ramifications and what Cilium will do about it. In some cases there's nothing that a user can do and so for example, the msg can say to file an issue.

In this case, I'd suggest 

```
Initial node event handling failed. Cilium may be in a degraded state. See error message for more details.
```

In the future, we can improve the error msg to say that Cilium will retry, etc. 

Please apply the above to the rest of the PR for errors / warnings logs.

pkg/node/manager/manager.go
christarazi: Do we need both debug logs and health reporter logs? Maybe we can put a debug log inside the health reporter instead of calling `.Debug` at each call site of the health reporter.

pkg/datapath/linux/node.go
```diff
@@ -509,9 +509,15 @@ func upsertIPsecLog(err error, spec string, loc, rem *net.IPNet, spi uint8) {
 	})
 	if err != nil {
 		scopedLog.WithError(err).Error("IPsec enable failed")
+		return fmt.Errorf("create endpoint: %s (local_ip=%s, rem=%s, spi=%d): %w",
+			spec,
+			loc.String(),
+			rem.String(),
+			spi, err)
 	} else {
```
christarazi: I think we should avoid the custom formatting (`local_ip=...`) and instead make the error msg itself more naturally flow:

```
return fmt.Errorf("failed to create endpoint for %q with local IP %q and remote IP %q: %w", spec, loc, rem)
```

christarazi: Not sure what "create endpoint" is referring to here though

christarazi: Also, if we are already going to log when this error occurs, then we don't need to stuff the error msg with the same context again. It would also be good to use the newly created error within the log msg, i.e. `WithError()`

pkg/datapath/linux/node.go
```diff
@@ -319,6 +319,15 @@ func (n *linuxNodeHandler) updateDirectRoutes(oldCIDRs, newCIDRs []*cidr.CIDR, o
...
 			log.WithError(err).Warningf("Unable to install direct node route %s", routeSpec.String())
+			// In the current implementation, this often fails because updates are tried for both ip families
+			// regardless if the Node has either ip types.
+			// At the time of this change we are only interested in bubbling up errors without affecting execution flow.
+			// Thus we are ignoring the error here for now.
+			//
+			// TODO(Tom): In the future we will want to avoid attempting to do the update if we know it will fail.
+			if newIP == nil && errors.Is(err, unix.ERANGE) {
+				return nil
+			}
```
christarazi, line 330: My concern with this is that we end up ignoring legitimate errors, but maybe this stems from the fact that I don't understand fully why this commit is needed.

pkg/datapath/linux/node.go
```diff
@@ -1473,102 +1622,117 @@ func (n *linuxNodeHandler) createNodeExternalIPSecOutRoute(ip *net.IPNet, dflt b
...
 	} else {
 		if !n.nodeConfig.EnableIPv6 {
-			return
+			return nil
 		}
 	}
 
 	if err := route.Upsert(n.createNodeExternalIPSecOutRoute(ip, true)); err != nil {
 		log.WithError(err).WithField(logfields.CIDR, ip).Error("Unable to replace the IPSec route OUT the default routing table")
+		errs = errors.Join(errs, fmt.Errorf("replace ipsec default route OUT: %w", err))
```
christarazi: One nit about the new error msgs: they should state exactly what went wrong so that they can be joined together in the natural way that Go errors are aggregated.

Here, it should say:

```suggestion
		errs = errors.Join(errs, fmt.Errorf("failed to replace ipsec default route in the OUT direction: %w", err))
```

In theory, error msgs should be read like:

```
failed to validate Linux node: failed to process new Linux node event: failed to replace ipsec default route in the OUT direction: ENOMEM
```

Please ensure all new error msgs adhere to the above.

christarazi (commented):
Nice work!

I have a smattering of random comments related to formatting the errors and logs so that we aren't introducing new signals into the tree which may be incorrectly interpreted by us as developers or users.

In regards to the discussion about control flow, I agree that we should avoid as many side effects as possible. My inclination is that this change can be done in a way that does not contain any additional side effects. If that's the case, I'd like to understand more deeply why that's the case. However, as long as we avoid changing the control flow and this PR effectively becomes just a change that better reports the status / errors of the node manager, then it can still be considered for v1.14.

pkg/node/manager/manager.go
tommyp1ckles: Probably not actually, the health reporter has debug logging on emitting status updates (which we may consider making into info at some point).

pkg/datapath/linux/node.go
tommyp1ckles: I was in between two priorities in many of these:

* keeping the error messages as terse/short as possible as to avoid excessively long errors once they've made their way up the stack.
* Providing enough context info such that users aren't left scratching their heads about why exactly the error occured.

pkg/datapath/linux/node.go
tommyp1ckles: Yeah, it's possible, but right now this error happens every time in some situations.  There's a better solution to this but it involves avoiding trying certain calls if ip4/ip6 isn't enabled.

pkg/datapath/linux/node.go
tommyp1ckles: When I was working on this I was concerned about creating unnecessary redundancy in errors, for example the "failed to..." gets repeated a lot down the stack and doesn't add much information.

I would argue that an error implies that something has failed, IMO to make things more clear we should add this wording closer to the top level:

```
failed to validate Linux node: processing new event: replacing ipsec default route in the OUT direction: ENOMEM
```

Where the top level explicitly indicates that there has been a failure/error/problem etc.  Following this you have essentially a sort of stack trace indicating error context all the way down.  With finally the error terminating with the core error.

(I also was trying to avoid having words like "node" or "route" repeated all the way down the stack).

pkg/hive/hivetest/health.go
tommyp1ckles: Re: this suggestion and the two above I think are more pointed at the hive/cell/HealthReporter interface.

I'm going to re-frame from changing this as part of this PR, I don't think that API has stabilized yet so we can still make changes.  I'm working on some design proposals that will likely include this API so I think its best we leave that to there.

pkg/hive/hivetest/health.go
tommyp1ckles: See: comment below

pkg/datapath/linux/node.go
tommyp1ckles: Cleaned this up to be more consistent with the rest of the changes.

pkg/datapath/linux/route/route_linux.go
pippolo84: Are you referring to the return at line 276? If so, that won't be affected, since the scope of `err2` is limited to the inner `if`. [Here](https://go.dev/play/p/EOEmPxuSQKD) a similar example where the "outer" `err` gets shadowed with an inner `err` whose scope is bounded to the inner `if`.

Anyway, I agree that shadowing might be trickier to read, so if you prefer to use `err2` instead that's completely fine :+1:

pkg/datapath/linux/node.go
pippolo84: > (I believe the newlines would be escaped by logrus though?).

TBH, I'm not sure.

> Wrapping doesn't really make sense in this context since the %w directive is supposed to add more context on top of an existing error, whereas the goal here is to provide a list of errors to indicate where reconciliation procedure calls have failed.

Actually, since Go 1.20 you can use `fmt.Errorf` with multiple `%w` directives, just to wrap multiple errors and customize the wrapped error string representation, like this:

`fmt.Errorf("%w-%w", err1, err2)`

That's basically equivalent of just wrapping `err1` and `err2` with `errors.Join`, but with the additional possibility of getting rid of the newline and specify a custom separator.

From the Go 1.20 release notes:

> The [fmt.Errorf](https://tip.golang.org/pkg/fmt/#Errorf) function now supports multiple occurrences of the %w format verb, which will cause it to return an error that wraps all of those error operands.
>
> The new function [errors.Join](https://tip.golang.org/pkg/errors/#Join) returns an error wrapping a list of errors.

Relevant code for [fmt.Errorf](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/fmt/errors.go) and for [errors.Join](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/fmt/errors.go).

> But I want to do more discussion/proposal work for that first.

Sounds good, then, let's go with this logic for now :+1:

pkg/datapath/linux/node.go
```diff
@@ -1165,6 +1253,10 @@ func (n *linuxNodeHandler) subnetEncryption() bool {
 // Must be called with linuxNodeHandler.mutex held.
 func (n *linuxNodeHandler) nodeUpdate(oldNode, newNode *nodeTypes.Node, firstAddition bool) error {
 	var (
+		// Don't stop executing the function if we get an error. Instead we
+		// log and aggregate errors in accumulator.
+		acc error
```
pippolo84: I'd change this one too to `errs`

pkg/node/manager/manager_test.go
```diff
@@ -829,3 +840,50 @@ func (s *managerTestSuite) TestNode(c *check.C) {
 	// Needs to be the same as n2
 	c.Assert(n, checker.DeepEquals, *n1V2)
 }
+
+func TestNodeManagerEmitStatus(t *testing.T) {
```
pippolo84, line 843: Trying the test locally, it seems to be flaky:

```
$ go test -count=5 -v ./pkg/node/manager/... -run="TestNodeManagerEmitStatus"
=== RUN   TestNodeManagerEmitStatus
--- FAIL: TestNodeManagerEmitStatus (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1fe4738]

goroutine 22 [running]:
...
github.com/cilium/cilium/pkg/node/manager.TestNodeManagerEmitStatus(0x0?)
	/home/pippolo/go/src/github.com/cilium/cilium/pkg/node/manager/manager_test.go:869 +0x618
...
FAIL	github.com/cilium/cilium/pkg/node/manager	0.692s
FAIL
```

The panic happens when `m.syncObs` has not yet been initialized (that's because `backgroundSync` is called asynchronously from the workerpool). A solution might be to initialize syncObs in `*manager.Start`, save a reference to the returned `tick` and `complete` functions inside the manager struct to be called later where needed.

Also, I think the test can be greatly simplified avoiding the creation of multiple channels and doing this:

```
	ch := make(chan struct{})
	m.syncObs.Observe(ctx, func(_ any) {
		ch <- struct{}{}
	}, func(err error) {
		close(ch)
	})
	tick := func() {
		<-ch
	}
```

As a final consideration, I don't like too much the idea of adding code into the manager that has the sole purpose of testing the integration with the health reporter (unless we are planning to use `syncObs` for something else in the future).
What about removing the `stream.Multicast` and write the test like this:

```
	// Start the manager
	assert.NoError(m.Start(ctx))

	require.Eventually(t, func() bool { return hr.Err() != nil }, time.Minute, 5*time.Millisecond)
	nh1.NodeValidateImplementationEventError = nil
	require.Eventually(t, func() bool { return hr.Err() == nil }, time.Minute, 5*time.Millisecond)

	// Stop the manager
	assert.NoError(m.Stop(ctx))
```

pippolo84 (changes requested):
Thanks for addressing the previous comments. I've left a comment about the manager test.

pkg/node/manager/manager_test.go
tommyp1ckles: Hmm yeah I didn't know about require.Eventually, I was hesitant to add the stream but I wanted to avoid non determinism by just using `time.Sleep()`

I'm going to change this over to just use the require approach, should clean things up quite a bit

tommyp1ckles: Oof this one is tricker than expected, upon further inspection, even this is a bit racey and lead to flakes.

I ended up just using the existing test handler fixture, but syncing on handler updates, this method proved a lot more reliable since we can sync on both the handler update and the health status update.

I ran it for a million iterations without a flake so hopefully this is the best way forward.
```
go test -failfast -count=1000000 -v -run=TestNodeManagerEmitStatus
```

pkg/datapath/linux/node.go
christarazi: I think my main principle here is to be consistent with the rest of the Cilium code base in regards to how errors are formatted. We need to think about whether what the PR code does is also how each module should structure / format their errors. Otherwise, if we follow how the rest of the code base is, then there's no "controversy" so to speak. WDYT?

tommyp1ckles: makes sense, ill keep things consistent with the broader code base here for now, we can continue to figure out the best way to do these messages going forward.

tommyp1ckles:
@pippolo84 Reworked the manager test, should be much more reliable now PTAL.

pippolo84 (approved):
Awesome, thanks! :rocket:

tommyp1ckles:
/test-1.24-5.4

tommyp1ckles:
/test-1.26-net-next

brb (approved):
WireGuard changes :ok_hand:

ti-mo:
Looks like this needs a rebase!

rolinh (approved):

tommyp1ckles:
@pippolo84 Had to make some clustermesh changes in last rebase, mind taking a look/re-approving for those changes?

tommyp1ckles:
@jibi Mind taking a look at this for cilium/ipsec 🥺

pippolo84 (approved):
Recent changes LGTM :rocket:

christarazi (approved):

tommyp1ckles:
@brb could you take a look at this as well for sig-ipsec? jibi is no longer on that review team.

pkg/datapath/linux/node.go
```diff
@@ -509,9 +509,15 @@ func upsertIPsecLog(err error, spec string, loc, rem *net.IPNet, spi uint8) {
 	})
 	if err != nil {
 		scopedLog.WithError(err).Error("IPsec enable failed")
+		return fmt.Errorf("failed to enable ipsec with %s using local IP %s, rem %s, spi %d: %w",
```
brb, line 551: I think the error logging above (and in other places) is no longer needed, as it will be logged by the node manager.

pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -536,42 +537,48 @@ func matchesOnNodeID(mark *netlink.XfrmMark) bool {
...
+	var errs error
 	scopedLog := log.WithFields(logrus.Fields{
 		logfields.NodeID: nodeID,
 	})
 
 	xfrmStateList, err := netlink.XfrmStateList(netlink.FAMILY_ALL)
 	if err != nil {
 		scopedLog.WithError(err).Warning("Failed to list XFRM states for deletion")
-		return
+		return err
```
brb, line 549: Ditto with the above.

brb (approved):
Nice changes, thanks! I guess we can remove the redundant logging in a follow-up PR.

dylandreimerink:
@tommyp1ckles I will hold off on merging for now given the open nits of Martynas, if you don't want to address them now and just merge thats also fine by me, just give the signal.

tommyp1ckles:
@dylandreimerink @brb leaving in the logging was a conscious choice, this is already a large PR so I wanted to avoid making too many changes that might affect debugging ipsec.  I'd like to do that in a follow up PR.

joestringer:
Depending on how far down this track we think it's viable to go during the v1.15 cycle, we may want to nominate this for `release-note/major` to highlight during v1.15 final release communications about how we're making Cilium easier to understand, operate and debug through more structured health visibility.

aanm:
@tommyp1ckles can you write a release note that is user-facing? Thanks! Don't forget to use the right format so that the automation picks it up.

PR #25934, opened 2023-06-06
resource: implement stream.Observable

mhofstetter:
This commit lets resource.Resource implement stream.Observable to easily use k8s resources as observable without wrapping them with stream.FromChannel.

This is especially useful with hive jobs and make direct use of the context handling.

thanks @joamaki for recommending this

PR #26018, opened 2023-06-08
auth: add missing stream package import

giorio94:
This PR adds back the stream package import, which appears to be missing due to a merge race between #25927 and #25934.

Fixes: eb653b60338e ("auth: use Resource.Observe for jobs")
Fixes: ebb6fc38803a ("auth: implement re-authentication in case of rotated certificates")


mhofstetter (approved):
thanks!

giorio94:
All images built correctly, and linting passed. Marking as ready to merge given that main is currently broken.

PR #26019, opened 2023-06-08
.github: refactor job matrix generation into YAML files

aanm:
This commit addresses two main issues and introduces a more streamlined approach to generating job matrices for GitHub Actions. By migrating the job matrix definition into YAML files, we gain the ability to separate scheduled jobs and pull requests (PRs) into specific files, ensuring comprehensive testing coverage similar to what we have with Jenkins.

The first benefit of this approach is the improved organization of job definitions. With separate files for scheduled jobs and PRs, it becomes easier to manage and maintain testing configurations independently. This modular structure allows us to make changes to specific job types without impacting others, enhancing overall maintainability.

The second advantage is with help of jq, we can define kernel versions for all desired Kubernetes (k8s) versions to be tested without manually commenting out unwanted versions. Previously, when using the "include" mechanism, certain k8s versions like 1.24 had to be commented out to prevent their execution. However, with jq, we can selectively include only the desired k8s versions by mapping them with the definitions in main-prs.yaml or main-scheduled.yaml.

Lastly, this refactor enables per-branch configuration through prefixed filenames. By structuring configuration files with branch-specific prefixes, we can easily define and manage job configurations tailored to stable branches.

### DIFF
--- a/.github/workflows/conformance-ginkgo.yaml
+++ b/.github/workflows/conformance-ginkgo.yaml
@@ -79,7 +79,15 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       tested: ${{ steps.tested-tree.outputs.src }}
+      matrix_sha: ${{ steps.sha.outputs.sha }}
+      base_branch: ${{ steps.sha.outputs.base_branch }}
       sha: ${{ steps.sha.outputs.sha }}
+      #
+      # For bisect uncomment the base_branch and 'sha' lines below and comment
+      # the two lines above this comment
+      #
+      #base_branch: <replace with the base branch name, should be 'main', not your branch name>
+      #sha: <replace with the SHA of an existing docker image tag that you want to bisect>
     steps:
       # Because we run on issue comments, we need to checkout the code for
       # paths-filter to work.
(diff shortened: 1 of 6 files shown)


brlbil (approved):

.github/workflows/conformance-ginkgo.yaml
```diff
@@ -211,8 +221,61 @@ jobs:
...
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - name: Checkout PR code
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+        with:
+          ref: ${{ needs.check_changes.outputs.matrix_sha }}
+          persist-credentials: false
+
+      - name: Convert YAML to JSON
```
nebril, line 237: Is this done because we need `jq`-specific features that `yq` is lacking?

nebril (approved):
This is awesome! I think one thing missing would be adding some kind of yaml schema for new `branch-*.yaml` files, that would validate their structure. I think that debugging errors made in these files would be pretty painful, so it might make sense to make these as obvious as possible.

I don't think it's blocking though, it could be tracked in a new issue.

.github/workflows/conformance-ginkgo.yaml
aanm: @nebril no, it's because of the there isn't a `fromYaml` function, similar to the `fromJson` on GitHub as documented https://docs.github.com/en/enterprise-cloud@latest/actions/learn-github-actions/expressions#fromjson

PR #26022, opened 2023-06-08
job,test: Improved reliability of `TestTimer_DoubleTrigger`

dylandreimerink:
There existed an edge case in this test which caused it to flake:
```
panic: close of closed channel

goroutine 106 [running]:
github.com/cilium/cilium/pkg/hive/job.TestTimer_DoubleTrigger.func1()
	/home/runner/work/cilium/cilium/pkg/hive/job/job_test.go:445 +0x1d
```

This PR moves the triggering of the timer to before the hive starts to ensure they are all processes before the timer has a chance to start consuming them. Also added a guard around the closing of the channel so we don't panic if the test were to fail for another reason.

```release-note
Improved reliability of pkg/hive/job timer double trigger unit test
```

### DIFF
--- a/pkg/hive/job/job_test.go
+++ b/pkg/hive/job/job_test.go
@@ -442,11 +442,13 @@ func TestTimer_DoubleTrigger(t *testing.T) {
 
 		g.Add(
 			Timer("on-interval", func(ctx context.Context) error {
-				defer func() { close(ran) }()
+				defer func(iter int) {
+					if iter == 0 {
+						close(ran)
+					}
+				}(i)
 
 				i++
-
-				time.Sleep(100 * time.Millisecond)
 
 				return nil
 			}, 1*time.Hour, WithTrigger(trigger)),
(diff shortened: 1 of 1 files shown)


pkg/hive/job/job_test.go
```diff
@@ -455,15 +459,20 @@ func TestTimer_DoubleTrigger(t *testing.T) {
...
 		t.Fatal(err)
 	}
 
-	trigger.Trigger()
-	trigger.Trigger()
-	trigger.Trigger()
 	<-ran
 
+	// Give the timer some time for bogus additional runs to come in after the first one
+	time.Sleep(time.Second)
```
ti-mo: Is this strictly necessary? Sleeping in tests is a big hammer.

ti-mo (approved):
Thanks! One concern around using time.Sleep().

pkg/hive/job/job_test.go
dylandreimerink: Ideally I would be able to check if the timer thread is currently blocked or not. I will remove this and the 100ms timer and then hope that when the behavior under test breaks that the callback is called more then once between `<-ran` and `Stop`.

Testing stuff like this is just hard without time

PR #25436, opened 2023-05-12
High-Scale IPcache: Chapter 2

pchaigno:
This is in the context of the high-scale ipcache feature described at https://github.com/cilium/design-cfps/pull/7.

This pull request simply adds a new CRD, used to describe IP addresses of entities outside the Cilium high-scale ipcache domain, typically outside the clustermesh.

Updates: https://github.com/cilium/cilium/issues/25243.
```release-note
New CiliumWorldCIDRSet CRD to define IP addresses outside the Cilium domain in the context of high-scale ipcache.
```

PR #26028, opened 2023-06-08
 High-Scale IPcache: Chapter 2

squeed:
This picks back up closed PR #25436, which needed some changes to get it past the finish line.

Differences from that PR:
- The manager now uses the Resource paradigm
- The BPF map is now exported as a Cell in the Hive

---

This pull request simply adds a new CRD, used to describe IP addresses of entities outside the Cilium high-scale ipcache domain, typically outside the clustermesh.

```release-note
New CiliumWorldCIDRSet CRD to define IP addresses outside the Cilium domain in the context of high-scale ipcache.
```


daemon/cmd/daemon.go
```diff
@@ -195,6 +196,8 @@ type Daemon struct {
 
 	ipamMetadata *ipamMetadata.Manager
 
+	worldCIDRsManager *worldcidrs.Manager
```
squeed: note to self: this can be removed.

pkg/worldcidrs/manager.go
```diff
@@ -0,0 +1,296 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
```
squeed, line 2: todo: add a quick intro to this file.

zacharysarah (approved):
Easiest review ever! 👍🏻

install/kubernetes/cilium/Chart.yaml
```diff
@@ -143,3 +143,11 @@ annotations:
       displayName: Cilium Pod IP Pool
       description: |
         CiliumPodIPPool defines an IP pool that can be used for pooled IPAM (i.e. the multi-pool IPAM mode).
+    - kind: CiliumWorldCIDRSet
```
sayboras, line 146: nit: we are having ciliumcidrgroups (i.e. groups vs sets), not sure if you are thinking about it and intentionally use `groups` vs `sets`.

sayboras (approved):
One small comment as per below, but it's mainly related to sig-k8s.

Other changes in helm look good to me :heavy_check_mark:

install/kubernetes/cilium/Chart.yaml
christarazi: @squeed Should we rename to CiliumWorldCIDRGroups?

install/kubernetes/cilium/Chart.yaml
```diff
@@ -143,3 +143,11 @@ annotations:
...
+      version: v2alpha1
+      name: ciliumworldcidrsets.cilium.io
+      displayName: Cilium World CIDR Set
+      description: |
+        WorldCIDRs is a list of IPv4 CIDRs that will be considered outside the
+        "high-scale ipcache domain" (typically outside the clustermesh) and for
+        which traffic won't be encapsulated.
```
christarazi: I think this isn't entirely accurate because the user still must specify in the CRD whether encap is true or false, even though false is the only supported now.

pkg/k8s/apis/cilium.io/v2alpha1/worldcidrset_types.go
```diff
@@ -0,0 +1,63 @@
...
+// CiliumWorldCIDRSetSpec configures the routing decision for a set of CIDRs.
+type CiliumWorldCIDRSetSpec struct {
+
+	// Encapsulate is whether or not the CIDRs in this set
+	// should be encapsulated.
+	//
+	// Currently, only false is supported.
+	//
+	// +kubebuilder:validation:Required
+	Encapsulate *bool `json:"encapsulate"`
```
christarazi, line 57: Does it need to be required? Why is it a pointer if it's required?

install/kubernetes/cilium/Chart.yaml
squeed: Ordinarily I would totally agree - I _hate_ API inconsistencies. In this case, however, I would be afraid that CIDRGroup and WorldCIDRGroup might be thought of as two almost-identical mechanisms that differ only in scope. For example, CiliumNetworkPolicy and CiliumClusterwideNetworkPolicy.

However, CiliumCIDRGroup affects policy, whereas CiliumWorldCIDRSet configures routing. I wouldn't want to confuse the two.

If we can come up with a better name for the type, I would love to. CiliumCIDRRoute? CiliumStaticRoute?

install/kubernetes/cilium/Chart.yaml
squeed: agreed, I'll reword.

pkg/k8s/apis/cilium.io/v2alpha1/worldcidrset_types.go
squeed: It's required for now, but we could add other options later, when this no longer becomes required. Then we need to distinguish between `false` and unset.

install/kubernetes/cilium/Chart.yaml
squeed: Aha, I thought this would be automatically updated from `make manifests`. Nope.

install/kubernetes/cilium/Chart.yaml
christarazi: How about going in a different direction: could we make CiliumWorldCIDR{Set,Group} expose a [CiliumCIDRGroups](https://github.com/cilium/cilium/blob/87ac446b902cf0f49177f87cc1cfc96a250eaded/pkg/policy/api/cidr.go#L49) reference instead of raw CIDRs? Then the only difference that CiliumWorldCIDR{Set,Group} would have is to provide routing configuration API. This makes the CRDs layer on each other such that the link between the two CRDs are explicit and then the naming question becomes much simpler.

Concretely, this is my proposal,

```
type CiliumWorldCIDRSetSpec struct {

	// Encapsulate is whether or not the CIDRs in this set
	// should be encapsulated.
	//
	// Currently, only false is supported.
	//
	Encapsulate *bool `json:"encapsulate"`

	// CIDRGroupRef is a list of CiliumCIDRGroup references.
	// Note: Currently only IPv4 is supported.
	//
	CIDRGroupRef []api.CIDRGroupRef `json:"cidrGroupRef"`
}
```

squeed: Hmm, seems reasonable. @pippolo84, is this a gross misuse of CIDRGroup?

That said, does the extra redirection gain us anything? Just shuffling around bags of CIDRs.

christarazi: It gains us better UX. We are inherently just "shuffling CIDRs" anyway and it is no different than how it is now in the PR. But the user would in theory only need to manage CIDRs in their CIDRGroups. The World CRD just piggybacks off of that. Interoperability, I guess if I were to use one word. 

With interoperability, we can mitigate concerns like

> Ordinarily I would totally agree - I hate API inconsistencies. In this case, however, I would be afraid that CIDRGroup and WorldCIDRGroup might be thought of as two almost-identical mechanisms that differ only in scope.

because it becomes more obvious that the two CRDs are not "identical mechanisms" for the same thing. In fact, the World CRD is explicitly tied to the CIDRGroups, making is easier to say that they are different CRDs altogether.

squeed: Hmm, yeah, that makes sense.

squeed: @christarazi updated based on this suggestion.

pkg/worldcidrs/manager.go
```diff
@@ -0,0 +1,290 @@
...
+
+		controllers: controller.NewManager(),
+	}
+
+	p.Lifecycle.Append(manager)
+
+	return manager
+}
+
+// Start performs the initial synchronization and
```
tklauser: This godoc comment seems incomplete.

pkg/k8s/watchers/watcher.go
```diff
@@ -422,6 +422,7 @@ var ciliumResourceToGroupMapping = map[string]watcherInfo{
 	synced.CRDResourceName(v2alpha1.CCGName):            {start, k8sAPIGroupCiliumCIDRGroupV2Alpha1},
 	synced.CRDResourceName(v2alpha1.L2AnnouncementName): {skip, ""}, // Handled by L2 announcement directly
 	synced.CRDResourceName(v2alpha1.CPIPName):           {skip, ""}, // Handled by multi-pool IPAM allocator
+	synced.CRDResourceName(v2alpha1.CWCIDRName):         {skip, ""}, // Hanled by world-cidr control plane
```
tklauser: nit:
```suggestion
	synced.CRDResourceName(v2alpha1.CWCIDRName):         {skip, ""}, // Handled by world-cidr control plane
```

pkg/worldcidrs/crd.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2alpha1"
+	"github.com/cilium/cilium/pkg/k8s/resource"
+	"github.com/cilium/cilium/pkg/logging/logfields"
+	"github.com/cilium/cilium/pkg/policy/api"
+)
+
+// allCIDRs computes the list of all cidrs requested by all
```
tklauser: nit:

```suggestion
// loadAllCIDRs computes the list of all CIDRs requested by all
```

pkg/worldcidrs/manager.go
```diff
@@ -0,0 +1,290 @@
...
+	"github.com/cilium/cilium/pkg/logging"
+	"github.com/cilium/cilium/pkg/logging/logfields"
+	"github.com/cilium/cilium/pkg/maps/worldcidrsmap"
+	"github.com/cilium/cilium/pkg/option"
+	"github.com/cilium/cilium/pkg/policy/api"
+)
+
+var (
+	log            = logging.DefaultLogger.WithField(logfields.LogSubsys, "world-cidr-manager")
+	controllerName = "world-cidr-manager"
```
tklauser: nit: this could be a const

tklauser (approved):
Nicely grouped into separate commits! LGTM, only nits inline. Feel free to ignore.

install/kubernetes/cilium/Chart.yaml
pippolo84: > Hmm, seems reasonable. @pippolo84, is this a gross misuse of CIDRGroup?
> 
> That said, does the extra redirection gain us anything? Just shuffling around bags of CIDRs.

From a UX perspective, I think the proposal is great. :+1: 

I'm just wondering how much the potentially increased usage of the CIliumCIDRGroup resource might impact the performance of the CNP/CCNP handling, given that we share [a single handler](https://github.com/cilium/cilium/blob/main/pkg/k8s/watchers/cilium_network_policy.go#L107-L204) for CNP/CCNP/CiliumCIDRGroup and we keep an [in-memory cache](https://github.com/cilium/cilium/blob/main/pkg/k8s/watchers/cilium_network_policy.go#L101) for all the CiliumCIDRGroups.
Do we have any kind of estimation on the numbers related to this feature?

Anyway, this should be solved rewriting the handling logic above with the statedb.

pkg/worldcidrs/manager.go
squeed: argh!

install/kubernetes/cilium/Chart.yaml
squeed: Right now, I don't expect this to be particularly large. Probably around 10 or so CIDRs. In fact, relative to heavier-weight things like pods or services, this will be a rounding error.

pkg/worldcidrs/crd.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/cilium/cilium/pkg/k8s/resource"
+	"github.com/cilium/cilium/pkg/logging/logfields"
+	"github.com/cilium/cilium/pkg/policy/api"
+)
+
+// loadAllCIDRs computes the set of cidrs and their groups that are referenced
+// by all WorldCIDRSets.
+func (m *Manager) loadAllCIDRs(setStore resource.Store[*v2alpha1.CiliumWorldCIDRSet], groupStore resource.Store[*v2alpha1.CiliumCIDRGroup]) (sets.Set[netip.Prefix], sets.Set[api.CIDRGroupRef]) {
+	wantPrefixes := make(sets.Set[netip.Prefix], len(m.mapState))
+	groupsInUse := make(sets.Set[api.CIDRGroupRef], len(m.groupsInUse))
```
christarazi, line 24: TIL about this type, very clean!

pkg/worldcidrs/manager.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	log.WithFields(logrus.Fields{
+		"added":   toAdd.UnsortedList(),
+		"removed": toRemove.UnsortedList(),
+	}).Debugf("Pushing changes in to BPF map %s", worldcidrsmap.MapName4)
+
+	// The order of the next 2 function calls matters, as by first adding
+	// missing CIDRs and only then removing obsolete ones we make sure there
+	// will be no connectivity disruption.
```
christarazi, line 285: :ok_hand:

test/k8s/manifests/high-scale-ipcache.yaml
```diff
@@ -108,3 +108,21 @@ spec:
...
         serviceName: http
+---
+apiVersion: cilium.io/v2alpha1
+kind: CiliumWorldCIDRSet
+metadata:
+  name: test-world-cidrs
+spec:
+  encapsulate: false
+  cidrGroupRef:
+  - no-encap-cidrs
```
christarazi, line 119: Nit: We should make this plural to be consistent with the fact that it's `[]CIDRGroupRef`.

install/kubernetes/cilium/Chart.yaml
christarazi: Thanks for implementing! Very clean!

Now I guess back to the hard question: should we rename to `CiliumWorldCIDRGroups` now that it references CIDR groups? :)

christarazi: Btw, small typo in the commit msg: mentions `cidrs` as the key in the CWCG, but now it's `cidrGroupRefs`.

pkg/worldcidrs/crd.go
```diff
@@ -0,0 +1,88 @@
...
+
+func parseCWCIDR(cwcidr *v2alpha1.CiliumWorldCIDRSet) (groupNames []api.CIDRGroupRef, err error) {
+	name := cwcidr.ObjectMeta.Name
+	if name == "" {
+		err = fmt.Errorf("CiliumWorldCIDRSet must have a name")
+		return
+	}
+
+	if cwcidr.Spec.Encapsulate != nil && *cwcidr.Spec.Encapsulate {
+		err = fmt.Errorf("invalid CiliumWorldCIDR: spec.encapsulate must be false")
```
tommyp1ckles, line 68: Why is it necessary to have this field?

pkg/worldcidrs/crd.go
```diff
@@ -0,0 +1,88 @@
...
+			if groupsInUse.Has(groupRef) {
+				continue
+			}
+
+			group, _, _ := groupStore.GetByKey(resource.Key{Name: string(groupRef)})
+			if group == nil {
+				log.WithError(err).WithFields(logrus.Fields{
+					logfields.CIDRGroupRef:           groupRef,
+					logfields.CiliumWorldCIDRSetName: cidrSet.Name,
+				}).Warning("Could not find CiliumCIDRGroup referenced by CiliumWorldCIDRSet")
```
tommyp1ckles, line 44: Not sure if there's anything we should do differently - but the one thing I wonder is it may be hard to tell if the CIDRSet/Group got correctly plumbed into the Agent state. The only indication is a couple warnings, it's also possible for partial applies to go through which may cause confusing issues?

Having a k8s status for this is probably overkill though... 

I wonder if this should be an error at least?

pkg/worldcidrs/crd.go
```diff
@@ -0,0 +1,88 @@
...
+	return cwcidr.Spec.CIDRGroupRef, nil
+}
+
+func parseCIDRGroup(group *v2alpha1.CiliumCIDRGroup) []netip.Prefix {
+	prefixes := make([]netip.Prefix, 0, len(group.Spec.ExternalCIDRs))
+	for _, cidrString := range group.Spec.ExternalCIDRs {
+		prefix, err := netip.ParsePrefix(string(cidrString))
+		if err != nil {
+			err := fmt.Errorf("invalid cidr %s: %w", cidrString, err)
+			log.WithField(logfields.CIDRGroupRef, group.Name).WithError(err).Warningf("Could not parse CIDR in CiliumCIDRGroup")
```
tommyp1ckles, line 81: Considering that the cidrgroup CRDs have a regex validation, a failure here seems worrying.  Perhaps should be an error?

pkg/maps/worldcidrsmap/worldcidrs.go
```diff
@@ -114,11 +129,15 @@ func (m *worldCIDRsMap) Add(cidrs ...netip.Prefix) error {
 	vals := make([]WorldCIDRVal, 0, len(cidrs))
 
 	for _, cidr := range cidrs {
+		// TODO: ipv6 support
+		if cidr.Addr().Is6() {
```
tommyp1ckles, line 133: should we handle/log this case further up the stack?

tommyp1ckles (commented):
Overall everything checks out for me,  very nice to read code 👍

Just a few questions here and there..

pkg/worldcidrs/crd.go
squeed: Good question. Right now, the initial driver of this is an end-user that wants to have all cidrs in-clustermesh by default, and cidrs set as external / non-encapsulated as needed.

I put this field in to allow for the opposite: assume that cidrs are external, and then in-clustermesh cidrs can be specified. The datapath plumbing for this does not yet exist, but without this, the CRD is overfitting a single use-case.

test/k8s/manifests/high-scale-ipcache.yaml
squeed: totally. RIP my green CI :p.

pkg/worldcidrs/crd.go
squeed: We can't put a status, since this field is read by all the agents :-(. But yeah, I can upgrade this to an error.

squeed: Given that we wait for both stores to sync before proceeding, we can't have partial applies to things we've already synced. So, we should be good in that regard.

pkg/worldcidrs/crd.go
squeed: I'm not sure that the regex is perfect; I think you can do `999.999.999.999/99`.

pkg/maps/worldcidrsmap/worldcidrs.go
squeed: Yeah, we should probably warn in the parsing code.

install/kubernetes/cilium/Chart.yaml
squeed: > should we rename to CiliumWorldCIDRGroups

No, since I don't want to somehow conflate these with CiliumCIDRGroups. If anything, I'd call it CiliumCIDRGroupRoute, but I *also* don't want to pretend that this does any routing at all.

I fully expect this type to be deprecated and removed in a few releases.

pkg/worldcidrs/crd.go
tklauser: Probably not for this PR but food for thought: could we use a `// +kubebuilder:validation:Format=cidr` validation on the CRD instead of the regex validation? This is used for other types where a CIDR is expected, e.g. most recently in the `CiliumPodIPPool` CRD.

pkg/worldcidrs/crd.go
squeed: Actually, this shouldn't be an error, just a warning: it means something is misconfigured but we will eventually coalesce.

pkg/maps/worldcidrsmap/worldcidrs.go
squeed: yeah, good idea.

squeed:
@tommyp1ckles @christarazi updated based on feedback, thanks!

pkg/worldcidrs/crd.go
squeed: I assume that CiliumCIDRGroup does that; is it a perfect matching CIDR regex?

Either way, this is unlikely and means we should just skip the entry.

aspsk (approved):
Map changes look ok

pkg/worldcidrs/crd.go
tommyp1ckles: Validation is hard... Fair enough.

pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumworldcidrsets.yaml
```diff
@@ -0,0 +1,86 @@
...
+      type: string
+    - jsonPath: .metadata.creationTimestamp
+      name: Age
+      type: date
+    name: v2alpha1
+    schema:
+      openAPIV3Schema:
+        description: CiliumWorldCIDRSet is a set of CIDRs and their routing destination.
+          It is used in certain modes (e.g. highscale-IPCache) to indicate whether
+          destinations require encapsulation.
```
tommyp1ckles, line 36: nit: the use of "encapsulation" can be a bit ambiguous, especially in the context of networking software. Might be clearer to specify _what_ encapsulation we're talking about here (i.e. vxlan overlay?)

pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumworldcidrsets.yaml
```diff
@@ -0,0 +1,86 @@
...
+                description: 'CIDRGroupRefs is a list of CiliumCIDRGroup references.
+                  The encapsulation configuration will apply to all CIDRs referenced
+                  by these groups. Note: Currently only IPv4 is supported.'
+                items:
+                  description: CIDRGroupRef is a reference to a CIDR Group. A CIDR
+                    Group is a list of CIDRs whose IP addresses should be considered
+                    as a same entity when applying fromCIDRGroupRefs policies on incoming
+                    network traffic.
+                  maxLength: 253
+                  pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
```
tommyp1ckles, line 64: thought: to the discussion about validation, we could in theory  make this more strict to reject 999 type scenarios at the apiserver level.

pkg/worldcidrs/crd.go
tommyp1ckles: Just wondering.  It does seem like a pretty trivial CRD schema upgrade here if we ever need this, is there a plan to implement such functionality?

tommyp1ckles (approved):
Just a some random nitpicks/questions. But if we're all happy with the CRD naming then the changes LGTM

PR #26031, opened 2023-06-08
mutual-auth: Support spire k8s service dns resolution

sayboras:
### Description

### Testing

Testing with the temp commit, the conformance-e2e is running successfully.

https://github.com/cilium/cilium/actions/runs/5220420030/jobs/9423492566?pr=26031

### DIFF
--- a/install/kubernetes/cilium/values.yaml.tmpl
+++ b/install/kubernetes/cilium/values.yaml.tmpl
@@ -3048,8 +3048,13 @@ authentication:
               country: "US"
               organization: "SPIRE"
               commonName: "Cilium SPIRE CA"
-      # -- SPIRE server address
-      serverAddress: spire-server.cilium-spire.svc:8081
+      # -- SPIRE server address used by Cilium Operator
+      #
+      # If k8s Service DNS along with port number is used (e.g. <service-name>.<namespace>.svc(.*):<port-number> format),
+      # Cilium Operator will resolve its address by looking up the clusterIP from Service resource.
+      #
+      # Example values: 10.0.0.1:8081, spire-server.cilium-spire.svc:8081
+      serverAddress: ~
       # -- SPIFFE trust domain to use for fetching certificates
       trustDomain: spiffe.cilium
       # -- SPIRE socket path where the SPIRE delegated api agent is listening
--- a/operator/auth/spire/client.go
+++ b/operator/auth/spire/client.go
@@ -6,6 +6,7 @@ package spire
 import (
 	"context"
 	"fmt"
+	"net"
 	"strings"
 	"time"
 
@@ -18,11 +19,13 @@ import (
 	"github.com/spiffe/spire-api-sdk/proto/spire/api/types"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/cilium/cilium/operator/auth/identity"
 	"github.com/cilium/cilium/pkg/backoff"
 	"github.com/cilium/cilium/pkg/hive"
 	"github.com/cilium/cilium/pkg/hive/cell"
+	k8sClient "github.com/cilium/cilium/pkg/k8s/client"
 	"github.com/cilium/cilium/pkg/logging/logfields"
 )
 
@@ -87,21 +90,30 @@ func (cfg ClientConfig) Flags(flags *pflag.FlagSet) {
 		"The trust domain for the SPIFFE identity.")
 }
 
+type params struct {
+	cell.In
+
+	K8sClient k8sClient.Clientset
+}
+
 type Client struct {
 	cfg   ClientConfig
 	log   logrus.FieldLogger
 	entry entryv1.EntryClient
+
+	k8sClient k8sClient.Clientset
 }
 
 // NewClient creates a new SPIRE client.
 // If the mutual authentication is not enabled, it returns a noop client.
-func NewClient(lc hive.Lifecycle, cfg ClientConfig, log logrus.FieldLogger) identity.Provider {
+func NewClient(params params, lc hive.Lifecycle, cfg ClientConfig, log logrus.FieldLogger) identity.Provider {
 	if !cfg.MutualAuthEnabled {
 		return &noopClient{}
 	}
 	client := &Client{
-		cfg: cfg,
-		log: log.WithField(logfields.LogSubsys, "spire-client"),
+		k8sClient: params.K8sClient,
+		cfg:       cfg,
+		log:       log.WithField(logfields.LogSubsys, "spire-client"),
 	}
 
 	lc.Append(hive.Hook{
@@ -135,6 +147,14 @@ func (c *Client) connect(ctx context.Context) (*grpc.ClientConn, error) {
 	timeoutCtx, cancelFunc := context.WithTimeout(ctx, c.cfg.SpireServerConnectionTimeout)
 	defer cancelFunc()
 
+	resolvedTarget, err := resolvedK8sService(ctx, c.k8sClient, c.cfg.SpireServerAddress)
+	if err != nil {
+		c.log.WithError(err).
+			WithField(logfields.URL, c.cfg.SpireServerAddress).
+			Warning("Unable to resolve SPIRE server address, using original value")
+		resolvedTarget = &c.cfg.SpireServerAddress
+	}
+
 	// This is blocking till the cilium-operator is registered in SPIRE.
 	source, err := workloadapi.NewX509Source(timeoutCtx,
 		workloadapi.WithClientOptions(
@@ -152,11 +172,20 @@ func (c *Client) connect(ctx context.Context) (*grpc.ClientConn, error) {
 	}
 
 	tlsConfig := tlsconfig.MTLSClientConfig(source, source, tlsconfig.AuthorizeMemberOf(trustedDomain))
-	conn, err := grpc.Dial(c.cfg.SpireServerAddress, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)))
+
+	c.log.WithFields(logrus.Fields{
+		logfields.Address: c.cfg.SpireServerAddress,
+		logfields.IPAddr:  resolvedTarget,
+	}).Info("Trying to connect to SPIRE server")
+	conn, err := grpc.Dial(*resolvedTarget, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)))
 	if err != nil {
 		return nil, fmt.Errorf("failed to create connection to SPIRE server: %w", err)
 	}
 
+	c.log.WithFields(logrus.Fields{
+		logfields.Address: c.cfg.SpireServerAddress,
+		logfields.IPAddr:  resolvedTarget,
+	}).Info("Connected to SPIRE server")
 	return conn, nil
 }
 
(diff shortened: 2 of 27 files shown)


squeed:
One point of confusion -- the PR seems to suggest that we need `ClusterFirstWithHostNet` if we wish to reach SPIRE via the service hostname. That also seems to be the default. So, do we need to set `ClusterFirstWithHostNet` automatically?

operator/auth/spire/client.go
```diff
@@ -277,6 +302,41 @@ func (c *Client) listEntries(ctx context.Context, id string) (*entryv1.ListEntri
...
+	// return the original address if it's not a Kubernetes service
+	name, namespace, err := splitK8sService(host)
+	if err != nil {
+		return nil, err
+	}
+	svc, err := client.CoreV1().Services(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	res := svc.Spec.ClusterIP + ":" + port
```
aanm: `net.JoinHostPort(svc.Spec.ClusterIP, port)`

aanm (changes requested):

aanm (approved):
One thing that I noticed is that it seems that we are duplicating some code for the resolution of k8s service since the code already exists for example https://github.com/cilium/cilium/blob/89bb0a3d7c1e339914fda4c9080fb9f0534d537e/operator/cmd/root.go#L506-L508

Could we avoid the code duplication?

operator/auth/spire/client.go
```diff
@@ -152,7 +172,12 @@ func (c *Client) connect(ctx context.Context) (*grpc.ClientConn, error) {
...
-	conn, err := grpc.Dial(c.cfg.SpireServerAddress, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)))
+
+	c.log.WithFields(map[string]interface{}{
+		"spire_service_address": c.cfg.SpireServerAddress,
+		"resolved_address":      resolvedTarget,
+	}).Info("Trying to connect to SPIRE server")
+	conn, err := grpc.Dial(*resolvedTarget, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)))
 	if err != nil {
 		return nil, fmt.Errorf("failed to create connection to SPIRE server: %w", err)
 	}
```
christarazi: Nit: Might be better to log after a successful connection. Also, please use `pkg/logging/logfields` for the fields. Some are already defined that we can reuse.

operator/auth/spire/client.go
```diff
@@ -277,6 +302,41 @@ func (c *Client) listEntries(ctx context.Context, id string) (*entryv1.ListEntri
...
+
+	// return the original address if it's not a Kubernetes service
+	name, namespace, err := splitK8sService(host)
+	if err != nil {
+		return nil, err
+	}
+	svc, err := client.CoreV1().Services(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
```
christarazi: Could we use the services informer / store here instead of making direct API calls?

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -3018,8 +3018,13 @@ authentication:
               country: "US"
               organization: "SPIRE"
               commonName: "Cilium SPIRE CA"
-      # -- SPIRE server address
-      serverAddress: spire-server.cilium-spire.svc:8081
+      # -- SPIRE server address used by Cilium Operator
+      #
+      # As Cilium Operator is running in host network, please adjust operator.dnsPolicy to
+      # ClusterFirstWithHostNet if DNS resolution is required.
```
squeed: This needs expansion, since the operator now looks up service names.

operator/auth/spire/client.go
```diff
@@ -152,7 +172,12 @@ func (c *Client) connect(ctx context.Context) (*grpc.ClientConn, error) {
 	}
 
 	tlsConfig := tlsconfig.MTLSClientConfig(source, source, tlsconfig.AuthorizeMemberOf(trustedDomain))
-	conn, err := grpc.Dial(c.cfg.SpireServerAddress, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)))
+
+	c.log.WithFields(map[string]interface{}{
+		"spire_service_address": c.cfg.SpireServerAddress,
+		"resolved_address":      resolvedTarget,
+	}).Info("Trying to connect to SPIRE server")
+	conn, err := grpc.Dial(*resolvedTarget, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)))
```
squeed, line 180: Aside: I was going to suggest proving a custom Dialer with the resolver overridden, but the current state of the Go module makes that much more difficult than necessary.

operator/auth/spire/client.go
```diff
@@ -277,6 +302,41 @@ func (c *Client) listEntries(ctx context.Context, id string) (*entryv1.ListEntri
 	})
 }
 
+// resolvedK8sService resolves the given address to the IP address.
+// This is the best effort to resolvedK8sService the Kubernetes service DNS to IP address.
+func resolvedK8sService(ctx context.Context, client k8sClient.Clientset, address string) (*string, error) {
```
squeed, line 312: To minimize surprising behavior, let's only support names of the format `<foo>.<bar>.svc(.the rest)` for automatic translation. That way we won't ever accidentally rewrite a name incorrectly. This should be documented in the Helm values.

operator/auth/spire/client.go
squeed: Though if that causes an annoying startup ordering issue, a single GET call isn't too bad.

operator/auth/spire/client.go
```diff
@@ -277,6 +302,41 @@ func (c *Client) listEntries(ctx context.Context, id string) (*entryv1.ListEntri
...
+	res := net.JoinHostPort(svc.Spec.ClusterIP, port)
+	return &res, nil
+}
+
+func splitK8sService(host string) (string, string, error) {
+	names := strings.Split(host, ".")
+	if len(names) >= 2 {
+		return names[0], names[1], nil
+	}
+	return "", "", fmt.Errorf("invalid service name. expecting <name>.<namespace>[optional], got: %s", host)
```
squeed: Don't return an error here; an empty name + namespace should be fine.

operator/auth/spire/client.go
squeed: Also, this function seems to have funny blind spots. I would re-structure it to do

- Is the name of the form `<foo>.<bar>.svc`? if so, try to rewrite.
- Otherwise, keep the name as-is.

operator/auth/spire/client.go
```diff
@@ -135,6 +147,14 @@ func (c *Client) connect(ctx context.Context) (*grpc.ClientConn, error) {
 	timeoutCtx, cancelFunc := context.WithTimeout(ctx, c.cfg.SpireServerConnectionTimeout)
 	defer cancelFunc()
 
+	resolvedTarget, err := resolvedK8sService(ctx, c.k8sClient, c.cfg.SpireServerAddress)
```
squeed, line 150: This function should only return an error when there is a genuine problem. In other words, only log if the supplied domain name is of the format `<foo>.<bar>.svc`, and we fail to look it up because the apiserver goes to bed or the name isn't found.

squeed (changes requested):

sayboras:
> If the operator needs to run on the same node as the SPIRE agent, then that should be expressed with PodAffinity. Or, we should include the spire container in the operator pod.

This was one of the approaches that we considered, however, cilium operator is running in HA mode with the leader election, so it's hard to control the node for both cilium operator and spire server.

operator/auth/spire/client.go
sayboras: thanks :+1:

operator/auth/spire/client.go
sayboras: yes, it's one of the reasons that I am using k8s client directly. Also, informer is good if we are watching multiple services. In this case, it's only limited to single one only.

install/kubernetes/cilium/values.yaml.tmpl
sayboras: It seems like this description is causing confusion now, removing the mention of operator.dnsPolicy.

operator/auth/spire/client.go
sayboras: thanks, this should be addressed along with other comments.

operator/auth/spire/client.go
sayboras: good point, the implementation seems much simpler and easier to read :+1:

operator/auth/spire/client.go
```diff
@@ -152,11 +172,20 @@ func (c *Client) connect(ctx context.Context) (*grpc.ClientConn, error) {
 	}
 
 	tlsConfig := tlsconfig.MTLSClientConfig(source, source, tlsconfig.AuthorizeMemberOf(trustedDomain))
-	conn, err := grpc.Dial(c.cfg.SpireServerAddress, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)))
+
+	c.log.WithFields(map[string]interface{}{
+		logfields.URL:    c.cfg.SpireServerAddress,
+		logfields.IPAddr: resolvedTarget,
+	}).Info("Trying to connect to SPIRE server")
```
christarazi: URL -> `logfields.Address` is probably better. Also `WithFields()` is usually called with `WithFields(logrus.Fields{...})`

christarazi (approved):

zacharysarah (approved):
LGTM for docs

squeed (approved):

sayboras:
Force push to address the log comment

<details>
<summary>Diff</summary>

```diff
diff --git a/operator/auth/spire/client.go b/operator/auth/spire/client.go
index 6af073b4c6..6d59e299e3 100644
--- a/operator/auth/spire/client.go
+++ b/operator/auth/spire/client.go
@@ -173,18 +173,18 @@ func (c *Client) connect(ctx context.Context) (*grpc.ClientConn, error) {
 
 	tlsConfig := tlsconfig.MTLSClientConfig(source, source, tlsconfig.AuthorizeMemberOf(trustedDomain))
 
-	c.log.WithFields(map[string]interface{}{
-		logfields.URL:    c.cfg.SpireServerAddress,
-		logfields.IPAddr: resolvedTarget,
+	c.log.WithFields(logrus.Fields{
+		logfields.Address: c.cfg.SpireServerAddress,
+		logfields.IPAddr:  resolvedTarget,
 	}).Info("Trying to connect to SPIRE server")
 	conn, err := grpc.Dial(*resolvedTarget, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)))
 	if err != nil {
 		return nil, fmt.Errorf("failed to create connection to SPIRE server: %w", err)
 	}
 
-	c.log.WithFields(map[string]interface{}{
-		logfields.URL:    c.cfg.SpireServerAddress,
-		logfields.IPAddr: resolvedTarget,
+	c.log.WithFields(logrus.Fields{
+		logfields.Address: c.cfg.SpireServerAddress,
+		logfields.IPAddr:  resolvedTarget,
 	}).Info("Connected to SPIRE server")
 	return conn, nil
 }
diff --git a/pkg/logging/logfields/logfields.go b/pkg/logging/logfields/logfields.go
index c4796bf951..3dcd56051d 100644
--- a/pkg/logging/logfields/logfields.go
+++ b/pkg/logging/logfields/logfields.go
@@ -121,6 +121,9 @@ const (
 	// NextHop is an IPV4 or IPv6 address for the next hop
 	NextHop = "nextHop"
 
+	// Address is an IPV4, IPv6 or FQDN address
+	Address = "address"
+
 	// IPAddr is an IPV4 or IPv6 address
 	IPAddr = "ipAddr"
 
```

</details>

sayboras:
Force pushed to rebase with main, and drop the last commit as cilium-cli version got upgraded in main branch now.

sayboras:
Reviews from all required teams are in, all tests passed, merging.

PR #26033, opened 2023-06-08
docu: add section about envoy daemonset deployment

mhofstetter:
This PR adds documentation for the new funcitonality to deploy the CIlium Envoy proxy as separate DaemonSet (https://github.com/cilium/cilium/pull/25081).

### DIFF
--- a/Documentation/security/network/proxy/envoy.rst
+++ b/Documentation/security/network/proxy/envoy.rst
@@ -16,6 +16,50 @@ for the cluster. Cilium proxy is distributed within the Cilium images.
 
 For more information on the version compatibility matrix, see `Cilium Proxy documentation <https://github.com/cilium/proxy#version-compatibility-matrix>`_.
 
+***********************
+Deployment as DaemonSet
+***********************
+
+Background
+==========
+
+When Cilium L7 functionality (Ingress, Gateway API, Network Policies with L7 functionality, L7 Protocol Visibility)
+is enabled or installed in a Kubernetes cluster, the Cilium agent starts an Envoy proxy as separate process within
+the Cilium agent pod.
+
+That Envoy proxy instance becomes responsible for proxying all matching L7 requests on that node.
+As a result, L7 traffic targeted by policies depends on the availability of the Cilium agent pod.
+
+Alternatively, it's possible to deploy the Envoy proxy as independently life-cycled DaemonSet called ``cilium-envoy``
+instead of running it from within the Cilium Agent Pod.
+
+The communication between Cilium agent and Envoy proxy takes place via UNIX domain sockets in both deployment modes.
+Be that streaming the access logs (e.g. L7 Protocol Visibility), updating the configuration via
+`xDS <https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol>`_ or accessing the admin interface.
+
+Enable and configure Envoy DaemonSet
+====================================
+
+To enable the dedicated Envoy proxy DaemonSet, install Cilium with the Helm value ``envoy.enabled`` set to ``true``.
+
+Please see the :ref:`helm_reference` (keys with ``envoy.*``) for detailed information on how to configure the Envoy proxy DaemonSet.
+
+Potential Benefits
+==================
+
+- Cilium Agent restarts (e.g. for upgrades) without impacts for the live traffic proxied via Envoy.
+- Envoy patch release upgrades without impacts for the Cilium Agent.
+- Separate CPU and memory limits for Envoy and Cilium Agent for performance isolation.
+- Envoy application log not mixed with the one of the Cilium Agent.
+- Dedicated health probes for the Envoy proxy.
+- Explicit deployment of Envoy proxy during Cilium installation (compared to on demand in the embedded mode).
+
+Known Limitations
+=================
+
+- Due to Pod-to-Pod communication with the Cilium Agent via UNIX domain sockets, Envoy DaemonSet isn't supported with SELinux enabled on the host. This is the default for Red Hat OpenShift.
+- Go Extensions (ProxyLib) aren't supported yet.
+
 *************
 Go Extensions
 *************


Documentation/security/network/proxy/envoy.rst
```diff
@@ -16,6 +16,49 @@ for the cluster. Cilium proxy is distributed within the Cilium images.
...
+
+Background
+==========
+
+When Cilium L7 functionality (Ingress, Gateway API, Network Policies with L7 functionality, L7 Protocol Visibility)
+is enabled or installed in a Kubernetes cluster, the Cilium agent starts an Envoy proxy as separate process within
+the Cilium agent pod.
+
+That Envoy proxy instance becomes responsible for proxying all matching L7 requests on that node.
+As a result, L7 traffic targeted by policies depend on the availability of the Cilium agent pod.
```
qmonnet:
```suggestion
As a result, L7 traffic targeted by policies depends on the availability of the Cilium agent pod.
```

Documentation/security/network/proxy/envoy.rst
```diff
@@ -16,6 +16,49 @@ for the cluster. Cilium proxy is distributed within the Cilium images.
...
+
+With Cilium 1.14 it's possible to deploy the Envoy proxy as independently life-cycled DaemonSet called ``cilium-envoy``
```
qmonnet: I'm not sure we need the Cilium version number here, since this text will only be available starting from the docs version accompanying 1.14.

I think this was discussed elsewhere before, but I can't remember the outcome :disappointed:. @zacharysarah do you remember, or do you have an opinion, about mentioning the version number here, knowing this will only be in version 1.14+ of the docs?

Documentation/security/network/proxy/envoy.rst
```diff
@@ -16,6 +16,49 @@ for the cluster. Cilium proxy is distributed within the Cilium images.
...
+The communication between Cilium agent and Envoy proxy takes place via UNIX domain sockets in both deployment modes.
+Be that streaming the access logs (e.g. L7 Protocol Visibility), updating the configuration via
+`xDS <https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol>`_ or accessing the admin interface.
+
+Enable and configure Envoy DaemonSet
+====================================
+
+To enable the dedicated Envoy proxy DaemonSet, install Cilium with the Helm value ``envoy.enabled`` set to ``true``.
+
+Please see the :ref:`helm_reference` (Keys with ``envoy.*``) for detailed information on how to configure the Envoy proxy DaemonSet.
```
qmonnet:
```suggestion
Please see the :ref:`helm_reference` (keys with ``envoy.*``) for detailed information on how to configure the Envoy proxy DaemonSet.
```

Documentation/security/network/proxy/envoy.rst
```diff
@@ -16,6 +16,49 @@ for the cluster. Cilium proxy is distributed within the Cilium images.
...
+
+Potential Benefits
+==================
```
qmonnet, line 48:
```suggestion
==================

```

Documentation/security/network/proxy/envoy.rst
```diff
@@ -16,6 +16,49 @@ for the cluster. Cilium proxy is distributed within the Cilium images.
...
+* Restarting Cilium Agent for upgrades does not impact live traffic proxied via Envoy.
+* Applying Envoy patch releases doesn't affect Cilium Agent.
+* Performance isolation between Envoy and Cilium Agent with the possibility to define separate CPU and memory limits.
+* Envoy application log doesn't get mixed with the one of the Cilium Agent.
+* Dedicated health probes for the Envoy proxy.
+* Explicit deployment of Envoy proxy during Cilium installation (compared to on demand in the embedded mode)
```
qmonnet: Please make the list consistent (with verbs in all items, or none; and with consistent punctuation at the end of the items).

Let's also use `-` rather than `*` for item bullets. I think both work, but the rest of this doc uses dashes.

Documentation/security/network/proxy/envoy.rst
```diff
@@ -16,6 +16,49 @@ for the cluster. Cilium proxy is distributed within the Cilium images.
...
+
+Known Limitations
+=================
+
+- Due to Pod-to-Pod communication with the Cilium Agent via UNIX domain sockets, there might be issues when running on Red Hat OpenShift.
```
qmonnet: `there might be` -> Do we know more than that, by any chance? Is it known to be incompatible, or we just haven't tested it?

qmonnet (changes requested):
Looks good overall, with a few nits and questions. Thanks!

Documentation/security/network/proxy/envoy.rst
mhofstetter: @qmonnet thanks - applied.

Documentation/security/network/proxy/envoy.rst
mhofstetter: @qmonnet thanks - applied.

Documentation/security/network/proxy/envoy.rst
mhofstetter: @qmonnet thanks - i tried to come up with a more consistent update. PTAL :pray:

Documentation/security/network/proxy/envoy.rst
mhofstetter: @qmonnet without having it tested explicitly - we know from other cases that it won't work with SELinux enabled on the host. And this is the default for Red Hat OpenShift. i tried to rephrase it

mhofstetter:
@qmonnet thanks for your review. the latest force-push addresses your feedback.

qmonnet (approved):
Thanks, look good now :+1:.

For next time: No need to @-mention me on every reply :)

jrajahalme (approved):

Documentation/security/network/proxy/envoy.rst
mhofstetter: Agree that the version isn't necessary - as this should be already given by the context / version of the documentation itself.

Changed it towards mentioning it as an _alternative_.

mhofstetter:
Reviews are in. Marking as `ready-to-merge` given that this is a docs-only PR.

PR #26038, opened 2023-06-08
ctmap: turn per-cluster CT maps into HashOfMaps

ti-mo:
This commit turns the tcp4/any4/tcp6/any6 outer maps into HashOfMaps instead of ArrayOfMaps. This makes iterating them significantly cheaper, as arrays are sparse and hashes are not.

Getting all inner maps is now done by iterating the outer maps until completion and opening all inner maps by their BPF ids.

This means inner maps no longer need to be pinned to bpffs, which avoids future headaches when any of these maps are unavoidably changed at some point.

```release-note
ctmap: turn per-cluster CT maps into HashOfMaps
```


ti-mo:
Closing this for now as this has already shipped and migration code needs to be added anyway.

PR #25809, opened 2023-05-31
BGPv1: Adds Peer Port Support

danehans:
A user can annotate a node with `local-port` to have BGPRouterManager listen on a port other than 179. However, the BgpServerInstance only creates a peering connection using port 179. This means either passive mode is required when one node specifies a non standard port or two Cilium nodes cannot establish a BGP neighbor relationship when both specify a non-standard port.

This PR adds an optional `PeerPort` field to the CiliumBGPVirtualRouter API type, allowing a user to specify the BGP peer's TCP port. When unspecified, the BgpServerInstance continues to use port 179.

Fixes: #24737 
Fixes: #25683 

```release-note
Adds `peerPort` field to CiliumBGPPeeringPolicy for specifying the port of a BGP neighbor. If unspecified, port 179 is used.
```

PR #25708, opened 2023-05-26
bgpv1: Add support for eBGP-multihop in BGP control plane

rastislavs:
Extends the `CiliumBGPNeighbor` configuration in the `CiliumBGPPeeringPolicy` CRD with a new configuration option: `EBGPMultihopTTL`. This can be used to enable multihop feature for eBGP peers, with explicit TTL hop limit count.

Example Configuration:
```yaml
apiVersion: "cilium.io/v2alpha1"
kind: CiliumBGPPeeringPolicy
metadata:
  name: tor
spec:
  virtualRouters:
  - localASN: 65001
    exportPodCIDR: true
    neighbors:
    - peerAddress: "172.0.0.1/32"
      peerASN: 65000
      eBGPMultihopTTL: 10   # >=1 means enabled, 0 means disabled, default=0.
```

Example state dump of peers after configuring:
```
# cilium bgp peers -o json
[
  {
    "applied-hold-time-seconds": 90,
    "applied-keep-alive-time-seconds": 30,
    "configured-hold-time-seconds": 90,
    "configured-keep-alive-time-seconds": 30,
    "connect-retry-time-seconds": 60,
    "ebgp-multihop-ttl": 10,
    "families": [
      {
        "advertised": 1,
        "afi": "ipv4",
        "received": 1,
        "safi": "unicast"
      },
      {
        "afi": "ipv6",
        "safi": "unicast"
      }
    ],
    "local-asn": 65001,
    "peer-address": "172.0.0.1",
    "peer-asn": 65000,
    "session-state": "established",
    "uptime-nanoseconds": 53443700101
  }
]
```

Example dump of a keepalive packet:

Screenshot from 2023-05-25 15-50-17

Fixes: #21753

```release-note
Add support for eBGP-multihop configuration for CiliumBGPNeighbor in CiliumBGPPeeringPolicy CRD
```

PR #25408, opened 2023-05-12
bgpv1: Add support for setting BGP timer parameters in CiliumBGPNeighbor CRD

rastislavs:
Extends the `CiliumBGPNeighbor` CRD with 3 new configuration options: `ConnectRetryTime`, `HoldTime` and `KeepAliveTime`. These can be used to fine-tune BGP peering, e.g. to achieve faster failover times. If not set, the default values for the affected BGP timers remain the same as before this change.

Example configuration:
```yaml
apiVersion: "cilium.io/v2alpha1"
kind: CiliumBGPPeeringPolicy
metadata:
  name: tor
spec:
  virtualRouters:
  - localASN: 65001
    exportPodCIDR: true
    neighbors:
    - peerAddress: "172.0.0.1/32"
      peerASN: 65000
      holdTime: 120s
      keepAliveTime: 30s
      connectRetryTime: 60s
```

Example state dump of peers after configuring:
```
# cilium bgp peers -o json
[
  {
    "applied-hold-time-seconds": 120,
    "applied-keep-alive-time-seconds": 30,
    "configured-hold-time-seconds": 120,
    "configured-keep-alive-time-seconds": 30,
    "connect-retry-time-seconds": 60,
    "families": [
      {
        "advertised": 1,
        "afi": "ipv4",
        "received": 1,
        "safi": "unicast"
      },
      {
        "afi": "ipv6",
        "safi": "unicast"
      }
    ],
    "local-asn": 65001,
    "peer-address": "172.0.0.1",
    "peer-asn": 65000,
    "session-state": "established",
    "uptime-nanoseconds": 8144057623
  }
]
```

Implementation notes:
- The default values for the timer params are the same as GoBGP uses internally, so there is no change in behaviour if they are not explicitly set.
- The timer params are defaulted in the `CiliumBGPPeeringPolicy` at the BGP Agent level, so that the same defaults can be consistently applied across all future BGP backends.
 - The BGP `Router` interface has been extended with a new method `UpdateNeighbor`, which is called from the BGP Manager's `neighborReconciler` when a neighbor needs to be updated.

```release-note
Add support for setting BGP timer parameters in CiliumBGPNeighbor CRD
```

PR #26040, opened 2023-06-08
bgpv1: Consolidate CRD API to follow K8s API Conventions

### LINKED ISSUE #25871: BGP CP: Move Default API Values to Type Definitions
Currently, default values are defined in the agent pkg. Instead, move these values to `pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go`. This allows the default value to be versioned with the API. This also provides consistency with the kubebuilder tag that defines the default value. If the value needs to be changed, it's updated in one place.

rastislavs:
This PR aims to consolidate the BGP control plane k8s API to follow the [K8s API Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md).

Although it modifies some filed names, it does it only for fields that have been added recently (not yet released in any Cilium version) as part of:
 - https://github.com/cilium/cilium/pull/25408
 - https://github.com/cilium/cilium/pull/25660
 - https://github.com/cilium/cilium/pull/25708
 - https://github.com/cilium/cilium/pull/25809

Already released API fields are modified only in a backward compatible way (e.g. changed internal type to a pointer, but no type change in the generated `CustomResourceDefinition`). The only changes of already released API in the generated [ciliumbgppeeringpolicies.yaml](https://github.com/cilium/cilium/pull/26040/files#diff-dddec5a0dbbf46faca6e3ea84b23aefe87d43e6ced346582f5d86d362936f204) are:
- `exportPodCIDR`: added default value (`false`, so not changing the default behaviour)
- `localASN` + `peerASN`: added `format` annotation, but without the change of the underlying `type` 

The changes are described in more detail in individual commits. 

Although the diff looks pretty big, most of the changes are in the unit / component tests, where the API is actually used.

Fixes https://github.com/cilium/cilium/issues/25871


pkg/bgpv1/agent/controller.go
```diff
@@ -431,62 +429,51 @@ func (c *Controller) FullWithdrawal(ctx context.Context) {
 }
 
 // applyPolicyDefaults applies default values on the CiliumBGPPeeringPolicy.
+// This defaulting is normally done by kube-apiserver (based on CRD default markers), but ensured here for non-k8s environments, tests, etc.
```
rastislavs: Actually, I am not sure if we still really need this runtime defaulting. Whole `applyPolicyDefaults` could be removed if we make sure we deference pointers carefully with applying defaults (see e.g. `pointer.Int32Deref(n.HoldTimeSeconds, v2alpha1api.DefaultBGPHoldTimeSeconds)`).

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -112,33 +123,32 @@ type CiliumBGPNeighbor struct {
...
+	// +kubebuilder:default=90
+	HoldTimeSeconds *int32 `json:"holdTimeSeconds,omitempty"`
+	// KeepaliveTimeSeconds defines the initial value for the BGP KeepaliveTimer (RFC 4271, Section 8).
+	// It can not be larger than HoldTimeSeconds. Updating this value will cause a session reset.
 	//
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:Format=duration
-	KeepAliveTime metav1.Duration `json:"keepAliveTime,omitempty"`
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:default=30
```
rastislavs, line 158: I thought it would be nicer to default `keepAliveTimeSeconds` to a static value here, not in the backend (we used to default to 1/3 of the HoldTime at runtime). That way all timers are defaulted consistently, e.g. if a user does not specify them in the CR for a neighbor, it would automatically default to this when they retrieve the CR the value:
```
    virtualRouters:
    - exportPodCIDR: true
      localASN: 65001
      neighbors:
      - connectRetryTimeSeconds: 120
        holdTimeSeconds: 90
        keepAliveTimeSeconds: 30
        peerASN: 65000
        peerAddress: 172.0.0.1/32
```

Otherwise `keepAliveTimeSeconds: 30` would be omitted from defaulting on the apiserver level.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -9,6 +9,19 @@ import (
...
+	// DefaultBGPHoldTimeSeconds defines the default initial value for the BGP HoldTimer (RFC 4271, Section 4.2).
+	DefaultBGPHoldTimeSeconds = 90
+	// DefaultBGPKeepAliveTimeSeconds defines the default initial value for the BGP KeepaliveTimer (RFC 4271, Section 8).
+	DefaultBGPKeepAliveTimeSeconds = 30
+	// DefaultBGPGRRestartTimeSeconds defines default Restart Time for graceful restart (RFC 4724, section 4.2)
+	DefaultBGPGRRestartTimeSeconds = 120
+	// DefaultBGPPeerPort defines the TCP port number of a CiliumBGPNeighbor when PeerPort is unspecified.
+	DefaultBGPPeerPort = 179
+)
+
```
rastislavs, line 31: What is still missing is the type change of: 
```go
LocalASN int `json:"localASN"

PeerASN int `json:"peerASN"`
```

I think these should become `int64` to allow for maximum value `4294967295` (extended 32bit ASNs).
(`uint32` would be best, but it seems unsigned integers are not really used in k8s API)

rastislavs: But note that it will probably be a larger change, as ASNs are used in various data structures in the backend, where they are manipulated as `int` (which actually could be an issue on 32bit systems, if Cilium runs on those)

rastislavs:
cc @danehans

pkg/bgpv1/agent/controller_test.go
```diff
@@ -143,11 +139,15 @@ func TestControllerSanity(t *testing.T) {
 				for _, r := range p.Spec.VirtualRouters {
 					for _, n := range r.Neighbors {
 						if n.PeerPort != nil &&
-							*n.PeerPort == types.DefaultPeerPort &&
-							n.ConnectRetryTime.Duration != 0 &&
-							n.HoldTime.Duration != 0 &&
-							n.KeepAliveTime.Duration != 0 &&
-							n.GracefulRestart.RestartTime.Duration != 0 {
+							*n.PeerPort == v2alpha1api.DefaultBGPPeerPort &&
```
harsimran-pabla: Consider refactoring this, this is bit hard to read.

pkg/bgpv1/agent/controller.go
harsimran-pabla: +1 to use `Int32Deref`, this method with a bunch of `if ( ptr == nil || *ptr ==0 ) {...}` will grow as more optional fields are added.

pkg/bgpv1/agent/controller.go
```diff
@@ -431,62 +429,51 @@ func (c *Controller) FullWithdrawal(ctx context.Context) {
...
-				// RFC4271 Sec 4.2 says that the hold time MUST be zero or at least 3 seconds.
-				// However, as GoBGP defaults the hold time for 0 value, it cannot be 0 in our case.
-				return fmt.Errorf("holdTime is lower than %v for peer ASN %d, IP %s", minHoldTime, n.PeerASN, n.PeerAddress)
-			}
-			if n.KeepAliveTime.Duration < 0 {
-				return fmt.Errorf("keepAliveTime is negative for peer ASN %d, IP %s", n.PeerASN, n.PeerAddress)
-			}
-			if n.KeepAliveTime.Duration > n.HoldTime.Duration {
+			// nil checks just in case, as everything should be defaulted in applyPolicyDefaults()
+			if n.KeepAliveTimeSeconds != nil && n.HoldTimeSeconds != nil &&
```
harsimran-pabla: I think we should add validateTimers method and move this logic into it.

pkg/bgpv1/gobgp/server.go
```diff
@@ -222,37 +218,38 @@ func (g *GoBGPServer) getPeerConfig(ctx context.Context, n *v2alpha1api.CiliumBG
...
 	}
 
 	// populate graceful restart config
 	if peer.GracefulRestart == nil {
 		peer.GracefulRestart = &gobgp.GracefulRestart{}
 	}
-	peer.GracefulRestart.Enabled = n.GracefulRestart.Enabled
-	peer.GracefulRestart.RestartTime = uint32(n.GracefulRestart.RestartTime.Round(time.Second).Seconds())
-
+	if n.GracefulRestart != nil {
```
harsimran-pabla: Two if statements can be merged into one - saves some indentation.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -154,7 +164,7 @@ type CiliumBGPVirtualRouter struct {
 	// to the configured neighbors.
 	//
 	// +kubebuilder:validation:Optional
-	ExportPodCIDR bool `json:"exportPodCIDR"`
+	ExportPodCIDR *bool `json:"exportPodCIDR"`
```
harsimran-pabla: Is there going to be any backward compatibility issues?

pkg/bgpv1/manager/reconcile.go
```diff
@@ -378,7 +379,7 @@ func exportPodCIDRReconciler(
 		ctx:       ctx,
 		name:      "pod CIDR",
 		component: "manager.exportPodCIDRReconciler",
-		enabled:   newc.ExportPodCIDR,
+		enabled:   pointer.BoolDeref(newc.ExportPodCIDR, false),
```
harsimran-pabla: Should the default value of ExportPodCIDR be explicitly defined in API? ( No strong opinion on this )

harsimran-pabla (changes requested):

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
harsimran-pabla: Ah read the description again - no concerns if these changes are backward compatible.

pkg/bgpv1/manager/reconcile.go
rastislavs: Yeah not sure either, but as this filed was released in a previous Cilium version, I would prefer the more conservative change to it (so, no?).

rastislavs: From [Changing The API - On Compatibility](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#on-compatibility): "An API change is considered compatible if it: does not change existing semantics, including the semantic meaning of default values and behavior"

But then, our BGP API is still alpha, which means we in theory could do [almost any changes](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions) - but I guess we don't want it.

pkg/bgpv1/agent/controller.go
rastislavs: Did you mean a method of `CiliumBGPNeighbor` in the `v2alpha1` package?

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: Done in https://github.com/cilium/cilium/pull/26040/commits/a811886ed98a9f856105b0fa35c267f4d12be357

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: Actually, I was thinking about this already before, so let me ask this, as now it is the last chance to change if we wanted it:

Would it make sense to wrap all timers config into its own struct (that would be optional in `CiliumBGPNeighbor`), similarly as we already do for `CiliumBGPNeighborGracefulRestart`, e.g.:
```go
type CiliumBGPNeighborTimers struct {
    ConnectRetryTimeSeconds *int32 `json:"connectRetryTimeSeconds,omitempty"`
    HoldTimeSeconds         *int32 `json:"holdTimeSeconds,omitempty"`
    KeepAliveTimeSeconds    *int32 `json:"keepAliveTimeSeconds,omitempty"`
}
```
That way, timers would not be automatically defaulted in the `CiliumBGPNeighbor`, until the user configures some of them. Then all of them would be defaulted. E.g. - without configuring timers:
```go
    virtualRouters:
    - exportPodCIDR: true
      localASN: 65001
      neighbors:
      - peerASN: 65000
        peerAddress: 172.0.0.1/32
```

Once some timer is configured:
```go
    virtualRouters:
    - exportPodCIDR: true
      localASN: 65001
      neighbors:
      - peerASN: 65000
        peerAddress: 172.0.0.1/32
        timers:
          connectRetryTimeSeconds: 120
          holdTimeSeconds: 90
          keepAliveTimeSeconds: 30
```
WDYT @harsimran-pabla & @YutaroHayakawa ? I am not sure, as that would also complicate the defaulting and nil checks in the backend though (e.g. we may still need that `applyPolicyDefaults` logic discussed here: https://github.com/cilium/cilium/pull/26040/files#r1223203259)...

pkg/bgpv1/agent/controller.go
YutaroHayakawa: > if we make sure we deference pointers carefully with applying defaults

Sorry, I didn't get what this means. Does this mean when we dereference the optional fields in the reconcilers, we need to use `Int32Deref`? If so, can't we set the default value at the very beginning and allow reconcilers to dereference without the risk of nil pointer dereference or having an invalid value?

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
YutaroHayakawa: I don't have a strong opinion here, but unlike graceful restart, these timer parameters exist internally anyways, so I'd prefer to always show it to users. Also, the grouping `timer` is too generic I think. For example, we have graceful restart-related timer value, but it belongs to another group.

pkg/bgpv1/manager/reconcile.go
YutaroHayakawa: Is it an incompatible change to set the default to `false` here? If not, I'd prefer to set it for consistency.

YutaroHayakawa (commented):
Thank you very much for woking on this! Now our API design looks very consistent. I left some nit picking comments.

pkg/bgpv1/agent/controller.go
rastislavs: I think it would make a lot of sense, done that in https://github.com/cilium/cilium/pull/26040/commits/956f08bb95b47ccaf134d2851b3ac03949d12cd0

pkg/bgpv1/agent/controller.go
rastislavs: Yeah, I think you understood correctly @YutaroHayakawa . We can do one of the following:

1. always deference with care, e.g. via `pointer.Int32Deref` and we could remove `applyPolicyDefaults()`
2. fully rely on `applyPolicyDefaults()` and always deference directly like `*this`
3. do both  `pointer.Int32Deref` and `applyPolicyDefaults()`, just in case

At the moment we are at the option 3) which I guess is the least ideal?

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: > Also, the grouping timer is too generic I think

good point, maybe we should leave it as it is then

pkg/bgpv1/manager/reconcile.go
rastislavs: > Is it an incompatible change to set the default to false here? 

Hmm, I think it should be fine as the default behavior remains the same. Let me change that.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
harsimran-pabla: No concerns about keeping timers as individual parameters.

pkg/bgpv1/manager/reconcile.go
rastislavs: Done in https://github.com/cilium/cilium/pull/26040/commits/4a064eb83a2fd1070bfdff21f41f0a1669e3ad17

rastislavs: Done in https://github.com/cilium/cilium/pull/26040/commits/aa3458df112c4d33516e0e281679b3a264c8062a

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -97,48 +113,47 @@ type CiliumBGPNeighbor struct {
...
 	// When empty or zero, eBGP multi-hop feature is disabled. The value is ignored for iBGP peers.
 	//
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:validation:Maximum=255
-	EBGPMultihopTTL int `json:"eBGPMultihopTTL,omitempty"`
-	// ConnectRetryTime defines the initial value for the BGP ConnectRetryTimer (RFC 4271, Section 8).
-	// The default value for the ConnectRetryTime (if empty or zero) is 120 seconds.
-	// Rounded internally to the nearest whole second.
+	EBGPMultihopTTL *int32 `json:"eBGPMultihopTTL,omitempty"`
```
rastislavs, line 136: Note that `EBGPMultihopTTL` is not defaulted in the API - I thought it would not make sense to default it, as it has no meaning for iBGP peers. Does it make sense or should we default it to `0` as well?

pkg/bgpv1/agent/controller.go
rastislavs: We discussed this a bit more with @YutaroHayakawa and decided to allow direct pointer dereferences. Apart from making it easier for reconciler developers, it is also safer if we always default before passing the API structs to the reconcilers (in case some developer would forget a nil check / safe dereference). In https://github.com/cilium/cilium/pull/26040/commits/e7faf3704e65edfd32559863310722a417043ad0 I cleaned this up and moved the defaulting logic to the API package, so that it can be used from other packages as well (was needed for unit tests).

harsimran-pabla: Sounds good to me.

rastislavs:
Thanks for all the comments so far @YutaroHayakawa & @harsimran-pabla! As I addressed them, I squashed relevant commits and moved this PR out of draft - ready for the final review!

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
danehans: Why is this field singular if a node can have multiple [PodCIDRs](https://github.com/cilium/cilium/blob/main/pkg/k8s/apis/cilium.io/client/crds/v2/ciliumnodes.yaml#L308-L314)?

From the k8s API conventions doc:
> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options.

Have you considered how this field could evolve where a bool type could be restrictive?

Although the field is being defaulted, I believe the `omitempty` tag should be used.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -65,19 +83,17 @@ type CiliumBGPPeeringPolicySpec struct {
 type CiliumBGPNeighborGracefulRestart struct {
 	// Enabled flag, when set enables graceful restart capability.
 	//
-	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Required
```
danehans, line 88: `Enabled` is a bool type. From the k8s API conventions:
> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options.

Would it be better to use `restartTimeSeconds: 0` to indicate that graceful restart is disabled? This is the approach of multihop:
> // When empty or zero, eBGP multi-hop feature is disabled. The value is ignored for iBGP peers.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -65,19 +83,17 @@ type CiliumBGPPeeringPolicySpec struct {
...
 	//
-	// Default is 120s if empty or zero.
-	// Rounded internally to the nearest whole second.
-	//
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:Format=duration
-	RestartTime metav1.Duration `json:"restartTime"`
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:default=120
+	RestartTimeSeconds *int32 `json:"restartTimeSeconds"`
```
danehans: Event hough this field is being defaulted, I believe the `omitempty` tag should be used. From the docs:
>In most cases, optional fields should also have the omitempty struct tag (the omitempty option specifies that the field should be omitted from the json encoding if the field has an empty value). However, If you want to have different logic for an optional field which is not provided vs. provided with empty values, do not use omitempty 

Thoughts?

danehans: https://book.kubebuilder.io/reference/markers.html is also a good read about `+optional` vs `+kubebuilder:validation:Optional` and states:
> The most reliable way in 1.x to get +optional is omitempty.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
danehans: Have you tested whether creating a `CiliumBGPPeeringPolicy` without this field set will cause validation to fail? If a field is optional but specifies min and max values for validation, then I believe it should be defaulted (`+kubebuilder:default=0`). Otherwise, you can drop the min validation and state "When unspecified, the default, ... is disabled".

__Note:__ This comment applies to other fields with similar properties.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -97,48 +113,47 @@ type CiliumBGPNeighbor struct {
...
+	EBGPMultihopTTL *int32 `json:"eBGPMultihopTTL,omitempty"`
+	// ConnectRetryTimeSeconds defines the initial value for the BGP ConnectRetryTimer (RFC 4271, Section 8).
 	//
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:Format=duration
-	ConnectRetryTime metav1.Duration `json:"connectRetryTime,omitempty"`
-	// HoldTime defines the initial value for the BGP HoldTimer (RFC 4271, Section 4.2).
-	// The default value for the HoldTime (if empty or zero) is 90 seconds.
-	// Rounded internally to the nearest whole second. Updating this value will cause a session reset.
+	// +kubebuilder:validation:Minimum=1
```
danehans, line 140: A maximum value should be considered.

danehans: Does updating this value cause a session reset?

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -97,48 +113,47 @@ type CiliumBGPNeighbor struct {
...
+	// HoldTimeSeconds defines the initial value for the BGP HoldTimer (RFC 4271, Section 4.2).
+	// Updating this value will cause a session reset.
 	//
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:Format=duration
-	HoldTime metav1.Duration `json:"holdTime,omitempty"`
-	// KeepaliveTime defines the initial value for the BGP KeepaliveTimer (RFC 4271, Section 8).
-	// The default value for the KeepaliveTime (if empty or zero) is 1/3 of the HoldTime.
-	// Rounded internally to the nearest whole second. Updating this value will cause a session reset.
+	// +kubebuilder:validation:Minimum=3
```
danehans, line 148: A maximum value should be considered.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -97,48 +113,47 @@ type CiliumBGPNeighbor struct {
...
+	// +kubebuilder:default=90
+	HoldTimeSeconds *int32 `json:"holdTimeSeconds,omitempty"`
+	// KeepaliveTimeSeconds defines the initial value for the BGP KeepaliveTimer (RFC 4271, Section 8).
+	// It can not be larger than HoldTimeSeconds. Updating this value will cause a session reset.
 	//
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:Format=duration
-	KeepAliveTime metav1.Duration `json:"keepAliveTime,omitempty"`
+	// +kubebuilder:validation:Minimum=1
```
danehans, line 156: A maximum value should be considered.

pkg/bgpv1/agent/annotations.go
```diff
@@ -79,7 +79,7 @@ func (e ErrAttrib) Error() string {
 //	local-port=port (int):  the local port to listen on for incoming BGP connections
 type Attributes struct {
 	// The local ASN of the virtual router these Attributes targets.
-	ASN int
+	ASN int64
```
danehans, line 82: Should `LocalPort` be an `int32` like `PeerPort`?

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -168,3 +184,60 @@ type CiliumBGPVirtualRouter struct {
 	// +kubebuilder:validation:MinItems=1
 	Neighbors []CiliumBGPNeighbor `json:"neighbors"`
 }
+
+// Default applies default values on the CiliumBGPPeeringPolicy.
+// This is normally done by kube-apiserver for fields with explicit static defaults,
+// the main use of this method is to avoid the need for nil-checks in the controller code.
+func (p *CiliumBGPPeeringPolicy) Default() {
```
danehans: Thoughts on naming `SetDefaults()` to make it explicit for the caller?

pkg/bgpv1/agent/controller_test.go
```diff
@@ -139,23 +137,24 @@ func TestControllerSanity(t *testing.T) {
...
+							n.KeepAliveTimeSeconds == nil ||
+							n.GracefulRestart.RestartTimeSeconds == nil {
+							t.Fatalf("policy: %v not defaulted properly (nil)", p)
+						}
+						if *n.PeerPort != v2alpha1api.DefaultBGPPeerPort ||
+							*n.ConnectRetryTimeSeconds != v2alpha1api.DefaultBGPConnectRetryTimeSeconds ||
+							*n.HoldTimeSeconds != v2alpha1api.DefaultBGPHoldTimeSeconds ||
+							*n.KeepAliveTimeSeconds != v2alpha1api.DefaultBGPKeepAliveTimeSeconds ||
+							*n.GracefulRestart.RestartTimeSeconds != v2alpha1api.DefaultBGPGRRestartTimeSeconds {
+							t.Fatalf("policy: %v not defaulted properly (invalid value)", p)
```
danehans: Thoughts on having a func that checks the defaults of a neighbor and lives by `Defaults()` to make it easier to maintain the list of defaults over time?

pkg/bgpv1/gobgp/server.go
```diff
@@ -22,8 +21,8 @@ const (
 	wildcardIPv4Addr = "0.0.0.0"
 	wildcardIPv6Addr = "::"
 
-	// defaultIdleHoldTimeAfterReset defines time BGP session will stay idle after neighbor reset.
-	defaultIdleHoldTimeAfterReset = 5 * time.Second
+	// defaultIdleHoldTimeAfterResetSeconds defines time BGP session will stay idle after neighbor reset.
```
danehans: nit: Not sure if this level of verbosity is needed for a local var.

pkg/bgpv1/manager/workdiff.go
```diff
@@ -16,31 +16,31 @@ import (
...
 	// local ASN.
-	seen map[int]*v2alpha1api.CiliumBGPVirtualRouter
+	seen map[int64]*v2alpha1api.CiliumBGPVirtualRouter
 	// the state of the bgp control plane at the time of this reconcileDiff's
 	// creation.
 	state *agent.ControlPlaneState
 	// Local ASNs which BgpServers must be instantiated, configured,
 	// and added to the manager. Intended key for `seen` map.
-	register []int
+	register []int64
```
danehans, line 25: From the k8s api conventions:
> All public integer fields MUST use the Go (u)int32 or Go (u)int64 types, not (u)int (which is ambiguous depending on target platform). Internal types may use (u)int.

Since this is an internal type, should we continue using `int`?

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: > Have you tested whether creating a CiliumBGPPeeringPolicy without this field set will cause validation to fail?

Sure, it will not cause a validation error. Validation only applies if the value is set.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: good catch, this should indeed have the `omitempty` tag (as well as some other optional fields that are missing it)

pkg/bgpv1/manager/workdiff.go
rastislavs: I don't think so, it could cause issues on 32-bit systems - think of storing max 32-bit ASN (`4294967295`) in a 32-bit signed integer.

pkg/bgpv1/agent/controller_test.go
rastislavs: Hmm, I'm not sure about that - this is a unit test, something very internal. The API package is on the other hand something very external, and such a function would not have any practical use other than this test. But at the same time I understand the inconvenience.. Would like to know what the others think about it :)

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: Personally I do not have any preference between the two - what others think about it?

pkg/bgpv1/agent/annotations.go
rastislavs: Not sure, it is not part of the CRD API  - it is just our internal type - so it does not have to. Only if that would be our preference.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: Well, yeah I agree that we are now a bit inconsistent with regards to the `eBGPMultihopTTL`.
I remember we went this way (putting GR options into a struct) because `gracefulRestartRestartTimeSeconds` field looked weird to us :)

Thoughts @YutaroHayakawa / @harsimran-pabla ?

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: > Does updating this value cause a session reset?

No, this one is applied immediately without the need for session reset.

pkg/bgpv1/gobgp/server.go
rastislavs: any preference for a better name?

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
harsimran-pabla: I think bool field for enabling/disabling graceful restart cap makes sense.

Right now, we are only exposing Restart Time, but this may change in future as we expose more configuration options described [here](https://github.com/osrg/gobgp/blob/master/api/gobgp.pb.go#L7553-L7563)

We should not tie in enabling/disabling to restart time - going back from it will be a breaking change in API.

pkg/bgpv1/gobgp/server.go
harsimran-pabla: I think we can drop `default` from the prefix since it is not really changeable. I don't have any strong preference for this.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
harsimran-pabla: +1 on SetDefaults() makes it clear that receivers' values may change.

rastislavs: Actually, you changed my mind :) `SetDefaults()` will be better, as it is in fact even mutable.

pkg/bgpv1/agent/controller_test.go
harsimran-pabla: Two comments regarding this 
1. Do you think this test should move into bgpp_types_test.go? Since, we moved SetDefaults() method there, shouldn't the test to validate that Defaults are set correctly be moved within that pkg? This should make it easier to maintain, as both code and tests live in the same pkg ( harder to miss ).
2. I tend to agree with @rastislavs  on not adding a new method externally for something which is used solely in the test.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: Good points @harsimran-pabla . At the same time I don't think that `eBGPMultihopTTL` would need its own struct and `enabled` flag similar to GR - as there are no other possible options.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: Fixed in https://github.com/cilium/cilium/pull/26040/commits/df19bf07f4fb39b339257b477e17f4282c28976b

pkg/bgpv1/agent/controller_test.go
rastislavs: > Do you think this test should move into bgpp_types_test.go?

Yeah, that would make sense - with one disadvantage: it would not test that `SetDefaults()` has been indeed called during the policy reconciliation in the agent.

harsimran-pabla: I see, why not split the test into two parts? 

- Test_SetDefaults (in bgpp_types_test.go): Validates if values are correctly set to expected defaults 
- Test_Policy (in controller_test.go): Validates there are no nil values. 

No strong opinion, just bouncing ideas. :)

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
YutaroHayakawa: Agree to the conclusion of Pabla and Rastislav here 👍

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
YutaroHayakawa: Since we indicate this option is for eBGP only, maybe setting the default to 1 (and ignoring that value for iBGP internally) makes our life a bit easier and the API looks consistent?

rastislavs: > maybe setting the default to 1 (and ignoring that value for iBGP internally) makes our life a bit easier and the API looks consistent?

I like that idea, it will also address the inconsistency with the GR discussed in the other thread.

rastislavs: Done in https://github.com/cilium/cilium/pull/26040/commits/ed40377c9f8db1d866745805e1a87b02f0f4bf89

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: The inconsistency with `eBGPMultihopTTL` resolved in  https://github.com/cilium/cilium/pull/26040/commits/ed40377c9f8db1d866745805e1a87b02f0f4bf89

pkg/bgpv1/agent/controller_test.go
rastislavs: In https://github.com/cilium/cilium/pull/26040/commits/bc81751823d7d638426726f47f3ca861c786a081 I added unit tests to the API package and removed checks for default values in the controller unit test.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: `omitemty` fixed. As this PR aims to keep backward compatibility, renaming or changing the type other than pointer change is not possible on this field. If there will be a need for that we can do it later as a separate PR with some migration logic.

pkg/bgpv1/agent/annotations.go
rastislavs: Actually makes a lot of sense, as it is `int32` in the `BGPGlobal` API as well. Consolidated in https://github.com/cilium/cilium/pull/26040/commits/68b5271b792f9b9f93e2800d55d1c777cf400e18

rastislavs:
Thanks for the comments @danehans ! I have resolved all of them, ready for the final review.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -97,48 +116,51 @@ type CiliumBGPNeighbor struct {
...
 	//
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:Format=duration
-	KeepAliveTime metav1.Duration `json:"keepAliveTime,omitempty"`
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=2147483647
+	// +kubebuilder:default=30
+	KeepAliveTimeSeconds *int32 `json:"keepAliveTimeSeconds,omitempty"`
 	// GracefulRestart defines graceful restart parameters which are negotiated
 	// with this neighbor.
```
youngnick: I'd suggest adding in something here about how Graceful Restart is disabled if this is empty or nil (which seems to be the case AFAICT).

This does mean that:
```yaml
    gracefulRestart:
      enabled: true
```
is the simplest way to enable this on input, which is a little verbose, but given that it means the timeout values will then be included in the output, I think it's well worth it.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -97,48 +116,51 @@ type CiliumBGPNeighbor struct {
...
 	//
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:validation:Maximum=4294967295
-	PeerASN int `json:"peerASN"`
+	PeerASN int64 `json:"peerASN"`
 	// EBGPMultihopTTL controls the multi-hop feature for eBGP peers.
 	// Its value defines the Time To Live (TTL) value used in BGP packets sent to the neighbor.
-	// When empty or zero, eBGP multi-hop feature is disabled. The value is ignored for iBGP peers.
+	// The value 1 implies that eBGP multi-hop feature is disabled. This field is ignored for iBGP peers.
```
youngnick:
```suggestion
	// The value 1 implies that eBGP multi-hop feature is disabled (because only a single hop is allowed). This field is ignored for iBGP peers.
```

Small clarification request?

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -149,22 +171,80 @@ type CiliumBGPVirtualRouter struct {
...
+func (r *CiliumBGPVirtualRouter) SetDefaults() {
+	if r.ExportPodCIDR == nil {
+		r.ExportPodCIDR = pointer.Bool(DefaultBGPExportPodCIDR)
+	}
+	for i := range r.Neighbors {
+		r.Neighbors[i].SetDefaults()
+	}
+}
+
+// SetDefaults applies default values on the CiliumBGPNeighbor.
```
youngnick, line 217: This is great, because it is possible (using `kubectl apply --force` or similar) to bypass the validation checks. Nice work.

youngnick (approved):
LGTM with a few small comments, very nice work. Love seeing people run CRDs through the API conventions.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -97,48 +116,51 @@ type CiliumBGPNeighbor struct {
...
-	PeerASN int `json:"peerASN"`
+	PeerASN int64 `json:"peerASN"`
 	// EBGPMultihopTTL controls the multi-hop feature for eBGP peers.
 	// Its value defines the Time To Live (TTL) value used in BGP packets sent to the neighbor.
-	// When empty or zero, eBGP multi-hop feature is disabled. The value is ignored for iBGP peers.
+	// The value 1 implies that eBGP multi-hop feature is disabled. This field is ignored for iBGP peers.
 	//
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Minimum=1
```
harsimran-pabla, line 133: Do we need to update Documentation as well regarding this change?

harsimran-pabla (approved):

harsimran-pabla:
Thank you @rastislavs, this is a great change! Minor comment on documentation otherwise LGTM.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: Yes - not only that change, but I will do that in a separate PR to not postpone this PR any more: https://github.com/cilium/cilium/pull/26196

rastislavs:
/test

(rebased to fetch latest CI changes/fixes)

YutaroHayakawa:
API change is a bit of description changes. I think it's ready to merge.

danehans (approved):

rastislavs:
> API change is a bit of description changes. I think it's ready to merge.

Rebased due to a conflicting change in the main branch. All green after the rebase - marking as ready to merge again.

aanm (changes requested):
I did an overall review on the PR. It looks that pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go got a lot of modifications, did we consider on using `v2alpha2` with a new CRD? If not, is there any upgrade path from users using v2alpha1 bgp?

rastislavs:
@aanm 
> pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go got a lot of modifications, did we consider on using `v2alpha2` with a new CRD? If not, is there any upgrade path from users using v2alpha1 bgp?

I believe this PR does not introduce any breaking change to the types that were already released in Cilium 1.13. Most of the changes in this PR are on the fields that were added as recent "to be released" features in the past weeks. There are only 3 backward compatible changes on the previously released CRD fields - copying from this PR's description:

> Already released API fields are modified only in a backward compatible way (e.g. changed internal type to a pointer, but no type change in the generated `CustomResourceDefinition`).  The changes of already released API in the generated [ciliumbgppeeringpolicies.yaml](https://github.com/cilium/cilium/pull/26040/files#diff-dddec5a0dbbf46faca6e3ea84b23aefe87d43e6ced346582f5d86d362936f204) are:
> 
> - exportPodCIDR: added default value (false, so not changing the default behaviour)
> - localASN + peerASN: added format annotation, but without the change of the underlying type

aanm (approved):
Thanks for the clarification

PR #26045, opened 2023-06-08
daemon, maps/ipcache: Replace usage of `net.IP*` for ingress IPs

christarazi:
- Revert "ip: Add IPNetToPrefix() helper"
- daemon, maps/ipcache: Replace usage of `net.IP*` for ingress IPs

---

> daemon, maps/ipcache: Replace usage of `net.IP*` for ingress IPs
> 
> Following the previously reverted commit, replace the ingress IP
> restoration to use netip types.
> 
> Towards https://github.com/cilium/cilium/issues/24246.
> 
> Reported-by: Tobias Klauser <tobias@cilium.io>

### DIFF
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -53,7 +53,6 @@ import (
 	"github.com/cilium/cilium/pkg/hubble/observer"
 	"github.com/cilium/cilium/pkg/identity"
 	"github.com/cilium/cilium/pkg/identity/identitymanager"
-	"github.com/cilium/cilium/pkg/ip"
 	"github.com/cilium/cilium/pkg/ipam"
 	ipamMetadata "github.com/cilium/cilium/pkg/ipam/metadata"
 	ipamOption "github.com/cilium/cilium/pkg/ipam/option"
@@ -561,7 +560,7 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 
 	// Collect old CIDR identities
 	var oldNIDs []identity.NumericIdentity
-	var oldIngressIPs []*net.IPNet
+	var oldIngressIPs []netip.Prefix
 	if option.Config.RestoreState && !option.Config.DryMode {
 		if err := ipcachemap.IPCacheMap().DumpWithCallback(func(key bpf.MapKey, value bpf.MapValue) {
 			k := key.(*ipcachemap.Key)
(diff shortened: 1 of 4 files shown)


christarazi:
cc @tklauser

tklauser (commented):
Code changes LGTM, thanks for fixing this!

I'd suggest to reverse the commit order. Otherwise bisectability is broken on the first commit [`Revert "ip: Add IPNetToPrefix() helper"`](https://github.com/cilium/cilium/pull/26045/commits/6abc440422b223cc2efe1cadfdf1044bb7dd7aa3) because `ip.IPNetToPrefix` is still used and only replaced in the second commit.

daemon/cmd/daemon.go
```diff
@@ -570,12 +569,13 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 				d.restoredCIDRs = append(d.restoredCIDRs, k.Prefix())
 				oldNIDs = append(oldNIDs, nid)
 			} else if nid == identity.ReservedIdentityIngress && v.TunnelEndpoint.IsZero() {
-				oldIngressIPs = append(oldIngressIPs, k.IPNet())
-				ip := k.IPNet().IP
```
tklauser, line 575: One other thing that I just noticed: these are the only remaining call sites of `ipcachemap.Key.IPNet`, so it can probably be removed too as part of this PR or in a follow-up.

mhofstetter (approved):
LGTM - thanks!

aspsk (approved):

daemon/cmd/daemon.go
christarazi: `IPNet()` is already removed in this PR.

tklauser (approved):

christarazi:
/test-1.26-net-next

michi-covalent (approved):
reviewed tophat-owned files 🧹

joestringer:
```
$ git grep IPNet pkg/ | wc -l
890
```

We still have a *lot* of `*net.IPNet` usage in the tree, I think that 2db75775e66e69e105294059bc3d1c5771bea2b9 was premature to remove the helper.

tklauser:
> EDIT: Ah it looks like there's a `netipx.FromStdIPNet()` equivalent we can use instead from netipx package: https://pkg.go.dev/go4.org/netipx?utm_source=godoc#FromStdIPNet

Yep, see https://github.com/cilium/cilium/pull/25749#discussion_r1222873858 which I think triggered this PR.

PR #26047, opened 2023-06-08
bpf: nodeport: wire up trace struct for IPv6 RevDNAT

julianwiedmann:
When the NAT path in `to-overlay/to-netdev` is included as `tail_handle_nat_fwd_ipv6()`, skip the hop through `handle_nat_fwd_ipv6()` and call the actual RevDNAT implementation straight away.

This allows us to pass through a `trace_ctx` struct, so that the RevDNAT can return precise trace information from its CT lookup.

The IPv4 path already does this, so we must have missed the boat at some point.

### DIFF
--- a/bpf/lib/nodeport.h
+++ b/bpf/lib/nodeport.h
@@ -1400,7 +1400,7 @@ int tail_handle_nat_fwd_ipv6(struct __ctx_buff *ctx)
 	obs_point = TRACE_TO_NETWORK;
 #endif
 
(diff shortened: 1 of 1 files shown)


julianwiedmann:
Looks like this cleanly cherry-picks back to v1.13, so let's apply it there as well.

aspsk (approved):
Yes, this makes more sense

julianwiedmann:
(the one fail in the GH ginkgo tests is https://github.com/cilium/cilium/issues/25964)

PR #26048, opened 2023-06-08
Ldelossa/srv6 encap fib

ldelossa:
Original SRv6 encapsulation flow:

``` mermaid
 flowchart LR
p["pod"] 
v1["veth1"]
v2["veth2"]
hs["host stack"]
e["eth0"]
tx["tx"]

p-->|egress pkt|v1-->v2-->hs-->|fib lookup|e-->|SRv6 encap|tx
```
This results in the egress packet always choosing the output interface, source, and destination mac from the FIB lookup that occurred on the inner packet. 

SRv6 encapsulation flow corrected:

``` mermaid
 flowchart LR
p["pod"] 
v1["veth1"]
v2["veth2"]
hs["host stack"]
e["eth0"]
tx["tx"]

p-->|egress pkt|v1-->v2-->hs-->|fib lookup|e-->|SRv6 encap\n+\nfib lookup|tx
```

Now, we perform an additional fib lookup once the egress packet has been encapsulated, ensuring the encapsulated packet is forwarded correctly based on the host namespace's FIB.

To accomplish this a refactor of `lib/fib.h` was necessary. 
The fib library can now be used to perform a v4/v6 fib lookup and a redirect independently, where before the actions were coupled together. 

Performing these actions independently is currently necessary for this SRv6 datapath fix. 
This is because we perform the `fib_lookup` already at a egress interface and we **may or may not** need to redirect to a new interface for tx. 
This refactor does not change the semantics of the existing functions, they work exactly how they used to, sans removing the `iif` field from the `fib_redirect_v[4|6]` functions, as we can extract them from the passed `ctx` making the arguments redundant. 

It is obvious to see that two fib lookups are occurring here. 
It does seem to me that this can be reduced to one, however this involves changing where encapsulation occurs and will be a larger refactor, however I will also look into this. 

```release-note
Fixes an issue where SRv6 encapsulated packets are forwarded to the wrong layer 2 next hop.
```

### DIFF
--- a/bpf/bpf_lxc.c
+++ b/bpf/bpf_lxc.c
@@ -643,8 +643,7 @@ ct_recreate6:
 	if (is_defined(ENABLE_HOST_ROUTING)) {
 		int oif = 0;
 
-		ret = fib_redirect_v6(ctx, ETH_HLEN, ip6, false, ext_err,
-				      ctx_get_ifindex(ctx), &oif);
+		ret = fib_redirect_v6(ctx, ETH_HLEN, ip6, false, ext_err, &oif);
 		if (fib_ok(ret))
 			send_trace_notify(ctx, TRACE_TO_NETWORK, SECLABEL,
 					  *dst_sec_identity, 0, oif,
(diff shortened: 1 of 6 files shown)


julianwiedmann:
Ah, I hadn't realized this also needs to go into `v1.13`. That FIB code looks very different :grimacing:.

Could we then just do what's needed for the fix (to have a backportable PR), and leave the refactor for a follow-on PR?

I peeked at just the following applied, and that lgtm so far :+1:.
```
7007d6761a bpf,srv6: perform fib lookup after encap
75e06d8be4 bpf,fib: introduce fib_lookup_v4/6 functions
92a5475f63 bpf,fib: introduce fib_do_redirect function
```

ldelossa:
I think I see what you're saying. 

The SRv6 code only needs the fib_lookup* and the fib_do_redirect function to exist. 

I'll split this out like you're saying, just introducing the the pieces necessary for SRv6 fix, and backport this. 

Then open another PR with the refactor, which *wont* need to be back ported. 

Sound aitttte?

ldelossa:
Closing in favor of opening a PR which splits the SRv6 fixes and the refactor.

Issue #24442, opened 2023-03-18
Found two node IDs for the same node errors


dctrwatson:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

1. restarted cilium-agent daemonsets in multiple clusters
2. agents in multiple clusters all complain about `Found two node IDs` for various `nodeName` but different node IDs

Seems it may just be nodes being deleted from clusters, not just agents restarting.

### Cilium Version

`Daemon: 1.12.8 87d8561 2023-03-15T19:28:16+01:00 go version go1.18.10 linux/amd64`

### Kernel Version

`Linux ip-10-206-123-81 5.15.0-1030-aws #34~20.04.1-Ubuntu SMP Tue Jan 24 15:16:46 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`

### Kubernetes Version

`Server Version: version.Info{Major:"1", Minor:"22+", GitVersion:"v1.22.16-eks-ffeb93d", GitCommit:"52e500d139bdef42fbc4540c357f0565c7867a81", GitTreeState:"clean", BuildDate:"2022-11-29T18:41:42Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}`

### Relevant log output

```shell
`level=error msg="Found two node IDs (0 and 4176) for the same node" ipAddr=10.199.160.233 nodeName=ip-10-199-150-180.ec2.internal subsys=linux-datapath`
```

### Anything else?

Spot checking, it seems the nodeName being logged are for non-existent nodes so far

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

pchaigno:
> Spot checking, it seems the nodeName being logged are for non-existent nodes so far

That makes sense because this error is printed while trying to deallocate a node ID, typically when a remote Kubernetes node was deleted.

A sysdump would help to figure this out. It would also be great if you had the Kubernetes Node object for a node for which this happened once deleted. So maybe taking two sysdumps: one before and one after a node deletion?

pchaigno:
This may have been fixed by https://github.com/cilium/cilium/pull/25222, which will be released in the upcoming v1.12.10.

pchaigno:
@dctrwatson Would you be able to share a sysdump?

dctrwatson:
> @dctrwatson Would you be able to share a sysdump?

Unfortunately, I cannot share a full sysdump easily. If you tell me what parts you would like to see, I can share those though.

[...]

PR #26057, opened 2023-06-09
Only delete ipsec endpoint when node ID is not 0

sjdot:
After applying a backport of https://github.com/cilium/cilium/pull/25784 to 1.11.16 I noticed that we were getting occasional spikes of "no inbound state" xfrm errors (XfrmInNoStates). These lead to packet loss and brief outages for applications sending traffic to the node on which the spikes occur.

I noticed that the "No node ID found for node." logline would appear at the time of these spikes and from the code this is logged when the node ID cannot be resolved. Looking a bit further the call to `DeleteIPsecEndpoint` will end up deleting the xfrm state for any state that matches the node id as derived from the mark in the state. 

The problem seems to be that the inbound state for 0.0.0.0/0 -> node IP has a mark of `0xd00` which when shifted >> 16 in `getNodeIDFromXfrmMark` matches nodeID 0 and so the inbound state gets deleted and the kernel drops all the inbound traffic as it no longer matches a state.

I've updated the code to only call `DeleteIPsecEndpoint` when the nodeID != 0 as a first pass here. Initially wanted to return after the logline, but am not if we can skip the rest of the code below this call so raising this PR and looking for input from @pchaigno to see if any other changes are preferred/required.

### DIFF
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -1575,8 +1575,9 @@ func (n *linuxNodeHandler) deleteIPsec(oldNode *nodeTypes.Node) {
 	nodeID := n.getNodeIDForNode(oldNode)
 	if nodeID == 0 {
 		scopedLog.Warning("No node ID found for node.")
+	} else {
+		ipsec.DeleteIPsecEndpoint(nodeID)
 	}
-	ipsec.DeleteIPsecEndpoint(nodeID)
 
 	if n.nodeConfig.EnableIPv4 && oldNode.IPv4AllocCIDR != nil {
 		old4RouteNet := &net.IPNet{IP: oldNode.IPv4AllocCIDR.IP, Mask: oldNode.IPv4AllocCIDR.Mask}


pchaigno:
Thanks for the report and PR!

> I noticed that the "No node ID found for node." logline would appear at the time of these spikes and from the code this is logged when the node ID cannot be resolved. Looking a bit further the call to DeleteIPsecEndpoint will end up deleting the xfrm state for any state that matches the node id as derived from the mark in the state.

I arrived at the same conclusion from a different point:
- We know the errors happen since patch https://github.com/cilium/cilium/pull/25784 is applied.
- This patch removes XFRM states and policies belonging to deleted remote nodes.
- It identifies the states and policies based on the node ID in the XFRM mark.
- The XfrmInNoStates errors indicates a missing XFRM IN state.
- Our XFRM IN states don't have a node ID in the mark.
- Thus, the only way the above patch would result in those errors is if the node ID of the remote node is zero.

That nodeID=0 for a remote node matches another bug report we've had (cf. https://github.com/cilium/cilium/issues/24442). We haven't been able to figure it out yet due to missing sysdumps. That bug didn't seem to have any side effect besides the error message because it happened on node deletion. But with my patch now using the node IDs for cleanup, I believe I cause it to have the impact you noticed.

> I've updated the code to only call `DeleteIPsecEndpoint` when the nodeID != 0 as a first pass here. Initially wanted to return after the logline, but am not if we can skip the rest of the code below this call so raising this PR and looking for input from @pchaigno to see if any other changes are preferred/required.

I think that's the proper fix. We indeed shouldn't return after the logline. I will add some additional "hardening" to skip policies and states that don't match on the node ID, but your patch should be enough to fix the bug.

Do you want me to carry this through? I think we'll just need to add the context to the commit description and triage CI once it's finished. I'm happy to do that for you.

pchaigno (approved):

sjdot:
Thanks for the review @pchaigno. I'm happy to update the commit with the context here, would the PR comment suffice? Would you like it prefaced with `ipsec:` also?

pchaigno:
I actually sent a PR with your commit at https://github.com/cilium/cilium/pull/26072 after all. I added another commit on top with the "hardening" I mentioned and added the `ipsec` prefix and the PR description. CI is almost ready there and we should be able to include it in the next set of patch releases (which are due in a few days).

sjdot:
Ah great, thanks @pchaigno

pchaigno:
Closing since this commit was merged as part of https://github.com/cilium/cilium/pull/26072 and backported to all stable branches. Thanks Steve!

PR #26059, opened 2023-06-09
auth: fix initial k8s events sync in auth map gc

mhofstetter:
This PR fixes the issue where k8s resource upsert events of cilium identities are still triggering a cleanup of missing identities even though the sync event has already been received and processed.

### DIFF
--- a/pkg/auth/authmap_gc.go
+++ b/pkg/auth/authmap_gc.go
@@ -90,6 +90,7 @@ func (r *authMapGarbageCollector) handleCiliumIdentityEvent(_ context.Context, e
 		if err = r.cleanupMissingIdentities(); err != nil {
 			return fmt.Errorf("failed to cleanup missing identities: %w", err)
 		}
+		r.discoveredCiliumIdentities = nil
 	case resource.Delete:
 		r.logger.
 			WithField("key", e.Key).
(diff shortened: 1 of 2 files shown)


pkg/auth/authmap_gc.go
```diff
@@ -90,14 +90,14 @@ func (r *authMapGarbageCollector) handleCiliumIdentityEvent(_ context.Context, e
 		if err = r.cleanupMissingIdentities(); err != nil {
 			return fmt.Errorf("failed to cleanup missing identities: %w", err)
 		}
+		r.discoveredCiliumIdentities = nil
```
sayboras, line 93: should we remove the key one by one in cleanupMissingIdentities() function instead ?

mhofstetter: @sayboras i can change this if wanted. and then change the check in the "upsert" whether the map has entries intead of whether its nil?

mhofstetter: @sayboras would it be ok to bring add this "optimization" in an upcoming PR. so the actual fix can get merged because this one might help in tackling https://github.com/cilium/cilium/issues/25964

sayboras (approved):

PR #26068, opened 2023-06-09
auth: policy based auth map GC

### LINKED ISSUE #25925: Auth: Support auth map garbage collection based on deleted/updated Policies
Currenty, Auth Map garbage collection deletes entries based on the following criterias (implemented by https://github.com/cilium/cilium/issues/25213)

* identity gets deleted -> all entries referencing the deleted identity as local or remote identity should be deleted
* node gets deleted -> all entries referencing the deleted node as remote node should be deleted
* entry expires -> all expired entries should be deleted

By introducing re-authentication based on certificate rotation events (https://github.com/cilium/cilium/issues/25475) - garbage collection doesn't cover the case where identities & nodes still exist, but there's no longer a policy enforcing auth between two identities. Re-authentication will renew them forever - which results in keeping the entry in the auth map "forever".

With this issue, garbage collection based on mutated/deleted policies should be implemented.

mhofstetter:
Until now, auth map entries were garbage collected based on the following criterias:

* related identity has been deleted
* related node has been deleted
* entry has been expired

The initial goal was that expiration will cover the case where no longer a policy is enforcing authentication.

But the introduction of re-authentication (#25927) changed this, because the entries would have re-authenticated "forever" (until identity or node would have been deleted).

Therefore, this commit introduces some rudimentary garbage collection based on policies by periodically checking whether a policy is still enforcing authentication between two identities with the given auth type. If not, the auth map entry gets deleted.

Fixes: #25925

### DIFF
--- a/pkg/auth/authmap_gc.go
+++ b/pkg/auth/authmap_gc.go
@@ -16,22 +16,30 @@ import (
 	"github.com/cilium/cilium/pkg/k8s/resource"
 	"github.com/cilium/cilium/pkg/logging/logfields"
 	"github.com/cilium/cilium/pkg/node/addressing"
+	"github.com/cilium/cilium/pkg/policy"
 )
 
 type authMapGarbageCollector struct {
-	logger  logrus.FieldLogger
-	authmap authMap
-	ipCache ipCache
+	logger     logrus.FieldLogger
+	authmap    authMap
+	ipCache    ipCache
+	policyRepo policyRepository
 
 	discoveredCiliumNodeIDs    map[uint16]struct{}
 	discoveredCiliumIdentities map[identity.NumericIdentity]struct{}
 }
 
-func newAuthMapGC(logger logrus.FieldLogger, authmap authMap, ipCache ipCache) *authMapGarbageCollector {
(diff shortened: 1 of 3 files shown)


jrajahalme (commented):
Looks right, just a minor difference in the new policy repo function prototype in my PR (returning `policy.AuthType` instead of a bool).

joestringer:
Just adding the blocker label in context of https://github.com/cilium/cilium/pull/26116. I see that there's mocked out values in this PR which help it to compile but make it functionally incorrect until rebased :+1:

joestringer:
I agree with Jarno, logic looks good. Seems like a bugfix for the reauth PR. I consider this exempt from the feature freeze window on those grounds. (Still nice to get in sooner than later)

mhofstetter:
rebased to `main` to fetch changes from https://github.com/cilium/cilium/pull/26116

mhofstetter:
adapted changes that `GetAuthTypes` returns all applied auth types between two security identities and removed the mock with the actual implementation.

jrajahalme (approved):
LGTM

PR #26072, opened 2023-06-09
ipsec: Fix cleanup of XFRM states and policies

pchaigno:
See commits for details.

Fixes: https://github.com/cilium/cilium/pull/25784.


qmonnet (approved):

jschwinger233 (approved):

qmonnet:
Conformance Ginkgo hit https://github.com/cilium/cilium/issues/25964

pchaigno:
I performed an IPsec upgrade on EKS from v1.13.0 to this patch, just in case. I also checked for XFRM leaks by doing a scale up and down from 3 to 9 nodes and back. Connectivity tests also passed (although the CI should already cover that anyway).

PR #26077, opened 2023-06-09
hive,metrics: Add metrics for pkg/hive/job

dylandreimerink:
This PR adds a metric to that tracks error counts for job runs, a metric to track timer run performance and observer run performance.

For one shot jobs the metric collection is disabled by default, only jobs with the `WithMetrics` option will be measured. This way we can collect metrics for short running jobs and skip collection for high cardinality jobs or jobs that run forever.

```release-note
Added metrics for jobs
```

### DIFF
--- a/Documentation/observability/metrics.rst
+++ b/Documentation/observability/metrics.rst
@@ -506,6 +506,18 @@ Name                               Labels                           Default     
 ``fqdn_alive_zombie_connections``  ``endpoint``                     Disabled     Number of IPs associated with domains that have expired (by TTL) yet still associated with an active connection (aka zombie), per endpoint
 ================================== ================================ ============ ========================================================
 
+Jobs
+~~~~
+
+================================== ================================ ============ ========================================================
+Name                               Labels                           Default      Description
+================================== ================================ ============ ========================================================
+``jobs_errors_total``              ``job``                          Enabled      Number of jobs runs that returned an error
+``jobs_one_shot_run_seconds``      ``job``                          Enabled      Histogram of one shot job run duration
+``jobs_timer_run_seconds``         ``job``                          Enabled      Histogram of timer job run duration
+``jobs_observer_run_seconds``      ``job``                          Enabled      Histogram of observer job run duration
+================================== ================================ ============ ========================================================
+
 .. _metrics_api_rate_limiting:
 
 API Rate Limiting
--- a/pkg/hive/job/job.go
+++ b/pkg/hive/job/job.go
@@ -5,7 +5,7 @@ package job
 
 import (
 	"context"
-	"fmt"
+	"errors"
 	"runtime/pprof"
 	"sync"
 	"time"
@@ -17,6 +17,7 @@ import (
 	"github.com/cilium/cilium/pkg/hive/cell"
 	"github.com/cilium/cilium/pkg/hive/internal"
 	"github.com/cilium/cilium/pkg/lock"
+	"github.com/cilium/cilium/pkg/spanstat"
 	"github.com/cilium/cilium/pkg/stream"
 )
 
@@ -25,7 +26,12 @@ import (
 // sure multiple goroutines properly shutdown takes a lot of boilerplate. Job groups make it easy to queue, spawn, and
 // collect jobs with minimal boilerplate. The registry maintains references to all groups which will allow us to add
 // automatic metrics collection and/or status reporting in the future.
-var Cell = cell.Provide(newRegistry)
+var Cell = cell.Module(
+	"jobs",
+	"Jobs",
+	cell.Provide(newRegistry),
+	cell.Metric(newJobMetrics),
+)
 
 // A Registry creates Groups, it maintains references to these groups for the purposes of collecting information
 // centralized like metrics.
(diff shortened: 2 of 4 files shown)


pkg/hive/job/job.go
```diff
@@ -248,6 +263,7 @@ func (jos *jobOneShot) start(ctx context.Context, wg *sync.WaitGroup, options op
 
 	if jos.fn == nil {
```
joamaki: side-note: can we eliminate this case in the OneShot constructor by making it panic? (likewise for other `fn == nil` checks like timer)

pkg/hive/job/job.go
```diff
@@ -390,12 +410,19 @@ func (jt *jobTimer) start(ctx context.Context, wg *sync.WaitGroup, options optio
 
 		l.Debug("Timer job triggered")
 
+		stat.Start()
```
joamaki, line 431: We should do spans also for one-shot jobs. While they might be infinite loops they might also do some interesting one-shot initialization.

pkg/hive/job/job.go
```diff
@@ -454,18 +481,30 @@ func (jo *jobObserver[T]) start(ctx context.Context, wg *sync.WaitGroup, options
...
+
 		err := jo.fn(ctx, t)
+
+		sec := stat.End(true).Seconds()
+		options.metrics.ObserverRunDuration.WithLabelValues(jo.name).Observe(sec)
+		stat.Reset()
+
 		if err != nil {
 			l.WithError(err).Error("Observer job errored")
+			options.metrics.JobErrorsTotal.WithLabelValues(jo.name).Inc()
```
joamaki, line 523: btw, we're logging "context cancelled" errors when stopping here. I think it'd make sense to check `ctx.Err()` for cancellation and then not log the error (since it's expected behaviour when stopping).

Documentation/observability/metrics.rst
```diff
@@ -487,6 +487,17 @@ Name                               Labels                           Default
 ``fqdn_alive_zombie_connections``  ``endpoint``                     Disabled     Number of IPs associated with domains that have expired (by TTL) yet still associated with an active connection (aka zombie), per endpoint
 ================================== ================================ ============ ========================================================
 
+Jobs
+~~~~
+
+================================== ================================ ============ ========================================================
+Name                               Labels                           Default      Description
+================================== ================================ ============ ========================================================
+``jobs_errors_total``                                               Enabled      Number of jobs runs that returned an error
```
joamaki: Doesn't this also have the `job` label?

pkg/hive/job/job.go
dylandreimerink: I am worried about the carnality of one-shot jobs since they are also used to dynamically spawn goroutines like in https://github.com/cilium/cilium/blob/main/pkg/l2announcer/l2announcer.go#L702

joamaki: Wouldn't metrics be especially useful for those? Perhaps the solution is to not generate job names like that and use just "leader-election" as the name?

pkg/hive/job/job.go
dylandreimerink: Agreed, I moved the checks to the constructors, which makes more sense anyway.

pkg/hive/job/job.go
dylandreimerink: So far, we only have jobs that live until shutdown, and these which live a very long time. In both cases, knowing how long they were running is not meaningful. It would only be meaningful to know for short lived jobs. Perhaps for one shots it makes sense to define the metric but only add jobs which have been marked with a `WithMetrics` option? That way we don't have unnecessary carnality

joamaki: Sounds good to me!

pkg/hive/job/job.go
```diff
@@ -264,13 +290,24 @@ func (jos *jobOneShot) start(ctx context.Context, wg *sync.WaitGroup, options op
...
+			options.metrics.OneShotRunDuration.WithLabelValues(jos.name).Observe(sec)
+			stat.Reset()
+		}
+
 		if err == nil {
 			l.Debug("one-shot job finished")
 			return
 		} else {
 			l.WithError(err).Error("one-shot job errored")
+			options.metrics.JobErrorsTotal.WithLabelValues(jos.name).Inc()
```
pippolo84, line 310: Shouldn't we check for `context.Canceled` here too?

pkg/hive/job/job.go
```diff
@@ -390,12 +428,19 @@ func (jt *jobTimer) start(ctx context.Context, wg *sync.WaitGroup, options optio
...
 
+		sec := stat.End(true).Seconds()
+		options.metrics.TimerRunDuration.WithLabelValues(jt.name).Observe(sec)
+		stat.Reset()
+
 		if err == nil {
 			l.Debug("Timer job finished")
 		} else {
 			l.WithError(err).Error("Timer job errored")
+			options.metrics.JobErrorsTotal.WithLabelValues(jt.name).Inc()
```
pippolo84, line 443: Ditto

zacharysarah (approved):
LGTM

pippolo84 (approved):
Awesome! :rocket:

aditighag (approved):

PR #26078, opened 2023-06-09
metrics,endpointmanager: Register endpointmanager metrics via dependency

dylandreimerink:
The endpointmanager was the last package to use the MustRegister package scope function in metrics. This PR replaces the access via global function by providing the registry via dependency injection. This allows us to get rid of the metrics.MustRegister function.

```release-note
Register endpointmanager metrics via dependency injected registry
```

### DIFF
--- a/pkg/endpointmanager/cell.go
+++ b/pkg/endpointmanager/cell.go
@@ -18,6 +18,7 @@ import (
 	"github.com/cilium/cilium/pkg/k8s/client"
 	"github.com/cilium/cilium/pkg/k8s/watchers"
 	"github.com/cilium/cilium/pkg/k8s/watchers/subscriber"
+	"github.com/cilium/cilium/pkg/metrics"
 	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/policy"
 )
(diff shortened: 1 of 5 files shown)


dylandreimerink:
/test-1.25-4.19

chancez (approved):

pkg/endpointmanager/cell.go
```diff
@@ -106,7 +107,7 @@ type EndpointManager interface {
 
 	// InitMetrics hooks the EndpointManager into the metrics subsystem. This can
 	// only be done once, globally, otherwise the metrics library will panic.
-	InitMetrics()
+	InitMetrics(regitry *metrics.Registry)
```
pippolo84: Why we need to add `InitMetrics` to the `EndpointManager` interface?

Nit: typo on the param name `regitry` -> `registry`

pippolo84 (commented):
Thanks, looks good overall, I've just a question about the `EndpointManager` interface.

aditighag (approved):

pkg/endpointmanager/cell.go
dylandreimerink: That is a very good point. I did not think of that. It was likely added to the signature so it matches exactly, but it is never actually called on the interface type as far as I can see. So removed it :rocket:

pippolo84 (approved):
:100:

PR #26080, opened 2023-06-09
datapath: Convert IPSec xfer collector into modular metric

dylandreimerink:
The existing logic created a collector and used the metric.Register function to conditionally register it. This PR makes the same collector available as modular metric and makes the collector implement metric.WithMetadata.

The metric is now always registered, but conditionally enabled. Since the metric must now be provided as a dependency, we update the datapath creation and all tests that use the linux node handler.

This is one step towards getting rid of the metric.Register global func

```release-note
Convert IPSec xfer collector into modular metric
```

### DIFF
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -12,10 +12,8 @@ import (
 	"net"
 	"os"
 	"path/filepath"
-	"sync"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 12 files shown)


jibi:
(removing myself from reviews as I'm not in the IPSec team anymore)

pkg/datapath/linux/node.go
```diff
@@ -1213,7 +1212,7 @@ func (n *linuxNodeHandler) nodeUpdate(oldNode, newNode *nodeTypes.Node, firstAdd
 			n.enableSubnetIPsec(n.nodeConfig.IPv4PodSubnets, n.nodeConfig.IPv6PodSubnets)
 		}
 		if firstAddition && n.nodeConfig.EnableIPSec {
-			n.registerIpsecMetricOnce()
+			n.enableIpsecMetric()
```
tommyp1ckles, line 1215: why do we enable this all the way down here 😄

tommyp1ckles (approved):

pkg/datapath/linux/node.go
tommyp1ckles, line 1215: I wonder if we would want to start emitting this metric right away, so that we know what the xfrm counts are prior to any ipsec reconciliation (i.e. for potentially leftover state between restarts?).

tommyp1ckles (approved):
metrics changes make sense to me, nice cleanup!

aditighag (approved):
Approving for my codeowner for `pkg/endpoint`.

PR #26083, opened 2023-06-09
Introduce kvstoremesh, a clustermesh-apiserver companion component

giorio94:
Please review commit by commit. I'm reporting the description of the main commit here for your convenience:

```
In the current clustermesh implementation, each agent connects to the
kvstore in all remote clusters, pulling information about nodes,
services, identities and IPs. The number of etcd watchers can grow
significantly large (~ 5 * total nodes in the clustermesh), as well
as the number of events to be relayed by etcd to the agents (especially
if the churn rate is high).

kvstoremesh is a new component responsible for synchronizing the
information from the remote kvstores to the local one. In other words,
the local kvstore caches the state of the entire clustermesh, so that
all agents only need to connect to the local kvstore, and not the remote
ones. Overall, this provides better isolation (because the only component
connecting to remote kvstores is now the kvstoremesh operator) and
reduces the number of agents connecting to a single kvstore.

While the number of events to be processed by a single kvstore is
similar to the vanilla clustermesh case, with kvstoremesh the load
is less spiky (because a specific event is propagated to a much lower
number of agents) and it is also easier to rate limit (configuring the
maximum number of keys that can be written by kvstoremesh per second).
This allows to trade latency for reliability in case of high churn in
the clustermesh. Finally, given that the number of watchers is
proportional to the number of nodes in the local cluster only (rather
than the nodes in the entire clustermesh), the dimensioning of the
kvstore is also simplified.

The kvstoremesh concept has been initially proposed by the Cloud-Network
team @Trip.com, led by Arthur Chiao:
http://arthurchiao.art/blog/trip-first-step-towards-cloud-native-security/
```

The Dockerfile and CI modifications required to build the new image, as well as the appropriate extensions to the helm chart, will be introduced in a subsequent PR to limit the size of the current one.

```release-note
Introduce kvstoremesh, a clustermesh-apiserver companion component allowing to cache remote cluster information in the local kvstore for increased scalability and separation.
```

### DIFF
--- a/Documentation/observability/metrics.rst
+++ b/Documentation/observability/metrics.rst
@@ -179,9 +179,11 @@ metrics conform to the OpenMetrics requirements.
 Cluster Mesh API Server Metrics
 ===============================
 
-Cluster Mesh API Server metrics provide insights into both the state of the
-``clustermesh-apiserver`` process and the sidecar etcd instance.
+Cluster Mesh API Server metrics provide insights into the state of the
+``clustermesh-apiserver`` process, the ``kvstoremesh`` process (if enabled),
+and the sidecar etcd instance.
 Cluster Mesh API Server metrics are exported under the ``cilium_clustermesh_apiserver_``
+Prometheus namespace. KVStoreMesh metrics are exported under the ``cilium_kvstoremesh_``
 Prometheus namespace. Etcd metrics are exported under the ``etcd_`` Prometheus namespace.
 
 
@@ -920,3 +922,59 @@ Name                                     Labels                                 
 ``kvstore_sync_queue_size``              ``scope``, ``source_cluster``                Number of elements queued for synchronization in the kvstore
 ``kvstore_initial_sync_completed``       ``scope``, ``source_cluster``, ``action``    Whether the initial synchronization from/to the kvstore has completed
 ======================================== ============================================ ========================================================
+
+.. _kvstoremesh_metrics_reference:
+
+kvstoremesh
+-----------
+
+Configuration
+^^^^^^^^^^^^^
+
+To expose any metrics, invoke ``kvstoremesh`` with the
+``--prometheus-serve-addr`` option. This option takes a ``IP:Port`` pair but
+passing an empty IP (e.g. ``:9964``) binds the server to all available
+interfaces (there is usually only one interface in a container).
+
+Exported Metrics
+^^^^^^^^^^^^^^^^
+
+All metrics are exported under the ``cilium_kvstoremesh_`` Prometheus namespace.
+
+Remote clusters
+~~~~~~~~~~~~~~~
+
+==================================== ======================================= =================================================================
+Name                                 Labels                                                       Description
+==================================== ======================================= =================================================================
+``remote_clusters``                  ``source_cluster``                      The total number of remote clusters meshed with the local cluster
+``remote_cluster_failures``          ``source_cluster``, ``target_cluster``  The total number of failures related to the remote cluster
+``remote_cluster_last_failure_ts``   ``source_cluster``, ``target_cluster``  The timestamp of the last failure of the remote cluster
+``remote_cluster_readiness_status``  ``source_cluster``, ``target_cluster``  The readiness status of the remote cluster
+==================================== ======================================= =================================================================
+
+KVstore
+~~~~~~~
+
+======================================== ============================================ ========================================================
+Name                                     Labels                                       Description
+======================================== ============================================ ========================================================
+``kvstore_operations_duration_seconds``  ``action``, ``kind``, ``outcome``, ``scope`` Duration of kvstore operation
+``kvstore_events_queue_seconds``         ``action``, ``scope``                        Seconds waited before a received event was queued
+``kvstore_quorum_errors_total``          ``error``                                    Number of quorum errors
+``kvstore_sync_queue_size``              ``scope``, ``source_cluster``                Number of elements queued for synchronization in the kvstore
+``kvstore_initial_sync_completed``       ``scope``, ``source_cluster``, ``action``    Whether the initial synchronization from/to the kvstore has completed
+======================================== ============================================ ========================================================
+
+API Rate Limiting
+~~~~~~~~~~~~~~~~~
+
+============================================== ================================ ========================================================
+Name                                           Labels                           Description
+============================================== ================================ ========================================================
+``api_limiter_processed_requests_total``       ``api_call``, ``outcome``        Total number of API requests processed
+``api_limiter_processing_duration_seconds``    ``api_call``, ``value``          Mean and estimated processing duration in seconds
+``api_limiter_rate_limit``                     ``api_call``, ``value``          Current rate limiting configuration (limit and burst)
+``api_limiter_requests_in_flight``             ``api_call``  ``value``          Current and maximum allowed number of requests in flight
+``api_limiter_wait_duration_seconds``          ``api_call``, ``value``          Mean, min, and max wait duration
+============================================== ================================ ========================================================
--- a/kvstoremesh/.gitignore
+++ b/kvstoremesh/.gitignore
@@ -1,0 +1,1 @@
+kvstoremesh
--- a/kvstoremesh/metrics/metrics.go
+++ b/kvstoremesh/metrics/metrics.go
@@ -1,0 +1,130 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package metrics
+
+import (
+	"errors"
+	"net/http"
+	"regexp"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/collectors"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/spf13/pflag"
+
+	"github.com/cilium/cilium/pkg/hive"
+	"github.com/cilium/cilium/pkg/hive/cell"
+	"github.com/cilium/cilium/pkg/logging"
+	"github.com/cilium/cilium/pkg/logging/logfields"
+	"github.com/cilium/cilium/pkg/metrics"
+	"github.com/cilium/cilium/pkg/metrics/metric"
+	"github.com/cilium/cilium/pkg/option"
+)
+
+var (
+	log = logging.DefaultLogger.WithField(logfields.LogSubsys, "metrics")
+)
+
+var Cell = cell.Module(
+	"kvstoremesh-metrics",
+	"KVStoreMesh metrics",
+
+	cell.Config(MetricsConfig{}),
+	cell.Invoke(registerMetricsManager),
+)
+
+type MetricsConfig struct {
+	// PrometheusServeAddr IP:Port on which to serve prometheus metrics (pass ":Port" to bind on all interfaces, "" is off)
+	PrometheusServeAddr string
+}
+
+func (def MetricsConfig) Flags(flags *pflag.FlagSet) {
+	flags.String(option.PrometheusServeAddr, def.PrometheusServeAddr, "Address to serve Prometheus metrics")
+}
+
+type metricsManager struct {
+	registry *prometheus.Registry
+	server   http.Server
+
+	metrics []metric.WithMetadata
+}
+
+type params struct {
+	cell.In
+
+	MetricsConfig
+
+	Metrics []metric.WithMetadata `group:"hive-metrics"`
+}
+
+func registerMetricsManager(lc hive.Lifecycle, params params) error {
+	manager := metricsManager{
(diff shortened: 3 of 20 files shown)


kvstoremesh/metrics/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+		metrics.APILimiterRateLimit,
+		metrics.APILimiterAdjustmentFactor,
+	)
+
+	for _, metric := range mm.metrics {
+		mm.registry.MustRegister(metric.(prometheus.Collector))
+	}
+
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.HandlerFor(mm.registry, promhttp.HandlerOpts{}))
```
tommyp1ckles, line 108: The Agent server enables OpenMetrics in HandlerOpts, Im not that familiar with the option but might be something to consider to make sure metrics are consistent among the different components.

kvstoremesh/metrics/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return nil
+}
+
+func (mm *metricsManager) Start(hive.HookContext) error {
+	log.Info("Registering metrics")
+
+	mm.registry.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
+	mm.registry.MustRegister(collectors.NewGoCollector())
```
tommyp1ckles: In the Agent we recently enabled sched latency metrics: https://github.com/cilium/cilium/blob/main/pkg/metrics/registry.go#L116 by default.

The idea was that when people run Cilium in severely CPU constrained environments that we sometimes begin to see strange behaviour as goroutines get large scheduling latency.

Might be worth something to have here too, keep things consistent and we can easily build alerting around that.

kvstoremesh/metrics/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+		log.Info("Prometheus metrics are disabled")
+	}
+
+	return nil
+}
+
+func (mm *metricsManager) Start(hive.HookContext) error {
+	log.Info("Registering metrics")
+
+	mm.registry.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
```
tommyp1ckles, line 80: Should these be namespaced (i.e. `cilium_kvstoremesh`)?

kvstoremesh/metrics/metrics.go
giorio94: I could find `EnableOpenMetrics` being used as an option only for the [Hubble metrics](https://github.com/cilium/cilium/blob/7ec43fe8c6bf53bba6476132200aa1623806713b/pkg/hubble/metrics/metrics.go#L78-L79), but not for the [agent metrics](https://github.com/cilium/cilium/blob/7ec43fe8c6bf53bba6476132200aa1623806713b/pkg/metrics/registry.go#LL79C17-L79C17). Am I missing anything?

kvstoremesh/metrics/metrics.go
giorio94: It seems that there is currently a bit of inconsistency in the code base, since process-related metrics are currently namespaced in the agents and cilium-operator, and non-namespaced in hubble relay and in the clustermesh-apiserver.  Go related metrics instead are always non-namespaced. Do you feel it would be better change the process ones to be namespaced in clustermesh-apiserver/kvstoremesh?

kvstoremesh/metrics/metrics.go
giorio94: Added, thanks for the heads up.

kvstoremesh/metrics/metrics.go
tommyp1ckles: Good point, the agent/operator metrics declarations also all specify their namespace in-line as well. 
It looks like hubble metrics are just name spaced individually with the "hubble" namespace.

So from what I can tell, with reuising the existing pkg/metrics declarations this will result in using the "cilium_" namespace?

Given that we're reusing the same metric name I think it would make sense for this component to have its own namespace prefix to distinguish between agent/kvstoremesh controller metrics.

kvstoremesh/metrics/metrics.go
```diff
@@ -0,0 +1,129 @@
...
+	mm.registry.MustRegister(collectors.NewGoCollector(
+		collectors.WithGoCollectorRuntimeMetrics(
+			collectors.GoRuntimeMetricsRule{Matcher: regexp.MustCompile(`^/sched/latencies:seconds`)},
+		)))
+
+	// Constructing the legacy metrics and register them at the metrics global variable.
+	// This is a hack until we can unify this metrics manager with the metrics.Registry.
+	metrics.NewLegacyMetrics()
+	mm.registry.MustRegister(
+		metrics.KVStoreOperationsDuration,
```
tommyp1ckles, line 91: Small optional point, it's always nice to add a "version" metrics such as `cilium_version` for a couple reasons:

* Discover ability: low cardinality, generally always enabled metric that can be used to detect running deployments/labels etc.
* Monitoring component versions via Prometheus.

kvstoremesh/metrics/metrics.go
tommyp1ckles: Yeah not sure if this is really necessary.

Documentation/observability/metrics.rst
```diff
@@ -179,9 +179,11 @@ metrics conform to the OpenMetrics requirements.
 Cluster Mesh API Server Metrics
 ===============================
 
-Cluster Mesh API Server metrics provide insights into both the state of the
-``clustermesh-apiserver`` process and the sidecar etcd instance.
+Cluster Mesh API Server metrics provide insights into the state of the
+``clustermesh-apiserver`` process, the ``kvstoremesh`` process (if enabled)
```
zacharysarah: Serial comma nit
```suggestion
``clustermesh-apiserver`` process, the ``kvstoremesh`` process (if enabled),
```

Documentation/observability/metrics.rst
```diff
@@ -920,3 +922,59 @@ Name                                     Labels
...
+
+kvstoremesh
+-----------
+
+Configuration
+^^^^^^^^^^^^^
+
+To expose any metrics, invoke ``kvstoremesh`` with the
+``--prometheus-serve-addr`` option. This option takes a ``IP:Port`` pair but
+passing an empty IP (e.g. ``:9964``) will bind the server to all available
```
zacharysarah: Use present tense
```suggestion
passing an empty IP (e.g. ``:9964``) binds the server to all available
```

Documentation/observability/metrics.rst
```diff
@@ -920,3 +922,59 @@ Name                                     Labels
...
+interfaces (there is usually only one in a container).
```
zacharysarah: One what? Please verify.
```suggestion
interfaces (there is usually only one interface in a container).
```

zacharysarah (approved):
Some small changes required, otherwise LGTM. Approving to unblock with the understanding that changes are required prior to merge.

kvstoremesh/.gitignore
```diff
@@ -0,0 +1 @@
+kvstoremesh
```
YutaroHayakawa, line 1: Nit: It's a bit odd to me that the name of the executable is `kvstoremesh` maybe `kvstoremesh-apiserver` or `kvstore-agent`?

YutaroHayakawa (approved):
LGTM 👍 Beautiful code!

kvstoremesh/metrics/metrics.go
giorio94: >  So from what I can tell, with reuising the existing pkg/metrics declarations this will result in using the "cilium_" namespace?

As for the cilium specific metrics, the [namespace is changed to `cilium_kvstoremesh`](https://github.com/cilium/cilium/blob/96abcb95bc34356c9465de536d35749039419815/kvstoremesh/main.go#L45) (and `cilium_clustermesh_apiserver` in the clustermesh-apiserver case). The only metrics that are currently not namespaced in those two cases are the generic ones (go and process).

kvstoremesh/.gitignore
giorio94: Honestly I was doubtful about which name to pick, and I ended up using plain *kvstoremesh* mainly for brevity. I'm personally a bit against `kvstoremesh-apiserver` because it is not actually an apiserver (we already have the same problem with the clustermesh-apiserver, which makes understanding its role a bit difficult). The same applies for `kvstoremesh-operator`, because it is not strictly an operator in the kubernetes sense. I don't have strong opinion though.

kvstoremesh/metrics/metrics.go
giorio94: Added, thanks!

giorio94:
Rebased onto main to pick the fixes for conformance-ginkgo

pkg/defaults/defaults.go
```diff
@@ -26,6 +26,9 @@ const (
 	// GopsPortApiserver is the default value for option.GopsPort in the apiserver
 	GopsPortApiserver = 9892
 
+	// GopsPortKVStoreMesh is the default value for option.GopsPort in kvstoremesh
+	GopsPortKVStoreMesh = 9893
```
tklauser: This value is already used for Hubble Relay's gops port:

https://github.com/cilium/cilium/blob/68b3ae1918b1bd81e2f82d4fb83c896f8fae4f9f/Documentation/operations/system_requirements.rst?plain=1#L431

https://github.com/cilium/cilium/blob/68b3ae1918b1bd81e2f82d4fb83c896f8fae4f9f/pkg/hubble/relay/defaults/defaults.go#L20-L21

I guess this should be 9894 instead.

Would also be nice to document the new value in the table:

https://github.com/cilium/cilium/blob/68b3ae1918b1bd81e2f82d4fb83c896f8fae4f9f/Documentation/operations/system_requirements.rst?plain=1#L413-L436

kvstoremesh/option/config.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/cilium/cilium/pkg/option"
+)
+
+const (
+	// PprofAddress is the default value for pprof in kvstoremesh
+	PprofAddress = "localhost"
+
+	// PprofPort is the default value for pprof in kvstoremesh
+	PprofPort = 6064
```
tklauser, line 17: Would be nice to document this in the following table along with the other pprof ports:

https://github.com/cilium/cilium/blob/68b3ae1918b1bd81e2f82d4fb83c896f8fae4f9f/Documentation/operations/system_requirements.rst?plain=1#L413-L436

pkg/defaults/defaults.go
giorio94: KVStoreMesh is not running in host network, hence it does not risk conflicting with the ports used on the nodes. Do you think it is better to add it to the system requirements also in that case? (I see that clustermesh-apiserver is present there, although it's unclear to me whether that is intentional or due to historical reasons).

tklauser: I'd still suggest to keep the port values distinct for consistency.

Feel free to omit it from the table, that one admittedly is a bit misleading probably for historical reasons and could use a cleanup.

giorio94: I've changed the port value to 9894 and opened https://github.com/cilium/cilium/pull/26230 for the cleanup. PTAL

tklauser (approved):
Thanks!

giorio94:
Rebased onto main to fix CI failure

michi-covalent (approved):
mostly rubber stamping on behalf of the tophat team. lgtm assuming it's getting reviewed by the actual code owner (sig-clustermesh team) ✅

giorio94:
@tommyp1ckles PTAL

aditighag (approved):
@giorio94 It's not obvious if there are any policy related changes in the PR. It looks like sig-policy review was requested after this change - https://github.com/cilium/cilium/compare/9a82196551f3ce7b212c8884c3a5b62895d1d489..d5abdcadbb249cc901ae0511ab332d3346864ee8. 
Let me know if that's not the case.

kvstoremesh/metrics/metrics.go
tommyp1ckles: ah I see, missed that line, thought this would be the same as the agent. Ok looks good thanks.

tommyp1ckles (approved):
Metrics LGTM 🙏

giorio94:
> @giorio94 It's not obvious if there are any policy related changes in the PR. It looks like sig-policy review was requested after this change - https://github.com/cilium/cilium/compare/9a82196551f3ce7b212c8884c3a5b62895d1d489..d5abdcadbb249cc901ae0511ab332d3346864ee8. Let me know if that's not the case.

Yeah, there's no policy-related change. It's only the metrics documentation page which seems to be also owned by sig-policy. Thanks!

PR #26085, opened 2023-06-09
gha: fix conformance-ginkgo base branch retrieval

giorio94:
Retrieving the base branch from `github.base_branch` does not work when the action is triggered by a comment, because that field is not populated. Let's retrieve it from the json information about the PR. Additionally, let's make sure that we use the correct URL to retrieve the PR information also when the workflow is triggered by a pull_request event.

Link to running workflow: https://github.com/cilium/cilium/actions/runs/5224161919/jobs/9432015821?pr=26085


qmonnet (approved):
Looks good to me, thanks for fixing this!

brlbil (approved):
LGTM

aanm (approved):

PR #26086, opened 2023-06-09
.github: Fail if print-chart-version.sh fails or does not exist

chancez:
If `print-chart-version.sh` doesn't exist (ie: if it isn't back ported to a stable branch) the workflow does not fail, so make it fail so we don't push invalid charts.

### DIFF
--- a/.github/workflows/push-chart-ci.yaml
+++ b/.github/workflows/push-chart-ci.yaml
@@ -98,7 +98,14 @@ jobs:
         fetch-depth: 0
     - id: get-version
       run: |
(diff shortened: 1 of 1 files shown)


viktor-kurchenko (approved):

michi-covalent (approved):
magical pipe

joestringer (approved):

michi-covalent (approved):

michi-covalent:
chance tested this in a forked repo. no need to run the full CI since it only modifies .github/workflows/push-chart-ci.yaml. let's wait for the basic CI to finish, and i'll merge it.

PR #26092, opened 2023-06-09
docs: reword incorrect L7 policy description

peterj:
Remove the extra $ characters in the paths (I didn't see this documented anywhere, so I am assuming it might be a left over from previous versions?).

Also, I am fixing incorrect description of the GET /public policy in the L7 section.

### DIFF
--- a/Documentation/security/policy/language.rst
+++ b/Documentation/security/policy/language.rst
@@ -859,10 +859,10 @@ Headers
 Allow GET /public
 ~~~~~~~~~~~~~~~~~
 
-The following example allows ``GET`` requests to the URL ``/public`` to be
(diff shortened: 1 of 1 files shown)


qmonnet (changes requested):
The fix for the description of the `GET /public` sounds good.

For the dollars, I'm not sure this is a mistake, maybe just something that's missing documentation. I can see some other examples using it, including in tests (`pkg/policy/api/utils_test.go`). My guess is that it means the path should match until the end, and not just be a prefix. So please let's clarify whether this is actually something that needs an update before removing these `$`.

qmonnet (approved):
Looks good, thank you!

PR #26105, opened 2023-06-12
cli: add "cilium bpf config list"

mhofstetter:
This PR introduces the possibility to list the content of the BPF map `cilium_runtime_config` via the cilium cli command `cilium bpf config list`.

In addition, these information are dumped during sysdump with the bugtool.


pkg/maps/configmap/config_map.go
```diff
@@ -32,14 +32,21 @@ const (
 	AgentLiveness
```
mhofstetter, line 31: @dylandreimerink @jrajahalme it looks like the "marker" entry `UsedEntries` is unused and lost its meaning. When L2 Announcement has been introduced, a new entry has been added to the end (whereas the intention of the marker entry was to be the last entry).

should we delete the entry and move `AgentLiveness` to second position?

dylandreimerink: Moving `AgentLiveness` is alright with me, as long as we do it before the feature freeze. Moving after that could pose upgrade issues if the map is not recreated. I also have no idea what `UsedEntries` was supposed to be used for.

dylandreimerink (approved):

mhofstetter:
removed review-request for @aditighag - as @dylandreimerink covered sig-datapath. L2Announcement (by dylan) is also using the runtime config map.

mhofstetter:
rebased to `main` to get rid of jenkins checks :)

pkg/maps/configmap/config_map.go
mhofstetter: @dylandreimerink thanks - removed the constant (in agent & `common.h`)

asauber (approved):
Looks great. The de-duplication of the index strings is nice

PR #26113, opened 2023-06-12
ipsec: Split removeStaleXFRMOnce to fix deprioritization issue

jschwinger233:
We expect deprioritizeOldOutPolicy() to be executed for IPv4 and IPv6, but removeStaleXFRMOnce prevents the second call. If both IPv4 and IPv6 are enabled, v6 xfrm policy won't be deprioritized due to this issue.

This commit fixes it by spliting removeStaleXFRMOnce into removeStaleIPv4XFRMOnce and removeStaleIPv6XFRMOnce.

Fixes: https://github.com/cilium/cilium/commit/688dc9ac802b11f6c16a9cbc5d60baaf77bd6ed0

### DIFF
--- a/pkg/datapath/linux/ipsec/ipsec_linux.go
+++ b/pkg/datapath/linux/ipsec/ipsec_linux.go
@@ -123,7 +123,8 @@ var (
 
 	// To attempt to remove any stale XFRM configs once at startup, after
 	// we've added the catch-all default-drop policy.
-	removeStaleXFRMOnce sync.Once
+	removeStaleIPv4XFRMOnce sync.Once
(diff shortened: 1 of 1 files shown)


jibi (approved):

pchaigno (approved):
Nice catch!

Could you add in the commit description a reference to the commit that introduced the bug?

jschwinger233:
Thanks Paul, will make it more clear. 

CI seems to have caught a bug cause by this change, let me investigate it.

pchaigno:
> CI seems to have caught a bug cause by this change, let me investigate it.

If it is the `with bpf_host` ginkgo IPsec test failing, do check open flake issues. We have a flake on that.

PR #23831, opened 2023-02-16
Move ct_lookup in bpf_host.c to a separate tailcall

gentoo-root:
Move ct_lookup in bpf_host.c to a separate tailcall to reduce verifier complexity.

Fixes: #17142

PR #26118, opened 2023-06-12
bpf: fix error handling for invoke_tailcall_if()

julianwiedmann:
`invoke_tailcall_if()` either inlines the target function, or calls it via
tail-call. In the inline variant, the `return` causes us to skip over any
subsequent error handling (eg drop notification) in the caller.

Rework the macro so that it properly assigns the `ret` value.

Also fix up one user that didn't handle returned errors.

### DIFF
--- a/bpf/bpf_host.c
+++ b/bpf/bpf_host.c
@@ -407,13 +407,13 @@ tail_handle_ipv6(struct __ctx_buff *ctx, const bool from_host)
 	if (ret == CTX_ACT_OK) {
 		ctx_store_meta(ctx, CB_SRC_LABEL, proxy_identity);
 		if (from_host)
-			invoke_tailcall_if(is_defined(ENABLE_HOST_FIREWALL),
-					   CILIUM_CALL_IPV6_CONT_FROM_HOST,
-					   tail_handle_ipv6_cont_from_host);
+			ret = invoke_tailcall_if(is_defined(ENABLE_HOST_FIREWALL),
+						 CILIUM_CALL_IPV6_CONT_FROM_HOST,
+						 tail_handle_ipv6_cont_from_host);
 		else
-			invoke_tailcall_if(is_defined(ENABLE_HOST_FIREWALL),
-					   CILIUM_CALL_IPV6_CONT_FROM_NETDEV,
-					   tail_handle_ipv6_cont_from_netdev);
+			ret = invoke_tailcall_if(is_defined(ENABLE_HOST_FIREWALL),
+						 CILIUM_CALL_IPV6_CONT_FROM_NETDEV,
+						 tail_handle_ipv6_cont_from_netdev);
 	}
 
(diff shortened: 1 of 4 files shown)


julianwiedmann:
Bit of a funky fail in the runtime tests (https://github.com/cilium/cilium/issues/26150), re-running.

bpf/bpf_host.c
```diff
@@ -409,11 +409,13 @@ tail_handle_ipv6(struct __ctx_buff *ctx, const bool from_host)
 		if (from_host)
 			invoke_tailcall_if(is_defined(ENABLE_HOST_FIREWALL),
 					   CILIUM_CALL_IPV6_CONT_FROM_HOST,
-					   tail_handle_ipv6_cont_from_host);
+					   tail_handle_ipv6_cont_from_host,
```
aspsk: How about
```
ret = invoke_tailcall_if(is_defined(ENABLE_HOST_FIREWALL),
					   CILIUM_CALL_IPV6_CONT_FROM_HOST,
					   tail_handle_ipv6_cont_from_host);
```
?

julianwiedmann: That would be easier to read, but I couldn't think of a pretty way to make it work for when the macro expands into a tail-call.

aspsk: But then it just doesn't return, so something like
```
#define __invoke_tailcall_if_1(NAME, FUNC)	({			\
		ep_tail_call(ctx, NAME);				\
                DROP_MISSED_TAIL_CALL;				\
})
```
should work

julianwiedmann: yes that should do the trick, thanks :pray:

julianwiedmann:
Picked up @aspsk's suggestion on how to build the macro, rebased. Also fixed up one user of the macro.

aspsk (approved):

nbusseneau:
@julianwiedmann Backporting the changes in this PR to v1.13 yields conflicts as they depend on changes made in #23831, which was not backported to v1.13. I'm not competent enough in BPF to fully understand what's going on, can you double-check if it's OK to backport #23831 to v1.13?

julianwiedmann:
> @julianwiedmann Backporting the changes in this PR to v1.13 yields conflicts as they depend on changes made in #23831, which was not backported to v1.13. I'm not competent enough in BPF to fully understand what's going on, can you double-check if it's OK to backport #23831 to v1.13?

Yep, v1.13 has less occurences of `invoke_tailcall_if()`. It's fine to ignore the non-existent ones.

But I see @tklauser has already decided how to progress, I'll queue up a backport PR then :grin: .

tklauser:
Was just about to write a reply :smile: Also happy to incorporate the patch in a 1.13 backport this week if it's reasonably easy. But I guess @julianwiedmann will have more context doing it.

Issue #25816, opened 2023-06-01
CI: ClusterMesh (ci-multicluster)


pippolo84:
## CI failure

https://github.com/cilium/cilium/actions/runs/5141598573/jobs/9254287333

Step: Setup & Test (7, vxlan, ipv4, wireguard, iptables)

Output of test failure:

```
📋 Test Report
connectivity test failed: 2 tests failed
❌ 2/43 tests failed (2/277 actions), 13 tests skipped, 1 scenarios skipped:
Test [all-ingress-deny]:
  ❌ all-ingress-deny/pod-to-cidr/external-1111-0: cilium-test/client2-76f4d7c5bc-zp4w4 (10.242.1.75) -> external-1111 (1.1.1.1:443)
Test [to-cidr-external-knp]:
  ❌ to-cidr-external-knp/pod-to-cidr/external-1111-1: cilium-test/client2-76f4d7c5bc-zp4w4 (10.242.1.75) -> external-1111 (1.1.1.1:443)
Error: Process completed with exit code 1.
```

Sysdump: [cilium-sysdumps-7.zip](https://github.com/cilium/cilium/files/11623519/cilium-sysdumps-7.zip)

learnitall:
Hey @giorio94, can you take a look?

rastislavs:
Hit in similar ones in https://github.com/cilium/cilium/pull/26040:

Installation and Connectivity Test (7, vxlan, ipv4, wireguard, iptables):
```
❌ 1/43 tests failed (1/277 actions), 13 tests skipped, 1 scenarios skipped:
Test [all-ingress-deny-knp]:
  ❌ all-ingress-deny-knp/pod-to-cidr/external-1111-0: cilium-test/client-6965d549d5-ksdvq (10.242.1.177) -> external-1111 (1.1.1.1:443)
```

Installation and Connectivity Test (6, vxlan, ipv4, disabled, none):
```
❌ 1/43 tests failed (1/288 actions), 13 tests skipped, 0 scenarios skipped:
Test [all-ingress-deny]:
  ❌ all-ingress-deny/pod-to-cidr/external-1111-0: cilium-test/client-6965d549d5-bmcb4 (10.242.1.164) -> external-1111 (1.1.1.1:443)
```

Sysdumps:
- [cilium-sysdumps-7.zip](https://github.com/cilium/cilium/files/11745648/cilium-sysdumps-7.zip)
[...]

PR #26124, opened 2023-06-12
clustermesh: Introduce ClusterID reservation mechanism

marseel:
Improve the lifecycle of cluster config by introducing the ClusterID reservation mechanism. More details in commit message

### DIFF
--- a/pkg/clustermesh/clustermesh.go
+++ b/pkg/clustermesh/clustermesh.go
@@ -5,12 +5,12 @@ package clustermesh
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/cilium/cilium/api/v1/models"
 	"github.com/cilium/cilium/pkg/allocator"
 	"github.com/cilium/cilium/pkg/clustermesh/internal"
 	"github.com/cilium/cilium/pkg/clustermesh/types"
-	cmtypes "github.com/cilium/cilium/pkg/clustermesh/types"
 	"github.com/cilium/cilium/pkg/hive"
 	"github.com/cilium/cilium/pkg/hive/cell"
 	"github.com/cilium/cilium/pkg/ipcache"
@@ -88,12 +88,45 @@ type ClusterMesh struct {
 	// internal implements the common logic to connect to remote clusters.
 	internal internal.ClusterMesh
 
+	usedIDs *ClusterMeshUsedIDs
 	// globalServices is a list of all global services. The datastructure
 	// is protected by its own mutex inside the structure.
 	globalServices *globalServiceCache
 
 	// nodeName is the name of the local node. This is used for logging and metrics
 	nodeName string
+}
+
+type ClusterMeshUsedIDs struct {
+	usedClusterIDs      map[uint32]struct{}
+	usedClusterIDsMutex lock.Mutex
+}
+
+func newClusterMeshUsedIDs() *ClusterMeshUsedIDs {
+	return &ClusterMeshUsedIDs{
+		usedClusterIDs: make(map[uint32]struct{}),
+	}
+}
+
+func (cm *ClusterMeshUsedIDs) reserveClusterID(clusterID uint32) error {
+	cm.usedClusterIDsMutex.Lock()
+	defer cm.usedClusterIDsMutex.Unlock()
+
+	if _, ok := cm.usedClusterIDs[clusterID]; ok {
+		// ClusterID already used
+		return fmt.Errorf("clusterID %d is already used", clusterID)
+	}
+
+	cm.usedClusterIDs[clusterID] = struct{}{}
+
+	return nil
+}
+
+func (cm *ClusterMeshUsedIDs) releaseClusterID(clusterID uint32) {
+	cm.usedClusterIDsMutex.Lock()
+	defer cm.usedClusterIDsMutex.Unlock()
+
+	delete(cm.usedClusterIDs, clusterID)
 }
 
 // NewClusterMesh creates a new remote cluster cache based on the
@@ -106,6 +139,7 @@ func NewClusterMesh(lifecycle hive.Lifecycle, c Configuration) *ClusterMesh {
 	nodeName := nodeTypes.GetName()
 	cm := &ClusterMesh{
 		conf:     c,
+		usedIDs:  newClusterMeshUsedIDs(),
 		nodeName: nodeName,
 		globalServices: newGlobalServiceCache(
 			c.Metrics.TotalGlobalServices.WithLabelValues(c.ClusterName, nodeName),
@@ -130,10 +164,11 @@ func NewClusterMesh(lifecycle hive.Lifecycle, c Configuration) *ClusterMesh {
 
 func (cm *ClusterMesh) newRemoteCluster(name string, status internal.StatusFunc) internal.RemoteCluster {
 	rc := &remoteCluster{
-		name:   name,
-		mesh:   cm,
-		status: status,
-		swg:    lock.NewStoppableWaitGroup(),
+		name:    name,
+		mesh:    cm,
+		usedIDs: cm.usedIDs,
+		status:  status,
+		swg:     lock.NewStoppableWaitGroup(),
 	}
 
 	rc.remoteNodes = store.NewRestartableWatchStore(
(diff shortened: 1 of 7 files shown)


pkg/clustermesh/types/types.go
```diff
@@ -38,25 +38,15 @@ type CiliumClusterConfigCapabilities struct {
 	SyncedCanaries bool `json:"syncedCanaries,omitempty"`
 }
 
-func (c0 *CiliumClusterConfig) IsCompatible(c1 *CiliumClusterConfig) error {
-	if c1 == nil {
-		// When remote cluster doesn't have cluster config, we
-		// currently just bypass the validation for compatibility.
-		// Otherwise, we cannot connect with older cluster which
-		// doesn't support cluster config feature.
```
giorio94: Nit. It could make sense to preserve the reasoning about tolerating a `nil` CiliumClusterConfig.

pkg/clustermesh/remote_cluster.go
```diff
@@ -129,3 +139,28 @@ func (rc *remoteCluster) Status() *models.RemoteCluster {
 	status.NumIdentities = int64(rc.remoteIdentityCache.NumEntries())
 	return status
 }
+
+func (rc *remoteCluster) onUpdateConfig(newConfig *cmtypes.CiliumClusterConfig) error {
+	oldConfig := rc.config
```
giorio94, line 156: Nit. I think it would be better to be consistent about guarding `rc.config` with a lock (it is safe to not protect `rc.config` with a lock given that it is always accessed by a single thread only).

pkg/clustermesh/remote_cluster.go
```diff
@@ -77,7 +81,7 @@ func (rc *remoteCluster) Run(ctx context.Context, backend kvstore.BackendOperati
 	defer remoteIdentityCache.Close()
 
 	rc.mutex.Lock()
-	rc.config = config
+	rc.config = newConfig
```
giorio94: I'd perform this operation inside `onUpdateConfig` for consistency, and to ensure that `rc.config` is always set after that the cluster ID has been acquired. Otherwise, it might not be correctly released if an error occurred in the meanwhile (e.g., when initializing the remote identities watcher).

pkg/clustermesh/remote_cluster.go
```diff
@@ -59,14 +59,18 @@ type remoteCluster struct {
 	swg *lock.StoppableWaitGroup
 }
 
-func (rc *remoteCluster) Run(ctx context.Context, backend kvstore.BackendOperations, config *cmtypes.CiliumClusterConfig) error {
-	if err := rc.mesh.canConnect(rc.name, config); err != nil {
+func (rc *remoteCluster) Run(ctx context.Context, backend kvstore.BackendOperations, newConfig *cmtypes.CiliumClusterConfig) error {
```
giorio94: Nit. I'd personally revert the parameter name change (matter of taste though).

pkg/clustermesh/remote_cluster.go
```diff
@@ -129,3 +139,28 @@ func (rc *remoteCluster) Status() *models.RemoteCluster {
...
+				return err
+			}
+		} else {
+			if err := rc.mesh.reserveClusterID(newConfig.ID); err != nil {
+				return err
+			}
+
+			rc.mesh.releaseClusterID(oldConfig.ID)
+		}
+	}
```
giorio94: This logic could be simplified to look like:

```go
	if newConfig != nil {
		if err := rc.mesh.reserveClusterID(newConfig.ID); err != nil {
			return err
		}
	}

	if oldConfig != nil {
		rc.mesh.releaseClusterID(oldConfig.ID)
	}
```

Additionally, we need to shortcircuit this logic if the cluster ID did not change, otherwise we would fail acquiring the cluster ID that we already own. Finally, it would be nice to add a unit test to ensure that this function handles all cases correctly.

pkg/clustermesh/clustermesh.go
```diff
@@ -173,6 +178,29 @@ func (cm *ClusterMesh) NumReadyClusters() int {
...
+		// ClusterID already used
+		return fmt.Errorf("clusterID %d is already used", clusterID)
+	}
+
+	cm.usedClusterIDs[clusterID] = struct{}{}
+
+	return nil
+}
+
+func (cm *ClusterMesh) releaseClusterID(clusterID uint32) error {
```
giorio94: I'd drop the `error` return parameter given that it is always nil, and never checked by the subsequent commit.

pkg/clustermesh/clustermesh.go
```diff
@@ -84,6 +84,9 @@ type ClusterMesh struct {
 	// internal implements the common logic to connect to remote clusters.
 	internal internal.ClusterMesh
 
+	usedClusterIDs      map[uint32]struct{}
+	usedClusterIDsMutex lock.Mutex
```
giorio94: Nit. It could make sense to extract this in a separate object, to loosen the circular dependency between `ClusterMesh` and `remoteCluster`.

pkg/clustermesh/types/types.go
```diff
@@ -38,25 +38,15 @@ type CiliumClusterConfigCapabilities struct {
...
-	} else {
-		// Remote cluster has cluster config. Do validations.
+	}
 
-		// ID shouldn't be duplicated
-		if c0.ID == c1.ID {
-			return fmt.Errorf("duplicated cluster id")
-		}
+	if err := ValidateClusterID(c.ID); err != nil {
+		return err
```
giorio94, line 60: This check is potentially problematic in the kvstoremesh case. Indeed, kvstoremesh could connect to a remote cluster which does not expose the cluster config. In that case, it would propagate a cluster config with the appropriate capabilities set, but the cluster ID unset (i.e., = 0). Hence, this check would fail.

marseel: Removed id = 0 case from validation

pkg/clustermesh/clustermesh.go
```diff
@@ -92,6 +93,38 @@ type ClusterMesh struct {
 	nodeName string
 }
 
+type ClusterMeshUsedIds struct {
```
giorio94:
```suggestion
type ClusterMeshUsedIDs struct {
```

Very minor nit, just for consistency. It also applies to sibling methods/fields which include `ID` in their name.

pkg/clustermesh/remote_cluster.go
```diff
@@ -129,3 +137,22 @@ func (rc *remoteCluster) Status() *models.RemoteCluster {
...
+	oldConfig := rc.config
+
+	if newConfig != nil && oldConfig != nil && newConfig.ID == oldConfig.ID {
+		return nil
+	}
+	if newConfig != nil {
+		if err := rc.usedIds.reserveClusterID(newConfig.ID); err != nil {
+			return err
+		}
+		rc.config = newConfig
```
giorio94: This assignment should be moved to the end of the function, as it needs to be executed also in case the new config is nil. Otherwise, we might end up attempting to release the same cluster ID a second time, although it could have been acquired by another cluster.

pkg/clustermesh/types/types.go
giorio94: I'd personally not drop the id == 0 check from `ValidateClusterID`, because it might break the expectations of already existing callers of that function. What about changing the check above:

```diff
--- if c == nil {
+++ if c == nil || c.ID == 0 {
```

At the end of the day, they are equivalent (they both essentially mean that the cluster ID is unknown), and if one is permitted also the other should be.

giorio94 (commented):
Looks mostly ready to me, just a couple of comments inline.

marseel:
/ci-multicluster

Similar flake to https://github.com/cilium/cilium/issues/25816

pkg/clustermesh/remote_cluster.go
```diff
@@ -61,10 +62,14 @@ type remoteCluster struct {
 }
 
 func (rc *remoteCluster) Run(ctx context.Context, backend kvstore.BackendOperations, config *cmtypes.CiliumClusterConfig, ready chan<- error) {
-	if err := rc.mesh.canConnect(rc.name, config); err != nil {
+	if err := config.Validate(); err != nil {
+		ready <- err
+		close(ready)
```
giorio94, line 67: The return statement is missing

pkg/clustermesh/remote_cluster.go
```diff
@@ -61,10 +62,14 @@ type remoteCluster struct {
...
+	}
+
+	if err := rc.onUpdateConfig(config); err != nil {
 		ready <- err
 		close(ready)
-		return
```
giorio94, line 67: The return statement should be restored.

giorio94 (approved):
/lgtm

marseel:
This was mostly refactoring of existing code and not really introducing any new functionality. 
The integration test now covers a case where we reserve multiple IDs, change the ID of the existing cluster, and also free IDs when removing clusters, but I agree it could cover more cases around the error handling.

PR #26127, opened 2023-06-12
k8s / policy: allow all services for toServices when using highscale ipcache

squeed:
Normally, toServices rules only allow Endpoints that are external to the cluster. This is to prevent conflict with entries already existing in the ipcache.

However, this is not relevant with highscale ipcache, where podips are normally excluded from the ipcache anyways. So, allow all services to be specified in toServices egress rules when highscale is enabled.

### DIFF
--- a/pkg/k8s/rule_translate.go
+++ b/pkg/k8s/rule_translate.go
@@ -9,6 +9,7 @@ import (
 	"net/netip"
 
 	"github.com/cilium/cilium/pkg/k8s/slim/k8s/apis/labels"
+	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/policy"
 	"github.com/cilium/cilium/pkg/policy/api"
 	"github.com/cilium/cilium/pkg/slices"
@@ -228,7 +229,14 @@ func PreprocessRules(r api.Rules, cache *ServiceCache) error {
 		}
 		for ns, ep := range cache.endpoints {
 			svc, ok := cache.services[ns]
-			if ok && svc.IsExternal() {
+			// Normally, only services without a label selector (i.e. empty services)
+			// are allowed as targets of a toServices rule.
+			// This is to minimize the chances of a pod IP being selected by this rule, which might
+			// cause conflicting entries in the ipcache.
+			//
+			// This requirement, however, is dropped for HighScale IPCache mode, because pod IPs are
+			// normally excluded from the ipcache regardless.
+			if ok && (option.Config.EnableHighScaleIPcache || svc.IsExternal()) {
 				eps := ep.GetEndpoints()
 				if eps != nil {
 					t := NewK8sTranslator(ns, Endpoints{}, *eps, false, svc.Labels)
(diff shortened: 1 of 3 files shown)


pkg/k8s/rule_translate.go
```diff
@@ -234,16 +235,25 @@ func PreprocessRules(r api.Rules, cache *ServiceCache) error {
...
+			//
+			// This requirement, however, is dropped for HighScale IPCache mode, because pod IPs are
+			// normally excluded from the ipcache regardless.
+			if !ok || (!option.Config.EnableHighScaleIPcache && !svc.IsExternal()) {
+				continue
+			}
+
+			eps := ep.GetEndpoints()
+			if eps != nil {
+				t := NewK8sTranslator(ns, *eps, false, svc.Labels, false)
```
christarazi: https://github.com/cilium/cilium/pull/25687 is nearing a merge so this will conflict.

christarazi (commented):
There are some lint errors regarding `option.Config.HighScaleIPcacheEnabled()`

pkg/k8s/rule_translate.go
squeed: rebased

pkg/k8s/rule_translate.go
```diff
@@ -238,6 +246,7 @@ func PreprocessRules(r api.Rules, cache *ServiceCache) error {
 					if err != nil {
 						return err
 					}
+
```
christarazi: Unintended diff?

christarazi (approved):

christarazi:
Checkpatch is complaining with 

```
  Error: ERROR:CUSTOM: Please avoid long commit subjects (max: 75, found: 76)
```

Maybe format the commit msg to

```
k8s, policy: enable toServices for high-scale ipcache
```

christarazi:
https://github.com/cilium/cilium/actions/runs/5259867178 is failing all for the same test :thinking:

squeed:
> https://github.com/cilium/cilium/actions/runs/5259867178 is failing all for the same test thinking

cloudflare blip :-/.

pkg/k8s/rule_translate.go
```diff
@@ -228,7 +229,14 @@ func PreprocessRules(r api.Rules, cache *ServiceCache) error {
...
 			svc, ok := cache.services[ns]
-			if ok && svc.IsExternal() {
+			// Normally, only services without a label selector (i.e. empty services)
+			// are allowed as targets of a toServices rule.
+			// This is to minimize the chances of a pod IP being selected by this rule, which might
+			// cause conflicting entries in the ipcache.
+			//
+			// This requirement, however, is dropped for HighScale IPCache mode, because pod IPs are
+			// normally excluded from the ipcache regardless.
+			if ok && (option.Config.EnableHighScaleIPcache || svc.IsExternal()) {
```
tklauser, line 239: Nit: might make sense to factor out the `option.Config.EnableHighScaleIPcache || svc.IsExternal()` check to a helper func and apply the comment there instead of duplicating it in `(*K8sWatcher) k8sServiceHandler()`?

tklauser (approved):

pkg/k8s/rule_translate.go
squeed: bah, if they were in the same package, I would. A little copying is better than a circular dependency :-p

tklauser: duh 🤦

joestringer:
Is there a test for this at all? Doesn't Cilium just ignore these rule sections if they're not translated?

PR #26132, opened 2023-06-12
operator: Fix default API server addr in metrics subcommand

pippolo84:
The operator API server has been modularized into its independent cell in a previous commit.

Doing that, the flag to specify a custom address for the server has been moved from the global operator config to the new cell config.

The `metrics list` subcommand needs to know the operator API server address to query the "/metrics" endpoint.  Before the modularization, the subcommand was at least able to query the server if that was started at the default address.

This commit adds a new flag to specify the address of the operator API server. Also, the default value for this flag is set to the same default used in the operator/api server cell.  Doing that fixes the bug introduced in the modularization and also allows to specify another address if the operator API server is listening on a non-default one.

Fixes: 11d5856e17 ("operator/api: Modularize api server")

### DIFF
--- a/Documentation/cmdref/cilium-operator-alibabacloud_metrics_list.md
+++ b/Documentation/cmdref/cilium-operator-alibabacloud_metrics_list.md
@@ -11,9 +11,10 @@ cilium-operator-alibabacloud metrics list [flags]
 ### Options
 
 ```
-  -h, --help                   help for list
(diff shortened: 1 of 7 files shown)


tklauser (approved):

pippolo84:
Conformance E2E failure tracked in https://github.com/cilium/cilium/issues/26114, rerunning.

pippolo84:
/test-1.26-net-next

PR #26133, opened 2023-06-12
CIDRGroup reference metric will not count nonexistent CIDRGroups

### LINKED ISSUE #25022: CIDRGroup references metric should count live references only and not count nonexistent CIDRGroups
The `cilium_cidrgroup_policies` metric counts all references of CIDRGroups inside a CNP even if the CIDRGroup doesn't exist (i.e. reference points to a nonexistent object).

akstron:

      GitHub issue.

Fixes: #25022

Added a check on CIDRGroupRef to make sure that the reference is valid. 

```release note
CIDRGroup reference metric will not count nonexistent CIDRGroups
```

Signed-off-by: Alok Kumar Singh alokaks601@gmail.com

### DIFF
--- a/pkg/k8s/watchers/cilium_network_policy.go
+++ b/pkg/k8s/watchers/cilium_network_policy.go
@@ -252,9 +252,11 @@ func (k *K8sWatcher) onUpsert(
 		return nil
 	}
 
-	// check if this cnp was referencing or is now referencing a
+	// check if this cnp was referencing or is now referencing at least one non-empty
 	// CiliumCIDRGroup and update the relevant metric accordingly.
-	if len(getCIDRGroupRefs(cnp)) > 0 {
+	cidrGroupRefs := getCIDRGroupRefs(cnp)
+	cidrsSets, _ := cidrGroupRefsToCIDRsSets(cidrGroupRefs, cidrGroupCache)
+	if len(cidrsSets) > 0 {
 		cidrGroupPolicies[key] = struct{}{}
 	} else {
 		delete(cidrGroupPolicies, key)
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -314,6 +314,7 @@ var (
 	CIDRGroupTranslationTimeStats = NoOpHistogram
 
 	// CIDRGroupPolicies is the number of CNPs and CCNPs referencing at least one CiliumCIDRGroup.
+	// CNPs with empty or non-existing CIDRGroupRefs are not considered
 	CIDRGroupPolicies = NoOpGauge
 
 	// Identity


pippolo84 (commented):
Thanks for your PR. :pray: 

Could you please explain in more detail in which case you ended up with `rule.CIDR` and `rule.CIDRGroupRef` both empty?

Given the kubebuilder validation that should not be possible:

https://github.com/cilium/cilium/blob/fb48741b49648f9e828c25a835bd189ab24ed0f3/pkg/policy/api/cidr.go#L39-L49

In fact, trying to apply this:

```yaml
apiVersion: "cilium.io/v2"
kind: CiliumNetworkPolicy
metadata:
  name: "cidr-group-ref-cnp"
spec:
  endpointSelector:
    matchLabels:
      run: app
  ingress:
  - fromCIDRSet:
    - cidrGroupRef: ""
```

is failing as expected:

```
The CiliumNetworkPolicy "cidr-group-ref-cnp" is invalid: spec.ingress[0].fromCIDRSet[0].cidrGroupRef: Invalid value: "": spec.ingress[0].fromCIDRSet[0].cidrGroupRef in body should match '^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'
```

And a `null` value:

```yaml
spec:
  endpointSelector:
    matchLabels:
      run: app
  ingress:
  - fromCIDRSet:
    - cidrGroupRef: null
```

 is leading to a validation error, too.

christarazi:
@pippolo84 I think the case is when the user applies a policy with a ref that doesn't exist, like

```
apiVersion: "cilium.io/v2"
kind: CiliumNetworkPolicy
metadata:
  name: "cidr-group-ref-cnp"
spec:
  endpointSelector:
    matchLabels:
      run: app
  ingress:
  - fromCIDRSet:
    - cidrGroupRef: "doesn't-exist"
```

pippolo84:
> apiVersion: "cilium.io/v2"
> kind: CiliumNetworkPolicy
> metadata:
>   name: "cidr-group-ref-cnp"
> spec:
>   endpointSelector:
>     matchLabels:
>       run: app
>   ingress:
>   - fromCIDRSet:
>     - cidrGroupRef: "doesn't-exist"

If I'm not mistaken the proposed change is just checking the length of `rule.CIDRGroupRef`, that is, the length of the string `"doesn't-exist"` in your example policy.
If we want to avoid incrementing the metric for a non-existing reference I think we need to check `cidrGroupCache` to see if a CiliumCIDRGroup with that name actually exists.

akstron:
Hi @pippolo84, probably I misunderstood the issue. What I understand from the conversation above is that we can add a particular groupRef, even though the object itself does not exist and we need to check this existence to resolve the issue.

What I suggest is we can use `cidrGroupRefsToCIDRsSets`, to translate the cidrGroupRefs we receive during metric update to get valid CIDRSets and check its length instead of the length of cidrGroupRefs.

pkg/k8s/watchers/cilium_network_policy.go
```diff
@@ -254,7 +254,11 @@ func (k *K8sWatcher) onUpsert(
 
 	// check if this cnp was referencing or is now referencing a
 	// CiliumCIDRGroup and update the relevant metric accordingly.
-	if len(getCIDRGroupRefs(cnp)) > 0 {
+	cidrGroupRefs := getCIDRGroupRefs(cnp)
+	// converting cidrGroupRefs to corresponding cidrsSets
+	// invalid cidrGroupRefs would be skipped during conversion.
```
pippolo84: The term "invalid" cidrGroupRef might be a bit confusing, since we would like to skip both non-existent and empty references.

I suggest to just add a clarification to the previous comment like this:

```
	// check if this cnp was referencing or is now referencing at least one non-empty
	// CiliumCIDRGroup and update the relevant metric accordingly.
```

Also, I think it'd be be better to update the doc about the metric itself here:

https://github.com/cilium/cilium/blob/2716ff7555c4a8856c3da2ad9ba3f8b44f19d80d/pkg/metrics/metrics.go#L316

to make it clear that we won't consider CNPs with all empty or non-existent CIDRGroupRefs.

pippolo84 (changes requested):
I think the current approach is the correct one, thanks! I've left a suggestion to improve the comment.

Could you please squash this last commit into the previous one?
Also, you need to use your real name for the `Signed-off-by` line. I suggest to read the Cilium [Contribution Guidelines](https://docs.cilium.io/en/stable/contributing/development/contributing_guide/#submitting-a-pull-request) and the [Developer's Certificate of Origin](https://docs.cilium.io/en/stable/contributing/development/contributing_guide/#developer-s-certificate-of-origin) section.

ti-mo:
@akstron Looks like your second commit overwrites the first, please squash.

pkg/k8s/watchers/cilium_network_policy.go
akstron: Thanks for reviewing. I have updated the comments accordingly.

joestringer:
It's likely that the test failures could be resolved by rebasing against the tip of the main branch.

pippolo84 (approved):
Thanks! :rocket:

christarazi (approved):

michi-covalent (approved):

aditighag (approved):

PR #26136, opened 2023-06-12
bpf: SRv6 fib on encap

ldelossa:
This is the first PR in a set of two.

In this PR we introduce new functions in lib/fib.h to fix the issue described below

A follow up PR will be opened once this is merged which carries the non-backported refactor of fib/lib.h, removing the duplicate code introduced in this PR. 

see https://github.com/cilium/cilium/pull/26048#issuecomment-1587382600 for an explanation of why this occurs in two PRs.

Original SRv6 encapsulation flow:

``` mermaid
 flowchart LR
p["pod"] 
v1["veth1"]
v2["veth2"]
hs["host stack"]
e["eth0"]
tx["tx"]

p-->|egress pkt|v1-->v2-->hs-->|fib lookup|e-->|SRv6 encap|tx
```
This results in the egress packet always choosing the output interface, source, and destination mac from the FIB lookup that occurred on the inner packet. 

SRv6 encapsulation flow corrected:

``` mermaid
 flowchart LR
p["pod"] 
v1["veth1"]
v2["veth2"]
hs["host stack"]
e["eth0"]
tx["tx"]

p-->|egress pkt|v1-->v2-->hs-->|fib lookup|e-->|SRv6 encap\n+\nfib lookup|tx
```

Now, we perform an additional fib lookup once the egress packet has been encapsulated, ensuring the encapsulated packet is forwarded correctly based on the host namespace's FIB.

Performing these actions independently is currently necessary for this SRv6 datapath fix. 
This is because we perform the `fib_lookup` already at a egress interface and we **may or may not** need to redirect to a new interface for tx. 

It is obvious to see that two fib lookups are occurring here. 
It does seem to me that this can be reduced to one, however this involves changing where encapsulation occurs and will be a larger refactor, however I will also look into this. 

```release-note
Fixes an issue where SRv6 encapsulated packets are forwarded to the wrong layer 2 next hop.
```

### DIFF
--- a/bpf/lib/egress_policies.h
+++ b/bpf/lib/egress_policies.h
@@ -4,6 +4,7 @@
 #ifndef __LIB_EGRESS_POLICIES_H_
 #define __LIB_EGRESS_POLICIES_H_
 
+#include "lib/fib.h"
 #include "lib/identity.h"
 
 #include "maps.h"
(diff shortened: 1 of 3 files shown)


ldelossa:
/test-1.26-net-next

nickolaev:
Is ti worth considering a unit test?

bpf/lib/egress_policies.h
```diff
@@ -579,6 +580,59 @@ srv6_store_meta_sid(struct __ctx_buff *ctx, const union v6addr *sid)
...
+	__s8 fib_ret;
+
+	if (!revalidate_data(ctx, &data, &data_end, &ip6))
+		return DROP_INVALID;
+
+	fib_ret = (__s8)fib_lookup_v6(ctx, &params, ip6, BPF_FIB_LOOKUP_OUTPUT);
+
+	if (fib_ret && fib_ret != BPF_FIB_LKUP_RET_NO_NEIGH)
+		return DROP_NO_FIB;
+
```
julianwiedmann: You could just fold this into the `switch()`  statement as the `default` case :).

bpf/lib/egress_policies.h
```diff
@@ -590,14 +644,21 @@ int tail_srv6_encap(struct __ctx_buff *ctx)
 	vrf_id = ctx_load_meta(ctx, CB_SRV6_VRF_ID);
 
 	ret = srv6_handling(ctx, vrf_id, &dst_sid);
+	if (ret < 0)
+		return send_drop_notify_error(ctx, SECLABEL, ret, CTX_ACT_DROP,
+					      METRIC_EGRESS);
 
+#ifdef ENABLE_IPV6
+	ret = srv6_refib(ctx);
```
julianwiedmann: Please also pass an `ext_err` parameter here to capture the detailed FIB errors, and report it via `send_drop_notify_error_ext()`.

julianwiedmann (changes requested):
Please apply one small change to the error reporting (and maybe pick up the cosmetic change while at it :grin:). Otherwise all good, thanks!

bpf/lib/egress_policies.h
ldelossa: Yeah I like that, will change.

bpf/lib/egress_policies.h
```diff
@@ -579,25 +582,87 @@ srv6_store_meta_sid(struct __ctx_buff *ctx, const union v6addr *sid)
...
+		 * We may need to redirect to the appropriate oif, if not
+		 * rewrite the layer 2 and continue processing.
+		 */
+		if (old_oif != params.l.ifindex)
+			return fib_do_redirect(ctx, true, &params,
+					      (__s8 *)ext_err, (int *)&old_oif);
+
+		if (eth_store_daddr(ctx, params.l.dmac, 0) < 0)
+			return DROP_WRITE_ERROR;
+
```
julianwiedmann, line 621: This is needed as you pushed the SRv6 encap headers onto the packet, but nothing has actually filled the outer L2 header yet. Right? Where would the source MAC come from in this case?

(I don't have a full picture of the SRv6 path, so this might be obvious ...).

bpf/lib/fib.h
```diff
@@ -40,14 +40,132 @@ static __always_inline bool fib_ok(int ret)
...
+					     sizeof(nh_params.ipv6_nh));
+			nh = &nh_params;
+
+			/* The neigh_record_ip{4,6} locations are mainly from
+			 * inbound client traffic on the load-balancer where we
+			 * know that replies need to go back to them.
+			 */
+			dmac = fib_params->l.family == AF_INET ?
+			neigh_lookup_ip4(&fib_params->l.ipv4_dst) :
+			neigh_lookup_ip6((void *)&fib_params->l.ipv6_dst);
```
julianwiedmann: Is the compiler able to eliminate these lookups for the `neigh_resolver_available()` path?

julianwiedmann (approved):
lgtm, thank you!

Two small clarifying questions, but no blocker unless you realize that something is amiss :)

bpf/lib/fib.h
ldelossa: hmm, I see your point, and I don't think it would in this case. 

I dont see why I can't wrap these lookups in the same  `neigh_resolver_available()` check as below. 

Good call.

bpf/lib/egress_policies.h
ldelossa: Kinda, 

Traffic leaves the pod right, not tunneled. 

It hits the host net stack, where a FIB lookup is done on the **outer packet** and the l2 is decided then along with output interface, its then pushed to the output interface. 

On the output interface we have SRv6 code running. 
The encap actually just pushes the **existing**  l2 down for the outer packet forward, and does the IPIP encap between the the original l2 and the original l3 headers. Leaving the l2 untouched.

So, we were just keeping the original l2 header around. As an additional point, this also means we were always sending out the same native dev too, even if the encap'd packet should route on a different link.

bpf/lib/fib.h
ldelossa: @julianwiedmann I wrapped this in the same macro check, but negated. lmk if that looks good.

ldelossa:
@jibi latest push moves the fib_lookup functions to take src/dst addresses instead of the full header

bpf/lib/fib.h
julianwiedmann: yup, makes sense :+1:

bpf/lib/egress_policies.h
julianwiedmann: ah! And this is the case where we stay on the same egress interface, so the initial smac is still valid. All good then :).

ldelossa:
@nickolaev added tests: https://github.com/cilium/cilium/pull/26136/commits/4bce0213d35cd46709708fc3f55d3dcd3a5a42c0

bpf/lib/egress_policies.h
ldelossa: Yup! and if we redirect, we re-write both :)

nbusseneau:
@ldelossa Backporting the changes in this PR to v1.13 yields conflicts as they depend on changes made in #23884, which was not backported to v1.13. I'm not competent enough in BPF to fully understand what's going on, can you double-check if it's OK to backport #23884 to v1.13?

ldelossa:
No problem @nbusseneau 

Let me see if I can open a manual backport on v1.13 that does not rely on #23884, as this PR only adds functions, and we are probably running into line conflicts, not logic conflicts.

julianwiedmann:
> No problem @nbusseneau
> 
> Let me see if I can open a manual backport on v1.13 that does not rely on #23884, as this PR only adds functions, and we are probably running into line conflicts, not logic conflicts.

Note that at the very least, we'll need to pick up the change that introduced `BPF_FIB_MAP_NO_NEIGH`. So I think this needs a proper backport, adding the corresponding label.

julianwiedmann:
I believe there's no plan to backport this into `v1.13` anymore.

PR #25491, opened 2023-05-16
backport (v1.12): policy: Fix Deny Precedence Bug

nathanjsweet:
-  ip: Add a helper to assist net.IP -> netip.Addr conversion (from @YutaroHayakawa)
  
   net.IP to netip.Addr conversion using netip.AddrFromSlice has a pitfall
that it interprets an IPv4 net.IP created with net.ParseIP as IPv4
mapped IPv6 address (see code comment for more details). This helper
safely converts net.IP to netip.Addr with preserving an original address
family, but with assumption that given net.IP is not a "real" IPv4
mapped IPv6 address which practical enough for our use cases.

- Add Tests for Deny Precedence Bug:

  Currently, when a broad "Deny" policy is paired with a specific "Unmanaged CIDR" policy, then the "Unmanaged CIDR" policy will still be inserted into the policy map for an endpoint. This results in "Deny" policies not always taking precedence over "Allow" policies. This test confirms the bugs existence.

- Fix Deny Precedence Bug:

  When the policy map state is created CIDRs are now checked against one another to ensure that deny-rules that supersede allow-rules when they should.

  `DenyPreferredInsert` has been refactored to use utility methods that make the complex boolean logic of policy precedence more atomic.

  Add `NetsContainsAny` method to `pkg/ip` to compare cases where one set of networks conatins or is equal to any network in another set.

- endpoint: Add policy.Identity Implementation

  A `policy.Identity` implementation is necessary for the incremental update to the endpoint's policy map that can occur with L7 changes. Valid deny-policy entries may prohibit these L7 changes based on CIDR rules, which are only obtainable by looking up all potentially conflicting policies' labels. Thus `l4.ToMapState` needs access to the identity allocater to lookup "random" identity labels.

- Additional commit from @joestringer to fix concurrent access to the SelectorCache, a bug the original work introduced.

- Additional commit from @nathanjsweet to prevent a potential nil pointer dereference in the endpoint code.

```release-note
policy: Promote Deny Policies from Beta to Stable
```

PR #26147, opened 2023-06-13
backport (v1.13): docs: Promote Deny Policies out of Beta

nathanjsweet:


michi-covalent (approved):

joestringer (approved):
@michi-covalent this partly comes out of my discussion here and around the release note: https://github.com/cilium/cilium/pull/25491#issuecomment-1588191135 .

I don't think we've strictly documented the bar this should meet or the process to remove features from stable. Practically speaking the quality of the feature with recent bugfixes has been raised to the level that we believe to be "stable" - no known outstanding issues for instance, and users are already running the feature in production. So if we're happy to mark it as stable on the latest branch, all other branches logically would also follow suit.

In general I'm not the biggest fan of retrospectively marking features like this, partly because tracking which micro release brought "stability" is difficult to track, and partly because it encourages a deployment lifecycle where users can stay on older versions and still get newer features. But in this particular case, given the feature maturity that has been met and the lack of documented process to guide us, I'm open to making an exception here.

I also recognize that Nate did a lot of work on improving the policy engine to get this feature to this point, and I think it's worth celebrating his contributions through a well-advertised release note.

joestringer (approved):

PR #26151, opened 2023-06-13
bpf: test: fix pktgen for IPv6 NEXTHDR_DEST option

julianwiedmann:
When building packets with multiple IPv6 options, a missing `break` for the NEXTHDR_DEST option would cause us to set up the preceding option's `nexthdr` field incorrectly.


qmonnet (approved):
:ship:

julianwiedmann:
This only touches the BPF unit tests, no need for full e2e coverage.

PR #26154, opened 2023-06-13
Agent: add support for watching kvstoremesh prefixes

giorio94:
This PR introduces a set of minor modifications to enable Cilium agents watching the kvstore prefixes used by [kvstoremesh](https://github.com/cilium/cilium/pull/26083). Please refer to the individual commit descriptions for additional details.

### DIFF
--- a/pkg/ipcache/kvstore.go
+++ b/pkg/ipcache/kvstore.go
@@ -161,6 +161,7 @@ type IPIdentityWatcher struct {
 	store   storepkg.WatchStore
 	ipcache IPCacher
 
+	clusterName                string
 	clusterID                  uint32
 	withSelfDeletionProtection bool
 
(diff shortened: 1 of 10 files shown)


giorio94:
Rebased onto main to pick the fixes for conformance-ginkgo

giorio94:
/test-1.26-net-next

tklauser (approved):
LGTM

aditighag (approved):
>clustermesh: enable watching the "cached" ipcache prefix

Thanks for the detailed commit messages. Changes lgtm for my codeowner (ipcache), just a high-level question regarding upgrade/downgrade - will the ipcache entries will be stored at both the legacy and the new path (when kvstoremesh is enabled)?

pkg/ipcache/kvstore.go
```diff
@@ -161,6 +161,7 @@ type IPIdentityWatcher struct {
 	store   storepkg.WatchStore
 	ipcache IPCacher
 
+	clusterName                string
```
aditighag, line 164: Why encode cluster name in addition to cluster id?

giorio94: Essentially, when using the *cached prefixes*, the cluster name is appended to the base prefix to isolate the information referring to different clusters. The cluster ID, instead, is not part of the etcd key, nor of the corresponding value.

giorio94:
> Just a high-level question regarding upgrade/downgrade - will the ipcache entries will be stored at both the legacy and the new path (when kvstoremesh is enabled)?

The clustermesh-apiserver still uses the old prefixes to store the information concerning the local cluster, while kvstoremesh uses the new ones to cache the information retrieved from remote clusters. This is transparent from the agents POW, because they select which ones to watch based on the *Cached* capability part of the cluster config (defaulting to the old behavior if unspecified). When performing an upgrade enabling kvstoremesh, the agents will switch from watching the remote kvstores to watching the local one (because their `cilium-clustermesh` config is changed by Helm), hence using the new prefixes, while during a downgrade (or if kvstoremesh is disabled), the opposite would happen.

giorio94:
Rebased onto main to pick the changes already introduced in https://github.com/cilium/cilium/pull/26083

YutaroHayakawa (approved):
LGTM 👍

PR #26073, opened 2023-06-09
auth: temporarily disable node-based auth gc

mhofstetter:
Currently, auth map garbage collection based on deleted node events provokes issues in combination with enabled encryption.

The actual problem is that there are combinations where garbage collection unintentionally allocates node ids when trying to resolve the node IDs for a given node IP. This results in multiple node ids for the same ip, which breaks encryption.

Therefore, this commit temporarily disables auth map garbage collection based on deleted node events.

Fixes: #25964

PR #26155, opened 2023-06-13
node_ids: introduce GetNodeID

mhofstetter:
This PR introduces the possibility to retrieve the node id for a given node IP without having to use `AllocateNodeID` which comes with the drawback of actually allocating a new node id if one doesn't exist yet for the given IP.

Example: Using `ipcache.AllocateNodeID` to lookup the node ID for a node IP during auth gc initialisation results in unintended node ID allocations if the nodeids doesn't exist yet for the cilium nodes.

By using the new method `GetNodeID` we remove this unintended side-effect.

Therefore, the temporarily disabled auth map GC functionality has been re-enabled!

Related to: https://github.com/cilium/cilium/pull/26073 & https://github.com/cilium/cilium/issues/25964

### DIFF
--- a/pkg/datapath/linux/node_ids.go
+++ b/pkg/datapath/linux/node_ids.go
@@ -61,8 +61,8 @@ func (n *linuxNodeHandler) AllocateNodeID(nodeIP net.IP) uint16 {
 }
 
 func (n *linuxNodeHandler) GetNodeIP(nodeID uint16) string {
-	n.mutex.Lock()
-	defer n.mutex.Unlock()
+	n.mutex.RLock()
+	defer n.mutex.RUnlock()
 
 	// Check for local node ID explicitly as local node IPs are not in our maps!
 	if nodeID == 0 {
@@ -73,8 +73,8 @@ func (n *linuxNodeHandler) GetNodeIP(nodeID uint16) string {
 }
 
 func (n *linuxNodeHandler) GetNodeID(nodeIP net.IP) (uint16, bool) {
-	n.mutex.Lock()
-	defer n.mutex.Unlock()
+	n.mutex.RLock()
+	defer n.mutex.RUnlock()
 
 	return n.getNodeIDForIP(nodeIP)
 }
--- a/pkg/datapath/linux/node_ids.go
+++ b/pkg/datapath/linux/node_ids.go
@@ -80,8 +80,9 @@ func (n *linuxNodeHandler) GetNodeID(nodeIP net.IP) (uint16, bool) {
 }
 
 func (n *linuxNodeHandler) getNodeIDForIP(nodeIP net.IP) (uint16, bool) {
-	localNode := node.GetIPv4()
-	if localNode.Equal(nodeIP) {
+	localNodeV4 := node.GetIPv4()
+	localNodeV6 := node.GetIPv6()
+	if localNodeV4.Equal(nodeIP) || localNodeV6.Equal(nodeIP) {
 		return 0, true
 	}
 
--- a/pkg/datapath/linux/node_ids.go
+++ b/pkg/datapath/linux/node_ids.go
@@ -33,16 +33,11 @@ func (n *linuxNodeHandler) AllocateNodeID(nodeIP net.IP) uint16 {
 		return 0
 	}
 
-	// Don't allocate a node ID for the local node.
-	localNode := node.GetIPv4()
-	if localNode.Equal(nodeIP) {
-		return 0
-	}
-
 	n.mutex.Lock()
 	defer n.mutex.Unlock()
 
-	if nodeID, exists := n.nodeIDsByIPs[nodeIP.String()]; exists {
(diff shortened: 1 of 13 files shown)


pkg/datapath/linux/node_ids.go
```diff
@@ -77,8 +72,28 @@ func (n *linuxNodeHandler) GetNodeIP(nodeID uint16) string {
 	return n.nodeIPsByIDs[nodeID]
 }
 
+func (n *linuxNodeHandler) GetNodeID(nodeIP net.IP) (uint16, bool) {
```
mhofstetter, line 75: do we want to go with an `error` instead of the `bool` in the "public" facing API? my thoughts were that a `bool` fulfills the purpose.

pkg/wireguard/agent/agent_test.go
```diff
@@ -82,7 +83,7 @@ func containsIP(allowedIPs []net.IPNet, ipnet *net.IPNet) bool {
 func newTestAgent(ctx context.Context) (*Agent, *ipcache.IPCache) {
 	ipCache := ipcache.NewIPCache(&ipcache.Configuration{
 		Context:       ctx,
-		NodeIDHandler: &mockNodeHandler{},
```
mhofstetter, line 85: idea: re-using the "duplicate" from ipcache kind of reduces the blast radius (e.g. wireguard) when changing the nodeidhandler interface

mhofstetter:
/test-1.26-net-next

pkg/datapath/linux/node_ids.go
meyskens: I agree a bool will be fine as it mimics the map "interface" just fine

meyskens (approved):
LGTM for me

mhofstetter:
/test-1.26-net-next

mhofstetter:
rebased to `main` after jenkins based checks have been removed there

mhofstetter:
replaced `Mutex` with `RWMutex` and only use read-lock in case of `GetNodeIP` & `GetNodeID`

mhofstetter:
Added support for IPv6 in `getNodeIDForIP` - as discussed offline with @brb

brb (approved):

mhofstetter:
hit some [strange error](https://github.com/cilium/cilium/actions/runs/5276632153/jobs/9543977453) in `integration-test` (postcheck) 

rebasing to `main` might help :pray:

jrajahalme (approved):
LGTM

mhofstetter:
removed review-request from @aditighag. @jrajahalme already reviewed in behalf of team _ipcache_.

pkg/datapath/linux/node_ids.go
```diff
@@ -25,24 +25,19 @@ const (
 
 // AllocateNodeID allocates a new ID for the given node (by IP) if one wasn't
 // already assigned.
+// This should never happen. If it ever does, we may have an unexpected
+// call to AllocateNodeID.
+// Don't allocate a node ID for the local node.
 func (n *linuxNodeHandler) AllocateNodeID(nodeIP net.IP) uint16 {
 	if len(nodeIP) == 0 || nodeIP.IsUnspecified() {
-		// This should never happen. If it ever does, we may have an unexpected
```
jibi: should this comment stay here?

pkg/datapath/linux/node_ids.go
```diff
@@ -77,8 +72,28 @@ func (n *linuxNodeHandler) GetNodeIP(nodeID uint16) string {
...
+	n.mutex.Lock()
+	defer n.mutex.Unlock()
+
+	return n.getNodeIDForIP(nodeIP)
+}
+
+func (n *linuxNodeHandler) getNodeIDForIP(nodeIP net.IP) (uint16, bool) {
+	localNode := node.GetIPv4()
+	if localNode.Equal(nodeIP) {
+		return 0, true
```
jibi, line 86: why are we returning true if we are not suppose to allocate a node ID for the local node?

pkg/datapath/linux/node_ids.go
mhofstetter: @jibi good catch - yes it should stay here - thanks. seems to be reordered when moving this check (&warning) back to `AllocateNodeID` because it doesn't make sense to have this in `getNodeIDForIP`

pkg/datapath/linux/node_ids.go
mhofstetter: @jibi `0` is still the node id for the local node (and therefore has its meaning for users of this (e.g. whether not to allocate in `AllocateNodeID` or not just getting the info that its the local node itself via `GetNodeID` ).

pkg/datapath/linux/node_ids.go
mhofstetter: moved the comments back to its original place

mhofstetter:
addressed @jibi's input

jibi (approved):

PR #26159, opened 2023-06-13
Improve Makefile to ease debugging

pippolo84:
Add the ability to override variables in Makefile for debugging purposes. This is done including (when available) an additional `Makefile.override` file.
Also, refactor the operator Dockerfile to remove fixed paths and rely on the ones defined in `Makefile.defs`, just like we do for the agent Dockerfile.


tklauser (approved):

pippolo84:
Runtime test (privileged) failure tracked in https://github.com/cilium/cilium/issues/26233, rerunning

PR #26162, opened 2023-06-13
Remove custom iproute2 fork

### LINKED ISSUE #24579: Remove iproute2 fork
Cilium currently relies on patches on top of iproute2 to be able to load ELFs. We want to move away from a patched iproute2 so simplify our container build processes and reduce reliance on iproute2 for bpf-related operations in general.

Only a few things stand in the way of removing the fork:
- The verifier tests still use iproute2, current proposal: https://github.com/cilium/cilium/pull/24538
- Tunnel device setup is done using `bpf_load` in init.sh, documented here: https://github.com/cilium/cilium/issues/20699

Additionally, we may want to create a small CLI tool or add a command to the agent binary that invokes the loader code for quickly testing ELF loads during Cilium development.

ti-mo:
As of Cilium 1.14, it no longer uses the ip command to load BPF programs into the kernel. This means we no longer need to maintain our patches on top of iproute2 and we no longer depend on a custom build in order for Cilium to run.

Remove cilium-iproute2 references from the documentation.

Also remove the contrib/ansible directory that was partly dedicated to building and installing this custom iproute2.

Fixes: https://github.com/cilium/cilium/issues/24579.

```release-note
Remove custom iproute2 fork
```


zacharysarah (approved):
LGTM

ti-mo:
Continuing this in https://github.com/cilium/cilium/pull/26221 to allow automation to do its thing.

PR #26172, opened 2023-06-13
cilium: IPv4 BIG TCP support

borkmann:
IPv4 BIG TCP support for Cilium on 6.3+ kernels.

Link: https://lore.kernel.org/netdev/cover.1674921359.git.lucien.xin@gmail.com/


qmonnet:
Given that the last commit is also in https://github.com/cilium/cilium/pull/26205, and seems to be causing some CI issues in there, should we leave it out of this one?

borkmann:
> Given that the last commit is also in #26205, and seems to be causing some CI issues in there, should we leave it out of this one?

That actually went through fine, see the l4lb where big tcp is part of. Looks like some failures are from master branch (e.g. Cilium Runtime).

borkmann:
> > Given that the last commit is also in #26205, and seems to be causing some CI issues in there, should we leave it out of this one?
> 
> That actually went through fine, see the l4lb where big tcp is part of.

Turns out https://github.com/vishvananda/netlink/commit/acdc658b8613655ddb69f978e9fb4cf413e2b830 causes a regression in Cilium. I've reverted that specific commit in our local copy for now.

api/v1/openapi.yaml
```diff
@@ -2445,10 +2445,16 @@ definitions:
         additionalProperties:
           type: object
       GSOMaxSize:
-        description: Maximum GSO size on workload facing devices
+        description: Maximum IPv6 GSO size on workload facing devices
```
jrfastab, line 2451: I think I would have made GSOMaxSize the ipv4 defult but hey doesn't matter in the end.

jrfastab: ah but, I guess that would break existing stuff so ok

borkmann: Agree, I extended this here as users might have kernels with IPv6 BIG TCP but not yet IPv4 BIG TCP (for that the existing {GRO,GSO}MaxSize is set for IPv6 and the {GRO,GSO}IPv4MaxSize is 0). The other thought  that would be good in future is to have v4/v6 specific tunables for Cilium.

sayboras (approved):
LGTM for vendor changes 💯

jrfastab (approved):

dylandreimerink (approved):
LGTM!

tklauser (approved):
LGTM for my code owners. Thanks for nicely splitting it up into easy-to-review and well-documented commits.

borkmann:
(CI all green, pushing small improvement)

qmonnet (approved):
Looks all good to me, nice work!

PR #26175, opened 2023-06-13
ci: Add workflow for testing multi-pool IPAM

gandro:
This adds a new GitHub Actions workflow (with `pull_request` trigger) which tests the new multi-pool IPAM feature added in #22762. The feature does have unit test coverage as well, but since it does have implications on routing and masquerading (see `Notes:` section in workflow), we also want to test the feature end-to-end to ensure it interacts correctly with those systems.

The workflow sets up a regular kind cluster and installs Cilium with Multi-Pool IPAM enabled in it. It then runs a version of the connectivity test that allocates the test pod IPs from non-default pools. This tests that routing and masquerading works for non-default pools. The workflow then validates that the pod IPs have been allocated from the correct pool, to ensure we don't report false positives in case the connectivity test annotation did not take effect.

At the moment, this workflow runs on every PR. It is an open question if it should run on a comment trigger instead. During testing, it only had a run-time of about 15 minutes, 5 minutes of which were spent waiting on the images to be ready.

 - :heavy_check_mark: [**Click here to see recent runs of this workflow on this PR**](https://github.com/cilium/cilium/actions/workflows/conformance-multi-pool.yaml?query=branch%3Apr%2Fgandro%2Fmulti-pool-ci-workflow). The older ones are red as I was still writing and testing the workflow, but the more recent ones should all be green.

### DIFF
--- a/.github/workflows/conformance-multi-pool.yaml
+++ b/.github/workflows/conformance-multi-pool.yaml
@@ -1,0 +1,196 @@
+name: ConformanceMultiPoolIPAM
+
+# Any change in triggers needs to be reflected in the concurrency group.
+on:
+  pull_request:
+    paths-ignore:
+      - 'Documentation/**'
+      - 'test/**'
+  push:
+    branches:
+      - main
+      - ft/main/**
+    paths-ignore:
+      - 'Documentation/**'
+      - 'test/**'
+
+permissions: read-all
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.after }}
+  cancel-in-progress: true
+
+env:
+  # renovate: datasource=github-releases depName=cilium/cilium-cli
+  cilium_cli_version: v0.14.7
+  cilium_cli_ci_version:
+  # renovate: datasource=github-releases depName=kubernetes-sigs/kind
+  kind_version: v0.19.0
+  kind_config: .github/kind-config.yaml
+  timeout: 5m
+
+jobs:
+  multi-pool-ipam-conformance-test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+    steps:
+      - name: Checkout main branch to access local actions
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          persist-credentials: false
+      - name: Set Environment Variables
+        uses: ./.github/actions/set-env-variables
+
+      - name: Install Cilium CLI
+        uses: cilium/cilium-cli@9fcfef089e5b7dd3212f2eac21ba8cfae6f05cca # v0.14.7
+        with:
+          release-version: ${{ env.cilium_cli_version }}
+          ci-version: ${{ env.cilium_cli_ci_version }}
+
+      - name: Set image tag
+        id: vars
+        run: |
+          if [ ${{ github.event.pull_request }} ]; then
+            SHA=${{ github.event.pull_request.head.sha }}
+          else
+            SHA=${{ github.sha }}
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/conformance-multi-pool.yaml
```diff
@@ -0,0 +1,195 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.after }}
+  cancel-in-progress: true
+
+env:
+  # renovate: datasource=github-releases depName=cilium/cilium-cli
+  cilium_cli_version: v0.14.7
+  cilium_cli_ci_version:
+  kind_version: v0.17.0
```
tklauser: Any reason not to use the latest version? 
FWIW, renovate will take care of updating these in the future cf. https://github.com/cilium/cilium/pull/26312, so would be nice to add the annotation as well:

```suggestion
  # renovate: datasource=github-releases depName=kubernetes-sigs/kind
  kind_version: v0.20.0
```

.github/workflows/conformance-multi-pool.yaml
```diff
@@ -0,0 +1,195 @@
...
+          #  - Multi-pool IPAM only supports endpoint routes, thus we disable
+          #    the local-node-route.
+          #  - helm/kind-action does not support BPF host routing, so we fall
+          #    back on legacy host routing (#23283)
+          #  - iptables-based masquerading does not support multiple non-masquerade
+          #    CIDRs. Thus, we enable BPF masquerading where we can add multiple
+          #    non-masquerade CIDRs.
+          #  - TODO: We currently cannot use ip-masq-agent, because it is broken
+          #    on main (#26262). For now, we use a single native-routing-cidr in
+          #    10.0.0.0/8 subnet and have all pools be part of that subnet.
```
tklauser, line 73: Thanks a lot for adding these explanatory notes!

tklauser (approved):
Very nice! LGTM apart from a small comment on kind versioning.

.github/workflows/conformance-multi-pool.yaml
gandro: I wanted to use the same kind version as the other workflows currently in use. I have not yet tested newer versions in CI, but with `kind.sh` I had issues with newer versions of kind that I have not yet triaged (iirc they were dual-stack related).

In general I consider the kind upgrade to change unrelated to this PR.

gandro: Having said that, I'll make sure to add `# renovate: datasource=github-releases depName=kubernetes-sigs/kind` and check if v0.19 works (since we are using it in other workflows too)

tklauser: > In general I consider the kind upgrade to change unrelated to this PR.

Fully agree. Let's leave it at v0.17.0 and add the renovate annotation and then let's figure out the upgrade on the respective bot PR.

gandro:
Pushed the suggested `kind` annotation and rebased on main

```diff
diff --git a/.github/workflows/conformance-multi-pool.yaml b/.github/workflows/conformance-multi-pool.yaml
index 1ca629d72f64..5f836788a510 100644
--- a/.github/workflows/conformance-multi-pool.yaml
+++ b/.github/workflows/conformance-multi-pool.yaml
@@ -24,7 +24,8 @@ env:
   # renovate: datasource=github-releases depName=cilium/cilium-cli
   cilium_cli_version: v0.14.7
   cilium_cli_ci_version:
-  kind_version: v0.17.0
+  # renovate: datasource=github-releases depName=kubernetes-sigs/kind
+  kind_version: v0.19.0
   kind_config: .github/kind-config.yaml
   timeout: 5m
```

nbusseneau (approved):

gandro:
Reviews are in, the CI workflow added here is green (see PR description) and no other tests are affected by this. Marking ready-to-merge

PR #26177, opened 2023-06-13
bpf: xdp: use CT tuple hash for tunnel encap's source port

julianwiedmann:
(follow-on to https://github.com/cilium/cilium/pull/24422)

`hash_from_tuple_v*()` isn't perfect (as it ignores the .daddr), but much better than what we currently have.

The one path where we might notice the reduced entropy is in a config with EgressGW and native-routing, when processing EgressGW reply traffic with non-service protocol (ICMP or other exotic types). As here we currently don't extract any L4 ports either.

### DIFF
--- a/bpf/lib/encap.h
+++ b/bpf/lib/encap.h
@@ -212,21 +212,25 @@ encap_and_redirect_netdev(struct __ctx_buff *ctx, struct tunnel_key *k,
 }
 #endif /* TUNNEL_MODE || ENABLE_HIGH_SCALE_IPCACHE */
 
-static __always_inline __be16 tunnel_gen_src_port_v4(void)
+static __always_inline __be16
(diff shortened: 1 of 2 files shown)


julianwiedmann:
Dropped the manual update of the CT tuple after RevDNAT. `REV_NAT_F_TUPLE_SADDR` should already do that for us :amaze:.

ldelossa (approved):
LGTM.

joestringer:
> The one path where we might notice the reduced entropy is in a config with EgressGW and native-routing, when processing EgressGW reply traffic with non-service protocol (ICMP or other exotic types). As here we currently don't extract any L4 ports either.

Does this need further discussion before we merge? cc @jibi

jibi (approved):

PR #26179, opened 2023-06-13
daemon: Do not require native routing CIDR if ipmasq-agent is enabled

gandro:
Cilium's built-in [ipmasq-agent replacement](https://docs.cilium.io/en/v1.13/network/concepts/masquerading/#ebpf-based) acts as a replacement for the native routing CIDR. Therefore, it does not make sense to require the native routing CIDR if the ipmasq-agent is enabled, since the two flags are basically mutually exclusive.

In addition, this commit also slightly restructures and aligns the if condition with the error message, to make it a bit easier to read.

Fixes: e7d4f5c6af7d ("daemon: validate IPv4NativeRoutingCIDR value in DaemonConfig")

### DIFF
--- a/pkg/option/config.go
+++ b/pkg/option/config.go
@@ -3638,26 +3638,36 @@ func (c *DaemonConfig) checkMapSizeLimits() error {
 }
 
 func (c *DaemonConfig) checkIPv4NativeRoutingCIDR() error {
-	if c.GetIPv4NativeRoutingCIDR() == nil && c.EnableIPv4Masquerade && !c.TunnelingEnabled() &&
-		c.IPAMMode() != ipamOption.IPAMENI && c.EnableIPv4 && c.IPAMMode() != ipamOption.IPAMAlibabaCloud {
+	if c.GetIPv4NativeRoutingCIDR() == nil &&
+		c.EnableIPv4 && c.EnableIPv4Masquerade &&
+		!c.EnableIPMasqAgent &&
+		!c.TunnelingEnabled() &&
+		c.IPAMMode() != ipamOption.IPAMENI && c.IPAMMode() != ipamOption.IPAMAlibabaCloud {
 		return fmt.Errorf(
 			"native routing cidr must be configured with option --%s "+
-				"in combination with --%s --%s=%s --%s=%s --%s=true",
-			IPv4NativeRoutingCIDR, EnableIPv4Masquerade, RoutingMode, RoutingModeNative,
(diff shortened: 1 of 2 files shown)


pkg/option/config.go
```diff
@@ -3638,26 +3638,36 @@ func (c *DaemonConfig) checkMapSizeLimits() error {
 }
 
 func (c *DaemonConfig) checkIPv4NativeRoutingCIDR() error {
-	if c.GetIPv4NativeRoutingCIDR() == nil && c.EnableIPv4Masquerade && !c.TunnelingEnabled() &&
-		c.IPAMMode() != ipamOption.IPAMENI && c.EnableIPv4 && c.IPAMMode() != ipamOption.IPAMAlibabaCloud {
+	if c.GetIPv4NativeRoutingCIDR() == nil &&
+		c.EnableIPv4 && c.EnableIPv4Masquerade &&
```
qmonnet, line 3642: Nit: We could remove `c.EnableIPv4` here, given that `c.EnableIPv4Masquerade` already depends on it (same below for IPv6), see earlier in the file:

```go
     c.EnableIPv4Masquerade = vp.GetBool(EnableIPv4Masquerade) && c.EnableIPv4
     c.EnableIPv6Masquerade = vp.GetBool(EnableIPv6Masquerade) && c.EnableIPv6
```

qmonnet (approved):

pkg/option/config.go
gandro: Good point, I didn't realize that. However, since we do print `EnableIPv{4,6}` in the error message, I think I'll leave it as is. That way, the if condition and the error message stay in sync.

qmonnet: Sounds good to me :+1:

pippolo84 (approved):
LGTM :100:

gandro:
CI is failing with `BPF masquerade is not supported for IPv6.` :thinking:

gandro:
Moving this back to draft, because now ipmasq-agent will fall back on the pod CIDR if no native-routing-cidr is provided, which in my opinion is not the correct thing to do, as this also lead to bugs in regular iptables masquerading mode (e.g. in IPAM modes that don't have pod CIDR).

My plan is to disable the fallback on the pod CIDR if no native-routing-cidr is provided. This is fully backwards compatible.

gandro:
Closing in favor of https://github.com/cilium/cilium/pull/27747 - which allows me to test changes to the workflow because it's not running from fork.

PR #26180, opened 2023-06-13
docs: clarify that L3 DNS policies require L7 proxy enabled

wedaly:
Add a note to the L3 policy documentation clarifying that L3 DNS policies require the L7 proxy enabled and an L7 policy for DNS traffic so Cilium can intercept DNS responses.

Previously, the documentation linked to other sections describing the DNS Proxy, but I know at least a few people who were surprised that a policy under "L3 Examples" would require an L7 proxy. Hopefully adding a note near the beginning of the section will make this requirement more obvious.

### DIFF
--- a/Documentation/security/policy/language.rst
+++ b/Documentation/security/policy/language.rst
@@ -518,13 +518,18 @@ provided in DNS responses are allowed by Cilium in a similar manner to IPs in
 or are not know a priori, or when DNS is more convenient. To enforce policy on
 DNS requests themselves, see `Layer 7 Examples`_.
 
-IP information is captured from DNS responses per-Endpoint via a `DNS Proxy`_.
+.. note::
+
+	In order to associate domain names with IP addresses, Cilium intercepts
+	DNS responses per-Endpoint using a `DNS Proxy`_. This requires Cilium
+	to be configured with ``--enable-l7-proxy=true`` and an L7 policy allowing
+	DNS requests. For more details, see :ref:`DNS Obtaining Data`.
+
 An L3 `CIDR based`_ rule is generated for every ``toFQDNs``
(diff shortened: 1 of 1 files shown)


Documentation/security/policy/language.rst
```diff
@@ -518,13 +518,16 @@ provided in DNS responses are allowed by Cilium in a similar manner to IPs in
 or are not know a priori, or when DNS is more convenient. To enforce policy on
 DNS requests themselves, see `Layer 7 Examples`_.
 
-IP information is captured from DNS responses per-Endpoint via a `DNS Proxy`_.
+.. note:: In order to associate domain names with IP addresses, Cilium intercepts
+	DNS responses per-Endpoint using a `DNS Proxy`_. This requires Cilium
+	to be configured with ``--enable-l7-proxy=true`` and an L7 policy allowing
+	DNS requests. For more details, see :ref:`DNS Obtaining Data`.
```
zacharysarah: Formatting nit
```suggestion
.. note::

    In order to associate domain names with IP addresses, Cilium intercepts
    DNS responses per-Endpoint using a `DNS Proxy`_. This requires Cilium
    to be configured with ``--enable-l7-proxy=true`` and an L7 policy allowing
    DNS requests. For more details, see :ref:`DNS Obtaining Data`.
```

zacharysarah (approved):
@wedaly One formatting nit, otherwise LGTM. Approving to unblock with the understanding that changes are required prior to merge.

Documentation/security/policy/language.rst
wedaly: thanks! updated

ti-mo:
Thanks!

PR #26181, opened 2023-06-13
k8s: Use Resource[*Pod] in pod watcher for the local pod watching

joamaki:
Now that Resource[*Pod] has been added as a shared resource and it is seeing use from multiple modules, refactor the pod watcher to also use it for the normal case of watching just the local pods.

The "handover" case of watching all pods still creates a separate informer, but this should affect very small set of deployments.

### DIFF
--- a/pkg/k8s/watchers/pod.go
+++ b/pkg/k8s/watchers/pod.go
@@ -11,6 +11,7 @@ import (
 	"net"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


pkg/k8s/watchers/pod.go
```diff
@@ -931,7 +970,6 @@ func (k *K8sWatcher) deletePodHostData(pod *slim_corev1.Pod) (bool, error) {
 //   - false: returns any pod in the cluster received by the pod watcher.
 func (k *K8sWatcher) GetCachedPod(namespace, name string) (*slim_corev1.Pod, error) {
 	<-k.controllersStarted
-	k.WaitForCacheSync(resources.K8sAPIGroupPodV1Core)
```
bimmlerd: need to put this back (for all pods watcher case this matters)

bimmlerd:
Labelled this a `kind/bug` as duplicate informers on the pod resource is not great - I don't think the release-blocker label would be approriate though, as it only occurs if IPAM multi-pool is used which is beta.

aanm (approved):

PR #26184, opened 2023-06-13
.github: add dedicated job to wait for images

aanm:
We can have a dedicated job that waits for images and it shouldn't be part of the build-ginkgo-binary workflow.

Suggested-by: Nicolas Busseneau <nicolas@isovalent.com>

### DIFF
--- a/.github/workflows/conformance-ginkgo.yaml
+++ b/.github/workflows/conformance-ginkgo.yaml
@@ -169,9 +169,6 @@ jobs:
           ref: ${{ needs.check_changes.outputs.sha }}
           persist-credentials: false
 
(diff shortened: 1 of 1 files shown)


nbusseneau (approved):

nbusseneau (changes requested):
We need to do the same for lines 310-317 https://github.com/cilium/cilium/pull/26184/files#diff-7b1c1cd100727d47983442d5c5860d86fa34ebac932ca2380db80cf369915cc6R310-R317

nbusseneau (approved):

brlbil (approved):

PR #26186, opened 2023-06-13
gha: enable debug logs in conformance-clustermesh workflows

giorio94:
Let's enable debug-level logs in conformance-clustermesh workflows, as they might be useful when investigating possible flakes to gather more information.

Link to successful run: https://github.com/cilium/cilium/actions/runs/5256914992/jobs/9498930808?pr=26186


brlbil (approved):
My only concern was would this change affect test execution time.
The test run took 18m18s, well within the average run time.

giorio94:
> My only concern was would this change affect test execution time. The test run took 18m18s, well within the average run time.

I don't think it should affect the execution time, as it only changes the verbosity of the cilium logs, which are retrieved only in case of failures.

giorio94:
Marking as ready to merge, as reviews are in, and this is a gha change only.

PR #25029, opened 2023-04-21
workflows: add the kind-based clustermesh conformance test for stable branches

giorio94:
This PR adds a copy of the new clustermesh conformance test initially introduced in https://github.com/cilium/cilium/pull/23496 for each of the currently active stable branches (i.e., v1.11, v1.12 and v1.13). Trigger phrases and Kubernetes
versions are updated accordingly.

Fixes: #issue-number

```release-note
workflows: add the kind-based clustermesh conformance test for stable branches
```

PR #24996, opened 2023-04-20
Drop the GKE-based multicluster GitHub actions workflow in favor of the kind-based one

giorio94:
The new kind-based clustermesh workflow introduced in #23496 has now run for a while in parallel with the old one, and it seems comparably stable while covering more scenarios. Hence, let's drop the old one.

Related: https://github.com/cilium/cilium/issues/23322

```release-note
Drop the GKE-based multicluster GitHub actions workflow in favor of the kind-based one
```

PR #26188, opened 2023-06-13
gha: drop the GKE-based multicluster workflow from stable branches

giorio94:
The new kind-based clustermesh workflows ported to all stable branches in #25029 have now run for a while in parallel with the old ones, and they seems comparably stable while covering more scenarios. Hence, let's drop the old ones, as we already did in #24996 for main.

Fixes: #issue-number

```release-note
Drop the GKE-based multicluster GitHub actions workflow in favor of the kind-based ones (stable branches)
```


nbusseneau (approved):

giorio94:
Reviews are in, marking as ready to merge as gha changes only.

PR #26201, opened 2023-06-14
chore: pkg imported more than once

testwill:
packages is being imported more than once (ST1019)

### DIFF
--- a/pkg/bgpv1/test/adverts_test.go
+++ b/pkg/bgpv1/test/adverts_test.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	corev1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/core/v1"
-	slim_metav1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/apis/meta/v1"
 	slimv1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/apis/meta/v1"
 	"github.com/cilium/cilium/pkg/node"
 	"github.com/cilium/cilium/pkg/testutils"
@@ -330,8 +329,8 @@ func Test_LBEgressAdvertisement(t *testing.T) {
 	require.NoError(t, err)
 
 	// setup bgp policy with service selection
-	fixture.config.policy.Spec.VirtualRouters[0].ServiceSelector = &slim_metav1.LabelSelector{
-		MatchExpressions: []slim_metav1.LabelSelectorRequirement{
+	fixture.config.policy.Spec.VirtualRouters[0].ServiceSelector = &slimv1.LabelSelector{
+		MatchExpressions: []slimv1.LabelSelectorRequirement{
 			// always true match
 			{
 				Key:      "somekey",
--- a/pkg/hubble/parser/threefour/parser.go
+++ b/pkg/hubble/parser/threefour/parser.go
@@ -13,7 +13,6 @@ import (
 	"github.com/sirupsen/logrus"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
-	"github.com/cilium/cilium/api/v1/flow"
 	pb "github.com/cilium/cilium/api/v1/flow"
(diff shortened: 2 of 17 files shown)


dylandreimerink (approved):

pkg/bgpv1/test/adverts_test.go
```diff
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	corev1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/core/v1"
-	slim_metav1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/apis/meta/v1"
 	slimv1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/apis/meta/v1"
```
tklauser, line 12: I'd keep `slim_metav1` here and drop `slimv1` instead, as the name is more specific.

tklauser (changes requested):
I assume you found these using some linter (probably `staticcheck`?). Could we also enable that check in our CI so we catch these types of issues in the future?

giorio94 (approved):

pkg/hubble/parser/threefour/parser.go
```diff
@@ -15,7 +15,6 @@ import (
 	"github.com/sirupsen/logrus"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
-	"github.com/cilium/cilium/api/v1/flow"
 	pb "github.com/cilium/cilium/api/v1/flow"
```
youngnick, line 16: I think it's better to keep the unaliased version here maybe? `pb` is a pretty generic name that doesn't give much clue about what the package is actually doing.

youngnick (commented):
One extra thing from me, but I agree that we should add whatever linter was used here to the standard set of checks.

ti-mo:
I need to echo the others, this PR should come with a linter so this gets caught in the future.

ldelossa:
@testwill if you are still interested in merging this PR, can you please rebase and push back up?

viktor-kurchenko (approved):
LGTM.

gandro (approved):
The changes in my code owners look good to me. But the commit message needs to be changed.

Approving for just my code owners part, defaulting Tobias for questions around CI.

nathanjsweet (approved):

tommyp1ckles:
@testwill thanks for the PR,  the changes are pretty straightforward but could you add a more accurate description of the changes in the commit message?

christarazi (approved):

christarazi:
@testwill Please update the commit msg to have a more description title and msg. The title of the PR works just fine.

pkg/hubble/parser/threefour/parser.go
tommyp1ckles: Unresolving this as it appears to still be using "pb", I agree with nick - pb might get confusing given there are multiple protobuffer libraries.

tommyp1ckles:
@testwill 

Looks good, just waiting for one point that was missed regarding the flow pb library.

squeed:
@testwill needs rebase.

ldelossa:
@testwill can you please rebase this PR if you still intend of merging?

ti-mo:
@testwill Sorry, this code has a bit of churn, you'll need to rebase again.

pkg/bgpv1/test/adverts_test.go
joestringer: This wasn't yet addressed, but it was marked as resolved.

joestringer:
I think that in principle this has been approved by most codeowners, there's just a couple of specific comments where the other import is preferred since it's more descriptive. Looks like it needs a rebase too.

I agree that adding a linter would be a good step, though I don't see it as a strict requirement to merge this PR. I've filed https://github.com/cilium/cilium/issues/27676 to follow up on that request.

I’ve marked this PR as draft while you work through the feedback in the PR above. When you’re ready to continue with review & testing of the PR, please click “Ready for Review” at the bottom of the page.

PR #26205, opened 2023-06-14
cilium: enable bpf host routing with per endpoint routes for IPv6 as well

borkmann:
See commit descriptions.

### DIFF
--- a/daemon/cmd/kube_proxy_replacement.go
+++ b/daemon/cmd/kube_proxy_replacement.go
@@ -415,9 +415,6 @@ func finishKubeProxyReplacementInit() error {
 		// KPR=strict is needed or we might rely on netfilter.
 		case option.Config.KubeProxyReplacement != option.KubeProxyReplacementStrict:
 			msg = fmt.Sprintf("BPF host routing requires %s=%s.", option.KubeProxyReplacement, option.KubeProxyReplacementStrict)
-		// All cases below still need to be implemented ...
-		case option.Config.EnableEndpointRoutes && option.Config.EnableIPv6:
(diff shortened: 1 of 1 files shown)


julianwiedmann:
For the ep-routes change, the original patch said

> In case of endpoint routes & host routing the IPv6 breaks, so enable host
> routing only for IPv4 case.  **One reason** is that the host can't populate
> neighbour entries for endpoints.  This happens because in case of endpoint
> routes ICMPv6 messages originate on [host] lxc veth devices. Those devices are
> only configured with IPv6 link-local addresses, so the ICMPv6 answer from the
> endpoint is sent to a link-local address which is recognized "world" by cilium,
> but fib_redirect can't redirect it (as it should be delivered to stack, not
> forwarded).

Given this lists just **one** reason - have all the other reasons been resolved by now? :)

borkmann:
> For the ep-routes change, the original patch said
> 
> > In case of endpoint routes & host routing the IPv6 breaks, so enable host
> > routing only for IPv4 case.  **One reason** is that the host can't populate
> > neighbour entries for endpoints.  This happens because in case of endpoint
> > routes ICMPv6 messages originate on [host] lxc veth devices. Those devices are
> > only configured with IPv6 link-local addresses, so the ICMPv6 answer from the
> > endpoint is sent to a link-local address which is recognized "world" by cilium,
> > but fib_redirect can't redirect it (as it should be delivered to stack, not
> > forwarded).
> 
> Given this lists just **one** reason - have all the other reasons been resolved by now? :)

Good reminder, I'll check this out if its still the case and brainstorm what we could do about it.

zacharysarah (approved):
LGTM

qmonnet (approved):
Looks good, thank you!

aspsk (approved):

christarazi (approved):

PR #26167, opened 2023-06-13
helm: Correct the flag names in validate.yaml

sayboras:
The attributes in IF condition statements are updated in the below comment, but the error description was not. This commit is to perform simple corrections.

Fixes: e05bc80477bbf18f79909537dbfd0bb502a29bd7

PR #26208, opened 2023-06-14
auth: feature flag for authentication

mhofstetter:
Up to this point, only the SPIRE installation & integration were put behind a feature flag. Authentication, re-authentication & garbage collection were always active.

This PR introduces an additional feature flag to be able to disable the auth module if necessary. It is enabled by default.

Cilium config: `auth-mesh-enabled`
Helm value: `authentication.enabled`

Note that if disabled, policy enforcement will still block requests that require authentication. But the resulting authentication requests for these requests will not be processed, therefore the requests not be allowed.

### DIFF
--- a/install/kubernetes/cilium/templates/validate.yaml
+++ b/install/kubernetes/cilium/templates/validate.yaml
@@ -66,6 +66,12 @@
   {{- end }}
 {{- end }}
 
+{{- if .Values.authentication.mutual.spire.enabled }}
+  {{- if not .Values.authentication.enabled }}
+    {{ fail "SPIRE integration requires .Values.authentication.enabled=true and .Values.authentication.mutual.spire.enabled=true" }}
+  {{- end }}
+{{- end }}
+
 {{- if .Values.authentication.mutual.spire.install.enabled }}
   {{- if not .Values.authentication.mutual.spire.enabled }}
     {{ fail "SPIRE installation requires .Values.auth.mTLS.enabled=true and .Values.auth.mTLS.spire.enabled=true" }}
--- a/install/kubernetes/cilium/values.yaml.tmpl
+++ b/install/kubernetes/cilium/values.yaml.tmpl
@@ -2970,6 +2970,10 @@ sctp:
 
 # Configuration for types of authentication for Cilium
 authentication:
+  # -- Enable authentication processing and garbage collection.
+  # Note that if disabled, policy enforcement will still block requests that require authentication.
+  # But the resulting authentication requests for these requests will not be processed, therefore the requests not be allowed.
+  enabled: true
   # -- Buffer size of the channel Cilium uses to receive authentication events from the signal map.
   queueSize: 1024
   # -- Buffer size of the channel Cilium uses to receive certificate expiration events from auth handlers.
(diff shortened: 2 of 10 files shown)


install/kubernetes/cilium/templates/validate.yaml
```diff
@@ -66,9 +66,15 @@
...
+  {{- if not .Values.authentication.enabled }}
+    {{ fail "SPIRE integration requires .Values.authentication.enabled=true" }}
+  {{- end }}
+{{- end }}
+
 {{- if .Values.authentication.mutual.spire.install.enabled }}
-  {{- if not .Values.authentication.mutual.spire.enabled }}
-    {{ fail "SPIRE installation requires .Values.auth.mTLS.enabled=true and .Values.auth.mTLS.spire.enabled=true" }}
+  {{- if or ( not .Values.authentication.enabled ) ( not .Values.authentication.mutual.spire.enabled ) }}
+    {{ fail "SPIRE installation requires .Values.authentication.enabled=true and .Values.authentication.mutual.spire.enabled=true" }}
```
sayboras: this was part of https://github.com/cilium/cilium/pull/26167 as well.

sayboras (approved):
:ship:

install/kubernetes/cilium/templates/validate.yaml
mhofstetter: ah thanks, i guess i'll wait and rebase

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -2964,6 +2964,9 @@ sctp:
 
 # Configuration for types of authentication for Cilium
 authentication:
+  # -- Enable authentication processing and garbage collection.
+  # Note that this doesn't affect auth policy enforcement.
```
youngnick: I think this could use a little more explanation - maybe something about what happens when it's disabled? That is, what stops running, and what won't work?

As written, it sounds like disabling authentication won't affect that auth policy enforcement works properly, which doesn't seem right?

youngnick (commented):
Mostly LGTM with one small change.

install/kubernetes/cilium/templates/validate.yaml
mhofstetter: @sayboras i removed these changes. we should be safe to merge the two PRs without any conflicts :pray:

install/kubernetes/cilium/values.yaml.tmpl
mhofstetter: @youngnick good point. i tried to come up with a better comment. PTAL

install/kubernetes/cilium/templates/validate.yaml
sayboras: it's great idea :+1:

mhofstetter: saw that you're PR has been merged at the time of writing :heavy_check_mark:

sayboras: yeah, I am pretty much trying to push all pending PRs, so that we don't need to rush tmr :D

youngnick (approved):
Thanks for the extra comment, that's excellent.

PR #26211, opened 2023-06-14
bpf: Use "fallthrough;", compile with -Wimplicit-fallthrough

qmonnet:
> **Warning** to backporters:
>
> Only the first commit should be backported.

We can label intentional fall-throughs and tell clang to report unannotated ones, to catch bugs that would be caused by involuntarily falling through between switch labels.

The result is not super aesthetic in builtins.h or jhash.h, but at the same time it's just one keyword so not too much boilerplate either. And it catches bugs. The one fixed in commit 9a641116021a (`"bpf: test: fix pktgen for IPv6 NEXTHDR_DEST option"`), or the first commit in this PR, were found this way.

I wasn't able to pin down the exact version that got support for the relevant attribute, but clang 10 has it for sure and it seems to be much older than that, so we don't have to worry and test for its availability.

Inspired by checkpatch complaining about the `/* fall through */` comments.

### DIFF
--- a/bpf/lib/egress_policies.h
+++ b/bpf/lib/egress_policies.h
@@ -372,6 +372,7 @@ srv6_create_state_entry(struct __ctx_buff *ctx)
 
 		if (map_update_elem(&SRV6_STATE_MAP6, inner_ips, outer_ips, 0) < 0)
 			return DROP_INVALID;
+		break;
 	}
 #  ifdef ENABLE_IPV4
 	case IPPROTO_IPIP: {
@@ -384,6 +385,7 @@ srv6_create_state_entry(struct __ctx_buff *ctx)
 
 		if (map_update_elem(&SRV6_STATE_MAP4, inner_ips, outer_ips, 0) < 0)
 			return DROP_INVALID;
+		break;
 	}
 #  endif /* ENABLE_IPV4 */
 	}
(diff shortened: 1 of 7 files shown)


bpf/lib/egress_policies.h
```diff
@@ -371,6 +371,7 @@ srv6_create_state_entry(struct __ctx_buff *ctx)
 
 		if (map_update_elem(&SRV6_STATE_MAP6, inner_ips, outer_ips, 0) < 0)
 			return DROP_INVALID;
+		break;
```
lmb, line 375: These seem like genuine mistakes, maybe split them into a separate commit with a Fixes tag?

lmb (approved):
Nice!

bpf/lib/egress_policies.h
qmonnet: Absolutely, I meant to do that in the first place :facepalm:. Thanks! Will update.

julianwiedmann (approved):

borkmann (approved):

qmonnet:
Repeating here the warning for backporters that **only the first commit should be backported to 1.13**.

nbusseneau:
Thanks for the warning @qmonnet!

PR #26212, opened 2023-06-14
service unittests: reinit pointers on SetUp

strudelPi:
In pkg/service/service_test.go the unit tests share several pointer `backends` variables. Those are changed by some tests. Some tests don't seem to expect that, some try to solve this by DeepCopy() in the begining (so it doesn't affect others) and setting the values at the start of the unittest.

Constructing new objects with the default constructor values in `SetUpTest` should make sure the unit tests don't affect each other and the *backends values are always in a reliable state.

as @aditighag suggested, this should probably **not get backported**.

      GitHub issue.

```release-note
Improve service unit test robustness
```

### DIFF
--- a/pkg/service/service_test.go
+++ b/pkg/service/service_test.go
@@ -135,7 +135,15 @@ type ManagerTestSuite struct {
 	ipv6                        bool
 }
 
-var _ = Suite(&ManagerTestSuite{})
+var (
+	_           = Suite(&ManagerTestSuite{})
+	surrogateFE = *lb.NewL3n4AddrID(lb.TCP, cmtypes.MustParseAddrCluster("0.0.0.0"), 80, lb.ScopeExternal, 0)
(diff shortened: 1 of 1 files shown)


dylandreimerink (approved):

ti-mo:
@strudelPi Nit: in the future, we'll want to use test helpers to set up fixtures and `t.Cleanup()` to do teardown instead of relying on `{SetUp,TearDown}Test`. With the introduction of `checkmate`, a `check.C` now embeds a `testing.T`, so you could technically use `c.Cleanup()` already in this PR. We'd like to merge this after the 1.14 release branch has been cut.

ldelossa:
@strudelPi, can you rebase this PR and push up? Will fix any pending issue with tests.

strudelPi:
@ti-mo We need something to init the values before the test rather than cleanup after. If the cleanup is the only thing available right now, we would need to have the pointers defined twice or something I guess.

squeed (approved):

PR #26213, opened 2023-06-14
Publish the 2022 Cilium security audits

### LINKED ISSUE #23791: docs: link to security audits from docs
The 2022 Security Audits for Cilium were recently published [here](https://github.com/cilium/cilium.io/blob/main/Security-Reports/CiliumSecurityAudit2022.pdf) and [here](https://github.com/cilium/cilium.io/blob/main/Security-Reports/CiliumFuzzingAudit2022.pdf). We should add links to them from the Cilium documentation. 

I'm not sure where the best location for this would be, but there might be an obviously good place once https://github.com/cilium/cilium/pull/23599 is finalized (wdyt @zacharysarah?)

zacharysarah:
Fixes: #23791 

Because Sphinx is the way that it is, I had to add some logic to force newlines between file downloads. Also because reST gonna reST, I had to leave whitespace before each newline in order to mollify the linter.

Screenshot 2023-06-14 at 1 57 07 AM

The result is pretty OK.

Screenshot 2023-06-14 at 1 56 30 AM


qmonnet (commented):
> Maybe a naive question, but why not just use a regular list?

~The directives fail to render when they're bulleted.~  Bullets aren't enough to force newlines.

Screenshot 2023-06-14 at 3 08 11 AM

Screenshot 2023-06-14 at 3 08 20 AM

> Also, would it make sense to have a page dedicated to audits?

I thought about this. I'm open to it, but like I think you're saying, nothing feels quite right. 🤔 

UPDATE: Maybe a page banner limited to pages in the `security/` path?

qmonnet:
Note: You can do “quote reply” by clicking on the ellipsis menu of a comment to cite the previous message, which should be easier to follow than editing my own messages :)

> > Maybe a naive question, but why not just use a regular list?
> 
> ~The directives fail to render when they're bulleted.~ Bullets aren't enough to force newlines.

You need a blank line between your paragraph and the list items, otherwise it's all part of the same paragraph:

```rst

    The 2022 security audits for Cilium are available:

    - :download:`Cilium Security Audit 2022 <audits/CiliumSecurityAudit2022.pdf>`
    - :download:`Cilium Fuzzing Audit 2022 <audits/CiliumFuzzingAudit2022.pdf>`
```

> 
> > Also, would it make sense to have a page dedicated to audits?
> 
> I thought about this. I'm open to it, but like I think you're saying, nothing feels quite right. thinking
> 
> UPDATE: Maybe a page banner limited to pages in the `security/` path?

Could work, although I'm not sure it deserves this much attention (and not sure if worth spending time on how to set up the banner for some pages only - although I think Sphinx allows it rather easily). Your call

zacharysarah:
> Note: You can do “quote reply” by clicking on the ellipsis menu of a comment to cite the previous message, which should be easier to follow than editing my own messages :)

Is there any notification advantage? What's the benefit to quoting that specific way?

> You need a blank line between your paragraph and the list items, otherwise it's all part of the same paragraph:

So it does. I'll add a commit for code review.

> Your call

Let's do it this minimal way and observe how much traffic we get.

qmonnet (approved):
[New preview](https://deploy-preview-26213--docs-cilium-io.netlify.app/security/index.html) looks good, thanks!

image

qmonnet:
> > Note: You can do “quote reply” by clicking on the ellipsis menu of a comment to cite the previous message, which should be easier to follow than editing my own messages :)
> 
> Is there any notification advantage? What's the benefit to quoting that specific way? It's not a hardship for me to select and copypaste as opposed to selecting and making GH copypaste. :-)

Oh sure that works, too. No notification advantage. My point was simply that any of those is better than answering inside someone else's post, because the latter makes it harder to follow the conversation.

> 
> > You need a blank line between your paragraph and the list items, otherwise it's all part of the same paragraph:
> 
> So it is. I'll add a commit for code review.

Thanks! Would you mind squashing these three commits together, please?

> 
> > Your call
> 
> Let's do it this minimal way and observe how much traffic we get.

Sounds good to me :)

zacharysarah:
> My point was simply that any of those is better than answering inside someone else's post, because the latter makes it harder to follow the conversation.

I still don't follow. Are you saying you would prefer quoting an entire paragraph rather than only enough of a line to establish reference?

> Thanks! Would you mind squashing these three commits together, please?

Done!

qmonnet:
> > My point was simply that any of those is better than answering inside someone else's post, because the latter makes it harder to follow the conversation.
> 
> I still don't follow. Are you saying you would prefer quoting an entire paragraph instead of only an excerpt?

No. I mean that your did your first reply on this thread not by quoting, but by editing my message directly:

image

and this makes it difficult to follow the conversation. It looks like I wrote everything [in that comment box](https://github.com/cilium/cilium/pull/26213#pullrequestreview-1478953120), when in fact it contains both my comment and your replies. Makes it hard for other readers to follow the discussion (or for me, for that matter :sweat_smile:).

> 
> > Thanks! Would you mind squashing these three commits together, please?
> 
> Done!

Thanks! @lizrice would you mind validating, please?

lizrice (approved):

lizrice:
Looks fine to me, thank you

qmonnet:
Doc change only and passing the relevant workflows, this is ready to merge.

zacharysarah:
> I mean that your did your first reply on this thread not by quoting, but by editing my message directly:

Oh shoot, I'm so sorry about that, @qmonnet! I didn't even catch that I'd done it, so thank you for pointing it out. 😳

qmonnet:
> Oh shoot, I'm so sorry about that, @qmonnet! I didn't even catch that I'd done it, so thank you for pointing it out. :flushed:

I thought it was intentional at first, but figured that after all it had to be an accident :smile: All good, and apologies for the confusing discussion - let's move on! :rocket:

joestringer (approved):
Weird, I totally thought we already published these :-)

Thanks for adding this!

PR #26215, opened 2023-06-14
egressgw: steer traffic to the right interface using BPF

### LINKED ISSUE #23504: Investigate `bpf_fib_lookup` + `bpf_redirect` to replace egress gateway IP rules/routes
When the egress IP is assigned to a different interface than the one with the default route, we'll have to install additional IP rules and routes to steer the traffic to the correct interface.

That's because the network stack routes packets before the TC egress hook (i.e. where bpf SNAT happens), which means can't match traffic on the egress IP. We instead need to explicitly match each combination of source (endpoints) IPs and destination CIDRs.

@pchaigno suggested we might want to look into switching to `bpf_fib_lookup` + `bpf_redirect` after the packet has been SNATed with the egress IP as that would allow to match the source/egress IP without the need for all the special IP rules/routes.

jibi:
see commits

Fixes: https://github.com/cilium/cilium/issues/23504

```release-note
EgressGW: interface selection is now done with BPF, using --install-egress-gateway-routes is no longer needed.
```


bpf/lib/nat.h
```diff
@@ -572,6 +575,12 @@ static __always_inline int snat_v4_rewrite_egress(struct __ctx_buff *ctx,
 	if (csum.offset &&
 	    csum_l4_replace(ctx, off, &csum, 0, sum, flags) < 0)
 		return DROP_CSUM_L4;
+
+#if defined(ENABLE_EGRESS_GATEWAY)
+	if (target && target->egress_gateway)
+		return egress_gw_fib_lookup_and_redirect(ctx, state->to_saddr);
+#endif
+
```
julianwiedmann: Layering-wise, I think we should lift this into `nodeport_snat_fwd_ipv4()`. The `target` should have all the info we need from there (`target.addr == state->to_saddr`, right?)

And `tail_handle_snat_fwd_ipv4()` needs a check to only `send_trace_notify()` for `CTX_ACT_OK`.

jibi: > Layering-wise, I think we should lift this into `nodeport_snat_fwd_ipv4()`. The `target` should have all the info we need from there (`target.addr == state->to_saddr`, right?)

yep, I was a bit worried about that stray `send_drop_notify_ext` in lib/egress_policies.h :hmmyes:
now drop notifications should get handled by the `tail_handle_snat_fwd_ipv4` caller

> And `tail_handle_snat_fwd_ipv4()` needs a check to only `send_trace_notify()` for `CTX_ACT_OK`.

:bulb:

bpf/lib/egress_gateway.h
```diff
@@ -137,5 +139,22 @@ bool egress_gw_reply_needs_redirect(struct iphdr *ip4 __maybe_unused,
...
+
+	*ext_err = (__s8)fib_lookup_v4(ctx, &fib_params, egress_ip, 0, 0);
+
+	if (*ext_err != BPF_FIB_LKUP_RET_SUCCESS && *ext_err != BPF_FIB_LKUP_RET_NO_NEIGH)
+		return DROP_NO_FIB;
+
+	if (old_oif == fib_params.l.ifindex)
+		return CTX_ACT_OK;
+
+	return fib_do_redirect(ctx, true, &fib_params, ext_err, (int *)&old_oif);
```
lmb: Why is this an `int` here?

bpf/lib/nodeport.h
```diff
@@ -3055,7 +3060,8 @@ int tail_handle_snat_fwd_ipv4(struct __ctx_buff *ctx)
 		return send_drop_notify_error_ext(ctx, 0, ret, ext_err,
 						  CTX_ACT_DROP, METRIC_EGRESS);
 
-	send_trace_notify(ctx, obs_point, 0, 0, 0, 0, trace.reason, trace.monitor);
+	if (ret == CTX_ACT_OK)
```
lmb, line 3076: This could use a comment. I'm guessing that nodeport_snat_fwd_ipv4 can now also redirect in addition to just rewriting headers?

pkg/egressgateway/manager_privileged_test.go
```diff
@@ -51,13 +50,13 @@ const (
 	excludedCIDR1   = "1.1.1.22/32"
 	excludedCIDR2   = "1.1.1.240/30"
 
+	zeroIP4 = "0.0.0.0"
```
lmb: Nit: gratuitous change?

pkg/datapath/linux/linux_defaults/linux_defaults.go
```diff
@@ -19,13 +19,6 @@ const (
 	// RouteTableVtep is the default table ID to use for VTEP routing rules
 	RouteTableVtep = 202
 
-	// RouteTableEgressGatewayInterfacesOffset is the offset for the per-ENI
-	// egress gateway routing tables.
-	// Each ENI interface will have its own table starting with this offset. It
-	// is 300 because it is highly unlikely to collide with the main routing
-	// table which is between 253-255. See ip-route(8).
-	RouteTableEgressGatewayInterfacesOffset = 300
```
lmb, line 27: What abut `RulePriorityEgressGateway`?

lmb (commented):
I am very very very happy about this change. I don't know enough about the datapath to know whether the logic is sound though. Do we need someone in addition to Julian to review?

bpf/lib/egress_gateway.h
jibi: `fib_do_redirect()` was introduced in https://github.com/cilium/cilium/pull/26136, so I'll defer that to @ldelossa as I'm not quite sure about the context (especially given that in that function, `oif` is always used as a u32)

bpf/lib/nodeport.h
jibi: yeah it's just in the commit message for now, I'll add a comment also here

pkg/datapath/linux/linux_defaults/linux_defaults.go
jibi: missed it

bpf/lib/egress_gateway.h
julianwiedmann: We've been going back and forth on `u32` and `int`, depending on who is the last to touch the relevant FIB / redirect code :slightly_smiling_face:. I believe it originates from `ctx_redirect()` taking an int.

brb (approved):
Reviewed `bpf/` - nice work! Less moving parts FTW.

BTW, when doing upgrade from v1.14 -> v1.15, shouldn't we remove old IP rules/routes?

jibi:
> BTW, when doing upgrade from v1.14 -> v1.15, shouldn't we remove old IP rules/routes?

good point :bulb: yes, I'll add back some logic to do the cleanup when the manager is initialized :+1:

bpf/lib/egress_gateway.h
```diff
@@ -22,6 +24,23 @@
 #define EGRESS_GATEWAY_NO_GATEWAY (0)
 #define EGRESS_GATEWAY_EXCLUDED_CIDR bpf_htonl(1)
 
+static __always_inline
+int egress_gw_fib_lookup_and_redirect(struct __ctx_buff *ctx, __be32 egress_ip, __s8 *ext_err)
+{
+	struct bpf_fib_lookup_padded fib_params = {};
+	__u32 old_oif = ctx_get_ifindex(ctx);
+
+	*ext_err = (__s8)fib_lookup_v4(ctx, &fib_params, egress_ip, 0, 0);
```
julianwiedmann: Could you explain why it's ok to set `.daddr` to 0 here? I would have expected the actual cluster-external destination.

bpf/lib/nodeport.h
julianwiedmann: Just a nit, but maybe adding at least a comment in the corresponding IPv6 path would be good (explaining why this handling is *not* needed there at the moment).

bpf/bpf_host.c
```diff
@@ -1361,6 +1361,9 @@ int cil_to_netdev(struct __ctx_buff *ctx __maybe_unused)
 #endif
 
 #ifdef ENABLE_HOST_FIREWALL
+	if (ctx_snat_done(ctx))
+		goto skip_host_firewall;
+
```
julianwiedmann, line 1366: The reasoning makes sense, although the scope of the change is a bit wider (it also affects eg. NAT-forwarded LB connections in TC, and RevDNATed replies, ...).

Using the SNAT-done flag for such policy decisions feels a bit fragile. I would hope that we can come up with a better solution over time. But good enough for me atm :+1:.

pkg/egressgateway/manager.go
```diff
@@ -87,6 +87,7 @@ var defaultConfig = Config{
 
 func (def Config) Flags(flags *pflag.FlagSet) {
 	flags.Bool("install-egress-gateway-routes", def.InstallEgressGatewayRoutes, "Install egress gateway IP rules and routes in order to properly steer egress gateway traffic to the correct ENI interface")
+	flags.MarkDeprecated("install-egress-gateway-routes", "This option is deprecated, has no effect, and will be removed in v1.16")
```
julianwiedmann, line 90: I remember at some point we said that the "FIB lookup" logic would only get enabled iff this option is set. To ensure that the egress interface would have bpf_host attached. Did you come up with another solution, or do we just want to take the simple approach first?

julianwiedmann (commented):
Just a few smaller questions for clarification, overall lgtm :rocket:.

bpf/lib/egress_gateway.h
jibi: I thought it was ok to set just the egress/source IP as we want to identify the interface with that address assigned to (and that should be independent from the destination), or do you have a specific case in mind where this would not not work?

pkg/egressgateway/manager.go
jibi: I'm just a bit eager to rip out this code (as otherwise it will stay around for yet another stable release) and start testing the feature with fib lookups always enabled

> To ensure that the egress interface would have bpf_host attached

I thought we concluded that these interfaces will have already bpf_host attached anyway (otherwise it's not possible to reverse translate reply traffic) so that shouldn't be an issue

julianwiedmann: > I thought we concluded that these interfaces will have already bpf_host attached anyway (otherwise it's not possible to reverse translate reply traffic) so that shouldn't be an issue

Ah yes that sounds reasonable (and I probably just don't remember discussing it before :disappointed:). Agree with the sentiment to roll this out ASAP.

bpf/lib/egress_gateway.h
julianwiedmann: Hmm not in combination with how we do EgressGW. We already say that all selected traffic should exit through *one* specific interface.

I'm just a bit wary about having *one* caller in the codebase that doesn't set the `daddr`. That makes it easy to miss bugs / corner cases, because no other path would be exposed to it.

jibi: > I'm just a bit wary about having one caller in the codebase that doesn't set the daddr. That makes it easy to miss bugs / corner cases, because no other path would be exposed to it.

yeah good point, I'll just pass `tuple.daddr` as well :+1:

lmb (approved):

julianwiedmann (approved):

gandro:
Sorry for the drive-by post-merge comment/review. I just noticed this change while I was investigating how the old feature worked (looking into auto-direct-node-routes improvements). Something that confused me for a bit: The documentation still mentions this flag. Since the flag is now a no-op, should the docs be removed too? 

https://docs.cilium.io/en/latest/network/egress-gateway/#eks-s-eni-mode

julianwiedmann:
> Sorry for the drive-by post-merge comment/review. I just noticed this change while I was investigating how the old feature worked (looking into auto-direct-node-routes improvements). Something that confused me for a bit: The documentation still mentions this flag. Since the flag is now a no-op, should the docs be removed too?
> 
> https://docs.cilium.io/en/latest/network/egress-gateway/#eks-s-eni-mode

Yep, we'll still need to go through the whole doc update dance (also mention the deprecation etc) :+1:.

PR #26220, opened 2023-06-14
clustermesh-apiserver: don't wait for the presence of unused CRDs

giorio94:
Currently, the clustermesh-apiserver waits for the presence of all the CRDs before starting the main logic, way more than the resource types actually used afterwards. This ends up preventing the possibility of running a newer version of the clustermesh-apiserver alongside older versions of cilium/cilium-operator due to newly introduced CRDs.

This PR modifies the logic to only wait for the CRDs which are actually used (the new list matches the entries listed in the clustermesh-apiserver cluster role).

```release-note
clustermesh-apiserver: don't wait for the presence of unused CRDs
```

### DIFF
--- a/clustermesh-apiserver/main.go
+++ b/clustermesh-apiserver/main.go
@@ -526,7 +526,7 @@ func startServer(startCtx hive.HookContext, clientset k8sClient.Clientset, backe
 	}).Info("Starting clustermesh-apiserver...")
 
 	if mockFile == "" {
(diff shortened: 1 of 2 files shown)


YutaroHayakawa (approved):

aanm (approved):

giorio94:
Reviews from in, and tests are green. Marking as ready to merge

PR #26229, opened 2023-06-14
[DRAFT] Observable identity allocations

joamaki:


joestringer:
Marked as `kind/bug` and blocker for 1.14 as currently 1.14 does double-pull from k8s for CiliumIdentity objects, which could cause scale problems for users who deploy 1.14.

mhofstetter:
Closed in favor of https://github.com/cilium/cilium/pull/26373 and a follow up PR which will refactor the auth parts.

PR #26230, opened 2023-06-14
docs: remove clustermesh-apiserver gops port from system requirements

giorio94:
clustermesh-apiserver runs in pod network, hence there is no risk of port conflicts with services running on the underlying nodes. Hence, let's remove its gops port from the system requirements table.


gentoo-root (approved):

zacharysarah (approved):
LGTM

PR #26235, opened 2023-06-14
ci: add Ariane configuration file

nbusseneau:
Ariane is a new GitHub App intended to trigger Cilium CI workflows based on trigger phrases commented in pull requests, in order to replace the existing `issue_comment`-based workflows and simplify our CI stack.

This commit adds a temporary configuration setting up triggers such that existing workflows can be triggered with the same trigger phrases, and based on the same PR changelist match / ignore.

The only difference to the final configuration is that all triggers are prefixed with `ariane`, so that we can iterate on the workflows in the PR without interfering with the regular CI until we are ready to merge.

### DIFF
--- a/.github/ariane-config.yaml
+++ b/.github/ariane-config.yaml
@@ -1,0 +1,65 @@
+triggers:
+  /ariane-test:
+    workflows:
+    - conformance-aks.yaml
+    - conformance-aws-cni.yaml
+    - conformance-clustermesh.yaml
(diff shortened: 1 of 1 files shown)


nbusseneau:
This PR does nothing by itself, it's a pre-requisite for #26234. It does not need any CI to run and can be merged immediately.

nbusseneau:
I've found a way to work around this and have everything in the same PR, closing this.

PR #26237, opened 2023-06-14
ci: use Ariane to trigger workflows

nbusseneau:
Ariane is a new GitHub App intended to trigger Cilium CI workflows based on trigger phrases commented in pull requests, in order to replace the existing `issue_comment`-based workflows and simplify our CI stack.

Ariane uses `workflow_dipatch` events to trigger workflows based on trigger phrases found in PR comments (e.g. `/test`). Effectively, this commit reworks all `issue_comment`-based workflows to use `workflow_dipatch` instead.

In order to continue testing changes coming from the PR in a secure fashion as we did with `issue_comment`, `workflow_dipatch`-based workflows implement a similar principle to what we used to have before:

- The workflows run in the context of the PR target branch (e.g. `main`) and thus are using workflow files from a trusted source while having access to repository secrets.
- The PR branch itself is tested by deploying the images built off the PR in ephemeral self-contained environments (e.g. a short-lived cluster).

In effect, from the point of view of external contributors, there will be no change in how the CI works and how they interact with it.

However, this change brings a good number of QoL advantages:

1. As can be seen in the diffs, most of the boilerplate we used to have for handling triggers and detecting whether or not this specific workflow should run based on the PR changelist is gone, therefore reducing maintenance drag on workflows.
2. Building up on this, when it is determined that a workflow does not need to be run at all based on the PR changelist, the corresponding PR check will automatically be marked as `Skipped`, so that the PR is not blocked due to missing a required PR check. (We had some workflows with boilerplate addressing this by manually marking the check as OK, but this is now handled consistently for all workflows and using the proper `Skipped` status, which wasn't possible with the previous boilerplate.)
3. For contributors with write privileges to the repository, Ariane will transparently run workflows in the context of the PR branch instead of the context of the PR target branch. This allows easy testing of changes to workflow files directly from a PR without any edit tricks.
4. Since the GitHub App is able to determine properly in which context to run pipelines, we can store the workflow files directly in the proper branch (e.g. stable workflows will live in stable branches). This will greatly simplify backport duties:
   - No need to maintain multiple sets of workflows with some being duplicated in `main` and some living in the appropriate branches, now everything is stored in the same place.
   - Changes made to backport workflows can be tested without even more cumbersome workflow edit tricks than the ones regular workflows already required.
5. There is now immediate feedback that "something" is happening when `/test` is commented, since Ariane will react to the comment to ack on the request.

Additional commit notes:

- This refactor standardized commit status reporting at the beginning and end of all workflows using `myrotvorets/set-commit-status-action` instead of `Sibz/github-status-action`, since the former allows for less boilerplate. Going forward we plan to find a way to remove that boilerplate as well via the GitHub App.
- We had to move back to regular secret-based authentication on AKS as the OIDC subject claims currently available on AKS do not allow for authorizing arbitrary branches. Microsoft is aware of this limitation and a new system is being worked on to allow it (see [1]), we will move back to OIDC when it is rolled out.

[1]: https://github.com/Azure/azure-workload-identity/issues/373#issuecomment-1611854839


nbusseneau:
All the reworked CI workflows are passing, here is an example to one of the working runs based on the new `workflow_dispatch` event: https://github.com/cilium/cilium/actions/runs/5414079403

I will now split this PR into multiple PRs:
- The current PR will keep all changes to `main` workflows, as well as the Ariane configuration file for `main`, and will remove all stable workflows from the `main` branch.
- One PR will be made for each stable branch, with the workflows and Ariane configuration for that stable branch.

nbusseneau:
Reviews from @bimmlerd @jschwinger233 have been requested as CODEOWNERS respectively on `/.github/` and `/.github/workflows/*datapath*.yaml`, but these reviews are not actually required since these changes do not make any contributing nor datapath changes. We only make changes relevant for `github-sec` and `ci-structure`.

.github/workflows/conformance-aks.yaml
```diff
@@ -2,26 +2,24 @@ name: ConformanceAKS (ci-aks)
...
-  issue_comment:
-    types:
-      - created
+  workflow_dispatch:
+    inputs:
+      PR-number:
+        description: "Pull request number."
+        required: true
+      context-ref:
+        description: "Context in which the workflow runs. Will be either the PR branch itself (default), or the PR target branch (if user has write privileges on the repository)."
```
viktor-kurchenko: `if user has write` - > `if user has no write`?

bimmlerd (approved):
blind ack as per above

qmonnet (approved):
Same, here's an ack from datapath

.github/workflows/conformance-aks.yaml
nbusseneau: Yeah, you are correct. And it is also confusing to read... What do you think about rephrasing it like this instead? I think this makes it easier to understand.
> Context in which the workflow runs. If PR is from a fork, will be the PR target branch. If PR is NOT from a fork, will be the PR branch itself.

viktor-kurchenko: I like it.

viktor-kurchenko (approved):
Cool!

.github/workflows/conformance-aks.yaml
nbusseneau: Slightly more detailed version:
> Context in which the workflow runs. If PR is from a fork, will be the PR target branch (general case). If PR is NOT from a fork, will be the PR branch itself (this allows committers to test changes to workflows directly from PRs).

viktor-kurchenko: One more:
```
Workflow context:
 - PR branch itself for PR from the origin (for contributors with write privileges to test changes)
 - target branch for PR from a fork (general case)
```

nbusseneau: Can't do lists -- it's a single line description 😢

nbusseneau:
Do I need any more street cred or am I gucci?
image

nbusseneau:
> **Warning**
> ### Important note to merger: this PR and all related backport PRs must be merged at the same time to avoid issues.
> - https://github.com/cilium/cilium/pull/26578
> - https://github.com/cilium/cilium/pull/26579
> - https://github.com/cilium/cilium/pull/26580

I will try to list the implications of merging them:

- Required PR checks:
  - `main` should not need to be adjusted (except if maybe we want to re-mark `ci-aks` and `ci-gke` as required).
  - All stable branches need to be adjusted. The general idea is that all the existing `ci-{foo}-1.XX` checks need to be removed, and their corresponding `ci-{foo}` check need to marked as required instead.
    - 1.11:
      - Conformance AWS-CNI (ci-awscni)
      - ConformanceEKS (ci-eks)
      - External workloads (ci-external-workloads)
      - ConformanceGKE (ci-gke)
      - ClusterMesh (ci-multicluster) 
      - Cilium L4LB XDP (ci-l4lb)
    - 1.12:
      - Same checks as 1.11 above, and also
      - ConformanceAKS (ci-AKS)
    - 1.13:
      - Same checks as 1.12 above
- Outstanding PRs targeting `main`:
  - If PR branch comes from a fork, should not need to rebased to be able to use `/test`.
  - If PR branch comes from upstream, will need to be rebased to use `/test`.
    - If it's a PR making changes to `main` workflows, rebasing may induce shenanigans. I'm happy to help.
    - If it's a PR making changes to stable workflows, rebasing will not be possible. The PR should be closed and instead rebased and re-opened against the appropriate stable branch, which may induce shenanigans again. I'm happy to help as well.
- DataStudio test results monitoring:
  - All workflows will now only use the same workflow file and workflow name on all branches, instead of being split over multiple workflow files with different workflow names based on the branch. The differentiating factor is now the branch triggering the workflow. DataStudio filters should be adjusted accordingly.

aanm (approved):

bimmlerd:
I missed suggesting it on the PR, but was reminded by https://github.com/cilium/cilium/pull/26653 - Could/should we document the Ariane setup in the repo?

nbusseneau:
Yes, the CI documentation is in sore need of an overhaul... 😢 I will try to make some time for it.

Issue #25442, opened 2023-05-15
CI: ClusterMesh: curl: (28) Resolving timed out after 2000 milliseconds


YutaroHayakawa:
## CI failure

```
[.] Action [to-entities-world/pod-to-world/http-to-google.com-0: cilium-test/client-6965d549d5-cflp6 (10.242.1.35) -> google.com-http (google.com:80)]
  ❌ command "curl -w %{local_ip}:%{local_port} -> %{remote_ip}:%{remote_port} = %{response_code} --silent --fail --show-error --output /dev/null --connect-timeout 2 --max-time 10 --retry 3 --retry-all-errors --retry-delay 3 [http://google.com:80](http://google.com/)" failed: command failed: curl: (28) Resolving timed out after 2000 milliseconds

  ℹ️  curl output:
  
  
  📄 No flows recorded for peer cilium-test/client-6[96](https://github.com/cilium/cilium/actions/runs/4975949770/jobs/8903597823#step:21:97)5d549d5-cflp6 during action http-to-google.com-0
  📄 No flows recorded for peer google.com-http during action http-to-google.com-0

📋 Test Report
❌ 1/41 tests failed (1/272 actions), 9 tests skipped, 1 scenarios skipped:
Test [to-entities-world]:
  ❌ to-entities-world/pod-to-world/http-to-google.com-0: cilium-test/client-6965d549d5-cflp6 (10.242.1.35) -> google.com-http (google.com:80)
```

Link: https://github.com/cilium/cilium/actions/runs/4975949770
Sysdump: [cilium-sysdumps-8.zip](https://github.com/cilium/cilium/files/11474705/cilium-sysdumps-8.zip)

qmonnet:
[Hit](https://github.com/cilium/cilium/actions/runs/5271892714/jobs/9533415372) on https://github.com/cilium/cilium/pull/23219.

PR #26238, opened 2023-06-14
endpoint: moveNewFilesTo performance and error handling improvements

learnitall:
These commits make changes to the function moveNewFilesTo to reduce its complexity from O(n^2) to O(n), decrease its CPU and memory usage, and improve its error handling. The function moveNewFilesTo is a utility function that may be called in the endpoint regeneration hot path.

Details of the changes can be found in each commit.

      GitHub issue.

I believe this needs the following labels:

* `release-note/misc`
* `kind/enhancement`

Also, I think this could be backported, but I haven't taken the time yet to dig into how feasible this would be.

### DIFF
--- a/pkg/endpoint/directory.go
+++ b/pkg/endpoint/directory.go
@@ -48,27 +48,43 @@ func (e *Endpoint) backupDirectoryPath() string {
 
 // moveNewFilesTo copies all files, that do not exist in newDir, from oldDir.
 func moveNewFilesTo(oldDir, newDir string) error {
-	oldFiles, err := os.ReadDir(oldDir)
+	var err error
+
+	oldDirFile, err := os.Open(oldDir)
 	if err != nil {
 		return err
 	}
-	newFiles, err := os.ReadDir(newDir)
+	defer oldDirFile.Close()
+
+	oldFiles, err := oldDirFile.Readdirnames(-1)
 	if err != nil {
 		return err
 	}
 
+	newDirFile, err := os.Open(newDir)
+	if err != nil {
+		return err
+	}
+	defer newDirFile.Close()
+
+	newFiles, err := newDirFile.Readdirnames(-1)
+	if err != nil {
+		return err
+	}
+
+	newFilesHash := make(map[string]struct{}, len(newFiles))
+	for _, f := range newFiles {
+		newFilesHash[f] = struct{}{}
+	}
+
+	var ok bool
+
 	for _, oldFile := range oldFiles {
-		exists := false
-		for _, newFile := range newFiles {
-			if oldFile.Name() == newFile.Name() {
-				exists = true
-				break
-			}
-		}
-		if !exists {
-			os.Rename(filepath.Join(oldDir, oldFile.Name()), filepath.Join(newDir, oldFile.Name()))
+		if _, ok = newFilesHash[oldFile]; !ok {
+			os.Rename(filepath.Join(oldDir, oldFile), filepath.Join(newDir, oldFile))
 		}
 	}
+
 	return nil
 }
 
--- a/pkg/endpoint/directory.go
+++ b/pkg/endpoint/directory.go
@@ -47,29 +47,31 @@ func (e *Endpoint) backupDirectoryPath() string {
 }
 
 // moveNewFilesTo copies all files, that do not exist in newDir, from oldDir.
+// It assumes that oldDir and newDir are an endpoint's old and new state
+// directories (see synchronizeDirectories below).
 func moveNewFilesTo(oldDir, newDir string) error {
 	var err error
 
 	oldDirFile, err := os.Open(oldDir)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to open old endpoint state dir: %w", err)
 	}
 	defer oldDirFile.Close()
 
 	oldFiles, err := oldDirFile.Readdirnames(-1)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to list old endpoint state dir: %w", err)
 	}
 
 	newDirFile, err := os.Open(newDir)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to open new endpoint state dir: %w", err)
 	}
 	defer newDirFile.Close()
 
 	newFiles, err := newDirFile.Readdirnames(-1)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to list new endpoint state dir: %w", err)
 	}
 
 	newFilesHash := make(map[string]struct{}, len(newFiles))
@@ -81,7 +83,9 @@ func moveNewFilesTo(oldDir, newDir string) error {
 
 	for _, oldFile := range oldFiles {
 		if _, ok = newFilesHash[oldFile]; !ok {
-			os.Rename(filepath.Join(oldDir, oldFile), filepath.Join(newDir, oldFile))
+			if err := os.Rename(filepath.Join(oldDir, oldFile), filepath.Join(newDir, oldFile)); err != nil {
+				return fmt.Errorf("failed to move endpoint state file: %w", err)
+			}
 		}
 	}
 
--- a/pkg/endpoint/directory_test.go
+++ b/pkg/endpoint/directory_test.go
@@ -4,8 +4,11 @@
 package endpoint
 
 import (
+	"fmt"
+	"math"
 	"os"
 	"path/filepath"
+	"testing"
 
 	check "github.com/cilium/checkmate"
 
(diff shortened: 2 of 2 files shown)


pkg/endpoint/directory.go
```diff
@@ -48,27 +49,48 @@ func (e *Endpoint) backupDirectoryPath() string {
...
-		}
-		if !exists {
-			os.Rename(filepath.Join(oldDir, oldFile.Name()), filepath.Join(newDir, oldFile.Name()))
+		if ok := newFilesHash[oldFile]; !ok {
+			oldFile := oldFile
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				os.Rename(filepath.Join(oldDir, oldFile), filepath.Join(newDir, oldFile))
+			}()
```
christarazi: I'm not sure how many `oldFiles` there can be but this is spinning off as many goroutines as there are `oldFiles`. While the body of the goroutine is quite small--just renaming a file, this can put additional pressure onto the Go runtime if it is already managing many goroutines or if it is already under CPU pressure. We probably want to bound this number of goroutines to at least the number of CPUs, but even then this code is running for each endpoint that's regenerating, which can happen in parallel.

marseel: I agree it would be nice to understand how many `oldFiles` there can be. I don't think the number of goroutines matters that much tbh, but we can limit concurrency here for sure - even something constant should be fine, like 5 or 10, as this is mostly io operation. Also, it would be nice to handle error from `os.Rename` as we haven't done that before...

learnitall: I pushed some changes to use an `errgroup.Group` to limit the number of goroutines to five and allow for us to bubble-up errors from `os.Rename`. Five goroutines is a small enough amount that it won't have a drastic impact on the runtime, and since these operations are IO bound we don't need to worry about these goroutines contributing to CPU pressure.

@christarazi helped me look into the number of `oldFiles` that we could expect and it should stay less than 10. There's four static files that are present here with an additional three that could be created if debug mode is enabled.

Hey @aanm, it looks like you originally authored this function five years ago in https://github.com/cilium/cilium/commit/f6c4385a43943c7be71bd8ca3a0582cab24db20b. Do you remember if there was a reason for not checking the error on `os.Rename`?

aanm: @learnitall no specific reason.

aanm: I didn't look to much into this PR but from what I can see the total number of go routines will depend on the number of endpoints on the node times ~4 (number of old files).

learnitall: I did a quick test to see what would happen if we kept the other changes in this PR, but run `os.Rename` in serial, and I don't see a difference. The sec/op stays about the same, and memory usage looks better since we skip creating the goroutines:

```
goos: linux
goarch: amd64
pkg: github.com/cilium/cilium/pkg/endpoint
cpu: 12th Gen Intel(R) Core(TM) i7-12700H
                   │ MoveNewFilesTo1-before--.txt │   MoveNewFilesTo1-after.txt   │
                   │            sec/op            │   sec/op     vs base          │
MoveNewFilesTo1-20                    5.508µ ± 0%   5.514µ ± 1%  ~ (p=0.363 n=10)

                   │ MoveNewFilesTo1-before--.txt │     MoveNewFilesTo1-after.txt      │
                   │             B/op             │    B/op     vs base                │
MoveNewFilesTo1-20                     438.0 ± 0%   373.0 ± 0%  -14.84% (p=0.000 n=10)

                   │ MoveNewFilesTo1-before--.txt │     MoveNewFilesTo1-after.txt     │
                   │          allocs/op           │ allocs/op   vs base               │
MoveNewFilesTo1-20                     11.00 ± 0%   10.00 ± 0%  -9.09% (p=0.000 n=10)
goos: linux
goarch: amd64
pkg: github.com/cilium/cilium/pkg/endpoint
cpu: 12th Gen Intel(R) Core(TM) i7-12700H
                   │ MoveNewFilesTo5-before--.txt │   MoveNewFilesTo5-after.txt   │
                   │            sec/op            │   sec/op     vs base          │
MoveNewFilesTo5-20                    6.376µ ± 1%   6.383µ ± 1%  ~ (p=0.897 n=10)

                   │ MoveNewFilesTo5-before--.txt │     MoveNewFilesTo5-after.txt     │
                   │             B/op             │    B/op     vs base               │
MoveNewFilesTo5-20                     705.0 ± 0%   641.0 ± 0%  -9.08% (p=0.000 n=10)

                   │ MoveNewFilesTo5-before--.txt │     MoveNewFilesTo5-after.txt     │
                   │          allocs/op           │ allocs/op   vs base               │
MoveNewFilesTo5-20                     20.00 ± 0%   19.00 ± 0%  -5.00% (p=0.000 n=10)
goos: linux
goarch: amd64
pkg: github.com/cilium/cilium/pkg/endpoint
cpu: 12th Gen Intel(R) Core(TM) i7-12700H
                    │ MoveNewFilesTo10-before--.txt │  MoveNewFilesTo10-after.txt   │
                    │            sec/op             │   sec/op     vs base          │
MoveNewFilesTo10-20                     7.694µ ± 4%   7.644µ ± 1%  ~ (p=0.089 n=10)

                    │ MoveNewFilesTo10-before--.txt │     MoveNewFilesTo10-after.txt      │
                    │             B/op              │     B/op      vs base               │
MoveNewFilesTo10-20                    1.385Ki ± 0%   1.322Ki ± 0%  -4.51% (p=0.000 n=10)

                    │ MoveNewFilesTo10-before--.txt │    MoveNewFilesTo10-after.txt     │
                    │           allocs/op           │ allocs/op   vs base               │
MoveNewFilesTo10-20                      31.00 ± 0%   30.00 ± 0%  -3.23% (p=0.000 n=10)```

Comment (learnitall) [reply pkg/endpoint/directory.go]: Given the above, I'll go ahead and remove usage of goroutines to make this PR safer.

File: pkg/endpoint/directory.go [hunk 2]
```diff
@@ -48,27 +48,43 @@ func (e *Endpoint) backupDirectoryPath() string {
...
+
+	newFiles, err := newDirFile.Readdirnames(-1)
+	if err != nil {
+		return err
+	}
+
+	newFilesHash := make(map[string]bool, len(newFiles))
+	for _, f := range newFiles {
+		newFilesHash[f] = true
+	}
```
christarazi: `newFilesHash` can use `struct{}` as a value instead of `bool` since `true` is the only value that's ever set in the map.

pkg/endpoint/directory.go
```diff
@@ -81,7 +81,9 @@ func moveNewFilesTo(oldDir, newDir string) error {
 
 	for _, oldFile := range oldFiles {
 		if ok = newFilesHash[oldFile]; !ok {
-			os.Rename(filepath.Join(oldDir, oldFile), filepath.Join(newDir, oldFile))
+			if err := os.Rename(filepath.Join(oldDir, oldFile), filepath.Join(newDir, oldFile)); err != nil {
+				return err
```
christarazi: Could you add some context here like `fmt.Errorf("failed to rename new endpoint state files: %w", err)`

learnitall: Done

pkg/endpoint/directory.go
learnitall: Ah good catch, changed.

christarazi (approved):
LGTM thanks. There are some CI failures including lint failures.

learnitall:
Looks like we hit two flakes in the last test:

* K8sUpstreamNetConformance is full of `putEndpointIdTooManyRequests`, which is tracked in #24361. (Failed job: https://github.com/cilium/cilium/actions/runs/5284212566/jobs/9561415454?pr=26238)
* ClusterMesh failed on a pod-to-world test that curls google, which is tracked in #25442. (Failed job: https://github.com/cilium/cilium/actions/runs/5284646478/jobs/9562294636#step:22:428)

The K8sUpstreamNetConformance test seems like an important one to pass for this PR, since this function hasn't been touched in ~five years, but the changes are fairly trivial. @christarazi how would you like to go from here?

christarazi:
@learnitall Have you checked the sysdumps / logs of Cilium to see if any of the errors come from the newly added error handling that this PR introduces? That is one easy way to check if this PR is relevant to the CI failures.

christarazi:
Also marking to wait for release due to feature freeze.

learnitall:
I haven't checked sysdumps/logs. Looks like the artifacts for them expired so I'll retrigger and take a look.

learnitall:
After re-running and looking through the sysdump and logs for the failing net conformance test (here: https://github.com/cilium/cilium/actions/runs/5338165106/jobs/9675167448?pr=26238), I can't find any reference to the errors that I added. There are five new errors, and any one of them can be found with:

```
grep -E 'failed to (open|list|move)( old| new)? endpoint'
```

aditighag (approved):

learnitall:
Hit a weird timeout with ginkgo tests: https://github.com/cilium/cilium/issues/27165

PR #26239, opened 2023-06-14
add controller group runs metric

asauber:
Adds a metric which counts controller runs by "group" (currently the controller name).

Currently produces the following type of output:

```
cilium_controllers_group_runs_total{group_name="bpf-map-sync-cilium_lxc",status="success"} 25
cilium_controllers_group_runs_total{group_name="cilium-health-ep",status="success"} 3
cilium_controllers_group_runs_total{group_name="dns-garbage-collector-job",status="success"} 3
cilium_controllers_group_runs_total{group_name="endpoint-gc",status="success"} 1
cilium_controllers_group_runs_total{group_name="ipcache-inject-labels",status="success"} 10
cilium_controllers_group_runs_total{group_name="k8s-heartbeat",status="success"} 5
cilium_controllers_group_runs_total{group_name="link-cache",status="success"} 9
cilium_controllers_group_runs_total{group_name="metricsmap-bpf-prom-sync",status="success"} 26
cilium_controllers_group_runs_total{group_name="resolve-identity-2088",status="success"} 1
cilium_controllers_group_runs_total{group_name="restoring-ep-identity (1153)",status="success"} 1
cilium_controllers_group_runs_total{group_name="sync-host-ips",status="success"} 3
cilium_controllers_group_runs_total{group_name="sync-lb-maps-with-k8s-services",status="success"} 1
cilium_controllers_group_runs_total{group_name="sync-policymap-1153",status="success"} 5
cilium_controllers_group_runs_total{group_name="sync-policymap-2088",status="success"} 4
cilium_controllers_group_runs_total{group_name="sync-to-k8s-ciliumendpoint (1153)",status="success"} 13
cilium_controllers_group_runs_total{group_name="sync-to-k8s-ciliumendpoint (2088)",status="success"} 14
cilium_controllers_group_runs_total{group_name="sync-utime",status="success"} 3
cilium_controllers_group_runs_total{group_name="write-cni-file",status="success"} 1
```

Clearly, some of these controller names are dynamic based on the Cilium endpoint ID, which should be handled somehow.

At least three approaches to this:

1. One idea is to require the `controller` struct to have a new field `groupName`, set in order to recieve this metric.
(**likely best idea and going to try this**)

2. Another idea is to dynamically strip anything that looks like an ID in the group name.

3. Yet another idea is to prescriptively strip IDs based on known dynamically-named controllers, although this runs the risk of new such controllers being added without the corresponding aggregation.

Next steps here:

* Find a good way to logically group controllers
* Put this behind a flag.
* Address concerns about cardinality and memory usage with runtime data.

### DIFF
--- a/clustermesh-apiserver/users_mgmt.go
+++ b/clustermesh-apiserver/users_mgmt.go
@@ -97,6 +97,7 @@ func (us *usersManager) Start(hive.HookContext) error {
 	}
 
 	us.manager.UpdateController(usersMgmtCtrl, controller.ControllerParams{
+		Group:   usersMgmtCtrl,
 		Context: context.Background(),
 		DoFunc:  us.sync,
 	})
--- a/operator/cmd/ccnp_event.go
+++ b/operator/cmd/ccnp_event.go
@@ -128,6 +128,7 @@ func enableCCNPWatcher(ctx context.Context, wg *sync.WaitGroup, clientset k8sCli
 
 	mgr.UpdateController("ccnp-to-groups",
 		controller.ControllerParams{
+			Group: "ccnp-to-groups",
 			DoFunc: func(ctx context.Context) error {
 				groups.UpdateCNPInformation(clientset)
 				return nil
--- a/daemon/cmd/cni/config.go
+++ b/daemon/cmd/cni/config.go
@@ -179,6 +179,7 @@ func (c *cniConfigManager) Start(hive.HookContext) error {
 	// Install the CNI file controller
 	c.controller.UpdateController(cniControllerName,
 		controller.ControllerParams{
+			Group: cniControllerName,
 			DoFunc: func(ctx context.Context) error {
 				err := c.setupCNIConfFile()
 				if err != nil {
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -1213,6 +1213,7 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 	d.controllers.UpdateController(
 		syncHostIPsController,
 		controller.ControllerParams{
+			Group: syncHostIPsController,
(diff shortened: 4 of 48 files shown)


clustermesh-apiserver/users_mgmt.go
```diff
@@ -97,6 +97,7 @@ func (us *usersManager) Start(hive.HookContext) error {
 	}
 
 	us.manager.UpdateController(usersMgmtCtrl, controller.ControllerParams{
+		Group:   usersMgmtCtrl,
```
tommyp1ckles, line 100: I think for this component you would have to [explicitly enable the metric ](https://github.com/cilium/cilium/blob/main/clustermesh-apiserver/metrics/metrics.go#L75)

operator/cmd/ccnp_event.go
```diff
@@ -128,6 +128,7 @@ func enableCCNPWatcher(ctx context.Context, wg *sync.WaitGroup, clientset k8sCli
 
 	mgr.UpdateController("ccnp-to-groups",
 		controller.ControllerParams{
+			Group: "ccnp-to-groups",
```
tommyp1ckles, line 131: 💡 idea: since we're not bound by the controller identifier, maybe we could make these more readable: ```ciliumclusterwidenetworkpolicy-to-groups```

(might actually be too hard to type in this situation)

clustermesh-apiserver/users_mgmt.go
asauber: Good to know. I'll have to confirm this.

operator/cmd/ccnp_event.go
asauber: I like the idea of extra verbosity, I'll consider this when moving these names into constants.

asauber:
Closing in favor of https://github.com/cilium/cilium/pull/26850

PR #26244, opened 2023-06-15
Updates TransformXXX Functions in k8s pkg

### LINKED ISSUE #26240: K8s TransformXXX() Funcs Should Return SlimXXX Objects

### What happened?

With https://github.com/cilium/cilium/pull/25604, a nil object and error are returned when a controller, that is configured with a transformer, processes a SlimXXX object. Previous to this PR, the provided object would be returned. Although controllers that do not require object transformation should create an informer with a `nil` transformer, the TransformXXX() functions should be updated to check for `SlimXXX` objects and return them without transformation.

### Cilium Version

Latest

### Kernel Version

N/A

### Kubernetes Version

Latest

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

danehans:
Updates TransformXXX functions to not return an error if the provided and returned objects are the same type. Instead, the provided object is returned without any transformations.

Fixes: #26240

### DIFF
--- a/pkg/k8s/factory_functions.go
+++ b/pkg/k8s/factory_functions.go
@@ -472,11 +472,13 @@ func ConvertToSlimIngressLoadBalancerStatus(slimLBStatus *slim_corev1.LoadBalanc
 	}
 }
 
-// TransformToK8sService transforms a *v1.Service into a
-// *slim_corev1.Service or a cache.DeletedFinalStateUnknown into
-// a cache.DeletedFinalStateUnknown with a *slim_corev1.Service in its Obj.
-// If the given obj can't be cast into either *slim_corev1.Service
-// nor cache.DeletedFinalStateUnknown, an error is returned.
+// TransformToK8sService transforms a *v1.Service into a *slim_corev1.Service
+// or a cache.DeletedFinalStateUnknown into a cache.DeletedFinalStateUnknown
+// with a *slim_corev1.Service in its Obj. If obj is a *slim_corev1.Service
+// or a cache.DeletedFinalStateUnknown with a *slim_corev1.Service in its Obj,
+// obj is returned without any transformations. If the given obj can't be cast
+// into either *slim_corev1.Service nor cache.DeletedFinalStateUnknown, an error
+// is returned.
 func TransformToK8sService(obj interface{}) (interface{}, error) {
 	switch concreteObj := obj.(type) {
 	case *v1.Service:
@@ -511,7 +513,12 @@ func TransformToK8sService(obj interface{}) (interface{}, error) {
 				},
 			},
 		}, nil
+	case *slim_corev1.Service:
+		return obj, nil
 	case cache.DeletedFinalStateUnknown:
+		if _, ok := concreteObj.Obj.(*slim_corev1.Service); ok {
+			return obj, nil
+		}
 		svc, ok := concreteObj.Obj.(*v1.Service)
 		if !ok {
 			return nil, fmt.Errorf("unknown object type %T", concreteObj.Obj)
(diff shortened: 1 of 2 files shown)


danehans:
@odinuge since you commented in https://github.com/cilium/cilium/pull/25604, PTAL.

odinuge (approved):
Thanks! The code looks good to me!

Outside the scope of this PR; I think at some point it would be more useful to extraxct the godoc for these functions to a common description, so we avoid having it duplicated that much; given its so easy to miss one and get inconsistent comments - given they ~all implement the same thing, just for different types.

giorio94 (approved):
Thanks! /lgtm

giorio94:
> Outside the scope of this PR; I think at some point it would be more useful to extraxct the godoc for these functions to a common description, so we avoid having it duplicated that much; given its so easy to miss one and get inconsistent comments - given they ~all implement the same thing, just for different types.

One step further might be to leverage generics to extract all the common logic, having only a single function as parameter which takes care of the actual conversion from the source to the destination type. This would reduce the overall code duplication and ensure consistency across all implementations. Related: https://github.com/cilium/cilium/pull/26367

danehans:
@giorio94 thanks for the heads-up regarding #26367. Based on my https://github.com/cilium/cilium/pull/26367/files#r1237222348, we may be able to close this PR and have #26367 fix #26240.

giorio94:
> @giorio94 thanks for the heads-up regarding #26367. Based on [my comment](https://github.com/cilium/cilium/pull/26367/files#r1237222348), we may be able to close this PR and have #26367 fix #26240.

The two PRs are currently addressing two separate aspects (this one tackles the transformation functions which do an actual type transformation, while #26367 the ones which don't mutate the object type).

danehans:
@giorio94 thanks for the clarification. I noticed the `TransformToGeneric()` func in https://github.com/cilium/cilium/pull/26367 and thought it was using generics for all transformation funcs. Back to your https://github.com/cilium/cilium/pull/26244#issuecomment-1600388029, should I update this PR to use generics? Since https://github.com/cilium/cilium/issues/26240 is expected to be a release blocker, IMO it's better for this PR to introduce minimal changes and tackle the transition to generics in a separate PR that is merged after the release. Thoughts?

giorio94:
> Back to your [#26244 (comment)](https://github.com/cilium/cilium/pull/26244#issuecomment-1600388029), should I update this PR to use generics? Since #26240 is expected to be a release blocker, IMO it's better for this PR to introduce minimal changes and tackle the transition to generics in a separate PR that is merged after the release. Thoughts?

Yeah, I agree that it is better to drive this fix to completion and tackle more invasive refactorings in a followup.

joestringer:
I re-ran `K8sUpstreamNetConformance` tests. From the sheer number of tests that failed, I don't quite understand whether some of the test failures could be related to the PR or the workflow is just a bit broken sometimes. :(

tommyp1ckles:
@danehans 

These changes look very reasonable to me, however I am still a bit confused as to how this situation might arise?

In the usage of this transform function in the operator, the [k8s client service interface](https://github.com/cilium/cilium/blob/main/operator/cmd/root.go#L521) returns a `*v1.Service` type.

danehans:
@tommyp1ckles thanks for the review. I agree that `TransformToK8sService()` should only receive a k8s Service. However, this PR makes `TransformToK8sService()` and other transform functions [backward compatible](https://github.com/cilium/cilium/blob/v1.13.3/pkg/k8s/factory_functions.go#L453-L534) prior to the changes introduced by https://github.com/cilium/cilium/pull/25604. I also believe this PR follows the updated guidance for using [TransformFunc](https://github.com/cilium/cilium/blob/main/vendor/k8s.io/client-go/tools/cache/delta_fifo.go#L145-L148). I can revert the `TransformToK8sService()` changes, especially since it's not being used as a `TransformFunc`, but I think it's safer to keep it as-is at this point of the release cycle and reassess after the release. Thoughts?

danehans:
@joestringer thanks for rerunning the K8sUpstreamNetConformance test. It's now passing but now the ConformanceEKS test is failing due to:

```sh
Error reading Cilium logs: error getting cilium-agent logs for kube-system/cilium-zxm7g: Get "https://192.168.116.242:10250/containerLogs/kube-system/cilium-zxm7g/cilium-agent?sinceTime=2023-06-23T06%3A13%3A12Z&timestamps=true": dial tcp 192.168.116.242:10250: i/o timeout
[=] Test [all-entities-deny]
```

I don't believe the ConformanceEKS test failures are related to this PR. Are you able to rerun this test?

tommyp1ckles (approved):
@danehans Thanks for the clarification, lgtm

joestringer:
I tried retriggering tests to kick off the gateway-conformance jobs, but I think actually the PR needs to be rebased against the latest main branch. @danehans could you rebase and retrigger the tests?

PR #26245, opened 2023-06-15
docs: Upgrade Note For Deny Policy Fix

nathanjsweet:


Documentation/operations/upgrade.rst
```diff
@@ -333,6 +333,12 @@ Annotations:
   Egress rules in CiliumNetworkPolicy CRD. The old attribute name is no longer supported,
   please update your CiliumNetworkPolicy CRD accordingly. Also applicable values for this
   attribute are changed to ``disabled``, ``required`` and ``test-always-fail``.
+* Deny policies are now stable because of a fix to Cilium's internal CIDR policy
+  enforcement logic. Previously, a CIDR-based allow-policy would always allow traffic,
+  even if there was CIDR-based deny-policy in place that should have overridden it. Please
+  check all of your CIDR-based deny and allow policies to ensure that your cluster is not
+  accidentally relying on an allow-policy that should not work and will no longer work
+  after upgrading to 1.14.
```
joestringer: I understand what this text "Please check ..." is trying to communicate to the user, but I highly doubt that most users will be able to interpret this and take action. :thinking: 

Historically if we wanted to provide a more comprehensive explanation, we may add a section near the bottom that shares more detail: https://docs.cilium.io/en/stable/operations/upgrade/#advanced .

What I'm thinking about is something like two examples, "Here's an example policy that is too broad: ...", "On Cilium 1.14 or later, the following should accurately allow the traffic: ...".

zacharysarah: I agree with @joestringer that users need more information about how specifically to verify that their allow/deny policies work as intended. If there's content that already provides this, let's link to it here. If not, let's add that content and determine where the best place for it to live might be, whether on this page or in the upgrade guide linked by Joe. My preference is for the latter.

Edits for clarity, hyphen nit, use clear, plain language
```suggestion
* Due to a fix to Cilium's internal CIDR policy enforcement logic in v1.14, deny policies are now stable.
  Previously, a CIDR-based allow policy allowed all traffic,
  even if there was a CIDR-based deny policy to deny traffic.
  
  Verify that all of your CIDR-based deny and allow policies work as intended, so that your cluster does not
  accidentally rely on an overly permissive allow policy that will not work
  after upgrading to 1.14.
```

zacharysarah (changes requested):
@nathanjsweet Good start! Some edits for clarity, and see my concurrence with @joestringer about what else this PR requires.

joestringer (approved):

Documentation/operations/upgrade.rst
```diff
@@ -342,6 +342,40 @@ Annotations:
   to prevent the possibility of connectivity disruptions. Note: this issue does not
   affect setups using a persistent etcd cluster instead of the ephemeral one bundled
   with the *clustermesh-apiserver*.
+* Deny policies now always take precedence over allow policies. Previously, a CIDR-based
+  allow policy would always allow traffic, even if there was an overlapping CIDR-based deny policy
+  to deny the same traffic. Now, the CIDR-based deny policy will drop traffic when there is
```
zacharysarah: use present tense
```suggestion
  to deny the same traffic. Now, a CIDR-based deny policy drops traffic when there is
```

Documentation/operations/upgrade.rst
```diff
@@ -342,6 +342,40 @@ Annotations:
...
+  an allow policy for the same traffic. Verify that all of your CIDR-based deny and allow policies work
+  as intended. The following example shows an allow policy that would previously allow
+  all egress traffic to ``20.1.1.1`` for its selector, but that traffic will now be dropped
+  by the deny policy:
```
zacharysarah: Break this into a new paragraph
```suggestion
  an allow policy for the same traffic.
  
  Verify that all of your CIDR-based deny and allow policies work
  as intended. The following example shows an allow policy that would previously allow
  all egress traffic to ``20.1.1.1`` for its selector, but for which a deny policy now halts traffic:
```

zacharysarah (changes requested):
Some small nits

zacharysarah (approved):
🚀

PR #23202, opened 2023-01-19
Introduce node IDs

pchaigno:
This pull request introduces node ID, node-scoped 16-bits-long unique identifiers for remote nodes. Those IDs are allocated locally and plugged into the ipcache and tunnel maps. In a subsequent pull request, the datapath will use them to identify remote nodes.

Summary of commits:
1. Introduces the allocation and deallocation logic for the node IDs. IDs are attached to node IP addresses.
2. Exposes the node IDs and their associated IP addresses via the agent's API.
3. Introduces a new CLI command to dump the node IDs and their associated IP addresses.
4. Plugs the node IDs in the ipcache. Each pod gets the node ID corresponding to the node they run on.
5. Gives the tunnel map its own struct for keys and values, in preparation for the subsequent commit.
6. Plugs the node IDs in the tunnel map.

```release-note
Introduce node IDs in the datapath and the agent, so datapath can later use them to identify remote nodes
```

PR #26247, opened 2023-06-15
 pkg/datapath: Prefer v4 Node IP addr for NodeID allocation

brb:
Previously, node IDs were meant to be allocated per node regardless of node IP addr families. However, this became problematic when running Cilium with both v4 and v6 enabled. Sometimes Cilium could have allocated two different IDs for the same node. This confused XFRM states or IPcache \[1\].

A proper solution is to use node name or other than node IP addr as a reference for node ID allocations. Unfortunately, it's not possible, as the node ID allocation can happen in the pkg/ipcache (see AllocateNodeID), which is not aware of node object. Because of this limitation, we could allocate node IDs per each IP familiy for a given node. However, CiliumEndpoint prefers IPv4 address for host IP, so IPv6 of a node is not always accessible to the pkg/ipcache.

To fix the issue, let's prefer v4 node IP addr. This is the same as we do in CiliumEndpoint creation.

\[1\]: https://github.com/cilium/cilium/issues/26114#issuecomment-1587906785

Fixes: https://github.com/cilium/cilium/pull/23202

### DIFF
--- a/pkg/datapath/linux/node_ids.go
+++ b/pkg/datapath/linux/node_ids.go
@@ -93,14 +93,13 @@ func (n *linuxNodeHandler) getNodeIDForIP(nodeIP net.IP) (uint16, bool) {
 	return 0, false
 }
 
-// getNodeIDForNode gets the node ID for the given node if one was allocated
-// for any of the node IP addresses. If none is found, 0 is returned.
+// getNodeIDForNode gets the node ID for the given node if one was allocated.
+// If none if found, 0 is returned.
 func (n *linuxNodeHandler) getNodeIDForNode(node *nodeTypes.Node) uint16 {
 	nodeID := uint16(0)
-	for _, addr := range node.IPAddresses {
-		if id, exists := n.nodeIDsByIPs[addr.IP.String()]; exists {
-			nodeID = id
-		}
+	nodeIP := n.getNodeIPForIDAllocation(node)
+	if id, exists := n.nodeIDsByIPs[nodeIP]; exists {
+		nodeID = id
 	}
 	return nodeID
 }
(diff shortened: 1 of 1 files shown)


pkg/datapath/linux/node_ids.go
```diff
@@ -93,14 +99,13 @@ func (n *linuxNodeHandler) getNodeIDForIP(nodeIP net.IP) (uint16, bool) {
 	return 0, false
 }
 
-// getNodeIDForNode gets the node ID for the given node if one was allocated
-// for any of the node IP addresses. If none is found, 0 is returned.
+// getNodeIDForNode gets the node ID for the given node if one was allocated.
+// If none if found, 0 is returned.
```
ldelossa, line 97:
```suggestion
// If none is found, 0 is returned.
```

pkg/datapath/linux/node_ids.go
```diff
@@ -308,3 +301,23 @@ func (n *linuxNodeHandler) registerNodeIDAllocations(allocatedNodeIDs map[string
...
+
+// getNodeIPForAllocation is a helper function which tries to retrieve a node IP used
+// for node ID allocations from a given Node object.
+//
+// If the IPv4 is enabled, then IPv4 addr is preferred. This is because we use
+// node IPv4 addr in CiliumEndpoint objects. The node IP addr from CiliumEndpoint
+// is used to create IPcache entries. The latter's creation might invoke the
+// node ID allocation. So, we want to ensure that we always use the same IP addr
+// for the same node when allocating IDs.
+func (n *linuxNodeHandler) getNodeIPForAllocation(node *nodeTypes.Node) string {
```
ldelossa:
```suggestion
func (n *linuxNodeHandler) getNodeIPForIDAllocation(node *nodeTypes.Node) string {
```

pkg/datapath/linux/node_ids.go
```diff
@@ -308,3 +301,23 @@ func (n *linuxNodeHandler) registerNodeIDAllocations(allocatedNodeIDs map[string
...
+	for _, addr := range node.IPAddresses {
```
ldelossa: Just to confirm, the first ipv4 address found in this `node.IPAddresses` will be picked, this is desired?

ldelossa (approved):
Logic checks out. 

Few nits

pkg/datapath/linux/node_ids.go
brb: Your comment pointed to an issue I was not aware of. Actually, a node might have two IPv4 addrs - Internal and External. We need to prefer the former.

PR #26250, opened 2023-06-15
Replace client-go with private fork.

marseel:
Our private fork contains fix to exponential backoff in client-go for informers, that
wasn't backported in k8s: https://github.com/kubernetes/kubernetes/pull/118132


go.mod
```diff
@@ -263,6 +263,10 @@ replace (
 
 	go.universe.tf/metallb => github.com/cilium/metallb v0.1.1-0.20220829170633-5d7dfb1129f7
 
+	// Using private fork of client-go.
+	// Contains fix for exponential backoff for informers.
+	k8s.io/client-go => github.com/cilium/client-go v0.0.0-20230606101523-be41a63db9a0
```
odinuge: Small nit; It seems like the fork is missing the tags, making it harder to understand what version this is based off. I assume that if you push the `client-go` tags this version would be more readable. 😄 

This is the branch for reference; https://github.com/cilium/client-go/commits/kubernetes-1.27.2-fix

go.mod
```diff
@@ -263,6 +263,10 @@ replace (
 
 	go.universe.tf/metallb => github.com/cilium/metallb v0.1.1-0.20220829170633-5d7dfb1129f7
 
+	// Using private fork of client-go.
```
odinuge: nit; I see that the other pins also use the term `private`, but I think it would make more sense to just say `our fork` - given the fork is public and open source 😄

odinuge (commented):
Some small nits. Otherwise, the new backoff seems very neat! Thanks

go.mod
marseel: yup, you were right. Done.

go.mod
marseel: thanks, done!

tklauser (approved):

aanm (approved):

PR #26256, opened 2023-06-15
cilium statedb dump command & bugtool

joamaki:
As requested by `:manager:`:

Adds the API handler `/statedb/dump` and implements "cilium statedb dump"
on top to dump the contents of StateDB as JSON.

Extends sysdump to call "cilium statedb dump" to include the output in sysdump.


daemon/cmd/statedb.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/cilium/cilium/pkg/statedb"
+)
+
+type getStateDBDump struct {
+	db statedb.DB
+}
+
+func (h *getStateDBDump) Handle(params GetStatedbDumpParams) middleware.Responder {
+	return middleware.ResponderFunc(func(w http.ResponseWriter, _ runtime.Producer) {
+		h.db.WriteJSON(w)
```
bimmlerd, line 22: does swagger handle setting http response codes?

bimmlerd (approved):
yay, this'll make :manager: happy

daemon/cmd/statedb.go
joamaki: Yep and the content-type too (similar usage you can see in `daemon/cmd/map.go`. This could also be written with the `GetStatedbDumpOK` which takes an `io.Reader`, but then I'd need a pipe and etc. This is much shorter and more obvious.

qmonnet (approved):
Looks good

joestringer (approved):
Looks pretty trivially correct and will help with debugging the l2 responder / announcer feature in 1.14.

bimmlerd (approved):
foundations changes lgtm

bimmlerd:
@aanm is on PTO for a bit, api hence covered mostly by Nate - changing the review request to  him.

pippolo84 (approved):
LGTM :rocket:

pkg/statedb/db.go
```diff
@@ -64,22 +64,25 @@ func (db *stateDB) WriteJSON(w io.Writer) error {
 		if err != nil {
 			return err
 		}
-		buf.WriteString("\"" + table + "\": [\n")
+		buf.WriteString("  \"" + table + "\": [\n")
 		obj := iter.Next()
```
derailed, line 68: Might make sense to roll this initializer in the for loop below.

derailed (commented):
@joamaki Nice work! LGTM

Might be no opt here so feel free to disregard!
Not sure about the volumes at play but we're intermixing writeByte, writeString. 
If memory serves the later is a bit slower and given we're looping, figured I'll flag. 
Given our sysdump perf issues on some clusters might be worth seeing if there is a `diet plan` and put this on the bench?

bimmlerd:
> intermixing writeByte, writeString. If memory serves the later is a bit slower and given we're looping, figured I'll flag.

I'm not sure I follow your comment with `writeByte` vs `writeString` - writing the runes of the string individually seems slower, and casting to `[]byte{}` would potentially break runes apart? What's your suggestion instead?

In any case, we're writing to a buffered writer, afaik the writes should amortize to be comparably fast to a raw memcpy; the volume of data is not expected to be huge at this time, since there's just one statedb table.

> Given our sysdump perf issues on some clusters might be worth seeing if there is a `diet plan` and put this on the bench?

I agree that performance hasn't been the foremost concern of sysdumps. I'm reading your comment as a suggestion to put the collection of the statedb information behind a flag? I'm not opposed in general, but so far we've just slurped as much information as possible into a sysdump -- changing that will imo confuse debugging efforts of people unaware, due to missing information. If we want to do this, I'd advocate for doing it in some coordinated fashion over all of the sysdump information. I'd rather not have to go back and forth with requesting sysdumps with different configurations to get to the information needed to debug an issue, and usually I don't know which component is to blame before looking at a sysdump...

nathanjsweet (approved):

derailed:
@bimmlerd Thank you! I was not advocating introducing a flag, just a general comment on sysdump efficiency and benchmarking to optimize outputs.

derailed (approved):
LGTM

PR #26263, opened 2023-06-15
.github: rebuild ginkgo tests in case of cache miss

aanm:
When a test is re-trigger after a long period of time, the test binary might already been deleted from the GitHub cache. If we hit a cache miss we should rebuild the binary so that the test can continue to be executed.


mhofstetter (approved):
LGTM - thanks @aanm

giorio94 (approved):
Changes look good to me. Thanks! 

I wonder if it could make sense to add a comment to the original steps, to point out that there is another copy down here, and they should be kept in sync in case a modification is performed.

sayboras (approved):

brlbil (approved):

PR #26265, opened 2023-06-15
cilium: export intermediate cobra.Commands

lmb:
Export all cobra.Command which do not have a Run function set. This allows extending the CLI behaviour from outside the package without allowing modifications to individual commands.


jibi (approved):

derailed (commented):
@lmb Tx for this PR! 
Could be totally out of context here but not sure if I am missing the intent here...
It would be great to highlight some of the use cases this pr would enable to further understand the motivation?  
Especially the part about `without allowing modifications to individual commands`?
Isn't it the case that if we make these package globals then one can override the entire command structure since cobra exports all the interesting bits? 
Especially in light that rootCmd would now be exported.

Am I wrong about that?

lmb:
> It would be great to highlight some of the use cases this pr would enable to further understand the motivation?

The use case is to build a binary which exposes a superset `cilium` commands. Imagine I want to build a `sillyum` binary that has the `cowsay` subcommand or something similar. Exporting the Root command (and others) allows doing that without having to patch the source code.

> Isn't it the case that if we make these package globals then one can override the entire command structure since cobra exports all the interesting bits?

You can't change arbitrary subcommands from what I can tell (since those are unexported), but other bits are up for grabs, yes!

I have two ideas if you think this level of openness is problematic:
1. Move the exported commands into an `internal` subpackage. This means that only true forks of the repo can access the root command. Importing it in a different Go module is prohibited.
2. Instead of exporting the commands, export functions like `RootCmd() MyCommandType` where `MyCommandType` is something like the following:

```go
type MyCommandType interface{
  AddCommand(...)
 // Prevent others from implementing the interface. May not be needed.
  unexported()
}
```

lmb:
@derailed thoughts?

lmb:
After discussing on Slack, it turns out that exporting the root command is equivalent to exporting all cobra commands anyways, due to https://pkg.go.dev/github.com/spf13/cobra#Command.Commands So I think we're not making anything worse by exporting intermediate commands here.

derailed (approved):
@lmb LGTM

PR #26269, opened 2023-06-15
k8s.Resource metrics

dylandreimerink:
This PR makes the pkg/k8s/resources emit the same metrics as their
k8s watcher counterparts. The only difference is that due to the
asynchronous nature of resource consumers, we are not able to track when
all consumers have processed the same event, thus we increment the
processed event metric for every consumer.

```release-note
Added metrics for pkg/k8s/resource
```


rastislavs (approved):
LGTM for BGP

squeed (approved):

Issue #25758, opened 2023-05-30
CI: kubernetes-e2e (ipv4): ReplicaSet should serve a basic image on each replica with a public image failed


YutaroHayakawa:
## CI failure

Link: https://github.com/cilium/cilium/actions/runs/5116398094/jobs/9198525758?pr=25743
Sysdump: [cilium-sysdumps.zip](https://github.com/cilium/cilium/files/11598759/cilium-sysdumps.zip)
Kind Logs: [kind-logs.zip](https://github.com/cilium/cilium/files/11598761/kind-logs.zip)

```
2023-05-30T00:59:16.0791494Z • [FAILED] [17.190 seconds]
2023-05-30T00:59:16.0797004Z [sig-apps] ReplicaSet [It] should serve a basic image on each replica with a public image  [Conformance]
2023-05-30T00:59:16.0800210Z test/e2e/apps/replica_set.go:111
2023-05-30T00:59:16.0800429Z
2023-05-30T00:59:16.0806465Z   Timeline >>
2023-05-30T00:59:16.0807247Z   STEP: Creating a kubernetes client @ 05/30/23 00:58:58.801
2023-05-30T00:59:16.0818090Z   May 30 00:58:58.801: INFO: >>> kubeConfig: /home/runner/work/cilium/cilium/_artifacts/kubeconfig.conf
2023-05-30T00:59:16.0820605Z   STEP: Building a namespace api object, basename replicaset @ 05/30/23 00:58:58.805
2023-05-30T00:59:16.0829388Z   STEP: Waiting for a default service account to be provisioned in namespace @ 05/30/23 00:58:58.928
2023-05-30T00:59:16.0838817Z   STEP: Waiting for kube-root-ca.crt to be provisioned in namespace @ 05/30/23 00:58:58.947
2023-05-30T00:59:16.0846998Z   May 30 00:58:58.971: INFO: Creating ReplicaSet my-hostname-basic-844c69c5-d5ca-4dbb-8fce-5500c01348cd
2023-05-30T00:59:16.0857832Z   May 30 00:58:59.026: INFO: Pod name my-hostname-basic-844c69c5-d5ca-4dbb-8fce-5500c01348cd: Found 0 pods out of 1
2023-05-30T00:59:16.0863733Z   May 30 00:59:04.096: INFO: Pod name my-hostname-basic-844c69c5-d5ca-4dbb-8fce-5500c01348cd: Found 1 pods out of 1
[...]

PR #26271, opened 2023-06-15
clustermesh: ensure that the status of the remote clusters controller is correcty reported

giorio94:
150de13bd6c6 ("clustermesh: delete stale node/service entries on reconnect/disconnect") and follow-ups changed the behavior of the controller which wraps the logic used to connect to the kvstore in a remote cluster. More specifically, we previously used to return from the controller DoFunc after completing the setup process (while executing in background the actual synchronization tasks). With that commit, instead, we don't return until the context is closed (which means that the connection needed to be restarted/stopped).

While this simplifies the implementation of the cleanup logic, the change turned out to cause issues in the controller health reporting logic. In particular, given that we don't return on success, a previous failure is never cleared out. Which means incorrect metrics and status reporting through the `cilium status` commands.

This PR fixes this issue reworking the logic so that we return from the controller DoFunc as soon as the initialization tasks completed, while executing the long running logic in background.

### DIFF
--- a/pkg/clustermesh/internal/config_test.go
+++ b/pkg/clustermesh/internal/config_test.go
@@ -34,8 +34,8 @@ var _ = Suite(&ClusterMeshTestSuite{})
 
 type fakeRemoteCluster struct{}
 
-func (*fakeRemoteCluster) Run(context.Context, kvstore.BackendOperations, *types.CiliumClusterConfig) error {
-	return nil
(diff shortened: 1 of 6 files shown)


YutaroHayakawa (approved):

giorio94:
ConformanceK8sKind hit https://github.com/cilium/cilium/issues/25758

PR #26277, opened 2023-06-15
Operator: Add missing observability for Azure API calls

hemanthmalla:
Currently we're not capturing response status and duration for some of the calls made to Azure. This commit calls `ObserveAPICall()` with status and duration for the operations missing them.

### DIFF
--- a/pkg/azure/api/api.go
+++ b/pkg/azure/api/api.go
@@ -383,7 +383,9 @@ func (c *Client) AssignPrivateIpAddressesVMSS(ctx context.Context, instanceID, v
 	var netIfConfig *compute.VirtualMachineScaleSetNetworkConfiguration
 
 	c.limiter.Limit(ctx, "VirtualMachineScaleSetVMs.Get")
+	sinceStart := spanstat.Start()
 	result, err := c.vmss.Get(ctx, c.resourceGroup, vmssName, instanceID, compute.InstanceViewTypesInstanceView)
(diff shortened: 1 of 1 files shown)


hemanthmalla:
@tommyp1ckles / @gandro Gentle nudge, can I get a quick review on this ? 
CI is green and the PR only adds some missing observability. Thank you.

pkg/azure/api/api.go
```diff
@@ -438,22 +440,27 @@ func (c *Client) AssignPrivateIpAddressesVMSS(ctx context.Context, instanceID, v
 	}
 
 	c.limiter.Limit(ctx, "VirtualMachineScaleSetVMs.Update")
+	sinceStart = spanstat.Start()
 	future, err := c.vmss.Update(ctx, c.resourceGroup, vmssName, instanceID, result)
 	if err != nil {
+		c.metricsAPI.ObserveAPICall("VirtualMachineScaleSetVMs.Update", deriveStatus(err), sinceStart.Seconds())
```
gandro: nit: Have you considered using defer for these?

gandro (approved):

tommyp1ckles (approved):
looks good - same question about the defer

pkg/azure/api/api.go
hemanthmalla: Good idea. Updated this.

PR #26281, opened 2023-06-15
Don't orphan CEPs when node IP changes.

### LINKED ISSUE #25094: Cilium can fail to take over ownership of CEP if internal node IP of the local node changes on Cilium restart
# Description

It is possible for a node to change internal node IPs (observed on GKE, with preemptive nodes) and the "sync-to-k8s-ciliumendpoint" controller will be forever stuck on the following error msg:

```
level=warning msg="could not take ownership of existing ciliumendpoint" containerID=afc34c5868 controller="sync-to-k8s-ciliumendpoint (1412)" datapathPolicyRevision=162 desiredPolicyRevision=162 endpointID=1412 error="endpoint sync cannot take ownership of CEP that is not local (\"10.168.0.21\")" identity=1377 ipv4=10.28.0.101 ipv6= k8sPodName=default/curl-84f95b5f4b-mbk9v subsys=endpointsynchronizer
```

See the offending code here: https://github.com/cilium/cilium/blob/404bf818518ab9f1961cfe67cd43a3b1bf040130/pkg/k8s/watchers/endpointsynchronizer.go#L358

# User impact

* If Cilium was installed with the Cilium CLI, then installation will fail and trigger a rollback because the CLI detects that controllers are failing.
* Cilium will never be able to manage these workloads because of the error bailing out

# Additional thoughts

* It seems that the above code relies on the CEP UID and if not found, falls back to the node IP check.
* It's possible that the CEP UID was not restored properly in this case
  * Has it ever worked and have we always just been falling back to node IPs because those don't often change? Not obvious from the logs.
* I think the cases covered in `Test_updateCEPUID` from my read don't cover the case where a node IP is mismatching

### LINKED ISSUE #26279: Changing Node IP orphans CEPs

### What happened?

Repro: 

1. In an environment where cilium serializes EPs into tmpfs (eg `/var/run` - which is currently the case on some cloud providers)
2. When node IP changes (typically on reboot or suspend/resume in the cloud)
3. CEPs are orphaned by the restarted cilium 
4. This results in ip cache corruption and a disruption of pod to pod traffic 

Root cause: 

The [endpointsynchronizer](https://github.com/cilium/cilium/blame/main/pkg/k8s/watchers/endpointsynchronizer.go) is responsible for updating the status of the CEP on cilium restart. 

When the cilium restart is because of a node restart which results in a node IP change, the endpoint synchronizer gets into a soft deadlock. 

* The [UID check](https://github.com/cilium/cilium/blob/main/pkg/k8s/watchers/endpointsynchronizer.go#L344) fails because it compares the CEP's UID with the previously serialized CEP's UID. Since the CEP was serialized into tmpfs, this state is lost. 
* The [Node IP check](https://github.com/cilium/cilium/blob/main/pkg/k8s/watchers/endpointsynchronizer.go#L357) fails because the node IP has changed. 

Instead of comparing the CEP node ip with the node's ip, [this](https://github.com/cilium/cilium/pull/26281) poc looks up the pod for the CEP and uses its host IP. The latter is updated by the kubelet and hence avoids the deadlock.

Opening this issue for discussion on what the right fix should be. 
@tommyp1ckles 

### Cilium Version

Tested on 
```
Client: 1.12.6 098dd3b4ba 2023-05-15T14:15:53-07:00 go version go1.20.1 linux/amd64
Daemon: 1.12.6 098dd3b4ba 2023-05-15T14:15:53-07:00 go version go1.20.1 linux/amd64
```
But should repro on head based on code inspection 

### Kernel Version

5.15.65+

### Kubernetes Version

```
Client Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.3", GitCommit:"25b4e43193bcda6c7328a6d147b1fb73a33f1598", GitTreeState:"clean", BuildDate:"2023-06-14T09:53:42Z", GoVersion:"go1.20.5", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v5.0.1
Server Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.3-gke.1000", GitCommit:"f14ee8d868842856ce474ba2ebadece18a6ca6cf", GitTreeState:"clean", BuildDate:"2023-03-30T09:31:19Z", GoVersion:"go1.19.7 X:boringcrypto", Compiler:"gc", Platform:"linux/amd64"}
```

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

bprashanth:
See https://github.com/cilium/cilium/issues/26279 for context.

      GitHub issue.

Fixes: #26279

```release-note
Handles nodeIP changes when CEPs are checkpointed to tmpfs and the nodeIP changes across a reboot. 
```

### DIFF
--- a/pkg/k8s/watchers/endpointsynchronizer.go
+++ b/pkg/k8s/watchers/endpointsynchronizer.go
@@ -339,23 +339,38 @@ func updateCEPUID(scopedLog *logrus.Entry, e *endpoint.Endpoint, localCEP *ciliu
 	// If the Endpoint already owns the CEP (by holding the matching CEP UID reference) then we don't have to
 	// worry about other ownership checks.
 	//
-	// This will cover cases such as if the NodeIP changes (as with a reboot). In which case we can safely
-	// take ownership and overwrite the CEPs status.
+	// This will cover cases such as if the NodeIP changes (as with a reboot).
+	// In which case we can safely take ownership and overwrite the CEPs
+	// status. However if the cilium endpoints are lost on restart (eg the
+	// state files were previously checkpointed into tmpfs) this check will
+	// fail and we will rely on the next check to prevent us from hijacking
+	// CEPs.
 	cepUID := e.GetCiliumEndpointUID()
 	if cepUID == localCEP.UID {
 		return nil
 	}
 
-	var nodeIP string
-	if netStatus := localCEP.Status.Networking; netStatus == nil {
-		return fmt.Errorf("endpoint sync cannot take ownership of CEP that has no nodeIP status")
-	} else {
-		nodeIP = netStatus.NodeIP
-	}
-
 	// We do not want to take ownership of CEPs created on other Nodes.
-	if nodeIP != node.GetCiliumEndpointNodeIP() {
-		return fmt.Errorf("endpoint sync cannot take ownership of CEP that is not local (%q)", nodeIP)
+	// However we can't directly compare the CEP node ip with the node, because
+	// the node ip can change, orphaning the CEP. So we retrieve the pod for
+	// the CEP and compare its node IP with that of the node. The kubelet on
+	// this node will update the pod object appropriately, allowing this check
+	// to eventually go through.
+	//
+	// The intent here is to check if a given pod is running on the same node
+	// this cilium is running on before taking over its CEP.
+	nodeIP := node.GetCiliumEndpointNodeIP()
+	pod := e.GetPod()
+	if pod == nil {
+		return fmt.Errorf("endpoint sync cannot take ownership of CEP: no pod")
+	}
+	podHostIP := pod.Status.HostIP
+	if podHostIP == "" {
+		return fmt.Errorf("endpoint sync cannot take ownership of CEP: no pod HostIP")
+	}
+	if podHostIP != nodeIP {
+		return fmt.Errorf("endpoint sync cannot take ownership of CEP that is not local: CEP's pod %q, pod's hostIP %q, cilium nodeIP %q)",
+			e.GetK8sPodName(), podHostIP, nodeIP)
 	}
 
 	// If the endpoint has a CEP UID, which does not match the current CEP, we cannot take
--- a/pkg/k8s/watchers/endpointsynchronizer_test.go
+++ b/pkg/k8s/watchers/endpointsynchronizer_test.go
@@ -5,6 +5,7 @@ package watchers
 
 import (
 	"fmt"
+	"net"
 	"testing"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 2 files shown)


pkg/k8s/watchers/endpointsynchronizer.go
```diff
@@ -339,23 +339,29 @@ func updateCEPUID(scopedLog *logrus.Entry, e *endpoint.Endpoint, localCEP *ciliu
...
-	if nodeIP != node.GetCiliumEndpointNodeIP() {
-		return fmt.Errorf("endpoint sync cannot take ownership of CEP that is not local (%q)", nodeIP)
+	// However we can't directly compare the CEP node ip with the node, because
+	// the node ip can change, orphaning the CEP. So we retrieve the pod for
+	// the CEP and compare its node IP with that of the node. The kubelet on
+	// this node will update the pod object appropriately, allowing this check
+	// to eventually go through.
+	// The intent here is to check if a given pod is running on the same node
+	// this cilium is running on before taking over its CEP.
+	nodeIP := node.GetCiliumEndpointNodeIP()
```
aojea, line 362: ~this function does not have any unit test~ I checked in the wrong place :)

, can we add one test case to cover this bug and avoid regression?

bprashanth: Yep, will add by EOD. Was just waiting for first pass comments.

bprashanth: Fixed and added unittests.

pkg/k8s/watchers/endpointsynchronizer.go
```diff
@@ -41,6 +41,21 @@ type EndpointSynchronizer struct {
...
+	NodeIP() string
+}
+
+// nodeAddressGetter implements NodeAddress.
+type nodeAddressGetter struct{}
+
+// NodeIP returns the node IP that will be referenced by CiliumEndpoints
+// with endpoints running on this node.
+func (n nodeAddressGetter) NodeIP() string {
+	return node.GetCiliumEndpointNodeIP()
```
aojea: I think that this can be mocked without need to do this dependency injection

`pkg/node/local_node_store.go`

> newTestLocalNodeStore

aojea:
@joestringer @christarazi  this is a pretty contained and small change that fix a bug we are hitting it in production, can you assign a reviewer?

joestringer:
@aojea is this ready for review? It's marked as draft. It looks like Chris is already assigned to review when it's ready for review.

aojea:
> @aojea is this ready for review? It's marked as draft. It looks like Chris is already assigned to review when it's ready for review.

upps sorry, didn't notice , @bprashanth ^^^

bprashanth:
@joestringer @christarazi it's ready for review. I removed "wip" from the title a while back, did i miss something else indicating the draft status?

bprashanth:
Ah apologies, I think I just had to click the button that said "this is not a draft". 
PATL.

pkg/k8s/watchers/endpointsynchronizer.go
```diff
@@ -334,28 +349,41 @@ func (epSync *EndpointSynchronizer) RunK8sCiliumEndpointSync(e *endpoint.Endpoin
...
+	// this node will update the pod object appropriately, allowing this check
+	// to eventually go through.
+	//
+	// The intent here is to check if a given pod is running on the same node
+	// this cilium is running on before taking over its CEP.
+	nodeIP := nadd.NodeIP()
+	if e.GetPod() == nil {
+		return fmt.Errorf("endpoint sync cannot take ownership of CEP: no pod")
+	}
+	podHostIP := e.GetPod().Status.HostIP
```
christarazi: Nit: 
```suggestion
	pod := e.GetPod() 
	if pod == nil {
		return fmt.Errorf("endpoint sync cannot take ownership of CEP: no pod")
	}
	podHostIP := pod.Status.HostIP
```

christarazi (approved):
LGTM pending on minor nit and Antonio's comment about not needing an extra dependency injection

pkg/k8s/watchers/endpointsynchronizer.go
```diff
@@ -339,23 +339,37 @@ func updateCEPUID(scopedLog *logrus.Entry, e *endpoint.Endpoint, localCEP *ciliu
...
-		nodeIP = netStatus.NodeIP
-	}
-
 	// We do not want to take ownership of CEPs created on other Nodes.
-	if nodeIP != node.GetCiliumEndpointNodeIP() {
-		return fmt.Errorf("endpoint sync cannot take ownership of CEP that is not local (%q)", nodeIP)
+	// However we can't directly compare the CEP node ip with the node, because
+	// the node ip can change, orphaning the CEP. So we retrieve the pod for
+	// the CEP and compare its node IP with that of the node. The kubelet on
+	// this node will update the pod object appropriately, allowing this check
```
tommyp1ckles, line 357: Just a thought, what are the core assumptions of how the Kubelet works that we'd be basing this on.  Are we confident that won't change in future releases?

pkg/k8s/watchers/endpointsynchronizer.go
```diff
@@ -339,23 +339,37 @@ func updateCEPUID(scopedLog *logrus.Entry, e *endpoint.Endpoint, localCEP *ciliu
 	// If the Endpoint already owns the CEP (by holding the matching CEP UID reference) then we don't have to
 	// worry about other ownership checks.
 	//
-	// This will cover cases such as if the NodeIP changes (as with a reboot). In which case we can safely
-	// take ownership and overwrite the CEPs status.
+	// This will cover cases such as if the NodeIP changes (as with a reboot).
+	// In which case we can safely take ownership and overwrite the CEPs
+	// status. However if the cilium endpoints were previously checkpointed
+	// into tmpfs this check will fail and we will rely on the next check to
```
tommyp1ckles: nit: For future readers, we could make this comment more general: any situation where Cilium might lose its state after a restart would cause similar problems.

pkg/k8s/watchers/endpointsynchronizer.go
bprashanth: The main assumption is that when/if the host ip changes, the kubelet will take notice and populate the updated hostIP in the Kubernetes pod object. Since hostIP is part of the api, I feel like if this isn't happening it would be a kubelet bug.

bprashanth:
@tommyp1ckles udpated comments, PTAL. I think I still need maintainer approval.

ldelossa:
All required tests passing. Merging.

pkg/k8s/watchers/endpointsynchronizer.go
aojea: the kubelet HostIP value is set at startup and inmutable after that, in order to change the IP, kubelet has to restart ref https://github.com/kubernetes/kubernetes/pull/93898

This affects hostNetwork pods, that inherit their PodIPs from the kubelet HostIPs.

If kubelet restarts with a new IP, all hostNetwork pods will have the new IP in the pod.status.PodIPs and pod.status.HostIP and the non-hostNetwork pods will have the new IP in pod.status.HostIP

julianwiedmann:
It's been a while - I'll go ahead and drop the `needs-backport` for `v1.13`. Please speak up if anyone still wants to tackle the backport.

PR #26289, opened 2023-06-16
cilium: nodeId, node update may remove IPAddresses so sync nodeMap also

jrfastab:
NodeID list can get out of sync with reality when nodeUpdates  remove old addresses. The result is we can hit the default drop policy in the xfrm policy handling. See patch for details.


jrfastab:
superceded

PR #26291, opened 2023-06-16
clustermesh: unbreak test

julianwiedmann:
Error: pkg/clustermesh/remote_cluster_test.go:155:54: too many arguments in call to newGlobalServiceCache
	have (string, string)
	want ("github.com/cilium/cilium/pkg/metrics/metric".Gauge)

PR #26294, opened 2023-06-16
clustermesh: unbreak test

giorio94:
Due to a merge race, one of the clustermesh tests is failing to compile:

```
Error: pkg/clustermesh/remote_cluster_test.go:155:54: too many arguments in call to newGlobalServiceCache
have (string, string)
want ("github.com/cilium/cilium/pkg/metrics/metric".Gauge)
```

Additionally, it seems that the newly introduced tests are badly interacting with each other (they use the same shared etcd instance), causing frequent flakes. Let's disable them for the moment to unblock main, and in the meanwhile I'll try to figure out how to make them more resilient.

Supersedes: https://github.com/cilium/cilium/pull/26291 (thanks @julianwiedmann)

### DIFF
--- a/pkg/clustermesh/clustermesh_test.go
+++ b/pkg/clustermesh/clustermesh_test.go
@@ -102,9 +102,7 @@ func (s *ClusterMeshTestSuite) TestClusterMesh(c *C) {
 
 	kvstore.SetupDummy("etcd")
 	defer func() {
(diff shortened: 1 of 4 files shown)


YutaroHayakawa (approved):

giorio94:
Integration tests passed. Marking as ready to merge to unblock main

PR #26301, opened 2023-06-16
bpf: tests: pktgen infra for tunneling + GENEVE-DSR test

julianwiedmann:
Add the necessary bits so that pktgen can build tunnel encapsulation headers (UDP, VXLAN, GENEVE). Exercise it with some initial tests - more to come.

### DIFF
--- a/bpf/tests/pktgen.h
+++ b/bpf/tests/pktgen.h
@@ -17,6 +17,7 @@
 #include <linux/if_arp.h>
 #include <linux/if_ether.h>
 #include <linux/tcp.h>
+#include <linux/udp.h>
 
 /* A collection of pre-defined Ethernet MAC addresses, so tests can reuse them
  * without having to come up with custom addresses.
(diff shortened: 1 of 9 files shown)


bpf/tests/pktgen.h
```diff
@@ -834,7 +886,20 @@ void pktgen__finish(const struct pktgen *builder)
...
+			 */
+			if (layer_off >= MAX_PACKET_OFF - sizeof(struct udphdr))
+				return;
+
+			udp_layer = ctx_data(builder->ctx) + layer_off;
+			if ((void *)udp_layer + sizeof(struct udphdr) >
+			    ctx_data_end(builder->ctx))
+				return;
+
+			if (i + 1 >= PKT_BUILDER_LAYERS)
```
gentoo-root: Is it needed? Won't it return because the loop finishes?

bpf/tests/pktgen.h
```diff
@@ -914,6 +1042,36 @@ void pktgen__finish(const struct pktgen *builder)
...
+			geneve_layer = ctx_data(builder->ctx) + layer_off;
+			if ((void *)geneve_layer + sizeof(struct genevehdr) >
+			    ctx_data_end(builder->ctx))
+				return;
+
+			if (i + 1 >= PKT_BUILDER_LAYERS)
+				return;
+			break;
+
+			switch (builder->layers[i + 1]) {
```
gentoo-root, line 1121: Looks like this code is unreachable because of the break above.

nbusseneau (approved):

bpf/tests/pktgen.h
julianwiedmann: :+1: copy & pasted from other cases that require the check before accessing the next layer, but for UDP it's not needed.

bpf/tests/pktgen.h
julianwiedmann: ha, thanks for catching that. I wonder if that explains some of the failures I was seeing in my test backlog ...

gentoo-root (approved):

PR #26312, opened 2023-06-16
.github/workflows: let renovate update kind

tklauser:
Add required renovate annotations such that `kind_version` will be updated automatically in GH action workflows.

### DIFF
--- a/.github/workflows/conformance-clustermesh-v1.11.yaml
+++ b/.github/workflows/conformance-clustermesh-v1.11.yaml
@@ -64,6 +64,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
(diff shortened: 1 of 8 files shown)


aanm (approved):
Nice! :partying_face: Is it difficult to add the same [here](https://github.com/cilium/little-vm-helper-images/blob/main/_data/bootstrap/kind.sh#L6)?

viktor-kurchenko (approved):

tklauser:
> Nice! partying_face Is it difficult to add the same [here](https://github.com/cilium/little-vm-helper-images/blob/main/_data/bootstrap/kind.sh#L6)?

We'd need to set up renovate for that repo but then it should be too difficult to add.

PR #26325, opened 2023-06-16
bgpv1: set N-bit in graceful restart

harsimran-pabla:
Set N-bit in graceful restart negotiation, so the peer can start graceful restart helper mode if HoldTime expires or there is a session reset because of configuration change. Further details in [RFC-8538](https://www.rfc-editor.org/rfc/rfc8538.html).

Open Message from Cilium : 

image

```release-note
BGPv1: Set N-bit in graceful restart capability negotiation.
```


rastislavs (approved):
LGTM

danehans:
[RFC-8538](https://www.rfc-editor.org/rfc/rfc8538.html) provides a nice summary of graceful restart negotiation:
>When a BGP session is reset, both speakers operate as "Receiving Speakers"
   according to [[RFC4724](https://www.rfc-editor.org/rfc/rfc4724)], meaning they retain each other's routes.
   This is also true for HOLDTIME expiration.  The functionality can be
   defeated by sending a BGP Cease NOTIFICATION message with the Hard
   Reset subcode.  If a Hard Reset is used, a full session reset is
   performed.

LGTM

PR #26327, opened 2023-06-16
Add checks to avoid use of logrus WithFields function in hot paths

learnitall:
This commit modifies identified hot paths that contain the use of logrus' WithFields function, to add extra conditions to only call this expensive function as necessary. The most common case was a scoped logger created using WithFields that is only used to emit debug messages.

These hot paths were identified using profiles taken from cilium agents in a 100 node ClusterLoader2 scale test. For example, the function `pkg/endpoint.(*Endpoint).removeDirectory` had a wasted call to WithFields which accounted for 1.5% of overall memory allocations and 0.13% of overall CPU time, as it is called during endpoint regeneration.

I believe this needs the following labels:

* `release-note/misc`
* `kind/bug`

### DIFF
--- a/pkg/endpoint/bpf.go
+++ b/pkg/endpoint/bpf.go
@@ -30,6 +30,7 @@ import (
 	"github.com/cilium/cilium/pkg/identity"
 	"github.com/cilium/cilium/pkg/labels"
 	"github.com/cilium/cilium/pkg/loadinfo"
+	"github.com/cilium/cilium/pkg/logging"
 	"github.com/cilium/cilium/pkg/logging/logfields"
 	"github.com/cilium/cilium/pkg/maps/bwmap"
 	"github.com/cilium/cilium/pkg/maps/ctmap"
@@ -631,11 +632,14 @@ func (e *Endpoint) regenerateBPF(regenContext *regenerationContext) (revnum uint
 func (e *Endpoint) realizeBPFState(regenContext *regenerationContext) (compilationExecuted bool, err error) {
 	stats := &regenContext.Stats
 	datapathRegenCtxt := regenContext.datapathRegenerationContext
+	debugEnabled := logging.CanLogAt(e.getLogger().Logger, logrus.DebugLevel)
 
-	e.getLogger().WithField(fieldRegenLevel, datapathRegenCtxt.regenerationLevel).Debug("Preparing to compile BPF")
+	if debugEnabled {
+		e.getLogger().WithFields(logrus.Fields{fieldRegenLevel: datapathRegenCtxt.regenerationLevel}).Debug("Preparing to compile BPF")
+	}
 
 	if datapathRegenCtxt.regenerationLevel > regeneration.RegenerateWithoutDatapath {
-		if e.Options.IsEnabled(option.Debug) {
+		if debugEnabled {
 			debugFunc := log.WithFields(logrus.Fields{logfields.EndpointID: e.StringID()}).Debugf
 			ctx, cancel := context.WithCancel(regenContext.parentContext)
 			defer cancel()
@@ -668,7 +672,7 @@ func (e *Endpoint) realizeBPFState(regenContext *regenerationContext) (compilati
 			return compilationExecuted, err
 		}
 		e.bpfHeaderfileHash = datapathRegenCtxt.bpfHeaderfilesHash
-	} else {
+	} else if debugEnabled {
 		e.getLogger().WithField(logfields.BPFHeaderfileHash, datapathRegenCtxt.bpfHeaderfilesHash).
 			Debug("BPF header file unchanged, skipping BPF compilation and installation")
 	}
@@ -755,7 +759,9 @@ func (e *Endpoint) runPreCompilationSteps(regenContext *regenerationContext, rul
 			return false, fmt.Errorf("Unable to write header file: %s", err)
 		}
 
-		log.WithField(logfields.EndpointID, e.ID).Debug("Skipping bpf updates due to dry mode")
+		if logging.CanLogAt(log.Logger, logrus.DebugLevel) {
+			log.WithField(logfields.EndpointID, e.ID).Debug("Skipping bpf updates due to dry mode")
+		}
 		return false, nil
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/endpoint/bpf.go
```diff
@@ -681,11 +682,14 @@ func (e *Endpoint) regenerateBPF(regenContext *regenerationContext) (revnum uint
 func (e *Endpoint) realizeBPFState(regenContext *regenerationContext) (compilationExecuted bool, err error) {
 	stats := &regenContext.Stats
 	datapathRegenCtxt := regenContext.datapathRegenerationContext
+	debugEnabled := logging.CanLogAt(e.getLogger().Logger, logrus.DebugLevel)
```
christarazi, line 635: Shouldn't this simply be 

```suggestion
	debugEnabled := e.Options.IsEnabled(option.Debug)
```

since that's the condition that it's replacing?

christarazi (approved):

christarazi:
(Accidentally submitted review too soon...but below is what I was going to say)

The change looks good to me. I wonder how we can prevent the need to do PRs like these in the future. At some point I'd imagine we'll have some performance testing, but that might be too late in the feedback cycle to inform whether new debug logs caused a performance impact. Food for thought I guess :).

pkg/endpoint/bpf.go
learnitall: I think either option works, but IMO I prefer using `logging.CanLogAt` because it checks the logger itself. If we check the options instead, it opens the door for a potential bug in the case that the logger is at a different level than what the options reflect. The function still should be efficient enough, as it checks the logger's level directly using an atomic load.

learnitall:
Definitely good for thought. Ideally, this will be a one-time PR for these changes. For 1.15, I am going to work on a CFP to explore some options to address the overhead we see here.

pkg/endpoint/bpf.go
christarazi: So the difference between `e.Options.IsEnabled(option.Debug)` I believe is that it checks the runtime options of a specific endpoint (i.e. https://github.com/cilium/cilium/blob/1a45357601326e10d86bcf0d1e11a3d982f359fe/pkg/option/endpoint.go#L10), whereas `logging.CanLogAt` checks the generic logging level for all of the Agent. A specific endpoint can toggle debug mode only for itself, while the rest of the Agent can not be in debug mode. This can be done via `cilium endpoint config <id> debug=true`, hence why I left the suggestion in my previous comment.

ysksuzuki (approved):
The changes look good to me. (The conflicts need to be resolved though)

Can I see the pprof result somewhere that indicates these paths which call logrus' WithFields function are expensive?

pkg/endpoint/bpf.go
learnitall: > whereas logging.CanLogAt checks the generic logging level for all of the Agent.

This isn't the case. `logging.CanLogAt` and `option.Debug` should always be in sync. `logging.CanLogAt` checks the logging level for the passed-in logger, it does not take into account the level for the Agent. Here's its definition:

```go
// CanLogAt returns whether a log message at the given level would be
// logged by the given logger.
func CanLogAt(logger *logrus.Logger, level logrus.Level) bool {
	return GetLevel(logger) >= level
}

// GetLevel returns the log level of the given logger.
func GetLevel(logger *logrus.Logger) logrus.Level {
	return logrus.Level(atomic.LoadUint32((*uint32)(&logger.Level)))
}
```

An endpoint's internal logger should have its level updated by `UpdateLogger` if it changes at runtime.

learnitall:
I haven't found a great way to share the flamegraphs yet, as go's pprof tool doesn't have a method for generating svg flamegraphs. I've attached tarballs of the CPU, Heap and Alloc 

profiles that were taken during the course of the 100 node test. After extracting each, you can run:

`go tool pprof -http :8080 *(CPU, Heap or Alloc)*`

To view CPU, Heap or Alloc flamegraphs respectively.

[alloc.tar.gz](https://github.com/cilium/cilium/files/11887467/alloc.tar.gz)
[cpu.tar.gz](https://github.com/cilium/cilium/files/11887466/cpu.tar.gz)
[heap.tar.gz](https://github.com/cilium/cilium/files/11887472/heap.tar.gz)

pkg/endpoint/bpf.go
christarazi: I'm not sure I'm following. `cilium endpoint config <id>` is handled by https://github.com/cilium/cilium/blob/1a45357601326e10d86bcf0d1e11a3d982f359fe/daemon/cmd/endpoint.go#L862. If you follow that code path, there's no call to `UpdateLogger` anywhere.

learnitall: Here's the path:

https://github.com/cilium/cilium/blob/1a45357601326e10d86bcf0d1e11a3d982f359fe/daemon/cmd/endpoint.go#L862

https://github.com/cilium/cilium/blob/1a45357601326e10d86bcf0d1e11a3d982f359fe/daemon/cmd/endpoint.go#L819-L829

https://github.com/cilium/cilium/blob/1a45357601326e10d86bcf0d1e11a3d982f359fe/pkg/endpoint/endpoint.go#L961-L988

https://github.com/cilium/cilium/blob/1a45357601326e10d86bcf0d1e11a3d982f359fe/pkg/endpoint/policy.go#L276-L281

https://github.com/cilium/cilium/blob/1a45357601326e10d86bcf0d1e11a3d982f359fe/pkg/endpoint/endpoint.go#L691-L695

https://github.com/cilium/cilium/blob/1a45357601326e10d86bcf0d1e11a3d982f359fe/pkg/endpoint/log.go#L97-L100

christarazi: I think this is ok in the end because ultimately, we are checking the endpoint's logger instance in 

```
debugEnabled := logging.CanLogAt(e.getLogger().Logger, logrus.DebugLevel)
```

I had missed this subtle detail previously and thought that we were considering `debugEnabled` if the global logger level was debug, rather than if the endpoint itself was configured for debug. :+1:

PR #26335, opened 2023-06-17
Chores: added ability to remove iptables

### LINKED ISSUE #25804: Minor improvement: automatically cleanup stale AWS-CNI iptables rules in nodeinit
In the [install documentation for EKS](https://github.com/cilium/cilium/blob/v1.13/Documentation/installation/k8s-install-helm.rst?plain=1#L208), it has you manually remove some stale AWS iptables rules. *Not* removing these rules results in subtle and random connectivity issues, most notably around host-network processes accessing the pod network. This typically results in a failure to reach the health check.

So, when we're sure these aren't in use, we should remove these in the [nodeinit script](https://github.com/cilium/cilium/blob/v1.13/install/kubernetes/cilium/files/nodeinit/startup.bash).

The challenge is knowing these rules are not in use. From reading the install documentation, I don't see a reliable way to detect it, so we'll have to add another helm value. Probably something like `nodeinit.removeAWSCNI`.

swastik959:

      GitHub issue.

Fixes: #25804

```release-note
```

### DIFF
--- a/install/kubernetes/cilium/files/nodeinit/startup.bash
+++ b/install/kubernetes/cilium/files/nodeinit/startup.bash
@@ -23,6 +23,11 @@ if ip link show cbr0; then
 fi
 {{- end }}
 
+{{- if .Values.nodeinit.removeAWSCNI }}
+echo "Removing stale AWS iptables rules..."
+sudo iptables -L
+{{- end }}
+
 {{- if .Values.nodeinit.reconfigureKubelet }}
 # Check if we're running on a GKE containerd flavor as indicated by the presence
 # of the '--container-runtime-endpoint' flag in '/etc/default/kubelet'.
--- a/install/kubernetes/cilium/values.yaml
+++ b/install/kubernetes/cilium/values.yaml
@@ -67,6 +67,7 @@ serviceAccounts:
     name: cilium-nodeinit
     automount: true
     annotations: {}
+    removeAWSCNI: false
   envoy:
     create: true
     name: cilium-envoy


nebril (commented):
Thanks for the PR!

I am not sure if these changes are actually required here, so let's wait for @squeed to answer to my comment in original issue: https://github.com/cilium/cilium/issues/25804#issuecomment-1597015255

install/kubernetes/cilium/files/nodeinit/startup.bash
```diff
@@ -23,6 +23,11 @@ if ip link show cbr0; then
 fi
 {{- end }}
 
+{{- if .Values.nodeinit.removeAWSCNI }}
+echo "Removing stale AWS iptables rules..."
+sudo iptables -L
```
ti-mo, line 28: iirc this code executes within the context of pid1 on the host, so the availability of `sudo` here is not a given, and there's no point in using it. Also, how does `iptables -L` remove iptables rules?

swastik959: @ti-mo sorry I think I got confused can you tell me if there is any command to remove the iptables

nebril:
@swastik959 please take a look at discussion in original issue and this comment, I think you should simply change the condition on this line https://github.com/cilium/cilium/blob/v1.13/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml#L227 to one described in this comment: https://github.com/cilium/cilium/issues/25804#issuecomment-1597338651 . `poststart-eni.bash` is a no-op on non-AWS clusters, so I think it's going to work fine.

squeed:
So, it seems like you'd like the option to flush all iptables rules on startup. That is... dangerous... but I see how it could be useful. My suggestions

1. Add a Helm option, `nodeinit.FlushIptables`, disabled by default, which warns of the potential dangers
2. Update the PR to use the correct iptables commands. The easiest way to flush all tables is to use `iptables-restore`, please experiment with that
3. Make sure you clear ip6tables as well :-)

nebril:
@squeed looking at the script that we are currently using, we would only run `iptables-restore` if there are rules matching `AWS-SNAT-CHAIN`: https://github.com/cilium/cilium/blob/v1.13/install/kubernetes/cilium/files/agent/poststart-eni.bash#L14

Does this behaviour really needs a flag to gate it?

christarazi:
Something to keep in mind here, but does it really make sense to have Cilium clear iptables that it did not create and therefore does not manage?

nebril:
@christarazi we already do that to mitigate an issue caused by these iptables rules, I feel like the OG issue is about making sure that we do it in every affected environment.

ldelossa:
@swastik959 if you still intend to push this PR forward, please include a release note on your original PR comment.

squeed:
> Does this behaviour really need a flag to gate it?

@nebril there might be cases where eni mode is disabled, but there still are stale aws rules lying around. Alternatively, we may be chaining off of AWS and thus should *not* touch any AWS rules. At this point, none of these options are good.

Also, we should not clear iptables if we just restart cilium, so anything like this should go in nodeinit, not in the agent.

nebril:
@squeed the reason this behaviour was moved to agent was that node-init didn't work on Bottlerocket OS :(.

I am not opposed to adding the flag, but it's one more config value to keep track off and if we can figure out a way without it, I would strongly prefer it.

> there might be cases where eni mode is disabled

You mean cases where eni mode is disabled/enabled during cluster lifecycle?

christarazi:
It's not clear how this PR helps remove iptables rules. Moving to draft. Please feel free to continue working on this PR and mark it ready to review when it's ready.

PR #26336, opened 2023-06-17
datapath: bigtcp: Fix the IPv4 BIG TCP may not work

haiyuewa:
The kernel will also update the IPv4 GRO/GSO setting if the new value of
"gso/gro_max_size" isn't greater than 65536:
  1. ip link set dev ens787np0 gso_max_size 130000
  2. ip link set dev ens787np0 gso_ipv4_max_size 128000
  3. ip -d -j link show dev ens787np0 | jq -c '.[0].gso_ipv4_max_size'
     128000
  4. ip link set dev ens787np0 gso_max_size 65536
  5. ip -d -j link show dev ens787np0 | jq -c '.[0].gso_ipv4_max_size'
     65536

Cilium will call IPv4 GRO/GSO setup firstly, then call IPv6. To handle the
above issue, setup the IPv6 GRO/GSO firstly, then IPv4 GRO/GSO.

### DIFF
--- a/pkg/datapath/linux/bigtcp/bigtcp.go
+++ b/pkg/datapath/linux/bigtcp/bigtcp.go
@@ -219,19 +219,9 @@ func InitBIGTCP(bigTCPConfig *Configuration) {
 		// modification and end up with an error, so best to be conservative
 		// and always reset all on error
 		modifiedDevices = append(modifiedDevices, device)
-		if haveIPv4 {
-			err = setGROGSOIPv4MaxSize(device,
-				bigTCPConfig.groIPv4MaxSize, bigTCPConfig.gsoIPv4MaxSize)
-			if err != nil {
-				log.WithError(err).WithField("device", device).Warnf("Could not modify IPv4 gro_max_size and gso_max_size%s",
-					disableMsg)
-				option.Config.EnableIPv4BIGTCP = false
-				option.Config.EnableIPv6BIGTCP = false
-				break
-			}
-			log.WithField("device", device).Infof("Setting IPv4 gso_max_size to %d and gro_max_size to %d",
-				bigTCPConfig.gsoIPv4MaxSize, bigTCPConfig.groIPv4MaxSize)
-		}
+		// For compatibility, the Kernel will also update the net device's "gso/gro_ipv4_max_size",
+		// if the new size of "gso/gro_max_size" isn't greater than 64KB.
+		// So it needs to set the IPv6 firstly, otherwide, the IPv4 BIG TCP value will be reset.
 		if haveIPv6 {
 			err = setGROGSOIPv6MaxSize(device,
 				bigTCPConfig.groIPv6MaxSize, bigTCPConfig.gsoIPv6MaxSize)
@@ -244,6 +234,19 @@ func InitBIGTCP(bigTCPConfig *Configuration) {
 			}
 			log.WithField("device", device).Infof("Setting IPv6 gso_max_size to %d and gro_max_size to %d",
 				bigTCPConfig.gsoIPv6MaxSize, bigTCPConfig.groIPv6MaxSize)
+		}
+		if haveIPv4 {
+			err = setGROGSOIPv4MaxSize(device,
+				bigTCPConfig.groIPv4MaxSize, bigTCPConfig.gsoIPv4MaxSize)
+			if err != nil {
+				log.WithError(err).WithField("device", device).Warnf("Could not modify IPv4 gro_max_size and gso_max_size%s",
+					disableMsg)
+				option.Config.EnableIPv4BIGTCP = false
+				option.Config.EnableIPv6BIGTCP = false
+				break
+			}
+			log.WithField("device", device).Infof("Setting IPv4 gso_max_size to %d and gro_max_size to %d",
+				bigTCPConfig.gsoIPv4MaxSize, bigTCPConfig.groIPv4MaxSize)
 		}
 	}
 


pkg/datapath/linux/bigtcp/bigtcp.go
```diff
@@ -219,31 +219,31 @@ func InitBIGTCP(bigTCPConfig *Configuration) {
 		// modification and end up with an error, so best to be conservative
 		// and always reset all on error
 		modifiedDevices = append(modifiedDevices, device)
-		if haveIPv4 {
-			err = setGROGSOIPv4MaxSize(device,
-				bigTCPConfig.groIPv4MaxSize, bigTCPConfig.gsoIPv4MaxSize)
+		if haveIPv6 {
```
PlatformLC, line 225: nit: should we leave an comment for this tricky order directly in code? Or hide such logic in setGROGSOIPv6MaxSize to make sure IPv6 operation will not impact IPv4 values.

pkg/datapath/linux/bigtcp/bigtcp.go
```diff
@@ -219,31 +219,31 @@ func InitBIGTCP(bigTCPConfig *Configuration) {
...
 				option.Config.EnableIPv4BIGTCP = false
 				option.Config.EnableIPv6BIGTCP = false
 				break
 			}
-			log.WithField("device", device).Infof("Setting IPv6 gso_max_size to %d and gro_max_size to %d",
-				bigTCPConfig.gsoIPv6MaxSize, bigTCPConfig.groIPv6MaxSize)
+			log.WithField("device", device).Infof("Setting IPv4 gso_max_size to %d and gro_max_size to %d",
+				bigTCPConfig.gsoIPv4MaxSize, bigTCPConfig.groIPv4MaxSize)
 		}
 	}
```
PlatformLC, line 252: nit: should we also modify the IPv4/IPv6 order for below err handling?

pkg/datapath/linux/bigtcp/bigtcp.go
haiyuewa: This will make 'setGROGSOIPv6MaxSize' not so clean, by my understanding. And this logic is in kernel, git commit message should be fine, no bugs, then peope will be fine about this. ; - )

pkg/datapath/linux/bigtcp/bigtcp.go
haiyuewa: 64KB will be OK. Just less code change to make things work.

pkg/datapath/linux/bigtcp/bigtcp.go
haiyuewa: After thinking more, looks like make sense to add some comments. Thanks!

test/bigtcp/test.sh
```diff
@@ -43,7 +43,7 @@ helm install cilium ${HELM_CHART_DIR} \
 kubectl -n kube-system rollout status ds/cilium --timeout=5m
 
 # check if BIG TCP is initialized
-kubectl -n kube-system logs ds/cilium 2>&1 | grep "Setting up IPv6 BIG TCP"
+kubectl -n kube-system logs ds/cilium 2>&1 | grep "Setting up BIG TCP"
```
borkmann: thanks for the fix. this could probably be submitted as a separate PR and we can already merge it.

pkg/datapath/linux/bigtcp/bigtcp.go
borkmann: Thanks for the fix, @haiyuewa! Were you able to test it on ice with localy tweaking default{GSO,GRO}MaxSize? Would be curious what numbers you see before/after on ice.

Do you plan to follow-up with a knob for tweaking internal default{GSO,GRO}MaxSize?

test/bigtcp/test.sh
haiyuewa: Done: https://github.com/cilium/cilium/pull/26377

pkg/datapath/linux/bigtcp/bigtcp.go
haiyuewa: I will re-submit with new logs after tweaking the MAX size.

And yes, if you don't start to code it. ;-) It should be about tweaking internal bigTCP{GRO/GSO}MaxSize, right ?

borkmann: Back in IPv6 testing, @NikAleksandrov found the 196608 to be a sweet spot for mlx5.

I think for ice, we could read out link.Attrs().TSOMaxSize and probably determine this automatically even without knob.

For example, if the default{GSO,GRO}MaxSize fails to be set, then if link.Attrs().TSOMaxSize is still larger than 65536 we know that the device supports BIG TCP.

So if larger than 65536, we could take min(default{GSO,GRO}MaxSize, link.Attrs().TSOMaxSize) to figure it out.

borkmann: @haiyuewa if you could help with testing on ice, I could code this up. But lets get first your fix from here in, so that we don't run into merge conflict.

haiyuewa: @borkmann Looks like more things need to be considered. Sure, I can help on test it.

borkmann: > Looks like more things need to be considered.

Could you elaborate?

borkmann (approved):
Awesome, looks good to me! Thanks!

pkg/datapath/linux/bigtcp/bigtcp.go
haiyuewa: > > Looks like more things need to be considered.
> 
> Could you elaborate?

I just thought it needs to do more things about "Back in IPv6 testing, @NikAleksandrov found the 196608 to be a sweet spot for mlx5." It's my misunderstanding.

Please let me know if the code is ready for ice test. Thanks.

borkmann: Btw, I just checked the veth side, and here we should be good as v6 is adjusted before v4: https://github.com/cilium/cilium/pull/26172/commits/0ab201b39b55ab36e6bd6ba98d60f5da851afe73

ldelossa (approved):

PR #26341, opened 2023-06-19
cilium: nodeId tracking fixes for ipcache and upsert

jrfastab:
This adds fixes for nodeID tracking to correctly trace Node updates where newNode and oldNode have different IP lists. This can happen when a IP is removed from the oldNode or otherwise changed.

The other fix is to correctly handle IP addresses being sent to the IPCache through kvstore and k8s watchers. In this case we were leaking idnodes because we never removed allocated ids through this path. At first this might simply sound like a small memory leak, but then those IPs may get reused by the IPAM and we have incorrect nodeID<->IP mappings that can result in incorrectly setting the skb->mark in the datapath, finally resulting in Policy Block policy dropping the traffic.

So too fix we refcnt the nodeIDs and add code to track IP deletes through ipcache so that we correctly release the nodeID<->IP mapping but also ensure we only release it when no more IPs are mapped to it.

TBD: Still need a fix for cilum agent restart the DumpListener code in the ipcache needs to be smart enough to correctly sync and refcnt the nodeID map on init. But, also this dump callback can be called through other non-init paths so in those cases needs to ensure we don't increment the refcnt which would pin the node id mapping in the map. I can either add that fix here or in another PR.

### DIFF
--- a/pkg/datapath/fake/node.go
+++ b/pkg/datapath/fake/node.go
@@ -55,6 +55,13 @@ func (n *FakeNodeHandler) AllocateNodeID(_ net.IP) uint16 {
 	return 0
 }
 
+func (n *FakeNodeHandler) DeallocateNodeID(_ net.IP) {
+}
+
+func (n *FakeNodeHandler) GetNodeID(_ net.IP) uint16 {
+	return 0
+}
+
 func (n *FakeNodeHandler) DumpNodeIDs() []*models.NodeID {
 	return nil
 }
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -49,6 +49,20 @@ type NeighLink struct {
 	Name string `json:"link-name"`
 }
 
+type nodeIDSource uint16
+
+const (
+	NewNodeCreatedID = iota
+	DatapathRestoreCreatedID
+	IPCacheCreatedID
+)
+
+type nodeID struct {
+	id     uint16
+	refcnt uint32
+	source nodeIDSource
+}
+
 type linuxNodeHandler struct {
 	mutex                lock.Mutex
 	isInitialized        bool
@@ -69,7 +83,7 @@ type linuxNodeHandler struct {
 	// Pool of available IDs for nodes.
 	nodeIDs idpool.IDPool
 	// Node-scoped unique IDs for the nodes.
-	nodeIDsByIPs map[string]uint16
+	nodeIDsByIPs map[string]nodeID
 
 	ipsecMetricCollector prometheus.Collector
 }
@@ -93,7 +107,7 @@ func NewNodeHandler(datapathConfig DatapathConfiguration, nodeAddressing datapat
 		neighByNextHop:         map[string]*netlink.Neigh{},
 		neighLastPingByNextHop: map[string]time.Time{},
 		nodeIDs:                idpool.NewIDPool(minNodeID, maxNodeID),
-		nodeIDsByIPs:           map[string]uint16{},
+		nodeIDsByIPs:           map[string]nodeID{},
 		ipsecMetricCollector:   ipsec.NewXFRMCollector(),
 	}
 }
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -1123,6 +1123,8 @@ func (n *linuxNodeHandler) nodeUpdate(oldNode, newNode *nodeTypes.Node, firstAdd
 		newIP6                 = newNode.GetNodeIP(true)
 		oldKey, newKey         uint8
 		isLocalNode            = false
+		oldNodeIPAddrs         []string
+		newNodeIPAddrs         []string
 	)
 
 	if oldNode != nil {
@@ -1131,12 +1133,23 @@ func (n *linuxNodeHandler) nodeUpdate(oldNode, newNode *nodeTypes.Node, firstAdd
 		oldIP4 = oldNode.GetNodeIP(false)
 		oldIP6 = oldNode.GetNodeIP(true)
 		oldKey = oldNode.EncryptionKey
+
+		// Delete the old node IP addresses from nodeID map if they are dropped by the update.
+		for _, address := range oldNode.IPAddresses {
+			oldNodeIPAddrs = append(oldNodeIPAddrs, address.IP.String())
+		}
+	}
+
+	for _, address := range newNode.IPAddresses {
+		newNodeIPAddrs = append(newNodeIPAddrs, address.IP.String())
 	}
 
 	if n.nodeConfig.EnableIPSec && !n.nodeConfig.EncryptNode {
 		n.enableIPsec(newNode)
 		newKey = newNode.EncryptionKey
 	}
+
+	n.deleteNodeIPs(oldNodeIPAddrs, newNodeIPAddrs)
 
 	if n.enableNeighDiscovery && !newNode.IsLocal() {
 		// Running insertNeighbor in a separate goroutine relies on the following
--- a/pkg/datapath/linux/node_ids.go
+++ b/pkg/datapath/linux/node_ids.go
@@ -43,7 +43,9 @@ func (n *linuxNodeHandler) AllocateNodeID(nodeIP net.IP) uint16 {
 	defer n.mutex.Unlock()
 
 	if nodeID, exists := n.nodeIDsByIPs[nodeIP.String()]; exists {
-		return nodeID
+		nodeID.refcnt++
+		n.nodeIDsByIPs[nodeIP.String()] = nodeID
+		return nodeID.id
 	}
 
 	nodeID := uint16(n.nodeIDs.AllocateID())
@@ -56,7 +58,7 @@ func (n *linuxNodeHandler) AllocateNodeID(nodeIP net.IP) uint16 {
 			logfields.IPAddr: nodeIP,
 		}).Debug("Allocated new node ID for node IP address")
 	}
-	if err := n.mapNodeID(nodeIP.String(), nodeID); err != nil {
+	if err := n.mapNodeID(nodeIP.String(), nodeID, IPCacheCreatedID); err != nil {
 		log.WithError(err).WithFields(logrus.Fields{
 			logfields.NodeID: nodeID,
 			logfields.IPAddr: nodeIP.String(),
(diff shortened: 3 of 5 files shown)


pkg/datapath/linux/node.go
```diff
@@ -1131,13 +1133,24 @@ func (n *linuxNodeHandler) nodeUpdate(oldNode, newNode *nodeTypes.Node, firstAdd
...
 		oldKey = oldNode.EncryptionKey
+
+		// Delete the old node IP addresses from nodeID map if they are dropped by the update.
+		for _, address := range oldNode.IPAddresses {
+			oldNodeIPAddrs = append(oldNodeIPAddrs, address.IP.String())
+		}
+	}
+
+	for _, address := range newNode.IPAddresses {
+		newNodeIPAddrs = append(newNodeIPAddrs, address.IP.String())
```
pchaigno, line 1158: Instead of having two additional loops on `x.IPAddresses`, we could pass `x.IPAddresses` directly to `deleteNodeIPs` and have it call `.IP.String()`.

pkg/datapath/linux/node_ids.go
```diff
@@ -77,6 +77,25 @@ func (n *linuxNodeHandler) getNodeIDForNode(node *nodeTypes.Node) uint16 {
...
+			if newAddr == oldAddr {
+				found = true
+				break
+			}
+		}
+		if !found {
+			if err := n.unmapNodeID(oldAddr); err != nil {
+				log.WithError(err).WithFields(logrus.Fields{
+					logfields.IPAddr: oldAddr,
+				}).Warn("DeleteNodeIPS failed to remove a node IP to node ID mapping")
```
pchaigno, line 154: Should this be an error given we expect a failure to remove the node IP mapping will eventually break connectivity once the IP is reused?

pkg/datapath/fake/node.go
```diff
@@ -55,6 +55,13 @@ func (n *FakeNodeHandler) AllocateNodeID(_ net.IP) uint16 {
 	return 0
```
pchaigno, line 55: > One complication is we can get a nodeUpdate after an IP upsert. In this case we promote the life cycle ownership over to the node delete path. This appears to be logical because the node update/delete path is more explicit.

Do you mean that the node manager becomes responsible for the node ID and we stop using the refcount? IIUC, that may not work because a node IP can still be used in the ipcache after the node itself has been deleted.

pkg/datapath/linux/node_ids.go
```diff
@@ -65,18 +67,96 @@ func (n *linuxNodeHandler) AllocateNodeID(nodeIP net.IP) uint16 {
...
+	}
+
+	// IPCache can allocate/deallocate nodeIDs but if this node was allocated
+	// through a node event lets wait for the paired update to remove it.
+	if nodeID.source != IPCacheCreatedID {
+		return
+	}
+
+	if nodeID.refcnt < 1 {
+		log.Warning("Attempt to dellocate node IP with refcnt < 1")
```
joestringer, line 112: Please ensure to include relevant information in `WithFields(logrus.Fields{ ... })` in such log statements. It would be very beneficial to understand which nodes / IPs are affected so you can correlate that with other events going on in the cluster.

pkg/datapath/linux/node_ids.go
```diff
@@ -65,18 +67,96 @@ func (n *linuxNodeHandler) AllocateNodeID(nodeIP net.IP) uint16 {
 	return nodeID
 }
 
+func (n *linuxNodeHandler) GetNodeID(nodeIP net.IP) uint16 {
```
joestringer, line 70: nit: I'd guess that `net.IP` is the convenient type here and probably makes sense for v1.11 branch, but if this were targeted at main branch then I'd suggest switching to `netip` library which is a better and more efficient underlying type that we use for all IP and prefix types in new code. (At the same time, I'm not sure the status of those types being used on this branch so I'm fine to leave this as-is for this branch).

pkg/datapath/linux/node_ids.go
```diff
@@ -65,18 +67,96 @@ func (n *linuxNodeHandler) AllocateNodeID(nodeIP net.IP) uint16 {
...
 	for _, addr := range node.IPAddresses {
 		if id, exists := n.nodeIDsByIPs[addr.IP.String()]; exists {
-			nodeID = id
+			nodeID = id.id
 		}
 	}
 	return nodeID
 }
 
+func (n *linuxNodeHandler) deleteNodeIPs(oldIPs, newIPs []string) {
```
joestringer, line 141: I agree that this should probably be shifted up further, this code looked rather familiar until I realized that this is specific to nodeID handling rather than just general node handling. I'm not sure how the decision of pkg/node vs. pkg/ipcache handling for this exactly plays out, but we should probably figure that out at some point. For now, to fix the issue, I'm nominally OK with this approach.

pkg/datapath/linux/node.go
```diff
@@ -49,6 +49,20 @@ type NeighLink struct {
...
+
+const (
+	NewNodeCreatedID = iota
+	DatapathRestoreCreatedID
+	IPCacheCreatedID
+)
+
+type nodeID struct {
+	id     uint16
+	refcnt uint32
```
joestringer, line 62: nit: From a review perspective it'd be easier to confirm that this patch is correct if one patch introduced the change to the the structure + map handling code which should be non-functional, then another patch added the refcount.

pkg/datapath/linux/node_ids.go
```diff
@@ -65,18 +67,96 @@ func (n *linuxNodeHandler) AllocateNodeID(nodeIP net.IP) uint16 {
...
+	defer n.mutex.Unlock()
+
+	nodeID, exists := n.nodeIDsByIPs[nodeIP.String()]
+	if !exists {
+		return 0
+	}
+	return nodeID.id
+}
+
+func (n *linuxNodeHandler) DeallocateNodeID(nodeIP net.IP) {
```
joestringer, line 81: Why doesn't `DeallocateNodeID()` call `deallocateNodeIDLocked()`?

pkg/datapath/node.go
```diff
@@ -141,9 +141,16 @@ type NodeNeighbors interface {
 
 type NodeIDHandler interface {
 	// AllocateNodeID allocates a new ID for the given node (by IP) if one wasn't
-	// already assigned.
+	// already assigned or increments the refcnt if it exists.
 	AllocateNodeID(net.IP) uint16
 
+	// DeallocateNodeID decrements the refcnt of a node ID and releases the ID
+	// if the refcnt is zero.
+	DeallocateNodeID(net.IP)
```
joestringer, line 149: Can we unit test this? I don't see any test changes in this PR to confirm that alloc+dealloc semantics are properly held.

pkg/ipcache/ipcache.go
```diff
@@ -440,7 +449,10 @@ func (ipc *IPCache) DumpToListenerLocked(listener IPIdentityMappingListener) {
 		}
 		nodeID := uint16(0)
 		if hostIP != nil {
-			nodeID = NodeHandler.AllocateNodeID(hostIP)
+			nodeID = NodeHandler.GetNodeID(hostIP)
+			if nodeID == 0 {
+				nodeID = NodeHandler.AllocateNodeID(hostIP)
```
joestringer, line 454: Why does dump need to allocate node IDs? This code path can be used for debugging introspection randomly, it seems a bit sketchy to have allocations happening in this path.

joestringer: I guess.. under what circumstances would a hostIP end up in the ipcache without allocating a node ID? Surely we could do this upon insert instead..?

pkg/ipcache/ipcache.go
```diff
@@ -536,8 +548,22 @@ func (ipc *IPCache) deleteLocked(ip string, source source.Source) (namedPortsCha
...
-	if newHostIP != nil {
-		nodeID = NodeHandler.AllocateNodeID(newHostIP)
+	// If this is a straight delete this is simple, deallocate the nodeID. If its a
+	// Upsert then the deleted IP was shadowing another CIDR so we will restore
+	// that mapping. But, from NodeID side we already did an allocate for that and
+	// we don't want to bump the refcnt here so simply Get() the ID. Remember to
+	// delete the old IP through the Upsert path though.
+	if cacheModification == Delete {
+		NodeHandler.DeallocateNodeID(oldHostIP)
+	} else if cacheModification == Upsert {
```
joestringer, line 558: Incidentally we can probably be pretty confident that the upsert case here is when there is a CIDR prefix `/32` in policy that is shadowed by a real pod Identity, so if we delete the pod then restore the CIDR prefix (hence `cacheModification == Upsert`), that CIDR prefix has no host association so the "newHostIP" case should not be possible. Probably also the newHostIP is always != oldHostIP. Furthermore, probably (though I say this with a little less confidence), the oldHostIP is always set.

It's not expressible in the current ipcache structure to have two competing pods with the same IP such that when you delete one of those pods, we put the other one (with a corresponding hostIP) into the ipcache.

All that said, the code here would theoretically do the right thing, I just think that a couple of the conditional statements here are likely never hit.

joestringer (commented):
Some initial comments.

I didn't go through the lifecycle to try to balance out and ensure (a) all interactions with the nodeID system always goes through consistent Allocate/Deallocate function APIs to ensure the refcounting is tracked correctly, nor did I go through (b) exactly how the ipcache refcount tracking can't somehow end up leaking references. Probably needs a little more attention there.

pkg/datapath/node.go
jrfastab: I have a patch on top of this with unit tests but because we call directly into the bpf map update helpers I had to comment out the bpf map updates to get it to work. This is another reason to move this code up out of datapath/linux into ipcache or elsewhere and build a fake handler to add/delete node ids. I'm trying to minimize the changes here though. I could add the unit test as a patch 3 here though for review, it might help understand the life cycle.

pkg/datapath/linux/node_ids.go
jrfastab: well they are not the same logic :/ I guess a small core could be pulled out but doesn't look like much to me.

pkg/datapath/linux/node_ids.go
jrfastab: sure

pkg/datapath/linux/node_ids.go
jrfastab: yeah I think this is where conflicts landed on main, point taken.

pkg/datapath/linux/node_ids.go
jrfastab: I'm trying to avoid a full refactor for a fix... Ideally we would uplevel this and a fake/real handlers and so on.

pkg/datapath/linux/node.go
jrfastab: ok I can do this if it helps.

pkg/ipcache/ipcache.go
jrfastab: I think Dump can happen when we restore the kernel state into userspace on a restart?

jrfastab: There are entries in the ipcache with hostIDs that never have node ID this is true in EKS env when running lots of node churn and pod churn. I didn't debug the root cause of this.

pkg/ipcache/ipcache.go
jrfastab: I agree, but I was trying to be a bit defensive to handle whatever the ipcache threw at us.

pkg/datapath/linux/node_ids.go
joestringer: Yep I think we're aligned here.

pchaigno (commented):
I switched to draft mode given this will need to be reopened again main.

pkg/datapath/fake/node.go
jrfastab: I guess its plausible that we could see the nodeDelete before pod delete events and continue to send traffic. Then without a nodeID mapping we would see policy block errors because the traffic would be dropped.

I really wanted to simply push all the handling of nodeIDs into ipcache logic so that it is not managed through addNode at all. WDYT? I hesitated thinking this promote logic would ensure we dropped it on node delete.  But, perhaps its better to just remove the node logic.

pkg/datapath/linux/node_ids.go
jrfastab: I guess. but its not obvious to me how to correct the situation. If the delete failed why did it fail, (1) it didn't exist, (2) the ip address is invalid, (3) BPF map delete failed.

It seems to me if (1) is OK then (2) really shouldn't happen and if we can reliably delete from the bpf map then we have other trouble. Maybe we could retry for case BPF map failure, but that seems unlikely to me.

If the entry doesn't exist then its a bug but we will recover because the bug deleted the entry to soon. So not much to do except log it and we fix it.

pkg/datapath/linux/node.go
jrfastab: oh sure can do that.

jrfastab:
@pchaigno you want this against main first it sounds like?

pchaigno:
> @pchaigno you want this against main first it sounds like? We can keep this open until main is merged I guess but last I checked a cherry-pick from main onto 1.11 wont work so will be best to use this as the backport.

Given this affects all stable branches and main, the usual process would be to open the fix against main and then backport it everywhere. That's what I'm doing for all other IPsec fixes.

brb:
@jrfastab I guess we can close this PR because of the https://github.com/cilium/cilium/pull/26725#issuecomment-1662397492?

PR #26342, opened 2023-06-19
auth: delete cache-entry on ErrKeyNotExist

mhofstetter:
In the unlikely event of trying to delete an auth map entry which is no longer present in the auth map, the entry should also be removed from the auth map cache. Currently it gets kept in the cache, which would result in unnecessary retries.


meyskens (approved):
Simple but sane change, LGTM

PR #26352, opened 2023-06-19
Use custom container to initialise clustermesh etcd

### LINKED ISSUE #23770: Update clustermesh-apiserver initContainer to configure etcd for v3.5.7 or later
Etcd v3.5.7 or later no longer ships with a shell in the default image.

Currently in the following clustermesh-apiserver deployment, we rely on the availability of a shell:

https://github.com/cilium/cilium/blob/105331bd133131c6f724fc9121fef1fb57c22ca9/install/kubernetes/cilium/templates/clustermesh-apiserver/deployment.yaml#L40

In order to bump the etcd dependency to v3.5.7, we need to update this line to either:
1. Rely on another image that has a shell that we can use in order to configure etcd, OR
2. Rework these instructions such that they work directly in the etcd 3.5.7 container.

Related: https://github.com/cilium/cilium/issues/23760

JamesLaverack:
This is an early draft of a fix for https://github.com/cilium/cilium/issues/23770

---

      GitHub issue.

Fixes: #23770

```release-note
```


etcd-init/.gitignore
```diff
@@ -0,0 +1,5 @@
+cilium-operator
+cilium-operator-aws
+cilium-operator-azure
+cilium-operator-alibabacloud
+cilium-operator-generic
```
JamesLaverack: TODO: Remove references to cilium-operator in this dir.

etcd-init/cmd/completion/completion.go
```diff
@@ -0,0 +1,71 @@
...
+# Installing zsh completion on Linux/macOS
+## Load the hubble-relay completion code for zsh into the current shell
+        source <(hubble-relay completion zsh)
+## Write zsh completion code to a file and source if from .zshrc
+        hubble-relay completion zsh > ~/.hubble-relay/completion.zsh.inc
+        printf "
+          # hubble-relay shell completion
+          source '$HOME/.hubble-relay/completion.zsh.inc'
+          " >> $HOME/.zshrc
+        source $HOME/.zshrc`
```
JamesLaverack: TODO Update this text, remove references to hubble-relay

etcd-init/cmd/version/version.go
```diff
@@ -0,0 +1,29 @@
...
+// Copyright Authors of Cilium
+
+package version
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/spf13/cobra"
+
```
JamesLaverack, line 11:
```suggestion
```

pkg/etcd/init/init.go
```diff
@@ -0,0 +1,121 @@
+package init
```
learnitall, line 1: There are a couple of variables in here that I think could be moved to the `defaults.go` file. See [1]

pkg/etcd/init/init.go
```diff
@@ -0,0 +1,121 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"go.etcd.io/etcd/client/v3"
+)
+
+const (
+	rootRoleName = "root"
```
learnitall: [1]

pkg/etcd/init/init.go
```diff
@@ -0,0 +1,121 @@
...
+	_, err = client.AuthEnable(ctx)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func setupRootUser(ctx context.Context, client *clientv3.Client) error {
+	rootUsername := "root"
```
learnitall: [1]

pkg/etcd/init/init.go
```diff
@@ -0,0 +1,121 @@
...
+	_, err = client.UserGrantRole(ctx, adminUsername, rootRoleName)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func setupExternalWorkloadUser(ctx context.Context, client *clientv3.Client) error {
+	externalWorkloadUsername := "externalworkload"
+	externalWorkloadRoleName := "externalworkload"
```
learnitall: [1]

pkg/etcd/init/init.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	_, err = client.UserGrantRole(ctx, rootUsername, rootRoleName)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func setupAdminUser(ctx context.Context, client *clientv3.Client, clusterName string) error {
+	adminUsername := fmt.Sprintf("admin-%s", clusterName)
```
learnitall: [1]. Could be something like `adminUsernamePrefix`

pkg/etcd/init/init.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func setupRemoteUser(ctx context.Context, client *clientv3.Client) error {
+	remoteUsername := "remote"
+	remoteRoleName := "remote"
```
learnitall: [1]

pkg/etcd/init/init.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"go.etcd.io/etcd/client/v3"
+)
+
+const (
+	rootRoleName = "root"
+)
+
+func InitEtcd(ctx context.Context, client *clientv3.Client, clusterName string) error {
+	_, err := client.Put(ctx, "cilium/.has-cluster-config", "true")
```
learnitall: I think it could make sense to add a constant for this key value. Also, I'm assuming that the goal is to eventually use this to stop init if this key is found?

learnitall (commented):
Super exciting work! I took a quick look and so far your implementation looks great. I have a couple of follow-up questions for you though:

1. Do we need pprof, gops or completion? Not a deal-breaker, but It does add some overhead. I'm not familiar with our other images and if this is standard.
2. Also not a deal breaker, but it might be a good idea to incorporate the hive framework here, as we might migrate this code over to it in the future anyways. This would also be helpful for potential testing on clustermesh in the future, as it would fit into the testing methodology we are working towards. For example, one could import the etcd-init Cell and use it to init a new etcd database within another integration test.
3. Any ideas or plans on how we want to write tests? I think starting with some kind of integration test where we compare an etcd database that has been initialized with this code to an etcd database that has been initialized using the shell script it will replace. We could even write a test that performs an init with this code and just checks the state of the database to make sure everything we expect is there.

JamesLaverack:
> Super exciting work! I took a quick look and so far your implementation looks great. I have a couple of follow-up questions for you though:

Thank you so much for the review!

> 1. Do we need pprof, gops or completion? Not a deal-breaker, but It does add some overhead. I'm not familiar with our other images and if this is standard.

Hrm, possibly not. I copied another binary to form this one, and in general I've tried to keep to existing style wherever possible. I'm indifferent on if we keep them or remove them.

> 2. Also not a deal breaker, but it might be a good idea to incorporate the hive framework here, as we might migrate this code over to it in the future anyways. This would also be helpful for potential testing on clustermesh in the future, as it would fit into the testing methodology we are working towards. For example, one could import the etcd-init Cell and use it to init a new etcd database within another integration test.

I must admit I'm not very familiar with hive. I'll do some research and see what can be done.

> 3. Any ideas or plans on how we want to write tests? I think starting with some kind of integration test where we compare an etcd database that has been initialized with this code to an etcd database that has been initialized using the shell script it will replace. We could even write a test that performs an init with this code and just checks the state of the database to make sure everything we expect is there.

Yeah, there's a few levels of test we can write

* A script comparison test is great for this implementation — but that means that the script will need to be mantained as future changes are made in order to keep the test passing. Is that an acceptable level of matinance?
* I assume there are existing end-to-end tests for cluster mesh? I presume they would fail if a mistake was made here?
* We could try to mock the etcd client for the init code, but I think that'd be rather brittle as a test, and not really actually **test** half the code here. So. I'd err away from that option, and certinally not as a replacement for the integration test.

giorio94:
I haven't yet had an in-depth look at the code. One aspect that worries me a bit, though, is that this binary will use a fixed version of etcd (the one specified in the go.mod file), which will be likely different from the one used for the real etcd container configured by the user, possibly leading to subtle bugs in case of incompatibilities. Moreover, we should also make multiple versions of this binary available (for different etcd versions), which is also problematic.   WDYT?

JamesLaverack:
> I haven't yet had an in-depth look at the code. One aspect that worries me a bit, though, is that this binary will use a fixed version of etcd (the one specified in the go.mod file), which will be likely different from the one used for the real etcd container configured by the user, possibly leading to subtle bugs in case of incompatibilities. Moreover, we should also make multiple versions of this binary available (for different etcd versions), which is also problematic. WDYT?

It's a good point. Do we have a documented set of etcd versions that we support?

Building multiple versions is of course possible. I'm not certian if we can import multiple different etcd versions in the same Go source tree — but I think it's possible. Failing that some makefile magic might be required.

I did consider not useing embedded etcd, and building the init binary as a client-only thing and layering it ontop of etcd. But we'd still need to build, test, and distribute multiple versions for different etcd versions so that doesn't really help. In a way this problem already exists — we're just presuming the etcdctl API wouldn't change in the existing shell script.

giorio94:
> It's a good point. Do we have a documented set of etcd versions that we support?

I don't think it is explicitly documented somewhere, but maybe I just don't know about that. Client side [we are checking](https://github.com/cilium/cilium/blob/main/pkg/kvstore/etcd.go#L64) that the etcd server version is >=3.1.0 though.

> 
> Building multiple versions is of course possible. I'm not certian if we can import multiple different etcd versions in the same Go source tree — but I think it's possible. Failing that some makefile magic might be required.

> 
> I did consider not useing embedded etcd, and building the init binary as a client-only thing and layering it ontop of etcd. But we'd still need to build, test, and distribute multiple versions for different etcd versions so that doesn't really help. In a way this problem already exists — we're just presuming the etcdctl API wouldn't change in the existing shell script.

Yeah, I also haven't found yet any good solution which does not require having to build multiple versions (unless etcd supported configuration files). Yet, the etcd docs mention that [minor version downgrade is not supported](https://etcd.io/docs/v3.5/upgrades/upgrade_3_5/#downgrade), and [minor version upgrades shall only performed one version at a time](https://etcd.io/docs/v3.5/upgrades/upgrade_3_5/#upgrade-requirements). Both of which are potential issues in case of mismatching etcd versions between the custom init container and the real etcd one.

JamesLaverack:
> This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

I'm still working on this, just been on PTO recently. 😅

pkg/etcd/init/init.go
JamesLaverack: >  Also, I'm assuming that the goal is to eventually use this to stop init if this key is found?

There's no specific future goal, this is just matching [existing behaviour](https://github.com/cilium/cilium/blob/5e6137cee001315b304add6646be8b541553daa6/install/kubernetes/cilium/templates/clustermesh-apiserver/deployment.yaml#L57) which sets this key.

It is used in the [kvstore code](https://github.com/cilium/cilium/blob/5e6137cee001315b304add6646be8b541553daa6/pkg/kvstore/kvstore.go#L50) though.

Speaking of, it's probably better to read this key from there.

giorio94:
> > Both of which are potential issues in case of mismatching etcd versions between the custom init container and the real etcd one.
> 
> We should have control of that though? We set the etcd image version in our install YAML right?

Kinda. We configure those as part of the helm values, which means that users can modify them if they wish.

PR #26353, opened 2023-06-19
docs: Update kvstore documentation with potential circular dependency.

### LINKED ISSUE #25632: Document recommendation for running kvstore as non-Cilium-managed pods
When Cilium is configured in kvstore mode ([docs](https://docs.cilium.io/en/latest/installation/k8s-install-external-etcd/#installation-with-external-etcd)), it uses the kvstore as the controlplane to manage state (see [reference](https://docs.cilium.io/en/latest/kvstore/)).

Commonly, users deploy etcd as their kvstore. It is possible to run etcd as pods within the Cilium cluster. If these etcd pods are using the pod network, then they rely on Cilium. However, Cilium also relies on the kvstore as the controlplane and uses the kvstore to fetch information needed to provision the datapath for pods. This creates a 🐔 and 🥚 problem.

The workaround and recommendation is to deploy etcd either outside of the cluster (or on a separately managed cluster) or to deploy the etcd pods with `hostNetwork: true`.

marseel:
Current documentation does not inform about potential circular dependency between Cilium and kvstore.
This commit explains how this dependency can happen and also gives two potential ways of resolving it.

Also, bump Etcd version requirement to 3.4.0+, as previous minor releases are no longer supported.

Fixes https://github.com/cilium/cilium/issues/25632

### DIFF
--- a/Documentation/installation/k8s-install-external-etcd.rst
+++ b/Documentation/installation/k8s-install-external-etcd.rst
@@ -37,7 +37,17 @@ when to use a kvstore:
 
 .. include:: requirements-intro.rst
 
-You will also need an external etcd version 3.1.0 or higher.
+You will also need an external etcd version 3.4.0 or higher.
+
+Kvstore and Cilium dependency
+=============================
+When using an external kvstore, it's important to break the circular dependency between Cilium and kvstore.
+If kvstore pods are running within the same cluster and are using a pod network then kvstore relies on Cilium.
+However, Cilium also relies on the kvstore, which creates a circular dependency.
+There are two recommended ways of breaking this dependency:
+
+ * Deploy kvstore outside of cluster or on separately managed cluster.
+ * Deploy kvstore pods with a host network, by specifying ``hostNetwork: true`` in the pod spec.
 
 Configure Cilium
 ===========================


Documentation/installation/k8s-install-external-etcd.rst
```diff
@@ -37,7 +37,17 @@ when to use a kvstore:
 
 .. include:: requirements-intro.rst
 
-You will also need an external etcd version 3.1.0 or higher.
+You will also need an external etcd version 3.4.0 or higher.
+
+Kvstore and Cilium dependency
+================================
```
qmonnet:
```suggestion
Kvstore and Cilium dependency
=============================

```

Documentation/installation/k8s-install-external-etcd.rst
```diff
@@ -37,7 +37,17 @@ when to use a kvstore:
...
+When using an external kvstore, it's important to break the circular dependency between Cilium and kvstore.
+If kvstore pods are running within the same cluster and are using a pod network then kvstore relies on Cilium.
+However, Cilium also relies on the kvstore, which creates a circular dependency.
+There are two recommended ways of breaking this dependency:
+
+ * Deploy kvstore outside of cluster or on separately managed cluster
```
qmonnet:
```suggestion
 * Deploy kvstore outside of cluster or on separately managed cluster.
```

qmonnet (changes requested):
Of the two methods to address the circular dependency, is there one that should be preferred? Do they have particular advantages/drawbacks?

(The underlying question is: if I'm a user facing this, how do I know which of the two methods I should follow?)

marseel:
I don't think there is any significant advantage/disadvantage between these two options.

qmonnet (approved):
OK, so let's users sort it out by themselves I guess :shrug: 

Looks all good

Issue #922, opened 2017-06-06
L4 policy with more than 40 ports doesn't work


michi-covalent:
Cilium endpoints go into 'not-ready' state when I apply an L4 policy with more than 40 ports. It would be nice if Cilium can validate the policy before applying it so that endpoints don't go into not-ready state. Here is the policy I'm using:

```
kind: NetworkPolicy
apiVersion: extensions/v1beta1
metadata:
  name: myapp
spec:
  podSelector:
    matchLabels:
      io.kubernetes.pod.namespace: mynamespace
  ingress:
  - from:
    - podSelector:
        matchLabels:
          io.kubernetes.pod.namespace: mynamespace
    ports:
    - protocol: TCP
      port: 1
    - protocol: TCP
      port: 2
    - protocol: TCP
      port: 3
    - protocol: TCP
      port: 4
    - protocol: TCP
      port: 5
    - protocol: TCP
      port: 6
    - protocol: TCP
      port: 7
    - protocol: TCP
      port: 8
    - protocol: TCP
      port: 9
    - protocol: TCP
      port: 10
    - protocol: TCP
      port: 11
    - protocol: TCP
      port: 12
    - protocol: TCP
      port: 13
    - protocol: TCP
      port: 14
    - protocol: TCP
      port: 15
    - protocol: TCP
      port: 16
    - protocol: TCP
      port: 17
    - protocol: TCP
      port: 18
    - protocol: TCP
      port: 19
    - protocol: TCP
      port: 20
    - protocol: TCP
      port: 21
    - protocol: TCP
      port: 22
    - protocol: TCP
      port: 23
    - protocol: TCP
      port: 24
    - protocol: TCP
      port: 25
    - protocol: TCP
      port: 26
    - protocol: TCP
      port: 27
    - protocol: TCP
      port: 28
    - protocol: TCP
      port: 29
    - protocol: TCP
      port: 30
    - protocol: TCP
      port: 31
    - protocol: TCP
      port: 32
    - protocol: TCP
      port: 33
    - protocol: TCP
      port: 34
    - protocol: TCP
      port: 35
    - protocol: TCP
      port: 36
    - protocol: TCP
      port: 37
    - protocol: TCP
      port: 38
    - protocol: TCP
      port: 39
    - protocol: TCP
      port: 40
    - protocol: TCP
      port: 41
```

michi-covalent:
... forgot to mention the version I'm using:

```
root@minikube:/# cilium version
[...]

PR #26355, opened 2023-06-19
Change maxPort 40 -> 100

damsien:

      GitHub issue.

Description of change:
Change the max ports authorized for one policy. Allow more than 40 ports in one policy.

It would be great if we could do that because '40' seems to be an arbitrary number.

#922 [specific message](https://github.com/cilium/cilium/issues/922#issuecomment-306678130)

### DIFF
--- a/pkg/policy/api/rule_validation.go
+++ b/pkg/policy/api/rule_validation.go
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	maxPorts      = 40
+	maxPorts      = 100
 	maxICMPFields = 40
 )
 


ti-mo:
@Damsien, thanks for your contribution! Please specify a valid email address in the DCO.

Also, what makes 100 less of an arbitrary number than 40? Could we improve this in a more meaningful way?

nathanjsweet:
@Damsien there are good reasons for keeping this number at a default of 40, as increasing it can explode the size of the policy table in the datapath. We would be open to making this number configurable via [a configuration option](https://docs.cilium.io/en/stable/configuration/), but simply changing the size is not something that is likely to merge.

Damsien:
Indeed, adding this number to the configmap is a good idea
Thank you for your quick feedback I'll go ahead and close this pull request

PR #26354, opened 2023-06-19
Documentation: Add graceful restart section in BGP documentation

harsimran-pabla:
Introduce BGP Graceful Restart documentation in bgp-control-plane.rst document.

Issue #26351, opened 2023-06-19
CI: Conformance E2E - echo-ingress-mutual-auth-spiffe/pod-to-pod/curl-ipv6-3


giorio94:
## CI failure

Hit on https://github.com/cilium/cilium/pull/26347: https://github.com/cilium/cilium/actions/runs/5310635377/jobs/9612898313
Matrix entry: (7, bpf-next-20230420.212204, none, strict, disabled, snat, true, testing-only)

```
[=] Test [echo-ingress-mutual-auth-spiffe]
........
  ℹ️  📜 Applying CiliumNetworkPolicy 'echo-ingress-mutual-authentication' to namespace 'cilium-test'..
  [-] Scenario [echo-ingress-mutual-auth-spiffe/pod-to-pod]
  [.] Action [echo-ingress-mutual-auth-spiffe/pod-to-pod/curl-ipv4-0: cilium-test/client-755fb678bd-9mwh4 (10.244.2.215) -> cilium-test/echo-other-node-7bc75565b9-vm7ql (10.244.3.140:8080)]
  [.] Action [echo-ingress-mutual-auth-spiffe/pod-to-pod/curl-ipv6-0: cilium-test/client-755fb678bd-9mwh4 (fd00:10:244:2::f66b) -> cilium-test/echo-other-node-7bc75565b9-vm7ql (fd00:10:244:3::bb80:8080)]
  [.] Action [echo-ingress-mutual-auth-spiffe/pod-to-pod/curl-ipv4-1: cilium-test/client-755fb678bd-9mwh4 (10.244.2.215) -> cilium-test/echo-same-node-77d94449fc-f9dwk (10.244.2.164:8080)]
  [.] Action [echo-ingress-mutual-auth-spiffe/pod-to-pod/curl-ipv6-1: cilium-test/client-755fb678bd-9mwh4 (fd00:10:244:2::f66b) -> cilium-test/echo-same-node-77d94449fc-f9dwk (fd00:10:244:2::faa5:8080)]
  [.] Action [echo-ingress-mutual-auth-spiffe/pod-to-pod/curl-ipv4-2: cilium-test/client2-5b97d7bc66-kttmv (10.244.2.90) -> cilium-test/echo-other-node-7bc75565b9-vm7ql (10.244.3.140:8080)]
[...]

PR #26356, opened 2023-06-19
CI Workflow: Add all Azure supported k8s versions

brlbil:
The current aks conformance tests run against one version of k8s either a specific version or the default version that is provided by the cloud provider.

This PR adds all the supported k8s versions by Azure in a matrix strategy.

Other cloud providers will be added by subsequent PRs.

Normal run examples:
:white_check_mark: [conformance-aks-v1.12.yaml](https://github.com/cilium/cilium/actions/runs/5313019853)
:white_check_mark: [conformance-aks-v1.13.yaml](https://github.com/cilium/cilium/actions/runs/5313019842)
:white_check_mark: [conformance-aks.yaml](https://github.com/cilium/cilium/actions/runs/5313019860)

Schedule run simulated:
:white_check_mark: [conformance-aks-v1.12.yaml](https://github.com/cilium/cilium/actions/runs/5313506446)
:x: [conformance-aks-v1.13.yaml](https://github.com/cilium/cilium/actions/runs/5314131467) ,  failures are not related cluster installation
:white_check_mark: [conformance-aks.yaml](https://github.com/cilium/cilium/actions/runs/5313506456)

### DIFF
--- a/.github/actions/azure/k8s-versions-schema.yaml
+++ b/.github/actions/azure/k8s-versions-schema.yaml
@@ -1,0 +1,6 @@
+include: list(include('includeItem'))
+---
+includeItem:
+  version: str()
+  location: str()
+  default: bool(required=False)
--- a/.github/actions/azure/k8s-versions.yaml
+++ b/.github/actions/azure/k8s-versions.yaml
@@ -1,0 +1,10 @@
+# List of k8s version for AKS tests
+---
+include:
+  - version: "1.24"
+    location: westeurope
+  - version: "1.25"
+    location: westus
+    default: true
+  - version: "1.26"
+    location: australiaeast
--- a/.github/workflows/conformance-aks-v1.11.yaml
+++ b/.github/workflows/conformance-aks-v1.11.yaml
@@ -67,9 +67,7 @@ concurrency:
 
 env:
   name: ${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
-  location: westeurope
   cost_reduction: --node-vm-size Standard_B2s --node-osdisk-size 30
-  k8s_version: 1.23
   # renovate: datasource=github-releases depName=cilium/cilium-cli
   cilium_cli_version: v0.14.7
   cilium_cli_ci_version:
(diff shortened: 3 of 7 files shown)


.github/workflows/lint-workflows.yaml
```diff
@@ -157,3 +157,30 @@ jobs:
...
+        run: |
+          for dir in aws azure gke;do
+            for name in k8s-versions test-config; do
+              file_base="src/github.com/cilium/cilium/.github/actions/${dir}/${name}"
+              if [ -f ${file_base}.yaml ];then
+                yamale -s ${file_base}-schema.yaml ${file_base}.yaml;
+              fi
+            done
+          done
+
```
nbusseneau: Ideally this would be handled in a separate PR on which all of #26356, #26351 and #26354 depend, but it's OK to just have the other PRs depend on this one as well. I think at the very least we should split this in a separate commit though.

nbusseneau (changes requested):

.github/workflows/conformance-aks-v1.11.yaml
```diff
@@ -114,11 +112,103 @@ jobs:
             src:
               - '!(test|Documentation)/**'
 
+  setup-report:
+    runs-on: ubuntu-latest
```
nickolaev, line 116: Do we always use latest? Shall we align to LTS (22.04)?

brlbil: Yes, all CI uses this

.github/workflows/lint-workflows.yaml
brlbil: It was good that this part was highlighted because tests did not cover some of the files.
Fixed it, and separated it as a commit.

.github/actions/azure/k8s-versions-schema.yaml
```diff
@@ -0,0 +1,6 @@
+include: list(include('includeItem'))
```
bimmlerd, line 1: somewhat meta, but should `.github/actions/` (maybe even all of `.github`) be CODEOWNED by by the `cilium/ci-structure` team instead of contributing? This feels far in the GHA action territory.

bimmlerd (approved):
I don't feel particularly qualified, see comment, but what is owned by contributing LGTM.

.github/actions/azure/k8s-versions.yaml
```diff
@@ -0,0 +1,10 @@
+# List of k8s version for AKS tests
+---
+include:
+  - version: "1.24"
```
tgraf, line 4: Are we unintentionally removing 1.23 test coverage here? The commit message says that this adds "all supported versions" so I'm not sure if the removal of 1.23 is intentional.

tgraf (changes requested):

.github/actions/azure/k8s-versions.yaml
nbusseneau: 1.23 is not supported on Azure anymore, so we removed it as part of our usual cycling (otherwise what happens the installation will fail consistently due the K8s version being refused by the Azure API on cluster creation) => https://learn.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli

nbusseneau (approved):

.github/actions/azure/k8s-versions.yaml
tgraf: OK, thanks. Not clear from the commit message.

tgraf (approved):

.github/actions/azure/k8s-versions.yaml
brlbil: This table was created from the supported k8s for AKS, https://learn.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli#aks-kubernetes-release-calendar

PR #25461, opened 2023-05-15
test/k8s: add host firewall workaround for svc host policy test.

tommyp1ckles:
Change 439a0a0 introduced workaround to common flake we've been seeing relating to issue #15455.

Any test enabling hostfw/host-policy will may suffer from the same issue.

Addresses: #25411

PR #26362, opened 2023-06-19
test: Fix the attempted fix for the hostfw flake

pchaigno:
Commit 0cfce97f31 ("test/k8s: add host firewall workaround for svc host policy test.") ported the fix from 439a0a059 ("test: Fix ACK and FIN+ACK policy drops in hostfw tests") to a new set of host firewall tests.

It is however not enough to call the same function. We also need to ensure that the same policy as in the tests themselves is loaded during the preparation.

This pull request makes that small change.

Related: https://github.com/cilium/cilium/pull/25461.

### DIFF
--- a/test/k8s/datapath_configuration.go
+++ b/test/k8s/datapath_configuration.go
@@ -450,7 +450,7 @@ var _ = Describe("K8sDatapathConfig", func() {
 			}
 			deploymentManager.DeployCilium(options, DeployCiliumOptionsAndDNS)
 
-			prepareHostPolicyEnforcement(kubectl)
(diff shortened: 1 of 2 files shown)


marseel:
Now test failed with:
```
daemon creation failed: unable to initialize kube-proxy replacement options: The high-scale IPcache mode requires Node Port mode \\\"dsr\\\" with geneve dispatch
```
and from logs:
```
--bpf-lb-mode='snat'"
```
which comes from the fact, that test runs with `kube-proxy-replacement=strict`, which enables node-port

christarazi (approved):

PR #26363, opened 2023-06-19
etcd: start the status checker only after establishing the initial session

giorio94:
Currently, the etcd status checker is started in a separate goroutine as soon as the client is created, without waiting for the establishment of the initial session. Yet, the status checker will never succeed until then, based on how it is implemented. This is problematic especially in the clustermesh case, because the errors propagated by the status checker cause a watchdog logic to restart the connection.

Let's consider the following situation: a new etcd client is created, starting the corresponding status checker, while the remote clustermesh-apiserver is not yet running. The status checker will likely fail a few times, propagating the corresponding error through the StatusCheckErrors channel. Eventually, the clustermesh-apiserver pod boots up, the initial session is established and the different watchers are started. At this point, the watchdog logic starts to read from the channel, and restarts an otherwise working connection. It is worth mentioning that this issue affects clustermesh only because the other components ignore the errors returned through the StatusCheckErrors channel.

This PR modifies the etcd client to start the status checker only after that the initial connection has been established (the same also for the heartbeat watcher), so that the above issue is prevented. Another alternative might be to ignore all errors reported through the StatusCheckErrors when the watchdog is initially started, but this would be subject to race conditions, and possibly show incorrect statuses through `cilium status`.

### DIFF
--- a/pkg/kvstore/etcd.go
+++ b/pkg/kvstore/etcd.go
@@ -699,33 +699,50 @@ func connectEtcdClient(ctx context.Context, config *client.Config, cfgPath strin
 		ec.RWMutex.Lock()
 		ec.sessionErr = err
 		ec.RWMutex.Unlock()
+
+		ec.statusLock.Lock()
+		ec.latestStatusSnapshot = "Failed to establish initial connection"
+		ec.latestErrorStatus = err
+		ec.statusLock.Unlock()
+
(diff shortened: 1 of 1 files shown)


tklauser (approved):

marseel (approved):
Looks good, thanks!

PR #26367, opened 2023-06-20
Use generics in k8s factory functions

AwesomePatrol:
There is a lot of code duplication and it is increasingly difficult to add new features to it. Go 1.18 introduced generics which can greatly help in such situations.

This touches a lot of places, so next time I will replace function body instead of function calls.


pkg/k8s/factory_functions.go
```diff
@@ -741,142 +741,22 @@ func TransformToNode(obj interface{}) (interface{}, error) {
...
-// *cilium_v2.CiliumEnvoyConfig or a cache.DeletedFinalStateUnknown into
-// a cache.DeletedFinalStateUnknown with a *cilium_v2.CiliumEnvoyConfig in its Obj.
-// If the given obj can't be cast into either *cilium_v2.CiliumEnvoyConfig
-// nor cache.DeletedFinalStateUnknown, an error is returned.
-func TransformToCiliumEnvoyConfig(obj interface{}) (interface{}, error) {
+// TransformToGeneric transforms a given obj into typ or
+// cache.DeletedFinalStateUnknown with typ in its Obj.
+// If the given obj can't be cast into either typ nor
+// cache.DeletedFinalStateUnknown, an error is returned.
+func TransformToGeneric[typ any](obj interface{}) (interface{}, error) {
```
giorio94: Is this transformer actually necessary, or could it be simply dropped? Is seems to me that no actual transformation of the object is being performed.

giorio94 (commented):
Thanks @AwesomePatrol,

* The ObjToGeneric change looks good to me (it would be great if you could just extend the commit message with a bit of description about the rationale behind the change).
* For the TransformToGeneric change, I wonder if it could be possible to drop it completely, given that it seems a no-op to me (and some informers are already not using it).

pkg/k8s/factory_functions.go
AwesomePatrol: It doesn't modify anything, but it checks if the correct type was used. I think it is also done with the `ObjTo` function later in the processing, so maybe we can drop it.

(some tests fail if I change it to `return obj, nil`)

giorio94: Yeah, error checking was introduced when switching to the new transformer implementation (https://github.com/cilium/cilium/pull/25604), which also updated the tests which I think you are referring to. As you mentioned this check is also performed by the `ObjTo` functions, so I would it personally just drop this one. But let's also hear the opinion of the other reviewers.

danehans: ~Note that some informers are configured with a transformer eventhough transformation of the provided object is unneeded. Refer to https://github.com/cilium/cilium/issues/26240 for additional details. Should the expected output type be included in the type parameters list when calling `TransformToGeneric()`?~ Nevermind, this PR is not using generics for TransformXX functions that perform type transformation.

giorio94: IMHO there are two separate categories. On the one hand the transformation functions which are actually no-ops (apart from error checking), and that are replaced here by `TransformToGeneric`. In this case the input and output types are the same, and for that reason I think that the best thing would be just to drop them. Then there are the ones which do an actual transformation, and that are not handled here.

AwesomePatrol: Removed `TransformToGeneric` and its references as it is already handled in other parts of the code.

giorio94: Makes sense to me. Could you please squash the first and the third commits together? Then I'll have a final look at it.

pkg/k8s/factory_functions.go
giorio94, line 1: There's still an `ObjToCiliumNode` function which could also be removed (usages have already been removed).

giorio94 (approved):
Looks good to me, thanks! Just one more function that could also be removed.

meyskens (approved):
Changes to service mesh look good :) and always yay for less code!

pkg/k8s/factory_functions.go
```diff
@@ -15,271 +15,29 @@ import (
...
-	slim_discover_v1beta1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/discovery/v1beta1"
 	slim_networkingv1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/networking/v1"
 	slim_metav1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/apis/meta/v1"
 	"github.com/cilium/cilium/pkg/k8s/types"
 	"github.com/cilium/cilium/pkg/logging/logfields"
 )
 
-func ObjToV1NetworkPolicy(obj interface{}) *slim_networkingv1.NetworkPolicy {
-	k8sNP, ok := obj.(*slim_networkingv1.NetworkPolicy)
+func ObjToGeneric[typ any](obj interface{}) *typ {
```
squeed: Can you add a quick docblock as to what this does? It's not a general purpose conversion function, it's just a safe cast around informer objects.

Not to bikeshed, but I'd call this function something else. Maybe CastEventObj or CastInformerEvent or something?

(I disagree with the existing codebase's naming convention :-) )

operator/cmd/ccnp_event.go
```diff
@@ -84,8 +85,8 @@ func enableCCNPWatcher(ctx context.Context, wg *sync.WaitGroup, clientset k8sCli
 			},
 			UpdateFunc: func(oldObj, newObj interface{}) {
 				k8sEventMetric(resources.MetricCCNP, resources.MetricUpdate)
-				if oldCNP := k8s.ObjToSlimCNP(oldObj); oldCNP != nil {
-					if newCNP := k8s.ObjToSlimCNP(newObj); newCNP != nil {
+				if oldCNP := k8s.ObjToGeneric[types.SlimCNP](oldObj); oldCNP != nil {
```
squeed: Suggestion: While you're in here, add a `InformerObjDeepEqual[t](a, b interface{})` function too? That should cut down on even more boilerplate :-)

squeed (approved):
A few suggestions, but nothing critical. Thanks for the cleanup!

ldelossa:
@AwesomePatrol can you rebase this and push? Two tests are not running and this will fix it.

pkg/k8s/factory_functions.go
AwesomePatrol: Changed name to `CastInformerEvent` and added docblock (based on similar docblocks for deleted functions).

operator/cmd/ccnp_event.go
AwesomePatrol: `DeepEqual`s are generated functions. I will leave it for another time, but would you prefer to implement `DeepEqual` in a generic way or have a call to `DeepEqual` in a generic function thus saving a few lines of code here and there?

squeed (approved):

squeed:
(this should get in soon)

AwesomePatrol:
Sorry, I had to rebase as name of one of the imports changed, so it would fail to build after merge.

squeed:
One non-required job was cancelled. Merging.

PR #26370, opened 2023-06-20
clustermesh: improve reliability of TestClusterMesh

giorio94:
Currently, TestClusterMesh is subject to possible race conditions caused by the conjoint usage of defer (i.e., to stop the identity allocator) and Cleanup (i.e., to execute the stop hook of the clustermesh subsystem). Additionally, some cluster ID reservation assertions may fail due race conditions as they are not "eventually" tested.

Let's fix this possible issues to improve the reliability of the test. While at it, let's also convert it to leverage the testing framework.

### DIFF
--- a/pkg/clustermesh/clustermesh_test.go
+++ b/pkg/clustermesh/clustermesh_test.go
@@ -9,10 +9,11 @@ import (
 	"fmt"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


YutaroHayakawa (approved):
LGTM 👍 Thanks!

PR #26372, opened 2023-06-20
gateway-api: Add tests for standard CRD

sayboras:
This is to improve the coverage with the standard CRD channel. Also, small changes on existing docs are done.

**Note**: required GHA jobs should be updated after this PR is merged.

### DIFF
--- a/.github/workflows/conformance-gateway-api.yaml
+++ b/.github/workflows/conformance-gateway-api.yaml
@@ -27,7 +27,7 @@ env:
   # renovate: datasource=github-releases depName=kubernetes-sigs/kind
   kind_version: v0.17.0
   kind_config: .github/kind-config.yaml
-  gateway_api_version: v0.6.1
+  gateway_api_version: v0.7.0
   metallb_version: 0.12.1
   timeout: 5m
 
@@ -35,6 +35,12 @@ jobs:
   gateway-api-conformance-test:
     runs-on: ubuntu-latest
     timeout-minutes: 120
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - crd-channel: experimental
+        - crd-channel: standard
     steps:
       - name: Checkout main branch to access local actions
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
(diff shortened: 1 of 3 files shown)


Documentation/network/servicemesh/gateway-api/installation.rst
```diff
@@ -5,21 +5,24 @@ Prerequisites
...
-        $ kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.5.1/config/crd/standard/gateway.networking.k8s.io_httproutes.yaml
-        $ kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.5.1/config/crd/experimental/gateway.networking.k8s.io_referencegrants.yaml
+        $ kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.7.0/config/crd/standard/gateway.networking.k8s.io_gatewayclasses.yaml
+        $ kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.7.0/config/crd/standard/gateway.networking.k8s.io_gateways.yaml
+        $ kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.7.0/config/crd/standard/gateway.networking.k8s.io_httproutes.yaml
+        $ kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.7.0/config/crd/standard/gateway.networking.k8s.io_referencegrants.yaml
+        $ kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.7.0/config/crd/standard/gateway.networking.k8s.io_referencegrants.yaml
```
mhofstetter: looks like `referencegrants` is a duplicate.

.github/workflows/conformance-gateway-api.yaml
```diff
@@ -105,11 +121,12 @@ jobs:
...
-          kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${{ env.gateway_api_version }}/config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
+          kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${{ env.gateway_api_version }}/config/crd/${{ matrix.crd-channel }}/gateway.networking.k8s.io_gatewayclasses.yaml
+          kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${{ env.gateway_api_version }}/config/crd/${{ matrix.crd-channel }}/gateway.networking.k8s.io_gateways.yaml
+          kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${{ env.gateway_api_version }}/config/crd/${{ matrix.crd-channel }}/gateway.networking.k8s.io_httproutes.yaml
+          kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${{ env.gateway_api_version }}/config/crd/${{ matrix.crd-channel }}/gateway.networking.k8s.io_referencegrants.yaml
+          ## TLSRoute is only available in experimental channel in v0.7.0
```
mhofstetter, line 128: shouldn't the experimental `TLSRoute` CRD only be applied in the corresponding matrix `experimental`? Or does the testrunner expect the CRD to be installed even though the feature isn't enabled?

sayboras: we don't have granular control in operator, so without tlsroute crd, the operator watcher will fail and panic.

Documentation/network/servicemesh/gateway-api/installation.rst
sayboras: oops, good catch :+1:

.github/workflows/conformance-gateway-api.yaml
mhofstetter: ok, so it's expected that a user installs all the experimental CRDs too? :thinking: 
might be worth to add a feature toggle to the operator the same way its supported by the testrunner.
but definitely out of scope for this PR - thanks for the info!

mhofstetter (approved):

qmonnet (approved):
Looks good, thanks!

nbusseneau (approved):

PR #26376, opened 2023-06-20
bpf: egressgw: refactor unit tests

jibi:
These changes were originally part of https://github.com/cilium/cilium/pull/26215 (to make it easier to add a new test) but I'm splitting them into a separate PR so that they can get backported without the fib_lookup changes, as it's always good to keep CI as much in sync as possible across all branches.

See individual commits for more context.


ldelossa (approved):
LGTM

PR #26378, opened 2023-06-20
Use generic Set instead of specified Set

bzsuni:
Use generic Set instead of specified Set
sets.String has been deprecated and the generic Set should be used instead

### DIFF
--- a/pkg/k8s/slim/k8s/apis/labels/selector.go
+++ b/pkg/k8s/slim/k8s/apis/labels/selector.go
@@ -267,8 +267,8 @@ func (r *Requirement) Operator() selection.Operator {
 }
 
 // Values returns requirement values
-func (r *Requirement) Values() sets.String {
-	ret := sets.String{}
+func (r *Requirement) Values() sets.Set[string] {
+	ret := sets.New[string]()
 	for i := range r.strValues {
 		ret.Insert(r.strValues[i])
 	}
@@ -686,7 +686,7 @@ func (p *Parser) parseRequirement() (*Requirement, error) {
 	if err != nil {
 		return nil, err
 	}
-	var values sets.String
+	var values sets.Set[string]
 	switch operator {
 	case selection.In, selection.NotIn:
 		values, err = p.parseValues()
@@ -696,7 +696,7 @@ func (p *Parser) parseRequirement() (*Requirement, error) {
 	if err != nil {
 		return nil, err
 	}
-	return NewRequirement(key, operator, values.List())
+	return NewRequirement(key, operator, sets.List(values))
 
 }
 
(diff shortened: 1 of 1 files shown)


ti-mo (approved):

giorio94:
Thanks @bzsuni for your contribution! The file which is modified by this PR appears to be part of a slim copy of the structures used in k8s ([readme](https://github.com/cilium/cilium/blob/e6748209b27ed24b319265ac45b517b9af9b516f/pkg/k8s/slim/README.md)). It is unclear to me whether we want to perform modifications which would cause it to diverge from the upstream version. Though, I'd defer it to other reviewers with more experience about this package.

jrajahalme (approved):
Given that `sets.String` says it's deprecated and the generic `Set` should be used instead, this is the right thing to do.

bzsuni:
it seems running e2e failed 
https://github.com/cilium/cilium/actions/runs/5329285324/jobs/9659206177?pr=26378#step:14:248

ldelossa:
@bzsuni looks like some tests are stuck. Can you rebase `main` and push this back up?

aditighag (approved):

bzsuni:
@ldelossa Thank you, I have ebased the `main`

rolinh (approved):

tklauser (approved):

tommyp1ckles (approved):
thanks for the cleanup!

qmonnet (changes requested):
Thanks! But looking at the commit, I can't find the reason for the change. Please add the motivation in the commit description.

bzsuni:
> Thanks! But looking at the commit, I can't find the reason for the change. Please add the motivation in the commit description.

yes, thanks! sets.String says it's deprecated and the generic Set should be used instead. I wonder if we should also make this change

qmonnet:
> > Thanks! But looking at the commit, I can't find the reason for the change. Please add the motivation in the commit description.
> 
> yes, thanks! sets.String says it's deprecated and the generic Set should be used instead.

Yes I got that from the discussion. Deprecated since v0.26.0, it seems.

> I wonder if we should also make this change

What other change do you mean?

What I'm saying is that your commit description is empty, and does not mention the deprecation. So when we look at the Git history in the future, to understand why this update happened, we'll have none of the context or motivation. Admittedly, it's possible to open the Pull Request discussion to get more info, but doing so shouldn't be necessary. Please state in the commit description the motivation for the change.

bzsuni:
> What I'm saying is that your commit description is empty (save for the sign-off), and does not mention the deprecation. So when we look at the Git history in the future, to understand why this update happened, we'll have none of the context or motivation. Admittedly, it's possible to open the Pull Request discussion to get more info, but doing so shouldn't be necessary. Please state in the commit description the motivation for the change.

Yes, you are right, I have stated it in the commit, thanks

qmonnet (approved):
Thank you

learnitall (approved):

julianwiedmann:
Overriding checkpatch's complaint about a typo in the patch description.

Issue #25508, opened 2023-05-17
CFP-2023-05-17: Hubble Flow Logs


AwesomePatrol:
## Cilium Feature Proposal
        
**Is your feature request related to a problem?**

Without Hubble Timescape it is impossible to get information about the past Hubble Flows (outside the history stored in the ring buffer).

**Describe the feature you'd like**

I want to have an option to output information on Flows for a period of time (maybe indefinite).

**(Optional) Describe your proposed solution**

I can't link to the doc, so I will copy the content in the first comment.

AwesomePatrol:
### Proposal

Add a new CRD that will be watched by `cilium-agent`. When `end` is in the future it will process Hubble flows, the ones that match the filter will be stripped from information outside the `field mask` and logged in a file. Most observability tools can be configured to ingest it.

#### Definition

```yaml
// Not a real CRD
kind: flowlogging.networking.gke.io
spec:
  properties:
    fieldmask:
      description: "List of field names that will be kept in the log output"
      type: array
      items:
        description: "Field name that will be kept in the output. Subfields are separated by dot"
        type: string
    filter:
      description: "Filter"
      type: object
      properties:
        allowlist:
          description: "Allowlist filter generated by CLI. Empty matches all. Flow passes the filter if any of the entries match. Only Flows matching Allowlist, but not matching Denylist will be logged"
          type: array
          items:
            description: "Allowlist entry"
            type: string
        denylist:
          description: "Denylist filter generated by CLI. Empty matches none. Flow passes the filter if none of the entries match. Only Flows matching Allowlist, but not matching Denylist will be logged"
          type: array
          items:
            description: "Denylist entry"
            type: string
    end:
      description: "End denotes the timestamp when logging will stop"
      type: string
```

#### Examples

Example of collecting Flow logs for a specific user-project.
* until `15:04:05` of `2023-05-10 PDT`
* from all pods in a `user-project` namespace
* but exclude `kube-dns` traffic in both directions
* keep only 4 fields: `time`, `src`, `dst`, `verdict` (forward/drop)

```yaml
spec:
  fieldmask:
    - "time"
    - "source.pod_name"
    - "destination.pod_name"
    - "verdict"
  filter:
    allowlist:
      - '{"source_pod":["user-project/"]}'
      - '{"destination_pod":["user-project/"]}'
    denylist:
      - '{"source_label":["k8s:k8s-app=kube-dns"]}'
[...]

PR #26379, opened 2023-06-20
Expand Hubble-exporter with filters and field mask

AwesomePatrol:
Add support for filters and field mask to Hubble-exporter

This is prerequisite for implementation of #25508

```release-note
Add an option to specify a filters and field mask for hubble-exporter
```

### DIFF
--- a/Documentation/cmdref/cilium-agent.md
+++ b/Documentation/cmdref/cilium-agent.md
@@ -176,6 +176,9 @@ cilium-agent [flags]
       --hubble-disable-tls                                        Allow Hubble server to run on the given listen address without TLS.
       --hubble-event-buffer-capacity int                          Capacity of Hubble events buffer. The provided value must be one less than an integer power of two and no larger than 65535 (ie: 1, 3, ..., 2047, 4095, ..., 65535) (default 4095)
       --hubble-event-queue-size int                               Buffer size of the channel to receive monitor events.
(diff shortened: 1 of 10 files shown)


AwesomePatrol:
Do you think that I should also add new fields to config-map/command-line to configure new options of the exporter?

If yes, how should they be specified? Is JSON for raw filters ok?

chancez:
@AwesomePatrol yes, we should add some flags for this. JSON is fine. 

And for adding to the configmap, that's also probably a good idea as well.

thorn3r (approved):
agent changes lgtm :+1:

asauber (approved):
CLI-related changes LGTM. Appreciate the clear variable names and parsing logic.

chancez (commented):
Overall the approach LGTM. What do you think about adding some documentation to go with the export feature?

AwesomePatrol:
> What do you think about adding some documentation to go with the export feature?

Sure, I can do that, but as there is a different PR relying on this can we have it done separately? Feel free to create an issue and assign it to me so that I won't forget.

chancez:
Created https://github.com/cilium/cilium/issues/26970 since we have existing flags + the new options that all need documenting.

chancez (approved):

AwesomePatrol:
I think the same tests failed again, but it is hard to tell if it may be the fault of this change or tests. Could you please take a look?

Issue #26075, opened 2023-06-09
CI: ConformanceAKS - Clean up Cilium


giorio94:
## CI failure

Hit on https://github.com/cilium/cilium/pull/25921: https://github.com/cilium/cilium/actions/runs/5220654175/jobs/9424008106

```
Run pkill -f "cilium.*hubble.*port-forward|kubectl.*port-forward.*hubble-relay"
  pkill -f "cilium.*hubble.*port-forward|kubectl.*port-forward.*hubble-relay"
  cilium uninstall --chart-directory=install/kubernetes/cilium --wait
  shell: /usr/bin/bash -e {0}
  env:
    name: cilium-cilium-5220654175
    location: westeurope
    cost_reduction: --node-vm-size Standard_B2s --node-osdisk-size 30
    cilium_cli_version: v0.14.5
    cilium_cli_ci_version: 
    check_url: https://github.com/cilium/cilium/actions/runs/5220654175
    job_name: Installation and Connectivity Test
    QUAY_ORGANIZATION: cilium
    QUAY_ORGANIZATION_DEV: cilium
    QUAY_CHARTS_ORGANIZATION_DEV: cilium-charts-dev
    AZURE_HTTP_USER_AGENT: 
    AZUREPS_HOST_ENVIRONMENT: 

Error: Process completed with exit code 1.
```

It seems that the background `cilium hubble port-forward` process terminated early for some reason, causing `pkill` to fail since no matching process was found (Exit code 1 means `No processes matched or none of them could be signalled` according to the man page).

Sysdump: [cilium-sysdumps.zip](https://github.com/cilium/cilium/files/11706108/cilium-sysdumps.zip)

aspsk:
Given the flakyness of the `kubectl port-forward`, we need to ignore the exit value `1` of `pkill` to prevent from this flake

tklauser:
Hit in https://github.com/cilium/cilium/pull/26390: https://github.com/cilium/cilium/actions/runs/5332672234/jobs/9662173594

PR #26380, opened 2023-06-20
bpf: refactor lib/fib.h to use new funcs

ldelossa:
- bpf,fib: use fib_do_redirect in fib_redirect
- bpf: refactor fib_redirect_v4/6 to use decoupled functions

This pull request is a follow up on #26136. 

In this pull request we refactor the existing lib/fib.h to use the modularized 
fib functions.

These changes should result in no functional differences in the datapath, instead it
removes the redundant code that is present prior to this commit.

```release-note
bpf,fib: refactor lib/fib.h to remove the now redundant code
```

### DIFF
--- a/bpf/lib/fib.h
+++ b/bpf/lib/fib.h
@@ -212,27 +212,32 @@ fib_lookup_v6(struct __ctx_buff *ctx, struct bpf_fib_lookup_padded *fib_params,
 
 static __always_inline int
 fib_redirect_v6(struct __ctx_buff *ctx, int l3_off,
-		struct ipv6hdr *ip6, const bool needs_l2_check,
-		__s8 *fib_err, int iif, int *oif)
+		struct ipv6hdr *ip6 __maybe_unused, const bool needs_l2_check,
+		__s8 *fib_err __maybe_unused, int *oif)
 {
-	struct bpf_fib_lookup_padded fib_params = {
-		.l = {
-			.family		= AF_INET6,
-			.ifindex	= iif,
-		},
-	};
+	struct bpf_fib_lookup_padded fib_params __maybe_unused = {0};
+	__s8 skip_fib __maybe_unused = BPF_FIB_LKUP_RET_NO_NEIGH;
 	int ret;
 
-	ipv6_addr_copy((union v6addr *)&fib_params.l.ipv6_src,
-		       (union v6addr *)&ip6->saddr);
-	ipv6_addr_copy((union v6addr *)&fib_params.l.ipv6_dst,
-		       (union v6addr *)&ip6->daddr);
+#ifndef ENABLE_SKIP_FIB
+	ret = fib_lookup_v6(ctx, &fib_params, &ip6->saddr, &ip6->daddr, 0);
+	*fib_err = (__s8)ret;
 
 	ret = ipv6_l3(ctx, l3_off, NULL, NULL, METRIC_EGRESS);
 	if (unlikely(ret != CTX_ACT_OK))
 		return ret;
 
-	return fib_redirect(ctx, needs_l2_check, &fib_params, fib_err, oif);
+	return fib_do_redirect(ctx, needs_l2_check, &fib_params, fib_err, oif);
+#else
+	ret = ipv6_l3(ctx, l3_off, NULL, NULL, METRIC_EGRESS);
+	if (unlikely(ret != CTX_ACT_OK))
+		return ret;
+
+	*fib_err = skip_fib;
+
+	*oif = DIRECT_ROUTING_DEV_IFINDEX;
+	return fib_do_redirect(ctx, needs_l2_check, NULL, fib_err, oif);
+#endif
 }
 #endif /* ENABLE_IPV6 */
 
@@ -256,24 +261,33 @@ fib_lookup_v4(struct __ctx_buff *ctx, struct bpf_fib_lookup_padded *fib_params,
 
 static __always_inline int
 fib_redirect_v4(struct __ctx_buff *ctx, int l3_off,
-		struct iphdr *ip4, const bool needs_l2_check,
-		__s8 *fib_err, int iif, int *oif)
+		struct iphdr *ip4 __maybe_unused, const bool needs_l2_check,
+		__s8 *fib_err __maybe_unused, int *oif)
 {
-	struct bpf_fib_lookup_padded fib_params = {
-		.l = {
-			.family		= AF_INET,
-			.ifindex	= iif,
-			.ipv4_src	= ip4->saddr,
-			.ipv4_dst	= ip4->daddr,
-		},
-	};
+	struct bpf_fib_lookup_padded fib_params __maybe_unused = {0};
+	__s8 skip_fib __maybe_unused = BPF_FIB_LKUP_RET_NO_NEIGH;
 	int ret;
+
+#ifndef ENABLE_SKIP_FIB
+	ret = fib_lookup_v4(ctx, &fib_params, ip4->saddr, ip4->daddr, 0);
+	*fib_err = (__s8)ret;
 
 	ret = ipv4_l3(ctx, l3_off, NULL, NULL, ip4);
 	if (unlikely(ret != CTX_ACT_OK))
 		return ret;
 
-	return fib_redirect(ctx, needs_l2_check, &fib_params, fib_err, oif);
+	return fib_do_redirect(ctx, needs_l2_check, &fib_params, fib_err, oif);
+#else
+	ret = ipv4_l3(ctx, l3_off, NULL, NULL, ip4);
+	if (unlikely(ret != CTX_ACT_OK))
+		return ret;
+
+	*fib_err = skip_fib;
+
+	*oif = DIRECT_ROUTING_DEV_IFINDEX;
+	return fib_do_redirect(ctx, needs_l2_check, NULL, fib_err, oif);
+#endif
+
 }
 #endif /* ENABLE_IPV4 */
 #endif /* __LIB_FIB_H_ */
--- a/bpf/lib/fib.h
+++ b/bpf/lib/fib.h
@@ -171,87 +171,20 @@ out_send:
 
 static __always_inline int
 fib_redirect(struct __ctx_buff *ctx, const bool needs_l2_check,
-	     struct bpf_fib_lookup_padded *fib_params, __s8 *fib_err, int *oif)
+	     struct bpf_fib_lookup_padded *fib_params __maybe_unused,
+	     __s8 *fib_err __maybe_unused, int *oif)
 {
-	struct bpf_redir_neigh nh_params;
-	struct bpf_redir_neigh *nh = NULL;
-	bool no_neigh = is_defined(ENABLE_SKIP_FIB);
-	int ret;
-
 #ifndef ENABLE_SKIP_FIB
+	int ret;
 	ret = fib_lookup(ctx, &fib_params->l, sizeof(fib_params->l), 0);
-	if (ret != BPF_FIB_LKUP_RET_SUCCESS) {
-		*fib_err = (__s8)ret;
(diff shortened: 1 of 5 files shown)


ldelossa:
Aha, cherry picked these and forgot we changed the signatures in #26136. 

Marking as draft until I update those sigs.

ldelossa:
Ping

aspsk:
Conformance AKS fails due to [supposedly closed] https://github.com/cilium/cilium/issues/26075 The fix was this: https://github.com/cilium/cilium/commit/f6600fd7c17facb4ca884a7acd2520274fc62e4a and I don't see it in the workflow from this PR. Maybe you need to rebase?

bpf/tests/session_affinity_test.c
```diff
@@ -187,7 +187,7 @@ int test1_check(__maybe_unused const struct __ctx_buff *ctx)
 
 	__u32 *status_code = data;
 
-	if (*status_code != XDP_TX) test_fatal("status code != XDP_TX");
+	if (*status_code != XDP_TX) test_fatal("status code != XDP_TX %d", *status_code);
```
aspsk: `%u` instead of `%d` for `__u32`

bpf/lib/fib.h
```diff
@@ -211,27 +211,31 @@ fib_lookup_v6(struct __ctx_buff *ctx, struct bpf_fib_lookup_padded *fib_params,
...
-	struct bpf_fib_lookup_padded fib_params = {
-		.l = {
-			.family		= AF_INET6,
-			.ifindex	= iif,
-		},
-	};
+	struct bpf_fib_lookup_padded fib_params __maybe_unused = {0};
+	void *data_end;
+	void *data;
+	__s8 skip_fib __maybe_unused = BPF_FIB_LKUP_RET_NO_NEIGH;
```
aspsk, line 219: Let's try to follow netdev kernel style in declarations (reverse xmass tree)

aspsk (approved):
LGTM, left two cosmetic comments

bpf/lib/fib.h
```diff
@@ -323,24 +259,32 @@ fib_lookup_v4(struct __ctx_buff *ctx, struct bpf_fib_lookup_padded *fib_params,
...
 
 	ret = ipv4_l3(ctx, l3_off, NULL, NULL, ip4);
 	if (unlikely(ret != CTX_ACT_OK))
 		return ret;
 
-	return fib_redirect(ctx, needs_l2_check, &fib_params, fib_err, oif);
+	/* we touched ttl above, revalidate for any header reads/writes. */
+	if (!revalidate_data(ctx, &data, &data_end, &ip4))
+		return DROP_INVALID;
+
```
julianwiedmann: Can we reshuffle the code so that this revalidation isn't needed? Do the FIB lookup first, then adjust the TTL, then do the redirect?

Otherwise this should respect the passed-in `l3_off`, and use `revalidate_data_l3_off()`.

bpf/lib/fib.h
```diff
@@ -279,27 +211,31 @@ fib_lookup_v6(struct __ctx_buff *ctx, struct bpf_fib_lookup_padded *fib_params,
...
-
 	ret = ipv6_l3(ctx, l3_off, NULL, NULL, METRIC_EGRESS);
 	if (unlikely(ret != CTX_ACT_OK))
 		return ret;
 
-	return fib_redirect(ctx, needs_l2_check, &fib_params, fib_err, oif);
+	/* we touched ttl above, revalidate for any header reads/writes */
+	if (!revalidate_data(ctx, &data, &data_end, &ip6))
+		return DROP_INVALID;
+
```
julianwiedmann: This shouldn't be required (there's no csum adjustment in the IPv6 case) ?

bpf/lib/fib.h
```diff
@@ -279,27 +211,31 @@ fib_lookup_v6(struct __ctx_buff *ctx, struct bpf_fib_lookup_padded *fib_params,
...
+#ifndef ENABLE_SKIP_FIB
+	ret = fib_lookup_v6(ctx, &fib_params, &ip6->saddr, &ip6->daddr, 0);
+	*fib_err = (__s8)ret;
+	return fib_do_redirect(ctx, needs_l2_check, &fib_params, fib_err, oif);
+#else
```
julianwiedmann: Note that one downside of this approach (deferring the error evaluation for `fib_lookup_v6()` into `fib_do_redirect()`) is that we dirty the `fib_err` even on success (with `BPF_FIB_LKUP_RET_NO_NEIGH`).

So if there's a subsequent error that doesn't set an `ext_err`, we might end up reporting inconsistent error data in the drop notification. But that's something to sort out in a follow-on change I suppose ...

julianwiedmann (changes requested):

bpf/lib/fib.h
ldelossa: Ack.

bpf/tests/session_affinity_test.c
ldelossa: ack

bpf/lib/fib.h
ldelossa: Yup I believe you're right, we aren't touching packet structure. I'll remove.

bpf/lib/fib.h
ldelossa: Stared at the code for a bit and I think I see what you're saying. 

My initial concern was, we touch TTL regardless, and we'll need to revalidate if we update mac addresses in `fib_do_redirect`. But after inspection of the helper methods we call in `fib_do_redirect` such as `eth_store_daddr`, it appears they do their own revalidation internally (getting the latest data and data_end pointers). 

Are we on the same page here about why the reshuffling works?

julianwiedmann: yep, a sequence like

```
fib_lookup_v4()
ipv4_l3()
fib_do_redirect()
```

should work, as the `fib_lookup()` has no reason to invalidate the context pointers. It operates on the FIB params block we provide.

bpf/lib/fib.h
julianwiedmann: I recently realized that `ipv6_dec_hoplimit()` is the *one* place where we use `BPF_F_RECOMPUTE_CSUM` though. So let's see :grin:.

ldelossa: Ah good point, but I guess if we reshuffle the code anyway, this can go away. We can reshuffle in the same manor as the ipv4 code you outlined above.

ldelossa: Implemented, lets see how the tests go.

bpf/lib/fib.h
ldelossa: implemented

bpf/lib/fib.h
ldelossa: Yes, let me do this in a follow up :notebook_with_decorative_cover:

bpf/lib/fib.h
```diff
@@ -171,87 +171,19 @@ fib_do_redirect(struct __ctx_buff *ctx, const bool needs_l2_check,
...
-		}
-	} else {
-		if (eth_store_daddr(ctx, fib_params->l.dmac, 0) < 0)
-			return DROP_WRITE_ERROR;
-		if (eth_store_saddr(ctx, fib_params->l.smac, 0) < 0)
-			return DROP_WRITE_ERROR;
-	}
-out_send:
-	return ctx_redirect(ctx, *oif, 0);
+	return fib_do_redirect(ctx, needs_l2_check, NULL, &skip_fib, oif);
```
julianwiedmann: When `fib_do_redirect()` sets an ext_err in its `fib_ret` parameter, it doesn't get propagated back to the caller from here. This means we lose the `BPF_FIB_MAP_NO_NEIGH` error info.

bpf/lib/fib.h
```diff
@@ -279,27 +211,30 @@ fib_lookup_v6(struct __ctx_buff *ctx, struct bpf_fib_lookup_padded *fib_params,
...
 
-	return fib_redirect(ctx, needs_l2_check, &fib_params, fib_err, oif);
+	return fib_do_redirect(ctx, needs_l2_check, &fib_params, fib_err, oif);
+#else
+	ret = ipv6_l3(ctx, l3_off, NULL, NULL, METRIC_EGRESS);
+	if (unlikely(ret != CTX_ACT_OK))
+		return ret;
+
+	*oif = DIRECT_ROUTING_DEV_IFINDEX;
+	return fib_do_redirect(ctx, needs_l2_check, NULL, &skip_fib, oif);
```
julianwiedmann: Same here - `skip_fib` is just a local variable, it doesn't propagate back to the caller.

bpf/lib/fib.h
```diff
@@ -323,24 +258,31 @@ fib_lookup_v4(struct __ctx_buff *ctx, struct bpf_fib_lookup_padded *fib_params,
...
 
-	return fib_redirect(ctx, needs_l2_check, &fib_params, fib_err, oif);
+	return fib_do_redirect(ctx, needs_l2_check, &fib_params, fib_err, oif);
+#else
+	ret = ipv4_l3(ctx, l3_off, NULL, NULL, ip4);
+	if (unlikely(ret != CTX_ACT_OK))
+		return ret;
+
+	*oif = DIRECT_ROUTING_DEV_IFINDEX;
+	return fib_do_redirect(ctx, needs_l2_check, NULL, &skip_fib, oif);
```
julianwiedmann: ditto.

julianwiedmann (changes requested):
The `ipv*_l3()` change looks good now :+1:.

Just some regression in error reporting left to fix.

bpf/lib/fib.h
ldelossa: Hmm, I think I had some feelings about this error handling. 

For one, the error message is really for the `fib_lookup` error, at least semantically from the caller, and not about the redirect. 

I don't love that the error is kinda overloaded, leaving the application to kinda "guess" what actually went wrong. 

I think my goal here was to make the passed in `fib_err` actually represent the error returned from a fib_lookup, while the return value from the function to classify that a different type of error took place (in this case, a redirect error, or an error writing new TTL). 

What's your thoughts here? I'm also happy to just make error handling occur like it did previously, where the error variable was overloaded.

julianwiedmann: Let's not mix subtle changes in behaviour into a "refactor" :). Or at least split them off into a separate patch that explains why this is better behaviour.

borkmann:
Sorry for being late for review. I was trying to catch up on https://github.com/cilium/cilium/pull/26136/commits which added the duplicated functions, and was wondering whether they are even needed at all?

Couldn't srv6_refib() have used just the plain old fib_redirect_v6(), or at least just a small wrapper to pass in the fib_params. 

Roughly something like this:
```
	ret = fib_redirect_params_v6(..., &params, ext_err);
        if (ret != <redirect_code>)
                   return DROP_NO_FIB;
	if (old_oif == params.l.ifindex)
	           return CTX_ACT_OK;
        return ret;
```

borkmann:
Something like this, not even compile tested, but to mention what I mean: https://github.com/cilium/cilium/commits/pr/fib-consolidation

ldelossa:
Spoke with @borkmann out of band. 

We will push this PR forward with the suggested error handling changes by Julian. Once this is merge I'll work with Daniel on ironing out and simplifying these changes further.

bpf/lib/fib.h
ldelossa: Okay @julianwiedmann latest commit fixes  this issue. 

You are right, we still dirty the fib_ret error in the case where 'ENABLE_SKIP_FIB' macro is defined. 

I think this is OK tho, as the caller should also check the return value of 'fib_do_redirect' and 'fib_redirect_v*' to determine what the issue was. 

The 'fib_ret' function now can pass the `BPF_FIB_MAP_NO_NEIGH` in case of error.

julianwiedmann (approved):
:rocket:

PR #26382, opened 2023-06-20
ci-aks, ci-external-workloads: Use cilium-cli Helm mode

michi-covalent:
migrating ci-aks and ci-external-workloads to use Helm mode for https://github.com/cilium/cilium/issues/25156

### DIFF
--- a/.github/workflows/conformance-aks.yaml
+++ b/.github/workflows/conformance-aks.yaml
@@ -71,6 +71,7 @@ env:
   # renovate: datasource=github-releases depName=cilium/cilium-cli
   cilium_cli_version: v0.14.7
   cilium_cli_ci_version:
+  CILIUM_CLI_MODE: helm
   check_url: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
 
 jobs:
@@ -250,18 +251,11 @@ jobs:
             --helm-set=debug.enabled=true \
             --helm-set loadBalancer.l7.backend=envoy \
             --helm-set tls.secretsBackend=k8s \
-            --azure-resource-group ${{ env.name }}-${{ matrix.location }} \
-            --wait=false \
-            --rollback=false \
-            --config monitor-aggregation=none \
-            --version="
-          HUBBLE_ENABLE_DEFAULTS="--chart-directory=install/kubernetes/cilium \
-            --relay-image=quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/hubble-relay-ci:${SHA} \
-            --relay-version=${SHA}"
+            --helm-set=azure.resourceGroup=${{ env.name }}-${{ matrix.location }} \
+            --helm-set=bpf.monitorAggregation=none"
           CONNECTIVITY_TEST_DEFAULTS="--flow-validation=disabled \
             --hubble=false --collect-sysdump-on-failure --external-target bing.com --external-cidr 8.0.0.0/8 --external-ip 8.8.4.4 --external-other-ip 8.8.8.8"
           echo cilium_install_defaults=${CILIUM_INSTALL_DEFAULTS} >> $GITHUB_OUTPUT
-          echo hubble_enable_defaults=${HUBBLE_ENABLE_DEFAULTS} >> $GITHUB_OUTPUT
           echo connectivity_test_defaults=${CONNECTIVITY_TEST_DEFAULTS} >> $GITHUB_OUTPUT
 
       - name: Install Cilium CLI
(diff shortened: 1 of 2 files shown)


michi-covalent:
✅ ci-aks: https://github.com/cilium/cilium/actions/runs/5324644428/jobs/9646166215?pr=26382
✅ ci-external-workloads: https://github.com/cilium/cilium/actions/runs/5324644411/jobs/9644243418?pr=26382

removing the temporary commit

.github/workflows/conformance-aks.yaml
```diff
@@ -260,7 +254,7 @@ jobs:
 
       - name: Enable Relay
         run: |
-          cilium hubble enable ${{ steps.vars.outputs.hubble_enable_defaults }}
+          cilium hubble enable
```
sayboras, line 326: should we remove this step and add enabled flag with helm-set ?

.github/workflows/conformance-aks.yaml
```diff
@@ -293,11 +287,12 @@ jobs:
...
         run: |
           cilium install ${{ steps.vars.outputs.cilium_install_defaults }} \
-            --encryption=ipsec
+            --helm-set encryption.enabled=true \
+            --helm-set encryption.type=ipsec
 
       - name: Enable Relay
         run: |
-          cilium hubble enable ${{ steps.vars.outputs.hubble_enable_defaults }}
+          cilium hubble enable
```
sayboras, line 365: same as above

sayboras (approved):
one comment on enabling hubble relay as part of cilium installation, but not a blocker.

LGTM :heavy_check_mark:

.github/workflows/conformance-aks.yaml
michi-covalent: that works too, i don't have a strong preference here 🧘

michi-covalent: let's do this as a follow up if we want to remove this step

.github/workflows/conformance-aks.yaml
```diff
@@ -250,18 +251,11 @@ jobs:
             --helm-set=debug.enabled=true \
             --helm-set loadBalancer.l7.backend=envoy \
             --helm-set tls.secretsBackend=k8s \
-            --azure-resource-group ${{ env.name }}-${{ matrix.location }} \
-            --wait=false \
-            --rollback=false \
```
joestringer, line 255: What happened to wait, rollback?

.github/workflows/conformance-aks.yaml
```diff
@@ -250,18 +251,11 @@ jobs:
...
-            --config monitor-aggregation=none \
-            --version="
```
joestringer, line 257: I guess we no longer need the empty version to trigger development versions because of the way helm CLI mode works..?

.github/workflows/conformance-aks.yaml
michi-covalent: - in helm mode we changed the default to false for `--wait` flag: https://github.com/cilium/cilium-cli/blob/main/internal/cli/cmd/install.go#L252
- we didn't implement `--rollback` flag in helm mode because it doesn't leave dangling resources when the installation fails. we needed this flag in the classic mode in case a resource creation failed midway during the installation to clean up resources that already got created.

.github/workflows/conformance-aks.yaml
michi-covalent: yeah in the helm mode, you either specify the version, or it'll use the default version.

... i'm not even sure if we needed this empty `--version` flag in the classic mode tbh 🤔

joestringer: My recollection was that in classic mode, if you didn't add an empty version flag, it would try to just use the default version hardcoded into the CLI instead of the local chart version.

joestringer: ^ I guess on that front, I see that you ran the tests already on the PR above by editing the workflow files to trigger the changes... Did you confirm that the correct version was set up?

michi-covalent: yeah it's using the local chart, and the image tags match the commit SHA:
```
Image versions         cilium             quay.io/cilium/cilium-ci:d6fb7cb1c5e6b155f704910736bb20252b74a55d: 2
                       hubble-relay       quay.io/cilium/hubble-relay-ci:d6fb7cb1c5e6b155f704910736bb20252b74a55d: 1
                       cilium-operator    quay.io/cilium/operator-generic-ci:d6fb7cb1c5e6b155f704910736bb20252b74a55d: 1
```

joestringer (approved):
There's still a little discussion going on below, but as long as we've validated that the new version does the same testing as the old one on the correct version, LGTM :+1:

christarazi (approved):
:ship: :ship: :ship:

Issue #26082, opened 2023-06-09
CI: FAIL: TestControlPlane/CiliumNodes/v1.24 


ldelossa:
## CI: FAIL: TestControlPlane/CiliumNodes/v1.24 [TravisCI]

```shell
-- FAIL: TestControlPlane (20.56s)

    --- pass: testcontrolplane/cnpstatusnodesgc (1.53s)
    --- pass: testcontrolplane/localnodestore (0.43s)
    --- pass: testcontrolplane/nodehandler (0.48s)
    --- fail: testcontrolplane/ciliumnodes (11.75s)
        --- fail: testcontrolplane/ciliumnodes/v1.24 (10.40s)
        --- pass: testcontrolplane/ciliumnodes/v1.25 (0.67s)
        --- pass: testcontrolplane/ciliumnodes/v1.26 (0.68s)
    --- pass: testcontrolplane/services/dualstack (1.44s)
        --- pass: testcontrolplane/services/dualstack/v1.24 (0.52s)
        --- pass: testcontrolplane/services/dualstack/v1.25 (0.41s)
        --- pass: testcontrolplane/services/dualstack/v1.26 (0.52s)
    --- pass: testcontrolplane/services/gracefultermination (0.62s)
    --- pass: testcontrolplane/services/nodeport (4.32s)
        --- pass: testcontrolplane/services/nodeport/v1.24/nodeport-control-plane (0.51s)
        --- pass: testcontrolplane/services/nodeport/v1.24/nodeport-worker (0.52s)
        --- pass: testcontrolplane/services/nodeport/v1.24/nodeport-worker2 (0.41s)
        --- pass: testcontrolplane/services/nodeport/v1.25/nodeport-control-plane (0.50s)
        --- pass: testcontrolplane/services/nodeport/v1.25/nodeport-worker (0.50s)
        --- pass: testcontrolplane/services/nodeport/v1.25/nodeport-worker2 (0.42s)
        --- pass: testcontrolplane/services/nodeport/v1.26/nodeport-control-plane (0.52s)
        --- pass: testcontrolplane/services/nodeport/v1.26/nodeport-worker (0.44s)
        --- pass: testcontrolplane/services/nodeport/v1.26/nodeport-worker2 (0.50s)
fail
	github.com/cilium/cilium/test/controlplane	coverage: 100.0% of statements
fail	github.com/cilium/cilium/test/controlplane	20.811s

--- fail: testcontrolplane (20.56s)

[...]

PR #26383, opened 2023-06-20
test/controlplane: Disable endpoint GC

pippolo84:
The Endpoint GC is currently not needed in controlplane testing.

Since that GC needs to lock the endpoints map in the endpoint manager, it can possibly delay the execution of the endpoint manager "node subscriber" callback, that needs to hold the same lock.  This in turn might lead to the slow down of the CiliumNodeUpdater callback that needs to wait for the previous callback before running.  The inability of the CiliumNodeUpdater callback to run in time seems to be the possible culprit of the controlplane CiliumNodes test flake.

To mitigate the flake, the endpoint GC is then disabled in controlplane testing.

Related: #26082

### DIFF
--- a/test/controlplane/suite/agent.go
+++ b/test/controlplane/suite/agent.go
@@ -97,6 +97,11 @@ func startCiliumAgent(t *testing.T, clientset k8sClient.Clientset, extraCell cel
 		}),
 	)
 
+	// Unlike global configuration options, cell-specific configuration options
+	// (i.e. the ones defined through cell.Config(...)) must be set to the *viper.Viper
+	// object bound to the test hive.
+	handle.hive.Viper().Set(option.EndpointGCInterval, 0)
+
 	if err := handle.hive.Start(context.TODO()); err != nil {
 		return nil, agentHandle{}, err
 	}


marseel (approved):

nebril (approved):

pippolo84:
> I'm not super convinced this will solve the flake as we are waiting 10s for labels to appear. Can you change `Fixes` to `Related` and let's keep the issue open for now?

Sure, makes sense. Let's keep the issue open for now :+1:

pippolo84:
Since this is just a change to the controlplane test environment setup, running the full set of CI tests is not needed. Marking ready-to-merge

marseel:
Ah yea, I wanted to trigger the `integration-test`, but apparently changes in "test" directory are ignored for these tests.

joestringer:
> Ah yea, I wanted to trigger the integration-test, but apparently changes in "test" directory are ignored for these tests.

Maybe we should change this to ensure that when we modify the control plane tsets, it will run the integration tests? Otherwise we're only relying on developers locally running the tests to confirm that the PR doesn't break something.

(I agree that for this PR the change looks innocent enough it shouldn't introduce a failure, but this is likely to come up again in future, perhaps with a more complicated change, and I don't want to encourage this pattern where we propose PRs, ignore CI and label the PR ready-to-merge to bypass all of the validation processes that apply for every other change that goes into Cilium)

marseel:
Yes, I agree. I was in the middle of opening PR for it :)

PR #26392, opened 2023-06-21
cilium, bigtcp: Add max gso/gro rates to sysdump

borkmann:
See commts.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -391,17 +391,25 @@ func FormatStatusResponse(w io.Writer, sr *models.StatusResponse, sd StatusDetai
 	}
 
 	if sr.IPV4BigTCP != nil {
-		status := "Enabled"
-		if !sr.IPV4BigTCP.Enabled {
-			status = "Disabled"
+		status := "Disabled"
+		if sr.IPV4BigTCP.Enabled {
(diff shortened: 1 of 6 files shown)


borkmann:
Cc @haiyuewa, fyi, might be interesting for you as well. :)

pkg/client/client.go
```diff
@@ -391,15 +391,23 @@ func FormatStatusResponse(w io.Writer, sr *models.StatusResponse, sd StatusDetai
...
 
 	if sr.IPV4BigTCP != nil {
-		status := "Enabled"
+		max := fmt.Sprintf("[%d, %d]", sr.IPV4BigTCP.MaxGRO, sr.IPV4BigTCP.MaxGSO)
+		if sr.IPV4BigTCP.MaxGRO == sr.IPV4BigTCP.MaxGSO {
+			max = fmt.Sprintf("[%d]", sr.IPV4BigTCP.MaxGSO)
+		}
+		status := fmt.Sprintf("Enabled\t%s", max)
 		if !sr.IPV4BigTCP.Enabled {
 			status = "Disabled"
```
aspsk: Looks like if `!sr.IPV4BigTCP.Enabled`, then we do three extra Sprintf's at this point. We could optimize this to one `fmt.Fprintf` per case. (Not a big deal though, as this is just `cilium status`, so ack.)

aspsk (approved):

Issue #22273, opened 2022-11-19
Node clusterpool Endpoint CIDRs used in AWS ENI mode erroneously


jkulesa:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

When using Cilium 1.12.4 in native AWS ENI mode (not chaining),  if you login to a cilium pod you can see the following:

```
$ cilium node list
Name                                           IPv4 Address     Endpoint CIDR   IPv6 Address   Endpoint CIDR
ip-10-196-160-103.us-west-2.compute.internal   10.196.160.103   10.103.0.0/16
ip-10-196-161-128.us-west-2.compute.internal   10.196.161.128   10.128.0.0/16
```

That endpoint CIDR is interesting, since its not any range we've specified (10.196.x.x or 10.10.x.x).   That range looks like its from here:
https://github.com/cilium/cilium/blob/1f342773e4f3a173510cbf1378b613c5f72180b4/pkg/defaults/node.go#L14

But since we are using ENI mode, there aren't Cluster pools to set - and nor can they BE set via the configmap (or helm).

The problem is that this node data is used in various places - one of them is iptables masquerading:

https://github.com/cilium/cilium/blob/1f342773e4f3a173510cbf1378b613c5f72180b4/pkg/datapath/iptables/iptables.go#L1435

...and so this inserts IPs into the masq iptables that don't route or don't participate in any way.  This would make sense when using cluster pools - but not ENI mode (BPF masq mode seems fine in this case)

another place is here:

https://github.com/cilium/cilium/blob/2fe3a92b003c5349df935dfe59de667fa92d73f8/pkg/datapath/linux/node.go#L282

In this case it doesn't break anything but it is trying to generate routes that shouldn't be generated from those 10.x.0.0/16 ranges as DEST:
```
level=warning msg="Unable to install direct node route {Ifindex: 0 Dst: 10.194.0.0/16 Src: <nil> Gw: 10.196.164.194 Flags: [] Table: 0 Realm: 0}" error="route to destination 10.196.164.194 contains gateway 10.196.162.1, must be directly reachable" subsys=linux-datapath
level=warning msg="Unable to install direct node route {Ifindex: 0 Dst: 10.103.0.0/16 Src: <nil> Gw: 10.196.160.103 Flags: [] Table: 0 Realm: 0}" error="route to destination 10.196.160.103 contains gateway 10.196.162.1, must be directly reachable" subsys=linux-datapath
[...]

PR #26397, opened 2023-06-21
iptables: Fix wrong use of podCIDR in cluster node NAT exclusion

gandro:
By default, in the iptables-based masquerading mode, Cilium will only masquerade traffic coming from the local pod CIDR (`allocRange` in `installMasqueradeRules`). However, many IPAM modes such as ENI or multi-pool IPAM do not have a single pod CIDR. Instead, those modes rely on the `egress-masquerade-interfaces` setting, which only masquerades traffic if it leaves one of the `egress-masquerade-interfaces` devices.

Therefore, the "exclude traffic to cluster nodes from masquerade" `CILIUM_POST_nat` rule should also respect the
`egress-masquerade-interfaces` setting and not masquerade traffic regardless of the value of `allocRange` (which will not be valid in settings such as ENI mode).

This likely has not manifested in ENI mode as an issue, because in ENI mode we derive the native routing CIDR (`snatDstExclusionCIDR` in `installMasqueradeRules`) from the EC2 VPC CIDR, which usually contains the node IPs too. However, we should not rely on that, since we are adding additional non-podCIDR based IPAM modes such as multi-pool where this will not be true.

Related: #22273

**Open question:** Should this be backported to v1.13? I'm not aware of any actual bugs this has caused so far in ENI mode (see above), but we should fix it nonetheless. I'd particularly would like to have this in v1.14, since it will fix iptables-based masquerading for the newly added Multi-Pool IPAM mode in v1.14.

### DIFF
--- a/pkg/datapath/iptables/iptables.go
+++ b/pkg/datapath/iptables/iptables.go
@@ -1193,13 +1193,26 @@ func (m *IptablesManager) installMasqueradeRules(prog iptablesInterface, ifName,
 			return err
 		}
 
-		if err := prog.runProg([]string{
+		progArgs := []string{
 			"-t", "nat",
 			"-A", ciliumPostNatChain,
-			"-s", allocRange,
+		}
+
+		// If EgressMasqueradeInterfaces is set, we need to mirror base condition
+		// of the "cilium masquerade non-cluster" rule below, as the allocRange might
+		// not be valid in such setups (e.g. in ENI mode).
+		if option.Config.EgressMasqueradeInterfaces != "" {
+			progArgs = append(progArgs, "-o", option.Config.EgressMasqueradeInterfaces)
+		} else {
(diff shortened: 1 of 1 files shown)


gandro:
ci-ginko failed with LVH provisioning errors https://github.com/cilium/cilium/actions/runs/5334904372/jobs/9667395901 https://github.com/cilium/cilium/actions/runs/5334904372/jobs/9667397270

gandro:
Louis says he does not have time to review

jibi (approved):

gandro:
Marking ready to merge. `gateway-api-conformance-test` are pending since the test was renamed since I ran CI, but the previous version of it has passed: https://github.com/cilium/cilium/actions/runs/5334861019/jobs/9667235765?pr=26397

PR #26402, opened 2023-06-21
CI Workflows: Fix sysdump file creation

brlbil:
Currently, conformance tests that run in matrix strategy do not create sysdump files with a unique name.
Only one report is uploaded in GitHub artifacts. This changes sysdump files names uniq per matrix run.

Test run, it generates sysdump per matrix run.
https://github.com/cilium/cilium/actions/runs/5336461842?pr=26402


viktor-kurchenko (approved):
LGTM

PR #26408, opened 2023-06-21
helm: add .extraEnv to cilium-agents config init container

nberlee:
All (init)containers in the cilium-agent daemonset with envs set allow to set extraEnv. The config initcontainer was the exception, this commit fixes this.

      GitHub issue.

### DIFF
--- a/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
+++ b/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
@@ -434,6 +434,9 @@ spec:
         - name: KUBERNETES_SERVICE_PORT
           value: {{ .Values.k8sServicePort | quote }}
         {{- end }}
+        {{- with .Values.extraEnv }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
         volumeMounts:
         - name: tmp
           mountPath: /tmp


gandro:
Thanks for the PR! I'm curious, was there an issue you were facing with the init container missing those flags? 

I'm not super familiar with the `cilium build-config` feature, but my worry here is that those (clusterwide) env variables might override the per-node `CiliumNodeConfig` which is read during that stage, which I think would not be what we want. But I'm not sure if that would even happen or not.

@squeed You added the `config` init container and the `CiliumNodeConfig` feature. Was there a particular reason why you didn't add the `extraEnv`?

nberlee:
@gandro cilium is automatically applied to all my clusters, clusters with the same use, can have the same cilium manifests, except for the vip.

I solved that by using configmap literals:
```
#k8sServiceHost: 172.26.37.250
extraEnv:
  - name: KUBERNETES_SERVICE_HOST
    valueFrom:
      configMapKeyRef:
        name: clusterconfig
        key: cluster.vip
```

This way I can create a configmap which consists the cluster vip and use the cilium yaml for all my clusters.
The config init container does have `KUBERNETES_SERVICE_HOST`, but doesn't have .extraEnv... I did't test if config actually uses ` KUBERNETES_SERVICE_HOST`, but since it is there if you configure k8sServiceHost, i figured i also needed to change this.

In the meantime I created a kustomization patch, which runs after helm template to workaround this issue...

gandro:
Ahh, that's indeed a very valid use-case. Thanks for providing it!

gandro (approved):

ldelossa:
@nberlee not sure why but there seems to be two stuck tests in this PR, can you rebase `main` and force push up?

gandro:
@tommyp1ckles PTAL

tommyp1ckles (approved):
Looks good, thanks 😄

PR #26410, opened 2023-06-21
vendor: Update vishvananda/netlink/ and x/sys

borkmann:
Pull in recent change to use IFLA_G{S,R}O_IPV4_MAX_SIZE constants from golang.org/x/sys/unix in the netlink library.

Link: https://github.com/vishvananda/netlink/pull/881


tklauser (approved):
Thanks!

PR #26414, opened 2023-06-22
Remove validation timeout in controlplane testing

pippolo84:
Due to the unpredictability of a CI environment, setting a validation timeout when running tests may lead to flakyness.  At the same time, a short timeout is useful when developing locally to get faster feedback from the tests.

The commit changes the controlplane test framework to run without a timeout by default.  An additional method `WithValidationTimeout` is added to set a custom one when developing locally.  Moreover, the result of each failed validation is always printed, to ease debuggability both locally and in a CI environment.

Related: #26082

### DIFF
--- a/test/controlplane/suite/testcase.go
+++ b/test/controlplane/suite/testcase.go
@@ -44,23 +44,20 @@ import (
 	"github.com/cilium/cilium/pkg/proxy"
 )
 
-const (
-	validationTimeout = 10 * time.Second
-)
-
 type trackerAndDecoder struct {
 	tracker k8sTesting.ObjectTracker
 	decoder k8sRuntime.Decoder
 }
 
 type ControlPlaneTest struct {
-	t              *testing.T
-	nodeName       string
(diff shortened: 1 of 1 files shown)


christarazi (commented):
The change makes sense to me. I was reading the discussion in the linked issue and came across [this](https://github.com/cilium/cilium/issues/26082#issuecomment-1602392471), which made me hesitate on approving this PR. Raising it here for us to discuss whether that comment impacts this PR.

pippolo84:
> The change makes sense to me. I was reading the discussion in the linked issue and came across [this](https://github.com/cilium/cilium/issues/26082#issuecomment-1602392471), which made me hesitate on approving this PR. Raising it here for us to discuss whether that comment impacts this PR.

Yep, I'm working to verify what has been suggested. But I think we should consider removing the timeout anyway, to avoid possible further flakyness.

christarazi (approved):

pippolo84:
The only affected tests successfully ran in Travis, so marking it ready-to-merge.

PR #26434, opened 2023-06-23
datapath: migrate tail call maps

### LINKED ISSUE #20691: Always migrate tail call maps

Jack-R-lantern:

      GitHub issue.

Fixes: #20691 

```release-note
```


ti-mo:
Closing in favor of https://github.com/cilium/cilium/pull/26459.

Issue #26074, opened 2023-06-09
Testing Follow-ups for BPF-based IPv6 Masquerading


qmonnet:
Follow-ups for https://github.com/cilium/cilium/pull/23165:

- [x] Add support for the ip-masq-agent
  - PR: https://github.com/cilium/cilium/pull/23219

- [x] Fix ip-masq-agent
  - Issue: https://github.com/cilium/cilium/issues/26262
  - PR: https://github.com/cilium/cilium/pull/26435

- [x] Look into the Host Firewall code and figure out how to run it with BPF IPv6 masquerading. What happens if we have masquerading enable for both IP versions? For one IP version only?
  - Context: https://github.com/cilium/cilium/pull/23165#discussion_r1222689761
  - After addressing it we want to lift the restriction on using BPF IPv6 Masquerading and Host Firewall in `daemon/cmd/daemon_main.go` and in the description of the conformance-e2e workflow
  - PR: https://github.com/cilium/cilium/pull/26323

- [x] Rework `snat_v6_needed()` in bpf/lib/nat.h to bring it closer to `snat_v4_prepare_state()`
  - Context: https://github.com/cilium/cilium/pull/23165#discussion_r1222700605
  - PR: https://github.com/cilium/cilium/pull/26236

- [x] Rework error handling from `ipv6_hdrlen()` and `ct_is_reply6()` in `snat_v6_needed()`, and `ct_is_reply4()` in `snat_v4_prepare_state()`
  - Context: https://github.com/cilium/cilium/pull/23165#discussion_r1222701356
  - PR: https://github.com/cilium/cilium/pull/26501

- [ ] Add BPF integration tests (`bpf/tests/`) for IPv6 Masquerading

- [ ] Enable IPv6 masquerading tests in `datapath_configuration.go` with KPR/BPF masquerading
  - PR: https://github.com/cilium/cilium/pull/26500

- [x] Lift BPF host routing for per-endpoint-routes wrt IPv6 support (for 1.14)
  - PR: https://github.com/cilium/cilium/pull/26205

- [x] IPv6 BIG TCP documentation: fix up ipv6 masquerade=false.. can be removed now
  - PR: https://github.com/cilium/cilium/pull/26172/

- [ ] Once the ip-masq-agent fixes are merged, find a way to move the new test away from Ginkgo

- [x] Mark IPv6 BPF Masquerading as Beta in documentation
  - PR: https://github.com/cilium/cilium/pull/26499

[...]

PR #26435, opened 2023-06-23
Fixes for ip-masq-agent

### LINKED ISSUE #26262: ip-masq-agent broken for IPv4-only
### What happened?

@gandro reports that, after https://github.com/cilium/cilium/pull/23219, he tested the ip-masq-agent in an IPv4-only cluster and observes the following error:

```
Controller Status:      21/22 healthy
  Name                            Last success   Last error   Count   Message
  bpf-map-sync-cilium_ipmasq_v6   never          9s ago       12      loading pinned map /sys/fs/bpf/tc/globals/cilium_ipmasq_v6: no such file or directory   
```

Adding that the ip-masq-agent seems broken, with the IPv4 map not containing any CIDRs anymore.

To be investigated as soon as possible - Worst case, we'll need to revert IPv6 ip-masq-agent before we cut 1.14.

### Cilium Version

`main` branch, on top of https://github.com/cilium/cilium/pull/23219

qmonnet:
It looks like I didn't test the ip-masq-agent quite enough when adding support for IPv6 - in particular, I didn't test with IPv4 only. Sebastian reported that when IPv6 is off, ip-masq-agent is broken. This PR contains a number of fixes to address that.

Fixes: #26262

```release-note
Fix some map handling logic as well as some issues with CLI commands related to ip-masq-agent, introduced with IPv6 support
```

### DIFF
--- a/cilium/cmd/bpf_ipmasq_list.go
+++ b/cilium/cmd/bpf_ipmasq_list.go
@@ -26,7 +26,15 @@ var bpfIPMasqListCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		common.RequireRootPrivilege("cilium bpf ipmasq list")
 
-		cidrs, err := (&ipmasq.IPMasqBPFMap{}).Dump()
+		// Depending on whether BPF masquerading is enabled for IPv4
+		// and IPv6, we may have zero, one, or two maps to dump, and we
+		// need to tell which ones to DumpForProtocols().
+		// Here we try to open the maps as a hack to avoid going
+		// through a full API request to check the config options from
+		// the agent.
+		ipv4Needed := ipmasq.IPMasq4Map().Open() == nil
+		ipv6Needed := ipmasq.IPMasq6Map().Open() == nil
+		cidrs, err := (&ipmasq.IPMasqBPFMap{}).DumpForProtocols(ipv4Needed, ipv6Needed)
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "error dumping contents of map: %s\n", err)
 			os.Exit(1)
--- a/daemon/cmd/datapath.go
+++ b/daemon/cmd/datapath.go
@@ -414,14 +414,16 @@ func (d *Daemon) initMaps() error {
 		datapathIpcache.NewListener(d, d, d.ipcache),
 	})
 
-	if option.Config.EnableIPv4 && option.Config.EnableIPMasqAgent {
-		if err := ipmasq.IPMasq4Map().OpenOrCreate(); err != nil {
-			return fmt.Errorf("initializing IPv4 masquerading map: %w", err)
+	if option.Config.EnableIPMasqAgent {
+		if option.Config.EnableIPv4Masquerade {
+			if err := ipmasq.IPMasq4Map().OpenOrCreate(); err != nil {
+				return fmt.Errorf("initializing IPv4 masquerading map: %w", err)
+			}
 		}
-	}
-	if option.Config.EnableIPv6 && option.Config.EnableIPMasqAgent {
-		if err := ipmasq.IPMasq6Map().OpenOrCreate(); err != nil {
-			return fmt.Errorf("initializing IPv6 masquerading map: %w", err)
+		if option.Config.EnableIPv6Masquerade {
+			if err := ipmasq.IPMasq6Map().OpenOrCreate(); err != nil {
+				return fmt.Errorf("initializing IPv6 masquerading map: %w", err)
+			}
 		}
 	}
 
--- a/pkg/datapath/maps/map.go
+++ b/pkg/datapath/maps/map.go
@@ -204,13 +204,12 @@ func (ms *MapSweeper) RemoveDisabledMaps() {
 		maps = append(maps, lbmap.SourceRange6MapName, lbmap.SourceRange4MapName)
 	}
 
-	if !option.Config.EnableIPMasqAgent {
-		if option.Config.EnableIPv4 {
-			maps = append(maps, ipmasq.MapNameIPv4)
-		}
-		if option.Config.EnableIPv6 {
-			maps = append(maps, ipmasq.MapNameIPv6)
-		}
+	if !(option.Config.EnableIPMasqAgent && option.Config.EnableIPv4Masquerade) {
+		maps = append(maps, ipmasq.MapNameIPv4)
+	}
+
+	if !(option.Config.EnableIPMasqAgent && option.Config.EnableIPv6Masquerade) {
+		maps = append(maps, ipmasq.MapNameIPv6)
 	}
 
 	if !option.Config.EnableXDPPrefilter {
--- a/pkg/maps/ipmasq/ipmasq.go
+++ b/pkg/maps/ipmasq/ipmasq.go
@@ -86,18 +86,28 @@ type IPMasqBPFMap struct{}
 
 func (*IPMasqBPFMap) Update(cidr net.IPNet) error {
 	if ip.IsIPv4(cidr.IP) {
-		return IPMasq4Map().Update(keyIPv4(cidr), &Value{})
+		if option.Config.EnableIPv4Masquerade {
+			return IPMasq4Map().Update(keyIPv4(cidr), &Value{})
+		}
 	} else {
-		return IPMasq6Map().Update(keyIPv6(cidr), &Value{})
+		if option.Config.EnableIPv6Masquerade {
+			return IPMasq6Map().Update(keyIPv6(cidr), &Value{})
+		}
 	}
+	return nil
 }
 
 func (*IPMasqBPFMap) Delete(cidr net.IPNet) error {
 	if ip.IsIPv4(cidr.IP) {
-		return IPMasq4Map().Delete(keyIPv4(cidr))
+		if option.Config.EnableIPv4Masquerade {
+			return IPMasq4Map().Delete(keyIPv4(cidr))
+		}
 	} else {
-		return IPMasq6Map().Delete(keyIPv6(cidr))
+		if option.Config.EnableIPv6Masquerade {
+			return IPMasq6Map().Delete(keyIPv6(cidr))
+		}
 	}
+	return nil
 }
 
 func (*IPMasqBPFMap) Dump() ([]net.IPNet, error) {
--- a/pkg/maps/ipmasq/ipmasq.go
+++ b/pkg/maps/ipmasq/ipmasq.go
@@ -110,18 +110,24 @@ func (*IPMasqBPFMap) Delete(cidr net.IPNet) error {
 	return nil
 }
 
-func (*IPMasqBPFMap) Dump() ([]net.IPNet, error) {
+// DumpForProtocols dumps the contents of the ip-masq-agent maps for IPv4
+// and/or IPv6, as requested by the caller.
+// Given that the package does not expose the maps directly, it's necessary to
+// specify which protocol we need when ipMasq4Map/ipMasq6Map, or config
+// options, have not been set, as is the case when calling from the CLI, for
+// example.
+func (*IPMasqBPFMap) DumpForProtocols(ipv4Needed, ipv6Needed bool) ([]net.IPNet, error) {
 	cidrs := []net.IPNet{}
-	if ipMasq4Map != nil {
-		if err := ipMasq4Map.DumpWithCallback(
+	if ipv4Needed {
+		if err := IPMasq4Map().DumpWithCallback(
 			func(keyIPv4 bpf.MapKey, _ bpf.MapValue) {
 				cidrs = append(cidrs, keyToIPNetIPv4(keyIPv4.(*Key4)))
 			}); err != nil {
 			return nil, err
 		}
 	}
-	if ipMasq6Map != nil {
-		if err := ipMasq6Map.DumpWithCallback(
+	if ipv6Needed {
+		if err := IPMasq6Map().DumpWithCallback(
 			func(keyIPv6 bpf.MapKey, _ bpf.MapValue) {
 				cidrs = append(cidrs, keyToIPNetIPv6(keyIPv6.(*Key6)))
 			}); err != nil {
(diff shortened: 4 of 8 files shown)


test/k8s/datapath_configuration.go
```diff
@@ -361,6 +361,18 @@ var _ = Describe("K8sDatapathConfig", func() {
 			testIPMasqAgent()
 		})
 
+		It("DirectRouting, IPv4 only", func() {
```
brb, line 364: JFYI, we want to move this to CLI connectivity test in v1.15.

brb (approved):

test/k8s/datapath_configuration.go
qmonnet: OK thanks, this shouldn't be an issue I think? Or do you have an alternative to suggest?

bimmlerd (approved):
contributing lgtm

learnitall (approved):
LGTM on changes in examples/

christarazi (approved):
LGTM for ci-structure. It would be good to have a tracking issue for migrating the test out of Ginkgo.

qmonnet:
Ok, I added an item to https://github.com/cilium/cilium/issues/26074. Where would be the best location for this test? Basically I want to make sure that Cilium deploys well with the ip-masq-agent on IPv4 only (and that we don't reintroduce this IPv6-map-related warning that would make it unhealthy) and that the feature works as expected. Would connectivity tests be the way to go (as a follow-up)?

test/k8s/datapath_configuration.go
brb: For this PR it's not an issue. I'm expecting the move to happen early in the v1.15 dev cycle.

pkg/datapath/maps/map.go
```diff
@@ -204,13 +204,12 @@ func (ms *MapSweeper) RemoveDisabledMaps() {
...
 	}
 
-	if !option.Config.EnableIPMasqAgent {
-		if option.Config.EnableIPv4 {
-			maps = append(maps, ipmasq.MapNameIPv4)
-		}
-		if option.Config.EnableIPv6 {
-			maps = append(maps, ipmasq.MapNameIPv6)
-		}
+	if !option.Config.EnableIPMasqAgent || !option.Config.EnableIPv4Masquerade {
```
lmb: Nit: took me a bit to parse this. Maybe `!(EnableIPMasqAgent && EnableIPv4Masquerade)` is easier?

cilium/cmd/bpf_ipmasq_list.go
```diff
@@ -26,7 +26,7 @@ var bpfIPMasqListCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		common.RequireRootPrivilege("cilium bpf ipmasq list")
 
-		cidrs, err := (&ipmasq.IPMasqBPFMap{}).Dump()
+		cidrs, err := (&ipmasq.IPMasqBPFMap{}).DumpForProtocols(getMasqueradingProtocols())
```
lmb: `getMasqueradingProtocols` seems like a pretty big hammer to me. Could you instead call `IPMasq6Map` and `IPMasq4Map` (ignoring any "file doesn't exist" error) before `Dump()`?

gandro (approved):
Looks good to me, thanks for fixing this!

cilium/cmd/bpf_ipmasq_list.go
qmonnet: As discussed offline -- This won't work, because we're trying to `Open()` the map in `DumpWithCallback()`, so if we initialise `ipmasq.ipMasq6Map` by calling `ipmasq.IPMasq6Map()` in the CLI, we get an error:

```console
root@kind-worker:/home/cilium# cilium bpf ipmasq list
error dumping contents of map: loading pinned map /sys/fs/bpf/tc/globals/cilium_ipmasq_v6: no such file or directory
```

I'll follow your alternate suggestion instead, using the result of `ipv4Needed := ipmasq.IPMasq4Map().Open() == nil` (and v6) to check whether the maps are open and get the booleans to pass to `DumpForProtocols()`, instead of going for the API request.

Thanks for the review!

lmb (approved):

daemon/cmd/datapath.go
```diff
@@ -414,12 +414,12 @@ func (d *Daemon) initMaps() error {
 		datapathIpcache.NewListener(d, d, d.ipcache),
 	})
 
-	if option.Config.EnableIPv4 && option.Config.EnableIPMasqAgent {
+	if option.Config.EnableIPv4Masquerade && option.Config.EnableIPMasqAgent {
```
derailed: We could refactor this a bit here as both checks only happen if EnableIPMasqAgent is set.

pkg/datapath/maps/map.go
```diff
@@ -204,13 +204,12 @@ func (ms *MapSweeper) RemoveDisabledMaps() {
...
 	}
 
-	if !option.Config.EnableIPMasqAgent {
-		if option.Config.EnableIPv4 {
-			maps = append(maps, ipmasq.MapNameIPv4)
-		}
-		if option.Config.EnableIPv6 {
-			maps = append(maps, ipmasq.MapNameIPv6)
-		}
+	if !(option.Config.EnableIPMasqAgent && option.Config.EnableIPv4Masquerade) {
```
derailed, line 207: Same here don't need to recheck EnableIPMasqAgent is not set.

pkg/maps/ipmasq/ipmasq.go
```diff
@@ -86,32 +86,42 @@ type IPMasqBPFMap struct{}
 
 func (*IPMasqBPFMap) Update(cidr net.IPNet) error {
 	if ip.IsIPv4(cidr.IP) {
```
derailed, line 88: Could combine the if and the one below and return if ipv4 is enable. No need for else clause

pkg/maps/ipmasq/ipmasq.go
```diff
@@ -86,32 +86,42 @@ type IPMasqBPFMap struct{}
...
+			return IPMasq6Map().Update(keyIPv6(cidr), &Value{})
+		}
 	}
+	return nil
 }
 
 func (*IPMasqBPFMap) Delete(cidr net.IPNet) error {
 	if ip.IsIPv4(cidr.IP) {
-		return IPMasq4Map().Delete(keyIPv4(cidr))
+		if option.Config.EnableIPv4Masquerade {
```
derailed, line 102: Same here.

pkg/maps/ipmasq/ipmasq.go
```diff
@@ -86,32 +86,42 @@ type IPMasqBPFMap struct{}
...
+		if option.Config.EnableIPv6Masquerade {
+			return IPMasq6Map().Delete(keyIPv6(cidr))
+		}
 	}
+	return nil
 }
 
-func (*IPMasqBPFMap) Dump() ([]net.IPNet, error) {
+func (*IPMasqBPFMap) DumpForProtocols(ipv4Needed, ipv6Needed bool) ([]net.IPNet, error) {
 	cidrs := []net.IPNet{}
```
derailed, line 120: nit: Prefer var cidrs []net.IPNet{} or pre-allocate if known size to ease up on GC

pkg/maps/ipmasq/ipmasq.go
```diff
@@ -121,6 +131,10 @@ func (*IPMasqBPFMap) Dump() ([]net.IPNet, error) {
 	return cidrs, nil
 }
 
+func (*IPMasqBPFMap) Dump() ([]net.IPNet, error) {
```
derailed, line 142: Still used? Also no comment on exported method. 

Perhaps a larger discussion here but this sounds like prime candidates for generics. 
There is very little use for this type as shown below as since this is just a `marker` type that just keep checking what kind of ips are at play and whether the correct matching configs are set. 
> Note: in this case we need an instance that is passed in on the fly.

pkg/maps/ipmasq/ipmasq.go
```diff
@@ -86,32 +86,42 @@ type IPMasqBPFMap struct{}
...
 func (*IPMasqBPFMap) Update(cidr net.IPNet) error {
 	if ip.IsIPv4(cidr.IP) {
-		return IPMasq4Map().Update(keyIPv4(cidr), &Value{})
+		if option.Config.EnableIPv4Masquerade {
+			return IPMasq4Map().Update(keyIPv4(cidr), &Value{})
+		}
 	} else {
-		return IPMasq6Map().Update(keyIPv6(cidr), &Value{})
+		if option.Config.EnableIPv6Masquerade {
+			return IPMasq6Map().Update(keyIPv6(cidr), &Value{})
```
derailed, line 94: Same here

pkg/maps/ipmasq/ipmasq.go
```diff
@@ -86,32 +86,42 @@ type IPMasqBPFMap struct{}
 
 func (*IPMasqBPFMap) Update(cidr net.IPNet) error {
 	if ip.IsIPv4(cidr.IP) {
-		return IPMasq4Map().Update(keyIPv4(cidr), &Value{})
+		if option.Config.EnableIPv4Masquerade {
+			return IPMasq4Map().Update(keyIPv4(cidr), &Value{})
```
derailed, line 90: We should be using the Value() constructor.

derailed (changes requested):
@qmonnet Nice work! Just a few picks...

pkg/datapath/maps/map.go
qmonnet: Are you sure about this one? I don't think I can group both IPv4 and IPv6 options under `if !option.Config.EnableIPMasqAgent {` here, the condition wouldn't be the same. I want to append the map if either `EnableIPMasqAgent` or `EnableIPv4Masquerade` is unset.

pkg/maps/ipmasq/ipmasq.go
qmonnet: Couldn't?

I understand you mean the following:

```go
	if ip.IsIPv4(cidr.IP) && option.Config.EnableIPv4Masquerade {
		return IPMasq4Map().Update(keyIPv4(cidr), &Value{})
	}
	if option.Config.EnableIPv6Masquerade {
		return IPMasq6Map().Update(keyIPv6(cidr), &Value{})
	}
	return nil
```

What if the CIDR is IPv4, `EnableIPv4Masquerade` is not set, but `EnableIPv6Masquerade` is set? In this case, without the `else`, we would attempt to add the CIDR to the IPv6 map, instead of skipping the entry.

Or did I get you wrong?

pkg/maps/ipmasq/ipmasq.go
qmonnet: Still used, in the agent. The method didn't have a comment so far and I didn't think of adding one; but I can do, for both.

As for the generics, it makes sense, but have we been using them so far in the code base? This PR is a set of fixes, I'd as well avoid moving things too much or introducing generics in here.

pkg/maps/ipmasq/ipmasq.go
qmonnet: OK but we don't know the size here, this depends on contents of the maps. The maps themselves have a fixed size at the moment and I could allocate for the maximal size, but that wouldn't be efficient in terms of memory usage. Do you think it's worth it?

pkg/maps/ipmasq/ipmasq.go
derailed: Totally Quentin, I understand. Just a quick note, for potential refacs at some other time.

pkg/maps/ipmasq/ipmasq.go
derailed: Thanks Quentin! I see what your are saying. You would know best here is there is something we could use based on the expected/average number of cidrs coming back.  Then this would avoid some GC churn. But this is an optimization and I understand you're trying to push thru bugs so feel free to disregard if not a good time to do so.

pkg/maps/ipmasq/ipmasq.go
derailed: Rats! I see what you are saying. You are totally correct!

pkg/maps/ipmasq/ipmasq.go
qmonnet: OK thank you! The only value we could use is the max size for the map, otherwise we'll get partial dumps if users ever fill their maps. Now I've got very little idea of whether the trade-off between GC churn and memory consumption would be interesting if we did so. As you say, I'll consider it beyond the scope of the PR and will leave it for another day. But thanks!

pkg/datapath/maps/map.go
derailed: I was thinking something along this line.

```go
  if !option.Config.EnableIPMaskAgent {
      switch {
      case !...EnableIPV4...:
      case !...EnableIPV6...:
      }
    }
```

qmonnet: But we also want to append the map if `option.Config.EnableIPv4Masquerade` is unset, _but_ `option.Config.EnableIPMaskAgent` is set. But with your suggestion we'll never reach the `switch` in that case?

[EDIT: Based on this comment, I left this check unchanged in the last revision.]

pkg/maps/ipmasq/ipmasq.go
qmonnet: Let's keep the `else`, same as above, as per the other discussion

qmonnet:
I addressed the feedback from Fernand - I factored a check in `daemon/cmd/datapath.go`, and added comments for the two exported `Dump*()` functions - and rebased.

pkg/datapath/maps/map.go
derailed: Ah makes sense.  Tx Quentin!

derailed (approved):
@qmonnet LGTM. Thank you for the updates!

qmonnet:
All codeowners are covered by reviews, and CI is green. Let's get these fixes merged

PR #26453, opened 2023-06-23
nat: fix missing L3 pseudo-header csum diff applied into L4

sahid:
In context of an ICMP need to frag received, the NAT happening in the embedded packet is missing applying the L3 pseudo-header csum diff into the TCP/UDP header.

This commit is fixing the bug by reusing same pattern used in other rewrite_ functions.

### DIFF
--- a/bpf/lib/nat.h
+++ b/bpf/lib/nat.h
@@ -359,18 +359,18 @@ snat_v4_icmp_rewrite_ingress_embedded(struct __ctx_buff *ctx,
 				      struct ipv4_nat_entry *state,
 				      __u32 l4_off, __u32 inner_l4_off)
 {
+	int ret;
+	int flags = BPF_F_PSEUDO_HDR;
 	struct csum_offset csum = {};
 	__be32 sum;
 
 	if (state->to_daddr == tuple->daddr &&
 	    state->to_dport == tuple->dport)
 		return 0;
-
 	sum = csum_diff(&tuple->daddr, 4, &state->to_daddr, 4, 0);
 	csum_l4_offset_and_flags(tuple->nexthdr, &csum);
-	if (state->to_dport != tuple->dport) {
-		__be32 suml4 = 0;
 
+	if (state->to_dport != tuple->dport) {
 		switch (tuple->nexthdr) {
 		case IPPROTO_TCP:
 		case IPPROTO_UDP:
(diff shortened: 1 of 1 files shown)


sahid:
o/ @jschwinger233, @ti-mo, @borkmann can you have a look? It's the following patch

bpf/lib/nat.h
```diff
@@ -359,30 +359,30 @@ snat_v4_icmp_rewrite_ingress_embedded(struct __ctx_buff *ctx,
 				      struct ipv4_nat_entry *state,
 				      __u32 l4_off, __u32 inner_l4_off)
```
ldelossa, line 360: Not part of your PR, but when we get into this function `l4_off` is really `icmphdr_off`, right? If so might be a nice change while this code is being edited here.

ldelossa (approved):
Changes LGTM.

bpf/lib/nat.h
```diff
@@ -421,6 +415,9 @@ snat_v4_icmp_rewrite_ingress_embedded(struct __ctx_buff *ctx,
 		return DROP_WRITE_ERROR;
 	if (ipv4_csum_update_by_diff(ctx, l4_off + sizeof(struct icmphdr), sum) < 0)
 		return DROP_CSUM_L3;
+	if (csum.offset &&
+	    csum_l4_replace(ctx, inner_l4_off, &csum, 0, sum, flags) < 0)
```
jschwinger233, line 419: In the case of UDP, this `csum_l4_replace` will run for twice. The first run is inside the `l4_modify_port` as you revised, the second run is here as `csum.offset` is non-zero.

bpf/lib/nat.h
```diff
@@ -395,23 +395,17 @@ snat_v4_icmp_rewrite_ingress_embedded(struct __ctx_buff *ctx,
...
 			break;
-		}
-		default:
-			return DROP_UNKNOWN_L4;
-		}
-		/* By recomputing L4 checksum of inner packet we avoid having
-		 * to recompute L4 of the ICMP Error.
-		 */
-		suml4 = csum_diff(&tuple->dport, 4, &state->to_dport, 4, 0);
-		if (csum_l4_replace(ctx, inner_l4_off, &csum, 0, suml4, 0) < 0)
```
jschwinger233, line 413: I understand the bug you want to fix: the original `csum_l4_replace` wasn't called with flag `BPF_F_PSEUDO_HDR`. Then how about simply passing a `flag` variable to it?

PR #24596, opened 2023-03-28
Operator: Improve removing stale CEPs from CESs entries on start.

alan-kut:
Handle situation when the CEP is duplicated in CESs.
Keep only one CEP if the CEP still exists or
delete all the instnaces if it doesn't exist anymore.
Prefer keeping CEP with the correct identity if such exists.

It's related to #[24581](https://github.com/cilium/cilium/issues/24581) but it doesn't fully fix it.
It only fixes the reconciliation at the startup.

```release-note
The operator now reconciles duplicate entries in a CiliumEndpointSlice on startup.
```

PR #26455, opened 2023-06-23
operator: Handle conflicts in CES update.

alan-kut:
If for some reason operator has outdated version of CES it will not be able to update such CES and it will never recover from such state.

This can happen not only due to some othe client updating CES but also when the update from operator succeeds but for some reason api-server does't return OK (it can fail after updating etcd).

Signed-off-by:  Alan Kutniewski <kutniewski@google.com>

```release-note
Fix operator entering broken state when it has outdated version of the CES in the cache.
```

### DIFF
--- a/operator/pkg/ciliumendpointslice/endpointslice.go
+++ b/operator/pkg/ciliumendpointslice/endpointslice.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"golang.org/x/time/rate"
+	"k8s.io/apimachinery/pkg/api/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
@@ -316,6 +317,14 @@ func (c *CiliumEndpointSliceController) handleErr(err error, key interface{}) {
 	// Increment error count for sync errors
 	if operatorOption.Config.EnableMetrics {
 		metrics.CiliumEndpointSliceSyncErrors.Inc()
+	}
+
+	if errors.IsConflict(err) {
+		// Update metadata of the object from store on conflict
+		obj, exists, err := c.ciliumEndpointSliceStore.GetByKey(key.(string))
+		if err == nil && exists {
+			c.Manager.updateCESInCache(obj.(*v2alpha1.CiliumEndpointSlice), false)
+		}
 	}
 
 	if c.queue.NumRequeues(key) < maxRetries {
--- a/operator/pkg/ciliumendpointslice/endpointslice_test.go
+++ b/operator/pkg/ciliumendpointslice/endpointslice_test.go
@@ -6,13 +6,16 @@ package ciliumendpointslice
 import (
 	"context"
 	"errors"
+	"fmt"
 	"strconv"
 	"sync"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/tools/cache"
 
 	v2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
@@ -261,4 +264,31 @@ func TestUsedIdentitiesInCESs(t *testing.T) {
 	wantIdentities["50"] = true
 	gotIdentities = usedIdentitiesInCESs(cesStore)
 	assertEqualIDs(t, wantIdentities, gotIdentities)
+}
+
+func TestHandleErr(t *testing.T) {
+	t.Run("Cache is updated on conflict", func(t *testing.T) {
+		_, client := client.NewFakeClientset()
+		cesController := NewCESController(context.Background(), &sync.WaitGroup{}, client, 5, cesIdentityBasedSlicing, 10, 20)
+		manager := cesController.Manager
+
+		key := "some-ces"
+		initialCES := createCESWithIDs(key, []int64{1, 2})
+		initialCES.Generation = 1
+		manager.createCES(key)
+		manager.updateCESInCache(initialCES, true)
+		updatedCES := createCESWithIDs(key, []int64{1, 2})
+		updatedCES.Generation = 2
+		cesController.ciliumEndpointSliceStore.Add(updatedCES)
+
+		var err error
+		err = k8s_errors.NewConflict(
+			schema.GroupResource{Group: "", Resource: "ciliumendpointslices"},
+			key,
+			fmt.Errorf("conflict"))
+		cesController.handleErr(err, key)
+
+		managerCES, _ := manager.getCESFromCache(key)
+		assert.Equal(t, updatedCES.Generation, managerCES.Generation)
+	})
 }


alan-kut:
\assign dlapcevic

sayboras:
Can you also help to correct the sign-off statement in the commit message?

```
Signed-off-by:  Alan Kutniewski <kutniewski@google.com>
```

dlapcevic (approved):
/lgtm

operator/pkg/ciliumendpointslice/reconciler.go
```diff
@@ -68,14 +73,43 @@ func (r *reconciler) reconcileCESUpdate(cesToUpdate string) (err error) {
...
+				wasConflict = true
+				// Update metadata of the object from store on conflict
+				obj, exists, getErr := r.cesStore.GetByKey(cesToUpdate)
+				if getErr != nil {
+					return getErr
+				}
+				if !exists {
+					return err
+				}
+				ces.ObjectMeta = obj.(*cilium_v2.CiliumEndpointSlice).ObjectMeta
```
chancez: This looks a little questionable. Might want to see if you can update the obj.Spec/obj.Status instead, otherwise you might overwrite values added by another update.

dlapcevic: I think it's fine because the operator doesn't want to modify CES ObjectMeta.
The operator creates CESs and then only modifies the Endpoints field.

The only remaining concern is that updates will still fail if another source (other than operator) modifies the Endpoints field, but the aim of this PR is to fix the issue of operator getting stuck on its own, just as a part of the k8s system.

chancez: I'm not worried about the ObjectMeta data, I'm worried that it's overwriting the entire rest of the object without concern for it's previous values.

alan-kut: I don't understand it. I intended to overwrite ObjectMeta only and I thought this is what I did.
This is the same as updateCESInCache code called later:
```
func (c *cesMgr) updateCESInCache(srcCES *cilium_v2.CiliumEndpointSlice, isDeepCopy bool) {
    if ces, ok := c.desiredCESs.getCESTracker(srcCES.GetName()); ok {
        ces.backendMutex.Lock()
        defer ces.backendMutex.Unlock()
        if !isDeepCopy {
            ces.ces.ObjectMeta = srcCES.ObjectMeta
        } else {
            ces.ces = srcCES
                for _, cep := range ces.ces.Endpoints {
                    // Update the desiredCESs, to reflect all CEPs are packed in a CES
                    c.desiredCESs.insertCEP(GetCEPNameFromCCEP(&cep, ces.ces.Namespace), srcCES.GetName())
                }
            }
        } else {
            log.WithFields(logrus.Fields{
            logfields.CESName: srcCES.GetName(),
            }).Debug("Attempted to updateCESInCache non-existent, skipping.")
    }
}
```

alan-kut: Ah, you are worried that in the apisever we ignore all other possible changes.

What is the case in which we shouldn't do this?
Operator should be the only controller for CES.

If some client changes k8s.Endpoints etc the controllers owning them will revert such changes. The same would happen here (to some extent) with the CiliumEndpointSlice.

Long term I think the operator should always reconcile the full state of CiliumEndpointSlice and make sure they are exactly as operator expects them to be.

squeed: Since the operator "owns" the CiliumEndpointSlice, it should just unset ResourceVersion from the Update request. Then the apiserver will never reject it for concurrency issues, and there will be no need to retry.

chancez: > Operator should be the only controller for CES.

If we run multiple workers (go routines) (I don't know if this is the case), then the operator could still be racing itself.

The problem here is that we're modifying our cache, which makes the control flow of state updates bi-directional. In a controller, we would ideally not mutate the cache directly but wait for the reconciliation and update after the cache is updated.

aojea: > In a controller, we would ideally not mutate the cache directly but wait for the reconciliation and update after the cache is updated.

totally agree, but my observation in the cilium codebase is that there is no implementation of the reconciler pattern (unfortunately) and everything is based on an event based pattern using `Upsert`calls and caches on top of caches (like this one, if we already have the informer cache with the state of the world we should not need an additional one)

Is the existing code implementing reconcile loops?

alan-kut: I'd like to go into full reconciler direction:
On event and every some time operator could compare expected state it knows based on CEPs, IDs with the existing state from the informer cache and make changes.

We are not there yet. Currently if CES is updated in apiserver but not in the cache (updated by other client or updated by operator but not updated cache (e.g. server updated but returned error so operator didn't updated cache) then the state for such single CES will be broken and won't reconcile until operator is restarted. It won't be able to remove old endpoints from the slice or add new endpoints to the slice.

aojea: > Since the operator "owns" the CiliumEndpointSlice, it should just unset ResourceVersion from the Update request.

resource version MUST be set on an Update 😄 , you can trigger a CreateOnUpdate for some specific objtect but that is not for update the object, @squeed are you referring to some special case?

The canonical solution to solve multiple controllers ownership in kubernetes is to use server side apply https://kubernetes.io/docs/reference/using-api/server-side-apply/ but I don't whink we have this problem here, I think we should try with a PATCH instead of an UPDATE

alan-kut: I'd keep it they way I did it. It keeps the original, current implementation until very rare case of conflict in which it would override the spec but keep the metadata from the apiserver.

IMO this is intended and expected behavior for operator to fully control the CESs.

squeed: > resource version MUST be set on an Update 😄 

It doesn't have to be :-). It's just good practice in the event of concurrency.

Given that we don't have multiple controllers, unsetting ResourceVersion is the functional equivalent to setting Force in a server-side apply. In this case, it is the safe thing to do.

squeed: > keep the metadata from the apiserver.

The apiserver will do that for you whether or not the ResourceVersion is set. Either the fields are not writable by clients and ignored (e.g. creationDate, UID, etc), or they are writable and need to be provided on Update.

ResourceVersion is *just* a way to ensure another client doesn't step on your toes in a read-update-write scenario.

aojea: seem Casey is right, but follow the discussion with api-machinery folks https://kubernetes.slack.com/archives/C0EG7JC6T/p1687868578131159 so we fully understand this

aojea: Ok, Casey and I asked for help and we found the explanation, long history in https://kubernetes.slack.com/archives/C0EG7JC6T/p1687868578131159

It turns out some types allow "UnconditionalUpdates" that is what Casey describe in his example, unfortunatly this is not available for CustomResources, and the recommendation is to use Patches

https://github.com/kubernetes/kubernetes/issues/83222#issuecomment-553408478

squeed: Yup, I'm wrong, but for an interesting reason :-).

aojea: I was right by chance, I didn't knew about "UnconditionalUpdates" , lucky we have Stefan to clarify these things :)

alan-kut: How can we resolve this conversation.

I see two options:
1) The way I did it. Keep current behavior until conflict, update metadata on conflict.
It's not nice and clean (I hope I'll have time to implement proper reconciliation logic but it would be much bigger change). The biggest advantage is that nothing is changed until conflict (which is really rare).

2) Use PATCH instead of UPDATE. It will look better probably. There will be no conflicts. But it would require much bigger change to the way how we update CESs in all the cases.

@aojea @squeed WDYT?

aojea: hmm ... there is a reconcile loop that will retry later

Or I miss something or for me the bug is that the cache is updated after doing the Update operation, and checking the `updateCESInCache` logic I think that this has to happen before, since is the one that gets the desired state

```diff
diff --git a/operator/pkg/ciliumendpointslice/endpointslice.go b/operator/pkg/ciliumendpointslice/endpointslice.go
index ce68044797..c515c9844e 100644
--- a/operator/pkg/ciliumendpointslice/endpointslice.go
+++ b/operator/pkg/ciliumendpointslice/endpointslice.go
@@ -351,6 +351,8 @@ func (c *CiliumEndpointSliceController) syncCES(key string) error {
                                return err
                        }
                } else {
+                       // Update local datastore CES with latest object metadata values.
+                       c.Manager.updateCESInCache(ces, false)
                        if err := c.reconciler.reconcileCESUpdate(key); err != nil {
                                return err
                        }
```

alan-kut: I don't think it would work better.

It would fix the problem with conflict eventually but it would introduce more problem.
It would always use metadata from the watcher cache so the metadata would be outdated after update before watch gets the event.

This is happening right now with creates. We determine create vs update based on the watch cache, we create if CES is not present in the watch cache. So we observed it happening that we create the CES and operator wants to update it before the watcher obeserves the event so it tries to create again and fails.

alan-kut: I could move my change to the handleErr - I think it makes more sense, will update the PR soon

aojea: > It would always use metadata from the watcher cache so the metadata would be outdated after update before watch gets the event.


that is how kubernetes controllers work, there is no pass through cache and you have to operate with the state of the informer, it eventually be consistent

aojea: > This is happening right now with creates. We determine create vs update based on the watch cache, we create if CES is not present in the watch cache

that is the part I don't understand, why are those two different operations?

alan-kut: PTAL

I changed the place for handling conflict errors to errorhandling in endpointslices.

squeed (approved):

brb:
Please rebase against the latest `main` branch to fix the `ci-e2e` failures.

alan-kut:
All tests passed except flaky K8sUpstreamNetConformance / kubernetes-e2e-net-conformance (ipv4)

aojea:
/lgtm

> K8sUpstreamNetConformance / kubernetes-e2e-net-conformance (ipv4) (pull_request

this job is flaky and not required and does not seem related, is on my backlog and I hope I can get to it as it is causing a lot of troubles

squeed:
removing vendor team from reviewers list now that the PR has no vendor changes.

sayboras (approved):
LGTM :heavy_check_mark:

alan-kut:
Thanks for the review. Can we merge it now?

alan-kut:
@squeed @sayboras Can we merge this PR?

squeed:
> @squeed @sayboras Can we merge this PR?

Yup, we'll make it happen.

Should this be backported? If so, back to which version?

alan-kut:
It is affecting all Cilium (at least v1.11+).
We can backport it all the way we can.
On the other side it's quite uncommon to hit this issue.

squeed:
All green!

nbusseneau:
@alan-kut @squeed This PR cannot be nicely backported to all stable branches as it depends on changes made in #24596, which was introduced recently and was not backported.

I am holding off on backporting this one for now, please advise as to whether we want to backport the dependent PR as well or skip backporting this one.

joestringer:
CiliumEndpointSlice is marked as beta, I don't see a good argument for backporting to 1.12 and 1.11 at this stage. Even 1.13 is a little bit questionable with 1.14 around the corner.

PR #26457, opened 2023-06-23
egressgw: retry getIdentityLabels on failure

### LINKED ISSUE #26158: CI: Conformance E2E - egress-gateway
## CI failure

Hit on https://github.com/cilium/cilium/pull/26109: https://github.com/cilium/cilium/actions/runs/5252755259/jobs/9489217504
Matrix entry: (7, bpf-next-20230420.212204, none, strict, disabled, snat, true, testing-only)

```
[=] Test [egress-gateway]

ℹ️  📜 Applying CiliumEgressGatewayPolicy 'cegp-sample-echo-service' to namespace ''..
ℹ️  📜 Applying CiliumEgressGatewayPolicy 'cegp-sample' to namespace ''..
[-] Scenario [egress-gateway/egress-gateway]
🟥 Failed to ensure egress gateway policy map is properly populated: timeout while waiting for condition, last error: Could not find egress gateway policy entry matching {SourceIP:10.244.3.62 DestCIDR:0.0.0.0/0 EgressIP:0.0.0.0 GatewayIP:172.18.0.5}
ℹ️  📜 Deleting CiliumEgressGatewayPolicy 'cegp-sample' from namespace ''..
ℹ️  📜 Deleting CiliumEgressGatewayPolicy 'cegp-sample-echo-service' from namespace ''..

connectivity test failed: 1 tests failed
📋 Test Report
❌ 1/50 tests failed (0/539 actions), 10 tests skipped, 0 scenarios skipped:
Test [egress-gateway]:
```

Sysdump: [cilium-sysdump-7-final.zip](https://github.com/cilium/cilium/files/11732749/cilium-sysdump-7-final.zip)

Looking at the sysdump, this error message looks suspicious:

```
level=error msg="Failed to get identity labels for endpoint, skipping update to egress policy." error="identity 28814 not found" k8sEndpointName=echo-other-node-7bc75565b9-287c5 k8sNamespace=cilium-test subsys=egressgateway
```

jibi:
when the manager fails to resolve the identity of an endpoint to a set of labels, instead of discarding the endpoint update, enqueue the event and retry identity labels resolution following the same exponential backoff logic used by the identity allocator

### testing 
I tested this locally by forcing the identity resolution to randomly fail:

```diff
diff --git a/pkg/egressgateway/manager.go b/pkg/egressgateway/manager.go
index ed7f19399c..0b867b2d12 100644
--- a/pkg/egressgateway/manager.go
+++ b/pkg/egressgateway/manager.go
@@ -7,6 +7,7 @@ import (
        "container/heap"
        "context"
        "fmt"
+       "math/rand"
        "net"
        "sort"
        "time"
@@ -171,6 +172,10 @@ func NewEgressGatewayManager(p Params) *Manager {
 // getIdentityLabels waits for the global identities to be populated to the cache,
 // then looks up identity by ID from the cached identity allocator and return its labels.
 func (manager *Manager) getIdentityLabels(securityIdentity uint32) (labels.Labels, error) {
+       if rand.Intn(16) > 0 {
+               return nil, fmt.Errorf("fake error")
+       }
+
```

```sh
➜  cilium git:(pr/jibi/egressgw-retry-identity) ks logs cilium-2dc5v --timestamps | grep subsys=egress | grep coredns-5d78c9869d-44vlf
2023-06-23T10:22:36.698593375Z level=warning msg="Failed to get identity labels for endpoint, will retry." error="fake error" k8sEndpointName=coredns-5d78c9869d-44vlf k8sNamespace=kube-system subsys=egressgateway
2023-06-23T10:22:37.590584472Z level=warning msg="Failed to get identity labels for endpoint, will retry." error="fake error" k8sEndpointName=coredns-5d78c9869d-44vlf k8sNamespace=kube-system subsys=egressgateway
2023-06-23T10:22:42.630015783Z level=warning msg="Failed to get identity labels for endpoint, will retry." error="fake error" k8sEndpointName=coredns-5d78c9869d-44vlf k8sNamespace=kube-system subsys=egressgateway
2023-06-23T10:22:44.710650728Z level=warning msg="Failed to get identity labels for endpoint, will retry." error="fake error" k8sEndpointName=coredns-5d78c9869d-44vlf k8sNamespace=kube-system subsys=egressgateway
2023-06-23T10:22:46.871488814Z level=warning msg="Failed to get identity labels for endpoint, will retry." error="fake error" k8sEndpointName=coredns-5d78c9869d-44vlf k8sNamespace=kube-system subsys=egressgateway
2023-06-23T10:22:52.190670670Z level=warning msg="Failed to get identity labels for endpoint, will retry." error="fake error" k8sEndpointName=coredns-5d78c9869d-44vlf k8sNamespace=kube-system subsys=egressgateway
2023-06-23T10:22:55.113841789Z level=warning msg="Failed to get identity labels for endpoint, will retry." error="fake error" k8sEndpointName=coredns-5d78c9869d-44vlf k8sNamespace=kube-system subsys=egressgateway
2023-06-23T10:22:58.952487945Z level=debug msg="Added CiliumEndpoint" k8sEndpointName=coredns-5d78c9869d-44vlf k8sNamespace=kube-system subsys=egressgateway
```

Fixes: #26158 (hopefully :crossed_fingers:)

### DIFF
--- a/pkg/egressgateway/manager.go
+++ b/pkg/egressgateway/manager.go
@@ -345,13 +345,6 @@ func (manager *Manager) addEndpoint(id types.NamespacedName) {
 	// encounter a failure it cannot be retried
 	delete(manager.pendingEndpointEvents, id)
 
-	if len(endpoint.Networking.Addressing) == 0 {
-		logger.WithError(err).
-			Error("Failed to get valid endpoint IPs, skipping update to egress policy.")
-
-		return
-	}
-
 	if epData, err = getEndpointMetadata(endpoint, identityLabels); err != nil {
 		logger.WithError(err).
 			Error("Failed to get valid endpoint metadata, skipping update to egress policy.")
--- a/pkg/egressgateway/manager.go
+++ b/pkg/egressgateway/manager.go
@@ -8,11 +8,14 @@ import (
 	"fmt"
 	"net"
 	"sort"
+	"sync"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/pflag"
 	"github.com/vishvananda/netlink"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/workqueue"
 
 	"github.com/cilium/cilium/pkg/datapath/linux/probes"
 	"github.com/cilium/cilium/pkg/hive"
@@ -101,6 +104,17 @@ type Manager struct {
 	// epDataStore stores endpointId to endpoint metadata mapping
 	epDataStore map[endpointID]*endpointMetadata
 
+	// pendingEndpointEvents stores the k8s CiliumEndpoint add/update events
+	// which still need to be processed by the manager, either because we
+	// just received the event, or because the processing failed due to the
+	// manager being unable to resolve the endpoint identity to a set of
+	// labels
+	pendingEndpointEvents map[endpointID]*k8sTypes.CiliumEndpoint
+
+	// endpointEventsQueue is a workqueue of CiliumEndpoint IDs that need to
+	// be processed by the manager
+	endpointEventsQueue workqueue.RateLimitingInterface
+
 	// identityAllocator is used to fetch identity labels for endpoint updates
 	identityAllocator identityCache.IdentityAllocator
 
@@ -130,16 +144,27 @@ func NewEgressGatewayManager(p Params) *Manager {
 		return nil
 	}
 
+	// here we try to mimic the same exponential backoff retry logic used by
+	// the identity allocator, where the minimum retry timeout is set to 20
+	// milliseconds and the max number of attempts is 16 (so 20ms * 2^16 ==
+	// ~20 minutes)
+	rateLimiter := workqueue.NewItemExponentialFailureRateLimiter(time.Millisecond*20, time.Minute*20)
+	endpointEventRetryQueue := workqueue.NewRateLimitingQueueWithConfig(rateLimiter, workqueue.RateLimitingQueueConfig{})
+
 	manager := &Manager{
 		cacheStatus:             p.CacheStatus,
 		nodeDataStore:           make(map[string]nodeTypes.Node),
 		policyConfigs:           make(map[policyID]*PolicyConfig),
 		policyConfigsBySourceIP: make(map[string][]*PolicyConfig),
 		epDataStore:             make(map[endpointID]*endpointMetadata),
+		pendingEndpointEvents:   make(map[endpointID]*k8sTypes.CiliumEndpoint),
+		endpointEventsQueue:     endpointEventRetryQueue,
 		identityAllocator:       p.IdentityAllocator,
 		installRoutes:           p.Config.InstallEgressGatewayRoutes,
 		policyMap:               p.PolicyMap,
 	}
+
+	var wg sync.WaitGroup
 
 	ctx, cancel := context.WithCancel(context.Background())
 	p.Lifecycle.Append(hive.Hook{
@@ -149,10 +174,13 @@ func NewEgressGatewayManager(p Params) *Manager {
 			}
 
 			manager.runReconciliationAfterK8sSync(ctx)
+			manager.processCiliumEndpoints(ctx, &wg)
 			return nil
 		},
 		OnStop: func(hc hive.HookContext) error {
 			cancel()
+
+			wg.Wait()
 			return nil
 		},
 	})
@@ -186,6 +214,51 @@ func (manager *Manager) runReconciliationAfterK8sSync(ctx context.Context) {
 			manager.reconcile(eventK8sSyncDone)
 			manager.Unlock()
 		case <-ctx.Done():
+		}
+	}()
+}
+
+// processCiliumEndpoints spawns a goroutine that:
+//   - consumes the endpoint IDs returned by the endpointEventsQueue workqueue
+//   - processes the CiliumEndpoints stored in pendingEndpointEvents for these
+//     endpoint IDs
+//   - in case the endpoint ID -> labels resolution fails, it adds back the
+//     event to the workqueue so that it can be retried with an exponential
+//     backoff
+func (manager *Manager) processCiliumEndpoints(ctx context.Context, wg *sync.WaitGroup) {
+	wg.Add(1)
+
+	go func() {
+		defer wg.Done()
+
+		retryQueue := manager.endpointEventsQueue
+		go func() {
+			select {
+			case <-ctx.Done():
+				retryQueue.ShutDown()
+			}
+		}()
+
+		for {
+			item, shutdown := retryQueue.Get()
+			if shutdown {
+				break
+			}
+			endpointID := item.(types.NamespacedName)
+
+			manager.addEndpoint(endpointID)
+
+			if manager.endpointEventIsPending(endpointID) {
+				// if the endpoint event is still pending it means the manager
+				// failed to resolve the endpoint ID to a set of labels, so add back
+				// the item to the queue
+				manager.endpointEventsQueue.AddRateLimited(endpointID)
+			} else {
+				// otherwise just remove it
+				manager.endpointEventsQueue.Forget(endpointID)
+			}
+
+			manager.endpointEventsQueue.Done(endpointID)
 		}
 	}()
 }
@@ -233,8 +306,15 @@ func (manager *Manager) OnDeleteEgressPolicy(configID policyID) {
 	manager.reconcile(eventDeletePolicy)
 }
 
-// OnUpdateEndpoint is the event handler for endpoint additions and updates.
-func (manager *Manager) OnUpdateEndpoint(endpoint *k8sTypes.CiliumEndpoint) {
+func (manager *Manager) endpointEventIsPending(id types.NamespacedName) bool {
+	manager.Lock()
+	defer manager.Unlock()
+
+	_, ok := manager.pendingEndpointEvents[id]
+	return ok
+}
+
+func (manager *Manager) addEndpoint(id types.NamespacedName) {
 	var epData *endpointMetadata
 	var err error
 	var identityLabels labels.Labels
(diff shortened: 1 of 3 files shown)


ti-mo:
I got pinged for TH review on `pkg/backoff`, could you assign a CODEOWNER to that package in this PR? :pray:

bimmlerd (approved):
CODEOWNERS change lgtm

pkg/backoff/backoff.go
```diff
@@ -201,3 +206,8 @@ func (b *Exponential) Duration(attempt int) time.Duration {
 
 	return t
 }
+
+// CurrentDuration returns the wait duration for the current number of attempt
```
giorio94:
```suggestion
// CurrentDuration returns the wait duration for the current number of attempts
```

Nit.

pkg/egressgateway/manager.go
```diff
@@ -149,6 +156,7 @@ func NewEgressGatewayManager(p Params) *Manager {
 			}
 
 			manager.runReconciliationAfterK8sSync(ctx)
+			manager.retryEndpointIdentityLabelsResolution(ctx)
 			return nil
 		},
 		OnStop: func(hc hive.HookContext) error {
```
giorio94, line 180: Nit. According to the [hive guidelines](https://docs.cilium.io/en/latest/contributing/development/hive/#guidelines), the stop hook should block until goroutines have terminated.

pkg/egressgateway/manager.go
```diff
@@ -190,6 +198,35 @@ func (manager *Manager) runReconciliationAfterK8sSync(ctx context.Context) {
...
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				manager.Lock()
+				queueLen := manager.endpointEventRetryQueue.Len()
+				manager.Unlock()
```
giorio94: There's the possibility of a race condition, given that the lock is released here, but then the heap is accessed again below to pop one item.

pkg/egressgateway/manager.go
```diff
@@ -190,6 +198,35 @@ func (manager *Manager) runReconciliationAfterK8sSync(ctx context.Context) {
...
+
+				if queueLen == 0 {
+					continue
+				}
+
+				endpointEventRetry := heap.Pop(&manager.endpointEventRetryQueue).(*endpointEventRetry)
+				endpointEventRetry.backoff.Wait(ctx)
```
giorio94: There's the possibility that this wait blocks also other "higher priority" retries. Let's say for instance that one endpoint has already been retried a few times, and needs to block here for a long period (t1). A new endpoint which is queued immediately afterwards (with a smaller wait period) will have to wait for t1 + t2, rather than just t2. The same happens in case multiple endpoints are queued, the latter will have to wait for n*t, although its own waiting period should have been only t.

More generally, I'd suggest to use a working queue rather than a custom implementation, given that it already automatically handles all these aspects.

pkg/egressgateway/manager.go
```diff
@@ -190,6 +198,35 @@ func (manager *Manager) runReconciliationAfterK8sSync(ctx context.Context) {
...
+				manager.addEndpoint(endpointEventRetry.endpoint, endpointEventRetry.backoff)
```
giorio94: There's the possibility that an endpoint which is added here after the wait period has already been deleted in parallel.

giorio94 (changes requested):

pkg/egressgateway/manager.go
jibi: ack, let me take a look at working queues then :+1:

jibi:
Thanks for the pointers @giorio94 , I switched this PR to a workqueue.

The approach I'm taking is the follow: there's an `endpointEventRetryQueue` workqueue that simply tracks the ID of the CiliumEndpoint that should be processed/reprocessed, and then in a separate `pendingEndpointEvents` object we store the ID -> CiliumEndpoint mapping.

This ensures we can deal nicely with CiliumEndpoints that get updated/deleted while in the workqueue: whenever we receive a new event, any existing one for the same ID gets removed from the queue (as otherwise we could end up retrying processing old revisions of a given CiliumEndpoint).

tested with a random fake error:

```
2023-06-30T08:04:33.847665474Z level=debug msg="Added CiliumEgressGatewayPolicy" ciliumEgressGatewayPolicyName=egress subsys=egressgateway
2023-06-30T08:04:34.300917210Z level=warning msg="Failed to get identity labels for endpoint" error="fake error" k8sEndpointName=ubuntu-1 k8sNamespace=default subsys=egressgateway
2023-06-30T08:04:34.300955321Z level=warning msg="Failed to get identity labels for endpoint" error="fake error" k8sEndpointName=ubuntu-1 k8sNamespace=default subsys=egressgateway
2023-06-30T08:04:34.321451715Z level=warning msg="Failed to get identity labels for endpoint" error="fake error" k8sEndpointName=ubuntu-1 k8sNamespace=default subsys=egressgateway
2023-06-30T08:04:34.361744042Z level=warning msg="Failed to get identity labels for endpoint" error="fake error" k8sEndpointName=ubuntu-1 k8sNamespace=default subsys=egressgateway
2023-06-30T08:04:34.442374668Z level=warning msg="Failed to get identity labels for endpoint" error="fake error" k8sEndpointName=ubuntu-1 k8sNamespace=default subsys=egressgateway
2023-06-30T08:04:34.602854941Z level=warning msg="Failed to get identity labels for endpoint" error="fake error" k8sEndpointName=ubuntu-1 k8sNamespace=default subsys=egressgateway
2023-06-30T08:04:34.923616503Z level=warning msg="Failed to get identity labels for endpoint" error="fake error" k8sEndpointName=ubuntu-1 k8sNamespace=default subsys=egressgateway
2023-06-30T08:04:35.564425666Z level=warning msg="Failed to get identity labels for endpoint" error="fake error" k8sEndpointName=ubuntu-1 k8sNamespace=default subsys=egressgateway
2023-06-30T08:04:36.845466966Z level=warning msg="Failed to get identity labels for endpoint" error="fake error" k8sEndpointName=ubuntu-1 k8sNamespace=default subsys=egressgateway
2023-06-30T08:04:39.405585974Z level=debug msg="Added CiliumEndpoint" k8sEndpointName=ubuntu-1 k8sNamespace=default subsys=egressgateway
2023-06-30T08:04:39.405952893Z level=debug msg="Egress gateway policy applied" destinationCIDR=0.0.0.0/0 egressIP=0.0.0.0 gatewayIP=0.0.0.0 sourceIP=10.244.2.32 subsys=egressgateway
2023-06-30T08:04:39.405959833Z level=debug msg="Egress gateway policy applied" destinationCIDR=8.8.8.8/32 egressIP=0.0.0.0 gatewayIP=0.0.0.1 sourceIP=10.244.2.32 subsys=egressgateway
```

pkg/egressgateway/manager.go
```diff
@@ -265,23 +355,63 @@ func (manager *Manager) OnUpdateEndpoint(endpoint *k8sTypes.CiliumEndpoint) {
...
+func (manager *Manager) OnUpdateEndpoint(endpoint *k8sTypes.CiliumEndpoint) {
+	id := types.NamespacedName{
+		Name:      endpoint.GetName(),
+		Namespace: endpoint.GetNamespace(),
+	}
+
+	// remove the ID from the workqueue in case an older revision of the
+	// same CiliumEndpoint was already present
+	manager.endpointEventRetryQueue.Forget(id)
+	manager.endpointEventRetryQueue.Done(id)
```
giorio94: It is a bit strange seeing these operations being performed outside of the reconcile loop, but I cannot spot anything that could go wrong with them. Still, I wonder if it could make sense to just add the entry to the pending events and push the ID in the queue, leaving the reconcile loop to trigger `addEndpoint`.

pkg/egressgateway/manager.go
```diff
@@ -265,23 +355,63 @@ func (manager *Manager) OnUpdateEndpoint(endpoint *k8sTypes.CiliumEndpoint) {
...
 	id := types.NamespacedName{
 		Name:      endpoint.GetName(),
 		Namespace: endpoint.GetNamespace(),
 	}
 
 	delete(manager.epDataStore, id)
 
+	delete(manager.pendingEndpointEvents, id)
+	manager.endpointEventRetryQueue.Forget(id)
+	manager.endpointEventRetryQueue.Done(id)
```
giorio94: Same here about just letting the event to be consumed by the reconcile loop rather than removing it explicitly.

giorio94 (approved):
The workqueue implementation looks good to me.
I wonder whether it could make sense to leverage it also during insertion for increased uniformity. WDYT?

pkg/egressgateway/manager.go
jibi: I think this should stay here as it has nothing to do with the reconciliation logic. More in general, the way the manager is structured is the following:
* we have a bunch of event handlers called by the k8s watchers (for policies, endpoints and nodes events) which simply parse the events and update the copies of these resources stored in the manager
* and then we have the reconciliation logic which goes through these objects and ensure the state of the node is in sync with these resources

because of this I think parsing an endpoint event (and retrying because we failed to resolve identity -> labels) should not be part of the reconciliation logic

giorio94: Yeah, I agree, and I think we might be referring to different things when mentioning "reconciliation logic". In particular, I was meaning the `retryEndpointIdentityLabelsResolution` loop, and not `manager.reconcile()` (I agree that the latter is independent and should not handle the endpoint parsing for symmetry with the other handlers). 

Still, my comment was mostly a nit, and I'm also fine with the current version.

jibi: Ah, I see what you mean now :+1: 
I changed the `On{Update,Delete}Endpoint` handlers to only add and remove items from the workqueue/pending events map, and let the retry loop do the rest (so call `addEnpoint` and deal with re-adding back items or marking them as done)

pkg/egressgateway/manager.go
```diff
@@ -190,6 +218,47 @@ func (manager *Manager) runReconciliationAfterK8sSync(ctx context.Context) {
 	}()
 }
 
+// retryEndpointIdentityLabelsResolution spawns a goroutine that retries to
+// add/update an endpoint for which we failed to resolve the identity to a set
+// of labels
+func (manager *Manager) retryEndpointIdentityLabelsResolution(ctx context.Context, wg *sync.WaitGroup) {
```
giorio94: Nit. The name of this function (and the comment) could be generalized now that handles all upsert events.

giorio94 (approved):
Looks great to me!

jibi:
that last change broke the unit tests :see_no_evil: 

```
--- FAIL: Test (0.02s)
    --- FAIL: Test/EgressGatewayTestSuite (0.02s)
        --- FAIL: Test/EgressGatewayTestSuite/TestEgressGatewayManager (0.02s)
            manager_privileged_test.go:630: 
                ... value *fmt.wrapError = &fmt.wrapError{msg:"lookup: key does not exist", err:(*sys.syscallError)(0x3b53c20)} ("lookup: key does not exist")
```

looking into

PR #26458, opened 2023-06-23
Fix content headers in Health Check Node Port response

cezarygerard:
Fix content headers in Health Check Node Port response. 

Before the fix 2 content headers are not returned, because `w.WriteHeader(http.StatusOK)` must be called after setting the headers map

```release-note
Fix: Return "Content-Type" and "X-Content-Type-Options" headers from  Health Check Node Port
```

### DIFF
--- a/pkg/service/healthserver/healthserver.go
+++ b/pkg/service/healthserver/healthserver.go
@@ -213,13 +213,14 @@ func (h *httpHealthServer) shutdown() {
 func (h *httpHealthServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Use headers and JSON output compatible with kube-proxy
 	svc := h.loadService()
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("X-Content-Type-Options", "nosniff")
+
 	if svc.LocalEndpoints == 0 {
 		w.WriteHeader(http.StatusServiceUnavailable)
 	} else {
 		w.WriteHeader(http.StatusOK)
 	}
-	w.Header().Set("Content-Type", "application/json")
-	w.Header().Set("X-Content-Type-Options", "nosniff")
 	if err := json.NewEncoder(w).Encode(&svc); err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 	}
--- a/pkg/service/healthserver/healthserver_test.go
+++ b/pkg/service/healthserver/healthserver_test.go
@@ -8,8 +8,8 @@ import (
 	"net/http/httptest"
 	"testing"
 
-	"github.com/google/go-cmp/cmp"
 	. "github.com/cilium/checkmate"
+	"github.com/google/go-cmp/cmp"
 )
 
 type ServiceHealthServerSuite struct{}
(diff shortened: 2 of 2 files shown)


dlapcevic (approved):
/lgtm

borkmann (approved):

pkg/service/healthserver/healthserver_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
```
aojea: you need to run gofmt on this file

```
ake: *** [Makefile:645: precheck] Error 1
Unformatted Go source code:
make: *** Waiting for unfinished jobs....
./pkg/service/healthserver/healthserver_test.go
diff ./pkg/service/healthserver/healthserver_test.go.orig ./pkg/service/healthserver/healthserver_test.go
--- ./pkg/service/healthserver/healthserver_test.go.orig
+++ ./pkg/service/healthserver/healthserver_test.go
@@ -8,8 +8,8 @@
 	"net/http/httptest"
 	"testing"
 
-	"github.com/google/go-cmp/cmp"
 	. "github.com/cilium/checkmate"
+	"github.com/google/go-cmp/cmp"
 )
 
 type ServiceHealthServerSuite struct{}
@@ -94,7 +94,7 @@
 	assertRespHeader(c, resp, "Content-Type", "application/json")
 	assertRespHeader(c, resp, "X-Content-Type-Options", "nosniff")
 	resp.Body.Close()
-	
+
```

pkg/service/healthserver/healthserver.go
```diff
@@ -213,13 +213,14 @@ func (h *httpHealthServer) shutdown() {
...
 	svc := h.loadService()
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("X-Content-Type-Options", "nosniff")
+
 	if svc.LocalEndpoints == 0 {
 		w.WriteHeader(http.StatusServiceUnavailable)
 	} else {
 		w.WriteHeader(http.StatusOK)
 	}
-	w.Header().Set("Content-Type", "application/json")
```
aojea, line 221: These are trailers https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer

https://pkg.go.dev/net/http#example-ResponseWriter-Trailers

I do not know if this is on purpose though

pkg/service/healthserver/healthserver_test.go
cezarygerard: done

pkg/service/healthserver/healthserver.go
aojea: does not look that using trailers instead of headers are on purpose, and using headers is coherent with kube-proxy https://github.com/kubernetes/kubernetes/blob/027ac5a426a261ba6b66a40e79e123e75e9baf5b/pkg/proxy/healthcheck/proxier_health.go#L159-L173 , since this is a "replacement" it makes sense to be consistent.
Also, the failing test indicates that the goal may have been to use headers in the first place

cezarygerard: similar code in kube-proxy uses regular headers, not trailers

https://github.com/kubernetes/kubernetes/blob/027ac5a426a261ba6b66a40e79e123e75e9baf5b/pkg/proxy/healthcheck/service_health.go#L236

brb (approved):

ysksuzuki (approved):

gandro (approved):
Looks good to me. Did this actually cause issues or how did you find this?

ldelossa:
All required tests passing, and multiple committers approved. Merging :rocket:

aojea:
> Looks good to me. Did this actually cause issues or how did you find this?

google loadbalancers can use the headers for passing the data of the existing endpoints on the node and this way use weights, the headers were not coming at all 😄

PR #26459, opened 2023-06-23
bpf: unconditionally migrate tail call maps

### LINKED ISSUE #20691: Always migrate tail call maps

ti-mo:
This patch makes it so tail call maps (program arrays) are never updated by user space after they've been created + populated.

A new version of an ELF may have some behaviour moved from one tail call to another, potentially leading to unexpected packet drops or a security bypass during the re-population phase of the array.

It is generally not safe from a logical perspective to manipulate a program array that is attached to a live endpoint, since packet processing might take place while the array is being updated.

Fixes: https://github.com/cilium/cilium/issues/20691

```release-note
Unconditionally migrate tail call maps when Endpoints are updated
```


ti-mo:
Superseded by https://github.com/cilium/cilium/pull/28740.

Issue #26430, opened 2023-06-22
update outdated installation docs


michi-covalent:
let's use this issue to track outdated documentation pages so that we can update them for v1.14 release ✨ 

- [x] cluster mesh: https://docs.cilium.io/en/latest/network/clustermesh/clustermesh/#gs-clustermesh let's document both helm-based and cilium-cli based instructions.
- [x] external workloads: https://docs.cilium.io/en/latest/network/external-workloads/#external-workloads document newly added helm values: https://github.com/cilium/cilium/pull/26607
- [x] cilium-cli: this is an open question, but i'd like to release cilium-cli v0.15 and update cilium v1.14 docs to consistently use Helm mode. otherwise we need to support the classic mode for another 6 monthds. cc @tklauser @asauber what do you think: https://github.com/cilium/cilium/pull/26606
- [x] openshift: https://docs.cilium.io/en/latest/installation/k8s-install-openshift-okd/#k8s-install-openshift-okd refers to https://github.com/cilium/cilium-olm, which has been deprecated: will clarify this in https://github.com/cilium/cilium/issues/26536

anything else?

### additional pages to check
- [x] Documentation/gettingstarted/hubble.rst
- [x] Documentation/gettingstarted/k8s-install-default.rst
- [x] Documentation/helm-values.rst
- [x] Documentation/installation/k3s.rst
- [x] Documentation/installation/k8s-install-migration.rst
- [x] Documentation/network/clustermesh/aks-clustermesh-prep.rst
- [x] Documentation/network/clustermesh/clustermesh.rst
- [x] Documentation/network/external-workloads.rst
- [x] Documentation/network/servicemesh/gateway-api/installation.rst
- [x] Documentation/network/servicemesh/installation.rst
- [x] Documentation/network/servicemesh/tls-default-certificate.rst
- [x] Documentation/operations/performance/tuning.rst
- [x] Documentation/security/network/encryption-ipsec.rst
- [x] Documentation/security/network/encryption-wireguard.rst

asauber:
I'm 100% onboard with Helm mode being the default documented CLI mode for 1.14.

For clarification, we would need to support classic mode for an additional 6 months otherwise, because the Helm-mode changes would not be an appropriate intra-minor-version change for 1.14?

asauber:
[...]

PR #26461, opened 2023-06-23
docs: Fix the cilium-cli default branch name

michi-covalent:
The default branch name for cilium/cilium-cli repo changed from master to main. The current instructions still work because the master branch gets synched with the main branch, but let's update the docs to use the new default branch name to avoid any confusion in the future.

Ref: #26430
Ref: https://github.com/cilium/cilium-cli/blob/main/.github/workflows/mirror.yaml


learnitall (approved):
Nice catch!

michi-covalent:
this pull request only modifies a file under `Documentation/` directory. merging 🚀

PR #26470, opened 2023-06-24
CiliumNetworkPolicy validation

### LINKED ISSUE #23152: CFP: CiliumNetworkPolicy validation
## Cilium Feature Proposal
        
Malformed/invalid CNPs (e.g., L7 policies without the `protocol: TCP` field set) get silently discarded by the agent, emitting only the corresponding log entry about the error encountered during validation (e.g., `L7 rules can only apply to TCP (not ANY) except for DNS rules`). For non-experts, this makes it tricky to understand why the CNP is not being enforced as expected, as no errors are reported while creating the resource, nor they are attached to it.

In my opinion, a great QoL addition would be the introduction of a validation webhook to prevent the creation of invalid network policies, making it immediately clear to the user that something is wrong with that. 

Alternatively, the feedback might be returned as part of the status stanza of the resource, or emitted as a Kubernetes event, which are typically looked for before the logs.

romanspb80:
This patch adds validation for CiliumNetworkPolicy with CEL

Fixes: #23152


gandro:
Thanks for the PR!

Could you rebase on `main`? It seems your branch is lagging quite a bit behind and thus causing CI Failures because of tests that have changed since.

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if and .Values.agent (not .Values.preflight.enabled) }}
```
gandro: I think we need to feature gate this via a Helm value. Otherwise this will be installed by default, which I don't think is what we want for an alpha K8s feature.

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicybinding.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if and .Values.agent (not .Values.preflight.enabled) }}
+---
+apiVersion: admissionregistration.k8s.io/v1alpha1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: "http-tcp-binding"
+spec:
+  policyName: "http-tcp"
+  matchResources:
```
gandro, line 8: I have not yet looked at the validation feature in detail. Will this empty match statement all resources?

contrib/vagrant/scripts/07-install-admissionpolicy.sh
```diff
@@ -0,0 +1,22 @@
...
+
+source "${dir}/helpers.bash"
+
+log "Installing L7 admission policy into kubectl cluster..."
+
+set -e
+
+deployments_dir="${dir}/../deployments"
+vap_file="${deployments_dir}/validatingadmissionpolicy.yaml"
+vapb_file="${deployments_dir}/validatingadmissionpolicybinding.yaml"
```
gandro: I'm a bit worried about the code duplication here. The two files (the one in `contrib/vagrant` and the one in `install/kubernetes`) will inevitability diverge, unless we have a script that ensures the two are kept in sync.

But I'm wondering, why is this needed? Can we not rely on `helm install` admission policy together with Cilium? Particularly, what will happen with `helm install` if the policy is already present here?

In general, the vagrant scripts are under-maintained and we're not relying on them for much anymore. Therefore I'd be curious for the motivation of adding the validation policy here, especially since the same can be apparently done via `helm install`.

gandro (changes requested):
Thanks for the PR. I think the feature itself would be a nice addition, but I think we first need to clarify the life cycle of the resource.

contrib/vagrant/deployments/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,22 @@
...
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups:   ["cilium.io"]
+        apiVersions: ["v2"]
+        operations:  ["CREATE", "UPDATE"]
+        resources:   ["ciliumnetworkpolicies"]
+  validations:
+    - expression:
+        (has( object.spec.ingress[0].toPorts[0].rules.http) &&
```
gandro: This is only testing the first element of the ingress or egress section. Shouldn't we test all elements?

contrib/vagrant/scripts/07-install-admissionpolicy.sh
romanspb80: > I'm a bit worried about the code duplication here. The two files (the one in `contrib/vagrant` and the one in `install/kubernetes`) will inevitability diverge, unless we have a script that ensures the two are kept in sync.
> 
> But I'm wondering, why is this needed? Can we not rely on `helm install` admission policy together with Cilium? Particularly, what will happen with `helm install` if the policy is already present here?
> 
> In general, the vagrant scripts are under-maintained and we're not relying on them for much anymore. Therefore I'd be curious for the motivation of adding the validation policy here, especially since the same can be apparently done via `helm install`.

Thank you for the code review.
Yes. This is an unnecessary action. I deleted this part of the commit.
Initially, I added these actions to Vagrant because helm didn't start under Vagrant

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicybinding.yaml
romanspb80: > I have not yet looked at the validation feature in detail. Will this empty match statement all resources?

Yep. I used an empty LabelSelector for "matchResources", which will match all objects

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
romanspb80: This is a good idea. I'll add a same validation

romanspb80:
> Thanks for the PR!
> 
> Could you rebase on `main`? It seems your branch is lagging quite a bit behind and thus causing CI Failures because of tests that have changed since you last synced your branch.

Done. Thanks

bimmlerd:
As far as I can tell, the latest version of this PR touches no files owned by contributing - removing review request.

contrib/vagrant/deployments/validatingadmissionpolicy.yaml
romanspb80: I have redone these checks

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,26 @@
+{{- if .Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1alpha1/ValidatingAdmissionPolicy" }}
```
gandro: This is much better, but I think we should also make this dependent on a Helm flag. Many users use `helm template`, in which case Capabilities won't work. And even if they have the capability, they might still want to turn off the validation (it's an alpha feature after all)

I would propose to add a Helm value, e.g. `validaingAdmissionPolicies.enabled` and just have the admission policy created based on that flag.

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,26 @@
...
+---
+apiVersion: admissionregistration.k8s.io/v1alpha1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: "http-tcp"
```
gandro: Let's use name here that indicates this comes from Cilium, e.g. `cilium-validate-network-policies` (I'm assuming we want to use additional validations later too)

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,26 @@
...
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups:   ["cilium.io"]
+        apiVersions: ["v2"]
+        operations:  ["CREATE", "UPDATE"]
+        resources:   ["ciliumnetworkpolicies"]
```
gandro: Should we also check for CiliumClusterWideNetworkPolicies here? I think the schema should match too (I have not verified that though).

gandro (commented):
Thanks, looking better!

Overall, this looks fine to me (I still have some feedback on the Helm side). 

There is the more fundamental question if Validating Admission Policies are really the right tool for the job here. It seems to me, their use case is mostly to enforce user-specific constraints. But I'm not familiar enough with CRD validation to understand if there is a better way to reject invalid CNPs earlier.

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,26 @@
...
+  validations:
+    - expression:
+        #ingress validating
```
gandro: I wonder if we could split ingress and egress into two separate expressions here?

romanspb80:
> 

In my opinion, [Validating Admission Policy](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) is the most promising because it uses the declarative principle, as well as the power of the [CEL language](https://github.com/google/cel-spec/blob/master/doc/langdef.md) and does not require intervention in already written code of controller/operator. This approach will allow us to develop new validation rules quickly and efficiently.

gandro:
> > 
> 
> In my opinion, [Validating Admission Policy](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) is the most promising because it uses the declarative principle, as well as the power of the [CEL language](https://github.com/google/cel-spec/blob/master/doc/langdef.md) and does not require intervention in already written code of controller/operator. This approach will allow us to develop new validation rules quickly and efficiently.

Sounds good to me (from sig-helm)! I'll defer to the sig-k8s code owners on that question.

squeed:
Should these validation rules go in the go type definition as [kubebuilder marks](https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/)? That would be my personal preference, so that all validation was in one place.

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,26 @@
...
+        (has(object.spec.ingress) && object.spec.ingress.exists(x, has(x.toPorts) && x.toPorts.exists(x, (has(x.ports) &&
```
chancez: Could we organize this with newlines/indentation to make it easier to read? Nested parens can make this tricky.

Here's an example. This is a just an example, it might make more sense to combine some lines.
```
(
  has(object.spec.ingress) 
  && object.spec.ingress.exists(x, has(x.toPorts) 
    && x.toPorts.exists(x, 
      (
        has(x.ports) && (
          x.ports.exists(x, x.protocol == 'TCP') 
          && (
            has(x.rules) && (
              has(x.rules.http) || has(x.rules.kafka)
            )
          ) 
        ) && (
          has(x.rules) && !has(x.rules.dns)
        )
      )
    )
  )
)
```

christarazi:
Given that reviewers are giving different directions for this PR, I think we may want to step back and think about how we could do CRD validation "2.0". In the community we do this via [CFPs](https://docs.cilium.io/en/stable/contributing/development/contributing_guide/#cilium-feature-proposals). I see that you @romanspb80 have already do a part of this which is to create a CFP issue. The next step would be to follow:

> For longer proposals, you might like to include a link to an external doc (e.g. a Google doc) where it’s easier for reviewers to make comments and suggestions in-line. The GitHub feature request template includes a link to the [Cilium Feature Proposal template](https://docs.google.com/document/d/1vtE82JExQHw8_-pX2Uhq5acN1BMPxNlS6cMQUezRTWg/edit) which you are welcome to use to help structure your proposal. Please make a copy of that template, fill it in with your ideas, and ensure it’s publicly visible, before adding the link into the GitHub issue.

The document should generally propose a few approaches (at the very least they are, (1) what's being implemented in this PR and (2) the suggestion from @squeed ([ref](https://github.com/cilium/cilium/pull/26470#issuecomment-1621949044))). With this document, we can use it as the basis for discussion so that we can ultimately decide what makes the most sense for the Cilium project. 

Does this makes sense @romanspb80? If you have any questions, feel free to reach out on #development channel in Cilium Slack. There's also the [Cilium community meeting](https://docs.google.com/document/d/1Y_4chDk4rznD6UgXPlPvn3Dc7l-ZutGajUv1eF0VDwQ/edit#) for synchronous discussion. You can put down an agenda item in the open session and we will discuss it from there.

squeed:
After chatting with @chancez, he convinced me that a separate validation object would be better; it allows us to incrementally enable this *and* means less chance of hitting the maximum CRD size limit :-).

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
christarazi, line 1: I would love to see validation that replaces our forked logic to support `oneOf` for validation of fields that are exclusive. ~~That would allow us to move off of the fork. For as long as the fork is alive, we pay an increasing cost each time we want to update it. A few community members have asked about the fork recently.~~

Edit: I realized that this change is for when an individual CNP is applied rather than the CRD validation, so this doesn't affect the status of the fork.

We have a few usages in CNPs today. Grepping for `oneOf` in the Cilium repo will show you all the results like LocalRedirectPolicy as well.

1. `EndpointSelector` and a `NodeSelector`
2. A CIDRRule can either have a CIDR or a CIDRGroupRef

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
romanspb80: > I wonder if we could split ingress and egress into two separate expressions here?

This is a good idea. But the expression needs for both items (ingress and egress) immediately

Documentation/spelling_wordlist.txt
```diff
@@ -1137,6 +1137,7 @@ vCPU
 vCPUs
 vX
 validateCNPs
+validatingAdmissionPolicy
```
qmonnet: Hi, and thanks!
You will need to rebase your PR on the main branch, there are conflicts to address.

You can now drop this change when doing so.

romanspb80:
dublicated with https://github.com/cilium/cilium/pull/26794

PR #26474, opened 2023-06-25
Computed and propagated the value of OldEndpoints field when merging …

### LINKED ISSUE #26176: Service cache: populate OldEndpoints when merging information from remote clusters
https://github.com/cilium/cilium/pull/25687 added an `OldEndpoints` field to the `ServiceEvent`, which is currently set to the same values as `Endpoints` when merging information from remote clusters. This followup issue proposes to compute and propagate the appropriate `OldEndpoints` value also in those cases, for consistency if it was needed to be relied upon. See https://github.com/cilium/cilium/pull/25687#discussion_r1223053368 for additional details.

akstron:
…information from remote clusters.

      GitHub issue.

- Updated `mergeServiceUpdateLocked` and `mergeExternalServiceDeleteLocked` to calculate and propagate oldEndpoints.

Fixes: #26176 

```release-note
Computed and propagated the value of OldEndpoints field when merging remote cluster information.
```
Signed-off-by: Alok Kumar Singh <alokaks601@gmail.com>

### DIFF
--- a/pkg/k8s/service_cache.go
+++ b/pkg/k8s/service_cache.go
@@ -550,6 +550,8 @@ func (s *ServiceCache) mergeServiceUpdateLocked(service *serviceStore.ClusterSer
 		s.externalEndpoints[id] = externalEndpoints
 	}
 
+	oldEPs, _ := s.correlateEndpoints(id)
+
 	// The cluster the service belongs to will match the current one when dealing with external
 	// workloads (and in that case all endpoints shall be always present), and not match in the
 	// cluster-mesh case (where remote endpoints shall be used only if it is shared).
@@ -579,7 +581,7 @@ func (s *ServiceCache) mergeServiceUpdateLocked(service *serviceStore.ClusterSer
 			Service:      svc,
 			OldService:   oldService,
 			Endpoints:    endpoints,
-			OldEndpoints: endpoints,
+			OldEndpoints: oldEPs,
 			SWG:          swg,
 		}
 	}
@@ -619,6 +621,8 @@ func (s *ServiceCache) mergeExternalServiceDeleteLocked(service *serviceStore.Cl
 	if ok {
 		scopedLog.Debug("Deleting external endpoints")
 
+		oldEPs, _ := s.correlateEndpoints(id)
+
 		delete(externalEndpoints.endpoints, service.Cluster)
 		if len(externalEndpoints.endpoints) == 0 {
 			delete(s.externalEndpoints, id)
@@ -636,7 +640,7 @@ func (s *ServiceCache) mergeExternalServiceDeleteLocked(service *serviceStore.Cl
 				ID:           id,
 				Service:      svc,
 				Endpoints:    endpoints,
-				OldEndpoints: endpoints,
+				OldEndpoints: oldEPs,
 				SWG:          swg,
 			}
 


squeed:
Hmm, interesting - it seems this is a behavior change. Is this fixing a bug? Can you explain more in your commit about why this change is needed? Thanks!

squeed:
I'm not as familiar with the inner workings of Clustermesh, asked @giorio94 to take a look too.

pkg/k8s/service_cache.go
```diff
@@ -550,10 +550,13 @@ func (s *ServiceCache) mergeServiceUpdateLocked(service *serviceStore.ClusterSer
 		s.externalEndpoints[id] = externalEndpoints
 	}
 
+	var oldEPs *Endpoints
```
giorio94: I'd suggest to use `s.correlateEndpoints(id)` to compute the old endpoints. This ensures that we actually propagate all old endpoints, and not only the ones associated with the given remote cluster, so that the difference between new and old endpoints corresponds to the ones that have been removed. The same commend applies to the other occurrence.

giorio94 (changes requested):
Thanks @akstron for your contribution!

In addition to the inline comment, could you also please extend the commit message with the rationale behind this change?

giorio94:
> Hmm, interesting - it seems this is a behavior change. Is this fixing a bug?

The propagation of the `OldEndpoints` field has been recently added to fix a bug in `toServices` policies (https://github.com/cilium/cilium/pull/25687). It is currently not used when merging global service information (and indeed it was set to the same value as the `Endpoints` one), but IMO it makes sense to get it fixed so that we can rely on it if needed.

pkg/k8s/service_cache.go
akstron: Got it. Shouldn't it be done in case of local cluster as well? As I can see this is not done in case of local cluster update and delete of earlier pr.

giorio94: Yes, I would uniform that as well. It would also be cool extending the existing tests to assert the correctness of `OldEndpoints`.

akstron:
Made the recommended changes.

giorio94:
> Made the recommended changes.

Modifications look good to me. Could you please extend the commit message to additionally state why this change is performed?

akstron:
> > Made the recommended changes.
> 
> Modifications look good to me. Could you please extend the commit message to additionally state why this change is performed?

Should I mention the PR link as well?

giorio94:
> > > Made the recommended changes.
> > 
> > 
> > Modifications look good to me. Could you please extend the commit message to additionally state why this change is performed?
> 
> Should I mention the PR link as well?

No, that is not necessary, as it gets automatically linked. 

While there, you should also shorten the commit subject, as the *checkpatch* check is currently failing with:
```
ERROR:CUSTOM: Please avoid long commit subjects (max: 75, found: 97)
```

akstron:
Hi @giorio94, let me know if I need to add anything else in the commit message. Thanks.

giorio94:
> Hi @giorio94, let me know if I need to add anything else in the commit message. Thanks.

Looks great to me. Thanks!

giorio94 (approved):

giorio94:
One of the Conformance E2E runs hit https://github.com/cilium/cilium/issues/26591. Rerunning

squeed (approved):

Issue #24617, opened 2023-03-29
Check scaling sizes for auth map design


youngnick:
Spun out from cilium/design-cfps#5:

Check that the sizes for the fields in the auth map will scale correctly for large numbers of identities.

youngnick:
From @joestringer:
> If we wanted to go whole-hog here then some diagrams could help - cases like:
>One local application connects to a single deployment that has two backends on two remote nodes. This results in four connections between two security identities:
Node A -> Node B for identity x->y (At Node A)
Node A -> Node B for identity x->y (At Node B)
Node A -> Node C for identity x->y (At Node A)
Node A -> Node C for identity x->y (At Node C)
Similarly, for one application that makes calls to two different remote Pods on other nodes with different identities, there would be four connections (even the same peer node) :
Node A -> Node B for identity x->y (At Node A)
Node A -> Node B for identity x->y (At Node B)
Node A -> Node B for identity x->z (At Node A)
Node A -> Node B for identity x->z (At Node B)

>We could probably explore even more cases, but this is probably enough info for now to consider the 'back of the hand' math on the impact.

So, a TODO is to consider how the number of rows will scale with number of identities and number of pods in each identity.

youngnick:
Another item from @joestringer:
> (implementation details: this is probably somehow padded/aligned on some word boundary so common case for 3x__u32 is probably equivalent to 4x__u32, ie ~16B, so about 1MB as the last sentence implies)

> EDIT: Ah plus value, so maybe another half MB for that assuming <=8B of data.

So we need also to consider how big each row will be, and ensure we keep a reasonable table size.

youngnick:
The last item will be to update the CFP in cilium/design-cfps with the details. (under the "Implementation Details" section inside the "Chosen alternative" section.

youngnick:
We've set this to ~512k, the same size as the conntrack table, and because of how CNPs are namespaced, we expect the limiting factors to be the number of nodes in the cluster and the number of identities in use. This should be big enough for all but the largest installations, but we'll have to wait and see.

tl;dr we don't think there will be scaling issues for some time, if ever.

Remaining work here is to check if there is a metric available for the auth map pressure, since it's configurable once you know that.

youngnick:
@sayboras was working on this in #26478, he will pick it up when he is back from leave.

PR #26478, opened 2023-06-26
metrics: Add map pressure metric for auth map

sayboras:
### Description

This commit is to make sure that we are having map pressure metric for auth map to improve observability, which can be useful for adjusting map size if required.

Relates: https://github.com/cilium/cilium/issues/24617

### Testing

Testing was done locally with some auth policies, and then check the metrics as per below

```
$ ksysex ds/cilium -- cilium metrics list | grep pressure
Defaulted container "cilium-agent" out of: cilium-agent, config (init), mount-cgroup (init), apply-sysctl-overwrites (init), wait-for-node-init (init), clean-cilium-state (init), install-cni-binaries (init)
cilium_bpf_map_pressure                               map_name="cilium_auth_map"                                                                                                                                                                                                    0.000004
cilium_bpf_map_pressure                               map_name="ipcache"                                                                                                                                                                                                            0.000021
cilium_bpf_map_pressure                               map_name="lb4_backends_v3"                                                                                                                                                                                                    0.000046
cilium_bpf_map_pressure                               map_name="lb4_reverse_nat"                                                                                                                                                                                                    0.000397
cilium_bpf_map_pressure                               map_name="lb4_services_v2"                                                                                                                                                                                                    0.000397
cilium_bpf_map_pressure                               map_name="lxc"                                                                                                                                                                                                                0.000107
```

### DIFF
--- a/pkg/auth/authmap_cache.go
+++ b/pkg/auth/authmap_cache.go
@@ -12,6 +12,8 @@ import (
 	"golang.org/x/exp/maps"
 
 	"github.com/cilium/cilium/pkg/lock"
+	"github.com/cilium/cilium/pkg/maps/authmap"
+	"github.com/cilium/cilium/pkg/metrics"
 )
 
 type authMapCache struct {
@@ -19,13 +21,15 @@ type authMapCache struct {
 	authmap           authMap
 	cacheEntries      map[authKey]authInfo
 	cacheEntriesMutex lock.RWMutex
+	pressureGauge     *metrics.GaugeWithThreshold
 }
 
-func newAuthMapCache(logger logrus.FieldLogger, authmap authMap) *authMapCache {
+func newAuthMapCache(logger logrus.FieldLogger, authMap authMap) *authMapCache {
 	return &authMapCache{
-		logger:       logger,
-		authmap:      authmap,
-		cacheEntries: map[authKey]authInfo{},
+		logger:        logger,
+		authmap:       authMap,
+		cacheEntries:  map[authKey]authInfo{},
+		pressureGauge: metrics.NewBPFMapPressureGauge(authmap.MapName, 0),
 	}
 }
 
@@ -60,7 +64,7 @@ func (r *authMapCache) Update(key authKey, info authInfo) error {
 	}
 
 	r.cacheEntries[key] = info
-
+	r.updatePressureMetric()
 	return nil
 }
 
@@ -79,8 +83,12 @@ func (r *authMapCache) Delete(key authKey) error {
 	}
 
 	delete(r.cacheEntries, key)
-
+	r.updatePressureMetric()
 	return nil
+}
+
+func (r *authMapCache) MaxEntries() uint32 {
+	return r.authmap.MaxEntries()
 }
 
 func (r *authMapCache) DeleteIf(predicate func(key authKey, info authInfo) bool) error {
@@ -102,7 +110,7 @@ func (r *authMapCache) DeleteIf(predicate func(key authKey, info authInfo) bool)
 			delete(r.cacheEntries, k)
 		}
 	}
-
+	r.updatePressureMetric()
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/auth/authmap_cache.go
```diff
@@ -12,20 +12,24 @@ import (
...
+	"github.com/cilium/cilium/pkg/maps/authmap"
+	"github.com/cilium/cilium/pkg/metrics"
 )
 
 type authMapCache struct {
 	logger            logrus.FieldLogger
 	authmap           authMap
 	cacheEntries      map[authKey]authInfo
 	cacheEntriesMutex lock.RWMutex
+	pressureGauge     *metrics.GaugeWithThreshold
```
mhofstetter, line 24: @sayboras is there a specific reason why the pressure gauge needs to live in the cache? why not in the authmap itself? only because we have the number of cached entries for pressure calculation? IMO wrong responsibility :confused:

sayboras: ah I can't find a good way to get the current estimated number of entries in authmap without iterating all entries (which is a costly operation). So I just take a look at other maps, seems like we are emitting metrics based on cache only.

Upon agent restart, we will restore auth maps from file to cache, so I think we still have a fairly accurate estimation.

mhofstetter: thought so - and i kind of understand. it's just that IMO this should be the responsibility of the auth map itself. even if that means that we somehow need to keep track of the entries via a different approach - e.g. keeping a counter in the auth map :thinking: 

it's just that one big purpose of these separate components is to separate their concerns.

pkg/auth/authmap_cache.go
```diff
@@ -12,20 +12,24 @@ import (
...
-func newAuthMapCache(logger logrus.FieldLogger, authmap authMap) *authMapCache {
+func newAuthMapCache(logger logrus.FieldLogger, authMap authMap) *authMapCache {
 	return &authMapCache{
-		logger:       logger,
-		authmap:      authmap,
-		cacheEntries: map[authKey]authInfo{},
+		logger:        logger,
+		authmap:       authMap,
+		cacheEntries:  map[authKey]authInfo{},
+		pressureGauge: metrics.NewBPFMapPressureGauge(authmap.MapName, 0),
```
mhofstetter, line 32: is it somehow possible to use the new metrics approach from hive introduced by @dylandreimerink  ? -> https://docs.cilium.io/en/latest/contributing/development/hive/#metric

sayboras: It will be awesome to use the new pattern, just a note that this is the existing metric, which is still not migrated to the new hive and cell structure (i.e. adding the same metric will cause a double registration error in prometheus colllector).

pkg/auth/authmap_cache.go
jschwinger233: Agree with Macro's responsibility concerns, just want to add a point that `authMapCache` to some extent can be seen as the facade of authMap, instead of a mere cache. For instance, we must acquire the mutex `cacheEntriesMutex` before accessing `authMap`, and `authMap` cannot be used outside of the `authMapCache`. From this point of view, it might be acceptable to put `pressureGauge` in the `authMapCache`.

mhofstetter: @jschwinger233 , yes it's a write&read-through cache - which acts as kind of a facade. the main purpose is to immediately reflect the changes in the BPF  map after successful authentication (without a delay to await a sync). 

But there are other "facades" too - like the authmap writer - which translates the userspace representation of the auth map to the bpf auth map representation.

Therefore my concerns whether the cache is the right place :)

pkg/auth/authmap_cache.go
```diff
@@ -117,8 +125,20 @@ func (r *authMapCache) restoreCache() error {
...
 		Debug("Restored entries")
 	return nil
 }
+
+func (r *authMapCache) updatePressureMetric() {
+	if r.pressureGauge == nil {
+		return
+	}
+	if !metrics.BPFMapPressure {
+		r.pressureGauge = nil
```
markpash, line 140: Could we do this just once in the constructor `newAuthMapCache()`?

PR #22263, opened 2022-11-18
Fix: share etcd logger accross all client fix memory leak, possibly due to cluster mesh

WingkaiHo:
What type of PR is this?
/kind bug

Which issue(s) this PR fixes:
https://github.com/cilium/cilium/issues/13446

What this PR does / why we need it:
Currently cilium creates one etcd client per cluster_mesh. When cluster_mesh reconnecting etcd, it will close old client, and create a new one.  If clients aren't provided a logger they'll each create their own. These loggers can account for >= 30% cilium-agent memory. 

such as in https://github.com/cilium/cilium/issues/13446
avatar

reference:
https://github.com/kubernetes/kubernetes/pull/111477

Issue #26479, opened 2023-06-26
CI: Runtime Test (privileged): Test/linuxPrivilegedIPv6OnlyTestSuite/TestArpPingHandling


qmonnet:
## CI failure

```
level=info msg="Detected devices" devices="[]" subsys=linux-datapath
level=info msg="Detected devices" devices="[dummy0]" subsys=linux-datapath
level=info msg="Detected devices" devices="[dummy0]" subsys=linux-datapath
level=info msg="Direct routing device detected" direct-routing-device=dummy1 subsys=linux-datapath
level=info msg="Detected devices" devices="[dummy0 dummy1 dummy2]" subsys=linux-datapath
level=info msg="Direct routing device detected" direct-routing-device=dummy1 subsys=linux-datapath
level=info msg="Detected devices" devices="[dummy0 dummy1 dummy2]" subsys=linux-datapath
level=info msg="Direct routing device detected" direct-routing-device=cilium_foo subsys=linux-datapath
level=info msg="IPv6 multicast device detected" ipv6-mcast-device=cilium_foo subsys=linux-datapath
level=info msg="Detected devices" devices="[cilium_foo dummy0 dummy1 dummy2]" subsys=linux-datapath
level=info msg="Direct routing device detected" direct-routing-device=cilium_foo subsys=linux-datapath
level=info msg="Detected devices" devices="[cilium_foo dummy0 dummy1 dummy2]" subsys=linux-datapath
level=info msg="Detected devices" devices="[cilium_foo dummy0 dummy1 dummy2 veth0]" subsys=linux-datapath
level=info msg="Detected devices" devices="[cilium_foo dummy0 dummy1 dummy2 dummy3 veth0]" subsys=linux-datapath
level=info msg="Detected devices" devices="[cilium_foo dummy0 dummy1 dummy2 dummy3 veth0]" subsys=linux-datapath
level=info msg="Detected devices" devices="[cilium_foo dummy0 dummy1 dummy2 veth0]" subsys=linux-datapath
level=info msg="Detected devices" devices="[bond0 cilium_foo dummy0 dummy1 veth0]" subsys=linux-datapath
level=info msg="Detected devices" devices="[dummy0 dummy1]" subsys=linux-datapath
level=info msg="Devices changed" devices="[dummy0]" subsys=linux-datapath
level=info msg="Listening for device changes" subsys=linux-datapath
level=info msg="Listening for device changes" subsys=linux-datapath
[...]

PR #26485, opened 2023-06-26
kvstore: share etcd client logger to reduce memory usage

giorio94:
Currently, the etcd client logger is not explicitly configured when a new etcd client is initialized, which means that a new one gets created for every client. Yet, this comes with a significant memory cost, which can be avoided explicitly initializing and configuring a shared logger. This change mimics a similar one implemented for the kubernetes apiserver: https://github.com/kubernetes/kubernetes/pull/111477

The following reports the top 3 nodes from the heap pprof profile of an idle process after initializing 1K etcd clients through `kvstore.NewClient`.

### Without the patch:

Showing nodes accounting for 268.62MB, 88.75% of 302.68MB total Dropped 124 nodes (cum <= 1.51MB)
Showing top 3 nodes out of 60
      flat  flat%   sum%        cum   cum%
  224.33MB 74.11% 74.11%   224.33MB 74.11%  go.uber.org/zap/zapcore.newCounters
   32.45MB 10.72% 84.83%    32.45MB 10.72%  google.golang.org/grpc/internal/transport.newBufWriter
   11.85MB  3.91% 88.75%    11.85MB  3.91%  bufio.NewReaderSize

### With the patch:

Showing nodes accounting for 65.20MB, 69.19% of 94.23MB total Showing top 3 nodes out of 137
      flat  flat%   sum%        cum   cum%
   43.62MB 46.29% 46.29%    43.62MB 46.29%  google.golang.org/grpc/internal/transport.newBufWriter
   19.08MB 20.25% 66.54%    19.08MB 20.25%  bufio.NewReaderSize
    2.50MB  2.65% 69.19%        3MB  3.19%  time.NewTimer

This change was initially proposed by @WingkaiHo as part of https://github.com/cilium/cilium/pull/22263

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -94,6 +94,7 @@ require (
 	go.uber.org/dig v1.17.0
 	go.uber.org/goleak v1.2.1
 	go.uber.org/multierr v1.11.0
+	go.uber.org/zap v1.24.0
 	go.universe.tf/metallb v0.11.0
 	go4.org/netipx v0.0.0-20230303233057-f1b76eb4bb35
 	golang.org/x/crypto v0.10.0
@@ -240,7 +241,6 @@ require (
 	go.mongodb.org/mongo-driver v1.11.3 // indirect
 	go.opentelemetry.io/otel/metric v1.16.0 // indirect
 	go.uber.org/atomic v1.9.0 // indirect
-	go.uber.org/zap v1.24.0 // indirect
 	golang.org/x/oauth2 v0.7.0 // indirect
 	golang.org/x/text v0.10.0 // indirect
 	golang.zx2c4.com/wintun v0.0.0-20230126152724-0fa3db229ce2 // indirect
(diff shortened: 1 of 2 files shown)


rolinh (approved):
I'm not fond of adding an explicit dependency to the zap logger but I guess that at this point it's the only option for configuring the etcd client. Given the benefits in terms of memory allocations, I think it's worth it.

giorio94:
Rebased onto main to pick the latest CI changes.

giorio94:
Runtime Test (privileged) hit https://github.com/cilium/cilium/issues/26479. Manually rerunning.

marseel (approved):
lgtm, thanks!

giorio94:
The PR has been approved by two reviewers (@marseel covered the kvstore-owned files), and the CI is green. Marking as ready-to-merge so that we can still include it in v1.14 without the need for backports.

PR #26486, opened 2023-06-26
routing: remove ip shellout from setUpRoutingTable()

### LINKED ISSUE #14383: Investigate workaround in privileged tests for netlink.RuleAdd() returning EINVAL
See https://github.com/cilium/cilium/pull/14337#discussion_r541360235. For now the workaround is to shell out to `ip rule add` instead of using the netlink library.

ti-mo:
The default value of a netlink.Rule does not seem to be useful. netlink.NewRule must be used instead.

Fixes: #14383

```release-note
Remove 'ip' shellout from setUpRoutingTable()
```


christarazi (approved):
Interesting TIL, thanks!

PR #26488, opened 2023-06-26
Replace LB-IPAM IP allocator

### LINKED ISSUE #22005: lb-ipam: Lift 65536 IPv6 allocation limit per CIDR
The initial LB-IPAM implementation as presented in #21764 has a notable limitation what mainly impacts users with large clusters, that is that a pool can only allocate up to 65536 IPv6 addresses from a single CIDR. The current work around is to break up larger CIDRs into multiple smaller CIDRs.

This limitation is hardcoded into the [ipallocator package](https://github.com/cilium/ipam/blob/46ef3d5567351d242e0c55a2be47c7be87ca4fb9/service/ipallocator/allocator.go#L258) used by LB-IPAM, there to limit memory usage.

This issue proposes to add an additional allocator to this package as alternative to the bitmap allocator with characteristics that allow us to utilize more of IPv6 ranges. 

An alternative would be to implement a allocator just for LB-IPAM which might make it easier to allocate a single IP for multiple services like suggested in #21776

### LINKED ISSUE #24351: LB-IPAM: Unable to set single addresses as a pool

### What happened?

I tried to create a pool with only single IPv4 (/32) and IPv6 (/128) addresses inside it.
I expected it to see that the pool has two IP addresses in total, but it doesn't seem to get that.

Instead it's having very hard time allocating IPs from it.

### Cilium Version

cilium-cli: v0.12.1 compiled with go1.18.5 on linux/amd64
cilium image (default): v1.12.0
cilium image (stable): v1.13.0
cilium image (running): v1.13.0

### Kernel Version

5.15.0-1025-raspi

### Kubernetes Version

```
{
  "clientVersion": {
    "major": "1",
    "minor": "26",
    "gitVersion": "v1.26.1",
    "gitCommit": "8f94681cd294aa8cfd3407b8191f6c70214973a4",
    "gitTreeState": "archive",
    "buildDate": "2023-02-06T14:43:47Z",
    "goVersion": "go1.19.4",
    "compiler": "gc",
    "platform": "linux/amd64"
  },
  "kustomizeVersion": "v4.5.7",
  "serverVersion": {
    "major": "1",
    "minor": "25",
    "gitVersion": "v1.25.6+k3s1",
    "gitCommit": "9176e03c5788e467420376d10a1da2b6de6ff31f",
    "gitTreeState": "clean",
    "buildDate": "2023-01-26T00:30:33Z",
    "goVersion": "go1.19.5",
    "compiler": "gc",
    "platform": "linux/arm64"
  }
}
```

### Sysdump

Too large; It's [here](https://s3.skym.fi/public/cilium-sysdump-20230313-191606.zip)

### Relevant log output

```shell
level=info msg="Notify ingress controller for service ingress" subsys=ingress-controller
level=error msg="Failed to process Ingress event, skipping: {oldIngress:0x4000171f00 newIngress:0x40015bc100}" subsys=ingress-controller
level=error msg="Unable to allocate IP" error="provided IP is not in the valid range. The range of valid IPs is 185.218.193.58/32" subsys=lbipam
level=error msg="Unable to allocate IP" error="provided IP is not in the valid range. The range of valid IPs is 2001:67c:1104:e03e::443/128" subsys=lbipam
level=info msg="Notify ingress controller for service ingress" subsys=ingress-controller
level=error msg="Unable to allocate IP" error="provided IP is not in the valid range. The range of valid IPs is 185.218.193.58/32" subsys=lbipam
level=error msg="Unable to allocate IP" error="provided IP is not in the valid range. The range of valid IPs is 2001:67c:1104:e03e::443/128" subsys=lbipam
```

### Anything else?

This is valid way for MetalLB, and I think it should work with Cilium too.

The pool manifest can be found [here](https://github.com/samip5/k3s-test-cluster/blob/6fa03734ed98edbaf4a86d79c7aa0bec3d70fe06/kubernetes/apps/cilium-system/cilium/cilium-lb-pool.yaml)

The cilium ingress svc looks like this:
```
apiVersion: v1
kind: Service
metadata:
  annotations:
    io.cilium/lb-ipam-ips: 185.218.193.58,2001:67c:1104:e03e::443
    meta.helm.sh/release-name: cilium
    meta.helm.sh/release-namespace: kube-system
  creationTimestamp: "2023-03-13T16:52:43Z"
  labels:
    app.kubernetes.io/managed-by: Helm
    cilium.io/ingress: "true"
    helm.toolkit.fluxcd.io/name: cilium
    helm.toolkit.fluxcd.io/namespace: kube-system
  name: cilium-ingress
  namespace: kube-system
  resourceVersion: "2315905"
  uid: 2b9c0a9c-b58a-45dd-bd47-1e00f21533cc
spec:
  allocateLoadBalancerNodePorts: true
  clusterIP: 10.46.159.107
  clusterIPs:
  - 10.46.159.107
  - 2001:67c:1104:e03e::c94b
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  - IPv6
  ipFamilyPolicy: RequireDualStack
  ports:
  - name: http
    nodePort: 31019
    port: 80
    protocol: TCP
    targetPort: 80
  - name: https
    nodePort: 31816
    port: 443
    protocol: TCP
    targetPort: 443
  sessionAffinity: None
  type: LoadBalancer
status:
  conditions:
  - lastTransitionTime: "2023-03-13T16:52:43Z"
    message: ""
    reason: satisfied
    status: "True"
    type: io.cilium/lb-ipam-request-satisfied
  loadBalancer: {}
```

### Code of Conduct

### LINKED ISSUE #28255: l2 announcer is 🤌 (but weird bugs with /32's)
I switched our cluster over from metallb to the l2 announcer in a single evening with 0-downtime. This is a very awesome feature!

However, it is issuing IP's outside of the IP pool...

Example pool:

```yaml
kind: CiliumLoadBalancerIPPool
metadata:
  name: pool
spec:
  cidrs:
    - cidr: 123.123.123.123/32
    - cidr: 123.123.123.125/32
  disabled: false
```

I'm seeing it issue IP's to services (LB type) to 123.123.123.122, for example. Deleting the status doesn't cause it to reissue the IP address from the pool (in fact, I couldn't figure out how to invalidate the assignment), and had to force an actual IP using the `io.cilium/lb-ipam-ips` annotation.

However, when setting that, I got the following status:

```
No pool exists with a CIDR containing '123.123.123.123'
```

I had to set the `/32` cidr on the annotation, which doesn't match the documentation here: https://docs.cilium.io/en/stable/network/lb-ipam/#requesting-ips

The only thing I can think of is that I'm using /32's because my IP block has some 'holes' in it (external services) that I don't want to be included in the pool, thus I had to enumerate the entire block using /32's.

```
# cilium version
cilium-cli: v0.15.8 compiled with go1.21.0 on linux/amd64
cilium image (default): v1.14.1
cilium image (stable): v1.14.2
cilium image (running): 1.14.2
```

```
# uname -a                                                                                                                                                                                                   
Linux capital 5.15.0-84-generic #93-Ubuntu SMP Tue Sep 5 17:16:10 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

```
#  kubectl version                                                                                                                                                                                            
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.6+k3s1", GitCommit:"bd04941a294793ec92e8703d5e5da14107902e88", GitTreeState:"clean", BuildDate:"2023-09-20T23:05:58Z", GoVersion:"go1.20.8", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v5.0.1
Server Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.6+k3s1", GitCommit:"bd04941a294793ec92e8703d5e5da14107902e88", GitTreeState:"clean", BuildDate:"2023-09-20T23:05:58Z", GoVersion:"go1.20.8", Compiler:"gc", Platform:"linux/amd64"}
```

[cilium-sysdump-20230924-231254.zip](https://github.com/cilium/cilium/files/12709847/cilium-sysdump-20230924-231254.zip)

dylandreimerink:
This PR introduces a new IP allocator, replaces the existing IP allocator used by LB-IPAM with this new allocator. This removes the limitation on IPv6 ranges, the full range can now be used (only limited by the resources required to book keep). It also removes the reservation of the base and broadcast IP address making it possible to allocate single IPs. Lastly, support is now added for non-CIDR IP ranges such as x.x.x.100-x.x.x.200.

Fixes: #24351
Fixes: #22005
Fixes: #28255

```release-note
Replace LB-IPAM IP allocator to remove limitations and enable additional features
```


w7team:
这个pr什么时候合并？

dylandreimerink:
> when will this pr merge?

We are currently stabilizing for the v1.14 version cut. This PR will be undrafted as soon as we can merge feature into the main branch again. First actual release is expected to be v1.15-snapshot.1 whenever that will be.

YutaroHayakawa:
Thanks for working on this! Wrt the network and broadcast address, I'm a bit worrying about enabling it by default. MetalLB supports disabling those addresses with `AvoidBuggyIPs` flag for dealing with buggy devices that filter those IPs (https://metallb.universe.tf/configuration/_advanced_ipaddresspool_configuration/).

Our current implementation is sort of accidentally setting `AvoidBuggyIPs: true` by default. If we start to include "buggy" addresses by default, some users that have such buggy devices may be affected.

Do you think it does make sense to make it opt-in?

dylandreimerink:
> Wrt the network and broadcast address, I'm a bit worrying about enabling it by default.

My original thinking was that, with ranges, it would not be difficult for users to declare a x.x.x.1-x.x.x.254 range if they do not want allocation of the base and broadcast IPs, keeping the CRD clean in that way

> If we start to include "buggy" addresses by default, some users that have such buggy devices may be affected.

Yea, makes sense to me.

> Do you think it does make sense to make it opt-in?

Yes, it makes sense to at least preserve existing behavior so users don't get surprised when updating. I think adding a `AllowBuggyIPs` (inverted) or a `AvoidBuggyIPs` which defaults to true isn't to difficult.

omniproc:
> > Wrt the network and broadcast address, I'm a bit worrying about enabling it by default.
> 
> My original thinking was that, with ranges, it would not be difficult for users to declare a x.x.x.1-x.x.x.254 range if they do not want allocation of the base and broadcast IPs, keeping the CRD clean in that way
> 
> > If we start to include "buggy" addresses by default, some users that have such buggy devices may be affected.
> 
> Yea, makes sense to me.
> 
> > Do you think it does make sense to make it opt-in?
> 
> Yes, it makes sense to at least preserve existing behavior so users don't get surprised when updating. I think adding a `AllowBuggyIPs` (inverted) or a `AvoidBuggyIPs` which defaults to true isn't to difficult.

It would be nice to get a propper error message why the first and last IP of a provided range is ignored in that case so user's don't need to wonder why cillium is not picking them up (unfortunately normal ppl don't calculate subnetmasks in their head on the fly).

dylandreimerink:
> It would be nice to get a propper error message why the first and last IP of a provided range is ignored in that case so user's don't need to wonder why cillium is not picking them up (unfortunately normal ppl don't calculate subnetmasks in their head on the fly).

I assume you mean in the status of a service when a service is specifically requesting a "buggy IP"? Or did you mean more broadly?

>> If we start to include "buggy" addresses by default, some users that have such buggy devices may be affected.
> Yea, makes sense to me.

I am also contemplating making a small exception here. For /32 or /31 ranges, having 1 or 2 IPs, it doesn't make sense under any circumstance to reserve the first and last IP because they might be "buggy". So only /30 and bigger are effected by the setting (/126 for IPv6). That should also help with the UX. And it should not break to many existing setups since /31 and /32 ranges currently are unusable anyway

omniproc:
> I assume you mean in the status of a service when a service is specifically requesting a "buggy IP"? Or did you mean more broadly?

I mean in the status of the  pool CRD when a CIDR was specified and the user tries to use the first or the last IP of that pool. Also logs in the operator should prbl. log an info that the first and last IP of the CIDR are ignored by default.
In contrast to that I think "range" notation is more useful then flags not _not_ ignore the first and last IP of a provided CIDR range. So, for example: 

- CIDR: x.x.x.x/30  -> first and last IP are ignored because CIDR notation was used
- Range: x.x.x.x - x.x.x.y -> first and last IP are nog ignored, it's the user's responsibility to not mess up the (human readable friendly) range

That's how other projects, e.g. kubevip, do it. See config map example here https://kube-vip.io/docs/usage/cloud-provider/

dylandreimerink:
So with regards to dodgy IPs I have settled on the following which I think is the best of all worlds:
* Dodgy IPs are reserved by default on CIDR ranges
* /32, /31 IPv4 and /128, /127 IPv6 CIDRs are fully available regardless of the Dodgy IP setting
* Start+end IP ranges have no reserved IPs 

With that I feel like this is where I want it to be, making ready for review.

Documentation/network/lb-ipam.rst
```diff
@@ -50,11 +53,26 @@ After adding the pool to the cluster, it appears like so.
...
-    The amount of available IPs in the pool is lower than the actual sum of all
-    usable IPs in the CIDRs because the allocation logic is limited to 65536 IPs
-    per CIDR. CIDRs containing more than 65536 IPs can be broken down into multiple
-    smaller CIDRs to achieve full utilization.
+CIDRs, Ranges and dodgy IPs
+---------------------------
+
+An IP pool can have multiple blocks of IPs. A block can be specified with CIDR
+notation (<prefix>/<bits>) or a range notation with a start and stop IP. As
+pictured above.
```
ghost: Can you link to the relevant section instead of saying "above"?

Documentation/network/lb-ipam.rst
```diff
@@ -50,11 +53,26 @@ After adding the pool to the cluster, it appears like so.
...
+"network address" and the last IP is the "broadcast address". In most networks
+these IPs are not usable. In LB-IPAM we refer to these as "Dodgy IPs" because
+they do not always play well with all network equipment. LB-IPAM will not assign
+these by default. Exceptions are /32 and /31 IPv4 ranges and /128 and /127 IPv6
+ranges since these only have 1 or 2 IPs respectively.
+
+If you wish to use these dodgy IPs, you can set the ``.spec.allowDodgyIPs`` field
+to ``true``.
+
+Since ranges are typically used to indicate sub-sections of routable IP ranges,
```
ghost:
```suggestion
Since Ranges are typically used to indicate subsections of routable IP ranges,
```
I would use an uppercase "Range" here to be consistent with the header and make it distinct from a more generic meaning of an IP range (that can be also a CIDR).

Documentation/network/lb-ipam.rst
```diff
@@ -50,11 +53,26 @@ After adding the pool to the cluster, it appears like so.
...
-    smaller CIDRs to achieve full utilization.
+CIDRs, Ranges and dodgy IPs
+---------------------------
+
+An IP pool can have multiple blocks of IPs. A block can be specified with CIDR
+notation (<prefix>/<bits>) or a range notation with a start and stop IP. As
+pictured above.
+
+CIDRs are often used to specify routable IP ranges. By convention, the first
+IP of a range and the last IP of a range are reserved. The first IP is the
```
ghost:
```suggestion
and the last IP of a CIDR are reserved. The first IP is the 
```
Using the word "range" here can be a bit confusing, as this section describes a concept of a range that's an alternative to a CIDR.

Documentation/network/lb-ipam.rst
```diff
@@ -50,11 +53,26 @@ After adding the pool to the cluster, it appears like so.
...
+notation (<prefix>/<bits>) or a range notation with a start and stop IP. As
+pictured above.
+
+CIDRs are often used to specify routable IP ranges. By convention, the first
+IP of a range and the last IP of a range are reserved. The first IP is the 
+"network address" and the last IP is the "broadcast address". In most networks
+these IPs are not usable. In LB-IPAM we refer to these as "Dodgy IPs" because
+they do not always play well with all network equipment. LB-IPAM will not assign
+these by default. Exceptions are /32 and /31 IPv4 ranges and /128 and /127 IPv6
+ranges since these only have 1 or 2 IPs respectively.
```
ghost:
```suggestion
these by default. Exceptions are /32 and /31 IPv4 CIDRs and /128 and /127 IPv6
CIDRs since these only have 1 or 2 IPs respectively.
```

ghost (changes requested):

nebril (approved):

ldelossa (approved):

ghost (approved):
docs good

Documentation/network/lb-ipam.rst
```diff
@@ -50,11 +55,26 @@ After adding the pool to the cluster, it appears like so.
...
+
+CIDRs are often used to specify routable IP ranges. By convention, the first
+and the last IP of a CIDR are reserved. The first IP is the 
+"network address" and the last IP is the "broadcast address". In most networks
+these IPs are not usable. In LB-IPAM we refer to these as "Dodgy IPs" because
+they do not always play well with all network equipment. LB-IPAM will not assign
+these by default. Exceptions are /32 and /31 IPv4 CIDRs and /128 and /127 IPv6
+CIDRs since these only have 1 or 2 IPs respectively.
+
+If you wish to use these dodgy IPs, you can set the ``.spec.allowDodgyIPs`` field
```
ysksuzuki: If we configure an expensive address pool such as global ipv4 addresses, we must set `allowDodgyIPs=true`, correct? If so, I'm not sure this spec makes sense because there are more cases where you want to set `allowDodgyIPs=false` from my experience. Plus, I'm confused because it's the opposite of the MetalLB `AvoidBuggyIPs` flag. 

We use `/28` for global and bastion networks. So I assume we need to set `allowDodgyIPs=true` accordingly.

dylandreimerink: > If we configure an expensive address pool such as global ipv4 addresses, we must set allowDodgyIPs=true, correct? If so, I'm not sure this spec makes sense because there are more cases where you want to set allowDodgyIPs=false from my experience.

Yea, that is correct, if you want to use the first and last IP of the global range then you have to set `allowDodgyIPs` to `true`. If unspecified, the value defaults to `false`, so unless you specify `allowDodgyIPs=true`, the first and last IPs are reserved, the same behavior we currently have.

> I'm confused because it's the opposite of the MetalLB `AvoidBuggyIPs` flag.

Yea, this can't really be helped because we need to maintain backwards compatibility for existing policies, otherwise we may break setups on upgrade.

> We use /28 for global and bastion networks. So I assume we need to set allowDodgyIPs=true accordingly.

Practical example for `10.0.0.128/28`, if set to `true`, the allocate range is `10.0.0.128-10.0.0.143`, if `false` or unspecified its `10.0.0.129-10.0.0.142`. So you should pick which is desirable for your scenario

ysksuzuki: Thank you for your comment!

I'm worried that users migrating from MetalLB(either MetalLB BGP ControlPlane or standalone) will be confused. With that said, it is understandable that backward compatibility must be maintained.

Wouldn't it be acceptable to break compatibility after reminding users in the upgrade guide in this case?

dylandreimerink: > Wouldn't it be acceptable to break compatibility after reminding users in the upgrade guide in this case?

It could be in some cases, but consider the worst scenario for this case if a user doesn't pay attention or misses the notice. Any user with existing pools will suddenly be able to allocate the dodgy IPs, hard to notice if you are not looking for it. Lets say a service gets assigned a broadcast IP in a network which has this feature enabled. At best the feature doesn't work, at worst it causes a DoS or security issue.

I think the confusion concern is valid, but I hope it isn't as big a deal as we might think. I would like to know what others think.

ysksuzuki: I think I'm slightly biased because I'm the user planning to migrate from MetalLB. So yeah, we should ask what others think.

networkop: I would change this behaviour and enable this by default. Here's my rationale:
network and broadcast addresses only make sense on a network segment. When we allocate and announce individual host prefixes, they have no notion of broadcast or network addresses. 
For example, 192.168.255.255 may be a broadcast address for a segment with prefix 192.168.0.0/16 but may be a normal address for prefix 192.168.0.0/15. 
Even if we announce individual host IPs and they get aggregated on a border router, they still are not assigned to any network segment (aggregate route is configured as null-route).

The only caveat may be with L2 announcer where, if the range is picked incorrectly, allocated IP can overlap with local subnet's broadcast address. While this is more of an operational problem, I don't think this will result in any loss of connectivity too.

---

I've got an extra ask, can we please not use the word `dodgy`? This is not a generally-used term and may get a lot of people confused. I can suggest `allowFirstLastlIPs` or `prefixAsRange` or `CIDRasRange`.

dylandreimerink: I brought this issue up in last weeks Community meeting to ask for advice regarding breakage of backwards compatibility. I think we settled on a decent middle ground, that is, in 1.15 we will not break anything, introduce a setting to change the behavior. Then next release, 1.16 we will switch over the default.

We will add warnings in the feature docs and upgrade guide, warning people of the upcoming change, and asking people to explicitly set what they want. That way, we have an upgrade path where we don't break setups on upgrade and change the behavior even if it takes a bit.

I am thinking of changing the setting from a boolean to a enum, so we get something like `allowFirstLastIPs: yes|no|default` with `default` being the value chosen if the field is not specified. This also should make the field different enough from MetalLB's implementation to not cause confusion.

Would this be acceptable?

ysksuzuki: @dylandreimerink Thank you for bringing this up!

So, in 1.15, we will introduce `allowFirstLastIPs=false` by default to ensure backward compatibility, and in 1.16 we will change its default, which is  `allowFirstLastIPs=true` by default. Is my understanding correct?

dylandreimerink: Yes. Currently there is no setting, so if we add the setting and change the default, then users have no chance to set the setting before the default switches. So by doing it in two versions, we allow users some time to explicitly set what they want the behavior to be for existing pools before we switch.

ysksuzuki: I understand. It makes sense to me. Thanks!

dylandreimerink: Alright, should be good now

ysksuzuki (approved):
LGTM. Thanks!

pkg/ipalloc/ipalloc.go
joestringer, line 1: Did we have a discussion about whether this IP allocator might overlap with some other IP allocators we have in the tree?

This file should also be added to CODEOWNERS under @cilium/sig-ipam .

joestringer:
Given IPAM is involved, I think that someone from @cilium/sig-ipam should give this a once-over review from a high level to ensure we're aligned in terms of the project direction and code reuse.

pkg/ipalloc/ipalloc.go
dylandreimerink: I don't think we have, but I am happy to

dylandreimerink:
> I think that someone from https://github.com/orgs/cilium/teams/sig-ipam should give this a once-over review from a high level to ensure we're aligned in terms of the project direction and code reuse.

Yea, makes sense.

> This file should also be added to CODEOWNERS under https://github.com/orgs/cilium/teams/sig-ipam .

We should also discuss this situation. I am currently the defacto owner of the code but I am not part of the IPAM sig, that just happend this way. But it is also a bit of an awkward situation. I don't want to put a burden on someone to maintain a piece of code out of the blue.

pkg/ipalloc/ipalloc.go
christarazi: Coming from sig-ipam, have we looked into how we could make this implementation backwards compatible with the existing users of `pkg/ipam/service/ipallocator` so that all users are using one IPAM implementation? It sounds like this new implementation can be a superset, is that right?

dylandreimerink: Hmm, so here are the two interfaces:

ipallocator:
```
type Interface interface {
	Allocate(net.IP) error
	AllocateNext() (net.IP, error)
	Release(net.IP) error
	ForEach(func(net.IP))
	CIDR() net.IPNet
	Has(ip net.IP) bool
}
```

New LB-IPAM allocator:
```
type Allocator[T any] interface {
	AllocAny(val T) (netip.Addr, error)
	Alloc(ip netip.Addr, val T) error
	Update(ip netip.Addr, val T) error
	Free(ip netip.Addr) error
	Get(ip netip.Addr) (T, bool)
	Stats() (allocated uint64, available *big.Int)
	Range() (from, to netip.Addr)
}
```

The additional features of the LB-IPAM allocator require certain changes in the API. But I think we can make a small adapter to go from the old interface to this one. Its mostly converting between types and placing some restrictions during construction.

```
Interface.Allocate(net.IP) error         -> Allocator.Alloc(ip netip.Addr, val T) error
Interface.AllocateNext() (net.IP, error) -> Allocator.AllocAny(val T) (netip.Addr, error)
Interface.Release(net.IP) error          -> Allocator.Free(ip netip.Addr) error
Interface.ForEach(func(net.IP))          -> // this one would require some modification, but can be done
Interface.CIDR() net.IPNet               -> Allocator.Range() (from, to netip.Addr)
Interface.Has(ip net.IP) bool            -> Allocator.Get(ip netip.Addr) (T, bool)
```

joestringer:
> We should also discuss this situation. I am currently the defacto owner of the code but I am not part of the IPAM sig, that just happend this way. But it is also a bit of an awkward situation. I don't want to put a burden on someone to maintain a piece of code out of the blue.

I'd suggest you join sig-ipam if you're doing work in the area of IPAM. Talk with the other members of the group and come to a shared understanding about where IPAM is going, why these changes are necessary, how to ensure that we are building out this functionality in a maintainable way. Whatever considerations there are for IPAM, I'd imagine that many of those core issues with bugs, maintainability, observability, scale and so on should be common across the general space of IPAM. No-one should have to do that maintenance alone, not you as the author of some piece of code, and not any other particular individual. We have SIGs in order to provide a way to have shared discussion about the future of areas of the code and coordinate together.

If you think that maintaining this piece of code is undue burden on the sig-ipam group, then I suggest identifying what makes code maintenance difficult for this PR and addressing those problems before we merge this code.

tommyp1ckles:
@dylandreimerink @joestringer I'll review for sig-ipam, I just added myself to that team somewhat recently so depending on the risk of these changes it may be worth asking another reviewer as well.

pkg/ipalloc/ipalloc.go
```diff
@@ -0,0 +1,617 @@
...
+	"strings"
+	"time"
+
+	"github.com/cilium/cilium/pkg/lock"
+
+	"golang.org/x/exp/slices"
+)
+
+// Allocator allocates IP addresses
+type Allocator[T any] interface {
```
tommyp1ckles, line 22: Could you be a bit more specific about the intention of the type variable here, I found that a little confused when first reading it.

pkg/ipalloc/ipalloc.go
```diff
@@ -0,0 +1,617 @@
...
+				to:   ip,
+				size: uint128{lo: 1},
+			},
+		})
+		return nil
+	}
+
+	// instead of a for {} loop, pick a unreasonably high iteration count and return an error
+	// when reaching it to prevent infinite loops in case of bugs
+	const maxIter = 1_000_000
```
tommyp1ckles, line 373: could we take the minimum of this and the max size of the allocator to optimize a bit?

pkg/ipalloc/ipalloc.go
```diff
@@ -0,0 +1,617 @@
...
+// Range returns the start and stop IP of the allocator.
+func (a *HashAllocator[T]) Range() (from, to netip.Addr) {
+	return a.start, a.stop
+}
+
+// scheduleSort schedules a sort to happen with one second delay, if a sort is already scheduled we do nothing.
+// This coalesces what otherwise would be multiple sort operations if lots of allocations happen in a short
+// time. Sorting of available blocks improves memory usage in the long term but is not critical to the
+// operation of the allocator, and thus its alright to delay the sort.
+func (a *HashAllocator[T]) scheduleSort() {
```
tommyp1ckles: Just a thought, perhaps we could use a priority queue to maintain these as an ordered set with lg(n) insertions

pkg/ipalloc/ipalloc.go
```diff
@@ -0,0 +1,617 @@
...
+	if availableBlock.to == ip {
+		availableBlock.to = availableBlock.to.Prev()
+		availableBlock.calcSize()
+		return nil
+	}
+
+	// At this point, it must be an IP in the middle of the block, so we need to split it
+
+	// Remove the current block from the list
+	*abl = slices.Delete(*abl, index, index+1)
```
tommyp1ckles: This seems potentially expensive for large, fragmented sets.

pkg/ipalloc/ipalloc.go
```diff
@@ -0,0 +1,617 @@
...
+
+	leftBlock := linkedBlock{
+		block: block{
+			from: availableBlock.from,
+			to:   ip.Prev(),
+			size: addrDiff(availableBlock.from, ip.Prev()),
```
tommyp1ckles: nit: we could use calcSize here to keep things consistent.

pkg/ipalloc/ipalloc.go
```diff
@@ -0,0 +1,617 @@
...
+
+	timer *time.Timer
+
+	allocations     map[netip.Addr]T
+	availableBlocks availableBlockList
+}
+
+var _ Allocator[bool] = (*HashAllocator[bool])(nil)
+
+// NewHashAllocator creates a new IP allocator which will allocate IPs between `start` and `stop`.
```
tommyp1ckles: Nit: Might be helpful to readers to mention that this is an inclusive set on both ends of the range.

pkg/ipalloc/ipalloc.go
```diff
@@ -0,0 +1,617 @@
...
+	prev *linkedBlock
+	next *linkedBlock
+}
+
+func (lb linkedBlock) String() string {
+	var sb strings.Builder
+	if lb.prev != nil {
+		fmt.Fprintf(&sb, "(%s - %s) <- ", lb.prev.from, lb.prev.to)
+	}
+	fmt.Fprintf(&sb, "(%s - %s)[%s]", lb.from, lb.to, lb.bigIntSize().String())
```
tommyp1ckles, line 594: Not sure if this is right, when I do something like:

```go
	alloc, err := NewHashAllocator[bool](
		netip.MustParseAddr("10.0.0.0"),
		netip.MustParseAddr("10.0.0.100"),
		10,
	)
	err = alloc.Alloc(netip.MustParseAddr("10.0.0.1"), true)
	fmt.Println(alloc.availableBlocks)
```
I get the output:

```
linked-block-list[
(10.0.0.0 - 10.0.0.0)[0] -> (10.0.0.2 - 10.0.0.100)
(10.0.0.0 - 10.0.0.0) <- (10.0.0.2 - 10.0.0.100)[98]
]
```

I think the size on the left side should be 1, not 0

pkg/ipalloc/ipalloc.go
```diff
@@ -0,0 +1,617 @@
...
+	availableBlocks availableBlockList
+}
+
+var _ Allocator[bool] = (*HashAllocator[bool])(nil)
+
+// NewHashAllocator creates a new IP allocator which will allocate IPs between `start` and `stop`.
+// The `start` IP should be lower then or equal to `stop`. The `caphint` is passed to the backing
+// hashmap to save on map resizing when initial IP count is broadly known.
+func NewHashAllocator[T any](start, stop netip.Addr, caphint int) (*HashAllocator[T], error) {
+	if start.BitLen() == 0 {
```
tommyp1ckles: This returns no error if you pass `netip.MustParseAddr("0.0.0.0")`, perhaps we'd be better off doing ` isValid()`?

operator/pkg/lbipam/lbipam.go
```diff
@@ -640,26 +641,26 @@ func (ipam *LBIPAM) stripOrImportIngresses(sv *ServiceView) (statusModified bool
...
-func getSVCRequestedIPs(svc *slim_core_v1.Service) []net.IP {
-	var ips []net.IP
+func getSVCRequestedIPs(svc *slim_core_v1.Service) []netip.Addr {
+	var ips []netip.Addr
 	if svc.Spec.LoadBalancerIP != "" {
-		ip := net.ParseIP(svc.Spec.LoadBalancerIP)
-		if ip != nil {
+		ip, err := netip.ParseAddr(svc.Spec.LoadBalancerIP)
+		if err == nil {
 			ips = append(ips, ip)
```
tommyp1ckles, line 652: we should probably log if these are failing to parse.

operator/pkg/lbipam/lbipam.go
```diff
@@ -1089,8 +1089,13 @@ func (ipam *LBIPAM) handleNewPool(ctx context.Context, pool *cilium_api_v2alpha1
 		if fromCidr && pool.Spec.AllowFirstLastIPs != "yes" {
 			// TODO: in 1.16 switch from default no to default yes.
 			from, to := lbRange.alloc.Range()
-			lbRange.alloc.Alloc(from, true)
-			lbRange.alloc.Alloc(to, true)
+
+			// If the first and last IPs are the same or adjacent, we would reserve the entire range.
+			// Only reserve first and last IPs for ranges /30 or /126 and larger.
+			if !(from.Compare(to) == 0 || from.Next().Compare(to) == 0) {
```
tommyp1ckles, line 1104: I'm wondering, if this is meant to be exclusive at both ends by default, should we just consider using `allowFirstLastIPs`when the ranges are /32?

tommyp1ckles:
Very nice example of library API design !

pkg/ipalloc/ipalloc.go
dylandreimerink: Yea, a priority queue indeed sounds like a better idea. That should allow us to eliminate this sorting routine which I didn't like to begin with. I will dedicate some time to try that out.

pkg/ipalloc/ipalloc.go
dylandreimerink: Yea, it could be. Though the list only contains pointers, so its just a copy of 8*#blockRightOfIndex amount of bytes, you would have to have things fragmented very heavily. But if we go for a heap to sort blocks this operation should become cheaper only a pointer update.

pkg/ipalloc/ipalloc.go
dylandreimerink: So you want to base this max iteration count on the side of the list? Something like `maxIter = len(*abl) + someMargin`?

operator/pkg/lbipam/lbipam.go
dylandreimerink: I don't understand the question sorry. This piece of code is to always allow first and last IP for /32 and /31 ranges because reserving the full range is always nonsensical

dylandreimerink:
Small update to prevent this PR from going stale. I have been caught up in some high priority work over the last weeks. And will be unavailable for the next 3 weeks. After that I will process the feedback and get this PR going again.

pkg/ipalloc/ipalloc.go
tommyp1ckles: Yeah was thinking that instead of setting some arbitrarily large max_iter we could find some upper bound based on the allocator size.

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
```diff
@@ -69,11 +69,20 @@ type CiliumLoadBalancerIPPoolSpec struct {
 	//
 	// +kubebuilder:validation:Optional
 	ServiceSelector *slimv1.LabelSelector `json:"serviceSelector"`
+	// AllowFirstLastIPs, if set to yes means that the first and list IPs of each CIDR will be allocatable.
```
danehans: s/list/last/

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
```diff
@@ -69,11 +69,20 @@ type CiliumLoadBalancerIPPoolSpec struct {
...
+	// If no, these IPs will be reserved.
```
danehans: s/no, these IPs/unspecified, the IPs defined in cidrs or blocks/

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
```diff
@@ -69,11 +69,20 @@ type CiliumLoadBalancerIPPoolSpec struct {
...
+	// +kubebuilder:validation:Enum=yes;no
+	AllowFirstLastIPs string `json:"allowFirstLastIPs,omitempty"`
 	// CiliumLoadBalancerIPPoolCIDRBlock is a list of CIDRs comprising this IP Pool
 	//
-	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:MinItems=1
-	Cidrs []CiliumLoadBalancerIPPoolCIDRBlock `json:"cidrs"`
+	// +kubebuilder:validation:Optional
+	Cidrs []CiliumLoadBalancerIPPoolIPBlock `json:"cidrs"`
+	// CiliumLoadBalancerIPPoolCIDRBlock is a list of CIDRs comprising this IP Pool
```
danehans: Why are two fields, `cidrs` and `blocks`, the same type with the same documented behavior? Instead, `cidrs` can continue to represent CIDR notation and then add `spec.ranges` that defines a list of IP ranges. For example:

```go
// CiliumLoadBalancerIPPoolIPRange describes a range of IP addresses.
type CiliumLoadBalancerIPPoolIPRange struct {
        // Start defines the first IP address of an IP pool range.
	// +required
	Start string `json:"start"`
        // Stop defines the last IP address of an IP pool range.
	// +required
	Stop string `json:"stop"`
}
```
Thoughts?

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
```diff
@@ -69,11 +69,20 @@ type CiliumLoadBalancerIPPoolSpec struct {
 	//
 	// +kubebuilder:validation:Optional
 	ServiceSelector *slimv1.LabelSelector `json:"serviceSelector"`
+	// AllowFirstLastIPs, if set to yes means that the first and list IPs of each CIDR will be allocatable.
+	// If no, these IPs will be reserved.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=yes;no
+	AllowFirstLastIPs string `json:"allowFirstLastIPs,omitempty"`
```
danehans: Why not use a `bool` type... are other enums expected in the future?

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
```diff
@@ -69,11 +69,20 @@ type CiliumLoadBalancerIPPoolSpec struct {
...
 	// CiliumLoadBalancerIPPoolCIDRBlock is a list of CIDRs comprising this IP Pool
```
danehans: s/CiliumLoadBalancerIPPoolCIDRBlock/Cidrs/

operator/pkg/lbipam/lbipam.go
```diff
@@ -991,14 +994,14 @@ func (ipam *LBIPAM) allocateIPAddress(
...
+		return newIp, lbRange, nil
 	}
 
 	if full {
-		return nil, nil, ipallocator.ErrFull
+		return netip.Addr{}, nil, ipalloc.ErrFull
 	}
 
-	return nil, nil, nil
+	return netip.Addr{}, nil, nil
```
danehans, line 1012: Why not define `newIP` as a pointer if the func returns an empty value?

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
```diff
@@ -69,11 +69,20 @@ type CiliumLoadBalancerIPPoolSpec struct {
...
 	//
-	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:MinItems=1
-	Cidrs []CiliumLoadBalancerIPPoolCIDRBlock `json:"cidrs"`
+	// +kubebuilder:validation:Optional
+	Cidrs []CiliumLoadBalancerIPPoolIPBlock `json:"cidrs"`
+	// CiliumLoadBalancerIPPoolCIDRBlock is a list of CIDRs comprising this IP Pool
+	//
+	// +kubebuilder:validation:Optional
+	Blocks []CiliumLoadBalancerIPPoolIPBlock `json:"blocks"`
```
danehans: Since the doc states "A block can be specified with CIDR
notation (<prefix>/<bits>) or a range notation with a start and stop IP.", consider making CiliumLoadBalancerIPPoolSpec a [union type](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#unions) and introducing a union discriminator to discriminate among supported allocation types. Using a union type with a discriminator can make validation and code clarity easier.

Example:
```go
// CiliumLoadBalancerIPPoolSpec is a human readable description for
// a load balancer IP pool.
// +union
type CiliumLoadBalancerIPPoolSpec struct {
  // AllocType defines the type of IP allocation to use for this pool.
  // Supported types are:
  //
  // "CIDR": An IP allocator based on CIDR notation, e.g. 1.1.1.0/24.
  // "Range": An IP allocator based on IP range notation, e.g. 1.1.1.1-1.1.1.10.
  //
  // If unset, defaults to CIDR.
  //
  // +unionDiscriminator
  // +required
  // +kubebuilder:default=CIDR
  // +kubebuilder:validation:Enum=CIDR,Range
  AllocType PoolAllocType `json:"allocType"`
  // CiliumLoadBalancerIPPoolCIDRBlock is a list of CIDRs that comprise this IP Pool.
  //
  // This field is only applicable if type is "CIDR".
  //
  Cidrs []CiliumLoadBalancerIPPoolIPBlock `json:"cidrs"`
  // CiliumLoadBalancerIPPoolIPRange is a list of IP ranges that comprise this IP Pool.
  //
  // This field is only applicable if type is "Range".
  //
  Blocks []CiliumLoadBalancerIPPoolIPRange `json:"ranges"`
  ...
}

// PoolAllocType is a type of IP pool allocation.
// +kubebuilder:validation:Enum=CIDR,Range
type PoolAllocType string

const (
  // CIDRPoolAllocType is an IP allocation type based on CIDR notation.
  CIDRPoolAllocType PoolAllocType = "CIDR"
  // RangePoolAllocType is an IP allocation type based on IP range notation.
  RangePoolAllocType PoolAllocType = "Range"
)
```

For backwards compatible upgrades, the union discriminator can instead be `+optional` or implement logic that adds this field to existing objects that were created with the previous version of this CRD.

operator/pkg/lbipam/lbipam.go
```diff
@@ -820,7 +823,7 @@ func (ipam *LBIPAM) satisfyService(sv *ServiceView) (statusModified bool, err er
 	for _, alloc := range sv.AllocatedIPs {
 		// If the allocated IP isn't found in the assigned list, assign it
 		if slices.IndexFunc(sv.Status.LoadBalancer.Ingress, func(in slim_core_v1.LoadBalancerIngress) bool {
-			return net.ParseIP(in.IP).Equal(alloc.IP)
+			return netip.MustParseAddr(in.IP).Compare(alloc.IP) == 0
```
danehans: According to the [docs](https://pkg.go.dev/net/netip#MustParseAddr), MustParseAddr should only be used in tests.

pkg/ipalloc/ipalloc.go
```diff
@@ -0,0 +1,617 @@
...
+}
+
+// uint128FromAddr extracts the uint128 value using a reflection hack.
+func uint128FromAddr(addr netip.Addr) uint128 {
+	addrStruct := reflect.ValueOf(addr).FieldByName("addr")
+	return uint128{
+		hi: addrStruct.FieldByName("hi").Uint(),
+		lo: addrStruct.FieldByName("lo").Uint(),
+	}
+}
```
danehans: Would it be more performant and maintainable to use the `encoding/binary` pkg? For example:
```go
func uint128FromAddr(addr netip.Addr) uint128 {
	bytes := addr.As16()
	return uint128{
		hi: binary.BigEndian.Uint64(bytes[:8]),
		lo: binary.BigEndian.Uint64(bytes[8:]),
	}
}
```

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
dylandreimerink: Because the value is optional, this provides a tri-state type which allows us to preserve existing behavior for users who are upgrading with an existing policy. This could also be done with a `*bool`, however when I brought this up in the community meeting this was preferred, also seems to be more inline with https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

operator/pkg/lbipam/lbipam.go
dylandreimerink: The empty/uninitialized IP is an intended feature of the package in combination with the `IsValid` method. Always passing around values and not pointer avoids a number of foot guns such as nil dereferences and modification of shared values.

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
danehans: @dylandreimerink thanks for the clarification. Note that the API conventions also recommend using string-type aliases instead of enums. For example:

```go
{
...
    AllowFirstLastIPs AllowFirstLastIPType `json:"allowFirstLastIPs,omitempty"`

}

type AllowFirstLastIPType string

const (
    NoAllowFirstLastIPType AllowFirstLastIPType = "No"
    YesAllowFirstLastIPType AllowFirstLastIPType = "Yes"
)
```

pkg/ipalloc/ipalloc.go
dylandreimerink: I added an adapter which converts from the old to the new interface, would be great if you can take a look to see if this is adequate.

pkg/ipalloc/ipalloc.go
dylandreimerink: Yes, its somewhat subtle but if 0 and 100 are allocatable, then 0-100 contains 101 usable IPs. The 98 in the .2-.100 should also be 99 IPs. I have corrected the off by one now.

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
dylandreimerink: > Why are two fields, cidrs and blocks, the same type with the same documented behavior?

We had `cidrs` before, but the idea is to rename it do `blocks`. So we currently have 2 fields which are aliased for backwards compatibility. 

> Instead, cidrs can continue to represent CIDR notation and then add spec.ranges that defines a list of IP ranges

Yes, that would have also worked. I think it comes to do style/preference.

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
dylandreimerink: That is a good point, thanks, I added string-type aliases :+1:

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
dylandreimerink: Perhaps we need to rephrase the docs. But I would like to keep the ability to mix CIDR and range notation in the same spec like:
```
spec:
  blocks:
  - cidr: "10.0.12.0/24"
  - start: "10.0.13.5"
    stop: "10.0.13.8"
```

I see no point in artificially limiting to one or the other.

pkg/ipalloc/ipalloc.go
dylandreimerink: Yes! Did a quick benchmark, this goes from `167.0 ns/op` to `7.397 ns/op`, did not expect that, thanks!

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
```diff
@@ -69,11 +69,21 @@ type CiliumLoadBalancerIPPoolSpec struct {
 	//
 	// +kubebuilder:validation:Optional
 	ServiceSelector *slimv1.LabelSelector `json:"serviceSelector"`
-	// CiliumLoadBalancerIPPoolCIDRBlock is a list of CIDRs comprising this IP Pool
+	// AllowFirstLastIPs, if set to `yes`` means that the first and last IPs of each CIDR will be allocatable.
+	// If `no` or undefined, these IPs will be reserved.
```
danehans, line 73: nit: According to `lb-ipam.rst`, LB-IPAM will ignore this field when specifying a /31-/32 for IPv4 and /131-/132 for IPv6. Not a bad idea to state this in the godocs here.

Documentation/network/lb-ipam.rst
```diff
@@ -128,8 +154,8 @@ CIDR, we will see the following:
 
     $ kubectl get ippools
     NAME        DISABLED   CONFLICTING   IPS AVAILABLE   AGE
-    blue-pool   false      False         254             25m
-    red-pool    false      True          254             11s
+    blue-pool   false      False         256             25m
+    red-pool    false      True          256             11s
```
danehans: This should still be `254` in v1.15 since `.spec.allowFirstLastIPs` defaults to `no`, correct? The same comment for L207.

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
danehans: nit: According to `lb-ipam.rst`, LB-IPAM will ignore this field when specifying a /31-/32 for IPv4 and /131-/132 for IPv6. Not a bad idea to state this in the godocs here.

operator/pkg/lbipam/lbipam.go
```diff
@@ -1070,13 +1081,32 @@ func (ipam *LBIPAM) handleNewPool(ctx context.Context, pool *cilium_api_v2alpha1
...
+		}
+
+		lbRange, err := NewLBRange(from, to, pool)
+		if err != nil {
+			return fmt.Errorf("Error making LB Range for '%s': %w", ipBlock.Cidr, err)
+		}
+
+		// If AllowFirstLastIPs is no or unspecified, mark the first and last IP as allocated upon range creation.
+		if fromCidr && pool.Spec.AllowFirstLastIPs != cilium_api_v2alpha1.AllowFirstLastIPYes {
+			// TODO: in 1.16 switch from default no to default yes.
```
danehans, line 1098: nit: Not a bad idea to create and link an issue here so this TODO is resolved for v1.16.

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
```diff
@@ -82,11 +92,22 @@ type CiliumLoadBalancerIPPoolSpec struct {
 	Disabled bool `json:"disabled"`
 }
 
-// CiliumLoadBalancerIPPoolCIDRBlock describes a single CIDR block.
-type CiliumLoadBalancerIPPoolCIDRBlock struct {
+type AllowFirstLastIPType string
+
+const (
+	AllowFirstLastIPNo  AllowFirstLastIPType = "no"
+	AllowFirstLastIPYes AllowFirstLastIPType = "yes"
```
danehans: nit: from my take on the k8s API conventions, alias string values should be capitalized.

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
```diff
@@ -82,11 +92,22 @@ type CiliumLoadBalancerIPPoolSpec struct {
...
+// CiliumLoadBalancerIPPoolIPBlock describes a single IP block.
+type CiliumLoadBalancerIPPoolIPBlock struct {
 	// +kubebuilder:validation:Format=cidr
-	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Optional
 	Cidr IPv4orIPv6CIDR `json:"cidr"`
+	// +kubebuilder:validation:Optional
+	Start string `json:"start"`
+	// +kubebuilder:validation:Optional
+	Stop string `json:"stop"`
```
danehans: If start/stop represents IP addresses, it's not a bad idea to include validation. For example:
```
// +kubebuilder:validation:Pattern=`((^\s*((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))\s*$)|(^\s*((([0-9A-Fa-f]{1,4}:){7}([0-9A-Fa-f]{1,4}|:))|(([0-9A-Fa-f]{1,4}:){6}(:[0-9A-Fa-f]{1,4}|((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){5}(((:[0-9A-Fa-f]{1,4}){1,2})|:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){4}(((:[0-9A-Fa-f]{1,4}){1,3})|((:[0-9A-Fa-f]{1,4})?:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){3}(((:[0-9A-Fa-f]{1,4}){1,4})|((:[0-9A-Fa-f]{1,4}){0,2}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){2}(((:[0-9A-Fa-f]{1,4}){1,5})|((:[0-9A-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){1}(((:[0-9A-Fa-f]{1,4}){1,6})|((:[0-9A-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(:(((:[0-9A-Fa-f]{1,4}){1,7})|((:[0-9A-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:)))(%.+)?\s*$))`
```

Optional fields should be pointers and include the `omitempty` JSON tag.

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
```diff
@@ -69,11 +69,21 @@ type CiliumLoadBalancerIPPoolSpec struct {
...
+	AllowFirstLastIPs AllowFirstLastIPType `json:"allowFirstLastIPs,omitempty"`
+	// Cidrs is a list of CIDRs comprising this IP Pool
+	// Deprecated: please use the `blocks` field instead. This field will be removed in a future release.
+	//
+	// +kubebuilder:validation:Optional
+	Cidrs []CiliumLoadBalancerIPPoolIPBlock `json:"cidrs"`
+	// Blocks is a list of CIDRs comprising this IP Pool
+	//
+	// +kubebuilder:validation:Optional
+	Blocks []CiliumLoadBalancerIPPoolIPBlock `json:"blocks"`
```
danehans: Optional fields should be omitted from JSON serialization by using the `omitempty` tag.

operator/pkg/lbipam/service_store.go
```diff
@@ -82,7 +83,7 @@ func (sv *ServiceView) isSatisfied() bool {
 		for _, reqIP := range sv.RequestedIPs {
 			// If reqIP doesn't exist in the list of assigned IPs
 			if slices.IndexFunc(sv.Status.LoadBalancer.Ingress, func(in slim_core_v1.LoadBalancerIngress) bool {
-				return net.ParseIP(in.IP).Equal(reqIP)
+				return netip.MustParseAddr(in.IP).Compare(reqIP) == 0
```
danehans: nit: according to the [docs](https://pkg.go.dev/net/netip#MustParseAddr) `MustParseAddr` should only be used in tests.

pkg/bgpv1/manager/route_policy_reconciler_test.go
```diff
@@ -31,7 +31,7 @@ var (
 			},
 		},
 		Spec: v2alpha1api.CiliumLoadBalancerIPPoolSpec{
-			Cidrs: []v2alpha1api.CiliumLoadBalancerIPPoolCIDRBlock{
+			Cidrs: []v2alpha1api.CiliumLoadBalancerIPPoolIPBlock{
```
danehans, line 34: Should the tests be updated to cover IP ranges too?

danehans (approved):
@dylandreimerink this is great work by you. Since my review comments are mostly nits, I `/approve`.

pkg/k8s/apis/cilium.io/v2alpha1/lbipam_types.go
dylandreimerink: Added the `omitempty`. I am holding off on the validation pattern. While I do understand, the reasoning, such regexes are near impossible to review/verify. They are also not very user friendly since kubectl will then spit back out: `{field} does not match {regex}`, which is huge and equally unreadable there. We have validation in place on the consuming side and report back bad IPs via the status (also not ideal). I would really like to see an improved version of this with https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook at some point.

pkg/bgpv1/manager/route_policy_reconciler_test.go
dylandreimerink: This test is part of the BGP package, I am honestly not sure why tests from that package are using LB-IPAM in these tests instead of just manually setting `.status.LoadBalancer.ingress[]`. I don't think its necessary to add extensive coverage of LB-IPAM features here.

pkg/ipalloc/ipalloc.go
```diff
@@ -0,0 +1,636 @@
...
+	"net/netip"
+	"strings"
+
+	"github.com/cilium/cilium/pkg/lock"
+
+	"golang.org/x/exp/slices"
+)
+
+// Allocator allocates IP addresses, `T` is the type of value associated with the IP.
+// The associated value can be used to store metadata about the IP, such as owner, purpose, etc.
```
tommyp1ckles, line 21: 👍

pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumloadbalancerippools.yaml
```diff
@@ -58,20 +58,47 @@ spec:
             description: Spec is a human readable description for a BGP load balancer
               ip pool.
             properties:
+              allowFirstLastIPs:
```
tommyp1ckles, line 61: Would there ever be a need to have this be per block?

dylandreimerink: I don't think so. You can always make 2 pools one with and one without this property set and put CIDRs in the corresponding pools

tommyp1ckles (approved):

clcc2019:
Hope to be merged into the new version as soon as possible, this is important to us

gandro:
> Hope to be merged into the new version as soon as possible, this is important to us

Please note as this is a new feature with API changes, it will land no earlier than Cilium 1.15 (i.e. it will not be backported to Cilium 1.14 in accordance with our [backporting policy](https://docs.cilium.io/en/latest/contributing/release/backports/#backport-criteria-for-current-minor-release)).

christarazi (approved):

kenlasko:
FWIW, I did extensive testing with this new feature in 1.15-pre.2 and it works exactly as expected. Great work everyone!

PR #26489, opened 2023-06-26
Add container scanning workflow with Grype

ferozsalam:
- Scans the same list of images as used in the Image CI build
- Runs daily + on demand
- Fails if any Critical issues are discovered

```release-note
Add container image scanning to Cilium images.
```

### DIFF
--- a/.github/workflows/container-scan.yaml
+++ b/.github/workflows/container-scan.yaml
@@ -1,0 +1,48 @@
+name: Container vulnerability scan
+on:
+  schedule:
+    - cron: "0 4 * * *"
+  workflow_dispatch:
+
+permissions: read-all
+
+jobs:
+  container-scan:
+    if: github.repository_owner == 'cilium'
+    name: scan-container
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        image: [
+          {name: cilium, dockerfile: ./images/cilium/Dockerfile},
+          {name: clustermesh-apiserver, dockerfile: ./images/clustermesh-apiserver/Dockerfile},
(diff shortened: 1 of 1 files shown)


.github/workflows/container-scan.yaml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+        ref: ${{ matrix.branch }}
+      - name: Set environment variables
+        uses: ./.github/actions/set-env-variables
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@4b4e9c3e2d4531116a6f8ba8e71fc6e2cb6e6c8c # v2.5.0
+      - name: Build local container
+        uses: docker/build-push-action@v2
```
viktor-kurchenko: Can we replace `v2` with commit SHA?
The same for the scan action.

viktor-kurchenko (approved):
LGTM.

.github/workflows/container-scan.yaml
ferozsalam: Done, thanks!

joestringer:
Do we need to do anything special in order to test this workflow before we merge it?

ferozsalam:
@joestringer I've tested the workflow using [act](https://github.com/nektos/act) and it works as intended. In any case, this will not block anyone from merging code, and is only set to run infrequently, so I think the chances of negative side effects are minimal.

Let me know if there's anything further you think I should test though!

ferozsalam:
The failing CI checks seem unrelated to my change, so have marked this as `ready-to-merge`

PR #21994, opened 2022-11-04
update generic-veth: set routes with RouteMTU

sofat1989:
```release-note
update generic-veth: set routes with RouteMTU
```

Signed-off-by: Su Fei <fesu@ebay.com>

When setting cilium cni-chaining-mode to generic-veth, our own cni plugin will set up the links and routes. It will not change the mtu values. So by default, it will be 1500. Cilium-cni is only to create the endpoint objects.
The configuration is like this.
```
{
  "name": "generic-veth",
  "cniVersion": "0.2.0",
  "plugins": [
    {
      "name": "ttnet",
      "type": "ttnet",
      "bridge": "cbr0",
      "addIf": "eth0",
      "isGateway": true,
      "ipMasq": true,
      "upstreamIf" : "eth0"
    },
    {
       "name": "cilium",
       "type": "cilium-cni",
       "enable-debug": true,
       "log-file": "/var/run/cilium/cilium-cni.log"
    }
  ]
}
```

host eth0 ( mtu 1500). <---> cilium_vlan (mtu 1500) ---> pod etho (mtu 1500)
At this time, the pod network cannot be pingable if the size is set to larger than 1500 (for example, 2000)
```
/# ping 172.18.253.36 -s 2000
PING 172.18.253.36 (172.18.253.36) 2000(2028) bytes of data
^C
--- 172.18.253.36 ping statistics ---
2 packets transmitted, 0 packets received, 100.0% packet loss
```
That is because when the packets arrive at the host eth0, the packets is larger than its mtu, it will drop this packets.

If the cilium-cni is a standalone plugin, this issue explained above will not happen, because cilium-cni will set the mtu which is gotten from `host eth0 mtu -  tunnel overhead`  for routes inside the containers' namespaces 

for example:
```
default via 172.18.253.1 dev eth0 mtu 1450
```

For cni-chaining-mode generic-veth, cilium-cni should set the MTU ( `host eth0 mtu -  tunnel overhead` ) for the routes inside the containers' namespaces in the same way.

PR #26495, opened 2023-06-27
Set RouteMTU for generic veth

sugangli:

      GitHub issue.

The PR inherits from https://github.com/cilium/cilium/pull/21994.  Once `enable-route-mtu` is true, route MTU will be applied to every route inside the pod. Also resolved some minor conflict in the second commit.

### DIFF
--- a/plugins/cilium-cni/chaining/generic-veth/generic-veth.go
+++ b/plugins/cilium-cni/chaining/generic-veth/generic-veth.go
@@ -67,6 +67,7 @@ func (f *GenericVethChainer) Add(ctx context.Context, pluginCtx chainingapi.Plug
 			return err
 		}
 
+		linkFound := false
 		for _, link := range links {
 			pluginCtx.Logger.Debugf("Found interface in container %+v", link.Attrs())
 
@@ -102,10 +103,49 @@ func (f *GenericVethChainer) Add(ctx context.Context, pluginCtx chainingapi.Plug
 					logfields.Interface: link.Attrs().Name}).Warn("No valid IPv6 address found")
 			}
 
-			return nil
+			linkFound = true
+			break
 		}
 
-		return fmt.Errorf("no link found inside container")
+		if !linkFound {
+			return fmt.Errorf("no link found inside container")
+		}
+
+		if pluginCtx.NetConf.EnableRouteMTU {
+			routes, err := netlink.RouteList(nil, netlink.FAMILY_V4)
+			if err != nil {
+				err = fmt.Errorf("unable to list the IPv4 routes: %s", err.Error())
+				return err
+			}
+			for _, rt := range routes {
+				if rt.MTU != int(pluginCtx.CiliumConf.RouteMTU) {
+					rt.MTU = int(pluginCtx.CiliumConf.RouteMTU)
+					err = netlink.RouteReplace(&rt)
+					if err != nil {
+						err = fmt.Errorf("unable to replace the mtu %d for the route %s: %s", rt.MTU, rt.String(), err.Error())
+						return err
+					}
+				}
+			}
+
+			routes, err = netlink.RouteList(nil, netlink.FAMILY_V6)
+			if err != nil {
+				err = fmt.Errorf("unable to list the IPv6 routes: %s", err.Error())
+				return err
+			}
+			for _, rt := range routes {
+				if rt.MTU != int(pluginCtx.CiliumConf.RouteMTU) {
+					rt.MTU = int(pluginCtx.CiliumConf.RouteMTU)
+					err = netlink.RouteReplace(&rt)
+					if err != nil {
+						err = fmt.Errorf("unable to replace the mtu %d for the route %s: %s", rt.MTU, rt.String(), err.Error())
+						return err
+					}
+				}
+			}
+		}
+
+		return nil
 	}); err != nil {
 		return
 	}
--- a/plugins/cilium-cni/main.go
+++ b/plugins/cilium-cni/main.go
@@ -397,6 +397,11 @@ func cmdAdd(args *skel.CmdArgs) (err error) {
 		return fmt.Errorf("unable to connect to Cilium daemon: %s", client.Hint(err))
 	}
 
+	conf, err := getConfigFromCiliumAgent(c)
+	if err != nil {
+		return err
+	}
+
 	// If CNI ADD gives us a PrevResult, we're a chained plugin and *must* detect a
 	// valid chained mode. If no chained mode we understand is specified, error out.
 	// Otherwise, continue with normal plugin execution.
@@ -405,10 +410,11 @@ func cmdAdd(args *skel.CmdArgs) (err error) {
 			var (
 				res *cniTypesV1.Result
 				ctx = chainingapi.PluginContext{
-					Logger:  logger,
-					Args:    args,
-					CniArgs: cniArgs,
-					NetConf: n,
+					Logger:     logger,
+					Args:       args,
+					CniArgs:    cniArgs,
+					NetConf:    n,
+					CiliumConf: conf,
 				}
 			)
 
@@ -438,11 +444,6 @@ func cmdAdd(args *skel.CmdArgs) (err error) {
 	if err = netns.RemoveIfFromNetNSIfExists(netNs, args.IfName); err != nil {
 		return fmt.Errorf("failed removing interface %q from namespace %q: %s",
 			args.IfName, args.Netns, err)
-	}
-
-	conf, err := getConfigFromCiliumAgent(c)
-	if err != nil {
-		return err
 	}
 
 	var ipam *models.IPAMResponse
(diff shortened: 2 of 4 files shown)


sugangli:
@aojea

plugins/cilium-cni/types/types.go
```diff
@@ -22,16 +22,21 @@ import (
...
+	Azure          azureTypes.AzureSpec   `json:"azure,omitempty"`
+	IPAM           IPAM                   `json:"ipam,omitempty"` // Shadows the JSON field "ipam" in cniTypes.NetConf.
+	AlibabaCloud   alibabaCloudTypes.Spec `json:"alibaba-cloud,omitempty"`
+	EnableDebug    bool                   `json:"enable-debug"`
+	LogFormat      string                 `json:"log-format"`
+	LogFile        string                 `json:"log-file"`
+	ChainingMode   string                 `json:"chaining-mode"`
+}
+
+type ChainingMode struct {
```
aojea: this type is not needed, is it?

plugins/cilium-cni/utils/utils.go
```diff
@@ -0,0 +1,20 @@
+package utils
```
aojea: I think you need the header

```
// SPDX-License-Identifier: Apache-2.0
// Copyright Authors of Cilium
```

plugins/cilium-cni/chaining/generic-veth/generic-veth.go
```diff
@@ -102,10 +110,33 @@ func (f *GenericVethChainer) Add(ctx context.Context, pluginCtx chainingapi.Plug
...
+			linkFound = true
+			break
+		}
+
+		if !linkFound {
+			return fmt.Errorf("no link found inside container")
+		}
+
+		if pluginCtx.NetConf.EnableRouteMTU {
+			routes, err = netlink.RouteList(nil, netlink.FAMILY_V4)
```
aojea: only ipv4?

aojea:
/assign @squeed 

this makes sense to me if cilium is allowed to modify the network inside the netnamespace after it was created on chaining mode, I do not know what the spec says about this.

cilium already has a configured route-mtu on the daemon and this opts-in to modify all the existing routes inside the network namespace to apply this value ... it seems that if not in chaining mode it already installs the routes with the configured mtu so the behavior should be the same

anyway, in the long term I think that setting the MTU in the cni configuration of routes will be better https://github.com/containernetworking/cni/issues/1004

plugins/cilium-cni/chaining/generic-veth/generic-veth.go
sugangli: Added the support for Ipv6 as well.

plugins/cilium-cni/utils/utils.go
sugangli: Done.

plugins/cilium-cni/types/types.go
sugangli: Deleted.

sugangli:
Looks like there is a build check for every commit. And the first commit is not self contained. Should I squash two commits together? @squeed

aojea:
> Looks like there is a build check for every commit. And the first commit is not self contained. Should I squash two commits together? @squeed

I would say yes, just retain the authors with the Co-Author-By

plugins/cilium-cni/chaining/generic-veth/generic-veth.go
```diff
@@ -49,11 +51,17 @@ func (f *GenericVethChainer) Add(ctx context.Context, pluginCtx chainingapi.Plug
 	}()
 	var (
 		hostMac, vethHostName, vethLXCMac, vethIP, vethIPv6 string
+		routes                                              []netlink.Route
 		vethHostIdx, peerIndex                              int
 		peer                                                netlink.Link
 		netNs                                               ns.NetNS
 	)
 
+	conf, err := ciliumcniutils.GetConfigFromCiliumAgent(cli)
```
tklauser: Instead of factoring this out into `ciliumcniutils` and calling the Cilium agent OpenAPI call a second time to retrieve the agent config here, how about storing the result of the initial call in `cmdAdd` as `*models.DaemonConfigurationStatus` in `PluginContext` and then use it here? The agent config is not expected to change in between the calls.

sugangli: Make sense.  I have updated accordingly.

plugins/cilium-cni/chaining/generic-veth/generic-veth.go
```diff
@@ -49,11 +49,16 @@ func (f *GenericVethChainer) Add(ctx context.Context, pluginCtx chainingapi.Plug
 	}()
 	var (
 		hostMac, vethHostName, vethLXCMac, vethIP, vethIPv6 string
+		routes                                              []netlink.Route
 		vethHostIdx, peerIndex                              int
 		peer                                                netlink.Link
 		netNs                                               ns.NetNS
 	)
 
+	if err != nil {
```
aojea: this does not make sense here, does it?

plugins/cilium-cni/main.go
```diff
@@ -397,6 +397,8 @@ func cmdAdd(args *skel.CmdArgs) (err error) {
 		return fmt.Errorf("unable to connect to Cilium daemon: %s", client.Hint(err))
 	}
 
+	conf, err := getConfigFromCiliumAgent(c)
+
```
aojea, line 404: you moved the line bot not the err check

plugins/cilium-cni/main.go
```diff
@@ -440,7 +443,6 @@ func cmdAdd(args *skel.CmdArgs) (err error) {
 			args.IfName, args.Netns, err)
 	}
 
-	conf, err := getConfigFromCiliumAgent(c)
```
aojea, line 443: https://github.com/cilium/cilium/pull/26495/files#r1245744566

plugins/cilium-cni/main.go
sugangli: Ah... Updated.

aojea:
@tklauser @squeed this should be ready for last review

tklauser (approved):

squeed (approved):
This looks good.

Please consider a follow-up PR that adds this value to the `cni` section in the Helm chart. This will then need to be plumbed through the entire configuration system, all the way down to the CNI file.

sugangli:
> This looks good.
> 
> Please consider a follow-up PR that adds this value to the `cni` section in the Helm chart. This will then need to be plumbed through the entire configuration system, all the way down to the CNI file.

Thanks for reviewing! Will update the helm chart later.

joestringer:
> Please consider a follow-up PR that adds this value to the cni section in the Helm chart. This will then need to be plumbed through the entire configuration system, all the way down to the CNI file.

Isn't that already the case? There's an `MTU` option in Helm that is set to autodetect by default, and that configures in the agent, then this code is pulling that configuration from the agent right? What more is there to configure?

joestringer:
Should we nominate this for v1.14 and/or v1.13 backport? This is currently proposed as a `release-note/misc` but I would have thought that incorrectly configured MTU would cause connectivity problems for folks using chaining mode.

joestringer:
> Once `enable-route-mtu` is true

What is this setting? I don't find any matches for this or similar variations in the tree (using `git grep -i ...`)

sugangli:
> > Once `enable-route-mtu` is true
> 
> What is this setting? I don't find any matches for this or similar variations in the tree (using `git grep -i ...`)
> 
> EDIT: Ah, I see it in the PR now. I had initially assumed that this just inherits the route configuration from the cilium-agent just like how other CNI modes would. I'm not sure I understand why this needs to be separately configured, why not just always pull it from the cilium-agent?

If there is a primary CNI,  it will be easier to put all CNI related configuration in one place.  For example, gke CNI has it own configuration, and passes along the config to cilium CNI through the file, and it does not interact with the cilium-agent.

aojea:
Since this is supposed to be used in chaining mode maybe the agent is not the source of truth?

@squeed WDYT?

squeed:
Agreed, this is reasonable -- there are cases where we don't or can't know the correct MTU. We're being a good CNI chained plugin citizen when we ensure that MTUs set up by the "primary" plugin are preserved.

joestringer:
> We're being a good CNI chained plugin citizen when we ensure that MTUs set up by the "primary" plugin are preserved.

If Cilium is imposing additional MTU constraints due to features enabled in Cilium (such as tunneling, encryption, etc.), then "respecting" the "primary" plugin's MTU is just going to break connectivity. While I appreciate that in principle the primary plugin's configuration is "correct" in such a regard, the user won't care if ultimately the MTU is misconfigured and that causes connectivity issues.

joestringer:
> What enable-route-mtu actually means is "tell the cilium cni to set route mtu if other CNI has already set the veth pairs."

I'm curious, why not always enable this option? Is there a scenario you have in mind where we *don't* want Cilium to configure route MTU?

PR #26500, opened 2023-06-27
test: Enable IPv6 datapath_configuration tests with BPF masquerading

qmonnet:
Following the introduction of IPv6 BPF masquerading, we can enable the related tests in datapath_configuration.go.

Logically, these tests should have been enabled when we added IPv6 BPF masquerading, but I missed them.


christarazi (approved):

qmonnet:
Ginko failure's likely related :disappointed: 

image

qmonnet:
Fails on bpf-next only, which is the only setup where we have the 3rd node without K8s, necessary condition to run the test.

Maybe worth trying to replicate _without_ masquerading, to check whether simple IPv6 connects at all in this configuration (thanks Julian for the suggestion).

PR #26501, opened 2023-06-27
bpf: nat: Handle errors from snat_v(4|6)_prepare_state()

qmonnet:
Functions `snat_v4_prepare_state()` and `snat_v6_prepare_state()` woule return a boolean, indicating whether SNAT should happen. They would not report errors that might occur when trying to determine that.

In this commit, we change they return type to an int, in order to report any error to the caller. The SNAT decision is reported via a pointer to a boolean, passed as a new argument.

We report errors if any of `revalidate_data()`, `ct_is_reply4()`, `ct_is_reply6()`, or `ipv6_hdrlen()` fails.

### DIFF
--- a/bpf/lib/nat.h
+++ b/bpf/lib/nat.h
@@ -711,11 +711,12 @@ static __always_inline void snat_v4_init_tuple(const struct iphdr *ip4,
  * - target->egress_gateway to true if the packet should be SNAT-ed because of
  *   an egress gateway policy
  *
- * The function will return true if the packet should be SNAT-ed, false
- * otherwise.
+ * On success, the function returns NAT_NEEDED if the packet should be SNAT-ed,
+ * or NAT_PUNT_TO_STACK if it should not. On failure, it returns a negative
+ * error code (distinct from NAT_PUNT_TO_STACK).
  */
-static __always_inline bool snat_v4_prepare_state(struct __ctx_buff *ctx,
-						  struct ipv4_nat_target *target)
+static __always_inline int
+snat_v4_prepare_state(struct __ctx_buff *ctx, struct ipv4_nat_target *target)
 {
 	void *data, *data_end;
 	struct iphdr *ip4;
@@ -725,7 +726,7 @@ static __always_inline bool snat_v4_prepare_state(struct __ctx_buff *ctx,
 	bool is_reply = false;
 
 	if (!revalidate_data(ctx, &data, &data_end, &ip4))
-		return false;
+		return DROP_INVALID;
 
 	/* Basic minimum is to only NAT when there is a potential of
 	 * overlapping tuples, e.g. applications in hostns reusing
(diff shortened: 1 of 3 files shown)


bpf/lib/nat.h
```diff
@@ -711,11 +711,15 @@ static __always_inline void snat_v4_init_tuple(const struct iphdr *ip4,
  * - target->egress_gateway to true if the packet should be SNAT-ed because of
  *   an egress gateway policy
  *
- * The function will return true if the packet should be SNAT-ed, false
+ * The function will also set *snat_needed to true if the packet should be
+ * SNAT-ed, to false otherwise.
+ *
```
julianwiedmann: Note that one common pattern in the NAT code is to return `NAT_PUNT_TO_STACK` for "no error, but also no NAT handling needed". Instead of having a distinct `snat_needed` return value.

Not sure if this would make things much easier here.

qmonnet: Do you mean, instead of returning an `int` plus passing the bool pointer? I hesitated between the two solutions, having error on < 0, no SNAT on 0 (or on `NAT_PUNT_TO_STACK`), and SNAT on > 0 (or >= 0). Passing the bool pointer is more verbose in `snat_v*_prepare_state()`, but I thought it might be easier to follow. I don't mind switching, though, if you think it would be more consistent?

julianwiedmann: > Do you mean, instead of returning an `int` plus passing the bool pointer?

Yep, exactly.

> Passing the bool pointer is more verbose in `snat_v*_prepare_state()`, but I thought it might be easier to follow.

Imho part of the current verbosity comes from all the (redundant?) `*snat_needed = false;` statements. If we expect the caller to initialize with `false`, do we still need these?

> I don't mind switching, though, if you think it would be more consistent?

Given that the callers already have awareness of what `NAT_PUNT_TO_STACK` means, it does feel a bit more consistent. But I know this code in way too much detail by now, hard to say what's easier to grasp on a casual glance :).

qmonnet: OK let's use `NAT_PUNT_TO_STACK` then. I'll use the comment on top of the IPv4 function to describe what the expectations are about the return value.

bpf/lib/nat.h
```diff
@@ -1698,26 +1701,26 @@ snat_v6_prepare_state(struct __ctx_buff *ctx, struct ipv6_nat_target *target)
 	struct ipv6hdr *ip6;
 
 	if (!revalidate_data(ctx, &data, &data_end, &ip6))
-		return false;
+		return NAT_PUNT_TO_STACK;
```
julianwiedmann:
```suggestion
		return DROP_INVALID;
```

bpf/lib/nodeport.h
```diff
@@ -104,11 +104,12 @@ static __always_inline int nodeport_snat_fwd_ipv6(struct __ctx_buff *ctx,
...
 	};
-	int ret = CTX_ACT_OK;
-	bool snat_needed;
+	int snat_needed, ret = CTX_ACT_OK;
 
 	snat_needed = snat_v6_prepare_state(ctx, &target);
-	if (snat_needed)
+	if (IS_ERR(snat_needed) && snat_needed != NAT_PUNT_TO_STACK)
+		return snat_needed;
+	if (snat_needed != NAT_PUNT_TO_STACK)
```
julianwiedmann: I *think* the following should be equivalent, and hopefully easier to read:

```suggestion
	ret = snat_v6_prepare_state(ctx, &target);
	if (ret == NAT_NEEDED)
		ret = snat_v6_nat(ctx, &target, ext_err);

	if (ret == NAT_PUNT_TO_STACK)
		ret = CTX_ACT_OK;

	ctx_snat_done_set(ctx);
	return ret;
```

julianwiedmann (changes requested):
:+1: that feels much easier to read. One fix-up needed, and one suggestion on how to handle the return value.

bpf/lib/nat.h
qmonnet: Right, I paid attention to it for IPv4 but missed this one. Good catch

bpf/lib/nodeport.h
qmonnet: Yes, I hesitated between keeping exactly the same checks or using the new `NAT_NEEDED`. I also believe this would be more readable, I'll go for that. Thanks!

julianwiedmann (approved):

PR #26502, opened 2023-06-27
helm: Allow node port allocation for Ingress LB service

sayboras:
### Description

This commit is to make sure that users can have the option to disable node port allocation for LB service used by Ingress.

https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-nodeport-allocation

### Testing
Testing was done locally for below cases:
- [x] Default case i.e. no helm option specified
- [x] Set helm flag ingressController.service.allocateLoadBalancerNodePorts=false
- [x] Set helm flag ingressController.service.allocateLoadBalancerNodePorts=true

### DIFF
--- a/install/kubernetes/cilium/README.md
+++ b/install/kubernetes/cilium/README.md
@@ -516,7 +516,8 @@ contributors across the globe, there is almost always someone available to help.
 | ingressController.secretsNamespace.create | bool | `true` | Create secrets namespace for Ingress. |
 | ingressController.secretsNamespace.name | string | `"cilium-secrets"` | Name of Ingress secret namespace. |
 | ingressController.secretsNamespace.sync | bool | `true` | Enable secret sync, which will make sure all TLS secrets used by Ingress are synced to secretsNamespace.name. If disabled, TLS secrets must be maintained externally. |
-| ingressController.service | object | `{"annotations":{},"insecureNodePort":null,"labels":{},"loadBalancerClass":null,"loadBalancerIP":null,"name":"cilium-ingress","secureNodePort":null,"type":"LoadBalancer"}` | Load-balancer service in shared mode. This is a single load-balancer service for all Ingress resources. |
+| ingressController.service | object | `{"allocateLoadBalancerNodePorts":null,"annotations":{},"insecureNodePort":null,"labels":{},"loadBalancerClass":null,"loadBalancerIP":null,"name":"cilium-ingress","secureNodePort":null,"type":"LoadBalancer"}` | Load-balancer service in shared mode. This is a single load-balancer service for all Ingress resources. |
+| ingressController.service.allocateLoadBalancerNodePorts | string | `nil` | Configure if node port allocation is required for LB service ref: https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-nodeport-allocation |
 | ingressController.service.annotations | object | `{}` | Annotations to be added for the shared LB service |
 | ingressController.service.insecureNodePort | string | `nil` | Configure a specific nodePort for insecure HTTP traffic on the shared LB service |
 | ingressController.service.labels | object | `{}` | Labels to be added for the shared LB service |
--- a/install/kubernetes/cilium/templates/cilium-ingress-service.yaml
+++ b/install/kubernetes/cilium/templates/cilium-ingress-service.yaml
@@ -28,6 +28,9 @@ spec:
   {{- if .Values.ingressController.service.loadBalancerClass }}
   loadBalancerClass: {{ .Values.ingressController.service.loadBalancerClass }}
   {{- end }}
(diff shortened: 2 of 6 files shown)


install/kubernetes/cilium/templates/cilium-ingress-service.yaml
```diff
@@ -28,6 +28,9 @@ spec:
   {{- if .Values.ingressController.service.loadBalancerClass }}
   loadBalancerClass: {{ .Values.ingressController.service.loadBalancerClass }}
   {{- end }}
+  allocateLoadBalancerNodePorts: {{ .Values.ingressController.service.allocateLoadBalancerNodePorts }}
```
gandro, line 32: According to [these docs](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-nodeport-allocation) this field is only stable in v1.24. What happens on older versions of Kubernetes?

Should we only add the key here if the Helm value is explicitly set to true or false? (i.e. via `hasKey`)?

sayboras: good catch 💯, let me make one update, thanks a lot.

gandro: As far as I understand, K8s defaults to `true` if not explicitly provided. With the updated code (`if .Values.ingressController.service.allocateLoadBalancerNodePorts`) there is no way to set it to `false`, no?

sayboras: thanks :+1:

install/kubernetes/cilium/templates/cilium-ingress-service.yaml
```diff
@@ -28,6 +28,9 @@ spec:
   {{- if .Values.ingressController.service.loadBalancerClass }}
   loadBalancerClass: {{ .Values.ingressController.service.loadBalancerClass }}
   {{- end }}
+  {{- if hasKey .Values.ingressController.service "allocateLoadBalancerNodePorts" }}
```
gandro: Won't this always be true if we have a default `~` value here? 

We should use `{{- if (not (kindIs "invalid" .Values.ingressController.service.allocateLoadBalancerNodePorts)) }}` instead here (c.f. `bpf.hostLegacyRouting` for a similar example).

install/kubernetes/cilium/README.md
```diff
@@ -512,7 +512,8 @@ contributors across the globe, there is almost always someone available to help.
...
 | ingressController.secretsNamespace.sync | bool | `true` | Enable secret sync, which will make sure all TLS secrets used by Ingress are synced to secretsNamespace.name. If disabled, TLS secrets must be maintained externally. |
-| ingressController.service | object | `{"annotations":{},"insecureNodePort":null,"labels":{},"loadBalancerClass":null,"loadBalancerIP":null,"name":"cilium-ingress","secureNodePort":null,"type":"LoadBalancer"}` | Load-balancer service in shared mode. This is a single load-balancer service for all Ingress resources. |
+| ingressController.service | object | `{"allocateLoadBalancerNodePorts":null,"annotations":{},"insecureNodePort":null,"labels":{},"loadBalancerClass":null,"loadBalancerIP":null,"name":"cilium-ingress","secureNodePort":null,"type":"LoadBalancer"}` | Load-balancer service in shared mode. This is a single load-balancer service for all Ingress resources. |
+| ingressController.service.allocateLoadBalancerNodePorts | string | `nil` | Configure if node port allocation is required https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-nodeport-allocation |
```
learnitall: Two changes that I'd recommend here. First, can you add `ref: ` before the external link? That is a common pattern used throughout this file and it would be great to keep things consistent. Second, can you adjust the description to explicitly include that this is node port allocation for load balancers. This would be helpful for users who may not be as familiar with this topic.

learnitall (changes requested):

gandro:
Does this need the `needs-backport/1.14` label?

install/kubernetes/cilium/templates/cilium-ingress-service.yaml
nathanjsweet: I agree.

nathanjsweet (changes requested):

install/kubernetes/cilium/README.md
sayboras: Thanks :100:

install/kubernetes/cilium/templates/cilium-ingress-service.yaml
sayboras: thanks for your pointer both :100:

learnitall (approved):
Looks good thanks!

gandro (approved):
Thanks!

sayboras:
Docs failure will be fixed in https://github.com/cilium/cilium/pull/26780

nathanjsweet (approved):

PR #26510, opened 2023-06-27
bpf: nat: fix build error in snat_v6_prepare_state()

julianwiedmann:
Without ENABLE_MASQUERADE_IPV6, the `remote_ep` variable isn't used. Fix up the build.

```
In file included from bpf_lxc.c:56:
./cilium/bpf/lib/nat.h:1694:31: error: variable 'remote_ep' set but not used [-Werror,-Wunused-but-set-variable]
        struct remote_endpoint_info *remote_ep;
                                     ^
1 error generated.
```

### DIFF
--- a/bpf/lib/nat.h
+++ b/bpf/lib/nat.h
@@ -1691,7 +1691,7 @@ snat_v6_prepare_state(struct __ctx_buff *ctx, struct ipv6_nat_target *target)
 {
 	union v6addr masq_addr __maybe_unused, router_ip __maybe_unused;
 	const union v6addr dr_addr __maybe_unused = IPV6_DIRECT_ROUTING;
-	struct remote_endpoint_info *remote_ep;
+	struct remote_endpoint_info *remote_ep __maybe_unused;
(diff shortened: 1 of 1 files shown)


qmonnet (approved):
Looks good, thank you!

Out of curiosity, where did you observe the warning? I'm surprised it didn't show up when I compiled locally or when this went through the CI :thinking:

julianwiedmann:
> Looks good, thank you!
> 
> Out of curiosity, where did you observe the warning? I'm surprised it didn't show up when I compiled locally or when this went through the CI 🤔

A local `make -C bpf build_all` was sufficient. And yes, I would also have expected that we run this in CI :)

qmonnet:
I did that, and checked again after I saw your PR, but got no error :open_mouth:. This is weird.

qmonnet:
Ah got it, this has to do with the clang version. I compiled on a machine with an old clang 10.0, when this warning [was added to LLVM 13](https://reviews.llvm.org/D109642).

PR #26517, opened 2023-06-27
docs: Add L2 Pod Announcements docs

markpash:
Add some brief documentation about the L2 Pod Annoucements feature to the existing L2 Annoucements docs page, as the features are very similar.

### DIFF
--- a/Documentation/network/l2-announcements.rst
+++ b/Documentation/network/l2-announcements.rst
@@ -330,3 +330,39 @@ has been reached.
 .. note::
    Since this feature has no IPv6 support yet, only ARP messages are sent, no 
    Unsolicited Neighbor Advertisements are sent.
+
+.. _l2_pod_announcements:
+
+L2 Pod Announcements
+####################
+
+L2 Pod Announcements announce Pod IP addresses on the L2 network using
+Gratuitous ARP replies. When enabled, the node transmits Gratuitous ARP
+replies for every locally created pod, on the configured network
+interface. This feature is enabled separately from the above L2
+announcements feature.
+
+To enable L2 Pod Announcements, set the following:
+
+.. tabs::
+    .. group-tab:: Helm
+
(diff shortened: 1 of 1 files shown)


Documentation/network/l2-announcements.rst
```diff
@@ -330,3 +330,40 @@ has been reached.
...
+
+.. _l2_pod_announcements:
+
+************************************
+L2 Pod Announcements
+************************************
+
+L2 Pod Announcements allows for announcing Pod IP addresses on the L2
+network using Gratuitous ARP replies. When enabled, the node will
+transmit Gratuitous ARP replies for every locally created pod, on the
```
qmonnet:
```suggestion
L2 Pod Announcements announce Pod IP addresses on the L2
network using Gratuitous ARP replies. When enabled, the node
transmits Gratuitous ARP replies for every locally created pod, on the
```

Documentation/network/l2-announcements.rst
```diff
@@ -330,3 +330,40 @@ has been reached.
 .. note::
    Since this feature has no IPv6 support yet, only ARP messages are sent, no 
    Unsolicited Neighbor Advertisements are sent.
+
+.. _l2_pod_announcements:
+
+************************************
+L2 Pod Announcements
+************************************
```
qmonnet: You're formatting this as a top-level title, at the same level as the page title. It would probably look more consistent to have just one such title on the page, have you considered decreasing the level? If it should be top-level, should it be on its own page?

qmonnet (changes requested):

Documentation/network/l2-announcements.rst
markpash: Good point, I've reduced the level.

qmonnet (approved):
Looks good, thank you!

PR #26524, opened 2023-06-28
treewide: wrap multiple errors using the standard library

rolinh:
Go 1.20 expands support for error wrapping to permit an error to wrap multiple other errors. Therefore, `go.uber.org/multierr` is no longer required. See https://go.dev/doc/go1.20#errors for more details.
 
 EDIT: I should have called this out: there's one difference in the error string when multiple errors are wrapped together: with `multierr`, the error strings are joined with `; ` while with the standard library `errors.Join()`, they are [joined with a newline](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/errors/join.go;l=42) `\n`. I think this _may_ be a problem when logging systems expect one line per log.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -71,6 +71,10 @@ linters-settings:
               - testing
               - github.com/stretchr/testify/assert
             reason: "gocheck has been deprecated, see https://docs.cilium.io/en/latest/contributing/testing/unit/#migrating-tests-off-of-gopkg-in-check-v1"
+        - go.uber.org/multierr:
+            recommendations:
+              - errors
+            reason: "Go 1.20+ has support for combining multiple errors, see https://go.dev/doc/go1.20#errors"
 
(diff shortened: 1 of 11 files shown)


sayboras (approved):

gandro (approved):
My codeowners look good to me!

ysksuzuki (approved):

rolinh:
I should have called this out: there's one difference in the error string when multiple errors are wrapped together: with `multierr`, the error strings are joined with `; ` while with the standard library `errors.Join()`, they are [joined with a newline](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/errors/join.go;l=42) `\n`. I think this _may_ be a problem when logging systems expect one line per log.

pippolo84 (approved):
LGTM

PR #26526, opened 2023-06-28
docs: Istio docs fix sidecar inject method

networkop:
Turns out `istioctl kube-inject` does not work with URLs, so this commit changes how the sidecards are injected in the demo application example.

### DIFF
--- a/Documentation/network/servicemesh/istio.rst
+++ b/Documentation/network/servicemesh/istio.rst
@@ -70,12 +70,11 @@ Start by deploying a set of web servers and client applications across three dif
 .. parsed-literal::
 
     kubectl create ns red
-    kubectl -n red apply -f <(istioctl kube-inject -f \ |SCM_WEB|\/examples/kubernetes-istio/httpbin.yaml)
-    kubectl -n red apply -f <(istioctl kube-inject -f \ |SCM_WEB|\/examples/kubernetes-istio/httpbin.yaml)
-    kubectl -n red apply -f <(istioctl kube-inject -f \ |SCM_WEB|\/examples/kubernetes-istio/netshoot.yaml)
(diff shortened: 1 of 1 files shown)


Documentation/network/servicemesh/istio.rst
```diff
@@ -70,15 +70,23 @@ Start by deploying a set of web servers and client applications across three dif
 .. parsed-literal::
 
     kubectl create ns red
-    kubectl -n red apply -f <(istioctl kube-inject -f \ |SCM_WEB|\/examples/kubernetes-istio/httpbin.yaml)
-    kubectl -n red apply -f <(istioctl kube-inject -f \ |SCM_WEB|\/examples/kubernetes-istio/httpbin.yaml)
-    kubectl -n red apply -f <(istioctl kube-inject -f \ |SCM_WEB|\/examples/kubernetes-istio/netshoot.yaml)
+    kubectl -n red apply -f \ |SCM_WEB|\/examples/kubernetes-istio/httpbin.yaml
```
mhofstetter: to keep the one-liner (and avoid a re-deployment) - it would also be possible to use curl to pipe it into the input of `istioctl kube-inject`:

```
kubectl -n red apply -f <(curl -s <URL> |  istioctl kube-inject -f -)
```

mhofstetter (approved):
LGTM, thanks @networkop 

non-blocking: personally i would prefer an approach to keep the one-liners and avoid a redeployment

Documentation/network/servicemesh/istio.rst
networkop: yeah, i like it too. will update in a sec

qmonnet (approved):

qmonnet:
Does this need a backport?

ldelossa:
Docs only PR, not running full suite.

qmonnet:
> Does this need a backport?

Seems like https://github.com/cilium/cilium/commit/0f64b959648ae8a9374fa09d4da8dd560068520a will be in 1.14, so let's backport to that branch maybe

PR #26527, opened 2023-06-28
bpf: ct: simplify ct_action parameter for CT lookup

julianwiedmann:
For a CT lookup, the calling paths currently need to determine a
`ct_action` parameter. This is typically ACTION_CREATE (for TCP/UDP/SCTP),
or ACTION_UNSPEC for most types of ICMP traffic.

But ever since https://github.com/cilium/cilium/commit/fd5ea2a3d2132c1f130be2ccf2b465e0e9754041 ("bpf: Don't reset TCP timer on final ACK"),
__ct_lookup() will only apply ACTION_CREATE handling to TCP packets with
the SYN flag set. Otherwise it has the same effect as ACTION_UNSPEC.

Simplify the logic by ignoring the passed-in value, and manually selecting
ACTION_CREATE for applicable traffic in the CT lookup helpers (same as we
already do for ACTION_CLOSE).

### DIFF
--- a/bpf/lib/conntrack.h
+++ b/bpf/lib/conntrack.h
@@ -47,6 +47,17 @@ struct ct_buffer6 {
 };
 #endif
 
+static __always_inline enum ct_action ct_tcp_select_action(union tcp_flags flags)
+{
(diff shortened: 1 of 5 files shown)


dylandreimerink (approved):
LGTM! :broom:

PR #26534, opened 2023-06-28
bgpv1: set correct upper limits to BPG timers and GR restart time

harsimran-pabla:
BGP holdtime is 16 bits unsigned int and GR restart time is 12 bits.
This limit is set in cilium BGP CRDs to validate configuration at the API layer.


borkmann:
should this be marked for 1.14 backport?

rastislavs:
> should this be marked for 1.14 backport?

I think it makes sense, as this enhances validation of the CRD fields that will be first released in 1.14 - I just marked it.
cc @YutaroHayakawa

squeed (approved):

YutaroHayakawa (approved):

PR #26537, opened 2023-06-28
Add client secret if clustermesh apiserver is enabled

georgew5656:

      GitHub issue.

Fixes: #[issue-number](https://github.com/cilium/cilium-cli/issues/1730)

```release-note
Fix issue with missing client secret when deploying clustermesh through helm.
```

### DIFF
--- a/install/kubernetes/cilium/templates/clustermesh-apiserver/tls-provided/client-secret.yaml
+++ b/install/kubernetes/cilium/templates/clustermesh-apiserver/tls-provided/client-secret.yaml
@@ -1,4 +1,4 @@
-{{- if and .Values.externalWorkloads.enabled (not .Values.clustermesh.apiserver.tls.auto.enabled) }}
+{{- if and (or .Values.externalWorkloads.enabled .Values.clustermesh.useAPIServer) (not .Values.clustermesh.apiserver.tls.auto.enabled) }}
 apiVersion: v1
 kind: Secret
 metadata:


georgew5656:
> @georgew5656, Thanks for the contribution! Your commit subject line is too long (max is 75 chars, yours is 92) for our checkpatch script. Can you amend your commit so that you have a shorter subject line. You can add more details you would like to the commit body.

just pushed an update

gandro (commented):
The change itself looks good to me (though I'd like to get an extra pair of eyes from the clustermesh codeowners to be sure)!

Do we need the same treatment in `tls-helm/client-secret.yaml` too?

giorio94 (changes requested):
Hi @georgew5656,

Thanks for your contribution!

Could you please clarify which is the problem that you are facing? The `clustermesh-apiserver-client-cert` is expected to be used for external workloads support (naming is unfortunate due to historical reasons), and in particular it is only retrieved by the Cilium CLI. For this reason, the current *if* statement appears correct to me, and it is consistent with the same one for the other generation methods for that certificate.

georgew5656:
> Hi @georgew5656,
> 
> Thanks for your contribution!
> 
> Could you please clarify which is the problem that you are facing? The `clustermesh-apiserver-client-cert` is expected to be used for external workloads support (naming is unfortunate due to historical reasons), and in particular it is only retrieved by the Cilium CLI. For this reason, the current _if_ statement appears correct to me, and it is consistent with the same one for the other generation methods for that certificate.

i may be misunderstanding as I was looking at https://github.com/aditighag/cilium-cli/blob/master/clustermesh/clustermesh.go, but does this mean that the CLI won't work for clusters installed with helm (e.g. cillium clustermesh status)

giorio94:
> i may be misunderstanding as I was looking at https://github.com/aditighag/cilium-cli/blob/master/clustermesh/clustermesh.go, but does this mean that the CLI won't work for clusters installed with helm? (e.g. cillium clustermesh status)

Historically, the name of those secrets diverged between helm and the cilium CLI, causing various incompatibilities.

This has been addressed by the new [Cilium CLI helm mode](https://github.com/cilium/cilium-cli/#helm-installation-mode), introduced opt-in in v1.14, and became the default one in v1.15. In practice, CLI operations creating/modifying kubernetes resources are now handled by Helm under the hood, so that it is possible to use the CLI and helm interchangeably. The CLI has also been adapted to recognize the different secret names.

Would you be able to give it a new try using the latest version of the CLI?

georgew5656:
> After i disabled hubble (it wasn't working at the time) i stopped getting the clustermesh-apiserver-client-cert issue and it started complaining about the secret "cilium-ca" not found. i suspect the client-cert issue is also still happening since the issue is not the naming of the secret but instead the absence of the client-cert in a helm installation?

(this is with cilium-cli: v0.15.0 compiled with go1.20.4 on darwin/amd64)

> > i may be misunderstanding as I was looking at https://github.com/aditighag/cilium-cli/blob/master/clustermesh/clustermesh.go, but does this mean that the CLI won't work for clusters installed with helm? (e.g. cillium clustermesh status)
> 
> Historically, the name of those secrets diverged between helm and the cilium CLI, causing various incompatibilities.
> 
> This has been addressed by the new [Cilium CLI helm mode](https://github.com/cilium/cilium-cli/#helm-installation-mode), introduced opt-in in v1.14, and became the default one in v1.15. In practice, CLI operations creating/modifying kubernetes resources are now handled by Helm under the hood, so that it is possible to use the CLI and helm interchangeably. The CLI has also been adapted to recognize the different secret names.
> 
> Would you be able to give it a new try using the latest version of the CLI?

After i disabled hubble (it wasn't working at the time) i stopped getting the clustermesh-apiserver-client-cert issue and it started complaining about the secret "cilium-ca" not found. i suspect the client-cert issue is also still happening since the issue is not the naming of the secret but instead the absence of the client-cert in a helm installation?

(this is with cilium-cli: v0.15.0 compiled with go1.20.4 on darwin/amd64)

georgew5656:
> The change itself looks good to me (though I'd like to get an extra pair of eyes from the clustermesh codeowners to be sure)!
> 
> Do we need the same treatment in `tls-helm/client-secret.yaml` too?

good catch, i will add this if we decide this patch is necessary

giorio94:
> After i disabled hubble (it wasn't working at the time) i stopped getting the clustermesh-apiserver-client-cert issue and it started complaining about the secret "cilium-ca" not found.

I'm guessing you are using Cilium v1.13, please correct me if I'm wrong. This is a separate issue. Essentially, you are configuring the certificates using `.clustermesh.apiserver.tls.(ca|server|admin|remote)`, which is fine. If Hubble is disabled, the `cilium-ca` certificate does not get created, because it is not strictly necessary for the Cilium functioning, but the CLI expects it to be present (as mentioned, helm installations and the CLI clustermesh commands never played well together until Cilium v1.14 and CLI v0.15). The easiest workaround is to configure `.tls.ca` with the same values of `.clustermesh.apiserver.tls.ca` (to uniform them and prevent these issues, we [deprecated](https://docs.cilium.io/en/latest/operations/upgrade/#helm-options) `.clustermesh.apiserver.tls.ca` in v1.14 in favor of `.tls.ca`).

> i suspect the client-cert issue is also still happening since the issue is not the naming of the secret but instead the absence of the client-cert in a helm installation?

The CLI tolerates the absence of the `client-cert`, so that should not be a problem. When I tried locally, the `cilium clustermesh status` command succeeded after ensuring the presence of the `cilium-ca` secret as mentioned above (even though the `client-cert` was not present.

PR #26538, opened 2023-06-28
ci-multi-pool: Use ip-masq-agent for masquerading

gandro:
Now that #26262 has been fixed, we can use disjunct CIDRs for the IP pools. This will make the testing a bit more realistic, as we expect users to define IP pools from different CIDR ranges.

CI run with these changes is green: 
 - :heavy_check_mark:  https://github.com/cilium/cilium/actions/runs/5403357970/jobs/9816045998?pr=26538


viktor-kurchenko (approved):
LGTM.

gandro:
The changed workflow is green, no need in running the rest of CI since it's unaffected. Marked ready to merge.

PR #25217, opened 2023-05-01
docs: HOWTO run cilium-cli e2e connectivity tests

brb:
This PR adds instructions how to run e2e conformance tests locally and on LVH VMs.

A follow-up PR will add info about `ci-e2e` and some tips how to develop those tests.

PR #26540, opened 2023-06-28
Add script to run GitHub ginkgo workflow locally

qmonnet:
Add a script to unroll the steps from the documentation for running locally the ginkgo tests on Kind, inside of a QEMU VM, to replicate the GitHub workflow.

Commits:

- docs: Address minor issues in doc for running ginkgo workflow locally
- contrib/scripts: Add script to run GH ginkgo workflow locally

### DIFF
--- a/contrib/scripts/run-gh-ginkgo-workflow.sh
+++ b/contrib/scripts/run-gh-ginkgo-workflow.sh
@@ -1,0 +1,332 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Authors of Cilium
+
+# See also Documentation/contributing/testing/e2e_legacy.rst for reference
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+ginkgo_matrix_yaml=".github/actions/ginkgo/main-k8s-versions.yaml"
+
+check_cmd() {
+    for cmd in "$@"; do
+        if ! (command -v "$cmd" >/dev/null); then
+            return 1
+        fi
+    done
+    return 0
+}
+
+check_cmd_or_fail() {
+    for cmd in "$@"; do
+        if ! (command -v "$cmd" >/dev/null); then
+            echo "Error: $cmd not found."
+            exit 1
+        fi
+    done
+}
+
+usage() {
+    echo "Usage: $0 [options]"
+    echo "Options:"
+    echo "  -k <version>        [mandatory] Kubernetes version to use"
+    echo "  -l <version>        [optional]  Linux kernel version to use, default: derived from Kubernetes version"
+    echo "  -c <sha>            [optional]  Cilium commit to use to fetch the images, default: current commit SHA"
+    echo "  -f <focus string>   [optional]  Ginkgo focus string to use, default: 'K8s'"
+    echo "  -t <directory>      [optional]  temporary directory to store images, default: '/tmp'"
+    echo "  -w                  [optional]  assume VM image is available, just wake it up"
+    echo "  -s                  [optional]  skip straight to tests (do not set up VM or provision Kind)"
+    echo "  -h                  [optional]  show this help message"
+    exit "$1"
+}
+
+report() {
+    printf '\033[1;34m%s\033[0m\n' "$*"
+}
+
+install_helm() {
+    local HELM_VERSION=3.7.0
+
+    pushd "${tmpdir}" >/dev/null
+    wget "https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz"
+    tar -xf "helm-v${HELM_VERSION}-linux-amd64.tar.gz"
+    mv linux-amd64/helm ./helm
+    popd >/dev/null
+}
+
+install_ginkgo() {
+    go install github.com/onsi/ginkgo/ginkgo@v1.16.5
+}
+
+get_kernel_tag_from_k8s() {
+    local k8s_version="${1}"
+
+    ${yq} ".include[] | select(.k8s-version == \"${k8s_version}\").kernel" "${ginkgo_matrix_yaml}"
+}
+
+get_k8s_image_from_version() {
+    local k8s_version="${1}"
+
+    ${yq} ".include[] | select(.k8s-version == \"${k8s_version}\").kube-image" "${ginkgo_matrix_yaml}"
+}
+
+get_kernel_tag_from_kernel() {
+    local kernel_version="${1}"
+
+    ${yq} "[ .include[] | select(.kernel | test(\"^${kernel_version}-\")).kernel ][0]" "${ginkgo_matrix_yaml}"
+}
+
+build_tests() {
+    pushd "${root_dir}/test" >/dev/null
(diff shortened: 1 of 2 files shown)


qmonnet:
I hear the plan is to move to ACT as soon as we can, to have an easy way to run GitHub Actions locally. So this PR will likely remain a draft until we reach that point.

qmonnet:
Now I'm also hearing that some people would like to have it in, so let's go for a round of reviews!

Documentation/contributing/testing/e2e.rst
```diff
@@ -62,6 +62,10 @@ The following example will provide the steps to run one of the tests of the
 focus ``f09-datapath-misc-2`` on Kubernetes ``1.27`` with the kernel ``net-next``
 for the commit SHA ``7b368923823e63c9824ea2b5ee4dc026bc4d5cd8``.
 
+You can also perform these steps automatically using the script
```
bimmlerd: While we're here: I think these instructions are Linux (and x86_64) only (at least the script is), but I don't see that called out explicitly.

contrib/scripts/run-gh-ginkgo-workflow.sh
```diff
@@ -0,0 +1,335 @@
...
+    helm="${tmpdir}/helm"
+fi
+
+if ! check_cmd ginkgo; then
+    install_ginkgo
+fi
+
+yq="yq"
+if ! check_cmd "${yq}"; then
+    yq="docker run --rm -v ${PWD}:/workdir --user $(id -u):$(id -g) mikefarah/yq:4.27.3"
```
bimmlerd: do we need a renovate reference to auto-update this image?

contrib/scripts/run-gh-ginkgo-workflow.sh
```diff
@@ -0,0 +1,335 @@
...
+    echo "  -h                  [optional]  show this help message"
+    exit "$1"
+}
+
+report() {
+    printf '\033[1;34m%s\033[0m\n' "$*"
+}
+
+install_helm() {
+    local HELM_VERSION=3.7.0
```
bimmlerd, line 50: do we need a renovate reference to update this version automatically?

contrib/scripts/run-gh-ginkgo-workflow.sh
```diff
@@ -0,0 +1,335 @@
...
+
+    pushd "${tmpdir}" >/dev/null
+    wget "https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz"
+    tar -xf "helm-v${HELM_VERSION}-linux-amd64.tar.gz"
+    mv linux-amd64/helm ./helm
+    popd >/dev/null
+}
+
+install_ginkgo() {
+    go install github.com/onsi/ginkgo/ginkgo@v1.16.5
```
bimmlerd, line 60: renovate reference to auto-update?

bimmlerd (approved):
couple of nits, but otherwise lgtm - if we can get `act` to do most of the heavy lifting for us that'll be more maintanable, but for now I think this is cool

Can't test it, however, because it assumes x86 :innocent:

Documentation/contributing/testing/e2e.rst
qmonnet: I'm happy to add an option for another arch, but _I_ can't test it on aarch64. Would you mind giving it a try, by replacing `x86`/`x86_64` by whatever values are needed, and letting me know what these values are, please? :innocent:

contrib/scripts/run-gh-ginkgo-workflow.sh
qmonnet: Probably not this one, as long as the few basic `yq` invocations we do in the script work fine, it probably doesn't matter what version we're using.

contrib/scripts/run-gh-ginkgo-workflow.sh
qmonnet: I think we use a fixed version for Ginkgo at the moment, updates seem disabled in renovate's config file because nobody has cycles to port things go Ginkgo v2+

contrib/scripts/run-gh-ginkgo-workflow.sh
qmonnet: Looks like updates are manual at the moment? https://github.com/cilium/cilium/pull/17488

contrib/scripts/run-gh-ginkgo-workflow.sh
bimmlerd: :grin: I ran into this while testing - probably best not to use local `yq` at all, since there's competing incompatible implementations apparently: 
- https://github.com/kislyuk/yq
- https://github.com/mikefarah/yq

Documentation/contributing/testing/e2e.rst
bimmlerd: Gave it a quick shot; I don't think it's easily doable- the docker images we reference from the github matrix don't seem to be multi-arch, so won't run on non-amd64.

contrib/scripts/run-gh-ginkgo-workflow.sh
qmonnet: Oh no :facepalm:
OK I'll update to just use the Docker container.

qmonnet:
I added `docker` and `qemu-system-x86_64` to the list of commands we check at the beginning of the script, so it will return early with an error message if these commands are not available.

I also remove the check on `yq` and only use the Docker image, to make sure we run the right tool (see David's comment on conflicting implementations).

Documentation/contributing/testing/e2e.rst
qmonnet: OK, so let's keep it x86-only for now. Thanks a lot for checking!
I added `qemu-system-x86_64` to the list of commands we check at the beginning, so we'll fail early if the command is not available.

aanm (approved):

qmonnet:
@learnitall, would you mind taking a quick look please?

Documentation/contributing/testing/e2e.rst
```diff
@@ -62,6 +62,10 @@ The following example will provide the steps to run one of the tests of the
...
+``contrib/scripts/run-gh-ginkgo-workflow.sh`` (run this script with ``-h`` for
+indications).
```
learnitall: It might make sense to move this instruction outside of the parentheses. Also, I don't think 'indications' would translate well here in this context, however, I could be wrong.

learnitall (approved):
One minor nit, but looks good! I apologize for the delay.

Documentation/contributing/testing/e2e.rst
qmonnet: OK, I'll fix this. Thanks a lot!

qmonnet:
I addressed the comments from Ryan, and rebased on top of https://github.com/cilium/cilium/pull/25217.

qmonnet:
Doc-related workflows passed successfully, and the new script is not tested in CI. This is ready to merge.

PR #26542, opened 2023-06-28
ci: Fix container scanning workflow

ferozsalam:
Using `act` to locally run the workflow didn't find a syntax error, which means that the workflow doesn't run correctly in GitHub Actions.

I fixed this issue, which brought to light some other minor issues with images not being available on specific branches, as well as one unnecessary step.

I've fixed all of the above and added v1.14 as another branch to scan.

```release-note
Fix container scanning workflow
```


brlbil (approved):

ferozsalam:
thanks @brlbil – didn't know that option was available!

PR #26547, opened 2023-06-28
hubble: Optimize namespace tracking

glibsm:
Avoiding the use of varargs saves some allocations

Before
```
Benchmark_TrackNamespaces-12    	 5122684	       231.8 ns/op	     192 B/op	       6 allocs/op
```

After
```
Benchmark_TrackNamespaces-12    	 6114062	       187.9 ns/op	     168 B/op	       4 allocs/op
```

### DIFF
--- a/pkg/hubble/observer/local_observer.go
+++ b/pkg/hubble/observer/local_observer.go
@@ -623,20 +623,18 @@ func (r *eventsReader) Next(ctx context.Context) (*v1.Event, error) {
 
 func (s *LocalObserverServer) trackNamespaces(flow *flowpb.Flow) {
 	// track namespaces seen.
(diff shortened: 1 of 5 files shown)


kaworu (changes requested):
Thanks for the patch @glibsm!

`pkg/hubble/relay/observer/server.go` needs to be adapted accordingly.

Note that after the patch, adding a flow's src/dst namespaces is no longer "atomic" from the `namespaceManager` point of view but I don't think that's an issue.

chancez:
We should back port this to 1.14 since this functionality is new in 1.14 and that memory increase could be viewed as a performance regression.

chancez (approved):

kaworu (approved):

PR #26550, opened 2023-06-29
bpf: nodeport: provide L4 ports for SNAT in LB egress path

julianwiedmann:
For LB traffic that gets forwarded to a remote backend in non-DSR mode, tail_nodeport_nat_egress_ipv*() calls snat_v*_nat() to perform SNAT on the packet. Under the covers, this extracts a fresh CT tuple to look up / build a SNAT entry.

But for LB traffic we don't require any of the ICMP handling in that code path, and we already extract a CT tuple for building tunnel headers in XDP mode. So we can optimize this code path, and provide a fully populated CT tuple to the SNAT helper.

One additional benefit is that we fix handling for fragmented IPv4 packets, as lb4_extract_tuple() knows how to extract their L4 ports while snat_v4_nat() doesn't.

```release-note
Fix SNAT by the N/S load-balancer for fragmented IPv4 requests.
```

### DIFF
--- a/bpf/lib/nat.h
+++ b/bpf/lib/nat.h
@@ -958,12 +958,37 @@ snat_v4_nat_handle_icmp_frag_needed(struct __ctx_buff *ctx, __u64 off,
 	return snat_v4_rewrite_egress(ctx, &tuple, state, off, has_l4_header);
 }
 
+static __always_inline int
+__snat_v4_nat(struct __ctx_buff *ctx, struct ipv4_ct_tuple *tuple,
+	      bool has_l4_header, int l4_off, enum ct_action action,
+	      bool update_tuple,
+	      const struct ipv4_nat_target *target, __s8 *ext_err)
+{
+	struct ipv4_nat_entry *state, tmp;
+	int ret;
+
+	ret = snat_v4_nat_handle_mapping(ctx, tuple, has_l4_header, action,
+					 &state, &tmp, l4_off, target, ext_err);
+	if (ret > 0)
+		return CTX_ACT_OK;
+	if (ret < 0)
+		return ret;
+
+	ret = snat_v4_rewrite_egress(ctx, tuple, state, l4_off, has_l4_header);
+
+	if (update_tuple) {
+		tuple->saddr = state->to_saddr;
+		tuple->sport = state->to_sport;
+	}
+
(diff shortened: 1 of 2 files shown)


gentoo-root (commented):
Looks good to me, but I suspect it may suffer from the same fragment double accounting issue as my #25340, because you also add a call of ipv4_handle_fragmentation. However, I stopped understanding how ipv4_handle_fragmentation leads to ct_lookup4. I'll take one more look tomorrow, maybe something changed in its implementation.

julianwiedmann:
> Looks good to me, but I suspect it may suffer from the same fragment double accounting issue as my #25340, because you also add a call of ipv4_handle_fragmentation. However, I stopped understanding how ipv4_handle_fragmentation leads to ct_lookup4. I'll take one more look tomorrow, maybe something changed in its implementation.

Ah, is that blocking you in #25340 ? It's most likely https://github.com/cilium/cilium/blob/28de75d593eb87f75fa66d2fc31f8cfaf2877567/bpf/lib/ipv4.h#L148 then. That part is a bit silly, I wouldn't expect it to produce accurate metrics at the moment ... we can hardly expect that every packet needs just a single L4 port lookup.

It's probably fine to just push this metrics update into `__ct_lookup4()` instead - there we should have more confidence that a given CT tuple only gets looked up (== accounted) once. Let's try that in a follow-on PR?

julianwiedmann:
> > It's probably fine to just push this metrics update into __ct_lookup4() instead - there we should have more confidence that a given CT tuple only gets looked up (== accounted) once.
> 
> I believe there is a scenario with host firewall where ct_lookup4 is called more than once.

Yep, I would fully expect that we still have cases where that's the case. But those we should fix over time :)

gentoo-root (approved):

julianwiedmann:
> > But those we should fix over time :)
> 
> OK, if I don't manage to reproduce it with host firewall, I'll try to move update_metrics in my pull request to see if it unblocks the failing test. Thanks for the suggestion! I was looking to fix it once and for all, but probably you are right.

Just FYI, I suspect that this might even produce slightly better metrics (more accurate `direction` variable, basically). I didn't have time to follow-up on https://github.com/cilium/cilium/pull/25880 yet.

PR #26553, opened 2023-06-29
bpf: Replace deprecated "-target bpf" with "--target=bpf" for clang

qmonnet:
Passing the target to clang with `-target <name>` has been [considered deprecated since clang 3.4][0], released in 2013, in favour of the more recent syntax `--target=<name>`. Let's update our code and documentation to use the recommended syntax.

This replicates [a similar patch submitted to the Linux kernel][1].

[0]: https://github.com/llvm/llvm-project/commit/274b6f0c87a6a1798de0a68135afc7f95def6277
[1]: https://lore.kernel.org/all/20230624001856.1903733-1-maskray@google.com/

### DIFF
--- a/Documentation/bpf/debug_and_test.rst
+++ b/Documentation/bpf/debug_and_test.rst
@@ -360,7 +360,7 @@ by ``pahole``:
 
 .. code-block:: shell-session
 
-     # clang [...] -O2 -target bpf -g -emit-llvm -c test_xdp_noinline.c -o - |
+     # clang [...] -O2 --target=bpf -g -emit-llvm -c test_xdp_noinline.c -o - |
        llc -march=bpf -mcpu=probe -mattr=dwarfris -filetype=obj -o test_xdp_noinline.o
      # pahole -J test_xdp_noinline.o
 
--- a/Documentation/bpf/toolchain.rst
+++ b/Documentation/bpf/toolchain.rst
@@ -319,7 +319,7 @@ It can then be compiled and loaded into the kernel as follows:
 
 .. code-block:: shell-session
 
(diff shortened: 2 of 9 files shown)


lmb (approved):

lmb:
FWIW, clang command line args seem like a mess. They mix double dashes, single dashes, space separated, equals separated. 

Another point to think about: what kind of flags does gcc accept? Might make it easier to experiment with bpf-gcc in the future if both compilers agree.

borkmann (approved):

qmonnet:
> FWIW, clang command line args seem like a mess. They mix double dashes, single dashes, space separated, equals separated.

@lmb Yep, I noticed :). I don't mind much between the two syntaxes to be honest, I simply thought it would be best to remain in sync with whatever the kernel goes with.

> Another point to think about: what kind of flags does gcc accept? Might make it easier to experiment with bpf-gcc in the future if both compilers agree.

That's a good question. So apparently there's no direct equivalent to `--target=bpf` for GCC, instead you compile your BPF-dedicated version of GCC and call that. See for example the flags used on godbolt.org:

image

To compare with those used with clang:

image

(There _is_ a `--target=bpf` flag mentioned on [the reference page for the BPF backend](https://gcc.gnu.org/wiki/BPFBackEnd), but it's used for compiling the BPF version of GCC, not to pick BPF with a generic GCC).

GCC does have a number of eBPF-related options though, none being identical to clang's, so the command line invocations won't be interchangeable anyway.

image

Thanks for the reviews!

gentoo-root (approved):

learnitall (approved):
Looks good!

ysksuzuki (approved):

PR #26557, opened 2023-06-29
helm: Updated description for Helm 'devices' flag

PhilipSchmid:
* Added hint regarding Cilium's autodetection of interfaces with non-local route
* Added recommendation to not use `devices` unless autodetection really doesn't work.

FYI: The `devices` flag isn't listed in `Documentation/helm-values.rst` and `install/kubernetes/cilium/README.md`. Should that be like this? I therefore also didn't update these two files.

Setting this flag to an interface that doesn't actually do the node's routing can cause severe connectivity issues up to full unreachability of the whole node as long as Cilium runs.

```release-note
helm: Updated description for Helm 'devices' flag
```

### DIFF
--- a/install/kubernetes/cilium/values.yaml
+++ b/install/kubernetes/cilium/values.yaml
@@ -629,7 +629,8 @@ daemon:
 # -- Specify which network interfaces can run the eBPF datapath. This means
 # that a packet sent from a pod to a destination outside the cluster will be
 # masqueraded (to an output device IPv4 address), if the output device runs the
-# program. When not specified, probing will automatically detect devices.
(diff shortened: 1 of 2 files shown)


gandro (approved):
Thanks!

nathanjsweet (approved):

gandro:
ci-e2e failed with `  ❌ no-policies/pod-to-world/https-to-one.one.one.one-1: cilium-test/client2-5b97d7bc66-9jch5 (10.244.2.127) -> one.one.one.one-https (one.one.one.one:443)` which is a known flake. Restarting failed test

PR #26565, opened 2023-06-30
make: drop redundant `go vet ./...` from integration tests

tklauser:
The Go toolchain will already run vet during `go test` since Go 1.10, see https://go.dev/doc/go1.10#test and
https://github.com/golang/go/issues/18084 for details.

This reduces total run time of `make integration-tests` on my machine from ~16min to ~18min.


markpash (approved):

qmonnet (approved):
> This reduces total run time of make integration-tests on my machine from ~16min to ~18min.

:scream_cat: Doesn't seem worth it then :stuck_out_tongue: 
Thanks!

PR #26582, opened 2023-06-30
docs: update KPR section on DSR

### LINKED ISSUE #26339: The DSR section in kubeproxy-free.rst may need to be updated
Under the https://github.com/cilium/cilium/blob/main/Documentation/network/kubernetes/kubeproxy-free.rst#direct-server-return-dsr section, the doc claims that "DSR currently requires Cilium to be deployed in [:ref:`arch_direct_routing`](https://github.com/cilium/cilium/blob/main/Documentation/network/kubernetes/kubeproxy-free.rst#id7), i.e. it will not work in either tunneling mode."

However, [the section after it](https://github.com/cilium/cilium/blob/main/Documentation/network/kubernetes/kubeproxy-free.rst#direct-server-return-dsr-with-geneve) claims that the DSR mode can work with tunnel protocol geneve.

Look like we need to update the former one?

julianwiedmann:
Rephrase the DSR section to clarify the differences between the two dispatch mechanisms (IP option/extension, or Geneve option).

Fixes: https://github.com/cilium/cilium/issues/26339

### DIFF
--- a/Documentation/network/kubernetes/kubeproxy-free.rst
+++ b/Documentation/network/kubernetes/kubeproxy-free.rst
@@ -474,32 +474,39 @@ This setting can be changed through the ``loadBalancer.mode`` Helm option to
 ``dsr`` in order to let Cilium's eBPF NodePort implementation operate in DSR mode.
 In this mode, the backends reply directly to the external client without taking
 the extra hop, meaning, backends reply by using the service IP/port as a source.
-DSR currently requires Cilium to be deployed in :ref:`arch_direct_routing`, i.e.
-it will not work in either tunneling mode.
 
 Another advantage in DSR mode is that the client's source IP is preserved, so policy
 can match on it at the backend node. In the SNAT mode this is not possible.
 Given a specific backend can be used by multiple services, the backends need to be
-made aware of the service IP/port which they need to reply with. Therefore, Cilium
-encodes this information in a Cilium-specific IPv4 option or IPv6 Destination Option
-extension header at the cost of advertising a lower MTU. For TCP services, Cilium
-only encodes the service IP/port for the SYN packet, but not subsequent ones. The
-latter also allows to operate Cilium in a hybrid mode as detailed in the later subsection
-where DSR is used for TCP and SNAT for UDP in order to avoid an otherwise needed MTU
-reduction.
+made aware of the service IP/port which they need to reply with. Cilium encodes this
+information into the packet (using one of the dispatch mechanisms described below),
+at the cost of advertising a lower MTU. For TCP services, Cilium
+only encodes the service IP/port for the SYN packet, but not subsequent ones. This
+optimization also allows to operate Cilium in a hybrid mode as detailed in the later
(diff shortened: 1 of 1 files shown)


Documentation/network/kubernetes/kubeproxy-free.rst
```diff
@@ -474,32 +474,40 @@ This setting can be changed through the ``loadBalancer.mode`` Helm option to
...
+made aware of the service IP/port which they need to reply with. Cilium encodes this
+information into the packet (using one of the dispatch mechanisms described below),
+at the cost of advertising a lower MTU. For TCP services, Cilium
+only encodes the service IP/port for the SYN packet, but not subsequent ones. This
+optimization also allows to operate Cilium in a hybrid mode as detailed in the later
+subsection where DSR is used for TCP and SNAT for UDP in order to avoid an otherwise
+needed MTU reduction.
+
+In some public cloud provider environments that implement source /
+destination IP address checking (e.g. AWS), the checking has to be disabled in
```
ysksuzuki, line 490: > source / destination IP address checking (e.g. AWS),

Is this about rp_filter or some other public cloud specific feature?

julianwiedmann: Yep, see eg. `Source/destination checking` in https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html.

ysksuzuki: I see. Thanks for sharing that!

ysksuzuki (approved):

Documentation/network/kubernetes/kubeproxy-free.rst
```diff
@@ -474,32 +474,40 @@ This setting can be changed through the ``loadBalancer.mode`` Helm option to
...
 first check whether the NodePort request actually arrived on the node
 containing the backend. If this was not the case, then consider either switching to
-DSR with Geneve which will be described in the later sections, or switching back to
-the default SNAT mode would be advised as a workaround.
-
-Also, in some public cloud provider environments, which implement a source /
-destination IP address checking (e.g. AWS), the checking has to be disabled in
-order for the DSR mode to work.
+DSR with Geneve (as described below), or switching back to the default SNAT mode would be
+advised as a workaround.
```
qmonnet: “consider either [...] or switching back to the default SNAT mode would be advised as a workaround” -> Doesn't read well, we want either “consider doing it” or “this is advised as a workaround”, not both. Can you please fix it since you're touching this part?

```suggestion
DSR with Geneve (as described below), or switching back to the default SNAT mode.
```

julianwiedmann: Done, ty!

qmonnet (approved):
Thanks!

qmonnet:
Doc change only, related workflow passed successfully, and Netlify preview builds OK. This is ready to merge.

spacewander (approved):

PR #26583, opened 2023-06-30
pkg/bpf: fix bug where second value can be skipped.

### LINKED ISSUE #26491: CI: Runtime Test (privileged): BPFPrivilegedTestSuite/TestDumpReliablyWithCallback - timed out waiting to send sub map event
## CI failure

```
level=info msg="Detected mounted BPF filesystem at /sys/fs/bpf" subsys=bpf
level=warning msg="Flags mismatch for BPF map" file-path=/sys/fs/bpf/tc/globals/cilium_test_upgrade new=0 old=1 subsys=bpf
level=warning msg="Removing map to allow for property upgrade (expect map data loss)" file-path=/sys/fs/bpf/tc/globals/cilium_test_upgrade subsys=bpf
level=info msg="Unpinning map with incompatible properties" file-path=/sys/fs/bpf/tc/globals/cilium_test new="Type:Hash KeySize:4 ValueSize:4 MaxEntries:16 Flags:0" old="Type:Hash KeySize:4 ValueSize:4 MaxEntries:16 Flags:1" subsys=bpf
level=info msg="Removed map pin at /sys/fs/bpf/tc/globals/cilium_test, recreating and re-pinning map cilium_test" file-path=/sys/fs/bpf/tc/globals/cilium_test name=cilium_test subsys=bpf
--- FAIL: Test (0.25s)
    --- FAIL: Test/BPFPrivilegedTestSuite (0.25s)
        --- FAIL: Test/BPFPrivilegedTestSuite/TestDumpReliablyWithCallback (0.24s)
level=warning msg="subscription channel buffer 0 was full, closing subscription" error="timed out waiting to send sub map event" subsys=bpf
FAIL
	github.com/cilium/cilium/pkg/bpf	coverage: 57.7% of statements
FAIL	github.com/cilium/cilium/pkg/bpf	0.285s
```

- [Link](https://github.com/cilium/cilium/actions/runs/5383112601/jobs/9769533777?pr=26435)
- [PR](https://github.com/cilium/cilium/pull/26435) (for `main` branch)
- No artifacts

Cc @tommyp1ckles regarding the code for subscribing to `bpf.Map` events

tommyp1ckles:
DumpReliablyWithCallback will skip a value callback in some situations.

In cases where the initial key is deleted just after being retrieved, there is no previous key to fallback on. The reliable dump will attempt to use the current nextKey (that was based on the deleted current key).

The local currentKey and nextKey Key types are being passed to NextKey which eventually writes the nextKey output to the nextKeys pointer location (via the bpf syscall).

The currentValue was simply being assinged by the equals operator, which was copying the underlying interface pointer.

Thus in this situation, the next iteration attempt was passing the same pointer twice to NextKey causing the currentKey to be set to the next key a second time - skipping a map element.

Fixes #26491

## Validating the fix

Because the test has [256 possible keys](https://github.com/cilium/cilium/blob/bd9c7cdcc8fe10ee6eb225992dee985b213a9367/pkg/bpf/map_linux_test.go#L505) the particular start condition for this failure seems to be quite rare in CI, by reducing this to 5 it happens much more often (presumably 20% of the time?) - making this much easier to reproduce.

I tested this change by continuously running the DumpReliablyWithCallback (modified with maxSize=5) test to reproduce, and after the fix to validate (ran it 1000 times post fix without failure).

```release-note
Fix bug where bpf map entries may not be reliably dumped or garbage collected when the map is actively being updated.
```

### DIFF
--- a/pkg/bpf/map_linux.go
+++ b/pkg/bpf/map_linux.go
@@ -672,6 +672,11 @@ func (m *Map) DumpReliablyWithCallback(cb DumpCallback, stats *DumpStats) error 
 				// map, nextKey may be the actual key element after the deleted
 				// one, or the first element in the map.
 				currentKey = nextKey
+				// To avoid having nextKey and currentKey pointing at the same memory
+				// we allocate a new key for nextKey. Without this currentKey and nextKey
+				// would be the same pointer value and would get double iterated on the next
+				// iterations m.NextKey(...) call.
+				nextKey = m.key.New()
 				stats.Interrupted++
 			}
 			continue
--- a/pkg/bpf/map_linux_test.go
+++ b/pkg/bpf/map_linux_test.go
@@ -4,6 +4,7 @@
 package bpf
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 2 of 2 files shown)


joestringer:
Minor nit: This is marked as a `bug` but "second value can be skipped" isn't very user-friendly. Perhaps "Fix bug where bpf map entries may not be reliably dumped or garbage collected, particularly affecting connection tracking and load balancing maps" would demonstrate the impact more clearly.

tommyp1ckles:
@joestringer Yeah makes sense - I'll make the user impact clearer.

pkg/bpf/map_linux.go
```diff
@@ -671,7 +671,11 @@ func (m *Map) DumpReliablyWithCallback(cb DumpCallback, stats *DumpStats) error
 				// Depending on exactly when currentKey was deleted from the
 				// map, nextKey may be the actual key element after the deleted
 				// one, or the first element in the map.
-				currentKey = nextKey
+
+				// We can't simply assign currentKey to equal nextKey because
+				// that will result in currentKey and nextKey pointing to the
+				// same memory (which causes unexpected behavior when calling ebpf.NextKey).
+				m.NextKey(currentKey, currentKey)
```
rgo3: What's the behavior of `m.NextKey` if you call it with a key that got deleted? Will it set `currentKey` to the first key in the map? And if so couldn't we just do `m.NextKey(nil, currentKey)`? I think the comment could explain this a bit more, in addition to explaining why `currentKey = nextKey` isn't working.

And shouldn't we check the returned error?

tommyp1ckles: Looking at this again, I think I was overthinking this.  I have a simpler solution that avoids the extra NetKey call now.

pkg/bpf/map_linux_test.go
```diff
@@ -501,6 +502,97 @@ func (s *BPFPrivilegedTestSuite) TestDump(c *C) {
...
+				c.Check(err, IsNil)
+			}
+		}
+	}()
+
+	// We expect that DumpReliablyWithCallback will iterate all odd key/value pairs
+	// even if the even keys are being deleted and reinserted.
+	expect := map[string]string{}
+	for i := uint32(0); i < maxEntries; i++ {
+		if i != 0 && i%2 != 0 {
```
rgo3: Nit, 0%2 == 0:
```suggestion
		if i%2 != 0 {
```

pkg/bpf/map_linux_test.go
```diff
@@ -501,6 +502,97 @@ func (s *BPFPrivilegedTestSuite) TestDump(c *C) {
...
+			expect[fmt.Sprintf("key=%d", i)] = fmt.Sprintf("value=%d", i+200)
+		}
+	}
+	close(start) // start testing.
+	for i := 0; i < iterations; i++ {
+		dump := map[string]string{}
+		ds := NewDumpStats(m)
+		err := m.DumpReliablyWithCallback(func(key MapKey, value MapValue) {
+			k := key.(*TestKey).Key
+			if k != 0 && k%2 != 0 {
```
rgo3:
```suggestion
			if k%2 != 0 {
```

pkg/bpf/map_linux_test.go
```diff
@@ -501,6 +502,97 @@ func (s *BPFPrivilegedTestSuite) TestDump(c *C) {
...
+		defer wg.Done()
+		<-start
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
+
+			for i := uint32(2); i < maxEntries; i += 2 {
```
rgo3: Why aren't we starting at 0? Even if the first key is deleted/reinserted it should work right?

pkg/bpf/map_linux_test.go
```diff
@@ -501,6 +502,97 @@ func (s *BPFPrivilegedTestSuite) TestDump(c *C) {
...
+			if k != 0 && k%2 != 0 {
+				k := key.(*TestKey).Key
+				ks := dump[fmt.Sprintf("key=%d", k)]
+				if _, ok := dump[ks]; ok {
+					c.FailNow()
+				}
+				dump[fmt.Sprintf("key=%d", key.(*TestKey).Key)] = fmt.Sprintf("value=%d", value.(*TestValue).Value)
+			}
+		}, ds)
+		if err == nil {
```
rgo3, line 584: Should the test fail if `err != nil`?

rgo3 (changes requested):
Just a couple of nits/comments, otherwise LGTM.

pkg/bpf/map_linux_test.go
tommyp1ckles: correct - i'll change it to 0.

pkg/bpf/map_linux_test.go
tommyp1ckles: No, the reliable dump has a max number of iterations it can fall back to previous keys before failing, however as with the other test I'll add the assertion:
```go
c.Check(err, Equals, ErrMaxLookup)
```

If an error is returned - which is the correct behavior.

tommyp1ckles:
Thanks for the review @rgo3 😄, ptal

rgo3 (approved):

Issue #26492, opened 2023-06-26
CI: kubernetes-e2e-net-conformance (ipv4) - NetworkPolicyLegacy [LinuxOnly] NetworkPolicy between server and client


qmonnet:
## CI failure

- [Link](https://github.com/cilium/cilium/actions/runs/5383112591/jobs/9769451566?pr=26435)
- [PR](https://github.com/cilium/cilium/pull/26435) (for `main` branch)
- No artifacts

```
 • [FAILED] [567.806 seconds]
[sig-network] NetworkPolicyLegacy [LinuxOnly] NetworkPolicy between server and client [It] should enforce policy based on PodSelector with MatchExpressions[Feature:NetworkPolicy]
test/e2e/network/netpol/network_legacy.go:270

[...]

  Jun 26 21:28:53.710: INFO: Running '/usr/local/bin/kubectl --server=https://127.0.0.1:33299/ --kubeconfig=/home/runner/work/cilium/cilium/_artifacts/kubeconfig.conf --namespace=network-policy-9493 logs client-a-t8xw7 --tail=100'
  Jun 26 21:28:54.425: INFO: stderr: ""
  Jun 26 21:28:54.425: INFO: stdout: "OTHER: dial tcp 10.96.174.86:80: connect: operation not permitted\nOTHER: dial tcp 10.96.174.86:80: connect: operation not permitted\nOTHER: dial tcp 10.96.174.86:80: connect: operation not permitted\nOTHER: dial tcp 10.96.174.86:80: connect: operation not permitted\nOTHER: dial tcp 10.96.174.86:80: connect: operation not permitted\n"
  Jun 26 21:28:54.425: INFO:
  Last 100 log lines of client-a-t8xw7:
  OTHER: dial tcp 10.96.174.86:80: connect: operation not permitted
  OTHER: dial tcp 10.96.174.86:80: connect: operation not permitted
  OTHER: dial tcp 10.96.174.86:80: connect: operation not permitted
  OTHER: dial tcp 10.96.174.86:80: connect: operation not permitted
  OTHER: dial tcp 10.96.174.86:80: connect: operation not permitted

[...]

  [FAILED] Pod client-a-t8xw7 should be able to connect to service svc-server, but was not able to connect.
  Pod logs:
  OTHER: dial tcp 10.96.174.86:80: connect: operation not permitted
  OTHER: dial tcp 10.96.174.86:80: connect: operation not permitted
[...]

PR #26590, opened 2023-07-03
bgpv1: Unconditionally select node when empty nodeSelector is given

YutaroHayakawa:
The CRD spec of CiliumBGPPeeringPolicy states

```
NodeSelector selects a group of nodes where this BGP Peering Policy
applies. If empty / nil this policy applies to all nodes.
```

However, our current implementation treats empty / nil selector as an error. This commit fixes it.

This issue was reported by an OSS community user on Slack.

スクリーンショット 2023-07-03 15 57 33

```release-note
bgpv1: Unconditionally select node when empty nodeSelector is given
```

### DIFF
--- a/pkg/bgpv1/agent/controller.go
+++ b/pkg/bgpv1/agent/controller.go
@@ -278,7 +278,8 @@ func (c *Controller) Run(ctx context.Context) {
 //
 // Policy selection follows the following rules:
 //   - A policy matches a node if said policy's "nodeSelector" field matches
-//     the node's labels
+//     the node's labels. If "nodeSelector" is omitted, it is unconditionally
+//     selected.
 //   - If (N > 1) policies match the provided *corev1.Node an error is returned.
 //     only a single policy may apply to a node to avoid ambiguity at this stage
 //     of development.
@@ -287,11 +288,10 @@ func PolicySelection(ctx context.Context, labels map[string]string, policies []*
 		l = log.WithFields(logrus.Fields{
 			"component": "PolicySelection",
 		})
-	)
-	// determine which policies match our node's labels.
-	var (
-		selected   *v2alpha1api.CiliumBGPPeeringPolicy
-		slimLabels = slimlabels.Set(labels)
+
+		// determine which policies match our node's labels.
+		selectedPolicy *v2alpha1api.CiliumBGPPeeringPolicy
+		slimLabels     = slimlabels.Set(labels)
 	)
 
 	// range over policies and see if any match this node's labels.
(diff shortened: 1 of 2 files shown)


YutaroHayakawa:
K8sUpstreamNetConformance: https://github.com/cilium/cilium/issues/26492
ClusterMesh: https://github.com/cilium/cilium/issues/25816
ConformanceAKS: https://github.com/cilium/cilium/issues/26075

pkg/bgpv1/agent/controller.go
```diff
@@ -300,26 +300,38 @@ func PolicySelection(ctx context.Context, labels map[string]string, policies []*
...
 		l.WithFields(logrus.Fields{
-			"policyNodeSelector": nodeSelector.String(),
+			"policyName":         policy.Name,
 			"nodeLabels":         slimLabels,
+			"policyNodeSelector": policy.Spec.NodeSelector.String(),
 		}).Debug("Comparing BGP policy node selector with node's labels")
-		if nodeSelector.Matches(slimLabels) {
-			if selected != nil {
+
+		if policy.Spec.NodeSelector == nil {
```
ldelossa, line 311: This logic becomes a bit windy. 

Can we simply early return here, and not do the if/else later? I think this is okay, right? Since if the policy has no selector, we know we are going to select it and return it. 

```suggestion
                if policy.Spec.NodeSelector == nil {
                    return policy, nil
                }
```
Then you can get rid of the else block, and do the NodeSelector processing like normal.

ldelossa (changes requested):

YutaroHayakawa:
> So in this change, what happens if multiple 'CiliumBGPPeeringPolicies' do not specify a node selector?
> We will pick the latest created 'CiliumBGPPeeringPolicy' and ignore all others, correct?

No, essentially the empty/nil selector is a selector that selects everything. So, when there are multiple policies select the same node, BGP sessions will be reset. This is the same behavior as before.

pkg/bgpv1/agent/controller.go
YutaroHayakawa: It is inconsistent with an existing behavior I think. An existing behavior is when multiple policies select the same node, `PolicySelection` returns an error and the controller resets the existing BGP session. Returning earlier here breaks this behavior.

ldelossa (approved):

pkg/bgpv1/agent/controller.go
ldelossa: Yup, I see, mis understood the initial intent here :)

PR #26592, opened 2023-07-03
auth: use NodeManager instead of k8s.CiliumNodeResource in auth gc

### LINKED ISSUE #25899: Auth: Support node based garbage collection in combination with ClusterMesh
Currenty, Auth Map garbage collection builds on top of K8s Resource of type CiliumIdentity to delete auth map entries whenever a K8s Node gets deleted. This only works if the Node belongs to the same Cluster and doesn't support ClusterMesh.

This issue should track this limitation and implement support for ClusterMesh (e.g. by using the observable `NodeManager` directly).

mhofstetter:
Auth map garbage collection is using the k8s.CiliumNodeResource to get notified about added and deleted nodes.

This comes with the drawback of not supporting ClusterMesh and introducing an additional k8s watcher for CiliumNodes (as long as the NodeManager is using the legacy k8s watcher).

Therefore, this commit is refactoring the auth map garbage collection to register itself as NodeHandler to the NodeManager to get notified about node events.

Fixes: #25899

### DIFF
--- a/pkg/auth/authmap_gc.go
+++ b/pkg/auth/authmap_gc.go
@@ -6,18 +6,18 @@ package auth
 import (
 	"context"
 	"fmt"
-	"net"
 	"time"
 
 	"github.com/sirupsen/logrus"
 
+	datapathTypes "github.com/cilium/cilium/pkg/datapath/types"
 	"github.com/cilium/cilium/pkg/identity"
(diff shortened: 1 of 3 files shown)


meyskens (approved):
LGTM

Even looks cleaner :)

PR #26594, opened 2023-07-03
bpf: nat: improve logic that creates the NAT entries

julianwiedmann:
The logic in `snat_v*_new_mapping()` is super tricky, and I start afresh every time I need to look at it. Re-organize the code, and add comments to the crucial pieces.

### DIFF
--- a/bpf/lib/nat.h
+++ b/bpf/lib/nat.h
@@ -45,13 +45,13 @@ struct nat_entry {
 # define SNAT_SIGNAL_THRES		16
 #endif
 
-static __always_inline __be16 __snat_clamp_port_range(__u16 start, __u16 end,
-						      __u16 val)
+static __always_inline __u16 __snat_clamp_port_range(__u16 start, __u16 end,
+						     __u16 val)
 {
 	return (val % (__u16)(end - start)) + start;
 }
 
-static __always_inline __maybe_unused __be16
+static __always_inline __maybe_unused __u16
 __snat_try_keep_port(__u16 start, __u16 end, __u16 val)
 {
 	return val >= start && val <= end ? val :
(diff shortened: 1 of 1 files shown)


julianwiedmann:
Looks like this also nicely reduces complexity in the relevant code paths (from pulling the NAT entry creation out of the loop?).

bpf/lib/nat.h
```diff
@@ -221,7 +220,13 @@ static __always_inline int snat_v4_new_mapping(struct __ctx_buff *ctx,
 
 #pragma unroll
 	for (retries = 0; retries < SNAT_COLLISION_RETRIES; retries++) {
+		rtuple.dport = bpf_htons(port);
+
+		/* Check if the selected port is already in use by a RevSNAT
+		 * entry for some other connection with the same src/dst:
+		 */
 		if (!__snat_lookup(map, &rtuple)) {
+			ostate->to_sport = rtuple.dport;
```
gentoo-root: >Also only update the SNAT tuple if the check was successful and the
>currently selected port is available.

__snat_update can still fail, and ostate will be changed, but I guess it's fine, given that we used to update it unconditionally.

bpf/lib/nat.h
```diff
@@ -171,28 +171,15 @@ struct ipv4_nat_entry *snat_v4_lookup(const struct ipv4_ct_tuple *tuple)
 	return __snat_lookup(&SNAT_MAPPING_IPV4, tuple);
 }
```
gentoo-root, line 173: >remove *swat_tuple() helpers

Typo in the commit title.

gentoo-root (approved):
Looks good, no bugs spotted.

bpf/lib/nat.h
julianwiedmann: Yep, I think reverting the change in that case would be excessive (given that the caller should be dropping the packet on such an error, not looking at the SNAT entry).

I should fix up the patch description though:
>Also only update the SNAT ~~tuple~~ entry if the check was successful and the
currently selected port is available.

julianwiedmann:
No-change push to fix up the patch descriptions.

PR #24378, opened 2023-03-15
bpf, datapath: unconditionally assume support for LRU hash maps

tklauser:
LRU hash maps are available since kernel version 4.10 [1], [2], [3]. For Cilium v1.14, the minimum required kernel version is 4.19.57 [4]. Thus we can now unconditionally assume support for that map type.

[1] https://github.com/iovisor/bcc/blob/master/docs/kernel-versions.md#map-types
[2] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=29ba732acbeece1e34c68483d1ec1f3720fa1bb3
[3] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=3a08c2fd763450a927d1130de078d6f9e74944fb
[4] https://docs.cilium.io/en/latest/operations/system_requirements/#base-requirements

For #22116

PR #26597, opened 2023-07-03
datapath: remove fill-up notification for CT maps

julianwiedmann:
With https://github.com/cilium/cilium/pull/24378 merged, CT always uses LRU maps. So we no longer have to deal with fill-up of the CT maps.

### DIFF
--- a/pkg/signal/signal.go
+++ b/pkg/signal/signal.go
@@ -30,7 +30,7 @@ type SignalType uint32
 const (
 	// SignalNatFillUp denotes potential congestion on the NAT table
 	SignalNatFillUp SignalType = iota
-	// SignalCTFillUp denotes potential congestion on the CT table
+	// SignalCTFillUp denotes potential congestion on the CT table. Unused.
 	SignalCTFillUp
 	// SignalAuthRequired denotes a connection dropped due to missing authentication
 	SignalAuthRequired
--- a/pkg/signal/signal_test.go
+++ b/pkg/signal/signal_test.go
@@ -68,7 +68,6 @@ func (t *signalSuite) TestSignalSet(c *C) {
 	sm := &signalManager{events: events}
 	c.Assert(sm.isMuted(), Equals, true)
 	c.Assert(sm.isSignalMuted(SignalNatFillUp), Equals, true)
-	c.Assert(sm.isSignalMuted(SignalCTFillUp), Equals, true)
 	c.Assert(sm.isSignalMuted(SignalAuthRequired), Equals, true)
 
 	// invalid signal, nothing changes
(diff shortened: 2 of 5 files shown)


pkg/signal/signal.go
```diff
@@ -30,7 +30,7 @@ type SignalType uint32
 const (
 	// SignalNatFillUp denotes potential congestion on the NAT table
 	SignalNatFillUp SignalType = iota
-	// SignalCTFillUp denotes potential congestion on the CT table
+	// SignalCTFillUp denotes potential congestion on the CT table. Unused.
```
qmonnet, line 33: I'd maybe expand here, “Unused, given that the datapath no longer sends this signal”.

pkg/signal/signal_test.go
```diff
@@ -77,16 +76,14 @@ func (t *signalSuite) TestSignalSet(c *C) {
...
 	c.Assert(sm.isSignalMuted(SignalAuthRequired), Equals, true)
 
 	// 2 active signals
-	err = sm.UnmuteSignals(SignalNatFillUp, SignalCTFillUp)
+	err = sm.UnmuteSignals(SignalNatFillUp, SignalAuthRequired)
 	c.Assert(err, IsNil)
 	c.Assert(sm.isMuted(), Equals, false)
 	c.Assert(sm.isSignalMuted(SignalNatFillUp), Equals, false)
-	c.Assert(sm.isSignalMuted(SignalCTFillUp), Equals, false)
-	c.Assert(sm.isSignalMuted(SignalAuthRequired), Equals, true)
```
qmonnet, line 89: Nit: We're losing this assert here, how about using another signal?

Same comment at a few other locations below

qmonnet (approved):
Looks good! Just a few, non-blocking nits below from my side

pkg/signal/signal_test.go
julianwiedmann: There's no other signal left to test  :grin:. This is the best I could come up with on a quick look.

qmonnet: OK fair enough :)

ldelossa (approved):
LGTM

squeed (approved):

joestringer (changes requested):
LRU map updates can get contended leading to this signal, we saw it in one production environment before. I can't share the details publicly but feel free to reach out to me for a private run-down.

I discussed this publicly a bit during LPC last year as well:
https://lpc.events/event/16/contributions/1368/

joestringer:
Julian and I discussed this a bit out-of-band. My perspective on this is that the signal itself is high value with no noise when triggered with a LRU-backed connection tracking map. For example, under high load scenarios it triggers relatively rarely with LRU, but if you see this signal then it immediately indicates that there is very high contention. Users can react to this my looking for ways to better distribute traffic flows across their cluster.

At the same time, there are other considerations like how userspace reacts to this event. With a CT map backed by a hashtable, this may occur more frequently and reacting to this signal by triggering CT garbage collection likely makes sense. It may also trigger more often. On the other hand, for the LRU case, it's far less likely to trigger and when it does trigger it's likely that the load is very high on the system. Hence in that scenario it may not make sense to trigger GC any more.

julianwiedmann:
> Julian and I discussed this a bit out-of-band. My perspective on this is that the signal itself is high value with no noise when triggered with a LRU-backed connection tracking map. For example, under high load scenarios it triggers relatively rarely with LRU, but if you see this signal then it immediately indicates that there is very high contention. Users can react to this metric my looking for ways to better distribute traffic flows across their cluster.
> 
> At the same time, there are other considerations like how userspace reacts to this event. With a CT map backed by a hashtable, this may occur more frequently and reacting to this signal by triggering CT garbage collection likely makes sense. On the other hand, for the LRU case, it's far less likely to trigger and when it does trigger it's likely that the load is very high on the system. Hence in that scenario it may not make sense to trigger GC any more.

thanks Joe for summarizing! Marking as draft for now - seems like we want to keep the signal around, but maybe still adjust whether it should (unconditionally) trigger GC.

PR #26600, opened 2023-07-03
.github: dump buddyinfo and pagetypeinfo when ci-e2e fails

lmb:
There is a persistent CI flake where the kernel returns ENOMEM from BPF map
updates. Grab buddyinfo and pagetypeinfo from /proc to see whether the
allocation failures are related to memory fragmentation.

Updates #26062


lmb:
Works in the sense that we get the expected output: https://github.com/cilium/cilium/actions/runs/5445805364/jobs/9906896104#step:11:225

viktor-kurchenko (approved):

PR #26619, opened 2023-07-04
proxy: refactor package global vars to proxy fields

mhofstetter:
Currently, `GetProxyPort` is exposed as global function which requires some variables to be defined as package global variables.

* `allocatedPorts`
* `proxyPorts`

This commit refactors this logic to be methods of the `proxy.Proxy` itself. Therefore, the variables can be refactored towards being fields of the proxy.

In addition, the mutex `proxyPortsMutex` can be replaced with the already existing mutex `mutex`.

This PR serves the purpose of sharpening the responsibility of the proxy before splitting it into more fine grained components.


jrajahalme (approved):
Thanks for the clean-up :-)

thorn3r (approved):
looks great!

Issue #26622, opened 2023-07-04
A null pointer occurs while reading netConf file


Hawkzilla:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

A bug happened!
[This operation does not read the CNI configuration file to *netConf
 ](https://github.com/cilium/cilium/blob/c068a443049b5dedb34608069df4b1721b1e4669/daemon/cmd/daemon.go#L420)
In ipamAWS、ipamAlibaba mode，netConf will be empty，and cannot work properly.
![image](https://github.com/cilium/cilium/assets/83360112/b158f238-4bee-416c-8b59-c5c0edb44ab9)

### Cilium Version

v1.14.0-rc.0"

### Kernel Version

Linux cilium-ipam-test-node 6.4.0-1.el8.elrepo.x86_64 #1 SMP PREEMPT_DYNAMIC Mon Jun 26 14:28:50 EDT 2023 x86_64 x86_64 x86_64 GNU/Linux

### Kubernetes Version

Client Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.4", GitCommit:"f89670c3aa4059d6999cb42e23ccb4f0b9a03979", GitTreeState:"clean", BuildDate:"2023-04-12T12:13:53Z", GoVersion:"go1.19.8", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7
Server Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.4", GitCommit:"f89670c3aa4059d6999cb42e23ccb4f0b9a03979", GitTreeState:"clean", BuildDate:"2023-04-12T12:05:35Z", GoVersion:"go1.19.8", Compiler:"gc", Platform:"linux/amd64"}

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

squeed:
Hmm, the stacktrace you've posted doesn't seem to touch the CNI code at all; can you confirm that is the issue?

PR #26623, opened 2023-07-04
daemon:fixed an error that the CNI configuration file could not be read

Hawkzilla:
We hope fix this bug , which may cause some problems in IPAMAWS、IPAMAlibaba mode.
RT： #26622 
Signed-off-by: bumblebeeK <[gleemanx@126.com](mailto:gleemanx@126.com)>

### DIFF
--- a/daemon/cmd/cni/config.go
+++ b/daemon/cmd/cni/config.go
@@ -42,13 +42,26 @@ type cniConfigManager struct {
 
 	// watcher watches for changes in the CNI configuration directory
 	watcher *fsnotify.Watcher
+
+	// mtu obtained from CNI Configuration file
+	mtu int
 }
 
 // GetMTU returns the MTU as written in the CNI configuration file.
 // This is one way to override the node's MTU.
 func (c *cniConfigManager) GetMTU() int {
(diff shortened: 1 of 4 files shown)


daemon/cmd/cni/config.go
```diff
@@ -42,29 +42,35 @@ type cniConfigManager struct {
...
+// GetNetConf returns the *NetConf obtained from CNI configuration file
+func (c *cniConfigManager) GetNetConf() *cnitypes.NetConf {
 	if c.config.ReadCNIConf == "" {
-		return 0
+		return nil
 	}
 
 	conf, err := cnitypes.ReadNetConf(c.config.ReadCNIConf)
 	if err != nil {
 		c.log.WithField("path", c.config.ReadCNIConf).WithError(err).Warnf("Failed to parse existing CNI configuration file")
```
squeed, line 69: Oops, need to return '0' here.

Hawkzilla: The return of this function is of type *NetConf，so we can't return '0' here

here is the new Pr: https://github.com/cilium/cilium/pull/26624

PR #26624, opened 2023-07-04
daemon:fixed an error that the CNI configuration file could not be read

### LINKED ISSUE #26622: A null pointer occurs while reading netConf file

Hawkzilla:
Signed-off-by: bumblebeek <gleemanx@126.com>

Fixes: #26622
We hope fix this bug , which may cause some problems in IPAMAWS、IPAMAlibaba mode.

### DIFF
--- a/bpf/lib/nat.h
+++ b/bpf/lib/nat.h
@@ -726,6 +726,8 @@ snat_v4_needs_masquerade(struct __ctx_buff *ctx __maybe_unused,
 	if (target->cluster_id != 0 &&
 	    target->cluster_id != CLUSTER_ID) {
 		target->addr = IPV4_INTER_CLUSTER_SNAT;
+		target->from_local_endpoint = true;
+
 		return NAT_NEEDED;
 	}
 # endif


squeed (changes requested):
Rather than adding a new function, can we just fix GetMTU to do the right thing?

daemon/cmd/cni/config.go
```diff
@@ -42,29 +42,35 @@ type cniConfigManager struct {
...
+func (c *cniConfigManager) GetNetConf() *cnitypes.NetConf {
 	if c.config.ReadCNIConf == "" {
-		return 0
+		return nil
 	}
 
 	conf, err := cnitypes.ReadNetConf(c.config.ReadCNIConf)
 	if err != nil {
 		c.log.WithField("path", c.config.ReadCNIConf).WithError(err).Warnf("Failed to parse existing CNI configuration file")
 	}
```
squeed: there is a bug here in the existing codebase; this should return 0 here as well.

bumblebeeK:
> Rather than adding a new function, can we just fix GetMTU to do the right thing?

okay

bumblebeeK:
> Rather than adding a new function, can we just fix GetMTU to do the right thing?

Do we need to change the GetMTU function signature

squeed:
> Do we need to change the GetMTU function signature ?

I don't think so. But I'm afraid I don't currently understand the bug this is trying to fix.

bumblebeeK:
> don't

https://github.com/cilium/cilium/blob/7726560faf2193a9a7bf06560aec830dcf49414a/daemon/cmd/daemon.go#L519
If not fixed, *netconf is nil in this function

thorn3r:
thanks for the PR, @bumblebeeK! I'm in the same position, I think the bug @squeed pointed out is worth fixing, but I'm not seeing how this fixes the linked issue https://github.com/cilium/cilium/issues/26622

bumblebeeK:
> 

Ok ,thx for your reply, i will do it!

PR #26627, opened 2023-07-04
loader: check enabled L7 proxy via config property

mhofstetter:
Currently, the check whether l7 proxy functionality is enabled is based on whether the passed proxy reference is `nil` or not.

Due to Go's nil-handling on Interfaces, this result in calls to `proxy.ReinstallRules` even though l7 proxy functionality isn't enabled. It doesn't result in nil panics, because the method doesn't access the proxy itself. (I detected the actual issue in a [separate PR](https://github.com/cilium/cilium/pull/26619) which accesses fields of the proxy struct in this method)

Therefore, this commit changes the check towards checking the config property.

As an alternative, it would be possible to add a nil check based on `reflect.ValueOf(p).IsNil()` - but i thought it's better to check the config property.

### DIFF
--- a/pkg/datapath/loader/base.go
+++ b/pkg/datapath/loader/base.go
@@ -541,7 +541,7 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner, 
 	}
 
 	// Reinstall proxy rules for any running proxies if needed
-	if p != nil {
+	if option.Config.EnableL7Proxy {
 		if err := p.ReinstallRules(ctx); err != nil {
 			return err
 		}


pkg/datapath/loader/base.go
```diff
@@ -541,7 +541,7 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner,
 	}
 
 	// Reinstall proxy rules for any running proxies if needed
-	if p != nil {
```
lmb, line 544: Why do we get a typed nil here in the first place? Your fix is pragmatic, but I wonder whether we can't get rid of the untyped nil instead. This probably isn't the only case where it's problematic and we really ought to move away from global variables influencing behaviour.

mhofstetter: @lmb When I introduced the hive cell for the Proxy ([PR](https://github.com/cilium/cilium/pull/25779)), the existing behaviour has been kept - to keep the PR as small as possible. -> If L7 proxy is disabled, the hive provider returns nil.

But i see it exactly like you - it's confusing and this logic bloats into all the modules - even though it seems as this is the only problematic case, all other cases are already checking with the config property or doing the reflection-based check.

It's just that I thought it's best to handle this as a bug and just fix it.

I plan to extract an interface for the proxy in a follow up PR anyway. This might be the chance to return an implementation which acts accordingly when L7 proxy functionality is disabled.

lmb (approved):

pkg/datapath/loader/base.go
lmb: Yeah, makes sense. I've run into this as well :(

jrajahalme (approved):
Good catch, having been bitten by `nil` checks on interfaces it's better to avoid them!

PR #26290, opened 2023-06-16
bpf: fib: delay smac selection until fib_do_redirect() has picked the oif

julianwiedmann:
fib_do_redirect() potentially takes the `oif` from the fib_params. But we currently don't consider this when selecting the smac.

Fix this by delaying the smac selection until we actually need it.

Fixes: 5fff05daf9f0 ("bpf,fib: introduce fib_do_redirect function")

PR #26638, opened 2023-07-05
bpf: nodeport: add RevDNAT-based FIB lookup for reply traffic 

### LINKED ISSUE #26166: DSR broken with VLANs on 1.13.3

### What happened?

We use cilium on tagged VLAN. Main interface has static address and VLAN interface has dynamically assigned (keepalived)  external addresses.

The following commit breaks DSR on our setup: https://github.com/cilium/cilium/commit/78f93188ae0e4a74c3794b517047e33b9c3984c5

Before 1.13.3 packets were output as it should be - on tagged interface (`wan0.4000` in our case) and with 1.13.3 packets are outputed:
* through `wan0.4000` interface if source address (packet was received on) currently directly belongs to this server (this is correct)
* through `wan0` interface in other situation and this breaks connectivity

It seems that new code ignores ip rules.

For test in VM (full setup: https://github.com/kvaster/k8s-test-guide):
10.117.10.x/24 - static internet addresses assigned to `wan0`. This addresses are used to connect via ssh to servers and to connect to internet from inside kubernetes.
10.117.12.x/24 - static addresses assigned to tagged vlan `wan0.4000`. This addresses are internal static network - used by kubernetes nodes
10.117.11.x/24 - dynamic internet addresses assigned `wan0.4000` devices via keepalived. This addresses are used by clients to connect to kubernetes ingress from outside of the world.

Following additional routes applied:
```
10.117.11.0/24 dev wan0.4000
10.117.11.0/24 dev wan0.4000 table vswitch
10.117.12.0/24 dev wan0.4000 table vswitch
```

Following additional rules applied:
```
from all to 10.244.0.0/16 lookup main priority 200
from 10.117.11.0/24 table vswitch priority 201
from 10.117.12.100 lookup vswitch priority 202
from 10.117.12.0/24 lookup vswitch priority 202
```

### Cilium Version

1.13.3

### Kernel Version

6.3.4

### Kubernetes Version

1.26.4

### Sysdump

[cilium-sysdump-20230613-124152.zip](https://github.com/cilium/cilium/files/11733377/cilium-sysdump-20230613-124152.zip)

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

julianwiedmann:
When moving the RevDNAT for backend replies from `bpf_lxc` into `bpf_host`, we lost the routing based on the post-RevDNAT layout of the packet. Depending on the installed routing rules, we then end up sending the (RevDNATed) packet from the wrong egress interface.

So as part of the RevDNAT processing in `to-netdev`, do a FIB lookup that considers the desired source IP.  And redirect the packet if it's currently on the wrong interface. We intentionally delay the actual RevDNAT rewrite until the packet has reached the correct interface, so that the packet looks as expected to components in `to-netdev` (in particular HostFW).

Long-term we would want
- a `FIB_DONE` mark on the packet (similar to `SNAT_DONE`),  to skip additional lookups.
- improve the FIB lookup  in `from-container` (for BPF Host Routing) to already consider service RevDNAT.

The initial idea was to have this check at the very top of the to-netdev program (see https://github.com/cilium/cilium/pull/26635). But we're hitting program size limits in 4.19 there, and improving that is too much of a refactor for a fix PR.

Fixes: https://github.com/cilium/cilium/issues/26166

### DIFF
--- a/bpf/lib/nodeport.h
+++ b/bpf/lib/nodeport.h
@@ -124,6 +124,29 @@ static __always_inline bool dsr_is_too_big(struct __ctx_buff *ctx __maybe_unused
 	return false;
 }
 
+static __always_inline int
+nodeport_fib_lookup_and_redirect(struct __ctx_buff *ctx,
+				 struct bpf_fib_lookup_padded *fib_params,
+				 __s8 *ext_err)
+{
+	int oif = NATIVE_DEV_IFINDEX;
+	int ret;
+
+	ret = fib_lookup(ctx, &fib_params->l, sizeof(fib_params->l), 0);
+	*ext_err = (__s8)ret;
+
+	switch (ret) {
+	case BPF_FIB_LKUP_RET_SUCCESS:
+	case BPF_FIB_LKUP_RET_NO_NEIGH:
+		if ((__u32)oif == fib_params->l.ifindex)
+			return CTX_ACT_OK;
+
+		return fib_do_redirect(ctx, true, fib_params, ext_err, &oif);
+	default:
+		return DROP_NO_FIB;
+	}
+}
+
 #ifdef ENABLE_IPV6
 static __always_inline bool nodeport_uses_dsr6(const struct ipv6_ct_tuple *tuple)
 {
@@ -1269,8 +1292,10 @@ redo:
 }
 
 static __always_inline int
-nodeport_rev_dnat_fwd_ipv6(struct __ctx_buff *ctx, struct trace_ctx *trace)
+nodeport_rev_dnat_fwd_ipv6(struct __ctx_buff *ctx, struct trace_ctx *trace,
+			   __s8 *ext_err __maybe_unused)
 {
+	struct bpf_fib_lookup_padded fib_params __maybe_unused = {};
 	struct lb6_reverse_nat *nat_info;
 	struct ipv6_ct_tuple tuple = {};
 	struct ct_state ct_state = {};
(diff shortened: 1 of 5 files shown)


gentoo-root (approved):

brb:
Before merging this PR, it would be good to get `Tests with secondary NodePort device` running. Currently, it's disabled in `ci-ginkgo` (as it doesn't create a secondary network, while the Vagrant setup did), and CLI's connectivity tests used in `ci-e2e` don't test NodePort service access via multiple networks.

Anyway, the suggested testing plan for `ci-e2e`:

1. Extend some matrix configurations (with KPR=true) in the conformance-e2e.yaml with `secondary-network: true`.
2. When provisioning a Kind cluster, pass `--secondary-network` (introduced in https://github.com/cilium/cilium/pull/26338) depending on the option above. 
3. Extend [the feature detection](https://github.com/cilium/cilium-cli/blob/main/connectivity/check/features.go) to detect that Kind has multiple networks.
4. In the [NodePort from outside test case](https://github.com/cilium/cilium-cli/blob/main/connectivity/tests/service.go#L253) consider all node IP addresses.

julianwiedmann:
Rebased to resolve trivial `bpf/tests` conflicts with https://github.com/cilium/cilium/pull/27134.

bpf/lib/nodeport.h
```diff
@@ -3012,12 +3073,14 @@ int tail_handle_nat_fwd_ipv4(struct __ctx_buff *ctx)
...
+	ret = __handle_nat_fwd_ipv4(ctx, cluster_id, &trace, &ext_err);
 	if (IS_ERR(ret))
-		return send_drop_notify_error(ctx, 0, ret, CTX_ACT_DROP, METRIC_EGRESS);
+		return send_drop_notify_error_ext(ctx, 0, ret, ext_err,
+						  CTX_ACT_DROP, METRIC_EGRESS);
 
-	send_trace_notify(ctx, obs_point, 0, 0, 0, 0, trace.reason,
-			  trace.monitor);
+	if (ret == CTX_ACT_OK)
+		send_trace_notify(ctx, obs_point, 0, 0, 0, 0, trace.reason,
```
gentoo-root, line 3082: For my education... do we only trace the forwarded packets at the destination interface?

julianwiedmann: I'm not sure if we have a strict precedent for it at the moment.

I could be convinced that throwing notifications for the intermediate interfaces has value. But then we should also apply proper CT-driven aggregration, and include the corresponding ifindex?

julianwiedmann:
No intentions to backport this into v1.13 anymore. Doing so would require https://github.com/cilium/cilium/pull/26136 and https://github.com/cilium/cilium/pull/26290.

PR #26640, opened 2023-07-05
.clang-format: Re-write and re-license .clang-format

qmonnet:
As part of the effort to harmonise licensing for the different files in the project, rewrite the .clang-format from scratch, and remove the attached GPL v2 license.

I went through all options for clang-format <= 15 currently described on https://clang.llvm.org/docs/ClangFormatStyleOptions.html, set them to the value we desire, and commented out the ones that are not relevant to Cilium's BPF code.

Tweaks for penalties were kept from commit b482af0f2de9 ("Apply initial cilium-specific tweaks").

Obviously, there are some similarities with the previous version, given that it's a config file and that we still want more or less the same config as before.

I didn't use any particular license marker, because I don't see it relevant for a config file; if one is necessary, the file now falls under the license global to the repository (Apache v2.0).

Cc: @bleggett, could you please ACK if you agree with the change?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,23 +1,19 @@
-# SPDX-License-Identifier: GPL-2.0
+# Configuration file for clang-format.
+# Intended for clang-format >= 15.
 #
-# clang-format configuration file. Intended for clang-format >= 15.
+# The list and meaning of the options is available at:
 #
-# For more information, see:
-#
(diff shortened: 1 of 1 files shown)


bimmlerd (approved):
Thanks!

aanm (approved):

PR #26642, opened 2023-07-05
auth: optimize log output for pending auth

mhofstetter:
This commit optimizes the logoutput by given insights when an authentication request is skipped due to a pending auth request. Without this log entry, it might be quite irritating why so many auth requests are logged without further insights.

In addition, a log output is added when the pending auth is cleared.


meyskens (approved):
LGTM

PR #26643, opened 2023-07-05
[highscale] fqdn/policy: expand service frontends IPs to backend IPs 

squeed:
For more details, see the individual commit messages.

As part of enabling egress policy for high-scale cache, we would like to enable toFQDN policy to reference kubernetes service IPs. However, this presently does not work because policy enforcement always sees the "real" destination IP, not the virtual / service IP.

So, we need to configure the policy engine to expand frontend IPs to their backend IPs without changing the DNS response. Additionally, we need to recompute FQDN policy whenever a service's backend IPs have changed.

```release-note
TODO
```

### DIFF
--- a/pkg/fqdn/cache.go
+++ b/pkg/fqdn/cache.go
@@ -14,6 +14,8 @@ import (
 	"time"
 	"unsafe"
 
+	"k8s.io/apimachinery/pkg/util/sets"
+
 	"github.com/cilium/cilium/pkg/fqdn/matchpattern"
 	"github.com/cilium/cilium/pkg/fqdn/re"
 	ippkg "github.com/cilium/cilium/pkg/ip"
@@ -139,6 +141,14 @@ type DNSCache struct {
 	// sent in the Update is lower, the TTL will be owerwritten to this value.
 	// Due is only read-only is not protected by the mutex.
 	minTTL int
+
+	// vipToBackends is the mapping from a frontend IP to one or more backend IPs.
+	// It is used for fqdn policies that select frontend IPs -- we need to allow the
+	// backend IPs when computing allowed identities
+	vipToBackends map[netip.Addr][]netip.Addr
+	// backendsToVips is the mapping from a backend IP to one or more frontend IPs.
+	// It is used for reverse lookups.
+	backendsToVips map[netip.Addr][]netip.Addr
 }
 
 // NewDNSCache returns an initialized DNSCache
@@ -151,6 +161,9 @@ func NewDNSCache(minTTL int) *DNSCache {
 		overLimit:    map[string]bool{},
 		perHostLimit: 0,
 		minTTL:       minTTL,
+
+		vipToBackends:  map[netip.Addr][]netip.Addr{},
+		backendsToVips: map[netip.Addr][]netip.Addr{},
 	}
 	return c
 }
@@ -406,15 +419,15 @@ func (c *DNSCache) Lookup(name string) (ips []netip.Addr) {
 	return c.lookupByTime(c.lastCleanup, name)
 }
 
-// lookupByTime takes a timestamp for expiration comparisons, and is only
-// intended for testing.
+// lookupByTime takes a timestamp for expiration comparisons
 func (c *DNSCache) lookupByTime(now time.Time, name string) (ips []netip.Addr) {
 	entries, found := c.forward[name]
 	if !found {
 		return nil
 	}
 
-	return entries.getIPs(now)
+	ips = entries.getIPs(now)
+	return c.expandVirtualIPs(ips)
 }
 
 // LookupByRegexp returns all non-expired cache entries that match re as a map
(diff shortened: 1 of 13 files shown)


pkg/fqdn/cache.go
```diff
@@ -699,6 +746,105 @@ func (c *DNSCache) UnmarshalJSON(raw []byte) error {
...
+			}
+		}
+	}
+	return updatedNames
+}
+
+// setVirtualIP updates a single frontend -> backend mapping. If mergeExisting is true, then
+// merge the new state with the existing state.
+// Returns true if the state was updated.
+func (c *DNSCache) setVirtualIP(frontend netip.Addr, backends []netip.Addr, mergeExisting bool) bool {
```
christarazi, line 788: Let's document that it must be called with the DNSCache mutex locked

pkg/k8s/watchers/watcher.go
```diff
@@ -613,10 +632,39 @@ func (k *K8sWatcher) k8sServiceHandler() {
...
 		case k8s.UpdateService:
+			if k.dnsNameManager != nil {
+				// Update the the frontend -> backend mapping in the fqdn policy manager.
+				// This may trigger policy updates if any fqdn policies reference the frontends.
+				// This update merges any existing backends with the new ones
+				wg, usedIdentities, newlyAllocatedIdentities, err := k.dnsNameManager.UpdateLBIPMapping(ippkg.MustAddrsFromIPs(svc.FrontendIPs), event.Endpoints.Addrs(), true)
+				if err != nil {
+					scopedLog.WithError(err).Error("Failed to update FQDN policies with new frontend -> backend mappings")
+				} else {
+					if err := waitTimeout(wg, option.Config.FQDNProxyResponseMaxDelay); err != nil {
```
christarazi, line 649: This could delay responding to updates for services/endpoints up to `FQDNProxyResponseMaxDelay` amount of time. Given https://github.com/cilium/cilium/pull/26643/files#r1254127031, any unoptimized code in this hotpath will definitely risk hitting more timeouts of `FQDNProxyResponseMaxDelay`  (consequence is transient policy drops). This is especially concerning if there is heavy churn with services/endpoints because it requires the DNSCache lock to be taken which is in the hotpath of the DNS proxy handling DNS requests. Meaning it could also delay the DNS proxy as well. One slight mitigating circumstance is that the DNSCache object is per endpoint, which is better than a global DNSCache because that would mean global lock contention. However, if an endpoint has many connections to many different services that all experience churn, it would still be heavy lock contention on that endpoint's specific DNSCache.

pkg/fqdn/cache.go
```diff
@@ -137,6 +141,14 @@ type DNSCache struct {
...
 	// Due is only read-only is not protected by the mutex.
 	minTTL int
+
+	// vipToBackends is the mapping from a frontend IP to one or more backend IPs.
+	// It is used for fqdn policies that select frontend IPs -- we need to allow the
+	// backend IPs when computing allowed identities
+	vipToBackends map[netip.Addr][]netip.Addr
+	// backendsToVips is the mapping from a backend IP to one or more frontend IPs.
+	// It is used for reverse lookups.
+	backendsToVips map[netip.Addr][]netip.Addr
```
christarazi, line 151: When the TTL expires and GC runs, don't we also need to clean these maps up? I only see these maps touched from the `UpdateLBIPMapping` code path.

pkg/fqdn/cache.go
```diff
@@ -699,6 +746,105 @@ func (c *DNSCache) UnmarshalJSON(raw []byte) error {
...
+
+	// If nothing changed; we're done
+	if len(frontendsWithChanges) == 0 {
+		return nil
+	}
+
+	updatedNames := []string{}
+	for name, entry := range c.forward {
+		for ip := range entry {
+			if frontendsWithChanges.Has(ip) {
```
squeed: This loop looks kind of expensive, and we need to call it on every single service update. We may wish to consider another tracking map instead.

pkg/k8s/watchers/watcher.go
christarazi: One idea to resolve this would be for the DNS name manager to have a queue where the code here enqueues an event to be processed without blocking, and a separate goroutine of the DNS name manager will handle processing the event.

christarazi (changes requested):
Architecturally, I have a worry that side-loading the DNS cache with updates that are fed by the services/endpoints watcher may not scale, but we don't have any official way to measure the impact, yet at least. I am wondering whether it's worth investing in making quite drastic changes to the `toFQDNs` logic when we know it's a policy model that scales with number of unique IPs, rather than identities. Moving forward a more ideal model would be to use `toServices` with an ability to leverage service labels for use in identities so that we can switch back to a model that scales with number of unique identities (services), rather than IPs. This is how the current label-based policy model works in Cilium today. Let's table this for now though until we discuss this further.

With that said, I have a couple of concerns with the implementation as it is today. I've mostly focused on the hotpath-affecting code changes.

pkg/k8s/watchers/watcher.go
squeed: That would work but lead to policy drops between applying the changes to the load balancers and pushing out the policy map updates.

pkg/fqdn/cache.go
squeed: Good question. No, these maps are just for service ip -> backend IPs (and the reverse); they're populated regardless of whether or not names reference them.

PR #26645, opened 2023-07-05
statedb: extract REST API handler to pkg

bimmlerd:
With the newly modularized API server, we can pull out individual handlers into their packages; in this case the statedb Dump handler.


pkg/statedb/api_handler.go
```diff
@@ -0,0 +1,30 @@
...
+	"net/http"
+
+	"github.com/go-openapi/runtime"
+	"github.com/go-openapi/runtime/middleware"
+
+	. "github.com/cilium/cilium/api/v1/server/restapi/statedb"
+	restapi "github.com/cilium/cilium/api/v1/server/restapi/statedb"
+)
+
+func createHandler(db DB) restapi.GetStatedbDumpHandler {
```
joamaki: Let's use a more descriptive name, either s/createHandler/newDumpHandler/, or if we'll use one constructor for potential future API handlers then something like `newAPIHandlers`

pkg/statedb/api_handler.go
```diff
@@ -0,0 +1,30 @@
...
+	return &dumphandler{db}
+}
+
+// REST API handler for the StateDB, handles the JSON Dumping.
```
joamaki: Hey, let's add comment here that this is available as "cilium statedb dump" and that this output is included in the sysdump.

bimmlerd:
CI triage:
- ci-e2e seems to have hit https://github.com/cilium/cilium/issues/26158 which was fixed recently - rerunning once, if I hit it again I'll rebase.

pkg/statedb/cell.go
```diff
@@ -22,6 +22,7 @@ var Cell = cell.Module(
 	"In-memory database",
 
 	cell.Provide(New),
+	cell.Provide(newDumpHandler), // REST API Handler
```
joamaki, line 25: nit(in case you rebase): you can collapse the provide's into one which I find
easier to read:

```
cell.Provide(
  New,
  newDumpHandler, // The /statedb/dump REST API handler
)
```

joamaki (approved):

PR #26656, opened 2023-07-06
bpf: Suppress Key Mismatch Log for Certain Maps

### LINKED ISSUE #25224: Key-size mismatch warning on 1.14 upgrades/downgrades
https://github.com/cilium/cilium/pull/23885 introduced a warning on upgrades/downgrades to/from v1.14:
```
"Key-size mismatch for BPF map" file-path=/sys/fs/bpf/tc/globals/cilium_policy_00835 new=8 old=12 subsys=bpf
```
This warning can't be avoided by users today and it is expected/non-actionable.

We need to fix that before we release v1.14.
cc @jrajahalme

nathanjsweet:
Some bpf maps do not affect any data loss or
require any action by operators when there is a
key size mismatch between bpf map versions.
When that is the case the warning log message
should be suppressed so as not to alarm
operators.

Fixes: #25224

### DIFF
--- a/pkg/bpf/bpf_linux.go
+++ b/pkg/bpf/bpf_linux.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"strings"
 
 	"github.com/cilium/ebpf"
 	"github.com/sirupsen/logrus"
@@ -19,6 +20,13 @@ import (
 	"github.com/cilium/cilium/pkg/metrics"
 	"github.com/cilium/cilium/pkg/spanstat"
 )
+
+var keySizeExceptions = []string{
+	// "cilium_policy" maps are populated in userspace
+	// and will not cause data loss when there is change
+	// in key size.
+	"cilium_policy",
+}
 
 // createMap wraps a call to ebpf.NewMapWithOptions while measuring syscall duration.
 func createMap(spec *ebpf.MapSpec, opts *ebpf.MapOptions) (*ebpf.Map, error) {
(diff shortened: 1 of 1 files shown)


pkg/bpf/bpf_linux.go
```diff
@@ -52,11 +56,26 @@ func objCheck(m *ebpf.Map, path string, mapType ebpf.MapType, keySize, valueSize
 		mismatch = true
 	}
 
+	if mi, err := m.Info(); err != nil {
+		scopedLog.Error("Could not get mapinfo for map %+v", *m)
```
dylandreimerink: The linter points it out already, but just to be sure, this should be an `Errorf` not an `Error` given the format option

pkg/bpf/bpf_linux.go
```diff
@@ -20,6 +21,8 @@ import (
 	"github.com/cilium/cilium/pkg/spanstat"
 )
 
+var keySizeExceptions = []string{"cilium_policy"}
```
dylandreimerink: Could you add a comment to explain when maps are exempt from the key size check? so it is clear why this code exists.

dylandreimerink (changes requested):

dylandreimerink (approved):
:rocket:

joestringer:
I know that @ti-mo and @jrajahalme had discussed approaches to solve this in the past, so I'll CC them for their input on the long term maintainability approach here.

ti-mo:
This is a band-aid, and objCheck is going to be removed soon. Maps that tolerate data loss (typically maps that are write-only from the agent side like ipcache and policy) should use `Map.Recreate()` instead of `Map.OpenOrCreate()` to explicitly indicate this fact. Maybe this PR could take that approach instead?

I have patches floating around that do this for most maps, along with changing `OpenOrCreate()` to no longer implicitly recreate maps, but haven't had the time to polish them and put them up for review, ETA would be early August. This is part of the ongoing bpf.Map work.

nathanjsweet:
A couple of things:
1. https://github.com/cilium/cilium/issues/26793 is filed to track managing this problem more effectively/comprehensively.
2. For some reason, CI ran Jenkins jobs for this PR, even though they no longer exist. The Jenkins jobs are the only failures for this CI. Marking ready-to-merge.

nathanjsweet:
Closing. See [this issue comment](https://github.com/cilium/cilium/issues/25224#issuecomment-1633136799) for why.

PR #26657, opened 2023-07-06
proxy: introduce envoy cell

mhofstetter:
This PR introduces a dedicated hive cell for the Envoy proxy and its control plane.

xDS- and accesslog server are moved from the proxy cell to the new Envoy proxy cell.

In addition, an interface has been extracted for the xds server. This will be used in upcoming refactorings to decouple the proxy from the xds server.

Follow up of https://github.com/cilium/cilium/pull/25779


pkg/envoy/cell.go
```diff
@@ -0,0 +1,86 @@
...
+
+func newEnvoyXDSServer(params xdsServerParams) (*XDSServer, error) {
+	xdsServer, err := newXDSServer(GetSocketDir(option.Config.RunDir), params.IPCache)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create Envoy xDS server: %w", err)
+	}
+
+	if !option.Config.EnableL7Proxy {
+		log.Debug("L7 proxies are disabled - don't start Envoy xDS server")
+		return xdsServer, nil
```
joamaki, line 40: What's the behavior of the public API methods of `XDSServer` when it hasn't been started?

Also could we hide `EnableL7Proxy` option inside this package (or pkg/proxy) and refactor the users of this API to not care whether or not it's been enabled? E.g. by either always calling a no-op method or adding `IsEnabled` or similar if there's expensive work needed prior to a call. I think it'd be preferably to reduce reliance on checking "option.Config.EnableFoo" and instead make APIs that are safe to use regardless of whether they're enabled or not.

pkg/envoy/cell.go
```diff
@@ -0,0 +1,86 @@
...
+type accessLogServerParams struct {
+	cell.In
+
+	Lifecycle hive.Lifecycle
+	XdsServer *XDSServer
+}
+
+func registerEnvoyAccessLogServer(params accessLogServerParams) {
+	if !option.Config.EnableL7Proxy {
+		log.Debug("L7 proxies are disabled - don't start Envoy AccessLog server")
```
joamaki: nit: we're not telling the user not to start it, so I'd word this as "not starting"

pkg/envoy/xds_server.go
```diff
@@ -217,24 +217,19 @@ func newXDSServer(envoySocketDir string, ipCache IPCacheEventSource) (*xdsServer
...
-	if err = os.Chmod(s.socketPath, 0660); err != nil {
-		return fmt.Errorf("failed to change mode of xDS listen socket at %s: %w", s.socketPath, err)
-	}
-	// Change the group to ProxyGID allowing access from any process from that group.
-	if err = os.Chown(s.socketPath, -1, option.Config.ProxyGID); err != nil {
-		log.WithError(err).Warningf("Envoy: Failed to change the group of xDS listen socket at %s, sidecar proxies may not work", s.socketPath)
-	}
+	resourceConfig := s.initializeXdsConfigs()
 
+	s.stopFunc = startXDSGRPCServer(socketListener, resourceConfig, 5*time.Second)
```
joamaki, line 227: I know it's the old code, but if it's not too much trouble, lift the `5*time.Second` into a constant and document what it's about. Also wondering how resilient this is.. what happens if the timeout is hit? How do we recover from it?

joamaki (approved):

jrajahalme (approved):

pkg/envoy/cell.go
mhofstetter: thanks, reworded.

pkg/envoy/xds_server.go
mhofstetter: good hint - which should lead to deletion of some code - because eventually this duration is used to create a context with timeout... which isn't used at all :) 

will open a separate PR if this is ok for you

pkg/envoy/cell.go
mhofstetter: > What's the behavior of the public API methods of XDSServer when it hasn't been started?

it's just updating it's internal state without actually exposing the config via gRPC.  (and for the time being it's access is "secured" by the feature flag anyway)

> Also could we hide EnableL7Proxy option inside this package (or pkg/proxy) and refactor the users of this API to not care whether or not it's been enabled? E.g. by either always calling a no-op method or adding IsEnabled or similar if there's expensive work needed prior to a call. I think it'd be preferably to reduce reliance on checking "option.Config.EnableFoo" and instead make APIs that are safe to use regardless of whether they're enabled or not.

I'd like to refactor this in a follow up PR too. I already pointed this out here https://github.com/cilium/cilium/pull/26627#discussion_r1252180307. But this needs to be handled in the proxy cell - which should return a noop-proxy in case of L7 proxy functionality is disabled.

mhofstetter:
@joamaki thanks for the feedback and questions. i answered them inline. most will be handled in planned follow ups.

mhofstetter:
rebased to `main`

pkg/envoy/xds_server.go
mhofstetter: filed the PR: https://github.com/cilium/cilium/pull/26747

Issue #26513, opened 2023-06-27
CI: ClusterMesh (ci-multicluster) connectivity test failed: timeout reached waiting lookup for localhost from pod cilium-test/client2


gandro:
## CI failure

Error:
```
 ⌛ [kind-cluster1-5390656331] Waiting for pod cilium-test/client2-76f4d7c5bc-ssxp7 to reach DNS server on cilium-test/echo-other-node-545c9b778b-btrlh pod...
connectivity test failed: timeout reached waiting lookup for localhost from pod cilium-test/client2-76f4d7c5bc-ssxp7 to server on pod cilium-test/echo-other-node-545c9b778b-btrlh to succeed (last error: context deadline exceeded)
Error: Process completed with exit code 1.
```

Observed in two separate PRs [here](https://github.com/cilium/cilium/actions/runs/5391374377/jobs/9788172243) and [here](https://github.com/cilium/cilium/actions/runs/5390656331/jobs/9786352042)

Looking at the sysdump, I don't see any obvious errors with that pod and no hubble drops either.

marseel:
I think we are missing hubble logs - I took a look at [this](https://github.com/cilium/cilium/actions/runs/5390656331/jobs/9786352042) 
and `hubble-flows-cilium-r6qnp-20230627-135702.json` has the first log with timestamp `2023-06-27T13:56:34.982006946Z` which is already after the test was run: `13:56:28 GMT`

marseel:
I took a look at multiple runs, and it seems that the only thing in common is kvstoremesh. 
First failed run that I found with this error: https://github.com/cilium/cilium/actions/runs/5290437383/jobs/9574599699
@giorio94  could you take a look, please?

giorio94:
I've checked a few runs, including the ones initially mentioned by @gandro and [another from a scheduled run](https://github.com/cilium/cilium/actions/runs/5360483511/jobs/9725417855), and all flakes seem to be caused by unlucky timing. 

The failure of the run mentioned by @marseel, instead, was fixed by https://github.com/cilium/cilium/pull/26363.

Taking the scheduled run as an example.

The check was performed between :24 and :44:
```
21:05:24 GMT ⌛ [kind-cluster1-5360483511] Waiting for pod cilium-test/client2-76f4d7c5bc-h4dxl to reach DNS server on cilium-test/echo-other-node-545c9b778b-4dxsv pod...
[...]

Issue #26321, opened 2023-06-16
CI: ClusterMesh: no-policies/pod-to-world/http-to-google.com-0: curl: (28) Resolving timed out after 2000 milliseconds


qmonnet:
## CI failure

`Installation and Connectivity Test (8, vxlan, ipv4, ipsec, iptables)`

```
🏃 Running tests...
[=] Test [no-policies]
...............

[...]

  [-] Scenario [no-policies/pod-to-world]
  [.] Action [no-policies/pod-to-world/http-to-google.com-0: cilium-test/client-6965d549d5-d8jbx (10.242.1.115) -> google.com-http (google.com:80)]
  ❌ command "curl -w %{local_ip}:%{local_port} -> %{remote_ip}:%{remote_port} = %{response_code} --silent --fail --show-error --output /dev/null --connect-timeout 2 --max-time 10 --retry 3 --retry-all-errors --retry-delay 3 [http://google.com:80](http://google.com/)" failed: command failed: curl: (28) Resolving timed out after 2000 milliseconds

  ℹ️  curl output:
  
  
  📄 No flows recorded for peer cilium-test/client-6965d549d5-d8jbx during action http-to-google.com-0
  📄 No flows recorded for peer google.com-http during action http-to-google.com-0
```

- Hit on https://github.com/cilium/cilium/pull/25746
- Link: https://github.com/cilium/cilium/actions/runs/5278695238/jobs/9548390837
- Sysdump: [cilium-sysdumps-8.zip](https://github.com/cilium/cilium/files/11772391/cilium-sysdumps-8.zip)

PR #26663, opened 2023-07-06
ipam: Fix invalid PodCIDR in CiliumNode in ENI/Azure/MultiPool mode

### LINKED ISSUE #9409: ENI: Do not fill ciliumnodes with allocated podCIDRs in ENI mode
In ENI mode, `spec.ipam.podCIDRs` of `ciliumnodes` is filled with CIDR allocated by cilium and it's never used. Remove `spec.ipam.podCIDRs` to avoid confusion with ENI IP addresses allocated by operator.

gandro:
This commit fixes an issue where Cilium-Agent would announce its internal auto-generated IPv4/IPv6AllocRange by writing it into the local CiliumNode resource.

Cilium-Agents IPv4/IPv6AllocRange is Cilium's internal representation of the local node's PodCIDR. The concept of a single allocation CIDR however is outdated, since in many IPAM modes (ENI, Azure, AlibabaCloud, MultiPool) there either is no PodCIDR to begin with (and pod IPs are allocated using a different scheme), or there are multiple PodCIDRs (in the case of MultiPool).

Before this commit, cilium-agent used to copy its internal PodCIDR into the CiliumNode resource, but this behavior is only correct in Kubernetes IPAM mode. In Kubernetes IPAM mode, the IPv4/IPv6AllocRange is taken from the Kubernetes Node resource, and thus it makes sense to announce it to other cilium-agent instances via the CiliumNode resource. This field then is for example used by `auto-direct-node-routes` to install routes for direct routing mode.

However, in all other IPAM modes that we have, copying the internal PodCIDR into the the CiliumNode resource does not make sense:

 - In ClusterPool mode, the CiliumNode PodCIDR field is populated by cilium-operator. Therefore, we should try to overwrite it ourselves. This behavior is unchanged by this commit.
 - In all other current IPAM modes, the CiliumNode PodCIDR field is not used. Instead we use fields specific to those IPAM modes (c.f. CiliumNode.Spec.IPAM). In those modes, the previous code therefore wrote a auto-generated but otherwise IPv4/IPv6AllocRange into the field. That auto-generated PodCIDR is not used for pod IP allocation in those modes, so it does not make sense to announce it either.

The announced auto-generated PodCIDR is causing issues in IPAM MultiPool mode. With this commit, we will no longer install invalid routes for the auto-generated PodCIDR. This was mostly harmless (since the auto-generated PodCIDR rarely overlapped with the real ones), but still a potential cause for confusion and bugs.

In addition, ENI-mode users have reported that they are confused by the this "fake" PodCIDR, which is understandable, since that PodCIDR is meaningless in ENI mode (#9409).

Long term, we want to remove the concept of the IPv4/IPv6AllocCIDR completely (and remove the auto-generation code), but this is a first incremental step towards that goal.

Fixes: #9409

```release-note
Fix potential cross-node connectivity issue when IPsec is enabled with ENI or Azure IPAM modes.
```

### DIFF
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -1081,11 +1081,10 @@ func (n *linuxNodeHandler) enableIPsecIPv4(newNode *nodeTypes.Node, nodeID uint1
 	var spi uint8
 	var errs error
 
-	if !n.nodeConfig.EnableIPv4 || newNode.IPv4AllocCIDR == nil {
+	if !n.nodeConfig.EnableIPv4 || (newNode.IPv4AllocCIDR == nil && !n.subnetEncryption()) {
 		return nil
 	}
 
-	new4Net := newNode.IPv4AllocCIDR.IPNet
 	wildcardIP := net.ParseIP(wildcardIPv4)
 	wildcardCIDR := &net.IPNet{IP: wildcardIP, Mask: net.IPv4Mask(0, 0, 0, 0)}
 
@@ -1093,14 +1092,18 @@ func (n *linuxNodeHandler) enableIPsecIPv4(newNode *nodeTypes.Node, nodeID uint1
 	errs = errors.Join(errs, upsertIPsecLog(err, "default-drop IPv4", wildcardCIDR, wildcardCIDR, spi))
 
 	if newNode.IsLocal() {
-		errs = errors.Join(errs, n.replaceNodeIPSecInRoute(new4Net))
-
 		localIP := newNode.GetCiliumInternalIP(false)
 		if localIP == nil {
 			return errs
 		}
 
 		if n.subnetEncryption() {
+			// FIXME: Remove the following four lines in Cilium v1.16
+			if localCIDR := n.nodeAddressing.IPv4().AllocationCIDR(); localCIDR != nil {
+				// This removes a bogus route that Cilium installed prior to v1.15
+				_ = route.Delete(n.createNodeIPSecInRoute(localCIDR.IPNet))
+			}
+
 			localNodeInternalIP, err := getV4LinkLocalIP()
 			if err != nil {
 				log.WithError(err).Error("Failed to get local IPv4 for IPsec configuration")
@@ -1126,6 +1129,7 @@ func (n *linuxNodeHandler) enableIPsecIPv4(newNode *nodeTypes.Node, nodeID uint1
 			}
 
 			localCIDR := n.nodeAddressing.IPv4().AllocationCIDR().IPNet
+			errs = errors.Join(errs, n.replaceNodeIPSecInRoute(localCIDR))
 			spi, err = ipsec.UpsertIPsecEndpoint(localCIDR, wildcardCIDR, localIP, wildcardIP, 0, ipsec.IPSecDirIn, false)
 			errs = errors.Join(errs, upsertIPsecLog(err, "in IPv4", localCIDR, wildcardCIDR, spi))
 		}
@@ -1154,7 +1158,7 @@ func (n *linuxNodeHandler) enableIPsecIPv4(newNode *nodeTypes.Node, nodeID uint1
 			}
 		} else {
 			remoteCIDR := newNode.IPv4AllocCIDR.IPNet
-			n.replaceNodeIPSecOutRoute(new4Net)
+			n.replaceNodeIPSecOutRoute(remoteCIDR)
 			spi, err = ipsec.UpsertIPsecEndpoint(wildcardCIDR, remoteCIDR, localIP, remoteIP, nodeID, ipsec.IPSecDirOut, false)
 			errs = errors.Join(errs, upsertIPsecLog(err, "out IPv4", wildcardCIDR, remoteCIDR, spi))
 		}
@@ -1166,11 +1170,10 @@ func (n *linuxNodeHandler) enableIPsecIPv6(newNode *nodeTypes.Node, nodeID uint1
 	var errs error
 	var spi uint8
 
-	if !n.nodeConfig.EnableIPv6 || newNode.IPv6AllocCIDR == nil {
+	if !n.nodeConfig.EnableIPv6 || (newNode.IPv6AllocCIDR == nil && !n.subnetEncryption()) {
 		return nil
 	}
 
-	new6Net := newNode.IPv6AllocCIDR.IPNet
 	wildcardIP := net.ParseIP(wildcardIPv6)
 	wildcardCIDR := &net.IPNet{IP: wildcardIP, Mask: net.CIDRMask(0, 128)}
 
@@ -1181,14 +1184,18 @@ func (n *linuxNodeHandler) enableIPsecIPv6(newNode *nodeTypes.Node, nodeID uint1
 	errs = errors.Join(errs, upsertIPsecLog(err, "default-drop IPv6", wildcardCIDR, wildcardCIDR, spi))
 
 	if newNode.IsLocal() {
-		errs = errors.Join(errs, n.replaceNodeIPSecInRoute(new6Net))
-
 		localIP := newNode.GetCiliumInternalIP(true)
 		if localIP == nil {
 			return errs
 		}
 
 		if n.subnetEncryption() {
+			// FIXME: Remove the following four lines in Cilium v1.16
+			if localCIDR := n.nodeAddressing.IPv6().AllocationCIDR(); localCIDR != nil {
+				// This removes a bogus route that Cilium installed prior to v1.15
+				_ = route.Delete(n.createNodeIPSecInRoute(localCIDR.IPNet))
+			}
+
 			localNodeInternalIP, err := getV6LinkLocalIP()
 			if err != nil {
 				log.WithError(err).Error("Failed to get local IPv6 for IPsec configuration")
@@ -1204,6 +1211,7 @@ func (n *linuxNodeHandler) enableIPsecIPv6(newNode *nodeTypes.Node, nodeID uint1
 			}
 		} else {
 			localCIDR := n.nodeAddressing.IPv6().AllocationCIDR().IPNet
+			errs = errors.Join(errs, n.replaceNodeIPSecInRoute(localCIDR))
 			spi, err = ipsec.UpsertIPsecEndpoint(localCIDR, wildcardCIDR, localIP, wildcardIP, 0, ipsec.IPSecDirIn, false)
 			errs = errors.Join(errs, upsertIPsecLog(err, "in IPv6", localCIDR, wildcardCIDR, spi))
 		}
@@ -1233,7 +1241,7 @@ func (n *linuxNodeHandler) enableIPsecIPv6(newNode *nodeTypes.Node, nodeID uint1
 			}
 		} else {
 			remoteCIDR := newNode.IPv6AllocCIDR.IPNet
-			n.replaceNodeIPSecOutRoute(new6Net)
+			n.replaceNodeIPSecOutRoute(remoteCIDR)
 			spi, err := ipsec.UpsertIPsecEndpoint(wildcardCIDR, remoteCIDR, localIP, remoteIP, nodeID, ipsec.IPSecDirOut, false)
 			errs = errors.Join(errs, upsertIPsecLog(err, "out IPv6", wildcardCIDR, remoteCIDR, spi))
 		}
--- a/pkg/nodediscovery/nodediscovery.go
+++ b/pkg/nodediscovery/nodediscovery.go
@@ -507,17 +507,16 @@ func (n *NodeDiscovery) mutateNodeResource(nodeResource *ciliumv2.CiliumNode) er
 		}
 	}
 
-	switch option.Config.IPAM {
-	case ipamOption.IPAMClusterPool, ipamOption.IPAMClusterPoolV2:
-		// We want to keep the podCIDRs untouched in these IPAM modes because
-		// the operator will verify if it can assign such podCIDRs.
-		// If the user was running in non-IPAM Operator mode and then switched
-		// to IPAM Operator, then it is possible that the previous cluster CIDR
-		// from the old IPAM mode differs from the current cluster CIDR set in
-		// the operator.
-		// There is a chance that the operator won't be able to allocate these
-		// podCIDRs, resulting in an error in the CiliumNode status.
-	default:
+	if option.Config.IPAM == ipamOption.IPAMKubernetes {
+		// We only want to copy the PodCIDR from the Kubernetes Node resource to
+		// the CiliumNode resource in IPAM Kubernetes mode. In other PodCIDR
+		// based IPAM modes (such as ClusterPool or MultiPool), the operator
+		// will set the PodCIDRs of the CiliumNode and those might be different
+		// from the ones assigned by Kubernetes.
+		// For non-podCIDR based IPAM modes (e.g. ENI, Azure, AlibabaCloud), there
+		// is no such thing as a podCIDR to begin with. In those cases, the
+		// IPv4/IPv6AllocRange is auto-generated and otherwise unused, so it does not
+		// make sense to copy it into the CiliumNode it either.
 		nodeResource.Spec.IPAM.PodCIDRs = []string{}
 		if cidr := node.GetIPv4AllocRange(); cidr != nil {
 			nodeResource.Spec.IPAM.PodCIDRs = append(nodeResource.Spec.IPAM.PodCIDRs, cidr.String())


doniacld (approved):

gandro:
- [CI Multicluster](https://github.com/cilium/cilium/actions/runs/5474058352) failed with #26513 - restarting
- [CI AKS](https://github.com/cilium/cilium/actions/runs/5474058322) failed due to quota (reported to infra team)
- [CI EKS](https://github.com/cilium/cilium/actions/runs/5474058443/jobs/9968346055) seems to have hit a new issue? I've created https://github.com/cilium/cilium/issues/26665 Unlikely that it is related, but I'll investigate in the issue and restart

thorn3r (approved):
thanks for the detailed commit msg, looks great :+1:

gandro:
The IPSec failures seem legitimate. It seems we do not setup IPSec if the node does not have a podCIDR, even in ENI mode:

https://github.com/cilium/cilium/blob/9c78f2a3c5c50d8d37619af8be4e8f225d791d28/pkg/datapath/linux/node.go#L1006

mathpl:
Does this relate to https://github.com/cilium/cilium/issues/22273 @gandro?

gandro:
> Does this relate to #22273 @gandro?

Yes, that issue too (we have to many issues caused by the invalid PodCIDRs :disappointed: )

gandro:
This is currently in draft because it affects the IPSec code base, which is currently also undergoing a refactor. Paul asked me to hold off with this change until the IPSec refactor dust has settled.

gandro:
> This is currently in draft because it affects the IPSec code base, which is currently also undergoing a refactor. Paul asked me to hold off with this change until the IPSec refactor dust has settled.

@pchaigno Are more changes to this part of the IPSec code base planned or is now a good time to try to get the changes in this PR merged?

In particular, if I recall correctly, you said the changes themselves were fine, but we need to also consider the upgrade/downgrade path. I see we now also have new IPSec pipelines. Do they test upgrade/downgrade or do I need to di manual testing still for this?

pchaigno:
> @pchaigno Are more changes to this part of the IPSec code base planned or is now a good time to try to get the changes in this PR merged?
> 
> In particular, if I recall correctly, you said the changes themselves were fine, but we need to also consider the upgrade/downgrade path. I see we now also have new IPSec pipelines. Do they test upgrade/downgrade or do I need to di manual testing still for this?

Nice timing! I was planning to ping you today :) Gray's PR to cover IPsec key rotation was merged this morning, so I thought now is probably as good a time as ever to merge this. We have some up/downgrade coverage (minus patch versions, minus error logs) and good key rotation coverage. We also extended the BPF and Golang unit tests. I think this doesn't need additional manual testing (the IPsec team will perform some before releases anyway). I'm also happy to review once it's ready for reviews!

gandro:
Image build failed due to an infra error:

```
 ERROR: failed to solve: failed to compute cache key: failed to copy: httpReadSeeker: failed open: unexpected status code https://quay.io/v2/cilium/cilium-runtime/blobs/sha256:3975c531ca1b6c65f021a92957af42ca1f01a690d3864fdeef48dbbe0da9d380: 502 Bad Gateway
Error: buildx failed with: ERROR: failed to solve: failed to compute cache key: failed to copy: httpReadSeeker: failed open: unexpected status code https://quay.io/v2/cilium/cilium-runtime/blobs/sha256:3975c531ca1b6c65f021a92957af42ca1f01a690d3864fdeef48dbbe0da9d380: 502 Bad Gateway
```

Restarting.

pkg/nodediscovery/nodediscovery.go
```diff
@@ -507,17 +507,16 @@ func (n *NodeDiscovery) mutateNodeResource(nodeResource *ciliumv2.CiliumNode) er
 		}
```
pchaigno, line 507: *(On commit description)*

(Not sure if it's worth repushing just for that.)

> In ClusterPool mode, the CiliumNode PodCIDR field is populated by cilium-operator. Therefore, we should try to overwrite it ourselves.

Did you mean "we shouldn't"?

> the previous code therefore wrote a auto-generated but otherwise IPv4/IPv6AllocRange into the field

I think there's a word missing here.

pkg/datapath/linux/node.go
```diff
@@ -1081,26 +1081,29 @@ func (n *linuxNodeHandler) enableIPsecIPv4(newNode *nodeTypes.Node, nodeID uint1
...
 
-	new4Net := newNode.IPv4AllocCIDR.IPNet
 	wildcardIP := net.ParseIP(wildcardIPv4)
 	wildcardCIDR := &net.IPNet{IP: wildcardIP, Mask: net.IPv4Mask(0, 0, 0, 0)}
 
 	err := ipsec.IPsecDefaultDropPolicy(false)
 	errs = errors.Join(errs, upsertIPsecLog(err, "default-drop IPv4", wildcardCIDR, wildcardCIDR, spi))
 
 	if newNode.IsLocal() {
-		errs = errors.Join(errs, n.replaceNodeIPSecInRoute(new4Net))
```
pchaigno, line 1096: IIUC, on EKS & AKS, we would install a route using the internal auto-generated IPv4/IPv6AllocRange. Do we have an idea of the potential consequences of that? The second commit suggests that the auto-generated CIDR could clash with real CIDR, although unlikely.

pchaigno (approved):
Thanks a lot for delaying this until we can have enough confidence into the IPsec CI and changes :pray: I know it's a pain to have to postpone PRs like that.

I have one comment below to try to understand the potential impact of the IPsec change. But regardless of that, I see the PR is marked as release-note/bug, but doesn't have a release note or a backport label. Outside of the IPsec part, is there any user-visible impact?

pkg/nodediscovery/nodediscovery.go
gandro: > Did you mean "we shouldn't"?

> I think there's a word missing here.

Yes, good catches! Will fix, thanks!

pkg/datapath/linux/node.go
gandro: So yes, before this PR installed a route for the auto-generated pod CIDR. It used an auto-generated pod CIDR, because that's Cilium's ancient fall-back if no pod CIDR is configured. That auto-generated pod CIDR is not used anywhere (see also https://github.com/cilium/cilium/issues/27018) and we're planning to eventually remove it, since it causes confusion for both engineers and users.

Particularly in EKS and AKS, IPs are not allocated from any pod CIDR (be it auto-generated or manually configured), and so using a node's pod CIDR in AKS or EKS is making assumptions about the IP range that does not hold true. By installing a route for each node's (auto-generated) pod CIDR (as we did before the PR), we routed all traffic belonging to those auto-generated CIDRs to the encrypt device.  

I'm not sure, what the consequence of this were. We have not received bug reports about this, I think mostly because the auto-generated CIDR is by design unlikely to clash with a real CIDR. Since the encrypt device is probably also the default egress device, the route might also not have affected any issues for CIDRs that were routable on the default egress device anyway.

In any case, my primary motivation for this PR is the clean up unconditional uses of "pod CIDR" in the agent code, because such code will not be compatible with pod-CIDR-less IPAM modes such as EKS/AKS or Multi-Pool. IPSec already has "subnet encryption" as a pod CIDR replacement, but for some reason it still unconditionally used the pod CIDR, which this PR attempts to fix.

gandro:
- ci-ipsec-e2e https://github.com/cilium/cilium/actions/runs/6010063414/job/16300738637 failed with `one.one.one.one:443` - I suspect this is a infra flake, though I'll investigate
- https://github.com/cilium/cilium/actions/runs/6010063515/job/16300786495 hit https://github.com/cilium/cilium/issues/27688, restarting

gandro:
I did a quick triage of the IPSec failiures. I don't see anything wrong, but if anything stands out @pchaigno let me know.

Sympton: The DNS lookup fails as follows:
```
[.] Action [no-policies/pod-to-world/http-to-one.one.one.one-0: cilium-test/client2-646b88fb9b-n948t (10.244.2.191) -> one.one.one.one-http (one.one.one.one:80)]
Tue, 29 Aug 2023 09:16:46 GMT   [.] Action [no-policies/pod-to-world/https-to-one.one.one.one-0: cilium-test/client2-646b88fb9b-n948t (10.244.2.191) -> one.one.one.one-https (one.one.one.one:443)]
Tue, 29 Aug 2023 09:16:46 GMT   ❌ command "curl -w %{local_ip}:%{local_port} -> %{remote_ip}:%{remote_port} = %{response_code} --silent --fail --show-error --output /dev/null --connect-timeout 2 --max-time 10 --retry 3 --retry-all-errors --retry-delay 3 [https://one.one.one.one:443](https://one.one.one.one/)" failed: command failed: curl: (28) Resolving timed out after 2000 milliseconds
```

It is a bit suspicious that it's not the connection to 1.1.1.1 that fails, but the DNS lookup. In Hubble, I don't see any dropped UDP port 53 traffic.

I also checked the routes in routing table 200 (the one affected by this PR) and it looks like it should on all three nodes, one route per remote node:

```
10.244.0.0/24 dev cilium_host proto kernel mtu 1450 
local 10.244.1.0/24 dev cilium_vxlan proto kernel scope host 
10.244.3.0/24 dev cilium_host proto kernel mtu 1450 
```

So I suspect this is just a flake where DNS somehow timed out, we've seen apparently similar flakes before: https://github.com/cilium/cilium/issues/26321

pkg/datapath/linux/node.go
pchaigno: I did a quick test on EKS to simulate the impact of a clash between the auto-generated CIDR and the pod subnet. Pod subnet was 192.168.0.0/16 and I manually installed a route like the one removed in this PR but with CIDR 192.168.128.0/17.

That caused some existing connections to be interrupted and prevented me from running the connectivity tests. I didn't debug further. There were no drops reported by Cilium or XFRM errors.

I have no idea how unlikely it is for the CIDRs to clash, but maybe we need to add a release note (something like below or can we make the "potential" more precise?) and backport to just v1.14?
```
Fix potential cross-node connectivity issue when IPsec is enabled with ENI or Azure IPAM modes.
```

pchaigno:
> So I suspect this is just a flake where DNS somehow timed out, we've seen apparently similar flakes before: #26321

I agree it's unlikely caused by this PR. Nevertheless, the fact that it happens in both cases in the key rotation section which we've just added makes me wonder if it's not a new flake caused by the key rotation changes. Might be worth opening a flake issue, just so we can track it.

pkg/datapath/linux/node.go
gandro: Awesome, thanks for that test :rocket: 

> I have no idea how unlikely it is for the CIDRs to clash, but maybe we need to add a release note (something like below or can we make the "potential" more precise?) and backport to just v1.14?

In the end, it depends on user configuration. [The auto-generated CIDR](https://github.com/cilium/cilium/blob/a8d80e25ca2d26affe4f7f7a8350bb537ddbe388/pkg/node/address.go#L123) is of the form "10.x.0.0/16" where x is derived from the node's IP. So I think clashes can happen - but I'm not aware of any bug reports, probably because most users are using IPs in the range of "10.0.0.0/16".

I think we can absolutely backport this to v1.14. I removed the backport label because there have not been bug reports and this has likely been present for years, but given the risk should be low, it does fit our backporting policy.

gandro:
> > So I suspect this is just a flake where DNS somehow timed out, we've seen apparently similar flakes before: #26321
> 
> I agree it's unlikely caused by this PR. Nevertheless, the fact that it happens in both cases in the key rotation section which we've just added makes me wonder if it's not a new flake caused by the key rotation changes. Might be worth opening a flake issue, just so we can track it.

I've created an issue. Pretty sure it's unrelated since I also found scheduled runs with the same error. https://github.com/cilium/cilium/issues/27799

---

Triage of new failures:

[Setup & Test (1, 4.19-20230810.091425, iptables, false, vxlan, ipsec, false)](https://github.com/cilium/cilium/actions/runs/6011556981/job/16305158986#logs)

```
   [.] Action [echo-ingress-knp/pod-to-pod/curl-ipv6-3: cilium-test/client2-646b88fb9b-g945g (fd00:10:244:3::9397) -> cilium-test/echo-same-node-59dc776cf5-qhn8t (fd00:10:244:3::7973:8080)]
  ❌ command "curl -w %{local_ip}:%{local_port} -> %{remote_ip}:%{remote_port} = %{response_code} --silent --fail --show-error --output /dev/null --connect-timeout 2 --max-time 10 http://[fd00:10:244:3::7973]:8080" failed: command terminated with exit code 28
```

There are some weird RSTs coming `from-host` with the host identity. This is weird:

```
Aug 29 11:49:03.669: cilium-test/client2-646b88fb9b-g945g:49480 (ID:4992) <> cilium-test/echo-same-node-59dc776cf5-qhn8t:8080 (ID:41938) from-endpoint FORWARDED (TCP Flags: SYN)
Aug 29 11:49:03.669: cilium-test/client2-646b88fb9b-g945g:49480 (ID:4992) -> cilium-test/echo-same-node-59dc776cf5-qhn8t:8080 (ID:41938) policy-verdict:L3-Only INGRESS ALLOWED (TCP Flags: SYN)
Aug 29 11:49:03.669: cilium-test/client2-646b88fb9b-g945g:49480 (ID:4992) -> cilium-test/echo-same-node-59dc776cf5-qhn8t:8080 (ID:41938) to-endpoint FORWARDED (TCP Flags: SYN)
Aug 29 11:49:03.669: cilium-test/echo-same-node-59dc776cf5-qhn8t:8080 (ID:41938) <> cilium-test/client2-646b88fb9b-g945g:49480 (ID:4992) from-endpoint FORWARDED (TCP Flags: SYN, ACK)
Aug 29 11:49:03.669: cilium-test/client2-646b88fb9b-g945g:49480 (host) <> cilium-test/echo-same-node-59dc776cf5-qhn8t:8080 (ID:41938) from-host FORWARDED (TCP Flags: RST)
Aug 29 11:49:03.669: cilium-test/client2-646b88fb9b-g945g:49480 (ID:4992) -> cilium-test/echo-same-node-59dc776cf5-qhn8t:8080 (ID:41938) to-endpoint FORWARDED (TCP Flags: RST)
```

---

[Setup & Test (2, 5.4-20230810.091425, iptables, false, disabled, ipsec, false)](https://github.com/cilium/cilium/actions/runs/6011556981/job/16305159139#logs)

```
[.] Action [echo-ingress-mutual-auth-spiffe/pod-to-pod/curl-ipv4-2: cilium-test/client2-646b88fb9b-vgmc9 (10.244.2.188) -> cilium-test/echo-other-node-f884df9df-cf7mg (10.244.1.142:8080)]
  ❌ command "curl -w %{local_ip}:%{local_port} -> %{remote_ip}:%{remote_port} = %{response_code} --silent --fail --show-error --output /dev/null --connect-timeout 2 --max-time 10 http://10.244.1.142:8080/" failed: command terminated with exit code 28
```

There are no RSTs in this trace (at least not related to IP `10.244.2.188`), but I see that the the connection eventually succeeds, so I'm not sure why the test failed:

```
Aug 29 11:35:27.757: cilium-test/client2-646b88fb9b-vgmc9:33116 (ID:22938) <> cilium-test/echo-other-node-f884df9df-cf7mg:8080 (ID:10042) to-host FORWARDED (TCP Flags: SYN)
Aug 29 11:35:27.757: cilium-test/client2-646b88fb9b-vgmc9:33116 (ID:9) <> cilium-test/echo-other-node-f884df9df-cf7mg:8080 (ID:10042) from-host FORWARDED (TCP Flags: SYN)
Aug 29 11:35:27.757: cilium-test/client2-646b88fb9b-vgmc9:33116 (ID:22938) <> cilium-test/echo-other-node-f884df9df-cf7mg:8080 (ID:10042) policy-verdict:L3-L4 INGRESS DENIED (TCP Flags: SYN; Auth: SPIRE)
Aug 29 11:35:27.757: cilium-test/client2-646b88fb9b-vgmc9:33116 (ID:22938) <> cilium-test/echo-other-node-f884df9df-cf7mg:8080 (ID:10042) Authentication required DROPPED (TCP Flags: SYN)
Aug 29 11:35:28.784: cilium-test/client2-646b88fb9b-vgmc9:33116 (ID:22938) <> cilium-test/echo-other-node-f884df9df-cf7mg:8080 (ID:10042) from-stack FORWARDED (TCP Flags: SYN)
Aug 29 11:35:28.784: cilium-test/client2-646b88fb9b-vgmc9:33116 (ID:22938) <> cilium-test/echo-other-node-f884df9df-cf7mg:8080 (ID:10042) to-host FORWARDED (TCP Flags: SYN)
Aug 29 11:35:28.784: cilium-test/client2-646b88fb9b-vgmc9:33116 (ID:9) <> cilium-test/echo-other-node-f884df9df-cf7mg:8080 (ID:10042) from-host FORWARDED (TCP Flags: SYN)
Aug 29 11:35:28.784: cilium-test/client2-646b88fb9b-vgmc9:33116 (ID:22938) -> cilium-test/echo-other-node-f884df9df-cf7mg:8080 (ID:10042) to-endpoint FORWARDED (TCP Flags: SYN)
Aug 29 11:35:28.784: cilium-test/echo-other-node-f884df9df-cf7mg:8080 (ID:10042) <> cilium-test/client2-646b88fb9b-vgmc9:33116 (ID:22938) from-endpoint FORWARDED (TCP Flags: SYN, ACK)
Aug 29 11:35:28.784: cilium-test/client2-646b88fb9b-vgmc9:33116 (ID:22938) <- cilium-test/echo-other-node-f884df9df-cf7mg:8080 (ID:10042) to-stack FORWARDED (TCP Flags: SYN, ACK)
```

gandro:
Given that I don't see anything wrong with the routes, and both symptoms have been observed before (see links), I'll restart again. I don't know how to otherwise troubleshoot those failures further

PR #26674, opened 2023-07-06
bpf: snat: DSR-eligible traffic can skip check for Nodeport NAT conflict

julianwiedmann:
Right now the SNAT logic in `to-overlay` / `to-netdev` tracks all (host-originating) connections that potentially conflict with NAT connections by the N/S LB to remote backends (ie. `tail_nodeport_nat_egress_ipv4()`).

If the LB uses DSR for the packet's L4 protocol type, there's no risk of conflicting with a NATed LB connection. So we don't need to track such a connection.

### DIFF
--- a/bpf/lib/nodeport.h
+++ b/bpf/lib/nodeport.h
@@ -150,6 +150,7 @@ static __always_inline int nodeport_snat_fwd_ipv6(struct __ctx_buff *ctx,
 	l4_off = ETH_HLEN + hdrlen;
 
 	if (lb_is_svc_proto(tuple.nexthdr) &&
+	    !nodeport_uses_dsr6(&tuple) &&
 	    nodeport_has_nat_conflict_ipv6(ip6, &target))
 		goto apply_snat;
 
@@ -1529,6 +1530,7 @@ static __always_inline int nodeport_snat_fwd_ipv4(struct __ctx_buff *ctx,
 	l4_off = ETH_HLEN + ipv4_hdrlen(ip4);
 
 	if (lb_is_svc_proto(tuple.nexthdr) &&
+	    !nodeport_uses_dsr4(&tuple) &&
 	    nodeport_has_nat_conflict_ipv4(ip4, &target))
 		goto apply_snat;
 
--- a/bpf/lib/nodeport.h
+++ b/bpf/lib/nodeport.h
@@ -149,7 +149,8 @@ static __always_inline int nodeport_snat_fwd_ipv6(struct __ctx_buff *ctx,
 	snat_v6_init_tuple(ip6, NAT_DIR_EGRESS, &tuple);
 	l4_off = ETH_HLEN + hdrlen;
 
-	if (nodeport_has_nat_conflict_ipv6(ip6, &target))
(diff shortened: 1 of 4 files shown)


bpf/lib/nodeport.h
```diff
@@ -1504,13 +1516,21 @@ static __always_inline int nodeport_snat_fwd_ipv4(struct __ctx_buff *ctx,
...
+	struct ipv4_ct_tuple tuple = {};
 	void *data, *data_end;
 	struct iphdr *ip4;
-	int ret;
+	int l4_off, ret;
 
 	if (!revalidate_data(ctx, &data, &data_end, &ip4))
 		return DROP_INVALID;
 
+	tuple.flags = TUPLE_F_OUT;
```
gentoo-root: Nit: why not use snat_v4_init_tuple?

bpf/lib/nodeport.h
```diff
@@ -1424,6 +1465,32 @@ static __always_inline bool nodeport_uses_dsr4(const struct ipv4_ct_tuple *tuple
...
+{
+#if defined(TUNNEL_MODE) && defined(IS_BPF_OVERLAY)
+	if (ip4->saddr == IPV4_GATEWAY) {
+		target->addr = IPV4_GATEWAY;
+		return true;
+	}
+#endif /* TUNNEL_MODE && IS_BPF_OVERLAY */
+
+#if defined(IS_BPF_HOST)
+    /* NATIVE_DEV_IFINDEX == DIRECT_ROUTING_DEV_IFINDEX cannot be moved into
```
gentoo-root: Nit: tab for indent.

bpf/lib/nodeport.h
```diff
@@ -1531,7 +1532,8 @@ static __always_inline int nodeport_snat_fwd_ipv4(struct __ctx_buff *ctx,
 
 	l4_off = ETH_HLEN + ipv4_hdrlen(ip4);
 
-	if (nodeport_has_nat_conflict_ipv4(ip4, &target))
+	if (lb_is_svc_proto(tuple.nexthdr) && !nodeport_uses_dsr4(&tuple) &&
```
gentoo-root: Could you please elaborate why we skip apply_snat for ICMP?

bpf/lib/nodeport.h
julianwiedmann: Hmm can't think of a good reason. Thanks, that also reduces the churn in the tests quite a bit.

bpf/lib/nodeport.h
julianwiedmann: Ack. While at it I also added a temporary variable that holds `DIRECT_ROUTING_DEV_IFINDEX` (to appease checkpatch).

bpf/lib/nodeport.h
julianwiedmann: Thanks for pointing that out :). I should do even better, and split that change off into a separate patch. Done so now.

We're not forwarding ICMP packets from the LB to a backend, so there's no risk of conflicting with such traffic.

gentoo-root (approved):

PR #26682, opened 2023-07-07
docs: Fix unintentional boolean value in YAML

dgl:
This used an unquoted "yes", which becomes boolean in YAML, it should be a string as it's a label value.

Fixes: 959fb0d74c4

Fixes warning from make:

```
/src/examples/policies/l3/service/service-labels.yaml
  14:23     warning  truthy value should be one of [false, true]  (truthy)
```

### DIFF
--- a/examples/policies/l3/service/service-labels.yaml
+++ b/examples/policies/l3/service/service-labels.yaml
@@ -11,4 +11,4 @@ spec:
     - k8sServiceSelector:
         selector:
           matchLabels:
-            external: yes
+            external: "yes"


dgl:
cc @odinuge

odinuge (approved):
Ooh, that was news to me 😅 Thanks for catching!

ghost (approved):
nice catch

squeed:
(I'm running CI because some of these examples are used in tests, so we need to make sure this didn't somehow double-unfix something)

PR #26683, opened 2023-07-07
opt cmdDel: ignore pod's netns has gone

cyclinder:

      GitHub issue.

opt cmdDel: ignore pod's netns has gone

```release-note
```

### DIFF
--- a/plugins/cilium-cni/main.go
+++ b/plugins/cilium-cni/main.go
@@ -438,6 +438,10 @@ func cmdAdd(args *skel.CmdArgs) (err error) {
 
 	netNs, err := ns.GetNS(args.Netns)
 	if err != nil {
+		if errors.As(err, &ns.NSPathNotExistErr{}) {
+			logger.Errorf("Pod's netns %s has gone, nothing to do", args.Netns)
+			return nil
+		}
 		return fmt.Errorf("failed to open netns %q: %w", args.Netns, err)
 	}
 	defer netNs.Close()


squeed (changes requested):
This is a good idea, but I think we can do one better.

Rather than trying to interpret every error message, the code should do the following:

1. Submit the deletion request to the server. If that fails, return error
2. Try and open the netns and delete the interface. If any of that fails, give up and return no error.

Make sense? Thanks!

cyclinder:
> 2. Try and open the netns and delete the interface. If any of that fails, give up and return no error.

I don't understand this, Why don't return an error if any of that fails? Which cause the pod to be deleted successfully, but its netns resources not be cleaned up, right?

squeed:
> > 3. Try and open the netns and delete the interface. If any of that fails, give up and return no error.
> 
> I don't understand this, Why don't return an error if any of that fails? This causes the pod to be deleted successfully, but its netns resources not be cleaned up, right?

Good question! No, deleting the netns itself is handled by the container runtime. First, however, the runtime requests that CNI plugins perform any cleanup.

The most important thing to do is clean up any external allocations, e.g. IP reservations. Then, it is considered a good practice to remove the interface in the netns. However, it is OK if this fails since the netns itself is going away, and the kernel will take care of deleting anything (routes, interfaces, etc) left behind.

So, it's 100% safe to continue if the netns doesn't exist. The right thing to do would be to clean up as much as possible and swallow any errors.

cyclinder:
> So, it's 100% safe to continue if the netns doesn't exist. The right thing to do would be to clean up as much as possible and swallow any errors.

I've developed some CNI meta-plugins, such as [coordinator](https://github.com/spidernet-io/spiderpool/tree/main/cmd/coordinator),This plugin helps coordinate policy route between interfaces to ensure consistence data path of request and reply packets when a pod is attached to multiple networks via multus.  It also create some routing table or neighbor table into the host. 

In the beginning, the CmdDel function did not do any cleanup, I believe that when the pod is deleted, its netns are reclaimed, and all netns resources will be purged. In the testing and R&D environment, there are no problems. When a pod is deleted, all netns are properly recycled.

But the other day in our production environment, we encountered pods deleted but netns resources were not properly reclaimed, I confirmed that netns have been recycled, causing pods to be unable to access the node they are on. 

I had to rethink the logic of CmdDel.  So I opened this, In the case of uncertain root cause, I think CmdDel should ignore errors like non-existent and then return an error for the runtime to retry the deletion. And we can also try to retry the deletion in CmdDel. This is my opinion, How do you think?

cyclinder:
Hi @squeed, How do you think about my comment https://github.com/cilium/cilium/pull/26683#issuecomment-1625353428? Thanks a lot.

squeed:
> > So, it's 100% safe to continue if the netns doesn't exist. The right thing to do would be to clean up as much as possible and swallow any errors.
>
> But the other day in our production environment, we encountered pods deleted but netns resources were not properly reclaimed, I confirmed that netns have been recycled, causing pods to be unable to access the node they are on.
> 
> I had to rethink the logic of CmdDel. So I opened this, In the case of uncertain root cause, I think CmdDel should ignore errors like non-existent and then return an error for the runtime to retry the deletion. And we can also try to retry the deletion in CmdDel. This is my opinion, How do you think?

What do you meant the netns wasn't properly reclaimed? As in, the network namespace continued to exist and there were still processes running in it? While that may be a problematic scenario, it's hard to imagine how CNI plugins should somehow help prevent that.

Network namespace management is not the responsibility of the CNI plugin. If the namespace provided for CNI DEL does not exist, the CNI spec is quite clear that the plugin should return success.

Out of curiosity, what runtime were you using? Containerd? CRI-O? Something else?

squeed (approved):

pkg/netns/netns.go
```diff
@@ -23,7 +22,7 @@ func RemoveIfFromNetNSIfExists(netNS ns.NetNS, ifName string) error {
 	return netNS.Do(func(_ ns.NetNS) error {
 		l, err := netlink.LinkByName(ifName)
 		if err != nil {
-			if errors.As(err, &netlink.LinkNotFoundError{}) {
+			if _, ok := err.(netlink.LinkNotFoundError); ok {
```
squeed: why did you change the `errors.As`? Was it not working as intended?

cyclinder:
> What do you meant the netns wasn't properly reclaimed? As in, the network namespace continued to exist and there were still processes running in it? 

Yes, I think there were still processes running in it, But I used `ip netns list` to check if the network namespace exists, and I don't find it indeed.  I don't know which process is still using this network namespace.

> Network namespace management is not the responsibility of the CNI plugin. If the namespace provided for CNI DEL does not exist, the CNI spec is quite clear that the plugin should return success.

So I think we should  delete network namespace resources as much as possible, rather than relying entirely on network namespace to be removed. Of course, we should also ignore some non-existent errors, and return meaningful errors,make the runtime retry the deletion.
 
> Out of curiosity, what runtime were you using? Containerd? CRI-O? Something else?

We using Containerd as runtime.

pkg/netns/netns.go
cyclinder: ah, I don't remember it, but you're right. updated now.

joamaki (approved):

plugins/cilium-cni/main.go
```diff
@@ -438,6 +438,10 @@ func cmdAdd(args *skel.CmdArgs) (err error) {
 
 	netNs, err := ns.GetNS(args.Netns)
 	if err != nil {
+		if errors.As(err, &ns.NSPathNotExistErr{}) {
```
ti-mo, line 441: Let me get this straight: if we don't manage to open the netns pin from nsfs, we return from `cmdAdd` successfully as if a CNI ADD was successfully completed? I don't think that's a good idea, this signals the container runtime that everything's peachy. Which problem does this patch solve, exactly?

ti-mo (changes requested):
> Yes, I think there were still processes running in it, But I used `ip netns list` to check if the network namespace exists, and I don't find it indeed. I don't know which process is still using this network namespace.

Namespaces are reclaimed when the last process using them exits. That's the only condition. I'm not sure what you were trying to find with `ip netns list` but that command scans `/var/run/ns` for netns handles, which the majority of modern container runtimes do not write to. The canonical way of enumerating all namespace handles on the system is by iterating `/proc/*/ns` in the root namespace.

>> Network namespace management is not the responsibility of the CNI plugin. If the namespace provided for CNI DEL does not exist, the CNI spec is quite clear that the plugin should return success.

> So I think we should delete network namespace resources as much as possible, rather than relying entirely on network namespace to be removed. Of course, we should also ignore some non-existent errors, and return meaningful errors,make the runtime retry the deletion.

Which network namespace resources do you mean here? And how does the proposed patch accomplish this? Cilium isn't going to directly kill processes, so it has zero impact on the lifecycle of a netns managed by the container runtime (short of returning an unsuccessful CNI ADD, of course).

joestringer:
At minimum, it would help a lot to refresh the PR's title, description, commit title and message to describe exactly what problem this is solving and why the proposed approach is a good way to solve it. Right now the PR describes doing something with `cmdDel` but the code is changing `cmdAdd`. Clearly there's still some ongoing discussion in the threads above as well which will need to be resolved before this PR can proceed.

cyclinder:
Thank you all for your comments, I think I made a mistake 😅. My intention was to check if Netns doesn't exist when CmdDel is an error, but I made the code changes for CmdAdd. I just confirmed the logic of CmdDel on the main branch and I think that's what I want to do, so I'm going to close this PR. Sorry again and thank you all.

cyclinder:
> Namespaces are reclaimed when the last process using them exits. That's the only condition. I'm not sure what you were trying to find with `ip netns list` but that command scans `/var/run/ns` for netns handles, which the majority of modern container runtimes do not write to. The canonical way of enumerating all namespace handles on the system is by iterating `/proc/*/ns` in the root namespace.

When I use `containerd` as CRI, the CmdArgs.Netns parameter passed to CNI is "/var/run/..."(docker is /pro/<Pid>/ns/net), so I use the 'ip netns list' to detect if this netns exists. 

> Which network namespace resources do you mean here? 

Mainly veth device on the host side.  Actually, I wanted to clean up the lxc* device on the host side when netns didn't exist, which seemed redundant, but we ran into netns not existing (/var/run/... does not exist), but the lxc* on the host still exists, but the pod has been deleted.

plugins/cilium-cni/main.go
cyclinder: Sorry, Ignore this please, this is my mistake.

PR #26684, opened 2023-07-07
docs: optimize ingress default tls secret documentation

mhofstetter:
Currently, the example Cilium CLI command to configure ingress with a default tls secret doesn't contain the flag to enable the ingress functionality. That's why the command actually doesn't configure the default secret too.

Therefore, this commit will add the missing flag to the command.


ghost (approved):

meyskens (approved):
LGTM

mhofstetter:
simple docs change only - marking as `ready-to-merge`

PR #26685, opened 2023-07-07
test/controlplane: Fix hostport test after API change

pippolo84:
Commit https://github.com/cilium/cilium/pull/25986/commits/9c00b80a9bed4824a508b8986a969631f4161a7e changed the controlplane tests API to account for the removal of the global daemon hive.  The Pod/HostPort test should then be changed according to the new API.

Related: https://github.com/cilium/cilium/pull/25986
Fixes: https://github.com/cilium/cilium/pull/22587

/cc @yasz24

### DIFF
--- a/test/controlplane/pod/hostport/hostport.go
+++ b/test/controlplane/pod/hostport/hostport.go
@@ -8,8 +8,7 @@ import (
 	"path"
 	"testing"
 
-	operatorOption "github.com/cilium/cilium/operator/option"
-	agentOption "github.com/cilium/cilium/pkg/option"
+	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/test/controlplane"
 	"github.com/cilium/cilium/test/controlplane/services/helpers"
 	"github.com/cilium/cilium/test/controlplane/suite"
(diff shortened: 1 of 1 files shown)


mhofstetter (approved):

julianwiedmann:
Adjusting labels, as it's not fixing a bug in a released version. Not sure about the backport - https://github.com/cilium/cilium/pull/22587 is tagged, but the conflicting PR (https://github.com/cilium/cilium/pull/25986) is not.

aanm (approved):

borkmann (approved):

borkmann:
Wrt backport, given the https://github.com/cilium/cilium/pull/22587 is a fix, I would love to have it in 1.14 backported.

pippolo84:
> Adjusting labels, as it's not fixing a bug in a released version. Not sure about the backport - #22587 is tagged, but the conflicting PR (#25986) is not.

I think it should **not** be backported, since #25986 should not land in v1.14.

borkmann:
> I think this should **not** be backported, since #25986 should not land in v1.14.

In that case we could probably ask the author to do a custom backport of the fix.

pippolo84:
> In that case we could probably ask the author to do a custom backport of the fix.

I think that #22587 can be backported to v1.14 as is, without leading to a failure in the controlplane test. That's because v1.14 has the "old" version of the controlplane tests API, so, in that case, there's no need for this fix :+1:

PR #26706, opened 2023-07-07
ipam: remove always-nil NewCIDRRange error return value

tklauser:
NewCIDRRange will never return an error, so omit the error return value to simplify some call sites.

Also fold the implementations of NewCIDRRange and NewAllocatorCIDRRange as the latter is only called by the former.


christarazi (approved):

gandro (approved):

tommyp1ckles (approved):
looks good

PR #26708, opened 2023-07-07
Fix upgrade for IPsec with tunneling

pchaigno:
Fix various packet drops impacting upgrades when running with IPsec and overlay mode. See commits for details.

```release-note
Fix bug that caused transient IPsec packet drops on upgrades when tunneling is enabled. 
```

### DIFF
--- a/bpf/lib/encap.h
+++ b/bpf/lib/encap.h
@@ -127,12 +127,12 @@ __encap_and_redirect_lxc(struct __ctx_buff *ctx, __be32 tunnel_endpoint,
 						seclabel);
 #endif
 
-#if !defined(ENABLE_NODEPORT) && (defined(ENABLE_IPSEC) || defined(ENABLE_HOST_FIREWALL))
-	/* For IPSec and the host firewall, traffic from a pod to a remote node
-	 * is sent through the tunnel. In the case of node --> VIP@remote pod,
-	 * packets may be DNATed when they enter the remote node. If kube-proxy
-	 * is used, the response needs to go through the stack on the way to
-	 * the tunnel, to apply the correct reverse DNAT.
+#if !defined(ENABLE_NODEPORT) && defined(ENABLE_HOST_FIREWALL)
+	/* For the host firewall, traffic from a pod to a remote node is sent
+	 * through the tunnel. In the case of node --> VIP@remote pod, packets may
+	 * be DNATed when they enter the remote node. If kube-proxy is used, the
+	 * response needs to go through the stack on the way to the tunnel, to
+	 * apply the correct reverse DNAT.
 	 * See #14674 for details.
 	 */
 	ret = __encap_with_nodeid(ctx, 0, 0, tunnel_endpoint, seclabel, dstid,
@@ -146,7 +146,7 @@ __encap_and_redirect_lxc(struct __ctx_buff *ctx, __be32 tunnel_endpoint,
 #else
 	return __encap_and_redirect_with_nodeid(ctx, 0, tunnel_endpoint,
 						seclabel, dstid, NOT_VTEP_DST, trace);
-#endif /* !ENABLE_NODEPORT && (ENABLE_IPSEC || ENABLE_HOST_FIREWALL) */
+#endif /* !ENABLE_NODEPORT && ENABLE_HOST_FIREWALL */
 }
 
 #if defined(TUNNEL_MODE) || defined(ENABLE_HIGH_SCALE_IPCACHE)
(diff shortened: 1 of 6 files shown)


brb:
Nice find! Thanks for the PR.

I'm a bit puzzled because the following part:
```
None of this is an issue on AKS and ENI IPAM modes because there, the
XFRM OUT policies look like:

src 0.0.0.0/0 dst 10.0.0.0/16
    dir out priority 0 ptype main
    mark 0x66d11e00/0xffffff00
    tmpl src 10.0.1.13 dst 10.0.0.214
        proto esp spi 0x00000001 reqid 1 mode tunnel

Thus, hostns -> remote pod traffic is matched regardless of the source
IP being selected and packets are not dropped by the default-drop rule.

We can therefore avoid the upgrade drops by changing the XFRM OUT
policies to never match on the source IPs, as on AKS and ENI IPAM modes.
```

Are you saying that after applying this commit we will start encrypting host netns -> remote pod packets?

jschwinger233:
The traffic from hostns to remote pod doesn't get a `0xe00/0xf00` mark, right? Previously it was marked by host_bpf on cilium_host, but you removed that on https://github.com/cilium/cilium/commit/5fe2b2d6da76d6b1d334f3ce9c0c59371b239892.

On the other hand, xfrm on the host has default-drop policy:
```
src 0.0.0.0/0 dst 0.0.0.0/0 
	dir out action block priority 100 
	mark 0xe00/0xf00 
```

The `mark 0xe00/0xf00` in the policy means, xfrm blocks out-dir IPv4 only if skb has mark `0xe00`, does it? (Can't access my IPsec note on the isovalent slack :crying_cat_face: ) If my understanding is correct, the hostns -> remote pod traffic shouldn't be blocked by xfrm default-drop policy, regardless of the source IP? Do I miss something?

pchaigno:
@brb @jschwinger233 I think you're both puzzled by the same thing. You have to consider all of this is to fix the upgrade. During the upgrade, we get into a weird, intermediate state where the catch-all default-drop rule exist but the hostns <> pod encryption hasn't been removed yet. That is, the XFRM OUT policy was added, but the BPF programs were not reloaded yet nor was the ipcache updated.

This weird state happens mostly on purpose: we want to install the default-drop rule before everything else. Upgrades are complex and we want to ensure we will not leak plain-text traffic, whatever else happens.

Now to answer each of you:

> Are you saying that after applying this commit we will start encrypting host netns -> remote pod packets?

@brb Yes, we will encrypt hostns -> remote pod traffic during the upgrade, to avoid dropping that traffic. We will do so as soon as the XFRM config is installed and until the upgrade is finished. Once the upgrade is finished, the new BPF programs and new ipcache content will be loaded and hostns <> pod traffic won't be encrypted anymore.

> The traffic from hostns to remote pod doesn't get a `0xe00/0xf00` mark, right? Previously it was marked by host_bpf on cilium_host, but you removed that on https://github.com/cilium/cilium/commit/5fe2b2d6da76d6b1d334f3ce9c0c59371b239892.

@jschwinger233 During the upgrade, there will be a time when the new XFRM config is used, but https://github.com/cilium/cilium/commit/5fe2b2d6da76d6b1d334f3ce9c0c59371b239892 hasn't taken effect yet. Hence, hostns -> remote pod traffic will have the `0xe00` mark and will match XFRM OUT policies.

jschwinger233 (approved):
IPsec upgrade is subtle, I think your split cluster test is valuable to add into the ci, instead of counting on the random upgrade order to catch bugs, which may be wrongly seen as flakes because they could be gone on re-run.

jrfastab (approved):
nice cleanup/fixes.

brb (approved):

pchaigno:
This passed CI, has been reviewed by three engineers with datapath and IPsec experience, and was tested manually with an upgrade and a scale up & down. The changes that are not covered by codeowners are for minor agent changes that are very specific to IPsec. Marking ready to merge.

julianwiedmann:
> This passed CI, has been reviewed by three engineers with datapath and IPsec experience, and was tested manually with an upgrade and a scale up & down. The changes that are not covered by codeowners are for minor agent changes that are very specific to IPsec. Marking ready to merge.

Agree :+1:. Is it sad that I'm most excited about the `lib/encap.h` cleanup? :grin:

pchaigno:
> > This passed CI, has been reviewed by three engineers with datapath and IPsec experience, and was tested manually with an upgrade and a scale up & down. The changes that are not covered by codeowners are for minor agent changes that are very specific to IPsec. Marking ready to merge.
> 
> Agree +1. Is it sad that I'm most excited about the `lib/encap.h` cleanup? grin

@julianwiedmann That's how you ended up being "the encapsulation expert" :rofl:

PR #26713, opened 2023-07-07
Bump allowed Golang version for v1.11 and v1.12

ferozsalam:
Allow v1.11 and v1.12 to use a supported Golang version, which will remove several CVEs
currently being reported by scanners.

Bumping based on the recently added documentation in: https://docs.cilium.io/en/latest/contributing/development/dev_setup/#minor-version

Signed-off-by: Feroz Salam <feroz.salam@isovalent.com>

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -230,7 +230,7 @@
         "docker.io/library/golang",
         "go"
       ],
-      "allowedVersions": "<1.19",
+      "allowedVersions": "<1.20",
       "matchBaseBranches": [
         "v1.12"
       ]
@@ -240,7 +240,7 @@
         "docker.io/library/golang",
         "go"
       ],
-      "allowedVersions": "<1.18",
+      "allowedVersions": "<1.20",
       "matchBaseBranches": [
         "v1.11"
       ]


bimmlerd:
I think there was a previous discussion at some point about this, a bump from Go 1.17 -> Go 1.20 is a massive change in terms of potential regressions - I think this deserves wider discussion. I'll raise a thread with the maintainers; maybe you could talk about this in a community meeting as well?

ferozsalam:
For some extra context, I made this change off the back of a discussion with @aanm @michi-covalent and @joestringer regarding the CVEs in older versions of Golang. The consensus was that we wanted to bump the version of Golang on maintained branches to a supported version.

I chose 1.20 because 1.19 is about to fall out of support, so 1.20 gives a longer runway for future releases.

michi-covalent:
/test-backport-1.11

michi-covalent (approved):

bimmlerd (approved):

akhilles:
There's an issue with Cilium v1.11 + Go 1.20 due to a breaking change. From Go 1.20 [release notes](https://tip.golang.org/doc/go1.20):
> Attempts to read from a SHT_NOBITS section using Section.Data or the reader returned by Section.Open now return an error.

which leads to:
```
level=warning msg="Error: error starting map migration for \"bpf_overlay.o\": file bpf_overlay.o: load data sections: data section .bss: can't get contents: unexpected read from SHT_NOBITS section" subsys=datapath-loader
```

This should be addressed by https://github.com/cilium/ebpf/pull/740/commits/0393df6446f405dfe1afaf7eadc47e75324bcc49. So, it'll require a cilium/ebpf to be bumped to >= 0.9.2.

ferozsalam:
Hi @akhilles – the change in this PR allows versions up to but not including Golang 1.20 ( '<1.20' as opposed to '<=1.20').

The v1.11 branch uses Golang v1.19 now as a result: https://github.com/cilium/cilium/commit/ad34c5a374e6dea863cabceb22c2aeb9f4491703

Did you see this error when testing the v1.11 branch, or were you testing v1.11 with Golang v1.20?

I will check if v1.12 and v1.13 might be affected by this – we will be bumping them to Golang v1.20 soon, so it's possible they will need the cilium/ebpf bump.

akhilles:
> the change in this PR allows versions up to but not including Golang 1.20 ( '<1.20' as opposed to '<=1.20').

Ah, missed that :). Since Go 1.19 is EOL now, should Go 1.20 be allowed for v1.11 to fix any future vulnerabilities?

> Did you see this error when testing the v1.11 branch, or were you testing v1.11 with Golang v1.20?

We saw this error on v1.11 +  Go 1.20. I don't think v1.12 and v1.13 are affected because they're already using `cilium/ebpf` >= 0.9.2.

joestringer:
Cilium v1.11 is also EOL now.

PR #26714, opened 2023-07-07
docs: fix minor TOC issues

networkop:
* The first issue is that lb-ipam is present in both BGP and Networking TOC. This change removes it from the top-level Networking TOC
* The second issue is the Kubernetes compatibility TOC displayed a subsection (CRD validation) at the same level as its parent.


ghost (approved):

joamaki (approved):

joamaki:
Marking ready-to-merge as doc builds pass.

PR #26715, opened 2023-07-07
ariane: don't skip verifier and l4lb tests on vendor/ changes

tklauser:
Both of these workflows use binaries that are built in CI making use of various vendored dependencies, so run them as well on PRs only changing vendor/.


nbusseneau (approved):
Blind approve.

nbusseneau:
Should we backport? The L4LB test is available on all stable branches, and the verifier test is also there on 1.14.

tklauser:
> Should we backport? The L4LB test is available on all stable branches, and the verifier test is also there on 1.14.

Yeah, good point. I'll add the respective labels.

bimmlerd (approved):

PR #26719, opened 2023-07-07
services: Fix metric from not publishing

### LINKED ISSUE #26511: Missing `services_events_total` metric

### What happened?

I deployed cilium 1.13.4 on kubernetes.
I am scraping both the operator and the agents with prometheus.
I deployed the default grafana dashboard.

I don't have data for one chart, turns out the `cilium_services_events_total` is not exposed by either components.

Looking at the code it looks like the metric should always be exposed

### Cilium Version

1.13.4

### Kernel Version

5.15.111-flatcar

### Kubernetes Version

v1.24.15

### Sysdump

_No response_

### Relevant log output

```shell
# curl -s localhost:9962/metrics|grep services_events_total
#
```

```
# curl -s localhost:9963/metrics|grep services_events_total
#
```
```


### Anything else?

it's not about the metric being without data, the description is not there in the exporter outputs.

### Code of Conduct

### PR DESCRIPTION
**Pull request by christarazi**
- metrics, service: Rename metric for clarity
- services: Fix metric from not publishing

```release-note
Fix missing metric "cilium_services_events_total"
```

Fixes: https://github.com/cilium/cilium/issues/26511

### DIFF
--- a/pkg/service/service.go
+++ b/pkg/service/service.go
@@ -34,9 +34,9 @@ import (
 const anyPort = "*"
 
 var (
-	updateMetric = metrics.ServicesCount.WithLabelValues("update")
-	deleteMetric = metrics.ServicesCount.WithLabelValues("delete")
-	addMetric    = metrics.ServicesCount.WithLabelValues("add")
+	updateMetric = metrics.ServicesEventsCount.WithLabelValues("update")
+	deleteMetric = metrics.ServicesEventsCount.WithLabelValues("delete")
+	addMetric    = metrics.ServicesEventsCount.WithLabelValues("add")
 )
 
 // ErrLocalRedirectServiceExists represents an error when a Local redirect
--- a/pkg/service/service.go
+++ b/pkg/service/service.go
@@ -33,12 +33,6 @@ import (
 
 const anyPort = "*"
 
-var (
-	updateMetric = metrics.ServicesEventsCount.WithLabelValues("update")
-	deleteMetric = metrics.ServicesEventsCount.WithLabelValues("delete")
-	addMetric    = metrics.ServicesEventsCount.WithLabelValues("add")
-)
-
 // ErrLocalRedirectServiceExists represents an error when a Local redirect
 // service exists with the same Frontend.
 type ErrLocalRedirectServiceExists struct {
@@ -719,9 +713,9 @@ func (s *Service) upsertService(params *lb.SVC) (bool, lb.ID, error) {
 	}
 
 	if new {
-		addMetric.Inc()
+		metrics.ServicesEventsCount.WithLabelValues("add").Inc()
 	} else {
-		updateMetric.Inc()
+		metrics.ServicesEventsCount.WithLabelValues("update").Inc()
 	}
 
 	s.notifyMonitorServiceUpsert(svc.frontend, svc.backends,
@@ -1617,7 +1611,7 @@ func (s *Service) deleteServiceLocked(svc *svcInfo) error {
 		s.healthServer.DeleteService(lb.ID(svc.frontend.ID))
 	}
 
-	deleteMetric.Inc()
+	metrics.ServicesEventsCount.WithLabelValues("delete").Inc()
 	s.notifyMonitorServiceDelete(svc.frontend.ID)
 
 	return nil
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -412,8 +412,8 @@ var (
 
 	// Services
 
-	// ServicesCount number of services
-	ServicesCount = NoOpCounterVec
+	// ServicesEventsCount counts the number of services
+	ServicesEventsCount = NoOpCounterVec
 
 	// Errors and warnings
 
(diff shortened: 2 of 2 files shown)


chancez:
Odd. I'm reading the docs on `WithLabelValues` to figure out the cause of the behavior described in your commit message. It refers to https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#CounterVec.GetMetricWithLabelValues which says:

> Keeping the Counter for later use is possible (and should be considered if performance is critical), but keep in mind that Reset, DeleteLabelValues and Delete can be used to delete the Counter from the CounterVec. In that case, the Counter will still exist, but it will not be exported anymore, even if a Counter with the same label values is created later.

Do we do anything like call Reset or Delete*?

pkg/service/service.go
```diff
@@ -33,12 +33,6 @@ import (
 
 const anyPort = "*"
 
-var (
-	updateMetric = metrics.ServicesCount.WithLabelValues("update")
```
ysksuzuki, line 37: This code is called before `NewLegacyMetrics` initializes ServicesCount, and calls `NoOpCounterVec.WithLabelValues` (returns `NoOpGauge`). That's the reason `services_events_total` is missing. Is my understanding correct?

christarazi: Yes, I think you're right. I opened a small thread about this in #development and that's what Andre suspected as well. :+1:

christarazi: cc @chancez

chancez: Gotcha. How can we prevent this from reoccurring?

christarazi: I'm not too sure to be honest. We have https://github.com/cilium/customvet where we can write our own linting checkers. However, this is the only case that I've found in the code base. Moving forward there is an effort to move metrics into their modules (see https://github.com/cilium/cilium/pull/25651) which should help, but techincally not a guarantee.

ysksuzuki: For example, we can configure a workflow for the metrics check in which we set up Prometheus and scrape the agent's metrics, fetch the metrics key list using promql-cli or Prometheus API, and then check if all metrics are available. But I'm not sure if such an expensive test suite is worth the cost.

ysksuzuki (approved):
The changes look good to me

pkg/service/service.go
christarazi: In the past, I've seen metrics not published unless Cilium modified its value. At least, this is what I observed with `cilium metrics list` and not sure if the same would apply to simply hitting the Prometheus endpoint. If they are not published, then this will make this workflow suggested a bit more complicated.

chancez: The option is to get rid of these NoOp metrics potentially. The approach we're using is a bit odd from my perspective but I haven't really dug into why we do it this way. There's a lot of history probably in how we decided that approach.

We could choose to always create the 'real' metrics (removing NoOps), and then change the logic to not register them instead when the metrics are disabled. However it's probably a bigger/harder change to make (and shouldn't block this PR). Then the metrics would still exist (they're just atomic counters) but they won't be exported (because they aren't registered).

christarazi: That seams quite reasonable to me. I guess I didn't question the structure, but now that I think about it a bit, they're likely defined this way (NoOp) because of the problem of having global metrics defined in one place rather than per module. Due to them being exposed as global vars, they can be accessed at any time without any guarantee that they've been initialized, so NoOp wrapper was created to prevent panics / crashes in case that happened. That's my guess.

cc @aanm (for adding NoOp) and @dylandreimerink (for whether we still need the idea of "NoOp" metrics going forward)

dylandreimerink: > The option is to get rid of these NoOp metrics potentially. The approach we're using is a bit odd from my perspective but I haven't really dug into why we do it this way

The NoOp metrics exist as a stopgap, currently a lot of the code base still uses global variables to get metrics which is subject to ordering. 

> they're likely defined this way (NoOp) because of the problem of having global metrics defined in one place rather than per module. Due to them being exposed as global vars, they can be accessed at any time without any guarantee that they've been initialized, so NoOp wrapper was created to prevent panics / crashes in case that happened.

Yes, that is indeed correct.

Ideally, all metrics are provided to components via parameters during construction, but this conversion takes some time. We can remove NoOps once all global metric variables are gone.

chancez: There's an intermediate step I mentioned that wasn't addressed: You can have globals, without NoOp metrics, but not call `registry.Register(yourMetric)`. This avoids ordering issues and nil globals.

christarazi:
@chancez Is there anything that you think is blocking merging this PR?

pkg/service/service.go
christarazi: Btw, I've updated the second commit's description to describe the global variable vs initialization ordering as the root cause of the bug.

chancez (approved):

pkg/service/service.go
dylandreimerink: > There's an intermediate step I mentioned that wasn't addressed: You can have globals, without NoOp metrics, but not call registry.Register(yourMetric). This avoids ordering issues and nil globals.

Yes, you could init the global var and then upon creating the registry determine if you want to add that metric to the registry or not. 

I am not sure why NoOps were chosen over that approach, it is not something invented for the modular metrics, it was already the way things worked. The main thing that changed was init ordering.

I would argue that removing all global variables in favor of dependency injection ASAP is a better use of cycles then refactoring the NoOps away first.

dylandreimerink (approved):

PR #26721, opened 2023-07-08
egressgw: improve reconciliation for IP routes

julianwiedmann:
For configs with `--install-egress-gateway-routes`, optimize the logic that reconciles the IP routes.

### DIFF
--- a/pkg/egressgateway/manager.go
+++ b/pkg/egressgateway/manager.go
@@ -630,21 +630,27 @@ nextIpRule:
 		}
 	}
 
-	// Then go through each interface on the node
-	links, err := netlink.LinkList()
+	// Fetch all IP routes, and delete the unused EgressGW-specific routes:
+	routes, err := netlink.RouteList(nil, netlink.FAMILY_V4)
 	if err != nil {
-		logger.WithError(err).Error("Cannot list interfaces")
+		logger.WithError(err).Error("Cannot list IP routes")
 		return
 	}
 
-	for _, l := range links {
-		// If egress gateway is active for this interface, move to the next interface
-		if _, ok := activeEgressGwIfaceIndexes[l.Attrs().Index]; ok {
+	for _, route := range routes {
+		linkIndex := route.LinkIndex
+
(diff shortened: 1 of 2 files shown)


pkg/egressgateway/manager.go
```diff
@@ -563,16 +559,27 @@ func (manager *Manager) addMissingIpRulesAndRoutes(isRetry bool) (shouldRetry bo
...
-		}
-		logger.Debug("Added IP routes")
 	}
 
 	for _, policyConfig := range manager.policyConfigs {
-		policyConfig.forEachEndpointAndDestination(addIPRulesAndRoutesForConfig)
+		gwc := &policyConfig.gatewayConfig
+
+		if gwc.localNodeConfiguredAsGateway &&
+			len(policyConfig.matchedEndpoints) > 0 {
```
jibi, line 568: nit, can we keep the early return (in this case continue) to reduce the indentation?

```
if !gwc.localNodeConfiguredAsGateway || len(policyConfig.matchedEndpoints) == 0 {
    continue
}
```

pkg/egressgateway/manager.go
```diff
@@ -630,21 +630,27 @@ nextIpRule:
 		}
 	}
 
-	// Then go through each interface on the node
-	links, err := netlink.LinkList()
+	// Fetch all IP routes, and delete the unused EgressGW-specific routes:
+	routes, err := netlink.RouteList(nil, netlink.FAMILY_V4)
```
joamaki, line 634: Hey, heads up that after https://github.com/cilium/cilium/pull/24677 is merged this can be done more efficiently by querying the new in-memory route table.

joamaki (approved):
LGTM!

jibi (approved):
nice improvements! :rocket: just one nit, feel free to ignore it

pkg/egressgateway/manager.go
julianwiedmann: ack, will pick that up in follow-on PR :+1:

PR #26724, opened 2023-07-09
Add option to redact http headers

ChrsMark:
This PR adds the `http-headers` part of https://github.com/cilium/cilium/issues/23887.

This is to provide the option to redact the values of http headers in http flows.

The specific feature can be configured by providing the following values for example:

```yml
hubble:
  redact:
    enabled: true
    http:
      headers:
        allow:
          - traceparent
          - tracestate
          - Cache-Control
        deny: []
```

Note_1: `allow` list means that these headers will not be redacted
Note_2: `deny` list means that these headers will be redacted
Note_3: only one of these 2 lists can be defined. If both defined then `allow` list has priority and `deny` will be ignored.

The filtering logic/algorithm is described at: https://github.com/cilium/cilium/pull/26724/commits/4c3d701f92a4dbeb533b71d404bf0fbaadce352f#diff-1df644554f9d2598342e7c4d8f9f3ae5af5fd29b52ed0dc88e01ea986f218a36R88-R98

I was not 100% sure what should be the default behavior when just `hubble.redact` is enabled but none of the lists are defined so I had it to redact everything by default. That's just my proposal but let's see what we can get from the review here.

### How to test this manually

1. In the root directory of the project edit https://github.com/cilium/cilium/blob/23bac97b3aa7a34e9d03f7d901bd77a124378446/contrib/testing/kind-values.yaml to add the following lines:
```yaml
hubble:
  redact:
    enabled: true
    http:
      headers:
        allow:
          - traceparent
          - tracestate
          - Cache-Control
```
2. Deploy cilium from source with `make kind-image && make kind-install-cilium`
3. Enable hubble with `cilium hubble enable --chart-directory=/home/chrismark/go/src/github.com/cilium/cilium/install/kubernetes/cilium`
4. Apply a L7 example endpoint with `kubectl apply -f https://raw.githubusercontent.com/cilium/cilium/1.13.3/examples/minikube/sw_l3_l4_l7_policy.yaml`
5. Port forward hubble relay svc with `kubectl port-forward -n kube-system svc/hubble-relay --address 0.0.0.0 --address :: 4245:80`
6. Deploy the sample app with `kubectl create -f https://raw.githubusercontent.com/cilium/cilium/1.13.3/examples/minikube/http-sw-app.yaml`. From https://docs.cilium.io/en/stable/gettingstarted/demo/#starwars-demo
7. Observe the flows with `hubble observe -o jsonpb --pod deathstar --protocol http -f | jq`
8. Execute a curl request with `kubectl exec tiefighter -- curl -H "Content-Type2: application/json" -s -XPOST -v "http://deathstar.default.svc.cluster.local/v1/request-landing" -H "X-Powered-By: ReqBin HTTP Client" -H "tracestate: foobar"`
9. Verify that headers' values are replaced with `HUBBLE_REDACTED` except of the `Accept` one like in the following:
```json
{
  "flow": {
    "time": "2023-09-02T08:14:26.498165607Z",
    "uuid": "5b70099c-e482-4c43-bb74-93e05933c21d",
    "verdict": "FORWARDED",
    "IP": {
      "source": "10.244.1.231",
      "destination": "10.244.1.227",
      "ipVersion": "IPv4"
    },
    "l4": {
      "TCP": {
        "source_port": 53704,
        "destination_port": 80
      }
    },
    "source": {
      "ID": 3464,
      "identity": 27221,
      "namespace": "default",
      "labels": [
        "k8s:app.kubernetes.io/name=tiefighter",
        "k8s:class=tiefighter",
        "k8s:io.cilium.k8s.policy.cluster=kind-kind",
        "k8s:io.kubernetes.pod.namespace=default",
        "k8s:org=empire"
      ],
      "pod_name": "tiefighter"
    },
    "destination": {
      "ID": 3,
      "identity": 56737,
      "namespace": "default",
      "labels": [
        "k8s:app.kubernetes.io/name=deathstar",
        "k8s:class=deathstar",
        "k8s:io.cilium.k8s.policy.cluster=kind-kind",
        "k8s:io.kubernetes.pod.namespace=default",
        "k8s:org=empire"
      ],
      "pod_name": "deathstar-7848d6c4d5-7dwtj",
      "workloads": [
        {
          "name": "deathstar",
          "kind": "Deployment"
        }
      ]
    },
    "Type": "L7",
    "node_name": "kind-kind/kind-worker",
    "l7": {
      "type": "REQUEST",
      "http": {
        "method": "POST",
        "url": "http://deathstar.default.svc.cluster.local/v1/request-landing",
        "protocol": "HTTP/1.1",
        "headers": [
          {
            "key": ":scheme",
            "value": "HUBBLE_REDACTED"
          },
          {
            "key": "Accept",
            "value": "HUBBLE_REDACTED"
          },
          {
            "key": "Content-Type2",
            "value": "HUBBLE_REDACTED"
          },
          {
            "key": "Tracestate",
            "value": "foobar"
          },
          {
            "key": "User-Agent",
            "value": "HUBBLE_REDACTED"
          },
          {
            "key": "X-Envoy-Expected-Rq-Timeout-Ms",
            "value": "HUBBLE_REDACTED"
          },
          {
            "key": "X-Envoy-Internal",
            "value": "HUBBLE_REDACTED"
          },
          {
            "key": "X-Powered-By",
            "value": "HUBBLE_REDACTED"
          },
          {
            "key": "X-Request-Id",
            "value": "HUBBLE_REDACTED"
          }
        ]
      }
    },
    "event_type": {
      "type": 129
    },
    "is_reply": false,
    "Summary": "HTTP/1.1 POST http://deathstar.default.svc.cluster.local/v1/request-landing"
  },
  "node_name": "kind-kind/kind-worker",
  "time": "2023-09-02T08:14:26.498165607Z"
}
```


pkg/hubble/parser/seven/http.go
```diff
@@ -24,7 +35,15 @@ func decodeHTTP(flowType accesslog.FlowType, http *accesslog.LogRecordHTTP, opts
 	sort.Strings(keys)
 	for _, key := range keys {
 		for _, value := range http.Headers[key] {
-			headers = append(headers, &flowpb.HTTPHeader{Key: key, Value: value})
+			if opts.RedactHTTPHeaders {
```
ghost: A structure like
```
if opts.RedactHTTPHeaders && !slices.Contains(allowedHeaders, strings.ToLower(key)) {
    value = defaults.SensitiveValueRedacted
}
headers = append(headers, &flowpb.HTTPHeader{Key: key, Value: value})
```
would be more readable I think.

Also, instead of checking `slices.Contains` every time, I would represent the allow list as a `map[string]struct{}` and check headers like `if _, ok := set[header]; ok ...`, see [here](https://emersion.fr/blog/2017/sets-in-go/). I'm not sure if compiler would similarly optimize this check?

pkg/hubble/parser/seven/http.go
```diff
@@ -7,15 +7,26 @@ import (
...
 	"github.com/cilium/cilium/pkg/proxy/accesslog"
 )
 
 func decodeHTTP(flowType accesslog.FlowType, http *accesslog.LogRecordHTTP, opts *options.Options) *flowpb.Layer7_Http {
+	allowedHeaders := []string{
+		"traceparent",
+		"tracestate",
+		"cache-control",
+		"content-encoding",
+		"accept",
```
ghost: What's the reasoning behind this list? In my opinion it should be longer, otherwise I foresee multiple requests to add a particular header that someone needs.

I'm not sure what's the best source for those, I would go through the [Wikipedia list](https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Standard_request_fields) and allow standard (non-deprecated) and common headers that are not sensitive. But maybe there is a better list somewhere. Some Envoy headers might be needed too.

In addition to `traceparent` and `tracecontext`, Hubble should allow headers for other common propagation formats: [baggage](https://www.w3.org/TR/baggage/#baggage-http-header-format), [X-B3-*](https://github.com/openzipkin/b3-propagation), [X-Datadog-*](https://docs.datadoghq.com/real_user_monitoring/connect_rum_and_traces/?tab=datadog#how-are-rum-resources-linked-to-traces), [X-Amzn-Trace-Id](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html). Hubble doesn't parse these headers itself, but they are used to correlate traces with flows in external systems.

pkg/hubble/parser/seven/http.go
```diff
@@ -7,15 +7,26 @@ import (
...
+	"golang.org/x/exp/slices"
+
 	flowpb "github.com/cilium/cilium/api/v1/flow"
 	"github.com/cilium/cilium/pkg/hubble/defaults"
 	"github.com/cilium/cilium/pkg/hubble/parser/options"
 	"github.com/cilium/cilium/pkg/proxy/accesslog"
 )
 
 func decodeHTTP(flowType accesslog.FlowType, http *accesslog.LogRecordHTTP, opts *options.Options) *flowpb.Layer7_Http {
+	allowedHeaders := []string{
```
ghost: Can you make it a package constant instead of a function-scoped var? Unless there is a better place to define it.

Documentation/observability/visibility.rst
```diff
@@ -84,8 +84,11 @@ sensitive information present in Layer 7 flows. More specifically, it offers
 the following features and values for supported Layer 7 protocols:
 
 * For HTTP: redacting URL query (GET) parameters (``--hubble-redact=http-url-query``)
+* For HTTP: redacting all headers except of specific ones (``--hubble-redact=http-headers``).
+  Traceparent, Tracestate, Cache-Control, Content-Encoding, and Accept headers are not redacted.
 * For Kafka: redacting API key (``--hubble-redact=kafka-api-key``)
 
+
```
ghost, line 92:
```suggestion
```

Documentation/observability/visibility.rst
```diff
@@ -84,8 +84,11 @@ sensitive information present in Layer 7 flows. More specifically, it offers
 the following features and values for supported Layer 7 protocols:
 
 * For HTTP: redacting URL query (GET) parameters (``--hubble-redact=http-url-query``)
+* For HTTP: redacting all headers except of specific ones (``--hubble-redact=http-headers``).
+  Traceparent, Tracestate, Cache-Control, Content-Encoding, and Accept headers are not redacted.
```
ghost:
```suggestion
  The following headers are not redacted: <list here>
```

ghost (commented):
Thank you for tackling that, I left some comments.

pkg/hubble/parser/seven/http_test.go
```diff
@@ -189,6 +189,54 @@ func TestDecodeL7HTTPRequestRemoveUrlQuery(t *testing.T) {
...
+	f := &flowpb.Flow{}
+	err = parser.Decode(lr, f)
+	require.NoError(t, err)
+	assert.Equal(t, &flowpb.HTTP{
+		Code:     0,
+		Method:   "POST",
+		Url:      "http://myhost/some/path?foo=bar",
+		Protocol: "HTTP/1.1",
+		Headers: []*flowpb.HTTPHeader{
+			{Key: "Host", Value: "HUBBLE_REDACTED"},
```
chancez: The host header being redacted gives me pause about this approach. I made a comment on the CFP https://github.com/cilium/cilium/issues/23887#issuecomment-1629275365 to discuss the overall approach being taken.

ldelossa:
@lambdanis hola, looks like this is still in review, but just a heads up, it needs a rebase.

ChrsMark:
Hey folks I have rebased and refactored the PR based on https://github.com/cilium/cilium/issues/23887#issuecomment-1632167762.

That is ready for review now so feel free :).

ghost (changes requested):
Thanks @ChrsMark for the update.

One thing that confused me a bit is the behaviour when both allow and deny lists are non-empty. You wrote in the PR description that:
> only one of these 2 lists can be defined. If both defined then allow list has priority and deny will be ignored.

but the example in `install/kubernetes/cilium/values.yaml.tmpl` has both lists defined, and there is no validation or test case for such config.

I agree with the point from the PR description that defining both lists shouldn't be allowed, it can be confusing. I would suggest fixing the example values and adding validation: in the [Helm chart](https://github.com/cilium/cilium/tree/main/install/kubernetes/cilium/templates/validate.yaml) and in the daemon.

If allowing both lists to be non-empty is justified somehow (feel free to chime in everyone), then could you add a test case for it? Also with one header contained in both lists, it's easier to understand what's happening if there's an example in a test imo.

pkg/hubble/parser/seven/http.go
```diff
@@ -80,3 +84,36 @@ func (p *Parser) httpSummary(flowType accesslog.FlowType, http *accesslog.LogRec
...
+// Based on the allow/deny lists of the provided HttpHeadersList it returns the original value
+// or the redacted constant "HUBBLE_REDACTED" accordingly:
+//  1. If both allow/deny lists are empty but hubble.redact feature is enabled the value of the
+//     header is redacted by default.
+//  2. If an allow list is provided, then if the header's key is contained in the allow list the value
+//     of the header will not be redacted.
+//  3. If a deny list is provided, then if the header's key is contained in the deny list the value
+//     of the header will be redacted.
+//  4. If none of the above happens, then if there is any allow list defined then the value will be redacted
+//     otherwise if there is a deny list defined the value will not be redacted.
```
ghost:
```suggestion
//  1. If HttpHeadersList is not nil (meaning that hubble.redact feature is enabled) but both allow/deny lists are empty then the value of the
//     header is redacted by default.
//  2. If the header's key is contained in the allow list then the value
//     of the header will not be redacted.
//  3. If the header's key is contained in the deny list then the value
//     of the header will be redacted.
//  4. If none of the above happens, then if there is any allow list defined then the value will be redacted
//     otherwise if there is a deny list defined the value will not be redacted.
```

I would change the wording a bit to avoid double ifs in points 2 and 3 and clarify how it's checked that hubble.redact is enabled.

ChrsMark:
Hey @lambdanis! Thanks for the review! The validation you propose makes sense. I have added the validation methods with 
70f1d8b. One thing that I'm not sure about though is if we should fail fatally in case of both lists defined 🤔 Maybe that's too "aggressive" but if that's common practice for the project then that's fine  :).  The execution+validation examples provided below:


### Helm chart validation
With 

```yml
hubble:
  redact:
    enabled: true
    http:
      headers:
        allow:
          - traceparent
          - tracestate
          - Cache-Control
        deny:
          - bar
```

the chart fails to get validated and we get

```console
Error: Unable to enable Hubble: execution error at (cilium/templates/validate.yaml:34:5): Hubble requires only one of .Values.hubble.redact.http.headers.allow, .Values.hubble.redact.http.headers.deny to be specified
```


### Daemon validation

With
```yml
extraArgs: ["--hubble-redact-http-headers-allow=['some']","--hubble-redact-http-headers-deny=['bar']","--hubble-redact-enabled=true"]
```
the cilium Pods fail to get started (`CrashLoopBackOff`) and in the logs we get:
```console
level=fatal msg="invalid daemon configuration: Hubble requires only one of --hubble-redact-http-headers-allow and --hubble-redact-http-headers-deny to be specified" subsys=daemon
```

install/kubernetes/cilium/templates/validate.yaml
```diff
@@ -30,6 +30,10 @@
   {{- end }}
 {{- end }}
 
+{{- if and .Values.hubble.redact.http.headers.allow .Values.hubble.redact.http.headers.deny }}
+  {{ fail "Hubble requires only one of .Values.hubble.redact.http.headers.allow, .Values.hubble.redact.http.headers.deny to be specified"}}
```
ghost:
```suggestion
  {{ fail "Hubble allows only one of .Values.hubble.redact.http.headers.allow, .Values.hubble.redact.http.headers.deny to be specified"}}
```

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1041,19 +1041,44 @@ hubble:
...
   #     http:
   #       urlQuery: true
+  #       headers:
+  #         allow:
+  #           - traceparent
+  #           - tracestate
+  #           - Cache-Control
+  #         deny:
+  #           - Authorization
+  #           - Proxy-Authorization
```
ghost:
```suggestion
```
Could we keep only the allow list in this example to make the intent clear?

pkg/option/config.go
```diff
@@ -2829,6 +2841,13 @@ func (c *DaemonConfig) validateIPv6NAT46x64CIDR() error {
 	return nil
 }
 
+func (c *DaemonConfig) validateHubbleRedact() error {
+	if len(c.HubbleRedactHttpHeadersAllow) > 0 && len(c.HubbleRedactHttpHeadersDeny) > 0 {
+		return fmt.Errorf("Hubble requires only one of --hubble-redact-http-headers-allow and --hubble-redact-http-headers-deny to be specified")
```
ghost:
```suggestion
		return fmt.Errorf("Hubble allows only one of --hubble-redact-http-headers-allow and --hubble-redact-http-headers-deny to be specified")
```

ghost (changes requested):
Thank you @ChrsMark for the update. Looks good to me overall, just requested some small tweaks (can we make the example clear?).

>  One thing that I'm not sure about though is if we should fail fatally in case of both lists have been defined 🤔 Maybe that's too "aggressive" but if that's common practice for the project then that's fine :)

It is a common practice for the project to strive for a conflict-free intent from the user. But I'm interested to hear others' opinions on that, arguably issuing a warning and just ignoring the deny list in case both are provided seems good too.

install/kubernetes/cilium/values.yaml.tmpl
ChrsMark: Ah, right that's a leftover. Thanks!

ghost (approved):
Looks good to me.

ChrsMark:
Thank's @lambdanis ! I see some tests are failing but most probably are unrelated to the patch (some are provisioning issues)?

lambdanis:
yeah, the CI failures look unrelated. Could you try rebasing with main @ChrsMark? I recall at least one CI workflow was fixed on main recently.

ChrsMark:
Hey @lambdanis , it seems that after the rebases the base CI's tests went green again. Could we trigger the rest of them again and move this forward?

lambdanis:
CI seems to be hitting https://github.com/cilium/cilium/issues/28088, but otherwise is green.

@joamaki @nathanjsweet @rolinh @chancez Can you review?

pkg/hubble/parser/options/options.go
```diff
@@ -27,10 +34,11 @@ func CacheSize(size int) Option {
 }
 
 // Redact configures which data Hubble will redact.
-func Redact(logger logrus.FieldLogger, httpQuery bool, kafkaApiKey bool) Option {
+func Redact(logger logrus.FieldLogger, httpQuery, kafkaApiKey bool, allowHeaders, denyHeaders []string) Option {
 	return func(opt *Options) {
 		opt.RedactHTTPQuery = httpQuery
 		opt.RedactKafkaAPIKey = kafkaApiKey
+		opt.RedactHttpHeaders = &HttpHeadersList{allowHeaders, denyHeaders}
```
chancez: Don't use unkeyed fields in struct literals.

pkg/hubble/parser/seven/http.go
```diff
@@ -80,3 +84,36 @@ func (p *Parser) httpSummary(flowType accesslog.FlowType, http *accesslog.LogRec
...
+//     header is redacted by default.
+//  2. If the header's key is contained in the allow list then the value
+//     of the header will not be redacted.
+//  3. If the header's key is contained in the deny list then the value
+//     of the header will be redacted.
+//  4. If none of the above happens, then if there is any allow list defined then the value will be redacted
+//     otherwise if there is a deny list defined the value will not be redacted.
+func filterHeader(key string, value string, headersList *options.HttpHeadersList) string {
+	if headersList != nil {
+		if len(headersList.Allow) == 0 && len(headersList.Deny) == 0 {
```
chancez: The way this is written feels somewhat error prone. I worry that a future refactor will result in all headers being redacted by accident. Perhaps `options.HttpHeadersList` should be `options.HubbleRedactSettings` with a boolean in there for when `hubble.redact` is enabled, so that we do not need to rely on nulls and this somewhat implicit logic.

pkg/hubble/parser/seven/http_test.go
```diff
@@ -423,3 +423,49 @@ func TestGetL7HTTPResponseTraceID(t *testing.T) {
 	_, ok = parser.traceContextCache.Get(requestID)
 	assert.False(t, ok, "request id should not be in the cache")
 }
+
+func TestFilterHeader(t *testing.T) {
```
chancez, line 493: I think an improvement would be to test `decodeHTTP` and verify the behavior with multiple headers. This test case can still exist, but I would like to see a test that has a flow with multiple headers, since it's possible `FilterHeader` could still be incorrectly used within `decodeHTTP` and result in all headers being redacted if not careful.

pkg/hubble/parser/seven/http.go
ChrsMark: Hey, this comes directly from https://github.com/cilium/cilium/blob/96265caefa91a788ec7d1407de8b2607b091ec5e/pkg/hubble/parser/options/options.go#L20. Do you suggest that we should replace this with a new field called `options.HubbleRedactSettings`? Since it's not quite clear how this should look like could you elaborate more with an example?

Also keep in mind that `filterHeader` function only depends on the http headers allow/deny lists and hence that's why a generic `hubble.redact` boolean might be a bit redundant.

In any case I can ofc change it accordingly if we clarify what exactly this should like.

ChrsMark:
Hey @chancez , I have pushed some improvements according to your comments. Only your [comment](https://github.com/cilium/cilium/pull/26724#discussion_r1336366536) about `filterHeader()`'s input is still open since it's not 100% clear to what would be the expected change. Could you please take another look?

pkg/hubble/parser/options/options.go
```diff
@@ -17,6 +17,13 @@ type Options struct {
 	CacheSize         int
 	RedactHTTPQuery   bool
 	RedactKafkaAPIKey bool
+	RedactHttpHeaders *HttpHeadersList
+}
+
+// HttpHeadersList contains the allow/deny list of headers
+type HttpHeadersList struct {
+	Allow []string
```
nathanjsweet: These should be maps.

pkg/hubble/parser/seven/http.go
```diff
@@ -80,3 +84,36 @@ func (p *Parser) httpSummary(flowType accesslog.FlowType, http *accesslog.LogRec
...
+			// That's the default case, where redact is generally enabled but not headers' lists
+			// have been specified. In that case we redact everything by default.
+			return defaults.SensitiveValueRedacted
+		}
+		if slices.Contains(headersList.Allow, strings.ToLower(key)) {
```
nathanjsweet: This should be a map lookup.

nathanjsweet (changes requested):

pkg/hubble/parser/seven/http.go
chancez: > Do you suggest that we should replace this with a new field called options.HubbleRedactSettings

Yes, a new struct, something like:
```
type HubbleRedactSettings struct {
  Enabled bool
  Headers *HttpHeadersList
}
```

> Also keep in mind that filterHeader function only depends on the http headers allow/deny lists and hence that's why a generic hubble.redact boolean might be a bit redundant.

Well, it also relies on the `*options.HttpHeadersList`, whether or not it's nil makes a difference. While pointers _can_ be used to enable/disable the feature, it feels like a foot-gun in this case, which is why I'm suggesting adding a new type with a boolean field to distinguish between "default redaction behavior" and "explicit redaction via allowed/denied lists".


What do you think?

pkg/hubble/parser/seven/http.go
ChrsMark: Done in [d60d196](https://github.com/cilium/cilium/pull/26724/commits/d60d1960959b9b23e13a49b475f94cb0a3489de5)

pkg/hubble/parser/options/options.go
ChrsMark: Done in [d60d196](https://github.com/cilium/cilium/pull/26724/commits/d60d1960959b9b23e13a49b475f94cb0a3489de5)

pkg/hubble/parser/seven/http.go
ChrsMark: @chancez I see! Thanks for the additional explanation. I tried to cover this with [d60d196](https://github.com/cilium/cilium/pull/26724/commits/d60d1960959b9b23e13a49b475f94cb0a3489de5). Let me know if that's what you had in mind :).

ChrsMark:
@chancez @nathanjsweet thank's for your reviews! I have pushed some changes to cover your comments. Feel free to take another look.

chancez (approved):
New tests and updated settings struct looks great. Thanks!

joamaki (approved):

ChrsMark:
@nathanjsweet @rolinh could you have a look please?

nathanjsweet (approved):

lambdanis:
I believe this PR is pending only a review from @cilium/helm, could someone take a look?

install/kubernetes/cilium/templates/cilium-configmap.yaml
```diff
@@ -882,6 +882,20 @@ data:
 {{- if .Values.hubble.redact.http }}
   # Enables redaction of the http URL query part in flows
   hubble-redact-http-urlquery: {{ .Values.hubble.redact.http.urlQuery | quote }}
+{{- if .Values.hubble.redact.http.headers }}
+  #  Enables redaction of http headers by specifying which headers will not be redacted
+{{- if .Values.hubble.redact.http.headers.allow }}
+  hubble-redact-http-headers-allow: {{- range .Values.hubble.redact.http.headers.allow }}
+    {{ . }}
```
nbusseneau, line 887: I think we need to prepend with a dash in order for `range` to generate a proper list here 🤔 
```suggestion
    - {{ . }}
```
Otherwise I think it generates an invalid YAML list:
```yaml
  hubble-redact-http-headers-allow:
    traceparent
    tracestate
    Cache-Control
```

install/kubernetes/cilium/templates/cilium-configmap.yaml
```diff
@@ -882,6 +882,20 @@ data:
...
+{{- end }}
+{{- end }}
+{{- if .Values.hubble.redact.http.headers.deny }}
+  # Enables redaction of http headers by specifying which headers will be redacted
+  hubble-redact-http-headers-deny: {{- range .Values.hubble.redact.http.headers.deny }}
+    {{ . }}
```
nbusseneau, line 893: Same here of course.

install/kubernetes/cilium/templates/cilium-configmap.yaml
```diff
@@ -882,6 +882,20 @@ data:
 {{- if .Values.hubble.redact.http }}
   # Enables redaction of the http URL query part in flows
   hubble-redact-http-urlquery: {{ .Values.hubble.redact.http.urlQuery | quote }}
+{{- if .Values.hubble.redact.http.headers }}
+  #  Enables redaction of http headers by specifying which headers will not be redacted
+{{- if .Values.hubble.redact.http.headers.allow }}
```
nbusseneau: I would swap the comment to be inside the `if` block, so that it only is added to the configmap when the value is actually present :)
```suggestion
{{- if .Values.hubble.redact.http.headers.allow }}
  #  Enables redaction of http headers by specifying which headers will not be redacted
```

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1055,19 +1055,41 @@ hubble:
...
   #
+  #   A deny list for the https headers is also supported (cannot be used along with the allow list):
+  #   --set hubble.redact.http.headers.deny="Authorization,Proxy-Authorization"
   redact:
     enabled: false
     http:
         urlQuery: false
+        headers:
+          allow: []
+          deny: []
```
nbusseneau: I think the `redact` docstring here is getting too beefy and confusing here, what with the caveats and all. At this point I would suggest we instead move the appropriate bits to the individual options (e.g. document `hubble.redact.http.headers.allow` with its own docstring), and keep the main `redact` short and descriptive. No need for a "full example" if the sub-options are properly documented.

install/kubernetes/cilium/templates/validate.yaml
```diff
@@ -30,6 +30,10 @@
   {{- end }}
 {{- end }}
 
+{{- if and .Values.hubble.redact.http.headers.allow .Values.hubble.redact.http.headers.deny }}
+  {{ fail "Hubble allows only one of .Values.hubble.redact.http.headers.allow, .Values.hubble.redact.http.headers.deny to be specified"}}
+{{- end }}
+
```
nbusseneau: 👍🏻

nbusseneau (changes requested):

install/kubernetes/cilium/templates/cilium-configmap.yaml
chancez: I believe this is intentional. The config option corresponds to a StringSlice flag, which takes a single string, with space separated values. We do this in a few other spots in the config like in the prometheus metrics list, etc.

nbusseneau: Ah, I see! In this case, disregard.

install/kubernetes/cilium/templates/cilium-configmap.yaml
joestringer: minor nit here, the description is quite hard to parse. I think the following would be simpler?

`Redact all http headers that do not match this list`

install/kubernetes/cilium/templates/cilium-configmap.yaml
ChrsMark: Yeap it's intentional, it's the same with the `hubble-metrics` from a few lines above: https://github.com/cilium/cilium/blob/8c86d079949bee3c91567d9f98a02399a8193086/install/kubernetes/cilium/templates/cilium-configmap.yaml#L868-L872

install/kubernetes/cilium/templates/cilium-configmap.yaml
ChrsMark: Same as https://github.com/cilium/cilium/pull/26724#discussion_r1364486431

install/kubernetes/cilium/values.yaml.tmpl
ChrsMark: Tried to address it at bd708da. Let me know if I got the suggestion correctly :)

nbusseneau: I was thinking more along the lines of "now we have too many options for `redact`, so its description needs to be split" => that would apply to the whole thing, not just the new sub-keys added in this PR.

So basically the `redact` documentation would be stripped down to just `# -- Configures the redact options for Hubble.` and then the rest of the docs moved to the appropriate subkeys, akin to what's done for the `metrics` entry right above `redact`.

Does this make sense?

ChrsMark: Allright got it! Pushed [41675df](https://github.com/cilium/cilium/pull/26724/commits/41675df9f8244094603a3a2673ad79e5b3d8d7a3) to apply this :)

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1049,39 +1049,51 @@ hubble:
...
-  #     kafka:
-  #       apiKey: false
-  #
-  # You can specify the options from the helm CLI:
-  #
-  #   --set hubble.redact.enabled="true"
-  #   --set hubble.redact.http.urlQuery="true"
-  #   --set hubble.redact.kafka.apiKey="false"
   redact:
     enabled: false
```
nbusseneau, line 1061: Let's document the top level `redact` key one as well, here's a proposal:
```
# -- Enables redacting sensitive information present in Layer 7 flows.
```

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1049,39 +1049,51 @@ hubble:
...
     http:
+        # -- Configures the http urlQuery redact option.
```
nbusseneau:
```suggestion
        # -- Enables redacting URL query (GET) parameters.
```

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1049,39 +1049,51 @@ hubble:
...
-          #
-          # or by using helm CLI:
-          #   --set hubble.redact.http.headers.allow="traceparent,tracestate,Cache-Control"
           allow: []
           # A deny list for the https headers is supported (cannot be used along with the allow list):
+          #   --set hubble.redact.enabled="true"
           #   --set hubble.redact.http.headers.deny="Authorization,Proxy-Authorization"
           deny: []
     kafka:
+        # -- Configures the kafka apiKey redact option.
```
nbusseneau:
```suggestion
        # -- Enables redacting Kakfa's API key.
```

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1049,39 +1049,51 @@ hubble:
...
         # -- Configures the http headers redact allow/deny lists
+        # Example:
+        #   hubble.redact.http.allow:
+        #     - traceparent
+        #     - tracestate
+        #     - Cache-Control
+        #
+        # You can specify the options from the helm CLI:
+        #   --set hubble.redact.enabled="true"
+        #   --set hubble.redact.http.headers.allow="traceparent,tracestate,Cache-Control"
```
nbusseneau: I think all of this should be moved to the `allow` key itself, instead of the `headers` key, and then if we agree on this I would suggest sticking with the same format as for other keys and providing a full example including parent keys:
```
        # -- List of HTTP headers to allow -- headers not matching will be redacted. Note: `allow` and `deny` lists cannot be used both at the same time, only one can be present.
        # Example:
        #   redact:
        #     enabled: true
        #     http:
        #       headers:
        #         allow:
        #           - traceparent
        #           - tracestate
        #           - Cache-Control
        #
        # You can specify the options from the helm CLI:
        #   --set hubble.redact.enabled="true"
        #   --set hubble.redact.http.headers.allow="traceparent,tracestate,Cache-Control"
```

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1049,39 +1049,51 @@ hubble:
...
-          #     - traceparent
-          #     - tracestate
-          #     - Cache-Control
-          #
-          # or by using helm CLI:
-          #   --set hubble.redact.http.headers.allow="traceparent,tracestate,Cache-Control"
           allow: []
           # A deny list for the https headers is supported (cannot be used along with the allow list):
+          #   --set hubble.redact.enabled="true"
           #   --set hubble.redact.http.headers.deny="Authorization,Proxy-Authorization"
```
nbusseneau: Now that we split the docstrings to individual keys I think we can reintroduce the full `deny` example, likewise sticking to the same format as for other keys:
```suggestion
        # -- List of HTTP headers to deny -- matching headers will be redacted. Note: `allow` and `deny` lists cannot be used both at the same time, only one can be present.
        # Example:
        #   redact:
        #     enabled: true
        #     http:
        #       headers:
        #         deny:
        #           - Authorization
        #           - Proxy-Authorization
        #
        # You can specify the options from the helm CLI:
        #   --set hubble.redact.enabled="true"
        #   --set hubble.redact.http.headers.deny="Authorization,Proxy-Authorization"
```

Documentation/observability/visibility.rst
```diff
@@ -88,6 +88,7 @@ More specifically, it offers the following features for supported Layer 7 protoc
 
 * For HTTP: redacting URL query (GET) parameters (``--hubble-redact-http-urlquery``)
 * For Kafka: redacting API key (``--hubble-redact-kafka-apikey``)
+* For HTTP headers: redacting headers defined in ``--hubble-redact-http-headers-allow`` list or redacting all headers except of those defined in ``--hubble-redact-http-headers-deny`` list
```
nbusseneau: I'm just noticing this now but I think this should be the other way around:
```suggestion
* For HTTP headers: redacting all headers except those defined in the ``--hubble-redact-http-headers-allow`` list or redacting only the headers defined in the ``--hubble-redact-http-headers-deny`` list
```

nbusseneau (commented):
Thanks, we're quite close, added some more comments below :)

install/kubernetes/cilium/values.yaml.tmpl
nbusseneau: Thanks, made some comments below.

Documentation/observability/visibility.rst
ChrsMark: Good catch, thanks!

ChrsMark:
Thank's for the detailed review @nbusseneau! Tried to address your comments with [c2394b5](https://github.com/cilium/cilium/pull/26724/commits/c2394b5579c61cbc89a6ace53985d1da96990ad0). Feel free to have another look :).

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1048,28 +1048,67 @@ hubble:
...
+      #     http:
+      #       urlQuery: true
+      #
+      # You can specify the options from the helm CLI:
+      #
+      #   --set hubble.redact.enabled="true"
+      #   --set hubble.redact.http.urlQuery="true"
+      urlQuery: false
+      headers:
+        # -- List of HTTP headers to allow -- headers not matching will be redacted. Note: `allow` and `deny` lists cannot be used both at the same time, only one can be present.
```
nbusseneau: Having another `--` in the docstring breaks generation of the README and documentation, hence why they were not present in the new version of the README after update, so we need to remove that:
```suggestion
        # -- List of HTTP headers to allow: headers not matching will be redacted. Note: `allow` and `deny` lists cannot be used both at the same time, only one can be present.
```

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1048,28 +1048,67 @@ hubble:
...
+        #         allow:
+        #           - traceparent
+        #           - tracestate
+        #           - Cache-Control
+        #
+        # You can specify the options from the helm CLI:
+        #   --set hubble.redact.enabled="true"
+        #   --set hubble.redact.http.headers.allow="traceparent,tracestate,Cache-Control"
+        allow: []
+        # -- List of HTTP headers to deny -- matching headers will be redacted. Note: `allow` and `deny` lists cannot be used both at the same time, only one can be present.
```
nbusseneau:
```suggestion
        # -- List of HTTP headers to deny: matching headers will be redacted. Note: `allow` and `deny` lists cannot be used both at the same time, only one can be present.
```

nbusseneau (changes requested):
Aaah I'm sorry, it appears I made a mistake in my previous suggestion. I added the details below. After updating the template file, please run `make -C install/kubernetes all && make -C Documentation update-helm-values` to ensure that both `install/kubernetes/cilium/README.md` and `Documentation/helm-values.rst` get properly updated 🚀 

Once this is done, I would suggest to squash together the commits and add a nice commit description, and this will be a A++ LGTM for my part 😄

daemon/cmd/daemon_main.go
```diff
@@ -1009,6 +1009,12 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	flags.Bool(option.HubbleRedactKafkaApiKey, defaults.HubbleRedactKafkaApiKey, "Hubble redact Kafka API key from flows")
 	option.BindEnv(vp, option.HubbleRedactKafkaApiKey)
 
+	flags.StringSlice(option.HubbleRedactHttpHeadersAllow, []string{}, "HTTP headers to not be redacted from flows")
+	option.BindEnv(vp, option.HubbleRedactHttpHeadersAllow)
+
+	flags.StringSlice(option.HubbleRedactHttpHeadersDeny, []string{}, "HTTP headers to be redacted from flows")
```
rolinh: nit
```suggestion
	flags.StringSlice(option.HubbleRedactHttpHeadersDeny, []string{}, "HTTP headers to redact from flows")
```

daemon/cmd/daemon_main.go
```diff
@@ -1009,6 +1009,12 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	flags.Bool(option.HubbleRedactKafkaApiKey, defaults.HubbleRedactKafkaApiKey, "Hubble redact Kafka API key from flows")
 	option.BindEnv(vp, option.HubbleRedactKafkaApiKey)
 
+	flags.StringSlice(option.HubbleRedactHttpHeadersAllow, []string{}, "HTTP headers to not be redacted from flows")
```
rolinh: nit:
```suggestion
	flags.StringSlice(option.HubbleRedactHttpHeadersAllow, []string{}, "HTTP headers to keep visible in flows")
```

install/kubernetes/cilium/templates/validate.yaml
```diff
@@ -30,6 +30,10 @@
   {{- end }}
 {{- end }}
 
+{{- if and .Values.hubble.redact.http.headers.allow .Values.hubble.redact.http.headers.deny }}
+  {{ fail "Hubble allows only one of .Values.hubble.redact.http.headers.allow, .Values.hubble.redact.http.headers.deny to be specified"}}
```
rolinh: nit: use more direct wording
```suggestion
  {{ fail "Only one of .Values.hubble.redact.http.headers.allow, .Values.hubble.redact.http.headers.deny can be specified"}}
```

pkg/hubble/parser/options/options.go
```diff
@@ -27,14 +41,27 @@ func CacheSize(size int) Option {
...
 		if logger != nil {
 			logger.WithField(
 				"options",
 				fmt.Sprintf("%+v", opt)).Info("configured Hubble with redact options")
 		}
 	}
 }
+
+func headerSliceToMap(headerList []string) map[string]struct{} {
+	headerMap := map[string]struct{}{}
```
rolinh: Given that we know the size of the map in advance, we can preallocate memory for the map to avoid multiple re-allocation when adding new keypair values.
```suggestion
	headerMap := make(map[string]struct{}, len(headerList))
```

pkg/hubble/parser/seven/http.go
```diff
@@ -80,3 +82,37 @@ func (p *Parser) httpSummary(flowType accesslog.FlowType, http *accesslog.LogRec
...
+			return defaults.SensitiveValueRedacted
+		} else {
+			if len(redactSettings.RedactHttpHeaders.Allow) > 0 {
+				return defaults.SensitiveValueRedacted
+			} else if len(redactSettings.RedactHttpHeaders.Deny) > 0 {
+				return value
+			}
+		}
+	}
+	return value
```
rolinh: Generally speaking, returning early makes the code easier to read as it limits indentation and cognitive overhead. What do you think about the following changes for the function body?
```go
	if !redactSettings.Enabled {
		return value
	}
	if len(redactSettings.RedactHttpHeaders.Allow) == 0 && len(redactSettings.RedactHttpHeaders.Deny) == 0 {
		// That's the default case, where redact is generally enabled but not headers' lists
		// have been specified. In that case we redact everything by default.
		return defaults.SensitiveValueRedacted
	}

	if _, ok := redactSettings.RedactHttpHeaders.Allow[strings.ToLower(key)]; ok {
		return value
	}
	if _, ok = redactSettings.RedactHttpHeaders.Deny[strings.ToLower(key)]; ok {
		return defaults.SensitiveValueRedacted
	}

	if len(redactSettings.RedactHttpHeaders.Allow) > 0 {
		return defaults.SensitiveValueRedacted
	}
	return value
```

pkg/option/config.go
```diff
@@ -2846,6 +2858,13 @@ func (c *DaemonConfig) validateIPv6NAT46x64CIDR() error {
 	return nil
 }
 
+func (c *DaemonConfig) validateHubbleRedact() error {
+	if len(c.HubbleRedactHttpHeadersAllow) > 0 && len(c.HubbleRedactHttpHeadersDeny) > 0 {
+		return fmt.Errorf("Hubble allows only one of --hubble-redact-http-headers-allow and --hubble-redact-http-headers-deny to be specified")
```
rolinh: nit
```suggestion
		return errors.New("only one of --hubble-redact-http-headers-allow and --hubble-redact-http-headers-deny can be specified, not both")
```

rolinh (commented):
Thanks for your work and patience! I've left suggestions below to make both the code and doc strings easier to parse but the PR looks good to me otherwise!

pkg/hubble/parser/seven/http.go
ChrsMark: Nice one! Looks more tidy like this, thanks!

rolinh (approved):
LGTM, let's ship it!

nbusseneau (approved):

nbusseneau (changes requested):
Just noticed a binary file `cilium/cilium` was accidentally added in the PR, please make sure to remove it. Probably it was an output leftover on your end following the rename of `cilium/cilium` to `cilium/cilium-dbg`.

ChrsMark:
> Just noticed a binary file `cilium/cilium` was accidentally added in the PR, please make sure to remove it. Probably it was an output leftover on your end following the rename of `cilium/cilium` to `cilium/cilium-dbg`.

Ah my bad, it somehow made it through the rebases. Thank's for spotting this :).

nbusseneau (approved):

PR #26725, opened 2023-07-09
cilium: Add nodeID object with reference counts for OOO events

jrfastab:
Handle OOO pod/kvstore/node events.

### Assumptions and Requirements:

We require same event types for single objects to be ordered. For example:

  NodeAdd(1) -> NodeDelete(1)

is valid. But we would not handle this order currently,

  NodeDelete(1) -> NodeAdd(1)

I'm assuming the API server and watchers can at least maintain this constraint. We don't however require any ordering between Node events and Pod events nor between different objects. So these are valid,

   PodAdd(hostA.IP) -> NodeAdd(hostA) -> NodeDelete(hostA.IP) -> PodDelete(hostA.IP)

   NodeAdd(hostA) -> NodeDelete(hostA) -> PodAdd(hostA.IP) -> PodDelete(hostA.IP)

Also noteworthy we require the Delete events to eventually reach us to be eventually consistent. We also require the delete events to reach us before the system reuses the IP. If the system were to try to reuse the hostIP before we received a delete for the same IP we wouldn't be able to identify if the next delete event was for the old object or the latest object.

### NodeID Issue

The following flow can currently cause an orphaned nodeID, meaning  a nodeID is associated with an IP in the nodeID table but it maps to a stale IP.

   NodeDelete(hostA) -> Pod{Add|Delete}(hostA.IP)

the reason is nodeIDs are removed from the nodeID map on Node Delete. But, in order to handle the case where pod or kvstore events are received before the NodeAdd ipcacheUpdates will also create an id if no id is found. So we get,
```
     Event                                              Mapping
 ---------------------------------------------------------------
     NodeAdd(hostA)                           id:1, hostA.IP
     NodeDelete(hostA)                      --- // empty now
     PodAdd(hostA.IP)                          id:2, hostA.IP
     PodDelete(hostA.IP)                     id.2, hostA.IP                        
```
After the last PodDelete we have a stale entry and no future node delete to clean up the table. Couple things might happen now. We can leak nodeIds until nodeID cache runs out. If the IPAM reuses an IP we can collide and use the old id which would be incorrect.

### Fix

The first naive attempt is to just delete the id on PodDelete() but this trivially breaks the expected case,

 NodeAdd() -> PodAdd() -> PodDelete() /// broken -> NodeDelete()

So we can't simply remove on PodDelete. This series instead adds refcnt and delete from pod flow so that the broken case is resolved as follows,
```
     Event                                              Mapping                     RefCnt:
 ----------------------------------------------------------------------------------------------
     NodeAdd(hostA)                           id:1, hostA.IP                 1
     NodeDelete(hostA)                      --- // empty now            0
     PodAdd(hostA.IP)                          id:2, hostA.IP                1
     PodDelete(hostA.IP)                    --- // empty now             0        
```
And the good case continues to work as expected,
```
     Event                                              Mapping                     RefCnt:
 ----------------------------------------------------------------------------------------------
     NodeAdd(hostA)                           id:1, hostA.IP                 1
     PodAdd(hostA.IP)                          id:2, hostA.IP                2
     PodDelete(hostA.IP)                     id:2, hostA.IP                 1 
     NodeDelete(hostA)                      --- // empty now             0
```

### Nice to do

I didn't do all the nice to have things in this series couple items on the todo list,
* Lift nodeID logic into its own pkg so we can fake the map updates and write unit tests
* Lift nodeID allocation out of ipsec and encap specific logic and use in all cases

### Testing

I wrote a set of unit tests to handle the cases attached here. But, unfortunately it requires commenting out the bpf map updates because there is no way to "fake" those from this layer easily. Also I didn't catch if there was a way to fake the GetLocalIP calls they were panic'ing because I didn't setup those layers correctly from the unit tests.

Then I tested by hand in an EKS ENI setup with 30nodes. Where I jumped down/up between nodes and had a stress test nginx server that ran on all nodes as they came online and would terminate and restart every few minutes. This let me get lots of interesting order on events posted for node add/delete and pod add/delete while having traffic in-flight which would trigger warnings/bugs when hitting nodeid mismatches. Also doing this shows that we get IP reuse so that we get collisions. With this PR I've managed to run this scenario for 24hrs or so without any further issues.

To test if we have "bad" id mappings I check the clusters ipcache for any entries that are not local objects without a nodeID or with a nodeID not in the nodeID list. And then check the other way to ensure the nodeID list only has IP addresses that exist. Before the PR the nodeID list would grow over long running tests as well. Now it correctly stays roughly aligned with number of nodes in the system

Note I didn't test this PR exactly but the backport onto released version.

### DIFF
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -1184,6 +1184,8 @@ func (n *linuxNodeHandler) nodeUpdate(oldNode, newNode *nodeTypes.Node, firstAdd
 		oldIP4 = oldNode.GetNodeIP(false)
 		oldIP6 = oldNode.GetNodeIP(true)
 		oldKey = oldNode.EncryptionKey
+
+		n.deleteNodeIPs(oldNode.IPAddresses, newNode.IPAddresses)
 	}
 
 	if n.nodeConfig.EnableIPSec && !n.nodeConfig.EncryptNode {
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -52,6 +52,10 @@ type NeighLink struct {
 	Name string `json:"link-name"`
 }
 
+type nodeID struct {
+	id uint16
+}
+
 type linuxNodeHandler struct {
 	mutex                lock.RWMutex
 	isInitialized        bool
@@ -73,7 +77,7 @@ type linuxNodeHandler struct {
 	// Pool of available IDs for nodes.
 	nodeIDs idpool.IDPool
 	// Node-scoped unique IDs for the nodes.
-	nodeIDsByIPs map[string]uint16
+	nodeIDsByIPs map[string]*nodeID
 	// reverse map of the above
 	nodeIPsByIDs map[uint16]string
 
@@ -101,7 +105,7 @@ func NewNodeHandler(datapathConfig DatapathConfiguration, nodeAddressing datapat
 		neighLastPingByNextHop: map[string]time.Time{},
 		nodeMap:                nodeMap,
 		nodeIDs:                idpool.NewIDPool(minNodeID, maxNodeID),
-		nodeIDsByIPs:           map[string]uint16{},
+		nodeIDsByIPs:           map[string]*nodeID{},
 		nodeIPsByIDs:           map[uint16]string{},
 		ipsecMetricCollector:   ipsec.NewXFRMCollector(),
 	}
@@ -1575,10 +1579,10 @@ func (n *linuxNodeHandler) deleteIPsec(oldNode *nodeTypes.Node) {
 	scopedLog.Debugf("Removing IPsec configuration for node")
 
 	nodeID := n.getNodeIDForNode(oldNode)
-	if nodeID == 0 {
+	if nodeID == nil {
 		scopedLog.Warning("No node ID found for node.")
 	} else {
-		ipsec.DeleteIPsecEndpoint(nodeID)
+		ipsec.DeleteIPsecEndpoint(nodeID.id)
 	}
 
 	if n.nodeConfig.EnableIPv4 && oldNode.IPv4AllocCIDR != nil {
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -53,7 +53,8 @@ type NeighLink struct {
 }
 
 type nodeID struct {
-	id uint16
+	id     uint16
+	refcnt uint32
 }
 
 type linuxNodeHandler struct {
--- a/pkg/datapath/linux/node_ids.go
+++ b/pkg/datapath/linux/node_ids.go
@@ -105,6 +105,25 @@ func (n *linuxNodeHandler) getNodeIDForNode(node *nodeTypes.Node) uint16 {
 	return nodeID
 }
 
+func (n *linuxNodeHandler) deleteNodeIPs(oldIPs, newIPs []nodeTypes.Address) {
+	for _, oldAddr := range oldIPs {
+		found := false
+		for _, newAddr := range newIPs {
+			if newAddr.IP.String() == oldAddr.IP.String() {
+				found = true
+				break
+			}
+		}
+		if !found {
+			if err := n.unmapNodeID(oldAddr.IP.String()); err != nil {
+				log.WithError(err).WithFields(logrus.Fields{
+					logfields.IPAddr: oldAddr,
+				}).Warn("DeleteNodeIPS failed to remove a node IP to node ID mapping")
+			}
+		}
+	}
+}
+
 // allocateIDForNode allocates a new ID for the given node if one hasn't already
 // been assigned. If any of the node IPs have an ID associated, then all other
 // node IPs receive the same. This might happen if we allocated a node ID from
--- a/pkg/datapath/linux/node_ids.go
+++ b/pkg/datapath/linux/node_ids.go
@@ -23,6 +23,12 @@ const (
 	maxNodeID = idpool.ID(^uint16(0))
 )
 
+var (
+	localNodeID = nodeID{
+		id: 0,
+	}
+)
+
 // AllocateNodeID allocates a new ID for the given node (by IP) if one wasn't
 // already assigned.
 func (n *linuxNodeHandler) AllocateNodeID(nodeIP net.IP) uint16 {
@@ -36,20 +42,23 @@ func (n *linuxNodeHandler) AllocateNodeID(nodeIP net.IP) uint16 {
 	n.mutex.Lock()
 	defer n.mutex.Unlock()
 
-	if nodeID, exists := n.getNodeIDForIP(nodeIP); exists {
+	if nodeID := n.getNodeIDForIP(nodeIP); nodeID != nil {
 		// Don't allocate a node ID if one already exists.
-		return nodeID
+		return nodeID.id
 	}
 
-	nodeID := uint16(n.nodeIDs.AllocateID())
-	if nodeID == uint16(idpool.NoID) {
+	id := uint16(n.nodeIDs.AllocateID())
+	if id == uint16(idpool.NoID) {
 		log.Error("No more IDs available for nodes")
-		return nodeID
+		return id
 	} else {
 		log.WithFields(logrus.Fields{
-			logfields.NodeID: nodeID,
+			logfields.NodeID: id,
 			logfields.IPAddr: nodeIP,
 		}).Debug("Allocated new node ID for node IP address")
+	}
+	nodeID := &nodeID{
+		id: id,
 	}
 	if err := n.mapNodeID(nodeIP.String(), nodeID); err != nil {
 		log.WithError(err).WithFields(logrus.Fields{
@@ -57,7 +66,7 @@ func (n *linuxNodeHandler) AllocateNodeID(nodeIP net.IP) uint16 {
 			logfields.IPAddr: nodeIP.String(),
 		}).Error("Failed to map node IP address to allocated ID")
 	}
-	return nodeID
+	return nodeID.id
 }
 
 func (n *linuxNodeHandler) GetNodeIP(nodeID uint16) string {
@@ -76,33 +85,36 @@ func (n *linuxNodeHandler) GetNodeID(nodeIP net.IP) (uint16, bool) {
 	n.mutex.RLock()
 	defer n.mutex.RUnlock()
 
-	return n.getNodeIDForIP(nodeIP)
+	nodeID := n.getNodeIDForIP(nodeIP)
+	if nodeID == nil {
+		return 0, false
+	}
+	return nodeID.id, true
 }
 
-func (n *linuxNodeHandler) getNodeIDForIP(nodeIP net.IP) (uint16, bool) {
+func (n *linuxNodeHandler) getNodeIDForIP(nodeIP net.IP) *nodeID {
 	localNodeV4 := node.GetIPv4()
 	localNodeV6 := node.GetIPv6()
 	if localNodeV4.Equal(nodeIP) || localNodeV6.Equal(nodeIP) {
-		return 0, true
+		return &localNodeID
 	}
 
 	if nodeID, exists := n.nodeIDsByIPs[nodeIP.String()]; exists {
-		return nodeID, true
+		return nodeID
 	}
 
-	return 0, false
+	return nil
 }
 
 // getNodeIDForNode gets the node ID for the given node if one was allocated
 // for any of the node IP addresses. If none is found, 0 is returned.
-func (n *linuxNodeHandler) getNodeIDForNode(node *nodeTypes.Node) uint16 {
-	nodeID := uint16(0)
+func (n *linuxNodeHandler) getNodeIDForNode(node *nodeTypes.Node) *nodeID {
 	for _, addr := range node.IPAddresses {
-		if id, exists := n.nodeIDsByIPs[addr.IP.String()]; exists {
-			nodeID = id
+		if nodeID, exists := n.nodeIDsByIPs[addr.IP.String()]; exists {
+			return nodeID
 		}
 	}
-	return nodeID
+	return nil
 }
 
 func (n *linuxNodeHandler) deleteNodeIPs(oldIPs, newIPs []nodeTypes.Address) {
@@ -130,17 +142,20 @@ func (n *linuxNodeHandler) deleteNodeIPs(oldIPs, newIPs []nodeTypes.Address) {
 // the ipcache, where we don't have all node IPs but only one.
 func (n *linuxNodeHandler) allocateIDForNode(node *nodeTypes.Node) uint16 {
 	// Did we already allocate a node ID for any IP of that node?
-	nodeID := n.getNodeIDForNode(node)
+	nID := n.getNodeIDForNode(node)
 
-	if nodeID == 0 {
-		nodeID = uint16(n.nodeIDs.AllocateID())
-		if nodeID == uint16(idpool.NoID) {
+	if nID == nil {
+		id := uint16(n.nodeIDs.AllocateID())
+		if id == uint16(idpool.NoID) {
 			log.WithField(logfields.NodeName, node.Name).Error("No more IDs available for nodes")
 		} else {
 			log.WithFields(logrus.Fields{
-				logfields.NodeID:   nodeID,
+				logfields.NodeID:   id,
 				logfields.NodeName: node.Name,
 			}).Debug("Allocated new node ID for node")
+		}
+		nID = &nodeID{
+			id: id,
 		}
 	}
 
@@ -149,23 +164,29 @@ func (n *linuxNodeHandler) allocateIDForNode(node *nodeTypes.Node) uint16 {
 		if _, exists := n.nodeIDsByIPs[ip]; exists {
 			continue
 		}
-		if err := n.mapNodeID(ip, nodeID); err != nil {
+		if err := n.mapNodeID(ip, nID); err != nil {
 			log.WithError(err).WithFields(logrus.Fields{
-				logfields.NodeID: nodeID,
+				logfields.NodeID: nID.id,
 				logfields.IPAddr: ip,
 			}).Error("Failed to map node IP address to allocated ID")
 		}
 	}
-	return nodeID
+	return nID.id
 }
 
 // deallocateIDForNode deallocates the node ID for the given node, if it was allocated.
 func (n *linuxNodeHandler) deallocateIDForNode(oldNode *nodeTypes.Node) {
 	nodeID := n.getNodeIDForNode(oldNode)
+	if nodeID == nil {
+		log.WithFields(logrus.Fields{
+			logfields.NodeName: oldNode.Name,
+		}).Warn("deallocate ID for node that does not have an existing ID")
+		return
+	}
 
 	for _, addr := range oldNode.IPAddresses {
 		id := n.nodeIDsByIPs[addr.IP.String()]
-		if nodeID != id {
+		if nodeID.id != id.id {
 			log.WithFields(logrus.Fields{
 				logfields.NodeName: oldNode.Name,
 				logfields.IPAddr:   addr.IP,
@@ -173,19 +194,20 @@ func (n *linuxNodeHandler) deallocateIDForNode(oldNode *nodeTypes.Node) {
 		}
 	}
 
-	n.deallocateNodeIDLocked(nodeID)
+	n.deallocateNodeIDLocked(nodeID.id)
 }
 
 // DeallocateNodeID deallocates the given node ID, if it was allocated.
-func (n *linuxNodeHandler) DeallocateNodeID(nodeID uint16) {
+func (n *linuxNodeHandler) DeallocateNodeID(id uint16) {
 	n.mutex.Lock()
 	defer n.mutex.Unlock()
-	n.deallocateNodeIDLocked(nodeID)
+
+	n.deallocateNodeIDLocked(id)
 }
 
 func (n *linuxNodeHandler) deallocateNodeIDLocked(nodeID uint16) {
 	for ip, id := range n.nodeIDsByIPs {
-		if nodeID == id {
+		if nodeID == id.id {
 			if err := n.unmapNodeID(ip); err != nil {
 				log.WithError(err).WithFields(logrus.Fields{
 					logfields.NodeID: nodeID,
@@ -204,7 +226,7 @@ func (n *linuxNodeHandler) deallocateNodeIDLocked(nodeID uint16) {
 // mapNodeID adds a node ID <> IP mapping into the local in-memory map of the
 // Node Manager and in the corresponding BPF map. If any of those map updates
 // fail, both are cancelled and the function returns an error.
-func (n *linuxNodeHandler) mapNodeID(ip string, id uint16) error {
+func (n *linuxNodeHandler) mapNodeID(ip string, id *nodeID) error {
 	if _, exists := n.nodeIDsByIPs[ip]; exists {
 		return fmt.Errorf("a mapping for node IP %s already exists", ip)
 	}
@@ -214,14 +236,14 @@ func (n *linuxNodeHandler) mapNodeID(ip string, id uint16) error {
 		return fmt.Errorf("invalid node IP %s", ip)
 	}
 
-	if err := n.nodeMap.Update(nodeIP, id); err != nil {
+	if err := n.nodeMap.Update(nodeIP, id.id); err != nil {
 		return err
 	}
 
 	// We only add the IP <> ID mapping in memory once we are sure it was
 	// successfully added to the BPF map.
 	n.nodeIDsByIPs[ip] = id
-	n.nodeIPsByIDs[id] = ip
+	n.nodeIPsByIDs[id.id] = ip
 
 	return nil
 }
@@ -244,7 +266,7 @@ func (n *linuxNodeHandler) unmapNodeID(ip string) error {
 	}
 	if id, exists := n.nodeIDsByIPs[ip]; exists {
 		delete(n.nodeIDsByIPs, ip)
-		delete(n.nodeIPsByIDs, id)
+		delete(n.nodeIPsByIDs, id.id)
 	}
 
 	return nil
@@ -257,12 +279,12 @@ func (n *linuxNodeHandler) DumpNodeIDs() []*models.NodeID {
 
 	nodeIDs := map[uint16]*models.NodeID{}
 	for ip, id := range n.nodeIDsByIPs {
-		if nodeID, exists := nodeIDs[id]; exists {
+		if nodeID, exists := nodeIDs[id.id]; exists {
 			nodeID.Ips = append(nodeID.Ips, ip)
-			nodeIDs[id] = nodeID
+			nodeIDs[id.id] = nodeID
 		} else {
-			i := int64(id)
-			nodeIDs[id] = &models.NodeID{
+			i := int64(id.id)
+			nodeIDs[id.id] = &models.NodeID{
 				ID:  &i,
 				Ips: []string{ip},
 			}
@@ -280,13 +302,16 @@ func (n *linuxNodeHandler) DumpNodeIDs() []*models.NodeID {
 // BPF map and into the node handler in-memory copy.
 func (n *linuxNodeHandler) RestoreNodeIDs() {
 	// Retrieve node IDs from the BPF map to be able to restore them.
-	nodeIDs := make(map[string]uint16)
+	nodeIDs := make(map[string]*nodeID)
 	parse := func(key *nodemap.NodeKey, val *nodemap.NodeValue) {
 		address := key.IP.String()
 		if key.Family == bpf.EndpointKeyIPv4 {
 			address = net.IP(key.IP[:net.IPv4len]).String()
 		}
-		nodeIDs[address] = val.NodeID
+		nid := &nodeID{
+			id: val.NodeID,
+		}
+		nodeIDs[address] = nid
 	}
 	if err := n.nodeMap.IterateWithCallback(parse); err != nil {
 		log.WithError(err).Error("Failed to dump content of node map")
@@ -297,7 +322,7 @@ func (n *linuxNodeHandler) RestoreNodeIDs() {
 	log.Infof("Restored %d node IDs from the BPF map", len(nodeIDs))
 }
 
-func (n *linuxNodeHandler) registerNodeIDAllocations(allocatedNodeIDs map[string]uint16) {
+func (n *linuxNodeHandler) registerNodeIDAllocations(allocatedNodeIDs map[string]*nodeID) {
 	n.mutex.Lock()
 	defer n.mutex.Unlock()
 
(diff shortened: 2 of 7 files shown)


pkg/datapath/linux/node_ids.go
```diff
@@ -23,6 +23,13 @@ const (
 	maxNodeID = idpool.ID(^uint16(0))
 )
 
+var (
+	localNodeID = nodeID{
+		id:     0,
+		refcnt: 1,
```
gandro, line 29: Is this refcount here ever increased/decreased, or is this basically a constant?

If the contents can here (i.e. if this is stateful), then we should not make this a global variable and instead have it be part of the `linuxNodeHandler`

pkg/datapath/linux/node.go
```diff
@@ -101,7 +106,7 @@ func NewNodeHandler(datapathConfig DatapathConfiguration, nodeAddressing datapat
 		neighLastPingByNextHop: map[string]time.Time{},
 		nodeMap:                nodeMap,
 		nodeIDs:                idpool.NewIDPool(minNodeID, maxNodeID),
-		nodeIDsByIPs:           map[string]uint16{},
+		nodeIDsByIPs:           map[string]*nodeID{},
```
gandro, line 109: We already have a ref-counting type in `pkg/counter` that we should probably use here, unless there is a good reason to have custom ref counting logic.

pkg/datapath/linux/node_ids.go
```diff
@@ -76,97 +87,174 @@ func (n *linuxNodeHandler) GetNodeID(nodeIP net.IP) (uint16, bool) {
...
+				log.WithError(err).WithFields(logrus.Fields{
+					logfields.IPAddr: oldAddr,
+				}).Warn("DeleteNodeIPS failed to remove a node IP to node ID mapping")
+			}
 		}
 	}
-	return nodeID
+}
+
+func (n *linuxNodeHandler) AllocateIDForNode(node *nodeTypes.Node) uint16 {
```
gandro: Why does this need to be exported? I don't see it being accessed out of the package here

pkg/datapath/linux/node_ids.go
```diff
@@ -76,97 +87,174 @@ func (n *linuxNodeHandler) GetNodeID(nodeIP net.IP) (uint16, bool) {
...
 	}
 
-	n.deallocateNodeIDLocked(nodeID)
+	n.deallocateNodeIDLocked(nodeID.id)
 }
 
 // DeallocateNodeID deallocates the given node ID, if it was allocated.
-func (n *linuxNodeHandler) DeallocateNodeID(nodeID uint16) {
+func (n *linuxNodeHandler) DeallocateNodeID(nodeIP net.IP) {
+	//localNode := node.GetIPv4()
```
gandro: left over comments?

pkg/ipcache/ipcache.go
```diff
@@ -555,7 +564,12 @@ func (ipc *IPCache) DumpToListenerLocked(listener IPIdentityMappingListener) {
 		}
 		nodeID := uint16(0)
 		if hostIP != nil {
-			nodeID = ipc.AllocateNodeID(hostIP)
+			exists := false
+
+			nodeID, exists = ipc.GetNodeID(hostIP)
+			if !exists {
+				nodeID = ipc.AllocateNodeID(hostIP)
```
gandro: Why do we need to bump the ref count on dump? Who will decrement it? 

Also, since the `linuxNodeHandler` is not locked here, it can happen that the nodeID appears between the time of check and the call to allocate here. Not sure if that is a problem, but I wanted to point it out nonetheless.

gandro (commented):
Some low-level comments. I have not reviewed the high-level interaction between IPCache and Linux Datapath yet, but it seems this PR does not fundamentally change the dependencies the between the two

pkg/datapath/linux/node_ids.go
jrfastab: No refcnt never changes its just a static id so I had something to say was the localNode.

pkg/datapath/linux/node.go
jrfastab: Are you suggesting have a ./pkg/counter/nodeId.go file and move the definitions there? I wasn't aware of the pkg is the main reason its not used.

pkg/datapath/linux/node_ids.go
jrfastab: It doesn't I'll stop exporting it. The idea was to create a ./pkg/nodeid but I decided it was a bit too heavy handed for something that needed backports. And wanted to get logic in first and then move to a new pkg second.

pkg/datapath/linux/node_ids.go
jrfastab: yep unfortunately GetIPV4() caused unit tests to panic because there are no *fake* handlers for it so I commented it out for testing and forgot to undo it. Thanks for catching.

pkg/ipcache/ipcache.go
jrfastab: My thought process on this is if we find an entry in the ipcache with a hostIP and no nodeID then we somehow got in a state where we have an entry that is missing a nodeID. So we should try to fix it by creating an ID for it.

The AllocateNodeID will lock the linuxNodeHandler and then rechecks if the nodeIP has a NodeID so I think we are OK from locking side because the AllocateNodeID would just return it if something allocated it between the check and call to Alocate.

What deletes it? Interesting question I guess it depends on how it got here in the first place. But a NodeDelete with the hostIP associated with it would then delete the entry. Or the podDelete for the ipcache entry would also do it.

I don't think this is essential to have so we can either remove it or leave it. I guess this logic is called from cmd on daemon side so really doesn't add much value and adds extra complexity we have to reason about. I'm tempted to just delete it.

jrfastab: I replaced the Allocate with a Warning so we get some indication this is happening. I think this makes sense. wdyt

pkg/datapath/linux/node.go
gandro: Oh, nevermind. I thought we had a counter map, but it's only a counter set. So that's not relevant here, sorry for the noise.

pkg/datapath/linux/node.go
```diff
@@ -1184,6 +1184,8 @@ func (n *linuxNodeHandler) nodeUpdate(oldNode, newNode *nodeTypes.Node, firstAdd
 		oldIP4 = oldNode.GetNodeIP(false)
 		oldIP6 = oldNode.GetNodeIP(true)
 		oldKey = oldNode.EncryptionKey
+
+		n.deleteNodeIPs(oldNode.IPAddresses, newNode.IPAddresses)
```
pchaigno, line 1193: > Then  we only add the newNode IP addresses and fail to delete the
old IP addresses we can get stale IP addresses in the nodeID map that
are not mapped to any real node except for in the nodeID map. Things
are broken now, but may not be visibly throwing errors until something
recycles that IP addresses that we failed to delete properly. When its
reused (say on a nodeAdd) the datapath logic will try to add it to the
nodeID map and fail because it already exists. But, being added from
some older instance will have the wrong node ID. Now the ipcache OTOH
will handle it correctly and add it. So we get ipcache and nodeID list
out of sync. The nodeID map is then used to create policy/state entries
in the xfrm logic so we will be missing an entry in the policy. Finally,
the result is skb falls through policy to drop rule and is free'd. The
end result is that traffic broken between some nodes.

I'm not sure I understand what you're saying here. Let's define a few things for clarity:
- OldIP is the IP address that was removed from node A and later reused by node B.
- NewIP is the IP address that was added to node A, to replace OldIP.
- OldNodeID is the node ID allocated for OldIP.

Now, are you saying that the ipcache entries for node B will have OldNodeID while the XFRM config for that node will be missing?

---

Which IP address is being updated here? The NodeInternalIP or the CiliumInternalIP?

pkg/datapath/linux/node_ids.go
```diff
@@ -124,39 +144,47 @@ func (n *linuxNodeHandler) deleteNodeIPs(oldIPs, newIPs []nodeTypes.Address) {
 	}
 }
 
+func (n *linuxNodeHandler) AllocateIDForNode(node *nodeTypes.Node) uint16 {
```
pchaigno: Why is this function needed?

pkg/datapath/linux/node_ids.go
```diff
@@ -23,6 +23,12 @@ const (
 	maxNodeID = idpool.ID(^uint16(0))
 )
 
+var (
+	localNodeID = nodeID{
+		id: 0,
+	}
```
pchaigno, line 30: Should this be a constant?

pkg/datapath/linux/node.go
```diff
@@ -1574,7 +1578,7 @@ func (n *linuxNodeHandler) deleteIPsec(oldNode *nodeTypes.Node) {
 	scopedLog := log.WithField(logfields.NodeName, oldNode.Name)
 	scopedLog.Debugf("Removing IPsec configuration for node")
 
-	nodeID := n.getNodeIDForNode(oldNode)
+	nodeID := n.GetNodeIDForNode(oldNode)
 	if nodeID == 0 {
```
pchaigno:
```suggestion
	nodeID := n.getNodeIDForNode(oldNode)
	if nodeID == nil {
```
Then we can get rid of function `GetNodeIDForNode`.

pkg/datapath/linux/node_ids.go
```diff
@@ -76,33 +85,44 @@ func (n *linuxNodeHandler) GetNodeID(nodeIP net.IP) (uint16, bool) {
...
+		return &localNodeID, true
 	}
 
 	if nodeID, exists := n.nodeIDsByIPs[nodeIP.String()]; exists {
 		return nodeID, true
 	}
 
-	return 0, false
+	return nil, false
+}
```
pchaigno: Given we are now only returning nil when the node ID doesn't exist, we can remove the second, bool return value and simplify all callers by checking for nil instead.

pkg/datapath/linux/node_ids.go
```diff
@@ -205,11 +205,26 @@ func (n *linuxNodeHandler) deallocateIDForNode(oldNode *nodeTypes.Node) {
 }
 
 // DeallocateNodeID deallocates the given node ID, if it was allocated.
-func (n *linuxNodeHandler) DeallocateNodeID(id uint16) {
+func (n *linuxNodeHandler) DeallocateNodeID(nodeIP net.IP) {
+	localNode := node.GetIPv4()
+	if localNode.Equal(nodeIP) {
+		log.WithFields(logrus.Fields{
+			logfields.IPAddr: nodeIP,
+		}).Warning("Attempt to deallocate local node IP")
```
pchaigno:
```suggestion
		}).Warning("Attempt to deallocate local node ID")
```

pkg/datapath/linux/node_ids.go
```diff
@@ -205,11 +205,26 @@ func (n *linuxNodeHandler) deallocateIDForNode(oldNode *nodeTypes.Node) {
...
+
 	n.mutex.Lock()
 	defer n.mutex.Unlock()
 
-	n.deallocateNodeIDLocked(id)
+	nodeID, exists := n.nodeIDsByIPs[nodeIP.String()]
+	if !exists {
+		log.WithFields(logrus.Fields{
+			logfields.IPAddr: nodeIP,
+		}).Warning("Attempt to dellocate unknown node IP")
```
pchaigno:
```suggestion
		}).Warning("Attempt to deallocate node ID for unknown IP")
```

pkg/datapath/linux/node.go
```diff
@@ -1061,7 +1062,7 @@ func (n *linuxNodeHandler) enableIPsecIPv4(newNode *nodeTypes.Node, zeroMark boo
 		}
 
 		localIP := n.nodeAddressing.IPv4().Router()
-		remoteNodeID := n.AllocateIDForNode(newNode)
```
pchaigno: This change (`AllocateIDForNode` back to `allocateIDForNode`) is in the wrong commit. It's in the last commit instead of the commit that made the change `allocateIDForNode` to `AllocateIDForNode`.

pkg/datapath/linux/node_ids.go
```diff
@@ -144,16 +146,16 @@ func (n *linuxNodeHandler) deleteNodeIPs(oldIPs, newIPs []nodeTypes.Address) {
 	}
 }
 
-func (n *linuxNodeHandler) AllocateIDForNode(node *nodeTypes.Node) uint16 {
-	nodeID := n.allocateIDForNode(node)
+func (n *linuxNodeHandler) allocateIDForNode(node *nodeTypes.Node) uint16 {
```
pchaigno: This revert change to not expose the function is in the wrong commit.

pkg/datapath/linux/node_ids.go
```diff
@@ -144,16 +146,16 @@ func (n *linuxNodeHandler) deleteNodeIPs(oldIPs, newIPs []nodeTypes.Address) {
...
 	return nodeID.id
 }
 
-// allocateIDForNode allocates a new ID for the given node if one hasn't already
+// allocateNodeID allocates a new ID for the given node if one hasn't already
 // been assigned. If any of the node IPs have an ID associated, then all other
 // node IPs receive the same. This might happen if we allocated a node ID from
 // the ipcache, where we don't have all node IPs but only one.
-func (n *linuxNodeHandler) allocateIDForNode(node *nodeTypes.Node) *nodeID {
+func (n *linuxNodeHandler) allocateNodeID(node *nodeTypes.Node) *nodeID {
```
pchaigno: I think this function was renamed by mistake.

pkg/datapath/linux/node_ids.go
```diff
@@ -190,9 +195,24 @@ func (n *linuxNodeHandler) allocateIDForNode(node *nodeTypes.Node) *nodeID {
 // deallocateIDForNode deallocates the node ID for the given node, if it was allocated.
 func (n *linuxNodeHandler) deallocateIDForNode(oldNode *nodeTypes.Node) {
 	nodeID := n.getNodeIDForNode(oldNode)
+	if nodeID == nil {
+		return
```
pchaigno: Should we throw a warning in this case? AFAIU, it's not a case we expect to ever happen, right?

pkg/datapath/linux/node_ids.go
```diff
@@ -224,7 +244,18 @@ func (n *linuxNodeHandler) DeallocateNodeID(nodeIP net.IP) {
...
+		log.WithFields(logrus.Fields{
+			logfields.IPAddr: nodeIP,
+		}).Warning("Attempt to deallocate node IP with refcnt <1")
+	}
+
+	nodeID.refcnt--
+	if nodeID.refcnt == 0 {
+		n.deallocateNodeIDLocked(nodeID.id)
+	}
+	return
```
pchaigno: I think the `return` here can be omitted.

pkg/datapath/linux/node_ids.go
```diff
@@ -224,7 +244,18 @@ func (n *linuxNodeHandler) DeallocateNodeID(nodeIP net.IP) {
...
+	if nodeID.refcnt < 1 {
+		log.WithFields(logrus.Fields{
+			logfields.IPAddr: nodeIP,
+		}).Warning("Attempt to deallocate node IP with refcnt <1")
+	}
+
+	nodeID.refcnt--
+	if nodeID.refcnt == 0 {
+		n.deallocateNodeIDLocked(nodeID.id)
+	}
```
pchaigno: Can't we move lines 248 to 256 inside `deallocateNodeIDLocked`? That would avoid the duplicate code with `deallocateIDForNode`.

It would also ensure we perform the same checks in both cases (`deallocateIDForNode` is missing the check on `< 1`).

pkg/ipcache/ipcache.go
```diff
@@ -422,8 +422,17 @@ func (ipc *IPCache) upsertLocked(
...
+		if hostIP != nil {
+			hostID, _ = ipc.GetNodeID(hostIP)
+		}
+	} else {
+		if hostIP != nil {
+			hostID = ipc.AllocateNodeID(hostIP)
+		}
+		if oldHostIP != nil {
+			ipc.DeallocateNodeID(oldHostIP)
+		}
```
pchaigno, line 435: I think we still have a bug when updating node IP addresses.

In the case of `hostIP != nil` AND `oldHostIP != nil`, we have two cases:
1. **The node update was received before the pod update.** `AllocateNodeID(hostIP)` will return the node ID associated with `oldHostIP` and increment its refcount. `DeallocateNodeID(oldHostIP)` will then decrement the refcount back.
2. **The node update is received after the pod update.** `AllocateNodeID(hostIP)` will allocate a new node ID with refcount of 1. `DeallocateNodeID(oldHostIP)` will decrement the refcount of the previous node ID and deallocate it.

In case 2, we end up with 2 different node IDs allocated for the node, depending on the IP addresses. We may end up with XFRM configs using the first node ID while the pod entries in the ipcache are using the second. Hence, XfrmOutPolBlock drops again.

pkg/ipcache/ipcache.go
```diff
@@ -555,7 +564,14 @@ func (ipc *IPCache) DumpToListenerLocked(listener IPIdentityMappingListener) {
...
 		nodeID := uint16(0)
 		if hostIP != nil {
-			nodeID = ipc.AllocateNodeID(hostIP)
+			exists := false
+
+			nodeID, exists = ipc.GetNodeID(hostIP)
+			if !exists {
+				log.WithFields(logrus.Fields{
+					logfields.IPAddr: hostIP,
+				}).Warn("Dump ToIdentityCache has host without nodeID")
```
pchaigno, line 573:
```suggestion
				}).Warn("IPcache has entry with a node IP but no node ID")
```

pkg/ipcache/ipcache.go
```diff
@@ -653,8 +669,24 @@ func (ipc *IPCache) deleteLocked(ip string, source source.Source) (namedPortsCha
...
+	// delete the old IP through the Upsert path though.
+	if oldHostIP != nil {
+		if cacheModification == Delete {
+			ipc.DeallocateNodeID(oldHostIP)
+		} else if cacheModification == Upsert {
+			if oldHostIP.Equal(newHostIP) {
+				nodeID, _ = ipc.GetNodeID(oldHostIP)
+			} else {
+				if newHostIP != nil {
+					nodeID, _ = ipc.GetNodeID(newHostIP)
```
pchaigno, line 685: I don't understand why we don't want to allocate a node ID or increment the refcount on this specific line.

pkg/datapath/linux/node_ids.go
```diff
@@ -76,97 +87,180 @@ func (n *linuxNodeHandler) GetNodeID(nodeIP net.IP) (uint16, bool) {
...
+				logfields.NodeID:   id,
 				logfields.NodeName: node.Name,
 			}).Debug("Allocated new node ID for node")
 		}
+		nID = &nodeID{
+			id:     id,
+			refcnt: 1,
+		}
+	} else {
+		nID.refcnt++
```
pchaigno, line 164: IIUC, we increment the refcount once per node (regardless of number of node IPs) and once per ipcache entry? I wonder if incrementing the refcount once per node IP on node updates/creations would make things a bit more consistent.

pchaigno (commented):
I have a bunch of comments and there's a bit of everything: lots of minor stuff, some questions, a potential bug.

I really like the new checks with warnings on the various unexpected conditions :-)

pkg/datapath/linux/node.go
jrfastab: for deleteNodeIPs I was concerned about the case where a node updates its IP addresses and remove any no longer needed IPAddress. In existing code we would orphan the IP->nodeID mapping. So flow might look like this,

   1 NoadeA removes OldIP
   2 OldIP is no longer in NodeA->Addressses[]
   3 NodeA is deleted, unmap NodeA->Addresses[]
   4 OldIP from (1) is still in ip->nodeID map

Now that is at least a leak of nodeIDs. If,

   5. NodeB is added reusing OldIP NoceB->Addresses[... OldIP ...]
   6. allocateNodeID() finds the old nodeID
 
But events come out of order weird things might happen,

    5. PodAdd event before node update event with hostIP != OldIP we would create new nodeID and insert it
    6. NodeB is added we would scan Addresses[] and have two differet nodeIDs
 
once this happens something has to break because we can only have one xfrm nodeID mapping.

The text I wrote in the commit above is horrible by the way that should be rewritten. And also issue of how to actually trigger above hypothetical cases. I think adding/removing network interfaces might due it?

pkg/datapath/linux/node_ids.go
jrfastab: I thought so to I had the code to do this here but put it in a patch behind this I can add it I agree it does seem more consistent.

pkg/ipcache/ipcache.go
jrfastab: I'll need to double check... I thought in this case the newHostIP already exists with a nodeID so we don't want to inc it again because we will only get a single delete for it. I'll double check and clarify shortly.

pkg/datapath/linux/node_ids.go
jrfastab: Correct means we somehow tried to deallocate something that either was previously released or we never added it. Either way we are out of sync and should throw a warning

pkg/datapath/linux/node_ids.go
jrfastab: Yep

pkg/datapath/linux/node.go
pchaigno: I agree this is a bug to fix regardless of the impact. I'd just like to get to have some explanation for the XfrmOutPolBlock errors to be sure we have caught all observed bugs.

So let's look at what will happen if the changing node IP address is the NodeInternalIP. I'll detail the case with CiliumInternalIP tomorrow.

All is as seen from a node C, different from nodes A and B.

1. On the update to node A with change from OldIP to NewIP:
  a. We will reinstall XFRM configs for node A. This won't change anything because the NodeInternalIP isn't used in the XFRM configs.
  b. We will call `allocateIDForNode(nodeA)` as part of the node update. That will map NewIP to OldNodeID.
  c. We won't delete the old mapping, so OldIP will stay associated to OldNodeID as well.
2. On the update of pods on node A:
  a. We will update the ipcache entries for all pods of node A. While doing so, we will try to allocate a node ID for NewIP. That will return OldNodeID and ipcache entries for node A will thus have OldNodeID as before.
3. On the creation of node B with OldIP:
  a. We will call `allocateIDForNode(nodeB)` as part of the node creation. That will map all IPs of nodeB, including OldIP, to OldNodeID since the mapping OldIP <> OldNodeID still exists.
  b. We will install XFRM configs for node B. Those configs will use OldNodeID.
4. On the creation of pods on node B:
  a. We will create ipcache entries for all pods of node B. They will be using OldNodeID.

AFAICS, none of this will cause XfrmOutPolBlock.

The state after 4 will however cause us to encapsulate some encrypted packets with the wrong outer IP address (e.g., packets going to node B encapsulated with the IP of node A).

What could explain XfrmOutPolBlock is if we receive the update for pods on node A (2) before the update for node A (1). Then, we will allocate a new, different node ID for NewIP. Let's call it NewNodeID. When sending packets to pods of node A, we will set NewNodeID in the skb->mark, but the XFRM configs will still be matching on OldNodeID. Hence, XfrmOutPolBlock. It's the same bug I mentioned in [the comment below](https://github.com/cilium/cilium/pull/26725#discussion_r1260164732).

pkg/datapath/linux/node_ids.go
jrfastab: I guess golang does not understand const structs.

pkg/ipcache/ipcache.go
jrfastab: I missing the steps to get two nodeIDs. Considering this flow,

First we get a pod update]
    PodUpdate -> upsertLocked(hostIP) -> AllocateNodeID(hostIP)  -> creates hostIP:idA:1

this allocates new ID 'idA' with refcnt 1 and maps it to hostIP. Then in example we get nodeUpdate.

    NodeUpdate->deleteNodeIPs->enableIPsecIPv4->allocateIDForNode

So I think the concern is the deleteNodeIPs in case 2 when a podupdate happens first. (just trying to be sure I understand) For deleteNodeIPs to be called we need an 'oldNode != nil' If this is the case then we didn't come from NodeAdd() we must have come from NodeUpdate.

Can we assume Node events for a single node are ordered so that NodeAdd always happens before NodeUpdate. I think so. Then there shouldn't be a problem because the NodeAdd will increment the refcnt and the following NodeUpdate could decrement it but that would be from 2->1.

So the case 2 would be: podUpdate -> NodeAdd -> NodeUpdate?

Please correct me I think I missed something in your comment?

pkg/ipcache/ipcache.go
jrfastab: I'm not confident so maybe need an ipcache expert here? Here is my reasoning.

This is the 'cacheModification = Upsert' case  with newHostIP != oldHostIP.  So we got here through this,

```
                // Check whether the deleted endpoint IP was shadowing that CIDR, and                                                                                                                                                                                          
                // restore its mapping with the listeners if that was the case.                                                                                                                                                 
                cidrClusterStr := cidrCluster.String()                                                                                                                                                                                                                                        
                if cidrIdentity, cidrFound := ipc.ipToIdentityCache[cidrClusterStr]; cidrFound {                                                                                                                                                
                        newHostIP, _ = ipc.getHostIPCache(cidrClusterStr)                                                                                                                                                       
                        if cidrIdentity.ID != cachedIdentity.ID || !oldHostIP.Equal(newHostIP) {                                                                                                                                              
                                scopedLog.Debug("Removal of endpoint IP revives shadowed CIDR to identity mapping")                                                                                                                                                                          
                                cacheModification = Upsert                                                                                                                                                                                                                                   
                                cidrIdentity.shadowed = false                                                                                                                                                                                         
                                ipc.ipToIdentityCache[cidrClusterStr] = cidrIdentity                                                                                                                                                                                             
                                oldIdentity = &cachedIdentity                                                                                                                                                                                                                       
                                newIdentity = cidrIdentity          
```
so we got the newHostIP with getHostIPCache which is returning a mapping already in the ipcache? So if we already put it in the ipcache then we don't need to allocate for it because its already been refcnt'd by the thing that put it there. And then when its removed it will either get a nodeDelete or ipcache delete to decrement the refcnt. I'm pretty sure this lines up with experimentation where having an alloc here (which is what I originally did) got some hostIDs stuck, but would need to double check that to be sure.

pkg/datapath/linux/node_ids.go
jrfastab: it should but I'm not sure golang knows how to do const structs?

pkg/datapath/linux/node.go
jrfastab: yep nice catch.

pkg/datapath/linux/node_ids.go
jrfastab: The idea was to have a function to allocate that returns the uint16 and a function to call to get the nodeID struct. Then we should only export the uint16 so that nodeid encapsulates the structure. Anyways I didn't do it very neatly so will clean it up and hopefully it becomes more obvious.

pkg/datapath/linux/node_ids.go
jrfastab: removed it.

pkg/datapath/linux/node_ids.go
jrfastab: seems reasonable to me doing now.

pkg/datapath/linux/node.go
jrfastab: Yeah cleaned up patches so they make sense in order and build correctly.

pkg/ipcache/ipcache.go
pchaigno: > So the case 2 would be: podUpdate -> NodeAdd -> NodeUpdate?

Case 2 is NodeAdd -> PodAdd -> PodUpdate -> NodeUpdate. We only care about the later two.

I detailed how we get into a broken state (in many ways) below. My previous comment wasn't very clear, sorry about that.

---

Let me keep the syntax from the other thread:
- `OldIP` is the IP address that was removed from node A and later reused by node B.
- `NewIP` is the IP address that was added to node A, to replace OldIP.
- `OldNodeID` is the node ID allocated for OldIP.

All is as seen from a node C, different from nodes A and B.

0. Initial state on node C.
    - <ins>State:</ins> `OldNodeID` has refcount 3 (1 for the node, +2 for two pods).
2. A pod update is received with `OldIP` -> `NewIP`.
    - We will decrement refcount for `OldNodeID`. 
    - We will allocate a new node ID for `NewIP`. Let's call it `NewNodeID`.
    - <ins>State:</ins> `OldNodeID` has refcount 2. `NewNodeID` has refcount 1.
    - After this pod update, I believe we will already have XfrmOutPolBlock drops because the ipcache for this pod is updated but the XFRM config isn't. Is that wrong?
3. A node update for node A is received with `OldIP` -> `NewIP`.
    - We will unmap `OldIP` from `OldNodeID`. 
    - We will call `allocateIDForNode(NodeA)`. That will call `getNodeIDForNode(NodeA)`. Nodes have several IP addresses. At this point, node A will have NewIP associated to `NewNodeID`, but it will also have other IP addresses still associated to `OldNodeID`. Which node ID `getNodeIDForNode(NodeA)` will pick is random because it relies on a map iteration.
    - Let's consider `OldNodeID` is picked. Then, in `allocateIDForNode`, all IP addresses of the node, including `NewIP` will be associated to `OldNodeID`. We will also increment the refcount for `OldNodeID`.
    - The XFRM config will remain the same since `OldNodeID` was selected.
    - <ins>State:</ins> `OldNodeID` has refcount 3 (!!). `NewNodeID` has refcount 1.
    - At this point, I believe the ipcache entry for the pod that was updated is still using `NewNodeID` while the XFRM config is using `OldNodeID`. That will result in XfrmOutPolBlock drops.
    - In addition, the refcount for `OldNodeID` is now incorrect (3 vs 1 node + 1 pod) and `OldNodeID` will thus never be deallocated.

We have big issues already, but let's keep going :sweat_smile: 
Let's introduce node B.

4. Node B is created with OldIP as one of its IP addresses.
    - We will call `allocateIDForNode(NodeB)` as part of the node creation. That will map all IPs of node B, including `OldIP`, to |OldNodeID` since the mapping OldIP <> OldNodeID still exists.
    - We will install XFRM configs for node B. Those configs will use `OldNodeID`. Small difference between ENI and overlay here: on ENI, we will overwrite the XFRM configs for node A. On overlay we will install additional entries that use the same node ID but different CIDRs.
    - <ins>State:</ins> `OldNodeID` has refcount 4. `NewNodeID` has refcount 1.
    - Both node A and B are using `OldNodeID` in XFRM configs, but the ipcache entries are a mix of both node IDs.
5. On the creation of a pod on node B.
    - We will create ipcache entries for all pods of node B. They will be using `OldNodeID`.
    - <ins>State:</ins> `OldNodeID` has refcount 5. `NewNodeID` has refcount 1.

jrfastab:
For reference the primary blocking issue that resulted in this patch is how to restore the map on agent restart. We restore the nodeID<->IP mappings from the bpf map, but that currently doesn't include a reference count which would be necessary to work correctly. Then we would need to do some other work to recover if we missed any pod events and correct the refcnt. At minimum it seems the refcnt would have to be part of the BPF map and this convolutes agent concerns and datapath concerns. Overall I don't like it and Paul's solution to just use the bpf nodeid map feels better to me.

margamanterola:
This PR was superseded by https://github.com/cilium/cilium/pull/27029, right? Can it be closed or is there some other case that this would be addressing?

jrfastab:
Correct closing. superseded by #27029

PR #26726, opened 2023-07-10
bpf: nodeport: only set outer src IP for tunnel encap in XDP

julianwiedmann:
As part of introducing native tunnel support for XDP, we started passing a `src_ip` parameter to `__encap_with_nodeid()`. This was needed to manually build the packet's outer headers.

But for TC context we actually don't want to specify the outer src IP, and defer its selection to the kernel stack instead. Worse, specifying the outer src IP means that the skb implementation for `ctx_set_encap_info()` starts to use the `local_ipv4` field in the `bpf_tunnel_key`. And that's not supported on older kernels, resulting in drops.

Fix this by only setting the `src_ip` parameter in XDP context.

Reported-by: Yusuke Suzuki <yusuke-suzuki@cybozu.co.jp>
Fixes: 43dffb298173 ("bpf: encap: manually set src IP and port from nodeport XDP paths")

### DIFF
--- a/bpf/lib/nodeport.h
+++ b/bpf/lib/nodeport.h
@@ -72,6 +72,39 @@ bpf_skip_recirculation(const struct __ctx_buff *ctx __maybe_unused)
 #endif
 }
 
+#ifdef HAVE_ENCAP
+static __always_inline int
+nodeport_add_tunnel_encap(struct __ctx_buff *ctx, __u32 src_ip, __be16 src_port,
+			  __be32 dst_ip, __u32 src_sec_identity, __u32 dst_sec_identity,
+			  enum trace_reason ct_reason, __u32 monitor, int *ifindex)
+{
+	/* Let kernel choose the outer source ip */
+	if (ctx_is_skb())
+		src_ip = 0;
+
+	return __encap_with_nodeid(ctx, src_ip, src_port, dst_ip,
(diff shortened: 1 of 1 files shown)


ysksuzuki (approved):
Thank you for the fix!

I tested this fix with flatcar stable 3510.2.4 and confirmed that it's working.
https://www.flatcar.org/releases#release-3510.2.4

```
$ uname -a
Linux rack0-cs5 5.15.117-flatcar

$ cilium config view
agent-not-ready-taint-key                      node.cilium.io/agent-not-ready
arping-refresh-period                          30s
auto-direct-node-routes                        false
bgp-announce-lb-ip                             true
bpf-ct-timeout-regular-any                     1h0m0s
bpf-ct-timeout-service-any                     1h0m0s
bpf-lb-acceleration                            disabled
bpf-lb-algorithm                               maglev
bpf-lb-dsr-dispatch                            geneve
bpf-lb-dsr-l4-xlate                            backend
bpf-lb-external-clusterip                      false
bpf-lb-map-max                                 65536
bpf-lb-mode                                    dsr
bpf-lb-sock                                    true
bpf-lb-sock-hostns-only                        true
bpf-map-dynamic-size-ratio                     0.0025
bpf-policy-map-max                             65536
bpf-root                                       /sys/fs/bpf
cgroup-root                                    /sys/fs/cgroup
```

gentoo-root (approved):

PR #26728, opened 2023-07-10
service: respond to healthchecks on LBips

### LINKED ISSUE #26727: GCP healthchecks fail in KubeProxyReplacement mode

### What happened?

Clusters hosted on GCP in kube-proxy-less mode, setting up a loadbalancer-service using `externalTrafficPolicy: Local` triggers failures in GCP healthchecks. This leads to ingress traffic being distributed among all Kubernetes nodes.

The [cloud-provider-gcp](https://github.com/kubernetes/cloud-provider-gcp) sets up these healthchecks based on the services' `spec.healthCheckNodePort`. This port is made accessible by Cilium on the nodeIP. However, GCP healthchecks have a destination IP towards the forwarding rule IP (equivalent to `spec.loadBalancerIP`). This results in the GCP LB randomly sends these packets to any nodes within the GCP instance group.

The problem emerges since Cilium's HealthServer binds to local IPs and no forwarding rules for the LBips, hence, it doesn't acknowledge the GCP LB healthcheck packets. It's noteworthy that Kube-Proxy has [forwarding rules](https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/ipvs/proxier.go#L1683) set up for such scenarios.

### Cilium Version

1.13.4

### Kernel Version

6.1.35

### Kubernetes Version

1.26.6

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

See https://github.com/kubernetes/kubernetes/issues/117621#issuecomment-1531059530 for a packetcapture of this behaviour. And a more detailed discussion on what GCP healthchecks are doing.

I will create a PR to fix this issue

### Code of Conduct

nberlee:
This Pull Request introduces the 'EnableHealthCheckLoadBalancerIP' flag, which exposes the healthCheckNodePort on the LoadBalancerIP (LB IP). This addition aims to resolve an issue where clusters hosted on GCP in KubeProxyReplacement mode, using a loadbalancer-service with `externalTrafficPolicy: Local`, were triggering failures in GCP health checks. This, in turn, was causing ingress traffic to be distributed among all Kubernetes nodes, which is not compatible with the Local traffic policy.

Specifically, GCP's Load Balancer performs the health checks using the LB IP as the destination IP rather than the node IP. While the cloud-provider-gcp sets up these healthchecks based on the services' spec.healthCheckNodePort which is accessible by Cilium on the nodeIP, it doesn't acknowledge the GCP LB healthcheck packets because Cilium's HealthServer binds to local IPs and doesn't have forwarding rules for the LB IPs.

      GitHub issue.

Fixes: #26727

```release-note
Added the EnableHealthCheckLoadBalancerIP flag to address health checks on LoadBalancerIP in Google Cloud Platform using KubeProxyReplacement.
```

### DIFF
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -237,6 +237,9 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	flags.Bool(option.EnableHealthCheckNodePort, defaults.EnableHealthCheckNodePort, "Enables a healthcheck nodePort server for NodePort services with 'healthCheckNodePort' being set")
 	option.BindEnv(vp, option.EnableHealthCheckNodePort)
 
+	flags.Bool(option.EnableHealthCheckLoadBalancerIP, defaults.EnableHealthCheckLoadBalancerIP, "Enable access of the healthcheck nodePort on the LoadBalancerIP. Needs --enable-health-check-nodeport to be enabled")
+	option.BindEnv(vp, option.EnableHealthCheckLoadBalancerIP)
+
 	flags.StringSlice(option.EndpointStatus, []string{},
 		"Enable additional CiliumEndpoint status features ("+strings.Join(option.EndpointStatusValues(), ",")+")")
 	option.BindEnv(vp, option.EndpointStatus)
--- a/install/kubernetes/cilium/templates/cilium-configmap.yaml
+++ b/install/kubernetes/cilium/templates/cilium-configmap.yaml
@@ -697,6 +697,11 @@ data:
 {{- if hasKey .Values.nodePort "enableHealthCheck" }}
   enable-health-check-nodeport: {{ .Values.nodePort.enableHealthCheck | quote}}
 {{- end }}
+{{- if .Values.gke.enabled }}
+  enable-health-check-loadbalancer-ip: "true"
+{{- else if hasKey .Values.nodePort "enableHealthCheckLoadBalancerIP" }}
+  enable-health-check-loadbalancer-ip: {{ .Values.nodePort.enableHealthCheckLoadBalancerIP | quote}}
+{{- end }}
   node-port-bind-protection: {{ .Values.nodePort.bindProtection | quote }}
   enable-auto-protect-node-port-range: {{ .Values.nodePort.autoProtectPortRange | quote }}
 {{- end }}
--- a/pkg/service/service.go
+++ b/pkg/service/service.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"net/netip"
 	"strconv"
 	"sync/atomic"
 	"time"
@@ -28,6 +29,7 @@ import (
 	"github.com/cilium/cilium/pkg/maps/lbmap"
 	"github.com/cilium/cilium/pkg/metrics"
 	monitorAPI "github.com/cilium/cilium/pkg/monitor/api"
+	"github.com/cilium/cilium/pkg/node"
 	nodeTypes "github.com/cilium/cilium/pkg/node/types"
 	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/service/healthserver"
@@ -93,6 +95,7 @@ type svcInfo struct {
 	sessionAffinity           bool
 	sessionAffinityTimeoutSec uint32
 	svcHealthCheckNodePort    uint16
+	healthcheckFrontendHash   string
 	svcName                   lb.ServiceName
 	loadBalancerSourceRanges  []*cidr.CIDR
 	l7LBProxyPort             uint16   // Non-zero for egress L7 LB services
@@ -697,7 +700,7 @@ func (s *Service) upsertService(params *lb.SVC) (bool, lb.ID, error) {
 	// Only add a HealthCheckNodePort server if this is a service which may
 	// only contain local backends (i.e. it has externalTrafficPolicy=Local)
 	if option.Config.EnableHealthCheckNodePort {
-		if svc.isExtLocal() && filterBackends {
+		if svc.isExtLocal() && filterBackends && svc.svcHealthCheckNodePort > 0 {
 			// HealthCheckNodePort is used by external systems to poll the state of the Service,
 			// it should never take into consideration Terminating backends, even when there are only
 			// Terminating backends.
@@ -709,11 +712,24 @@ func (s *Service) upsertService(params *lb.SVC) (bool, lb.ID, error) {
 			}
 			s.healthServer.UpsertService(lb.ID(svc.frontend.ID), svc.svcName.Namespace, svc.svcName.Name,
 				activeBackends, svc.svcHealthCheckNodePort)
+
+			if err = s.upsertNodePortHealthService(svc, &nodeMetaCollector{}); err != nil {
+				return false, lb.ID(0), fmt.Errorf("upserting NodePort health service failed: %w", err)
+			}
+
 		} else if svc.svcHealthCheckNodePort == 0 {
 			// Remove the health check server in case this service used to have
 			// externalTrafficPolicy=Local with HealthCheckNodePort in the previous
 			// version, but not anymore.
 			s.healthServer.DeleteService(lb.ID(svc.frontend.ID))
+
+			if svc.healthcheckFrontendHash != "" {
+				healthSvc := s.svcByHash[svc.healthcheckFrontendHash]
+				if healthSvc != nil {
+					s.deleteServiceLocked(healthSvc)
+				}
+				svc.healthcheckFrontendHash = ""
+			}
 		}
 	}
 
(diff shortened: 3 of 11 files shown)


ghost (approved):
docs (autogenerated) ok

install/kubernetes/cilium/templates/cilium-configmap.yaml
```diff
@@ -429,6 +429,7 @@ data:
   routing-mode: "native"
   enable-endpoint-routes: "true"
   enable-local-node-route: "false"
+  enable-health-check-loadbalancer-ip: "true"
```
kaworu: Seems like this section of the ConfigMap is for routing / tunneling options, see my next comment.

install/kubernetes/cilium/templates/cilium-configmap.yaml
```diff
@@ -692,6 +693,9 @@ data:
 {{- end }}
 {{- if hasKey .Values.nodePort "enableHealthCheck" }}
   enable-health-check-nodeport: {{ .Values.nodePort.enableHealthCheck | quote}}
+{{- end }}
+{{- if hasKey .Values.nodePort "enableHealthCheckLoadBalancerIP" }}
+  enable-health-check-loadbalancer-ip: {{ .Values.nodePort.enableHealthCheckLoadBalancerIP | quote}}
```
kaworu: IMO it would be cleaner to have all value logic tied to `enable-health-check-loadbalancer-ip` here, like e.g for `install-no-conntrack-iptables-rules`

https://github.com/cilium/cilium/blob/8e37f1fc845c8148bae77df3c68751111b813f8a/install/kubernetes/cilium/templates/cilium-configmap.yaml#L595-L600

```suggestion
{{- end }}
{{- if .Values.gke.enabled }}
  enable-health-check-loadbalancer-ip: "true"
{{- else if hasKey .Values.nodePort "enableHealthCheckLoadBalancerIP" }}
  enable-health-check-loadbalancer-ip: {{ .Values.nodePort.enableHealthCheckLoadBalancerIP | quote}}
```

kaworu (changes requested):
Thanks for the PR @nberlee, I left a cosmetic comment but other than that Helm changes LGTM.

install/kubernetes/cilium/templates/cilium-configmap.yaml
nberlee: Added your suggestion in latest commit

kaworu (approved):
Thanks for the update @nberlee, Helm changes LGTM now.

daemon/cmd/daemon_main.go
```diff
@@ -237,6 +237,9 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	flags.Bool(option.EnableHealthCheckNodePort, defaults.EnableHealthCheckNodePort, "Enables a healthcheck nodePort server for NodePort services with 'healthCheckNodePort' being set")
 	option.BindEnv(vp, option.EnableHealthCheckNodePort)
 
+	flags.Bool(option.EnableHealthCheckLoadBalancerIP, defaults.EnableHealthCheckLoadBalancerIP, "Enable access of the healthcheck nodePort on the LoadBalancerIP. Needs 'EnableHealthCheckNodePort' to be enabled")
```
asauber:
```suggestion
	flags.Bool(option.EnableHealthCheckLoadBalancerIP, defaults.EnableHealthCheckLoadBalancerIP, "Enable access of the healthcheck nodePort on the LoadBalancerIP. Needs --enable-health-check-nodeport to be enabled")
```

pkg/service/service.go
```diff
@@ -729,6 +744,72 @@ func (s *Service) upsertService(params *lb.SVC) (bool, lb.ID, error) {
 	return new, lb.ID(svc.frontend.ID), nil
 }
 
+// upsertHealthService makes the HealthCheckNodePort available to the external IP of the service.
+func (s *Service) upsertHealthService(svc *svcInfo) error {
```
asauber: Would this be better named `upsertNodePortHealthService`?

nberlee: Probably, I will fix it

nberlee: Fixed in https://github.com/cilium/cilium/compare/f56980fca4f8a79d20f1453ea975e825e049c111..f240423a10e417a7e1ef3259f40c9cf221586c28

daemon/cmd/daemon_main.go
nberlee: fixed in https://github.com/cilium/cilium/compare/f56980fca4f8a79d20f1453ea975e825e049c111..f240423a10e417a7e1ef3259f40c9cf221586c28

pkg/service/service.go
```diff
@@ -729,6 +744,72 @@ func (s *Service) upsertService(params *lb.SVC) (bool, lb.ID, error) {
...
+		IntTrafficPolicy: lb.SVCTrafficPolicyLocal,
+		Backends:         healthCheckBackends,
+		LoopbackHostport: true,
+	}
+	_, _, err := s.upsertService(healthCheckSvc)
+	if err != nil {
+		return err
+	}
+	svc.healthcheckFrontendHash = healthCheckFrontend.Hash()
+	log.Debugf("Created healtcheck service %s for frontend", healthCheckSvcName.Name)
```
joamaki: s/healtcheck/healthcheck/. Log using the same pattern as elsewhere in pkg/service:
```
	log.WithFields(logrus.Fields{
		logfields.ServiceName:      serviceName.Name,
		logfields.ServiceNamespace: serviceName.Namespace,
	}).Debug("Created ...")
```

pkg/service/service.go
```diff
@@ -710,11 +713,23 @@ func (s *Service) upsertService(params *lb.SVC) (bool, lb.ID, error) {
 			}
 			s.healthServer.UpsertService(lb.ID(svc.frontend.ID), svc.svcName.Namespace, svc.svcName.Name,
 				activeBackends, svc.svcHealthCheckNodePort)
+
+			if err = s.upsertNodePortHealthService(svc); err != nil {
+				return false, lb.ID(0), err
```
joamaki: Please wrap `err` with `fmt.Errorf` to add some more context, e.g. `fmt.Errorf("upserting NodePort health service failed: %w", err)`. Would be otherwise hard to figure out where the error is coming from when it's eventually logged.

joamaki (changes requested):
Few nits, but other than that implementation looks good to me.
I would however like to see a test for this. It should be fairly straightforward to add to
`pkg/service/service_test.go`. `TestHealthCheckNodePort` is likely a good starting point.
Remember to restore `option.Config.EnableHealthCheckLoadBalancerIP` after the test case.

nberlee:
> I would however like to see a test for this. It should be fairly straightforward to ...

I agree that a test would be nice, but its not straightforward for me when it come to mocking `node.GetIPv4().To4()` / `node.GetIPv6()`

pkg/service/service.go
```diff
@@ -729,6 +744,72 @@ func (s *Service) upsertService(params *lb.SVC) (bool, lb.ID, error) {
 	return new, lb.ID(svc.frontend.ID), nil
 }
 
+// upsertNodePortHealthService makes the HealthCheckNodePort available to the external IP of the service
+func (s *Service) upsertNodePortHealthService(svc *svcInfo) error {
```
asauber: One approach to adding a test here would be to parameterize this function with a new interface, defined in this file, similar to the following

```
type NodeMetaCollector interface {
	GetIPv4() net.IP
	GetIPv6() net.IP
}

type nodeMetaCollector struct{}

func (n *nodeMetaCollector) GetIPv4() net.IP {
	return node.GetIPv4()
}

func (n *nodeMetaCollector) GetIPv6() net.IP {
	return node.GetIPv6()
}
```

A test could then provide an alternate implementation for `NodeMetaCollector`

Ideally, this package, and the node package, would be refactored into a `hive.Cell` in order to allow dependency injection of `node`. However, that's not necessary for the purposes of this PR.

pkg/service/service.go
nberlee: Fixed

pkg/service/service.go
nberlee: Fixed

pkg/service/service.go
nberlee: Ty for showing the way @asauber, I have created the tests that @joamaki suggested

nberlee:
@joamaki friendly reminder :)
I've added the requested tests and have changed the requested changes

thorn3r (approved):
agent changes lgtm :+1: nice work!

joamaki (approved):

nberlee:
Great that the PR is approved, could anymody issue a
/test
Please? also I do not think the `needs-rebase` label is warranted anymore. (rebased yesterday again)

asauber (approved):

PR #26729, opened 2023-07-10
ci: skip post-test info gathering in case of skipped cilium installation

mhofstetter:
Currently, getting the Cilium status (in steps like "Post-test information gathering" (or similar ones)) always fail if the Cilium installation is skipped due to missing images. As a consequence, these steps are confusingly listed in CI analysis.

IMO we should try to avoid having "post-processing" steps fail due to some missing pre-requisites (like Cilium being installed).

Therefore, this commit skips these steps if Cilium installation has been skipped.

### DIFF
--- a/.github/workflows/conformance-aks.yaml
+++ b/.github/workflows/conformance-aks.yaml
@@ -221,6 +221,7 @@ jobs:
           persist-credentials: false
 
       - name: Install Cilium
+        id: install-cilium
         run: |
           cilium install ${{ steps.vars.outputs.cilium_install_defaults }}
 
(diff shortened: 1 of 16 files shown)


mhofstetter:
Current state of this PR illustrates the failing tests (due to cancelled image builds) which results in skipped post-processing steps.

will properly run the tests if we decide that we want to go this way.

aanm (approved):

viktor-kurchenko (approved):
LGTM.

mhofstetter:
MLH still seems to be asleep - adding `ready-to-merge`manually

Issue #26350, opened 2023-06-19
Testing IPsec key rotation


pchaigno:
We want to test IPsec key rotation in our pull request CI. The test should deploy an application sensitive to drops (e.g., `test/k8s/manifests/migrate-svc-*.yaml`), perform the key rotation (https://docs.cilium.io/en/latest/security/network/encryption-ipsec/#key-rotation), and check:
- That the drop-sensitive application didn't restart (it restarts if there is any drop).
- That the CLI's connectivity tests pass.
- That there are no errors in logs.
- That the XFRM counters didn't increase.

pchaigno:
This isn't exactly fixed yet. Workflows still need to be updated.

PR #26730, opened 2023-07-10
CI: Move IPsec CI jobs into separate pipelines

jschwinger233:
This PR moves IPsec test cases from `conformance-e2e.yaml` to `conformance-ipsec-e2e.yaml`. It would be beneficial to separate the IPsec test from normal ones as we will have more test specific for IPsec (e.g. #26350 ) and more IPsec features.

### DIFF
--- a/.github/workflows/conformance-ipsec-e2e.yaml
+++ b/.github/workflows/conformance-ipsec-e2e.yaml
@@ -1,4 +1,4 @@
-name: Cilium Conformance E2E (ci-e2e)
+name: Cilium Conformance IPsec E2E (ci-ipsec-e2e)
 
 # Any change in triggers needs to be reflected in the concurrency group.
 on:
@@ -52,7 +52,7 @@ concurrency:
 
 env:
   # renovate: datasource=github-releases depName=cilium/cilium-cli
-  cilium_cli_version: v0.15.4
+  cilium_cli_version: v0.14.8
   cilium_cli_ci_version:
   check_url: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
 
@@ -67,7 +67,6 @@ jobs:
 
   setup-and-test:
     runs-on: ubuntu-latest-4cores-16gb
-    name: 'Setup & Test'
     env:
       job_name: 'Setup & Test'
     strategy:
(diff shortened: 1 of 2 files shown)


aanm (approved):

giorio94:
I'm personally a bit against moving the IPSec tests away from the clustermesh workflow, unless there's a strong reason. I mean, I perfectly understand the rationale behind doing this change for other workflows, but the goal of the clustermesh ones is verify that cross-cluster communication works in a variety of scenarios

giorio94:
I'm personally against moving the IPSec tests away from the clustermesh workflow, unless there's a strong reason, as it increases the overall maintenance burden and likelihood of drifts. I mean, I perfectly understand the rationale behind doing this change for other workflows, but the goal of the clustermesh ones is to verify that cross-cluster communication works in a variety of scenarios, for different control-plane and datapath configurations. Detailed datapath tests are expected to be performed in other workflows (such as conformance E2E) as at the end of the day remote nodes are treated pretty much in the same way as the local ones.

pchaigno:
@jschwinger233 Did you test the two changed workflows? You might want to reopen this with a branch on cilium/cilium if you haven't already.

.github/workflows/conformance-ipsec-e2e.yaml
```diff
@@ -0,0 +1,326 @@
...
+      - name: Checkout context ref
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          ref: ${{ inputs.context-ref || github.sha }}
+          persist-credentials: false
+
+      - name: Set Environment Variables
+        uses: ./.github/actions/set-env-variables
+
+      - name: Set up job variables
```
brb, line 124: Maybe we could benefit from https://docs.github.com/en/actions/using-workflows/reusing-workflows to avoid the same code in ci-e2e, ci-ipsec and upcoming ci-upgrade?

christarazi (approved):
Pending Martynas's suggestion

.github/workflows/conformance-ipsec-e2e.yaml
pchaigno: I don't think we should block an improvement on making it a bigger improvement. Let's ship this and iterate after.

jschwinger233: Agree with Paul. Let me figure out how to make good use of reusing-workflows for https://github.com/cilium/cilium/pull/26983 after.

brb: Yeah, we can do it as a follow-up.

jschwinger233:
Two effected workflows (`ci-ipsec-e2e` and `ci-e2e`) has been tested from https://github.com/cilium/cilium/pull/27145, both have passed:
ci-ipsec-e2e: https://github.com/cilium/cilium/actions/runs/5715863236
ci-e2e: https://github.com/cilium/cilium/actions/runs/5711043643

jschwinger233:
Only ci-e2e and ci-ipsec-e2e matter. Since both tests are green, I labelled this PR as ready-to-merge.

PR #26732, opened 2023-07-10
daemon: fix issue where IPAM options in custom CNI confs was ignored

### LINKED ISSUE #26731: Custom IPAM section in CNI config is ignored
@joamaki stumbled upon the fact that NetConf isn't set anymore in the daemon: https://github.com/cilium/cilium/blob/main/daemon/cmd/daemon.go#L405. This is passed down to NodeDiscovery which then pulls some IPAM related things out of it.

This breaks the ability to configure IPAM parameters such as e.g. the ones used for ENI [documented here](https://docs.cilium.io/en/stable/network/concepts/ipam/eni/#custom-eni-configuration). Since this is a critical ability for many ENI users, this is a release-blocking bug that we need to address. The IPAM parameters need to be present in the CiliumNode CRD when it is created, otherwise the operator will take meaningful decisions (e.g. attaching network interface from the wrong subnet) based on absent configuration.

In addition, we should probably also add some CI testing around this, for example by using custom subnets in the CI-EKS workflow.

squeed:
As part of a CNI refactor, a bug was introduced where node-specific IPAM overrides were no longer being loaded. Specifically, the parsed network configuration wasn't properly being passed down to the NodeDiscovery manager.

Fixes: #26731
Fixes: 1254bf403f

### DIFF
--- a/daemon/cmd/cni/config.go
+++ b/daemon/cmd/cni/config.go
@@ -47,24 +47,32 @@ type cniConfigManager struct {
 // GetMTU returns the MTU as written in the CNI configuration file.
 // This is one way to override the node's MTU.
 func (c *cniConfigManager) GetMTU() int {
-	if c.config.ReadCNIConf == "" {
+	conf := c.GetCustomNetConf()
+	if conf == nil {
 		return 0
 	}
-
-	conf, err := cnitypes.ReadNetConf(c.config.ReadCNIConf)
-	if err != nil {
-		c.log.WithField("path", c.config.ReadCNIConf).WithError(err).Warnf("Failed to parse existing CNI configuration file")
-	}
-
-	if conf.MTU > 0 {
-		return conf.MTU
-	}
-	return 0
+	return conf.MTU
 }
 
 // GetChainingMode returns the configured chaining mode.
 func (c *cniConfigManager) GetChainingMode() string {
 	return c.config.CNIChainingMode
+}
+
+// GetCustomNetConf returns the parsed custom CNI configuration, if provided
+// (In other words, the value to --read-cni-conf).
+// Otherwise, returns nil.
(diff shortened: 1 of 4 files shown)


squeed:
tentatively setting this as release-note misc, since we will (hopefully) ship no releases with this label!

joamaki (approved):

daemon/cmd/cni/config.go
```diff
@@ -67,6 +67,22 @@ func (c *cniConfigManager) GetChainingMode() string {
 	return c.config.CNIChainingMode
 }
 
+// GetCustomNetConf returns the parsed custom CNI configuration, if provided
+// (In other words, the value to --read-cni-conf).
+// Otherwise, returns nil.
+func (c *cniConfigManager) GetCustomNetConf() *cnitypes.NetConf {
```
gentoo-root, line 65: GetMTU can reuse this new function, the code is almost the same.

daemon/cmd/daemon.go
```diff
@@ -414,11 +407,6 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 		return nil, nil, fmt.Errorf("CRD Identity allocation mode requires k8s to be configured")
 	}
 
-	if mtu := params.CNIConfigManager.GetMTU(); mtu > 0 {
```
gentoo-root, line 417: If GetMTU returned 0, configuredMTU used to be set to option.Config.MTU, but not anymore after this patch. Is it intentional?

squeed: no, that's a bug. Oops

squeed: Trying to be too clever, I'll back out the MTU changes.

squeed:
Thanks for the review; fixed.

gentoo-root (approved):

squeed:
All green, merging

PR #26741, opened 2023-07-10
Egress Gateway: make CiliumEndpoint reconciliation asynchronous from k8s watcher 

### LINKED ISSUE #23976: Pod identity propagation delay to remote IP cache when using egress gateways
We seem to be experiencing some problems with Cilium Endpoint propagation on a GKE cluster with 2600 pods, 17 nodes and 80 cronjobs. Short-lived pods either created manually or via a cronjob sometimes experience connectivity problems when connecting to other internal services protected by network policies.

The failures often happen when there is a high pod churn on the cluster, this might be due to a number of cronjobs triggering at once or pods being deleted/created at the time the affected pod spawns.

The large number of pod updates seem to cause propagation delays of new pod IPs, which leads to the cilium IP cache not being updated in time on the remote nodes. If the delay is long enough it seems that the cache is not updated before the initial network connection is attempted, this leads to cilium classifying the identity of the cronjob pod as `reserved:world` which then causes the connection to be blocked by the network policy in that namespace. For example from cilium monitor logs during a failure event:

```
Policy verdict log: flow 0x0 local EP ID 902, remote ID world, proto 6, ingress, action deny, match none, 10.4.36.247:52942 -> 10.4.58.66:3306 tcp SYN
xx drop (Policy denied) flow 0x0 to endpoint 902, file bpf_lxc.c line 2003, , identity world->48169: 10.4.36.247:52942 -> 10.4.58.66:3306 tcp SYN
```

vs when the cronjob connects as expected:

```jsx
Policy verdict log: flow 0x0 local EP ID 902, remote ID 35916, proto 6, ingress, action allow, match L3-Only, 10.4.31.130:43190 -> 10.4.58.66:3306 tcp SYN
```

The network policy in that namespace looks like so:

```jsx
kind: NetworkPolicy
metadata:
  name: ns-block
  namespace: ns-a
spec:
  ingress:
  - from:
    - namespaceSelector:
        matchLabels:
          netpolicylabel: ns-a
  - from:
    - namespaceSelector:
        matchLabels:
          name: istio
  - from:
    - namespaceSelector:
        matchLabels:
          name: forwarding
  - from:
    - namespaceSelector:
        matchLabels:
          name: logging
  podSelector: {}
  policyTypes:
  - Ingress
status: {}
```

I wrote a quick tool to watch the kubernetes event stream for cilium endpoint resource changes and then run `cilium bpf ipcache get <ip-addr>` on each node in the cluster to determine how long the propagation takes. For the majority of the time the delay is below 1 second, but in the event of 30/50/100 cilium endpoint updates the propagation delays can become in excess of 10s, leading to the issue.

We recently upgraded cilium to 1.13.0 and also reduced the identity labels at the same time as this upgrade. This has improved the situation and the problem occurs now less frequently, but it is still something which we would prefer to avoid happening at all.

Since the upgrade I have noticed that the failure timestamps for the jobs correspond with a large increase of:

`2023-02-21T12:30:06.348176806Z stderr F level=info msg="Trace[1237947675]: \"DeltaFIFO Pop Process\" ID:ns-a/test-db-swzfx,Depth:15,Reason:slow event handlers blocking the queue (21-Feb-2023 12:30:05.768) (total time: 579ms):" subsys=klog`

log lines, which is demonstrated in the below image:

image

The green line is the rate of the `slow event handlers blocking the queue` logs and the yellow line corresponds to the creation/deletion logs of some of the recently failed jobs.

I believe the slow event handlers are down to the time it takes to process the updates in the cilium egress gateway related code. At the time of the second set of failures (the rightmost red highlight on the graph) a flamegraph of the cilium agent process shows a lot of time spent in `egressgateway.(*Manager).OnUpdateEndpoint` and `egressgateway.(*Manager).OnDeleteEndpoint` functions.

image

Here is the left side of the graph following a cilium endpoint `OnAdd`:

image

and the right side following a cilium endpoint `OnDelete`:

image

Regarding the egress gateway setup, we have two cilium egress gateway entries on the cluster which are responsible for SNATing the connections for ~40 pods, with one node acting as the designated gateway. The designated node runs no workloads apart from the daemonsets on the clusters, the cilium daemonset and any kubernetes components.

jibi:
see individual commits

Fixes: #23976

### DIFF
--- a/pkg/egressgateway/manager.go
+++ b/pkg/egressgateway/manager.go
@@ -66,6 +66,11 @@ const (
 	eventDeleteEndpoint
 )
 
+type endpointEvent struct {
+	eventType eventType
+	endpoint  *k8sTypes.CiliumEndpoint
+}
+
 type Config struct {
 	// Install egress gateway IP rules and routes in order to properly steer
 	// egress gateway traffic to the correct ENI interface
@@ -117,7 +122,11 @@ type Manager struct {
 	// just received the event, or because the processing failed due to the
 	// manager being unable to resolve the endpoint identity to a set of
 	// labels
-	pendingEndpointEvents map[endpointID]*k8sTypes.CiliumEndpoint
+	pendingEndpointEvents map[endpointID]endpointEvent
+
+	// pendingEndpointEventsLock protects the access to the
+	// pendingEndpointEvents map
+	pendingEndpointEventsLock lock.RWMutex
 
 	// endpointEventsQueue is a workqueue of CiliumEndpoint IDs that need to
 	// be processed by the manager
@@ -179,7 +188,7 @@ func NewEgressGatewayManager(p Params) (*Manager, error) {
 		policyConfigs:                 make(map[policyID]*PolicyConfig),
 		policyConfigsBySourceIP:       make(map[string][]*PolicyConfig),
 		epDataStore:                   make(map[endpointID]*endpointMetadata),
-		pendingEndpointEvents:         make(map[endpointID]*k8sTypes.CiliumEndpoint),
+		pendingEndpointEvents:         make(map[endpointID]endpointEvent),
 		endpointEventsQueue:           endpointEventRetryQueue,
 		identityAllocator:             p.IdentityAllocator,
 		installRoutes:                 p.Config.InstallEgressGatewayRoutes,
@@ -306,7 +315,18 @@ func (manager *Manager) processCiliumEndpoints(ctx context.Context, wg *sync.Wai
 			}
 			endpointID := item.(types.NamespacedName)
 
-			manager.addEndpoint(endpointID)
+			manager.pendingEndpointEventsLock.RLock()
+			epEvent, ok := manager.pendingEndpointEvents[endpointID]
+			manager.pendingEndpointEventsLock.RUnlock()
+
+			if ok {
+				switch epEvent.eventType {
+				case eventUpdateEndpoint:
+					manager.addEndpoint(endpointID)
+				case eventDeleteEndpoint:
+					manager.deleteEndpoint(endpointID)
+				}
+			}
 
 			if manager.endpointEventIsPending(endpointID) {
 				// if the endpoint event is still pending it means the manager
@@ -369,8 +389,8 @@ func (manager *Manager) OnDeleteEgressPolicy(configID policyID) {
 }
 
 func (manager *Manager) endpointEventIsPending(id types.NamespacedName) bool {
-	manager.Lock()
-	defer manager.Unlock()
+	manager.pendingEndpointEventsLock.RLock()
+	defer manager.pendingEndpointEventsLock.RUnlock()
 
 	_, ok := manager.pendingEndpointEvents[id]
 	return ok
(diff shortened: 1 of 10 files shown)


kaworu (approved):
Helm changes LGTM.

pkg/egressgateway/manager_privileged_test.go
```diff
@@ -136,7 +136,7 @@ func (k *EgressGatewayTestSuite) SetUpTest(c *C) {
 		egressmap.Cell,
 		cell.Provide(NewEgressGatewayManager),
 		cell.Provide(
-			func() Config { return Config{true} },
+			func() Config { return Config{true, 1 * time.Millisecond} },
```
julianwiedmann, line 139: This could still result in CI flakes, right? We're checking for the new state, but the agent hasn't caught up yet.

Thinking we might still want a sync-mode (even only for testing), or wait for some signal that the agent has gone through one reconciliation.

pkg/egressgateway/manager.go
```diff
@@ -185,7 +236,23 @@ func NewEgressGatewayManager(p Params) *Manager {
...
+		manager.eventsBitmap |= e
+	}
+}
+
+func (manager *Manager) eventBitmapIsSet(events ...eventType) bool {
+	for _, e := range events {
+		if manager.eventsBitmap&e != 0 {
+			return true
+		}
+	}
```
julianwiedmann, line 253: You're telling me there is no `return (manager.eventsBitmap & mask) != 0` ? :smile:

julianwiedmann (approved):
Thank you! One CI-related comment below, would be good to understand if we have issues lurking there.

pkg/egressgateway/manager.go
jibi: eh, you have to build it first that `mask` :see_no_evil: (now, we could discuss if `eventBitmapIsSet` _really_ needs to be variadic, but I'd say it makes the invocation a tiny bit more readable when you have multiple events)

pkg/egressgateway/manager_privileged_test.go
jibi: > This could still result in CI flakes, right? We're checking for the new state, but the agent hasn't caught up yet.

no, I had to solve this already in a past/related PR:

https://github.com/cilium/cilium/blob/2899038e494c61d9686c7a35d8823878ddae74bd/pkg/egressgateway/manager_privileged_test.go#L643-L653

since in https://github.com/cilium/cilium/pull/26457 I changed `OnUpdateEndpoint` to be not fully synced anymore, I had to update the assertions in the unit tests to allow for a few retries

julianwiedmann: May the retry gods have mercy with our CI :pray: .

pkg/egressgateway/manager.go
```diff
@@ -125,6 +142,20 @@ type Manager struct {
...
 	// policyMap communicates the active policies to the dapath.
 	policyMap egressmap.PolicyMap
+
+	// reconciliationTriggerInterval is the amount of time between triggers
+	// of reconciliations are invoked
+	reconciliationTriggerInterval time.Duration
+
+	// eventsBitmap is a bitmap that tracks which type of events has been
+	// received by the manager (e.g. node added or policy removed) since the
+	// last invokation of the reconciliation logic
```
qmonnet:
```suggestion
	// last invocation of the reconciliation logic
```

Documentation/spelling_wordlist.txt
```diff
@@ -915,6 +915,7 @@ rebase
 rebased
 rebasing
 recompiles
+reconciliationTriggerInterval
```
qmonnet: We need to stop this madness, I've created https://github.com/cilium/cilium/pull/26759. But OK for this PR :+1:

qmonnet (approved):
One spelling nit, but looks good otherwise from my side

Documentation/spelling_wordlist.txt
qmonnet: Sorry, #26759 was merged and created a conflict for your PR :grimacing:. Can you please rebase, re-generate the Helm value update, and remove this new entry from the `spelling_wordlist.txt`?

jibi: Updated :heavy_check_mark:

aanm (approved):

qmonnet:
Doc workflow failure is my fault, should be fixed with https://github.com/cilium/cilium/pull/26780.

PR #26742, opened 2023-07-10
map: fix reconciliation failure caused by out of sync errors number

giorio94:
Cached maps come with a controller that retries Update/Delete operations in case an error occurred while synchronizing the given entry with the kernel. Currently, it relies on the `outstandingErrors` counter to determine whether reconciliation is necessary, as well as if any entry still needs to be processed in a subsequent operation.

Yet, it is possible that this counter gets out of sync, in particular in case the given error is resolved automatically when a subsequent operation acting on the same key succeed. If this happens, the reconciliation function will never complete successfully, as there will continue to be an error that cannot be resolved (as it no longer exists). The given controller will this continue failing forever until the agent gets restarted.

Let's fix this inferring the number of outstanding errors from the cache itself. The `outstandingErrors` variable is preserved (although converted to a boolean) to avoid iterating over the full cache in case it is known that no error occurred. Still, if this flag gets out of sync, the only consequence will be that the cache is iterated once to determine that there's actually no failure, ensuring that the reconciliation logic still converges properly.

```release-note
Fix issue which caused the map reconciliation process to never complete successfully if the error resolved automatically
```

### DIFF
--- a/pkg/bpf/map_linux.go
+++ b/pkg/bpf/map_linux.go
@@ -84,9 +84,12 @@ type Map struct {
 	// was last scheduled
 	errorResolverLastScheduled time.Time
 
-	// outstandingErrors is the number of outsanding errors syncing with
-	// the kernel
-	outstandingErrors int
+	// outstandingErrors states whether there are outstanding errors, occurred while
+	// syncing an entry with the kernel, that need to be resolved. This variable exists
+	// to avoid iterating over the full cache to check if reconciliation is necessary,
+	// but it is possible that it gets out of sync if an error is automatically
+	// resolved while performing a subsequent Update/Delete operation on the same key.
+	outstandingErrors bool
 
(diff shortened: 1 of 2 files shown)


giorio94:
> @giorio94 Thanks for the patch! Would you be able to craft a regression test for this? This code may undergo significant changes throughout 1.15/1.16.

I've added a unit test to assert that the error resolver works properly (and succeeds) both in case an element needs to be synchronized, and in case the issue already resolved automatically. The second test consistently fails without the patch introduced in this commit. @ti-mo PTAL.

giorio94:
Hey @ti-mo, would you have time to take another look at this?

ti-mo (approved):
Not super familiar with this code in particular, but the approach and code seem reasonable to me. I don't want to hold this up any longer, thanks for adding the test!

PR #26745, opened 2023-07-11
bgpv1: move Signaler to BGPCPSignaler cell

ldelossa:
In order to promote modularity, move BGPCPSignaler to its own cell.

This will allow other subsystems to add this Cell to their dependencies without relying on the BGP-CP-Controller.

Subsystems can then use this event source as a way to notify the BGP-CP it should perform a reconciliation.

```release-note
bpgv1: move the internal BGP signaler to a cell and allow other cells to depend on it.
```

### DIFF
--- a/pkg/bgpv1/agent/controller.go
+++ b/pkg/bgpv1/agent/controller.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/cilium/workerpool"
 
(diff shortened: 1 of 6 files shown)


rastislavs (approved):
LGTM

dylandreimerink (approved):
LGTM :rocket:

PR #26747, opened 2023-07-11
proxy: remove unused xds resource access timeout

mhofstetter:
xds cache acepts a resource access timeout which isn't actually used.

It's only used as a timeout on a context.Context that isn't used in the code path.

Therefore, this commit deleted the parameter.


jrajahalme (approved):
Makes sense as `GetResources()` was not using the `ctx` parameter, and actually does not do any (potentially) blocking operations, so timeout is not even needed.

PR #26753, opened 2023-07-11
make: rework kind-install-cilium-clustermesh for Cilium CLI Helm mode

giorio94:
Cilium CLI v0.15 leverages the Helm-based installation mode by default, and it is the one which supports Cilium v1.14 and above. Let's adapt the `kind-install-cilium-clustermesh` Makefile target to work properly with it.

While being there, let's also add the possibility to enable kvstoremesh if the dedicated variable is set.

Marking for backport to v1.14 because Cilium CLI v0.15 supports Cilium 1.14 and newer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -497,32 +497,34 @@ kind-clustermesh-images: kind-clustermesh-ready kind-build-clustermesh-apiserver
 	$(QUIET)kind load docker-image $(LOCAL_KVSTOREMESH_IMAGE) --name clustermesh1
 	$(QUIET)kind load docker-image $(LOCAL_KVSTOREMESH_IMAGE) --name clustermesh2
 
+ENABLE_KVSTOREMESH ?= false
 $(eval $(call KIND_ENV,kind-install-cilium-clustermesh))
 kind-install-cilium-clustermesh: kind-clustermesh-ready ## Install a local Cilium version into the clustermesh clusters and enable clustermesh.
 	@echo "  INSTALL cilium on clustermesh1 cluster"
-	kubectl config use kind-clustermesh1
-	-$(CILIUM_CLI) uninstall >/dev/null
-	$(CILIUM_CLI) install \
+	-$(CILIUM_CLI) --context=kind-clustermesh1 uninstall >/dev/null
+	$(CILIUM_CLI) --context=kind-clustermesh1 install \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -497,32 +497,34 @@ kind-clustermesh-images: kind-clustermesh-ready kind-build-clustermesh-apiserver
...
+		--set=clustermesh.apiserver.kvstoremesh.enabled=$(ENABLE_KVSTOREMESH)
+
 	@echo "  INSTALL cilium on clustermesh2 cluster"
-	kubectl config use kind-clustermesh2
-	-$(CILIUM_CLI) uninstall >/dev/null
-	$(CILIUM_CLI) install \
-		--inherit-ca kind-clustermesh1 \
+	-$(CILIUM_CLI) --context=kind-clustermesh2 uninstall >/dev/null
+	$(KUBECTL) --context=kind-clustermesh1 get secret -n kube-system cilium-ca -o yaml | \
+        $(KUBECTL) --context=kind-clustermesh2 replace --force -f -
```
qmonnet: Nit: Let's indent this line to show that it's related to the one above, since we're piping?

Makefile
```diff
@@ -497,32 +497,34 @@ kind-clustermesh-images: kind-clustermesh-ready kind-build-clustermesh-apiserver
 	$(QUIET)kind load docker-image $(LOCAL_KVSTOREMESH_IMAGE) --name clustermesh1
 	$(QUIET)kind load docker-image $(LOCAL_KVSTOREMESH_IMAGE) --name clustermesh2
 
+ENABLE_KVSTOREMESH=false
```
qmonnet: Should you leave the possibility to change it from the environment?

```suggestion
ENABLE_KVSTOREMESH ?= false
```

Makefile
giorio94: Fixed, thanks! It was actually indented, but spaces and tabs got mixed :-(

qmonnet: Ah, maybe I just missed it because of the GH interface, I should maybe have checked locally. OK that will make it cleaner anyway :+1:

qmonnet (approved):
Looks good, thanks!

Makefile
giorio94: Yeah, it's definitely better now that it also displays correctly on GH.

PR #26759, opened 2023-07-11
docs: Ignore Helm values, update spelling list

qmonnet:
It's a pain for developers to have to update the spelling list each time they add a new Helm value. Let's ignore these when running the spell-checker. Then we can also update (trim down) the list of exceptions.

- docs: Ignore Helm value names for spellcheck
- docs: Trim down spelling list


ghost (approved):
Nice, looks good to me.

gandro:
I'm struggling to backport this to v1.13. It seems our tooling on that branch does not understand this new directive. Running `make -C Documentation` on the first commit yields the following error: `helm-values.rst:8: ERROR: Unknown interpreted text role "spelling:ignore"`

qmonnet:
> Running `make -C Documentation` on the first commit yields the following error: `helm-values.rst:8: ERROR: Unknown interpreted text role "spelling:ignore"`

This should be addressed by the change in this PR on Documentation/check-build.sh. Unless the rstcheck version in 1.13 is too old? I'll look into it and backport.

qmonnet:
Turns out this was more complex than I thought; but backports to 1.12, 1.13, 1.14 are now all done.

PR #16293, opened 2021-05-25
docs: Fix WireGuard spelling

gandro:
According to WireGuard's Trademark Usage Policy [1], WireGuard must be
written with a capital W and a capital G.

[1] https://www.wireguard.com/trademark-policy/

PR #26764, opened 2023-07-11
Fix spelling for "WireGuard"

qmonnet:
Late follow-up to https://github.com/cilium/cilium/pull/16293.

We had a custom filter for the spell checker in the docs, to ensure we'd never write "Wireguard" again. It's been broken since we added "wireguard" (considered case-insensitive) to the list of exceptions for the spell checker.

Let's fix this, and let's clean up the occurrences of "Wireguard" or "wireguard" where relevant in docs, comments, output strings that are not API.


squeed (approved):

gentoo-root (approved):

ghost (approved):

gandro (approved):
Thanks!

nathanjsweet (approved):

PR #26775, opened 2023-07-12
ci: fix checking `github.event.pull_request.head.sha`

mhofstetter:
Currently, checks like `if [ ${{ github.event.pull_request.head.sha }} != "" ]; then` are resulting in errors (`[: !=: unary operator expected`) if the variable is empty. This is the case if the workflow is triggered via push on main (and not via pull_request).

This commit fixes this by wrapping it with a string.


viktor-kurchenko (approved):
LGTM.

PR #17623, opened 2021-10-17
Speed up build image process for PRs

aanm:
Most of the PRs use the same code base. Thus, we can store Golang's
build cache on GitHub and re-use it for PR builds to speed up the
building process of the PRs code.

Since Docker buildx share the mount of type 'cache' across image builds,
the GitHub workflow to build images will populate its cache using a
'dummy' Dockerfile located in 'images/cache/Dockerfile'. This cache will
always be updated every time a build is executed in the master branch
and all PRs will reuse this cache to build their images.

The times that take each build process are:
```
Current build process     - master branch - ~ 11m3s
Build with cache hit      - master branch - ~ 9m26s (-14.63%)
Build with cache miss [1] - master branch - ~ 12m (+8.59%)

Current build process     - Pull Request  - ~ 11m1s
Cache hit PR [2]          - Pull Request  - ~ 4m17 (-61.11%)
```
[1] A cache miss will only happen if the cache is cleaned up / deleted
will only be deliberate.

[2] This was tested on a source code where there was a new line added
in the cilium-agent source code. Adding a new line meant that most of
the build cache was reused. The best case scenario are PRs that don't
change a single line of the Golang's source code, where the build
process will take less than 4m17s. The worst case scenario will be a PR
that change all of the source code for which the build time will be no
worse than 11m1s.

The reason why it takes more time on a cache hit on the master branch
than on a PR cache hit is due to the fact that the master branch will
spend the time storing the new cache where the PR will not store any
cache

Fixes https://github.com/cilium/cilium/issues/14928

PR #26786, opened 2023-07-12
ci: skip cosign / sbom in case of building images during cache rebuild

mhofstetter:
Docker cache re-build is triggered via event `workflow_run`. In this case, the container image isn't pushed to a docker registry, therefore `outputs.digest` isn't set. This results in cosign and sbom steps to fail because they depend on this.

This commit fixes this by only executing these steps in case of a github event of type `push`.

Cosign & sbom support has been introduced after the [cache](https://github.com/cilium/cilium/pull/17623). It looks like the cache-refresh use-case hasn't been considered.

Failing runs: https://github.com/cilium/cilium/actions/workflows/build-images-ci.yaml?query=event%3Aworkflow_run

### DIFF
--- a/.github/workflows/build-images-ci.yaml
+++ b/.github/workflows/build-images-ci.yaml
@@ -208,14 +208,22 @@ jobs:
             OPERATOR_VARIANT=${{ matrix.name }}
 
       - name: Sign Container Images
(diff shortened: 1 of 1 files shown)


nbusseneau (approved):
Nice catch.

Issue #26617, opened 2023-07-04
CI: azure/ipam: TestIpamManyNodes fails


lmb:
## CI failure

```
--- FAIL: Test/IPAMSuite/TestIpamManyNodes (0.08s)

...

level=info msg="Resolving IP deficit of node" available=0 availableForAllocation=256 emptyInterfaceSlots=0 instanceID=vm88 maxIPsToAllocate=10 name=node88 neededIPs=10 remainingInterfaces=1 selectedInterface=/subscriptions/xxx/resourceGroups/g1/providers/Microsoft.Compute/virtualMachineScaleSets/vmss11/virtualMachines/vm88/networkInterfaces/vmss11 selectedPoolID=s-1 subsys=ipam used=0
level=info msg="Synchronized Azure IPAM information" numInstances=100 numSubnets=3 numVirtualNetworks=1 subsys=azure
level=info msg="Synchronized Azure IPAM information" numInstances=100 numSubnets=3 numVirtualNetworks=1 subsys=azure
level=info msg="Synchronized Azure IPAM information" numInstances=100 numSubnets=3 numVirtualNetworks=1 subsys=azure
level=info msg="Synchronized Azure IPAM information" numInstances=100 numSubnets=3 numVirtualNetworks=1 subsys=azure
level=info msg="Synchronized Azure IPAM information" numInstances=100 numSubnets=3 numVirtualNetworks=1 subsys=azure
level=info msg="Synchronized Azure IPAM information" numInstances=100 numSubnets=3 numVirtualNetworks=1 subsys=azure
level=info msg="Synchronized Azure IPAM information" numInstances=100 numSubnets=3 numVirtualNetworks=1 subsys=azure
level=info msg="Synchronized Azure IPAM information" numInstances=100 numSubnets=3 numVirtualNetworks=1 subsys=azure
level=info msg="Synchronized Azure IPAM information" numInstances=100 numSubnets=3 numVirtualNetworks=1 subsys=azure
level=info msg="Synchronized Azure IPAM information" numInstances=100 numSubnets=3 numVirtualNetworks=1 subsys=azure
level=info msg="Synchronized Azure IPAM information" numInstances=100 numSubnets=3 numVirtualNetworks=1 subsys=azure
level=info msg="Synchronized Azure IPAM information" numInstances=100 numSubnets=3 numVirtualNetworks=1 subsys=azure
[...]

PR #26788, opened 2023-07-12
Extend clustermesh status reporting with remote configuration and synchronization information 

### LINKED ISSUE #26532: Expose initial synchronization of ClusterMesh data in Cilium Status

Currently,  cilium status does not expose information about initial synchronization of data from remote clusters in clustermesh. 
When connecting large clusters, initial synchronization can take significant amount of time. 
Since introduction of sync canaries in https://github.com/cilium/cilium/pull/25905  and https://github.com/cilium/cilium/pull/25499 we could expose this information in cilium status. 

This will help us uderstant current status of data synchronization in Clustermesh, but also allow us to make `cilium clustermesh status --wait` wait for initial synchronization.

giorio94:
This PR enriches the remote cluster status to additionally report information concerning the configuration advertised by the remote cluster and the synchronization status. In particular, a cluster is not reported as ready only after that the initial list of entries for each resource type has been retrieved. This is intended to provide additional information while troubleshooting possible issues (as also captured by sysdumps), as well as to allow the `cilium clustermesh status` command  to better reflect the actual status. 

Example (with the `--all-clusters` flag set, otherwise only not-ready clusters are expanded):

```
ClusterMesh:            2/3 clusters ready, 0 global-services
   hedgehog: ready, 2 nodes, 4 endpoints, 3 identities, 0 services, 0 failures (last: never)
   └  etcd: 1/1 connected, leases=0, lock lease-ID=7c02894a36550948, has-quorum=true: https://hedgehog.mesh.cilium.io:32380 - 3.5.4 (Leader)
   └  remote configuration: expected=false, retrieved=false
   └  synchronization status: nodes=true, endpoints=true, identities=true, services=true
   lobster: ready, 2 nodes, 4 endpoints, 3 identities, 0 services, 0 failures (last: never)
   └  etcd: 1/1 connected, leases=0, lock lease-ID=7c02894a1ee69fa3, has-quorum=true: https://lobster.mesh.cilium.io:32379 - 3.5.4 (Leader)
   └  remote configuration: expected=true, retrieved=true cluster-id=4, kvstoremesh=false, sync-canaries=true
   └  synchronization status: nodes=true, endpoints=true, identities=true, services=true
   ocelot: not-ready, 2 nodes, 4 endpoints, 3 identities, 0 services, 0 failures (last: never)
   └  etcd: 1/1 connected, leases=0, lock lease-ID=7c02894a22a8bc93, has-quorum=true: https://ocelot.mesh.cilium.io:32379 - 3.5.4 (Leader)
   └  remote configuration: expected=false, retrieved=true cluster-id=3, kvstoremesh=false, sync-canaries=true
   └  synchronization status: nodes=true, endpoints=true, identities=false, services=true
```

Please refer to the individual commit messages for additional information.
Tentatively marking for backport to v1.14, given that it improves the investigation of possible issues around clustermesh.

Fixes: #26532
Related: https://github.com/cilium/cilium-cli/pull/1834

```release-note
Extend clustermesh status reporting with remote configuration and synchronization information 
```


tklauser (approved):
kvstore changes LGTM, thanks.

ghost (approved):
docs ok

aditighag (approved):
>ipcache: expose synchronization status of the underlying watcher

This commit, and the earlier preparatory commit look good.

pkg/client/client.go
```diff
@@ -392,6 +392,18 @@ func FormatStatusResponse(w io.Writer, sr *models.StatusResponse, sd StatusDetai
...
 					cluster.NumFailures, timeSince(time.Time(cluster.LastFailure)))
 				fmt.Fprintf(w, "   └  %s\n", cluster.Status)
+
+				fmt.Fprint(w, "   └  remote configuration: ")
+				if cluster.Config != nil {
+					fmt.Fprintf(w, "expected=%t, retrieved=%t", cluster.Config.Required, cluster.Config.Retrieved)
+					if cluster.Config.Retrieved {
+						fmt.Fprintf(w, ", cluster-id=%d, kvstoremesh=%t, sync-canaries=%t",
+							cluster.Config.ClusterID, cluster.Config.Kvstoremesh, cluster.Config.SyncCanaries)
+					}
```
YutaroHayakawa, line 402: Is it hard to generate this output from the cluster-config itself? Otherwise, everytime we expose a new config, we need to edit here manually.

pkg/clustermesh/internal/remote_cluster.go
```diff
@@ -251,6 +255,10 @@ func (rc *remoteCluster) getClusterConfig(ctx context.Context, backend kvstore.B
 		}
 	}
 
+	rc.mutex.Lock()
+	rc.config = &models.RemoteClusterConfig{Required: requireConfig}
+	rc.mutex.Unlock()
+
```
YutaroHayakawa, line 261: I think these changes to the `restartRemoteCluster` and `getClusterConfig` should be explained in the commit message.

YutaroHayakawa (approved):
Some small suggestions. Overall, looks great.

pkg/client/client.go
giorio94: 🤔 I see your point, and I'm not an expert of the API definition part, but AFAIU this is not possible because we need to update the `openapi.yaml` definitions to include all the properties to be propagated. Hence, explicitly requiring to manually extend the definition and the usages in case a new entry gets added to the cluster config.

pkg/clustermesh/internal/remote_cluster.go
giorio94: I've extended the commit message with the additional information.

giorio94:
Last force-push rebased onto main to make tests happy.

giorio94:
Travis hit https://github.com/cilium/cilium/issues/26617. Rerunning

pkg/client/client.go
joestringer: I think that maybe what Yutaro is suggesting is that the details of how to serialize the cluster-config option should be defined in the cluster-config object, then this call should just call out to the stringify method on that object to get the canonical string representation?

giorio94: The tricky part (if I'm not missing anything) is that the cluster-config struct here is the one generated from the openapi definition, which is different from the one retrieved from etcd. Requiring to manually propagate the fields from one to another, and print them here. We could move the serialization function to somewhere else, but it would probably not help much (and it could not be a method of that object as autogenerated).

giorio94:
Removing the blocked label, given that reviews are almost all in, and https://github.com/cilium/cilium-cli/pull/1834 also got an initial feedback.

joestringer (approved):
Approving on behalf of api. Getting this in will help debuggability, so it's a nice-to-have for v1.14.0. Merging.

PR #26789, opened 2023-07-12
Add guidance for bumping the Golang version in Cilium

ferozsalam:


Documentation/contributing/development/dev_setup.rst
```diff
@@ -539,6 +539,54 @@ Making Changes
...
+
+Minor version
+^^^^^^^^^^^^^
+
+Each Cilium release is tied to a specific version of Golang via an explicit constraint
+in our Renovate configuration.
+
+We aim to build and release all maintained Cilium branches using a Golang version
+that is actively supported. This needs to be balanced against the desire to avoid
+regressions in Golang that may imapct Cilium. A new minor version of Golang should
```
michi-covalent: golang supports 2 minor versions at any given point of time. should we just always pick the older stable version?

ferozsalam: I think that could lead to unnecessary overhead – if I understand correctly that could have meant releasing Cilium 1.14 on Golang 1.19, which will be the older supported stable version at the time of release (I know that won't happen because we're already on Golang 1.20 for Cilium 1.14, but in the general case).

Assuming that we did stick to the oldest stable version, that means that we would have had to bump its Golang version three times in its lifetime (once to 1.20 in Aug '23, then again to 1.21 in Feb '24, finally to 1.22 in Aug '24), based on the Golang release cycle. By already being on 1.20, we skip an iteration of updates. I think being on the latest version that has been out for 3 months (as Kubernetes does) would be helpful from that perspective.

michi-covalent: not sure if i follow. a new go minor version gets released every 6 months. regardless of whether we pick up a new version after 3 months or picking up the older minor release, the overhead is the same, no? i don't think it's necessary to optimize for a specific cilium release.

having said that, i'm ok with picking up a new release after 3 months. renovate configuration would probably look a bit more complicated but i don't think it's a big deal.

ferozsalam: I think we're potentially talking about slightly different things but agreed it doesn't significantly matter 👍 happy to change the wording.

michi-covalent (approved):

joestringer (approved):

aanm (approved):

PR #26790, opened 2023-07-12
contrib: add check for new files in check-(api|k8s)-code-gen scripts

giorio94:
Currently, the `check-(api|k8s)-code-gen` scripts verify that generated files are up to date, running a full regeneration and checking the resulting output of `git diff`. Yet, `git diff` does not include unstaged files, meaning that brand new files would be missed. Let's fix this running `git add --intent-to-add .` first to make them appear in the diff.


aspsk (approved):
Thanks, TIL

aditighag (approved):
TIL!

PR #25440, opened 2023-05-14
bpf: Don't encrypt on path `hostns -> remote pod`

pchaigno:
In pod-to-pod encryption with IPsec and tunneling, Cilium currently encrypts traffic on the path `hostns -> remote pod` even though traffic is in plain-text on the path `remote pod -> hostns`. When using native routing, neither of those paths is encrypted because traffic from the hostns doesn't go through the bpf_host BPF program.

Cilium's Transparent Encryption with IPsec aims at encrypting pod-to-pod traffic. It is therefore unclear why we are encrypting traffic from the hostns. The simple fact that only one direction of the connection is encrypted begs the question of its usefulness. It's possible that this traffic was encrypted by mistake: some of this logic is necessary for node-to-node encryption with IPsec (not supported anymore) and pod-to-pod encryption may have been somewhat simplified to encrypt `*-to-pod` traffic.

Encrypting traffic from the hostns nevertheless creates several issues. First, this situation creates a path asymmetry between the forward and reply paths of `hostns<>remote pod` connections. Path asymmetry issues are well known to be a source of bugs, from of `--ctstate INVALID -j DROP` iptables rules to NAT issues.

Second, Gray recently uncovered a separate bug which, when combined with this encryption from hostns, can prevent Cilium from starting. That separate bug is still being investigated but it seems to cause the reload of bpf_host to depend on Cilium connecting to etcd in a clustermesh context. If this etcd is a remote pod, Cilium connects to it on hostns -> remote pod path. The bpf_host program being unloaded[1], it fails. We end up in a cyclic dependency: bpf_host requires connectivity to etcd, connectivity to etcd requires bpf_host.

This pull request therefore removes encryption with IPsec for the path `hostns -> remote pod` when using tunneling (already unencrypted when using native routing).

1 - More specifically, in Gray's case, the bpf_host program is already loaded, but it needs to be reloaded because the IPsec XFRM  config changed. Without this reload, encryption fails.

```release-note
Fix path asymmetry when using pod-to-pod encryption with IPsec and tunnel mode.
```

PR #26792, opened 2023-07-12
v1.13 backports 2023-07-12

pchaigno:
- [x] #25440 -- bpf: Don't encrypt on path hostns -> remote pod (@pchaigno)
     - Many small conflicts all around. Usually easy to resolve because we are removing code anyway.
 - [x] #26708 -- Fix upgrade for IPsec with tunneling (@pchaigno)
     - Minor conflicts for code that was removed anyway or code that was moved in the same file.

```upstream-prs
$ for pr in 25440 26708; do contrib/backporting/set-labels.py $pr done 1.13; done
```

### DIFF
--- a/bpf/bpf_host.c
+++ b/bpf/bpf_host.c
@@ -22,9 +22,6 @@
 /* Pass unknown ICMPv6 NS to stack */
 #define ACTION_UNKNOWN_ICMP6_NS CTX_ACT_OK
 
-/* CB_PROXY_MAGIC overlaps with CB_ENCRYPT_MAGIC */
-#define ENCRYPT_OR_PROXY_MAGIC 0
-
 /* Controls the inclusion of the CILIUM_CALL_SEND_ICMP6_ECHO_REPLY section in
  * the bpf_lxc object file.
  */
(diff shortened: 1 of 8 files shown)


pchaigno:
/test-backport-1.13

pchaigno:
I manually tested:
- [ ] Scale up and down
- [x] Upgrade from v1.12.7
- [ ] Upgrade from v1.13.0
- [ ] Key rotation

bpf/bpf_host.c
```diff
@@ -1266,10 +1241,7 @@ int cil_to_host(struct __ctx_buff *ctx)
 	bool traced = false;
 	__u32 src_id = 0;
 
-	if ((magic & MARK_MAGIC_HOST_MASK) == MARK_MAGIC_ENCRYPT) {
-		ctx->mark = magic; /* CB_ENCRYPT_MAGIC */
-		src_id = ctx_load_meta(ctx, CB_ENCRYPT_IDENTITY);
```
pchaigno, line 1271: For reviewers: On a previous backport of this change, Julian commented:
> Just to note that in v1.13 the wireguard implementation is different, and `from-container` uses `set_encrypt_mark()`. So I hope our reasoning why we were able to remove this section in `main` also applies to v1.13...

Martynas confirmed that this code isn't used in the case of WireGuard on v1.13 and older. Instead, traffic marked for WireGuard encryption is sent to cilium_wg0 and then traverses `to-host` without a mark.

pchaigno:
/test-1.20-4.19

jschwinger233 (approved):
:+1:

bpf/bpf_host.c
julianwiedmann: I must have not been paying sufficient attention during that review. In the `from-container` path, Wireguard uses `set_encrypt_mark()` - so the `MARK_MAGIC_ENCRYPT` ends up in `skb->mark`. But here we obtained the `magic` value from the `skb->cb` (note the `ctx_load_meta()`).

So unless there's some magic at play that transfers the `skb->mark` to the `skb->cb`, there's no way this could have even conflicted. Sorry for burning your cycles on this :(.

julianwiedmann (approved):
lgtm, thank you!

bpf/bpf_host.c
pchaigno: No worries at all. Better have unnecessary checks than missing something.

PR #26794, opened 2023-07-12
CiliumNetworkPolicy validation

### LINKED ISSUE #23152: CFP: CiliumNetworkPolicy validation

romanspb80:
This patch adds validation for CiliumNetworkPolicy with CEL

Fixes: #23152

### DIFF
--- a/install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
+++ b/install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
@@ -1,0 +1,59 @@
+{{- if and .Values.validatingAdmissionPolicy.enabled (.Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1alpha1/ValidatingAdmissionPolicy") }}
+apiVersion: admissionregistration.k8s.io/v1alpha1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: "validate-cilium-network-policies"
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups:   ["cilium.io"]
+        apiVersions: ["v2"]
+        operations:  ["CREATE", "UPDATE"]
+        resources:   ["ciliumnetworkpolicies", "ciliumclusterwidenetworkpolicies"]
+  validations:
+    - expression:
+        (
+          has(object.spec.ingress)
+          && object.spec.ingress.exists(x, has(x.toPorts)
+            && x.toPorts.exists(x,
+              (
+                has(x.ports) && (
+                  x.ports.exists(x, x.protocol == 'TCP')
+                  && (
+                       has(x.rules) && (
+                         has(x.rules.http) || has(x.rules.kafka)
+                       )
+                  )
+                ) && (
+                       has(x.rules) && !has(x.rules.dns)
+                )
+              )
+            )
+          )
+        ) ||
+        (
+          has(object.spec.egress)
+          && object.spec.egress.exists(x, has(x.toPorts)
+            && x.toPorts.exists(x,
+              (
+                has(x.ports) && (
+                  x.ports.exists(x, x.protocol == 'TCP')
+                  && (
+                       has(x.rules) && (
+                         has(x.rules.http) || has(x.rules.kafka)
+                       )
+                  )
+                ) || (
+                       has(x.rules) && has(x.rules.dns)
+                )
+              )
+            )
+          )
+        ) ||
+        (
+          !has(object.spec.ingress) && !has(object.spec.egress)
+        )
+
+      message: "L7 rules can only apply to TCP (not UDP) except for egress DNS rules"
+{{- end }}
--- a/install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicybinding.yaml
+++ b/install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicybinding.yaml
@@ -1,0 +1,10 @@
+{{- if and .Values.validatingAdmissionPolicy.enabled (.Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1alpha1/ValidatingAdmissionPolicyBinding") }}
+apiVersion: admissionregistration.k8s.io/v1alpha1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: "validate-cilium-network-policies-binding"
+spec:
+  policyName: "validate-cilium-network-policies"
+  validationActions: [Deny]
+  matchResources:
+{{- end }}
--- a/install/kubernetes/cilium/values.yaml
+++ b/install/kubernetes/cilium/values.yaml
@@ -3214,3 +3214,7 @@ k8sClientRateLimit:
   # -- The burst request rate in requests per second.
   # The rate limiter will allow short bursts with a higher rate.
   burst: 10
+
+validatingAdmissionPolicy:
+  # -- Enable support for Validating Admission Policy
+  enabled: true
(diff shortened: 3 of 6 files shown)


install/kubernetes/cilium/values.yaml
kaworu, line 1: I don't think removing `k8sClientRateLimit` was intended, also please report your changes to `install/kubernetes/cilium/values.yaml.tmpl` and then run:

```
make -C install/kubernetes cilium/values.yaml
```

to update this file.

kaworu (changes requested):
Thanks for the PR @romanspb80, please see my comment on `value.yaml`. Also pulling @cilium/sig-policy for the review.

install/kubernetes/cilium/values.yaml
joamaki: `k8sClientRateLimit` was just introduced, so rebase and then regenerate `values.yaml`.

joamaki:
Adding @christarazi and @giorio94 as extra reviewers due to https://github.com/cilium/cilium/issues/23152

giorio94:
Adding @gandro as extra reviewer, who has already reviewed an [earlier iteration of this PR](https://github.com/cilium/cilium/pull/26470). 
@romanspb80 Please do not open separate PRs for the same patch as it becomes difficult to follow the comments history.

gandro (approved):
In terms of Helm structure, this looks good to me.

I'll defer the review of the actual validation logic to other subteams though.

gandro:
@romanspb80 Could you rebase on main again? CI is failing due to #26883 not being part of you branch. Thanks.

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
giorio94, line 1: I'd suggest to move these two files to a dedicated directory, renaming the files to reflect the name of the policy.

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicybinding.yaml
giorio94, line 1: When trying to apply this policy, an error is returned:

```
The ValidatingAdmissionPolicyBinding "cilium-validate-network-policies-binding" is invalid: spec.validationActions: Required value: at least one validation action is required
```

giorio94 (changes requested):
I made a try applying this validation policy, but it appears to reject even empty policies:

```
cat <<EOF | kubectl apply -f - 
apiVersion: cilium.io/v2
kind: CiliumNetworkPolicy
metadata:
  name: test
spec:
  endpointSelector: {}
EOF
```

Returns: 

```
The ciliumnetworkpolicies "test" is invalid: : ValidatingAdmissionPolicy 'cilium-validate-network-policies' with binding 'cilium-validate-network-policies-binding' denied request: L7 rules can only apply to TCP (not UDP) except for egress DNS rules
```

My understanding is that an operation is rejected if the expression evaluates to false, and not when it evaluates to true.

I wonder how we could have tests to ensure that this (and other policies which we may introduce in the future) actually work as expected, and to prevent possible future regressions. 

I'd be also personally hesitant in saying that this PR fixes https://github.com/cilium/cilium/issues/23152. While being a step forward for sure, I feel that there may be other combinations of settings which would benefit from a validation, as invalid.

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,56 @@
...
+        (
+          has(object.spec.egress)
+          && object.spec.egress.exists(x, has(x.toPorts)
+            && x.toPorts.exists(x,
+              (
+                has(x.ports) && (
+                  x.ports.exists(x, x.protocol == 'TCP')
+                  && (
+                       has(x.rules) && (
+                         has(x.rules.http) || has(x.rules.kafka)
```
aditighag, line 44: Can you add a reference link to the CNP spec here so that readers know where these rules come from?

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,56 @@
+{{- if and .Values.validatingAdmissionPolicy.enabled (.Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1alpha1/ValidatingAdmissionPolicy") }}
+apiVersion: admissionregistration.k8s.io/v1alpha1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: "cilium-validate-network-policies"
```
aditighag:
```suggestion
  name: "validate-cilium-network-policies"
```

aditighag (changes requested):
Very cool! 

>This patch adds validation for CiliumNetworkPolicy with CEL

Please spell out CEL in the description, and add an official k8s documentation link. It'll be great add a reference link to the CNP spec where the rules are specified.

Can you also add an example in the PR description showing how the validation fails an incorrect CNP?

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,59 @@
...
+    resourceRules:
+      - apiGroups:   ["cilium.io"]
+        apiVersions: ["v2"]
+        operations:  ["CREATE", "UPDATE"]
+        resources:   ["ciliumnetworkpolicies", "ciliumclusterwidenetworkpolicies"]
+  validations:
+    - expression:
+        (
+          has(object.spec.ingress)
+          && object.spec.ingress.exists(x, has(x.toPorts)
```
nathanjsweet, line 18: Members of `ingress` do not need `toPorts`. My proposed change checks if `toPorts` does not exist and then proceeds to validate `toPorts` if it does.
```suggestion
          && object.spec.ingress.exists(x, !has(x.toPorts)
```

install/kubernetes/cilium/templates/cilium-agent/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,59 @@
...
+            && x.toPorts.exists(x,
```
nathanjsweet, line 19: With the above changes this converts to an "or".
```suggestion
            x.toPorts.exists(x,
```

nathanjsweet (changes requested):
Your expression is not structured to cover both `egress` and `ingress` if they are both present (and if they are both present and one of them is invalid, then this validation logic will miss the invalid object), the `toPorts` validation incorrectly assumes that `toPorts` must be present but there are other membership possibilities to consider, and it doesn't validate `endpointSelectors`.

The structure of the expression should look something like this:
```cel
has(object.spec.endpointSelectors) &&
// validate endpointSelectors
(... ) &&
 // validate ingress object
(!has(object.spec.ingress) ||
    ...
) &&
// validate egress object
(!has(object.spec.egress) ||
    ...
)
```

I'm converting this PR to a draft as it needs more work.

Feel free to reach out on slack for more detailed feedback.

PR #26800, opened 2023-07-13
client, health/client: set dummy host header on unix:// local communication

tklauser:
Go 1.20.6 added a security fix [1] which leads to stricter sanitization of the HTTP host header in the net/http client. Cilium's pkg/client currently sets the Host header to the UDS path (e.g. /var/run/cilium/cilium.sock), however the slashes in that Host header now lead net/http to reject it.

RFC 7230, Section 5.4 states [2]:

> If the authority component is missing or undefined for the target URI,
> then a client MUST send a Host header field with an empty field-value.

The authority component is undefined for the unix:// scheme. Thus, the correct value to use would be the empty string. However, this does not work due to OpenAPI runtime using the same value for the URL's host and the http client's host header. Thus, use a dummy value "localhost".

[1] https://go.dev/issue/60374
[2] https://datatracker.ietf.org/doc/html/rfc7230#section-5.4

(The same security fix was backported to Go 1.19.11, so requesting backport to v1.13 as well given that the branch uses Go 1.19)

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -120,6 +120,8 @@ func NewRuntime(host string) (*runtime_client.Runtime, error) {
 		return nil, fmt.Errorf("invalid host format '%s'", host)
 	}
 
+	hostHeader := tmp[1]
+
 	switch tmp[0] {
 	case "tcp":
 		if _, err := url.Parse("tcp://" + tmp[1]); err != nil {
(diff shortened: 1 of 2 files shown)


tklauser:
From https://github.com/cilium/cilium/pull/26713 it looks like we'll use Go 1.19 or Go 1.20 on other stable branches as well, so marking this PR to be backported to all stable branches.

tklauser:
Conformance AWS-CNI failed with https://github.com/cilium/cilium-cli/issues/1794: https://github.com/cilium/cilium/actions/runs/5541124605 re-triggering

aanm (approved):

PR #26802, opened 2023-07-13
bpf: tests: improve CT checks for observed TCP flags

julianwiedmann:
The code in conntrack_test.c is very confused. It takes a __be32 value (TCP_FLAG_SYN), and then clamps it to __be16 before byte-swapping to host-order (it should use bpf_ntohl() instead). The u16 then gets compared to .rx_flags_seen, which is actually an __u8.

Amazingly all this magic works out well, because the flag bits are *just* in the right position of the TCP header's flag word. At least on little-endian.

Clean it up by introducing a little helper that accesses the right part of the flag word. Use it to add some checks in the other CT tests.

### DIFF
--- a/bpf/lib/l4.h
+++ b/bpf/lib/l4.h
@@ -24,6 +24,11 @@ union tcp_flags {
 	__u32 value;
 };
 
+static __always_inline __u8 tcp_flags_to_u8(__be32 value)
+{
+	return ((union tcp_flags)value).lower_bits;
+}
+
 /**
  * Modify L4 port and correct checksum
  * @arg ctx:      packet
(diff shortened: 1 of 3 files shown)


aspsk (approved):
Looks good. (I would split the change into two commits though--splitting the `Use it to add some checks in the other CT tests` into a second commit.)

julianwiedmann:
> Looks good. (I would split the change into two commits though--splitting the `Use it to add some checks in the other CT tests` into a second commit.)

Hmm yes, that would have been prettier. Most likely an artifact of how this whole journey started (I was chasing down an issue in `test_ct4_rst1_check()` initially).

But now CI is green, so proceed we must.

PR #26803, opened 2023-07-13
bgpv1: Add GetRoutes method to Router interface and generic Path type

rastislavs:
Adds `GetRoutes` method to the Router interface and generic `Path` and `Route` types, that can be used to retrieve existing routes from the RIB of underlying router.

This is the first step towards exposing the BGP routes via cilium API + CLI.

### DIFF
--- a/pkg/bgpv1/gobgp/conversions_test.go
+++ b/pkg/bgpv1/gobgp/conversions_test.go
@@ -1,0 +1,57 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package gobgp
+
+import (
+	"context"
+	"testing"
+
+	gobgp "github.com/osrg/gobgp/v3/api"
+	"github.com/osrg/gobgp/v3/pkg/server"
+	"github.com/stretchr/testify/require"
+)
+
+// Test common Path type conversions appearing in the codebase
+func TestPathConversions(t *testing.T) {
+	for _, tt := range commonPaths {
+		t.Run(tt.name, func(t *testing.T) {
+			s := server.NewBgpServer()
+			go s.Serve()
+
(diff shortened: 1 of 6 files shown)


pkg/bgpv1/gobgp/conversions_test.go
```diff
@@ -0,0 +1,44 @@
...
+			res, err := s.AddPath(context.TODO(), &gobgp.AddPathRequest{
+				Path: path,
+			})
+			require.NoError(t, err)
+			require.NotZero(t, res.Uuid)
+		})
+	}
+}
+
+// TODO: Test ToAgentPath
```
YutaroHayakawa: Could you address this TODO?

YutaroHayakawa (changes requested):
Thanks for taking over this! Overall looks good except some minor things. Could you consider consolidating last two commits into first two? So that we will have a clean commit history.

pkg/bgpv1/gobgp/conversions_test.go
rastislavs: I believe it is addressed, I extended the existing test to perform the conversion in both directions and compare the result with the original Path - see: https://github.com/cilium/cilium/blob/7fb8fb5dbc342014b77dfec5e5549aaca10a5472/pkg/bgpv1/gobgp/conversions_test.go#L48-L52

rastislavs:
@YutaroHayakawa I consolidated the commits

pkg/bgpv1/gobgp/conversions_test.go
YutaroHayakawa: Oh, sorry. I have missed that.

YutaroHayakawa (approved):
Now LGTM 👍 Thanks!

PR #24811, opened 2023-04-11
ipsec, option: Make the IPsec key rotation delay configurable

pchaigno:
We currently have a maximum IPsec key rotation delay hardcoded to 5 minutes. After that time, we remove the old key from the node. Therefore, if the new key wasn't propagated to all nodes before then, communication between nodes will break.

5 minutes is usually enough because all agents will pick up the new key as soon as it's loaded and takes them a few seconds maximum to reload their IPsec configuration in the kernel.

For clustermesh scenarios however, 5 minutes may not be enough. In those cases, the key needs to be updated separately on all clusters. Depending on the number of clusters and how this is done, it may take longer than 5 minutes.

This pull request therefore makes this 5 minutes delay configurable via an agent flag.

PR #26809, opened 2023-07-13
docs/ipsec: Extend troubleshooting for long key rotations

pchaigno:
This is documentation for the flag introduced in https://github.com/cilium/cilium/pull/24811 and the problem it addresses.

### DIFF
--- a/Documentation/security/network/encryption-ipsec.rst
+++ b/Documentation/security/network/encryption-ipsec.rst
@@ -211,14 +211,24 @@ Troubleshooting
 
  * All XFRM errors correspond to a packet drop in the kernel. Except for
    ``XfrmFwdHdrError`` and ``XfrmInError``, all XFRM errors indicate a bug in
-   Cilium or an operational mistake. ``XfrmOutStateSeqError`` and
-   ``XfrmInStateProtoError`` may be caused by operational mistakes, as detailed
-   in the following points.
+   Cilium or an operational mistake. ``XfrmOutStateSeqError``,
+   ``XfrmInStateProtoError``, and ``XfrmInNoStates`` may be caused by
+   operational mistakes, as detailed in the following points.
 
  * If the sequence number reaches its maximum value for any XFRM OUT state, it
    will result in packet drops and XFRM errors of type
    ``XfrmOutStateSeqError``. A key rotation resets all sequence numbers.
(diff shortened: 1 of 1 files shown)


gentoo-root (approved):

Documentation/security/network/encryption-ipsec.rst
```diff
@@ -211,15 +211,25 @@ Troubleshooting
...
 
+ * After a key rotation, if the old key is cleaned up before the
+   configuration of the new key is installed on all nodes, it will result in
+   ``XfrmInNoStates`` errors. The old key is removed from nodes after 5 minutes
+   by default. By default, all agents watch for key updates and update their
+   configuration within 1 minute after the key is changed, leaving plenty of
+   time before the old key is removed. If you expect the key rotation to take
+   longer for some reason (e.g., in case of clustermesh several clusters need
+   to be updated), you can increase the delay before cleanup with agent flag
+   ``ipsec-key-rotation-duration``.
```
zacharysarah: Edits for clarity, future tense, no Latin abbreviations. Normally I would review for passive voice, but I think active voice alternatives wouldn't be nearly as clear.
```suggestion
   configuration of the new key is installed on all nodes, it results in
   ``XfrmInNoStates`` errors. The old key is removed from nodes after a default interval
   of 5 minutes. By default, all agents watch for key updates and update their
   configuration within 1 minute after the key is changed, leaving plenty of
   time before the old key is removed. If you expect the key rotation to take
   longer for some reason (for example, in the case of Cluster Mesh if several clusters need
   to be updated), you can increase the delay before cleanup with agent flag
   ``ipsec-key-rotation-duration``.
```

zacharysarah (changes requested):
@pchaigno Some mild edits for the usual suspects

Documentation/security/network/encryption-ipsec.rst
pchaigno: > future tense

I really need to be more careful on that. I know it and you've pointed it before :disappointed: 

> no Latin abbreviations

FYI, regarding that and "clustermesh" vs. "Cluster Mesh", the docs are not consistent (even this page isn't).

zacharysarah (approved):
@pchaigno Thanks for the updates!

PR #26812, opened 2023-07-13
certloader: Correctly support RequestClientCert in WatchedClientConfig

chancez:
Only configure GetClientCertificate if client keypair is configured, allowing servers to Request ClientCertificates without requiring them.

In docs for `GetClientCertificate` it specifies:

>  GetClientCertificate must return a non-nil Certificate. If
>  Certificate.Certificate is empty then no certificate will be sent to the
>  server.

If a nil certificate is sent when the server requests a client certificate, the client will return an error. Instead, only configure GetClientCertificate if certificates are provided and the server may choose to how to handle the lack of a client certificate.

This is needed primarily for when the server is using RequestClientCert, which requests a certificate, but does not require the client to send one.

Previously, you would see this log message produced by the clients using `WatchedClientConfig.GetClientCertificate`:

```
transport: authentication handshake failed: mTLS client certificate requested, but not configured
```

Now, if a server requires a client certificate it should reject the TLS connection and the client will receive the error from the server.

I also discovered the same issue in Hubble CLI here: https://github.com/cilium/hubble/pull/1123, though it doesn't use this package.

```release-note
certloader: Correctly support RequestClientCert in WatchedClientConfig
```

### DIFF
--- a/pkg/crypto/certloader/client.go
+++ b/pkg/crypto/certloader/client.go
@@ -5,7 +5,6 @@ package certloader
 
 import (
 	"crypto/tls"
-	"fmt"
 
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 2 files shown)


chancez:
Went with release-note/misc since it seems the only user of the modified method is in our tests.

rolinh (approved):

rolinh:
This is technically a bug, marking for backport to v1.14 and v1.13 as per our backport policy.

kaworu (changes requested):
Thanks for the patch, LGTM but I'd like to see a simple unit test added for this change please.

chancez:
@kaworu good point. I wrote a simple unit test verifying GetClientCertificate is nil when no keypair is provided. I considered doing an httpserver test, but wasn't sure that much was needed since this behavior is well documented and should be tested in the stdlib.

kaworu (approved):
Thanks!

aditighag (commented):
I don't have any context for the package, so please go ahead with getting this merged based on reviews from the Hubble folks.

PR #26826, opened 2023-07-14
bpf: ct: clean up tuple swapping for forward lookups

julianwiedmann:
While working on https://github.com/cilium/cilium/pull/25826, @gentoo-root suggested that we can also stream-line some of the CT tuple swapping. For SCOPE_FORWARD lookups, it doesn't make sense to first build a tuple in "standard" (== reverse) layout, when it immediately gets swapped again.


gentoo-root (approved):
Thank you for keeping track of all follow-ups!

PR #26834, opened 2023-07-14
k8s/apis: refactor CRD registration helpers into a separate package

tklauser:
The first commit factors out the helper functions used to create and install CRD objects with k8s. The second commit changes the factored out helpers to no longer depend on a specific API version but take all values as parameters.


pippolo84 (approved):
Thanks! :rocket:

joamaki (approved):

PR #22892, opened 2022-12-29
Refactor k8s identities GC into a cell.Module

pippolo84:
Port the current implementation of the CiliumIdentity GC to a separate `cell.Module`.

This makes possible to improve the testing of the GC code through unit tests, without relying on more complicated (and slow) controlplane tests.
Besides, the modularization of the GC cleans up the operator bootstrap code, since all the GC specific parts are now moved to their cell.

The new cell contains the code for both GC modes: CRD allocated identities and kvstore allocated identities.
The GC is moved inside the `identity` package and has the following structure:

```
operator/identity/
├── cell.go
├── crd_gc.go
├── gc.go
├── heartbeat.go
└── kvstore_gc.go
```

where:

- `cell.go` contains all the exported API for the new cell, alongside its dependencies and configuration.
- `gc.go` contains the common part of both gc modes.
- `crd_gc.go` and `heartbeat.go` contain the CRD allocation mode specific GC part.
- `kvstore`contains the kvstore allocation mode specific GC part.

The new cell relies on `resource.Resource[*v2.CiliumIdentity]` to listen for CiliumIdentity related events and react accordingly.

**Note for the reviewers**

This is probably better to review commit by commit. The most impactful one is 1a482f8d3a08147410e65d91adbc6e49515a49b4, but despite the number of files modified, it aims to add the minimum amount of changes needed to modularize the identities GC code in a separate cell.
Further cleanups are left in the subsequent commits.

PR #26836, opened 2023-07-14
Full CES Reconciliation in the Operator

### LINKED ISSUE #24581: Cilium Operator doesn't fully reconcile CiliumEndpointSlices on startup

### What happened?

Cluster with Cilium Endpoint Slices enabled with identity based batching.
Label on namespace was changed. It led to change of identities for Cilium Endpoints in this namespace - Cilium Endpoints were updated with new identity.
Operator moved the endpoints from one slice to another.

First operator queued the insert of the cep to the new ces.
Then, delayed, operator queued the removed of the old cep from the old ces.
Operator crashed after inserting, before removing.

The old cep was never removed from the old ces.

In general there is no reconciliation of Cilium Endpoint Slices. If for some reason the state becomes bad it stays that way. In this particular case it should be enough to have full reconciliation during startup. 

### Cilium Version

all

### Kernel Version

all

### Kubernetes Version

all

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

alan-kut:
Implement full CES reconciliation logic in the Cilium Operator.

Refactor and unify CEPs and CESs watchers in the operator.
Refactor CES Controller:
- Limit the functionality of the Manager to maintain mapping between CEPs and CESs.
- Implement reconciliation in the Reconcilar based on the mapping from the Manager and CEPs and CESs from the watchers stores.
- React to the CESs events in addition to the CEPs events.

This PR should fix all the problems related to CESs out of sync.
If there is a broken state (stale, duplicated, missing CESs) the operator should always fully reconcile it on startup and maintain good state.
If different client breaks CES operator would fix it back.

Fixes: #24581

```release-note
Implement full CES reconciliation logic in the operator
```

### DIFF
--- a/operator/cmd/root.go
+++ b/operator/cmd/root.go
@@ -30,7 +30,7 @@ import (
 	operatorK8s "github.com/cilium/cilium/operator/k8s"
 	operatorMetrics "github.com/cilium/cilium/operator/metrics"
 	operatorOption "github.com/cilium/cilium/operator/option"
-	ces "github.com/cilium/cilium/operator/pkg/ciliumendpointslice"
+	"github.com/cilium/cilium/operator/pkg/ciliumendpointslice"
 	gatewayapi "github.com/cilium/cilium/operator/pkg/gateway-api"
 	"github.com/cilium/cilium/operator/pkg/ingress"
 	"github.com/cilium/cilium/operator/pkg/lbipam"
@@ -127,6 +127,14 @@ var (
 			}
 		}),
 
+		cell.Provide(func(
+			daemonCfg *option.DaemonConfig,
+		) ciliumendpointslice.SharedConfig {
+			return ciliumendpointslice.SharedConfig{
+				EnableCiliumEndpointSlice: daemonCfg.EnableCiliumEndpointSlice,
+			}
+		}),
+
 		api.HealthHandlerCell(
 			kvstoreEnabled,
 			isLeader.Load,
@@ -157,6 +165,13 @@ var (
 			// setup operations. This is a hacky workaround until the kvstore is
 			// refactored into a proper cell.
 			identitygc.Cell,
+
+			// CiliumEndpointSlice controller depends on the CiliumEndpoint and
+			// CiliumEndpointSlice resources. It reconciles the state of CESs in the
+			// cluster based on the CEPs and CESs events.
+			// It is disabled if CiliumEndpointSlice is disabled in the cluster -
+			// when --enable-cilium-endpoint-slice is false.
+			ciliumendpointslice.Cell,
 		),
 	)
 
@@ -420,29 +435,6 @@ func (legacy *legacyOnLeader) onStop(_ hive.HookContext) error {
 // in HA mode.
 func (legacy *legacyOnLeader) onStart(_ hive.HookContext) error {
 	isLeader.Store(true)
-
-	// If CiliumEndpointSlice feature is enabled, create CESController, start CEP watcher and run controller.
-	// Knowing CES are enabled only if CiliumEndpoint CRD are enabled too.
-	if legacy.clientset.IsEnabled() && option.Config.EnableCiliumEndpointSlice {
-		log.Info("Create and run CES controller, start CEP watcher")
-		// Initialize  the CES controller
-		cesController := ces.NewCESController(
-			legacy.ctx,
-			&legacy.wg,
-			legacy.clientset,
-			operatorOption.Config.CESMaxCEPsInCES,
-			operatorOption.Config.CESSlicingMode,
-			operatorOption.Config.CESWriteQPSLimit,
-			operatorOption.Config.CESWriteQPSBurst)
-		// Start CEP watcher
-		operatorWatchers.CiliumEndpointsSliceInit(legacy.ctx, &legacy.wg, legacy.clientset, cesController)
-		// Start the CES controller, after current CEPs are synced locally in cache.
-		legacy.wg.Add(1)
-		go func() {
-			defer legacy.wg.Done()
-			cesController.Run(operatorWatchers.CiliumEndpointStore, legacy.ctx.Done())
-		}()
-	}
 
 	// Restart kube-dns as soon as possible since it helps etcd-operator to be
 	// properly setup. If kube-dns is not managed by Cilium it can prevent
--- a/operator/identitygc/crd_gc_test.go
+++ b/operator/identitygc/crd_gc_test.go
@@ -1,0 +1,104 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package identitygc
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/cilium/cilium/operator/k8s"
+	"github.com/cilium/cilium/pkg/hive"
+	"github.com/cilium/cilium/pkg/hive/cell"
+	cilium_v2a1 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2alpha1"
+	k8sClient "github.com/cilium/cilium/pkg/k8s/client"
+	"github.com/cilium/cilium/pkg/k8s/resource"
+	"github.com/cilium/cilium/pkg/testutils"
+)
+
+func TestUsedIdentitiesInCESs(t *testing.T) {
+	var fakeClient k8sClient.FakeClientset
+	var ciliumEndpointSlice resource.Resource[*cilium_v2a1.CiliumEndpointSlice]
+	hive := hive.New(
+		k8sClient.FakeClientCell,
+		k8s.ResourcesCell,
+		cell.Invoke(func(c *k8sClient.FakeClientset, ces resource.Resource[*cilium_v2a1.CiliumEndpointSlice]) error {
+			fakeClient = *c
+			ciliumEndpointSlice = ces
+			return nil
+		}),
+	)
+	err := hive.Start(context.Background())
+	if err != nil {
+		t.Fatalf("unable to start hive for the test: %s", err)
+	}
+
+	cesStore, _ := ciliumEndpointSlice.Store(context.Background())
+
+	// Empty store.
+	gotIdentities := usedIdentitiesInCESs(cesStore)
+	wantIdentities := make(map[string]bool)
+	assertEqualIDs(t, wantIdentities, gotIdentities)
+
+	// 5 IDs in the store.
+	cesA := createCESWithIDs("cesA", []int64{1, 2, 3, 4, 5})
+	fakeClient.CiliumV2alpha1().CiliumEndpointSlices().Create(context.Background(), cesA, meta_v1.CreateOptions{})
+	err = testutils.WaitUntil(isCESPresent("cesA", cesStore), time.Second)
+	if err != nil {
+		t.Fatalf("cesA not present in the store after timeout: %s", err)
+	}
+	wantIdentities["1"] = true
+	wantIdentities["2"] = true
+	wantIdentities["3"] = true
+	wantIdentities["4"] = true
+	wantIdentities["5"] = true
+	gotIdentities = usedIdentitiesInCESs(cesStore)
+	assertEqualIDs(t, wantIdentities, gotIdentities)
+
+	// 10 IDs in the store.
+	cesB := createCESWithIDs("cesB", []int64{10, 20, 30, 40, 50})
+	fakeClient.CiliumV2alpha1().CiliumEndpointSlices().Create(context.Background(), cesB, meta_v1.CreateOptions{})
+	err = testutils.WaitUntil(isCESPresent("cesB", cesStore), time.Second)
+	if err != nil {
+		t.Fatalf("cesB not present in the store after timeout: %s", err)
+	}
+	wantIdentities["10"] = true
+	wantIdentities["20"] = true
+	wantIdentities["30"] = true
+	wantIdentities["40"] = true
+	wantIdentities["50"] = true
+	gotIdentities = usedIdentitiesInCESs(cesStore)
+	assertEqualIDs(t, wantIdentities, gotIdentities)
+
+	err = hive.Stop(context.Background())
+	if err != nil {
+		t.Fatalf("unable to stop hive for the test: %s", err)
+	}
+}
+
+func isCESPresent(cesName string, cesStore resource.Store[*cilium_v2a1.CiliumEndpointSlice]) testutils.ConditionFunc {
+	return func() bool {
+		_, exists, _ := cesStore.GetByKey(resource.Key{Name: cesName})
+		return exists
+	}
+}
+
+func createCESWithIDs(cesName string, ids []int64) *cilium_v2a1.CiliumEndpointSlice {
+	ces := &cilium_v2a1.CiliumEndpointSlice{ObjectMeta: meta_v1.ObjectMeta{Name: cesName}}
+	for _, id := range ids {
+		cep := cilium_v2a1.CoreCiliumEndpoint{IdentityID: id}
+		ces.Endpoints = append(ces.Endpoints, cep)
+	}
+	return ces
+}
+
+func assertEqualIDs(t *testing.T, wantIdentities, gotIdentities map[string]bool) {
+	t.Helper()
+	if diff := cmp.Diff(wantIdentities, gotIdentities); diff != "" {
+		t.Errorf("Unexpected Identites in the CES store (-want +got): \n%s", diff)
+	}
+}
--- a/operator/pkg/ciliumendpointslice/cell.go
+++ b/operator/pkg/ciliumendpointslice/cell.go
@@ -1,0 +1,69 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package ciliumendpointslice
+
+import (
+	"github.com/spf13/pflag"
+
+	"github.com/cilium/cilium/pkg/hive/cell"
+)
+
+const (
+	// CESMaxCEPsInCES is the maximum number of cilium endpoints allowed in single
+	// a CiliumEndpointSlice resource.
+	CESMaxCEPsInCES = "ces-max-ciliumendpoints-per-ces"
+
+	// CESSlicingMode instructs how CEPs are grouped in a CES.
+	CESSlicingMode = "ces-slice-mode"
+
+	// CESWriteQPSLimit is the rate limit per second for the CES work queue to
+	// process  CES events that result in CES write (Create, Update, Delete)
+	// requests to the kube-apiserver.
+	CESWriteQPSLimit = "ces-write-qps-limit"
+
+	// CESWriteQPSBurst is the burst rate per second used with CESWriteQPSLimit
+	// for the CES work queue to process CES events that result in CES write
+	// (Create, Update, Delete) requests to the kube-apiserver.
+	CESWriteQPSBurst = "ces-write-qps-burst"
+)
+
+// Cell is a cell that implements a Cilium Endpoint Slice Controller.
+// The controller subscribes to cilium endpoint and cilium endpoint slices
+// events and reconciles the state of the cilium endpoint slices in the cluster.
+var Cell = cell.Module(
+	"k8s-ces-controller",
+	"Cilium Endpoint Slice Controller",
+	cell.Config(defaultConfig),
+	cell.Invoke(registerController),
+)
+
+type Config struct {
+	CESMaxCEPsInCES  int     `mapstructure:"ces-max-ciliumendpoints-per-ces"`
+	CESSlicingMode   string  `mapstructure:"ces-slice-mode"`
+	CESWriteQPSLimit float64 `mapstructure:"ces-write-qps-limit"`
+	CESWriteQPSBurst int     `mapstructure:"ces-write-qps-burst"`
+}
+
+var defaultConfig = Config{
+	CESMaxCEPsInCES:  100,
+	CESSlicingMode:   "cesSliceModeIdentity",
+	CESWriteQPSLimit: 10,
+	CESWriteQPSBurst: 20,
+}
+
+func (def Config) Flags(flags *pflag.FlagSet) {
+	flags.Int(CESMaxCEPsInCES, def.CESMaxCEPsInCES, "Maximum number of CiliumEndpoints allowed in a CES")
+	flags.String(CESSlicingMode, def.CESSlicingMode, "Slicing mode define how ceps are grouped into a CES")
+	flags.Float64(CESWriteQPSLimit, def.CESWriteQPSLimit, "CES work queue rate limit")
+	flags.Int(CESWriteQPSBurst, def.CESWriteQPSBurst, "CES work queue burst rate")
+}
+
+// SharedConfig contains the configuration that is shared between
+// this module and others.
+// It is a temporary solution meant to avoid polluting this module with a direct
+// dependency on global operator and daemon configurations.
+type SharedConfig struct {
+	// EnableCiliumEndpointSlice enables the cilium endpoint slicing feature and the CES Controller.
+	EnableCiliumEndpointSlice bool
+}
--- a/operator/pkg/ciliumendpointslice/ceptocesmap.go
+++ b/operator/pkg/ciliumendpointslice/ceptocesmap.go
@@ -8,104 +8,137 @@ import "github.com/cilium/cilium/pkg/lock"
 type CEPName string
 type CESName string
 
-// CESToCEPMapping is used to map CiliumEndpointSlice name to cesTracker object
-// which in turn consists of CEPs, a CES name to list of all CEPs.
-// Also, it manages a map CEP name to CES name.
-// These maps are used by the CES manager, primarily used for storing and retrieving
-// the desired CESs by thread-safe.
+// CESToCEPMapping is used to map Cilium Endpoints to CiliumEndpointSlices and
+// retrieving all the Cilium Endpoints mapped to the given CiliumEndpointSlice.
 // This map is protected by lock for consistent and concurrent access.
 type CESToCEPMapping struct {
-	cesMutex lock.RWMutex
-	// cepNametoCESName is used to map CiliumEndpoint name to CiliumEndpointSlice name.
-	cepNametoCESName map[CEPName]CESName
-	// desiredCESs is used to map cesName to cesTracker object.
-	desiredCESs map[CESName]*cesTracker
+	mutex lock.RWMutex
+	// cepNameToCESName is used to map CiliumEndpoint name to CiliumEndpointSlice name.
+	cepNameToCESName map[CEPName]CESName
+	// cesNameToCEPNameSet is used to map CiliumEndpointSlice name to all CiliumEndpoints names it contains.
+	cesNameToCEPNameSet map[CESName]map[CEPName]struct{}
+	cesData             map[CESName]CESData
+}
+
+// CESData contains all CES data except endpoints.
+// CES is reconicled to have endpoints equal to CEPs mapped to it
+// and other fields set from the CESData.
+type CESData struct {
+	ns string
 }
 
 // Creates and intializes the new CESToCEPMapping
-func newDesiredCESMap() *CESToCEPMapping {
+func newCESToCEPMapping() *CESToCEPMapping {
 	return &CESToCEPMapping{
-		desiredCESs:      make(map[CESName]*cesTracker),
-		cepNametoCESName: make(map[CEPName]CESName),
+		cepNameToCESName:    make(map[CEPName]CESName),
+		cesNameToCEPNameSet: make(map[CESName]map[CEPName]struct{}),
+		cesData:             make(map[CESName]CESData),
 	}
 }
 
 // Insert the CEP in cache, map CEP name to CES name
-func (c *CESToCEPMapping) insertCEP(cepName, cesName string) {
-	c.cesMutex.Lock()
-	defer c.cesMutex.Unlock()
-	c.cepNametoCESName[CEPName(cepName)] = CESName(cesName)
+func (c *CESToCEPMapping) insertCEP(cepName CEPName, cesName CESName) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	c.cepNameToCESName[cepName] = cesName
+	c.cesNameToCEPNameSet[cesName][cepName] = struct{}{}
 }
 
 // Remove the CEP entry from map
-func (c *CESToCEPMapping) deleteCEP(cepName string) {
-	c.cesMutex.Lock()
-	defer c.cesMutex.Unlock()
-	delete(c.cepNametoCESName, CEPName(cepName))
+func (c *CESToCEPMapping) deleteCEP(cepName CEPName) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	delete(c.cesNameToCEPNameSet[c.cepNameToCESName[cepName]], cepName)
+	delete(c.cepNameToCESName, cepName)
 }
 
-func (c *CESToCEPMapping) getCESName(cepName string) (string, bool) {
-	c.cesMutex.RLock()
-	defer c.cesMutex.RUnlock()
-	name, ok := c.cepNametoCESName[CEPName(cepName)]
-	return string(name), ok
+// Return CES to which the given CEP is assigned
+func (c *CESToCEPMapping) getCESName(cepName CEPName) (CESName, bool) {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	name, ok := c.cepNameToCESName[cepName]
+	return name, ok
 }
 
-func (c *CESToCEPMapping) hasCEP(cepName string) bool {
-	c.cesMutex.RLock()
-	defer c.cesMutex.RUnlock()
-	_, ok := c.cepNametoCESName[CEPName(cepName)]
+func (c *CESToCEPMapping) hasCEP(cepName CEPName) bool {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	_, ok := c.cepNameToCESName[cepName]
 	return ok
 }
 
 // Return total number of CEPs stored in cache
 func (c *CESToCEPMapping) countCEPs() int {
-	c.cesMutex.RLock()
-	defer c.cesMutex.RUnlock()
-	return len(c.cepNametoCESName)
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	return len(c.cepNameToCESName)
 }
 
-// Insert the CES tracker in map
-func (c *CESToCEPMapping) insertCES(cesName string, ces *cesTracker) {
-	c.cesMutex.Lock()
-	defer c.cesMutex.Unlock()
-	c.desiredCESs[CESName(cesName)] = ces
+// Return total number of CEPs mapped to the given CES
+func (c *CESToCEPMapping) countCEPsInCES(ces CESName) int {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	return len(c.cesNameToCEPNameSet[ces])
 }
 
-// Remove the CES tracker from map
-func (c *CESToCEPMapping) deleteCES(cesName string) {
-	c.cesMutex.Lock()
-	defer c.cesMutex.Unlock()
-	delete(c.desiredCESs, CESName(cesName))
+// Return CEP Names mapped to the given CES
+func (c *CESToCEPMapping) getCEPsInCES(ces CESName) []CEPName {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	ceps := make([]CEPName, 0, len(c.cesNameToCEPNameSet[ces]))
+	for cep := range c.cesNameToCEPNameSet[ces] {
+		ceps = append(ceps, cep)
+	}
+	return ceps
 }
 
-func (c *CESToCEPMapping) getCESTracker(cesName string) (*cesTracker, bool) {
-	c.cesMutex.RLock()
-	defer c.cesMutex.RUnlock()
-	ces, ok := c.desiredCESs[CESName(cesName)]
-	return ces, ok
+// Initializes mapping structure for CES
+func (c *CESToCEPMapping) insertCES(cesName CESName, ns string) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	c.cesNameToCEPNameSet[cesName] = make(map[CEPName]struct{})
+	c.cesData[cesName] = CESData{
+		ns: ns,
+	}
 }
 
-func (c *CESToCEPMapping) getAllCESs() []*cesTracker {
-	c.cesMutex.RLock()
-	defer c.cesMutex.RUnlock()
-	var cess []*cesTracker
-	for _, ces := range c.desiredCESs {
+// Remove mapping structure for CES
+func (c *CESToCEPMapping) deleteCES(cesName CESName) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	delete(c.cesNameToCEPNameSet, cesName)
+	delete(c.cesData, cesName)
+}
+
+func (c *CESToCEPMapping) hasCESName(cesName CESName) bool {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	_, ok := c.cesNameToCEPNameSet[cesName]
+	return ok
+}
+
+// Return the total number of CESs.
+func (c *CESToCEPMapping) getCESCount() int {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	return len(c.cesNameToCEPNameSet)
+}
+
+// Return names of all CESs.
+func (c *CESToCEPMapping) getAllCESs() []CESName {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	cess := make([]CESName, 0, len(c.cesNameToCEPNameSet))
+	for ces := range c.cesNameToCEPNameSet {
 		cess = append(cess, ces)
 	}
 	return cess
 }
 
-func (c *CESToCEPMapping) hasCESName(cesName string) bool {
-	c.cesMutex.RLock()
-	defer c.cesMutex.RUnlock()
-	_, ok := c.desiredCESs[CESName(cesName)]
-	return ok
-}
-
-// Return the total number of desired CESs.
-func (c *CESToCEPMapping) getCESCount() int {
-	c.cesMutex.RLock()
-	defer c.cesMutex.RUnlock()
-	return len(c.desiredCESs)
+// Return the CES data
+func (c *CESToCEPMapping) getCESData(name CESName) CESData {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	data := c.cesData[name]
+	return data
 }
--- a/operator/pkg/ciliumendpointslice/ceptocesmap.go
+++ b/operator/pkg/ciliumendpointslice/ceptocesmap.go
@@ -3,10 +3,14 @@
 
 package ciliumendpointslice
 
-import "github.com/cilium/cilium/pkg/lock"
+import (
+	capi_v2a1 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2alpha1"
+	"github.com/cilium/cilium/pkg/k8s/resource"
+	"github.com/cilium/cilium/pkg/lock"
+)
 
-type CEPName string
-type CESName string
+type CEPName resource.Key
+type CESName resource.Key
 
 // CESToCEPMapping is used to map Cilium Endpoints to CiliumEndpointSlices and
 // retrieving all the Cilium Endpoints mapped to the given CiliumEndpointSlice.
@@ -141,4 +145,32 @@ func (c *CESToCEPMapping) getCESData(name CESName) CESData {
 	defer c.mutex.RUnlock()
 	data := c.cesData[name]
 	return data
+}
+
+func (ces CESName) key() resource.Key {
+	return resource.Key(ces)
+}
+
+func (cep CEPName) key() resource.Key {
+	return resource.Key(cep)
+}
+
+func (ces CESName) string() string {
+	return ces.key().String()
+}
+
+func (cep CEPName) string() string {
+	return cep.key().String()
+}
+
+func NewCESName(name string) CESName {
+	return CESName(resource.Key{Name: name})
+}
+
+func NewCEPName(name, ns string) CEPName {
+	return CEPName(resource.Key{Name: name, Namespace: ns})
+}
+
+func GetCEPNameFromCCEP(cep *capi_v2a1.CoreCiliumEndpoint, namespace string) CEPName {
+	return NewCEPName(cep.Name, namespace)
 }
--- a/operator/pkg/ciliumendpointslice/controller.go
+++ b/operator/pkg/ciliumendpointslice/controller.go
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package ciliumendpointslice
+
+import (
+	"context"
+	"time"
+
+	"github.com/cilium/workerpool"
+	"github.com/sirupsen/logrus"
+	"k8s.io/client-go/util/workqueue"
+
+	"github.com/cilium/cilium/pkg/hive"
+	"github.com/cilium/cilium/pkg/hive/cell"
+	v2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
+	"github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2alpha1"
+	k8sClient "github.com/cilium/cilium/pkg/k8s/client"
+	"github.com/cilium/cilium/pkg/k8s/resource"
+	"github.com/cilium/cilium/pkg/lock"
+)
+
+// params contains all the dependencies for the CiliumEndpointSlice controller.
+// They will be provided through dependency injection.
+type params struct {
+	cell.In
+
+	Logger    logrus.FieldLogger
+	Lifecycle hive.Lifecycle
+
+	Clientset           k8sClient.Clientset
+	CiliumEndpoint      resource.Resource[*v2.CiliumEndpoint]
+	CiliumEndpointSlice resource.Resource[*v2alpha1.CiliumEndpointSlice]
+
+	Cfg       Config
+	SharedCfg SharedConfig
+}
+
+type Controller struct {
+	logger        logrus.FieldLogger
+	context       context.Context
+	contextCancel context.CancelFunc
+
+	// Cilium kubernetes clients to access V2 and V2alpha1 resources
+	clientset           k8sClient.Clientset
+	ciliumEndpoint      resource.Resource[*v2.CiliumEndpoint]
+	ciliumEndpointSlice resource.Resource[*v2alpha1.CiliumEndpointSlice]
+
+	// reconciler is an util used to reconcile CiliumEndpointSlice changes.
+	reconciler *reconciler
+
+	// Manager is used to create and maintain a local datastore. Manager watches for
+	// cilium endpoint changes and enqueues/dequeues the cilium endpoint changes in CES.
+	// It maintains the desired state of the CESs in dataStore
+	manager      operations
+	slicingMode  string
+	maxCEPsInCES int
+
+	// workqueue is used to sync CESs with the api-server. this will rate-limit the
+	// CES requests going to api-server, ensures a single CES will not be proccessed
+	// multiple times concurrently, and if CES is added multiple times before it
+	// can be processed, this will only be processed only once.
+	queue           workqueue.RateLimitingInterface
+	queueTerminated chan struct{}
+	writeQPSLimit   float64
+	writeQPSBurst   int
+
+	enqueuedAt     map[CESName]time.Time
+	enqueuedAtLock lock.Mutex
+
+	wp *workerpool.WorkerPool
+}
+
+// registerController creates and initializes the CES controller
+func registerController(p params) {
+	if !p.Clientset.IsEnabled() || !p.SharedCfg.EnableCiliumEndpointSlice {
+		return
+	}
+
+	cesController := &Controller{
+		logger:              p.Logger,
+		clientset:           p.Clientset,
+		ciliumEndpoint:      p.CiliumEndpoint,
+		ciliumEndpointSlice: p.CiliumEndpointSlice,
+		slicingMode:         p.Cfg.CESSlicingMode,
+		maxCEPsInCES:        p.Cfg.CESMaxCEPsInCES,
+		writeQPSLimit:       p.Cfg.CESWriteQPSLimit,
+		writeQPSBurst:       p.Cfg.CESWriteQPSBurst,
+		enqueuedAt:          make(map[CESName]time.Time),
+	}
+
+	p.Lifecycle.Append(cesController)
+}
--- a/operator/pkg/ciliumendpointslice/controller_test.go
+++ b/operator/pkg/ciliumendpointslice/controller_test.go
@@ -1,0 +1,135 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package ciliumendpointslice
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/goleak"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/cilium/cilium/operator/k8s"
+	"github.com/cilium/cilium/pkg/hive"
+	"github.com/cilium/cilium/pkg/hive/cell"
+	cilium_v2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
+	cilium_v2a1 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2alpha1"
+	k8sClient "github.com/cilium/cilium/pkg/k8s/client"
+	"github.com/cilium/cilium/pkg/k8s/resource"
+	"github.com/cilium/cilium/pkg/testutils"
+)
+
+func TestRegisterController(t *testing.T) {
+	defer goleak.VerifyNone(
+		t,
+		// To ignore goroutine started by the workqueue. It reports metrics
+		// on unfinished work with default tick period of 0.5s - it terminates
+		// no longer than 0.5s after the workqueue is stopped.
+		goleak.IgnoreTopFunction("k8s.io/client-go/util/workqueue.(*Type).updateUnfinishedWorkLoop"),
+	)
+	var fakeClient k8sClient.FakeClientset
+	var ciliumEndpoint resource.Resource[*cilium_v2.CiliumEndpoint]
+	var ciliumEndpointSlice resource.Resource[*cilium_v2a1.CiliumEndpointSlice]
+	hive := hive.New(
+		k8sClient.FakeClientCell,
+		k8s.ResourcesCell,
+		cell.Provide(func() Config {
+			return defaultConfig
+		}),
+		cell.Provide(func() SharedConfig {
+			return SharedConfig{
+				EnableCiliumEndpointSlice: true,
+			}
+		}),
+		cell.Invoke(func(p params) error {
+			registerController(p)
+			return nil
+		}),
+		cell.Invoke(func(c *k8sClient.FakeClientset, cep resource.Resource[*cilium_v2.CiliumEndpoint], ces resource.Resource[*cilium_v2a1.CiliumEndpointSlice]) error {
+			fakeClient = *c
+			ciliumEndpoint = cep
+			ciliumEndpointSlice = ces
+			return nil
+		}),
+	)
+	if err := hive.Start(context.Background()); err != nil {
+		t.Fatalf("failed to start: %s", err)
+	}
+	cesCreated, err := createCEPandVerifyCESCreated(fakeClient, ciliumEndpoint, ciliumEndpointSlice)
+	if err != nil {
+		t.Fatalf("Couldn't verify if CES is created: %s", err)
+	}
+	// Verify CES is created when CES features is enabled
+	assert.Equal(t, true, cesCreated)
+	if err := hive.Stop(context.Background()); err != nil {
+		t.Fatalf("failed to stop: %s", err)
+	}
+}
+
+func TestNotRegisterControllerWithCESDisabled(t *testing.T) {
+	defer goleak.VerifyNone(
+		t,
+		// To ignore goroutine started by the workqueue. It reports metrics
+		// on unfinished work with default tick period of 0.5s - it terminates
+		// no longer than 0.5s after the workqueue is stopped.
+		goleak.IgnoreTopFunction("k8s.io/client-go/util/workqueue.(*Type).updateUnfinishedWorkLoop"),
+	)
+	var fakeClient k8sClient.FakeClientset
+	var ciliumEndpoint resource.Resource[*cilium_v2.CiliumEndpoint]
+	var ciliumEndpointSlice resource.Resource[*cilium_v2a1.CiliumEndpointSlice]
+	h := hive.New(
+		k8sClient.FakeClientCell,
+		k8s.ResourcesCell,
+		cell.Provide(func() Config {
+			return defaultConfig
+		}),
+		cell.Provide(func() SharedConfig {
+			return SharedConfig{
+				EnableCiliumEndpointSlice: false,
+			}
+		}),
+		cell.Invoke(func(p params) error {
+			registerController(p)
+			return nil
+		}),
+		cell.Invoke(func(c *k8sClient.FakeClientset, cep resource.Resource[*cilium_v2.CiliumEndpoint], ces resource.Resource[*cilium_v2a1.CiliumEndpointSlice]) error {
+			fakeClient = *c
+			ciliumEndpoint = cep
+			ciliumEndpointSlice = ces
+			return nil
+		}),
+	)
+	if err := h.Start(context.Background()); err != nil {
+		t.Fatalf("failed to start: %s", err)
+	}
+	cesCreated, err := createCEPandVerifyCESCreated(fakeClient, ciliumEndpoint, ciliumEndpointSlice)
+	if err != nil {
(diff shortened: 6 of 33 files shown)


dlapcevic:
cc @dlapcevic @aojea

operator/pkg/ciliumendpointslice/ceptocesmap.go
```diff
@@ -8,104 +8,136 @@ import "github.com/cilium/cilium/pkg/lock"
...
+	delete(c.cesNameToCEPNameSet[c.cepNameToCESName[cepName]], cepName)
+	delete(c.cepNameToCESName, cepName)
 }
 
-func (c *CESToCEPMapping) getCESName(cepName string) (string, bool) {
-	c.cesMutex.RLock()
-	defer c.cesMutex.RUnlock()
-	name, ok := c.cepNametoCESName[CEPName(cepName)]
-	return string(name), ok
+// Return CES to which the given CEP is amm
```
dlapcevic: typo: CEP is assigned

operator/pkg/ciliumendpointslice/ceptocesmap.go
```diff
@@ -8,104 +8,136 @@ import "github.com/cilium/cilium/pkg/lock"
...
-	var cess []*cesTracker
-	for _, ces := range c.desiredCESs {
-		cess = append(cess, ces)
-	}
-	return cess
+// Remove mapping structure for CES
+func (c *CESToCEPMapping) deleteCES(cesName CESName) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	delete(c.cesNameToCEPNameSet, cesName)
```
dlapcevic, line 113: Should it also be removed from `c.cesData`?

operator/pkg/ciliumendpointslice/manager_test.go
```diff
@@ -6,241 +6,108 @@ package ciliumendpointslice
...
-
-	// Insert CEPs in Cache and count total number of CES and CEP
-	t.Run("Test Inserting CEPs in cache and count number of CEPs and CESs", func(*testing.T) {
-		m := newCESManagerFcfs(newQueue(), 2)
-		m.InsertCEPInCache(cep1, "kube-system")
-		m.InsertCEPInCache(cep2, "kube-system")
-		m.InsertCEPInCache(cep3, "kube-system")
-		assert.Equal(t, m.getCESCount(), 2, "Total number of CESs allocated is 2")
-		assert.Equal(t, m.getTotalCEPCount(), 3, "Total number of CEPs inserted is 3")
+func TestFCFSManager(t *testing.T) {
```
dlapcevic, line 21: Consider adding a test with different namespaces, to verify that each CES can only take CEPs from a single namespace.
`UpdateCEPMapping(getCEP("c1", 0), "ns-a")`
`UpdateCEPMapping(getCEP("c2", 0), "ns-b")`

Same for `TestIdentityManager()` below.

operator/pkg/ciliumendpointslice/endpointslice.go
```diff
@@ -53,11 +52,14 @@ const (
 	// Default CES Synctime, multiple consecutive syncs with k8s-apiserver are
 	// batched and synced together after a short delay.
 	DefaultCESSyncTime = 500 * time.Millisecond
+	UpdateEvent        = 1
```
dlapcevic: nit: Should it be exported (capital first letter)?
Anyway maybe it's better to make it like this:
```
type EventType int

const (
	AddEvent EventType = iota
	UpdateEvent
	DeleteEvent
)
```

And then update EndpointEvent.event to be EventType.

operator/pkg/ciliumendpointslice/endpointslice.go
```diff
@@ -102,24 +104,47 @@ func GetCEPNameFromCCEP(cep *capi_v2a1.CoreCiliumEndpoint, namespace string) str
...
+		slicingMode:                    slicingMode,
+		workerLoopPeriod:               1 * time.Second,
+		enqueuedAt:                     make(map[string]time.Time),
+		preInitEnqueuedEndpointsEvents: make([]EndpointEvent, 0),
+		endpointsMappingInitialized:    false,
+	}
+
+	cepWatcher.AddOnUpdateHandler(controller.onEndpointUpdate)
+	cepWatcher.AddOnDeleteHandler(controller.onEndpointDelete)
+	cesWatcher.AddOnDeleteHandler(controller.onSliceUpdate)
```
dlapcevic: typo: Should be `AddOnUpdateHandler`

operator/pkg/ciliumendpointslice/manager.go
```diff
@@ -214,396 +99,100 @@ func uniqueCESliceName(desiredCESs *CESToCEPMapping) string {
 //     with an empty name, it generates a random unique name and assign it to the CES.
 //  2. During operator warm boot [after crash or software upgrade], slicing manager
 //     creates a CES, by passing unique name.
-func (c *cesMgr) createCES(name string) *cesTracker {
-	var cesName string = name
+func (c *cesMgr) createCES(name, ns string) CESName {
+	var cesName = name
```
dlapcevic: nit: `cesName := name`

operator/pkg/ciliumendpointslice/manager.go
```diff
@@ -214,396 +99,100 @@ func uniqueCESliceName(desiredCESs *CESToCEPMapping) string {
...
+			logfields.CEPName: cepName,
+			logfields.CESName: cesName,
+		}).Debug("CEP already mapped to CES")
+		return []CESName{cesName}
 	}
 
 	// If given cep object isn't packed in any of the CES. find a new ces
 	// to pack this cep.
-	cb, cesName := func() (*cesTracker, string) {
+	for {
```
dlapcevic: Is the loop needed because `getLargestAvailableCESForNamespace()` can return a `cesName` that can get full before locking the cache, even though the function returns the name of the largest non-full CES?

operator/pkg/ciliumendpointslice/manager.go
```diff
@@ -214,396 +99,100 @@ func uniqueCESliceName(desiredCESs *CESToCEPMapping) string {
...
-		logfields.CEPCount: len(ces.ces.Endpoints),
-	}).Debug("Removed CEP from CES")
-
-	// Increment the cepRemove counter
-	ces.cepRemoved += 1
-	c.insertCESInWorkQueue(ces, baseDelay)
+	return ""
 }
 
 // getLargestAvailableCESForNamespace returns the largest CES from cache for the
```
dlapcevic, line 162: Update the comment that it returns the name of the largest, and that it doesn't guarantee that the CESs is not full by the time it's read from cache by name.

operator/pkg/ciliumendpointslice/manager.go
```diff
@@ -612,133 +201,109 @@ func (c *cesManagerIdentity) InsertCEPInCache(cep *cilium_v2.CoreCiliumEndpoint,
...
-func (c *cesMgr) getCESQueueDelayInSeconds(cesName string) (diff float64) {
-	ces, exists := c.desiredCESs.getCESTracker(cesName)
-	if !exists {
-		return
+func (c *cesManagerIdentity) removeCESToIdentity(id int64, cesName CESName) {
+	c.identityLock.Lock()
+	cesSlice := c.identityToCES[id]
+	for i, ces := range cesSlice {
+		if ces == cesName {
+			cesSlice[i] = cesSlice[len(cesSlice)-1]
```
dlapcevic, line 277: I don't understand this. How is cesSlice ever reducing the number of elements?

operator/pkg/ciliumendpointslice/reconciler.go
```diff
@@ -29,71 +35,174 @@ func newReconciler(client clientset.CiliumV2alpha1Interface, cesMgr operations)
...
-	if ces, err = r.cesManager.getCESCopyFromCache(cesToUpdate); err != nil {
-		return
+	// Iterate through existing CEPs in the CES
+	// remove stale ones
+	// update outdated
+	// Iterate through expected CEPs in the CES
+	// insert missing
+	ceps := r.cesManager.getCEPinCES(cesName)
+	desiredEndpoints := make([]cilium_v2a1.CoreCiliumEndpoint, 0, len(ceps))
+	desiredObjects := make(map[CEPName]*cilium_v2a1.CoreCiliumEndpoint)
```
dlapcevic: It's a little difficult to follow `desiredEndpoints` and `desiredObjects`.
Consider renaming `desiredEndpoints` to `updatedEndpoints`, and `desiredObjects` to `desiredEndpoints`.

operator/watchers/ciliumendpoint/utils.go
```diff
@@ -0,0 +1,39 @@
...
+		return false
+	}
+	ces, _ := CiliumEndpointStore.IndexKeys(identityIndex, identity)
+
+	return len(ces) != 0
+}
+
+// HasCE returns true or false if the Cilium Endpoint store has the endpoint
+// with the given name.
+func HasCE(ns, name string) (*ciliumApiV2.CiliumEndpoint, bool, error) {
```
dlapcevic: The function name and description isn't matching. It should be `GetCE`.

pippolo84:
Thanks for the PR!

Currently, there is an ongoing effort to modularize each component of the operator, in order to improve the readability and the testability of the codebase. To do that we introduced the [Hive](https://docs.cilium.io/en/latest/contributing/development/hive/#id1) package and we started to organize each component into a separate cell.
As an example, that kind of modularization has already been done in the operator for the [CiliumIdentity GC](https://github.com/cilium/cilium/pull/22892).
There is also a [Work-In-Progress PR](https://github.com/cilium/cilium/pull/27037) for the CiliumEndpoint GC.

Would you be willing to follow the same approach for the CES controller? I think that the linked PR for the CiliumEndpoint GC should be a good example to follow, since it shows how to handle events for k8s resources through the `Resource[T]`-based providers. Besides, it is strictly related to the CES controller code.

I'm more than happy to help if you need any kind of clarifications and/or further examples.

alan-kut:
I was occupied by other things but I plan to fix the current issues with the current version soon and then I will upgrade the code to follow the same modularization apprach for the CES controller

pippolo84:
> I was occupied by other things but I plan to fix the current issues with the current version soon and then I will upgrade the code to follow the same modularization apprach for the CES controller

That's great, thanks! :rocket:

operator/pkg/ciliumendpointslice/ceptocesmap.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/ceptocesmap.go
alan-kut: Yes, fixed.

operator/pkg/ciliumendpointslice/manager_test.go
alan-kut: Added

operator/pkg/ciliumendpointslice/endpointslice.go
alan-kut: Done

operator/pkg/ciliumendpointslice/endpointslice.go
alan-kut: Fixed.

operator/pkg/ciliumendpointslice/manager.go
alan-kut: Yes, that's the reason

alan-kut: Locked sooner so the loop is not needed here.

operator/pkg/ciliumendpointslice/manager.go
alan-kut: I think it should be clear that if this only returns a name of CES the state may change in the meantime.

operator/pkg/ciliumendpointslice/manager.go
alan-kut: CES is deleted when all the elements inside are deleted.

operator/pkg/ciliumendpointslice/reconciler.go
alan-kut: Simplified a code a little bit. It may reshuffle the CEPs on update now but it's much easier to track

operator/watchers/ciliumendpoint/utils.go
alan-kut: Done

operator/pkg/ciliumendpointslice/manager.go
alan-kut: Fixed the code:
- removed second mutex (everything is now on one mutex)
- added deleting CES from the local operator cache

alan-kut:
Resolved comments and fixed checks.

Next step: upgrade the code to follow the same modularization apprach for the CES controller

dlapcevic (approved):
/lgtm

I'll review the changes for "Next step: upgrade the code to follow the same modularization apprach for the CES controller" when it's ready.
Please make it a separate commit.

alan-kut:
@pippolo84 Please take a look. I'm not sure yet that everything is fully correct and passes the tests but at least hive part is done.

pippolo84:
> @pippolo84 Please take a look. I'm not sure yet that everything is fully correct and passes the tests but at least hive part is done.

Thanks for all the modularization work @alan-kut !
Right now I'm on PTO, but I'm happy to review this as soon as I'll be back at the end of August.

operator/cmd/root.go
```diff
@@ -153,6 +153,8 @@ var (
 			// setup operations. This is a hacky workaround until the kvstore is
 			// refactored into a proper cell.
 			identitygc.Cell,
+
+			ciliumendpointslice.Cell,
```
joamaki, line 174: Add a short one or two line comment here describing the functionality of this cell. It helps getting an overview of what each top-level cell is doing and how they relate to each other.

operator/pkg/ciliumendpointslice/cell.go
```diff
@@ -0,0 +1,57 @@
...
+	// requests to the kube-apiserver.
+	CESWriteQPSLimit = "ces-write-qps-limit"
+
+	// CESWriteQPSBurst is the burst rate per second used with CESWriteQPSLimit
+	// for the CES work queue to process CES events that result in CES write
+	// (Create, Update, Delete) requests to the kube-apiserver.
+	CESWriteQPSBurst = "ces-write-qps-burst"
+)
+
+var Cell = cell.Module(
```
joamaki, line 34: Add descriptive comment here describing what this cell does. E.g. think of the `var Cell` definition as something someone new to the codebase would be hovering over in the `root.go` file and trying to piece together what's going on. Having clear description here about what this implements makes the codebase easier to understand and navigate.

(Incidentally have an issue to "document how to document" hive cells :smile:)

operator/pkg/ciliumendpointslice/controller.go
```diff
@@ -0,0 +1,111 @@
...
+
+	Logger    logrus.FieldLogger
+	Lifecycle hive.Lifecycle
+
+	Clientset           k8sClient.Clientset
+	CiliumEndpoint      resource.Resource[*v2.CiliumEndpoint]
+	CiliumEndpointSlice resource.Resource[*v2alpha1.CiliumEndpointSlice]
+
+	Cfg Config
+	//daemonCfg *option.DaemonConfig
```
joamaki: Drop this?

operator/pkg/ciliumendpointslice/controller.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// registerController creates and initializes the CES controller
+func registerController(p params) {
+	if !p.Clientset.IsEnabled() && option.Config.EnableCiliumEndpointSlice {
+		return
+	}
+
+	cesController := &Controller{
+		logger:                         p.Logger.WithField(logfields.LogSubsys, "ces-controller"),
```
joamaki: The `p.Logger` already has `LogSubsys` set to `k8s-ces-controller` via `cell.Module`, so no need to set it again.

operator/pkg/ciliumendpointslice/controller.go
```diff
@@ -0,0 +1,111 @@
...
+		writeQPSBurst:                  p.Cfg.CESWriteQPSBurst,
+		workerLoopPeriod:               1 * time.Second,
+		enqueuedAt:                     make(map[CESName]time.Time),
+		preInitEnqueuedEndpointsEvents: make([]EndpointEvent, 0),
+		endpointsMappingInitialized:    false,
+	}
+
+	p.Lifecycle.Append(hive.Hook{
+		OnStart: func(context hive.HookContext) error {
+			return cesController.Start(context)
```
joamaki: It would be preferable to do `p.Lifecycle.Append(cesController)` as then the `cilium-agent hive` output that shows the start and stop hooks would have nice function names rather than anonymous functions.

`Append` takes a `HookInterface` containing the `Start` and `Stop` methods. `Hook` is an alternative way of implementing the hooks, but it's usually preferably to have an object implement the `HookInterface` directly.

operator/pkg/ciliumendpointslice/reconciler.go
```diff
@@ -7,93 +7,209 @@ import (
...
 	cesManager operations
+	cepStore   resource.Store[*cilium_v2.CiliumEndpoint]
+	cesStore   resource.Store[*cilium_v2a1.CiliumEndpointSlice]
 }
 
 // newReconciler creates and initializes a new reconciler.
-func newReconciler(client clientset.CiliumV2alpha1Interface, cesMgr operations) *reconciler {
+func newReconciler(client clientset.CiliumV2alpha1Interface, cesMgr operations,
+	logger logrus.FieldLogger, ciliumEndpoint resource.Resource[*cilium_v2.CiliumEndpoint], ciliumEndpointSlice resource.Resource[*cilium_v2a1.CiliumEndpointSlice]) *reconciler {
+	cepStore, _ := ciliumEndpoint.Store(context.Background())
```
joamaki: Do not call blocking (Store blocks until resource has synchronized) or side-effectful methods from a constructor as this is also called when inspecting hive (e.g. `go run ./operator hive`). Instead call these from a `Start` hook and use the start context to allow aborting the start.

https://docs.cilium.io/en/latest/contributing/development/hive/#guidelines

operator/watchers/ciliumendpoint/cilium_endpoint.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/cilium/cilium/pkg/logging"
+	"github.com/cilium/cilium/pkg/logging/logfields"
+)
+
+const identityIndex = "identity"
+
+type cepHanlderFunc func(cep *ciliumApiV2.CiliumEndpoint)
+
+var (
+	log      = logging.DefaultLogger.WithField(logfields.LogSubsys, "watchers")
```
joamaki: s/watchers/ciliumendpoint/

operator/watchers/ciliumendpoint/cilium_endpoint.go
```diff
@@ -0,0 +1,153 @@
...
+	once = sync.Once{}
+	CiliumEndpointsSynced = make(chan struct{})
+	CiliumEndpointsInit(ctx, wg, clientset)
+}
+
+// CiliumEndpointsInit starts a CiliumEndpointWatcher
+func CiliumEndpointsInit(ctx context.Context, wg *sync.WaitGroup, clientset k8sClient.Clientset) {
+	once.Do(func() {
+		CiliumEndpointStore = cache.NewIndexer(cache.DeletionHandlingMetaNamespaceKeyFunc, indexers)
+		ciliumEndpointInformer := informer.NewInformerWithStore(
```
joamaki: We should not have two informers for the same object kind. Is this still needed after adding `Resource[*CiliumEndpoint]`? Would it be hard to convert the remaining uses? If it is consider converting this internally to use `Resource[*CiliumEndpoint]` as an intermediate step towards removing it. https://github.com/cilium/cilium/blob/main/pkg/k8s/watchers/pod.go#L124

operator/watchers/ciliumendpointslice/cilium_endpoint_slice.go
```diff
@@ -0,0 +1,119 @@
...
+func CiliumEndpointSliceReinitialize(ctx context.Context, wg *sync.WaitGroup, clientset k8sClient.Clientset) {
+	once = sync.Once{}
+	CiliumEndpointSlicesSynced = make(chan struct{})
+	CiliumEndpointSliceInit(ctx, wg, clientset)
+}
+
+func CiliumEndpointSliceInit(ctx context.Context, wg *sync.WaitGroup, clientset k8sClient.Clientset) {
+	once.Do(func() {
+		var ciliumEndpointSliceInformer cache.Controller
+		CiliumEndpointSliceStore, ciliumEndpointSliceInformer = informer.NewInformer(
```
joamaki: Same here, we really don't want duplicate informers.

joamaki (changes requested):
Hey, this is looking great!

Any feedback on `pkg/hive` and `pkg/k8s/resource`? Did you find anything confusing or documentation lacking?

alan-kut:
> Any feedback on pkg/hive and pkg/k8s/resource? Did you find anything confusing or documentation lacking?
https://docs.cilium.io/en/latest/contributing/development/hive/ is very good and I managed to learn how to use hive quite quickly.

I can think about two problems - please treat it as 
I had problems with testing - it was hard for me to provide resources and fake client - I ended up using hive to provide those for test.

It wasn't also clear how to wait for syncing resource cache.
I needed to go through the code to make sure I'll not get any event before getting cache sync event.
Then I learned that I can block on cache sync by getting the store.

operator/cmd/root.go
alan-kut: Done

operator/pkg/ciliumendpointslice/controller.go
alan-kut: Fixed. I wasn't sure how to propagate option.Config.EnableCiliumEndpointSlice to CES Controller but at the end I used it directly so at the end it's not needed and I forgot to remove it.

operator/pkg/ciliumendpointslice/controller.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/controller.go
alan-kut: Fixed, thanks. It makes sense.

operator/pkg/ciliumendpointslice/reconciler.go
alan-kut: This is called on Start - CES Controller creates reconciler (using newReconciler).
I moved this to Start just because of this.

I see that newReconciler may be confusing. I don't have much better idea as it doesn't do anything else except for getting those stores, logger and client.

operator/watchers/ciliumendpoint/cilium_endpoint.go
alan-kut: Done

operator/watchers/ciliumendpoint/cilium_endpoint.go
alan-kut: When I started working on the Full CES Reconciliation I was unaware of the hive.
I started by fixing the watchers, implemented full reconciliation and then I learned about the hive and I moved CES Controller to hive.

If I knew about the hive I wouldn't refactor legacy watchers. I can revert this refactoring if needed.
I'll try to revert this refactoring.
This won't help with the main issue - there are still use cases on legacy watchers.

I talked to @pippolo84 about this problem and it's tricky one. It may be hard to move everything out of legacy watchers at once. He agreed that we may have some period of time in which we would have two informers.

I'm trying to move other use case to hive in the #27458
I could work on all remaining use cases but there is already some work done by @pippolo84 so I'm not sure whether it's ok to hijack it.

operator/pkg/ciliumendpointslice/cell.go
alan-kut: Added

alan-kut:
@joamaki Thank you for the review.

I replied to all the comments.

I also:
- reverted the refactoring on the watchers - should've done this when I moved CES Controller to hive
- due to this I needed to move Identity GC to CES resource so I decided to combine the other PR with this one
- I fixed the leaks from hive.Stop() and added such test for the CES Controller as well

alan-kut:
I will be out till Wednesday

operator/pkg/ciliumendpointslice/reconciler.go
joamaki: Ah sorry, I missed that it was called from a start hook. In that case I don't have any problem with this. I think the naming etc. is fine here. I should've checked how it was used and not assume.

operator/identitygc/crd_gc_test.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+func TestUsedIdentitiesInCESs(t *testing.T) {
+	wait := make(chan struct{})
+	var fakeClient k8sClient.FakeClientset
+	var ciliumEndpointSlice resource.Resource[*cilium_v2a1.CiliumEndpointSlice]
+	hive := hive.New(
+		k8sClient.FakeClientCell,
+		k8s.ResourcesCell,
+		cell.Invoke(func(c *k8sClient.FakeClientset, cep resource.Resource[*cilium_v2.CiliumEndpoint], ces resource.Resource[*cilium_v2a1.CiliumEndpointSlice]) error {
```
pippolo84: `cep` parameter is unused, it can be removed

operator/identitygc/crd_gc_test.go
```diff
@@ -0,0 +1,88 @@
...
+			fakeClient = *c
+			ciliumEndpointSlice = ces
+			close(wait)
+			return nil
+		}),
+	)
+	hive.Start(context.Background())
+	<-wait
```
pippolo84: `Invoke` calls are executed synchronously before `Start` returns, so there's no need to wait here (but I suggest to add a check for the error returned by `hive.Start`).

operator/identitygc/crd_gc_test.go
```diff
@@ -0,0 +1,88 @@
...
+	time.Sleep(50 * time.Millisecond)
+	wantIdentities["10"] = true
+	wantIdentities["20"] = true
+	wantIdentities["30"] = true
+	wantIdentities["40"] = true
+	wantIdentities["50"] = true
+	gotIdentities = usedIdentitiesInCESs(cesStore)
+	assertEqualIDs(t, wantIdentities, gotIdentities)
+
+	hive.Stop(context.Background())
```
pippolo84: Same here, better to check for the error returned by `hive.Stop`

operator/identitygc/crd_gc_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// Empty store.
+	gotIdentities := usedIdentitiesInCESs(cesStore)
+	wantIdentities := make(map[string]bool)
+	assertEqualIDs(t, wantIdentities, gotIdentities)
+
+	// 5 IDs in the store.
+	cesA := createCESWithIDs("cesA", []int64{1, 2, 3, 4, 5})
+	fakeClient.CiliumV2alpha1().CiliumEndpointSlices().Create(context.Background(), cesA, meta_v1.CreateOptions{})
+	time.Sleep(50 * time.Millisecond)
```
pippolo84: One-shot `time.Sleep`s with fixed duration don't play nice with the unpredictability of the CI environment. To avoid introducing flakiness, I suggest to use `testutils.WaitUntil` or `testutils.WaitUntilWithSleep` with a very large timeout.

operator/identitygc/crd_gc_test.go
```diff
@@ -0,0 +1,88 @@
...
+	wantIdentities["3"] = true
+	wantIdentities["4"] = true
+	wantIdentities["5"] = true
+	gotIdentities = usedIdentitiesInCESs(cesStore)
+	assertEqualIDs(t, wantIdentities, gotIdentities)
+
+	// 10 IDs in the store.
+	cesB := createCESWithIDs("cesB", []int64{10, 20, 30, 40, 50})
+	fakeClient.CiliumV2alpha1().CiliumEndpointSlices().Create(context.Background(), cesB, meta_v1.CreateOptions{})
+	time.Sleep(50 * time.Millisecond)
```
pippolo84: Same here, use `testutils.WaitUntil` or `testutils.WaitUntilWithSleep`

operator/pkg/ciliumendpointslice/ceptocesmap.go
```diff
@@ -8,104 +8,137 @@ import "github.com/cilium/cilium/pkg/lock"
...
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	return len(c.cesNameToCEPNameSet)
+}
+
+// Return names of all CESs.
+func (c *CESToCEPMapping) getAllCESs() []CESName {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	var cess []CESName
```
pippolo84: Nit: `cess := make([]CESName, 0, len(c.cesNameToCEPNameSet))`

operator/pkg/ciliumendpointslice/ceptocesmap.go
```diff
@@ -8,104 +8,137 @@ import "github.com/cilium/cilium/pkg/lock"
...
-// Remove the CES tracker from map
-func (c *CESToCEPMapping) deleteCES(cesName string) {
-	c.cesMutex.Lock()
-	defer c.cesMutex.Unlock()
-	delete(c.desiredCESs, CESName(cesName))
+// Return CEP Names mapped to the given CES
+func (c *CESToCEPMapping) getCEPsInCES(ces CESName) []CEPName {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	ceps := []CEPName{}
```
pippolo84: Nit: `ceps := make([]CEPName, 0, len(c.cesNameToCEPNameSet[ces]))`

operator/pkg/ciliumendpointslice/controller.go
```diff
@@ -0,0 +1,97 @@
...
+	enqueuedAt                     map[CESName]time.Time
+	preInitEnqueuedEndpointsEvents []EndpointEvent
+	endpointsMappingInitialized    bool
+
+	wp *workerpool.WorkerPool
+}
+
+// registerController creates and initializes the CES controller
+func registerController(p params) {
+	if !p.Clientset.IsEnabled() && option.Config.EnableCiliumEndpointSlice {
```
pippolo84: Shouldn't this be `if !p.Clientset.IsEnabled() || !option.Config.EnableCiliumEndpointSlice {` ?

operator/pkg/ciliumendpointslice/endpointslice_test.go
```diff
@@ -51,115 +52,137 @@ func createStoreEndpointSlice(name string, namespace string, endpoints []capi_v2
...
+		cell.Invoke(func(c *k8sClient.FakeClientset, cep resource.Resource[*cilium_v2.CiliumEndpoint], ces resource.Resource[*cilium_v2a1.CiliumEndpointSlice]) error {
+			fakeClient = *c
+			ciliumEndpoint = cep
+			ciliumEndpointSlice = ces
+			close(wait)
+			return nil
+		}),
+	)
+	hive.Start(context.Background())
+	<-wait
```
pippolo84: No need to wait here.

operator/pkg/ciliumendpointslice/endpointslice_test.go
```diff
@@ -51,115 +52,137 @@ func createStoreEndpointSlice(name string, namespace string, endpoints []capi_v2
...
```
pippolo84: No need to wait here.

pippolo84 (changes requested):
Thanks for the PR and for the additional effort to refactor the identitygc.
I did a first pass and left some suggestions inline, I'm planning to do another pass focusing on the CES controller inner workings tomorrow.

operator/pkg/ciliumendpointslice/endpointslice.go
```diff
@@ -48,263 +48,237 @@ const (
...
+// start the worker thread, reconciles the modified CESs with api-server
+func (c *Controller) Start(ctx hive.HookContext) error {
+	c.logger.Info("Bootstrap ces controller")
+	defer utilruntime.HandleCrash()
+	if c.slicingMode == cesIdentityBasedSlicing {
+		c.manager = newCESManagerIdentity(c.maxCEPsInCES, c.logger)
+	} else {
+		c.manager = newCESManagerFcfs(c.maxCEPsInCES, c.logger)
+	}
+	c.reconciler = newReconciler(c.clientset.CiliumV2alpha1(), c.manager, c.logger, c.ciliumEndpoint, c.ciliumEndpointSlice)
```
pippolo84: We should pass down the `hive.HookContext` here in order to use it when Creating/Updating/Deleting CESs in place of `context.TODO()`.

operator/pkg/ciliumendpointslice/reconciler.go
```diff
@@ -7,93 +7,214 @@ import (
...
+		}
 	}
 
 	// Call the client API, to Create CES
-	if retCES, err = r.client.CiliumEndpointSlices().Create(
-		context.TODO(), ces, metav1.CreateOptions{}); err != nil {
-		log.WithError(err).WithFields(logrus.Fields{
-			logfields.CESName: cesToCreate,
+	if _, err = r.client.CiliumEndpointSlices().Create(
+		context.TODO(), newCES, meta_v1.CreateOptions{}); err != nil {
```
pippolo84: Following the above comment, we should use the context from the upper level and extend the check as `err != nil && !errors.Is(err, context.Canceled)`.
Same for the update and deletion.

operator/pkg/ciliumendpointslice/endpointslice_test.go
```diff
@@ -54,240 +52,137 @@ func createStoreEndpointSlice(name string, namespace string, endpoints []capi_v2
...
+	cesN, _ = mapping.getCESName(NewCEPName("cep3", "ns"))
+	assert.Equal(t, cesN, NewCESName("ces1"))
+	cesN, _ = mapping.getCESName(NewCEPName("cep4", "ns"))
+	assert.Equal(t, cesN, NewCESName("ces1"))
+	cesN, _ = mapping.getCESName(NewCEPName("cep5", "ns"))
+	assert.Equal(t, cesN, NewCESName("ces2"))
+	cesN, _ = mapping.getCESName(NewCEPName("cep6", "ns"))
+	assert.Equal(t, cesN, NewCESName("ces2"))
+	cesN, _ = mapping.getCESName(NewCEPName("cep7", "ns"))
+	assert.Equal(t, cesN, NewCESName("ces2"))
```
pippolo84, line 116: The call to hive.Stop is missing, this test will leak goroutines.

operator/pkg/ciliumendpointslice/endpointslice_test.go
```diff
@@ -54,240 +52,137 @@ func createStoreEndpointSlice(name string, namespace string, endpoints []capi_v2
...
+	assert.Equal(t, 0, cesController.queue.Len())
+	assert.Equal(t, 4, len(cesController.preInitEnqueuedEndpointsEvents))
+	assert.Equal(t, false, cesController.endpointsMappingInitialized)
+
+	cesController.syncCESsInLocalCache(context.Background())
+	assert.Equal(t, true, cesController.endpointsMappingInitialized)
+
+	cesController.processEnqueuedPreInitEndpoints()
+	time.Sleep(DefaultCESSyncTime * 2) // elements are added to the queue with a default delay
+	assert.Equal(t, 2, cesController.queue.Len())
```
pippolo84: Missing call to hive.Stop

operator/pkg/ciliumendpointslice/reconciler_test.go
```diff
@@ -4,217 +4,226 @@
...
+	m.mapping.insertCEP(NewCEPName("cep2", "ns"), NewCESName("ces1"))
+	m.mapping.insertCEP(NewCEPName("cep3", "ns"), NewCESName("ces2"))
+	r.reconcileCES(NewCESName("ces1"))
+
+	assert.Equal(t, "ces1", createdSlice.Name)
+	assert.Equal(t, 2, len(createdSlice.Endpoints))
+	assert.Equal(t, "ns", createdSlice.Namespace)
+	eps := []string{createdSlice.Endpoints[0].Name, createdSlice.Endpoints[1].Name}
+	assert.Contains(t, eps, "cep1")
+	assert.Contains(t, eps, "cep2")
```
pippolo84, line 73: Missing call to hive.Stop

operator/pkg/ciliumendpointslice/reconciler_test.go
```diff
@@ -4,217 +4,226 @@
...
+	// ces1 contains cep1 and cep3, but it's mapped to cep1 and cep2
+	// so it's expected that after update it would contain cep1 and cep2
+	r.reconcileCES(NewCESName("ces1"))
+
+	assert.Equal(t, "ces1", updatedSlice.Name)
+	assert.Equal(t, 2, len(updatedSlice.Endpoints))
+	assert.Equal(t, "ns", updatedSlice.Namespace)
+	eps := []string{updatedSlice.Endpoints[0].Name, updatedSlice.Endpoints[1].Name}
+	assert.Contains(t, eps, "cep1")
+	assert.Contains(t, eps, "cep2")
```
pippolo84, line 128: Missing call to hive.Stop

operator/pkg/ciliumendpointslice/reconciler_test.go
```diff
@@ -4,217 +4,226 @@
...
+	cesStore.CacheStore().Add(ces1)
+	m.mapping.insertCES(NewCESName("ces1"), "ns")
+	m.mapping.insertCES(NewCESName("ces2"), "ns")
+	m.mapping.insertCEP(NewCEPName("cep1", "ns"), NewCESName("ces2"))
+	m.mapping.insertCEP(NewCEPName("cep2", "ns"), NewCESName("ces2"))
+	m.mapping.insertCEP(NewCEPName("cep3", "ns"), NewCESName("ces2"))
+	// ces1 contains cep1 and cep3, but it's mapped to nothing so it should be deleted
+	r.reconcileCES(NewCESName("ces1"))
+
+	assert.Equal(t, "ces1", deletedSlice)
```
pippolo84, line 177: Missing call to hive.Stop

operator/pkg/ciliumendpointslice/reconciler_test.go
```diff
@@ -4,217 +4,226 @@
...
+	cepStore.CacheStore().Add(cep3)
+	m.mapping.insertCES(NewCESName("ces1"), "ns")
+	m.mapping.insertCES(NewCESName("ces2"), "ns")
+	m.mapping.insertCEP(NewCEPName("cep1", "ns"), NewCESName("ces2"))
+	m.mapping.insertCEP(NewCEPName("cep2", "ns"), NewCESName("ces2"))
+	m.mapping.insertCEP(NewCEPName("cep3", "ns"), NewCESName("ces2"))
+	// ces1 contains cep1 and cep3, but it's mapped to nothing so it should be deleted
+	r.reconcileCES(NewCESName("ces1"))
+
+	assert.Equal(t, true, noRequest)
```
pippolo84, line 222: Missing call to hive.Stop

operator/pkg/ciliumendpointslice/controller.go
```diff
@@ -0,0 +1,97 @@
...
+	Lifecycle hive.Lifecycle
+
+	Clientset           k8sClient.Clientset
+	CiliumEndpoint      resource.Resource[*v2.CiliumEndpoint]
+	CiliumEndpointSlice resource.Resource[*v2alpha1.CiliumEndpointSlice]
+
+	Cfg Config
+}
+
+type EndpointEvent struct {
```
pippolo84: No need to export this type.

operator/pkg/ciliumendpointslice/endpointslice.go
```diff
@@ -48,263 +48,237 @@ const (
...
-		// Create new CES locally, with the given cesName
-		manager.createCES(ces.GetName())
-
-		// Deep copy the ces, we got from api-server to local datastore.
-		manager.updateCESInCache(ces, true)
-
+func (c *Controller) processEnqueuedPreInitEndpoints() {
+	for _, e := range c.preInitEnqueuedEndpointsEvents {
+		if e.event == updateEvent {
+			c.onEndpointUpdate(e.cep)
```
pippolo84: I suggest to add a Debug log here too, as already done in `runCiliumEndpointsUpdater`, it might be convenient. Same for the deletion.

operator/identitygc/crd_gc_test.go
alan-kut: Fixed

operator/identitygc/crd_gc_test.go
alan-kut: Fixed. I think I had problems with it at some point but it works now. Removed

operator/identitygc/crd_gc_test.go
alan-kut: Fixed

operator/identitygc/crd_gc_test.go
alan-kut: Glad I could learn something new. Fixed.

operator/identitygc/crd_gc_test.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/ceptocesmap.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/ceptocesmap.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/controller.go
alan-kut: You are right. Fixed. Moved the EnableCiliumEndpointSlice to shared config and added test to verify it is used correctly.

operator/pkg/ciliumendpointslice/endpointslice_test.go
alan-kut: Fixed here and in the other places in the CES controller tests

operator/pkg/ciliumendpointslice/endpointslice_test.go
alan-kut: fixed

operator/pkg/ciliumendpointslice/endpointslice_test.go
```diff
@@ -54,240 +52,131 @@ func createStoreEndpointSlice(name string, namespace string, endpoints []capi_v2
...
+	cesN, _ = mapping.getCESName(NewCEPName("cep3", "ns"))
+	assert.Equal(t, cesN, NewCESName("ces1"))
+	cesN, _ = mapping.getCESName(NewCEPName("cep4", "ns"))
+	assert.Equal(t, cesN, NewCESName("ces1"))
+	cesN, _ = mapping.getCESName(NewCEPName("cep5", "ns"))
+	assert.Equal(t, cesN, NewCESName("ces2"))
+	cesN, _ = mapping.getCESName(NewCEPName("cep6", "ns"))
+	assert.Equal(t, cesN, NewCESName("ces2"))
+	cesN, _ = mapping.getCESName(NewCEPName("cep7", "ns"))
+	assert.Equal(t, cesN, NewCESName("ces2"))
```
pippolo84, line 116: Also, before stopping the hive, you need to call `cesController.queue.ShutDown()` to stop the internal client-go goroutine managing the delaying queue. That is because you are manually managing (outside of the hive lifecycle) the CES controller (and so controller `Stop` is not called).
After doing that, you can get rid of the goleak.IgnoreCurrent, since there will be no other goroutine leak (ultimately making the test better in finding possible leaks).

operator/pkg/ciliumendpointslice/endpointslice_test.go
```diff
@@ -54,240 +52,131 @@ func createStoreEndpointSlice(name string, namespace string, endpoints []capi_v2
...
+	assert.Equal(t, 0, cesController.queue.Len())
+	assert.Equal(t, 4, len(cesController.preInitEnqueuedEndpointsEvents))
+	assert.Equal(t, false, cesController.endpointsMappingInitialized)
+
+	cesController.syncCESsInLocalCache(context.Background())
+	assert.Equal(t, true, cesController.endpointsMappingInitialized)
+
+	cesController.processEnqueuedPreInitEndpoints()
+	time.Sleep(DefaultCESSyncTime * 2) // elements are added to the queue with a default delay
+	assert.Equal(t, 2, cesController.queue.Len())
```
pippolo84: Same here, we need to call `cesController.queue.ShutDown()`before stopping the hive.

operator/pkg/ciliumendpointslice/utils.go
```diff
@@ -0,0 +1,31 @@
...
+// of a CES name is similar to pod k8s naming convention "ces-123456789-abcde".
+// First 3 letters indicates ces resource, followed by random letters.
+func uniqueCESliceName(mapping *CESToCEPMapping) string {
+	var cesName string
+	for {
+		cesName = fmt.Sprintf("%s-%s-%s", cesNamePrefix, randomName(9), randomName(5))
+		if !mapping.hasCESName(NewCESName(cesName)) {
+			return cesName
+		}
+	}
```
pippolo84: Not sure if this truly qualifies as a "hot path", but I suggest to avoid the usage of `fmt.Sprintf` in favor of `strings.Builder` (it is also possible to preallocate the underlying buffer with the right size):
```
	var sb strings.Builder
	for {
		rn1, rn2 := randomName(9), randomName(5)

		sb.Reset()
		sb.Grow(len(cesNamePrefix) + 1 + len(rn1) + 1 + len(rn2))
		sb.WriteString(cesNamePrefix)
		sb.WriteRune('-')
		sb.WriteString(rn1)
		sb.WriteRune('-')
		sb.WriteString(rn2)
		cesName := sb.String()

		if !mapping.hasCESName(NewCESName(cesName)) {
			return cesName
		}
	}
```

See https://github.com/cilium/cilium/issues/19571

operator/pkg/ciliumendpointslice/manager.go
```diff
@@ -211,394 +97,99 @@ func uniqueCESliceName(desiredCESs *CESToCEPMapping) string {
...
+			logfields.CEPName: cepName.string(),
+			logfields.CESName: cesName.string(),
+		}).Debug("Removing CEP from CES")
+		c.mapping.deleteCEP(cepName)
+		if c.mapping.countCEPsInCES(cesName) == 0 {
+			c.mutex.Lock()
+			defer c.mutex.Unlock()
+			if c.mapping.countCEPsInCES(cesName) == 0 {
+				c.mapping.deleteCES(cesName)
+			}
```
pippolo84: I suggest to add a brief comment to explain that we're doing the first check with the read lock to avoid the performance penalty of acquiring the write lock every time. The double check might be confusing for a casual reader.

operator/pkg/ciliumendpointslice/manager.go
```diff
@@ -607,133 +198,125 @@ func (c *cesManagerIdentity) InsertCEPInCache(cep *cilium_v2.CoreCiliumEndpoint,
...
-	if ces.cesInsertedAt.IsZero() {
-		ces.cesInsertedAt = time.Now()
+func (c *cesManagerIdentity) removeCESToIdentity(id int64, cesName CESName) {
+	cesSlice := c.identityToCES[id]
+	for i, ces := range cesSlice {
+		if ces == cesName {
+			cesSlice[i] = cesSlice[len(cesSlice)-1]
+		}
 	}
+	c.identityToCES[id] = cesSlice
```
pippolo84: Shouldn't we keep track of how many elements we've overwritten in cesSlice at line 291 and then assign here `cesSlice[:len(cesSlice)-removed]`? Otherwise the slices in the `c.identityToCES` won't never shrink.

pippolo84 (changes requested):
I've completed a second round of review.
Overall LGTM, my suggestions are mainly related to the tests. Thanks again!

operator/pkg/ciliumendpointslice/endpointslice.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/reconciler.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/endpointslice_test.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/endpointslice_test.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/reconciler_test.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/reconciler_test.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/reconciler_test.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/reconciler_test.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/controller.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/endpointslice.go
alan-kut: Added

operator/pkg/ciliumendpointslice/endpointslice_test.go
alan-kut: Fixed stopping the queue.

Regarding IgnoreCurrent - Shouldn't we always use it?
Otherwise it's easy to add a test for A (e.g. some CES controller test leaking) and change some other part B (e.g. Identity GC) that does have such test and the test for B would start failing even though the problem is totally outside of B.

operator/pkg/ciliumendpointslice/endpointslice_test.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/utils.go
alan-kut: I just moved this code but there is no harm in changing it now do string builder.

operator/pkg/ciliumendpointslice/manager.go
alan-kut: Added, you are right, it may  be confusing

operator/pkg/ciliumendpointslice/manager.go
alan-kut: Fixed. Good catch.

operator/pkg/ciliumendpointslice/endpointslice_test.go
pippolo84: That's true but, if we keep that, it's easier to add a new test that might leak a goroutine, and let the leak go unnoticed. I find this to be even worse, although you're right that having a failure in another test is not ideal.
The hive based tests are like integration tests for multiple cells even if they are written as unit. I suggest to first add goleak to each hive based test (like the ones in `endpointslice_test.go` and in `reconciler_test.go`) and then check that we are not leaking anything (except for anything that is due to a dependency like client-go). If so, I still suggest to avoid `IgnoreCurrent` for now.

alan-kut: Also - those test would fail if some other test fail and doesn't clean up everything.
```
hive.Start()

err := checkSth()
if err != nil {
  t.Fatalf("failed")
}

hive.Stop()
```

Removed IgnoreCurrent for now.

It would be good to have some kind of a leak test at the end if all other test passed, maybe?

alan-kut:
I think I responded to all the comments. If some test/check fails I'll fix it on Monday

operator/pkg/ciliumendpointslice/endpointslice_test.go
pippolo84: > Also - those test would fail if some other test fail and doesn't clean up everything.

That's true but that is less harmful, since it'll be clear that the root cause is the first test failure.

> It would be good to have some kind of a leak test at the end if all other test passed, maybe?

I think that right now is not necessary, since we can check every single hive-based test with goleak, excluding what we already know is leaking (and out of our control), like `k8s.io/client-go/util/workqueue.(*Type).updateUnfinishedWorkLoop`.

alan-kut:
Thanks, fixed it.

I read that start context should be used for all potentially long running operations during start but I thought that it would harm to use it later after I got your comment.

> This came up after a round of manual testing in a kind cluster
Did you test this PR specifically or are there some kind of other tests not shown here?

I have prepared a manual test for identities so I would've caught it as well before merging.
I'll run my test as well once we are close to accepting the PR.

pippolo84:
> Did you test this PR specifically or are there some kind of other tests not shown here?

Yep, I've played a bit with this branch.

> I have prepared a manual test for identities so I would've caught it as well before merging.
> I'll run my test as well once we are close to accepting the PR.

Sounds good, thanks.

operator/pkg/ciliumendpointslice/controller_test.go
```diff
@@ -0,0 +1,135 @@
...
+		}),
+	)
+	if err := hive.Start(context.Background()); err != nil {
+		t.Fatalf("failed to start: %s", err)
+	}
+	cesCreated, err := createCEPandVerifyCESCreated(fakeClient, ciliumEndpoint, ciliumEndpointSlice)
+	if err != nil {
+		t.Fatalf("Couldn't verify if CES is created: %s", err)
+	}
+	// Verify CES is created when CES features is disabled
```
pippolo84: Typo: `disabled` -> `enabled`

operator/pkg/ciliumendpointslice/endpointslice.go
```diff
@@ -48,263 +48,244 @@ const (
...
-		logfields.CESSliceMode: c.slicingMode,
-	}).Info("Starting CES controller reconciler.")
+func (c *Controller) enqueueCESReconciliation(cess []CESName) {
+	for _, ces := range cess {
+		c.logger.WithFields(logrus.Fields{
+			logfields.CESName: ces.string(),
+		}).Debug("Enqueueing CES (if not empty name)")
+		if ces.Name != "" {
+			if c.enqueuedAt[ces].IsZero() {
+				c.enqueuedAt[ces] = time.Now()
```
pippolo84, line 109: I think `enqueuedAt` needs to be protected from concurrent writes. It is written here (from the workerpool `cilium-endpoint-slices-updater` goroutine) and from the `worker` goroutine in `getAndResetCESProcessingDelay`.

rolinh (changes requested):
I just did a review of the vendor changes. Please, see suggestion below.

operator/pkg/ciliumendpointslice/controller_test.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func createCEPandVerifyCESCreated(fakeClient k8sClient.FakeClientset, ciliumEndpoint resource.Resource[*cilium_v2.CiliumEndpoint], ciliumEndpointSlice resource.Resource[*cilium_v2a1.CiliumEndpointSlice]) (bool, error) {
+	cep := createStoreEndpoint("cep1", "ns", 1)
+	fakeClient.CiliumV2().CiliumEndpoints("ns").Create(context.Background(), cep, meta_v1.CreateOptions{})
+	cepStore, _ := ciliumEndpoint.Store(context.Background())
+	if err := testutils.WaitUntil(func() bool {
+		return len(cepStore.List()) > 0
+	}, time.Second); err != nil {
+		return false, errors.Errorf("failed to get CEP: %s", err)
```
rolinh: I don't see a good argument for having `github.com/pkg/errors` as a new direct dependency. 
```suggestion
		return false, fmt.Errorf("failed to get CEP: %s", err)
```

rolinh (changes requested):

operator/pkg/ciliumendpointslice/manager.go
```diff
@@ -23,75 +18,31 @@ var (
...
 	// as well as CEPName to CESName.
-	desiredCESs *CESToCEPMapping
+	mapping *CESToCEPMapping
 
-	// workqueue is used to sync CESs with the api-server. this will rate-limit the
-	// CES requests going to api-server, ensures a single CES will not be proccessed
-	// multiple times concurrently, and if CES is added multiple times before it
-	// can be processed, this will only be processed only once.
-	queue workqueue.RateLimitingInterface
+	mutex lock.Mutex
```
pippolo84: This lock can be avoided since `mapping` is assigned once at the manager startup and then every ops on it is protected by its internal lock. Same for `maxCEPsInCES`, it is assigned at startup and then only read.

operator/pkg/ciliumendpointslice/controller.go
```diff
@@ -0,0 +1,100 @@
...
+	// multiple times concurrently, and if CES is added multiple times before it
+	// can be processed, this will only be processed only once.
+	queue           workqueue.RateLimitingInterface
+	queueTerminated chan struct{}
+	writeQPSLimit   float64
+	writeQPSBurst   int
+
+	enqueuedAt                     map[CESName]time.Time
+	preInitEnqueuedEndpointsEvents []endpointEvent
+	endpointsMappingInitialized    bool
```
pippolo84: I think the access to `endpointsMappingInitialized` needs to be synchronized (for example using an `atomic.Bool`). That's because it is written by the goroutine that executes `syncCESsInLocalCache` and it is read by the goroutine that executes `onEndpointUpdate` and `onEndpointDelete` (that is, the goroutine `cilium-endpoints-updater` from the workerpool).
IOW, the second goroutine is doing a busy wait on that boolean, and doing that without synchronization is incorrect, as reported [here](https://go.dev/ref/mem) (look for the relevant example regarding "busy waiting").

pippolo84 (changes requested):
Thanks for addressing all the comments! I've focused this last round of review on the usage of locks. I think there is something that needs to be addressed.
Apart from that, I think now the PR is starting to looking great! :+1:

operator/pkg/ciliumendpointslice/controller_test.go
alan-kut: Fixed, thanks

operator/pkg/ciliumendpointslice/controller_test.go
alan-kut: Fixed

operator/pkg/ciliumendpointslice/manager.go
alan-kut: It is used in the following cases:

Get largest not full CES and create new one if there is none. Without lock there may be too many CEPs in CES or too many CESs (all gets the same CES and fill it above capacity or all create new one when one would be enough)
```go
cesName = c.getLargestAvailableCESForNamespace(ns)
if cesName.Name == "" {
	cesName = c.createCES("", ns)
}
c.mapping.insertCEP(cepName, cesName)
```

Check if CES can be deleted. Without this check new CEP may be added to the CES and CES then deleted.
```go
c.mutex.Lock()
defer c.mutex.Unlock()
if c.mapping.countCEPsInCES(cesName) == 0 {
	c.mapping.deleteCES(cesName)
}
```

The same, a little bit more complicated, for identity.

I think I could rewrite this code without this lock. The mapping could do the checks and then return error. But it means the code would need to be changed to the loop.

operator/pkg/ciliumendpointslice/controller.go
alan-kut: I don't think it's busy waiting.

```go
func (c *Controller) onEndpointEvent(cep *cilium_api_v2.CiliumEndpoint) {
  if c.endpointsMappingInitialized {
    process
  } else {
    add to queue
}
```
Nothing else read it. It's changed to true later.

I agree we need synchronization to avoid situation in which we miss an event - check for c.endpointsMappingInitialized return false but the queue is started being processed before new element is added.

I think we need synchronization on the preInitEnqueuedEndpointsEvents even more. If two threads change it at the same time (it's not possible right now I think as we have 1 thread processing the events) then we will miss some events.

It's too bad we will have quite complicated code here just for short period of time.
I'll try to remove it all together. I think we can start processing the events only after the initial mapping is initialized.

alan-kut: Done it. It is quite simple with the hive as we can start processing event when we want to.

operator/pkg/ciliumendpointslice/manager.go
pippolo84: > It is used in the following cases:
> 
> Get largest not full CES and create new one if there is none. Without lock there may be too many CEPs in CES or too many CESs (all gets the same CES and fill it above capacity or all create new one when one would be enough)
> 

But the only operations that can insert or remove a CEP are:
- `UpdateCEPMapping` and `RemoveCEPMapping`, but these are serialized by the channel returned by `Events` in `runCiliumEndpointsUpdater` (IOW it's not possible to handle multiple CiliumEndpoint related events at the same time)
- `initializeMappingCEPtoCES` that is called only before initial sync is completed

So, it should not be possible to have multiple goroutines concurrently reading the CEPs in CES counter and trying to  insert or remove a CEP in a CES. Or am I missing something?

operator/pkg/ciliumendpointslice/endpointslice.go
alan-kut:
```go
func enqueueCES(ces CESName) {
  if c.enqueuedAt[ces].IsZero() {
    c.enqueuedAt[ces] = time.Now()
  }
}

func getAndReset(ces CESName) {
  enqueued, exists := c.enqueuedAt[ces]
  if exists && !enqueued.IsZero() {
    c.enqueuedAt[ces] = time.Time{}
    return time.Since(enqueued).Seconds()
  }
}
```

So when we enqueue a CES and there is no timestamp we set it to current time. When we process the CES and there is a timestamp we reset it and report time.

The conflict and race may happen if we try to enqueue and process a CES in the same time but the result doesn't matter.

Case I: enqueueAt[ces] was not zero before the race.
In this case the result reported from getAndReset will be correct delay.
At the end enqueueAt[ces] may be zero (if getAndReset finished later than enqueue) or time.Now() otherwise. The same can happen if either enqueue happens fully before or after getAndReset so synchronization doesn't change anything.

Case II: enqueueAt[ces] was zero.
In this case race may happen that enqueue first sees it's zero, changes it to time.now() and then getAndReset sees it's not zero, reports the delay and reset it. The same can happen if enqueueAt happens fully before getAndReset.

I don't see a case in which the result without synchronization cannot be achieved with synchronization. I'd rather not add a lock here just to protect queue delay metric.

That said this may be wrong. The code should always show the time really spent in the queue. I'm not sure how to fix it. It would be great to add new field to the queue. However this would break the behavior of not processing the same items multiple times as all entries would be unique now.
We can't also just prepare a queue of enqueued at as not all add will add element to the queue.

I will keep thinking about it.

operator/pkg/ciliumendpointslice/controller.go
pippolo84: > I'll try to remove it all together. I think we can start processing the events only after the initial mapping is initialized.

Yep, this sounds better from a readability perspective too. :+1:

operator/pkg/ciliumendpointslice/manager.go
alan-kut: Yes. As long as we are sure we cannot process multiple events from the CEP at the same time we are good here. I'm removing those locks for now.

I thought that the CES controller worker also changes mapping but it's not the case, it's:

CEP event => update mapping => enqueue CES => worker processing CESs from the queue
CES event => enqueue CES => the same worker as above processing CESs from the queu

I'll remove the locks.

alan-kut: Removed

rolinh (approved):
No more vendor changes so the PR lgtm in that respect. Thanks for your contribution!

operator/pkg/ciliumendpointslice/endpointslice.go
pippolo84: > I don't see a case in which the result without synchronization cannot be achieved with synchronization. I'd rather not add a lock here just to protect queue delay metric.

The write operations themselves (`c.enqueuedAt[ces] = time.Now()` and `c.enqueuedAt[ces] = time.Time{}`) are not atomic. So we can end up with a value in `c.enqueuedAt[ces]` that is neither `time.Now()` nor `time.Time{}`.
That's because `time.Time` is defined as a struct:

```
type Time struct {
	wall uint64
	ext  int64

	loc *Location
}
```

And if that happens, `getAndReset` might return a totally bogus time interval when doing `time.Since(enqueued).Seconds()` the next time is called for the same CES.

alan-kut: Added a lock

operator/pkg/ciliumendpointslice/endpointslice.go
```diff
@@ -48,263 +41,223 @@ const (
...
+		}
+	}
+}
+
+func (c *Controller) getAndResetCESProcessingDelay(ces CESName) float64 {
+	enqueued, exists := c.enqueuedAt[ces]
+	if !exists {
+		return 0
+	}
+	c.enqueuedAtLock.Lock()
```
pippolo84: I think we should take the lock before checking the key existence, too. That's because it is possible to have a concurrent write to the `c.enqueuedAt` map at line 109 and that can potentially reallocate the map internal buckets (e.g: think about when the map grows) giving us incorrect results.

alan-kut: Done

pippolo84 (approved):
Awesome, thanks! :rocket:

alan-kut:
Can we merge it now?

I see two tests are failing now but I don't see how this change can cause it especially when all the tests were passing just moments ago before small review changes.

pippolo84:
ci-ginkgo hit a known [flake](https://github.com/cilium/cilium/issues/13071#issuecomment-1711671271), rerunning.

pippolo84:
> Can we merge it now?
> 
> I see two tests are failing now but I don't see how this change can cause it especially when all the tests were passing just moments ago before small review changes.

Let's wait for the other assigned reviewers to give their feedback. Once the PR has all the approvals and all the tests marked as "Required" have been passed, a repository owner will merge it.
No need to change anything in the meantime.

asauber (approved):
Config-related changes LGTM. 

A great reference for Hive considerations as well.

joamaki (approved):

PR #26841, opened 2023-07-14
init.sh: move ipip device creation to Go

### LINKED ISSUE #20740: init.sh: move ip-in-ip device creation to separate Go package
This could land in a new host(lb)-related package. This needs to integrate with the socket-lb (host-lb): https://github.com/cilium/cilium/issues/20739.

rgo3:
vendor: update vishvananda/netlink:
```
This pulls in the following recent changes:

- vishvananda/netlink@88c0de
- vishvananda/netlink@16d31d

These are needed to move ipip device creation from init.sh to Go.

Signed-off-by: Robin Gögge <r.goegge@isovalent.com>
```
init.sh: move ipip device creation to Go:
```
This refactors the creation of the cilium_ipip{4,6} devices to a
Go based implementation and adds a simple device creation test.

One behavioural change was made: no longer try to repurpose
tunl0 by renaming it and enabling collect_md, as that dance is
rather tricky and fragile due to the kernel module needing to be
loaded first.

With this change, we still prefix the fallback devices with cilium_,
meaning cilium_tunl0 for v4 and cilium_ip6tnl0 for v6, but we no
longer try to use or reconfigure these devices. cilium_ipip4/6 are
always created with the correct parameters from the start, reducing
the amount of moving parts.

Signed-off-by: Robin Gögge <r.goegge@isovalent.com>
```
Fixes: #20740

### DIFF
--- a/pkg/datapath/loader/base.go
+++ b/pkg/datapath/loader/base.go
@@ -319,9 +319,6 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner, 
 		encapProto = option.Config.TunnelProtocol
 	case option.Config.EnableHealthDatapath:
 		mode = option.DSRDispatchIPIP
-		sysSettings = append(sysSettings,
-			sysctl.Setting{Name: "net.core.fb_tunnels_only_for_init_net",
-				Val: "2", IgnoreErr: true})
 	default:
 		mode = directMode
 	}
@@ -351,6 +348,18 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner, 
 	}
 	args[initArgHostDev1] = hostDev1.Attrs().Name
 	args[initArgHostDev2] = hostDev2.Attrs().Name
+
+	if option.Config.EnableHealthDatapath {
+		sysSettings = append(
+			sysSettings,
+			sysctl.Setting{
+				Name: "net.core.fb_tunnels_only_for_init_net", Val: "2", IgnoreErr: true,
+			},
+		)
+		if err := setupIPIPDevices(option.Config.IPv4Enabled(), option.Config.IPv6Enabled()); err != nil {
+			return fmt.Errorf("unable to create ipip encapsulation devices for health datapath")
+		}
+	}
 
 	if option.Config.IPAM == ipamOption.IPAMENI {
 		var err error
--- a/pkg/datapath/loader/netlink.go
+++ b/pkg/datapath/loader/netlink.go
@@ -226,7 +226,9 @@ func RemoveTCFilters(ifName string, tcDir uint32) error {
 	return nil
 }
 
-func setupDev(link netlink.Link) error {
+// configureDevice puts the given link into the up state and applies
+// link-specific sysctls.
+func configureDevice(link netlink.Link) error {
 	ifName := link.Attrs().Name
 
 	if err := netlink.LinkSetUp(link); err != nil {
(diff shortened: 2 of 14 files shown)


pkg/datapath/loader/base.go
```diff
@@ -317,11 +317,6 @@ func (l *Loader) Reinitialize(ctx context.Context, o datapath.BaseProgramOwner,
 	case option.Config.TunnelingEnabled():
 		mode = tunnelMode
 		encapProto = option.Config.TunnelProtocol
-	case option.Config.EnableHealthDatapath:
-		mode = option.DSRDispatchIPIP
```
ti-mo, line 321: This statement has gone missing in this PR. I'll push my local branch with this readded in a bit.

rgo3: I figured it's no longer needed given it sets a variable that's used as the input arg for init.sh with a value that's no longer used in init.sh after this PR.

ti-mo: Ah! I figured as much, thanks for the context. I'll leave it in, and we'll remove all of them when init.sh is dead. :+1:

pkg/datapath/loader/netlink.go
```diff
@@ -444,3 +446,127 @@ func addHostDeviceAddr(hostDev netlink.Link, ipv4, ipv6 net.IP) error {
...
+//
+// In versions of Cilium prior to 1.15, the behaviour was as follows:
+//   - Repurpose the default tunl0 by setting it into collect_md mode and renaming
+//     it to cilium_ipip4. Use the interface for production traffic.
+//   - The same cannot be done for ip6tunl0, as collect_md cannot be enabled on
+//     this interface. Leave it unused.
+//   - Rename sit0 to cilium_sit, if present. This was potentially a mistake,
+//     as the sit module is not involved with ip6tnl interfaces.
+//
+// As of Cilium 1.15, if present, tunl0 is renamed to cilium_tunl and ip6tnl0 is
```
lmb, line 468: This could use a rationale why we're doing this, and at which point we can stop doing this (if ever).

pkg/datapath/loader/netlink.go
```diff
@@ -444,3 +446,127 @@ func addHostDeviceAddr(hostDev netlink.Link, ipv4, ipv6 net.IP) error {
...
+		if err := setupDevice(&netlink.Iptun{
+			LinkAttrs: netlink.LinkAttrs{Name: defaults.CiliumIPIPv4Device},
+			FlowBased: true,
+		}); err != nil {
+			return fmt.Errorf("creating %s: %w", defaults.CiliumIPIPv4Device, err)
+		}
+
+		// Rename fallback device created by potential kernel module load after
+		// creating tunnel interface.
+		if err := renameDevice("tunl0", "cilium_tunl"); err != nil {
```
lmb, line 489: Meta: we should stop relying on device names so much and instead track indices. That saves us from renames breaking stuff.

lmb (approved):

aanm (approved):

pkg/datapath/loader/netlink.go
ti-mo: Good point, added.

ti-mo:
Sorry @nebril please ignore. I cherry-picked https://github.com/cilium/cilium/pull/27339 to see if it contains the kernel module needed for the tests added here. Converted to draft in the meantime.

ti-mo:
Closing this in favor of https://github.com/cilium/cilium/pull/27082 since it makes more sense to track the changes together.

Issue #23353, opened 2023-01-25
Cilium agent opens ports in all interfaces (maybe unnecessarily?)


cortopy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

I've noticed that cilium agent opens several ports on all interfaces, while some ports only listen on localhost.

```
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 127.0.0.1:9879          0.0.0.0:*               LISTEN      1345831/cilium-agen 
tcp        0      0 127.0.0.1:9890          0.0.0.0:*               LISTEN      1345831/cilium-agen 
tcp6       0      0 :::4244                 :::*                    LISTEN      1345831/cilium-agen 
tcp6       0      0 :::4240                 :::*                    LISTEN      1345831/cilium-agen 
tcp6       0      0 :::36195                :::*                    LISTEN      1345831/cilium-agen 
tcp6       0      0 :::9962                 :::*                    LISTEN      1345831/cilium-agen 
udp6       0      0 :::36195                :::*                                1345831/cilium-agen
````

I'm trying to secure a cluster where all traffic between nodes goes through a private network, but then Cilium agent is listening on all interfaces. It seems to me that this makes security more difficult to implement. I wasn't able to find documentation about what those ports do, but I would expect them to listen on the interface of the node's InternalIP.

I found that a similar question was asked in [Slack](https://cilium.slack.com/archives/C1MATJ5U5/p1670496809639599), and it was suggested that a new issue should be opened if using a recent version of cilium.

### Cilium Version

cilium-cli: v0.12.11 compiled with go1.19.3 on linux/amd64
cilium image (default): v1.12.2
cilium image (stable): v1.12.5
cilium image (running): v1.12.3

### Kernel Version

Linux node1 5.15.0-52-generic #58-Ubuntu SMP Thu Oct 13 08:03:55 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

### Kubernetes Version

1.23

### Anything else?

I've tried using firewall, but setting up an external firewall with cilium is breaking connectivity. It feels like this solution is bound to create many issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ydp:
I have a similar question, does cilium support configure interface or node CIDR to use for pod to pod traffic? Like what Antrea did in this issue https://github.com/antrea-io/antrea/pull/2704

[...]

PR #26845, opened 2023-07-14
health: Update Cilium agent to listen on nodeip

tamilmani1989:

      GitHub issue.

This PR updates agent health server to listen on nodeip instead of all interfaces.
Fixes:
[#23353](https://github.com/cilium/cilium/issues/23353)

```release-note
```

### DIFF
--- a/cilium-health/responder/main.go
+++ b/cilium-health/responder/main.go
@@ -31,7 +31,7 @@ func main() {
 	// Shutdown gracefully to halt server and remove pidfile
 	ctx, cancel := signal.NotifyContext(context.Background(), unix.SIGINT, unix.SIGHUP, unix.SIGTERM, unix.SIGQUIT)
 
-	srv := responder.NewServer(listen)
+	srv := responder.NewServers([]string{""}, listen)
 	defer srv.Shutdown()
 	go func() {
 		if err := srv.Serve(); !errors.Is(err, http.ErrServerClosed) {
--- a/pkg/health/probe/responder/responder.go
+++ b/pkg/health/probe/responder/responder.go
@@ -7,7 +7,9 @@ package responder
 // as this package typically runs in its own process
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"time"
 )
@@ -17,29 +19,53 @@ var defaultTimeout = 30 * time.Second
 
 // Server wraps a minimal http server for the /hello endpoint
 type Server struct {
-	httpServer http.Server
+	httpServers []*http.Server
 }
 
 // NewServer creates a new server listening on the given port
-func NewServer(port int) *Server {
-	return &Server{
-		http.Server{
-			Addr:    fmt.Sprintf(":%d", port),
-			Handler: http.HandlerFunc(serverRequests),
-		},
+func NewServers(address []string, port int) *Server {
+	if len(address) == 0 {
+		address = []string{""}
 	}
+
+	server := &Server{}
+	for _, ip := range address {
+		addr := net.JoinHostPort(ip, fmt.Sprintf("%v", port))
+		hs := http.Server{
+			Addr:    addr,
+			Handler: http.HandlerFunc(serverRequests),
+		}
+		server.httpServers = append(server.httpServers, &hs)
+	}
+
+	return server
 }
 
 // Serve http requests until shut down
 func (s *Server) Serve() error {
-	return s.httpServer.ListenAndServe()
+	errors := make(chan error)
+	for _, hs := range s.httpServers {
+		tmpHttpServer := hs
+		go func() {
+			errors <- tmpHttpServer.ListenAndServe()
+		}()
+	}
+
+	// Block for the first error, then return.
+	err := <-errors
+	return err
 }
 
 // Shutdown server gracefully
 func (s *Server) Shutdown() error {
 	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
 	defer cancel()
-	return s.httpServer.Shutdown(ctx)
+	errs := make([]error, 0, len(s.httpServers))
+	for _, hs := range s.httpServers {
+		errs = append(errs, hs.Shutdown(ctx))
+	}
+
+	return errors.Join(errs...)
 }
 
 func serverRequests(w http.ResponseWriter, r *http.Request) {
--- a/pkg/health/probe/responder/responder_test.go
+++ b/pkg/health/probe/responder/responder_test.go
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package responder
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewServersInitialization(t *testing.T) {
+	tests := []struct {
+		name                   string
+		address                []string
+		expectedServerCount    int
+		exptectedServerAddress []string
+	}{
+		{
+			name:                   "Initialize http server listening on all ports",
+			address:                []string{""},
+			expectedServerCount:    1,
+			exptectedServerAddress: []string{":4240"},
+		},
+		{
+			name:                   "Initialize http server listening on ipv4 address",
+			address:                []string{"192.168.1.4"},
+			expectedServerCount:    1,
+			exptectedServerAddress: []string{"192.168.1.4:4240"},
+		},
+		{
+			name:                   "Initialize http server listening on ipv4 and ipv6 address",
+			address:                []string{"192.168.1.4", "fc00:c111::2"},
+			expectedServerCount:    2,
+			exptectedServerAddress: []string{"192.168.1.4:4240", "[fc00:c111::2]:4240"},
+		},
+		{
+			name:                   "Initialize http server with nil address",
+			address:                []string{},
+			expectedServerCount:    1,
+			exptectedServerAddress: []string{":4240"},
+		},
+	}
+
+	for _, tt := range tests {
+		s := NewServers(tt.address, 4240)
+		assert.NotNil(t, s)
+		assert.Equal(t, len(s.httpServers), tt.expectedServerCount, "Number of listen address doesn't match")
+		for i, s := range s.httpServers {
+			assert.Equal(t, tt.exptectedServerAddress[i], s.Addr)
+		}
+	}
+}
--- a/pkg/health/server/server.go
+++ b/pkg/health/server/server.go
@@ -23,6 +23,8 @@ import (
 	"github.com/cilium/cilium/pkg/logging"
 	"github.com/cilium/cilium/pkg/logging/logfields"
 	"github.com/cilium/cilium/pkg/metrics"
+	"github.com/cilium/cilium/pkg/node"
+	"github.com/cilium/cilium/pkg/option"
 )
 
 var (
@@ -417,7 +419,38 @@ func NewServer(config Config) (*Server, error) {
 	server.Client = cl
 	server.Server = *server.newServer(config.HealthAPISpec)
 
-	server.httpPathServer = responder.NewServer(config.HTTPPathPort)
+	server.httpPathServer = responder.NewServers(getAddresses(), config.HTTPPathPort)
 
 	return server, nil
+}
+
+// Get internal node ipv4/ipv6 addresses based on config enabled.
+// If it fails to get either of internal node address, it returns "0.0.0.0" if ipv4 or "::" if ipv6.
+func getAddresses() []string {
+	addresses := make([]string, 0, 2)
+
+	// listen on all interfaces and all families in case of external-workloads
+	if option.Config.JoinCluster {
+		return []string{""}
+	}
+
+	if option.Config.EnableIPv4 {
+		if ipv4 := node.GetInternalIPv4(); ipv4 != nil {
+			addresses = append(addresses, ipv4.String())
+		} else {
+			// if Get ipv4 fails, then listen on all ipv4 addr.
+			addresses = append(addresses, "0.0.0.0")
+		}
+	}
+
+	if option.Config.EnableIPv6 {
+		if ipv6 := node.GetInternalIPv6(); ipv6 != nil {
+			addresses = append(addresses, ipv6.String())
+		} else {
+			// if Get ipv6 fails, then listen on all ipv6 addr.
+			addresses = append(addresses, "::")
+		}
+	}
+
+	return addresses
 }
--- a/pkg/node/types/node.go
+++ b/pkg/node/types/node.go
@@ -351,6 +351,34 @@ func (n *Node) GetK8sNodeIP() net.IP {
 	return externalIP
 }
 
+// GetNodeInternalIP returns the Internal IPv4 of node or nil.
+func (n *Node) GetNodeInternalIPv4() net.IP {
+	for _, addr := range n.IPAddresses {
+		if addr.IP.To4() == nil {
+			continue
+		}
+		if addr.Type == addressing.NodeInternalIP {
+			return addr.IP
+		}
+	}
+
+	return nil
+}
+
+// GetNodeInternalIP returns the Internal IPv6 of node or nil.
+func (n *Node) GetNodeInternalIPv6() net.IP {
+	for _, addr := range n.IPAddresses {
+		if addr.IP.To4() != nil {
+			continue
+		}
+		if addr.Type == addressing.NodeInternalIP {
+			return addr.IP
+		}
+	}
+
+	return nil
+}
+
 // GetCiliumInternalIP returns the CiliumInternalIP e.g. the IP associated
 // with cilium_host on the node.
 func (n *Node) GetCiliumInternalIP(ipv6 bool) net.IP {
--- a/pkg/node/address.go
+++ b/pkg/node/address.go
@@ -283,6 +283,18 @@ func GetIPv4() net.IP {
 	return clone(n.GetNodeIP(false))
 }
 
+// GetInternalIPv4 returns node internal ipv4 address else return nil.
+func GetInternalIPv4() net.IP {
+	n := getLocalNode()
+	return clone(n.GetNodeInternalIPv4())
+}
+
+// GetInternalIPv6 returns node internal ipv6 address else return nil.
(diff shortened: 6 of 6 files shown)


cilium-health/responder/main.go
```diff
@@ -31,7 +32,7 @@ func main() {
 	// Shutdown gracefully to halt server and remove pidfile
 	ctx, cancel := signal.NotifyContext(context.Background(), unix.SIGINT, unix.SIGHUP, unix.SIGTERM, unix.SIGQUIT)
 
-	srv := responder.NewServer(listen)
+	srv := responder.NewServer(node.GetCiliumEndpointNodeIP(), listen)
```
pippolo84: This is not going to work, because cilium-health-responder is a separate binary from the agent and here we don't start the `node` machinery that allows us to get the node IP:

```
# cilium-health-responder   
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x11c8687]

goroutine 1 [running]:
github.com/cilium/cilium/pkg/node.(*LocalNodeStore).Observe(0x16cf988?, {0x16cf950?, 0xc000426000?}, 0x13845a0?, 0x40d201?)
	<autogenerated>:1 +0x27
github.com/cilium/cilium/pkg/stream.First[...]({_, _}, {_, _})
	/go/src/github.com/cilium/cilium/pkg/stream/sinks.go:25 +0x1a3
github.com/cilium/cilium/pkg/node.(*LocalNodeStore).Get(...)
	/go/src/github.com/cilium/cilium/pkg/node/local_node_store.go:116
github.com/cilium/cilium/pkg/node.getLocalNode()
	/go/src/github.com/cilium/cilium/pkg/node/address.go:49 +0xad
github.com/cilium/cilium/pkg/node.GetIPv4()
	/go/src/github.com/cilium/cilium/pkg/node/address.go:282 +0x2e
github.com/cilium/cilium/pkg/node.GetCiliumEndpointNodeIP()
	/go/src/github.com/cilium/cilium/pkg/node/address.go:290 +0x29
main.main()
	/go/src/github.com/cilium/cilium/cilium-health/responder/main.go:35 +0x19a
```

Being a binary meant for testing purposes only, I think we can simply let it bind to all interfaces using an empty string as address.

pippolo84 (changes requested):
Thanks for your PR @tamilmani1989 . Changes look good on the agent side, but I think we should leave the cilium-health-responder behavior as is.

cilium-health/responder/main.go
tamilmani1989: my bad.. didnt realise `cilium-health-responder` separate binary.. let me fix this. Thanks for pointing out

pippolo84 (commented):
Changes look good, thanks!

One last thing: could you please change the commit message to adhere to the guide [here](https://docs.cilium.io/en/latest/contributing/development/contributing_guide/#submitting-a-pull-request)? Maybe a brief description of the solved issue and a reference to it in a `Fixes: ...` line. For the title, I suggest something like `health: Update daemon to listen on node ip`.
Thanks! :pray:

pippolo84:
Hi @tamilmani1989 , what I meant was to change the commit message like this:

```
health: Update daemon to listen on node ip

Update agenthealth to listen on internal node ip instead of listening on
all interfaces.

Fixes: #23353
    
Signed-off-by: Tamilmani <tamanoha@microsoft.com>
```

I was not referring to the PR title, but to the first line of the commit message. Sorry for the confusion.
Can you please update the commit message one last time so I can approve the PR? Thanks in advance! :pray:

pippolo84 (approved):
Nice! :rocket:

pippolo84:
@tamilmani1989 CI failures are unrelated to the PR. It has already been fixed in main, could you please rebase?

tamilmani1989:
@pippolo84 anything else required for this PR?

pippolo84:
CI failures seem legit, are you up to investigate them? I think it should be the same root cause for all of those.

tamilmani1989:
I see requests to <ipv6>:port. since cilium agent is listening only on ipv4+port its failing:

```
18:26:59 STEP: Cilium is not ready yet: connectivity health is failing: cilium-agent 'cilium-7wwml': connectivity to path 'kind-control-plane.host.secondary-addresses[*].http.status' is unhealthy: 'Get "http://[fc00:c111::2]:4240/hello": dial tcp [fc00:c111::2]:4240: connect: connection refused'
```

should i update code to listen both on ipv4 and ipv6 on port 4240. @pippolo84 what do you suggest?

pippolo84:
> should i update code to listen both on ipv4 and ipv6 on port 4240 based on `EnableIPv4` and `EnableIPv6` config setting. @pippolo84 what do you suggest?

I see that `nodeGetCiliumEndpointNodeIP()` returns only the ipv4 address even in case of dual stack. So I think what you proposed is correct, we should listen on both ipv4 and ipv6 in case both options are set.

tamilmani1989:
@pippolo84 not sure why external workloads CI failing. It doesn't look like regular cilium agent running on a node. Is there anyway I can repro it locally and get access to node which runs `cilium-cilium-5707680440-1-vm-6`? Rest of failures looks flaky.

This is the failure:
```
cilium-health validation failed: "cilium-agent 'kube-system/cilium-w7lzk': connectivity to path 'cilium-cilium-5707680440-1-vm/cilium-cilium-5707680440-1-vm-6.host.primary-address.http.status' is unhealthy: 'Get \"[http://10.202.0.30:4240/hello\](http://10.202.0.30:4240/hello/)": dial tcp 10.202.0.30:4240: connect: connection refused'", retrying...
```
```
NAME                              CILIUM ID   IP
cilium-cilium-5707680440-1-vm-6   21875       10.202.0.30
```

pkg/health/probe/responder/responder.go
```diff
@@ -17,29 +18,62 @@ var defaultTimeout = 30 * time.Second
...
+		if ip != "" {
+			ipBytes, _ := netip.ParseAddr(ip)
+			if ipBytes.Is6() {
+				// if ipv6 address, then listen address should be in format of [ipv6]:port
+				ip = "[" + ip + "]"
+			}
+		}
+
+		hs := http.Server{
+			Addr:    fmt.Sprintf("%s:%d", ip, port),
```
pippolo84: You can avoid this using [net.JoinHostPort](https://pkg.go.dev/net#JoinHostPort)

pkg/health/probe/responder/responder.go
```diff
@@ -17,29 +18,62 @@ var defaultTimeout = 30 * time.Second
...
-	httpServer http.Server
+	httpServer []*http.Server
 }
 
 // NewServer creates a new server listening on the given port
-func NewServer(port int) *Server {
-	return &Server{
-		http.Server{
-			Addr:    fmt.Sprintf(":%d", port),
+func NewServer(address []string, port int) *Server {
```
pippolo84: Since the code is now returning more servers, I think it'd be better to change the naming (e.g: `NewServer` -> `NewServers`, `httpServer` -> `httpServers` and so on), alongside their comments.

pkg/health/probe/responder/responder.go
```diff
@@ -17,29 +18,62 @@ var defaultTimeout = 30 * time.Second
...
 	defer cancel()
-	return s.httpServer.Shutdown(ctx)
+	var err error
+	for _, hs := range s.httpServer {
+		if tmpError := hs.Shutdown(ctx); tmpError != nil {
+			err = tmpError
+		}
+	}
+
+	return err
```
pippolo84: I suggest to join all the errors like this:

```go
	errs := make([]error, 0, len(s.httpServer))
	for _, hs := range s.httpServer {
		errs = append(errs, hs.Shutdown(ctx))
	}
	return errors.Join(errs...)
```
[errors,Join](https://pkg.go.dev/errors#Join) will discard any nil error, if any.

pkg/health/probe/responder/responder_test.go
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package responder
+
+import (
+	"testing"
+
+	. "github.com/cilium/checkmate"
```
pippolo84: gopkg.in/check.v1 package was added to ease the transition out of `check.v1` in favor of the standard `testing` library. So, I suggest to use the std library for such a simple test (or, if you want to use assertion, you can try https://github.com/stretchr/testify that is already used widely in the codebase).

pkg/health/server/server.go
```diff
@@ -417,7 +419,24 @@ func NewServer(config Config) (*Server, error) {
...
+		}
+	}
+
+	if option.Config.EnableIPv6 {
+		nodeIpv6 := node.GetInternalIPv6()
+		if nodeIpv6 != nil {
+			address = append(address, nodeIpv6.String())
+		} else {
+			address = nil
+		}
```
pippolo84: I'd prefer to avoid these `else` and leave the slice as is, in case the ipv4 address has been already appended.
Maybe we should add a `log` entry in case `option.Config.EnableIPv4 == true && nodeIpv4 == nil` or `option.Config.EnableIPv6 == true && nodeIpv6 == nil`.

pkg/node/types/node.go
```diff
@@ -351,6 +351,21 @@ func (n *Node) GetK8sNodeIP() net.IP {
 	return externalIP
 }
 
+// GetNodeInternalIP returns the Internal IP of node or nil.
+func (n *Node) GetNodeInternalIP(ipv6 bool) net.IP {
```
pippolo84: The boolean flag in`GetNodeInternalIP(true)` and `GetNodeInternalIP(false)` is not so straightforward, I suggest to add two separate helpers (one for IPv4 and another one for IPv6) and get rid of it, so to improve the readability.
Also, that may simplify the `if` condition inside the helper itself.

pippolo84 (changes requested):
Thanks for this next iteration! :pray: 
I've left some comments, mainly around the style.

To setup a local cluster with external workloads you can refer to the doc [here](https://docs.cilium.io/en/latest/network/external-workloads/#external-workloads).
I think you can check `option.Config.JoinCluster` to understand if Cilium is running in a VM, since the feature is enabled based on that config flag.

pkg/health/server/server.go
tamilmani1989: the reason I added this `else` is that if for some reason `GetInternalIPv6` returns nil, then health server should listen on all address (`:4240`)and not on just appended ipv4 otherwise `ipv6:port` would not be reachable. 

The other option is to listen on specific ipv4 address and listen on all ipv6 addr(`::`). The same should go for v4(`0.0.0.0`) if `GetInternalIPv4` returns nil

pkg/node/types/node.go
tamilmani1989: sure will update it. I just followed existing api semantics `GetNodeIP` and `GetCiliumInternalIP` https://github.com/cilium/cilium/blob/6ba173d9ba392e05ff58930b482d697e00c70805/pkg/node/types/node.go#L356

pkg/health/probe/responder/responder.go
tamilmani1989: Thanks for pointing this out.

tamilmani1989:
> Thanks for this next iteration! 🙏 I've left some comments, mainly around the style.
> 
> To setup a local cluster with external workloads you can refer to the doc [here](https://docs.cilium.io/en/latest/network/external-workloads/#external-workloads). I think you can check `option.Config.JoinCluster` to understand if Cilium is running in a VM, since the feature is enabled based on that config flag.

I tried creating external workload following that doc but I lost connection to VM once i run this script `./install-external-workload.sh
` as specified in that step. I couldn't access VM after that. I see `cew` resource created for that external worload though
```
$ k get cew
NAME     CILIUM ID   IP
extvm3   29819       10.10.0.7
```

tamilmani1989:
These CI failures look flaky as it succeeded in previous run

pkg/health/server/server.go
pippolo84: I'd go for the second option, seems more natural in terms of user expectations.

pippolo84:
> These CI failures look flaky as it succeeded in previous run

Agree. Let's try a re-run and in case flakyness is confirmed I'm gonna open an issue.

pippolo84:
CI is green now, awesome! :100: 
Apart from [this](https://github.com/cilium/cilium/pull/26845#discussion_r1279032318) last feedback to address, one minor thing left: could you please add a comment where we check for `option.Config.JoinCluster` to explain why that is needed? :pray: Thanks!

pippolo84:
Hey @tamilmani1989 , I took a closer look at the last changes and I think we can avoid the introduction of the helpers `GetInternalIPv4` and `GetInternalIPv6`, relying on the already existing `node.GetIPv4` and `node.GetIPv6` instead.
Regarding the external workloads, I think that what you observed is due to the fact that the agent is running on a VM. In that case, we just have to leave the old logic since there is no such thing as an InternalIP. To do that, I suggest to introduce a helper to get the right addresses based on the global options. I did it [here](https://github.com/cilium/cilium/pull/27214/commits/6846adc0304462ff6fc3c2074f8f93f15278cd96) and it is working fine (don't mind the failures, they are just flakes, the health related tests are passing).
So, I suggest to:
- remove the helpers `GetInternalIPv4` and `GetInternalIPv6`
- refactor the code to introduce a helper for the server addresses as done [here](https://github.com/cilium/cilium/pull/27214/commits/6846adc0304462ff6fc3c2074f8f93f15278cd96)
- Change the comment removing the reference to the issue and stating that in the external workload case we listen on all interfaces and all address families
- squash the commits and fix all their messages to be compliant to the guidelines [here](https://github.com/cilium/cilium/pull/27214/commits/6846adc0304462ff6fc3c2074f8f93f15278cd96) (please take a careful look at point 5)

Thanks in advance!

tamilmani1989:
> Hey @tamilmani1989 , I took a closer look at the last changes and I think we can avoid the introduction of the helpers `GetInternalIPv4` and `GetInternalIPv6`, relying on the already existing `node.GetIPv4` and `node.GetIPv6` instead. Regarding the external workloads, I think that what you observed is due to the fact that the agent is running on a VM. In that case, we just have to leave the old logic since there is no such thing as an InternalIP. To do that, I suggest to introduce a helper to get the right addresses based on the global options. I did it [here](https://github.com/cilium/cilium/pull/27214/commits/6846adc0304462ff6fc3c2074f8f93f15278cd96) and it is working fine (don't mind the failures, they are just flakes, the health related tests are passing). So, I suggest to:
> 
> * remove the helpers `GetInternalIPv4` and `GetInternalIPv6`
> * refactor the code to introduce a helper for the server addresses as done [here](https://github.com/cilium/cilium/pull/27214/commits/6846adc0304462ff6fc3c2074f8f93f15278cd96)
> * Change the comment removing the reference to the issue and stating that in the external workload case we listen on all interfaces and all address families
> * squash the commits and fix all their messages to be compliant to the guidelines [here](https://github.com/cilium/cilium/pull/27214/commits/6846adc0304462ff6fc3c2074f8f93f15278cd96) (please take a careful look at point 5)
> 
> Thanks in advance!

The reason i added new api `GetInternalIPv4 ` is that existing `node.GetIPv4` may return node External IP if internal ip not exist and cilium would be listening on <external_ip>:4240. is that desired behavior? 

The comment for that function reads:

// GetIPv4 returns one of the IPv4 node address available with the following
// priority:
// - NodeInternalIP
// - NodeExternalIP
// - other IP address type.
// It must be reachable on the network.

pippolo84:
> The reason i added new api `GetInternalIPv4 ` is that existing `node.GetIPv4` may return node External IP if internal ip doesn't exist and cilium would be listening on <external_ip>:4240. is that desired behavior?
> 
> The comment for that function reads:
> 
> // GetIPv4 returns one of the IPv4 node address available with the following // priority: // - NodeInternalIP // - NodeExternalIP // - other IP address type. // It must be reachable on the network.

The only case I can think of where the InternalIP won't be available if the external workload, but we already take care of that separately. Anyway, there's no harm in having extra care, so let's leave your helpers as you suggested :+1:

tamilmani1989:
If GetIPv4 or GetIPV6 fails for some reason, then cilium agent would not listen on that ip family as its not appended to address list. so we may need this piece of code as well to address it (it was removed in your version)
```
	if len(addresses) != 0 {
		if option.Config.EnableIPv4 && ipv4 == nil {
			addresses = append(addresses, "0.0.0.0")
		}
		if option.Config.EnableIPv6 && ipv6 == nil {
			addresses = append(addresses, "::")
		}
	}
```

tamilmani1989:
@pippolo84  I updated PR based on above comments and my perspective. I'm open to hear if there any alternate suggestions and can update accordingly. Thanks for taking time to review this. Appreciate it. 👍

pkg/health/server/server.go
```diff
@@ -417,7 +420,46 @@ func NewServer(config Config) (*Server, error) {
...
+	)
+	// listen on all interfaces and all families in case of external-workloads
+	if option.Config.JoinCluster {
+		return []string{""}
+	}
+
+	if option.Config.EnableIPv4 {
+		if ipv4 = node.GetInternalIPv4(); ipv4 != nil {
+			addresses = append(addresses, ipv4.String())
+		}
```
pippolo84, line 443: You can simply use a `else` here to account for the case `ipv4 == nil`, so to avoid the `if len(addresses) != 0` later.

pkg/health/server/server.go
```diff
@@ -417,7 +420,46 @@ func NewServer(config Config) (*Server, error) {
...
+	}
+
+	if option.Config.EnableIPv6 {
+		if ipv6 = node.GetInternalIPv6(); ipv6 != nil {
+			addresses = append(addresses, ipv6.String())
+		}
```
pippolo84: Same here for `ipv6`

pippolo84 (changes requested):
Thank you for the further iteration :100: 
I agree with your analysis regarding the `GetInternalIPv4` and `GetInternalIPv4` returnng `nil`. I've left a suggestion to improve the readability of the code when handling those cases.

pippolo84 (approved):
Thanks! This turned out to be trickier than expected, great job! :rocket:

ldelossa:
All required tests passing. Merging

PR #26846, opened 2023-07-15
egressgw: consider egress interface when deleting stale IP rules

julianwiedmann:
When an EgressGW policy was updated to select a different egress interface, delete all IP rules that still point to the old interface.

```release-note
Fix a bug in the Egress Gateway feature when using the --install-egress-gateway-routes option. Delete stale IP rules after a CiliumEgressGatewayPolicy is updated and selects a different egress network interface.
```

### DIFF
--- a/pkg/egressgateway/manager_privileged_test.go
+++ b/pkg/egressgateway/manager_privileged_test.go
@@ -40,6 +40,7 @@ import (
 
 const (
 	testInterface1 = "cilium_egw1"
+	testInterface2 = "cilium_egw2"
 
 	node1 = "k8s1"
 	node2 = "k8s2"
@@ -56,6 +57,8 @@ const (
 
 	egressIP1   = "192.168.101.1"
 	egressCIDR1 = "192.168.101.1/24"
+	egressIP2   = "192.168.102.1"
+	egressCIDR2 = "192.168.102.1/24"
 
 	zeroIP4 = "0.0.0.0"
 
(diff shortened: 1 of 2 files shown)


pkg/egressgateway/manager_privileged_test.go
```diff
@@ -56,6 +57,8 @@ const (
 
 	egressIP1   = "192.168.101.1"
 	egressCIDR1 = "192.168.101.1/24"
+	egressIP2   = "192.168.102.1"
+	egressCIDR2 = "192.168.102.1/24"
```
aspsk, line 61: Not that it matters, but maybe `192.168.102.0/24` looks more canonical?

aspsk (approved):
lgtm, one minor cosmetic comment

pkg/egressgateway/manager_privileged_test.go
julianwiedmann: I would agree, but right now that change causes the tests to fail :thinking:. Needs a closer eye why that is.

aspsk: Ok, I see. I think code parses this CIDR as an IP address in ```func parseIPRule(sourceIP, destCIDR, egressIP string, ifaceIndex int)``` (it parses it as a CIDR, but then only uses the IP address part), didn't look too precise

gandro:
Marking as `backport/author`. The `matchFunc` looks different on v1.13. I could probably resolve them, but it looks like it might miss a dependent PR?

PR #26848, opened 2023-07-15
bpf: nat: small Masquerading improvements

julianwiedmann:
More chipping away at the SNAT / Masquerading code. Trying to get to a point where we can pass a fully populated CT tuple to `snat_v*_nat()`, and not repeat the same work there.

### DIFF
--- a/bpf/lib/conntrack.h
+++ b/bpf/lib/conntrack.h
@@ -321,6 +321,7 @@ ct_is_reply ## FAMILY(const void *map, struct __ctx_buff *ctx, int off,		\
 		      struct ipv ## FAMILY ## _ct_tuple *tuple,			\
 		      bool *is_reply)						\
 {										\
+	__u8 flags = tuple->flags;						\
 	int err = 0;								\
 										\
 	err = ct_extract_ports ## FAMILY(ctx, off, CT_EGRESS, tuple, NULL);	\
@@ -330,6 +331,12 @@ ct_is_reply ## FAMILY(const void *map, struct __ctx_buff *ctx, int off,		\
 	tuple->flags = TUPLE_F_IN;						\
 										\
 	*is_reply = map_lookup_elem(map, tuple) != NULL;			\
+										\
+	/* restore initial flags */						\
+	tuple->flags = flags;							\
+	/* callers (ie. SNAT code) have their own port extraction logic: */	\
+	tuple->dport = 0;							\
+	tuple->sport = 0;							\
 										\
 	return 0;								\
 }
--- a/bpf/lib/conntrack.h
+++ b/bpf/lib/conntrack.h
@@ -316,29 +316,22 @@ ct_lookup_select_tuple_type(enum ct_dir dir, enum ct_scope scope)
  * flow is a reply.
  */
 #define DEFINE_FUNC_CT_IS_REPLY(FAMILY)						\
-static __always_inline int							\
-ct_is_reply ## FAMILY(const void *map, struct __ctx_buff *ctx, int off,		\
-		      struct ipv ## FAMILY ## _ct_tuple *tuple,			\
-		      bool *is_reply)						\
+static __always_inline bool							\
+ct_is_reply ## FAMILY(const void *map,						\
+		      struct ipv ## FAMILY ## _ct_tuple *tuple)			\
 {										\
 	__u8 flags = tuple->flags;						\
-	int err = 0;								\
-										\
-	err = ct_extract_ports ## FAMILY(ctx, off, CT_EGRESS, tuple, NULL);	\
-	if (err < 0)								\
-		return err;							\
+	bool is_reply = false;							\
 										\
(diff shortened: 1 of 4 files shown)


bpf/lib/nat.h
```diff
@@ -1720,9 +1717,10 @@ snat_v6_needs_masquerade(struct __ctx_buff *ctx __maybe_unused,
 		 * is on 4 (after pfx.lpm.prefixlen), and we can't use memcpy()
 		 * on the whole field or the verifier complains.
 		 */
-		memcpy(pfx.lpm.data, &ip6->daddr, 4);
-		memcpy(pfx.lpm.data + 4, (__u8 *)&ip6->daddr + 4, 8);
-		memcpy(pfx.lpm.data + 12, (__u8 *)&ip6->daddr + 12, 4);
+		memcpy(pfx.lpm.data, &tuple->daddr.p1, 4);
+		memcpy(pfx.lpm.data + 4, &tuple->daddr.p2, 4);
+		memcpy(pfx.lpm.data + 8, &tuple->daddr.p3, 4);
+		memcpy(pfx.lpm.data + 12, &tuple->daddr.p4, 4);
```
aspsk, line 1721: The comment above doesn't make sense after this change. Can we leave that optimization here? If not, then we need to remove the comment.

bpf/lib/conntrack.h
```diff
@@ -316,22 +316,13 @@ ct_lookup_select_tuple_type(enum ct_dir dir, enum ct_scope scope)
...
-	err = ct_extract_ports ## FAMILY(ctx, off, CT_EGRESS, tuple, NULL);	\
-	if (err < 0)								\
-		return err;							\
-										\
 	tuple->flags = TUPLE_F_IN;						\
 										\
-	*is_reply = map_lookup_elem(map, tuple) != NULL;			\
-										\
-	return 0;								\
+	return map_lookup_elem(map, tuple) != NULL;				\
```
aspsk: bot complained about this one ^

bpf/lib/nat.h
```diff
@@ -746,19 +748,21 @@ snat_v4_needs_masquerade(struct __ctx_buff *ctx __maybe_unused,
...
+		err = ct_is_reply4(get_ct_map4(tuple), ctx, l4_off, tuple,
+				   &is_reply);
 		if (IS_ERR(err))
 			return err;
+
+		/* restore initial flags */
+		tuple->flags = flags;
+		/* SNAT code has its own port extraction logic: */
+		tuple->dport = 0;
+		tuple->sport = 0;
```
aspsk, line 763: maybe to wrap this into an inline function defined next to `ct_is_reply*` and comment that it should be in sync?

aspsk (changes requested):
looks great, a few minor comments

bpf/lib/nat.h
julianwiedmann: I *think* the comment still makes sense - we're still limited by `pfx.lpm.data`'s alignment on the stack. But now the tuple is on the stack as well, so that "middle" 8-byte copy no longer works either.

I slightly updated the comment to reflect that.

bpf/lib/nat.h
julianwiedmann: Hm yes - the `tuple->flags` reset should be done by `ct_is_reply()`. Resetting the ports needs to be aligned with `ct_extract_ports().`

I'll shuffle things around a bit :).

bpf/lib/conntrack.h
julianwiedmann: checkpatch is silly :grin:. Surely it would be even worse style to define the return value as `bool` and then return a pointer.

But this should go away when your comments below are addressed :+1:.

bpf/lib/nat.h
aspsk: Ok, this works. (I also checkedm just in case, that this will translate to 4 `u32` writes and tnothing more.)

bpf/lib/nat.h
```diff
@@ -752,10 +752,15 @@ snat_v4_needs_masquerade(struct __ctx_buff *ctx __maybe_unused,
...
-				   &is_reply);
-		if (IS_ERR(err))
+		err = ct_extract_ports4(ctx, l4_off, CT_EGRESS, tuple, NULL);
+		if (err < 0)
 			return err;
+
+		is_reply = ct_is_reply4(get_ct_map4(tuple), tuple);
+
+		/* SNAT code has its own port extraction logic: */
+		tuple->dport = 0;
```
aspsk, line 762: This looks like an extra operation. Isn't `ct_is_reply4` always resetting ports now?

julianwiedmann: I shuffled it out in the last patch - `ct_is_reply()` isn't extracting the ports anymore (that happens in the caller now), so it doesn't reset them either.

aspsk (approved):

PR #26850, opened 2023-07-15
Controller Group Metrics

asauber:
### Controller Group Metrics

Add a ControllerGroup for all controllers. ControllerGroup is used to
aggregate controllers which currently have a dynamic name, for example,
per-Endpoint and per-CiliumNetworkPolicy controllers.

ControllerGroups are used to expose a success/failure counter
Prometheus metric for each group. Each ControllerGroup also has a flag
which indicates if the metric should be enabled by default.

A config option is added which allows the user to specify a set of
controller group name strings, which override the default set of
controller metrics which are enabled. The special values of "all" and
"none" are supported.

### Example Usage

With `controller-group-metrics: "all"`

```
root@kind-control-plane:/home/cilium# curl localhost:9968/metrics 2>/dev/null | grep controllers_group
# HELP cilium_controllers_group_runs_total Number of times that a controller group was run labeled by completion status and controller name
# TYPE cilium_controllers_group_runs_total counter
cilium_controllers_group_runs_total{group_name="bpf-map-sync",status="success"} 29
cilium_controllers_group_runs_total{group_name="cilium-health",status="success"} 3
cilium_controllers_group_runs_total{group_name="dns-garbage-collector-job",status="success"} 3
cilium_controllers_group_runs_total{group_name="endpoint-gc",status="success"} 1
cilium_controllers_group_runs_total{group_name="initial-global-identities",status="success"} 3
cilium_controllers_group_runs_total{group_name="ipcache-inject-labels",status="failure"} 12
cilium_controllers_group_runs_total{group_name="ipcache-inject-labels",status="success"} 14
cilium_controllers_group_runs_total{group_name="k8s-heartbeat",status="success"} 5
cilium_controllers_group_runs_total{group_name="link-cache",status="success"} 10
cilium_controllers_group_runs_total{group_name="metricsmap-bpf-prom-sync",status="success"} 30
cilium_controllers_group_runs_total{group_name="resolve-identity",status="success"} 1
cilium_controllers_group_runs_total{group_name="resolve-labels",status="success"} 3
cilium_controllers_group_runs_total{group_name="restore-endpoint-identity",status="success"} 4
cilium_controllers_group_runs_total{group_name="sync-host-ips",status="success"} 3
cilium_controllers_group_runs_total{group_name="sync-lb-maps-with-k8s-services",status="success"} 1
cilium_controllers_group_runs_total{group_name="sync-policymap",status="success"} 15
cilium_controllers_group_runs_total{group_name="sync-to-k8s-ciliumendpoint",status="success"} 76
cilium_controllers_group_runs_total{group_name="sync-utime",status="success"} 3
cilium_controllers_group_runs_total{group_name="write-cni-file",status="success"} 1
```

With `controller-group-metrics: "sync-policymap sync-host-ips restore-endpoint-identity"`

```
root@kind-worker:/home/cilium# curl localhost:9968/metrics 2>/dev/null | grep group_runs_total
# HELP cilium_controllers_group_runs_total Number of times that a controller group was run labeled by completion status and controller name
# TYPE cilium_controllers_group_runs_total counter
cilium_controllers_group_runs_total{group_name="restore-endpoint-identity",status="success"} 1
cilium_controllers_group_runs_total{group_name="sync-host-ips",status="success"} 4
cilium_controllers_group_runs_total{group_name="sync-policymap",status="success"} 8
```

With `controller-group-metrics: "none all sync-policymap sync-host-ips restore-endpoint-identity"`

```
root@kind-worker:/home/cilium# curl localhost:9968/metrics 2>/dev/null | grep group_runs_total
root@kind-worker:/home/cilium#
```

Release note:

```release-note
Add per-controller success/failure count metrics and a config option for these
```


pkg/controller/controller.go
```diff
@@ -15,11 +15,25 @@ import (
...
+	noControllerMetricsEnabled  = "none"
+)
+
+var (
+	// groupMetricsEnabled indicates if the given
+	// ControllerGroup name has metrics enabled.
+	//
+	// Only if this map empty then the ControllerGroup
+	// MetricsEnabledDefault is used instead.
+	groupMetricsEnabled = populateGroupMetricsConfig()
```
asauber: This is likely a race with option.Config

clustermesh-apiserver/users_mgmt.go
```diff
@@ -34,6 +34,11 @@ var usersManagementCell = cell.Module(
 	cell.Invoke(registerUsersManager),
 )
 
+var usersManagementControllerGroup = controller.ControllerGroup{
+	Name:                  "clustermesh-users-management",
+	MetricsEnabledDefault: false,
```
giorio94: I'd be personally in favor of dropping the `MetricsEnabledDefault` parameter here, directly using the `controller-group-metrics` flag (and the corresponding helm setting which I suppose should be added) to enable the default ones. That would make it clear for users which are the enabled metrics (and allow different defaults for different components) as well as allow to simplify the creation here to something like `= controller.NewGroup("clustermesh-users-management")`.

clustermesh-apiserver/users_mgmt.go
```diff
@@ -96,7 +101,7 @@ func (us *usersManager) Start(hive.HookContext) error {
 		return fmt.Errorf("unable to setup config watcher: %w", err)
 	}
 
-	us.manager.UpdateController(usersMgmtCtrl, controller.ControllerParams{
+	us.manager.UpdateController(usersManagementControllerGroup, usersMgmtCtrl, controller.ControllerParams{
```
giorio94: This controller is run as part of the clustermesh-apiserver. To expose new metrics, you have to enable them here: https://github.com/cilium/cilium/blob/main/clustermesh-apiserver/metrics/metrics.go#L75-L87. The same goes for kvstoremesh here: https://github.com/cilium/cilium/blob/main/kvstoremesh/metrics/metrics.go#L89-L101.

daemon/cmd/daemon_main.go
```diff
@@ -789,6 +789,9 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	flags.Int(option.AuthMapEntriesName, option.AuthMapEntriesDefault, "Maximum number of entries in auth map")
 	option.BindEnv(vp, option.AuthMapEntriesName)
 
+	flags.StringSlice(option.ControllerGroupMetrics, []string{}, "List of controller group names for which to to enable metrics. Accepts 'all' and 'none'")
+	option.BindEnv(vp, option.ControllerGroupMetrics)
```
giorio94: Given the current modularization effort, I think it would be useful to move those to a separate `Config`, to allow registering the same flags in all components (agent, operator, clustermesh-apiserver, kvstoremesh). I guess that  the same could also apply for the metrics themselves: https://docs.cilium.io/en/latest/contributing/development/hive/#metric

daemon/cmd/cni/config.go
```diff
@@ -149,6 +149,11 @@ const chainedCNIEntry = `
 
 const cniControllerName = "write-cni-file"
 
+var cniControllerGroup = controller.ControllerGroup{
+	Name:                  "write-cni-file",
+	MetricsEnabledDefault: false,
```
squeed: heck it, enable this one :-)

pkg/controller/manager.go
```diff
@@ -41,17 +41,22 @@ func GetGlobalStatus() models.ControllerStatuses {
...
+// If a controller with the name already exists, the controller
+// will be shut down and replaced with the provided controller.
+//
+// A controller's group is used for aggregate metrics collection.
+// The group.Name must NOT be dynamically generated from a
+// resource identifier, in order to limit metrics cardinality.
+//
+// Updating a controller will cause the DoFunc to be run immediately regardless
+// of any previous conditions. It will also cause any statistics to be reset.
+func (m *Manager) UpdateController(group ControllerGroup, name string, params ControllerParams) {
```
squeed, line 55: Rather than changing the signature of `UpdateController`, what about adding the option to set a group for an entire Manager?

If that doesn't make sense / needs to be overridden, then the ControllerGroup should be a member in ControllerParams

squeed: Rather than changing the signature of `UpdateController`, what about adding the option to set a group for an entire Manager?

If that doesn't make sense / needs to be overridden, then the ControllerGroup should be a member in ControllerParams

asauber: Two thoughts there.

Having taken a look at all of the controllers, it seems that many of them do not have an associated Manager, so the group would have to be associated with Controller directly.

As for making ControllerGroup part of the ControllerParams, I had a previous iteration of this that did that, but upon implementation it started to not make sense: `ControllerParams` is passed as the sole parameter to `runController`, and it's expected that these parameters can be "updated" between runs. Since Group is expected to be static for the lifetime of the controller, and is not a parameter of the control loop itself, that approach didn't make sense to me.

daemon/cmd/daemon_main.go
asauber: I added a cell.Config for the controller package. Working on possibly using a cell.Metrics now. It looks like this might involve a large refactor that would be outside the scope of this PR. If you have any suggestions based on my initial approach, let me know.

pkg/controller/manager.go
squeed: I see what you mean, but given that `UpdateController()` is the only API we have, the group is by-implication updatable anyways. My vote is for both -- `params` and a per-Manager default in the case where that makes sense.

daemon/cmd/daemon_main.go
asauber: The same approach worked for cell.Metrics! https://github.com/cilium/cilium/pull/26850/commits/d1b385be0a9113898bcf148f24aa285e3ba89c13#diff-ac54f170f197d1a1607b60527697e2c8228d0a3ca57046a7814f76393e6d7a50R64

Let me know if you have any qualms about this "package-scope shuffling" using Invoke.

lmb (approved):
`pkg/bpf` LGTM.

tommyp1ckles:
small nit, first first commit has a typo: "aggregate cotnrollers"

pkg/controller/cell.go
```diff
@@ -7,35 +7,65 @@ import (
...
+
+	// labelControllerGroupName is the label used
+	// to identify controller-specific metrics
+	labelControllerGroupName = "group_name"
 )
 
-var ControllerGroupMetrics = map[string]bool{}
+var (
+	// GroupMetricEnabled is populated with the set of ControllerGroups for which metrics are enabled
+	GroupMetricEnabled = map[string]bool{}
```
tommyp1ckles: Should this be exported, seems dangerous if someone updates this map from another package async.

pkg/controller/cell.go
```diff
@@ -7,35 +7,65 @@ import (
...
+type Metrics struct {
+	ControllerGroupRuns metric.Vec[metric.Counter]
+}
+
+func NewMetrics() Metrics {
+	return Metrics{
+		ControllerGroupRuns: metric.NewCounterVec(metric.CounterOpts{
+			ConfigName: metrics.Namespace + "_controllers_group_runs_total",
+			Namespace:  metrics.Namespace,
+			Name:       "controllers_group_runs_total",
```
tommyp1ckles, line 48: How do we know any particular rate() of these is the correct one, since controllers are invoked on different periods and sometimes in exponential backoffs?

pkg/controller/cell.go
```diff
@@ -7,35 +7,65 @@ import (
...
+			Help:       "Number of times that a controller group was run, labeled by completion status and controller name",
+		}, []string{labelControllerGroupName, metrics.LabelStatus}),
```
tommyp1ckles, line 50: With respect to this being a gauge, will we be able to know what the current state of the controller group is based on this?

pkg/controller/cell.go
asauber: agreed, it should not be exported

pkg/controller/cell.go
asauber: The best info that we can get out of this appears to be based on the rate of the controller. My feeling is that this is fine for the purposes of these metrics. We can iterate with a proposed maximum period at some point. Any given alert can take this into account.

pkg/controller/manager.go
asauber: The group name has been moved back to `ControllerParams`. 

Adding it to `Manager` didn't make sense, at least the way that Manager is currently used. There is no name currently associated with a Manager, and many different existing groups (with quite varied purposes) are currently managed by the same Manager. So, adding a group name that would cross-cut these various controllers didn't make sense, given the goal of adding these metrics:

1. The metric should be specific enough that an alert can be constructed with associated documentation on how to recover from the specific type of controller failure.
2. The labels on this metric must be a bounded set of strings, and not be dynamically generated based on Endpoint ID, Policy Name, or similar.

Adding name to controller.Manager doesn't work towards these goals, at least with the current semantics implied by its usage.

squeed: Gotcha, understood.

There definitely are managers that do meet that definition; but if they are few and far between, there's no need to also add that to the manager.

asauber: Okay. I appreciate the second look

pkg/controller/cell.go
```diff
@@ -0,0 +1,71 @@
...
+	cell.Config(Config{}),
+	cell.Metric(NewMetrics),
+	cell.Invoke(Init),
+)
+
+type Metrics struct {
+	ControllerGroupRuns metric.Vec[metric.Counter]
+}
+
+func NewMetrics() Metrics {
```
christarazi, line 43: We should add the new metrics in the metrics reference in the docs: https://docs.cilium.io/en/v1.14/observability/metrics/#metrics-reference

Documentation/cmdref/cilium-agent.md
```diff
@@ -70,6 +70,7 @@ cilium-agent [flags]
       --config string                                             Configuration file (default "$HOME/ciliumd.yaml")
       --config-dir string                                         Configuration directory that contains a file for each option
       --conntrack-gc-interval duration                            Overwrite the connection-tracking garbage collection interval
+      --controller-group-metrics strings                          List of controller group names for which to to enable metrics. Accepts 'all' and 'none'
```
christarazi: One thing that came to mind is that controller group names may change over time and these names will be reference  in this flag here. Should we be concerned that we might break users? Do we want to call this out and say "please don't rely on the group names" or do we want to have a small process to deprecate names and define aliases to the new ones, similar to what we already do with deprecated metrics?

One approach might be to default to `all` and if a user wants to be selective, then we call out that they're subject to change, and to monitor the docs for the configurable groups. This would require keeping the [metrics reference](https://docs.cilium.io/en/v1.14/observability/metrics/#metrics-reference) up to date.

Open to thoughts here.

pkg/controller/manager.go
```diff
@@ -77,6 +78,7 @@ func (m *Manager) updateController(name string, params ControllerParams) *manage
 		ctrl = &managedController{
 			controller: controller{
 				name:       name,
+				group:      params.Group,
```
christarazi, line 87: Should we enforce that a group is always set?

pkg/endpoint/endpoint.go
```diff
@@ -1566,8 +1573,7 @@ type MetadataResolverCB func(ns, podName string) (pod *slim_corev1.Pod, _ []slim
 // will handle updates (such as pkg/k8s/watchers informers).
 func (e *Endpoint) RunMetadataResolver(resolveMetadata MetadataResolverCB) {
 	done := make(chan struct{})
-	const controllerPrefix = "resolve-labels"
-	controllerName := fmt.Sprintf("%s-%s", controllerPrefix, e.GetK8sNamespaceAndPodName())
+	controllerName := fmt.Sprintf("%s-%s", resolveLabels, e.GetK8sNamespaceAndPodName())
```
christarazi: We can simplify this to just be a string concat (see https://github.com/cilium/cilium/issues/19571)

christarazi (commented):
LGTM, some comments to address.

pkg/controller/cell.go
```diff
@@ -0,0 +1,71 @@
...
+	labelControllerGroupName = "group_name"
+)
+
+var (
+	// GroupMetricEnabled is populated with the set of ControllerGroups for which metrics are enabled
+	groupMetricEnabled = map[string]bool{}
+
+	// GroupRuns is a Prometheus-compatible metric for Controller
+	// runs, labeled by completion status and Group name
+	GroupRuns metric.Vec[metric.Counter]
```
pippolo84: What about initializing this to `metrics.NoOpCounterVec`? Doing so will allow to remove all the checks `GroupRuns != nil` in `pkg/controller/controller.go`.

pkg/controller/controller.go
```diff
@@ -95,6 +117,29 @@ func NoopFunc(ctx context.Context) error {
...
+		// "none" takes precedence over "all"
+		return false
+	} else if metricsEnabled[allControllerMetricsEnabled] {
+		return true
+	} else if len(metricsEnabled) > 0 {
+		// if any configuration was provided,
+		// then do not use any of the defaults
+		return metricsEnabled[g.Name]
+	}
+	return false
```
pippolo84: I think there's no need to check the map length. Also, all the `else` can be removed, since we are early-returning:
```
	if metricsEnabled[noControllerMetricsEnabled] {
		// "none" takes precedence over "all"
		return false
	}
	if metricsEnabled[allControllerMetricsEnabled] {
		return true
	}
	// if any configuration was provided,
	// then do not use any of the defaults
	return metricsEnabled[g.Name]
```
Besides, the last comment about the defaults is unclear to me. By "defaults" you mean "all" and "none"? If so, I think the comment is misleading, since we are looking for them first, in the map.

pippolo84 (commented):
Thanks for the PR! Left a couple of comments inline.

pkg/controller/manager.go
asauber: Previously, this was enforced by Group being a new additional parameter to `UpdateController`. 

This was reverted per this discussion: https://github.com/cilium/cilium/pull/26850#discussion_r1267893592.

Do you have a suggested technique for enforcing it as a struct-param-field?

pkg/controller/controller.go
asauber: Thanks for the pointer. There were previously some hardcoded defaults, so I can see how the comments are confusing.

pkg/controller/manager.go
christarazi: I can only think of a panic. When a developer submits a change for a controller without a group defined, it will show up in CI or when they do local testing. Otherwise, I would also document it but that isn't enough for enforcing. 

My worry is that we forget about the group parameter after this PR and then no new controllers use it. The only way to catch it would then be in code review, which would require everyone to be aware of the group field.

asauber: I would prefer it as a separate parameter rather than use a panic. @squeed What do you think about moving it back to a parameter for this reason?

Documentation/cmdref/cilium-agent.md
asauber: I would consider this a more advanced feature of Cilium, useful in diagnosing failure scenarios for high-volume use cases. Therefore, I think a stability note as you suggested is useful in the flag description and in the Helm documentation.

#### re: The cases where the set of group names changes

##### Group Name is Removed 

If a controller group name is removed, then specifying it in the configuration is a no-op (it's added to the map but never gets hit). If a user was using this metric to monitor that controller group, then the metric is no longer useful, because the controller has been removed or replaced by a new controller. In this case, the removal itself alerts the user that they need to revisit their use of the metric (an alias is not useful for them). Combined with the above stability note, I think that this is reasonable.

##### Group Name is Changed

Similar to the removal case, if a controller group name is changed, then the meaning of the metric has changed as well; something about the implementation of the controller has changed. If a user was using the metric to monitor a controller, then the name change alerts the user that they need to revisit their use of the metric (an alias is not useful for them).

##### Group Name is Added

The most useful documentation for this case would be an exhaustive list of all controller group names available to the Cilium component. However, generating this list is more difficult than it seems, and I would argue that it is not currently worth the effort. 

The `controller.Cell` which enables this feature is made active by this PR for `cilium-agent`, `clustermesh-apiserver`, and `kvstoremesh`. This dynamically includes the metric for all controller groups which are linked-to by that binary, for which there may be overlap between Cilium components. Writing a docsgen which identifies all of the group names present in the binary would be useful, and I think its a good follow-up once we start to use these metrics as operators.

#### re: The default set of group names

Given cardinality concerns, I don't think that defaulting to "all" is the way to go. I would prefer to point users to the default set in the Helm chart.

Given the above, I've added some additional documentation. Do you think it meets our current needs?

pkg/clustermesh/internal/remote_cluster.go
```diff
@@ -27,6 +27,11 @@ import (
 	"github.com/cilium/cilium/pkg/option"
 )
 
+var (
+	remoteConnectionControllerGroup = controller.NewGroup("clustermesh-remote-etcd")
```
giorio94:
```suggestion
	remoteConnectionControllerGroup = controller.NewGroup("clustermesh-remote-cluster")
```

Nit., as it does not only handle the etcd connection.

pkg/controller/cell.go
```diff
@@ -0,0 +1,71 @@
...
+	ControllerGroupRuns metric.Vec[metric.Counter]
+}
+
+func NewMetrics() Metrics {
+	return Metrics{
+		ControllerGroupRuns: metric.NewCounterVec(metric.CounterOpts{
+			ConfigName: metrics.Namespace + "_controllers_group_runs_total",
+			Namespace:  metrics.Namespace,
+			Name:       "controllers_group_runs_total",
+			Help:       "Number of times that a controller group was run, labeled by completion status and controller name",
```
giorio94:
```suggestion
			Help:       "Number of times that a controller group was run, labeled by completion status and controller group name",
```

Nit.

pkg/kvstore/consul.go
```diff
@@ -52,6 +52,8 @@ var (
 
 	// ErrNotImplemented is the error which is returned when a functionality is not implemented.
 	ErrNotImplemented = errors.New("not implemented")
+
+	consuleLeaseKeepaliveControllerGroup = controller.NewGroup("consul-lease-keepalive")
```
giorio94:
```suggestion
	consulLeaseKeepaliveControllerGroup = controller.NewGroup("consul-lease-keepalive")
```

Nit.

contrib/testing/kind-values.yaml
```diff
@@ -11,6 +11,10 @@ operator:
   nodeSelector:
     kubernetes.io/os: linux
     kubernetes.io/hostname: kind-worker
+prometheus:
+  enabled: true
+  controllerGroupMetrics:
+    - all
```
giorio94: Should this be reverted?

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1819,6 +1819,14 @@ prometheus:
   # ref: https://docs.cilium.io/en/stable/observability/metrics/
   metrics: ~
 
+  # --- Enable controller group metrics for monitoring specific Cilium subsystems. The list
+  # is a list of controller group names. The special values of "all" and "none" are supported.
+  controllerGroupMetrics:
+    - write-cni-file
+    - clustermesh-users-management
```
giorio94: This will not take effect, as the clustermesh-apiserver/kvstoremesh containers do not mount the `cilium-config` ConfigMap. I'd suggest to directly specify the argument (enabling all controllers) in their deployment.

pkg/controller/cell.go
```diff
@@ -0,0 +1,71 @@
...
+var Cell = cell.Module(
+	"controller",
+	"Controllers and Controller Lifecycle management",
+	cell.Config(Config{}),
+	cell.Metric(NewMetrics),
+	cell.Invoke(Init),
+)
+
+type Metrics struct {
+	ControllerGroupRuns metric.Vec[metric.Counter]
```
giorio94, line 40: While being here, it could make sense to extract also the other controller-related metrics, for consistency. That would make sure that they are also exposed by the clustermesh-apiserver/kvstoremesh.

Documentation/observability/metrics.rst
```diff
@@ -426,9 +426,19 @@ Name                                     Labels
 ======================================== ================================================== ========== ========================================================
 ``controllers_runs_total``               ``status``                                         Enabled    Number of times that a controller process was run
 ``controllers_runs_duration_seconds``    ``status``                                         Enabled    Duration in seconds of the controller process
+``controllers_group_runs_total``         ``status`` ``group_name``                          Enabled    Number of times that a controller process was run, labeled by controller group name (configurable)
```
giorio94: I would mention it also in the clustermesh-apiserver and kvstoremesh tables.

giorio94:
> The controller.Cell which enables this feature is made active by this PR for cilium-agent, clustermesh-apiserver, and kvstoremesh. 

Should it be enabled also for the cilium-operator?

install/kubernetes/cilium/values.yaml.tmpl
asauber: One of the concerns with this feature was an increase in metrics cardinality. Given your experience with those components, would you still recommend "all"?

pkg/controller/cell.go
asauber: This cell can definitely be made more useful. I think we should leave that to a follow-up PR so as not to expand the scope of this one too much.

install/kubernetes/cilium/values.yaml.tmpl
asauber: I've set both of these to "all" in the Helm chart.

asauber:
@giorio94 The cell is now enabled for cilium-operator.

@christarazi The stability note was added to the cmdref and the Helm docs.

install/kubernetes/cilium/values.yaml.tmpl
giorio94: Thanks! I'm not personally concerned in case of these two components, as they are very limited in scope, and make use of only a few controllers (mainly the ones used for the etcd client).

giorio94: It could be additionally removed from here, as this controller is not used by the agent.

install/kubernetes/cilium/templates/clustermesh-apiserver/deployment.yaml
```diff
@@ -161,6 +161,7 @@ spec:
         - --enable-external-workloads={{ .Values.externalWorkloads.enabled }}
         {{- if .Values.clustermesh.apiserver.metrics.enabled }}
         - --prometheus-serve-addr=:{{ .Values.clustermesh.apiserver.metrics.port }}
+        - --controller-group-metrics all
```
giorio94:
```suggestion
        - --controller-group-metrics=all
```

This should fix the currently failing tests. Same for the occurrence below.

pkg/controller/manager.go
tommyp1ckles: If we add a warn log it should similarly also fail in CI, seems safer than panic.

tommyp1ckles (approved):
looks good!

giorio94 (approved):
/lgtm

christarazi (approved):

Documentation/cmdref/cilium-agent.md
christarazi: > Given cardinality concerns, I don't think that defaulting to "all" is the way to go. I would prefer to point users to the default set in the Helm chart.

Agreed, I realized I said `all` in my previous but I will correct that to agree with the default set already provided in the PR.

> Group Name is Changed
> 
> Similar to the removal case, if a controller group name is changed, then the meaning of the metric has changed as well; something about the implementation of the controller has changed. If a user was using the metric to monitor a controller, then the name change alerts the user that they need to revisit their use of the metric (an alias is not useful for them).

That's an important point that I didn't realize. I would agree then that if a metric is "renamed", then it's likely a new metric anyway.

Documentation/observability/metrics.rst
learnitall: Can you add a comma between the labels and remove the `(configurable)` from the Description? The text you added below the table explains that it is configurable already. [1]

```suggestion
``controllers_group_runs_total``         ``status``, ``group_name``                          Enabled    Number of times that a controller process was run, labeled by controller group name
```

Documentation/observability/metrics.rst
```diff
@@ -592,6 +602,24 @@ Name                                     Labels
 ``lbipam_services_unsatisfied_total``                                                                      Enabled    Number of services which did not get requested IPs
 ======================================== ================================================================= ========== ========================================================
 
+Controllers
+~~~~~~~~~~~
+
+======================================== ================================================== ========== ========================================================
+Name                                     Labels                                             Default    Description
+======================================== ================================================== ========== ========================================================
+``controllers_group_runs_total``         ``status`` ``group_name``                          Enabled    Number of times that a controller process was run, labeled by controller group name (configurable)
```
learnitall: See [1]

Documentation/observability/metrics.rst
```diff
@@ -981,6 +1009,24 @@ Name                                           Labels
 ``api_limiter_wait_duration_seconds``          ``api_call``, ``value``          Mean, min, and max wait duration
 ============================================== ================================ ========================================================
 
+Controllers
+~~~~~~~~~~~
+
+======================================== ================================================== ========== ========================================================
+Name                                     Labels                                             Default    Description
+======================================== ================================================== ========== ========================================================
+``controllers_group_runs_total``         ``status`` ``group_name``                          Enabled    Number of times that a controller process was run, labeled by controller group name (configurable)
```
learnitall: See [1]

Documentation/observability/metrics.rst
```diff
@@ -1036,3 +1082,21 @@ Name                                           Labels
...
 ``api_limiter_wait_duration_seconds``          ``api_call``, ``value``          Mean, min, and max wait duration
 ============================================== ================================ ========================================================
+
+Controllers
+~~~~~~~~~~~
+
+======================================== ================================================== ========== ========================================================
+Name                                     Labels                                             Default    Description
+======================================== ================================================== ========== ========================================================
+``controllers_group_runs_total``         ``status`` ``group_name``                          Enabled    Number of times that a controller process was run, labeled by controller group name (configurable)
```
learnitall: See [1]

Documentation/observability/metrics.rst
```diff
@@ -592,6 +602,24 @@ Name                                     Labels
...
+======================================== ================================================== ========== ========================================================
+
+The ``controllers_group_runs_total`` metric reports the success
+and failure count of each controller within the system, labeled by
+controller group name and completion status. Enabling this metric is
+on a per-controller basis. This is configured using an allow-list
+which is passed as the ``controller-group-metrics`` configuration
+option. The current default set for ``cilium-operator`` found in
+the Cilium Helm chart is "all", which enables the metric for all
+controller groups. The special names "all" and "none" are supported.
```
learnitall: Can you reword this last sentence a little bit to help introduce that "all" is a special name before describing it as the default value? I was thinking of something like this:

```suggestion
the Cilium Helm chart is the special name "all", which enables the
metric for all controller groups. The special name "none" is also
supported.
```

[2]

Documentation/observability/metrics.rst
```diff
@@ -981,6 +1009,24 @@ Name                                           Labels
...
+======================================== ================================================== ========== ========================================================
+
+The ``controllers_group_runs_total`` metric reports the success
+and failure count of each controller within the system, labeled by
+controller group name and completion status. Enabling this metric is
+on a per-controller basis. This is configured using an allow-list
+which is passed as the ``controller-group-metrics`` configuration
+option. The current default set for ``clustermesh-apiserver`` found
+in the Cilium Helm chart is "all", which enables the metric for all
+controller groups. The special names "all" and "none" are supported.
```
learnitall: See [2]

Documentation/observability/metrics.rst
```diff
@@ -1036,3 +1082,21 @@ Name                                           Labels
...
+======================================== ================================================== ========== ========================================================
+
+The ``controllers_group_runs_total`` metric reports the success
+and failure count of each controller within the system, labeled by
+controller group name and completion status. Enabling this metric is
+on a per-controller basis. This is configured using an allow-list
+which is passed as the ``controller-group-metrics`` configuration
+option. The current default set for ``kvstoremesh`` found in
+the Cilium Helm chart is "all", which enables the metric for all
+controller groups. The special names "all" and "none" are supported.
```
learnitall: See [2]

Documentation/observability/metrics.rst
```diff
@@ -592,6 +602,24 @@ Name                                     Labels
...
+======================================== ================================================== ========== ========================================================
+``controllers_group_runs_total``         ``status`` ``group_name``                          Enabled    Number of times that a controller process was run, labeled by controller group name (configurable)
+======================================== ================================================== ========== ========================================================
+
+The ``controllers_group_runs_total`` metric reports the success
+and failure count of each controller within the system, labeled by
+controller group name and completion status. Enabling this metric is
+on a per-controller basis. This is configured using an allow-list
+which is passed as the ``controller-group-metrics`` configuration
+option. The current default set for ``cilium-operator`` found in
```
learnitall: I can't find a change to the `cilium-operator` deployment in this PR. Is the operator reading this from the `cilium-configmap` like the agent? If so, can you use the same language that you used for the agent here?

Documentation/observability/metrics.rst
```diff
@@ -426,9 +426,19 @@ Name                                     Labels
...
 ``controllers_runs_duration_seconds``    ``status``                                         Enabled    Duration in seconds of the controller process
+``controllers_group_runs_total``         ``status`` ``group_name``                          Enabled    Number of times that a controller process was run, labeled by controller group name (configurable)
 ``controllers_failing``                                                                     Enabled    Number of failing controllers
 ======================================== ================================================== ========== ========================================================
 
+The ``controllers_group_runs_total`` metric reports the success and failure
+count of each controller within the system, labeled by controller group name
+and completion status. Due to the large number of controllers, enabling this
+metric is on a per-controller basis. This is configured using an allow-list
+which is passed as the ``controller-group-metrics`` configuration option,
```
learnitall: In all of these descriptions that you added, can you change `option` to `flag` here, to make it clear to the user that this is something they need to pass to the binary? The word `option` might make users think this is a helm value that they can look for, especially in the description for clustermesh and kvstoremesh where the `prometheus.controllerGroupMetrics` helm value isn't listed. [3]

```suggestion
which is passed as the ``controller-group-metrics`` configuration flag,
```

Documentation/observability/metrics.rst
learnitall: Also see [3]

Documentation/observability/metrics.rst
```diff
@@ -981,6 +1009,24 @@ Name                                           Labels
...
+======================================== ================================================== ========== ========================================================
+``controllers_group_runs_total``         ``status`` ``group_name``                          Enabled    Number of times that a controller process was run, labeled by controller group name (configurable)
+======================================== ================================================== ========== ========================================================
+
+The ``controllers_group_runs_total`` metric reports the success
+and failure count of each controller within the system, labeled by
+controller group name and completion status. Enabling this metric is
+on a per-controller basis. This is configured using an allow-list
+which is passed as the ``controller-group-metrics`` configuration
+option. The current default set for ``clustermesh-apiserver`` found
```
learnitall: See [3]

Documentation/observability/metrics.rst
```diff
@@ -1036,3 +1082,21 @@ Name                                           Labels
...
+======================================== ================================================== ========== ========================================================
+``controllers_group_runs_total``         ``status`` ``group_name``                          Enabled    Number of times that a controller process was run, labeled by controller group name (configurable)
+======================================== ================================================== ========== ========================================================
+
+The ``controllers_group_runs_total`` metric reports the success
+and failure count of each controller within the system, labeled by
+controller group name and completion status. Enabling this metric is
+on a per-controller basis. This is configured using an allow-list
+which is passed as the ``controller-group-metrics`` configuration
+option. The current default set for ``kvstoremesh`` found in
```
learnitall: See [3]

learnitall (changes requested):
Looks great! Thank you! I just have a couple of comments about the changes to the metrics page.

pippolo84 (approved):
LGTM, thanks! :rocket:

pkg/controller/manager.go
squeed: We fail CI when error-level messages are detected (except for known-good ones). So I would say it's safe to log an Error if a controller is updated without a group being set. Then it will fail CI.

There could also be an "unknown" group, and we should fail CI if any metrics are published in that group.

asauber: A couple reasons why a default group name doesn't seem like the way to go here: We are attempting to limit cardinality, and the metric would have essentially zero practical purpose. The default set of group-names is opt-in (based on the Helm chart), so we would have to add logic to the agent CI in order for it to be visible.

Given that, I've added the Error log as suggested as the current solution here. The "log error and continue executing" pattern doesn't feel great to me, but in this case it does almost no harm, because if you haven't yet thought about the metrics of a new controller (and the name of those metrics), as a developer, then the dynamics of those metrics are likely entirely unknown, and they can be expanded on at a later time.

Documentation/observability/metrics.rst
```diff
@@ -1036,3 +1082,21 @@ Name                                           Labels
...
+======================================== ================================================== ========== ========================================================
+``controllers_group_runs_total``         ``status``, ``group_name``                         Enabled    Number of times that a controller process was run, labeled by controller group name
+======================================== ================================================== ========== ========================================================
+
+The ``controllers_group_runs_total`` metric reports the success
+and failure count of each controller within the system, labeled by
+controller group name and completion status. Enabling this metric is
+on a per-controller basis. This is configured using an allow-list
+which is passed as the ``controller-group-metrics`` configuration
+option. The current default set for ``kvstoremesh`` found in the
```
learnitall:
```suggestion
flag. The current default set for ``kvstoremesh`` found in the
```

learnitall (changes requested):
There was one s/option/flag that was missed, but other than that lgtm. Thanks!

asauber:
Hit known flake https://github.com/cilium/cilium/issues/27431

asauber:
rebased again to pick up fix for https://github.com/cilium/cilium/issues/27431

learnitall (approved):

squeed (approved):

PR #26852, opened 2023-07-16
bpf: nodeport: consolidate packet rewrite in RevDNAT path

julianwiedmann:
This adds some smarts to the nodeport helpers, so that it's possible to split up the RevDNAT lookup (what IP/port to use for the RevDNAT) and the actual packet rewrite. This allows us to use the same helper code for the rewrite, regardless whether the reply comes from a local or a DSR backend.

It's a nice simplification on its own, but will also enable a follow-on fix in the RevDNAT path (https://github.com/cilium/cilium/pull/26638) that would otherwise be rejected by the verifier. So we'll backport this PR accordingly, once that fix is ready to merged.

### DIFF
--- a/bpf/lib/conntrack.h
+++ b/bpf/lib/conntrack.h
@@ -1021,6 +1021,19 @@ err_ct_fill_up:
 	return DROP_CT_CREATE_FAILED;
 }
 
+static __always_inline bool
+__ct_has_nodeport_egress_entry(const struct ct_entry *entry,
+			       __u16 *rev_nat_index, bool check_dsr)
+{
+	if (entry->node_port) {
+		if (rev_nat_index)
+			*rev_nat_index = entry->rev_nat_index;
+		return true;
+	}
+
+	return check_dsr && entry->dsr;
+}
+
 /* The function tries to determine whether the flow identified by the given
  * CT_INGRESS tuple belongs to a NodePort traffic (i.e., outside client => N/S
  * LB => local backend).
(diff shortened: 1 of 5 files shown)


julianwiedmann:
https://github.com/cilium/cilium/pull/26638 needs a few more cycles until it's ready for primetime. But this part is good on its own already, so let's get started on review.

bpf/lib/conntrack.h
```diff
@@ -1042,10 +1042,19 @@ ct_has_nodeport_egress_entry4(const void *map,
...
+		return false;
 
-	return 0;
+	if (entry->node_port) {
+		if (rev_nat_index)
+			*rev_nat_index = entry->rev_nat_index;
+		return true;
+	}
+
+	if (check_dsr && entry->dsr)
```
aspsk: Maybe just `return check_dsr && entry->dsr;` ?

bpf/lib/conntrack.h
```diff
@@ -1076,10 +1085,19 @@ ct_has_nodeport_egress_entry6(const void *map,
...
+		return true;
+
+	return false;
```
aspsk: The code is identical with the v4 case. Maybe you can wrap the error checking and saving the `rev_nat_index` in an inline function?

aspsk (approved):

bpf/lib/conntrack.h
julianwiedmann: I like it :+1: See the pushed diff - two birds, one stone.

PR #26853, opened 2023-07-17
bpf: nat: set .from_local_endpoint for all inter-cluster SNAT traffic

julianwiedmann:
Otherwise snat_v4_nat() bails out in snat_v4_nat_can_skip() for traffic that uses a source port < NAT_MIN_EGRESS.


julianwiedmann:
@YutaroHayakawa does that make sense? I suppose we want to SNAT *all* outbound traffic. Either way, it would be good to have an BPF integration test that clarifies the expected behaviour :).

YutaroHayakawa (approved):
@julianwiedmann Sorry for the delay. I think it makes sense. Thanks for finding it out!

PR #26856, opened 2023-07-17
gh/workflows: Optionally enable dual stack in ci-e2e

brb:
Otherwise, when the IPv6 is disabled, the check-log-errors fails with:

    Error while inserting service in LB map" error="Unable to upsert
    service [fd00:10:96::8f2f]:8080 as IPv6 is disabled"
    k8sNamespace=cilium-test k8sSvcName=echo-same-node


sayboras (approved):
LGTM :heavy_check_mark:

nbusseneau (approved):

PR #26858, opened 2023-07-17
egressgw: ignore endpoint events with .endpoint = nil

aanm:
The commit 9fb24de93ebe added a check for pendingEndpointEvents that got
deleted by checking if the `id` was in the map. However, on delete
events, we were also inserting the request into the map but with an
explicit empty endpoint. This commit also checks if the endpoint is not
set for that case which we should skip the 'addEndpoint' events.
    
Fixes: 9fb24de93ebe ("egressgw: retry getIdentityLabels on failure")

### DIFF
--- a/pkg/egressgateway/manager.go
+++ b/pkg/egressgateway/manager.go
@@ -408,7 +408,7 @@ func (manager *Manager) addEndpoint(id types.NamespacedName) {
 	epEvent, ok := manager.pendingEndpointEvents[id]
 	manager.pendingEndpointEventsLock.RUnlock()
 
-	if !ok {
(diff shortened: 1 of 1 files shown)


jibi (approved):
I think it's good to be extra defensive here and check if endpoint is set, but when an endpoint gets deleted, `epEvent.endpoint` is rightfully set to nil as `(*Manager)deleteEndpoint()` will not try to access it (so the issue we are fixing here is for add/update events, as we assumed in that case the endpoint would always be set)

aanm:
Closed in favor of https://github.com/cilium/cilium/pull/26901

PR #26863, opened 2023-07-17
mlh: disable remove PR to project

mhofstetter:
Currently, trying to add new PRs to the 1.15 project dashboard fails because the column `In progress` doesn't exist. It seems as this feature has not been used for quite some time (1.14 timeframe).

To prevent MLH from failing, this feature gets disabled.


aanm (approved):
FYI @joestringer it looks that we no longer use this

PR #26867, opened 2023-07-17
Documentation: Fix Envoy LB docs incorrect supported annotation values

rauanmayemir:
There is a one-liner example that advices adding correct `service.cilium.io/lb-l7=enabled` annotation, but the table with supported annotations has a typo with incorrect `envoy` option instead of `enabled`.

### DIFF
--- a/Documentation/network/servicemesh/envoy-load-balancing.rst
+++ b/Documentation/network/servicemesh/envoy-load-balancing.rst
@@ -115,8 +115,8 @@ Supported Annotations
      - Default Value
    * - ``service.cilium.io/lb-l7``
      - Enable L7 Load balancing for kubernetes service.
-     - ``envoy``, ``disabled``
-     - Default to ``disabled``
+     - ``enabled``, ``disabled``
+     - Defaults to ``disabled``
    * - ``service.cilium.io/lb-l7-algorithm``
      - The LB algorithm to be used for services.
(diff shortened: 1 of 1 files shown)


zacharysarah (approved):
LGTM for docs

meyskens (approved):
Good catch! LGTM this is indeed the correct value

qmonnet (approved):
The mistake probably comes from the fact that the Helm value `loadBalancer.l7.backend` currently accepts either `envoy` or `disabled`. But the correct value for the `lb-l7` annotation is indeed `enabled`, as can be seen [here](https://github.com/cilium/cilium/blob/main/operator/pkg/ciliumenvoyconfig/annotations.go#L32).

Thanks!

qmonnet:
Could you please rebase on top of the current `main` branch? I believe the documentation builds were broken (then fixed) recently, and it would be nice to ensure the Netlify preview builds correctly before merging this PR.

rauanmayemir:
I messed it up with the merge commit, that's what you get for using github UI. Will try to fix the tree.

rauanmayemir:
Nice. https://deploy-preview-26867--docs-cilium-io.netlify.app/network/servicemesh/envoy-load-balancing.html#supported-annotations

qmonnet (approved):
All good, thank you!

Issue #24840, opened 2023-04-12
v1.12: CI: K8sDatapathConfig MonitorAggregation Checks that monitor aggregation flags send notifications


jschwinger233:
### Test Name

```shell
Suite-k8s-1.16.K8sDatapathConfig MonitorAggregation Checks that monitor aggregation flags send notifications
```

### Failure Output

```shell
FAIL: Timed out after 240.000s.
```

### Stack Trace

```shell
/home/jenkins/workspace/cilium-master-k8s-1.16-kernel-4.19/src/github.com/cilium/cilium/test/ginkgo-ext/scopes.go:515
Timed out after 240.000s.
monitor aggregation did not result in correct number of TCP notifications

Expected
    <bool>: false
to be true
/home/jenkins/workspace/cilium-master-k8s-1.16-kernel-4.19/src/github.com/cilium/cilium/test/k8s/datapath_configuration.go:139
```

### Standard Output

```shell
06:15:45 STEP: Installing Cilium
06:15:47 STEP: Waiting for Cilium to become ready
06:16:32 STEP: Validating if Kubernetes DNS is deployed
06:16:32 STEP: Checking if deployment is ready
06:16:32 STEP: Checking if kube-dns service is plumbed correctly
06:16:32 STEP: Checking if DNS can resolve
06:16:32 STEP: Checking if pods have identity
06:16:36 STEP: Kubernetes DNS is up and operational
06:16:36 STEP: Validating Cilium Installation
06:16:36 STEP: Performing Cilium status preflight check
06:16:36 STEP: Performing Cilium controllers preflight check
06:16:36 STEP: Checking whether host EP regenerated
06:16:36 STEP: Performing Cilium health check
06:16:43 STEP: Performing Cilium service preflight check
06:16:43 STEP: Performing K8s service preflight check
06:16:43 STEP: Cilium is not ready yet: connectivity health is failing: Cluster connectivity is unhealthy on 'cilium-bspz9': Exitcode: 1 
Err: exit status 1
Stdout:
 	 
Stderr:
 	 Error: Cannot get status/probe: Put "http://%2Fvar%2Frun%2Fcilium%2Fhealth.sock/v1beta/status/probe": dial unix /var/run/cilium/health.sock: connect: no such file or directory
	 
	 command terminated with exit code 1
	 

06:16:43 STEP: Performing Cilium controllers preflight check
06:16:43 STEP: Performing Cilium health check
06:16:43 STEP: Performing Cilium status preflight check
[...]

PR #26872, opened 2023-07-17
v1.11 backports 2023-07-17

pchaigno:
- [x] #25440 -- bpf: Don't encrypt on path hostns -> remote pod (@pchaigno)
     - Many small conflicts all around. Usually easy to resolve because we are removing code anyway.
 - [x] #26708 -- Fix upgrade for IPsec with tunneling (@pchaigno)
     - Minor conflicts for code that was removed anyway or code that was moved in the same file.
 - [x] #26810 -- docs/ipsec: Clarify limitation on number of nodes (@pchaigno)
     - Trivial conflict due to one line of docs not being backported to v1.12.

I manually tested:
- [x] Upgrade from v1.11.14 on EKS + overlay
- [x] Upgrade from v1.11.14 on EKS + ENI

Scale downs & ups as well as key rotations were not tested as we still have known issues on those operations (interruption of connections).

```upstream-prs
$ for pr in 25440 26708 26810; do contrib/backporting/set-labels.py $pr done 1.11; done
```

### DIFF
--- a/pkg/node/manager/manager.go
+++ b/pkg/node/manager/manager.go
@@ -427,8 +427,7 @@ func (m *Manager) NodeUpdated(n nodeTypes.Node) {
 		// to encrypt something we know does not have an encryption policy installed
 		// in the datapath. By setting key=0 and tunnelIP this will result in traffic
 		// being sent unencrypted over overlay device.
-		if !m.conf.NodeEncryptionEnabled() &&
-			(address.Type == addressing.NodeExternalIP || address.Type == addressing.NodeInternalIP) {
+		if !m.conf.NodeEncryptionEnabled() {
 			key = 0
 		}
 
--- a/pkg/node/manager/manager.go
+++ b/pkg/node/manager/manager.go
@@ -60,7 +60,6 @@ type Configuration interface {
 	TunnelingEnabled() bool
 	RemoteNodeIdentitiesEnabled() bool
 	NodeEncryptionEnabled() bool
-	EncryptionEnabled() bool
 }
 
 // Notifier is the interface the wraps Subscribe and Unsubscribe. An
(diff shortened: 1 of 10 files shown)


pchaigno:
/test-backport-1.11

pchaigno:
/test-1.16-netnext

pchaigno:
/test-1.21-5.4

pchaigno:
/test-1.16-netnext

pkg/node/manager/manager.go
```diff
@@ -409,7 +403,7 @@ func (m *Manager) NodeUpdated(n nodeTypes.Node) {
 		// through the tunnel to preserve the source identity as part of the
 		// encapsulation. In encryption case we also want to use vxlan device
 		// to create symmetric traffic when sending nodeIP->pod and pod->nodeIP.
-		if address.Type == addressing.NodeCiliumInternalIP || m.conf.EncryptionEnabled() ||
```
julianwiedmann, line 412: Upstream this was changed to `m.conf.NodeEncryptionEnabled()`, but not completely dropped. Before I start digging deeper, is there an easy reasoning for that diff?

julianwiedmann (commented):
lgtm so far, but one comment below.

pkg/node/manager/manager.go
pchaigno: I think upstream is correct and this is wrong. I'll fix it.

The same mistake went into the already-merged v1.13 backport PR and is in the draft v1.12 PR. I don't think it matters much given node-to-node encryption doesn't work anyway, but I'll send a fix to be consistent.

pchaigno: Fixed here and in the v1.12 PR (which is also now ready). I will send a PR to fix it in v1.13 for consistency.

pchaigno: Thanks for spotting this!!

julianwiedmann (approved):

pchaigno:
/test-backport-1.11

pchaigno:
k8s-1.18-kernel-4.9 hit known flake https://github.com/cilium/cilium/issues/24840. Marking ready to merge given monitor aggregation is unaffected here.

PR #26889, opened 2023-07-18
bpf: nat: SNAT cleanups

julianwiedmann:
Rip out a bunch of small things that on closer look turned out to be unused. I suspect most of this is leftovers from splitting the SNAT code into ingress/egress paths.

Also clean up two small inconsistencies in the ICMPv6 NS / NA handling.

### DIFF
--- a/bpf/lib/nat.h
+++ b/bpf/lib/nat.h
@@ -1590,13 +1590,12 @@ static __always_inline int snat_v6_rewrite_ingress(struct __ctx_buff *ctx,
 }
 
 static __always_inline bool
-snat_v6_nat_can_skip(const struct ipv6_nat_target *target, const struct ipv6_ct_tuple *tuple,
-		     bool icmp_echoreply)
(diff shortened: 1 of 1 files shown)


julianwiedmann:
Add two more small patches, and rebase to pick up https://github.com/cilium/cilium/pull/26641 for instant gratification.

bpf/lib/nat.h
```diff
@@ -1794,16 +1792,14 @@ snat_v6_nat(struct __ctx_buff *ctx, struct ipv6_ct_tuple *tuple, int off,
 			tuple->sport = icmp6hdr.icmp6_dataun.u_echo.identifier;
 			ct_action = ACTION_CREATE;
 		} else {
```
gentoo-root: Wanted to suggest to simplify the `if`s above using a switch, then I saw you did it in the last commit :D

gentoo-root (approved):
Looks good, thanks!

PR #26890, opened 2023-07-18
ci-gke: Set `useDigest=false` for Hubble Relay

gandro:
As with all other *-ci images, we should not use check the digest when pulling the image, as it is different for CI builds. On main this has not been an issue, because useDigest already defaults to false, but on release branches (such as v1.14), we do check the digest by default. This caused failing workflows on the v1.14 branch which should be fixed by this commit.

This extends the previous PR which was missing this change for the GKE workflow.

Fixes: 98003d5fef7b ("ci/github: Set `useDigest=false` for Hubble Relay")


michi-covalent (approved):

nbusseneau (approved):

michi-covalent:
the button is not green 😿

nbusseneau:
It can't be green, but it's still `ready-to-merge` in our hearts (but I will leave the privilege to @gandro).

PR #26894, opened 2023-07-18
cni: Follow CNI spec by using `(containerID, ifName)` as unique endpoint identifier

gandro:
The [CNI spec states](https://www.cni.dev/docs/spec/#section-3-execution-of-network-configurations) that a attachment is uniquely identified by the tuple consisting of both the container runtime container ID and the container interface name. Since a Cilium endpoint corresponds to a network attachment, the  container ID alone is not sufficient to uniquely identify an endpoint.

Therefore, this commit deprecates the concept of the container ID as a unique identifier for an endpoint, and instead introduces the "CNI attachment ID", which consists of both the containerID *and* container interface name, which is guaranteed to be unique according to the CNI spec.

Consequently, this PR also removes the restriction that there is a 1:1 mapping between a pod and an endpoint. To achieve this, we remove the `LookupPodName` helper in the agent and replace it with a `LookupCEPName` (for correlating with a CEP) and `GetEndpointsByPodName` (for correlating with a pod) helpers instead.

This change should be fully backwards compatible for endpoints which have a single interface.

 - Review per commit.
 - Co-authored-by @tklauser


gandro:
This is now ready to review (it already passed CI once)

asauber (approved):
CLI-owned changes LGTM. I haven't reviewed for correctness of the new endpoint logic.

api/v1/openapi.yaml
```diff
@@ -1796,23 +1804,32 @@ definitions:
...
+        description: Name of network device in host netns
         type: string
       interface-index:
-        description: Index of network device
+        description: Index of network device in host netns
         type: integer
+      container-interface-name:
+        description: Name of network device in container netns
+        type: string
+      container-interface-is-secondary:
```
asauber: Naming this "not primary" rather than "secondary", seems to make more sense. Not a blocker for me.

nathanjsweet (approved):

api/v1/openapi.yaml
```diff
@@ -1820,11 +1821,14 @@ definitions:
       +deepequal-gen=true
     type: object
     properties:
+      cni-attachment-id:
+        description: ID assigned to this attachment by container runtime
+        type: string
       container-id:
-        description: ID assigned by container runtime
+        description: ID assigned by container runtime (deprecated, may not be unique)
```
tommyp1ckles, line 1874: After we go up to version v1.16 could we remove this altogether?

pkg/endpoint/identifiers.go
```diff
@@ -41,6 +41,21 @@ func (e *Endpoint) GetK8sNamespaceAndPodName() string {
 	return e.K8sNamespace + "/" + e.K8sPodName
 }
 
+// getCNIAttachmentIDLocked returns the endpoint's unique CNI attachment ID
+func (e *Endpoint) getCNIAttachmentIDLocked() string {
+	if e.containerIfName != "" {
+		return e.containerID + ":" + e.containerIfName
+	}
+	return e.containerID
```
tommyp1ckles, line 70: should we ensure that this is a primary iface endpoint in this case?

pkg/endpoint/identifiers.go
```diff
@@ -41,6 +41,23 @@ func (e *Endpoint) GetK8sNamespaceAndPodName() string {
 	return e.K8sNamespace + "/" + e.K8sPodName
 }
 
+// GetK8sCEPResourceName returns the corresponding K8s CiliumEndpoint resource name
+// for this endpoint (without the namespace)
+func (e *Endpoint) GetK8sCEPResourceName() string {
```
tommyp1ckles: Can we document that its expected that this must return an empty string if the endpoint hasn't synced its PodName yet?

pkg/endpoint/api.go
```diff
@@ -145,17 +145,16 @@ func (e *Endpoint) getModelEndpointIdentitiersRLocked() *models.EndpointIdentifi
...
-	// primary attachment, they must not be used for endpoints of secondary
-	// attachments
+	// These fields only uniquely identify the endpoint of the primary attachment.
+	// They must not be used for endpoints of secondary attachments.
 	if !e.containerIfIsSecondary {
 		identifiers.ContainerID = e.containerID
 		identifiers.ContainerName = e.containerName
+		identifiers.PodName = e.GetK8sNamespaceAndPodName()
+		identifiers.K8sPodName = e.K8sPodName
+		identifiers.K8sNamespace = e.K8sNamespace
```
tommyp1ckles, line 156: Not really related specifically to this but this struct model is kind of strange, I expected everything in here to uniquely identify the endpoint but namespace does not do that on its own.

Is this even used anymore?

pkg/endpoint/identifiers.go
tommyp1ckles: small thought: the CEPResource name seems a bit redundant to me

pkg/k8s/watchers/endpointsynchronizer.go
```diff
@@ -95,11 +95,11 @@ func (epSync *EndpointSynchronizer) RunK8sCiliumEndpointSync(e *endpoint.Endpoin
...
 
-				// K8sPodName and K8sNamespace are not always available when an
-				// endpoint is first created, so we collect them here.
-				podName := e.GetK8sPodName()
-				if podName == "" {
-					scopedLog.Debug("Skipping CiliumEndpoint update because it has no k8s pod name")
+				// The CEP name is derived from the K8sPodName and K8sNamespace.
+				// Those are not always available when an endpoint is first created, so we collect them here.
+				cepName := e.GetK8sCEPResourceName()
+				if cepName == "" {
```
tommyp1ckles, line 101: The function comment says that K8sPodName is const but this would imply that it gets set later.

api/v1/openapi.yaml
gandro: I don't think we can remove this until we remove the docker plugin (ref https://github.com/cilium/cilium/issues/27022)

pkg/endpoint/identifiers.go
gandro: Not sure I follow. The pod name is set upon initial creation of the endpoint, so this should never be empty if the corresponding endpoint is created for a pod In CNI ADD.

It will always be empty for endpoints without a pod (e.g. host endpoint, health endpoint), and it will always be populated for endpoints backed by a pod, _even_ if we haven't synced the corresponding pod yet.

See also https://github.com/cilium/cilium/commit/e20b16d70843530aefad9ac8c3ae0834c7b6b057

pkg/endpoint/identifiers.go
gandro: Yes good point, I'll look into it.

pkg/endpoint/api.go
gandro: Yeah, this is a bit odd. The K8sNamespace is indeed not unique (and in fact we don't have a way to just pass a namespace into the `Lookup` method). I think this exists so K8s pod and K8s namespace are kept separate in the EndpointIndentifiers struct (they replaced the deprecated PodName field).

But I'm not sure they are actually used, I see no usages in our code. But they are part of the CEP CRD, so I didn't want to change them.

pkg/k8s/watchers/endpointsynchronizer.go
gandro: Oh yeah good catch. This comment is false since https://github.com/cilium/cilium/commit/e20b16d70843530aefad9ac8c3ae0834c7b6b057, I'll fix it.

zacharysarah (approved):
LGTM for docs

gandro:
Based on all the feedback, I have decided to remove the `ContainerInterfaceIsSecondary` flag and instead changed to code to fall-back on legacy lookups if the container interface is not set. Other than that, no significant changes, just addressed the other feedback and fixed a couple of source code comments.

pkg/k8s/watchers/endpointsynchronizer.go
gandro: Fixed by updating the comment to reflect reality

pkg/endpoint/identifiers.go
gandro: Fixed by renaming to `GetK8sCEPName`

pkg/endpoint/api.go
gandro: Unless you feel otherwise, I'll leave this as is for now.

pkg/endpoint/identifiers.go
gandro: Documented that it can return an empty string!

pkg/endpoint/identifiers.go
gandro: Fixed by removing the `secondary interface` flag instead

gandro:
> Based on all the feedback, I have decided to remove the `ContainerInterfaceIsSecondary` flag and instead changed to code to fall-back on legacy lookups if the container interface is not set. Other than that, no significant changes, just addressed the other feedback and fixed a couple of source code comments.

I had to undo that change and bring back the boolean flag. Instead of calling it  `ContainerInterfaceIsSecondary` though, I decided to call it by what it effectively does, i.e. `DisableLegacyIdentifiers`. It's called `disable` instead of `enable`, because this way it defaults to false if not provided and I want endpoints to explicitly opt-in for backwards compatibility reasons.

joamaki (approved):
operator changes LGTM

pkg/endpointmanager/manager.go
```diff
@@ -330,14 +333,27 @@ func (mgr *endpointManager) LookupIP(ip netip.Addr) (ep *endpoint.Endpoint) {
...
 	mgr.mutex.RLock()
-	ep := mgr.lookupPodNameLocked(name)
+	ep := mgr.lookupCEPNameLocked(namespacedName)
 	mgr.mutex.RUnlock()
 	return ep
 }
 
+// GetEndpointsByPodName looks up endpoints by namespace + pod name
+func (mgr *endpointManager) GetEndpointsByPodName(namespacedName string) []*endpoint.Endpoint {
+	mgr.mutex.RLock()
```
tommyp1ckles, line 346: nit: can use defer here.

pkg/endpointmanager/manager.go
```diff
@@ -330,14 +333,27 @@ func (mgr *endpointManager) LookupIP(ip netip.Addr) (ep *endpoint.Endpoint) {
...
+	eps := make([]*endpoint.Endpoint, 0, 1)
+	for _, ep := range mgr.endpoints {
```
tommyp1ckles, line 349: this is used by watcher Pod updates I believe, we're now iterating every endpoint upon each (local) update.  Probably is fine - but I wonder if there's cases where this might cause slowdowns.

pkg/k8s/watchers/endpointsynchronizer.go
```diff
@@ -95,11 +95,11 @@ func (epSync *EndpointSynchronizer) RunK8sCiliumEndpointSync(e *endpoint.Endpoin
...
-				// K8sPodName and K8sNamespace are not always available when an
-				// endpoint is first created, so we collect them here.
-				podName := e.GetK8sPodName()
-				if podName == "" {
-					scopedLog.Debug("Skipping CiliumEndpoint update because it has no k8s pod name")
+				// The CEP name is derived from the K8sPodName and K8sNamespace.
+				// They should always be available if an endpoint belongs to a pod.
+				cepName := e.GetK8sCEPName()
+				if cepName == "" {
+					scopedLog.Debug("Skipping CiliumEndpoint update because it has no k8s cep name")
```
tommyp1ckles, line 102: It seems like this case would should now be unexpected, and potentially a bug?  Perhaps we should error log this (and maybe return err to the controller?)

tommyp1ckles:
nit: commit 602668b0299ca7a46852093b2ba60df84e392bcd has a type "which requies endpoints"

gandro:
> nit: commit [602668b](https://github.com/cilium/cilium/commit/602668b0299ca7a46852093b2ba60df84e392bcd) has a type "which requies endpoints"

Fixed!

api/v1/models/endpoint_change_request.go
```diff
@@ -40,6 +40,11 @@ type EndpointChangeRequest struct {
 	// ID of datapath tail call map
 	DatapathMapID int64 `json:"datapath-map-id,omitempty"`
 
+	// Disables lookup using legacy endpoint identifiers (container name,
+	// container id, pod name) for this endpoint
+	//
```
tommyp1ckles: nit: extra comment line here

tommyp1ckles (approved):
k8s changes lgtm, just a few minor things.

pkg/k8s/watchers/endpointsynchronizer.go
gandro: Good point, I just checked and we are running this controller for the health and host endpoint which do not have a pod backing it. So emitting a warning or error would cause a lot of unnecessary noise.

We should probably fix that (i.e. not start the controller for such endpoints), but I'll leave that up for future work. I've opened an issue: https://github.com/cilium/cilium/issues/27128

api/v1/models/endpoint_change_request.go
gandro: Took me a while to figure this out, since it's auto generated. For future reference, the trick is to use ` >-` in openapi.yaml for the description

pkg/endpointmanager/manager.go
gandro: That is a valid concern. The current indexing unfortunately is not powerful enough to avoid that. I looked into it, and we would have to introduce another `endpointsByPodName map[string]map[uint16]*endpoint.Endpoint` (or similar) index in the endpoint manager to fix it, which seems a bit intrusive. This could be a great use-case for statedb.

Given that the number of endpoints on a node should be in the order of few hundred pods at most I'm not too worried and will leave it as is for now.

squeed (approved):

PR #26896, opened 2023-07-18
Add SPIRE connection to status

### LINKED ISSUE #1821: Mutual Auth: `cilium status --wait` needs to wait for SPIRE installation
Currently, `cilium status` doesn't take the installation of SPIRE into account (context: Cilium mutual Auth). This results in `cilium status --wait` not waiting until SPIRE is successfully installed.

Proposal: `cilium status` should include the status of the SPIRE server (K8s DaemonSet) and SPIRE agent (K8s StatefulSet).

meyskens:
This adds the SPIRE connection to cilium status, this then can be used by the CLI tool to surface errors and/or wait for SPIRE to be ready. If Auth is disabled it will surface the disabled status.

Fixes: https://github.com/cilium/cilium-cli/issues/1821

```release-note
Add SPIRE connection to `cilium status`
```


pkg/auth/spire/certificate_provider.go
```diff
@@ -106,3 +107,22 @@ func (s *SpireDelegateClient) SNIToNumericIdentity(sni string) (identity.Numeric
 func (s *SpireDelegateClient) SubscribeToRotatedIdentities() <-chan certs.CertificateRotationEvent {
 	return s.rotatedIdentitiesChan
 }
+
+func (s *SpireDelegateClient) Status() *models.Status {
+	s.connectedMutex.RLock()
+	defer s.connectedMutex.RUnlock()
+	if !s.connected {
+		msg := "not connected to SPIRE server"
```
mhofstetter: nit: Start with uppercase to be consistent with the message below

daemon/cmd/status.go
```diff
@@ -1045,6 +1045,26 @@ func (d *Daemon) startStatusCollector(cleaner *daemonCleanup) {
 				}
 			},
 		},
+		{
+			Name: "spire",
```
mhofstetter: The status of this probe depends on the implementation of `certs.CertificateProvider` which might or might not be implemented by `SPIRE` (`SpireDelegateClient`). Changing this implementation detail would affect the status api (which is unnecessary)

-> change to something more generic like `auth`, `auth-certificate-provider`, ...

daemon/cmd/daemon.go
```diff
@@ -217,6 +218,9 @@ type Daemon struct {
 	// statedb for implementing /statedb/dump
 	db statedb.DB
 
+	// authCertificateProvider for reporting the status of the auth system certificates
+	authCertificateProvider certs.CertificateProvider
```
mhofstetter: IMO it would be better if status reporting doesn't depend on internals of the auth module. It's a current implementation detail that there is only one auth handler which makes use of a certificate provider - hence there is only one of them. But this doesn't need to be the case.

It would be nicer if the auth module itself would report the status and aggregate all the details of its internals. In this case the authmanager would expose the status and would fetch the status from its internals (e.g. calling the authhandlers; that itself would call the certificateprovider, ...)

mhofstetter (changes requested):
@meyskens Thanks for the PR. I like the idea of leveraging the status for the cilium status instead of checking the SPIRE deployment directly.

I left some thoughts inline.

Overall i think the concept of the recently introduced `HealthReporter` would fit here quite well. It looks really promising. It already provides an easy way to report the health status of a hive module by injecting an instance of a [HealthReporter](https://github.com/cilium/cilium/blob/main/pkg/hive/cell/health.go#L39-L56). AFAIK, until now, it only lacks support for exposing the reported health status (e.g. via cilium status API).

But the bits for fetching the data are already prepared in the [Health interface](https://github.com/cilium/cilium/blob/main/pkg/hive/cell/health.go#L60-L77). IMO it should be possible that the daemon depends on the Health interface and adds a status probe for all registered modules by calling `All()`. This way the daemon doesn't have to be directly dependent on something "auth-related" at all.

I'm aware of the fact that this should land in v1.14. But I thought it's worth to bring this up and I'd like to have the input of @tommyp1ckles (maybe there's already something in the pipeline)

api/v1/models/status_response.go
```diff
@@ -104,6 +104,9 @@ type StatusResponse struct {
 
 	// List of stale information in the status
 	Stale map[string]strfmt.DateTime `json:"stale,omitempty"`
+
+	// Status of Authentication Spire connection
+	Spire *Status `json:"spire,omitempty"`
```
mhofstetter: > This file was generated by the swagger tool.

-> you need to edit the respective `openapi.yaml` and generate the code with `make generate-api generate-health-api generate-hubble-api generate-operator-api`

meyskens: oops i saw the notice but apparently went ahead anyway... thx for the reminder

daemon/cmd/daemon.go
```diff
@@ -217,6 +218,9 @@ type Daemon struct {
 	// statedb for implementing /statedb/dump
 	db statedb.DB
 
+	// authCertificateProvider for reporting the status of the auth system certificates
```
mhofstetter: wrong comment

pkg/auth/cell.go
```diff
@@ -24,7 +24,7 @@ import (
 	"github.com/cilium/cilium/pkg/stream"
 )
 
-// Cell invokes authManager which is responsible for request authentication.
+// Cell invokes AuthManager which is responsible for request authentication.
```
mhofstetter: nit: `Cell provides...`

mhofstetter (changes requested):
some small optimizations regarding comments. 

But it looks like there's a lint-error -> `Please run make generate-k8s-api & make manifests and submit your changes`

tommyp1ckles:
@meyskens @mhofstetter From the changes this does look a lot like a module health status,  the only difference being that the status is checked when the status endpoint is called, the health reporter status would likely be emitted when there is a change of health status state or on some periodic loop (preferably on a reconciliation loop/controller).

Right now the status of the health reporting is that we don't have the new Agent API health reporter endpoint in v1.14.

However the health status reporter updates can still be accessed via debug logging so there may be value in doing that (we could make the case for bumping that to info level in that case).

If you are interested in eventually adding health reporting lmk and I can provide more details.  For the time being I probably recommend doing the changes in this PR and having the health reporting be supplementary.

mhofstetter (approved):
LGTM, thanks for the changes @meyskens

meyskens:
Rebased for the conflict with ipcahce changes on main

nathanjsweet (approved):

youngnick:
The runtime test failed to copy the Ginkgo binary? I'll retest.

/test

tklauser:
@meyskens I've tried to backport this PR to `v1.14` as part of the backporter rotation but hit some non-trivial merge conflicts. Would you be ok to backport this yourself given you're more familiar with the code in question? Thanks

meyskens:
Will do!

michi-covalent:
i guess this didn't get moved to `backport-done/1.14`. trying to figure out how to manually fix it 🧠

PR #26898, opened 2023-07-18
 option: add LoadBalancerUsesDSR() helper

### LINKED ISSUE #25769: Consolidate checks for ad-hoc TUNNEL_PROTOCOL setting
We currently have multiple copies of the "special handling for ad-hoc tunnel" to create parameters for init.sh or set config options: https://github.com/cilium/cilium/blob/6b84bef94ba46ed3c09bf7a242a711449f3003bd/pkg/datapath/loader/base.go#L397 https://github.com/cilium/cilium/blob/6b84bef94ba46ed3c09bf7a242a711449f3003bd/pkg/option/config.go#L3141 and will add another one after completion of #24422 to write a correct value to node_config.h.

As a follow-up we should find a way to consolidate these checks in one place as explained by @julianwiedmann in this https://github.com/cilium/cilium/pull/24422#discussion_r1210202828_

julianwiedmann:
Spelling out the different modes that imply DSR is more robust than saying
"if it's not SNAT, it must be DSR".

As discussed in https://github.com/cilium/cilium/pull/25553#pullrequestreview-1454817088.

### DIFF
--- a/daemon/cmd/kube_proxy_replacement.go
+++ b/daemon/cmd/kube_proxy_replacement.go
@@ -215,17 +215,17 @@ func initKubeProxyReplacementOptions(tunnelConfig tunnel.Config) error {
 
 	if option.Config.EnableNodePort {
 		if option.Config.TunnelingEnabled() && tunnelConfig.Protocol() == tunnel.VXLAN &&
-			option.Config.NodePortMode != option.NodePortModeSNAT {
+			option.Config.LoadBalancerUsesDSR() {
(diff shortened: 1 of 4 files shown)


YutaroHayakawa (approved):
LGTM 👍

thorn3r (approved):
code clarity++ :+1:

ti-mo (approved):

squeed (approved):

PR #26901, opened 2023-07-18
egressgw: fix race with endpoint deletion

jibi:
with the current implementation of the workqueue used in the egressgw
manager to handle CiliumEndpoint events and retries it's possible to
trigger a race where by the time we process an add/update event, the
related CiliumEndpoint has already been deleted from the
pendingEndpointEvents, resulting in the agent panicing due to a nil
access.

This commit simplifies how the delete events are handled in order to
eliminate the race.

Fixes: 9fb24de93ebe ("egressgw: retry getIdentityLabels on failure")

### DIFF
--- a/pkg/egressgateway/manager.go
+++ b/pkg/egressgateway/manager.go
@@ -842,20 +842,14 @@ func (manager *Manager) reconcileLocked() {
 		return
 	}
 
-	if manager.eventBitmapIsSet(eventUpdateEndpoint, eventDeleteEndpoint) {
-		manager.updatePoliciesMatchedEndpointIDs()
-		manager.updatePoliciesBySourceIP()
-	}
-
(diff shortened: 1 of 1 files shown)


bimmlerd (approved):
tried pretty hard to find another race, but I think this is safe - my assumption is that `OnUpdateEndpoint` and `OnDeleteEndpoint` are not called concurrently (otherwise it's possible to reorder Add/Delete in such a way that the EP _should_ be there, but isn't). I think that's an ok assumption though.

jibi:
removed the explicit deletion of the event from `pendingEndpointEvents` https://github.com/cilium/cilium/blob/728e5777b804d268043b58a719a80380870c367d/pkg/egressgateway/manager.go#L333-L340 as discussed offline

aanm (approved):

PR #26905, opened 2023-07-18
BGP CP: Adds Support for LoadBalancerClass

### LINKED ISSUE #26556: L2-announcements and BGP compatibility
Currently, nighter the L2 announcement feature nor the BGP Control Plane have support for service [loadbalancer classes](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class). While it is unlikely that either feature will run in a cloud environment where another LB implementation could pick services up. Chances a user enabling both L2 announcement and BGP Control Plane in a on-prem environment are higher. So we might want to consider adding support for service loadbalancer classes to both features.

Currently LB-IPAM recognizes the `io.cilium/bgp-control-plane` loadbalancerClass, so the BGP control plane should use this class as well. The L2 announcement feature does not have a loadbalancerClass yet, so we should add one.

Doing so will allow users to pick on a per service basis which LB should be used, which this can technically already be done with label selectors, it makes sense to honor the actual service object.

danehans:
Updates BGP CP to support advertising services configured with a [LoadBalancerClass](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class).

Fixes: #26556

```release-note
When BGP control plane is enabled and configured for service announcements, it will only advertise a matching service that has an unspecified loadbalancerClass or set for "io.cilium/bgp-control-plane".
```

### DIFF
--- a/pkg/bgpv1/manager/reconcile.go
+++ b/pkg/bgpv1/manager/reconcile.go
@@ -626,26 +626,33 @@ func (r *LBServiceReconciler) svcDiffReconciliation(ctx context.Context, sc *Ser
 // desired state.
 func (r *LBServiceReconciler) svcDesiredRoutes(newc *v2alpha1api.CiliumBGPVirtualRouter, svc *slim_corev1.Service, ls localServices) ([]netip.Prefix, error) {
 	if newc.ServiceSelector == nil {
-		// If the vRouter has no service selector, there are no desired routes
+		// If the vRouter has no service selector, there are no desired routes.
 		return nil, nil
 	}
 
-	// Ignore non-loadbalancer services
+	// Ignore non-loadbalancer services.
 	if svc.Spec.Type != slim_corev1.ServiceTypeLoadBalancer {
 		return nil, nil
 	}
 
+	// The vRouter has a service selector, so determine the desired routes.
 	svcSelector, err := slim_metav1.LabelSelectorAsSelector(newc.ServiceSelector)
 	if err != nil {
 		return nil, fmt.Errorf("labelSelectorAsSelector: %w", err)
 	}
 
-	// Ignore non matching services
+	// Ignore non matching services.
 	if !svcSelector.Matches(serviceLabelSet(svc)) {
 		return nil, nil
 	}
 
-	// Ignore externalTrafficPolicy == Local && no local endpoints
+	// Ignore service managed by an unsupported LB class.
+	if svc.Spec.LoadBalancerClass != nil && *svc.Spec.LoadBalancerClass != v2alpha1api.BGPLoadBalancerClass {
+		// The service is managed by a different LB class.
+		return nil, nil
+	}
+
+	// Ignore externalTrafficPolicy == Local && no local endpoints.
 	if svc.Spec.ExternalTrafficPolicy == slim_corev1.ServiceExternalTrafficPolicyLocal &&
 		!hasLocalEndpoints(svc, ls) {
 		return nil, nil
--- a/pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumbgppeeringpolicies.yaml
+++ b/pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumbgppeeringpolicies.yaml
@@ -220,8 +220,12 @@ spec:
                       type: array
                     serviceSelector:
                       description: "ServiceSelector selects a group of load balancer
-                        services which this virtual router will announce. \n If empty
-                        / nil no services will be announced."
+                        services which this virtual router will announce. The loadBalancerClass
+                        for a service must be nil or specify a class supported by
+                        Cilium, e.g. \"io.cilium/bgp-control-plane\". Refer to the
+                        following document for additional details regarding load balancer
+                        classes: \n   https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class
+                        \n If empty / nil no services will be announced."
                       properties:
                         matchExpressions:
                           description: matchExpressions is a list of label selector
--- a/operator/pkg/lbipam/lbipam.go
+++ b/operator/pkg/lbipam/lbipam.go
@@ -68,7 +68,7 @@ func newLBIPAM(params LBIPAMParams) *LBIPAM {
 
 	var lbClasses []string
 	if params.DaemonConfig.EnableBGPControlPlane {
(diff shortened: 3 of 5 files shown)


danehans:
The `integration-test` job is failing due to:
```
    node_manager_test.go:608: 
        node_manager_test.go:593:
            c.Errorf("Node %s allocation mismatch. expected: %d allocated: %d", s.name, minAllocate, node.Stats().AvailableIPs)
        ... Error: Node node10 allocation mismatch. expected: 10 allocated: 20
        
        ... obtained int = 1010
        ... expected int = 1000
```

However, this test passes when run locally:
```
--- PASS: Test (13.39s)
    --- PASS: Test/IPAMSuite (13.39s)
...
        --- PASS: Test/IPAMSuite/TestNodeManagerManyNodes (0.04s)
```

rastislavs (approved):
LGTM, thanks!

rastislavs:
@danehans Could you also update the [docs for service-announcements](https://docs.cilium.io/en/latest/network/bgp-control-plane/#service-announcements) accordingly?

tommyp1ckles:
> The `integration-test` job is failing due to:

This is a known flake https://github.com/cilium/cilium/issues/26617

pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumbgppeeringpolicies.yaml
```diff
@@ -220,8 +220,12 @@ spec:
                       type: array
                     serviceSelector:
                       description: "ServiceSelector selects a group of load balancer
-                        services which this virtual router will announce. \n If empty
-                        / nil no services will be announced."
+                        services which this virtual router will announce. The loadBalancerClass
+                        for a service must be nil or specify a class supported by
+                        Cilium, e.g. \"io.cilium/bgp-control-plane\". Refer to the
+                        following document for additional details regarding load balancer
+                        classes: \n   https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class
```
tommyp1ckles, line 227: Are these newlines supported?

tommyp1ckles (approved):
K8s changes look ok to me, but I'm not that familiar with BGP so it may be worth having another reviewer take another look.

pkg/bgpv1/manager/reconcile.go
```diff
@@ -626,26 +626,33 @@ func (r *LBServiceReconciler) svcDiffReconciliation(ctx context.Context, sc *Ser
...
 
-	// Ignore non matching services
+	// Ignore non matching services.
 	if !svcSelector.Matches(serviceLabelSet(svc)) {
 		return nil, nil
 	}
 
-	// Ignore externalTrafficPolicy == Local && no local endpoints
+	// Ignore service managed by an unsupported LB class.
+	if svc.Spec.LoadBalancerClass != nil && *svc.Spec.LoadBalancerClass != v2alpha1api.BGPLoadBalancerClass {
```
ysksuzuki, line 650: Does this mean that BGP CP handles services where `.spec.loadBalancerClass` is not set as well? If so, wouldn't it interfere with the services that the cloud provider is supposed to be handling?

https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class

> By default, .spec.loadBalancerClass is not set and a LoadBalancer type of Service uses the cloud provider's default load balancer implementation if the cluster is configured with a cloud provider using the --cloud-provider component flag.

MetalLB handles only services whose spec.loadBalancerClass matches the given lb class. Have we considered an option like that?

https://github.com/metallb/metallb/blob/4672d27ba6168e833c33a4f3ace337c079de57b3/speaker/main.go#L83

> load balancer class. When enabled, metallb will handle only services whose spec.loadBalancerClass matches the given lb class

https://github.com/metallb/metallb/blob/5385d5c4aa96da3aeb6a72ef6b0d69e1a47cb38a/internal/k8s/controllers/service_controller.go#L164-L181

rastislavs: > Does this mean that BGP CP handles services where .spec.loadBalancerClass is not set as well? If so, wouldn't it interfere with the services that the cloud provider is supposed to be handling?

Yes, I think that is the current behavior as well due to user friendliness. As LB IPAM + BGP announcements is mostly an on-prem feature, most likely there will be no cloud provider controller handling the services when it is enabled - and in that case making the Cilium the default load balancer implementation probably makes sense.

ysksuzuki: Thank you for your comment!

> in that case making the Cilium the default load balancer implementation probably makes sense.

So What motivates users to set LoadBalancerClass to services in the on-prem environment where Cilium is the default load balancer implementation?

rastislavs: One motivation is described in the [original issue](https://github.com/cilium/cilium/issues/26556) - apart from BGP announcements, Cilium now also supports L2 announcement of LB services, so it can be used to distinguish between the two.

Apart from that, on-prem clusters may still have some other on-prem load-balancer deployed (such as MetalLB), for example if they used it in the past and now want to migrate to Cilium.

ysksuzuki: From the original issue
> The L2 announcement feature does not have a loadbalancerClass yet, so we should add one.

Are we going to add a loadbalancerClass for L2 mode in a separate PR? So that users can set it to the services which they expect to be handled by the L2 mode (not BGP mode).

rastislavs: Good point @ysksuzuki , I think it makes sense to have a separate PR for that - but that means we should not fully close the original issue with this PR @danehans

ysksuzuki: Thank you for your explanation. It's clear now.

ysksuzuki (approved):

pkg/k8s/apis/cilium.io/client/crds/v2alpha1/ciliumbgppeeringpolicies.yaml
danehans: The newlines are not reflected in `kubectl explain`. However, existing fields have the same issue and the newlines improve readability.

pkg/bgpv1/manager/reconcile.go
danehans: @rastislavs https://github.com/cilium/cilium/issues/26556 was reopened so L2 mode can be resolved.

PR #26912, opened 2023-07-19
services: prevent temporary connectivity loss on agent restart

### LINKED ISSUE #23823: clustermesh global services: temporary connectivity loss on agent restart

### What happened?

Consider a clustermesh scenario composed of two clusters, and a global service with backends present only in one cluster (to better highlight the issue). The restart of the cilium agent on the other cluster causes a brief period in which pods hosted on the same node cannot connect to the global service, while connectivity to local services and local/remote pods works without any interruption.

### Cilium Version

cilium image (running): v1.13.0
I expect this to affect also older versions, but I haven't tested them yet 

### Kernel Version

Linux 6.1.0-3-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.8-1 (2023-01-29) x86_64 GNU/Linux

### Kubernetes Version

Client Version: v1.26.1
Kustomize Version: v4.5.7
Server Version: v1.25.3

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

From a quick look, the issue seems to occur at startup since the remote backends are not recognized to be part of any service, and get removed:

```
level=debug msg="Removing obsolete backend" backendID=50 backends="[]" l7LBFrontendPorts="[]" l7LBProxyPort=0 loadBalancerSourceRanges="[]" serviceID=13 serviceIP="{10.21.243.156 {TCP 80} 0}" serviceName=global-svc serviceNamespace=app sessionAffinity=false sessionAffinityTimeout=0 subsys=service svcExtTrafficPolicy=Cluster svcHealthCheckNodePort=0 svcIntTrafficPolicy=Cluster svcType=ClusterIP
```

The issue gets fixed when the agent receives an update from the remote etcd server, and the service gets restored.

### Code of Conduct

### LINKED ISSUE #26944: During Cilium restart DualStack service might temporarily remove one IP family endpoints
Observed during the upgrade from `v1.13.4` to the `main`:
```
level=debug msg="Kubernetes service definition changed" action=service-updated endpoints="[fd00:10:244:1::3b68]:8000/TCP,[fd00:10:244:3::3153]:8000/TCP,[fd00:10:244::4fee]:8000/TCP" k8sNamespace=cilium-test k8sSvcName=test-conn-disrupt old-endpoints= old-service=nil service="frontends:[10.96.141.20 fd00:10:96::22c1]/ports=[http]/selector=map[app:test-conn-disrupt-server]" subsys=k8s-watcher
...
level=debug msg="Upserting service" backends="[]" l7LBFrontendPorts="[]" l7LBProxyPort=0 loadBalancerSourceRanges="[]" serviceIP="{10.96.141.20 {TCP 8000} 0}" serviceName=test-conn-disrupt serviceNamespace=cilium-test sessionAffinity=false sessionAffinityTimeout=0 subsys=service svcExtTrafficPolicy=Cluster svcHealthCheckNodePort=0 svcIntTrafficPolicy=Cluster svcType=ClusterIP
...
```

Eventually, the agent received an update for the service with IPv4 endpoints. Until then the IPv4 endpoints were removed which interrupted open IPv4 connections to that service.

The issue reproduced in https://github.com/cilium/cilium/pull/26398 when enabling `ipv6`.

I haven't checked whether other versions are impacted too.

giorio94:
Cilium already implements a restore path to prevent dropping existing connections on agent restart. Yet, there's currently an issue which causes the removal of valid backends from a service when receiving incomplete service updates, either because backends are spread across multiple endpointslices or some belong to remote clusters. Indeed, all previously known backends get replaced with the ones we just heard about (and present as part of the service cache event), possibly causing connectivity disruptions.

Let's prevent this behavior keeping a list of restored backends for each service, and continuing merging them with the ones we received an update for, until the bootstrap phase completes. After synchronization, an update is triggered for each service still associated with stale backends, so that they can be removed.

Fixes: #23823
Fixes: #26944

```release-note
Fix possible connection drops on agents restart when a service is associated with multiple endpointslices or has backends across multiple clusters
```

### DIFF
--- a/pkg/service/service.go
+++ b/pkg/service/service.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/cilium/cilium/pkg/bpf"
 	"github.com/cilium/cilium/pkg/cidr"
@@ -20,6 +21,7 @@ import (
 	datapathOpt "github.com/cilium/cilium/pkg/datapath/option"
 	"github.com/cilium/cilium/pkg/datapath/types"
 	datapathTypes "github.com/cilium/cilium/pkg/datapath/types"
+	"github.com/cilium/cilium/pkg/k8s"
 	lb "github.com/cilium/cilium/pkg/loadbalancer"
 	"github.com/cilium/cilium/pkg/lock"
 	"github.com/cilium/cilium/pkg/logging/logfields"
@@ -104,6 +106,9 @@ type svcInfo struct {
 	LoopbackHostport          bool
 
 	restoredFromDatapath bool
+	// The hashes of the backends restored from the datapath and
+	// not yet heard about from the service cache.
+	restoredBackendHashes sets.Set[string]
 }
 
 func (svc *svcInfo) isL7LBService() bool {
@@ -1072,7 +1077,27 @@ func (s *Service) restoreAndDeleteOrphanSourceRanges() error {
 //
 // The removal is based on an assumption that during the sync period
 // UpsertService() is going to be called for each alive service.
-func (s *Service) SyncWithK8sFinished() error {
+func (s *Service) SyncWithK8sFinished(ensurer func(k8s.ServiceID, *lock.StoppableWaitGroup) bool) error {
+	servicesWithStaleBackends := sets.New[lb.ServiceName]()
+
+	// We need to trigger the stale services refresh while not holding the
+	// lock, to ensure that the generated events can be processed, and to
+	// prevent a possible deadlock in case the events channel is already full.
+	defer func() {
+		swg := lock.NewStoppableWaitGroup()
+
+		for svc := range servicesWithStaleBackends {
+			ensurer(k8s.ServiceID{
+				Cluster:   svc.Cluster,
+				Namespace: svc.Namespace,
+				Name:      svc.Name,
+			}, swg)
+		}
+
+		swg.Stop()
+		swg.Wait()
+	}()
+
 	s.Lock()
 	defer s.Unlock()
 
@@ -1086,7 +1111,18 @@ func (s *Service) SyncWithK8sFinished() error {
 			if err := s.deleteServiceLocked(svc); err != nil {
 				return fmt.Errorf("Unable to remove service %+v: %s", svc, err)
 			}
+		} else if svc.restoredBackendHashes.Len() > 0 {
+			// The service is still associated with stale backends
+			servicesWithStaleBackends.Insert(svc.svcName)
+			log.WithFields(logrus.Fields{
+				logfields.ServiceID:      svc.frontend.ID,
+				logfields.ServiceName:    svc.svcName.String(),
+				logfields.L3n4Addr:       logfields.Repr(svc.frontend.L3n4Addr),
+				logfields.OrphanBackends: svc.restoredBackendHashes.Len(),
+			}).Info("Service has stale backends: triggering refresh")
 		}
+
+		svc.restoredBackendHashes = nil
 	}
 
 	// Remove no longer existing affinity matches
@@ -1533,6 +1569,14 @@ func (s *Service) restoreServicesLocked(svcBackendsById map[lb.BackendID]struct{
 			svcBackendsById[backend.ID] = struct{}{}
 		}
 
+		if len(newSVC.backendByHash) > 0 {
+			// Indicate that these backends were restored from BPF maps,
+			// so that they are not removed until SyncWithK8sFinished()
+			// is executed (if not observed in the meanwhile) to prevent
+			// disrupting valid connections.
+			newSVC.restoredBackendHashes = sets.KeySet(newSVC.backendByHash)
+		}
+
 		// Recalculate Maglev lookup tables if the maps were removed due to
 		// the changed M param.
 		ipv6 := newSVC.frontend.IsIPv6() || (svc.NatPolicy == lb.SVCNatPolicyNat46)
@@ -1651,6 +1695,10 @@ func (s *Service) updateBackendsCacheLocked(svc *svcInfo, backends []*lb.Backend
 				backends[i].ID = s.backendByHash[hash].ID
 			}
 		} else {
+			// We observed this backend, hence let's remove it from the list
+			// of the restored ones.
+			svc.restoredBackendHashes.Delete(hash)
+
 			backends[i].ID = b.ID
 			// Backend state can either be updated via kubernetes events,
 			// or service API. If the state update is coming via kubernetes events,
@@ -1684,6 +1732,14 @@ func (s *Service) updateBackendsCacheLocked(svc *svcInfo, backends []*lb.Backend
 
 	for hash, backend := range svc.backendByHash {
 		if _, found := backendSet[hash]; !found {
+			if svc.restoredBackendHashes.Has(hash) {
+				// Don't treat backends restored from the datapath and not yet observed as
+				// obsolete, because that would cause connections targeting those backends
+				// to be dropped in case we haven't fully synchronized yet.
+				backends = append(backends, backend)
+				continue
+			}
+
 			obsoleteSVCBackendIDs = append(obsoleteSVCBackendIDs, backend.ID)
 			if s.backendRefCount.Delete(hash) {
 				DeleteBackendID(backend.ID)
--- a/pkg/service/service_test.go
+++ b/pkg/service/service_test.go
@@ -10,13 +10,17 @@ import (
 
 	. "github.com/cilium/checkmate"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/exp/maps"
 
 	"github.com/cilium/cilium/pkg/checker"
 	"github.com/cilium/cilium/pkg/cidr"
 	cmtypes "github.com/cilium/cilium/pkg/clustermesh/types"
 	datapathOpt "github.com/cilium/cilium/pkg/datapath/option"
 	datapathTypes "github.com/cilium/cilium/pkg/datapath/types"
+	"github.com/cilium/cilium/pkg/k8s"
 	lb "github.com/cilium/cilium/pkg/loadbalancer"
+	"github.com/cilium/cilium/pkg/lock"
 	nodeTypes "github.com/cilium/cilium/pkg/node/types"
 	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/service/healthserver"
@@ -611,7 +615,7 @@ func (m *ManagerTestSuite) TestSyncWithK8sFinished(c *C) {
 
 	// cilium-agent finished the initialization, and thus SyncWithK8sFinished
 	// is called
-	err = m.svc.SyncWithK8sFinished()
+	err = m.svc.SyncWithK8sFinished(func(k8s.ServiceID, *lock.StoppableWaitGroup) bool { return true })
 	c.Assert(err, IsNil)
 
 	// svc1 should be removed from cilium while svc2 is synced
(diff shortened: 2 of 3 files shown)


brb:
This PR resolves https://github.com/cilium/cilium/issues/26944.

pkg/service/service.go
```diff
@@ -1086,7 +1111,18 @@ func (s *Service) SyncWithK8sFinished() error {
...
 			}
+		} else if svc.restoredBackendHashes.Len() > 0 {
+			// The service is still associated with stale backends
+			withStaleBackends.Insert(svc.svcName)
+			log.WithFields(logrus.Fields{
+				logfields.ServiceID:      svc.frontend.ID,
+				logfields.ServiceName:    svc.svcName.String(),
+				logfields.L3n4Addr:       logfields.Repr(svc.frontend.L3n4Addr),
+				logfields.OrphanBackends: svc.restoredBackendHashes.Len(),
+			}).Info("Service has stale backends: triggering regeneration")
```
christarazi: Nit: "regeneration" is probably not the right terminology here and I'd rather avoid the conflation of this "regeneration" and Cilium endpoint regeneration. Maybe "refresh" is better word?

pkg/service/service.go
```diff
@@ -1072,7 +1077,27 @@ func (s *Service) restoreAndDeleteOrphanSourceRanges() error {
 //
 // The removal is based on an assumption that during the sync period
 // UpsertService() is going to be called for each alive service.
-func (s *Service) SyncWithK8sFinished() error {
+func (s *Service) SyncWithK8sFinished(ensurer func(k8s.ServiceID, *lock.StoppableWaitGroup) bool) error {
```
christarazi, line 1080: Did you consider making this function return the set of services and then call `ensurer` on the return of `SyncWithK8sFinished` instead? Might make it easier to understand what `ensurer` is for. Anonymous functions can make the code a bit more convenient to write, but hinders readability. This is a minor comment though because there's only one call site of `SyncWithK8sFinished`, so mostly just wondering what you've considered.

christarazi (approved):
LGTM, good find! 

Minor comments below.

pkg/service/service.go
```diff
@@ -1684,13 +1732,20 @@ func (s *Service) updateBackendsCacheLocked(svc *svcInfo, backends []*lb.Backend
...
+				// to be dropped in case we haven't fully synchronized yet.
+				backends = append(backends, backend)
+			} else {
+				obsoleteSVCBackendIDs = append(obsoleteSVCBackendIDs, backend.ID)
+				if s.backendRefCount.Delete(hash) {
+					DeleteBackendID(backend.ID)
+					delete(s.backendByHash, hash)
+					obsoleteBackendIDs = append(obsoleteBackendIDs, backend.ID)
+				}
+				delete(svc.backendByHash, hash)
```
aditighag: Since there is no change in the original code to delete obsolete backends, can we wrap the special restore case inside `if` block and break?

pkg/service/service.go
```diff
@@ -1072,7 +1077,27 @@ func (s *Service) restoreAndDeleteOrphanSourceRanges() error {
 //
 // The removal is based on an assumption that during the sync period
 // UpsertService() is going to be called for each alive service.
-func (s *Service) SyncWithK8sFinished() error {
+func (s *Service) SyncWithK8sFinished(ensurer func(k8s.ServiceID, *lock.StoppableWaitGroup) bool) error {
+	withStaleBackends := sets.New[lb.ServiceName]()
+
+	// We need to trigger the stale services regeneration while not holding
```
aditighag: I don't think regeneration is the right word here. How about services `update`?

pkg/service/service_test.go
```diff
@@ -611,7 +615,7 @@ func (m *ManagerTestSuite) TestSyncWithK8sFinished(c *C) {
 
 	// cilium-agent finished the initialization, and thus SyncWithK8sFinished
 	// is called
-	err = m.svc.SyncWithK8sFinished()
+	err = m.svc.SyncWithK8sFinished(func(k8s.ServiceID, *lock.StoppableWaitGroup) bool { return true })
```
aditighag, line 618: `SyncWithK8sFinished` is called conditionally (see `initRestore`). Can we add one more test with restore flag disabled, just to make sure we don't have any oops moments?

aditighag (changes requested):
Nice find! Fix looks good to me overall with some requested changes.

Do we want this logic to execute in case of standalone LB?

>removal of valid backends from a service when receiving incomplete service updates, either because backends are spread across multiple endpointslices or some belong to remote clusters.

I wonder how this issue go unnoticed until now. Is there a scale aspect to it, or did something change on k8s side?
I'm not too familiar with clustermesh, so this might be a newbie question. But does clustermesh run multiple instances of a k8s api server? If not, why would it matter that backends belong to remote clusters?

pkg/service/service.go
```diff
@@ -1072,7 +1077,27 @@ func (s *Service) restoreAndDeleteOrphanSourceRanges() error {
 //
 // The removal is based on an assumption that during the sync period
 // UpsertService() is going to be called for each alive service.
-func (s *Service) SyncWithK8sFinished() error {
+func (s *Service) SyncWithK8sFinished(ensurer func(k8s.ServiceID, *lock.StoppableWaitGroup) bool) error {
+	withStaleBackends := sets.New[lb.ServiceName]()
```
ysksuzuki: How about `serviceWithStaleBackends`? It's clearer to me.

ysksuzuki (changes requested):
The scenario in which the problem occurs is as follows, and this PR fixes this problem by delaying the deletion of the restored backends until the synchronization with kube-apiserver is complete. Is my understanding correct?

- Restore services and backends from the BPF map
- Sync with kube-apiserver and reflect the actual state
  - If the k8sServiceHandler receives an incomplete event like the following(event1 is incomplete), a valid backend is removed prematurely
    - event1{service1: backend1 } <- a valid backend2 is removed (the informer cache sync delays, perhaps?)
    - event2{service1: backend1, backend2 } <- the removed backend2 is back

Is this scenario that the handler receives an incomplete event likely to happen if the backends spread across multiple endpointslices or some belong to remote clusters?

brb:
> Do we want this logic to execute in case of standalone LB?

Today, the standalone LB mode doesn't assume any connectivity to kube-apiserver. Services are programmed through Cilium API.

pkg/service/service.go
giorio94: I've changed it to *refresh* for consistency with the updated log message below.

giorio94:
> I wonder how this issue go unnoticed until now. Is there a scale aspect to it, or did something change on k8s side?

Yes, in the sense that this issue doesn't occur in case all the backends are contained in a single endpointslice (given that we see only a single atomic update). By default, an endpointslice contains 100 endpoints, hence Cilium may (depending on events ordering this might also not occur) currently remove valid backends only when a service is associated with more than 100 pods.

> I'm not too familiar with clustermesh, so this might be a newbie question. But does clustermesh run multiple instances of a k8s api server? If not, why would it matter that backends belong to remote clusters?

Essentially, each cluster exposes an etcd cluster and synchronizes there a subset of the information available as Kubernetes resources, including the list of the services which are marked as shared, and the associated backends. Remote agents connect to these etcd clusters to pull the various objects and merge them with the local state. In the services case, the remote backends get merged into the service cache, as if they were just a different endpointslice. I personally discovered the issue in this scenario, as a single backend is enough to trigger the connectivity drop (opposed to the local case).

brb:
> I wonder how this issue go unnoticed until now. Is there a scale aspect to it, or did something change on k8s side?

How I hit this - add DualStack services, and then do Cilium upgrade.

giorio94:
> The scenario in which the problem occurs is as follows, and this PR fixes this problem by delaying the deletion of the restored backends until the synchronization with kube-apiserver is complete. Is my understanding correct?
> 
>     * Restore services and backends from the BPF map
> 
>     * Sync with kube-apiserver and reflect the actual state
>       
>       * If the k8sServiceHandler receives an incomplete event like the following(event1 is incomplete), a valid backend is removed prematurely
>         
>         * event1{service1: backend1 } <- a valid backend2 is removed (the informer cache sync delays, perhaps?)
>         * event2{service1: backend1, backend2 } <- the removed backend2 is back
> 
> 

Exactly. Essentially, consider the case in which a given service *foo* is associated with the *foo-1* and *foo-2* epslices each containing a set of backends. 

As you mentioned, upon restart, Cilium restores services and backends from the BPF map. Then it starts the service and epslice informers, both of which propagate the event to the service cache. Let's say that we first receive the event about the service: at this point the service is considered not ready (as we have not yet seen any epslice), and nothing gets propagated. Then, we receive the event for the *foo-1* epslice, the service cache processes it and, given that the service has now backends, propagates an event down to the service subsystem for the service *foo*, including all backends part of *foo-1*. At this point, all previously known backends get replaced by the new ones in the BPF maps, dropping the connections targeting the backends that were part of the *foo-2* epslice. Once an event for that epslice is also seen, then the backends will be merged and restored. 

The only case in which this issue doesn't happen is when we first receive both *foo-1* and *foo-2* events from kubernetes, and only afterwards the one for the service itself.  

The modification implemented in this PR ensures that in the service subsystem we continue preserving the backends restored from the BPF maps also once we receive the first service update, given that at that point some backends might still be missing.

> Is this scenario that the handler receives an incomplete event likely to happen if the backends spread across multiple endpointslices or some belong to remote clusters?

Yes, it doesn't happen in case of a single endpointslice, because all backends would be part of a single event atomically received from the k8s informer. And the clustermesh case essentially corresponds to having another epslice for each remote cluster.

giorio94:
> How I hit this - add DualStack services, and then do Cilium upgrade.

TIL that endpointslices are created per family. Hence, if the service is of type dualstack, we'll have two separate endpointslices (regardless of whether pods are dualstack or not).

pkg/service/service.go
giorio94: I personally don't have a strong opinion here. IMO the current approach has the advantage that the entire synchronization logic is self-contained in a single function, which is then easier to be tested. Another possibility might be to propagate the entire service cache (rather than the single function) as parameter (but that would make testing harder), or replace the anonymous function with an interface (I initially discarded this option as being more verbose, but it might improve clarity). WDYT?

pkg/service/service_test.go
giorio94: I've made a few attempts here, but I couldn't find any good way to test this condition without a large refactoring. The main issue is that the synchronization logic is triggered asynchronously through a controller, hence making it very difficult to wait for its termination. Given that the `initRestore` function is already untested (hence there's no check that the other operations are also skipped when `RestoreStatus` is unset) and the condition is pretty trivial, I'd personally skip the implementation of this test for the moment.

giorio94:
@aditighag, @ysksuzuki PTAL

ysksuzuki (approved):
Thank you for the clear explanation! Looks good to me

giorio94:
Last force-push was to extend the commit message with the information from the discussion above. No code changes have been introduced.

giorio94:
I've double-checked, and this issue additionally affects v1.11 (hence also v1.12).

pkg/service/service.go
christarazi: Given that from an offline conversation about this function we talked about it being executed in a controller although it's one-off, maybe we just defer the refactoring of this function anyway (into a hive job or something). I'm fine as it is.

pkg/service/service_test.go
aditighag: It isn't so much about testing `initRestore`, as about making sure there are no deadlocks, or we don't assume that `syncWithK8sFinished` is always invoked (e.g., standalone LB).

aditighag (approved):
> > Do we want this logic to execute in case of standalone LB?
> 
> Today, the standalone LB mode doesn't assume any connectivity to kube-apiserver. Services are programmed through Cilium API.

Right, so should this logic be skipped for standalone LB case? 

Do you also want to mention the ipv{4, 6} case in the commit description that Martynas linked above?

Approving the change, as I'll be on PTO next week. Please check on unit testing the standalone case where `syncWithK8sFinished` is a no-op. Thanks!

giorio94:
>>>  Do we want this logic to execute in case of standalone LB?
>>
>>  Today, the standalone LB mode doesn't assume any connectivity to kube-apiserver. Services are programmed through Cilium API.
>
> Right, so should this logic be skipped for standalone LB case?

AFAIU, this logic is skipped when running in standalone LB mode, because the k8s client is not enabled in that case, which is [a requirement ](https://github.com/giorio94/cilium/blob/mio%2Fservice-obsolete-backends/daemon/cmd/state.go#L452-L477)for the k8s to lb maps synchronization. More in general, the synchronization logic was already present, and the changes in this PR only extend it with an additional cleanup step, but do not modify how it is triggered. 

> Do you also want to mention the ipv{4, 6} case in the commit description that Martynas linked above?

Sure, the updated commit message mentions that the issue can also occur in case of dual stack services, and links to the Martynas's issue.

pkg/service/service_test.go
giorio94: Yeah, that logic is already triggered only in case that `RestoreState` is set, and the clientset is enabled (which excludes the standalone LB case). It is currently tricky to further test these aspects, because the `syncWithKsFinished` function (which by itself is covered by unit tests) is run asynchronously. 

I'd propose to defer the introduction of additional tests to a subsequent refactoring which extracts this logic to be executed by an hive job, as mentioned by Chris. I'd not include the refactoring in this PR though, to reduce churn in older versions and because that would complicate backporting given that hive jobs support has only been introduced recently.

giorio94:
Further refactoring and related tests are tracked by https://github.com/cilium/cilium/issues/27012. Marking as ready to merge.

PR #26933, opened 2023-07-19
ci: fix Azure cluster names sometimes being too long

nbusseneau:
Azure does not allow having multiple clusters with the same name in the same subscription even if they are hosted on different locations.

In order to avoid name conflicts, we previously added the location name to the cluster name, however in some cases this leads to cluster names exceeding the maximum length.

As a quick fix, we replace the location name with a simple index.

### DIFF
--- a/.github/actions/azure/k8s-versions.yaml
+++ b/.github/actions/azure/k8s-versions.yaml
@@ -3,8 +3,11 @@
 include:
   - version: "1.24"
     location: westeurope
+    index: 1
   - version: "1.25"
     location: westus
+    index: 2
   - version: "1.26"
     location: eastasia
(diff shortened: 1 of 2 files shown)


nbusseneau:
Link to working run: https://github.com/cilium/cilium/actions/runs/5601174013
The run hit flake #26008 but the name change worked as expected.

This PR does not need to run any other CI, marking `ready-to-merge`.

viktor-kurchenko (approved):
LGTM.

michi-covalent (approved):
magical

nbusseneau:
This PR does not need be backported to 1.11 because we don't test AKS on 1.11 anymore.

PR #26620, opened 2023-07-04
Refactor duplicate imports for Cilium v2alpha1 API

dlapcevic:
A couple of go files had `github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2alpha1` imported twice.
Remove the duplicate and make the import name consistent.

Signed-off-by: Dorde Lapcevic <[dordel@google.com](mailto:dordel@google.com)>

PR #26938, opened 2023-07-19
endpointslice: fix EndpointSlice import

mhofstetter:
This commit fixes the import of EndpointSlice.

Merge race between #26455 & #26620

Fixes: #26455


aditighag (approved):
Thanks!

PR #26946, opened 2023-07-20
USERS: Add Trendyol

eminaktas:
Starting from 1.26, Trendyol is now using Cilium as default CNI.
Also, we published an article about our process [here](https://t.ly/FDCZK)

### DIFF
--- a/USERS.md
+++ b/USERS.md
@@ -585,6 +585,11 @@ Users (Alphabetically)
       L: https://teuto.net/managed-kubernetes
       Q: @cwrau
 
+    * N: Trendyol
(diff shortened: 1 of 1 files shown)


eminaktas:
> Thanks for the addition @eminaktas ! The blog article is great too 💯 I'll add it to the blog and newsletter
> 
> Would you also be interested in doing a case study about your use of Cilium? You can see examples here https://www.cncf.io/case-studies?_sft_lf-project=cilium We would just need a 30 minute interview with you

Thanks @xmulligan. We would love to be part of the case study. I will ping you on Slack. ❤️

xmulligan (approved):

aditighag:
❤️

PR #26953, opened 2023-07-20
v1.13: node: Fix node encryption condition in incorrect backport

pchaigno:
Julian noticed that backported commit 7dbd40f280 ("ipsec: Remove workarounds for path asymmetry that was removed") didn't match the upstream commit due to a wrong conflict resolution. This commit fixes it.

We don't expect much impact here as node encryption was never stable, barely used, and eventually removed, but best to fix it, if only for consistency with v1.11 and v1.12.

Fixes: https://github.com/cilium/cilium/pull/26792.


pchaigno:
/test-backport-1.13

julianwiedmann (approved):
Thank you Paul!

PR #26960, opened 2023-07-20
egressgw: use Resource[T] to consume CiliumEgressGatewayPolicy

lmb:
egressgw: manually construct Manager in tests

    Don't rely on hive to construct a Manager, it's more trouble than it's
    worth. Instead create the necessary dependencies manually.

    This gives us much more control over which types are used for certain
    interfaces without having to fiddle with hive rules.

egressgw: use tb.Cleanup to tear down dummy interfaces

    The testsuite leaves around dummy interfaces if the test aborts for some
    reason. Use tb.Cleanup to avoid this behaviour.

egressgw: use Resource[T] to consume CiliumEgressGatewayPolicy

    Replace the copy pasta'd k8s watcher for CiliumEgressGatewayPolicy with 
    resource.Resource.


pkg/egressgateway/manager_privileged_test.go
```diff
@@ -179,7 +174,7 @@ func (k *EgressGatewayTestSuite) TestEgressGatewayManager(c *C) {
 
 	// Create a new policy
 	policy1 := newEgressPolicyConfigWithNodeSelector("policy-1", ep1Labels, destCIDR, []string{}, nodeGroup1Selector, testInterface1)
-	egressGatewayManager.OnAddEgressPolicy(policy1)
+	egressGatewayManager.onAddEgressPolicy(policy1)
```
lmb: I think we shouldn't call the method directly, but instead go via k.policies by sending an event. That means constructing `CEGP` in tests however. Currently there is only a helper for `PolicyConfig`

@jibi should I change this to something like the following?
```go
k.policies.process(resource.Event[*Policy]{Obj: CEGP{...}})
```

pkg/egressgateway/manager_privileged_test.go
```diff
@@ -402,40 +399,37 @@ func TestCell(t *testing.T) {
...
-	if err != nil {
-		return err
+	if err := netlink.AddrAdd(link, a); err != nil {
+		tb.Fatal(err)
 	}
 
-	if err := netlink.LinkDel(link); err != nil {
-		return err
-	}
+	tb.Cleanup(func() {
```
markpash: Nit: Could we register the cleanup function a little earlier? (Just after calling `netlink.LinkByName(iface)`)

pkg/egressgateway/manager.go
```diff
@@ -267,20 +273,59 @@ func (manager *Manager) getIdentityLabels(securityIdentity uint32) (labels.Label
...
+				event.Done(nil)
+			} else {
+				manager.handlePolicyEvent(event)
+			}
 		}
-	}()
+	}
+}
+
+func (manager *Manager) handlePolicyEvent(event resource.Event[*Policy]) {
```
markpash, line 320: Nit: this looks like it could just be merged into the above function.

lmb: I split it out since `processEvents` is going to get more complicated as we migrate away from the monolithic k8s watchers. AFAIK we'll add at least Endpoint and Node. So I'd prefer to keep them split.

pkg/egressgateway/manager_privileged_test.go
lmb: Ah good catch. Maybe it should even be after `LinkAdd`?

lmb: NVM, LinkDel needs the interface index.

pkg/egressgateway/manager.go
```diff
@@ -353,23 +403,25 @@ func (manager *Manager) OnAddEgressPolicy(config PolicyConfig) {
...
+	configID := ParseCEGPConfigID(policy)
+
 	manager.Lock()
 	defer manager.Unlock()
 
 	logger := log.WithField(logfields.CiliumEgressGatewayPolicyName, configID.Name)
 
 	if manager.policyConfigs[configID] == nil {
 		logger.Warn("Can't delete CiliumEgressGatewayPolicy: policy not found")
-		return
```
jibi, line 372: why do we need to go on with the reconciliation?

pkg/egressgateway/manager.go
```diff
@@ -267,20 +273,59 @@ func (manager *Manager) getIdentityLabels(securityIdentity uint32) (labels.Label
 	return identity.Labels, nil
 }
 
-// runReconciliationAfterK8sSync spawns a goroutine that waits for the agent to
+// processEvents spawns a goroutine that waits for the agent to
```
jibi, line 276: nit: comment is outdated

pkg/egressgateway/manager.go
```diff
@@ -267,20 +273,59 @@ func (manager *Manager) getIdentityLabels(securityIdentity uint32) (labels.Label
...
 // sync with k8s and then runs the first reconciliation.
-func (manager *Manager) runReconciliationAfterK8sSync(ctx context.Context) {
-	go func() {
-		select {
-		case <-manager.cacheStatus:
-			manager.Lock()
-			manager.setEventBitmap(eventK8sSyncDone)
-			manager.Unlock()
+func (manager *Manager) processEvents(ctx context.Context, cacheStatus k8s.CacheStatus) {
+	var globalSync, policySync bool
```
jibi, line 279: maybe it's just me, but it took me a bit to understand how `globalSync`, `policySync` and `allCachesSynced` fit together.

Could I suggest to simplify this a bit like:
* move `globalSync` and `policySync` inside the manager
* make `allCachesSynced` a method that just checks `globalSync && policySync`

(yes, after this we need to grab and release the manager lock a couple more times to set these 2 variables, but imo it makes it overall easier to follow the logic).

What do you think?

lmb: That changes the semantics a little bit: `allCachesSynced` is also used to keep track of whether we've triggered the first reconciliation. Removing the separate boolean means that additional `Sync` events will trigger additional reconciliation. Not sure if it's even possible to receive multiple Sync, but the API doesn't make any guarantees about it either way.

pkg/egressgateway/manager.go
lmb: Oops!

pkg/egressgateway/manager.go
jibi: > allCachesSynced is also used to keep track of whether we've triggered the first reconciliation.

looks like we could also decouple this and make it more explicit

> Not sure if it's even possible to receive multiple Sync, but the API doesn't make any guarantees about it either way.

fair, I assumed it wouldn't  be the case :+1: 

anyway, I think this PR is already good as it is and it's not worth blocking it on this, we can always try to simplify this logic in the future

jibi (approved):

youngnick (approved):

PR #26962, opened 2023-07-20
cilium node chain refactor

bimmlerd:
The intent of this PR is to remove the CiliumNodeChain.

Review commit by commit.

bgp/speaker: use Resource[T] instead of NodeChains
    
    As part of a refactoring effort to remove the CiliumNodeChain, remove
    the dependency of the MetalLBSpeaker on it, by directly using the
    LocalCiliumNodeResource.
    
    Note that the speaker always discarded events not pertaining to the
    local Node, hence no functional change results from the fact that the
    CiliumNodeChain proagated events from all cluster CiliumNodes.
    
    As an added benefit, the CiliumNode events now don't stop coming once we
    connect to KVStore, since the resource doesn't stop.
   

ipam: multipool: use resource for node events

```
As part of a refactoring effort to remove the CiliumNodeChain, remove
the multipool IPAM dependency on it.

Since IPAM isn't yet part of the hive, we need to pass the resource down
to IPAM through the daemon, in `startIPAM`. Once IPAM code is
modularised, this is no longer necessary.

In order to keep the tests largely unchanged, we reimplement the
fakeNodeK8sAPI testing infrastructure in terms of the resource API. This
leads to code duplication, but a commit later in this series removes the
duplication again, by also converting the other usages to the resource style.
```

watchers: remove ciliumnodechain

    Previous commits in the series have removed all subscribers to the
    CiliumNodeChain by moving them over to use the LocalCiliumNodeResource.
    This commit reaps the benefits by deleting all of the now unused code.


pkg/k8s/watchers/subscriber/ciliumnode.go
```diff
@@ -49,46 +49,31 @@ func (l *CiliumNodeChain) Register(s CiliumNode) {
 func (l *CiliumNodeChain) OnAddCiliumNode(node *ciliumv2.CiliumNode) error {
 	l.RLock()
 	defer l.RUnlock()
-	errs := []error{}
+	var err error
 	for _, s := range l.subs {
-		if err := s.OnAddCiliumNode(node); err != nil {
-			errs = append(errs, err)
-		}
+		err = errors.Join(s.OnAddCiliumNode(node))
```
joamaki: Doesn't this only return the error from last subscriber since we overwrite `err`?
`err = errors.Join(err, s.OnAddCiliumNode(node))` would probably not work since we get nested `joinError`s. Perhaps just do the same as the original except use `errors.Join(errs...)` instead of `fmt.Errorf`.

pkg/k8s/watchers/subscriber/ciliumnode.go
```diff
@@ -49,46 +49,31 @@ func (l *CiliumNodeChain) Register(s CiliumNode) {
...
-	errs := []error{}
+	var err error
 	for _, s := range l.subs {
-		if err := s.OnUpdateCiliumNode(oldNode, newNode); err != nil {
-			errs = append(errs, err)
-		}
-	}
-	if len(errs) > 0 {
-		return fmt.Errorf("Errors: %v", errs)
+		err = errors.Join(s.OnUpdateCiliumNode(oldNode, newNode))
```
joamaki: Same issue as above.

pkg/k8s/watchers/subscriber/ciliumnode.go
```diff
@@ -49,46 +49,31 @@ func (l *CiliumNodeChain) Register(s CiliumNode) {
...
-	errs := []error{}
+	var err error
 	for _, s := range l.subs {
-		if err := s.OnDeleteCiliumNode(node); err != nil {
-			errs = append(errs, err)
-		}
-	}
-	if len(errs) > 0 {
-		return fmt.Errorf("Errors: %v", errs)
+		err = errors.Join(s.OnDeleteCiliumNode(node))
```
joamaki: Same issue as above.

pkg/k8s/watchers/subscriber/ciliumnode.go
```diff
@@ -28,19 +34,49 @@ type CiliumNode interface {
 type CiliumNodeChain struct {
```
joamaki: Should we just rename this to `LocalCiliumNodeChain` so there won't be confusion in case refactoring takes a bit longer?

joamaki (changes requested):
~~The approach LGTM. Fix the error joining :)~~

The approach is flawed as we're now retrying an event that single subscriber fails to process on all of the subscribers in CiliumNodeChain. I'm not sure they're idempotent nor would it be the behavior we want. See my last comment.

pkg/k8s/watchers/subscriber/ciliumnode.go
```diff
@@ -28,19 +34,49 @@ type CiliumNode interface {
...
+				err = l.OnAddCiliumNode(node)
+			} else {
+				err = l.OnUpdateCiliumNode(existing, node)
+			}
+			existing = node
+		case resource.Delete:
+			err = l.OnDeleteCiliumNode(ev.Object)
+		}
+
+		ev.Done(err)
```
joamaki: This will now retry the event for all subscribers even if only one of them failed. I'm not sure we can rely on them being idempotent. It might be better to keep the original semantics of not doing much with the error (it was just marking up the EventProcessed metric). Though then we won't have `Resource[T]` marking up the metric and will lose metric coverage :(

Take a look at the subscribers and see what options we have.

bimmlerd: Thanks for this catch. Rewrote the PR to drop CiliumNodeChain completely

daemon/cmd/ipam.go
```diff
@@ -527,11 +528,11 @@ func (d *Daemon) configureIPAM() {
 	}
 }
 
-func (d *Daemon) startIPAM() {
+func (d *Daemon) startIPAM(nodes agentK8s.LocalCiliumNodeResource) {
```
gandro: nit: Maybe we should call this `node`

pkg/ipam/multipool.go
```diff
@@ -239,14 +240,32 @@ func newMultiPoolManager(conf Configuration, nodeWatcher nodeWatcher, owner Owne
...
 
+func (m *multiPoolManager) eventloop(res agentK8s.LocalCiliumNodeResource) {
+	// Don't retry events - the downstream code isn't setup to handle retries.
+	// Ensure that the subscrption happens synchronously (but process async), so
+	// that there's no window in which we may miss events.
+	evs := res.Events(context.TODO(), resource.WithErrorHandler(resource.RetryUpTo(0)))
+	go func() {
+		for ev := range evs {
+			switch ev.Kind {
+			case resource.Upsert:
```
gandro: Question: Does the resource watcher issue an event if the `resourceVersion` has changed, but the object is otherwise semantically equivalent? The current IPAM code relies on this (though this can be changed if the nodeStore gives us the most recent version)

gandro (approved):
Thanks so much for tackling this. One question around upsert events, otherwise this looks excellent to me

rastislavs (approved):
LGTM for BGP

pkg/ipam/multipool.go
bimmlerd: Yes.

Resource[T] doesn't do any deep equal stuff. The only deduplication that happens is if an event hasn't been observed, and a newer event regarding the same object comes in, you only get one event. This isn't resource code, this is just the k8s workqueue implementation. Since `Resource[T].Events()` fetches the newest state for an object (by key) from the store for each event it delivers anyway, this just means you don't get two identical events.

pkg/bgp/speaker/speaker.go
```diff
@@ -247,43 +247,53 @@ func (s *MetalLBSpeaker) notifyNodeEvent(op Op, nodeMeta metaGetter, podCIDRs *[
...
-// OnDeleteCiliumNode notifies the Speaker of a CiliumNode deletion.
-//
-// When the speaker discovers the node that it is running on
-// is shuttig down it will send a BGP message to its peer
-// instructing it to withdrawal all previously advertised
-// routes.
-func (s *MetalLBSpeaker) OnDeleteCiliumNode(node *ciliumv2.CiliumNode, swg *lock.StoppableWaitGroup) error {
-	return s.notifyNodeEvent(Delete, node, ciliumNodePodCIDRs(node), true)
+func eventLoop[T metaGetterObject](s *MetalLBSpeaker, res resource.Resource[T], nodePodCIDRs func(T) *[]string) {
+	ctx, cancel := context.WithCancel(context.Background())
```
joamaki: shouldn't this use the context passed to `New`? Otherwise only way to stop here is to stop the resource.

pkg/ipam/multipool.go
```diff
@@ -239,14 +240,32 @@ func newMultiPoolManager(conf Configuration, nodeWatcher nodeWatcher, owner Owne
...
 	c.waitForAllPools()
 
 	return c
 }
 
+func (m *multiPoolManager) eventloop(res agentK8s.LocalCiliumNodeResource) {
+	// Don't retry events - the downstream code isn't setup to handle retries.
+	// Ensure that the subscrption happens synchronously (but process async), so
+	// that there's no window in which we may miss events.
+	evs := res.Events(context.TODO(), resource.WithErrorHandler(resource.RetryUpTo(0)))
```
joamaki: Do we have a proper context to use here that is cancelled when IPAM stops? This might be dangerous as Resource[T] might be emitting events while systems that this is calling into are being stopped.

pkg/ipam/clusterpool.go
```diff
@@ -154,12 +155,31 @@ func newCRDWatcher(conf Configuration, nodeWatcher nodeWatcher, owner Owner, nod
...
+	c.eventloop(nodes)
 	owner.UpdateCiliumNodeResource()
 
 	return c
 }
 
+func (c *crdWatcher) eventloop(res agentK8s.LocalCiliumNodeResource) {
+	// Ensure that the subscrption happens synchronously (but processing
+	// asynchronously), so that there's no window in which we miss events.
+	// FIXME is that actually a race?
```
joamaki: Please address the FIXME

pkg/ipam/clusterpool.go
```diff
@@ -154,12 +155,31 @@ func newCRDWatcher(conf Configuration, nodeWatcher nodeWatcher, owner Owner, nod
...
+	// Don't retry events - the downstream code isn't setup to handle retries.
+	evs := res.Events(context.TODO(), resource.WithErrorHandler(resource.RetryUpTo(0)))
```
joamaki: Same here as in previous comments: use a proper context that is cancelled when this sub-system is stopping. Or if there is no such thing, validate that it's fine that this only stops when the `Resource[T]` stops and it's fine handling events from it while everything is being stopped & add comment to explain this.

joamaki (changes requested):

pkg/bgp/speaker/speaker.go
```diff
@@ -16,12 +16,15 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
 
+	k8sRuntime "k8s.io/apimachinery/pkg/runtime"
+
+	agentK8s "github.com/cilium/cilium/daemon/k8s"
```
danehans, line 21: nit: For consistency, consider using `_` for named imports, e.g. `agent_k8s`.

daemon/cmd/ipam.go
```diff
@@ -14,6 +14,7 @@ import (
 
 	"github.com/cilium/cilium/api/v1/models"
 	ipamapi "github.com/cilium/cilium/api/v1/server/restapi/ipam"
+	agentK8s "github.com/cilium/cilium/daemon/k8s"
```
danehans, line 17: nit: For consistency, named imports should use underscores.

pkg/ipam/multipool.go
```diff
@@ -15,11 +15,12 @@ import (
 	"golang.org/x/exp/slices"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	agentK8s "github.com/cilium/cilium/daemon/k8s"
```
danehans, line 18: Same feedback as https://github.com/cilium/cilium/pull/26962/commits/d288a7590c0d2c6e198457c39c7974366f823a9a#r1307537605.

pkg/ipam/multipool.go
```diff
@@ -205,7 +206,7 @@ type multiPoolManager struct {
 
 var _ Allocator = (*multiPoolAllocator)(nil)
 
-func newMultiPoolManager(conf Configuration, nodeWatcher nodeWatcher, owner Owner, clientset nodeUpdater) *multiPoolManager {
+func newMultiPoolManager(conf Configuration, nodes agentK8s.LocalCiliumNodeResource, owner Owner, clientset nodeUpdater) *multiPoolManager {
```
danehans: s/nodes/node/

pkg/ipam/clusterpool.go
bimmlerd: clusterpool was removed

pkg/ipam/clusterpool.go
bimmlerd: clusterpool was removed

pkg/bgp/speaker/speaker.go
bimmlerd: I've chosen to be consistent with the above, `k8sRuntime` instead. We don't seem to have consistency in the codebase here, hence unclear which would be correct.

pkg/bgp/speaker/speaker.go
bimmlerd: It's the daemon context which ends up being passed in - changes such that it is passed to the subscription too.

pkg/ipam/multipool.go
bimmlerd: We do not. Previous code was getting called via the k8s informer, hence I believe this to be equivalent. Added a comment to that effect.

bimmlerd:
Finally got around to updating this. Significant changes since last review:

1. clusterpool has been dropped upstream, hence dropping changes to it in this PR.
2. improved context handling in the speaker, and added an explanation which in IPAM there isn't a signal available to stop.

PTAL.

gandro (approved):
Looks good to me from my code owners! Thanks a lot for taking care of this!

rastislavs (approved):
LGTM

pkg/bgp/speaker/speaker.go
```diff
@@ -247,43 +247,53 @@ func (s *MetalLBSpeaker) notifyNodeEvent(op Op, nodeMeta metaGetter, podCIDRs *[
...
+			existing = &node
+		case resource.Delete:
+			// When the speaker discovers the node that it is running on
+			// is shuttig down it will send a BGP message to its peer
+			// instructing it to withdrawal all previously advertised
+			// routes.
+			node := ev.Object
+			err = s.notifyNodeEvent(Delete, node, nodePodCIDRs(node), true)
+		}
+		ev.Done(err)
```
joamaki, line 290: Resource[T] doesn't log failed events (as it doesn't have enough context, so it's left to subscribers). Could you check whether how this now differs from K8sWatcher and whether logging behavior changed?

pkg/ipam/multipool.go
```diff
@@ -251,14 +247,39 @@ func newMultiPoolManager(conf Configuration, nodeWatcher nodeWatcher, owner Owne
 	}
 
 	// Subscribe to CiliumNode updates
-	nodeWatcher.RegisterCiliumNodeSubscriber(c)
+	c.eventloop(node)
```
joamaki: nit: Perhaps call this `startCiliumNodeEventLoop` so it's clear it's not blocking and what it's handling?

pkg/ipam/multipool.go
joamaki: Thanks!

joamaki (approved):

bimmlerd:
want to address Jussi's nits before merging

pkg/bgp/speaker/speaker.go
bimmlerd: The logging behaviour did not change AFAICT - cilium node chain did not log by itself; and the metric of k8s event success is baked into resource, so no change there.

danehans:
@bimmlerd the ipsec upgrade job is [failing](https://github.com/cilium/cilium/actions/runs/6393490426/job/17352832845#step:16:123) due to this error:

```sh
Unable to contact Hubble Relay, disabling Hubble telescope and flow validation: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp: lookup localhost on 1.1.1.1:53: no such host"
```

I don't understand why Cilium Agent is using a DNS server to resolve `localhost`. It appears to be related to [this](https://github.com/cilium/cilium/issues/27799) flake.

bimmlerd:
> @bimmlerd the ipsec upgrade job is [failing](https://github.com/cilium/cilium/actions/runs/6393490426/job/17352832845#step:16:123) due to this error:
> 
> ```shell
> Unable to contact Hubble Relay, disabling Hubble telescope and flow validation: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp: lookup localhost on 1.1.1.1:53: no such host"
> ```
> 
> I don't understand why Cilium Agent is using a DNS server to resolve `localhost`. It appears to be related to [this](https://github.com/cilium/cilium/issues/27799) flake.

I don't think so. AFAICT it's failing because of `  🟥 Pod test-conn-disrupt-client-54df7db847-nqmzw flow was interrupted (restart count does not match 2 != 3)`
In any case, this isn't a required test, which is typically a strong signal that the test is either still in development or was deemed too flaky. The new failure is different than the two before that, so I'm not considering it as a signal of breakage.

PR #26963, opened 2023-07-20
ipam: Fix race in NodeManager.Resync

jaffcheng:
please see commit message

Related: #26617

### DIFF
--- a/pkg/ipam/node_manager.go
+++ b/pkg/ipam/node_manager.go
@@ -417,19 +417,18 @@ func (n *NodeManager) Get(nodeName string) *Node {
 	return node
 }
 
-// GetNodesByIPWatermark returns all nodes that require addresses to be
+// GetNodesByIPWatermarkLocked returns all nodes that require addresses to be
 // allocated or released, sorted by the number of addresses needed to be operated
 // in descending order. Number of addresses to be released is negative value
 // so that nodes with IP deficit are resolved first
-func (n *NodeManager) GetNodesByIPWatermark() []*Node {
-	n.mutex.RLock()
+// The caller must hold the NodeManager lock
+func (n *NodeManager) GetNodesByIPWatermarkLocked() []*Node {
 	list := make([]*Node, len(n.nodes))
 	index := 0
 	for _, node := range n.nodes {
 		list[index] = node
 		index++
 	}
-	n.mutex.RUnlock()
 
 	sort.Slice(list, func(i, j int) bool {
 		valuei := list[i].GetNeededAddresses()
@@ -509,12 +508,14 @@ func (n *NodeManager) resyncNode(ctx context.Context, node *Node, stats *resyncS
 // watermarks. Any updates to the node resource are synchronized to the
 // Kubernetes apiserver.
 func (n *NodeManager) Resync(ctx context.Context, syncTime time.Time) {
+	n.mutex.Lock()
+	defer n.mutex.Unlock()
 	n.metricsAPI.IncResyncCount()
 
 	stats := resyncStats{}
 	sem := semaphore.NewWeighted(n.parallelWorkers)
 
-	for _, node := range n.GetNodesByIPWatermark() {
+	for _, node := range n.GetNodesByIPWatermarkLocked() {
 		err := sem.Acquire(ctx, 1)
 		if err != nil {
 			continue
--- a/pkg/ipam/node_manager_test.go
+++ b/pkg/ipam/node_manager_test.go
@@ -25,8 +25,7 @@ import (
 )
 
 var (
-	k8sapi     = &k8sMock{}
-	metricsapi = metricsmock.NewMockMetrics()
+	k8sapi = &k8sMock{}
 )
 
 const testPoolID = ipamTypes.PoolID("global")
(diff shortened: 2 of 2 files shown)


jaffcheng:
@tommyp1ckles @christarazi Please take a look. This flake clearly happens more often lately, but I'm not quite sure why recent changes would introduce this.

tommyp1ckles:
@jaffcheng Interesting, I've been looking into this as well,  I'm pretty sure its related to racey code related to appending to the AzureInterface Addresses array in the mock API implementation.

I have a local reproduction that can cause the failure to occur fairly quickly, let me validate if this fixes it.

tommyp1ckles:
Even with these changes, I stil run into the failure.

jaffcheng:
> Even with these changes, I stil run into the failure.

Thanks for trying, could you share your reproduction method and test output? My local test is running pkg/ipam unit tests a lot of times with/without -race and lockdebug. But haven’t tried pkg/azure yet.

tommyp1ckles:
@jaffcheng Sure, my branch is quite messy, ill push that up later today.  The main part is here in the Upsert function in node_manager.go. Basically just removing the exponential backoff between triggers and running the poolMaintainer trigger in a loop causes this to occur usually in a couple seconds.  I'm running the test with

```
go test -count=100 -v -failfast -race
```

```go
		poolMaintainer, err := trigger.NewTrigger(trigger.Parameters{
			Name:            fmt.Sprintf("ipam-pool-maintainer-%s", resource.Name),
			MinInterval:     10 * time.Millisecond,
			MetricsObserver: n.metricsAPI.PoolMaintainerTrigger(),
			TriggerFunc: func(reasons []string) {
				if err := node.MaintainIPPool(ctx); err != nil {
					node.logger().WithError(err).Warning("Unable to maintain ip pool of node")
					//backoff.Wait(ctx) Remove exponential backoff wait to increase chance of race occuring.
				}
			},
			ShutdownFunc: cancel,
		})
                // Trigger poolMaintainer continuously!
		go func() {
			for {
				<-time.After(time.Millisecond)
				poolMaintainer.Trigger()
			}
		}()
```

tommyp1ckles:
Race reveals a couple data races in this,  I was hoping fixing those would solve the problem but yet it still persists 😕

jaffcheng:
Although I think this patch should significantly lower the possibility of flakes, there are still some races found in /pkg/ipam with -race on. Need more digging...

tommyp1ckles:
> Although I think this patch should significantly lower the possibility of flakes, there are still some races found in /pkg/ipam with -race on. Need more digging...

there appears to be many races there 😄 I have a few race fixes as well I'll put PRs for.  Considering how hard im finding it to get to a single cause, its possible the issue is caused by multiple races.  We can start fixing these and see if the flakes improve.

jaffcheng:
Change `Fixes:` to `Related:` since this does not fix all races.

pkg/ipam/node_manager.go
```diff
@@ -418,15 +418,14 @@ func (n *NodeManager) Get(nodeName string) *Node {
 // allocated or released, sorted by the number of addresses needed to be operated
 // in descending order. Number of addresses to be released is negative value
 // so that nodes with IP deficit are resolved first
+// The caller must hold the NodeManager lock
 func (n *NodeManager) GetNodesByIPWatermark() []*Node {
```
tommyp1ckles: In such cases, the convention is to call the function xxxLocked(...) instead of just having a comment.

ex. https://github.com/tommyp1ckles/cilium/blob/main/pkg/aws/eni/node.go#L92

tommyp1ckles:
When I was looking at this, another place I thought it might be racing was here: https://github.com/cilium/cilium/blob/main/pkg/ipam/node.go#L430

i.e. the interfaces are checked in parallel and then on the next line they will race to get the Node lock, so in theory a stale set of values could overwrite the latest ones.

But... fixing this didn't fix the flake in practice. 

I think these changes make sense, lock changes in this code scare me a bit, I'll do another review to double check that we aren't creating a deadlock somewhere in shortly.

tommyp1ckles:
I think these changes make sense, lock changes in this code scare me a bit, I'll do another review to double check that we aren't creating a deadlock somewhere in shortly.

tommyp1ckles (approved):

pkg/ipam/node_manager.go
```diff
@@ -418,15 +418,14 @@ func (n *NodeManager) Get(nodeName string) *Node {
...
-	n.mutex.RLock()
```
christarazi, line 422: Why is it safe to drop this lock?

christarazi, line 425: Why is it safe to drop this lock?

christarazi (dismissed):

pkg/ipam/node_manager.go
tommyp1ckles: This is only called from nodemanager.Resync: https://github.com/cilium/cilium/pull/26963/files#diff-28c460f7ee57d8ce6d40f9b55d3e627821a0e9158d7089e526fc75138b45096eR508

So the lock is being moved up to there.

jaffcheng:
Changed to `GetNodesByIPWatermarkLocked` and slightly changed the commit message given that this only fixes one source of the flake.

tommyp1ckles:
@jaffcheng The integration tests are failing I believe to CI using a newer version of Go,  can you rebase your branch - that should fix it.

joestringer:
What's the consequence of this bugfix? Should we backport it to v1.14 per the [backport criteria](https://docs.cilium.io/en/stable/contributing/release/backports/#id1)? Does the race condition affect other branches too?

joestringer:
Reviews are in, CI is passing. LGTM to merge.

tommyp1ckles:
@joestringer The changes that causes this don't appear to be in the v1.14 branch

PR #26973, opened 2023-07-21
egressgw: always set ifaceName in deriveFromPolicyGatewayConfig()

julianwiedmann:
Make sure that we always have a fully populated `gatewayconfig` when `deriveFromPolicyGatewayConfig()` returns.


julianwiedmann:
(looks like we don't actually use this right now :grin:. Stumbled over this while trying to make use of the field)

jibi (approved):

PR #26975, opened 2023-07-21
ci: increase connectivity test timeout in GHA external workload

mhofstetter:
This commit increases the timeout of the job `Installation and Connectivity Test` within the worklfow `External Workload` from 30m to 45m.

The current timeout is too low:

> Installation and Connectivity Test (1.26, europe-north1-b, 5) 
> The job running on runner GitHub Actions 52 has exceeded the maximum execution time of 30 minutes.`

https://github.com/cilium/cilium/actions/runs/5604532568


nbusseneau (approved):

nbusseneau:
What's funny here is that it's always this specific combo that times out, are the instances in this region slower? :p

PR #26976, opened 2023-07-21
ci: replace GHA action Sibz/github-status-action

mhofstetter:
This commit replaces the GitHub Action `Sibz/github-status-action` to set the commit status with `myrotvorets/set-commit-status-action`.

Sibz/github-status-action is only used in the Helm Chart Push and builds on top of node12 which is deprecated in GHA.

> The following actions uses node12 which is deprecated and will be forced to run on node16:
Sibz/github-status-action@650dd1a882a76dbbbc4576fb5974b8d22f29847f. For more info:
https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

Example workflow run with deprecation warning: https://github.com/cilium/cilium/actions/runs/5619983977

### DIFF
--- a/.github/workflows/push-chart-ci.yaml
+++ b/.github/workflows/push-chart-ci.yaml
@@ -27,7 +27,7 @@ permissions:
   contents: read
   # To allow retrieving information from the PR API
   pull-requests: read
-  # So that Sibz/github-status-action can write into the status API
+  # To be able to set commit status
   statuses: write
 
 concurrency:
@@ -37,9 +37,6 @@ concurrency:
       (github.event_name == 'workflow_run' && github.event.workflow_run.head_sha)
     }}
   cancel-in-progress: true
-
-env:
-  check_url: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
 
 jobs:
   push-charts:
@@ -79,14 +76,11 @@ jobs:
         fi
 
     - name: Set commit status to pending
-      uses: Sibz/github-status-action@650dd1a882a76dbbbc4576fb5974b8d22f29847f # v1.1.6
+      uses: myrotvorets/set-commit-status-action@243b4f7e597f62335408d58001edf8a02cf3e1fd # v1.1.7
       with:
-        authToken: ${{ secrets.GITHUB_TOKEN }}
         sha: ${{ steps.get-ref.outputs.sha }}
-        context: ${{ github.workflow }}
+        status: pending
         description: Helm push in progress
-        state: pending
-        target_url: ${{ env.check_url }}
 
     - name: Checkout Source Code
       uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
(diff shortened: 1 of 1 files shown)


nbusseneau (changes requested):
Can we take this opportunity to rework the jobs as well? `myrotvorets/set-commit-status-action` allows us to do it in a simpler way, like we do in other workflows:

```
  commit-status-start:
    runs-on: ubuntu-latest
    steps:
      - name: Set initial commit status
        uses: myrotvorets/set-commit-status-action@243b4f7e597f62335408d58001edf8a02cf3e1fd # v1.1.7  
        with:
          sha: ${{ steps.get-ref.outputs.sha }}

[main job `foo` goes here...]

  commit-status-final:
    if: ${{ always() }}
    needs: foo
    runs-on: ubuntu-latest
    steps:
      - name: Set final commit status
        uses: myrotvorets/set-commit-status-action@243b4f7e597f62335408d58001edf8a02cf3e1fd # v1.1.7  
        with:
          sha: ${{ steps.get-ref.outputs.sha }}
          status: ${{ needs.foo.result }}
```

mhofstetter:
> Can we take this opportunity to rework the jobs as well? `myrotvorets/set-commit-status-action` allows us to do it in a simpler way, like we do in other workflows:
> 
> ```
>   commit-status-start:
>     runs-on: ubuntu-latest
>     steps:
>       - name: Set initial commit status
>         uses: myrotvorets/set-commit-status-action@243b4f7e597f62335408d58001edf8a02cf3e1fd # v1.1.7  
>         with:
>           sha: ${{ steps.get-ref.outputs.sha }}
> 
> [main job `foo` goes here...]
> 
>   commit-status-final:
>     if: ${{ always() }}
>     needs: foo
>     runs-on: ubuntu-latest
>     steps:
>       - name: Set final commit status
>         uses: myrotvorets/set-commit-status-action@243b4f7e597f62335408d58001edf8a02cf3e1fd # v1.1.7  
>         with:
>           sha: ${{ steps.get-ref.outputs.sha }}
>           status: ${{ needs.foo.result }}
> ```

good idea, it's just that the step `get-ref` is required by steps from all 3 resulting jobs `commit-status-start`, `<main>` & `commit-status-final`. :/ 

Before doing that: Actually I don't really understand why this workflow isn't just a simple `push` event based one - that doesn't support `workflow_run` & `workflow_dispatch` :man_shrugging:  :thinking:

nbusseneau:
Yea I just did a draft by hand but of course add any other job dependencies, then you can access that jobs' outputs in the dependencies.

mhofstetter:
> Yea I just did a draft by hand but of course add any other job dependencies, then you can access that job's outputs in the dependent jobs.

If it's ok I might come back to that later. Given the bigger restructuring and resulted testing (especially the manual cases) ->  would it be possible to merge this as is (with the goal to get rid of the deprecation warning). (It's my last day on CI-Health duty...)

nbusseneau (approved):

PR #26979, opened 2023-07-21
docs/ipsec: Document RSS limitation

pchaigno:
All IPsec traffic between two nodes is always send on a single IPsec flow (defined by outer source and destination IP addresses). As a consequence, RSS on such traffic is ineffective and throughput will be limited to the decryption performance of a single core.

### DIFF
--- a/Documentation/security/network/encryption-ipsec.rst
+++ b/Documentation/security/network/encryption-ipsec.rst
@@ -272,3 +272,6 @@ Limitations
     * IPsec encryption is not currently supported in combination with IPv6-only clusters.
     * IPsec encryption is not supported on clusters or clustermeshes with more
       than 65535 nodes.
+    * Decryption with Cilium IPsec is limited to a single CPU core per IPsec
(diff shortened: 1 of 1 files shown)


Documentation/security/network/encryption-ipsec.rst
```diff
@@ -272,3 +272,6 @@ Limitations
     * IPsec encryption is not currently supported in combination with IPv6-only clusters.
     * IPsec encryption is not supported on clusters or clustermeshes with more
       than 65535 nodes.
+    * Decryption with Cilium IPsec is limited to a single CPU core per IPsec
+      tunnel. That may affect performance in case of high throughput between
```
learnitall: One nit: `That` -> `This`. My apologies for missing this beforehand.

learnitall (approved):
One nit I saw when re-reading (sorry about that), but LGTM. Thanks @pchaigno!

gentoo-root (approved):

PR #26981, opened 2023-07-21
Temporarily disable bpf clock probe

### LINKED ISSUE #26955: Disable --enable-bpf-clock-probe
When doing Cilium upgrade from v1.13.3 to v1.13.4 we noticed drops in long-lived connections to a ClusterIP (the LB done in bpf_lxc). Closer look revealed that the valid service CT entry was removed during the upgrade leading to the interruptions.

The notable change between v1.13.3 and v1.13.4 was https://github.com/cilium/cilium/pull/26197 which influences the CT GC decisions. Disabling `--enable-bpf-clock-probe` resolved the drops.

The `--enable-bpf-clock-probe` got implicitly enabled by https://github.com/cilium/cilium/pull/25795 (previously, `/proc/timer_list` was not accessible to the cilium-agent pods, so the jiffies were disabled).

As a temp fix we should disable `--enable-bpf-clock-probe`. It gets enabled here https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/templates/cilium-configmap.yaml#L22.

The upgrade path from v1.13.3 to upcoming v1.14.0 is affected too.

cc @ti-mo

margamanterola:
When enabling the bpf clock probe, the garbage collector is sometimes removing valid conntrack entries, likely due to some error in the jiffies calculation.

Until we figure out a permanent solution, let's disable the the bpf-clock-probe by default.

Fixes: #26955

```release-note
Temporarily disable bpf-clock-probe to avoid causing interruptions for long-lived connections during upgrades
```

### DIFF
--- a/install/kubernetes/cilium/values.yaml.tmpl
+++ b/install/kubernetes/cilium/values.yaml.tmpl
@@ -421,9 +421,6 @@ bpf:
   # -- Configure the mount point for the BPF filesystem
   root: /sys/fs/bpf
 
-  # -- Enable BPF clock source probing for more efficient tick retrieval.
-  clockProbe: false
-
   # -- Enables pre-allocation of eBPF map values. This increases
   # memory usage but can reduce latency.
   preallocateMaps: false
@@ -497,6 +494,9 @@ bpf:
   # [0] will allow all VLAN id's without any filtering.
   # @default -- `[]`
   vlanBypass: ~
+
+# -- Enable BPF clock source probing for more efficient tick retrieval.
+bpfClockProbe: false
 
 # -- Clean all eBPF datapath state from the initContainer of the cilium-agent
 # DaemonSet.
--- a/Documentation/helm-values.rst
+++ b/Documentation/helm-values.rst
@@ -224,10 +224,6 @@
      - Enable automatic mount of BPF filesystem When ``autoMount`` is enabled, the BPF filesystem is mounted at ``bpf.root`` path on the underlying host and inside the cilium agent pod. If users disable ``autoMount``\ , it's expected that users have mounted bpffs filesystem at the specified ``bpf.root`` volume, and then the volume will be mounted inside the cilium agent pod at the same path.
      - bool
      - ``true``
-   * - :spelling:ignore:`bpf.clockProbe`
-     - Enable BPF clock source probing for more efficient tick retrieval.
-     - bool
-     - ``false``
    * - :spelling:ignore:`bpf.ctAnyMax`
      - Configure the maximum number of entries for the non-TCP connection tracking table.
      - int
@@ -296,6 +292,10 @@
      - Configure explicitly allowed VLAN id's for bpf logic bypass. [0] will allow all VLAN id's without any filtering.
      - list
      - ``[]``
+   * - :spelling:ignore:`bpfClockProbe`
+     - Enable BPF clock source probing for more efficient tick retrieval.
+     - bool
+     - ``false``
    * - :spelling:ignore:`certgen`
      - Configure certificate generation for Hubble integration. If hubble.tls.auto.method=cronJob, these values are used for the Kubernetes CronJob which will be scheduled regularly to (re)generate any certificates not provided manually.
      - object
--- a/install/kubernetes/cilium/README.md
+++ b/install/kubernetes/cilium/README.md
@@ -106,7 +106,6 @@ contributors across the globe, there is almost always someone available to help.
 | bgpControlPlane.enabled | bool | `false` | Enables the BGP control plane. |
 | bpf.authMapMax | int | `524288` | Configure the maximum number of entries in auth map. |
 | bpf.autoMount.enabled | bool | `true` | Enable automatic mount of BPF filesystem When `autoMount` is enabled, the BPF filesystem is mounted at `bpf.root` path on the underlying host and inside the cilium agent pod. If users disable `autoMount`, it's expected that users have mounted bpffs filesystem at the specified `bpf.root` volume, and then the volume will be mounted inside the cilium agent pod at the same path. |
-| bpf.clockProbe | bool | `false` | Enable BPF clock source probing for more efficient tick retrieval. |
 | bpf.ctAnyMax | int | `262144` | Configure the maximum number of entries for the non-TCP connection tracking table. |
 | bpf.ctTcpMax | int | `524288` | Configure the maximum number of entries in the TCP connection tracking table. |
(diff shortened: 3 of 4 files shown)


install/kubernetes/cilium/templates/cilium-configmap.yaml
```diff
@@ -19,7 +19,7 @@
   {{- $defaultEnableCnpStatusUpdates = "false" -}}
   {{- $defaultBpfMapDynamicSizeRatio = 0.0025 -}}
   {{- $defaultBpfMasquerade = "true" -}}
-  {{- $defaultBpfClockProbe = "true" -}}
+  {{- $defaultBpfClockProbe = "false" -}}
```
brb: Not sure whether it's the right place to change it :-/ Maybe someone from the @cilium/helm could review?

squeed: Yes, this is correct; in the daemon itself it defaults to false.

squeed (approved):

install/kubernetes/cilium/templates/cilium-configmap.yaml
pchaigno: I think Martynas wasn't asking Helm vs. agent, but this section of the Helm template vs. the section above.

Won't a change of time measurement for existing installations cause all existing connections to be interrupted? I would expect that's why we have this here to ensure the setting is preserved is the cluster was initially deployed with it, no?

gandro: Yeah this logic here is very confusing. The way this `upgradeCompatibility`  section here works ([docs here](https://docs.cilium.io/en/stable/operations/upgrade/#step-2-use-helm-to-upgrade-your-cilium-deployment)) is that we have the old pre-v1.8 default in the top of the file, and then for each version where we change the default value, we add a new section.

So this section here is for default values changed in v1.8 (where we apparently turned bpf clock probe on by default). So I would argue it's not the right place to change. 

If we just want to turn it off by default again for new installations in v1.14+, we probably should add a v1.14 section where we do `$defaultBpfClockProbe = "false"`. If people then explicitly declare they want to use v1.12 default values, then bpf clock probe would still be enabled, as it was in v1.12.

If, on the other hand, we want to turn of bpf clock probe even for existing installations, then we should probably add it to the `values.yaml` and document this in the upgrade impact section of the docs. That way, users can set it back to `true` if they know what they are doing.

To summarize:
- If we want to turn of bpf clock probe forcefully *for existing installations, regardless of `upgradeCompatibility`*, then let's add it to `values.yaml` and document it in the upgrade notes so users can opt out
- If we want to turn of bpf clock probe *only for new installations*, then let's add a `upgradeCompatibility` section for v1.14 or v1.15 (depending on the version where we want to turn it on by default)

gandro (changes requested):

install/kubernetes/cilium/templates/cilium-configmap.yaml
margamanterola: I think the current understanding is that this feature had been effectively turned off everywhere until https://github.com/cilium/cilium/pull/25795. Quoting from the PR desc: 

> The Cilium agent has been throwing the 'Auto-disabling "enable-bpf-clock-probe" feature since kernel doesn't expose /proc/timer_list' warning for a while now.

(admittedly, I don't know how long this "for a while now" is)

margamanterola: It seems it's already false in the values.yaml file, which I think was what Casey meant by his previous message:
https://github.com/cilium/cilium/blob/32118ccaa8677a3b836e60542a52a2af94693533/install/kubernetes/cilium/values.yaml#L428

gandro: Hm. So I did a quick check (`helm template cilium ./install/kubernetes/cilium --namespace kube-system  | grep enable-bpf-clock-probe`) and we have a bug in our Helm charts. BPF probes **are currently turned on by default**, because the above value in `values.yaml` is not used and so we fall back on the v1.8 defaults.

The value in `values.yaml` is called `bpf.clockProbe`:
https://github.com/cilium/cilium/blob/32118ccaa8677a3b836e60542a52a2af94693533/install/kubernetes/cilium/values.yaml#L428

But the way it is set in the config map is using `bpfClockProbe` (no dot):
https://github.com/cilium/cilium/blob/ffc60dd087d55f0466772dba16710c11dbcf5c5d/install/kubernetes/cilium/templates/cilium-configmap.yaml#L297

I think we should rename the unused variable in `values.yaml` to `bpfClockProbe`, or make the configmap use both the defacto-used `bpfClockProbe` and the intended `bpf.clockProbe` value instead.

pchaigno: > (admittedly, I don't know how long this "for a while now" is)

AFAIK, for a while is since that flag was added.

> I think the current understanding is that this feature had been effectively turned off everywhere until https://github.com/cilium/cilium/pull/25795.

Unfortunately, that fix got shipped into v1.13.4, which means some customers may be running with that feature enabled now.

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -498,6 +495,9 @@ bpf:
   # @default -- `[]`
   vlanBypass: ~
 
+# -- Enable BPF clock source probing for more efficient tick retrieval.
+bpfClockProbe: false
```
gandro, line 499: With this change here, the above change to `$defaultBpfClockProbe` is no longer needed

margamanterola: Uhm, if the other one is left as true, then new deployments will have it enabled, I'm not sure that's what we intend.

joestringer: Really? I was under the impression that the values.yaml values override the defaults.

I'm fine with the `defaultBpfClockProbe` change though. If we want to set a new default for newer versions then we will just add another stanza with an upgradeCompatibility check for the newer version in order to enable it only on new installations and not for existing ones. (We should do this separately in another submission, no need to address this for the backport/bugfix here)

margamanterola: Ok, Sebastian confirmed what you said, Joe, that values.yaml overrides the configmap, so I've dropped that changed and kept only the values.yaml one.

gandro (approved):
Helm change looks good to me now. But I do think we should also document this in upgrade impact Helm section here: 

https://github.com/cilium/cilium/blob/main/Documentation/operations/upgrade.rst#helm-options

joestringer (approved):

joestringer:
I agree that we should add some docs to describe this behaviour to users, but that can be submitted separately.

Issue #17459, opened 2021-09-23
UDP connections may be incorrectly dropped because of former DNS redirect policy


pchaigno:
If we used to have a policy redirecting traffic to a UDP proxy (typically the DNS proxy), after this policy is removed new connections may still be redirected to the proxy because they match an old, redirected connection in the conntrack map. As a result packets are dropped by the userspace proxy.

This incorrect redirection to the proxy can happen for up to 6 minutes after the DNS/L7 policy was removed in default settings. Every 5 minutes, we garbage collect UDP connections that didn't have activity for 1 minutes.

We may need to implement a grace period in the proxy where packets are not dropped to avoid this.

See https://github.com/cilium/cilium-cli/issues/367#issuecomment-920218323 and https://github.com/cilium/cilium-cli/issues/367#issuecomment-922937121 for details.

joestringer:
Minor question, where does the `Every 5 minutes` come from? Conntrack GC is typically dynamic and may take much longer. Is there another mechanism that is ensuring that this will be analyzed within 5 minutes?

If there is a grace period for disabling the proxy, we will still need to ensure that the proxy is correctly implementing the new forwarding policy during that time, for instance if the policy changes to simply drop the traffic, the proxy would need to also follow that policy. And presumably that grace period should be tied into the conntrack GC interval (for instance, by triggering a new GC iteration, then after GC runs, re-evaluating whether the proxy can disable its logic)

Alternative options are possibly more invasive, eg the datapath could consult the policymap again for the original direction flow to understand whether the current policy is to forward the traffic towards the proxy, rather than relying on conntrack.

pchaigno:
> Minor question, where does the `Every 5 minutes` come from? Conntrack GC is typically dynamic and may take much longer. Is there another mechanism that is ensuring that this will be analyzed within 5 minutes?

No, you're correct. I didn't measure this, just checked the code to see what I could expect for a default deployment. So I guess it's more of a: if you want to *maximize* chances of reproducing, you should ensure the two connections are within 6 minutes of each other.

> If there is a grace period for disabling the proxy, we will still need to ensure that the proxy is correctly implementing the new forwarding policy during that time, for instance if the policy changes to simply drop the traffic, the proxy would need to also follow that policy.

Yep, I'm not sure how feasible this grace period idea is. I'm a bit out of depth here (w.r.t. proxy internals). We should probably set up an ad-hoc call to discuss this at some point.

[...]

PR #26983, opened 2023-07-21
ci-ipsec-upgrade: Add IPsec upgrade tests

brb:
This commit adds an option IPsec e2e upgrade test. The test does the following:

* Install the latest stable release version of Cilium.
* Run the CLI's connectivity tests.
* Upgrade Cilium to a PR version.
* Run the CLI's connectivity tests.
* Downgrade to the previous version.
* Run the CLI's connectivity tests.

After each connectivity test case we flush CT to avoid the interference \[1\]\[2\] after running L7 proxy tests.

The test runs pods which established long-lived connections.
The test checks whether they are not interupted during upgrade /
downgrade.

\[1\]: https://github.com/cilium/cilium-cli/issues/367
\[2\]: https://github.com/cilium/cilium/issues/17459

### DIFF
--- a/.github/workflows/tests-ipsec-upgrade.yaml
+++ b/.github/workflows/tests-ipsec-upgrade.yaml
@@ -1,0 +1,445 @@
+name: Cilium IPsec upgrade (ci-ipsec-upgrade)
+
+# Any change in triggers needs to be reflected in the concurrency group.
+on:
+  workflow_dispatch:
+    inputs:
+      PR-number:
+        description: "Pull request number."
+        required: true
+      context-ref:
+        description: "Context in which the workflow runs. If PR is from a fork, will be the PR target branch (general case). If PR is NOT from a fork, will be the PR branch itself (this allows committers to test changes to workflows directly from PRs)."
+        required: true
+      SHA:
+        description: "SHA under test (head of the PR branch)."
+        required: true
+      extra-args:
+        description: "[JSON object] Arbitrary arguments passed from the trigger comment via regex capture group. Parse with 'fromJson(inputs.extra-args).argName' in workflow."
+        required: false
+        default: '{}'
+  # Run every 6 hours
+  schedule:
+    - cron:  '0 5/6 * * *'
+
+# By specifying the access of one of the scopes, all of those that are not
+# specified are set to 'none'.
+permissions:
+  # To be able to access the repository with actions/checkout
+  contents: read
+  # To allow retrieving information from the PR API
+  pull-requests: read
+  # To be able to set commit status
+  statuses: write
+
+concurrency:
+  # Structure:
+  # - Workflow name
+  # - Event type
+  # - A unique identifier depending on event type:
+  #   - schedule: SHA
+  #   - workflow_dispatch: PR number
+  #
+  # This structure ensures a unique concurrency group name is generated for each
+  # type of testing, such that re-runs will cancel the previous run.
+  group: |
+    ${{ github.workflow }}
+    ${{ github.event_name }}
+    ${{
+      (github.event_name == 'schedule' && github.sha) ||
+      (github.event_name == 'workflow_dispatch' && github.event.inputs.PR-number)
+    }}
+  cancel-in-progress: true
+
+env:
+  # renovate: datasource=github-releases depName=cilium/cilium-cli
+  cilium_cli_version:
+  # until https://github.com/cilium/cilium-cli/pull/1854 has been released
+  cilium_cli_ci_version: 8c624e141844a19a6b2e21d5255f3ca6195c996d
+  check_url: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
+  upgrade_cilium_from_vsn: 1.13.4
+  upgrade_cilium_to_vsn: 1.14.0-rc.1 # empty means upgrade to CI build
+
+jobs:
+  commit-status-start:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set initial commit status
+        uses: myrotvorets/set-commit-status-action@243b4f7e597f62335408d58001edf8a02cf3e1fd # v1.1.7
+        with:
+          sha: ${{ inputs.SHA || github.sha }}
+
+  setup-and-test:
+    runs-on: ubuntu-latest-4cores-16gb
+    name: 'Setup & Test'
+    env:
+      job_name: 'Setup & Test'
+    strategy:
+      fail-fast: false
+      max-parallel: 16
+      matrix:
+        include:
+          - name: '1'
+            kernel: '5.4-20230420.212204'
+            kube-proxy: 'iptables'
+            kpr: 'disabled'
+            tunnel: 'disabled'
+            encryption: 'ipsec'
+            encryption-node: 'false'
+            debug: 'true'
+            test-flow-interrupts: 'true'
+            ipv6: 'false' # until https://github.com/cilium/cilium/issues/26944 resolved
+
+          - name: '2'
+            kernel: '5.10-20230420.212204'
+            kube-proxy: 'iptables'
+            kpr: 'disabled'
+            tunnel: 'disabled'
+            encryption: 'ipsec'
+            encryption-node: 'false'
+            endpoint-routes: 'true'
+            test-flow-interrupts: 'true'
+            ipv6: 'false' # until https://github.com/cilium/cilium/issues/26944 resolved
+
+          - name: '3'
+            kernel: 'bpf-next-20230420.212204'
+            kube-proxy: 'iptables'
+            kpr: 'disabled'
+            tunnel: 'vxlan'
+            encryption: 'ipsec'
+            encryption-node: 'false'
+            endpoint-routes: 'true'
+            test-flow-interrupts: 'true'
+            ipv6: 'false' # until https://github.com/cilium/cilium/issues/26944 resolved
+
+    timeout-minutes: 60
+    steps:
+      - name: Checkout context ref
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          ref: ${{ inputs.context-ref || github.sha }}
+          persist-credentials: false
+
+      - name: Set Environment Variables
+        uses: ./.github/actions/set-env-variables
+
+      - name: Set up job variables
+        id: vars
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            SHA="${{ inputs.SHA }}"
+          else
+            SHA="${{ github.sha }}"
+          fi
+
+          # TODO(brb) move the settings derivation into a reusable GH workflow
+          CILIUM_MAIN_IMAGE_SETTINGS="--chart-directory=./install/kubernetes/cilium \
+            --helm-set=image.repository=quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/cilium-ci \
+            --helm-set=image.useDigest=false \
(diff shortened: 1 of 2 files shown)


.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -0,0 +1,415 @@
...
+            --helm-set=hubble.relay.image.tag=${SHA}"
+          echo "cilium_main_image_settings=${CILIUM_MAIN_IMAGE_SETTINGS}" >> $GITHUB_OUTPUT
+
+          CILIUM_INSTALL_DEFAULTS="--wait \
+            --helm-set=debug.enabled=true \
+            --helm-set=cni.uninstall=false \
+            --helm-set=cluster.name=default \
+            --helm-set=hubble.eventBufferCapacity=65535 \
+            --helm-set=bpf.monitorAggregation=none \
+            --helm-set=authentication.mutual.spire.enabled=true \
```
pchaigno: Any reason this is enabled? It's a beta feature.

.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -0,0 +1,415 @@
...
+
+          TUNNEL="--helm-set-string=tunnelProtocol=${{ matrix.tunnel }}"
+          if [ "${{ matrix.tunnel }}" == "disabled" ]; then
+            TUNNEL="--helm-set-string=routingMode=native --helm-set-string=autoDirectNodeRoutes=true --helm-set-string=ipv4NativeRoutingCIDR=10.244.0.0/16 --helm-set-string=tunnel=disabled"
+            TUNNEL="${TUNNEL} --helm-set-string=ipv6NativeRoutingCIDR=fd00:10:244::/56"
+          fi
+          LB_MODE=""
+          if [ "${{ matrix.lb-mode }}" != "" ]; then
+            LB_MODE="--helm-set-string=loadBalancer.mode=${{ matrix.lb-mode }}"
+          fi
```
pchaigno, line 167: This doesn't seem to be used here.

pchaigno: I see a bunch of matrix configurations that are used here actually. Is the plan to later extend this workflow to more than just IPsec?

.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -0,0 +1,415 @@
...
+      fail-fast: false
+      max-parallel: 16
+      matrix:
+        include:
+          - name: '1'
+            kernel: '5.4-20230420.212204'
+            kube-proxy: 'iptables'
+            kpr: 'disabled'
+            tunnel: 'disabled'
+            encryption: 'ipsec'
```
nbusseneau, line 86: I see we are only testing with IPsec enabled here, is there any reason for that? Do we plan on having upgrades tests without IPsec enabled added at a later point?

nbusseneau (approved):
I only checked the "boilerplate" parts of the workflow, LGTM. The meat of the workflow itself around the upgrade tests passed the smell test on a quick read but I did not do a deep review.

.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -0,0 +1,415 @@
...
+      - name: Wait for images to be available
+        if:  ${{ env.upgrade_cilium_to_vsn-upgrade == '' }}
+        timeout-minutes: 10
+        shell: bash
+        run: |
+          for image in cilium-ci operator-generic-ci hubble-relay-ci ; do
+            until docker manifest inspect quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/$image:${{ steps.vars.outputs.sha }} &> /dev/null; do sleep 45s; done
+          done
+
+      - name: Run upgrade tests (${{ join(matrix.*, ', ') }})
```
pchaigno: I would split this into separate install, connectivity tests, upgrade, etc. steps, to make what is failing a lot more obvious when we hit a failure.

.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -0,0 +1,415 @@
...
+            kubectl -n kube-system exec daemonset/cilium -- cilium status
+
+            ./cilium-cli connectivity test --include-unsafe-tests --collect-sysdump-on-failure \
+              \$EXTRA_CONNECTIVITY_TEST \
+              --flush-ct \
+              --test '!no-missed-tail-calls' \
+              --sysdump-hubble-flows-count=1000000 --sysdump-hubble-flows-timeout=5m \
+              --sysdump-output-filename "cilium-sysdump-${{ matrix.name }}-<ts>" \
+              --junit-file "cilium-junits/${{ env.job_name }} (${{ join(matrix.*, ', ') }}).xml" \
+              --junit-property github_job_step="Run tests upgrade 2 (${{ join(matrix.*, ', ') }})"
```
pchaigno, line 363: Is the `--include-conn-disrupt-test` part of this connectivity tests always executed last? If not, we may want to separate it from the rest to ensure it is executed last. The rationale is that connectivity tests themselves could cause XFRM errors and missed tail calls, so we'd want to check for those after running the tests.

.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -0,0 +1,415 @@
...
+
+            # --flush-ct interrupts the flows, so we need to set up again.
+            ./cilium-cli connectivity test --include-conn-disrupt-test --conn-disrupt-test-setup
+
+            # Dowgrade
```
pchaigno:
```suggestion
            # Downgrade
```

.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -0,0 +1,415 @@
...
+            CILIUM_CLI_MODE=helm ./cilium-cli upgrade \
+              --version ${{ env.upgrade_cilium_from_vsn }} \
+              ${{ steps.vars.outputs.cilium_install_defaults }} \
+              \$EXTRAS_V13
+
+            ./cilium-cli status --wait
+            kubectl get pods --all-namespaces -o wide
+            kubectl -n kube-system exec daemonset/cilium -- cilium status
+
+            ./cilium-cli connectivity test --test '!check-log-errors' --include-unsafe-tests --collect-sysdump-on-failure \
```
pchaigno: I'm guessing we have `--test '!check-log-errors'` here because of a bug. Is there an issue for it?

.github/workflows/tests-ipsec-upgrade.yaml
brb: The idea is to introduce a reusable GH workflow which derives Cilium configuration from a given matrix.

pchaigno (commented):
Super happy to see this! :heart_eyes: :rocket: 

I'm aware you're soon to be OOO, so I'm happy to take over or find someone who will. Considering that, I'm more interested in the answers to my comments below than in the actual changes that may follow. I can implement the changes myself if any are needed and I understand the rationales.

.github/workflows/tests-ipsec-upgrade.yaml
brb: Too much copy-paste

.github/workflows/tests-ipsec-upgrade.yaml
brb: Too much stock market

.github/workflows/tests-ipsec-upgrade.yaml
brb: Too much copy paste

.github/workflows/tests-ipsec-upgrade.yaml
brb: These are executed first.

If so, we can run the connectivity test twice - first connectivity w/o upgrade tests, and then only the upgrade tests. I didn't do that due to the L7 test issues :grimacing:

.github/workflows/tests-ipsec-upgrade.yaml
brb: Yep - https://github.com/cilium/cilium/pull/26983#discussion_r1270670830.

.github/workflows/tests-ipsec-upgrade.yaml
brb: Yeah, good idea. Let me do that.

.github/workflows/tests-ipsec-upgrade.yaml
pchaigno: The upgrade tests don't actually perform any connectivity tests, right? They just check various counters IIUC. So should be fine w.r.t. L7 CT reuse.

brb: The upgrade test checks whether the connections are not interrupted too. So, the L7 tests might interrupt them :-/

pchaigno: The L7 tests would interrupt existing connections because of the CT reuse bug? I wasn't aware of that :disappointed:

brb: I think I just spread FUD about L7 tests. It might be fine, but maybe let's try to change the order after merging this PR. Also, I haven't tested the suggested order yet.

.github/workflows/tests-ipsec-upgrade.yaml
brb: Came too quick to the conclusion. https://github.com/cilium/cilium-cli/issues/1858 :-(

.github/workflows/tests-ipsec-upgrade.yaml
pchaigno: :+1:

brb:
@pchaigno BTW, this job tests v1.13.4 -> v1.14-rc.1 -> v1.13.4. If you want to test against a custom CI PR build, then `upgrade_cilium_to_vsn` needs to be set to empty.

brb:
@pchaigno All green https://github.com/cilium/cilium/actions/runs/5624295113/job/15240865325 :tada:

brb:
Only `ci-ipsec-upgrade` matters. The rest can be red when merging.

pchaigno (approved):

pchaigno:
The new workflow is passing (cf. https://github.com/cilium/cilium/actions/runs/5694700139) and it's reviewed so marking ready to merge.

jschwinger233:
@brb If I understand correctly, we are going to make this .github/workflows/tests-ipsec-upgrade.yaml reusable in future. For example, I can call this upgrade test from conformance-ipsec-e2e.yaml (https://github.com/cilium/cilium/pull/27145). 

If that's the plan, maybe the test name can be more general? Right now it's `ci-ipsec-upgrade`, and we can make it `ci-upgrade` as a general upgrade test, then conformance-ipsec-e2e.yaml (ci-ipsec-e2e) will take care of the ipsec scenario by calling this workflow with properly arguments.

brb:
> If I understand correctly, we are going to make this .github/workflows/tests-ipsec-upgrade.yaml reusable in future. For example, I can call this upgrade test from conformance-ipsec-e2e.yaml 

@jschwinger233 I'd like to have `reusable-ipsec.yaml` with configuration matrix, Cilium installation params derivation, etc. Then this workflow could be  embedded by ci-ipsec-updare, ci-ipsec-e2e, etc.

jschwinger233:
@brb I noticed this PR didn't add `tests-ipsec-upgrade` to `/test`, which means `/test` won't trigger the test. Was it done by purpose, or we just forgot it?

https://github.com/cilium/cilium/blob/main/.github/ariane-config.yaml#L2-L18

PR #26987, opened 2023-07-21
ipsec: reject validation of empty keys

tommyp1ckles:
Previously, a ipsec xfrm config missing a key would pass validation and be inserted into node configuration.

This would go on to silently fail down the line with EINVAL when node managers: linux datapath handler attempted to create xfrm state with this key.  

This checks if the key is empty upon validation, and causes Cilium to terminate if the key is not valid.

As well, this removes the deprecated functionality of allowing keys without the spi as the first parameter validation.

Finally, when watching for key changes if that functionality is enabled, keys being rejected due to validation errors will be logged.


tommyp1ckles:
don't have time for this one atm, going to convert to draft for the time being.

PR #26989, opened 2023-07-21
Refactor hubble redact settings schema

ChrsMark:
This PR implements the refactoring of the `--hubble-redact` option in order to provide more flexibility as per https://github.com/cilium/cilium/issues/23887#issuecomment-1633015416.

Specifically we change 

```yml
hubble:
 redact:
  - http-url-query
  - kafka-api-key
```

to 

```yml
hubble:
  redact:
    enabled: true
    http:
      urlQuery: true
    kafka:
      apiKey: true
```

and 

```console
--hubble-redact=http-url-query,kafka-api-key
```

to

```console
--hubble-redact=true
--hubble-redact-http-url-query=true
--hubble-redact-kafka-api-key=true
```

With this change we aim to support the option for user provided http-headers list.

This will look like the following (as per https://github.com/cilium/cilium/issues/23887#issuecomment-1632167762):

```yml
hubble:
  redact:
    enabled: false
    http:
      userInfo: true
      urlQuery: true
      headers:
        allow:
          - traceparent
          - tracestate
          - Cache-Control
          - ...
    kafka:
      apiKey: true
```

cc: @ioandr

### DIFF
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -977,8 +977,14 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	)
 	option.BindEnv(vp, option.HubbleMonitorEvents)
 
-	flags.StringSlice(option.HubbleRedact, []string{}, "List of Hubble redact options")
+	flags.Bool(option.HubbleRedact, defaults.HubbleRedact, "Hubble redact sensitive information from flows")
 	option.BindEnv(vp, option.HubbleRedact)
+
+	flags.Bool(option.HubbleRedactHttpURLQuery, defaults.HubbleRedactHttpURLQuery, "Hubble redact http URL query from flows")
+	option.BindEnv(vp, option.HubbleRedactHttpURLQuery)
+
+	flags.Bool(option.HubbleRedactKafkaApiKey, defaults.HubbleRedactKafkaApiKey, "Hubble redact Kafka API key from flows")
+	option.BindEnv(vp, option.HubbleRedactKafkaApiKey)
 
 	flags.StringSlice(option.DisableIptablesFeederRules, []string{}, "Chains to ignore when installing feeder rules.")
 	option.BindEnv(vp, option.DisableIptablesFeederRules)
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -86,6 +86,15 @@ const (
 	// HubbleRecorderSinkQueueSize is the queue size for each recorder sink
 	HubbleRecorderSinkQueueSize = 1024
 
+	// HubbleRedact controls if sensitive information will be redacted from L7 flows
+	HubbleRedact = false
+
+	// HubbleRedactHttpURLQuery controls if the URL query will be redacted from flows
+	HubbleRedactHttpURLQuery = false
+
+	// HubbleRedactKafkaApiKey controls if the Kafka API key will be redacted from flows
+	HubbleRedactKafkaApiKey = false
+
 	// MonitorSockPath1_2 is the path to the UNIX domain socket used to
 	// distribute BPF and agent events to listeners.
 	// This is the 1.2 protocol version.
--- a/Documentation/cmdref/cilium-agent.md
+++ b/Documentation/cmdref/cilium-agent.md
@@ -187,7 +187,9 @@ cilium-agent [flags]
       --hubble-prefer-ipv6                                        Prefer IPv6 addresses for announcing nodes when both address types are available.
       --hubble-recorder-sink-queue-size int                       Queue size of each Hubble recorder sink (default 1024)
       --hubble-recorder-storage-path string                       Directory in which pcap files created via the Hubble Recorder API are stored (default "/var/run/cilium/pcaps")
-      --hubble-redact strings                                     List of Hubble redact options
+      --hubble-redact                                             Hubble redact sensitive information from flows
+      --hubble-redact-http-urlquery                               Hubble redact http URL query from flows
+      --hubble-redact-kafka-apikey                                Hubble redact Kafka API key from flows
       --hubble-skip-unknown-cgroup-ids                            Skip Hubble events with unknown cgroup ids (default true)
       --hubble-socket-path string                                 Set hubble's socket path to listen for connections (default "/var/run/cilium/hubble.sock")
       --hubble-tls-cert-file string                               Path to the public key file for the Hubble server. The file must contain PEM encoded data.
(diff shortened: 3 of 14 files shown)


pkg/defaults/defaults.go
```diff
@@ -86,6 +86,15 @@ const (
 	// HubbleRecorderSinkQueueSize is the queue size for each recorder sink
 	HubbleRecorderSinkQueueSize = 1024
 
+	// HubbleReact controls is sensitive information will be redacted from L7 flows
```
pippolo84: Couple of typos: `HubbleReact` -> `HubbleRedact` and `is` -> `if`

pippolo84 (changes requested):
Left a comment about a couple of typos. Apart from that, agent related changes LGTM.
CI Failures are legit, you need to run `make -C Documentation/ update-cmdref` to regenerate the docs.
Thanks!

daemon/cmd/daemon_main.go
```diff
@@ -977,9 +977,15 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	)
 	option.BindEnv(vp, option.HubbleMonitorEvents)
 
-	flags.StringSlice(option.HubbleRedact, []string{}, "List of Hubble redact options")
+	flags.Bool(option.HubbleRedact, defaults.HubbleRedact, "Hubble redact sensitive information from flows")
```
chancez, line 980: Do we even need a separate enabled option? The other options seem like they would be sufficient.

chancez (changes requested):
Looks like the tests need updating. One question about if we need all 3 options now or if we could just have the 2 new ones.

daemon/cmd/daemon_main.go
ChrsMark: That's based on "this approach would provide the best flexibility for the user yet with reasonable default, all the user has to do is to set `hubble.redact.enabled=true`" from https://github.com/cilium/cilium/issues/23887#issuecomment-1632167762.

chancez: Ah gotcha, so the defaults for the individual options are true, but the entire feature is off by default, that makes sense to me.

ChrsMark: Right! Note that the defaults are still `false` in this PR, but will be fixed accordingly based on the plan I suggested at https://github.com/cilium/cilium/issues/23887#issuecomment-1646650350.

zacharysarah (approved):
LGTM

chancez (approved):
Overall LGTM.

gandro (approved):
Helm looks good to me. I'll defer the Hubble changes to Chance

pippolo84 (approved):
Thanks! :100:

ChrsMark:
@derailed @tommyp1ckles @nathanjsweet could you please review this one when you get the chance 🙏🏽 ?

rolinh (approved):

ldelossa:
All required tests are passing and have critical mass on review approvals. Merging.

joestringer:
Both of the tests that failed on this PR prior to merging had this error, which is directly related to the changes in this PR:

```
connectivity test failed: unmarshaling cilium runtime config json: json: cannot unmarshal bool into Go struct field DaemonConfig.HubbleRedact of type []string
```

ldelossa:
@ChrsMark as explained here: https://github.com/cilium/cilium/pull/27352#issuecomment-1670007214

this pr was merged prematurely. Can you please open a subsequent PR with any suggested fixes mentioned in this one

ChrsMark:
Hey @ldelossa , sure I will have a look soon.
In the meantime, I wonder why we hit these compatability issues,  since https://github.com/cilium/cilium/pull/25844 should only be landed in `main` by now based on the discussions happened at https://github.com/cilium/cilium/issues/23887#issuecomment-1645784815 and the respective revert from `v1.14`.
Any pointer here would be appreciated :).

joestringer:
@ChrsMark Ah, I didn't recognize that this is only changing code in the main branch. The reason for the incompatibility is https://github.com/cilium/cilium-cli/issues/1896 . It looks like maybe the cilium/cilium-cli repo pulled in vendor code from cilium/cilium main after https://github.com/cilium/cilium/pull/25844, so the type it internally uses is incompatible with the type that was changed in this PR.

ChrsMark:
Hey @joestringer! So what is the best way to move this one forward? 

Given that the "breaking" change only applies on `main` would that make sense to make it now and then cilium-cli can pull an updated version of the dependency? 

Otherwise we can just rename `HubbleRedact` to `HubbleRedactEnabled` in this current patch but this would be a bit overkill just to tackle a non actual breaking change.

joestringer:
We're in a bit of a pickle because Cilium CLI is incorrectly relying on internal Cilium structures. Either we need to solve https://github.com/cilium/cilium-cli/issues/1896, or maybe use a different name for the field in `DaemonConfig` and open a fresh PR for these changes. We might even be able to also delete the current `HubbleRedact` field. However, those changes would need to be merged into main and then pulled into cilium-cli as a dependency before cilium/cilium main could have a different field called `HubbleRedact` inside `DaemonConfig`.

cc @cilium/cilium-cli-maintainers

michi-covalent:
yeah we still don't have a good answer for https://github.com/cilium/cilium-cli/issues/1896#issuecomment-1671605371.

in the meantime if we pick up the latest cilium code from main branch in cilium-cli, this error `connectivity test failed: unmarshaling cilium runtime config json: json: cannot unmarshal bool into Go struct field DaemonConfig.HubbleRedact of type []string` will go away right?

michi-covalent:
... so more concretely what i'm suggesting as a short term fix is:

- revert this revert: https://github.com/cilium/cilium/pull/27352
- and then immediately pick up cilium from the latest main branch in cilium-cli repo 💭

joestringer:
@michi-covalent We can't run CI properly and confirm that this PR doesn't break _other_ CI jobs without resolving this problem. Hence I suggest we just use a different field name so that the PR can pass the full CI and go through review *before* we merge it.

joestringer:
> in the meantime if we pick up the latest cilium code from main branch in cilium-cli, this error connectivity test failed: unmarshaling cilium runtime config json: json: cannot unmarshal bool into Go struct field DaemonConfig.HubbleRedact of type []string will go away right?

No, the problem is that the flag was first introduced with a bool type, then this PR changes it to a string. Now that this PR is reverted, it is back to a bool type. Cilium-CLI is still incorrectly deserializing that option as a bool, so updating Cilium-CLI will not fix that problem.

michi-covalent:
ok yeah using a different field name would work for sure if we are ok with changing the field name 👍

PR #26990, opened 2023-07-21
egressgw: use route.Upsert() for inserting nexthop / prefix IP route

julianwiedmann:
Upsert() handles the installation of the relevant nexthop route internally, and rolls it back if the prefix route fails.

It also contains a workaround for an issue in the kernel, when the nexthop route is not available immediately.

### DIFF
--- a/pkg/egressgateway/manager.go
+++ b/pkg/egressgateway/manager.go
@@ -668,7 +668,7 @@ func (manager *Manager) addMissingIpRulesAndRoutes(isRetry bool) (shouldRetry bo
 
 		policyConfig.forEachEndpointAndDestination(addIPRulesForConfig)
 
-		if err := addEgressIpRoutes(gwc.egressIP, gwc.ifaceIndex); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/egressgateway/manager.go
```diff
@@ -628,6 +628,12 @@ func (manager *Manager) addMissingIpRulesAndRoutes(isRetry bool) (shouldRetry bo
 			logfields.LinkIndex: gwc.ifaceIndex,
 		})
 
+		if err := addEgressIpRoutes(gwc); err != nil {
+			logger.WithError(err).Warn("Can't add IP routes")
+		} else {
+			logger.Debug("Added IP routes")
+		}
+
```
julianwiedmann, line 636: btw, I think we should `continue` here on an error (and not try to add any useless IP rules). And maybe even set `shouldRetry`, although I would need to refresh my memory for what cases we expect it to help ...

joamaki (approved):

PR #26998, opened 2023-07-21
ipam: deepcopy interface resource correctly.

tommyp1ckles:
ipam: deepcopy interface resource correctly.

Currently only used by azure ipam test mocks, was resulting in race
condition prone code where underlying interface addresses where being
shared across resources.

This ensures that the revision interface is correctly deepcopied such
that the underlying resource is also safely copied.

This seems at least partially related to flake https://github.com/cilium/cilium/issues/26617.

Addresses: https://github.com/cilium/cilium/issues/26617

Signed-off-by: Tom Hadlaw <tom.hadlaw@isovalent.com>

### DIFF
--- a/pkg/alibabacloud/eni/types/types.go
+++ b/pkg/alibabacloud/eni/types/types.go
@@ -126,6 +126,10 @@ type ENI struct {
 	Tags map[string]string `json:"tags,omitempty"`
 }
 
+func (e *ENI) DeepCopyInterface() types.Interface {
+	return e.DeepCopy()
+}
+
 // InterfaceID returns the identifier of the interface
 func (e *ENI) InterfaceID() string {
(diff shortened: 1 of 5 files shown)


christarazi (approved):

gandro (approved):

tommyp1ckles:
failures are flake this is partially trying to address

tommyp1ckles:
Think this is still desired, if tests pass following the last fix for the referenced test.

tommyp1ckles:
GKE test hit: https://github.com/cilium/cilium/issues/31385

christarazi:
@hemanthmalla Could you take a look on behalf of cilium/azure?

hemanthmalla (approved):
LGTM for Azure

Issue #21508, opened 2022-09-29
Extend cilium_bpf_map_pressure metric for all maps


joestringer:
Currently the cilium_bpf_map_pressure metric does not support all map types. We should extend it to support all map types.

Tasks:
- [x] Add support for CT / NAT map types (https://github.com/cilium/cilium/issues/20069)
  - [x] https://github.com/cilium/cilium/pull/28183
  - [x] https://github.com/cilium/cilium/pull/27001
- [x] Add policymap to `cilium_bpf_map_pressure_metric` https://github.com/cilium/cilium/pull/14131/files#diff-aeac8ae95e325c39fbeca8b97427e24b733ac5d11e549c7e78279872b05c3533R1082 
  - [x] Investigate why it doesn't show up in `cilium metrics list` output
  - [x] #27332
- [ ] Evaluate whether there are any other missing map types, and add support for those

ysksuzuki:
Hi @joestringer , do you have any plans to work on this issue? I'm planning to work on the enhancement of our monitoring system and want to monitor the map pressure of the CT, NAT, and policy map. Actually, we encountered an issue the other day in which Cilium couldn't create new pods because the policy map was full. Also, we would expect that the communication with Loadbalacner will be disrupted if the CT and NAT map are full. I believe those are very important metrics we should monitor.

ysksuzuki:
> Investigate why it doesn't show up in cilium metrics list output

I think that's because of the threshold.
https://github.com/cilium/cilium/blob/d1d8e7a35b35d3420a33251767ae2696d664da53/pkg/maps/policymap/policymap.go#L44

It shows up in my env if I set the threshold 0. Maybe we should make the threshold configurable?

```
cilium_bpf_map_pressure                               map_name="ipcache"                                                                       0.001301
cilium_bpf_map_pressure                               map_name="lb4_backends_v2"                                                               0.000107
cilium_bpf_map_pressure                               map_name="lb4_reverse_nat"                                                               0.002884
cilium_bpf_map_pressure                               map_name="lb4_services_v2"                                                               0.002884
cilium_bpf_map_pressure                               map_name="lb_affinity_match"                                                             0.000122
cilium_bpf_map_pressure                               map_name="lxc"                                                                           0.001190
cilium_bpf_map_pressure                               map_name="policy_00036"                                                                  0.007553
cilium_bpf_map_pressure                               map_name="policy_00063"                                                                  0.007553
[...]

Issue #20069, opened 2022-06-03
CFP: Extend bpf_map_pressure metric to cover connection tracking, NAT maps


joestringer:
## Cilium Feature Proposal

**Is your feature request related to a problem?**

If the CT or NAT maps become full, this can cause packet drops. Provide better visibility into the datapath when this occurs.

**Describe the feature you'd like**

The work in https://github.com/cilium/cilium/pull/14131 provided `bpf_map_pressure` metrics into maps managed purely by userspace. It would be nice to extend this to also provide feedback on the pressure on CT, NAT maps.

CT and NAT maps are more tricky than the other map types, partially because these maps are managed by the datapath, and also due to the use of LRU map type. It's not possible to accurately count adds and deletes for LRU maps, because some deletes are hidden. More specifically, when the map becomes full, you can still insert, but that insert has the side effect of bumping out an old entry, and this behaviour is not made visible to eBPF programs.

**(Optional) Describe your proposed solution**

One option for this is to just periodically perform a dump of these maps to count the number of entries at a point in time and export this. While this is not entirely accurate given that this takes time and the map is changing while the dump occurs, it could at least give some sense of how full the maps are. Variability in the metrics could also be insightful.

In order to do this efficiently, this feature could make use of newer kernel features like [batched map lookups](https://github.com/torvalds/linux/blob/50fd82b3a9a9335df5d50c7ddcb81c81d358c4fc/include/uapi/linux/bpf.h#L585) (available in Linux 5.6 or later).

Jiang1155:
This CFP looks interesting. If no one already worked on this, I will spend some time on this one.

aanm:
@Jiang1155 feel free to give it a go. I'll assign the issue to you. Thank you! cc @joestringer

ysksuzuki:
@Jiang1155 May I know the current status of this issue? If there's anything I can do, I'd be happy to help.

ysksuzuki:
I will work on this after finishing my on-call shift.

ysksuzuki:
Can I assign this issue to myself?

ysksuzuki:
Unassigned myself, since others are already working on this issue

PR #27001, opened 2023-07-21
Add support for tracking map pressure for NAT map

derailed:
Taps into LRU maps GC cycle to surface map pressure metric based on the resulting computed map size.

Add initial map pressure metrics support for the following LRU maps:
- NAT

> Note: This is a temporary solution until LRU map counts are surfaced
in the kernel.

Related: [#20069](https://github.com/cilium/cilium/issues/20069), https://github.com/cilium/cilium/issues/21508

```release-note
Add Prometheus map pressure metrics for NAT maps
```

### DIFF
--- a/pkg/maps/ctmap/ctmap.go
+++ b/pkg/maps/ctmap/ctmap.go
@@ -660,10 +660,10 @@ func PurgeOrphanNATEntries(ctMapTCP, ctMapAny *Map) *NatGCStats {
 			if !ctEntryExist(ctMap, ctKey) {
 				// No egress CT entry is found, delete the orphan ingress SNAT entry
 				if deleted, _ := natMap.Delete(natKey); deleted {
-					stats.IngressDeleted += 1
+					stats.IngressDeleted++
 				}
 			} else {
-				stats.IngressAlive += 1
+				stats.IngressAlive++
 			}
 		} else if natKey.GetFlags()&tuple.TUPLE_F_OUT == tuple.TUPLE_F_OUT {
 			ingressCTKey := ingressCTKeyFromEgressNatKey(natKey)
@@ -675,15 +675,19 @@ func PurgeOrphanNATEntries(ctMapTCP, ctMapAny *Map) *NatGCStats {
 				!ctEntryExist(ctMap, dsrCTKey) {
 				// No relevant CT entries were found, delete the orphan egress NAT entry
 				if deleted, _ := natMap.Delete(natKey); deleted {
-					stats.EgressDeleted += 1
+					stats.EgressDeleted++
 				}
 			} else {
-				stats.EgressAlive += 1
+				stats.EgressAlive++
 			}
 		}
 	}
 
-	natMap.DumpReliablyWithCallback(cb, stats.DumpStats)
+	if err := natMap.DumpReliablyWithCallback(cb, stats.DumpStats); err != nil {
+		log.WithError(err).Error("NATmap dump failed during GC")
+	} else {
+		natMap.UpdatePressureMetricWithSize(int32(stats.IngressAlive + stats.EgressAlive))
+	}
 
 	return &stats
 }
--- a/pkg/bpf/map_linux.go
+++ b/pkg/bpf/map_linux.go
@@ -288,7 +288,8 @@ func (m *Map) WithPressureMetric() *Map {
 	return m.WithPressureMetricThreshold(0.0)
 }
 
-func (m *Map) updatePressureMetric() {
+// UpdatePressureMetricWithSize updates map pressure metric using the given map size.
+func (m *Map) UpdatePressureMetricWithSize(size int32) {
 	if m.pressureGauge == nil {
 		return
 	}
@@ -303,8 +304,17 @@ func (m *Map) updatePressureMetric() {
 		return
 	}
 
-	pvalue := float64(len(m.cache)) / float64(m.MaxEntries())
+	pvalue := float64(size) / float64(m.MaxEntries())
 	m.pressureGauge.Set(pvalue)
+}
+
+func (m *Map) updatePressureMetric() {
+	// Skipping pressure metric gauge updates for LRU map as the cache size
+	// does not accurately represent the actual map sie.
+	if m.spec != nil && m.spec.Type == ebpf.LRUHash {
+		return
+	}
+	m.UpdatePressureMetricWithSize(int32(len(m.cache)))
 }
 
 func (m *Map) FD() int {
--- a/pkg/maps/nat/nat.go
+++ b/pkg/maps/nat/nat.go
@@ -97,7 +97,9 @@ func NewMap(name string, v4 bool, entries int) *Map {
 			mapValue,
 			entries,
 			0,
-		).WithCache().WithEvents(option.Config.GetEventBufferConfig(name)),
+		).WithCache().
+			WithEvents(option.Config.GetEventBufferConfig(name)).
+			WithPressureMetric(),
 		v4: v4,
 	}
 }
@@ -188,6 +190,7 @@ func (m *Map) Flush() int {
 	if m.v4 {
 		return int(doFlush4(m).deleted)
 	}
+
 	return int(doFlush6(m).deleted)
 }
 


ysksuzuki:
@derailed Thank you for the PR! 

Some random thoughts:

By default, the GC interval for the ct map is calculated dynamically based on maxDeleteRatio. From the GC perspective, that's OK, but it's not OK from the perspective of the user who wants to monitor the map pressure. This is not entirely accurate given that this takes time, and the map is changing while the dump occurs. I think we should use a fixed interval for the GC when the map pressure for the ct map is enabled. 
https://github.com/cilium/cilium/blob/3eaca5d3b0ac680acc2c7bf28dad9c0324b477f7/pkg/maps/ctmap/ctmap.go#L840

How do we export the map pressure of the nat map regularly? As far as I know, the flush nat map doesn't occur regularly, so we can't use the same approach as for the ct map. One possible solution is adding a go routine that exports the map pressure of the nat map regularly.

Since it is predicted that the CPU load will increase due to these changes, it seems necessary to perform them efficiently. I want to try BPF_MAP_LOOKUP_BATCH and see if we can do these efficiently.

pkg/maps/ctmap/ctmap.go
```diff
@@ -575,11 +575,17 @@ func (f *GCFilter) doFiltering(srcIP, dstIP netip.Addr, srcPort, dstPort uint16,
 }
 
 func doGC(m *Map, filter *GCFilter) int {
+	if m.mapType.isIPv4() {
+		stats := doGC4(m, filter)
+		m.UpdatePressureMetricWithSize(int32(stats.aliveEntries))
+		return int(stats.deleted)
+	}
 	if m.mapType.isIPv6() {
```
christarazi: IIUC, this code didn't exist before, as in we didn't have GC for IPv4 maps?

christarazi:
> By default, the GC interval for the ct map is calculated dynamically based on maxDeleteRatio. From the GC perspective, that's OK, but it's not OK from the perspective of the user who wants to monitor the map pressure. This is not entirely accurate given that this takes time, and the map is changing while the dump occurs. I think we should use a fixed interval for the GC when the map pressure for the ct map is enabled.

@ysksuzuki I believe the challenge with a fixed interval is that it will put extra pressure on the CPU. Walking these maps is very expensive. While counting the entries during the GC interval is not perfect, the significant advantage is that we can now have visibility into the map entries count without any extra CPU overhead, which is a really nice property. Plus, due to the dynamic GC interval, updating the metric occurs as pressure increases on the map. So the metric is updated essentially when the user should start paying attention to the map count.

ysksuzuki:
@christarazi 
> I believe the challenge with a fixed interval is that it will put extra pressure on the CPU. Walking these maps is very expensive on the CPU. 

I think that's why @joestringer suggests trying BPF_MAP_LOOKUP_BATCH  to perform the calculation efficiently, and I'm planning to do some PoC with it. I'm not so sure how much that could reduce the CPU cost though.

https://github.com/cilium/cilium/issues/20069
> In order to do this efficiently, this feature could make use of newer kernel features like batched map lookups (available in Linux 5.6 or later).

> the significant advantage is that we can now have visibility into the entry count without any extra CPU overhead, which is a really nice property

But we can already visualize the entry count with `cilium_datapath_conntrack_gc_entries{status="alive"}`, right? Given that we've 
already had the similar metric, I don't see the significant value in it.

Shouldn't we start by writing a design document, discussing it, and aligning our expectations for this issue?

pkg/maps/ctmap/ctmap.go
derailed: Tx Chris! No ipv4 gc checks was there.

derailed:
@ysksuzuki @chris Thank you both for weighting in!

1. I did see that a size metrics are reported for both ct and nat maps but the mechanic for map_pressure is a little bit different as this metric is registerered/unregistered when over a given threshold which would be used to trigger an alert when present vs just reporting on a map size. Not to mention, consistency as it uses the very same mechanic as other map types currently use.
2. There is a correlation between ct and nat map so tho the gc cycles are different, sanitization would be performed on nat maps to ensure the entries are in sync as part of the ct gc.
3. As stated in the PR this won't accurately represent pressure but thought might be viable in the interim...
4. I think @ti-mo has already experimented with this as he has a decent prototype using BPF_MAP_LOOKUP_BATCH and got very good benchmarked results. The eternal trade off here cpu vs mem as the map need to be pre-allocated for that call to return the count. 
5. Also there is great progress per the excellent work of @aspsk to get the counts directly from the kernel which is going to be ideal but will take time to trickle down to cilium and its customers. So we need something viable in the interim...

ysksuzuki:
@derailed Wow, some people are already working on this issue. I didn't know that. Can I see the information about those efforts somewhere? (BPF_MAP_LOOKUP_BATCH prototype and get the counts directly from the kernel)

If you need a temporary solution in the interim, I don't want to block you. Should I unassign myself from this issue for now? Please let me know if there is anything I can do to help!

derailed:
@ysksuzuki - I'll let @ti-mo and @aspsk pipe in here with the gory details. 
As far as the issue, I think you should still own it as you are much more knowledgeable than I in this area. 
It would be great if you could review this PR and see if I missed anything for this interim solution. 

Thank you!!

pkg/maps/ctmap/ctmap.go
```diff
@@ -575,11 +575,17 @@ func (f *GCFilter) doFiltering(srcIP, dstIP netip.Addr, srcPort, dstPort uint16,
...
-		return int(doGC6(m, filter).deleted)
-	} else if m.mapType.isIPv4() {
-		return int(doGC4(m, filter).deleted)
+		stats := doGC6(m, filter)
+		m.UpdatePressureMetricWithSize(int32(stats.aliveEntries))
+		return int(stats.deleted)
 	}
+
 	log.Fatalf("Unsupported ct map type: %s", m.mapType.String())
 	return 0
```
christarazi: So it looks like the CT map GC already has some equivalent metrics tracking in `stats.finish()`, see https://github.com/cilium/cilium/blob/47724684e520154e66001bc3383eac3bccf7bde2/pkg/maps/ctmap/metrics.go#L86-L101. The metric is the one pointed out by Yusuke in https://github.com/cilium/cilium/pull/27001#issuecomment-1646525210.

So it seems we don't need these changes highlighted here.

christarazi (commented):
Given the below and @ysksuzuki's comment, I can see that it doesn't quite give us much value if we just duplicate whatever we do for the existing CT GC metric and `cilium_bpf_map_pressure`. Now I understand the proposal to rather do the counting on a fixed interval and utilized batch lookups to mitigate CPU spike concerns.

christarazi:
@ysksuzuki Does it make sense to keep this PR just for the NAT map since it doesn't have any metrics? Then in another PR, we can propose batch lookups + fixed interval counting.

christarazi:
@derailed Thanks for updating the PR. 

I took a deep look at what this PR is doing and the NAT GC cycle. I think the changes made in `pkg/bpf/map_linux.go` are good to keep as those are the foundational pieces that we need to build off of. However, the changes in `pkg/maps/nat/nat.go` are not needed as is. The current problem with it is that the pressure metric calculation logic relies on `Flush()` being called, which upon investigation is only called from the CLI in response to a user issuing `cilium bpf ... nat flush`. Taking a closer look at how the NAT map works, we find that the datapath will insert into the map, but deletion happens during CT map GC. The fact that deletion happens during CT map GC is a bit confusing, but the code for that is [here](https://github.com/cilium/cilium/blob/551e8cd837dec24e96bb4ff95d03c5b220aee505/pkg/maps/ctmap/ctmap.go#L628). You'll [notice](https://github.com/cilium/cilium/blob/551e8cd837dec24e96bb4ff95d03c5b220aee505/pkg/maps/ctmap/ctmap.go#L662) that the delete is issued for the NAT map in that code as well. Conveniently, the logic is also [counting](https://github.com/cilium/cilium/blob/551e8cd837dec24e96bb4ff95d03c5b220aee505/pkg/maps/ctmap/ctmap.go#L666) the number of alive entries, which is exactly what we need to report as the pressure metric.

One complication though with the current implementation is that internal calls to `updatePressureMetric()` will now call `UpdatePressureMetricWithSize(len(cache))` which will overwrite the value that we would pass in from `PurgeOrphanNATEntries()`. So I think we need some mechanism to prevent this from happening. Thoughts? Once we do that, then I think this should accurately count the entries in the NAT map.

derailed:
@christarazi Great catch! So let me see if I understood this correctly. 

The ctmap manages the content of an associated nat map during its GC cycles. 
When the purgeOrphan call is made, we end up calling DumpReliablyWithCallback which performs house keeping on the associated nat map. Tho we are tracking stats for the number of alive ingress/egress these don't seem relevant when establishing map pressure?? But since pressure metrics is now enabled on nat maps and we're maintaining a cache, deletes would trigger a [nat map pressure updates](https://github.com/cilium/cilium/blob/main/pkg/bpf/map_linux.go#L873). 
Or am I missing it?

derailed:
@christarazi I think I understand but still unclear... From what I can tell during the orphan call we do keep a tally only for ingress/egress entries but these are only 2 of the types of entries that could be in the nat map? 
Looking at the bit flag there could be other types of entries and thus triggering the map pressure update call would not yield the correct size ie UpdatePressureWithSize(stats.IngressAlive+stats.EgressAlive). 
If this is correct,  we would need to keep another tally of alive-entries and make the call with that computed total.
Am I wrong about that?

christarazi:
@derailed Ah, now I understand your question. It's basically, can there be more than 2 unique flags per NAT map key. From looking at the datapath code (`bpf/lib/nat.h`) for NAT and tracing the call paths, I only see `TUPLE_F_OUT` (egress) and `TUPLE_F_IN` (ingress) being set. Which makes sense given that there can only be two directions. While there's theoretically [more flags](https://github.com/cilium/cilium/blob/b3171b99441a84ad307fdf588e1c954fc7bcfd88/bpf/lib/common.h#L813-L816) that can be set, the non-directional ones are only used in the CT map keys and not the NAT map keys. It's a bit confusing because both the keys types are the same for both maps.

So in summary, 

```
UpdatePressureWithSize(stats.IngressAlive+stats.EgressAlive).
```

should be accurate.

derailed:
@christarazi - Brilliant! Thank you for the clarification. I'll make it so.

christarazi:
@derailed I don't see any code that addresses my previous point:

> Hence why we need to stop LRU maps from having the [automatic map pressure updates triggered](https://github.com/cilium/cilium/blob/main/pkg/bpf/map_linux.go#L873)

otherwise, the call to `UpdatePressureMetricWithSize()` inside `PurgeOrphanNATEntries()` will get overwritten.

derailed:
@christarazi I don't think that is the case as the UpdatePressureMetricWithSize is called at the end of this process (aka after the potential deletions) and sets the metric to what's left in the map. If we don't want delete to update the metrics and given delete is implemented in the base we would need to introduce a new check to prevent the map pressure updates specifically for nats. Is this what we want?

christarazi:
@derailed Yes, because AFAIU, LRU maps cannot be represented accurately in userspace because the datapath can trigger eviction and eviction does not trigger the userspace to update its cache. Therefore, we must only rely on walking the map, a la `PurgeOrphanNATEntries()` and a manual call to `UpdatePressureMetricWithSize()`. So I think we need logic to check the map type in https://github.com/cilium/cilium/blob/e657234cdd358572cb303236c8b81e8eeb445f7a/pkg/bpf/map_linux.go#L291 and return early if the map type is LRU.

christarazi:
@derailed It seems that most of CI is failing presumably due to some fatal crash or something of the sort. I would suggest downloading a sysdump from one of the runs and investigate the cause. For example, this is the smoke test's sysdump: https://github.com/cilium/cilium/suites/14891492852/artifacts/848345104

christarazi (approved):

dylandreimerink (approved):

ti-mo:
I'm not in favor of adding anything to `struct Map`, but given the lack of alternative and the change is small, it's fine for now.

ldelossa:
all required test passing. Merging

PR #27002, opened 2023-07-21
Trigger required workflows using Ariane

### LINKED ISSUE #26968: some github workflows don't get marked "skipped" when skipped
some github workflows don't get marked "skipped" when skipped, and they get stuck in "pending" state. for example:

- https://github.com/cilium/cilium/blob/main/.github/workflows/conformance-gateway-api.yaml
- https://github.com/cilium/cilium/blob/main/.github/workflows/conformance-ingress.yaml
- https://github.com/cilium/cilium/blob/main/.github/workflows/conformance-runtime.yaml
- https://github.com/cilium/cilium/blob/main/.github/workflows/integration-test.yaml

there is a rumor that converting these workflow to use Ariane will make it easy to mark skipped workflows 💡

michi-covalent:
Use Ariane to trigger all the required workflows so that they get marked
as skipped when they get skipped.

After this gets merged, we need to update required status checks:

- Replace `gateway-api-conformance-test *` with `ConformanceGatewayAPI`
- Replace `ingress-conformance-test *` with `ConformanceIngress`
- Replace `integration-test` with `IntegrationTests`
- Replace `Runtime Test *` with `Cilium Runtime`

Fixes https://github.com/cilium/cilium/issues/26968

### DIFF
--- a/.github/workflows/conformance-gateway-api.yaml
+++ b/.github/workflows/conformance-gateway-api.yaml
@@ -2,10 +2,22 @@ name: ConformanceGatewayAPI
 
 # Any change in triggers needs to be reflected in the concurrency group.
 on:
-  pull_request:
-    paths-ignore:
-      - 'Documentation/**'
-      - 'test/**'
+  workflow_dispatch:
+    inputs:
+      PR-number:
+        description: "Pull request number."
+        required: true
+      context-ref:
+        description: "Context in which the workflow runs. If PR is from a fork, will be the PR target branch (general case). If PR is NOT from a fork, will be the PR branch itself (this allows committers to test changes to workflows directly from PRs)."
+        required: true
+      SHA:
+        description: "SHA under test (head of the PR branch)."
+        required: true
+      extra-args:
+        description: "[JSON object] Arbitrary arguments passed from the trigger comment via regex capture group. Parse with 'fromJson(inputs.extra-args).argName' in workflow."
+        required: false
+        default: '{}'
+
   push:
     branches:
       - main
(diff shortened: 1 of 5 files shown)


bimmlerd (approved):
contributing looks good - I think Nicolas is the main review here.

.github/workflows/conformance-gateway-api.yaml
```diff
@@ -31,8 +66,17 @@ env:
   gateway_api_version: v0.7.0
   metallb_version: 0.12.1
   timeout: 5m
+  check_url: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
```
nbusseneau: I don't think `check_url` is used anymore, is it? We should probably just remove it from other workflows if it's still there.

.github/workflows/conformance-gateway-api.yaml
```diff
@@ -60,11 +104,7 @@ jobs:
       - name: Set image tag
         id: vars
         run: |
-          if [ ${{ github.event.pull_request }} ]; then
-            SHA=${{ github.event.pull_request.head.sha }}
-          else
-            SHA=${{ github.sha }}
-          fi
+          SHA=${{ inputs.SHA || github.sha }}
```
nbusseneau: Can we standardize to the same as other workflows?
```suggestion
          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
            SHA="${{ inputs.SHA }}"
          else
            SHA="${{ github.sha }}"
          fi
```
I know it's kinda stupid here since there is only one variable but I think it helps to keep a consistent structure.

nbusseneau (approved):

nbusseneau:
Note for maintainers:

The following checks will need to be unmarked as required on `main`:

- Runtime Test (agent)
- Runtime Test (datapath)
- Runtime Test (privileged)
- gateway-api-conformance-test (experimental)
- gateway-api-conformance-test (standard)
- ingress-conformance-test (With Default Ingress Controller, disabled, dedicated, true)
- ingress-conformance-test (With Shared LB, disabled, shared, false)
- ingress-conformance-test (With XDP, native, dedicated, false)
- ingress-conformance-test (Without XDP, disabled, dedicated, false)
- integration-test

And the following checks will need to be marked as required instead:

- Cilium Runtime
- ConformanceGatewayAPI
- ConformanceIngress
- IntegrationTests

nbusseneau (approved):

joestringer:
@nbusseneau should we backport this to v1.14?

nbusseneau:
I think yes, and to 1.13 and 1.12 as well as some of the workflows were already there.

PR #27007, opened 2023-07-23
metrics: fix potential conflict on metrics registration

### LINKED ISSUE #25610: Missing client-go Metrics and Potential Conflict with controller-runtime on Metrics Registration

### What happened?

The client-go metrics such as `cilium_k8s_client_api_latency_time_seconds` and `cilium_k8s_client_api_calls_total` are missing. I suspect that there's a conflict with controller-runtime on the metrics registration. controller-runtime also registers the client-go request result metrics.

controller-runtime also calls the client-go metrics registration
https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/metrics/client_go_adapter.go#L52

Subsequent calls will be ignored
https://github.com/kubernetes/client-go/blob/master/tools/metrics/metrics.go#L124

#### v1.12.8

112_latency

112-total

#### v1.13.2

113-latency

113-total

### Cilium Version

Client: 1.13.2 8cb94c70f7 2023-04-17T23:19:21+02:00 go version go1.19.8 linux/amd64
Daemon: 1.13.2 8cb94c70f7 2023-04-17T23:19:21+02:00 go version go1.19.8 linux/amd64

### Kernel Version

Linux rack0-cs4 5.15.106-flatcar

### Kubernetes Version

Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.9", GitCommit:"a1a87a0a2bcd605820920c6b0e618a8ab7d117d4", GitTreeState:"clean", BuildDate:"2023-04-12T12:16:51Z", GoVersion:"go1.19.8", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7
Server Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.9", GitCommit:"a1a87a0a2bcd605820920c6b0e618a8ab7d117d4", GitTreeState:"archive", BuildDate:"2023-05-02T00:33:10Z", GoVersion:"go1.19.8", Compiler:"gc", Platform:"linux/amd64"}

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

ysksuzuki:
In Cilium v1.13, K8s rest client metrics, such as cilium_k8s_client_api_latency_time_seconds and cilium_k8s_client_api_calls_total, are missing because of a conflict with controller-runtime on the metrics registration.

controller-runtime(sigs.k8s.io/controller-runtime/pkg/metrics) also registers the client-go rest client metrics on its registry using metrics.Register method in its init function. The metrics.Register can be called only once, and subsequent calls will be ignored.

cilium-agent doesn't use controller-runtime, but there's an indirect dependency from github.com/cilium/cilium/daemon to github.com/cilium/cilium/operator/metrics package. (operator uses controller-runtime in its Gateway API implementation) For example,
github.com/cilium/cilium/daemon/cmd
-> github.com/cilium/cilium/pkg/ipam
-> github.com/cilium/cilium/pkg/ipam/metrics
-> github.com/cilium/cilium/operator/metrics

The init function of sigs.k8s.io/controller-runtime/pkg/metrics will be called because of this indirect dependency.

The issue(missing client-go metrics) doesn't happen with v1.14 and later, but it can happen if the order of imports changes again. This commit is to prevent the issue from happning.

This commit is a forward port of　https://github.com/cilium/cilium/pull/26412/commits/e4ff641fb30fa0e711dc9a7e2fcce92dd6d510b6

fixes: #25610

      GitHub issue.

### DIFF
--- a/pkg/k8s/watchers/watcher.go
+++ b/pkg/k8s/watchers/watcher.go
@@ -93,13 +93,29 @@ func init() {
 		k8s.K8sErrorHandler,
 	}
 
-	k8s_metrics.Register(k8s_metrics.RegisterOpts{
+	// The client-go Register function can be called only once,
+	// but there's a possibility that another package calls it and
+	// registers the client-go metrics on its own registry.
+	// The metrics of one who called the init first will take effect,
+	// and which one wins depends on the order of package initialization.
+	// Currently, controller-runtime also calls it in its init function
+	// because there's an indirect dependency on controller-runtime.
+	// Given the possibility that controller-runtime wins, we should set
+	// the adapters directly to override the metrics registration of
+	// controller-runtime as well as calling Register function.
+	// Without calling Register function here, controller-runtime will have
+	// a chance to override our metrics registration.
+	registerOps := k8s_metrics.RegisterOpts{
 		ClientCertExpiry:      nil,
 		ClientCertRotationAge: nil,
 		RequestLatency:        &requestLatencyAdapter{},
 		RateLimiterLatency:    &rateLimiterLatencyAdapter{},
 		RequestResult:         &resultAdapter{},
-	})
+	}
+	k8s_metrics.Register(registerOps)
+	k8s_metrics.RequestLatency = registerOps.RequestLatency
+	k8s_metrics.RateLimiterLatency = registerOps.RateLimiterLatency
+	k8s_metrics.RequestResult = registerOps.RequestResult
 }
 
 var (


ysksuzuki:
context: https://github.com/cilium/cilium/pull/26412#pullrequestreview-1512518470

pkg/k8s/watchers/watcher.go
```diff
@@ -93,13 +93,27 @@ func init() {
 		k8s.K8sErrorHandler,
 	}
 
-	k8s_metrics.Register(k8s_metrics.RegisterOpts{
+	// The client-go Register function can be called only once,
+	// but controller-runtime also calls it in its init function
```
tommyp1ckles: nit: we can maybe generalize that any in case of transitive imports can cause this condition to happen (while also mentioning the specific controller-runtime scenario that was the motivation here).

pkg/k8s/watchers/watcher.go
```diff
@@ -93,13 +93,27 @@ func init() {
...
+	// and which one wins depends on the order of package initialization.
+	// Given the possibility that controller-runtime wins, we should set
+	// the adapters directly to override the metrics registration of
+	// controller-runtime as well as calling Register function.
+	// Without calling Register function here, controller-runtime will have
+	// a chance to override our metrics registration.
+	registerOps := k8s_metrics.RegisterOpts{
 		ClientCertExpiry:      nil,
 		ClientCertRotationAge: nil,
 		RequestLatency:        &requestLatencyAdapter{},
```
tommyp1ckles, line 111: do we still need these if we're just overriding them directly right afterwards?

tommyp1ckles (commented):
nice fix, I think this makes sense but I had a few questions

pkg/k8s/watchers/watcher.go
ysksuzuki: Yes, we still need to call `Register`, otherwise controller-runtime will override our requestResultAdapter.
https://github.com/kubernetes-sigs/controller-runtime/blob/7f0c6dc440f334849b25866dd9990ac1a5a777c2/pkg/metrics/client_go_adapter.go#L52-L54

We need to block it by calling `Register` here.

pkg/k8s/watchers/watcher.go
ysksuzuki: I rephrased it. Please take a look.

ysksuzuki:
ginkgo build cache seems broken
https://github.com/cilium/cilium/actions/runs/5706051069/job/15462054023

pkg/k8s/watchers/watcher.go
tommyp1ckles: I think I'm still a bit confused, my understanding is that the problem is caused by two places racing to call Register.

Is it not possible for the client_go_adapter callsite to call register prior to this?

ysksuzuki: > Is it not possible for the client_go_adapter callsite to call register prior to this?

It's possible if the import order changes.

client-go `Register` is called from these two init functions, and which will be called first depends on the package initialization order (import order).

`init` in `sigs.k8s.io/controller-runtime/pkg/metrics` 
https://github.com/kubernetes-sigs/controller-runtime/blob/7f0c6dc440f334849b25866dd9990ac1a5a777c2/pkg/metrics/client_go_adapter.go#L43

`init` in `github.com/cilium/cilium/pkg/k8s/watchers`
https://github.com/cilium/cilium/blob/fded1abcf310160a9c518b6e25f1ac6059e08a97/pkg/k8s/watchers/watcher.go#L96

ysksuzuki: @tommyp1ckles Did I answer your question?

ysksuzuki:
@tommyp1ckles Do you have any other feedback or question?

pkg/k8s/watchers/watcher.go
tommyp1ckles: @ysksuzuki sorry for the delay, I missed your notification 😞 

What I was getting at is if we look at the [Register code](https://github.com/cilium/cilium/blob/main/vendor/k8s.io/client-go/tools/metrics/metrics.go#L132), as you mentioned this is only done once and the register logic is called inside a sync.Once.

So if this is called after the go-client init one, the Register will still be a no-op which what I think the issue we're trying to fix?

i.e. It seems like this still only works if the pkgs/watchers/watchers call to Registers "wins" and calls Register first

ysksuzuki: > So if this is called after the go-client init one, the Register will still be a no-op which what I think the issue we're trying to fix?

Yes. In the case the controller-runtime wins, we need to override the adapters directly. On the other hand, in the case we win, we need to call `Register`, otherwise controller-runtime will call `Register` and override our adapters. (The former case happens if the import order changes)
This code works for both scenarios.

tommyp1ckles: Ok I see, sorry I was a bit confused still.  The metrics are also exported which lets us override them, so in the case where the client-go library sets them we override, otherwise the client-go library just uses Register which is a no-op.

Thanks for the extended clarification 👍

tommyp1ckles (approved):

pkg/k8s/watchers/watcher.go
ysksuzuki: Yeah, it's very confusing.  I'm considering whether we can remove the dependency on the controller-runtime's metrics package.

borkmann:
@tommyp1ckles do we need backport labels?

tommyp1ckles:
@borkmann 👍 This fixues a bug in 1.13, but this sounds like it may prevent future issues in 1.14 as well.

PR #26958, opened 2023-07-20
[v1.12] Delete IP Label metadata on delete from ipcache

meyskens:
This adds a delete call to the metadata map of the ipcache. On a delete from ipcache this info was left bhind. In case a node gets deleted the IP would keep receiving the remote-host label from this cache incorrectly.

This issue came out of an issue where we saw EKS nodes being recycled and the IPs getting re-used and receiving an incorrect remote-host label in identity cache and ip cache.

```release-note
Delete IP Label metadata on delete from ipcache
```

- v1.12: this PR
- v1.13: https://github.com/cilium/cilium/pull/27010
- v1.14: not affected thanks to a refactor :)

PR #27010, opened 2023-07-24
[v1.13] Delete IP Label metadata on delete from ipcache

meyskens:
v1.13 backport of https://github.com/cilium/cilium/pull/26958 

```release-note
Delete IP Label metadata on delete from ipcache
```

### DIFF
--- a/pkg/node/manager/manager.go
+++ b/pkg/node/manager/manager.go
@@ -57,6 +57,7 @@ type IPCache interface {
 	Upsert(ip string, hostIP net.IP, hostKey uint8, k8sMeta *ipcache.K8sMetadata, newIdentity ipcache.Identity) (bool, error)
 	Delete(IP string, source source.Source) bool
 	UpsertLabels(prefix netip.Prefix, lbls labels.Labels, src source.Source, rid ipcacheTypes.ResourceID)
+	RemoveLabels(prefix netip.Prefix, lbls labels.Labels, resource ipcacheTypes.ResourceID)
 }
 
 // Configuration is the set of configuration options the node manager depends
@@ -508,6 +509,7 @@ func (m *Manager) NodeUpdated(n nodeTypes.Node) {
 		entry.mutex.Lock()
 		m.mutex.Unlock()
 		oldNode := entry.node
+		oldResource := ipcacheTypes.NewResourceID(ipcacheTypes.ResourceKindNode, "", oldNode.Name)
 		entry.node = n
 		if dpUpdate {
 			m.Iter(func(nh datapath.NodeHandler) {
@@ -532,7 +534,7 @@ func (m *Manager) NodeUpdated(n nodeTypes.Node) {
 			}
 			oldNodeIPAddrs = append(oldNodeIPAddrs, prefix.String())
 		}
-		m.deleteIPCache(oldNode.Source, oldNodeIPAddrs, ipsAdded)
+		m.deleteIPCache(oldNode.Source, oldNodeIPAddrs, ipsAdded, remoteHostIdentity, oldResource)
 
 		// Delete the old health IP addresses if they have changed in this node.
 		oldHealthIPs := []string{}
@@ -542,7 +544,7 @@ func (m *Manager) NodeUpdated(n nodeTypes.Node) {
 		if oldNode.IPv6HealthIP != nil {
 			oldHealthIPs = append(oldHealthIPs, oldNode.IPv6HealthIP.String())
 		}
-		m.deleteIPCache(oldNode.Source, oldHealthIPs, healthIPsAdded)
+		m.deleteIPCache(oldNode.Source, oldHealthIPs, healthIPsAdded, remoteHostIdentity, oldResource)
 
 		// Delete the old ingress IP addresses if they have changed in this node.
 		oldIngressIPs := []string{}
@@ -552,7 +554,7 @@ func (m *Manager) NodeUpdated(n nodeTypes.Node) {
 		if oldNode.IPv6IngressIP != nil {
 			oldIngressIPs = append(oldIngressIPs, oldNode.IPv6IngressIP.String())
 		}
-		m.deleteIPCache(oldNode.Source, oldIngressIPs, ingressIPsAdded)
+		m.deleteIPCache(oldNode.Source, oldIngressIPs, ingressIPsAdded, remoteHostIdentity, oldResource)
 
 		entry.mutex.Unlock()
 	} else {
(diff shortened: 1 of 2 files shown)


meyskens:
This change breaks in 1.13, converting to draft

christarazi:
I think there should be a call to `RemoveLabels` in `(m *Manager) deleteIPCache` to be symmetric with `(m *Manager) upsertIntoIDMD` in `pkg/node/manager/manager.go` because that will properly clean up the metadata map. The `RemoveLabels` function does not exist in v1.12 version btw.

youngnick (approved):

christarazi (changes requested):
https://github.com/cilium/cilium/pull/27010#issuecomment-1652559801

meyskens:
I modified RemoveMetadata to remove all entries of an IP instead of a select set, the function was not used anywhere so did not cause any issues.

pkg/ipcache/ipcache.go
```diff
@@ -417,9 +417,10 @@ func (ipc *IPCache) UpsertMetadata(prefix netip.Prefix, src source.Source, resou
 	ipc.TriggerLabelInjection()
 }
 
-func (ipc *IPCache) RemoveMetadata(prefix netip.Prefix, resource ipcacheTypes.ResourceID, aux ...IPMetadata) {
+// RemoveMetadata removes the metadata of a given IP from the ipcache metadata
+func (ipc *IPCache) RemoveMetadata(prefix netip.Prefix) {
 	ipc.metadata.Lock()
-	ipc.metadata.remove(prefix, resource, aux...)
+	ipc.metadata.removeAllForPrefix(prefix)
 	ipc.metadata.Unlock()
```
christarazi: This will actually break the intended behavior of the metadata resolution. Essentially, there should be an equal number of upserts to removals. In other words, for every upsert-metadata, there is a remove-metadata. This breaks that assumption, which will remove metadata from a prefix prematurely.

pkg/ipcache/ipcache.go
```diff
@@ -436,10 +437,6 @@ func (ipc *IPCache) UpsertLabels(prefix netip.Prefix, lbls labels.Labels, src so
 	ipc.UpsertMetadata(prefix, src, resource, lbls)
 }
 
-func (ipc *IPCache) RemoveLabels(cidr netip.Prefix, lbls labels.Labels, resource ipcacheTypes.ResourceID) {
-	ipc.RemoveMetadata(cidr, resource, lbls)
```
christarazi: I don't think we should do this because it is a public API which may get used in future versions of v1.13.

christarazi (changes requested):
My previous review comment still applies.

meyskens:
/test-backport-1.13

meyskens:
Thrown away all work and started fresh given the architecture is way different than 1.12 :)

pkg/node/manager/manager.go
```diff
@@ -651,15 +652,9 @@ func (m *Manager) NodeDeleted(n nodeTypes.Node) {
...
 
-	for _, address := range []net.IP{
-		entry.node.IPv4HealthIP, entry.node.IPv6HealthIP,
-		entry.node.IPv4IngressIP, entry.node.IPv6IngressIP,
-	} {
-		if address != nil {
-			m.ipcache.Delete(address.String(), n.Source)
-		}
+		resource := ipcacheTypes.NewResourceID(ipcacheTypes.ResourceKindNode, "", n.Name)
+		m.ipcache.RemoveLabels(prefix, labels.Labels{}, resource)
```
christarazi: Strictly only this change will cause other bugs because we should be calling `RemoveLabels` on the inverse of wherever `upsertIntoIDMD` is called. In other words, `upsertIntoIDMD` and `RemoveLabels` need to be symmetric.

We already have a convenient function analogous to `upsertIntoIDMD` called `deleteIPCache`. I think the call to `RemoveLabels` should go inside `deleteIPCache` without removing the existing ipcache calls.

joestringer:
@meyskens heads up that I ended up preparing a subtly different fix on v1.12 here, may be worth comparing for the approach on v1.13: https://github.com/cilium/cilium/pull/27406. I also extended your testing script [here](https://gist.github.com/joestringer/d7f55da5dd2f33be6148dd52a96a4a54), which has two tests that now pass and two tests that still fail on all branches. For this bugfix it's fine to fix the first two test cases, then we can separately take a look at the second two test cases.

I know that there's more differences with v1.13 and this code has been heavily changing over time, so we'll need to be a bit careful around how we introduce the changes.

pkg/node/manager/manager.go
```diff
@@ -651,15 +681,12 @@ func (m *Manager) NodeDeleted(n nodeTypes.Node) {
 			prefix = ip.IPToNetPrefix(address.IP.To16())
 		}
 		m.ipcache.Delete(prefix.String(), n.Source)
-	}
 
-	for _, address := range []net.IP{
-		entry.node.IPv4HealthIP, entry.node.IPv6HealthIP,
-		entry.node.IPv4IngressIP, entry.node.IPv6IngressIP,
```
joestringer, line 658: When the node gets deleted, this code still needs to remove the association for the old health and ingress IPs. Was it intentional to remove the iteration here for these additional addresses?

pkg/node/manager/manager.go
```diff
@@ -651,15 +681,12 @@ func (m *Manager) NodeDeleted(n nodeTypes.Node) {
...
-	} {
-		if address != nil {
-			m.ipcache.Delete(address.String(), n.Source)
-		}
+		ip := address.IP.String()
+		resource := ipcacheTypes.NewResourceID(ipcacheTypes.ResourceKindNode, "", n.Name)
+
+		m.removeFromIDMD(prefix, remoteHostIdentity, resource)
+		m.ipcache.Delete(ip, n.Source)
```
joestringer, line 689: This seems to be using the non-prefix representation of the string like `192.0.2.3`, as compared with the prefix representation like `192.0.2.3/32` as like the call six lines up. However, `NodeUpdated()` doesn't seem to do an `ipcache.Upsert(...)` with this representation. So this seems to be deleting an association in the ipcache that was never created when the node was initially discovered? I don't think that this is correct.

joestringer (changes requested):

qmonnet:
I'm turning the PR to draft, please mark as ready again once it's unblocked and ready to move forwards.

christarazi:
Superseded by https://github.com/cilium/cilium/pull/28972

PR #14610, opened 2021-01-14
bpf: add LB ipip health check datapath

borkmann:
(See commit msgs.)

PR #27015, opened 2023-07-24
bpf: l4lb: fix IPIP health-check path

julianwiedmann:
If `handle_nat_fwd()` expands into a tail-call (eg. because both ENABLE_IPV4 and ENABLE_IPV6 are set), control doesn't return to the main `to-netdev` program. Thus the packet path never passes through the L4LB health-check processing.

This is a regression caused by 54a86312782f - previously, we would skip the NAT check in a full-DSR config (and the L4LB IPIP mode is always full-DSR).

Fix it by pulling the health-check section up, so that it gets applied before walking down the NAT path. Also add a minimal integration test to cover the expected encapsulation behaviour.

Fixes: 54a86312782f ("bpf: nodeport: handle revDNAT for local backends at to-netdev/to-overlay")

```release-note
Fix a bug that affected the health-check feature in Stand-alone L4LB mode. For certain configurations (eg if both IPv4 and IPv6 support is enabled) health-check traffic would not get IPIP-encapsulated.
```

### DIFF
--- a/.github/workflows/lint-bpf-checks.yaml
+++ b/.github/workflows/lint-bpf-checks.yaml
@@ -162,7 +162,7 @@ jobs:
         env:
           # Disable coverage report for these test cases since they are hitting
           # https://github.com/cilium/coverbee/issues/7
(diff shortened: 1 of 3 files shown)


viktor-kurchenko (approved):
LGTM.

julianwiedmann:
tiny update to the test, to bring it more in-line with how the feature works

julianwiedmann:
For some context: https://github.com/cilium/cilium/pull/14610

ldelossa (approved):
Changes LGTM.

PR #27023, opened 2023-07-24
gh: feature template: s/request/proposal

julianwiedmann:
Make the wording consistent. While we are happy to discuss proposals for new features, there should be no false expectation that features can be simply requested.


qmonnet (approved):
Thanks, looks good!

I didn't mind the “request”, submitting a request doesn't mean it will be granted? Anyway, it was “Feature Proposal” in the section title already and I'm all in for consistency. :+1:

PR #27029, opened 2023-07-24
ipsec, auth: Avoid out-of-order events that lead to inconsistent node IDs

### LINKED ISSUE #26954: Ever increasing Xfrm error count when ipsec is enabled

### What happened?

We recently deployed Cilium with IPsec enabled in a cluster comprising approximately 250 pods. Based on the service-level testing we conducted, everything appears to be functioning as expected. Additionally, when we inspected the `cilium_vxlan` interface using `tcpdump`, we could observe the proper flow of `esp` packets.

However, during our assessment of the encryption status using `cilium encrypt status`, we encountered a somewhat concerning number of Xfrm errors, particularly `XfrmOutPolBlock`. In one of the nodes with a higher number of pods, we observed this specific error to be at 709,924. Notably, this number increases every time we execute the `cat /proc/net/xfrm_stat` command on the Cilium agent daemonset.

For the installation of Cilium on our AKS cluster, we utilized the following Helm command in conjunction with the bring your own CNI mode:

```sh
helm upgrade --install cilium cilium/cilium --version 1.13.4 \
  --namespace kube-system \
  --set aksbyocni.enabled=true \
  --set nodeinit.enabled=true \
  --set encryption.enabled=true \
  --set encryption.type=ipsec \
  --set hubble.enabled=true \
  --set hubble.relay.enabled=true \
  --set hubble.ui.enabled=true 
```

The error counter stop as soon as we set the `encryption.enabled` false. So any insight on this issue would be appreciated.  

### Cilium Version

1.13.2 and 1.13.4

### Kernel Version

5.4.0-1106-azure

### Kubernetes Version

1.24.9

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

cilium encrypt status
```
Encryption: IPsec                     
Keys in use: 1                         
Max Seq. Number: 0x8b194/0xffffffff
Errors: 710577                    
	XfrmInNoStates: 646                       
	XfrmOutPolBlock: 709930                    
	XfrmInError: 1   
```

/proc/net/xfrm_stat
```
XfrmInError             	1
XfrmInBufferError       	0
XfrmInHdrError          	0
XfrmInNoStates          	646
XfrmInStateProtoError   	0
XfrmInStateModeError    	0
XfrmInStateSeqError     	0
XfrmInStateExpired      	0
XfrmInStateMismatch     	0
XfrmInStateInvalid      	0
XfrmInTmplMismatch      	0
XfrmInNoPols            	0
XfrmInPolBlock          	0
XfrmInPolError          	0
XfrmOutError            	0
XfrmOutBundleGenError   	0
XfrmOutBundleCheckError 	0
XfrmOutNoStates         	0
XfrmOutStateProtoError  	0
XfrmOutStateModeError   	0
XfrmOutStateSeqError    	0
XfrmOutStateExpired     	0
XfrmOutPolBlock         	709930
XfrmOutPolDead          	0
XfrmOutPolError         	0
XfrmFwdHdrError         	0
XfrmOutStateInvalid     	0
XfrmAcquireError        	0
```

### Code of Conduct

pchaigno:
We found out that the logic to allocate and deallocate node IDs had multiple bugs because it wasn't handling out-of-order events properly. In this context, out-of-order events are e.g. when pod objects are received before their corresponding node object. We discussed those issues at length in https://github.com/cilium/cilium/pull/26725 (see review) and concluded we should try to avoid having to deal with such out-of-order events altogether.

This pull request therefore removes all handling of node IDs on pod events. Node IDs are now entirely managed from node events. The node IDs are removed from the ipcache (`ep->node_id`). That simplifies the agent code a lot (all ipcache is uncoupled from node ID logic) and only slightly complicated the datapath logic (one more map lookup).

Summary of commits:
- Commits 1-4 are preparatory work in the datapath and the agent, without functional changes.
- Commit 5 moves the node ID allocation to be consistent with the deallocation.
- Commit 6 introduces a new DROP reason.
- Commits 7-8 remove the two consumers of `ep->node_id` in the datapath and replace them with an additional map lookup.
- Commits 9-10 covers the new logic of the previous commit in BPF unit tests.
- Commits 11-12 remove the node IDs from the ipcache and clean up related agent logic.
- Commit 13 fixes a leak of node IDs when the IP addresses of a node are updated.
- Commit 14 adds a runtime check for the bogus state fixed in the previous commit.
- Commits 15-16 cover the node ID lifecycle in Golang integration tests.

See commits for details.

This PR is an alternative to the solution tried at https://github.com/cilium/cilium/pull/26725.

```release-note
Fix a bug that could cause packet drops of type XfrmOutPolBlock when IPsec is enabled and node are recycled.
Fix a bug that could cause IPsec-encrypted packets to be sent to the wrong destination node when node churn is high.
```


jrfastab:
actually now that it is internal to datapath/node can we get rid of all the exported GetNodeIP, GetNodeID. But DumpNodeIDs is used from cmd so might still need that.

And one small question. On restart can we miss a NodeDelete? If that happens do we need some logic to correct or should the normal nodeDelete logic eventually get called and nodeID map will also be fixed up  (I think this is true?).

jrfastab (approved):
This is amazing!

jrfastab:
I agree with the overall assessment trying to save a map lookup in the datapath introduces too much complexity in user space to warrant the effort/complexity/risk and anyways this is going to an ipsec stack that is going to encrypt the entire thing anyways so a single map lookup is optimizing something with no/little gain.

pkg/ipcache/ipcache.go
```diff
@@ -653,14 +643,10 @@ func (ipc *IPCache) deleteLocked(ip string, source source.Source) (namedPortsCha
 		namedPortsChanged = namedPortsChanged && ipc.needNamedPorts.Load()
 	}
 
-	if newHostIP != nil {
-		nodeID = ipc.AllocateNodeID(newHostIP)
```
jrfastab, line 657: by the way I suspect this allocateNodeID from the delete path is the bug most often hit. Shouldn't have this been a deleteNodeID? Or rather don't we need some delete logic here as added in the linked PR. Anyways now we don't need to reason about it.

pkg/maps/ipcache/ipcache.go
```diff
@@ -124,14 +123,13 @@ func NewKey(ip net.IP, mask net.IPMask, clusterID uint8) Key {
 type RemoteEndpointInfo struct {
 	SecurityIdentity uint32     `align:"sec_identity"`
 	TunnelEndpoint   types.IPv4 `align:"tunnel_endpoint"`
-	NodeID           uint16     `align:"node_id"`
```
jrfastab, line 127: Do we need to patch the hole here with `Pad uint16` then when upgrade/downgrade we don't have to adjust the maps formats? Seems this just pushes the pad to the end.

pkg/ipcache/ipcache.go
pchaigno: I pushed your commit (with a new commit description) to fix this.

pchaigno:
> actually now that it is internal to datapath/node can we get rid of all the exported GetNodeIP, GetNodeID. But DumpNodeIDs is used from cmd so might still need that.

The Mutual Auth feature is also still using `GetNodeID` & co. It's probably doable to replace that, but I'd rather avoid touching Mutual Auth too much in this PR. It's already quite big in terms of commits.

> And one small question. On restart can we miss a NodeDelete? If that happens do we need some logic to correct or should the normal nodeDelete logic eventually get called and nodeID map will also be fixed up (I think this is true?).

AFAIK, yes, the NodeDelete logic gets called on restart in that case. We should manually test it though.

pkg/ipcache/types/types.go
```diff
@@ -63,7 +63,6 @@ func NewResourceID(kind ResourceKind, namespace, name string) ResourceID {
 
 // NodeIDHandler is responsible for the management of node identities.
 type NodeIDHandler interface {
```
gandro, line 65: This IPCache type is no longer needed, is it? Can we remove it and remove it from the IPCache `Configuration` completely? Thanks

pkg/datapath/linux/node_ids.go
```diff
@@ -187,6 +187,25 @@ func (n *linuxNodeHandler) unmapNodeID(ip string) error {
...
 }
 
+// diffAndUnmapNodeIPs takes two lists of node IP addresses: new and old ones.
+// It unmaps the node IP to node ID mapping for all the old IP addresses that
+// are not in the list of new IP addresses.
+func (n *linuxNodeHandler) diffAndUnmapNodeIPs(oldIPs, newIPs []nodeTypes.Address) {
+nextOldIP:
+	for _, oldAddr := range oldIPs {
+		for _, newAddr := range newIPs {
+			if newAddr.IP.String() == oldAddr.IP.String() {
```
gandro: Any reason we are not using `net.IP.Equal()` here?

gandro (commented):
My code owners mostly look excellent to me, thanks a lot for reducing the complexity in the agent code

I had two minor points that stood out to me

pkg/datapath/linux/node_ids.go
pchaigno: None. I pushed an update with only rebase + that change.

pkg/ipcache/types/types.go
pchaigno: The Mutual Auth feature still uses it:
```
$ git grep -P "ipCache.GetNodeI[PD]" pkg/auth/
pkg/auth/authmap_gc.go:                 nodeID, exists := r.ipCache.GetNodeID(addr.IP)
pkg/auth/manager.go:    nodeIP := a.ipCache.GetNodeIP(key.remoteNodeID)
```
I suspect that could be avoided by giving the auth package a reference to the node manager directly, but I'm a bit hesitant to make such changes to a feature I have very little context on. I've pinged Mutual Auth folks about this PR and Marco will review, so I'm hoping they can carry additional simplifications as follow ups.

mhofstetter: Yep, should be easy to get rid of this and depend directly on the nodemanager. Will come up with a follow up.

gandro: Ah, I see. Thanks!

gandro (approved):

bpf/lib/auth.h
```diff
@@ -10,17 +10,30 @@
...
 		.local_sec_label = local_id,
 		.remote_sec_label = remote_id,
-		.remote_node_id = remote_node_id,
 		.auth_type = auth_type,
 		.pad = 0,
 	};
+	__u16 *node_id;
+
+	if (remote_node_ip) {
+		node_ip.family = ENDPOINT_KEY_IPV4;
```
mhofstetter, line 27: just for clarification: the tunnel IP is always IPv4?

mhofstetter (approved):
proxy & servicemesh related bits LGTM! thanks a lot @pchaigno

rgo3 (approved):

bpf/lib/auth.h
pchaigno: Yep, that's correct.

pchaigno:
John already reviewed for the datapath and ipsec teams. We're missing a review for the api team, but the only change there is a new DROP reason and a bunch of generated files. CI is passing. I'm merging.

pkg/ipcache/types/types.go
mhofstetter: @gandro @pchaigno see https://github.com/cilium/cilium/pull/27106

PR #27042, opened 2023-07-25
metrics: Add workqueue metrics

### LINKED ISSUE #26122: CFP: Add workqueue metrics for cilium-agent
## Cilium Feature Proposal
        
It would be helpful to add workqueue metrics such as the depth, current depth of a workqueue, and the latency, how long an item stays in a workqueue to monitor the agent's k8s watcher performance.
https://github.com/kubernetes/client-go/blob/5a019202120ab4dd7dfb3788e5cb87269f343ebe/util/workqueue/metrics.go#L73-L90

`Resource[T k8sRuntime.Object]` relies on workqueue(enqueue items on cache.ResourceEventHandler), and some watchers, such as the EndpointSlice watcher, are using it with the latest implementation.
https://github.com/cilium/cilium/blob/9dc8dae85b50415ce4483962e2b0efb9d170b39b/pkg/k8s/resource/resource.go#L339

We sometimes see the agent's sync delay when a large number of Pods restart, and the following logs from client-go is appearing when delays occur.(v1.12) With the new `Resource[T k8sRuntime.Object]` implementation, we should monitor the workqueue instead of DeletaFIFO.

```
2023-04-17 19:09:48 level=info msg="Trace[1730758035]: \"DeltaFIFO Pop Process\" ID:app/moco-mysql-675dn,Depth:1026,Reason:slow event handlers blocking the queue (17-Apr-2023 10:09:48.471) (total time: 109ms):" subsys=klog
```

Add workqueue metrics using [workqueue.SetProvidor](https://github.com/kubernetes/client-go/blob/5a019202120ab4dd7dfb3788e5cb87269f343ebe/util/workqueue/metrics.go#L264)

Currently, there's a conflict with the controller-runtime metrics registration. So we need to fix it.

ysksuzuki:
This PR adds client-go workqueue metrics for cilium-agent

- workqueue_depth
- workqueue_adds_total
- workqueue_queue_duration_seconds
- workqueue_work_duration_seconds
- workqueue_unfinished_work_seconds
- workqueue_longest_running_processor_seconds
- workqueue_retries_total

The name label of the workqueue metrics is set GroupVersionKind.Kind resolved from a reconciled object, which is the same approach as the controller-runtime.
https://github.com/kubernetes-sigs/controller-runtime/blob/7f0c6dc440f334849b25866dd9990ac1a5a777c2/pkg/builder/controller.go#L332-L340

Without setting a name to the workqueue, client-go doesn't export the workqueue metrics.
https://github.com/kubernetes/client-go/blob/c5b1c13ccbedeb03c00ba162ef27566b0dfb512d/util/workqueue/metrics.go#L247-L260

      GitHub issue.

Fixes: #26122

### DIFF
--- a/Documentation/observability/metrics.rst
+++ b/Documentation/observability/metrics.rst
@@ -480,6 +480,21 @@ Name                                          Labels                            
 ``k8s_client_api_calls_total``                ``host``, ``method``, ``return_code``         Enabled    Number of API calls made to kube-apiserver labeled by host, method and return code
 ============================================= ============================================= ========== ===========================================================
 
+Kubernetes workqueue
+~~~~~~~~~~~~~~~~~~~~
+
+==================================================== ============================================= ========== ===========================================================
+Name                                                 Labels                                        Default    Description
+==================================================== ============================================= ========== ===========================================================
+``k8s_workqueue_depth``                              ``name``                                      Enabled    Current depth of workqueue
+``k8s_workqueue_adds_total``                         ``name``                                      Enabled    Total number of adds handled by workqueue
+``k8s_workqueue_queue_duration_seconds``             ``name``                                      Enabled    Duration in seconds an item stays in workqueue prior to request
+``k8s_workqueue_work_duration_seconds``              ``name``                                      Enabled    Duration in seconds to process an item from workqueue
+``k8s_workqueue_unfinished_work_seconds``            ``name``                                      Enabled    Duration in seconds of work in progress that hasn't been observed by work_duration. Large values indicate stuck threads. You can deduce the number of stuck threads by observing the rate at which this value increases.
+``k8s_workqueue_longest_running_processor_seconds``  ``name``                                      Enabled    Duration in seconds of the longest running processor for workqueue
+``k8s_workqueue_retries_total``                      ``name``                                      Enabled    Total number of retries handled by workqueue
+==================================================== ============================================= ========== ===========================================================
+
 IPAM
 ~~~~
 
--- a/pkg/k8s/resource/resource.go
+++ b/pkg/k8s/resource/resource.go
@@ -6,14 +6,17 @@ package resource
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"runtime"
 	"strconv"
+	"strings"
 	"sync"
 
 	corev1 "k8s.io/api/core/v1"
 	k8sRuntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 
 	"github.com/cilium/cilium/pkg/hive"
 	k8smetrics "github.com/cilium/cilium/pkg/k8s/metrics"
@@ -125,6 +128,7 @@ type options struct {
 	sourceObj   func() k8sRuntime.Object // prototype for the object before it is transformed
 	indexers    cache.Indexers           // map of the optional custom indexers to be added to the underlying resource informer
 	metricScope string                   // the scope label used when recording metrics for the resource
+	name        string                   // the name label used for the workqueue metrics
 }
 
 type ResourceOption func(o *options)
@@ -168,6 +172,13 @@ func WithMetric(scope string) ResourceOption {
 func WithIndexers(indexers cache.Indexers) ResourceOption {
 	return func(o *options) {
 		o.indexers = indexers
+	}
+}
+
+// WithName sets the name of the resource. Used for workqueue metrics.
+func WithName(name string) ResourceOption {
+	return func(o *options) {
+		o.name = name
 	}
 }
 
@@ -395,8 +406,9 @@ func (r *resource[T]) Events(ctx context.Context, opts ...EventsOpt) <-chan Even
 
 	// Create a queue for receiving the events from the informer.
 	queue := &keyQueue{
-		RateLimitingInterface: workqueue.NewRateLimitingQueue(options.rateLimiter),
-		errorHandler:          options.errorHandler,
+		RateLimitingInterface: workqueue.NewRateLimitingQueueWithConfig(options.rateLimiter,
+			workqueue.RateLimitingQueueConfig{Name: r.resourceName()}),
+		errorHandler: options.errorHandler,
 	}
 	r.mu.Lock()
 	subId := r.subId
@@ -530,6 +542,25 @@ func (r *resource[T]) Events(ctx context.Context, opts ...EventsOpt) <-chan Even
 	}()
 
 	return out
+}
+
+func (r *resource[T]) resourceName() string {
+	if r.opts.name != "" {
+		return r.opts.name
+	}
+
+	// We create a new pointer to the reconciled resource type.
+	// For example, with resource[*cilium_api_v2.CiliumNode] new(T) returns **cilium_api_v2.CiliumNode
+	// and *new(T) is nil. So we create a new pointer using reflect.New()
+	o := *new(T)
+	sourceObj := reflect.New(reflect.TypeOf(o).Elem()).Interface().(T)
+
+	gvk, err := apiutil.GVKForObject(sourceObj, scheme)
+	if err != nil {
+		return ""
+	}
+
+	return strings.ToLower(gvk.Kind)
 }
 
 // keyQueue wraps the workqueue to implement the error retry logic for a single subscriber,
--- a/pkg/k8s/resource/scheme.go
+++ b/pkg/k8s/resource/scheme.go
@@ -1,0 +1,33 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package resource
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+
+	cilium_api_v2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
+	cilium_api_v2alpha1 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2alpha1"
+	corev1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/core/v1"
+	discoveryv1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/discovery/v1"
+	discoveryv1beta1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/discovery/v1beta1"
+	networkingv1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/networking/v1"
+)
+
+var scheme = runtime.NewScheme()
+
+var localSchemeBuilder = runtime.SchemeBuilder{
+	corev1.AddToScheme,
+	discoveryv1beta1.AddToScheme,
+	discoveryv1.AddToScheme,
+	networkingv1.AddToScheme,
+	cilium_api_v2.AddToScheme,
+	cilium_api_v2alpha1.AddToScheme,
+}
+
+var AddToScheme = localSchemeBuilder.AddToScheme
+
+func init() {
+	utilruntime.Must(AddToScheme(scheme))
+}
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -54,6 +54,9 @@ const (
 	// SubsystemK8sClient is the subsystem to scope metrics related to the kubernetes client.
 	SubsystemK8sClient = "k8s_client"
 
+	// SubsystemWorkQueue is the subsystem to scope metrics related to the workqueue.
+	SubsystemWorkQueue = "k8s_workqueue"
+
 	// SubsystemKVStore is the subsystem to scope metrics related to the kvstore.
 	SubsystemKVStore = "kvstore"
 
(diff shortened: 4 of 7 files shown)


ysksuzuki:
workqueue_depth
workqueue_depth

workqueue_adds_total
workqueue_adds_total

workqueue_queue_duration_seconds
workqueue_queue_duration_seconds

workqueue_work_duration_seconds
workqueue_work_duration_seconds

workqueue_unfinished_work_seconds
workqueue_unfinished_work_seconds

workqueue_longest_running_processor_seconds
workqueue_longest_running_processor_seconds


workqueue_retries_total
workqueue_retries_total

ysksuzuki:
In v1.14 and later, we can't opt out of the client-go metrics such as `cilium_k8s_client_api_calls_total` because all metrics are registered to the registry, even if wet set `-cilium_k8s_client_api_calls_total` in the cilium-config cm.
https://github.com/cilium/cilium/blob/fded1abcf310160a9c518b6e25f1ac6059e08a97/pkg/metrics/registry.go#L172-L176

And all the metrics are initialized here regardless of the metrics flag.
https://github.com/cilium/cilium/blob/fded1abcf310160a9c518b6e25f1ac6059e08a97/pkg/metrics/metrics.go#L652

client-go isn't aware of `WithMetadata.IsEnabled`.
https://github.com/cilium/cilium/blob/fded1abcf310160a9c518b6e25f1ac6059e08a97/pkg/metrics/metric/metric.go#L13C1-L17

Ref: https://github.com/cilium/cilium/pull/25651

We can't opt out of the workqueue metrics as well.  Personally, I don't feel the need to opt out of the client-go metrics. What do you think?

Documentation/observability/metrics.rst
```diff
@@ -461,6 +461,21 @@ Name                                          Labels
...
 
+Kubernetes workqueue
+~~~~~~~~~~~~~~~~~~~~
+
+================================================ ============================================= ========== ===========================================================
+Name                                             Labels                                        Default    Description
+================================================ ============================================= ========== ===========================================================
+``workqueue_depth``                              ``name``                                      Enabled    Current depth of workqueue
+``workqueue_adds_total``                         ``name``                                      Enabled    Total number of adds handled by workqueue
+``workqueue_queue_duration_seconds``             ``name``                                      Enabled    How long in seconds an item stays in workqueue before being requested
```
zacharysarah: Clarity nits
```suggestion
``workqueue_queue_duration_seconds``             ``name``                                      Enabled    Duration in seconds an item stays in workqueue prior to request
```

Documentation/observability/metrics.rst
```diff
@@ -461,6 +461,21 @@ Name                                          Labels
...
+``workqueue_work_duration_seconds``              ``name``                                      Enabled    How long in seconds processing an item from workqueue takes
```
zacharysarah:
```suggestion
``workqueue_work_duration_seconds``              ``name``                                      Enabled    Duration in seconds to process an item from workqueue
```

Documentation/observability/metrics.rst
```diff
@@ -461,6 +461,21 @@ Name                                          Labels
...
+``workqueue_unfinished_work_seconds``            ``name``                                      Enabled    How many seconds of work has been done that is in progress and hasn't been observed by work_duration. Large values indicate stuck threads. One can deduce the number of stuck threads by observing the rate at which this increases
```
zacharysarah:
```suggestion
``workqueue_unfinished_work_seconds``            ``name``                                      Enabled    Duration in seconds of work in progress that hasn't been observed by work_duration. Large values indicate stuck threads. You can deduce the number of stuck threads by observing the rate at which this value increases.
```

Documentation/observability/metrics.rst
```diff
@@ -461,6 +461,21 @@ Name                                          Labels
...
+``workqueue_longest_running_processor_seconds``  ``name``                                      Enabled    How many seconds has the longest running processor for workqueue been running
```
zacharysarah:
```suggestion
``workqueue_longest_running_processor_seconds``  ``name``                                      Enabled    Duration in seconds of the longest running processor for workqueue
```

zacharysarah (changes requested):
@ysksuzuki I've added some changes to improve clarity, otherwise LGTM for docs.

ysksuzuki:
@zacharysarah Thank you for the review! I've reflected your suggestion accordingly.

zacharysarah (approved):
Thanks for the updates, LGTM for docs

youngnick (approved):
LGTM for k8s, nice work.

I think the naming is fine, and that it's fine that we can't turn these metrics off.

pkg/metrics/metrics.go
```diff
@@ -53,6 +53,9 @@ const (
 	// SubsystemK8sClient is the subsystem to scope metrics related to the kubernetes client.
 	SubsystemK8sClient = "k8s_client"
 
+	// SubsystemWorkQueue is the subsystem to scope metrics related to the workqueue.
+	SubsystemWorkQueue = "workqueue"
```
christarazi: I think it would be useful to disambiguate here because there are a few queues in Cilium. I think calling out that this is a K8s construct is useful, so my suggestion would be `k8s_workqueue`.

pkg/metrics/metrics.go
```diff
@@ -1291,6 +1375,20 @@ func NewLegacyMetrics() *LegacyMetrics {
...
+
+		WorkQueueLatency: WorkQueueLatency,
+
+		WorkQueueDuration: WorkQueueDuration,
+
+		WorkQueueUnfinishedWork: WorkQueueUnfinishedWork,
+
+		WorkQueueLongestRunningProcessor: WorkQueueLongestRunningProcessor,
+
+		WorkQueueRetries: WorkQueueRetries,
```
christarazi: Nit: unnecessary newlines between each line?

christarazi:
> In v1.14 and later, we can't opt out of the client-go metrics such as `cilium_k8s_client_api_calls_total` because all metrics are registered to the registry, even if wet set `metrics: -cilium_k8s_client_api_calls_total` in the cilium-config cm.
> 
> https://github.com/cilium/cilium/blob/fded1abcf310160a9c518b6e25f1ac6059e08a97/pkg/metrics/registry.go#L172-L176
> 
> And all the metrics are initialized here regardless of the metrics flag.
> 
> https://github.com/cilium/cilium/blob/fded1abcf310160a9c518b6e25f1ac6059e08a97/pkg/metrics/metrics.go#L652
> 
> client-go isn't aware of `WithMetadata.IsEnabled`. https://github.com/cilium/cilium/blob/fded1abcf310160a9c518b6e25f1ac6059e08a97/pkg/metrics/metric/metric.go#L13C1-L17
> 
> Ref: #25651
> 
> We can't opt out of the workqueue metrics as well. Personally, I don't feel the need to opt out of the client-go metrics. What do you think?

I don't think it's a problem for this PR or generally. However, if someone does want to disable them, then it should be doable by wrapping the client-go metrics with a type we control in order to implement the `WithMetadata` interface, right?

pkg/k8s/resource/resource.go
```diff
@@ -532,6 +543,25 @@ func (r *resource[T]) Events(ctx context.Context, opts ...EventsOpt) <-chan Even
...
+func (r *resource[T]) resourceName() string {
+	if r.opts.name != "" {
+		return r.opts.name
+	}
+
+	// We create a new pointer to the reconciled resource type.
+	// For example, with resource[*cilium_api_v2.CiliumNode] new(T) returns **cilium_api_v2.CiliumNode
+	// and *new(T) is nil. So we create a new pointer using reflect.New()
+	o := *new(T)
+	sourceObj := reflect.New(reflect.TypeOf(o).Elem()).Interface().(T)
```
christarazi, line 556: I'm not sure I understand what would happen if we don't have this code.

pkg/k8s/resource/scheme.go
```diff
@@ -0,0 +1,107 @@
...
+
+func init() {
+	utilruntime.Must(AddToScheme(scheme))
+}
+
+// referenced from https://github.com/kubernetes-sigs/controller-runtime/blob/93cd3cda0fdddc1f703634a881097c246dcabbd3/pkg/client/apiutil/apimachinery.go#L113
+// We are doing it this way to avoid the direct dependency on controller-runtime, otherwise conflicts in metric registration will occur
+// controller-runtime automatically registers metrics such as k8s rest client and workqueue metrics on its registry,
+// and cilium-agent also registers the same metrics on its registry. Those metrics registrations are exclusive.
+func gvkForObject(obj runtime.Object) (schema.GroupVersionKind, error) {
```
christarazi: Where does [apiutil](https://github.com/kubernetes-sigs/controller-runtime/tree/93cd3cda0fdddc1f703634a881097c246dcabbd3/pkg/client/apiutil) automatically register its metrics?

christarazi (commented):
Thanks for the PR, very useful!

A few comments below.

pkg/k8s/resource/scheme.go
ysksuzuki: On second thoughts, depending on `sigs.k8s.io/controller-runtime/pkg/client/apiutil` doesn't cause the metrics conflict because the workqueue metrics are registered in the different package
`sigs.k8s.io/controller-runtime/pkg/metrics`. So we can drop the gvkForObject function here and use the original function. Thank you for your comment!

https://github.com/kubernetes-sigs/controller-runtime/blob/f30e11d97fd19564e71b4f672efcdc0ecad3a8d1/pkg/metrics/workqueue.go#L90-L100

pkg/k8s/resource/resource.go
ysksuzuki: Without this code, we won't get the GVK for the reconciled object. GVKForObject finds the GVK associated with the given object and if we pass a nil pointer it can't find it.

https://github.com/kubernetes/apimachinery/blob/3e2600dc79feea6cdc8a9224bc8a6a7fcfee1466/pkg/runtime/scheme.go#L243

christarazi (approved):

ysksuzuki:
@derailed @nathanjsweet Would you be able to take a look at this PR?

pkg/k8s/resource/resource.go
```diff
@@ -532,6 +544,25 @@ func (r *resource[T]) Events(ctx context.Context, opts ...EventsOpt) <-chan Even
...
+
+	gvk, err := apiutil.GVKForObject(sourceObj, scheme)
+	if err != nil {
+		return ""
+	}
+
+	return strings.ToLower(gvk.Kind)
```
derailed, line 563: I could be off here but shouldn't the resource name be the pluralize version of kind besides just lower?

pkg/metrics/metrics.go
```diff
@@ -565,6 +568,80 @@ var (
...
+	// We set actual metrics here instead of NoOp for the workqueue metrics
+	// because these metrics will be registered with workqueue.SetProvider
+	// by init function in watcher.go. Otherwise, we will register NoOps.
+	//
+	WorkQueueDepth = metric.NewGaugeVec(metric.GaugeOpts{
+		ConfigName: Namespace + "_" + SubsystemWorkQueue + "_depth",
+		Namespace:  Namespace,
+		Subsystem:  SubsystemWorkQueue,
+		Name:       "depth",
+		Help:       "Current depth of workqueue",
```
derailed: nit: Consistency. Some help msgs contain punctuation others don't

pkg/metrics/metrics.go
```diff
@@ -565,6 +568,80 @@ var (
...
 	APILimiterProcessedRequests = NoOpCounterVec
+
+	// WorkQueueDepth is the depth of the workqueue
+	//
+	// We set actual metrics here instead of NoOp for the workqueue metrics
+	// because these metrics will be registered with workqueue.SetProvider
+	// by init function in watcher.go. Otherwise, we will register NoOps.
+	//
+	WorkQueueDepth = metric.NewGaugeVec(metric.GaugeOpts{
+		ConfigName: Namespace + "_" + SubsystemWorkQueue + "_depth",
```
derailed, line 564: nit: Perhaps a helper here might be in order to dry out the config naming?

pkg/metrics/metrics.go
```diff
@@ -565,6 +568,80 @@ var (
...
+		Help:       "Total number of adds handled by workqueue",
+	}, []string{"name"})
+
+	// WorkQueueLatency is the latency of how long an item stays in the workqueue
+	WorkQueueLatency = metric.NewHistogramVec(metric.HistogramOpts{
+		ConfigName: Namespace + "_" + SubsystemWorkQueue + "_queue_duration_seconds",
+		Namespace:  Namespace,
+		Subsystem:  SubsystemWorkQueue,
+		Name:       "queue_duration_seconds",
+		Help:       "How long in seconds an item stays in workqueue before being requested",
```
derailed: nit: requested vs processed?

pkg/metrics/metrics.go
```diff
@@ -647,6 +724,13 @@ type LegacyMetrics struct {
 	APILimiterRateLimit              metric.Vec[metric.Gauge]
 	APILimiterAdjustmentFactor       metric.Vec[metric.Gauge]
 	APILimiterProcessedRequests      metric.Vec[metric.Counter]
+	WorkQueueDepth                   metric.Vec[metric.Gauge]
```
derailed, line 708: Might be clearer to encapsulate these into their own type aka `workQueueMetrics` and aggregate into LegacyMetrics?

derailed (changes requested):
@ysksuzuki Thank you for your PR! Nice work Yusuke!! Just a few small picks...

pkg/k8s/resource/resource.go
ysksuzuki: How can we pluralize the kind here? `GroupVersionKind` doesn't have a pluralized version of kind.
https://github.com/cilium/cilium/blob/6a70af21c3b6985096fca3f8240139b548ac4760/vendor/k8s.io/apimachinery/pkg/runtime/schema/group_version.go#L142-L146

Personally, I don't see the need to make it plural here because this function returns the resource name that the workqueue handles. If a caller passes the type `CiliumNetworkPolicyList`, then the kind is plural, but it wouldn't be used that way.

pkg/metrics/metrics.go
ysksuzuki: I think we can't define our own type here. It breaks the assumption of `cell.Metrics`

https://github.com/cilium/cilium/blob/84ea38311084fffc40a58d26c05386eedb04e6a7/pkg/hive/cell/metric.go#L28-L31

pkg/metrics/metrics.go
ysksuzuki: Sorry, I'm not following. Can you elaborate?

ysksuzuki: If you are talking about the order, the order here is the same as the order of fields in `MetricsProvidor`
https://github.com/cilium/cilium/blob/e6748209b27ed24b319265ac45b517b9af9b516f/vendor/k8s.io/client-go/util/workqueue/metrics.go#L173-L181

pkg/metrics/metrics.go
derailed: Ah... I stand corrected. Good catch! Thank you Yusuke!! Please disregard.

pkg/k8s/resource/resource.go
youngnick: I'm sorry, I missed this before, thanks for catching it  @derailed.

By the k8s API standards, yes, the resource name should _generally_ be the lower-case, pluralized Kind (that is, `Pod` Kind = `pods` resource). But this is a convention, not an invariant rule, so it's not _guaranteed_. (The difference is actually that you can put a _resource_ name directly into a GET URL to the apiserver, and it will work, while a Kind will not.)

But, as @ysksuzuki says, you can't just add an "s" and call it a day. The upstream k8s code will actually do lookups to the apiserver to get the canonical resource name for a GVK - because the apiserver holds mappings for every GVK to resource (so that it can generate URLs itself).

That means, however, that you need a dynamic client, and to use the restmapper package to generate a dynamic restmapper, that you then pass the GVK to, that supplies you with a mapper object that contains the GVR for the given GVK.

I couldn't find old code where I've done this before, but this example, taken from (https://ymmt2005.hatenablog.com/entry/2020/04/14/An_example_of_using_dynamic_client_of_k8s.io/client-go) seems about right:
```go
import (
    "k8s.io/apimachinery/pkg/api/meta"
    "k8s.io/apimachinery/pkg/runtime/schema"
    "k8s.io/client-go/discovery"
    "k8s.io/client-go/discovery/cached/memory"
    "k8s.io/client-go/rest"
    "k8s.io/client-go/restmapper"
)

// find the corresponding GVR (available in *meta.RESTMapping) for gvk
func findGVR(gvk *schema.GroupVersionKind, cfg *rest.Config) (*meta.RESTMapping, error) {

    // DiscoveryClient queries API server about the resources
    dc, err := discovery.NewDiscoveryClientForConfig(cfg)
    if err != nil {
        return nil, err
    }
    mapper := restmapper.NewDeferredDiscoveryRESTMapper(memory.NewMemCacheClient(dc))

    return mapper.RESTMapping(gvk.GroupKind(), gvk.Version)
}
```

That's the only way to _guarantee_ that you've got the correct values completely generically.

_However_, in this case, if we're controlling all the objects this is used for, we _may_ be able to get away with what we have here.

ysksuzuki: @youngnick Thank you for your comment!

To be clear, the `resourceName()` function here resolves GVK from a given Go struct type to add `GVK.Kind` to a name label of the workqueue metrics. It's not for finding GVR and requesting resources from kube-apiserver using a dynamic client.　https://github.com/cilium/cilium/pull/27042#issuecomment-1658419028

Regarding the rest API client, the current implementation of `resource[T k8sRuntime.Object]` requires to pass ListWatcher with a typed client to callers.
https://github.com/cilium/cilium/blob/9cdcddc67061c59df7b517ef52a1d5a8f0b700c1/pkg/k8s/resource/resource.go#L71-L74

youngnick: 🤦 Sorry for the noise then, I guess that's on me for not reading the rest of the code properly and assuming what `resource` meant.

pkg/metrics/metrics.go
ysksuzuki: WorkQueueLatency is the latency of how long an item stays in workqueue before a consumer requests the item from it. The consumer will start processing the item after that.
https://github.com/kubernetes/client-go/blob/21b4553e4404bae3fec9ff9ef7204beabcfd3c27/util/workqueue/metrics.go#L112

derailed (approved):
Nice work! LGTM

pkg/metrics/metrics.go
ysksuzuki: I have added punctuation to all the help messages.

michi-covalent:
i'm merging this. the button is green. review from nate is pending, but the review requirement from sig-policy team is covered by the review from chris.

PR #27047, opened 2023-07-25
Remove unnecessary type conversions in fqdn zombies handling

giorio94:
Let's drop the netip.Addr to string to netip.Addr roundtrip conversions, which introduce unnecessary performance penalties, and circumvent type safety relying on functions which might panic.


mhofstetter (approved):
:+1: thanks @giorio94!

youngnick (approved):

PR #27048, opened 2023-07-25
[1.13] Add documentation about bpf-clock breakage

margamanterola:
We've disabled bpfClockProbe everywhere with #26981 and backports. Document that long lived connections will be interrupted when upgrading to and from 1.13.4, the only release to ship with this enabled.

```release-note
Document that upgrades to 1.13.4 may experience interruptions of existing connections, while upgrades from 1.13.4 may encounter lingering connections.
```

### DIFF
--- a/Documentation/operations/upgrade.rst
+++ b/Documentation/operations/upgrade.rst
@@ -286,7 +286,9 @@ from ``1.10.x`` to ``1.11.y`` first, and to ``1.12.z`` only afterwards).
 +-----------------------+-----------------------+-------------------------+---------------------------+
 | Current version       | Target version        | L3/L4 impact            | L7 impact                 |
 +=======================+=======================+=========================+===========================+
-| ``1.12.x``            | ``1.13.y``            | Minimal to None         | Clients must reconnect[1] |
+| ``1.13.x (x <= 3)``   | ``1.13.y (y >= 5)``   | Minimal to None         | Clients must reconnect[1] |
++-----------------------+-----------------------+-------------------------+---------------------------+
+| ``1.12.x``            | ``1.13.y (y >= 5)``   | Minimal to None         | Clients must reconnect[1] |
 +-----------------------+-----------------------+-------------------------+---------------------------+
 | ``1.11.x``            | ``1.12.y``            | Minimal to None         | Clients must reconnect[1] |
 +-----------------------+-----------------------+-------------------------+---------------------------+
@@ -303,6 +305,15 @@ Annotations:
 .. _current_release_required_changes:
 
 .. _1.13_upgrade_notes:
+
+1.13.4 (to and from) Upgrade Notes
+----------------------------------
+
+* Existing connections may be interrupted when upgrading to v1.13.4, while they
+  may linger for too long when upgrading from v1.13.4.  This is due to a
+  difference in how the age of the connections is calculated when using
+  bpf-clock-probe (:gh-issue:`26955`). To minimize disruption during upgrades,
+  we recommend skipping this release and upgrading directly to a newer version.
 
 1.13.3+ Upgrade Notes
 ---------------------
@@ -435,6 +446,11 @@ Helm Options
   domain socket will be removed and the Kubernetes peer-svc mandatory when
   Hubble Relay is enabled. Users overriding ``hubble.peerService.enabled`` to
   ``false`` are encouraged to migrate to ``true`` with the Cilium v1.13 release.
+* ``bpfClockProbe`` has been set to ``false`` everywhere to avoid interrupting
+  existing connections. It can still be manually set to ``true`` if the
+  BPF clock probe behavior is desired.
+* The previous ``bpf.clockProbe`` flag used in ``values.yaml`` was incorrectly
+  named. The correct name is ``bpfClockProbe``.
 
 CRD Changes
 ~~~~~~~~~~~


qmonnet (approved):
Looks good, thanks

pchaigno (approved):

Documentation/operations/upgrade.rst
```diff
@@ -435,6 +450,9 @@ Helm Options
   domain socket will be removed and the Kubernetes peer-svc mandatory when
   Hubble Relay is enabled. Users overriding ``hubble.peerService.enabled`` to
   ``false`` are encouraged to migrate to ``true`` with the Cilium v1.13 release.
+* ``bpfClockProbe`` has been set to ``false`` everywhere to avoid interrupting
+  existing connections. It can still be manually set to ``true`` if the
+  ``bpf-clock-probe`` behavior is desired.
```
aanm: I'm confused about this. `bpfClockProbe` is the helm option, shouldn't it be "`bpfClockProbe` behavior is desired."?

Documentation/operations/upgrade.rst
```diff
@@ -283,27 +283,42 @@ combination is not listed in the table below, then it may not be safe. In that
...
 .. _1.13_upgrade_notes:
 
+1.13.4 (to and from) Upgrade Notes
+----------------------------------
+
+* Existing connections may be interrupted when upgrading to v1.13.4, while they
+  may linger for too long when upgrading from v1.13.4.  This is due to a
+  difference in how the age of the connections is calculated when using
+  bpf-clock-probe (:gh-issue:`26955`).
+
```
aanm: We should have a place documenting the helm flag change `bpf.clockProbe` -> `bpfClockProbe`

aanm (changes requested):

Documentation/operations/upgrade.rst
margamanterola: I think it should actually be "if the BPF clock probe behavior is desired" (i.e. no code text, this is talking about the underlying behavior), what do you think?

Documentation/operations/upgrade.rst
margamanterola: The previous flag never worked. If we are writing something, it would be something like:
"the previous Helm flag `bpf.clockProbe` used in `values.yaml` didn't work, the correct flag name is `bpfClockProbe`"

I'm not against documenting it, but I don't know where I should put something like that.

Documentation/operations/upgrade.rst
```diff
@@ -283,27 +283,44 @@ combination is not listed in the table below, then it may not be safe. In that
...
-| ``1.10.x``            | ``1.11.y``            | Minimal to None         | Clients must reconnect[1] |
-+-----------------------+-----------------------+-------------------------+---------------------------+
++-----------------------+-----------------------+----------------------------+---------------------------+
+| Current version       | Target version        | L3/L4 impact               | L7 impact                 |
++=======================+=======================+============================+===========================+
+| ``1.13.4``            | ``1.13.y``            | Lingering Connections[2]   | Clients must reconnect[1] |
++-----------------------+-----------------------+----------------------------+---------------------------+
+| ``1.13.x``            | ``1.13.4``            | Interrupted Connections[3] | Clients must reconnect[1] |
++-----------------------+-----------------------+----------------------------+---------------------------+
+| ``1.12.x``            | ``1.13.y``            | Minimal to None            | Clients must reconnect[1] |
```
joestringer: Couple of notes:

- For 1.13.x (x <= 3) upgrade to 1.13.y (y >= 5), it should also be minimal to none.
- For 1.12.x to 1.13.y we should probably recommend y >= 5 

Given the complexity of enumerating these, and given that we'll have a version with minimal disruption coming soon, I'm not sure it is worth documenting in this specific table. It gets awfully complicated to list the combinations.

Currently for any combinations that have minimal to no impact, we list the versions. I think that implies there are no guarantee for other versions.

One counter-proposal here would be just to update the existing table to:
1. Add 1.13.x (x <= 3) to 1.13.y (y >= 5) has minimal to none disruption
2. Update the 1.12.x to 1.13.y to specify y >= 5.

Documentation/operations/upgrade.rst
joestringer: Usually we would document such changes here: https://docs.cilium.io/en/v1.13/operations/upgrade/#helm-options

Some phrasing similar to this could work:

> The Helm option `bpf.clockProbe` was not functional previously. The setting `bpfClockProbe` now configures the clock probe mechanism for BPF timestamps.

joestringer: Never mind my suggestion, it looks like you already addressed this. LGTM.

Documentation/operations/upgrade.rst
joestringer: Related, we never assert support for older patch releases. If someone is running 1.13.0 today, it's clear that they should upgrade to the latest 1.13.y release in order to adopt all bugfixes. After newer patch releases are out, it's unusual for users to decide to upgrade to the N-1 patch release if the N patch release is out. From that perspective, I think it makes sense to (a) assume that users will attempt to upgrade to the latest patch release for the series and document well for that scenario, and (b) have a separate note that clearly defines what happens if they upgrade to the version that has the issues (1.13.4). The rest of the PR mostly covers (b) already. My feedback above on this table is trying to improve (a).

Documentation/operations/upgrade.rst
margamanterola: Reworded as mentioned. PTAL

Documentation/operations/upgrade.rst
margamanterola: I've added your feedback (I think). Can you verify this is what you wanted?

joestringer (approved):
LGTM, thanks.

youngnick (approved):

aanm (approved):

PR #27054, opened 2023-07-25
cilium: use absolute path to include Makefile.defs

lmb:
cilium/Makefile currently can't be included from elsewhere, since it uses a relative path to Makefile.defs. Copy the approach used for daemon and operator.

At the same time, avoid a possible shell splitting issue by quoting the arguments to dirname.


joamaki (approved):

derailed (approved):
LGTM

PR #27062, opened 2023-07-25
Fix NatDumpCreated

gentoo-root:
See the commit messages.

```release-note
Fix the Created timestamps in `cilium bpf nat list` that used to display the same values.
```

### DIFF
--- a/pkg/maps/nat/nat.go
+++ b/pkg/maps/nat/nat.go
@@ -9,9 +9,11 @@ import (
 
 	"github.com/cilium/ebpf"
 
+	"github.com/cilium/cilium/api/v1/models"
 	"github.com/cilium/cilium/pkg/bpf"
 	"github.com/cilium/cilium/pkg/logging"
 	"github.com/cilium/cilium/pkg/logging/logfields"
+	"github.com/cilium/cilium/pkg/maps/timestamp"
 	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/tuple"
 )
@@ -47,7 +49,7 @@ type NatEntry interface {
 	ToHost() NatEntry
 
 	// Dumps the Nat entry as string.
-	Dump(key NatKey, start uint64) string
+	Dump(key NatKey, toDeltaSecs func(uint64) string) string
 }
 
 // A "Record" designates a map entry (key + value), but avoid "entry" because of
@@ -66,14 +68,6 @@ type NatMap interface {
 	Path() (string, error)
 	DumpEntries() (string, error)
 	DumpWithCallback(bpf.DumpCallback) error
-}
-
-// NatDumpCreated returns time in seconds when NAT entry was created.
-func NatDumpCreated(dumpStart, entryCreated uint64) string {
-	tsecCreated := entryCreated / 1000000000
-	tsecStart := dumpStart / 1000000000
-
-	return fmt.Sprintf("%dsec", tsecStart-tsecCreated)
 }
 
 // NewMap instantiates a Map.
@@ -117,22 +111,50 @@ func (m *Map) DumpReliablyWithCallback(cb bpf.DumpCallback, stats *bpf.DumpStats
 	return (&m.Map).DumpReliablyWithCallback(cb, stats)
 }
 
-// DoDumpEntries iterates through Map m and writes the values of the
-// nat entries in m to a string.
-func DoDumpEntries(m NatMap) (string, error) {
+// DumpEntriesWithTimeDiff iterates through Map m and writes the values of the
+// nat entries in m to a string. If clockSource is not nil, it uses it to
+// compute the time difference of each entry from now and prints that too.
+func DumpEntriesWithTimeDiff(m NatMap, clockSource *models.ClockSource) (string, error) {
+	var toDeltaSecs func(uint64) string
 	var sb strings.Builder
 
-	nsecStart, _ := bpf.GetMtime()
+	if clockSource == nil {
+		toDeltaSecs = func(t uint64) string {
+			return fmt.Sprintf("? (raw %d)", t)
+		}
+	} else {
+		now, err := timestamp.GetCTCurTime(clockSource)
+		if err != nil {
+			return "", err
+		}
+		tsConverter, err := timestamp.NewCTTimeToSecConverter(clockSource)
+		if err != nil {
+			return "", err
+		}
+		tsecNow := tsConverter(now)
+		toDeltaSecs = func(t uint64) string {
+			tsec := tsConverter(uint64(t))
+			diff := int64(tsecNow) - int64(tsec)
+			return fmt.Sprintf("%dsec ago", diff)
+		}
+	}
+
 	cb := func(k bpf.MapKey, v bpf.MapValue) {
 		key := k.(NatKey)
 		if !key.ToHost().Dump(&sb, false) {
 			return
 		}
 		val := v.(NatEntry)
-		sb.WriteString(val.ToHost().Dump(key, nsecStart))
+		sb.WriteString(val.ToHost().Dump(key, toDeltaSecs))
 	}
 	err := m.DumpWithCallback(cb)
 	return sb.String(), err
+}
+
+// DoDumpEntries iterates through Map m and writes the values of the
+// nat entries in m to a string.
+func DoDumpEntries(m NatMap) (string, error) {
+	return DumpEntriesWithTimeDiff(m, nil)
 }
 
 // DumpEntries iterates through Map m and writes the values of the
--- a/pkg/maps/timestamp/timestamp.go
+++ b/pkg/maps/timestamp/timestamp.go
@@ -1,0 +1,98 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package timestamp
+
+import (
+	"fmt"
+
+	"github.com/cilium/cilium/api/v1/client/daemon"
+	"github.com/cilium/cilium/api/v1/models"
+	"github.com/cilium/cilium/pkg/bpf"
+	"github.com/cilium/cilium/pkg/datapath/linux/probes"
+	"github.com/cilium/cilium/pkg/option"
+	"github.com/cilium/cilium/pkg/time"
+)
+
+const (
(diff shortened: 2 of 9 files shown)


cilium/cmd/bpf_ct_list.go
```diff
@@ -104,7 +104,11 @@ func getMaps(t string, id uint32) []*ctmap.Map {
 func getClockSource() (*models.ClockSource, error) {
 	switch timeDiffClockSourceMode {
 	case "":
-		return timestamp.GetClockSourceFromAgent(client.Daemon)
+		clockSource, err := timestamp.GetClockSourceFromAgent(client.Daemon)
+		if err != nil {
+			clockSource, err = timestamp.GetClockSourceFromRuntimeConfig()
```
aspsk: Maybe print error the same way as below?

youngnick (approved):

cilium/cmd/bpf_ct_list.go
```diff
@@ -102,26 +101,14 @@ func getMaps(t string, id uint32) []*ctmap.Map {
...
-	return resp.Payload.ClockSource, nil
-}
-
 func getClockSource() (*models.ClockSource, error) {
 	switch timeDiffClockSourceMode {
 	case "":
-		return getClockSourceFromAgent()
+		clockSource, err := timestamp.GetClockSourceFromAgent(client.Daemon)
+		if err != nil {
+			clockSource, err = timestamp.GetClockSourceFromRuntimeConfig()
```
derailed: Might as well return directly here.

pkg/maps/nat/nat.go
```diff
@@ -115,24 +109,51 @@ func (m *Map) DumpReliablyWithCallback(cb bpf.DumpCallback, stats *bpf.DumpStats
...
+		now, err := timestamp.GetCTCurTime(clockSource)
+		if err != nil {
+			return "", err
+		}
+		tsecNow, err := timestamp.ConvertCTTimeToSec(now, clockSource)
+		if err != nil {
+			return "", err
+		}
+		toDeltaSecs = func(t uint64) string {
+			tsec, _ := timestamp.ConvertCTTimeToSec(t, clockSource)
```
derailed: Missing err check??

pkg/maps/nat/nat.go
```diff
@@ -115,24 +109,51 @@ func (m *Map) DumpReliablyWithCallback(cb bpf.DumpCallback, stats *bpf.DumpStats
...
 	cb := func(k bpf.MapKey, v bpf.MapValue) {
 		key := k.(NatKey)
 		if !key.ToHost().Dump(&sb, false) {
 			return
 		}
 		val := v.(NatEntry)
-		sb.WriteString(val.ToHost().Dump(key, nsecStart))
+		sb.WriteString(val.ToHost().Dump(key, toDeltaSecs))
 	}
 	err := m.DumpWithCallback(cb)
```
derailed, line 150: nit: might as well return directly `return sb.String(), m.DumpWithCallback(cb)`

pkg/maps/timestamp/timestamp.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Get the clocksource from the agent config file in case agent is not running.
+func GetClockSourceFromRuntimeConfig() (*models.ClockSource, error) {
+	var config option.DaemonConfig
+
+	agentConfigFile := filepath.Join(defaults.RuntimePath, defaults.StateDir,
+		"agent-runtime-config.json")
+
+	if byteValue, err := os.ReadFile(agentConfigFile); err == nil {
+		err = json.Unmarshal(byteValue, &config)
```
derailed, line 66: if unmarshal fails do we still want to make the call with a zero value config??

pkg/maps/timestamp/timestamp.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		return j >> bpfMonoScaler, nil
+	default:
+		return 0, fmt.Errorf("invalid clocksource: %s", clockSource.Mode)
+	}
+}
+
+// Converts a CT timestamp from clocksource units into seconds.
+func ConvertCTTimeToSec(timestamp uint64, clockSource *models.ClockSource) (uint64, error) {
+	switch clockSource.Mode {
```
derailed, line 106: This will bomb is clockSource is nil.

pkg/maps/timestamp/timestamp.go
derailed: Also might be clearer to return right away is the read fail.

pkg/maps/timestamp/timestamp.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Get current clocksource - to be used in the agent context.
+func GetClockSourceFromOptions() *models.ClockSource {
+	return getClockSourceFromConfig(option.Config)
+}
+
+// Connect to the agent via API and get its current clocksource.
+func GetClockSourceFromAgent(svc daemon.ClientService) (*models.ClockSource, error) {
+	params := daemon.NewGetHealthzParamsWithTimeout(5 * time.Second)
+	brief := false
```
derailed, line 44: nit: better to use zero val aka `var brief bool`

pkg/maps/timestamp/timestamp.go
derailed: Also might be great if we can to add some tests (if possible) around this stuff.

derailed (changes requested):
Nice work! just a few picks...

aspsk (approved):

cilium/cmd/bpf_ct_list.go
gentoo-root: ACK.

pkg/maps/nat/nat.go
gentoo-root: Not at all! If clockSource was faulty, the invocation of ConvertCTTimeToSec above would return an error, where it would be handled. If the invocation above succeeded, this one can't fail. toDeltaSecs can't return an error anyway.

pkg/maps/nat/nat.go
gentoo-root: That's existing code, I didn't touch it in my patches.

pkg/maps/timestamp/timestamp.go
gentoo-root: ACK.

pkg/maps/timestamp/timestamp.go
gentoo-root: > This will bomb is clockSource is nil.

Why would it be nil? Did you catch some flow where nil is passed here? Or do you have an idea how to make it impossible to pass nil here?

> Also might be great if we can to add some tests (if possible) around this stuff.

Makes sense, ACK.

pkg/maps/timestamp/timestamp.go
gentoo-root: Could you explain why it's better? Or if it's some standard coding style guideline, could you point me at it?

In my opinion, this conveys the idea much better:

```
brief := false
params.SetBrief(&brief)
```

than this:

```
var brief bool
params.SetBrief(&brief)
```

If I saw the second snippet, I would be confused and would have thought `SetBrief` returns a value into `brief`.

cilium/cmd/bpf_ct_list.go
gentoo-root: Like where below? I don't seem to understand what you are referring to...

getIpv6EnableStatus() does a similar thing (querying the agent and falling back to the config file) and falls back silently if an error on the first step occurs.

pkg/maps/timestamp/timestamp.go
derailed: This function is exported hence we can't control who is going to call it and what params will be passed in. 
We're making the assumption that clockSource is legit but should we?

pkg/maps/nat/nat.go
derailed: We are checking for errs right above with a presumably valid clocksource and a known timestamp. 
That func can take adhoc timestamp so perhaps log the error since we can't return it?

pkg/maps/timestamp/timestamp_test.go
```diff
@@ -0,0 +1,37 @@
...
+func Test(t *testing.T) {
+	TestingT(t)
+}
+
+type TimestampSuite struct{}
+
+var _ = Suite(&TimestampSuite{})
+
+func (s *TimestampSuite) TestConvert(c *C) {
+	clockSource := &models.ClockSource{Mode: models.ClockSourceModeKtime}
```
derailed: We are only testing happy path here. What if it isn't so?

derailed (commented):
@gentoo-root Thank you for the updates. Just a few minor comments

pkg/maps/nat/nat.go
gentoo-root: A different timestamp can't lead to a new error, and the clockSource is the same, so all possible errors have been checked. I'd really like to avoid cluttering the code with dead code. There is no use of adding a few more lines of code that cannot be possibly called. It's also a source of confusion for whoever will read it.

Though, I have an idea on how to drop the implicit assumption without introducing dead code.

pkg/maps/timestamp/timestamp.go
gentoo-root: It feels like checking `if (this == NULL)` in the beginning of C++ methods.

> We're making the assumption that clockSource is legit

I am only assuming it's not nil. In my vision, it's a sane assumption. I can adjust, though, it's not a big deal.

pkg/maps/timestamp/timestamp_test.go
gentoo-root: Not only the happy path; I also test the error flow (Hertz == 0), see lines 29-31. Let me also add a test for an invalid Mode.

derailed: Perhaps we could leverage table testing here would clarifying the various use cases and DRY up this test?

derailed (approved):
@gentoo-root Thank you for the updates Maxim!

pkg/maps/timestamp/timestamp_test.go
gentoo-root: Updated the test to use a table.

PR #26934, opened 2023-07-19
docs: Specify Helm chart version in "cilium install" commands

michi-covalent:
- For the main branch latest docs, clone the Cilium GitHub repo and use
 "--chart-directory ./install/kubernetes/cilium" flag.
- For stable branches, set "--version" flag to the version in the
  top-level VERSION file.

Fixes: #26931

Signed-off-by: Michi Mutsuzaki <michi@isovalent.com>

PR #27066, opened 2023-07-25
docs: Use kubeProxyReplacement=true for Gateway API docs

### LINKED ISSUE #27060: `partial` and `strict` are deprecated for helm installation

### What happened?

When trying to install v1.14 rc with helm, the cilium-operator and agents never initialize correctly
`partial` and `strict` values for `kubeProxyReplacement` are deprecated, and the GatewayAPI show that as a [prereq](https://docs.cilium.io/en/latest/network/servicemesh/gateway-api/gateway-api/) 

### Cilium Version

v1.14 rc and greater

### Kernel Version

not relevant

### Kubernetes Version

v1.27.3

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

michi-covalent:
- Use `kubeProxyReplacement=true` as `partial` and `strict` settings have been deprecated in v1.14.
- Specify the same Helm values for Helm and Cilium-CLI instructions.
- Move the Cilium CLI installation instructions out of the tabs. It's required for both Helm and Cilium-CLI instructions.

Fixes: #27060

deploy preview: https://deploy-preview-27066--docs-cilium-io.netlify.app/network/servicemesh/gateway-api/gateway-api.html

### DIFF
--- a/Documentation/network/servicemesh/gateway-api/installation.rst
+++ b/Documentation/network/servicemesh/gateway-api/installation.rst
@@ -1,8 +1,8 @@
 Prerequisites
 #############
 
-* Cilium must be configured with ``kubeProxyReplacement`` as partial
-  or strict. Please refer to :ref:`kube-proxy replacement <kubeproxy-free>`
+* Cilium must be configured with ``kubeProxyReplacement=true``.
+  Please refer to :ref:`kube-proxy replacement <kubeproxy-free>`
   for more details.
 * Cilium must be configured with the L7 proxy enabled using the ``--enable-l7-proxy`` flag (enabled by default).
 * The below CRDs from Gateway API v0.7.0 ``must`` be pre-installed.
@@ -29,6 +29,8 @@ Prerequisites
 Installation
 ############
 
+.. include:: ../../../installation/cli-download.rst
+
 .. tabs::
 
     .. group-tab:: Helm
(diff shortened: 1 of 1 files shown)


sayboras (approved):
:rocket:

Documentation/conf.py
```diff
@@ -87,7 +87,7 @@
 if not branch or branch == 'latest':
     branch = 'HEAD'
     archive_name = 'main'
-    chart_release = './cilium'
+    chart_release = './install/kubernetes/cilium'
```
qmonnet: Can you please split this out, please? https://github.com/cilium/cilium/pull/26934 is being backported to all supported branches, so this change should likely be as well?

I'd also prefer having separate commits for the different issues that your PR addresses. Not a blocker because the commit remains small, but that would makes it easier to track the changes.

qmonnet (changes requested):

Documentation/conf.py
qmonnet: (The updates themselves look good, by the way.)

michi-covalent: updated 🥰

youngnick (approved):

qmonnet (approved):
Thanks!

PR #27079, opened 2023-07-26
bpf: nat: let caller determine whether SNATed connection needs CT

julianwiedmann:
Right now this is decided at the lowest level of the SNAT path. But actually the callers know much better. In particular this avoids one case where we bake EgressGW knowledge deep into the SNAT code.


julianwiedmann:
Ah sweet, this also nicely reduces the instruction count for the `tail_nodeport_nat_egress_ipv*()` paths. That traffic never needs CT, and now the compiler understands that it can eliminate all the CT-related code when it pulls in `__snat_v*_nat()`.

julianwiedmann:
@gentoo-root FYI

dylandreimerink (approved):
LGTM

PR #27083, opened 2023-07-26
CODEOWNERS: assign bpf/lib/auth.h to sig-servicemesh

mhofstetter:
sig-servicemesh is responsible mutual authentication and therefore should review `bpf/lib/auth.h`.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -246,6 +246,7 @@ Makefile* @cilium/build
 /bpf/Makefile* @cilium/loader
 /bpf/init.sh @cilium/loader
 /bpf/custom/Makefile* @cilium/build @cilium/loader
+/bpf/lib/auth.h @cilium/sig-datapath @cilium/sig-servicemesh
 /bpf/lib/encrypt.h @cilium/ipsec
 /bugtool/ @cilium/tophat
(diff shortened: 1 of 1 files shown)


bimmlerd (approved):

CODEOWNERS
```diff
@@ -246,6 +246,7 @@ Makefile* @cilium/build
 /bpf/Makefile* @cilium/loader
 /bpf/init.sh @cilium/loader
 /bpf/custom/Makefile* @cilium/build @cilium/loader
+/bpf/lib/auth.h @cilium/sig-servicemesh
```
joestringer: Should this have @cilium/sig-datapath _and_ @cilium/sig-servicemesh ?

I don't mind either way, but I'd like to see an ack from datapath side before merging.

pchaigno: I agree it would be good to keep datapath in the loop here given the servicemesh team isn't a datapath team at all.

mhofstetter: definitely agree. it was meant to stay up to date if something changes. -> included sig-datapath

pchaigno (approved):

joestringer (approved):

PR #27089, opened 2023-07-26
docs: update roadmap after 1.14 release

lizrice:
Updates to the project roadmap page

### DIFF
--- a/Documentation/community/roadmap.rst
+++ b/Documentation/community/roadmap.rst
@@ -50,9 +50,9 @@ Major Feature Status
 ++-------------------------------------------------+----------------------------------------------------------+
 || :ref:`gs_ingress`                               | Stable                                                   |
 ++-------------------------------------------------+----------------------------------------------------------+
-|| :ref:`gs_gateway_api`                           | Stable                                                   |
+|| :ref:`gs_gateway_api`                           | Beta                                                     |
 ++-------------------------------------------------+----------------------------------------------------------+
-|| :ref:`gs_l7_traffic_management`                 | Stable                                                   |
+|| :ref:`gs_l7_traffic_management`                 | Beta                                                     |
 ++-------------------------------------------------+----------------------------------------------------------+
 || :ref:`gs_mutual_authentication`                 | Beta                                                     |
 ++-------------------------------------------------+----------------------------------------------------------+
@@ -67,7 +67,7 @@ features may still be in beta or in development).
 Release Cadence
 ~~~~~~~~~~~~~~~
 
-We aim to make 2-3 `point releases`_ per year of Cilium and its core components
+We aim to make 2 to 3 `point releases`_ per year of Cilium and its core components
 (Hubble, Cilium CLI, Tetragon, etc). We also make patch releases available as
 necessary for security or urgent fixes. 
 
@@ -97,9 +97,11 @@ Cilium has applied for `CNCF Graduation`_, please add your support on the PR!
 Cilium Service Mesh 
 ~~~~~~~~~~~~~~~~~~~
 
-As Cilium Service Mesh gets wider adoption, the next steps we'd like to take in
-this area (in no particular order) are: 
+As Cilium Service Mesh gets wider adoption, the next steps in
+this area (in no particular order) include: 
 
+* Graduating Gateway API and L7 Traffic management to stable
+* Graduating next-gen mutual authentication and SPIFFE integration to stable
 * Graduating Prometheus metrics and OpenTelemetry collector to stable
 * Extended sample Grafana dashboards for L7 visibility
 * SMI integration 
@@ -132,6 +134,7 @@ following features to graduate to stable:
 * :ref:`Local Redirect Policy<local-redirect-policy>`
 * :ref:`CiliumEndpointSlice<gsg_ces>`
 * :ref:`Multi-Pool IPAM<ipam_crd_multi_pool>`
+* :ref:`Node-to-node WireGuard encryption<node-node-wg>`
 
 .. _rm-hubble-observability:
 
@@ -162,8 +165,8 @@ Tetragon Security
 CLI for things like process execution, file access, network observability, and
 privileged execution.
 
-Although Tetragon is still in Beta stage, it is already in
-production use with a set of adopters.
+Although Tetragon is still in Beta stage, a set of adopters already use it in
+production.
 
 Codebase modularization
 ~~~~~~~~~~~~~~~~~~~~~~~
--- a/Documentation/community/roadmap.rst
+++ b/Documentation/community/roadmap.rst
@@ -14,27 +14,52 @@ broader community. You'll also find here some pointers on how you can
 
 Major Feature Status
 --------------------
-+-----------------------------+------------------------------------------------------------+
-| eBPF Networking             | Stable (:ref:`Roadmap Details<rm-advanced-networking>`)    |
-++----------------------------+------------------------------------------------------------+
-|| Kubernetes CNI             | Stable                                                     |
-++----------------------------+------------------------------------------------------------+
-|| Load Balancing             | Stable                                                     |
-++----------------------------+------------------------------------------------------------+
-|| Network Policy             | Stable                                                     |
-++----------------------------+------------------------------------------------------------+
-|| Kube-proxy Replacement     | Stable                                                     |
-++----------------------------+------------------------------------------------------------+
-|| Egress Gateway             | Stable                                                     |
-++----------------------------+------------------------------------------------------------+
-| Multi-Cluster (ClusterMesh) | Stable (:ref:`Roadmap Details<rm-clustermesh>`)            |
-+-----------------------------+------------------------------------------------------------+
-| Hubble Observability        | Stable (:ref:`Roadmap Details<rm-hubble-observability>`)   |
-+-----------------------------+------------------------------------------------------------+
-| Service Mesh                | Stable (:ref:`Roadmap Details<rm-cilium-service-mesh>`)    |
-+-----------------------------+------------------------------------------------------------+
-| Tetragon Security           | Beta                                                       |
-+-----------------------------+------------------------------------------------------------+
+
++--------------------------------------------------+----------------------------------------------------------+
+| eBPF Networking                                  | Stable (:ref:`Roadmap Details<rm-advanced-networking>`)  |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`Kubernetes CNI<k8s_network_root>`         | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| `Load Balancing`_                               | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`k8s_policy`                               | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`Kube-proxy Replacement<kubeproxy-free>`   | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`egress-gateway`                           | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`gsg_encryption` (IPSec and WireGuard)     | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`bandwidth-manager`                        | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+| Cilium Mesh                                      | Stable (:ref:`Roadmap Details<rm-clustermesh>`)          |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`Multi-Cluster (ClusterMesh)<clustermesh>` | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`External Workloads<external_workloads>`   | Beta                                                     |
+++-------------------------------------------------+----------------------------------------------------------+
+| Hubble Observability                             | Stable (:ref:`Roadmap Details<rm-hubble-observability>`) |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`Hubble CLI<hubble_cli>`                   | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`hubble_ui`                                | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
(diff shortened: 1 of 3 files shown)


Documentation/community/roadmap.rst
```diff
@@ -14,35 +14,60 @@ broader community. You'll also find here some pointers on how you can
...
+++-------------------------------------------------+----------------------------------------------------------+
+| Cilium Service Mesh                              | Stable (:ref:`Roadmap Details<rm-cilium-service-mesh>`)  |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`gs_ingress`                               | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`gs_gateway_api`                           | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`gs_l7_traffic_management`                 | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`gs_mutual_authentication`                 | Beta                                                     |
```
qmonnet, line 57: This one says `Mutual Authentication (Beta) | Beta` so “Beta” appears twice - although it probably doesn't matter.

qmonnet (approved):
Looks great, thanks a lot! It will be good to have this in for the 1.14 release.

I checked that all links in the table work and point to the correct location.

Direct link to preview: https://deploy-preview-27089--docs-cilium-io.netlify.app/community/roadmap.html

Documentation/community/roadmap.rst
```diff
@@ -72,43 +97,28 @@ Cilium has applied for `CNCF Graduation`_, please add your support on the PR!
...
-* Using Kubernetes as service mesh control plane 
- 
-  * Simple to use sidecar-free service mesh configured using Kubernetes Services
-    and Ingress with support for additional annotations
-
-* Graduating EnvoyConfig CRD to stable
 * Extended sample Grafana dashboards for L7 visibility
 * SMI integration 
-* SPIFFE integration
-* Gateway API Integration
```
joestringer, line 91: Probably worth retaining Gateway API in some form since the spec isn't finalized yet but we're working to implement newer versions of the spec?

Documentation/community/roadmap.rst
```diff
@@ -72,43 +97,28 @@ Cilium has applied for `CNCF Graduation`_, please add your support on the PR!
...
+As Cilium Service Mesh gets wider adoption, the next steps we'd like to take in
+this area (in no particular order) are: 
 
 * Graduating Prometheus metrics and OpenTelemetry collector to stable
-* Using Kubernetes as service mesh control plane 
- 
-  * Simple to use sidecar-free service mesh configured using Kubernetes Services
-    and Ingress with support for additional annotations
-
-* Graduating EnvoyConfig CRD to stable
```
joestringer, line 87: Was this dropped because we abandoned making EnvoyConfig stable? Because I'm not sure we went through any process to decide that it's stable.

Documentation/community/roadmap.rst
```diff
@@ -14,35 +14,60 @@ broader community. You'll also find here some pointers on how you can
...
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`k8s_policy`                               | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`Kube-proxy Replacement<kubeproxy-free>`   | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`egress-gateway`                           | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`gsg_encryption` (IPSec and WireGuard)     | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`bandwidth-manager`                        | Stable                                                   |
```
joestringer, line 33: Do we have input from users on stability of bandwidth manager? Are users enabling it? cc @cilium/sig-datapath

Documentation/community/roadmap.rst
```diff
@@ -14,35 +14,60 @@ broader community. You'll also find here some pointers on how you can
...
+++-------------------------------------------------+----------------------------------------------------------+
+|| `Load Balancing`_                               | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`k8s_policy`                               | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`Kube-proxy Replacement<kubeproxy-free>`   | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`egress-gateway`                           | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`gsg_encryption` (IPSec and WireGuard)     | Stable                                                   |
```
joestringer, line 31: There's some caveats here I think, for instance for node-to-node encryption? I'm not sure whether the level of testing really validates a stable tag yet or not cc @cilium/sig-encryption

Documentation/community/roadmap.rst
```diff
@@ -14,35 +14,60 @@ broader community. You'll also find here some pointers on how you can
...
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`Prometheus metrics<metrics>`              | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+| Cilium Service Mesh                              | Stable (:ref:`Roadmap Details<rm-cilium-service-mesh>`)  |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`gs_ingress`                               | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`gs_gateway_api`                           | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`gs_l7_traffic_management`                 | Stable                                                   |
```
joestringer: Is this really stable? 2/3 of the [examples](https://docs.cilium.io/en/v1.14/network/servicemesh/l7-traffic-management/#examples) are using CiliumEnvoyConfig which I don't think we make any stability guarantees about, and the other example which doesn't use CEC is explicitly marked beta.

joestringer (changes requested):
Should be marked as requesting changes to ensure discussion on above feedback occurs.

Documentation/community/roadmap.rst
```diff
@@ -14,35 +14,60 @@ broader community. You'll also find here some pointers on how you can
...
++--------------------------------------------------+----------------------------------------------------------+
 
 "Stable" means that the feature is in use in production (though advanced
 features may still be in beta or in development).
 
 Release Cadence
 ~~~~~~~~~~~~~~~
 
-We aim to make 2-3 point releases per year of Cilium and its core components
+We aim to make 2-3 `point releases`_ per year of Cilium and its core components
```
zacharysarah: nit
```suggestion
We aim to make 2 to 3 `point releases`_ per year of Cilium and its core components
```

Documentation/community/roadmap.rst
```diff
@@ -145,13 +153,18 @@ currently working on `CI improvements`_ to make these more reliable and easier t
...
 Tetragon Security
 ~~~~~~~~~~~~~~~~~
 
-Tetragon provides security observability and runtime enforcement through the JSON events and the Tetragon
+`Tetragon`_ is a sub-project within the Cilium family. It provides security observability and runtime enforcement through the JSON events and the Tetragon
 CLI for things like process execution, file access, network observability, and
 privileged execution.
 
+Although Tetragon is still in Beta stage, it is already in
+production use with a set of adopters.
```
zacharysarah: Use active voice
```suggestion
Although Tetragon is still in Beta stage, a set of adopters use it in production.
```

Documentation/community/roadmap.rst
```diff
@@ -72,43 +97,28 @@ Cilium has applied for `CNCF Graduation`_, please add your support on the PR!
 Cilium Service Mesh 
 ~~~~~~~~~~~~~~~~~~~
 
-Our eBPF-accelerated Service Mesh is the main focus for
-major enhancement, and it's a natural evolution of Cilium's networking
-capabilities. We released a beta at the end of 2021 and had very valuable
-feedback from our user community. The next steps we'd like to take for Cilium
-Service Mesh (in no particular order) are: 
+As Cilium Service Mesh gets wider adoption, the next steps we'd like to take in
+this area (in no particular order) are:
```
zacharysarah: Use clear, plain language
```suggestion
As Cilium Service Mesh gets wider adoption, next steps (in no particular order) include: 
```

zacharysarah (changes requested):
@lizrice Some nits, otherwise LGTM

Documentation/community/roadmap.rst
youngnick: Yes, I think that we're better to keep EnvoyConfig as _mostly_ an implementation detail that brave folks may make use of if they're game. And then find things that people want to do with their own EnvoyConfigs and do them in Gateway API instead.

Documentation/community/roadmap.rst
youngnick: Agreed, I'd like to see Gateway API integration stay in next steps until it's GA.

Documentation/community/roadmap.rst
youngnick: Yeah, I would have thought this should be a bit lower for now too.

lizrice: good catch, I meant to mark Gateway API as Beta so that was my mistake. Sounds like L7 traffic mgmt should be too, so I’ll change that as well

Documentation/community/roadmap.rst
```diff
@@ -14,35 +14,60 @@ broader community. You'll also find here some pointers on how you can
...
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`hubble_ui`                                | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`Prometheus metrics<metrics>`              | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+| Cilium Service Mesh                              | Stable (:ref:`Roadmap Details<rm-cilium-service-mesh>`)  |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`gs_ingress`                               | Stable                                                   |
+++-------------------------------------------------+----------------------------------------------------------+
+|| :ref:`gs_gateway_api`                           | Stable                                                   |
```
lizrice:
```suggestion
|| :ref:`gs_gateway_api`                           | Beta                                                     |
```

Documentation/community/roadmap.rst
lizrice:
```suggestion
|| :ref:`gs_l7_traffic_management`                 | Beta                                                     |
```

Documentation/community/roadmap.rst
lizrice: I took these out because I think they're largely covered by the table above (if I correct GW API and L7 traffic management to be marked as Beta). But you're right, I should leave something here about graduating both to stable

Documentation/community/roadmap.rst
gandro: So yeah,

- node to node encryption for WireGuard is beta
- node to node encryption for IPSec is not supported

where as regular pod to pod encryption should be stable for both. Node to node encryption is opt-in.

Documentation/community/roadmap.rst
pchaigno: It's hard to say. We've had it enabled by default for a bit, disabled it because we were hitting issues, and it has been disabled by default ever since. We don't have many (any?) bug reports, but I also think we don't have many reports of people enabling it.

lizrice: FWIW it's not currently marked as beta in the [docs](https://docs.cilium.io/en/v1.14/network/kubernetes/bandwidth-manager/) (but that could be an oversight?)

Documentation/community/roadmap.rst
lizrice: Added bullets for graduating Gateway API, L7 Traffic Management, next-gen mutual authn and SPIFFE integration

Documentation/community/roadmap.rst
lizrice: yep, added bullet for this

Documentation/community/roadmap.rst
lizrice: I think fixing this would mean re-sizing the table (because RST is so awesome!) and I don't think it's worth the effort

Documentation/community/roadmap.rst
lizrice: I added a bullet in the roadmap details below to cover graduating node-to-node WireGuard to stable. I considered adding something in the table referring to the docs for details, but it seemed like that might well also apply to other features, and it would end up being too much detail for this high-level view. There is a caveat already under the table saying:

> “Stable” means that the feature is in use in production (though advanced features may still be in beta or in development).

joestringer (approved):

zacharysarah (approved):
LGTM

gandro (approved):

Documentation/community/roadmap.rst
```diff
@@ -14,35 +14,60 @@ broader community. You'll also find here some pointers on how you can
...
++--------------------------------------------------+----------------------------------------------------------+
 
 "Stable" means that the feature is in use in production (though advanced
 features may still be in beta or in development).
 
 Release Cadence
 ~~~~~~~~~~~~~~~
 
-We aim to make 2-3 point releases per year of Cilium and its core components
+We aim to make 2 to 3 `point releases`_ per year of Cilium and its core components
```
margamanterola, line 70: We are rather consistently making 2 releases per year, with patch releases every month.

lizrice: That's consistent with the text as written? We could document patch releases here too, but I think that would probably be more detail than needed for this roadmap page

youngnick (approved):

Documentation/community/roadmap.rst
lizrice: @borkmann what's your view, should this be marked as stable or beta? Edit: don't worry, we found where it was previously promoted to stable

dylandreimerink: We have 2 enterprise users that use the feature according to internal data, including IKEA (openly states this on the USERS.md)

qmonnet: Note: the Bandwidth Manager was publicly promoted to “stable” [in Cilium 1.12's announcement blog](https://isovalent.com/blog/post/cilium-release-112/#bm-stable).

lizrice: Sounds clear that we should be treating it as stable

PR #27090, opened 2023-07-26
Add note for changing IPAM settings

darox:
This  PR adds notes to https://docs.cilium.io/en/v1.14/network/concepts/ipam/#address-management and https://docs.cilium.io/en/v1.14/network/concepts/ipam/cluster-pool/ what should be considered when changing IPAM settings.

### DIFF
--- a/Documentation/network/concepts/ipam/cluster-pool.rst
+++ b/Documentation/network/concepts/ipam/cluster-pool.rst
@@ -50,7 +50,7 @@ Expanding the cluster pool
 
 Don't change any existing elements of the ``clusterPoolIPv4PodCIDRList`` list, as
 changes cause unexpected behavior. If the pool is exhausted,
-add a new element to the list. The minimum mask length is ``/30``, with a recommended minimum mask 
+add a new element to the list instead. The minimum mask length is ``/30``, with a recommended minimum mask 
 length of at least ``/29``. The reason to add new elements rather than change existing elements is that
 the allocator reserves 2 IPs per CIDR block for the network and broadcast addresses.
 Changing ``clusterPoolIPv4MaskSize`` is also not possible. 
--- a/Documentation/network/concepts/ipam/cluster-pool.rst
+++ b/Documentation/network/concepts/ipam/cluster-pool.rst
@@ -45,6 +45,16 @@ Configuration
 For a practical tutorial on how to enable this mode in Cilium, see
 :ref:`gsg_ipam_crd_cluster_pool`.
 
+Expanding the cluster pool
+==========================
+
+Don't change any existing elements of the ``clusterPoolIPv4PodCIDRList`` list, as
+changes cause unexpected behavior. If the pool is exhausted,
+add a new element to the list. The minimum mask length is ``/30``, with a recommended minimum mask 
+length of at least ``/29``. The reason to add new elements rather than change existing elements is that
+the allocator reserves 2 IPs per CIDR block for the network and broadcast addresses.
+Changing ``clusterPoolIPv4MaskSize`` is also not possible. 
+
 ***************
 Troubleshooting
 ***************
--- a/Documentation/network/concepts/ipam/index.rst
+++ b/Documentation/network/concepts/ipam/index.rst
@@ -29,7 +29,7 @@ Dynamic CIDR/IP allocation   ❌                     ❌                       �
 Don't change the IPAM mode of an existing cluster. Changing the IPAM mode in
 a live environment may cause persistent disruption of connectivity for existing workloads.
 The safest path to change IPAM mode is to install a fresh Kubernetes cluster with the new IPAM configuration.
-
+If you are interested in extending Cilium to support migration between IPAM modes, see :gh-issue:`27164`.
 
 .. toctree::
    :maxdepth: 1
--- a/Documentation/network/concepts/ipam/index.rst
+++ b/Documentation/network/concepts/ipam/index.rst
@@ -26,6 +26,10 @@ Multiple CIDRs per node      ❌                     ❌                       �
 Dynamic CIDR/IP allocation   ❌                     ❌                       ✅          ✅          ✅              ✅                ❌
 ============================ ====================== ======================== =========== =========== =============== ================= ===============
 
+Don't change the IPAM mode of an existing cluster. Changing the IPAM mode in
+a live environment may cause persistent disruption of connectivity for existing workloads.
+The safest path to change IPAM mode is to install a fresh Kubernetes cluster with the new IPAM configuration.
+
 
 .. toctree::
    :maxdepth: 1


Documentation/network/concepts/ipam/cluster-pool.rst
```diff
@@ -45,6 +45,12 @@ Configuration
 For a practical tutorial on how to enable this mode in Cilium, see
 :ref:`gsg_ipam_crd_cluster_pool`.
 
+.. warning::
+    An existing element of the ``clusterPoolIPv4PodCIDRList`` list shouldn't be removed or changed. 
+    Doing so will cause unexpected behaviour. If you face a situation where the pool is exhausted,
+    add a new element to the list. The minimum mask length is ``/30``, but a mask of at least ``/29`` is recommended.
```
gandro: I think we should document the reason why that is:

The first and last IP address in each pod CIDR block is reserved for the network and broadcast address. This restriction might be lifted in a future release.

gandro (commented):
Looks good to me overall! One small comment

gandro:
Check patch is failing because the commit subject is overly long:

```
  Error: ERROR:CUSTOM: Please avoid long commit subjects (max: 75, found: 80)
```

Documentation/network/concepts/ipam/cluster-pool.rst
darox: Sure, I can easily add that. Do you think a warning is still the right way of documenting it in the docs, or shall I change it to "not a warning".

gandro: I think `.. note` will do the trick too

darox: done

darox:
> Check patch is failing because the commit subject is overly long:
> 
> ```
>   Error: ERROR:CUSTOM: Please avoid long commit subjects (max: 75, found: 80)
> ```

fixed

gandro (approved):

Documentation/network/concepts/ipam/cluster-pool.rst
```diff
@@ -45,6 +45,13 @@ Configuration
 For a practical tutorial on how to enable this mode in Cilium, see
 :ref:`gsg_ipam_crd_cluster_pool`.
 
+.. note::
```
joestringer: Does this need to use a `.. note::` or can we just create a dedicated section title or just put the text directly in without a callout box?

It is very common for contributors who are new to the docs to think that they're just adding one more note to the docs so it should be separate like this, but if we don't guard against this then eventually every single page is just filled with `..note` boxes that draw user attention to them.

This specific change seems to be related to "Expanding the cluster pool", so maybe this text should just be under a new title section with that title?
```suggestion
Expanding the cluster pool
==========================
```

Documentation/network/concepts/ipam/cluster-pool.rst
```diff
@@ -69,3 +76,5 @@ You can solve it in two ways:
 
   - Explicitly set ``clusterPoolIPv4PodCIDRList`` to a non-conflicting CIDR
   - Use a different CIDR for your nodes
+
+
```
joestringer: Extraneous whitespace, please remove.
```suggestion
```

Documentation/network/concepts/ipam/index.rst
```diff
@@ -39,3 +39,10 @@ Dynamic CIDR/IP allocation   ❌                     ❌                       
    gke
    crd
    deep_dive
+
+.. note::
+
```
joestringer: How does this render?

Again I think this is a case where `.. note::` is unsuitable. Either create a section title for modifying IPAM configuration or just embed the text directly without a note.

Documentation/network/concepts/ipam/cluster-pool.rst
```diff
@@ -45,6 +45,13 @@ Configuration
 For a practical tutorial on how to enable this mode in Cilium, see
 :ref:`gsg_ipam_crd_cluster_pool`.
 
+.. note::
+    An existing element of the ``clusterPoolIPv4PodCIDRList`` list shouldn't be removed or changed. 
+    Doing so will cause unexpected behaviour. If you face a situation where the pool is exhausted,
+    add a new element to the list. The minimum mask length is ``/30``, but a mask of at least ``/29`` 
+    is recommended. The reason for this is that the allocator will reserve 2 IPs per CIDR block for 
+    the network and broadcast addresses. Changing ``clusterPoolIPv4MaskSize`` is also not supported.
```
joestringer: nit: "not supported" gives the impression that for the rest of the project, there is some sort of formal support agreement. However there is not, all of Cilium is provided `on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND` as per the license.

Perhaps @zacharysarah has some sort of suggestion as to how to appropriately phrase that this sort of change in configuration is not tested / not validated by the project and hence the risk is entirely on the user if they decide to make such changes (but also we would happily accept patches to improve this!).

joestringer: Same "not supported" wording shows up in `index.rst` as well, whatever we decide this feedback should apply to that case as well.

joestringer (changes requested):
Thanks for the submission, this helps to clarify how IPAM is / is not safe to change at runtime. I have a few minor wording+formatting suggestions below.

joestringer:
Adding backport labels back to 1.13 since this wording should apply to all previous releases. Skipping 1.12 just because it's likely that the filepaths have changed between 1.12->1.13 which will likely cause backport conflicts.

Documentation/network/concepts/ipam/cluster-pool.rst
zacharysarah: I think it's better to omit any mention of support for the reasons @joestringer mentions. If we need to give advice about use cases, then let's speak to the specific use case. For example: "Changing ``value`` isn't recommended for production environments." We could also just say: "Changing ``value`` isn't recommended."

zacharysarah: Use present tense, use US English, use clear, plain language
```suggestion
    Don't change any existing elements of the ``clusterPoolIPv4PodCIDRList`` list, as
    changes cause unexpected behavior. If the pool is exhausted,
    add a new element to the list. The minimum mask length is ``/30``, with a recommended minimum mask 
    length of at least ``/29``. The reason to add new elements rather than change existing elements is that
    the allocator reserves 2 IPs per CIDR block for 
    the network and broadcast addresses.
    
    Changing ``clusterPoolIPv4MaskSize`` is also not recommended.
```

Documentation/network/concepts/ipam/index.rst
```diff
@@ -39,3 +39,10 @@ Dynamic CIDR/IP allocation   ❌                     ❌                       
...
+    Changing the IPAM mode of an existing cluster is not supported. 
+    Cilium must be reinstalled to change the IPAM mode. Before reinstalling
+    Cilium, make sure that all CiliumNode CRD objects are deleted.
```
zacharysarah: s/supported/recommended per ☝🏻, use active voice
```suggestion
    Changing the IPAM mode of an existing cluster isn't recommended. 
    You must reinstall Cilium to change the IPAM mode. 
    
    Before reinstalling Cilium, delete all CiliumNode CRD objects.
```

zacharysarah (changes requested):
@darox 👋🏻 Good start. Here are some things to fix in addition to a strong +1 to everything @joestringer mentioned.

Documentation/network/concepts/ipam/cluster-pool.rst
gandro: Maybe to clarify, "not supported" in this context here means "it will not work"/"it will break your cluster" (recoverable, I think). So it's not only not recommended, but something that our code simply cannot deal with and so users _must_ not do this.

 Ideally of course, we would enforce this, and eventually we will probably have the ability to do prevent the user from doing this (I think Dario has some ideas), but for now we want to at least document it.

Documentation/network/concepts/ipam/cluster-pool.rst
darox: I thought about that, I even started with putting a warning before. I'm with you that a new section is a better idea.

Documentation/network/concepts/ipam/cluster-pool.rst
darox: I changed it to "Note that changing the IPAM mode of an existing cluster is not possible."

darox: Thank you very much for the correction.

joestringer: May be worth considering something like: "Changing the IPAM mode of an existing cluster will break Pod IP assignment for existing workloads"? This would be a bit more clear about the consequences.

darox: My understanding of the behaviour is that nothing will happen if you change the IPAM mode. Is my understanding wrong?

joestringer: I don't have specific knowledge in this regard, but I was teeing off @gandro's comment above that it will "break your cluster" in a possibly recoverable way. My guess is that if you reconfigure then upon Cilium restart it will try to re-allocate the previously-used IPs from the new pool and that may not be possible, so then I don't know if Cilium will fail restore of existing workloads or let them persist with IPs that may or may not be usable.

gandro: The "is not possible" statement was with regards to the "changing `clusterPoolIPv4MaskSize`" sentence in the highlighted patch line. Doing this will cause the operator to reject any existing pod CIDRs on a node and allocate new pod CIDRs with the new mask size. I also think this means that some of the IP space of a node with the old mask size can be reused by a different node, which then leads to duplicate IPs on pods.

Changing the IPAM mode in a cluster with running pods (which also in the category of "is not possible") will cause agents and operator to ignore the old IPAM IPs and start allocating new ones. Existing pods still have IPs from the old IPAM mode and likely be in a defunct state (e.g. routes might not point to them anymore, endpoint restoration will not work etc, the list here is rather long).

joestringer: :+1: thanks for the details. To summarize, that sounds a bit more like:

> Changing the IPAM mode of an existing cluster is strongly discouraged as it may cause persistent connectivity disruption for existing workloads.

Documentation/network/concepts/ipam/index.rst
```diff
@@ -26,6 +26,10 @@ Multiple CIDRs per node      ❌                     ❌                       
 Dynamic CIDR/IP allocation   ❌                     ❌                       ✅          ✅          ✅              ✅                ❌
 ============================ ====================== ======================== =========== =========== =============== ================= ===============
 
+Changing the IPAM mode of an existing cluster is impossible.
```
joestringer: Maybe..?

```suggestion
Changing the IPAM mode of an existing cluster is strongly discouraged as it may cause persistent connectivity disruption for existing workloads.
```

joestringer: This plus the "you must reinstall Cilium ..." should make it clear both why and also how to make the change if necessary.

Documentation/network/concepts/ipam/index.rst
```diff
@@ -26,6 +26,10 @@ Multiple CIDRs per node      ❌                     ❌                       
...
+You must reinstall Cilium to change the IPAM mode.
+Before reinstalling Cilium, delete all CiliumNode CRD objects.
```
joestringer: I struggle a little bit with the step to delete all CiliumNode CRD objects. Exactly when should they delete the CilumNode CRD objects? Before uninstalling? After uninstalling?

I think that the intent here is to state what is necessary to switch modes, but at the same time it also feels like it's half way between documenting limitations and trying to provide a migration guide. 

Strictly speaking this is probably correct, but also if we want to provide a path for users to change the IPAM mode then we could remove the need to delete CiliumNodeCRD objects in the implementation, and probably we would come up with a more comprehensive migration guide rather than dropping three terse sentences on how it could be done.

joestringer: Here's another alternative proposal:
```suggestion
Do not change the IPAM mode of an existing cluster. Changing the IPAM mode in
a live environment may cause persistent connectivity disruption for existing workloads.
The safest path to change IPAM mode is to install a fresh Kubernetes cluster with the new IPAM configuration.
```

I feel this would make it clear that the Cilium community has not attempted to solve this problem yet, and if we want to solve this problem then we can open issues/CFPs/PRs and work through the process to provide a path.

Documentation/network/concepts/ipam/index.rst
joestringer: (I prefer my proposal [below](https://github.com/cilium/cilium/pull/27090#discussion_r1276688974) instead, but I'm also open to this proposal. In particular I think that the previous text that states "it is impossible" followed by "here's how to do it" is confusing.)

Documentation/network/concepts/ipam/index.rst
darox: Yeah, I agree.

darox: Is a fresh cluster needed? I guess removing Cilium and run https://docs.cilium.io/en/stable/cmdref/cilium_cleanup/ will do the trick?

joestringer: Not quite, mainly because that cilium cleanup only cleans local node state, not state in k8s, so the step that was referenced earlier about deleting CiliumNode resources won't happen AFAIK.

darox: Yes, that's also my understanding. I'm refering to "install a fresh cluster" So in theory deleting CiliumNode CRD + run the cleanup.

joestringer: Ah yes that should technically work as well. We could alternatively describe it as completely removing Cilium from the cluster including all custom k8s resources and then freshly installing Cilium.

darox: Yes, I think that's the best approach. Then we can link from "changing IPAM mode" to this section. I could work on that next week. Maybe you can give me some hints on where I should add that.

joestringer: Ah, what I had in mind was just to tersely describe here that Cilium must be completely removed and reinstalled. But yeah it seems reasonable to document that too. I'm not seeing an obvious place in the hierarchy for this right now, maybe @zacharysarah might have a thought on this?

Documentation/network/concepts/ipam/cluster-pool.rst
```diff
@@ -45,6 +45,16 @@ Configuration
 For a practical tutorial on how to enable this mode in Cilium, see
 :ref:`gsg_ipam_crd_cluster_pool`.
 
+Expanding the cluster pool
+==========================
+
+Don't change any existing elements of the ``clusterPoolIPv4PodCIDRList`` list, as
+changes cause unexpected behavior. If the pool is exhausted,
+add a new element to the list. The minimum mask length is ``/30``, with a recommended minimum mask
```
zacharysarah: One quick wording nit
```suggestion
add a new element to the list instead. The minimum mask length is ``/30``, with a recommended minimum mask 
```

Documentation/network/concepts/ipam/index.rst
zacharysarah: Would it make sense to add that heading instead under troubleshooting? It's not explicitly a troubleshoot per se, so that's something to consider. We could also add a heading for `Known issues`.

Minor tweaks to @joestringer's proposed update:
```suggestion
Don't change the IPAM mode of an existing cluster. Changing the IPAM mode in
a live environment may cause persistent disruption of connectivity for existing workloads.
The safest path to change IPAM mode is to install a fresh Kubernetes cluster with the new IPAM configuration.
```

> I feel this would make it clear that the Cilium community has not attempted to solve this problem yet, and if we want to solve this problem then we can open issues/CFPs/PRs and work through the process to provide a path.

Would adding explicit language to that effect be useful, or is the statement sufficiently self-disclosing in context? (I think your language here would work great, with some tweaks.)

darox: implemented the recommendations

joestringer: > > I feel this would make it clear that the Cilium community has not attempted to solve this problem yet, and if we want to solve this problem then we can open issues/CFPs/PRs and work through the process to provide a path.
>
> Would adding explicit language to that effect be useful, or is the statement sufficiently self-disclosing in context? (I think your language here would work great, with some tweaks.)

I think that it would help to call this out explicitly. Even though it's Open Source and hence there's always an opportunity to change the way things work, I think that it's often difficult for users to understand whether something is impossible by design or just that no-one solved it yet. I'll propose some wording to this effect.

Documentation/network/concepts/ipam/index.rst
```diff
@@ -26,6 +26,10 @@ Multiple CIDRs per node      ❌                     ❌                       
 Dynamic CIDR/IP allocation   ❌                     ❌                       ✅          ✅          ✅              ✅                ❌
 ============================ ====================== ======================== =========== =========== =============== ================= ===============
 
+Don't change the IPAM mode of an existing cluster. Changing the IPAM mode in
+a live environment may cause persistent disruption of connectivity for existing workloads.
+The safest path to change IPAM mode is to install a fresh Kubernetes cluster with the new IPAM configuration.
+
```
joestringer:
```suggestion
If you are interested in extending Cilium to support migration between IPAM modes, see :gh-issue:`27164`.
```

Documentation/network/concepts/ipam/cluster-pool.rst
joestringer: Un-resolving this comment since the text doesn't include the change yet.

joestringer (changes requested):
I've filed https://github.com/cilium/cilium/issues/27164 and added some suggested text below to reference it in case community members are interested in engaging with the problem space of migrating IPAM modes.

joestringer (approved):
I'm fine with this proposal as-is. I think we can always iterate a bit more on the best way to do IPAM migration, but these instructions are correct, indicate the safest approaches to users, and provide a pointer to somewhere where the community can collaborate on improving the way that Cilium handles this use case.

Thanks!

tommyp1ckles (approved):
changes look good, thanks!

Documentation/network/concepts/ipam/cluster-pool.rst
zacharysarah: @darox 👋🏻 Please address this comment and re-request review.

joestringer: @zacharysarah I see this change in the PR now, do you see it?

zacharysarah: @joestringer I do now, thanks for the prompt.

zacharysarah (approved):

Issue #17240, opened 2021-08-25
Support tunneling over IPv6


AkkyOrz:
## Bug report

Hello. First of all, thank you for an amazing product.

I am trying to configure an IPv6 only kubernetes network(, which means that all the interfaces on my node are not given ipv4 addresses except for the loopback interface.)
Under such circumstances, I tried to introduce cilium based on these articles.

- [Installation using Helm — Cilium 1\.10\.3 documentation](https://docs.cilium.io/en/v1.10/gettingstarted/k8s-install-helm/)
- [CRD\-backed by Cilium cluster\-pool IPAM — Cilium 1\.10\.3 documentation](https://docs.cilium.io/en/v1.10/gettingstarted/ipam-cluster-pool/)

However, I got the following error.

```
...
level=fatal msg="postinit failed" error="external IPv4 node address could not be derived, please configure via --ipv4-node" subsys=daemon
```

I had assumed that I would be tunneling IPv6 packets with IPv6 packets.
However, from what I see in this code, it seems to assume that an IPv4 interface is required.

- [cilium/address\.go#L453-L457 at a8e3fa283cf40b330035265bf83d63800b296bcf · cilium/cilium](https://github.com/cilium/cilium/blob/a8e3fa283cf40b330035265bf83d63800b296bcf/pkg/node/address.go#L453-L457)

https://github.com/cilium/cilium/blob/a8e3fa283cf40b330035265bf83d63800b296bcf/pkg/node/address.go#L453-L457

Is this protected from being configured by this code because there is no implementation that encapsulates it in ipv6?
Or, does the ipv6 encapsulation function exist, but the error is caused by a bug in the implementation?

If the latter is the case, I would like to see this conditional branch modified if it would allow tunneling with ipv6.
Please let me know if there is any information you need.
Thanks.

**General Information**

- Cilium version ( `cilium version` => `v1.10.3`)
- Kernel version (`uname -a` => `Linux <hostname> 5.4.0-81-generic #91-Ubuntu SMP Thu Jul 15 19:09:17 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`)
[...]

PR #27093, opened 2023-07-26
Fix ValidatePostInit for IPv6-only + tunnel

meyskens:
This fixes a check in ValidatePostInit in tunnel mode where it would error on not having a v4 host address. This adds a check if a v6 host address was found. 

```release-note
Fix Cilium startup for IPv6 only and routingMode=tunnel
```

### DIFF
--- a/pkg/node/address.go
+++ b/pkg/node/address.go
@@ -540,9 +540,9 @@ const mismatchRouterIPsMsg = "Mismatch of router IPs found during restoration. T
 // ValidatePostInit validates the entire addressing setup and completes it as
 // required
 func ValidatePostInit() error {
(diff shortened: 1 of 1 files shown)


julianwiedmann:
I believe this is intended - we don't have IPv6 tunnel support, the outer headers are always IPv4.

meyskens:
In that case i'd close it, i could not find any documentation to confirm this so it was confusing.

julianwiedmann:
> In that case i'd close it, i could not find any documentation to confirm this so it was confusing.

https://github.com/cilium/cilium/issues/17240 might be your best :grin:.

PR #27103, opened 2023-07-26
Extend API/CLI to support updating policy by label

deverton-godaddy:
This change extends the HTTP API to allow updating the agent policy repository by label. This is existing functionality that this change just exposes over the HTTP API as well.

The `cilium policy import` command has been updated to make this functionality available from the CLI.

```release-note
cli: Update `cilium policy import` to allow policy replacement by label
```

### DIFF
--- a/api/v1/openapi.yaml
+++ b/api/v1/openapi.yaml
@@ -572,6 +572,8 @@ paths:
       - policy
       parameters:
       - "$ref": "#/parameters/policy-rules"
+      - "$ref": "#/parameters/policy-replace"
+      - "$ref": "#/parameters/policy-replace-with-labels"
       responses:
         '200':
           description: Success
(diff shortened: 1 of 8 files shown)


api/v1/openapi.yaml
```diff
@@ -1126,6 +1128,20 @@ parameters:
     in: body
     schema:
       type: string
+  policy-replace:
+    name: replace
+    description: If true indicates that existing rules with identical labels should be replaced
```
nathanjsweet:
```suggestion
    description: If true, indicates that existing rules with identical labels should be replaced.
```

api/v1/openapi.yaml
```diff
@@ -1126,6 +1128,20 @@ parameters:
...
+    required: false
+    in: query
+    type: boolean
+  policy-replace-with-labels:
+    name: replace-with-labels
+    description: If present indicates that existing rules with the given labels should be deleted.
```
nathanjsweet:
```suggestion
    description: If present, indicates that existing rules with the given labels should be deleted.
```

nathanjsweet (changes requested):
Thanks! This is good. Just some nits. Otherwise, LGTM.

nathanjsweet (approved):

PR #27106, opened 2023-07-27
auth: depend on nodeIDHandler directly 

mhofstetter:
After removing the node IDs from the IPCache with #27029, the auth module should directly depend on the NodeIDHandler instead of the IPCache.

This PR refactors the dependency from the IPCache to the NodeIDHandler.

In addition, the unused NodeIDHandler gets deleted from the IPCache Configuration.


pkg/auth/manager_test.go
```diff
@@ -175,6 +176,13 @@ type fakeIPCache struct {
 	nodeIdMappings map[uint16]string
 }
 
+func (r *fakeIPCache) DumpNodeIDs() []*models.NodeID {
```
gandro: nit: Consider renaming this type to `fakeNodeIDHandler`

gandro (approved):
Excellent! Thanks a lot for this!

meyskens (approved):
Apart from the nit already noted this is an LGTM for me!

christarazi (approved):

pkg/auth/manager_test.go
mhofstetter: @meyskens thanks a lot for dealing with that during my PTO!

PR #27110, opened 2023-07-27
Add docs on first and last IP of LB-IPAM pool

darox:
The first and last IP in a LB-IPAM pool are reserved for the network and broadcast address and can thus not be requested.


zacharysarah (approved):
LGTM

Issue #10601, opened 2020-03-17
CI: Links in docs are not tested


errordeveloper:
I have attempted fixing some of the most trivial issues with links in our docs (https://github.com/cilium/cilium/pull/10576), but some links are not pass the link checker right now and I'm not sure how to fix these.
Perhaps we could instruct the checker to ignore these somehow? Or explore using an alternative link checker.

Here are some of the main examples of what's broken:

Anchors work fine in the browser, but not in the link checker:

```
(line   35) broken    https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#cli-quick-configuration - Anchor 'cli-quick-configuration' not found
(line   19) broken    https://github.com/kubernetes-sigs/kubespray#requirements - Anchor 'requirements' not found
```

Commands that contain URL substrings:

```
(line   32) broken    https://github.com/weaveworks/eksctl/releases/download/latest_release/eksctl_$(uname - 404 Client Error: Not Found for url: https://github.com/weaveworks/eksctl/releases/download/latest_release/eksctl_$(uname
(line   39) broken    https://github.com/kubernetes/kops/releases/download/$(curl - 404 Client Error: Not Found for url: https://github.com/kubernetes/kops/releases/download/$(curl
```

This one needs auth:
```
(line   41) broken    https://github.com/cilium/cilium/projects/new - 404 Client Error: Not Found for url: https://github.com/cilium/cilium/projects/new
```

We also have a few localhost links like this:
```
(line   42) broken    http://localhost:9080 - HTTPConnectionPool(host='localhost', port=9080): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f31cddb0d10>: Failed to establish a new connection: [Errno 111] Connection refused'))
(line  238) broken    http://127.0.0.1:15000/logging?level=debug - HTTPConnectionPool(host='127.0.0.1', port=15000): Max retries exceeded with url: /logging?level=debug (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f31ce248bd0>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

qmonnet:
[The linkcheck builder](http://www.sphinx-doc.org/en/master/_modules/sphinx/builders/linkcheck.html) can take a set of options:

``` python
    app.add_config_value('linkcheck_ignore', [], None)
    app.add_config_value('linkcheck_auth', [], None)
[...]

PR #27116, opened 2023-07-27
CI: Link checker script

vipul-21:

      GitHub issue.

Enabling the link checker for the CI.
- There is know issue of warning: Circular inclusion in the include directive example: WARNING: circular inclusion in "include" directive. Not able to find the circular inclusion in the docs files. There is an open issue on the same : https://github.com/sphinx-doc/sphinx/issues/10951
- myst not able to parse the #anchor in the Documentation. example: 'myst' reference target not found. This has been disabled using the `ref.myst` in the suppress_warning for sphinx.
Note: This is dependent on merge of PR: https://github.com/cilium/cilium/issues/10601 to make sure CI passes

```release-note
ci: Enable link checker to ensure that all links in documentation are valid
```

Related: #10601

### DIFF
--- a/Documentation/check-build.sh
+++ b/Documentation/check-build.sh
@@ -87,10 +87,6 @@ build_with_spellchecker() {
         -E -n --color -w "${warnings}" -W --keep-going 2>/dev/null
 }
 
-build_with_linkchecker() {
-    sphinx-build -b linkcheck -d "${build_dir}/doctrees" . "${spelldir}" -q -E
-}
-
 run_linter() {
     local CONF_PY_ROLES CONF_PY_SUBSTITUTIONS ignored_messages
 
@@ -154,15 +150,6 @@ else
   fi
 fi
 
-# TODO: Fix broken links and re-enable this
-# (https://github.com/cilium/cilium/issues/10601)
-# echo "Checking links..."
-# if ! build_with_linkchecker ; then
-#     echo "Link check failed!"
-#     exit 1
-# fi
-
-echo ""
 echo "Building documentation (${target})..."
 sphinx-build -M "${target}" "${script_dir}" "${build_dir}" $@ \
     ${read_all_opt} -n --color -w "${warnings}" 2>/dev/null
--- a/Documentation/check-links.sh
+++ b/Documentation/check-links.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+echo "WARNING: There is a bug in docutils which may result in failure of this script. For ref: https://github.com/cilium/cilium/pull/27116#issuecomment-1752760611"
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+build_dir="${script_dir}/_build"
+warnings="${build_dir}/links_warnings.txt"
+filtered_warnings="${build_dir}/filtered_links_warnings.txt"
+linksdir="${build_dir}/links"
+
+cd "${script_dir}"
+mkdir -p "${build_dir}"
+rm -f -- "${warnings}"
+rm -f -- "${filtered_warnings}"
+
+filter_warnings() {
+    [ -s "${warnings}" ] || return
+    grep -v -E \
+        -e "/_api/v1/.*/README\.md:[0-9]+: WARNING: 'myst' reference target not found:" \
+        -e "/.*:[0-9]+: WARNING: circular inclusion in \"include\" directive" \
+        "${warnings}"
+}
+
+has_build_warnings() {
+    filter_warnings > /dev/null
+}
+
+build_with_linkchecker() {
+    rm -rf "${linksdir}"
+
+    sphinx-build -b linkcheck -d "${build_dir}/doctrees" . "${linksdir}" \
+        --color -E -w "${warnings}" -W --keep-going 2>/dev/null
+}
+
+echo "Checking links..."
+if ! build_with_linkchecker ; then
+    if has_build_warnings; then
+        printf "\nPlease fix the following documentation warnings:\n"
+        filter_warnings
+        printf "\nAdding the warnings to the file.\n"
+        filter_warnings | sed 's/\x1B\[[0-9;]*[a-zA-Z]//g' > "${filtered_warnings}"
+        exit 1
+    fi
+fi
+echo "Done checking links."
--- a/Documentation/conf.py
+++ b/Documentation/conf.py
@@ -270,6 +270,31 @@ http_strict_mode = False
 # Try as hard as possible to find references
 default_role = 'any'
 
+# -- Options for Link Check output ----------------------------------------
+
+linkcheck_ignore = [
+    # Local links
+    'http://127.0.0.1',
+    'http://localhost',
+    'http://192.168.60.13',
+    'https://192.168.60.11',
+
+    # URLs that are not reachable(invalid or test links)
+    'https://bookinfo.cilium.rocks',
+    'dns:query',
+
+    # Needs authentication
+    'https://github.com/cilium/cilium/projects/new',
+    'https://github.com/cilium/cilium/settings/branches',
+    'https://jenkins.cilium.io/job/',
+    'https://jenkins.cilium.io/view/'
+]
+
+linkcheck_anchors = False
+linkcheck_retries = 5
+
+tls_verify = False
+
 def setup(app):
     app.add_css_file('parsed-literal.css')
     app.add_css_file('copybutton.css')
--- a/Documentation/community/governance/commit_access.rst
+++ b/Documentation/community/governance/commit_access.rst
@@ -136,7 +136,7 @@ Code of Conduct violator is a Maintainer, the Maintainers will instead
 designate two Maintainers to work with the `CNCF CoC Committee`_.
 
 .. _Code of Conduct: https://github.com/cilium/cilium/blob/main/CODE_OF_CONDUCT.md
-.. _CNCF CoC Committee: https://www.cncf.io/conduct/procedures/
+.. _CNCF CoC Committee: https://www.cncf.io/conduct/committee/
 
 Granting Commit Access
 ----------------------
--- a/Documentation/configuration/verify-image-signatures.rst
+++ b/Documentation/configuration/verify-image-signatures.rst
@@ -53,5 +53,5 @@ Let's verify a Cilium image's signature using the ``cosign verify`` command:
     from the Github build images release workflow.
     
 
-.. _`Keyless Signatures`: https://docs.sigstore.dev/cosign/keyless/
+.. _`Keyless Signatures`: https://docs.sigstore.dev/cosign/overview/#keyless-signing-of-a-container
 .. _`Cilium workflows`: https://github.com/cilium/cilium/tree/main/.github/workflows
--- a/Documentation/contributing/development/reviewers_committers/review_process.rst
+++ b/Documentation/contributing/development/reviewers_committers/review_process.rst
@@ -134,7 +134,7 @@ to add or remove yourself from any team, please submit a PR against the
 `community repository`_.
 
 Once a contributor opens a PR, GitHub automatically picks which `teams
-<cilium_teams>`_ should review the PR using the ``CODEOWNERS`` file. Each
+<cilium_teams_>`_ should review the PR using the ``CODEOWNERS`` file. Each
 reviewer can see the PRs they need to review by filtering by reviews
 requested. A good filter is provided in this `link <user_review_filter_>`_ so
 make sure to bookmark it.
--- a/Documentation/installation/alibabacloud-eni.rst
+++ b/Documentation/installation/alibabacloud-eni.rst
@@ -1,5 +1,3 @@
-.. _k8s_alibabacloud_eni:
-
 To install Cilium on `ACK (Alibaba Cloud Container Service for Kubernetes) <https://www.alibabacloud.com/help/doc-detail/86745.htm>`_, perform the following steps:
 
 **Disable ACK CNI (ACK Only):**
@@ -106,7 +104,6 @@ As soon as you have the access tokens, the following secret needs to be added,
 with each empty string replaced by the associated value as a base64-encoded string:
 
 .. code-block:: yaml
-    :name: cilium-secret.yaml
 
     apiVersion: v1
     kind: Secret


qmonnet:
The change looks good, now we need to see how it looks like in practice in the CI:

- How long does the workflow take to run (I can already notice it's longer than usual, it's becoming one of the longest among workflows that are triggered automatically)
- How well does it work, in particular, can it validate all links or does it hit some rate limiting (which is likely, especially if several instances of the workflow run in parallel on different PRs)
- How stable it is, meaning, does it fail often to validate a link because of external error - We'll need to run the workflow several times to get an idea.

qmonnet:
Turning to draft given it's blocked on https://github.com/cilium/cilium/pull/26880 for now.

qmonnet:
Documentation workflow took 11 minutes to complete. I hope it gets faster once more links are fixed and it doesn't have to wait for timeouts.

vipul-21:
@qmonnet Need some help regarding fixing the issue:

``` 
warning: Circular inclusion in the include directive example: WARNING: circular inclusion in "include" directive. " I don't see any circular inclusion.
```

I went through the issues in the docutils and found this: https://sourceforge.net/p/docutils/bugs/459/ . It seems to have been resolved in version 0.18 of docutils  and only exixts in version 0.17.
I tried upgrading the version of docutils and see if the relevant packages dependencies could be resolved, but the `pip -r` time out due to maxdepth reached while resolving dependencies.

qmonnet:
I'm not sure we can update to Docutils 0.18 yet, last time I checked we were still stuck with some dependencies on 0.17 somewhere. Possibly for the theme, I don't think we've rebased our fork on top of https://github.com/readthedocs/sphinx_rtd_theme/pull/1304 yet.

If it's just dependent on 0.17, let's filter out this warning with a comment in the script telling to remove it after we've upgraded to a newer docutils version?

vipul-21:
@qmonnet So i updated the PR.
1. Enabled the link checker: Had to make `tls_verify` as false since the SSL certificates verification for a one site was failing. We might need to add cert as a part of other PR.
2. Fixed few broken links as a part of different commit.

The issue which is happening is 429 too many request error from the requests. As we would be running this a lot(for each PR) the chances of throttling increases(current failure in the checks).

qmonnet:
> @qmonnet So i updated the PR.
> 
>     1. Enabled the link checker: Had to make `tls_verify` as false since the SSL certificates verification for a one site was failing. We might need to add cert as a part of other PR.

OK, I think we can live with that. Out of curiosity, what link is it?

>     2. Fixed few broken links as a part of different commit.

Cool, thanks!

> The issue which is happening is 429 too many request error from the requests. As we would be running this a lot(for each PR) the chances of throttling increases(current failure in the checks).

That's annoying :confused:. Have you tried playing with [`linkcheck_rate_limit_timeout`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_rate_limit_timeout)? Although it will be painful if the workflow takes too long.

Thinking again about the checks, I'm not sure it makes sense to run it on every PR. When authors add new links, the links are (presumably) valid, or it's easy enough to check them during the PR review. Detecting broken links unrelated to the PR doesn't really makes sense in the PR context; this is something that we might just want for pushes to the `main` branch (the current workflow does that, _in addition_ to running on PRs). This would still make quite a lot of requests, especially when someone merges several PRs in a row. There are probably ways to rate-limit it further, we could look into it.

Another thing that could be interesting: Do you know if the rate-limiting on GitHub happens to authenticated users as well? I see [we can pass headers to the probes](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_request_headers), so maybe checking the links by passing a GitHub token would be less impacted? I honestly don't know.

Unrelated: Can we raise `linkcheck_retries` to 2 or 3 (doc says the default is 1), to make the checks a bit more resilient in case a request is lost on the network?

vipul-21:
> I think we can live with that. Out of curiosity, what link is it?

https://netdevconf.info

> Maybe we should just run it [on a given schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule).]

I think this can help us to limit the number of runs and address the issue of 11 mins or so for each PR check.

I did try [linkcheck_rate_limit_timeout](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_rate_limit_timeout). with the `linkcheck_retries` as 2 and timeout as 15.0 ( Can experiment more on this)
Not sure about the rate-limiting on github - will take a look.

vipul-21:
@qmonnet  I played around with the `linkcheck_rate_limit_timeout` but didn't help much.
I found the `-Q`(https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-Q)  flag for sphinx that suppress warning. I removed the `-W`( which I added) that made each warning an error.

Wondering if we should take this PR as a step forward towards enabling the linkchecker only for errors and create an issue for handling the warnings ?
 Also regarding the time for the check in the PR, it took 8 mins earlier. But seems like it is taking more now. Let me know if enabling it for each PR is the path we want to go forward with. I can make a small PR to update the check to run on a given schedule

qmonnet:
> @qmonnet I played around with the `linkcheck_rate_limit_timeout` but didn't help much.

Too bad, but thanks a lot for trying!

> I found the `-Q`(https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-Q) flag for sphinx that suppress warning. I removed the `-W`( which I added) that made each warning an error.

Can you please remind me what are the warnings that we observe? Are the rate-limit events displayed as warnings? Is there anything else?

I've got no objection to ignoring warnings (and removing `-W`) for now, but adding `-Q` is maybe too much - I'd rather keep these warnings in the logs, if they don't make the workflow fail.

> Wondering if we should take this PR as a step forward towards enabling the linkchecker only for errors and create an issue for handling the warnings ? Also regarding the time for the check in the PR, it took 8 mins earlier. But seems like it is taking more now. Let me know if enabling it for each PR is the path we want to go forward with. I can make a small PR to update the check to run on a given schedule

This is probably something we need to discuss with the folks who focus on the CI. I can raise this at today's [Community meeting](https://docs.cilium.io/en/latest/community/community/) (feel free to join). My feeling is that we'd be better off running it as a cron task only, rather than on all PRs. I see Tetragon does both (https://github.com/cilium/tetragon/tree/main/.github/workflows, `check-links-*.yaml`), but their docs are smaller with fewer links I suppose.

We'll need to make the linkchecker run in check-build.sh optional anyway, because we also don't want to spend 10 minutes checking all links when running it to validate doc changes locally. I wonder if we could just reuse the existing `sphinx-build -M "${target}" ...` from this script, by passing `linkcheck` as a target, without having to otherwise modify the script? :thinking: Depends on the options we've got to pass I suppose, it would probably not work if we use different options for different targets

Documentation/check-build.sh
```diff
@@ -153,13 +154,20 @@ else
...
+        printf "\nPlease fix the following documentation warnings:\n"
+        filter_warnings
+        status_ok=1
+    fi
+
+    if [ "${status_ok}" -ne 0 ] ; then
+        echo "Link check failed!"
+        exit 1
+    fi
+fi
```
qmonnet: I think we should maybe just move this to a different script (say `Documentation/check-links.sh`), that we could call from `Documentation/Makefile`:

```makefile

linkcheck:
	@$(ECHO_CHECK) links
	$(QUIET)$(DOCKER_RUN) ./check-links.sh $(SPHINX_OPTS)
```

or directly from a new GitHub workflow, like we do for `check-build.sh` in `.github/workflows/documentation.yaml`.

qmonnet (commented):
So we discussed about that at the community meeting, and the current consensus would be to do just like Tetragon does: a cron-task to check external links and to open GH issues when it finds any, but we would _not_ run the workflow on every PR.

So we need to have a way to run the link checker independently of the rest of the script. Let's maybe move it to a different script? (see below)

vipul-21:
@qmonnet  Sorry for the delayed response( was busy with other stuff)

> Can you please remind me what are the warnings that we observe? Are the rate-limit events displayed as warnings? Is there anything else?

The warnings were related to 429 errors.( throttling due to too many request)

Sure will make the appropriate changes and move it to a different script.
Do we want to enable the script in the github workflow in the same PR OR we should keep them separate ?

Documentation/check-build.sh
vipul-21: I have made the script as a separate one.

vipul-21: I did try to run the github workflow but that error due permissions issue.
Do we want to keep that in the same PR ?

Documentation/check-links.sh
```diff
@@ -0,0 +1,50 @@
...
+has_build_warnings() {
+    filter_warnings > /dev/null
+}
+
+build_with_linkchecker() {
+    set +o nounset
+    if [[ -n "$MAKE_GIT_REPO_SAFE" ]]; then
+        git config --global --add safe.directory "${root_dir}"
+    fi
+    set -o nounset
```
qmonnet: The spell-checker needed this part because it runs some Git commands (to retrieve the list of authors and consider them as spelling exceptions), but I don't expect the linkchecker to do that, so we can probably remove this bit.

Documentation/check-links.sh
```diff
@@ -0,0 +1,50 @@
...
+
+    rm -rf "${linksdir}"
+    
+    sphinx-build -b linkcheck -d "${build_dir}/doctrees" . "${linksdir}" \
+     -q -E -w "${warnings}" 2>/dev/null
```
qmonnet: Nit: indent

```suggestion
    sphinx-build -b linkcheck -d "${build_dir}/doctrees" . "${linksdir}" \
        -q -E -w "${warnings}" 2>/dev/null
```

Documentation/conf.py
```diff
@@ -270,6 +270,31 @@
...
+    'https://github.com/cilium/cilium/projects/new',
+    'https://github.com/cilium/cilium/settings/branches',
+    'https://jenkins.cilium.io/job/',
+    'https://jenkins.cilium.io/view/'
+]
+
+linkcheck_anchors = False
+linkcheck_retries = 5
+
+tls_verify=False
```
qmonnet:
```suggestion
tls_verify = False
```

Do you happen to know whether this option is used for anything else than the linkchecker? It's fine to disable for the linkchecker, but I wouldn't want to lower security if it were used for something else :thinking:

qmonnet (changes requested):
Hi, sorry for the delay (PTO). The separate script looks good, thanks! I just have some minor comments below. Ideally, could you squash all changes to `check-build.sh`/`check-links.sh` into a single commit (and keep the fixes for the broken links in a separate one)?

Ideally, yes I would prefer having the new script enable in the CI in this PR, so we can make sure the script works fine in CI.

Documentation/check-build.sh
qmonnet: Preferably in the same PR, so we can validate that the script works fine with the CI.

Documentation/conf.py
vipul-21: https://github.com/sphinx-doc/sphinx/pull/3156/files#diff-aac07aefec754086f5c344e2c233862f6a611c739b5647fd456c1b77d3577c79R10-R12, it is used for `intersphinx` as well

vipul-21:
@qmonnet Thanks for the review. I updated the script with the changes and modified the commits as well. Have couple of question related to workflow:
1. I tried adding the step of link checker in the workflow but it failed with timeout. ? How can I test the script works in the workflow.
2. Regarding the cron job, what should be the timing for the cron job ? Also, do we need to add the code similar to Tetragon in terms of creating the github issue on link failure ? (Same question, how can I check that the workflow is working as intended, if the permission issue exists.)

Documentation/conf.py
qmonnet: They should really document that. Thanks a lot for looking it up! OK, intersphinx is not a concern and I don't see us using it anytime soon, so all good :+1:.

qmonnet:
> @qmonnet Thanks for the review. I updated the script with the changes and modified the commits as well. Have couple of question related to workflow:

Thanks!

> 1. I tried adding the step of link checker in the workflow but it failed with timeout. ? How can I test the script works in the workflow.

You don't have to do anything to test this workflow, given that it runs on the code from the pull request (that's the `pull_request` trigger at the top; it would be different if the workflow used `pull_request_target` instead).

Currently it fails with: `./Documentation/check-links.sh: Permission denied`. This is likely because the script doesn't have the right permissions: try running `chmod +x Documentation/check-links.sh` and amending your commit.

> 2. Regarding the cron job, what should be the timing for the cron job ?

Good question. It's probably OK to check that links are still valid every few days, let's say once per week?

> Also, do we need to add the code similar to Tetragon in terms of creating the github issue on link failure ?

Yes please. We can probably reuse most of the existing code in https://github.com/cilium/tetragon/blob/main/.github/workflows/check-links-cron.yaml.

> (Same question, how can I check that the workflow is working as intended, if the permission issue exists.)

Another good question, I see no secret being used in that workflow, so presumably nothing required? Cc @mtardy, would you know?

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,69 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          persist-credentials: false 
+      - name: Build HTML
+        run: |
+          set -ex
+          # chmod +x ./Documentation/build.sh
+          # ./Documentation/build.sh
```
qmonnet: We probably don't need to build the docs?

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Check Documentation links
+        run: |
+          set -ex
+          cd Documentation && sudo make html && cd ..
```
qmonnet: Same here

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+          chmod +x ./Documentation/check-links.sh
```
qmonnet: No, this should not be part of the workflow. You need to do it locally and then amend your commit to have the file with the updated permissions.

You can check locally the permissions for this file as known to Git:

```console
$ git ls-files -s Documentation/check-links.sh
100644 <SHA>
```

We want execution rights:

```console
$ chmod +x ./Documentation/check-links.sh
$ git add ./Documentation/check-links.sh
$ git commit --amend <REF>
```

Now we can check again then update the PR:

```console
$ git ls-files -s Documentation/check-links.sh
100755 <SHA>
```

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,69 @@
...
+          ./Documentation/check-links.sh
+          
+          ls ./Documentation/_build
+          sudo touch ./Documentation/_build/links_warnings.txt
```
qmonnet: This shouldn't be necessary? I take it you don't get any warnings in the file?

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,69 @@
...
+          ./Documentation/_build/links_warnings.txt > $GITHUB_OUTPUT
```
qmonnet:
```suggestion
          cat ./Documentation/_build/links_warnings.txt > $GITHUB_OUTPUT
```

qmonnet (commented):
Looking at the workflow code in passing, here are some observations in case it helps.

Anything in particular we can help with?

mtardy:
> Another good question, I see no secret being used in that workflow, so presumably nothing required? Cc @mtardy, would you know?

In the Tetragon one, we use the `secrets.GITHUB_TOKEN` to avoid a ban from the github.com domain after polling all those URLs. You can technically do without but since we have a lot of github.com URLs, we would be banned instantly. Maybe it's different when you run from GitHub runners, I didn't try.

.github/workflows/check-links-cron.yaml
vipul-21: Hey sorry forgot to make it a draft. Was testing a bunch of things to make sure workflow works properly. Will ping you once I am done with the changes. 
Should be done by tomorrow

.github/workflows/check-links-cron.yaml
vipul-21: Hey, thanks for this updated it.

vipul-21:
@qmonnet Sorry for not making it a draft earlier. I was actually testing the workflow, that is why there were few unrelated code segments in the PR.

I have added the code but I am still not able to create the warning.txt in the workflow whereas locally it gets created.
Also, the amount of time it takes to run the check links is very less. I wonder if it is actually running?(Given we have no warning.txt getting created)

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,55 @@
+name: Periodically Check Documentation Links
```
qmonnet:
```suggestion
name: Check Documentation Links
```

It's only periodical when we trigger it with the cron job; but the workflow itself is not necessarily periodical, especially if we can trigger it from PRs touching the workflow file. I'd remove this word from the title

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,55 @@
...
+    name: Documentation links checker
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          persist-credentials: false 
+      - name: Check Documentation links
+        run: |
+          ./Documentation/check-links.sh
```
qmonnet: I agree that the time elapsed for the test is not realistic. The script didn't run correctly, we could try a few things to check it:

- Run with `bash -x` to check where it exits
- Remove the `-q` option passed to Sphinx in `check-links.sh`. Actually, we should maybe do this even after we figured the issue. This would show that we effectively test the links, show how many rate-limit occur, or any warning that we skip, as well that all the redirects that happen.

My guess is simply that Sphinx (and the linkchecker) are not installed on the runner. We should run this command from the `docs-builder` container, which has all dependencies set up, just like we do for the documentation builds in CI (`documentation.yaml`):
```suggestion
      - name: Check Documentation links
        uses: docker://quay.io/cilium/docs-builder:4b194c0facc5d1d3b9c0dfef3529ee7c661a6b73@sha256:095cf9621cf46c6c878e25c3a98f02ef3d0aed1154936663e352fd3fe5e7613d
        with:
          entrypoint: ./Documentation/check-build.sh
```

Documentation/check-links.sh
```diff
@@ -0,0 +1,43 @@
...
+
+has_build_warnings() {
+    filter_warnings > /dev/null
+}
+
+build_with_linkchecker() {
+    rm -rf "${linksdir}"
+    
+    sphinx-build -b linkcheck -d "${build_dir}/doctrees" . "${linksdir}" \
+        -q -E -w "${warnings}" 2>/dev/null
```
qmonnet:
```suggestion
        -E -w "${warnings}" 2>/dev/null
```

.github/workflows/check-links-cron.yaml
vipul-21: I think this will require us to make changes to the script as well ?

qmonnet: My bad, obviously I meant `check-links.sh` and not `check-build.sh`. I realised my mistake after having logged off yesterday but forgot to update in the evening :disappointed:.

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,56 @@
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          persist-credentials: false 
+      - name: Check Documentation links
+        uses: docker://quay.io/cilium/docs-builder:4b194c0facc5d1d3b9c0dfef3529ee7c661a6b73@sha256:095cf9621cf46c6c878e25c3a98f02ef3d0aed1154936663e352fd3fe5e7613d
+        with:
+          entrypoint: ./Documentation/check-build.sh
```
qmonnet:
```suggestion
          entrypoint: ./Documentation/check-links.sh
```

My bad, wrong suggestion.

Documentation/check-links.sh
```diff
@@ -0,0 +1,43 @@
...
+
+has_build_warnings() {
+    filter_warnings > /dev/null
+}
+
+build_with_linkchecker() {
+    rm -rf "${linksdir}"
+    
+    sphinx-build -b linkcheck -d "${build_dir}/doctrees" . "${linksdir}" \
+        -E -w "${warnings}" 2>/dev/null
```
qmonnet:
```suggestion
        --color -E -w "${warnings}" 2>/dev/null
```

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+permissions: read-all
+
+jobs:
+  links-checker:
+    name: Documentation links checker
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
```
qmonnet:
```suggestion
        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
```

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          persist-credentials: false 
+      - name: Check Documentation links
+        uses: docker://quay.io/cilium/docs-builder:4b194c0facc5d1d3b9c0dfef3529ee7c661a6b73@sha256:095cf9621cf46c6c878e25c3a98f02ef3d0aed1154936663e352fd3fe5e7613d
```
qmonnet:
```suggestion
        uses: docker://quay.io/cilium/docs-builder:9eb38bbec80e56a8a498b0cac42456e42cfe7f04@sha256:9936a2f648cb25cad6cf1fe5772ad1a5800d0033eb88b0dce77ad70a083661d9
```

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          entrypoint: ./Documentation/check-links.sh
+      - name: Upload Documentation links report
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v3
```
qmonnet:
```suggestion
        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
```

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Search for existing issue number
+        id: search-issue
+        run: |
+          set -ex
+          encoded_issue_name=$(echo "$ISSUE_NAME" | sed 's/ /%20/g')
+          response=$(curl -s -X GET "https://api.github.com/search/issues?q=$encoded_issue_name+repo:cilium/cilium+state:open+label:automated-issue&type=Issues")
+          issue_number=$(echo "$response" | jq -r '.items[0].number // empty')
+          echo "issue_number=$issue_number" >> $GITHUB_OUTPUT
+          echo $issue_number
+      - uses: actions/checkout@v3
```
qmonnet:
```suggestion
      - uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
```

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,71 @@
...
+        name: Checkout
+      - uses: actions/download-artifact@v3
```
qmonnet:
```suggestion
      - uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
```

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,71 @@
...
+        id: search-issue
+        run: |
+          set -ex
+          encoded_issue_name=$(echo "$ISSUE_NAME" | sed 's/ /%20/g')
+          response=$(curl -s -X GET "https://api.github.com/search/issues?q=$encoded_issue_name+repo:cilium/cilium+state:open+label:automated-issue&type=Issues")
+          issue_number=$(echo "$response" | jq -r '.items[0].number // empty')
+          echo "issue_number=$issue_number" >> $GITHUB_OUTPUT
+          echo $issue_number
+      - uses: actions/checkout@v3
+        name: Checkout
```
qmonnet: Do we need to check out the code at this stage?

Documentation/check-links.sh
```diff
@@ -0,0 +1,47 @@
...
+
+has_build_warnings() {
+    filter_warnings > /dev/null
+}
+
+build_with_linkchecker() {
+    rm -rf "${linksdir}"
+    
+    sphinx-build -b linkcheck -d "${build_dir}/doctrees" . "${linksdir}" \
+        --color -E -q -w "${warnings}" -W --keep-going 2>/dev/null
```
qmonnet:
```suggestion
        --color -E -w "${warnings}" -W --keep-going 2>/dev/null
```

Documentation/conf.py
```diff
@@ -175,7 +175,7 @@
 spelling_filters = ["cilium_spellfilters.WireGuardFilter"]
 
 # Ignore some warnings from MyST parser
-suppress_warnings = ['myst.header', 'myst.xref_missing']
+suppress_warnings = ["myst.header", "myst.xref_missing"]
```
qmonnet: This seems unrelated? Probably because this line was edited recently and your initial version would add the same warning exception? Let's keep the quotes, the file already uses a mix of both anyway.

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,71 @@
...
+          title: ${{ env.ISSUE_NAME }}
+          content-filepath: ${{ github.workspace }}/filtered_links_warnings.txt
+          issue-number: ${{ steps.search-issue.outputs.issue_number }}
+          labels: automated-issue
+      - name: Close automated issue
+        if: ${{ needs.links-checker.outputs.exit_code == 0 && steps.search-issue.outputs.issue_number != '' }}
+        uses: peter-evans/close-issue@276d7966e389d888f011539a86c8920025ea0626 # v3.0.1
+        with:
+          issue-number: ${{ steps.search-issue.outputs.issue_number }}
+          comment: '[Periodic links check](https://github.com/cilium/cilium/actions/workflows/check-links-cron.yaml) no longer finds broken links, closing issue.'
```
qmonnet: Nit: There is no linebreak at the end of the file (this probably depends on your editor's configuration), can you please add one?

qmonnet (commented):
Looks mostly good, I've got some minor comments on Docker image pinning and a few other things. The link checker seems to run well, now we need to validate the issue creation on broken link detection.

> The checks for creating/updating the issue needs the write permission. I don't have the permission to assign that to the github workflow stage.

I think I should be able to check if I open another PR from the cilium repository? I'll give it a try.

.github/workflows/check-links-cron.yaml
```diff
@@ -0,0 +1,71 @@
...
+          encoded_issue_name=$(echo "$ISSUE_NAME" | sed 's/ /%20/g')
+          response=$(curl -s -X GET "https://api.github.com/search/issues?q=$encoded_issue_name+repo:cilium/cilium+state:open+label:automated-issue&type=Issues")
+          issue_number=$(echo "$response" | jq -r '.items[0].number // empty')
+          echo "issue_number=$issue_number" >> $GITHUB_OUTPUT
+          echo $issue_number
+      - uses: actions/checkout@v3
+        name: Checkout
+      - uses: actions/download-artifact@v3
+        name: Download Documentation links report
+        if: ${{ failure() && needs.links-checker.outputs.status == 'failure' }}
```
qmonnet: While the `failure()` here? Doesn't it mean that we run the step only if the previous step (checking out the code) fails? It seems to have prevented the download from happening on this test: https://github.com/cilium/cilium/actions/runs/6314074348/job/17143831502?pr=28284

.github/workflows/check-links-cron.yaml
vipul-21: I guess not.

.github/workflows/check-links-cron.yaml
vipul-21: Isn't the `failure` suppose to check if any of the steps fails of job ? An in case of chain, the dependent job's steps are also considered.

qmonnet: > Isn't the `failure` suppose to check if any of the steps fails of job ?

Honestly I'm not familiar enough to tell. All I know is something prevented the artifact download to run on my test. Let me try another time with your updated version

Documentation/check-links.sh
```diff
@@ -0,0 +1,47 @@
...
+}
+
+has_build_warnings() {
+    filter_warnings > /dev/null
+}
+
+build_with_linkchecker() {
+    rm -rf "${linksdir}"
+    
+    sphinx-build -b linkcheck -d "${build_dir}/doctrees" . "${linksdir}" \
```
qmonnet: Nit: trailing spaces

```suggestion
    rm -rf "${linksdir}"

    sphinx-build -b linkcheck -d "${build_dir}/doctrees" . "${linksdir}" \
```

qmonnet (changes requested):
Looking at [my latest run](https://github.com/cilium/cilium/actions/runs/6327248302/job/17183023483), it was maybe not the `failure()` that blocked the artifact download, but possibly the `needs.links-checker.outputs.status == 'failure'` instead: I don't think we're ever setting this output for `links-checker`? Looking at a similar usage of `outputs.status` in file `.github/workflows/conformance-ginkgo.yaml`, there is `echo "status=failure" >> $GITHUB_OUTPUT` to set the status, it's not automatic. So we probably need something similar.

Annoyingly, the link checker step also failed (consistently, it seems) on this new run, and I've no idea what's causing it:

```
[...]
reading sources... [ 58%] gettingstarted/hubble_setup
reading sources... [ 58%] gettingstarted/k8s-install-default
Please fix the following documentation warnings:

Exception occurred:
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2297, in comment
    self.document.include_log.pop()
IndexError: pop from empty list
The full traceback has been saved in /tmp/sphinx-err-1xnht2gf.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!

Adding the warnings to the file.
```

Did you observe this before, by any chance?

qmonnet:
I can reproduce the link checker issue locally and dump the logs, but that's not super useful. Looks like a Sphinx bug (we recently bumped the version and updated the `docs-builder` image). I'd like to check if this happens with the latest version, but I think we're stuck to the current one because of some dependencies :disappointed:.

```
$ docker run --workdir /github/workspace --rm --entrypoint "./Documentation/check-links.sh" -v $PWD:"/github/workspace" quay.io/cilium/docs-builder:9eb38bbec80e56a8a498b0cac42456e42cfe7f04@sha256:9936a2f648cb25cad6cf1fe5772ad1a5800d0033eb88b0dce77ad70a083661d
```

<details><summary>Logs from Sphinx</summary>

```
# Platform:         linux; (Linux-6.2.0-33-generic-x86_64-with)
# Sphinx version:   7.1.2
# Python version:   3.11.5 (CPython)
# Docutils version: 0.18.1
# Jinja2 version:   3.1.2
# Pygments version: 2.16.1

# Last messages:
#   gettingstarted/hubble_intro
#

#   reading sources... [ 58%]
#   gettingstarted/hubble_setup
#

#   reading sources... [ 58%]
#   gettingstarted/k8s-install-default
#

# Loaded extensions:
#   sphinx.ext.mathjax (7.1.2)
#   alabaster (0.7.13)
#   sphinxcontrib.applehelp (1.0.4)
#   sphinxcontrib.devhelp (1.0.2)
#   sphinxcontrib.htmlhelp (2.0.1)
#   sphinxcontrib.serializinghtml (1.1.5)
#   sphinxcontrib.qthelp (1.0.3)
#   myst_parser (2.0.0)
#   sphinx.ext.ifconfig (7.1.2)
#   sphinx.ext.githubpages (7.1.2)
#   sphinx.ext.extlinks (7.1.2)
#   sphinxcontrib.httpdomain (unknown version)
#   sphinxcontrib.openapi (0.8.1)
#   sphinx_tabs.tabs (unknown version)
#   sphinxcontrib.googleanalytics (0.3)
#   sphinxcontrib.spelling (8.0.0)
#   versionwarning.extension (1.1.2)

# Traceback:
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/sphinx/cmd/build.py", line 290, in build_main
    app.build(args.force_all, args.filenames)
  File "/usr/local/lib/python3.11/site-packages/sphinx/application.py", line 351, in build
    self.builder.build_update()
  File "/usr/local/lib/python3.11/site-packages/sphinx/builders/__init__.py", line 290, in build_update
    self.build(to_build,
  File "/usr/local/lib/python3.11/site-packages/sphinx/builders/__init__.py", line 310, in build
    updated_docnames = set(self.read())
                           ^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/sphinx/builders/__init__.py", line 417, in read
    self._read_serial(docnames)
  File "/usr/local/lib/python3.11/site-packages/sphinx/builders/__init__.py", line 438, in _read_serial
    self.read_doc(docname)
  File "/usr/local/lib/python3.11/site-packages/sphinx/builders/__init__.py", line 494, in read_doc
    publisher.publish()
  File "/usr/local/lib/python3.11/site-packages/docutils/core.py", line 217, in publish
    self.document = self.reader.read(self.source, self.parser,
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/sphinx/io.py", line 104, in read
    self.parse()
  File "/usr/local/lib/python3.11/site-packages/docutils/readers/__init__.py", line 78, in parse
    self.parser.parse(self.input, document)
  File "/usr/local/lib/python3.11/site-packages/sphinx/parsers.py", line 80, in parse
    self.statemachine.run(inputlines, document, inliner=self.inliner)
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 170, in run
    results = StateMachineWS.run(self, input_lines, input_offset,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/statemachine.py", line 240, in run
    context, next_state, result = self.check_line(
                                  ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/statemachine.py", line 452, in check_line
    return method(match, context, next_state)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 3018, in text
    self.section(title.lstrip(), source, style, lineno + 1, messages)
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 327, in section
    self.new_subsection(title, lineno, messages)
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 393, in new_subsection
    newabsoffset = self.nested_parse(
                   ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 281, in nested_parse
    state_machine.run(block, input_offset, memo=self.memo,
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 196, in run
    results = StateMachineWS.run(self, input_lines, input_offset)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/statemachine.py", line 240, in run
    context, next_state, result = self.check_line(
                                  ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/statemachine.py", line 452, in check_line
    return method(match, context, next_state)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2779, in underline
    self.section(title, source, style, lineno - 1, messages)
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 327, in section
    self.new_subsection(title, lineno, messages)
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 393, in new_subsection
    newabsoffset = self.nested_parse(
                   ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 281, in nested_parse
    state_machine.run(block, input_offset, memo=self.memo,
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 196, in run
    results = StateMachineWS.run(self, input_lines, input_offset)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/statemachine.py", line 240, in run
    context, next_state, result = self.check_line(
                                  ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/statemachine.py", line 452, in check_line
    return method(match, context, next_state)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2352, in explicit_markup
    nodelist, blank_finish = self.explicit_construct(match)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2364, in explicit_construct
    return method(self, expmatch)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2101, in directive
    return self.run_directive(
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2151, in run_directive
    result = directive_instance.run()
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/sphinx_tabs/tabs.py", line 103, in run
    self.state.nested_parse(self.content, self.content_offset, node)
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 281, in nested_parse
    state_machine.run(block, input_offset, memo=self.memo,
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 196, in run
    results = StateMachineWS.run(self, input_lines, input_offset)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/statemachine.py", line 240, in run
    context, next_state, result = self.check_line(
                                  ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/statemachine.py", line 452, in check_line
    return method(match, context, next_state)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2354, in explicit_markup
    self.explicit_list(blank_finish)
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2379, in explicit_list
    newline_offset, blank_finish = self.nested_list_parse(
                                   ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 318, in nested_list_parse
    state_machine.run(block, input_offset, memo=self.memo,
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 196, in run
    results = StateMachineWS.run(self, input_lines, input_offset)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/statemachine.py", line 240, in run
    context, next_state, result = self.check_line(
                                  ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/statemachine.py", line 452, in check_line
    return method(match, context, next_state)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2657, in explicit_markup
    nodelist, blank_finish = self.explicit_construct(match)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2364, in explicit_construct
    return method(self, expmatch)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2101, in directive
    return self.run_directive(
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2151, in run_directive
    result = directive_instance.run()
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/sphinx_tabs/tabs.py", line 220, in run
    node = super().run()
           ^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/sphinx_tabs/tabs.py", line 197, in run
    self.state.nested_parse(self.content[1:], self.content_offset, panel)
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 281, in nested_parse
    state_machine.run(block, input_offset, memo=self.memo,
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 196, in run
    results = StateMachineWS.run(self, input_lines, input_offset)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/statemachine.py", line 240, in run
    context, next_state, result = self.check_line(
                                  ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/statemachine.py", line 452, in check_line
    return method(match, context, next_state)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2352, in explicit_markup
    nodelist, blank_finish = self.explicit_construct(match)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2370, in explicit_construct
    nodelist, blank_finish = self.comment(match)
                             ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docutils/parsers/rst/states.py", line 2297, in comment
    self.document.include_log.pop()
IndexError: pop from empty list
```
</details>

qmonnet:
Looks like a bug caused by the bump of docutils from 0.17.1 to 0.18.1 - bisecting shows it is introduced by https://repo.or.cz/docutils.git/commitdiff/93dc79108625a3af6be11a3eb85e50be59b2c910?hp=0515f37ca519ecfde133040425a06b6e4887aea0.

Not sure how to handle this, though :thinking:

ti-mo:
Ping, anything we can do to move this along?

qmonnet:
Thanks Timo. I must admit I'm a bit stuck here. I talked with @zacharysarah about this PR, and we don't see a good path forward.

This bug in the link checker is clearly an issue. [We know the cause](https://github.com/cilium/cilium/pull/27116#issuecomment-1737813302), but it's tricky to address.

- We can't just ignore the bug if we enable the link checker, because it won't run at all, this is not a warning we could just skip.

- We could report or try to fix the bug upstream. However, we're already behind on docutils versions: we're limited to 0.18.1 because of the `sphinx-tabs` extension, when 0.20.0 is out already. This means that even if the issue is fixed upstream, we won't be able to pull it immediately, likely not before a few years. Even if `sphinx-tabs` got updated quickly, there's still an issue with Netlify not being aware of the latest versions of Sphinx and dependencies in its repo, forcing us to stay a few versions behind. So I don't see the fix available from upstream before a couple years at least.

- We could revert to a working version of docutils, but this would mean downgrading Sphinx and several other add-ons - in other words, reverting all the recent version changes, and this may get us stuck with older versions until the bug is fixed upstream - which will take a while, as explained above. I don't really expect security vulnerabilities in Sphinx or its add-ons to be exploitable, but being stuck on old versions does not sound great.

- The bug itself is easy to “fix” (although I'm not sure this is the right fix). We could [use a fork of docutils to add a fix](https://github.com/qmonnet/docutils/commit/b3d5049dd8e2fec0f5d0380e4922878360f46b55), which is not difficult to do but means more work for future upgrades.

Add to that that there are internal discussions about moving away from Sphinx in the future (but nothing decided yet), so there's little appetite for long-term commitments/solutions at the moment.

Maybe the simplest option for now would be to drop the CI workflow (sorry Vipul :cry:), and try to merge `check-links.sh` with the accompanying config update and doc fixes. Possibly adding an `echo` at the beginning of the script to print a warning about the fact it may be broken, and link to this discussion?

vipul-21:
I can remove the workflow and update the  script, if that is the suggested path forward. Let me know.

qmonnet (changes requested):
The changes look all good, thanks a lot.

On the second commit, can you please:

- Update the commit title, given that we no longer enable the link checker in the CI
- Add a commit description that summarises what we're doing here (and why we don't enable it in CI)
- Drop the `Co-authored-by:` tag, or updated it with my work email address (`<quentin@isovalent.com>`)

We should be all good after that

qmonnet (approved):
Thanks a lot for this work!

Documentation/check-links.sh
```diff
@@ -0,0 +1,48 @@
...
+
+has_build_warnings() {
+    filter_warnings > /dev/null
+}
+
+build_with_linkchecker() {
+    rm -rf "${linksdir}"
+
+    sphinx-build -b linkcheck -d "${build_dir}/doctrees" . "${linksdir}" \
+        --color -E -w "${warnings}" -W --keep-going 2>/dev/null
```
joestringer, line 35: Why is stderr being redirected here? Could this hide real errors?

qmonnet:
```console
$ sphinx-build --help
[...]
  -w FILE           write warnings (and errors) to given file
```

We keep (and print, when relevant) both warnings and errors in the provided file, so I think it's safe to ignore `stderr`. If we keep it, it makes duplicate entries with the file that we dump at the end. One motivation for dumping the file instead of simply using `stderr` is that the errors and warnings can easily get lost in the middle of the info-level output.

joestringer (approved):

joestringer:
Thanks for the contribution, merged!

PR #27117, opened 2023-07-27
ldelossa/register bgpserver fix

ldelossa:
Fixes an issue where an empty ControlPlaneState was used during registration of BGP speakers. 

```release-note
Fixes an issue where an empty ControlPlaneState was used during registration of BGP speakers. This would cause reconciliation issues as the current state would be unknown.
```

### DIFF
--- a/pkg/bgpv1/manager/manager.go
+++ b/pkg/bgpv1/manager/manager.go
@@ -224,7 +224,7 @@ func (m *BGPRouterManager) registerBGPServer(ctx context.Context, c *v2alpha1api
 				AdvertiseInactiveRoutes: true,
 			},
 		},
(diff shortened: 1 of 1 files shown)


rastislavs (approved):
LGTM

joestringer:
Is there a user-facing impact of this? I found it a bit difficult to really understand from the release note. If yes, could you clarify the release note further? If no, then `release-note/misc` or `release-note/bug` may be more appropriate (obviously with backport labels if relevant, based on backporting policy if this is a bugfix).

ldelossa:
@joestringer okay updated. No backporting, since the bug was introduced before after the last release.

PR #27122, opened 2023-07-27
docs: Replace non-portable "sed -i" in Makefile

qmonnet:
We recently added a call to `sed` in Documentation's Makefile, as part of the generation process for the Helm reference. We use the `-i` option to edit the file in-place; but this option is not portable, and the command fails when running BSD versions of sed, for example on MacOS, as opposed to GNU sed.

Let's fix the command by using an additional temporary file.

Fixes: 2e9b20fcf1a7 ("docs: Ignore Helm value names for spellcheck")

### DIFF
--- a/Documentation/Makefile
+++ b/Documentation/Makefile
@@ -105,13 +105,14 @@ M2R := $(DOCKER_CTR) $(HELM_TOOLBOX_IMAGE) python3 /usr/bin/m2r2
 .PHONY: update-helm-values FORCE
 $(HELM_VALUES): TMP_FILE_1 := helm-values.tmp
 $(HELM_VALUES): TMP_FILE_2 := helm-values.awk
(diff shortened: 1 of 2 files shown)


joestringer:
Heads up that `Makefile.defs` defines `SED` as `gsed` for platforms that don't have a GNU sed. We should probably be trying to replace usage of `sed` with `$(SED)`. That said, I'm not against using more portable expressions :+1:

qmonnet:
I didn't know that. OK I'll update with `$(SED)`, but I can also keep the temp file to avoid using `-i` anyway.

learnitall (approved):
Looks good! Just a heads up though, I'm on Linux and haven't tested BSD or Mac OS.

jibi (approved):

PR #27134, opened 2023-07-28
bpf: tests: add helpers for common patterns

julianwiedmann:
Clean up some of the copy & pasted boilerplate code that sets up a test scenario.

### DIFF
--- a/bpf/tests/lib/lb.h
+++ b/bpf/tests/lib/lb.h
@@ -1,0 +1,118 @@
+/* SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause) */
+/* Copyright Authors of Cilium */
+
+#ifdef ENABLE_IPV4
+static __always_inline void
+lb_v4_add_service(__be32 addr, __be16 port, __u16 backend_count, __u16 rev_nat_index)
+{
+	struct lb4_key svc_key = {
+		.address = addr,
+		.dport = port,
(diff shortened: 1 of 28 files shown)


bpf/tests/lib/lb.h
```diff
@@ -0,0 +1,118 @@
...
+	};
+	struct lb4_service svc_value = {
+		.count = backend_count,
+		.flags = SVC_FLAG_ROUTABLE,
+		.rev_nat_index = rev_nat_index,
+	};
+	map_update_elem(&LB4_SERVICES_MAP_V2, &svc_key, &svc_value, BPF_ANY);
+	/* Register with both scopes: */
+	svc_key.scope = LB_LOOKUP_SCOPE_INT;
+	map_update_elem(&LB4_SERVICES_MAP_V2, &svc_key, &svc_value, BPF_ANY);
```
gentoo-root, line 21: Not all tests used to register in the internal scope; I hope that shouldn't break anything.

gentoo-root (approved):
Nice reduction of repeating code! I hope there are no copy-paste mistakes :)

bpf/tests/lib/lb.h
julianwiedmann: Yep I think over time we should grow special helpers for the different kinds of service configurations - `eTP:local` etc. But at least nothing was specifically testing for the current behaviour.

PR #27114, opened 2023-07-27
update upgrade tests to test from v1.14.0 to main

aanm:

PR #27136, opened 2023-07-28
test: Enable IPv6 BPF masquerading for upgrade tests

qmonnet:
Now that Cilium v1.14, with support for IPv6 BPF masquerading, has been released, we can enable IPv6 BPF masquerading in the upgrade tests, and lift the restriction introduced in commit d6f4c53ba443.

To do that, we first need to update the version from which we upgrade: It's time to test upgrades from 1.14 in the CI.

### DIFF
--- a/test/helpers/cons.go
+++ b/test/helpers/cons.go
@@ -175,9 +175,9 @@ const (
 
 	// CiliumStableHelmChartVersion should be the chart version that points
 	// to the v1.X branch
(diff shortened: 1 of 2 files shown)


aanm (changes requested):
Thanks for the PR but I think it's a no-op since the changes were part of https://github.com/cilium/cilium/pull/27114?

qmonnet:
Yes. I missed that other PR when opening mine. Thanks!

PR #27137, opened 2023-07-28
refactor: stale services GC to be executed in hive jobs

### LINKED ISSUE #27012: Refactor stale services GC to leverage an hive job
Currently, the service GC process to ensure that lb maps and k8s services (and backends) are synchronized, and remove stale entries is [triggered through a controller run by a goroutine](https://github.com/cilium/cilium/blob/32118ccaa8677a3b836e60542a52a2af94693533/daemon/cmd/state.go#L451-L478). Given that this is a one-off operation performed when the agent bootstraps, it should be managed through an hive job. This would allow to explicit dependencies, simplify testing and improve visibility.

Related: https://github.com/cilium/cilium/pull/26912

vakalapa:

      GitHub issue.

This change will remove controller based retry to GC leftover k8s services instead uses a hive job to achieve the same. 
Also adds a jobGroup to daemon for any later needs of jobs to be done in daemon. 

Fixes: #27012 

```release-note
```

### DIFF
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -51,6 +51,7 @@ import (
 	"github.com/cilium/cilium/pkg/eventqueue"
 	"github.com/cilium/cilium/pkg/fqdn"
 	"github.com/cilium/cilium/pkg/hive/cell"
+	"github.com/cilium/cilium/pkg/hive/job"
 	"github.com/cilium/cilium/pkg/hubble/observer"
 	"github.com/cilium/cilium/pkg/identity"
 	"github.com/cilium/cilium/pkg/identity/identitymanager"
@@ -104,6 +105,8 @@ const (
 	ConfigModifyQueueSize = 10
 
 	syncHostIPsController = "sync-host-ips"
+
+	daemonJobs = "daemon-jobs"
 )
 
 // Daemon is the cilium daemon that is in charge of perform all necessary plumbing,
@@ -223,6 +226,9 @@ type Daemon struct {
 	// enable modules health support
 	healthProvider cell.Health
 	healthReporter cell.HealthReporter
+
+	// jobGroup to run any jobs needed by the daemon
+	jobGroup job.Group
 }
 
 func (d *Daemon) initDNSProxyContext(size int) {
@@ -511,6 +517,10 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 		return nil, nil, err
 	}
 
+	jobGroup := params.JobRegistry.NewGroup(
+		job.WithLogger(log.WithField(logfields.JobName, daemonJobs)),
+	)
+
 	d := Daemon{
 		ctx:               ctx,
 		clientset:         params.Clientset,
@@ -543,6 +553,7 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 		settings:             params.Settings,
 		healthProvider:       params.HealthProvider,
 		healthReporter:       params.HealthReporter,
+		jobGroup:             jobGroup,
 	}
 
 	d.configModifyQueue = eventqueue.NewEventQueueBuffered("config-modify-queue", ConfigModifyQueueSize)
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -105,8 +105,6 @@ const (
 	ConfigModifyQueueSize = 10
 
 	syncHostIPsController = "sync-host-ips"
-
-	daemonJobs = "daemon-jobs"
 )
 
 // Daemon is the cilium daemon that is in charge of perform all necessary plumbing,
@@ -518,8 +516,9 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 	}
 
 	jobGroup := params.JobRegistry.NewGroup(
-		job.WithLogger(log.WithField(logfields.JobName, daemonJobs)),
+		job.WithLogger(log),
 	)
+	params.Lifecycle.Append(jobGroup)
 
 	d := Daemon{
 		ctx:               ctx,
--- a/daemon/cmd/state.go
+++ b/daemon/cmd/state.go
@@ -14,9 +14,10 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/util/workqueue"
 
-	"github.com/cilium/cilium/pkg/controller"
 	"github.com/cilium/cilium/pkg/endpoint"
+	"github.com/cilium/cilium/pkg/hive/job"
 	"github.com/cilium/cilium/pkg/ipam"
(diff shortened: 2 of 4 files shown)


daemon/cmd/state.go
```diff
@@ -439,43 +444,52 @@ func (d *Daemon) allocateIPsLocked(ep *endpoint.Endpoint) (err error) {
...
-			if d.clientset.IsEnabled() {
-				// Also wait for all cluster mesh to be synchronized with the
-				// datapath before proceeding.
-				if d.clustermesh != nil {
-					err := d.clustermesh.ClustersSynced(d.ctx)
-					if err != nil {
-						log.WithError(err).Fatal("timeout while waiting for all clusters to be locally synchronized")
+		d.jobGroup.Add(
+			job.OneShot("empty restoreComplete", func(_ context.Context) error {
+				<-restoreComplete
```
tommyp1ckles: why do we need this job?

daemon/cmd/state.go
```diff
@@ -439,43 +444,52 @@ func (d *Daemon) allocateIPsLocked(ep *endpoint.Endpoint) (err error) {
...
-		}()
+				return nil
+			}),
+			// Start a job which removes any leftover Kubernetes
+			// services that may have been deleted while Cilium was not
+			// running. Since this is a OneShot job, it will not run again unless
+			// updated elsewhere. This means that if, for instance, a user manually
+			// adds a service via the CLI into the BPF maps, that it will
+			// not be cleaned up by the daemon until it restarts.
+			job.OneShot(
```
tommyp1ckles: Should we just avoid registering the job if the k8s client is disabled?

daemon/cmd/state.go
vakalapa: Removed this redundant job.

daemon/cmd/state.go
vakalapa: makes sense, wrapped the job creation with clientsent Enabled check

daemon/cmd/state.go
```diff
@@ -439,43 +444,44 @@ func (d *Daemon) allocateIPsLocked(ep *endpoint.Endpoint) (err error) {
...
-		go func() {
-			if d.clientset.IsEnabled() {
-				// Also wait for all cluster mesh to be synchronized with the
-				// datapath before proceeding.
-				if d.clustermesh != nil {
-					err := d.clustermesh.ClustersSynced(d.ctx)
-					if err != nil {
-						log.WithError(err).Fatal("timeout while waiting for all clusters to be locally synchronized")
+		if d.clientset.IsEnabled() {
+			d.jobGroup.Add(
```
tommyp1ckles: When I tested these locally (using the clustersmesh testing scripts), I didn't see these jobs running (similarly when I rain main these tasks where run).

Seems like there's potentially some issue with the jobGroup being invoked correctly?

vakalapa: Do you have any docs pointers for me to run these scripts locally ?

tommyp1ckles: Unfortunately, it doesn't seem like we document them very well, I ran:

```bash
make kind-clustermesh
make kind-clustermesh-images
kind-install-cilium-clustermesh
```

Which creates two clusters and installs cilium w/ clustermesh on both.

vakalapa: ok i tried creating clusters and running cilium in my WSL setup, and my agents are in crash loop, may be its not happy that i am running a 5.15 WSL ubuntu. Do you a recommendation on what OS i can test on. (may be real Ubuntu22 ?)

`iptables command: exit status 1 stderr=\"Warning: Extension CT revision 0 not supported, missing kernel module?`

giorio94: The issue here is that the registry is not registered. You can find one example [here](https://github.com/cilium/cilium/blob/main/pkg/l2announcer/l2announcer.go#L105-L138). I'd also extract this logic outside of this function, mimicking the example above.

daemon/cmd/state.go
```diff
@@ -439,43 +444,44 @@ func (d *Daemon) allocateIPsLocked(ep *endpoint.Endpoint) (err error) {
...
-				controller.NewManager().UpdateController("sync-lb-maps-with-k8s-services",
-					controller.ControllerParams{
-						DoFunc: func(ctx context.Context) error {
-							return d.svc.SyncWithK8sFinished(d.k8sWatcher.K8sSvcCache.EnsureService)
-						},
-						Context: d.ctx,
+				job.OneShot(
+					"sync-lb-maps-with-k8s-services",
+					func(ctx context.Context) error {
+						<-clusterSyncComplete
```
giorio94: I would not split the logic in two separate jobs, as they are strictly related. In particular, it is necessary to wait until all clusters are synchronized before triggering the service synchronization.

daemon/cmd/daemon.go
```diff
@@ -511,6 +517,10 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 		return nil, nil, err
 	}
 
+	jobGroup := params.JobRegistry.NewGroup(
+		job.WithLogger(log.WithField(logfields.JobName, daemonJobs)),
```
giorio94: The job name is already added automatically as field: https://github.com/cilium/cilium/blob/main/pkg/hive/job/job.go#L273-L276

giorio94 (changes requested):
Thanks @vakalapa for the PR! 

I've left you a few comments inline. Feel free to ask for any further clarification.

daemon/cmd/state.go
vakalapa: OK will combine them.

daemon/cmd/daemon.go
vakalapa: Ak okay.

daemon/cmd/state.go
vakalapa: do you recommend using this same daemon pkg or creating a new one under cilium/pkg? I like the earlier option because this might be a oneoff use case for only daemon on init ?

giorio94 (changes requested):
Hooks cannot be registered from another start hook. 

More in general, the idea with the Hive framework is to split the overall codebase in multiple components (i.e., cells), which can in turn depend on other components, populated through dependency injection. In this case, we could have a `Service` cell, including a provide function which depends on the set of parameters (e.g., the ones used [here](https://github.com/cilium/cilium/blob/main/daemon/cmd/daemon.go#L630)) and returns the initialized `Service` object. Additionally, it registers the appropriate start hooks (for instance the jobs to do stale services GC) to handle the lifecycle of the given object.

Converting a given submodule into a cell typically requires also some additional refactoring (e.g., to remove the dependency on the whole daemon object, replacing it with more specialized objects/interfaces). 

The ultimate goal is to get rid of the monolithic daemon initialization logic, making it the result of the composition of the different dependencies only.

lmb:
@vakalapa are you still planning to land this PR? If yes, please rebase and address the feedback. Otherwise feel free to close it!

joestringer:
(converting to draft while the contributor is addressing feedback. @vakalapa , when you are ready to continue with this PR please click "Ready for review" at the bottom of the page.)

PR #27141, opened 2023-07-29
fix: use proper helm param name for specifying pod cidr

yandzee:


learnitall (approved):
Nice catch! Thank you! One small nit that I have for you is the commit message, which could be a bit more specific. Something like:

```
docs: remove `clusterPoolIPv4PodCIDR option in k3s install

The option `clusterPoolIPv4PodCIDR option was deprecated in
Cilium 1.11 and removed from helm values in commit 1eef5ac.
This commit replaces an outdated reference to this option in
the k3s installation docs with the preferred option
`clusterPoolIPv4PodCIDRList`
```

qmonnet:
Introduced in d3f2bcb2364d ("Doc: Update k3s cilium installation to match k3s default podCIDR") which is in 1.14, I'm marking for backports for that release.

PR #27143, opened 2023-07-30
[Draft] Structured Health Reporter and EndpointManager Health

tommyp1ckles:


pkg/hive/cell/scoped/scoped.go
```diff
@@ -0,0 +1,177 @@
...
+	log.Warn("stopped called on scoped reporter")
+}
+
+func (s *scopedReporter) clear(id string) {
+	s.Lock()
+	defer s.Unlock()
+	delete(s.status, id)
+}
+
+func WithLabels(ctx context.Context, hr cell.HealthReporter, labels Labels) *scopedReporter {
```
asauber, line 58: This is a great concept, creating a tree of health reporters using a "create named child" mechanism.

At the moment, I don't think this is named appropriately.

The implied function name is `(hive.cell).scoped.WithLabels`. This indicates that this is a generic hive mechanism to extend (? copy? reference?) an existing cell (implicit from the signature). If we made this more generic for all types of cells (I don't think we want to), and made the semantics more clear, then this name might be appropriate.

A better name for this function given it's current location and parameters would be `NewHealthReporterFromParent`. 

I think the signature is almost exactly what we need. Keeping a set of Labels as a required parameter is important. Given that a caller could mistakenly provide an empty `map[string]string` of labels, adding a required `name string` parameter could be useful.

tommyp1ckles: Agreed, for this draft naming was quite fluid.  I'm hoping to converge on something better.

tommyp1ckles: > The implied function name is (hive.cell).scoped.WithLabels. This indicates that this is a generic hive mechanism to extend (? copy? reference?) an existing cell (implicit from the signature). If we made this more generic for all types of cells (I don't think we want to), and made the semantics more clear, then this name might be appropriate.

It doesn't need to belong in pkg//hive/cell.

> A better name for this function given it's current location and parameters would be NewHealthReporterFromParent.

> Given that a caller could mistakenly provide an empty map[string]string of labels, adding a required name string parameter could be useful.

Yeah I like both of these suggestions 😄 👍

PR #27147, opened 2023-07-31
Increase number of dnsproxy mutexes from 128 to 131.

marseel:
Using a prime number ensures that we are using all bits of IP for hashing.


pippolo84 (approved):
Nice, thanks! :rocket:

PR #27150, opened 2023-07-31
ci-ipsec-upgrade: Derive Cilium version to upgrade from / downgrade to

brb:


brlbil (approved):

.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -57,7 +57,7 @@ env:
   cilium_cli_ci_version: 8c624e141844a19a6b2e21d5255f3ca6195c996d
   check_url: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
   upgrade_cilium_from_vsn: 1.13.4
-  upgrade_cilium_to_vsn: 1.14.0-rc.1 # empty means upgrade to CI build
+  upgrade_cilium_to_vsn: 1.14.0 # empty means upgrade to CI build
```
pchaigno: Shouldn't this be always empty anyway? We want to cover the CI builds like we did in K8sUpdates. Otherwise, we'll always learn about regressions after merging.

pchaigno (approved):

.github/workflows/tests-ipsec-upgrade.yaml
brb: :+1:  Let me do this in this PR.

brb: Done. Once this PR is merged, we can try to backport it to the v1.1{1,2,3,4} branches.

.github/actions/derive-cilium-vsn/action.yml
```diff
@@ -0,0 +1,21 @@
...
+    required: true
+    default: ''
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      if: ${{ inputs.from_vsn != '' }}
+      run: |
+        VSN=$(curl -fsSL https://raw.githubusercontent.com/cilium/cilium/${{ inputs.from_vsn }}/VERSION)
+        echo "CILIUM_VSN=${VSN}" >> $GITHUB_ENV
```
pchaigno: In K8sUpdates, we used to up/downgrade from/to the stable branch and not any specific release. That was convenient when we needed to patch the branch to make either the upgrade or the downgrade work; we didn't have to wait for the patch to be released before seeing the test passing.

Is the same not possible here?

pchaigno (approved):
:+1: on the improvement.

.github/actions/derive-cilium-vsn/action.yml
brb: Ouch, this `from_vsn` naming might be a bit misleading. It should be read as _derive the latest stable version from a given vsn_. For example, `from_vsn=v1.13`, it will derive and set `env.CILIUM_VSN=v1.13.5`. Later on, the test does the following:

1. Install `env.CILIUM_VSN`, and test.
2. Upgrade to the PR built image, and test. 
3. Downgrade to `env.CILIUM_VSN`, and test.

pchaigno: Yeah, I understood that. I'm asking if we can upgrade from v1.13's latest build instead of v1.13.5.

brb: Ah, I see. Make sense. Then we can catch the regressions happening in backport branches too :+1:

Anyway, it seems that there is no easy way to instruct the CLI's install / upgrade to pick the latest unreleased v1.13. I will follow up on this. But first, let's merge this PR.

brb: Oh, found the way how to install. Please hold my beer, trying to do it in this PR :crossed_fingers:

.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -132,12 +131,22 @@ jobs:
...
+            --helm-set=operator.image.useDigest=false \
+            --helm-set=hubble.relay.image.repository=quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/hubble-relay-ci \
+            --helm-set=hubble.relay.image.tag=v${{ env.cilium_stable_version }}"
+          echo "cilium_stable_image_settings=${CILIUM_STABLE_IMAGE_SETTINGS}" >> $GITHUB_OUTPUT
+
           CILIUM_MAIN_IMAGE_SETTINGS="--chart-directory=./install/kubernetes/cilium \
             --helm-set=image.repository=quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/cilium-ci \
             --helm-set=image.useDigest=false \
             --helm-set=image.tag=${SHA} \
-            --helm-set=debug.enabled=true \
```
pchaigno, line 139: Why did you remove that? I understand removing Envoy debug logs, but we typically run all CI with basic debugging enabled.

pchaigno (approved):

.github/workflows/tests-ipsec-upgrade.yaml
brb: One debug enablement is enough :smile:  https://github.com/cilium/cilium/pull/27150/files#diff-07b1303f71b74ecfe10ad34472da7c7e9b79ac9274fd93fe833ecc1551898473R159

pchaigno: Oh, right :facepalm:

aanm (approved):

PR #27160, opened 2023-07-31
statedb v2.0 with per-table locks and delete tracking

joamaki:
"StateDB v2.0" reimplements pkg/statedb to introduce per-table locking, built-in revision numbers and ability to track objects being deleted from tables. It moves away from the go-memdb library and implements directly on top of the immutable-radix-tree library, with the biggest visible change being that indexes are no longer defined using reflection, but with type-safe functions.

To make reviewing easier the new version is first introduced as `pkg/statedb2` and then in a follow-up PR the uses of `pkg/statedb` are converted over to the new API and old code replaced.

See `pkg/statedb2/example` for an overview of how the new API looks like.

### DIFF
--- a/pkg/lock/sortable_mutex.go
+++ b/pkg/lock/sortable_mutex.go
@@ -7,6 +7,7 @@ import (
 	"sort"
 	"sync"
 	"sync/atomic"
+	"time"
 )
 
 // sortableMutexSeq is a global sequence counter for the creation of new
@@ -16,11 +17,20 @@ var sortableMutexSeq atomic.Uint64
 // sortableMutex implements SortableMutex. Not exported as the only way to
 // initialize it is via NewSortableMutex().
 type sortableMutex struct {
-	*sync.Mutex
-	seq uint64
+	sync.Mutex
+	seq             uint64
+	acquireDuration time.Duration
 }
 
-func (s sortableMutex) Seq() uint64 { return s.seq }
+func (s *sortableMutex) Lock() {
+	start := time.Now()
+	s.Mutex.Lock()
+	s.acquireDuration += time.Since(start)
+}
+
+func (s *sortableMutex) Seq() uint64 { return s.seq }
+
+func (s *sortableMutex) AcquireDuration() time.Duration { return s.acquireDuration }
 
 // SortableMutex provides a Mutex that can be globally sorted with other
 // sortable mutexes. This allows deadlock-safe locking of a set of mutexes
@@ -28,6 +38,7 @@ func (s sortableMutex) Seq() uint64 { return s.seq }
 type SortableMutex interface {
 	sync.Locker
 	Seq() uint64
+	AcquireDuration() time.Duration // The amount of time it took to acquire the lock
 }
 
 // SortableMutexes is a set of mutexes that can be locked in a safe order.
@@ -69,8 +80,7 @@ var _ sort.Interface = SortableMutexes{}
 
 func NewSortableMutex() SortableMutex {
 	seq := sortableMutexSeq.Add(1)
-	return sortableMutex{
-		Mutex: new(sync.Mutex),
-		seq:   seq,
+	return &sortableMutex{
+		seq: seq,
 	}
 }
--- a/pkg/lock/sortable_mutex.go
+++ b/pkg/lock/sortable_mutex.go
@@ -1,0 +1,76 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package lock
+
+import (
+	"sort"
+	"sync"
+	"sync/atomic"
+)
+
+// sortableMutexSeq is a global sequence counter for the creation of new
+// SortableMutex's with unique sequence numbers.
+var sortableMutexSeq atomic.Uint64
+
+// sortableMutex implements SortableMutex. Not exported as the only way to
+// initialize it is via NewSortableMutex().
+type sortableMutex struct {
+	*sync.Mutex
+	seq uint64
+}
+
+func (s sortableMutex) Seq() uint64 { return s.seq }
+
+// SortableMutex provides a Mutex that can be globally sorted with other
+// sortable mutexes. This allows deadlock-safe locking of a set of mutexes
+// as it guarantees consistent lock ordering.
+type SortableMutex interface {
+	sync.Locker
+	Seq() uint64
+}
+
+// SortableMutexes is a set of mutexes that can be locked in a safe order.
+// Once Lock() is called it must not be mutated!
+type SortableMutexes []SortableMutex
+
+// Len implements sort.Interface.
+func (s SortableMutexes) Len() int {
+	return len(s)
+}
+
+// Less implements sort.Interface.
+func (s SortableMutexes) Less(i int, j int) bool {
+	return s[i].Seq() < s[j].Seq()
+}
+
+// Swap implements sort.Interface.
+func (s SortableMutexes) Swap(i int, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+// Lock sorts the mutexes, and then locks them in order. If any lock cannot be acquired,
+// this will block while holding the locks with a lower sequence number.
+func (s SortableMutexes) Lock() {
+	sort.Sort(s)
+	for _, mu := range s {
+		mu.Lock()
+	}
+}
+
+// Unlock locks the sorted set of mutexes locked by prior call to Lock().
+func (s SortableMutexes) Unlock() {
+	for _, mu := range s {
+		mu.Unlock()
+	}
+}
+
+var _ sort.Interface = SortableMutexes{}
+
+func NewSortableMutex() SortableMutex {
+	seq := sortableMutexSeq.Add(1)
+	return sortableMutex{
+		Mutex: new(sync.Mutex),
+		seq:   seq,
+	}
+}
--- a/pkg/statedb2/cell.go
+++ b/pkg/statedb2/cell.go
@@ -1,0 +1,57 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package statedb2
+
+import (
+	"github.com/cilium/cilium/pkg/hive"
+	"github.com/cilium/cilium/pkg/hive/cell"
+)
+
+// This module provides an in-memory database built on top of immutable radix trees
+// As the database is based on an immutable data structure, the objects inserted into
+// the database MUST NOT be mutated, but rather copied first!
+//
+// For example use see pkg/statedb/example.
+var Cell = cell.Module(
+	"statedb",
+	"In-memory transactional database",
+
+	cell.Provide(newHiveDB),
+)
+
+type tablesIn struct {
+	cell.In
+
+	TableMetas []TableMeta `group:"statedb-tables"`
+}
+
+func newHiveDB(lc hive.Lifecycle, tablesIn tablesIn) (*DB, error) {
+	db, err := NewDB(tablesIn.TableMetas)
+	if err != nil {
+		return nil, err
+	}
+	lc.Append(db)
+	return db, nil
+}
+
+type tableOut[Obj any] struct {
+	cell.Out
+	Table Table[Obj]
+	Meta  TableMeta `group:"statedb-tables"`
+}
+
+// NewTableCell creates a cell for creating and registering a statedb Table[Obj].
+func NewTableCell[Obj any](
+	tableName TableName,
+	primaryIndexer Indexer[Obj],
+	secondaryIndexers ...Indexer[Obj],
+) cell.Cell {
+	return cell.Provide(func() (tableOut[Obj], error) {
+		if table, err := NewTable(tableName, primaryIndexer, secondaryIndexers...); err != nil {
+			return tableOut[Obj]{}, err
+		} else {
+			return tableOut[Obj]{Table: table, Meta: table}, nil
+		}
+	})
+}
--- a/pkg/statedb2/cell.go
+++ b/pkg/statedb2/cell.go
@@ -18,16 +18,18 @@ var Cell = cell.Module(
 	"In-memory transactional database",
 
 	cell.Provide(newHiveDB),
+	cell.Metric(NewMetrics),
 )
 
 type tablesIn struct {
 	cell.In
 
 	TableMetas []TableMeta `group:"statedb-tables"`
+	Metrics    Metrics
 }
 
 func newHiveDB(lc hive.Lifecycle, tablesIn tablesIn) (*DB, error) {
-	db, err := NewDB(tablesIn.TableMetas)
+	db, err := NewDB(tablesIn.TableMetas, tablesIn.Metrics)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/statedb2/db.go
+++ b/pkg/statedb2/db.go
@@ -1,0 +1,178 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package statedb2
+
+import (
+	"errors"
+	"fmt"
+	"sync/atomic"
+
+	iradix "github.com/hashicorp/go-immutable-radix/v2"
+
+	"github.com/cilium/cilium/pkg/hive"
+	"github.com/cilium/cilium/pkg/lock"
+)
+
+// DB provides an in-memory transaction database built on top of immutable radix
+// trees. The database supports multiple tables, each with one or more user-defined
+// indexes. Readers can access the data locklessly with a simple atomic pointer read
+// to obtain a snapshot. On writes to the database table-level locks are acquired
+// on target tables and on write transaction commit a root lock is taken to swap
+// in the new root with the modified tables.
+//
+// As data is stored in immutable data structures any objects inserted into
+// it MUST NOT be mutated afterwards.
+//
+// DB holds the "root" tree of tables with each table holding a tree of indexes:
+//
+//	           root
+//	          /    \
+//	         ba    T(foo)
+//	       /   \
+//	   T(bar)  T(baz)
+//
+//	      T(bar).indexes
+//		   /  \
+//		  i    I(byRevision)
+//		/   \
+//	   I(id)    I(ip)
+//
+//	          I(ip)
+//	          /  \
+//	        192  172
+//	        /     ...
+//	    bar(192.168.1.1)
+//
+// T = tableEntry
+// I = indexTree
+//
+// To lookup:
+//  1. Create a read (or write) transaction
+//  2. Find the table from the root tree
+//  3. Find the index from the table's index tree
+//  4. Find the object from the index
+//
+// To insert:
+//  1. Create write transaction against the target table
+//  2. Find the table from the root tree
+//  3. Create/reuse write transaction on primary index
+//  4. Insert/replace the object into primary index
+//  5. Create/reuse write transaction on revision index
+//  6. If old object existed, remove from revision index
+//  7. If old object existed, remove from graveyard
+//  8. Update each secondary index
+//  9. Commit transaction by committing each index to
+//     the table and then committing table to the root.
+//     Swap the root atomic pointer to new root and
+//     notify by closing channels of all modified nodes.
+//
+// To observe deletions:
+//  1. Create write transaction against the target table
+//  2. Create new delete tracker and add it to the table
+//  3. Commit the write transaction to update the table
+//     with the new delete tracker
+//  4. Query the graveyard by revision, starting from the
+//     revision of the write transaction at which it was
+//     created.
+//  5. For each successfully processed deletion, mark the
+//     revision to set low watermark for garbage collection.
+//  6. Periodically garbage collect the graveyard by finding
+//     the lowest revision of all delete trackers.
+type DB struct {
+	tables    map[TableName]TableMeta
+	mu        lock.Mutex // sequences modifications to the root tree
+	root      atomic.Pointer[iradix.Tree[tableEntry]]
+	gcTrigger chan struct{} // trigger for graveyard garbage collection
+	gcExited  chan struct{}
+}
+
+func NewDB(tables []TableMeta) (*DB, error) {
+	txn := iradix.New[tableEntry]().Txn()
+	db := &DB{
+		tables:    make(map[TableName]TableMeta),
+		gcTrigger: make(chan struct{}, 1),
+		gcExited:  make(chan struct{}),
+	}
+	for _, t := range tables {
+		name := t.Name()
+		if _, ok := db.tables[name]; ok {
+			return nil, fmt.Errorf("table %q already exists", name)
+		}
+		db.tables[name] = t
+		var table tableEntry
+		table.meta = t
+		table.deleteTrackers = iradix.New[deleteTracker]()
+		indexTxn := iradix.New[indexTree]().Txn()
+		indexTxn.Insert([]byte(t.primaryIndexer().name), iradix.New[object]())
+		indexTxn.Insert([]byte(RevisionIndex), iradix.New[object]())
+		indexTxn.Insert([]byte(GraveyardIndex), iradix.New[object]())
+		indexTxn.Insert([]byte(GraveyardRevisionIndex), iradix.New[object]())
+		for index := range t.secondaryIndexers() {
+			indexTxn.Insert([]byte(index), iradix.New[object]())
+		}
+		table.indexes = indexTxn.CommitOnly()
+		txn.Insert(t.tableKey(), table)
+	}
+	db.root.Store(txn.CommitOnly())
+
+	return db, nil
+}
+
+// ReadTxn constructs a new read transaction for performing reads against
+// a snapshot of the database.
+//
+// ReadTxn is not thread-safe!
+func (db *DB) ReadTxn() ReadTxn {
+	return &txn{
+		db:          db,
+		rootReadTxn: db.root.Load().Txn(),
+	}
+}
+
+// WriteTxn constructs a new write transaction against the given set of tables.
+// Each table is locked, which may block until the table locks are acquired.
+// The modifications performed in the write transaction are not visible outside
+// it until Commit() is called. To discard the changes call Abort().
+//
+// WriteTxn is not thread-safe!
+func (db *DB) WriteTxn(table TableMeta, tables ...TableMeta) WriteTxn {
+	allTables := append(tables, table)
+	smus := lock.SortableMutexes{}
+	for _, table := range allTables {
+		smus = append(smus, table.sortableMutex())
+	}
+	smus.Lock()
+
+	rootReadTxn := db.root.Load().Txn()
+	tableEntries := make(map[TableName]*tableEntry, len(tables))
+	for _, table := range allTables {
+		tableEntry, ok := rootReadTxn.Get(table.tableKey())
+		if !ok {
+			panic("BUG: Table '" + table.Name() + "' not found")
+		}
+		tableEntries[table.Name()] = &tableEntry
+	}
+	return &txn{
+		db:          db,
+		rootReadTxn: rootReadTxn,
+		tables:      tableEntries,
+		writeTxns:   make(map[tableIndex]*iradix.Txn[object]),
+		smus:        smus,
+	}
+}
+
+func (db *DB) Start(hive.HookContext) error {
+	go graveyardWorker(db)
+	return nil
+}
+
+func (db *DB) Stop(ctx hive.HookContext) error {
+	close(db.gcTrigger)
+	select {
+	case <-ctx.Done():
+		return errors.New("timed out waiting for graveyard worker to exit")
+	case <-db.gcExited:
+	}
+	return nil
+}
--- a/pkg/statedb2/db.go
+++ b/pkg/statedb2/db.go
@@ -6,9 +6,14 @@ package statedb2
 import (
 	"errors"
 	"fmt"
+	"reflect"
+	"runtime"
+	"strings"
 	"sync/atomic"
+	"time"
 
 	iradix "github.com/hashicorp/go-immutable-radix/v2"
+	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/cilium/cilium/pkg/hive"
 	"github.com/cilium/cilium/pkg/lock"
@@ -85,12 +90,14 @@ type DB struct {
 	root      atomic.Pointer[iradix.Tree[tableEntry]]
 	gcTrigger chan struct{} // trigger for graveyard garbage collection
 	gcExited  chan struct{}
+	metrics   Metrics
 }
 
-func NewDB(tables []TableMeta) (*DB, error) {
+func NewDB(tables []TableMeta, metrics Metrics) (*DB, error) {
 	txn := iradix.New[tableEntry]().Txn()
 	db := &DB{
-		tables: make(map[TableName]TableMeta),
+		tables:  make(map[TableName]TableMeta),
+		metrics: metrics,
 	}
 	for _, t := range tables {
 		name := t.Name()
@@ -135,28 +142,49 @@ func (db *DB) ReadTxn() ReadTxn {
 //
 // WriteTxn is not thread-safe!
 func (db *DB) WriteTxn(table TableMeta, tables ...TableMeta) WriteTxn {
+	callerPkg := callerPackage()
+
 	allTables := append(tables, table)
 	smus := lock.SortableMutexes{}
 	for _, table := range allTables {
 		smus = append(smus, table.sortableMutex())
 	}
+	lockAt := time.Now()
 	smus.Lock()
+	acquiredAt := time.Now()
 
 	rootReadTxn := db.root.Load().Txn()
 	tableEntries := make(map[TableName]*tableEntry, len(tables))
+	var tableNames []string
 	for _, table := range allTables {
 		tableEntry, ok := rootReadTxn.Get(table.tableKey())
 		if !ok {
 			panic("BUG: Table '" + table.Name() + "' not found")
 		}
 		tableEntries[table.Name()] = &tableEntry
+		tableNames = append(tableNames, table.Name())
+
+		db.metrics.TableContention.With(prometheus.Labels{
+			"table": table.Name(),
+		}).Set(table.sortableMutex().AcquireDuration().Seconds())
 	}
+
+	db.metrics.WriteTxnAcquisition.With(prometheus.Labels{
+		"package": callerPkg,
+		"tables":  strings.Join(tableNames, "+"),
+	}).Observe(acquiredAt.Sub(lockAt).Seconds())
+
 	return &txn{
-		db:          db,
-		rootReadTxn: rootReadTxn,
-		tables:      tableEntries,
-		writeTxns:   make(map[tableIndex]*iradix.Txn[object]),
-		smus:        smus,
+		db:                     db,
+		rootReadTxn:            rootReadTxn,
+		tables:                 tableEntries,
+		writeTxns:              make(map[tableIndex]*iradix.Txn[object]),
+		smus:                   smus,
+		acquiredAt:             acquiredAt,
+		tableNames:             strings.Join(tableNames, "+"),
+		packageName:            callerPkg,
+		pendingObjectDeltas:    make(map[string]float64),
+		pendingGraveyardDeltas: make(map[string]float64),
 	}
 }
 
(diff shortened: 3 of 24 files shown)


pkg/statedb/index/keyset.go
```diff
@@ -0,0 +1,61 @@
...
+	if len(ks.buf) < 2 {
+		return nil
+	}
+	length := uint16(ks.buf[0])<<8 | uint16(ks.buf[1])
+	return ks.buf[2 : 2+length]
+}
+
+func (ks *KeySet) Append(k []byte) {
+	if len(k) > 2<<16 {
+		panic("keyset.Append: key too long, maximum is 16k")
```
joestringer: Math is off here.
```suggestion
	if len(k) >= 2<<16 {
		panic("keyset.Append: key too long, maximum is 65k")
```

pkg/statedb/example/reconcile.go
```diff
@@ -0,0 +1,176 @@
...
+						r.Log.WithError(err).WithField("revision", rev).WithField("id", be.ID).Warn("Failed to insert backend")
+					}
+					return err
+				}
+			},
+		)
+
+		minRevision = newRevision
+
+		if processErr != nil {
```
tommyp1ckles: I think this aligns with how I've been thinking about health checking.  Only thing I'm wondering is if this is going to be a common pattern going forward where we can abstract out having to write the backoff logic/health reporting logic.

This is an example, but "backends" sound like the type of thing that might have a large amount of items being processed. I do see there's a rate limiter as well, even so I think health reporting shouldn't be subject to changing that much - I'm thinking we may need to come up with some kind of batching system for doing these types of health updates.

pkg/statedb/example/reconcile.go
```diff
@@ -0,0 +1,176 @@
...
+			r.Reporter.Degraded(fmt.Sprintf("Failure at revision %d, latest is %d", minRevision, tableRevision), processErr)
+			if err := backoff.Wait(ctx); err != nil {
```
tommyp1ckles: So in this case, each item has to complete processing before moving on with the next one - so this backoff would stall further reconciliation tasks?

pkg/statedb/index/keyset.go
joamaki: Thanks fixed!

pkg/statedb/example/reconcile.go
joamaki: Yep, that's the case in this example. 

I'm going to add another reconciler type which will do retrying of individual objects in the style of Resource[T]. Essentially we can do the same thing here as in Resource[T] with work queues by pushing the upsert and delete keys into the queue. Not yet sure how efficient it's going to be due to repeated creation of single-use ReadTxn and point-wise lookups. Perhaps a hybrid could work...

pkg/lock/sortable_mutex.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// sortableMutexSeq is a global sequence counter for the creation of new
+// SortableMutex's with unique sequence numbers.
+var sortableMutexSeq atomic.Uint64
+
+// SortableMutex provides a Mutex that can be globally sorted with other
+// sortable mutexes. This allows deadlock-safe locking of a set of mutexes
+// as it guarantees consistent lock ordering.
+type SortableMutex struct {
```
bimmlerd: Since that `sync.Mutex`es are usable by default, while this isn't, I think the struct should not be exported.

pkg/lock/sortable_mutex.go
```diff
@@ -0,0 +1,64 @@
...
+func (s SortableMutexes) Less(i int, j int) bool {
+	return s[i].seq < s[j].seq
+}
+
+// Swap implements sort.Interface.
+func (s SortableMutexes) Swap(i int, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+func (s SortableMutexes) Lock() {
```
bimmlerd, line 65: I think this deserves a comment to the effect of:

`Lock sorts the mutexes, and then locks them in order. If any lock cannot be acquired, this will block while holding the locks with a lower sequence number.`

pkg/statedb2/db.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+}
+
+// WriteTxn constructs a new write transaction against the given set of tables.
+// Each table is locked, which may block until the table locks are acquired.
+// The modifications performed in the write transaction are not visible outside
+// it until Commit() is called. To discard the changes call Abort().
+//
+// WriteTxn is not thread-safe!
+func (db *DB) WriteTxn(table TableMeta, tables ...TableMeta) WriteTxn {
```
bimmlerd, line 144: I think we'll need to be pretty careful about deadlocks when acquiring write transactions. Clearly, the total lock order of the tables prevents deadlocks between tables, but it's still possible for table locks to be included in a wait cycle. Here's an example:
Let's call the ipcache mutex `Mipcache`, and some table Ta's lock `MTa`. If any code path acquires `Mipcache` and then a write transaction including `Ta` (i.e. locking `MTa`), while any other path acquires `Mipcache` while holding `MTa`, we have a deadlock.
It may be prudent to prohibit acquiring _any_ external locks while holding a `WriteTxn`. (If not enforceable via some custom linter, we should at least document it).

pkg/statedb/example/reconcile.go
joamaki: Ok added that (see `pkg/statedb2/workqueue.go` and `pkg/statedb2/examples/reconcile_workqueue.go`). No health status updates for them yet.

I'm going to split this and the k8s reflector into their own PRs shortly.
EDIT: currently waiting in branch `pr/joamaki/statedb2-workqueue-and-reflector`

pkg/statedb2/db.go
joamaki: Yep that's a good idea to document! For somewhat simple cases that don't involve function pointers or interfaces we can probably catch this with a linter.

pkg/lock/sortable_mutex.go
joamaki: Thanks, added!

pkg/lock/sortable_mutex.go
joamaki: Ah good catch. I hid the struct behind an inteface, so it cannot now be initialized wrong.

pkg/statedb/example/reconcile.go
joamaki: https://github.com/joamaki/cilium/blob/pr/joamaki/statedb2-workqueue-and-reflector/pkg/statedb2/example/reconcile_workqueue.go

pkg/statedb2/db.go
```diff
@@ -0,0 +1,219 @@
...
+//     revision of the write transaction at which it was
+//     created.
+//  5. For each successfully processed deletion, mark the
+//     revision to set low watermark for garbage collection.
+//  6. Periodically garbage collect the graveyard by finding
+//     the lowest revision of all delete trackers.
+type DB struct {
+	tables    map[TableName]TableMeta
+	mu        lock.Mutex // sequences modifications to the root tree
+	root      atomic.Pointer[iradix.Tree[tableEntry]]
```
joamaki, line 90: Lookup performance can be significantly improved by switching to a flat array presentation for the tables and indexes, e.g. assign each table a position, each index a position within the table and store those positions within `genTable` and `Index`:
```
type DB struct {
  ...
  rootPtr atomic.Pointer[root]
  ... 
}

type root []tableEntry

type tableEntry struct {
  indexes []iradix.Tree[object]
  revision Revision
}

tables := db.root.Load()
index := tables[tablePos].indexes[indexPos]
```

This would be especially beneficial for "point-wise lookups" (`table.First(db.ReadTxn(), index.Query(key))`) that currently go through 3 radix trees. I'll leave this optimization for a follow-up PR for now though.

pkg/statedb2/metrics.go
```diff
@@ -0,0 +1,89 @@
...
+	// The lowest revision of a given table that has been processed by the graveyard garbage collector.
+	TableGraveyardLowWatermark metric.Vec[metric.Gauge]
+	// The time it took to clean the graveyard for a given table.
+	TableGraveyardCleaningDuration metric.Vec[metric.Observer]
+}
+
+func NewMetrics() Metrics {
+	return Metrics{
+		WriteTxnDuration: metric.NewHistogramVec(metric.HistogramOpts{
+			Namespace: metrics.CiliumAgentNamespace,
```
joamaki, line 35: The namespace should be configurable via `NewDB`... or via metrics registry?

pkg/statedb2/cell.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/cilium/cilium/pkg/hive/cell"
+)
+
+// This module provides an in-memory database built on top of immutable radix trees
+// As the database is based on an immutable data structure, the objects inserted into
+// the database MUST NOT be mutated, but rather copied first!
+//
+// For example use see pkg/statedb/example.
+var Cell = cell.Module(
+	"statedb",
```
bimmlerd, line 17: this should probably differentiate it-self more clearly from statedb

pkg/statedb2/index/int.go
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package index
+
+import "encoding/binary"
+
+func Uint64(n uint64) Key {
+	buf := make([]byte, 8)
+	binary.BigEndian.PutUint64(buf, n)
```
bimmlerd, line 10: A comment explaining why big endian lends itself better to this task would be nice

pkg/statedb2/index/keyset_test.go
```diff
@@ -0,0 +1,53 @@
...
+	ks.Foreach(func(_ []byte) {
+		t.Fatalf("Foreach on NewKeySet called function")
+	})
+	require.False(t, ks.Exists(nil))
+	require.False(t, ks.Exists([]byte{1, 2, 3}))
+
+	ks.Append([]byte("foo"))
+	require.EqualValues(t, "foo", ks.First())
+	ks.Foreach(func(bs []byte) {
+		require.EqualValues(t, "foo", ks.First())
```
bimmlerd, line 26: I think this should compare `bs`

pkg/statedb2/txn.go
```diff
@@ -0,0 +1,369 @@
...
+type lastIndexReadTxn struct {
+	table TableName
+	index IndexName
+	txn   *iradix.Txn[object]
+}
+
+var zeroTxn = txn{}
+
+func revisionKey(rev uint64, idKey []byte) []byte {
+	buf := make([]byte, 8+len(idKey))
```
bimmlerd, line 48: putting the 8 in a named constant would help with quicker understanding

pkg/statedb2/txn.go
```diff
@@ -0,0 +1,369 @@
...
+	return indexTxn
+}
+
+// indexWriteTxn returns a transaction to read/write to a specific index.
+// The created transaction is memoized and used for subsequent reads and/or writes.
+func (txn *txn) indexWriteTxn(name TableName, index IndexName) *iradix.Txn[object] {
+	if indexTreeTxn, ok := txn.writeTxns[tableIndex{name, index}]; ok {
+		return indexTreeTxn
+	}
+	table, ok := txn.tables[name]
```
bimmlerd, line 113: there's a couple of these panics based on not finding the table; I think there's `getTable` for this already

pkg/statedb2/txn.go
```diff
@@ -0,0 +1,369 @@
...
+	// Commit each individual changed index to each table.
+	// We don't notify yet (CommitOnly) as the root needs to be updated
+	// first.
+	for tableIndex, subTxn := range txn.writeTxns {
+		table, ok := txn.tables[tableIndex.table]
+		if !ok {
+			panic("BUG: Table " + tableIndex.table + " not cached")
+		}
+		table.indexes, _, _ =
+			table.indexes.Insert([]byte(tableIndex.index), subTxn.CommitOnly())
```
bimmlerd, line 326: you're explicitly ignoring the error here. I think it's non-obvious why this can't fail, so maybe an explanatory comment would help? :)

pkg/statedb2/txn.go
```diff
@@ -0,0 +1,369 @@
...
+
+	db := txn.db
+
+	// Acquire the lock on the root tree to sequence the updates to it.
+	db.mu.Lock()
+	rootTxn := db.root.Load().Txn()
+
+	// Insert the modified tables into the root.
+	for name, table := range txn.tables {
+		rootTxn.Insert([]byte(name), *table)
```
bimmlerd, line 337: remark on consistency: I think the following properties hold:

1. the critical section of updating the root is protected from concurrency by `db.mu`
2. Concurrent write transactions can only occur if the set of tables the write transactions cover is disjoint (thanks to the sortable mutexes), and hence there cannot be a conflict of writers here.
3. Read transactions can occur concurrently to a write transaction (even with overlapping tables), but will observe a consistent state (as long as the immutability of objects within the radix is upheld).

Something like this somewhere in the code would help; having the properties formulated makes it easier to think about how they can be broken

pkg/statedb2/txn.go
```diff
@@ -0,0 +1,369 @@
...
+	db.root.Store(newRoot)
+	db.mu.Unlock()
+
+	// Now that new root is available notify of the changes by closing the watch channels.
+	for _, subTxn := range txn.writeTxns {
+		subTxn.Notify()
+	}
+	rootTxn.Notify()
+
+	txn.smus.Unlock()
```
bimmlerd, line 351: do we need to keep the table locks before we notify? (I think this came up in the code walk too)

pkg/statedb2/index/keyset.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (ks KeySet) Foreach(fn func(Key)) {
+	for len(ks.buf) >= 2 {
+		length := uint16(ks.buf[0])<<8 | uint16(ks.buf[1])
+		fn(append([]byte(nil), ks.buf[2:2+length]...))
+		ks.buf = ks.buf[2+length:]
+	}
+}
+
+func (ks KeySet) Exists(k Key) bool {
```
bimmlerd, line 51: I think the bit banging is correct, but in general the keyset deserves some comments, maybe even some ascii art :innocent:

pkg/statedb2/db.go
```diff
@@ -90,9 +90,7 @@ type DB struct {
 func NewDB(tables []TableMeta) (*DB, error) {
 	txn := iradix.New[tableEntry]().Txn()
 	db := &DB{
-		tables:    make(map[TableName]TableMeta),
-		gcTrigger: make(chan struct{}, 1),
-		gcExited:  make(chan struct{}),
+		tables: make(map[TableName]TableMeta),
```
bimmlerd: these changes might want to fixedup into the initial commit?

pkg/lock/sortable_mutex.go
```diff
@@ -16,18 +17,28 @@ var sortableMutexSeq atomic.Uint64
...
+
+func (s *sortableMutex) AcquireDuration() time.Duration { return s.acquireDuration }
 
 // SortableMutex provides a Mutex that can be globally sorted with other
 // sortable mutexes. This allows deadlock-safe locking of a set of mutexes
 // as it guarantees consistent lock ordering.
 type SortableMutex interface {
 	sync.Locker
 	Seq() uint64
+	AcquireDuration() time.Duration // The amount of time it took to acquire the lock
```
bimmlerd, line 41: maybe `cumulative amount of time` in this case?

bimmlerd (approved):
it's a lot of code to read, so couldn't review it all thoroughly- mostly skipped the metrics, don't know much about those anyway.

Other than that, impressive feat of engineering.

pkg/lock/sortable_mutex.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// sortableMutexSeq is a global sequence counter for the creation of new
+// SortableMutex's with unique sequence numbers.
+var sortableMutexSeq atomic.Uint64
+
+// sortableMutex implements SortableMutex. Not exported as the only way to
+// initialize it is via NewSortableMutex().
+type sortableMutex struct {
+	*sync.Mutex
```
tommyp1ckles: we're not going to use our internal mutex type here?

pkg/statedb2/txn.go
```diff
@@ -0,0 +1,369 @@
...
+	table, ok := txn.tables[tableName]
+	if !ok {
+		return 0, fmt.Errorf("table %q not locked for writing", tableName)
+	}
+	table.revision++
+	return table.revision, nil
+}
+
+func (txn *txn) Insert(meta TableMeta, data any) (any, bool, error) {
+	if txn.rootReadTxn == nil {
```
tommyp1ckles, line 140: Where does this ever get set to nil?

pkg/statedb2/table.go
```diff
@@ -0,0 +1,214 @@
...
+	indexTxn := txn.getTxn().indexReadTxn(t.table, q.index)
+	if indexTxn == nil {
+		panic("BUG: Missing index " + q.index)
+	}
+	iter := indexTxn.Root().Iterator()
+	watchCh := iter.SeekPrefixWatch(q.key)
+	return &iterator[Obj]{iter}, watchCh
+}
+
+// Insert implements Table
```
tommyp1ckles, line 166: unfinished comment?

pkg/statedb2/index/keyset.go
```diff
@@ -0,0 +1,62 @@
...
+
+package index
+
+import "bytes"
+
+type Key = []byte
+
+// KeySet is a sequence of (length, byte slice) pairs.
+// length is encoded as 16-bit big-endian unsigned int.
+type KeySet struct {
```
tommyp1ckles, line 12: why does this need to be in a specific byte order instead of just an int type?

pkg/statedb2/table.go
```diff
@@ -0,0 +1,214 @@
...
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/cilium/cilium/pkg/lock"
+	"github.com/cilium/cilium/pkg/statedb2/index"
+)
+
+func NewTable[Obj any](
+	tableName TableName,
+	primaryIndexer Indexer[Obj],
```
tommyp1ckles, line 18: is it obvious that the primary indexer only ever use the first key value in the set?

tommyp1ckles (commented):
Few questions - might take me another pass or two before I understand this enough to make meaningful comments.

pkg/statedb2/cell.go
joamaki: Right yeah, since they'll live side-by-side for a few days, I'll rename this to statedb2.

pkg/statedb2/txn.go
joamaki: We don't, we can drop them and then notify.

pkg/statedb2/table.go
joamaki: Ah this was auto-filled by golsp. I'll clean these up.

pkg/statedb2/index/keyset.go
joamaki: sure, I might as well make it native 32-bit int and use `binary.NativeEndian`. I'm still wondering though if I'm prematurely optimizing here and `[][]byte` would be just fine. I wanted to keep memory allocations to somewhat minimum on inserts...

pkg/statedb2/table.go
joamaki: Primary index must not be a multi-index which is why only first key is used. I'll think about whether it'd be worth the extra code to split into SingleIndexer and MultiIndexer to make this explicit or see if I can figure out some other ways, perhaps allowing specifying either `FromObject func(Obj) index.Key` or `FromObjectMulti func(Obj) index.KeySet`, but not both.

pkg/statedb2/txn.go
joamaki: Abort and Commit set it to mark the txn as finished. I'll think about ways to clean this up a bit.

pkg/lock/sortable_mutex.go
joamaki: Ah good catch. I'll switch it to `lock.Mutex`.

pkg/statedb2/txn.go
joamaki: What error? Insert cannot fail, it returns `(newTree Tree, oldObject T, oldExisted bool)`

pkg/statedb2/db.go
```diff
@@ -0,0 +1,233 @@
...
+		tableNames = append(tableNames, table.Name())
+
+		db.metrics.TableContention.With(prometheus.Labels{
+			"table": table.Name(),
+		}).Set(table.sortableMutex().AcquireDuration().Seconds())
+	}
+
+	db.metrics.WriteTxnAcquisition.With(prometheus.Labels{
+		"package": callerPkg,
+		"tables":  strings.Join(tableNames, "+"),
```
dylandreimerink, line 174: We (I) forgot to sort the table names. If we don't do that then metrics for write transactions for `table1, table2` and `table2, table1` will end up in separate buckets.

pkg/statedb2/db.go
```diff
@@ -0,0 +1,233 @@
...
+
+func (db *DB) Stop(ctx hive.HookContext) error {
+	close(db.gcTrigger)
+	select {
+	case <-ctx.Done():
+		return errors.New("timed out waiting for graveyard worker to exit")
+	case <-db.gcExited:
+	}
+	return nil
+}
```
dylandreimerink, line 206: nit: not strictly necessary, but we could use a timer job + trigger here instead of managing the lifecycle of this goroutine manually.

pkg/statedb2/graveyard.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+			continue
+		}
+
+		// Dead objects found, do a write transaction against all tables with dead objects in them.
+		tablesToModify := maps.Keys(toBeDeleted)
+		txn = db.WriteTxn(tablesToModify[0], tablesToModify[1:]...).getTxn()
+		for meta, deadObjs := range toBeDeleted {
+			tableName := meta.Name()
+			cleaningTimes[tableName].Start()
```
dylandreimerink, line 93: I made another mistake here :disappointed: . I thought that running `.Start` on a timestat would accumulate, but it just resets the start timer. So this doesn't work as intended, we have to add a second set of timestats and combine them when reporting the CleaningDuration below

pkg/statedb2/metrics.go
dylandreimerink: The registry should set the namespace of very collected metric.

pkg/statedb2/txn.go
```diff
@@ -0,0 +1,400 @@
...
+	return indexTreeTxn
+}
+
+func (txn *txn) newRevision(tableName TableName) (Revision, error) {
+	table, ok := txn.tables[tableName]
+	if !ok {
+		return 0, fmt.Errorf("table %q not locked for writing", tableName)
+	}
+	table.revision++
+	txn.db.metrics.TableRevision.With(prometheus.Labels{
```
dylandreimerink, line 133: Uhm :thinking:, if we `Abort` this transaction, is the revision still incremented? Or are changes to the revision also discarded? If they are we should also switch to tracing a revision delta and only applying to the metric once we `Commit`.

pkg/statedb2/txn.go
```diff
@@ -0,0 +1,400 @@
...
+	db.mu.Unlock()
+
+	// Now that new root is available notify of the changes by closing the watch channels.
+	for _, subTxn := range txn.writeTxns {
+		subTxn.Notify()
+	}
+	rootTxn.Notify()
+
+	txn.smus.Unlock()
+	for name, delta := range txn.pendingObjectDeltas {
```
dylandreimerink, line 352: We proccess the pending object deltas here, as we should, but I forgot to add the same logic for the `pendingGraveyardDeltas` :face_exhaling:

pkg/statedb2/txn.go
```diff
@@ -0,0 +1,400 @@
...
+		}
+		root := indexTxn.Root()
+		iter := root.Iterator()
+
+		buf.WriteString("  \"" + table.Name() + "\": [\n")
+
+		_, obj, ok := iter.Next()
+		for ok {
+			buf.WriteString("    ")
+			bs, err := json.Marshal(obj.data)
```
dylandreimerink, line 384: If we are going to use reflection based Marshalling anyway, way do the whole manual JSON crafting instead of using it for the whole JSON message? Is this premature optimization, and if so, would another JSON library not be a better idea?

dylandreimerink (dismissed):
Nice work, though I spotted a number of mistakes I made myself in the metrics part

pkg/lock/sortable_mutex.go
```diff
@@ -16,18 +17,28 @@ var sortableMutexSeq atomic.Uint64
 // sortableMutex implements SortableMutex. Not exported as the only way to
 // initialize it is via NewSortableMutex().
 type sortableMutex struct {
-	*sync.Mutex
-	seq uint64
+	sync.Mutex
+	seq             uint64
+	acquireDuration time.Duration
```
giorio94, line 22: Nit. I guess that most (all?) changes in this file should belong to ff6d098eb54774db75288680be49f29675fd0562.

pkg/statedb2/types.go
```diff
@@ -0,0 +1,219 @@
...
+	// been committed or aborted.
+	//
+	// If the table is being tracked for deletions via DeleteTracker()
+	// the deleted object is inserted into a graveyard index and garbage
+	// collected when all delete trackers have consumed it. Each deleted
+	// object in the graveyard has unique revision allowing interleaved
+	// iteration of updates and deletions (see (*DeleteTracker[Obj]).Process).
+	Delete(WriteTxn, Obj) (oldObj Obj, hadOld bool, err error)
+
+	DeleteAll(WriteTxn) error
```
giorio94, line 80: Nit. The comment for `DeleteAll` is missing.

pkg/statedb2/types.go
```diff
@@ -0,0 +1,219 @@
...
+	Name() TableName                          // The name of the table
+	tableKey() index.Key                      // The radix key for the table in the root tree
+	primaryIndexer() anyIndexer               // The untyped primary indexer for the table
+	secondaryIndexers() map[string]anyIndexer // Secondary indexers (if any)
+	sortableMutex() lock.SortableMutex        // The sortable mutex for locking the table for writing
+}
+
+// Iterator for iterating objects returned from queries.
+type Iterator[Obj any] interface {
+	// Next returns the next object its revision if ok is true, otherwise
```
giorio94, line 99: Nit.

```diff
- // Next returns the next object its revision if ok is true, otherwise
+ // Next returns the next object and its revision if ok is true, otherwise
```

pkg/statedb2/types.go
```diff
@@ -0,0 +1,219 @@
...
+type Index[Obj any, Key any] struct {
+	Name       string
+	FromObject func(obj Obj) index.KeySet
+	FromKey    func(key Key) []byte
+	Unique     bool
+}
+
+var _ Indexer[struct{}] = &Index[struct{}, bool]{}
+
+//nolint:unused
```
giorio94, line 142: Nit. I'd suggest to expand the comment to mention that nolint is required because the linter doesn't understand that these methods are used to implement a generic interface.

pkg/statedb2/types.go
```diff
@@ -0,0 +1,219 @@
...
+	// 'Obj' type. Provides the database access to table's indexers.
+	TableMeta
+
+	// Revision of the table. Constant for a read transaction, but
+	// increments in a write transaction on each Insert and Delete.
+	Revision(ReadTxn) Revision
+
+	// All returns an iterator for all objects in the table and a watch
+	// channel that is closed when the table changes.
+	All(ReadTxn) (Iterator[Obj], <-chan struct{})
```
giorio94, line 32: Nit. I wonder if it could make sense to define a dedicated alias type instead of using `<-chan struct{}` to signal that a change has occurred.

pkg/statedb2/db.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	db.root.Store(txn.CommitOnly())
+
+	return db, nil
+}
+
+// ReadTxn constructs a new read transaction for performing reads against
+// a snapshot of the database.
+//
+// ReadTxn is not thread-safe!
```
giorio94, line 130: I think that it would be worth expanding the comment here (and for WriteTxn) about thread safeness. For instance, in the example there are two `WriteTxn` operations that could possibly occur in parallel, which seems to contradict what's written here.

pkg/statedb2/deletetracker.go
```diff
@@ -0,0 +1,111 @@
...
+
+}
+
+// Process is a helper to iterate updates and deletes to a table in revision order.
+//
+// The 'processFn' is called for each updated or deleted object in order. If an error
+// is returned by the function the iteration is stopped and the revision at which
+// processing failed and the error is returned. The caller can then retry processing
+// again from this revision by providing it as the 'minRevision'.
+func (dt *DeleteTracker[Obj]) Process(txn ReadTxn, minRevision Revision, processFn func(obj Obj, deleted bool, rev Revision) error) (Revision, <-chan struct{}, error) {
```
giorio94, line 80: Nit. It seems a bit strange to me that both updates and deletes are watched through a `DeleteTracker`. I wonder if it could make sense to have a dedicated interface with a more semantic name for that.

giorio94 (commented):
Looks really cool! I still have to fully digest all the logic, but I left a few minor comments inline.

pkg/statedb2/example/reconcile.go
```diff
@@ -0,0 +1,176 @@
...
+
+		// Wait until something changes or we're being stopped.
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-watch:
+		}
+
+		// Refresh the read transaction to read the new version of the state.
+		txn = r.DB.ReadTxn()
```
tommyp1ckles, line 132: Do we need to close the old transaction here?

pkg/statedb2/metrics.go
```diff
@@ -0,0 +1,89 @@
...
+		WriteTxnAcquisition: metric.NewHistogramVec(metric.HistogramOpts{
+			Namespace: metrics.CiliumAgentNamespace,
+			Subsystem: "statedb",
+			Name:      "write_txn_acquisition_seconds",
+			Help:      "How long it took to acquire a write transaction for all tables.",
+		}, []string{"tables", "package"}),
+		TableContention: metric.NewGaugeVec(metric.GaugeOpts{
+			Namespace: metrics.CiliumAgentNamespace,
+			Subsystem: "statedb",
+			Name:      "table_contention_seconds",
```
tommyp1ckles, line 49: Should this be `table_contention_duration_seconds`, similar to other metrics?

pkg/statedb2/metrics.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/cilium/cilium/pkg/metrics/metric"
+)
+
+type Metrics struct {
+	// How long a read transaction was held.
+	WriteTxnDuration metric.Vec[metric.Observer]
+	// How long it took to acquire a write transaction for all tables.
+	WriteTxnAcquisition metric.Vec[metric.Observer]
+	// How long writers were blocked while waiting to acquire a write transaction for a specific table.
+	TableContention metric.Vec[metric.Gauge]
```
tommyp1ckles, line 17: These are basically measuring db lock acquire latency, should we be using histograms for this?

pkg/statedb2/metrics.go
```diff
@@ -0,0 +1,89 @@
...
+	TableDeleteTrackerCount metric.Vec[metric.Gauge]
+	// The amount of objects in the graveyard for a given table.
+	TableGraveyardObjects metric.Vec[metric.Gauge]
+	// The lowest revision of a given table that has been processed by the graveyard garbage collector.
+	TableGraveyardLowWatermark metric.Vec[metric.Gauge]
+	// The time it took to clean the graveyard for a given table.
+	TableGraveyardCleaningDuration metric.Vec[metric.Observer]
+}
+
+func NewMetrics() Metrics {
```
tommyp1ckles, line 32: These metrics being all labelled by table (as well as possibly by package) would seem to have a pretty high metrics cardinality (up to: nodes X tables X package). 

Should these be enabled by default? and perhaps we should have a way to emit these on aggregate without tables for more general usage (otherwise these seem like they may get disabled and be largely used for developer debugging purposes)

pkg/statedb2/table.go
tommyp1ckles: documenting this behaviour might be sufficient

pkg/statedb2/metrics.go
```diff
@@ -0,0 +1,89 @@
...
+	return Metrics{
+		WriteTxnDuration: metric.NewHistogramVec(metric.HistogramOpts{
+			Namespace: metrics.CiliumAgentNamespace,
+			Subsystem: "statedb",
+			Name:      "write_txn_duration_seconds",
+			Help:      "How long a write transaction was held.",
+		}, []string{"tables", "package"}),
```
tommyp1ckles, line 39: Having a package label seems especially a bit worrying since there's a large set of values this could be.

pkg/statedb2/db.go
```diff
@@ -0,0 +1,233 @@
...
+//
+//	           root
+//	          /    \
+//	         ba    T(foo)
+//	       /   \
+//	   T(bar)  T(baz)
+//
+//	      T(bar).indexes
+//		   /  \
+//		  i    I(byRevision)
```
tommyp1ckles, line 42: I get why we store tables in the trie, why do we need to store indices in one as well?

pkg/statedb2/example/reconcile.go
dylandreimerink: Read transactions are like snapshots, they do not need to be closed. Only write transactions need to be Committed or Aborted.

pkg/statedb2/metrics.go
dylandreimerink: We already have a histogram above it which measures how long it takes for all locks in a transaction to be acquired, which should cover most information. The table contention metric is just slightly more specific since it is per table, using a gauge was an attempt to keep the amount of metric data down. This is just 1 value per table, histograms have lots of values per label.

pkg/statedb2/metrics.go
dylandreimerink: Yea, I see what you are saying, its a concern. A lot of these metrics are to track things that we as developers are worried about. The idea being that if something goes wrong and we get a sysdump that we can hopefully see if certain things are going wrong in statedb. 

But metrics might not be the correct tool for the job, or at least not in this way. Perhaps we should look into a setting where metrics are part of the CLI output and thus the sysdump and the prometheus server.

pkg/statedb2/metrics.go
dylandreimerink: The idea is that this can help identify a culprit if we ever run into high table contention situations. But using metrics as a debugging tool like this might not be the right way to go about it.

pkg/statedb2/db.go
dylandreimerink: We store the actual data in immutable radix tries which allow us to do read snapshots. But since we have multiple indecies and multiple tables which should all be part of the same read snapshot, we place the smaller tries in larger tries, until we have 1 single root node for the whole StateDB so we can get a transaction with a single atomic pointer read and update with a single atomic pointer write.

joamaki: I also said in an earlier comment that this can be optimized a lot by just using arrays for tables and indexes. I will follow up with that in a later PR.

joamaki:
This was merged to expedite the removal of old statedb and the go-memdb dependency. We will follow up with the reviews.

pkg/statedb2/metrics.go
tommyp1ckles: I think these might be nice to have, but maybe not enabled by default... which ultimately may make it hard to catch high table contention situations.

pkg/statedb2/metrics.go
tommyp1ckles: I think these metrics may be potentially useful to some users.  It may also help us reveal scaling issues in the future.

This is kind of part of a broader problem we have that I've been thinking about.  We find ourselves in a position where we walk a line between offering metrics that are high cardinality and providing ones that can be enabled by default, or we make compromises (i.e. see map_pressure metric).

Not sure what the answer is here, I think in this case we may just want to think about what we enable by default.

pkg/statedb2/metrics.go
dylandreimerink: Addressed part of the concern in https://github.com/cilium/cilium/pull/27657

pkg/statedb2/metrics.go
dylandreimerink: Would love your feedback on https://github.com/cilium/cilium/issues/27659

PR #27163, opened 2023-07-31
repository: shard rules by namespace

odinuge:
This central ruleSlice has for us been one of the main bottle necks for throughput, and one of the main contributors to cpu usage. With this patch, we will now shard it by namespace transparently, so that we don't need to evaluate rules unnecessary. The main benefit is that when calculating rules for pod X in namespace A, we don't need to evaluate rules only applying to pods in namespace B. The same applies to the rule add and remove; when a rule is removed in namespace B, we don't need to evaluate rules in namespace A.

This is all transparent to users, and an implementation detail, so the sharding mechanism can easily be changed at a later stage.

**TBD**: Still some stuff missing here, so mostly a proof-of-concept.

Based on the BenchmarkParseLabel with 10 namespaces/shards, this is the benchstat diff;
```
name          old time/op    new time/op    delta
ParseLabel-7     5.21s ±33%     1.02s ±40%  -80.36%  (p=0.000 n=10+10)

name          old alloc/op   new alloc/op   delta
ParseLabel-7    26.9MB ± 0%    26.9MB ± 0%   -0.33%  (p=0.001 n=9+10)

name          old allocs/op  new allocs/op  delta
ParseLabel-7      520k ± 0%      520k ± 0%   +0.02%  (p=0.000 n=9+9)
```

The more rules, the higher bigger the gain will be! We also see that the locking `(*Repository) getMatchingRules` is also a huge chuck of cpu time, as well as a lot of lock contention. 

```release-note
Improve performance of Network Policy calculation
```

### DIFF
--- a/pkg/policy/repository.go
+++ b/pkg/policy/repository.go
@@ -659,6 +659,24 @@ func (p *Repository) GetRulesMatching(lbls labels.LabelArray) (ingressMatch bool
 	return
 }
 
+func getNamespaceFromIdentity(securityIdentity *identity.Identity) string {
+	for _, lbl := range securityIdentity.LabelArray {
+		if lbl.Key == k8sConst.PodNamespaceLabel && lbl.Source == labels.LabelSourceK8s {
+			return lbl.Value
+		}
+	}
+	return ""
+}
+
+func isInit(securityIdentity *identity.Identity) bool {
+	for _, lbl := range securityIdentity.LabelArray {
+		if lbl.Source == labels.LabelSourceReserved && lbl.Key == labels.IDNameInit {
+			return true
+		}
+	}
+	return false
+}
+
 // getMatchingRules returns whether any of the rules in a repository contain a
 // rule with labels matching the given security identity, as well as
 // a slice of all rules which match.
@@ -669,11 +687,16 @@ func (p *Repository) getMatchingRules(securityIdentity *identity.Identity) (
 	matchingRules ruleSlice) {
 
 	matchingRules = []*rule{}
-	key, _ := getShardKey(securityIdentity.LabelArray)
+	key := getNamespaceFromIdentity(securityIdentity)
 	relevantShards := map[string]ruleSlice{
 		key: p.rulesByShard[key],
 		"":  p.rulesByShard[""],
 	}
+
+	if isInit(securityIdentity) {
+		relevantShards = p.rulesByShard
+	}
+
 	for _, rr := range relevantShards {
 		for _, r := range rr {
 			isNode := securityIdentity.ID == identity.ReservedIdentityHost
--- a/pkg/policy/repository.go
+++ b/pkg/policy/repository.go
@@ -141,16 +141,67 @@ type Repository struct {
 	getEnvoyHTTPRules func(certificatemanager.SecretManager, *api.L7Rules, string) (*cilium.HttpNetworkPolicyRules, bool)
 }
 
-// getShardKey returns the key to be used for the sharing of Repository.rulesByShard.
-// We currently use the K8s namespace label as the key, and if its not present, we fall back
+// getShardKey returns the key to be used for the sharding of Repository.rulesByShard.
+// We currently use the K8s namespace label as the key, and if it's not present, we fall back
 // to an empty string. Returns a true if it found a namespace label, and false if not.
 func getShardKey(lbls labels.LabelArray) (string, bool) {
+	key := ""
+	matched := false
 	for _, lbl := range lbls {
 		if lbl.Key == k8sConst.PolicyLabelNamespace && lbl.Source == labels.LabelSourceK8s {
-			return lbl.Value, true
+			// If we find two, ignore then and put them into the non-namespaced for easier housekeeping
+			if matched {
+				return "", false
+			}
+			key = lbl.Value
+			matched = true
 		}
 	}
-	return "", false
+	return key, matched
+}
+
+// getShardKeyFromRule finds the correct shard key to ensure that housekeeping works
+// as expected. A rule will get a shard key that is not the empty string iff
+//   - it has a single namespace label in its api.Rule.Labels
+//   - and has a single namespace match in its endpoint selector that match the namespace from the label.
+//
+// Otherwise, we always default to empty string.
+//
+// This means that when searching for rules that apply on an endpoint based on labels, one always has to search the
+// shard for the empty string, as well as the one for the namespace found in the label set or EndpointSelector.
+//
+// On the opposite side, when searching for rules that has a set of labels x, one has to search through all shards.
+// Only if the set of labels has a single namespace label, one can search through the empty string and the shard for
+// the given namespace
+func getShardKeyFromRule(r api.Rule) string {
+	if r.EndpointSelector.LabelSelector == nil {
+		return ""
+	}
+
+	if matches, found := r.EndpointSelector.GetMatch(labels.LabelSourceK8sKeyPrefix + k8sConst.PodNamespaceLabel); found {
+		key, ok := getShardKey(r.Labels)
+		if ok && len(matches) == 1 && matches[0] == key {
+			return key
+		}
+	}
+	return ""
+}
+
+// getNamespaceFromLabels returns the namespace found in the labels iff there is a single
+// namespace value. If zero or more than one is found, it returns an empty string.
+func getNamespaceFromLabels(lbls labels.LabelArray) string {
+	key := ""
+	matched := false
+	for _, lbl := range lbls {
+		if lbl.Key == k8sConst.PodNamespaceLabel && lbl.Source == labels.LabelSourceK8s {
+			if matched {
+				return ""
+			}
+			key = lbl.Value
+			matched = true
+		}
+	}
+	return key
 }
 
 // GetSelectorCache() returns the selector cache used by the Repository
@@ -406,6 +457,7 @@ func (p *Repository) SearchRLocked(lbls labels.LabelArray) api.Rules {
 	if ok {
 		relevantShards = map[string]ruleSlice{
 			key: p.rulesByShard[key],
+			"":  p.rulesByShard[""],
 		}
 	}
 
@@ -449,9 +501,9 @@ func (p *Repository) AddListLocked(rules api.Rules) (ruleSlice, uint64) {
 			Rule:     *rules[i],
 			metadata: newRuleMetadata(),
 		}
-		key, _ := getShardKey(rules[i].Labels)
 		newList = append(newList, newRule)
-		p.rulesByShard[key] = append(p.rulesByShard[key], newRule)
+		shard := getShardKeyFromRule(newRule.Rule)
+		p.rulesByShard[shard] = append(p.rulesByShard[shard], newRule)
 	}
 
 	p.BumpRevision()
@@ -552,11 +604,13 @@ func (p *Repository) DeleteByLabelsLocked(lbls labels.LabelArray) (ruleSlice, ui
 	deleted := 0
 	var deletedRules ruleSlice
 
-	key, ok := getShardKey(lbls)
 	relevantShards := p.rulesByShard
+
+	key, ok := getShardKey(lbls)
 	if ok {
 		relevantShards = map[string]ruleSlice{
 			key: p.rulesByShard[key],
+			"":  p.rulesByShard[""],
 		}
 	}
 	for shard := range relevantShards {
@@ -627,7 +681,7 @@ func (p *Repository) GetRulesMatching(lbls labels.LabelArray) (ingressMatch bool
 	ingressMatch = false
 	egressMatch = false
 
-	key, _ := getShardKey(lbls)
+	key := getNamespaceFromLabels(lbls)
 	relevantShards := map[string]ruleSlice{
 		key: p.rulesByShard[key],
 		"":  p.rulesByShard[""],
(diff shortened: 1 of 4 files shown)


odinuge:
cc @christarazi.

I'll do some more testing, but would be nice with some feedback on the overall idea. I think this should reduce our cilium-agent cpu load quite significantly, and it should also reduce the lock contention quite significantly as well.

pkg/policy/repository.go
```diff
@@ -139,6 +141,18 @@ type Repository struct {
 	getEnvoyHTTPRules func(certificatemanager.SecretManager, *api.L7Rules, string) (*cilium.HttpNetworkPolicyRules, bool)
 }
 
+// getShardKey returns the key to be used for the sharing of Repository.rulesByShard.
```
christarazi, line 144:
```suggestion
// getShardKey returns the key to be used for the sharding of Repository.rulesByShard.
```

Suble typo or did you mean `sharing`? If the latter, then somehow the sentence could be clarified as it will potentially confuse other readers if said typo was made. :)

christarazi:
```suggestion
// getShardKey returns the key to be used for the sharding of Repository.rulesByShard.
```

Suble typo or did you mean `sharing`? If the latter, then somehow the sentence could be clarified as it will potentially confuse other readers if said typo was made. :)

pkg/policy/repository.go
```diff
@@ -139,6 +141,18 @@ type Repository struct {
...
+// We currently use the K8s namespace label as the key, and if its not present, we fall back
+// to an empty string. Returns a true if it found a namespace label, and false if not.
+func getShardKey(lbls labels.LabelArray) (string, bool) {
+	for _, lbl := range lbls {
+		if lbl.Key == k8sConst.PolicyLabelNamespace && lbl.Source == labels.LabelSourceK8s {
+			return lbl.Value, true
```
christarazi: AFAICT, this returns the label value, but the function is called `getShardKey`. So the result returned from 

```
k8s:foo=bar
k8s:baz=net
```

these labels would be `bar` and `net`, depending on the order of the iteration.

Is the the expected behavior? Wouldn't this cause a problem for [1]?

pkg/policy/repository.go
```diff
@@ -367,9 +401,19 @@ func (p *Repository) AllowsEgressRLocked(ctx *SearchContext) api.Decision {
...
-	for _, r := range p.rules {
-		if r.Labels.Contains(lbls) {
-			result = append(result, &r.Rule)
+	key, ok := getShardKey(lbls)
+	relevantShards := p.rulesByShard
+	if ok {
+		relevantShards = map[string]ruleSlice{
+			key: p.rulesByShard[key],
+		}
+	}
```
christarazi: [1]

christarazi (commented):
Thanks for the PR, this is a great idea! 

A few comments on initial approach.

pkg/policy/repository.go
odinuge: Heh, yeah, `sharding` it should be!

pkg/policy/repository.go
```diff
@@ -399,16 +443,17 @@ func (p *Repository) Add(r api.Rule, localRuleConsumers []Endpoint) (uint64, map
...
 
-	newList := make(ruleSlice, len(rules))
+	var newList ruleSlice
 	for i := range rules {
 		newRule := &rule{
 			Rule:     *rules[i],
 			metadata: newRuleMetadata(),
 		}
-		newList[i] = newRule
+		key, _ := getShardKey(rules[i].Labels)
```
joestringer: I think that this imposes a requirement on the user to populate the `Labels` field of each rule with the namespace that the rule applies in, but at the same time the `rules[i].EndpointSelector` probably already has this information. Did you consider deriving the shard/namespace from the `EndpointSelector` instead?

One caveat is that CCNP rules don't have an `EndpointSelector`, they have a `NodeSelector` instead, and they won't have a namespace, but we can come up with a heuristic for that case.

pkg/policy/repository.go
```diff
@@ -109,7 +109,9 @@ func (p *policyContext) SetDeny(deny bool) bool {
 type Repository struct {
 	// Mutex protects the whole policy tree
 	Mutex lock.RWMutex
-	rules ruleSlice
+	// Internally we shard the rule list to improve performance and maximize throughput
```
joestringer, line 112: I think that this is specifically targeted for policy update performance, is that correct? Also, what's the difference between improving update performance vs. maximizing throughput? Aren't they the same thing?

```suggestion
	// Internally we shard the rule list to improve update performance
```

pkg/policy/repository.go
odinuge: Good point!

I _think_ it should be fine given it doesn't really make sense having more than one label with `k8s:io.cilium.k8s.policy.namespace=<xyz>`. It is set based on the namespace here;
https://github.com/cilium/cilium/blob/6ba173d9ba392e05ff58930b482d697e00c70805/pkg/k8s/apis/cilium.io/utils/utils.go#L345-L346 (and the same for the k8s version of `NetworkPolicy`). For `CCNP` its not present, and we can fallback to `""`.

But maybe the users can manually set the `spec.Labels` field manually with same source and key, resulting in two of them. I'll look at that case, and I guess we have to take a look at it and handle it properly if thats the case. 😄  The same applies to non-namespaced CCNPs then, if users can manually add a label to those "tricking" us into thinking its a namespaced policy while its not. This should not as I understand it affect what the policy applies to, just how its stored in this policy "repo".

pkg/policy/repository.go
```diff
@@ -562,28 +626,55 @@ func (p *Repository) GetJSON() string {
...
+			return lbl.Value
 		}
 	}
-	return
+	return ""
+}
+
+func isInit(securityIdentity *identity.Identity) bool {
+	for _, lbl := range securityIdentity.LabelArray {
+		if lbl.Source == labels.LabelSourceReserved && lbl.Key == labels.IDNameInit {
```
odinuge: Need to verify that this is ineed the correct logic.

pkg/policy/repository.go
joestringer: Ah, based on the changes in `getMatchingRules()` it looks like this should also improve policy evaluation performance by excluding rule iteration for namespaces that are known to be different from the "current" namespace under iteration.

So maybe 'update and evaluation performance'?

joestringer (commented):
Neat proposal! Couple of brief comments from initial look.

Also, the danger here would be if there could be a bug for some corner case of rule format that somehow ends up not getting selected by the rule engine and hence the policy doesn't apply. I don't have a good grasp on the corner cases but it'd be nice to see some testing to evaluate those cases. One basic one would be to continue to ensure that CCNPs that select all namespaces will apply to all Pods.

pkg/policy/repository.go
christarazi: Ah yeah it's a strange edge case. Maybe we can just document that if there are ever >1 namespace label, then the one that appears first will be used.

pkg/policy/repository.go
odinuge: Hi,

> I think that this imposes a requirement on the user to populate the Labels field of each rule with the namespace that the rule applies in, [...]

This is piggy backing on the label injected here; https://github.com/cilium/cilium/blob/6ba173d9ba392e05ff58930b482d697e00c70805/pkg/k8s/apis/cilium.io/utils/utils.go#L345-L346. The name label is captured here https://docs.cilium.io/en/stable/security/policy/intro/#rule-basics, so its ~kinda unknown if the UID+namespace label is part of the API or not.

> Did you consider deriving the shard/namespace from the EndpointSelector instead?

Yeah, i did consider it, but there seem to be some edge cases where only looking at the endpoint selector is not sufficient (without other menas). So it seems like the most sane solution is to actually look at both; but _maybe_ we can figure out a way without doing that.

 
> One caveat is that CCNP rules don't have an EndpointSelector, they have a NodeSelector instead, and they won't have a namespace, but we can come up with a heuristic for that case.

I personally think that keeping those in the empty string basket (`""`) is sufficient for now; and tackle that later if we want or have to. I think adding another heuristic will make the logic more complex and brittle. Does that make sense?

pkg/policy/repository.go
odinuge: Yeah, this stated as an experiment where I wanted to improve the policy addition/update/deletion flow, but yeah, it will also reduce the policy evaluation as well! I'll try updating the description when we all agree on the approach and implementation. 😄

odinuge:
> Neat proposal! Couple of brief comments from initial look.
>
> Also, the danger here would be if there could be a bug for some corner case of rule format that somehow ends up not getting selected by the rule engine and hence the policy doesn't apply. I don't have a good grasp on the corner cases but it'd be nice to see some testing to evaluate those cases. One basic one would be to continue to ensure that CCNPs that select all namespaces will apply to all Pods.

Thanks for looking! And thanks for the initial feedback (both @joestringer and @christarazi)!

> I don't have a good grasp on the corner cases but it'd be nice to see some testing to evaluate those cases.

Yeah, this is my main concern now; so I really want to poke at all those. My main concern is related to https://github.com/cilium/cilium/pull/27163#discussion_r1279817838 where users can add extra "namespace" labels to the `spec.labels` part of a CNP or CCNP, making it harder to reason about how those work. I'll push a new version now, but this is pretty far from production ready.

pkg/policy/repository.go
joestringer: Ah, if I understand correctly I was missing the context around GetPolicyLabels() and actually these labels are already configured automatically by Cilium, so no need for users to opt in to this functionality?

This sounds like a fine solution then. I can't think of any caveats that might impose the need for taking EndpointSelector input then.

I agree regarding CCNP, we can always further optimize that specific problem if it becomes an issue.

joestringer:
SGTM. Also when we've narrowed down on the user value it'll be worth updating the release note to say something about improving policy processing performance.

pkg/policy/repository.go
odinuge: > Ah, if I understand correctly I was missing the context around GetPolicyLabels() and actually these labels are already configured automatically by Cilium, so no need for users to opt in to this functionality?

yeah, cilium will add this. The plot twist is that users can add similar lables with either the same or a different namespace, or even add them on CCNPs; so we need to ensure we deal with that properly as well. Will probably require a huge set of extra tests

pkg/policy/repository.go
joestringer: If Cilium prepends the namespace label and we use the first one, then there may be no need to document it at all?

On one hand the threat actor who could set these labels is a fairly privileged [Kubernetes API server Attacker](https://docs.cilium.io/en/stable/security/threat-model/#kubernetes-api-server-attacker) and if they could customize the labels of the policy then they already have significant privileges to allow actual traffic. But on the other hand if they only have access to policies in one namespace, they shouldn't be able to "poison" policy shards for other namespaces, even if the result is only some form of additional policy cost (a cost which is currently borne by all users in all namespaces in current versions of Cilium)

pkg/policy/repository.go
```diff
@@ -139,6 +141,69 @@ type Repository struct {
...
+		if lbl.Key == k8sConst.PolicyLabelNamespace && lbl.Source == labels.LabelSourceK8s {
+			// If we find two, ignore then and put them into the non-namespaced for easier housekeeping
+			if matched {
+				return "", false
+			}
+			key = lbl.Value
+			matched = true
+		}
+	}
+	return key, matched
```
schlosna, line 160: The logic in `getNamespaceFromLabels` and `getShardKey` seems duplicative, might be worth implementing one in terms of the other.


```suggestion
	if key := getNamespaceFromLabels(lbls); key != "" {
		return key, true
	}
	return "", false
```

odinuge: Sadly those two functions use different labels; since there is one label that is present in endpoint selectors (`io.kubernetes.pod.namespace`), while there is another one in the `rule.Labels` field (`io.cilium.k8s.policy.namespace`).

schlosna: I was thinking we could reuse the k8s label search logic by adjusting the function definition for reuse. The refactor I had in mind is a bit more than fits in github suggestion, so I created https://github.com/odinuge/cilium/pull/94

schlosna:
> This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

I would still like to see this PR move forward

joestringer:
Hey @odinuge, was this waiting on further feedback or discussion? Or you just haven't had time to revisit? I think that in principle this can improve Cilium quite a bit, so it would be neat to see it move forward.

squeed:
@odinuge I'm picking this up to try and get it merged for v1.16.

odinuge:
Hey @squeed! Happy to push this myself. However we need a proper solution to #30338 first, and we need _a lot_ of tests for that to avoid security regressions. If you want to start with that first that would be very appreciated!

odinuge:
The easy alternative is reverting https://github.com/cilium/cilium/commit/4cb094060d5a6b93313e222516c644ef5c51da07, but not too sure how happy people would be with that. Imo. we should not introduce more "indexes" here until we are 100% the current state handles all the edge cases, with tests ensuring that is the case.

squeed:
> we should not introduce more "indexes"

This seems like a good chance to convert the PolicyRepository to StateDB over the coming months.

joestringer:
When converting more critical components over to StateDB, we'll need to exercise a lot of care and constraint. I'm open to the idea, but we should be rolling StateDB out in incremental manners through the tree that start non-critical components and move towards critical components, so we can ensure correctness and properly understand the performance characteristics.

squeed:
Also, @odinuge, it looks like what this PR does is improve the time to insert / replace rules in the repository, and doesn't particularly improve endpoint validation. Which is fine, but the description is misleading ;-).

From your tests, which is more critical? Rule addition / deletion time, or endpoint policy calculation?

squeed:
I filed #32703, which was inspired by this but takes a different tack.

squeed:
https://github.com/cilium/cilium/pull/32703 should have removed this bottleneck, I believe this can be closed, @odinuge.

PR #27170, opened 2023-07-31
Add WireGuard to the firewall rules documentation

joestringer:
- docs: Word-wrap firewall section
- docs: Add WireGuard into the firewall rules

### DIFF
--- a/Documentation/operations/system_requirements.rst
+++ b/Documentation/operations/system_requirements.rst
@@ -332,6 +332,11 @@ For IPSec enabled Cilium deployments, you need to ensure that the firewall
 allows ESP traffic through. For example, AWS Security Groups doesn't allow ESP
 traffic by default.
 
+If you are using WireGuard, you must allow UDP port 51871. Furthermore, if you
+have disabled node-to-node encryption and configured an overlay network mode
+(such as VXLAN or Geneve) in addition to WireGuard, then the overlay ports must
+also be allowed.
+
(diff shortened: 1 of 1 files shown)


brb (approved):
Thanks!

gentoo-root (approved):

learnitall (dismissed):
Nice catch, thank you for adding this! I think there are a couple of changes we could make to add more context to this.

* If WireGuard is enabled with tunnel mode, do users need to open firewall rules for VXLAN/Geneve and WireGuard, or just WireGuard, since packets bypass the VXLAN/Geneve tunnels when WireGuard is enabled? If users only need to open a port for WireGuard, can we add a statement saying so?
* If users need to open ports for WireGuard and VXLAN/Geneve, can we move the information for WireGuard into its own own paragraph, similar to IPSec?

joestringer:
@brb can you help to answer the above questions?

brb:
It's a bit tricky, as it depends on the encryption mode.

* With `encryption.nodeEncryption=true`, all traffic flows through the WireGuard tunnel device completely bypassing the VXLAN/Geneve tunnels.
* With `encryption.nodeEncryption=false`, the pod-to-pod traffic will bypass the VXLAN/Geneve tunnels, but the host-to-pod flows through the VXLAN/Geneve tunnels.

TL;DR with the node encryption enabled, there is no need to open ports for the VXLAN/Geneve tunnels.

learnitall (approved):
Looks great! Thanks!

PR #27180, opened 2023-08-01
datapath: configure linux node config writer through hive

giorio94:
Currently, the logic in charge of writing the header file containing the local node configuration is entirely defined as part of a single function, which has grown significantly in size, and depends on a large set of global variables. Overall, this leads to additional friction when extending it with new options, especially as it does not play well with the recently introduced hive framework.

This PR implements a first step to address this issue, enabling the possibility of providing through hive a set of extra functions returning subsets of configurations, which are then merged to form the final header file. Overall, this allows to break down the generation process in smaller chunks, with each module defining its own specific settings (e.g., based on user configurations and internal state), reducing at the same time the dependency on global variables.

```release-note
Configure the linux node config writer through Hive
```

### DIFF
--- a/pkg/datapath/linux/config/config.go
+++ b/pkg/datapath/linux/config/config.go
@@ -24,6 +24,7 @@ import (
 	"github.com/cilium/cilium/pkg/bpf"
 	"github.com/cilium/cilium/pkg/byteorder"
 	"github.com/cilium/cilium/pkg/datapath/link"
+	dpdef "github.com/cilium/cilium/pkg/datapath/linux/config/defines"
 	datapath "github.com/cilium/cilium/pkg/datapath/types"
 	"github.com/cilium/cilium/pkg/defaults"
 	"github.com/cilium/cilium/pkg/identity"
@@ -75,7 +76,9 @@ var (
 
 // HeaderfileWriter is a wrapper type which implements datapath.ConfigWriter.
 // It manages writing of configuration of datapath program headerfiles.
-type HeaderfileWriter struct{}
+type HeaderfileWriter struct {
+	nodeExtraDefines []dpdef.Fn
+}
 
 func writeIncludes(w io.Writer) (int, error) {
 	return fmt.Fprintf(w, "#include \"lib/utils.h\"\n\n")
@@ -83,8 +86,8 @@ func writeIncludes(w io.Writer) (int, error) {
 
 // WriteNodeConfig writes the local node configuration to the specified writer.
 func (h *HeaderfileWriter) WriteNodeConfig(w io.Writer, cfg *datapath.LocalNodeConfiguration) error {
-	extraMacrosMap := make(map[string]string)
-	cDefinesMap := make(map[string]string)
+	extraMacrosMap := make(dpdef.Map)
+	cDefinesMap := make(dpdef.Map)
 
 	fw := bufio.NewWriter(w)
 
@@ -719,6 +722,21 @@ func (h *HeaderfileWriter) WriteNodeConfig(w io.Writer, cfg *datapath.LocalNodeC
 		return err
 	}
 	cDefinesMap["EPHEMERAL_MIN"] = fmt.Sprintf("%d", ephemeralMin)
+
+	for _, fn := range h.nodeExtraDefines {
+		defines, err := fn()
+		if err != nil {
+			return err
+		}
+
+		for key, value := range defines {
+			if _, ok := cDefinesMap[key]; ok {
+				return fmt.Errorf("extra node define overwrites key %q", key)
+			}
+
+			cDefinesMap[key] = value
+		}
+	}
 
 	// Since golang maps are unordered, we sort the keys in the map
 	// to get a consistent written format to the writer. This maintains
--- a/pkg/datapath/linux/config/config_test.go
+++ b/pkg/datapath/linux/config/config_test.go
@@ -13,10 +13,12 @@ import (
 	"testing"
 
 	. "github.com/cilium/checkmate"
+	"github.com/stretchr/testify/require"
 	"github.com/vishvananda/netlink"
 
 	"github.com/cilium/ebpf/rlimit"
 
+	dpdef "github.com/cilium/cilium/pkg/datapath/linux/config/defines"
 	"github.com/cilium/cilium/pkg/datapath/loader"
 	datapath "github.com/cilium/cilium/pkg/datapath/types"
 	"github.com/cilium/cilium/pkg/maps/ctmap"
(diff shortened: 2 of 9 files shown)


pkg/datapath/linux/config/config.go
```diff
@@ -720,6 +730,17 @@ func (h *HeaderfileWriter) WriteNodeConfig(w io.Writer, cfg *datapath.LocalNodeC
...
 	cDefinesMap["EPHEMERAL_MIN"] = fmt.Sprintf("%d", ephemeralMin)
 
+	for _, fn := range h.nodeExtraDefines {
+		defines, err := fn()
+		if err != nil {
+			return err
+		}
+
+		for key, value := range defines {
+			cDefinesMap[key] = value
```
lmb, line 737: Can we return an error here if `key` already exists? It probably doesn't make sense to allow overwriting defines.

pkg/datapath/linux/config/config_test.go
```diff
@@ -98,6 +98,22 @@ func (s *ConfigSuite) TestWriteNodeConfig(c *C) {
 	})
 }
 
+func (s *ConfigSuite) TestWriteNodeConfigExtraDefines(c *C) {
+	cfg := HeaderfileWriter{nodeExtraDefines: []HeaderNodeDefinesFn{
+		func() (HeaderNodeDefinesMap, error) { return HeaderNodeDefinesMap{"FOO": "0x1", "BAR": "0x2"}, nil },
```
lmb: Can you add a test which returns an error?

pkg/datapath/linux/config/config.go
```diff
@@ -73,18 +73,28 @@ var (
 	}
 )
 
+// HeaderNodeDefinesMap is the type containing the key-value pairs representing
+// extra define directives for datapath node configuration.
+type HeaderNodeDefinesMap map[string]string
+
+// HeaderNodeDefinesFn is a function returning the key-value pairs representing
+// extra define directives for datapath node configuration.
+type HeaderNodeDefinesFn func() (HeaderNodeDefinesMap, error)
```
lmb: Assuming this is a pure function, why do we need the `error` return here? I imagine that whatever provides the `header-node-defines` group will just error out if it can't create the correct headers?

Somewhat related, could `NodeExtraDefines` just be `[]HeaderNodeDefinesMap` instead?

lmb (commented):
It'd be nice to have an example of how you want to use this in the PR itself. Makes it easier to discuss API and such.

pkg/datapath/linux/config/config.go
giorio94: The idea of returning a function rather than the materialized defines here was to support values not known at provisioning time, and that could change over time. For instance, different configs are currently configured accessing some local node property through the node global variable (and the associated accessor functions), which has not yet been modularized, and is initialized during daemon initialization. Even if the local node object was modularized, though, we would still need to return a function so that we materialize the corresponding values only once the corresponding start hook has completed. Given that the function may rely on external values, it might fail, which is the reason why it also returns a possible error value.

giorio94:
> It'd be nice to have an example of how you want to use this in the PR itself. Makes it easier to discuss API and such.

Makes sense. Let me see if I can craft an example without requiring a too extensive refactoring, which would be out of scope for this PR.

giorio94:
> > It'd be nice to have an example of how you want to use this in the PR itself. Makes it easier to discuss API and such.
> 
> Makes sense. Let me see if I can craft an example without requiring a too extensive refactoring, which would be out of scope for this PR.

I've prepared a [couple of examples](https://github.com/cilium/cilium/compare/4a7910134b42f5a0be1df002e486d14bc31053dc...3681fbce2943f56a86128cff802d663739b7ea27), one targeting the L2 announcer (extra configs depend on user configs only), and the other the wireguard subsystem (extra defines depend also on external operations). While not complete (as for instance additional refactorings would be required to get rid of the global options), I think that they should provide a complete usage example to discuss about the current API and implementation.

pkg/datapath/linux/config/config.go
lmb: > values not known at provisioning time, and that could change over time

Are there values that currently change over time? If not, I'm not sure we should add these via the templating mechanism. Changing the config dynamically through `node_config.h` will trigger a full recompilation. Rewriting constants in the ELF is much cheaper.
cc @ti-mo

pkg/datapath/linux/config/config_test.go
```diff
@@ -98,6 +98,41 @@ func (s *ConfigSuite) TestWriteNodeConfig(c *C) {
 	})
 }
 
+func (s *ConfigSuite) TestWriteNodeConfigExtraDefines(c *C) {
```
lmb: Nit: seems like you could just write this using `testing` instead of `gocheck`?

lmb (commented):
Thanks for providing the examples! I'm still unsure about dynamically changing defines, so I'm pulling in Timo for a second opinion from the loader side of things.

pkg/datapath/linux/config/config.go
giorio94: > Are there values that currently change over time?

Honestly I'm not sure about this. I saw that the config contains parameters such as IP addresses and the device MTU, which could theoretically change (I don't know though whether there's any logic to retrigger the generation, or an explicit agent restart is required for a change to take effect). That's why I wrote `[...] that could change over time`, but it was mostly a supposition. And I don't have plans to add any new config which would require that. 

I still think that the `values not known at provisioning time` part applies though, because we cannot retrieve and materialize parameters such as the Node IP from a provide function, but we need to defer that to a start hook (which is where the provided functions are executed with the current implementation). Unless we also rewrite these constants in the ELF, but that would likely only move the problem to a different place.

ti-mo: node_config.h is rewritten when the host datapath gets regenerated (see current usage of `WriteNodeConfig()` through `Loader.Reinitialize()`). It is also called from `loader.hashDatapath()`, so we need to make sure the output of `WriteNodeConfig()` stays stable throughout multiple calls during agent init, otherwise things may break in surprising ways. Do the functions in `h.nodeExtraDefines` block until their data is available? Or should they error out in that case?

The goal is to eventually remove our runtime dependency on clang and plumbing in all values at load time using constant rewriting, no longer relying on writing out node_config.h at all.

giorio94: > Do the functions in h.nodeExtraDefines block until their data is available? Or should they error out in that case?

The idea of `h.nodeExtraDefines` is that it behaves similarly to the global header generation function, that is when it gets called all the dependencies have already been resolved (for instance by start hooks) and can be directly accessed (e.g., local node information).

lmb (approved):

lmb:
I think this will cause frequent circular dependencies:

```
$ go test -c ./pkg/datapath/linux/config
# github.com/cilium/cilium/pkg/datapath/linux/config
package github.com/cilium/cilium/pkg/datapath/linux/config
	imports github.com/cilium/cilium/pkg/maps/egressmap
	imports github.com/cilium/cilium/pkg/datapath/alignchecker
	imports github.com/cilium/cilium/pkg/datapath/loader: import cycle not allowed in test
FAIL	github.com/cilium/cilium/pkg/datapath/linux/config [setup failed]
```

This happened when I tried to convert `egressmap` over. Maybe the `Out` structure needs to move to `datapath/types`? Not sure.

giorio94:
Last push shortened the type definitions based on @lmb's suggestion.

lmb:
@cilium/endpoint could someone please take a look at this PR? It's been waiting for review for a week now.

christarazi (approved):

tommyp1ckles (approved):

PR #27184, opened 2023-08-01
docs: Document `DROP_NO_NODE_ID` for IPsec

pchaigno:
This commit documents the new drop reason introduced for IPsec in https://github.com/cilium/cilium/pull/27029.

### DIFF
--- a/Documentation/security/network/encryption-ipsec.rst
+++ b/Documentation/security/network/encryption-ipsec.rst
@@ -267,6 +267,16 @@ Troubleshooting
                             allocate memory.
    =======================  ==================================================
 
+ * In addition to the above XFRM errors, packet drops of type ``No node ID
+   found`` (code 197) may also occur under normal operations. These drops can
+   happen if a pod attempts to send traffic to a pod on a new node for which
+   the Cilium agent didn't yet receive the CiliumNode object. It can also
+   happen if the IP address of the destination node changed and the agent
+   didn't receive the updated CiliumNode object yet. In both cases, the IPsec
+   configuration in the kernel isn't ready yet, so Cilium drops the packets at
+   the source. These drops will stop once the CiliumNode information is
+   propagated across the cluster.
+
 Disabling Encryption
 ====================
 


Documentation/security/network/encryption-ipsec.rst
```diff
@@ -267,6 +267,15 @@ Troubleshooting
...
    =======================  ==================================================
 
+ * In addition to the above XFRM errors, packet drops of type ``No node ID
+   found`` (code 197) may also occur under normal operations. These drops can
+   happen if a pod attempts to send traffic to a pod on a new node for which
+   the Cilium agent didn't yet receive the CiliumNode object. It can also
+   happen if the IP address of the destination node changed and the agent
+   didn't receive the updated CiliumNode object yet. In both cases, the IPsec
+   configuration in the kernel isn't ready yet, so Cilium drops the packets at
+   the source.
```
joestringer: I wonder if it's worth adding a final sentence giving some frame of reference for the expected recovery time from this state?

If the same traffic persistently hits this error message, it would be useful for users to be able to identify that this is actually an unexpected condition and report a bug that Cilium is not getting back out of this state. Or identify that the kube-apiserver is struggling and hence is unable to propagate the information to nodes in the cluster.

```suggestion
   the source. These drops will stop once the CiliumNode information is
   propagated across the cluster.
```

It might even be worth providing tips as to how you can tell if a node is synced or not, although I think in general we don't expect this to occur for longer than maybe a few seconds unless the cluster is really big(?).

pchaigno: I considered it, but wasn't sure what exactly to add. It's not clear to me what exactly can cause those out-of-order events (CiliumNode received before CiliumEndpoint) to happen more frequently or for longer.

I think your proposed sentence is good. It documents what we understand and can indeed be useful to users. I've added it.

jschwinger233 (approved):

joestringer (approved):

PR #27188, opened 2023-08-01
node: Tests should fail

pchaigno:
Check that the inverted test fails as expected.


pchaigno:
Fixed by Nicolas in https://github.com/cilium/cilium/pull/27193.

PR #26952, opened 2023-07-20
CODEOWNERS: assign egressgw control plane/datapath logic to egress-gateway team

jibi:

PR #27189, opened 2023-08-01
bpf: gate egressgw datapath on separate defines

lmb:
bpf: gate egressgw datapath on separate defines

    Add a new define ENABLE_EGRESS_GATEWAY_COMMON which enables the common
    egress gateway code. The existing ENABLE_EGRESS_GATEWAY flag now controls
    whether to check for egress gateway policies with static gateway IPs. This
    makes it easier to extend the codebase to with alternative implementations.

bpf: move CT checks into egress_gw_request_needs_redirect

    Move the CT reply check into egress_gw_request_needs_redirect so that
    alternative implementations can have their own CT state handling.

### DIFF
--- a/bpf/bpf_lxc.c
+++ b/bpf/bpf_lxc.c
@@ -1075,15 +1075,7 @@ ct_recreate4:
 		if (identity_is_cluster(*dst_sec_identity))
 			goto skip_egress_gateway;
 
-		/* If the packet is a reply or is related, it means that outside
-		 * has initiated the connection, and so we should skip egress
-		 * gateway, since an egress policy is only matching connections
-		 * originating from a pod.
-		 */
-		if (ct_status == CT_REPLY || ct_status == CT_RELATED)
-			goto skip_egress_gateway;
-
-		if (egress_gw_request_needs_redirect(ip4, &tunnel_endpoint)) {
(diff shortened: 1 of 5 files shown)


lmb:
Needs a rebase on top of #26952 since that already moves the code to a separate file.

bpf/bpf_lxc.c
```diff
@@ -1065,7 +1065,7 @@ static __always_inline int handle_ipv4_from_lxc(struct __ctx_buff *ctx, __u32 *d
 		}
 	}
 
-#ifdef ENABLE_EGRESS_GATEWAY
+#ifdef ENABLE_EGRESS_GATEWAY_COMMON
```
jibi, line 1068: nit: let's just switch all of them to `#if defined()`?

bpf/lib/egress_gateway.h
```diff
@@ -8,7 +8,11 @@
 
 #include "maps.h"
 
-#ifdef ENABLE_EGRESS_GATEWAY
+#if defined(ENABLE_EGRESS_GATEWAY)
+#define ENABLE_EGRESS_GATEWAY_COMMON
+#endif
+
+#ifdef ENABLE_EGRESS_GATEWAY_COMMON
```
jibi, line 15: ditto

jibi (approved):
LGTM! just a non blocking nit

PR #27192, opened 2023-08-01
srv6: Add constructor and CRUD helpers for SID map

ldelossa:
This commit adds some quality of life methods.

First is a constructor which creates a key for the SID map from a `net.IP`.

Second we add update and delete methods for the map to create a namespace for CRUD operations on the map.

```release-note
SRv6: Add quality of life methods for SID map usage. 
```

### DIFF
--- a/pkg/maps/srv6map/sid.go
+++ b/pkg/maps/srv6map/sid.go
@@ -4,6 +4,8 @@
 package srv6map
 
 import (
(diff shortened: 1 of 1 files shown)


YutaroHayakawa (approved):

gentoo-root (approved):
Are you planning to use these for something?

PR #27193, opened 2023-08-01
ci: fix and standardize checkouts in privileged workflows

nbusseneau:
Please see per commit. This PR does not touch upon non-privileged workflows (those that run on `pull_request` triggers), we might want to address them in a separate PR just to clean them up but there are no security concerns or specific issues to address on these so it's not urgent.

### DIFF
--- a/.github/workflows/build-images-base.yaml
+++ b/.github/workflows/build-images-base.yaml
@@ -27,7 +27,7 @@ jobs:
     environment: release-base-images
     runs-on: ubuntu-20.04
     steps:
-      - name: Checkout main branch to access local actions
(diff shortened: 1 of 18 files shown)


brb (approved):

brlbil (approved):

nbusseneau:
Note to backporter: 2nd commit is "just" tab aligning steps, so if you see big conflicts just feel free to drop the changes and manually re-align.

joestringer:
cc @nbusseneau there were conflicts for 1.12, so the PR got stuck in pending state. I set `backport/author`.

PR #27199, opened 2023-08-01
doc: update preferred ways to install cilium on aks

tamilmani1989:

      GitHub issue.

This PR updates preferred way to install cilium on AKS to include "Azure CNI Powered by Cilium" option in addition to byocni. Also added brief comparison between those options which would help AKS customers to decide based on their requirements.

Fixes: #issue-number

```release-note
```

### DIFF
--- a/Documentation/installation/requirements-aks.rst
+++ b/Documentation/installation/requirements-aks.rst
@@ -7,8 +7,14 @@ BYOCNI (``none``)             Encapsulation   Cluster Pool        Kubernetes CRD
 Legacy Azure IPAM (``azure``) Direct Routing  Azure IPAM          Kubernetes CRD
 ============================= =============== =================== ==============
 
-Using `Bring your own CNI <https://docs.microsoft.com/en-us/azure/aks/use-byo-cni?tabs=azure-cli>`_
-is the preferred way to run Cilium on `Azure Kubernetes Service (AKS) <https://docs.microsoft.com/en-us/azure/aks/>`_,
-however integration with the Azure stack via the :ref:`Azure IPAM<ipam_azure>`
-is not available and will only work with clusters not using BYOCNI. While still
-maintained for now, this mode is considered legacy.
+The preferred way to run Cilium on `Azure Kubernetes Service (AKS) <https://docs.microsoft.com/en-us/azure/aks/>`_ is 
+either `Bring your own CNI <https://docs.microsoft.com/en-us/azure/aks/use-byo-cni?tabs=azure-cli>`_ or 
+`Azure CNI Powered by Cilium <https://learn.microsoft.com/en-us/azure/aks/azure-cni-powered-by-cilium>`_. 
+Choosing between these options depends on your specific requirements and use case. 
+If you need full control over Cilium and its features, then Bring your own CNI approach might be preferred. 
+On the other hand, if you prefer a managed solution with simplified installation and 
+integration with Azure Virtual Network stack, then Azure CNI Powered by Cilium option might be suitable.
+
+Another way to install Cilium on AKS is through chaining with Azure CNI :ref:`chaining_azure` and 
+can be integrated with Azure Stack via :ref:`Azure IPAM<ipam_azure>`. However this integration will not work for clusters using BYOCNI. 
+While it is still maintained for now, it is considered a legacy mode.
--- a/Documentation/installation/requirements-aks.rst
+++ b/Documentation/installation/requirements-aks.rst
@@ -10,10 +10,15 @@ Legacy Azure IPAM (``azure``) Direct Routing  Azure IPAM          Kubernetes CRD
 The preferred way to run Cilium on `Azure Kubernetes Service (AKS) <https://docs.microsoft.com/en-us/azure/aks/>`_ is 
 either `Bring your own CNI <https://docs.microsoft.com/en-us/azure/aks/use-byo-cni?tabs=azure-cli>`_ or 
 `Azure CNI Powered by Cilium <https://learn.microsoft.com/en-us/azure/aks/azure-cni-powered-by-cilium>`_. 
-Choosing between these options depends on your specific requirements and use case. 
-If you need full control over Cilium and its features, then Bring your own CNI approach might be preferred. 
-On the other hand, if you prefer a managed solution with simplified installation and 
-integration with Azure Virtual Network stack, then Azure CNI Powered by Cilium option might be suitable.
+There are different benefits to each option.
+
+**Azure CNI powered by Cilium:**
+- Installation and upgrades are provided by AKS.
+- Cilium is configured with delegated Azure IPAM, so Pods receive IP addresses that are routable in the Azure Virtual Network Stack.
+
+**Bring your own CNI:**
+- You can customize the Cilium installation with a wider range of features.
+- Azure IPAM is not available in this mode.
 
 Another way to install Cilium on AKS is through chaining with Azure CNI :ref:`chaining_azure` and 
 can be integrated with Azure Stack via :ref:`Azure IPAM<ipam_azure>`. However this integration will not work for clusters using BYOCNI. 
--- a/Documentation/network/concepts/ipam/azure.rst
+++ b/Documentation/network/concepts/ipam/azure.rst
@@ -12,9 +12,11 @@ Azure IPAM
 
 .. note::
 
+   The preferred way to run Cilium on `Azure Kubernetes Service (AKS) <https://docs.microsoft.com/en-us/azure/aks/>`_ is
+   either `Bring your own CNI <https://docs.microsoft.com/en-us/azure/aks/use-byo-cni?tabs=azure-cli>`_ or 
+   `Azure CNI Powered by Cilium <https://learn.microsoft.com/en-us/azure/aks/azure-cni-powered-by-cilium>`_.
    While still maintained for now, Azure IPAM is considered legacy and is not
-   compatible with AKS clusters created in `Bring your own CNI <https://docs.microsoft.com/en-us/azure/aks/use-byo-cni?tabs=azure-cli>`_
-   mode. Using BYOCNI is the preferred way to install Cilium on AKS.
+   compatible with AKS clusters created in Bring your own CNI mode.
 
 The Azure IPAM allocator is specific to Cilium deployments running in the Azure
 cloud and performs IP allocation based on `Azure Private IP addresses


tamilmani1989:
cc: @nbusseneau @joestringer

Documentation/installation/requirements-aks.rst
```diff
@@ -7,8 +7,14 @@ BYOCNI (``none``)             Encapsulation   Cluster Pool        Kubernetes CRD
...
-however integration with the Azure stack via the :ref:`Azure IPAM<ipam_azure>`
-is not available and will only work with clusters not using BYOCNI. While still
-maintained for now, this mode is considered legacy.
+The preferred way to run Cilium on `Azure Kubernetes Service (AKS) <https://docs.microsoft.com/en-us/azure/aks/>`_ is 
+either `Bring your own CNI <https://docs.microsoft.com/en-us/azure/aks/use-byo-cni?tabs=azure-cli>`_ or 
+`Azure CNI Powered by Cilium <https://learn.microsoft.com/en-us/azure/aks/azure-cni-powered-by-cilium>`_. 
+Choosing between these options depends on your specific requirements and use case. 
+If you need full control over Cilium and its features, then Bring your own CNI approach might be preferred. 
+On the other hand, if you prefer a managed solution with simplified installation and 
+integration with Azure Virtual Network stack, then Azure CNI Powered by Cilium option might be suitable.
```
joestringer: Out of curiosity, what does "integration with Azure Virtual Network stack" mean? Is this referring to native IPAM from Azure?

joestringer: The wording here feels a bit conversational, I'm wondering if we can tighten it a bit.

One approach, less diff vs. current but it still seems a little bit loose, I'd prefer to still iterate further:

```suggestion
Choosing between these options depends on your specific requirements and use case. 
If you need full control over Cilium and its features, use Bring your own CNI. 
On the other hand, if you prefer a managed solution with simplified installation and 
integration with Azure Virtual Network stack, then use Azure CNI Powered by Cilium.
```

More of a factual approach:

```suggestion
There are different benefits to each option.

**Azure CNI powered by Cilium:**
- Installation and upgrades are provided by AKS.
- Cilium is configured with delegated Azure IPAM, so Pods receive IP addresses that are routable in the Azure Virtual Network Stack.

**Bring your own CNI:**
- You can customize the Cilium installation with a wider range of features.
- Azure IPAM is not available in this mode.
```

Documentation/installation/requirements-aks.rst
```diff
@@ -7,8 +7,14 @@ BYOCNI (``none``)             Encapsulation   Cluster Pool        Kubernetes CRD
 Legacy Azure IPAM (``azure``) Direct Routing  Azure IPAM          Kubernetes CRD
 ============================= =============== =================== ==============
```
joestringer, line 8: I'm not quite sure what "Default configuration" means above but we should probably remove legacy Azure IPAM and add APBC in this table?

Documentation/installation/requirements-aks.rst
```diff
@@ -7,8 +7,14 @@ BYOCNI (``none``)             Encapsulation   Cluster Pool        Kubernetes CRD
...
 
-Using `Bring your own CNI <https://docs.microsoft.com/en-us/azure/aks/use-byo-cni?tabs=azure-cli>`_
-is the preferred way to run Cilium on `Azure Kubernetes Service (AKS) <https://docs.microsoft.com/en-us/azure/aks/>`_,
-however integration with the Azure stack via the :ref:`Azure IPAM<ipam_azure>`
-is not available and will only work with clusters not using BYOCNI. While still
-maintained for now, this mode is considered legacy.
+The preferred way to run Cilium on `Azure Kubernetes Service (AKS) <https://docs.microsoft.com/en-us/azure/aks/>`_ is
```
joestringer, line 10: I'm also wondering if we should just drop the "preferred" step here and walk this back to just a statement that "AKS provides two methods to integrate Cilium CNI" and point to the two, then explain about the benefits in a minimal comparative way.

Documentation/installation/requirements-aks.rst
```diff
@@ -7,8 +7,14 @@ BYOCNI (``none``)             Encapsulation   Cluster Pool        Kubernetes CRD
...
+either `Bring your own CNI <https://docs.microsoft.com/en-us/azure/aks/use-byo-cni?tabs=azure-cli>`_ or 
+`Azure CNI Powered by Cilium <https://learn.microsoft.com/en-us/azure/aks/azure-cni-powered-by-cilium>`_. 
+Choosing between these options depends on your specific requirements and use case. 
+If you need full control over Cilium and its features, then Bring your own CNI approach might be preferred. 
+On the other hand, if you prefer a managed solution with simplified installation and 
+integration with Azure Virtual Network stack, then Azure CNI Powered by Cilium option might be suitable.
+
+Another way to install Cilium on AKS is through chaining with Azure CNI :ref:`chaining_azure` and 
+can be integrated with Azure Stack via :ref:`Azure IPAM<ipam_azure>`. However this integration will not work for clusters using BYOCNI. 
+While it is still maintained for now, it is considered a legacy mode.
```
joestringer, line 25: I'd be inclined to drop this part. Users who need this already know about it and are using it, and they can track down the specific docs page for this information. I don't think it's particularly valuable to point new users towards these modes.

joestringer (commented):
I've added some initial thoughts below, maybe that also helps @nbusseneau provide some feedback.

I know that with Azure CNI powered by Cilium there are also options for Enterprise upgrades via vendors from https://cilium.io/enterprise which can provide some of the benefits of BYOCNI, but I'm not necessarily sure how to integrate those into the docs currently so I've just left that feedback out for now.

Documentation/installation/requirements-aks.rst
tamilmani1989: yes that's right.

Documentation/installation/requirements-aks.rst
tamilmani1989: will wait for @nbusseneau response as well

nbusseneau (changes requested):
I agree with Joe in general. Rather than discuss the current diff, here is a high-level overview of what I think we should do:

- Remove advertisement of Azure IPAM in installation guides, and make sure to mark Azure IPAM as Legacy everywhere it still makes sense to have it referenced.
- Rework the installation guides to present BYOCNI and Azure CNI Powered by Cilium as the 2 options available for installing Cilium on an AKS cluster. However, since the latter is managed by Microsoft, and users can only interact with it (e.g. via Cilium CLI) but cannot fully control it, it should be documented as an External Installer.
- Rework the IPAM plugins list with a new entry for Azure CNI Powered by Cilium's delegated IPAM.

It's not a trivial modification in my opinion, apologies if this is way more than what you intended to implement 😅 I'm happy to try and set aside time to do it myself if need be.

In details, this would translate roughly to the following:

- Add a new Azure CNI Powered by Cilium entry to External Installers https://docs.cilium.io/en/latest/installation/external-toc/
  - This new page should be focused on pointing users to the appropriate Microsoft documentation and stick to only important commands (such as cluster creation with `--network-dataplane cilium`), as we do for other External Installers.
  - It should then contain a section on how can users interact with and leverage an Azure CNI Powered by Cilium cluster via the Cilium CLI, as we do for other External Installers.
  - It might contain a section dedicated to how the delegated IPAM works, or failing that it should redirect to a new dedicated page for the delegated IPAM (more details below when discussing IPAM plugins).
- Getting Started guide https://docs.cilium.io/en/latest/gettingstarted/k8s-install-default/
  - Rework the AKS tabs so that only instructions for BYOCNI remain. Completely remove Azure IPAM.
  - Add an introductory note pointing to the Azure CNI Powered by Cilium External Installer guide. The note should contain a brief high-level discussion of both options. I was thinking of something like this:
    > On AKS, Cilium can be installed either manually by administrators (Bring your own CNI) or automatically by AKS (Azure CNI Powered by Cilium). Bring your own CNI offers more flexibility and customization as administrators have full control over the installation, but does not integrate natively with the Azure network stack and administrators need to handle Cilium upgrades. Azure CNI Powered by Cilium integrates natively with the Azure network stack and upgrades are handled by AKS, but does not offer as much flexibility and customization as it is controlled by AKS. The following instructions assume Bring your own CNI, please refer to the Azure CNI Powered by Cilium external installer guide for dedicated instructions.
- Helm guide https://docs.cilium.io/en/latest/installation/k8s-install-helm/
  - Do the same as for Getting Started.
- IPAM guides https://docs.cilium.io/en/latest/network/concepts/ipam/
  - Rename `Azure IPAM` to `Azure IPAM (Legacy)`.
  - Add new entry for Azure CNI Powered by Cilium's delegated IPAM. This entry might either redirect to a new dedicated page, or just to the same new External Installer page for Azure CNI Powered by Cilium but pointing to a sub-section dedicated to how the delegated IPAM works (cf. note above).
- Azure IPAM https://docs.cilium.io/en/latest/network/concepts/ipam/azure/
  - Rename page to `Azure IPAM (Legacy)`.
  - Rework the note with more context around why it's legacy and point to the options we recommend instead.
- CNI chaining guides https://docs.cilium.io/en/latest/installation/cni-chaining/
  - Rename `Azure CNI` to `Azure CNI (Legacy)`.
- Azure CNI chaining  https://docs.cilium.io/en/latest/installation/cni-chaining-azure-cni/
  - Rename page to `Azure CNI (Legacy)`.
  - Rework the note in a similar fashion to the Azure IPAM dedicated page, making it explicit that Azure CNI chaining is legacy because Azure IPAM is legacy, and point to the options we recommend instead.
- NodePort XDP on Azure https://docs.cilium.io/en/latest/network/kubernetes/kubeproxy-free/#nodeport-xdp-on-azure
  - Not sure what to do about this but probably at least add a note that it's legacy because Azure IPAM is legacy.

joestringer:
Wow, that is a detailed proposal and I agree with all of it. For each bulletpoint there it could potentially be a different task in a different PR and maybe even different people working on it, depending on whether @tamilmani1989 would be interested in tackling all of those items or only some of them. Let me know if it would be helpful to file a tracking issue to list those individual tasks so we can better distribute the work across different folks, or whether you would be happy to address each of these items.

> - NodePort XDP on Azure https://docs.cilium.io/en/latest/network/kubernetes/kubeproxy-free/#nodeport-xdp-on-azure
>     - Not sure what to do about this but probably at least add a note that it's legacy because Azure IPAM is legacy.

There's two aspects to that guide that seem outdated:
- Recommendation to install Ubuntu 18.04. This distro reached EOL in May 2023.
- Recommendation to use Legacy Azure IPAM. I don't think this is necessary, my guess would be that it only needs to be configured in native routing mode. Given that BYOCNI is mainly viable with tunneling mode, this means that the guide should probably suggest users to follow the "external installer" (Azure CNI powered by Cilium) approach, though admittedly that also means that if we make the recommendation then we should probably validate the instructions before making such a docs change. If we aren't able to do that, the minimum we could do is to state that it requires direct routing. @cilium/sig-datapath or #sig-datapath on Slack should be able to help with confirming the underlying requirements for KPR XDP acceleration on Azure.

tamilmani1989:
Thanks for the detailed suggestions @nbusseneau. @joestringer we will take care of it. I'm planning to take some and distribute the rest within my team. We will open a separate PR for each specific doc.

tamilmani1989:
@ldelossa will open separate PR for each task listed in this issue https://github.com/cilium/cilium/issues/27289 . I'm fine with closing this PR

PR #27203, opened 2023-08-02
CI: Add IPsec key rotation test

### LINKED ISSUE #26350: Testing IPsec key rotation

jschwinger233:
This PR adds IPsec key rotation test in `conformance-ipsec-e2e.yaml` (ci-ipsec-e2e). A new github action `conn-disrupt-test` is also used to reduce duplication for this test as well as IPsec upgrade test.

Fixes: #26350

### DIFF
--- a/.github/workflows/conformance-ipsec-e2e.yaml
+++ b/.github/workflows/conformance-ipsec-e2e.yaml
@@ -284,7 +284,39 @@ jobs:
               --junit-file "cilium-junits/${{ env.job_name }} (${{ join(matrix.*, ', ') }}).xml" \
               --junit-property github_job_step="Run tests (${{ join(matrix.*, ', ') }})"
 
-            ./contrib/scripts/kind-down.sh
+      - name: Rotate IPsec Key & Test (${{ join(matrix.*, ', ') }})
+        uses: cilium/cilium/.github/actions/conn-disrupt-test@48023b9f6df4fffa80162f418ae2344d33e36b11
+        with:
+          job-name: conformance-ipsec-e2e-key-rotation-${{ matrix.name }}
+          operation-cmd: |
+            cd /host/
+
+            KEYID=\$(kubectl get secret -n kube-system cilium-ipsec-keys -o go-template --template={{.data.keys}} | base64 -d | cut -c 1)
+            if [[ \$KEYID -ge 15 ]]; then KEYID=0; fi
+            data=\$(echo "{\"stringData\":{\"keys\":\"\$(((\$KEYID+1))) "rfc4106\(gcm\(aes\)\)" 59f4d92cccede1b1abc920104ca61cd552782e12 128\"}}")
+            kubectl patch secret -n kube-system cilium-ipsec-keys -p="\$data" -v=1
+
+            # Wait until key rotation starts
+            while true; do
+              keys_in_use=\$(kubectl -n kube-system exec daemonset/cilium -- cilium encrypt status | awk '/Keys in use/ {print \$NF}')
+              if [[ \$keys_in_use == 2 ]]; then
+                break
+              fi
+              echo "Waiting until key rotation starts"
+              sleep 30s
+            done
+
+            # Wait until key rotation completes
+            # By default the key rotation cleanup delay is 5min, let's sleep 4min before actively polling
+            sleep \$((4*60))
+            while true; do
+              keys_in_use=\$(kubectl -n kube-system exec daemonset/cilium -- cilium encrypt status | awk '/Keys in use/ {print \$NF}')
+              if [[ \$keys_in_use == 1 ]]; then
+                break
+              fi
+              echo "Waiting until key rotation completes"
+              sleep 30s
+            done
 
       - name: Fetch artifacts
         if: ${{ !success() }}
--- a/.github/workflows/tests-ipsec-upgrade.yaml
+++ b/.github/workflows/tests-ipsec-upgrade.yaml
@@ -218,7 +218,7 @@ jobs:
 
           CONFIG="${CILIUM_INSTALL_DEFAULTS} ${TUNNEL} ${LB_MODE} ${ENDPOINT_ROUTES} ${IPV6} ${MASQ} ${EGRESS_GATEWAY} ${ENCRYPT} ${HOST_FW} ${LB_ACCELERATION}"
           echo "cilium_install_defaults=${CONFIG}" >> $GITHUB_OUTPUT
-          
+
           JUNIT=""
           for NAME in ${{ matrix.kube-proxy }} ${{ matrix.tunnel }} ${{ matrix.lb-mode }} ${{ matrix.encryption }} ${{ matrix.endpoint-routes }}; do
             if [[ "${NAME}" != "" ]] && [[ "${NAME}" != "disabled" ]] && [[ "${NAME}" != "none" ]]; then
@@ -330,11 +330,14 @@ jobs:
             # interruption in such flows.
             ./cilium-cli connectivity test --include-conn-disrupt-test --conn-disrupt-test-setup
 
-      - name: Upgrade Cilium (${{ matrix.name }})
-        uses: cilium/little-vm-helper@908ab1ff8a596a03cd5221a1f8602dc44c3f906d # v0.0.12
+      - name: Upgrade Cilium & Test (${{ matrix.name }})
+        uses: cilium/cilium/.github/actions/conn-disrupt-test@48023b9f6df4fffa80162f418ae2344d33e36b11
         with:
-          provision: 'false'
-          cmd: |
+          job-name: ipsec-upgrade-${{ matrix.name }}
+          # Disable no-missed-tail-calls due to https://github.com/cilium/cilium/issues/26739
+          # Disable check-log-errors due to https://github.com/cilium/cilium-cli/issues/1858
+          extra-connectivity-test-flags: --test '!no-missed-tail-calls,!check-log-errors'
+          operation-cmd: |
             cd /host/
 
             CILIUM_CLI_MODE=helm ./cilium-cli upgrade \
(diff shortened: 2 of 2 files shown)


jschwinger233:
IPsec key rotation test was green: https://github.com/cilium/cilium/actions/runs/5751901605/job/15591693145
IPsec upgrade test hit some flakes during downgrade and I'm looking into it: https://github.com/cilium/cilium/actions/runs/5746964893/job/15577230167

.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -342,32 +336,11 @@ jobs:
             kubectl get pods --all-namespaces -o wide
             kubectl -n kube-system exec daemonset/cilium -- cilium status
 
-      - name: Test Cilium after upgrade (${{ matrix.name }})
-        uses: cilium/little-vm-helper@6a3d1797af8fab1a49f1c3d09afe3230aea559b6 # v0.0.11
+      - name: Downgrade Cilium ${{ env.cilium_stable_version }} & Test (${{ matrix.name }})
```
bimmlerd: nit:
```suggestion
      - name: Downgrade Cilium to ${{ env.cilium_stable_version }} & Test (${{ matrix.name }})
```

.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -322,16 +321,11 @@ jobs:
...
 
-            # Create pods which establish long lived connections. It will be used by
-            # subsequent connectivity tests with --include-conn-disrupt-test to catch any
-            # interruption in such flows.
-            ./cilium-cli connectivity test --include-conn-disrupt-test --conn-disrupt-test-setup
-
-      - name: Upgrade Cilium (${{ matrix.name }})
-        uses: cilium/little-vm-helper@6a3d1797af8fab1a49f1c3d09afe3230aea559b6 # v0.0.11
+      - name: Upgrade Cilium & Test (${{ matrix.name }})
+        uses: ./.github/actions/conn-disrupt-test
```
bimmlerd: I'm under the impression this action gets executed from the PR branch. If so, as a malicious actor I can put stuff in the commands of the action and get code execution in a PR, no?

bimmlerd (commented):
looks good - one security concern I'm not entirely sure about, I'm sure github-sec will chime in :)

.github/workflows/tests-ipsec-upgrade.yaml
jschwinger233: I'm not sure but I expect only a [member of cilium](https://github.com/cilium/community/blob/main/ladder/members.yaml) has permission to create a branch?

bimmlerd: That's true, but I think it'll also checkout a branch of a fork, which anyone can create. See https://github.com/cilium/cilium/pull/27203/files#diff-b77d9c4420c99e9bbac5afc52278677e846739ed4b7b4f7c1efdb24f1b69a249L216

pchaigno: I believe David is correct. The only other time we execute a checked out script (`./contrib/scripts/kind.sh`) in this workflow is within a VM (LVH). Might be worth getting a confirmation from @nbusseneau.

If this is insecure, then I guess our (only?) option is to move this new GitHub action to its own repository. Or can we use a specific version (specific SHA) of this new GitHub action without moving it elsewhere?

.github/actions/conn-disrupt-test/action.yaml
```diff
@@ -0,0 +1,48 @@
...
+          ${{ inputs.operation-cmd }}
+
+    - name: Perform Conn Disrupt Test
+      uses: cilium/little-vm-helper@0fcaa3fed17811fcd8b6f1b0dc1f24e5f4ff6b13 # v0.0.7
+      with:
+        provision: 'false'
+        cmd: |
+          cd /host/
+          # Disable no-missed-tail-calls due to https://github.com/cilium/cilium/issues/26739
+          # Disable check-log-errors due to https://github.com/cilium/cilium-cli/issues/1858
```
pchaigno: Downside of the new GitHub action is that we end up disabling those two checks for key rotations as well :disappointed: Any way to avoid that?

.github/workflows/conformance-ipsec-e2e.yaml
```diff
@@ -278,7 +278,40 @@ jobs:
...
+            cd /host/
+
+            KEYID=\$(kubectl get secret -n kube-system cilium-ipsec-keys -o go-template --template={{.data.keys}} | base64 -d | cut -c 1)
+            if [[ \$KEYID -ge 15 ]]; then KEYID=0; fi
+            data=\$(echo "{\"stringData\":{\"keys\":\"\$(((\$KEYID+1))) "rfc4106\(gcm\(aes\)\)" 59f4d92cccede1b1abc920104ca61cd552782e12 128\"}}")
+            kubectl patch secret -n kube-system cilium-ipsec-keys -p="\$data" -v=1
+
+            # Re-create cilium pods to reload new ipsec key
+            kubectl -n kube-system delete pod -l k8s-app=cilium
+            ./cilium-cli status --wait
```
pchaigno: This is not needed. There is a watcher that will detect the updated key file and reload the key automatically. I believe it runs once per minute or so.

.github/workflows/conformance-ipsec-e2e.yaml
```diff
@@ -278,7 +278,40 @@ jobs:
...
+            done
+
+            while true; do
+              keys_in_use=\$(kubectl -n kube-system exec daemonset/cilium -- cilium encrypt status | awk '/Keys in use/ {print \$NF}')
+              if [[ \$keys_in_use == 1 ]]; then
+                break
+              fi
+              echo "Waiting until key rotation completes"
+              sleep 30s
+            done
```
pchaigno, line 319: By default the key rotation cleanup delay is 5min, so you could probably do a 4min sleep before starting to actively poll for the number of keys.

pchaigno (commented):
Apart from the potential security issue we need to figure out, this looks great!

The fact that the test is passing however makes me very suspicious. We know that key rotations currently cause XfrmOutPolBlock errors (which @jrfastab is working on fixing). How can the test be passing? :confused:

.github/actions/conn-disrupt-test/action.yaml
jschwinger233: Yeah I can avoid these two hardcoded disabling by introducing one more input parameter, so caller can pass flags like
```
uses: conn-disrupt-test
with:
  extra-connectivity-test-flags: "--test '!no-missed-tail-calls'"
```
Action can therefore run connectivity test with caller-specified flags.

For key rotation test, caller just don't specify any extra flags so as to run full tests.

.github/workflows/tests-ipsec-upgrade.yaml
jschwinger233: Seems we can references a subdirectory in a public GitHub repository at a specific branch, ref, or SHA.

nbusseneau: That is correct, we must be careful with how local `uses` are used in privileged contexts such as `workflow_dispatch`-based workflows, since they have access to repository secrets and also have extended permissions.

If you grep for `uses: .` you can inspect how do we do it for the existing composite actions (for env variables and EKS configuration): we make sure the local `uses` are only done in the "trusted" part of the workflow, and we do not use them in the "untrusted" part (which is everything after we checked out the PR code, see line 234 in this file for example).

I was thinking the other day that we should remove local `uses` entirely and switch to a static reference anyway, that would help with not having to concern ourselves with this specific issue for maintenance and reviews, so I would advocate for doing so here:

```suggestion
        uses: ${{ github.repository }}/.github/actions/conn-disrupt-test@${{ inputs.context-ref || github.sha }}
```

This should pull the composite actions from the appropriate context based on trigger (if `workflow_dispatch`, `input.context-ref` will be either the PR target branch or the PR branch itself depending on whether the PR comes from a fork or not, and if `schedule`, `github.sha` will be used as fallback).

jschwinger233:
Pending #27276

nbusseneau:
I've set this back to draft while we still iterate on the PR, so as to remove this PR from the stale queue.

.github/workflows/tests-ipsec-upgrade.yaml
jschwinger233: Unfortunately `uses: ${{ github.repository }}/.github/actions/conn-disrupt-test@${{ inputs.context-ref || github.sha }}` is invalid: https://github.com/cilium/cilium/actions/runs/5889885321

```
Invalid workflow file: .github/workflows/tests-ipsec-upgrade.yaml#L333
The workflow is not valid. .github/workflows/tests-ipsec-upgrade.yaml (Line: 333, Col: 15): Unrecognized named-value: 'github'. Located at position 1 within expression: github.repository .github/workflows/tests-ipsec-upgrade.yaml (Line: 352, Col: 15): Unrecognized named-value: 'github'. Located at position 1 within expression: github.repository
```

So I decide to create an independent PR https://github.com/cilium/cilium/pull/27567 to add this new action, then we can refer to the action using `cilium/cilium/.github/actions/conn-disrupt-test@SHA` to address the security concern.

.github/actions/conn-disrupt-test/action.yaml
pchaigno: Sounds good to me :+1:

jschwinger233:
Finally both ci-ipsec-e2e and ci-ipsec-upgrade pass!

pchaigno (approved):

brb (approved):

pchaigno:
@jschwinger233 My retrigger didn't work. I think you'll have to rebase first :disappointed:

PR #27212, opened 2023-08-02
datapath: Cover subnet encryption in XFRM leak test

### LINKED ISSUE #25793: Unit tests for XFRM config removal
We should extend the IPsec unit tests to prevent https://github.com/cilium/cilium/pull/25784 from happening again. We probably need to make sure that they cover both cluster-pool & co. IPAMs as well as Azure & ENI IPAMs. The latter are probably not covered today and there are still small differences in the way we handle these two groups of IPAM modes.

pchaigno:
This commit complements 27a8fb113b ("datapath: Integration test for XFRM leaks on node churn"). In that previous commit, subnet encryption (the XFRM config used for ENI and Azure IPAM modes) was not covered. This commit covers it.

There are two pre-conditions to subnet encryption: nodes must have a list of pod subnets (cf. `IPv{4,6}PodSubnets`) and the encryption interface must exist and have an IP address. The test configures these two preconditions before running the usual checks for XFRM config leaks.

### DIFF
--- a/pkg/datapath/linux/node_linux_test.go
+++ b/pkg/datapath/linux/node_linux_test.go
@@ -752,6 +752,38 @@ func (s *linuxPrivilegedBaseTestSuite) TestNodeUpdateIDs(c *check.C) {
 
 // Tests that we don't leak XFRM policies and states as nodes come and go.
 func (s *linuxPrivilegedBaseTestSuite) TestNodeChurnXFRMLeaks(c *check.C) {
+	externalNodeDevice := "ipsec_interface"
+
+	// Cover the XFRM configuration for IPAM modes cluster-pool, kubernetes, etc.
+	config := datapath.LocalNodeConfiguration{
+		EnableIPv4:  s.enableIPv4,
+		EnableIPv6:  s.enableIPv6,
+		EnableIPSec: true,
+	}
+	//s.testNodeChurnXFRMLeaksWithConfig(c, config)
+
+	// In the case of subnet encryption (tested below), the IPsec logic
+	// retrieves the IP address of the encryption interface directly so we need
+	// a dummy interface.
+	removeDevice(externalNodeDevice)
+	err := setupDummyDevice(externalNodeDevice, net.ParseIP("1.1.1.1"), net.ParseIP("face::1"))
+	c.Assert(err, check.IsNil)
+	defer removeDevice(externalNodeDevice)
+	option.Config.EncryptInterface = []string{externalNodeDevice}
+
+	// Cover the XFRM configuration for subnet encryption: IPAM modes AKS and EKS.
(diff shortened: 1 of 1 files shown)


jschwinger233 (commented):
I have to confess that I didn't know encryption interface before, but my initial understanding is:
1. it's used for non-tunneling mode, replacing the cilium_vxlan.
2. it's used for EKS because of its special IPAM mode: node IPs are within the pod CIDRs, so we don't need tunnels, but consequently we need to specify encryption interface.

In the test, encryption iface's address is 1.1.1.1, but pod subnet is 4.4.0.0/16. If my understanding is correct, maybe we should assign a IP address within the pod CIDR to encrption  interface?

pchaigno:
> I have to confess that I didn't know encryption interface before, but my initial understanding is:
> 1. it's used for non-tunneling mode, replacing the cilium_vxlan.
> 2. it's used for EKS because of its special IPAM mode: node IPs are within the pod CIDRs, so we don't need tunnels, but consequently we need to specify encryption interface.

It's actually a bit simpler: it's just the native device from which encrypted traffic will leave. Typically some eth0 interface. On EKS and AKS, if `option.Config.UseCiliumInternalIPForIPsec=false` (default), we then use the IP address of that interface as the outer IP address for IPsec encapsulation. 

> In the test, encryption iface's address is 1.1.1.1, but pod subnet is 4.4.0.0/16. If my understanding is correct, maybe we should assign a IP address within the pod CIDR to encrption interface?

It is therefore expected that the IP address of that eth0 interface isn't part of the pod subnet.

jschwinger233 (approved):

brb:
@pchaigno I am a bit puzzled about this PR - did you intend to add a new test case, or to replace the existing one to cover the subnet encryption? AFAICS, the latter was done.

pkg/datapath/linux/node_linux_test.go
```diff
@@ -752,6 +752,38 @@ func (s *linuxPrivilegedBaseTestSuite) TestNodeUpdateIDs(c *check.C) {
...
 func (s *linuxPrivilegedBaseTestSuite) TestNodeChurnXFRMLeaks(c *check.C) {
+	externalNodeDevice := "ipsec_interface"
+
+	// Cover the XFRM configuration for IPAM modes cluster-pool, kubernetes, etc.
+	config := datapath.LocalNodeConfiguration{
+		EnableIPv4:  s.enableIPv4,
+		EnableIPv6:  s.enableIPv6,
+		EnableIPSec: true,
+	}
+	//s.testNodeChurnXFRMLeaksWithConfig(c, config)
```
brb, line 763: Probably unintended disablement.

pchaigno: Yep, that's unintended. I did verify locally that the test works with and without this commented though. So should be easy to uncomment.

Issue #27183, opened 2023-08-01
Mutual Auth does not handle having auth table rules removed by non-agent actors


youngnick:
During some other work, @joestringer noted that, if an entry is deleted from the mutual auth table, then packets will be dropped with a "needs auth" signal to the agent, but because the agent has the authentication cached, it will not take any action.

The agent auth table code should be robust against auth table changes, and should be able to fix the situation.

meyskens:
Fix in https://github.com/cilium/cilium/pull/27241

PR #27216, opened 2023-08-02
Add a bpf auth flush command

meyskens:
This command in cilium-agent allows the user to flush the auth map. This can be used to force re-auths between all entities for debugging purposes.

This is useful to force an auth handshake on a given agent node when debugging. (once https://github.com/cilium/cilium/issues/27183 is fixed)

This function works similar to the `nat flush` command. 

```release-note
Add `cilium bpf auth flush` command for debugging purposes
```


cilium/cmd/bpf_auth_flush.go
```diff
@@ -0,0 +1,57 @@
...
+
+			Fatalf("Cannot load auth bpf map: %s", err)
+		}
+
+		entries := 0
+		deleteEntry := func(key *authmap.AuthKey, _ *authmap.AuthInfo) {
+			if key == nil {
+				return
+			}
+			authMap.Delete(*key)
```
sayboras: probably checking returned err here, and print out if not nil. Also, only do entries++ if there is no error.

sayboras:
Also you might need to run `make -C Documentation update-cmdref` to generate related docs

cilium/cmd/bpf_auth_flush.go
meyskens: good point!

cilium/cmd/bpf_auth_flush.go
```diff
@@ -0,0 +1,61 @@
...
+
+			Fatalf("Cannot load auth bpf map: %s", err)
+		}
+
+		entries := 0
+		deleteEntry := func(key *authmap.AuthKey, _ *authmap.AuthInfo) {
+			if key == nil {
+				return
+			}
+			err := authMap.Delete(*key)
```
mhofstetter, line 42: Re-authentication checks for existing (and valid) entries in the auth map cache. Therefore, deleting the entries from the auth map alone doesn't trigger a re-authentication and re-population of the auth map entry on subsequent requests.

https://github.com/cilium/cilium/blob/main/pkg/auth/manager.go#L127-L137

It is also visible from the log messages

```
level=debug msg="Handle authentication request" key="localIdentity=62189, remoteIdentity=64335, remoteNodeID=0, authType=spire" subsys=auth
level=debug msg="Already authenticated, skipping authentication" key="localIdentity=62189, remoteIdentity=64335, remoteNodeID=0, authType=spire" subsys=auth
level=debug msg="Clearing pending authentication" key="localIdentity=62189, remoteIdentity=64335, remoteNodeID=0, authType=spire" subsys=auth
```

Currently, we only read from the bpf auth map once at startup to populate the auth map cache.

-> To properly implement flushing the auth bpf map, the CLI would need to instruct the agent to flush the map AND the cache. (or push down the auth information in this case where an auth request for an already authenticated packet gets emitted)

mhofstetter (changes requested):

cilium/cmd/bpf_auth_flush.go
joestringer: Heads up that any `cilium bpf ...` commands should only talk directly to the bpf datapath, and not rely on the agent. This is to ensure that these commands can still operate even if the agent is unhealthy. If you want to properly flush the userspace cache as well, I suggest moving this to `cilium auth flush` or similar.

meyskens: > Re-authentication checks for existing (and valid) entries in the auth map cache. Therefore, deleting the entries from the auth map alone doesn't trigger a re-authentication and re-population of the auth map entry on subsequent requests.

This is why i wrote this to test a way to fix that, we should not rely on the cache always being correct as external things can delete the bpf entries. There is a 2nd PR open to fix that.

see https://github.com/cilium/cilium/issues/27183

sayboras (approved):
LGTM :heavy_check_mark:

cilium/cmd/bpf_auth_flush.go
mhofstetter: ok :) would be great to have such information in the PR description itself :)

mhofstetter (approved):

PR #27222, opened 2023-08-02
docs: Clean up prerequisites for the Ingress Controller

qmonnet:
This commit brings various minor improvements to the intro and prerequisites for the Ingress Controller documentation:

- Add a link to Kubernetes' documentation for Ingress, given that we never define or explain what Kubernetes Ingress is
- Instruct users to "enable" kube-proxy replacement, given that the values "strict" or "partial" are deprecated.
- Use the Helm value instead of the agent flag for the L7 proxy configuration, simply to remain consistent with the item above (KPR).
- Remove the requirement on Kubernetes 1.19+, given that 1.19 is now the minimum version supported by Cilium.

**NOTE**: `kubeProxyReplacement=true` is equivalent to the legacy `strict` mode, but not to the `partial` mode. It's likely that the full KPR is not necessary for running the Ingress Controller, but I wasn't able to find the exact subset of options that are required. @sayboras (you authored this requirement in the first place), do you know?

### DIFF
--- a/Documentation/network/servicemesh/ingress.rst
+++ b/Documentation/network/servicemesh/ingress.rst
@@ -10,7 +10,7 @@
 Kubernetes Ingress Support
 **************************
 
-Cilium uses the standard Kubernetes Ingress resource definition, with
+Cilium uses the standard `Kubernetes Ingress`_ resource definition, with
 an ``ingressClassName`` of ``cilium``. This can be used for path-based
 routing and for TLS termination. For backwards compatibility, the
 ``kubernetes.io/ingress.class`` annotation with value of ``cilium``
@@ -43,14 +43,17 @@ path prefix) between resources.
 This is a step-by-step guide on how to enable the Ingress Controller in
 an existing K8s cluster with Cilium installed.
 
+.. _Kubernetes Ingress: https://kubernetes.io/docs/concepts/services-networking/ingress/
+
 Prerequisites
 #############
 
-* Cilium must be configured with ``kubeProxyReplacement`` as partial
-  or strict. Please refer to :ref:`kube-proxy replacement <kubeproxy-free>`
-  for more details.
-* Cilium must be configured with the L7 proxy enabled using the ``--enable-l7-proxy`` flag (enabled by default).
-* The minimum supported Kubernetes version for Ingress is 1.19.
+* Cilium must be configured with NodePort enabled, using
+  ``nodePort.enabled=true`` or by enabling the kube-proxy replacement with
+  ``kubeProxyReplacement=true``. For more information, see :ref:`kube-proxy
+  replacement <kubeproxy-free>`.
+* Cilium must be configured with the L7 proxy enabled using ``l7Proxy=true``
+  (enabled by default).
 
 .. include:: installation.rst
 


Documentation/network/servicemesh/ingress.rst
```diff
@@ -43,14 +43,15 @@ path prefix) between resources.
...
 
-* Cilium must be configured with ``kubeProxyReplacement`` as partial
-  or strict. Please refer to :ref:`kube-proxy replacement <kubeproxy-free>`
-  for more details.
-* Cilium must be configured with the L7 proxy enabled using the ``--enable-l7-proxy`` flag (enabled by default).
-* The minimum supported Kubernetes version for Ingress is 1.19.
+* Cilium must be configured with ``kubeProxyReplacement`` enabled. Please refer
+  to :ref:`kube-proxy replacement <kubeproxy-free>` for more details.
+* Cilium must be configured with the L7 proxy enabled using ``l7Proxy``
+  (enabled by default).
```
zacharysarah: Minor edits for style guide format
```suggestion
* Cilium must be configured with ``kubeProxyReplacement`` enabled. For more
   information, see :ref:`kube-proxy replacement <kubeproxy-free>`.
* Cilium must be configured with the L7 proxy enabled using ``l7Proxy``
  (enabled by default).
```

zacharysarah (dismissed):
@qmonnet This is awesome. ✨ Minor formatting tweak, otherwise LGTM

Documentation/network/servicemesh/ingress.rst
qmonnet: Fixed, thank you!

Documentation/network/servicemesh/ingress.rst
```diff
@@ -43,14 +43,15 @@ path prefix) between resources.
...
+.. _Kubernetes Ingress: https://kubernetes.io/docs/concepts/services-networking/ingress/
+
 Prerequisites
 #############
 
-* Cilium must be configured with ``kubeProxyReplacement`` as partial
-  or strict. Please refer to :ref:`kube-proxy replacement <kubeproxy-free>`
-  for more details.
-* Cilium must be configured with the L7 proxy enabled using the ``--enable-l7-proxy`` flag (enabled by default).
-* The minimum supported Kubernetes version for Ingress is 1.19.
```
sayboras, line 53: I think we still need this one, before 1.19, Ingress resource is with a different API version (i.e. beta)

sayboras (approved):
> NOTE: kubeProxyReplacement=true is equivalent to the legacy strict mode, but not to the partial mode. It's likely that the full KPR is not necessary for running the Ingress Controller, but I wasn't able to find the exact subset of options that are required

Actually, partial along might not be enough, we also need `enable-node-port`  to complement.

Documentation/network/servicemesh/ingress.rst
sayboras: for k8s version 1.20+, the user can still use the old Ingress version, as the k8s api server will automatically perform the conversion.

qmonnet:
> Actually, partial along might not be enough, we also need `enable-node-port` to complement.

@sayboras "Partial" does not exist anymore, we have "enabled" (formerly "strict") or disabled (pick your options). My question in this case, is, is it enough to have KPR "disabled" and `enable-node-port` enabled for the Ingress Controller to work? Or is there something else that needs to be enabled?

Documentation/network/servicemesh/ingress.rst
qmonnet: But Cilium with K8s < 1.19 is no longer supported, is it?

qmonnet: From https://docs.cilium.io/en/v1.14/network/kubernetes/requirements/#kubernetes-version

sayboras: ah good point, I was under impression that we still have 1.16+ :heavy_check_mark:

sayboras:
> "Partial" does not exist anymore, we have "enabled" (formerly "strict") or disabled ("pick your options"). My question in this case, is, is it enough to have KPR "disabled" and `enable-node-port` enabled for the Ingress Controller to work? Or is there something else that needs to be enabled?

Ah right, let me do a couple of tests locally and get back to you on this :+1:

sayboras:
:wave: sorry for the late reply, seems like the combination of kpr=false and nodePort.enabled=true is working for Ingress (except direct traffic to Ingress LB service nodeport). I have created #27304 to add coverage in CI so that we don't break this behavior unintentionally.

Can you help to update the docs if possible, otherwise, I can send the follow-up PR?

```
$ ksysg cm cilium-config -o json | zgrep -E "replacement|enable-node-port" | head -3
        "enable-node-port": "true",
        "kube-proxy-replacement": "false",
        "kube-proxy-replacement-healthz-bind-address": "",

$ lb=$(kubectl get ingress basic-ingress -o jsonpath='{.status.loadBalancer.ingress[0].ip}')

$ curl -k http://$lb/details/1           
{"id":1,"author":"William Shakespeare","year":1595,"type":"paperback","pages":200,"publisher":"PublisherA","language":"English","ISBN-10":"1234567890","ISBN-13":"123-1234567890"}
```

sayboras (approved):
Thanks and LGTM ✅

qmonnet:
@sayboras: Thanks! Grepping for `strict` or `partial` in the docs, I see similar requirements on “[KPR] set to either partial or strict” in `Documentation/servicemesh/l7-traffic-management.rst` and `Documentation/operations/troubleshooting_servicemesh.rst`. Do you know if NodePort is the actual feature that we need there as well?

PR #27223, opened 2023-08-02
bgpv1: Use Path type in AdvertisePath & WithdrawPath

rastislavs:
This PR makes the `AdvertisePath()` and `WithdrawPath()` methods of the `Router` interface use the `Path` datatype introduced in https://github.com/cilium/cilium/pull/26803

This makes the `bgpv1/gobgp` package more generic and allows advertising routes with more fine-grained NLRI and Path Attributes in the future.

### DIFF
--- a/pkg/bgpv1/gobgp/conversions.go
+++ b/pkg/bgpv1/gobgp/conversions.go
@@ -43,6 +43,7 @@ func ToGoBGPPath(p *types.Path) (*gobgp.Path, error) {
 		Age:    ageTimestamp,
 		Best:   p.Best,
 		Family: family,
+		Uuid:   p.UUID,
 	}, nil
 }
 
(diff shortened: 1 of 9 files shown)


pkg/bgpv1/gobgp/conversions.go
```diff
@@ -69,6 +70,7 @@ func ToAgentPath(p *gobgp.Path) (*types.Path, error) {
 		PathAttributes: pattrs,
 		AgeNanoseconds: ageNano,
 		Best:           p.Best,
+		GoBGPPathUUID:  p.Uuid,
```
YutaroHayakawa: IIRC, we decided to delete this GoBGPPath prefix.

YutaroHayakawa (approved):
LGTM with a nit.

pkg/bgpv1/gobgp/conversions.go
rastislavs: You are right, just pushed the change with the rename.

rastislavs:
ConformanceAKS (ci-aks) hit a new flake: https://github.com/cilium/cilium/issues/27236

PR #27233, opened 2023-08-03
Clearify QPS and BURST are ment to be variables in the given l2-announcements documentation example

omniproc:
The current example gives the impression that the actual QPS and BURST variables are supposed to be provided within curly brackets while in fact in the Helm chart they're simply used as integers. I suppose this was a simple typo and both those params are ment to be variables in the given example, similar to the existing KUBERNETES_API_SERVER_ADDRESS and KUBERNETES_API_SERVER_PORT variables provided in the same example.


learnitall:
Hey @omniproc, any reason why you closed this? CC @dylandreimerink as the author of these docs

omniproc:
> Hey @omniproc, any reason why you closed this? CC @dylandreimerink as the author of these docs

Formal reasons (signing process).

PR #27234, opened 2023-08-03
CODEOWNERS: remove stale cilium_egress_gateway_policy.go entry

giorio94:
The `pkg/k8s/watchers/cilium_egress_gateway_policy.go` file has been deleted in ebabe7032b3f ("egressgw: use Resource[T] to consume CiliumEgressGatewayPolicy"). Hence, let's remove the corresponding entry in the CODEOWNERS file.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -440,7 +440,6 @@ jenkinsfiles @cilium/ci-structure
 /pkg/k8s/watchers/cilium_network_policy.go @cilium/sig-policy
 /pkg/k8s/watchers/cilium_cidr_group.go @cilium/sig-policy
 /pkg/k8s/watchers/cilium_cidr_group_test.go @cilium/sig-policy
-/pkg/k8s/watchers/cilium_egress_gateway_policy.go @cilium/egress-gateway @cilium/sig-k8s
 /pkg/k8s/apis/cilium.io/client/crds/v2/ @cilium/sig-k8s
 /pkg/k8s/apis/cilium.io/client/crds/v2/ciliumegressgatewaypolicies.yaml @cilium/egress-gateway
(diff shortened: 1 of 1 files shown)


julianwiedmann (approved):

bimmlerd:
We're losing the k8s approval as is though, should `egw/resource.go` be covered by k8s instead?

jibi:
> > We're losing the k8s approval as is though, should `egw/resource.go` be covered by k8s instead?
> 
> I'm not sure about the policy for that honestly. AFAIU all resource constructors are currently in the `pkg/k8s/resource_ctors.go` file. /cc @cilium/egress-gateway

ah, didn't know that, perhaps then we should just move `egw/resource.go` to that file as well :+1:

giorio94:
> > > We're losing the k8s approval as is though, should `egw/resource.go` be covered by k8s instead?
> > 
> > 
> > I'm not sure about the policy for that honestly. AFAIU all resource constructors are currently in the `pkg/k8s/resource_ctors.go` file. /cc @cilium/egress-gateway
> 
> ah, didn't know that, perhaps then we should just move `egw/resource.go` to that file as well +1

In that case, I'd propose to merge this PR to fix the CI failure, and let you open a followup one to move the function to that file. @bimmlerd @jibi WDYT?

bimmlerd (approved):
oh I didn't realise this is failing CI, sure, go ahead.

giorio94:
> oh I didn't realise this is failing CI, sure, go ahead.

Not a hard failure actually, since the check is not required. But the PR doesn't become green...

giorio94:
Marking as ready to merge, given that it touches the codeowners file only.

aanm (approved):

PR #27237, opened 2023-08-03
Fix agent panic in case of malformed objects retrieved from the kvstore, and improve validation

giorio94:
This PR adds extra validation to prevent processing invalid node and service objects retrieved from the kvstore, which could trigger unexpected behavior.

```release-note
Fix agent panic in case malformed objects are retrieved from the kvstore, and improve validation
```

### DIFF
--- a/pkg/node/types/node.go
+++ b/pkg/node/types/node.go
@@ -14,6 +14,7 @@ import (
 	"github.com/cilium/cilium/api/v1/models"
 	"github.com/cilium/cilium/pkg/annotation"
 	"github.com/cilium/cilium/pkg/cidr"
+	cmtypes "github.com/cilium/cilium/pkg/clustermesh/types"
 	"github.com/cilium/cilium/pkg/defaults"
 	ipamTypes "github.com/cilium/cilium/pkg/ipam/types"
 	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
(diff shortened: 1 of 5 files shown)


pkg/node/types/node.go
```diff
@@ -615,7 +616,23 @@ func (n *Node) Unmarshal(_ string, data []byte) error {
...
+func (n *Node) validate() error {
+	// Skip the ClusterID check if it matches the local one, as we assume that
+	// it has already been validated, and to allow it to be zero.
+	if n.ClusterID != option.Config.ClusterID {
+		if err := cmtypes.ValidateClusterID(n.ClusterID); err != nil {
+			return err
+		}
+	}
+
+	return nil
```
pippolo84, line 637: Possible style improvement:

```suggestion
	if n.ClusterID == option.Config.ClusterID {
		return nil
	}

	return cmtypes.ValidateClusterID(n.ClusterID)
```

pippolo84 (approved):
LGTM! :rocket: 
I've left inline a non-blocking style improvement advice, feel free to do as you please.

pkg/node/types/node.go
giorio94: Generally speaking, I agree that your proposal is cleaner. I'd personally prefer to preserve the current approach, though, as it simplifies the addition of further checks, and it is consistent with the services validation.

squeed (approved):

giorio94:
The last push reverted a minor modification for better consistency with the previous implementation, with no functional differences.

ysksuzuki (approved):

PR #27241, opened 2023-08-03
Re-auth if signal got send again after added to cache

### LINKED ISSUE #27183: Mutual Auth does not handle having auth table rules removed by non-agent actors

meyskens:
This change will re-trigger the auth mechanism if a signal got send but the entry was already in cache.
It adds a 1 second backoff time to allow for the backend map to finish updating, which is why it was added in the first place.

Why does this happen? If somebody removes the entry in the bpf map that is not notified to the cell cache (meaning an external actor) we currently refuse to do the auth handshake again till the entry in the cache is gone.

It does add a small backoff time to not cause un-needed handshakes when data arrives and signals auth_needed between the mutex being released and the datapath getting the update. The current time it will allow for that is 1 second which seems to be okay. It is currently not configurable as it is quite an internal complexity. However I am open to reasons we should export it.

Fixes: https://github.com/cilium/cilium/issues/27183

```release-note
Do mutual authentication handshake again if mismatch between bpf map and cached map happens
```

### DIFF
--- a/pkg/auth/cell.go
+++ b/pkg/auth/cell.go
@@ -45,23 +45,27 @@ var Cell = cell.Module(
 		newAlwaysFailAuthHandler,
 	),
 	cell.Config(config{
-		MeshAuthEnabled:    true,
-		MeshAuthQueueSize:  1024,
-		MeshAuthGCInterval: 5 * time.Minute,
+		MeshAuthEnabled:               true,
+		MeshAuthQueueSize:             1024,
+		MeshAuthGCInterval:            5 * time.Minute,
+		MeshAuthSignalBackoffDuration: 1 * time.Second, // this default is based on the default TCP retransmission timeout
 	}),
 	cell.Config(MutualAuthConfig{}),
 )
 
 type config struct {
-	MeshAuthEnabled    bool
-	MeshAuthQueueSize  int
-	MeshAuthGCInterval time.Duration
+	MeshAuthEnabled               bool
+	MeshAuthQueueSize             int
+	MeshAuthGCInterval            time.Duration
+	MeshAuthSignalBackoffDuration time.Duration
 }
 
 func (r config) Flags(flags *pflag.FlagSet) {
 	flags.Bool("mesh-auth-enabled", r.MeshAuthEnabled, "Enable authentication processing & garbage collection (beta)")
 	flags.Int("mesh-auth-queue-size", r.MeshAuthQueueSize, "Queue size for the auth manager")
 	flags.Duration("mesh-auth-gc-interval", r.MeshAuthGCInterval, "Interval in which auth entries are attempted to be garbage collected")
+	flags.Duration("mesh-auth-signal-backoff-duration", r.MeshAuthSignalBackoffDuration, "Time to wait betweeen two authentication required signals in case of a cache mismatch")
+	flags.MarkHidden("mesh-auth-signal-backoff-duration")
 }
 
 type authManagerParams struct {
@@ -93,7 +97,7 @@ func registerAuthManager(params authManagerParams) (*AuthManager, error) {
 	mapWriter := newAuthMapWriter(params.Logger, params.AuthMap)
 	mapCache := newAuthMapCache(params.Logger, mapWriter)
 
-	mgr, err := newAuthManager(params.Logger, params.AuthHandlers, mapCache, params.NodeIDHandler)
+	mgr, err := newAuthManager(params.Logger, params.AuthHandlers, mapCache, params.NodeIDHandler, params.Config.MeshAuthSignalBackoffDuration)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create auth manager: %w", err)
 	}
--- a/pkg/auth/manager.go
+++ b/pkg/auth/manager.go
@@ -28,10 +28,11 @@ func (key signalAuthKey) String() string {
 }
 
 type AuthManager struct {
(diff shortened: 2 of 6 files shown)


pkg/auth/manager.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/cilium/cilium/pkg/policy"
 )
 
+const authSignalBackoffTime = time.Second
```
joestringer: Could you add a comment with some rough context about why one second seems about the right amount of time? Is this corresponding to default TCP retry period?

joestringer: If we're unsure about the value, it can be an option to expose a hidden flag for this and just avoid documenting / avoid exposing in helm, just in case we might want to change it later in a production environment. But if we're pretty confident about the underlying logic for this time then that might not be necessary.

pkg/auth/manager.go
```diff
@@ -128,10 +130,15 @@ func handleAuthentication(a *authManager, k authKey, reAuth bool) {
...
+			// If the entry was cached more than authSignalBackoffTime
+			// it will authenticate again, this is to make sure that
+			// we re-authenticate if the authmap was updated by an
+			// external source.
+			if i, err := a.authmap.GetCacheInfo(key); err == nil && i.expiration.After(time.Now()) && time.Now().Before(i.storedAt.Add(authSignalBackoffTime)) {
 				a.logger.
 					WithField("key", key).
-					Debug("Already authenticated, skipping authentication")
+					WithField("storedAt", i.storedAt).
+					Debug("Already authenticated in the past second, skipping authentication")
```
joestringer: nit: the second is based on a variable, would be nice to `%s` it in case we ever decide to change the timer.

pkg/auth/manager.go
```diff
@@ -128,10 +130,15 @@ func handleAuthentication(a *authManager, k authKey, reAuth bool) {
 			// Check if the auth is actually required, as we might have
 			// updated the authmap since the datapath issued the auth
 			// required signal.
-			if i, err := a.authmap.Get(key); err == nil && i.expiration.After(time.Now()) {
+			// If the entry was cached more than authSignalBackoffTime
+			// it will authenticate again, this is to make sure that
+			// we re-authenticate if the authmap was updated by an
+			// external source.
```
joestringer, line 138: Maybe for my education, what is the harm in always authenticating? Is there a particular attack pattern or costly operation we are trying to protect the control plane against?

I'm thinking maybe if an adversarial app deploys and starts creating many many new connections immediately and it's subject to authentication, then probably in that case it is better to skip. Maybe that's the use case you have in mind?

joestringer (approved):
Looks good, thanks! I have a few questions that I think are inconsequential for the fix.

pkg/auth/manager.go
sayboras: not sure how the one second backoff works in practice. Do you have any idea how long auth will take, probably we should add the metric for the same :thinking: ? 

IMO, with the metrics and hidden flag, user can easily tweak the value.

sayboras (approved):

pkg/auth/manager.go
meyskens: Added a flag :) 
It doesn't really depend on how long auth will take but it will catch the scenario where many auth requests are created at once where it is finished but the datapath isn't up to date but the cache is. We specifically added this later as we saw this causing more unneeded handshakes and map updates. Howerver this causes manual removal of auth map entries to block traffic till the cache expired

pkg/auth/cell.go
```diff
@@ -45,23 +45,26 @@ var Cell = cell.Module(
...
+	MeshAuthQueueSize         int
+	MeshAuthGCInterval        time.Duration
+	MeshAuthSignalBackoffTime time.Duration
 }
 
 func (r config) Flags(flags *pflag.FlagSet) {
 	flags.Bool("mesh-auth-enabled", r.MeshAuthEnabled, "Enable authentication processing & garbage collection (beta)")
 	flags.Int("mesh-auth-queue-size", r.MeshAuthQueueSize, "Queue size for the auth manager")
 	flags.Duration("mesh-auth-gc-interval", r.MeshAuthGCInterval, "Interval in which auth entries are attempted to be garbage collected")
+	flags.Duration("mesh-auth-signal-backoff-time", r.MeshAuthSignalBackoffTime, "Time to wait betweeen two authentication required signals in case of a cache mismatch")
```
joestringer: nit, since "time" is a bit ambiguous whether it means a specific point in time or a duration:
```suggestion
	flags.Duration("mesh-auth-signal-backoff-duration", r.MeshAuthSignalBackoffTime, "Time to wait betweeen two authentication required signals in case of a cache mismatch")
```

I also wonder if we should mark this as hidden. Most users are not going to need to customize it, but I do think it's worth having here in case it turns out to be useful to modify.

joestringer: Related to the "hidden" comment, I see that CI failed specifically because the docs weren't yet update for this flag. I guess if we hide it then no docs updates are necessary, but either way the PR needs to fix up that CI failure.

meyskens: I marked it as hidden so the documentation will be correct

PR #27246, opened 2023-08-03
Enable strict validation of cluster config for clustermesh

giorio94:
Currently, the cluster config validation function always adopts a loose approach for backward compatibility, allowing nil/empty configurations as valid. As new features may depend on the config presence, let's make this configurable.

```release-note
Enable strict validation of cluster config for clustermesh
```


YutaroHayakawa (approved):

giorio94:
Rebased onto main to fix a test failure due to a mismatch between workflow and kind template.

PR #27247, opened 2023-08-03
Highscale chapter 2, redux redux

squeed:
This is high-scale IP cache, chapter 2, take 3. It is in response to commentary on the previous effort, #26028.

It allows for the configuration of which CIDRs should be considered in-mesh (and thus encapsulated), versus out-of-mesh and unencapsulated. In the high-scale ipcache mode, this must be explicitly set, since the ipcache itself does not contain that information.

**Differences from previous effort**:
- There is no longer a new CRD. Rather, encapsulation for a set of CIDRs is defined by an annotation on a CiliumCIDRGroup.
- Rather than assuming all CIDRs are in-cluster (and thus encapsulated), this assumes all CIDRs are out-of-cluster. Then, encapsulation is a configurable bit in the BPF map.

```release-note
Adds a new annotation for CiliumCIDRGroup resources, `network.cilium.io/high-scale-encapsulate`, which configures the routing engine in high-scale ipcache mode to force encapsulation to a give set of CIDRs.
```


squeed:
fyi @joestringer, who reviewed the previous PR.

joestringer:
Just briefly: My gut feeling is that the annotation is a useful, lightweight way to add the feature and start to gain user feedback on it without formalizing the API for it. I plan to propose an improvement to the way that CIDRs and routing information is expressed in Cilium CRDs going forward, but given that I don't even have a concrete proposal yet, it seems premature to try to block this effort upon my future intentions.

Two specific thoughts from that high level approach:
- One, should we use `beta.network.cilium.io` or something in the annotation to make it clear that we have not decided on this as the future long term approach but that users can start to use it and the interface may change in future?
- Two, CiliumCIDRGroup defines `externalCIDRs` in the spec currently. The exact semantics are not fully specified in the CRD but the usage for highscale ipcache seems like it is misaligned with the idea of externalCIDRs. This PR discusses enabling tunneling for peers, which implies that they are inside the cluster. This probably suggests that the long term approach is to add an `internalCIDRs` field there, or rename the field and define better semantics, or add this encapsulation configuration to another CRD, perhaps referencing one of these CIDRGroups.

squeed:
> Two specific thoughts from that high level approach:
> 
>     * One, should we use `beta.network.cilium.io` 

Very good idea

>     * Two, CiliumCIDRGroup defines `externalCIDRs` in the spec currently. (...) This PR discusses enabling tunneling for peers, which implies that they are inside the cluster.

AIUI this is used for a clustermesh-style scenario (but without the full control-plane; I'm fuzzy on the details). So, they really are all externalCIDRs. (@pchaigno can you confirm?). So the naming is not misleading, given the intended specifics.

Issue #25155, opened 2023-04-27
golangci-lint is slow and obtuse


joestringer:
```
run golangci-lint
  Running [/home/runner/golangci-lint-1.52.2-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
  Error: type `exists` is unused (unused)
  Error: func `createPrefixLengthCounter` is unused (unused)
  
  Error: issues found
  Ran golangci-lint in 480960ms
```

I've submitted a PR and golangci-lint complains like the above in the [failure run](https://github.com/cilium/cilium/actions/runs/4814339927/jobs/8571888780?pr=25153). 

There's two things that are suboptimal here:
(1) The errors have no context. Where is `exists`? The failure does not describe where the problem is. The second error at least has a useful symbol to track down with `grep`, but it would be much easier to deal with issues like this if the failure output just points to the actual file that has the problem.
(2) Running this target took 480s (8 minutes!)

Locally, I can build the daemon in around 20 seconds. Why is golangci-lint so slow?

```
$ time sh -c 'make clean && make -C daemon'
  CLEAN  proxylib
  CLEAN  bpf
  CLEAN  bpf/custom
  CLEAN  cilium
  CLEAN  daemon
  CLEAN  cilium-health
  CLEAN  cilium-health/responder
  CLEAN  bugtool
  CLEAN  tools/mount
  CLEAN  tools/sysctlfix
  CLEAN  operator
  CLEAN  plugins/cilium-docker
  CLEAN  plugins/cilium-cni
  CLEAN  tools/alignchecker
  CLEAN  tools/mount
  CLEAN  hubble-relay
make: Entering directory '/home/joe/git/cilium/daemon'
  GO     daemon/cilium-agent
make: Leaving directory '/home/joe/git/cilium/daemon'

real    0m13.508s
user    0m20.665s
sys     0m6.498s
```

Tasks:
- [x] Ensure that golangci-lint github workflows provide details about the file and line where the linter fails (#25157)
- [ ] Fix or delete the `golangci-lint-fix` makefile target
- [ ] Ensure that golangci-lint workflow can complete in <2 mins on CI
  - [ ] Explore whether we can leverage better caching to achieve this
  - [ ] Consider removing or replacing slow linters

joestringer:
Looks like in the repo we at least have `make golangci-lint-fix` to generate the fixes for the complaints, so that simplifies reacting to the issues reported by this workflow.

joestringer:
[...]

PR #27253, opened 2023-08-03
Add time wrapper to test agent delays in CI

joestringer:
Cilium's internal business logic relies on a highly parallel combination of reactive handlers for incoming information, "triggers" that ratelimit requests for processing to ensure Cilium does not over-consume resources, and "controllers" that periodically perform updates or resiliency checks of configured state. While in general most things are "eventually consistent", the presence of time-based triggers and controllers can introduce challenges when evaluating how Cilium will perform once the "eventual consistency" is resolved.

Taking an example of a situation where this eventual consistency caused issues that were not identified during pre-release testing, consider the issue fixed by https://github.com/cilium/cilium/pull/27327 . Cilium v1.14.0 was released with a bug where Cilium appeared to work correctly for the first several minutes following startup, then afterwards introduced connectivity disruption to specific peers. This was traced down to a logic delay tied to a 10 minute timer following startup which caused some state to be recomputed & configured in the system, ultimately causing the packet loss.

While better unit testing for the individual package could have caught this issue earlier, the issue was also triggered by the integration between the specific package and other logic in other packages. It is quite difficult to systematically identify time-based errors across the entire agent by relying purely on such testing in each package. This PR attempts to provide a more systematic safety net for timer-based issues by providing a hidden flag to override all timers within Cilium and ensuring that Cilium's CI runs with these timers set to a minimal value.

One of the interesting challenges with this PR is that it can be tempting for developers to rely on time-based tricks in order to ensure the execution order of specific pieces of logic. However, when the system is highly loaded, such mechanisms can become unreliable as an ordering enforcement mechanism. As a side-objective, this PR also hopes to make such tricks less viable on order to incentivize implementing better ordering mechanisms.

Review tips: The `treewide` commit has about 2/3 of the changes but is generated almost entirely from a script, so can be overlooked for initial review:
```
    treewide: Wrap time package with internal wrapper
...
178 files changed, 186 insertions(+), 191 deletions(-)
```

Related: https://github.com/cilium/cilium/issues/28844

### DIFF
--- a/contrib/scripts/check-time.sh
+++ b/contrib/scripts/check-time.sh
@@ -1,0 +1,110 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Authors of Cilium
+
+set -eu
+
+GOIMPORTS=("golang.org/x/tools/cmd/goimports" "-w")
+MATCH="^[^a-zA-Z]*\"time\"$"
+EXCLUDED_DIRS=(
+  # Wrapper directories
+  "time"
+
+  # Generated directories
+  "api"
+  "client"
+  "slim"
+
+  # Not Go source
+  ".git"
+  "_build"
+  "contrib"
+  "externalversions"
+  "examples"
+  "install"
+  "Documentation"
+
+  # Not for cilium-agent
+  "bugtool"
+  "cilium-dbg"
+  "clustermesh"
+  "clustermesh-apiserver"
+  "health"
+  "operator"
+  "plugins"
+  "tools"
+  "test"
+  "testutils"
+  "vendor"
+
+  # Source shared with other binaries
+  "hive"
+  "lock"
+
+  # Skip override (not applicable)
+  "defaults"
+  "loadinfo"
+  "logging"
+  "metric"
+  "probes"
+  "rand"
+  "types"
+
+  # May need subsequent evaluation to detect resiliency issues in CI
+  "contexthelpers"
+  "rate"
+  "resiliency"
+)
+
+find_match() {
+  local target="."
+
+  # shellcheck disable=2046
+  grep "$@" -r --include \*.go \
+       $(printf "%s\n" "${EXCLUDED_DIRS[@]}" \
+         | xargs -I{} echo '--exclude-dir={}') \
+       --exclude \*_test.go \
+       -i "$MATCH" \
+       "$target"
+}
+
+check() {
+  # Used to cause failure when pkg/time is not used
+  if find_match ; then
+    >&2 echo "Found match for '$MATCH'. Please use pkg/time instead to improve ordering and consistency testing.";
+    >&2 echo "Run '$0 update' to update most instances automatically."
+    exit 1
+  fi
+}
+
+update() {
+  local files=()
+
+  while IFS='' read -r line; do
+    files+=("$line");
+  done < <(find_match -l | sort -u)
+  if [ "${#files[@]}" -eq 0 ]; then
+      return
+  fi
+
+  # Add a cilium pkg/time input next to the other imports
+  sed -i '/'"$MATCH"'/d; /"github.com\/cilium\/cilium\/.*"/a\
+	"github.com/cilium/cilium/pkg/time"' \
+      "${files[@]}"
+
+  # Fix up imports formatting
+  printf "%s\n" "${files[@]}" \
+  | xargs dirname \
+  | sort -u \
+  | xargs go run "${GOIMPORTS[@]}"
+}
+
+main() {
+  if [ $# -ge 1 ] && [ "$1" == "update" ]; then
+      update
+  fi
+
+  check
+}
+
+main "$@"
--- a/pkg/time/time.go
+++ b/pkg/time/time.go
@@ -41,16 +41,12 @@ const (
 )
 
 var (
-	After                  = time.After
-	Sleep                  = time.Sleep
-	Tick                   = time.Tick
 	ParseDuration          = time.ParseDuration
 	Since                  = time.Since
 	Until                  = time.Until
 	FixedZone              = time.FixedZone
 	LoadLocation           = time.LoadLocation
 	LoadLocationFromTZData = time.LoadLocationFromTZData
-	NewTicker              = time.NewTicker
 	Date                   = time.Date
 	Now                    = time.Now
 	Parse                  = time.Parse
@@ -58,8 +54,6 @@ var (
 	Unix                   = time.Unix
 	UnixMicro              = time.UnixMicro
 	UnixMilli              = time.UnixMilli
-	AfterFunc              = time.AfterFunc
-	NewTimer               = time.NewTimer
 )
 
 type (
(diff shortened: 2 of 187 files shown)


joestringer:
🎣 `f04-agent-policy-multi-node-1` seems to be failing consistently across k8s platforms:

```
• Failure in Spec Teardown (AfterEach) [130.273 seconds]
K8sAgentPolicyTest
  Multi-node policy test
    validates fromEntities policies
      with remote-node identity enabled [AfterEach]
        Validates fromEntities remote-node policy

        Found 1 k8s-app=cilium logs matching list of errors that must be investigated:
        2023-08-03T21:42:49.751308617Z level=error msg="Unable to find identity of previously used CIDR 172.18.0.2/32" subsys=ipcache
```

pkg/time/time.go
```diff
@@ -0,0 +1,86 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+// package time is a wrapper for the upstream system library for time.Time that
+// aliases most underlying types, but allows overrides for testing purposes.
+//
+// Synced to go-1.20.7.
+package time
```
sayboras, line 8: we can add the check similar to lock-check.sh if this new time package is supposed to be used in most of the cases.

pkg/time/time.go
```diff
@@ -0,0 +1,86 @@
...
+	Month      = time.Month
+	ParseError = time.ParseError
+	Ticker     = time.Ticker
+	Time       = time.Time
+	Timer      = time.Timer
+	Weekday    = time.Weekday
+)
+
+var (
+	MaxInternalTimerDelay time.Duration
```
joestringer, line 71: Make this link-time specified

joestringer: Or maybe not, we don't want to accidentally break this for all CI images => can't get user input because the images diverge too far from prod images.

pkg/time/time.go
```diff
@@ -0,0 +1,123 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+// package time is a wrapper for the upstream system library for time.Time that
```
danehans: nit: space between `.` and `Time`

contrib/scripts/check-time.sh
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env bash
```
mhofstetter, line 1: I wonder whether we could replace this checkfile with an existing linter provided by golangci-lint.

We're already using `gomodguard` to prevent the usage of some go modules -> https://github.com/cilium/cilium/blob/main/.golangci.yaml#L66-L81.

[DepGuard](https://golangci-lint.run/usage/linters/#depguard) should be able to prevent access to std library packages and provides the possibility to define regexes for the directories.

Linter: https://github.com/OpenPeeDeeP/depguard

joestringer: Do any of those options provide a way to script updating the dependency?

joestringer: While writing this script, I did feel like it would be better to have a more generic mechanism for this sort of linting / updating, since there are various scripts that work in a similar way and they really just look for some pattern and reject it, and maybe there's a way to update. But the update is actually the most complicated case.

There was also a point where I thought that it was a mistake to write this linter/updater in bash instead of Go :sweat_smile: At this point I think this is functional enough that I'd like to move on. The incremental benefit of reworking this part of the solution seems minimal to me.

Other than that, while I'm _open_ to more golangci-lint, on a personal level I'm not a fan because of #25155. Taking 8-10 minutes to complain about something a good `grep` can do in less than a second is painful.

danehans: +1 on proceeding since the script is functional but it might be a good idea to capture the Go/gomodguard refactor in an issue.

mhofstetter: > Do any of those options provide a way to script updating the dependency?

No it would only cover the linting (check) part

> At this point I think this is functional enough that I'd like to move on.

Totally fine for me :+1: Just thought it's worth to drop a comment

joestringer: Yep good feedback and worth considering. Thanks for raising this discussion item.

pkg/time/time.go
joestringer: Actually this was originally intentional as `time.Time` but the lib ended up wrapping more than just `time.Time` so I'll fix it up to refer that it just wraps the whole upstream library.

test/helpers/kubectl.go
```diff
@@ -122,6 +122,8 @@ var (
 		"ipv6NativeRoutingCIDR":  IPv6NativeRoutingCIDR,
 
 		"ipam.operator.clusterPoolIPv6PodCIDRList": "fd02::/112",
+
+		"extraConfig.max-internal-timer-delay": "5s",
```
gandro, line 126: Should we also add this to the GitHub Actions default values here:
https://github.com/cilium/cilium/blob/main/.github/actions/helm-default/action.yaml

gandro (approved):

tklauser (approved):

mhofstetter (approved):

test/helpers/kubectl.go
joestringer: I think I'll file an issue with this and other followups if that's alright with you. There's a few more small iterative items we can do on top like:
- Apply this change to more than just cilium-agent (operator etc.)
- See if there are additional packages that can reuse the time wrapper.
- Pick a different value than 5s

The PR is currently pulling in a lot of owners so I wanted to make sure we get together an implementation that works and transitions the codebase over, then we can iterate and tweak the details.

joestringer:
(Dropping a couple of autorequested codeowners where their areas were already reviewed)

christarazi (approved):

pkg/time/time.go
```diff
@@ -72,3 +66,58 @@ type (
...
+
+var (
+	MaxInternalTimerDelay time.Duration
+)
+
+// After overrides the stdlib time.After to enforce maximum sleepiness via
+// option.MaxInternalTimerDelay.
+func After(d Duration) <-chan Time {
+	if MaxInternalTimerDelay > 0 && d > MaxInternalTimerDelay {
+		d = MaxInternalTimerDelay
```
tommyp1ckles, line 78: Will this only be used in testing, should we emit a log that something is having time.After/Sleep capped?

harsimran-pabla (approved):

pkg/time/time.go
joestringer: No harm emitting a debug log here yeah. Could help folks understand better in future if they are unaware of this change.

test/helpers/kubectl.go
joestringer: Filed https://github.com/cilium/cilium/issues/28844

pkg/time/time.go
joestringer: Added to https://github.com/cilium/cilium/issues/28844

derailed (approved):
@joestringer Nice work! Just wish there was a better way to handle this ;(

contrib/scripts/check-time.sh
```diff
@@ -0,0 +1,110 @@
...
+  # Skip override (not applicable)
+  "defaults"
+  "loadinfo"
+  "logging"
+  "metric"
+  "probes"
+  "rand"
+  "types"
+
+  # May need subsequent evaluation to detect resiliency issues in CI
```
tommyp1ckles, line 53: Is this a todo?

tommyp1ckles (approved):

lmb (approved):

contrib/scripts/check-time.sh
joestringer: More or less yeah. The PR provides a good basis for now, there's some packages I stopped short of since they may have more nuanced interactions. I figured that for those, it'd be better to iterate on top once the initial bigger PR is agreed upon.

Added to the tasklist in https://github.com/cilium/cilium/issues/28844 .

PR #27255, opened 2023-08-04
Convert daemon identity restore logic over to newer IPCache APIs

joestringer:
This PR introduces logic to restore identity references into the name manager cache during startup and converts one of the older ipcache API users over to the newer async ipcache interface. The goal is to align more of the logic towards the new model for interacting with the ipcache, which should eventually result in a simpler codebase with fewer bugs.

Tasks:
- [x] Test the fixes using #27253
  - #27253 fails on its own against the main branch.
  - The final commit fix from this PR by itself also fails, but for a different test around ToFQDNs.
  - With all commits, this resolves the initial report as well as the hidden ToFQDNs reference tracking bug that was revealed through testing on top of #27253 .
- [x] Manual testing
- [x] Rebase against main
- [x] Create hotfix image for user testing
- [x] List all related filed issues
- [x] Investigate more surgical fix (https://github.com/cilium/cilium/pull/27327)
- [x] Review release note
- [ ] Review hard dependencies between identity restore and ipcache
- [ ] Review hard dependencies between ipcache and datapath
- [ ] Revisit whether each individual change is comprehensive enough to improve the tree
- [ ] Rewrite each commit message now that this is a code health change and not a bugfix

Related: https://github.com/cilium/cilium/pull/27327

### DIFF
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -59,7 +59,7 @@ import (
 	ipamMetadata "github.com/cilium/cilium/pkg/ipam/metadata"
 	ipamOption "github.com/cilium/cilium/pkg/ipam/option"
 	"github.com/cilium/cilium/pkg/ipcache"
-	ipcachetypes "github.com/cilium/cilium/pkg/ipcache/types"
+	ipcacheTypes "github.com/cilium/cilium/pkg/ipcache/types"
 	"github.com/cilium/cilium/pkg/k8s"
 	k8sConst "github.com/cilium/cilium/pkg/k8s/apis/cilium.io"
 	k8sClient "github.com/cilium/cilium/pkg/k8s/client"
@@ -583,7 +583,7 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 	// same numeric IDs as before, but the restored identities are to be upsterted to the new
 	// (datapath) ipcache after it has been initialized below. This is accomplished by passing
 	// 'restoredCIDRidentities' to AllocateCIDRs() and then calling
-	// UpsertGeneratedIdentities(restoredCIDRidentities) after initMaps() below.
+	// UpsertPrefixes(d.restoredCIDRs) after initMaps() below.
 	restoredCIDRidentities := make(map[netip.Prefix]*identity.Identity)
 	if len(d.restoredCIDRs) > 0 {
 		log.Infof("Restoring %d old CIDR identities", len(d.restoredCIDRs))
@@ -695,8 +695,17 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 		return nil, nil, fmt.Errorf("error while opening/creating BPF maps: %w", err)
 	}
 	// Upsert restored CIDRs after the new ipcache has been opened above
-	if len(restoredCIDRidentities) > 0 {
-		d.ipcache.UpsertGeneratedIdentities(restoredCIDRidentities, nil)
+	restoredResource := ipcacheTypes.NewResourceID(ipcacheTypes.ResourceKindDaemon, "", "")
+	if len(d.restoredCIDRs) > 0 {
+		// Asynchronously upsert the prefixes into the new ipcache map.
+		// By the time the datapath gets initialized to consume this
+		// map, ipcache should have completed the async work. If this
+		// doesn't occur, it will result in temporary drops in the
+		// time between datapath initialization until this async
+		// process completes.
+		// TODO: Add signalling mechanism to pass to the datapath to
+		//       ensure this ordering property.
+		d.ipcache.UpsertPrefixes(d.restoredCIDRs, source.Restored, restoredResource)
 	}
 	// Upsert restored local Ingress IPs
 	for _, ingressIP := range oldIngressIPs {
@@ -704,7 +713,7 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 			ingressIP,
 			labels.LabelIngress,
 			source.Restored,
-			ipcachetypes.NewResourceID(ipcachetypes.ResourceKindDaemon, "", ""),
+			restoredResource,
 		)
 	}
 	if len(oldIngressIPs) > 0 {
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -59,6 +59,7 @@ import (
 	ipamMetadata "github.com/cilium/cilium/pkg/ipam/metadata"
 	ipamOption "github.com/cilium/cilium/pkg/ipam/option"
 	"github.com/cilium/cilium/pkg/ipcache"
+	ipcachetypes "github.com/cilium/cilium/pkg/ipcache/types"
 	"github.com/cilium/cilium/pkg/ipmasq"
 	"github.com/cilium/cilium/pkg/k8s"
 	k8sClient "github.com/cilium/cilium/pkg/k8s/client"
@@ -91,6 +92,7 @@ import (
 	"github.com/cilium/cilium/pkg/proxy"
 	"github.com/cilium/cilium/pkg/rate"
 	"github.com/cilium/cilium/pkg/resiliency"
+	"github.com/cilium/cilium/pkg/source"
 	"github.com/cilium/cilium/pkg/statedb"
 	"github.com/cilium/cilium/pkg/sysctl"
 	"github.com/cilium/cilium/pkg/version"
@@ -1855,7 +1857,17 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
 			// (re-)allocate the restored identities before we release them.
 			time.Sleep(option.Config.IdentityRestoreGracePeriod)
 			log.Debugf("Releasing reference counts for %d restored CIDR identities", len(d.restoredCIDRs))
-			d.ipcache.ReleaseCIDRIdentitiesByCIDR(d.restoredCIDRs)
+			// Release the CIDR identity references acquired by
+			// the corresponding AllocateCIDRs() call from when
+			// d.restoredCIDRs was initialized.
+			// Remove the ipcache CIDR associations from the
+			// corresponding UpsertPrefixes() call.
+			d.identityAllocator.ReleaseCIDRs(d.ctx, d.restoredCIDRs)
+			d.ipcache.RemovePrefixes(
+				d.restoredCIDRs,
+				source.Restored,
+				ipcachetypes.NewResourceID(ipcachetypes.ResourceKindDaemon, "", ""),
+			)
 			// release the memory held by restored CIDRs
 			d.restoredCIDRs = nil
 
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -1858,6 +1858,31 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
 			d.ipcache.ReleaseCIDRIdentitiesByCIDR(d.restoredCIDRs)
 			// release the memory held by restored CIDRs
 			d.restoredCIDRs = nil
+
+			wg, toRelease, newIdentities, err := d.dnsNameManager.ReinitializeIdentityReferences(d.ctx)
+			if err != nil {
+				log.WithFields(logrus.Fields{
+					logfields.CIDRS: logging.SampleSlice(d.restoredCIDRs),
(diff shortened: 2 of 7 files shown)


daemon/cmd/daemon.go
```diff
@@ -708,16 +708,25 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
...
-		d.ipcache.UpsertGeneratedIdentities(restoredCIDRidentities, nil)
+	restoredResource := ipcacheTypes.NewResourceID(ipcacheTypes.ResourceKindDaemon, "", "")
+	if len(d.restoredCIDRs) > 0 {
+		// Asynchronously upsert the prefixes into the new ipcache map.
+		// By the time the datapath gets initialized to consume this
+		// map, ipcache should have completed the async work. If this
+		// doesn't occur, it will result in temporary drops in the
+		// time between datapath initialization until this async
+		// process completes.
+		// TODO: Add signalling mechanism to pass to the datapath to
```
lmb, line 706: This should have a name associated with it? Is there a ticket?

pkg/slices/slices.go
```diff
@@ -145,3 +145,19 @@ func SubsetOf[S ~[]T, T comparable](a, b S) (bool, []T) {
 	d := Diff(a, b)
 	return len(d) == 0, d
 }
+
+// NFromMap takes a map of elements and returns a slice of N of them.
+func NFromMap[K comparable, V any](m map[K]V, n uint) []K {
```
lmb: Why not use `x/exp/maps.Keys` and slice the result?

daemon/cmd/daemon_main.go
```diff
@@ -1779,6 +1780,31 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
 			d.ipcache.ReleaseCIDRIdentitiesByCIDR(d.restoredCIDRs)
 			// release the memory held by restored CIDRs
 			d.restoredCIDRs = nil
+
+			wg, toRelease, newIdentities, err := d.dnsNameManager.ReinitializeIdentityReferences(d.ctx)
+			if err != nil {
+				log.WithFields(logrus.Fields{
+					logfields.CIDRS: d.restoredCIDRs[:10],
```
lmb: This will panic if there are less than 10 restoredCIDRs?

Why is it useful to know the first 10 restored CIDRs? If restoring failed due to a specific CIDR it would be better to decorate the error in `ReinitializeIdentityReferences`? If it's not specific to the set of cidrs, why dump this? Same comment for the other error logs.

pkg/fqdn/name_manager.go
```diff
@@ -194,6 +194,29 @@ func (n *NameManager) ForceGenerateDNS(ctx context.Context, namesToRegen []strin
...
+func (n *NameManager) ReinitializeIdentityReferences(ctx context.Context) (wg *sync.WaitGroup, toRelease []*identity.Identity, newlyAllocatedIdentities map[netip.Prefix]*identity.Identity, err error) {
+	n.RWMutex.Lock()
+	defer n.RWMutex.Unlock()
+
+	selectors := make(map[api.FQDNSelector]struct{}, 0)
+	for fqdnSel := range n.allSelectors {
+		selectors[fqdnSel] = struct{}{}
+	}
+
+	// Ignore newly allocated IDs (3rd result) as this is specifically
```
lmb, line 213: This isn't ignoring anything?

daemon/cmd/identity.go
```diff
@@ -92,3 +95,22 @@ func (c cachingIdentityAllocator) AllocateCIDRsForIPs(ips []net.IP, newlyAllocat
...
 }
+
+func (c *cachingIdentityAllocator) ReleaseCIDRs(ctx context.Context, prefixes []netip.Prefix) {
+	for _, prefix := range prefixes {
+		lbls := cidr.GetCIDRLabels(prefix)
+		id := c.LookupIdentity(ctx, lbls)
+		if id == nil {
+			log.WithFields(logrus.Fields{
+				logfields.CIDR: prefix.String(),
+			}).Warning("Restored Identity for CIDR cannot be found")
```
lmb, line 106: Should these be either errors or info? Warning is unspecific, and it seems kinda bad if this happens?

lmb (commented):
Left some meta comments, I don't know the reference counting logic well enough to be able to validate the fix.

daemon/cmd/daemon.go
```diff
@@ -708,16 +708,25 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 		return nil, nil, fmt.Errorf("error while opening/creating BPF maps: %w", err)
 	}
 	// Upsert restored CIDRs after the new ipcache has been opened above
-	if len(restoredCIDRidentities) > 0 {
-		d.ipcache.UpsertGeneratedIdentities(restoredCIDRidentities, nil)
+	restoredResource := ipcacheTypes.NewResourceID(ipcacheTypes.ResourceKindDaemon, "", "")
+	if len(d.restoredCIDRs) > 0 {
+		// Asynchronously upsert the prefixes into the new ipcache map.
+		// By the time the datapath gets initialized to consume this
+		// map, ipcache should have completed the async work. If this
```
squeed, line 702: The ipcache won't synchronize until after all k8s caches have synched -- will that be a problem?

daemon/cmd/daemon_main.go
```diff
@@ -1777,7 +1779,17 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
...
+			// the corresponding AllocateCIDRs() call from when
+			// d.restoredCIDRs was initialized.
+			// Remove the ipcache CIDR associations from the
+			// corresponding UpsertPrefixes() call.
+			d.identityAllocator.ReleaseCIDRs(d.ctx, d.restoredCIDRs)
+			d.ipcache.RemovePrefixes(
+				d.restoredCIDRs,
+				source.Restored,
+				ipcachetypes.NewResourceID(ipcachetypes.ResourceKindDaemon, "", ""),
+			)
```
christarazi, line 1870: I'm not sure I understand why there are effectively two refcount releases happening here. `RemovePrefixes` eventually calls down to `releaseCIDRIdentities`, which already should do what `d.identityAllocator.ReleaseCIDRs` does.

I see that your commit msg addresses this:

> As a matter of process to clearly delineate the codepaths for step (5),
this commit also moves that code into the daemon so that there is no
ambiguity over whether the daemon's identity references are tied to the
ipcache or not.


but we also have the delineation for the daemon via the fact that the daemon has a unique resource ID in the ipcache metadata.

Also how is the above code not a double-free to begin with?

daemon/cmd/daemon.go
christarazi: Hm, that's a good point. AFAIU, a cluster with sufficiently large number of K8s resources will arbitrarily delay CIDR restoration until K8s is synced. I'd guess that our CI is just small enough in terms of K8s cluster resources that Cilium just gets by without the delay being sufficiently long enough to cause policy / connectivity outages in our tests.

daemon/cmd/daemon_main.go
joestringer: The ten aspect was an attempt at bounding the output. But given that this could be a range of CIDRs potentially in the tens, hundreds, maybe even thousands and it's not particularly discriminating in providing specific information that's useful to interpret, maybe it's good enough to skip the variables here.

pkg/fqdn/name_manager.go
joestringer: Dropped the comment, this was a holdover from earlier iteration.

daemon/cmd/identity.go
joestringer: I think you're right. A while back Louis and I tried to collect some feedback on thresholds for using different log levels, and [these guidelines](https://docs.google.com/document/d/142qQ81YQbYysJ_GPQJtVRRnZ-zFCKH_2NsQ-FEnlK7Y/edit#bookmark=id.bsw8c4e8mrss) are where we got to. At some point we should formalize these, but the implications here could be that traffic to certain IPs is inhibited. I think that this could be an error with a prompt for the user to report the bug.

joestringer: I'll wrap this in a sync.Once too, there's no need to spam the logs for every IP.

daemon/cmd/daemon.go
joestringer: If the datapath restores with a partial ipcache that's not fully in sync with k8s, that in itself could trigger connectivity or policy enforcement issues for peers. I don't exactly know what the constraints are today but I think that the datapath should only be replaced after fully syncing with k8s. Whether it does this or not I am not quite sure at this time.

daemon/cmd/daemon_main.go
joestringer: My alternate attempt to solve the problem may reveal some of the answers to this question: https://github.com/cilium/cilium/pull/27327

joestringer: The way I see it is this: `ipcache.Upsert()` lays the responsibility for identity reference counting on the callers of ipcache, while `ipcache.UpsertMetadata()` lays this responsibility on the ipcache itself. Additionally, after commit https://github.com/cilium/cilium/commit/c96b9d81a8a679d77f9929e6ed32559a3d54108d the ipcache no longer properly handled transitioning the ownership of an identity from outside the ipcache into the ipcache.

The changes in this PR+file are decoupling the identity reference counting from the ipcache entirely to ensure that the daemon retains its own references and the ipcache acquires its own references, so there can be up to two different references on each identity during the `identity-restore-grace-period`. As long as those two different subsystems manage their own identity references based on the lifecycle that they expect, everything should be fine. In the case of the daemon, it expects to only keep the reference live during the `identity-restore-grace-period`, and after that time hopefully the rest of the agent is initialized and has taken its own references to keep the identities alive as long as they are used by policies.

There are three different subsystems involved. Here is the behaviour before this PR:

- The daemon would allocate its own reference and do an `ipcache.Upsert()`. The Upsert would not touch the identity's refcount.
- The policy engine for CIDR policy handling would not take a reference on identities, but request ipcache to do so via `ipcache.UpsertMetadata()`.
- The ipcache, while evaluating that request, could take its own reference. However due to c96b9d81a8a679d77f9929e6ed32559a3d54108d it would quickly release that reference, thinking that it already had a reference via the entry inserted by the daemon.
- After `identity-restore-grace-period`, the daemon releases the last references to those Identities, causing them to be freed & removing them from policymaps.

After this PR:

- The daemon would allocate its own reference, then request ipcache to start handling the CIDR via `ipcache.UpsertPrefixes()` (similar to `ipcache.UpsertMetadata()`).
- The policy engine does the same thing as before.
- The ipcache combines the two requests from policy and deamon, and given that there is not yet an ipcache entry, allocates its own reference on the identity that it will keep. The identity now has two live references.
- After `identity-restore-grace-period`, the daemon releases its initial reference, but the ipcache still has its own reference so the CIDR identity stays live.
- Given that the daemon no longer needs the ipcache mapping, it can remove its resource from ipcache, and that will not change the reference counting for the identity because the ipcache still needs to keep the identity alive for the policy user of ipcache.

daemon/cmd/daemon.go
joestringer: Yeah typically we would file a ticket. I'm exploring another option to solve the bug in the short term, so I'll leave this comment unresolved for now and revisit when picking this PR back up.

joestringer:
Thanks for the reviews, I've addressed some of the initial feedback. I think there's still a few steps that need to be investigated a bit more deeply before proceeding with this PR, specifically around some of the ordering constraints as raised by reviews from Lorenz and Casey.

After some discussion with the reviewers and re-evaluating the state of this PR, I think that it's going in the right longer term direction for several aspects, but for now there may be a simpler and more reliable fix in order to resolve the issue for v1.14. I've filed https://github.com/cilium/cilium/pull/27327 to follow up on that, and I'll pursue that option first before returning to this PR.

daemon/cmd/daemon.go
christarazi: Yeah, I guess there are security implications with restoring previous state, especially if the IP <-> identity mappings changed. It is more secure for Cilium to not assume that the state of the world was the same as it was when it was last running because it prevents reuse issues.

joestringer:
I'll go ahead and close this one out in favor of https://github.com/cilium/cilium/pull/28673 . I think that Casey has a bit more available bandwidth at the moment to push the proper solution across the line on this front, and his PR is not burdened with the commit messages assuming one bugfix or another may be applied within the changes due to historic versions of the change. We can take a closer look at any of the remaining tasks from this PR as part of review on that PR.

PR #27267, opened 2023-08-04
bpf: dsr: merge Ingress tail-calls into nodeport_lb*()

julianwiedmann:
The DSR Ingress path was initially implemented through a tail-call for program size reasons. But with the recent improvements to the CT handling, we have sufficient program space to run the DSR ingress handling straight from nodeport_lb*().

This reduces duplicated code (eg. extracting the CT tuple and DSR info twice), avoids additional error handling, and simplifies the overall code flow.


dylandreimerink (approved):
LGTM

PR #27269, opened 2023-08-04
improv: check for k8s backing before running sync

### LINKED ISSUE #27128: cleanup: Don't run `sync-to-k8s-ciliumendpoint` for endpoints which don't need it
During a [PR review](https://github.com/cilium/cilium/pull/26894#discussion_r1277291164) we noticed that we are starting the `sync-to-k8s-ciliumendpoint` controller for host and health endpoints. Those endpoints don't have or need a CiliumEndpoint object, so the controllers are running in vain.

We should not start those controllers for endpoints where we know that they don't need a CEP.

kwakubiney:
This change ensures that the synchronisation controller does not run at all when the endpoint has no `CEP` name, which implies that it also has no pod backing.

Fixes: #27128

### DIFF
--- a/pkg/k8s/watchers/endpointsynchronizer.go
+++ b/pkg/k8s/watchers/endpointsynchronizer.go
@@ -75,6 +75,14 @@ func (epSync *EndpointSynchronizer) RunK8sCiliumEndpointSync(e *endpoint.Endpoin
 		return
 	}
 
+	// The CEP name is derived from the K8sPodName and K8sNamespace.
+	// They should always be available if an endpoint belongs to a pod.
+	cepName := e.GetK8sCEPName()
+	if cepName == "" {
+		scopedLog.Debug("Skipping CiliumEndpoint update because it has no k8s cep name")
+		return
+	}
+
 	var (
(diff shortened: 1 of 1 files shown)


pkg/k8s/watchers/endpointsynchronizer.go
```diff
@@ -95,14 +103,6 @@ func (epSync *EndpointSynchronizer) RunK8sCiliumEndpointSync(e *endpoint.Endpoin
 					return fmt.Errorf("Kubernetes apiserver is not available")
```
squeed, line 103: Can this condition (k8s connectivity) change as well, or can we also hoist it out of the controller?

squeed (approved):
Looks good, just one possible small improvement. Thanks!

pkg/k8s/watchers/endpointsynchronizer.go
kwakubiney: I'd look at it but looks like PR has been merged?

squeed: Hah! Nevermind :-)

Issue #27268, opened 2023-08-04
CI: linuxPrivilegedIPv4AndIPv6TestSuite/TestNodeChurnXFRMLeaks


giorio94:
## CI failure

Hit on https://github.com/cilium/cilium/pull/27242: https://github.com/cilium/cilium/actions/runs/5752260375/job/15592941475

```
--- FAIL: Test (227.53s)
    --- FAIL: Test/linuxPrivilegedIPv4AndIPv6TestSuite (3.51s)
        --- FAIL: Test/linuxPrivilegedIPv4AndIPv6TestSuite/TestNodeChurnXFRMLeaks (1.06s)
            node_linux_test.go:809: 
                ... value *errors.joinError = &errors.joinError{errs:[]error{(*errors.joinError)(0xc0010e0180)}} ("failed to enable ipsec with out IPv6 using local IP ::/0, rem 2001:aaaa::/64, spi 0: unable to replace remote state: file exists")
```

julianwiedmann:
@pchaigno @tommyp1ckles FYI - I reproduced locally, this fails when https://github.com/cilium/cilium/pull/27212 and https://github.com/cilium/cilium/pull/25994 are put together.

brb:
Looks like https://github.com/cilium/cilium/pull/25994 helped to expose an error which previously was visible only in logs.

brb:
The test tries to insert the following XFRM states:

1. `destinationIP="<nil>" sourceIP=1.1.1.1 spi=6` 
2. `destinationIP="<nil>" sourceIP="face::1" spi=6`

The second conflicts with the first one, thus the test failure.

PR #27275, opened 2023-08-04
maps: do not depend on global variable to initialize CT maps

giorio94:
Currently, the creation of a new CT map depends on the `mapInfo` global variable. Yet, this is suboptimal, as it always requires that variable to be initialized beforehand, which might be tricky in combination with the Hive framework. Additionally, all the necessary information is either known at compile time or depends on user configurations only. Hence, let's update that logic to remove the dependency on the global variable.

### DIFF
--- a/pkg/datapath/loader/loader_test.go
+++ b/pkg/datapath/loader/loader_test.go
@@ -21,10 +21,8 @@ import (
 	"github.com/cilium/cilium/pkg/datapath/loader/metrics"
 	"github.com/cilium/cilium/pkg/elf"
 	"github.com/cilium/cilium/pkg/maps/callsmap"
-	"github.com/cilium/cilium/pkg/maps/ctmap"
 	"github.com/cilium/cilium/pkg/maps/policymap"
 	"github.com/cilium/cilium/pkg/node"
-	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/testutils"
 )
 
@@ -77,8 +75,6 @@ func (s *LoaderTestSuite) SetUpSuite(c *C) {
 		}
 		return os.RemoveAll(tmpDir)
 	}
-
-	ctmap.InitMapInfo(option.CTMapEntriesGlobalTCPDefault, option.CTMapEntriesGlobalAnyDefault, true, true, true)
 
 	SetTestIncludes([]string{
 		fmt.Sprintf("-I%s", bpfDir),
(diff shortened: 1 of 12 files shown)


giorio94:
Cilium Runtime failure is https://github.com/cilium/cilium/issues/27268

pkg/maps/ctmap/types_test.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+func TestMapKey(t *testing.T) {
+	for mapType := mapType(0); mapType < mapTypeMax; mapType++ {
+		assert.NotNil(t, mapType.key())
+	}
+}
+
+func TestMapBPFDefine(t *testing.T) {
+	for mapType := mapType(0); mapType < mapTypeMax; mapType++ {
```
asauber, line 24: A test for the panic case as well would be useful here

pkg/datapath/loader/loader_test.go
```diff
@@ -78,8 +76,6 @@ func (s *LoaderTestSuite) SetUpSuite(c *C) {
 		return os.RemoveAll(tmpDir)
 	}
 
-	ctmap.InitMapInfo(option.CTMapEntriesGlobalTCPDefault, option.CTMapEntriesGlobalAnyDefault, true, true, true)
```
rgo3, line 81: Why were we calling this function here? Were calls to `InitMapInfo()` in some of the test suites unnecessary in the first place or do the changes in this PR enable us to delete them?

giorio94: These calls are removed thanks to the changes in this PR. For instance, the [`TestWrap` ](https://github.com/cilium/cilium/blob/main/pkg/datapath/loader/template_test.go#L16-L44) test calls [`WriteBPFMacros`](https://github.com/cilium/cilium/blob/main/pkg/maps/ctmap/ctmap.go#L807-L819), which in turn create the [map objects](https://github.com/cilium/cilium/blob/main/pkg/maps/ctmap/ctmap.go#L333-L346), historically requiring the global variable to be initialized. But it is no longer the case now that the map objects creation doesn't depend on that global variable.

rgo3: Ah, thanks for the context :pray:.

rgo3 (approved):

asauber (approved):
LGTM

giorio94:
Could someone from @cilium/endpoint please take a look at this PR?

christarazi (approved):

giorio94:
Codeowner reviews are covered, and checks are green. Marking as ready to merge

ldelossa (approved):

giorio94:
Rebased onto main as GH was complaining about a conflict.

giorio94:
Tests are green. Marking as ready-to-merge again.

PR #27281, opened 2023-08-04
egressgw: detect conflicting configurations in ENI mode

julianwiedmann:
In ENI mode the manager creates EgressGW-specific routing tables for each selected interface. Currently we create one default route in this table, that sends all traffic to its nexthop.

If multiple policies with different egressIPs (== different nexthop) select the same interface, the resulting default routes would collide.

Detect such a configuration by tracking what egressIP is used on each interface, and reject if the egressIPs would differ.

### DIFF
--- a/pkg/egressgateway/manager_privileged_test.go
+++ b/pkg/egressgateway/manager_privileged_test.go
@@ -46,6 +46,7 @@ const (
 
 	ep1IP = "10.0.0.1"
 	ep2IP = "10.0.0.2"
+	ep3IP = "10.0.0.3"
 
 	destCIDR      = "1.1.1.0/24"
 	excludedCIDR1 = "1.1.1.22/32"
@@ -55,6 +56,8 @@ const (
 	egressCIDR1 = "192.168.101.1/24"
 	egressIP2   = "192.168.102.1"
 	egressCIDR2 = "192.168.102.1/24"
+	egressIP3   = "192.168.103.1"
+	egressCIDR3 = "192.168.103.1/24"
 
 	zeroIP4 = "0.0.0.0"
 
@@ -66,6 +69,7 @@ const (
 var (
 	ep1Labels = map[string]string{"test-key": "test-value-1"}
 	ep2Labels = map[string]string{"test-key": "test-value-2"}
+	ep3Labels = map[string]string{"test-key": "test-value-3"}
 
 	identityAllocator = testidentity.NewMockIdentityAllocator(nil)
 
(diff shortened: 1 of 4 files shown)


pkg/egressgateway/manager_privileged_test.go
```diff
@@ -237,6 +241,50 @@ func (k *EgressGatewayTestSuite) TestEgressGatewayManager(c *C) {
...
+	// that selects this IP. The conflicting policy should be ignored,
+	// resulting in a "no gateway found" egress entry and no IP rule
+	// for ep3.
+	link, err := netlink.LinkByName(testInterface1)
+	if err != nil {
+		panic(err)
+	}
+
+	a, _ := netlink.ParseAddr(egressCIDR3)
+	netlink.AddrAdd(link, a)
```
lmb: I think this can fail?

pkg/egressgateway/manager_privileged_test.go
```diff
@@ -237,6 +241,50 @@ func (k *EgressGatewayTestSuite) TestEgressGatewayManager(c *C) {
...
+
+	addPolicy(c, k.policies, &policyParams{
```
lmb: Nit: you can assign `policyParams` to `policy3 :=` or similar and cut down on the repetition to delete the policy.

pkg/egressgateway/manager_privileged_test.go
```diff
@@ -237,6 +241,50 @@ func (k *EgressGatewayTestSuite) TestEgressGatewayManager(c *C) {
...
+	// Delete the conflicting policy and IP again.
+	deletePolicy(c, k.policies, &policyParams{
+		name:            "policy-3",
+		endpointLabels:  ep3Labels,
+		destinationCIDR: destCIDR,
+		nodeLabels:      nodeGroup1Labels,
+		egressIP:        egressIP3,
+	})
+
+	netlink.AddrDel(link, a)
```
lmb: Same, can fail?

pkg/egressgateway/policy.go
```diff
@@ -115,14 +115,34 @@ func (config *PolicyConfig) regenerateGatewayConfig(manager *Manager) {
...
+			// We can only have one default route per interface. Enforce that all
+			// policies on an interface use the same EgressIP (== default route).
+			currentPolicy := policyByInterfaceIndex[gwc.ifaceIndex]
+			if currentPolicy == nil {
+				policyByInterfaceIndex[gwc.ifaceIndex] = config
+				break
+			}
+
+			egressIP := currentPolicy.gatewayConfig.egressIP.String()
+			if egressIP != gwc.egressIP.String() {
```
lmb: Why not use https://pkg.go.dev/net#IP.Equal?

julianwiedmann: Awkwardly it's a `net.IPNet`. So I was probably just too lazy to write up the proper comparison - done so now.

julianwiedmann:
Ta! Pushed with (hopefully) everything addressed.

lmb (approved):

PR #27283, opened 2023-08-04
.github: add workflow to track replied issues

aanm:
With this GH workflow we will be able to track which GH issues were replied by the users after we have requested a "need-more-info" label.

Once the user replies to the GH issue, the "need-more-info" label is removed and the "info-completed" label will be added.

### DIFF
--- a/.github/workflows/needs-more-info.yaml
+++ b/.github/workflows/needs-more-info.yaml
@@ -1,0 +1,18 @@
+name: Issue Info Complete
+
+# This workflow is triggered on issue comments.
+on:
+  issue_comment:
+    types: created
+
+jobs:
+  applyNeedsAttentionLabel:
+    name: Apply Info Complete Label
+    runs-on: ubuntu-latest
+    steps:
+      - name: Apply Needs Attention Label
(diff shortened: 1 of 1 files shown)


.github/workflows/needs-more-info.yaml
```diff
@@ -0,0 +1,18 @@
...
+  applyNeedsAttentionLabel:
+    name: Apply Info Complete Label
+    runs-on: ubuntu-latest
+    steps:
+      - name: Apply Needs Attention Label
+        uses: hramos/needs-attention@v1
+        with:
+            repo-token: ${{ secrets.GITHUB_TOKEN }}
+            response-required-label: 'need-more-info'
+            needs-attention-label: 'info-completed'
```
joestringer, line 18: LGTM, did you consider just adding needs/triage back?

joestringer (approved):

.github/workflows/needs-more-info.yaml
aanm: I did not but the needs/triage might be misleading since an issue could already be triaged but we still need more info? :thinking:

PR #27285, opened 2023-08-04
BGP-CP: Remove NodeSpecer and ControlPlaneState

ldelossa:
This PR removes the NodeSpecer abstraction and the ControlPlaneState structure from the BGP control plane. 

Now that LocalNodeStore is available it can be used in place of these two constructs. 

Additionally, since all ConfigReconcilers are now part of the Hive/Cell dependency injection graph, each ConfigReconciler can declare their own depedencies at time of construction. This provides much greater flexibility as a ConfigReconciler can now utilize any entities available in the DIG to perform their necessary reconciliation. 

```release-note
Remove NodeSpecer and ControlPlaneState from BGP-CP. Rely on Hive/Cell for further ConfigReconciler dependencies. 
```


pkg/bgpv1/agent/controller.go
```diff
@@ -352,46 +353,9 @@ func (c *Controller) Reconcile(ctx context.Context) error {
 		return fmt.Errorf("invalid BGP peering policy %s: %w", policy.Name, err)
 	}
 
-	// parse any virtual router specific attributes defined on this node via
```
ldelossa, line 355: This whole block goes away, since all this information can be gained just from looking at 'localNode'

joestringer:
Is this changing a user-facing interface or will it have user-facing consequences? If yes, release note seems fine though I'm not sure whether users will understand it. If no, this should probably be labelled with `release-note/misc`. Note that even if it's misc, that's not saying the PR is any less valuable, it's just that we use the major/minor/bug/misc as a way to order the release notes from most user-facing and prominent for all users to know about to least user-facing and prominent.

ldelossa:
@joestringer no user facing changes. I'll update the label to misc.

pkg/bgpv1/agent/controller.go
```diff
@@ -217,7 +172,12 @@ func (c *Controller) Run(ctx context.Context) {
 			"component": "Controller.Run",
 		})
 	)
-	l.Debug("Starting informers")
+	l.Info("Starting LocalNodeStore Observer")
+
+	// setup a reconciliation trigger on LocalNodeStore changes
+	c.LocalNodeStore.Observe(ctx, func(node node.LocalNode) { c.Sig.Event(struct{}{}) }, func(err error) {
+		l.WithError(err).Info("LocalNodeStore observe has yielded. Reconciliation with no longer be triggered for LocalNode changes")
```
rastislavs:
```suggestion
		l.WithError(err).Info("LocalNodeStore observe has yielded. Reconciliation will no longer be triggered for LocalNode changes")
```

pkg/bgpv1/manager/manager.go
```diff
@@ -201,18 +202,26 @@ func (m *BGPRouterManager) registerBGPServer(ctx context.Context, c *v2alpha1api
...
 		}
 	}
 
-	routerID, err := cstate.ResolveRouterID(c.LocalASN)
+	routerID, err := annoMap.ResolveRouterID(c.LocalASN)
 	if err != nil {
-		return err
+		nodeIP := node.GetNodeIP(false)
+		if nodeIP.IsUnspecified() {
+			return fmt.Errorf("failed to resolve router id")
```
rastislavs: Nit: maybe wrap the error returned from `ResolveRouterID` here?

pkg/bgpv1/agent/annotations.go
```diff
@@ -103,6 +104,15 @@ func (e ErrMulti) Error() string {
 	return s.String()
 }
 
+func (a AnnotationMap) ResolveRouterID(localASN int64) (string, error) {
+	if _, ok := a[localASN]; ok {
+		if parsed, err := netip.ParseAddr(a[localASN].RouterID); err == nil && !parsed.IsUnspecified() {
+			return parsed.String(), nil
+		}
```
rastislavs, line 113: Nit: Would be nicer to return a wrapped error from `ParseAddr` if it returns an error, so that in case that incorrect address is provided in the annotation, you would get a parse error instead of "router id not specified by annotation"

rastislavs (commented):
found just some nits, overall a great change!

pkg/bgpv1/agent/controller.go
```diff
@@ -217,7 +213,12 @@ func (c *Controller) Run(ctx context.Context) {
 			"component": "Controller.Run",
 		})
 	)
-	l.Debug("Starting informers")
+	l.Info("Starting LocalNodeStore Observer")
+
+	// setup a reconciliation trigger on LocalNodeStore changes
+	c.LocalNodeStore.Observe(ctx, func(node node.LocalNode) { c.Sig.Event(struct{}{}) }, func(err error) {
```
harsimran-pabla, line 178: Should this be part of Controller.Start,? 

Since policy-observer is defined there, it would also make sense to have node-observer in it.

pkg/bgpv1/agent/controller.go
```diff
@@ -217,7 +213,12 @@ func (c *Controller) Run(ctx context.Context) {
...
+		l.WithError(err).Info("LocalNodeStore observe has yielded. Reconciliation with no longer be triggered for LocalNode changes")
```
harsimran-pabla: Should this be error log ? `l.WithError(err).Info -> l.WithError(err).Error`

pkg/bgpv1/agent/controller.go
ldelossa: fixed

pkg/bgpv1/manager/manager.go
ldelossa: fixed

pkg/bgpv1/agent/annotations.go
ldelossa: fixed

pkg/bgpv1/manager/manager.go
```diff
@@ -201,18 +202,26 @@ func (m *BGPRouterManager) registerBGPServer(ctx context.Context, c *v2alpha1api
...
 	}
 
-	routerID, err := cstate.ResolveRouterID(c.LocalASN)
+	routerID, err := annoMap.ResolveRouterID(c.LocalASN)
 	if err != nil {
-		return err
+		nodeIP := node.GetNodeIP(false)
+		if nodeIP.IsUnspecified() {
+			return fmt.Errorf("failed to resolve router id: %w", err)
+		}
```
harsimran-pabla, line 224: Missing assignment of nodeIP to routerID 
225: `routerID = nodeIP.String()`

pkg/bgpv1/manager/reconcile.go
```diff
@@ -100,78 +92,86 @@ func preflightReconciler(
...
 	}
 
-	routerID, err := cstate.ResolveRouterID(newc.LocalASN)
+	routerID, err := annoMap.ResolveRouterID(p.DesiredConfig.LocalASN)
 	if err != nil {
-		return err
+		nodeIP := p.Node.GetNodeIP(false)
+		if nodeIP.IsUnspecified() {
+			return fmt.Errorf("failed to resolve router id")
+		}
```
harsimran-pabla, line 126: Missing routerID = nodeIP.String()

pkg/bgpv1/agent/annotations.go
```diff
@@ -103,6 +104,15 @@ func (e ErrMulti) Error() string {
 	return s.String()
 }
 
+func (a AnnotationMap) ResolveRouterID(localASN int64) (string, error) {
+	if _, ok := a[localASN]; ok {
+		if parsed, err := netip.ParseAddr(a[localASN].RouterID); err == nil && !parsed.IsUnspecified() {
+			return parsed.String(), fmt.Errorf("failed to parse RouterID for local ASN %v: %w", localASN, err)
```
harsimran-pabla: Something is not right here, on err == nil, we should not be returning err.

pkg/bgpv1/agent/controller.go
```diff
@@ -217,7 +172,12 @@ func (c *Controller) Run(ctx context.Context) {
 			"component": "Controller.Run",
 		})
 	)
-	l.Debug("Starting informers")
+	l.Info("Starting LocalNodeStore Observer")
+
+	// setup a reconciliation trigger on LocalNodeStore changes
+	c.LocalNodeStore.Observe(ctx, func(node node.LocalNode) { c.Sig.Event(struct{}{}) }, func(err error) {
```
harsimran-pabla, line 178: General observation, observe is not triggering event on changes to annotations in node object.

pkg/bgpv1/agent/controller.go
ldelossa: I'm not sure it matters, as long as we are observing before we enter the `run` loop. 

But maybe it's best to do it there for consistency?

pkg/bgpv1/agent/controller.go
ldelossa: Well, we get an error, but most likely, the system stopped the observer for some valid reason, which is probably not an error?

pkg/bgpv1/agent/controller.go
ldelossa: You're right, issue is here: https://github.com/ldelossa/cilium/blob/ldelossa/bgpv1-configreconciler-hivecell/pkg/endpointmanager/host.go#L70

We do not trigger any update to the LocalNodeStore on k8s node updates.

ldelossa: Fixed with latest commit.

pkg/bgpv1/agent/controller.go
harsimran-pabla: Yeah, it was mostly for consistency since Observe does not need to be started in its own goroutine.

ldelossa: @harsimran-pabla - I attempted this change, but it seems to have broken the Observer. I no longer see events if I place it in the Start hook. 

I'm inclined to leave it as is for now and Ill investigate it a bit further once a separate PR

harsimran-pabla: ack, no issues.

YutaroHayakawa (approved):
LGTM overall 👍 Thanks for your big effort! This makes code much simpler and easy to change.

One non-blocking suggestion. How about splitting each reconcilers into separated files? That makes the code easy to read, reduce the risk of conflict when we introduce a new reconciler, etc.

rastislavs (approved):
LGTM

pkg/endpointmanager/host.go
```diff
@@ -61,17 +63,37 @@ func (mgr *endpointManager) OnUpdateNode(oldNode, newNode *slim_corev1.Node,
...
-	}
+	oldNodeAnnotations := oldNode.GetAnnotations()
+	newNodeAnnotations := newNode.GetAnnotations()
 
-	node.SetLabels(newNodeLabels)
+	labelsEqual := comparator.MapStringEquals(oldNodeLabels, newNodeLabels)
+	annotationsEqual := comparator.MapStringEquals(oldNodeAnnotations, newNodeAnnotations)
+
+	// if labels have changed we need to recompute security ID.
+	if labelsEqual {
```
harsimran-pabla: Should this be `if !labelsEqual` ?

ldelossa: Yup! Good call.

harsimran-pabla (approved):

nebril (approved):

danehans:
I have a local rebase of https://github.com/cilium/cilium/pull/27100 with this PR. @joamaki @aditighag PTAL and merge when you have a moment.

aditighag:
I don't have much context for the endpoint related changes. As this is a BGP focused change, feel free to go ahead with the merge based on the existing reviews.

ldelossa:
@aditighag i think endpoint team was required due to the first and last commit, which arent exactly BGP related, rather, we now trigger LocalNodeStore updates on annotation changes. Can you review just those two commits?

pkg/endpointmanager/host.go
```diff
@@ -61,17 +63,37 @@ func (mgr *endpointManager) OnUpdateNode(oldNode, newNode *slim_corev1.Node,
...
-	if nodeEP == nil {
-		log.Error("Host endpoint not found")
-		return nil
-	}
+	oldNodeAnnotations := oldNode.GetAnnotations()
+	newNodeAnnotations := newNode.GetAnnotations()
 
-	node.SetLabels(newNodeLabels)
+	labelsEqual := comparator.MapStringEquals(oldNodeLabels, newNodeLabels)
+	annotationsEqual := comparator.MapStringEquals(oldNodeAnnotations, newNodeAnnotations)
```
aditighag, line 70: There is a factory function available to compare annotations - https://github.com/cilium/cilium/blob/main/pkg/k8s/factory_functions.go#L67. Any reason why we don't use it here? The watcher subsystem uses the same function.

pkg/endpointmanager/host.go
```diff
@@ -61,17 +63,37 @@ func (mgr *endpointManager) OnUpdateNode(oldNode, newNode *slim_corev1.Node,
...
+	// better to perform one update if multiple attributes change.
+	if !labelsEqual && !annotationsEqual {
+		node.SetMultiAttributes(func(n *node.LocalNode) {
+			n.Labels = newNodeLabels
+			n.Annotations = newNodeAnnotations
+		})
+	} else if !labelsEqual {
+		node.SetLabels(newNodeLabels)
+	} else if !annotationsEqual {
+		node.SetAnnotations(newNodeAnnotations)
```
aditighag, line 96: Can we pass the `labelsEqual` and `annotationsEqual` flags to `SetMultiAttribues`, and fold the if-else conditions?

pkg/k8s/watchers/node.go
```diff
@@ -34,7 +34,8 @@ type NodeUpdate interface {
 }
 
 func nodeEventsAreEqual(oldNode, newNode *slim_corev1.Node) bool {
-	return comparator.MapStringEquals(oldNode.GetLabels(), newNode.GetLabels())
+	return comparator.MapStringEquals(oldNode.GetLabels(), newNode.GetLabels()) &&
+		comparator.MapStringEquals(oldNode.GetAnnotations(), newNode.GetAnnotations())
```
aditighag, line 38: `nodeEventsAreEqual` is being called in a generic event loop that tracks changes to the node ep, and notifies the subscribers of node events. Your commit message only talks about the annotations update being tracked by the BGP subsystem, so how do we know that the updated function doesn't cause any regressions for other subsystems that are listening to node events?

aditighag (changes requested):
Thanks for highlighting the commits that are not BGP specific. I've left some review comments. Mainly, it's not clear to me how some of the changes may affect other subsystems subscribing to node events.

pkg/k8s/watchers/node.go
ldelossa: My presumed consequence of this change is that all existing code works as normal, but may now also receive an event when the k8s Node object associated with the current Cilium agent has its annotations changed. 

I'm not capable to testing every sub-system, but I do rely on tests (which are all passing) to obtain confidence in this change. 

I felt it was safe to assume that sub-systems that are listening for NodeEvents would assume the same semantics as a change to the k8s Node object itself which fires for **any** change? 

Do you feel there maybe an issue for a sub-system to receive a NodeEvent on both labels and annotations changes?

pkg/endpointmanager/host.go
ldelossa: That would make SetAnnotations pretty "specific", no? Right now the method doesn't care 'why' you want to set multiple attributes, its just an interface to do so. This seems appropriate to me? Since it may be used in a completely different setting for a completely different evaluation of Node fields.

pkg/endpointmanager/host.go
ldelossa: Good call, no reason here, I just stumbled upon the `comparator` pkg in the Watcher code before I actually saw that function.

pkg/k8s/watchers/node.go
aditighag: >I'm not capable to testing every sub-system, but I do rely on tests (which are all passing) to obtain confidence in this change.

Yes, that's fair. But the PR doesn't seem to add a test that updates node annotations, so how the other subsystems behave when they receive additional events is unknown?

>Do you feel there maybe an issue for a sub-system to receive a NodeEvent on both labels and annotations changes?

I can't say definitively, but I looked at the service cache subscriber, and it seems to assume that updates only pertain to labels - https://github.com/cilium/cilium/blob/main/pkg/k8s/service_cache.go#L739-L739. 

I think it's worthwhile to bring this up in the community meeting, or in the public Slack channel (maybe just as a heads-up). Wdyt?

ldelossa: Yup I do agree. Our sig was actually bringing up your point that some sub-systems may think an update **implicitly** indicates that only labels are updated and whether this is an issue or not. 

Bringing it up in next comm meeting sounds like a good idea and I'll add a message in public Slack before hand.

ldelossa: @aditighag I went through each implementation of the `subscriber.Node` interface to confirm that an event for annotations will not cause an issue. 

Following Chris's response in this thread: https://cilium.slack.com/archives/C2B917YHE/p1692050432666799, I think this change is rather safe. 

I do have a few other questions around our node-syncing design, but they are a bit orthogonal to this PR, and will be doing some investigating further.

pkg/endpointmanager/host.go
ldelossa: Looking at this again, I think I do prefer to use the `comparator` package here. 

1. There is no comparable `LabelsEqual` functions, so I'd be importing `comparator` anyway.
2. I don't need to filter the annotations for comparison.
3. The watcher package also uses `comparator` internally as well.

aditighag: Well, there is a difference in the implementation. The factory function checks if a specific annotation is equal or not. Maybe you need MapStringEquals here for comparing all annotations no a node.

pkg/endpointmanager/host.go
aditighag: If `MultiSetAttributes` were used in multiple places, that might've made sense. But only two attributes are being set here depending on the labels and annotations equal flags. Anyway it's your call.

pkg/k8s/watchers/node.go
aditighag: Thanks for checking! It would be nice to have a test that mutates node annotations: something to keep in mind for follow-ups.

aditighag (approved):

pkg/endpointmanager/host.go
ldelossa: Yes I think so, at this level, we simply want to know if any annotation has changed between the old and the new node.

pkg/k8s/watchers/node.go
ldelossa: I'm def not opposed to this, but I am having a bit of trouble understanding where that test would live and what is under test in that scenario. 

Is it, testing each sub-system can handle an event where a node annotation is updated or are you more thinking a test within the watcher code itself?

pkg/endpointmanager/host.go
ldelossa: Yes, I do think we *need* an opinionated way to set multiple values on LocalNode without causing LocalNodeStore subscribers to wake up multiple times.

pkg/k8s/watchers/node.go
aditighag: > testing each sub-system can handle an event where a node annotation is updated 

^^ Yes, mainly this. Maybe the control plane test framework is well suited to test such a functionality?

PR #27299, opened 2023-08-05
bpf: nodeport: update TTL / hop-limit on redirect

julianwiedmann:
https://github.com/cilium/cilium/pull/23884 and https://github.com/cilium/cilium/pull/24422 introduced some inconsistencies and bugs in how the TTL / hop-limit field is updated when a packet is redirected from the ingress path to a remote destination. Fix them by adding `ipv*_l3()` calls, or shuffling the existing calls into the correct place.

```release-note
Fix several paths in the North-South load-balancer where the TTL / hop-limit field of a forwarded packet was not updated.
```

### DIFF
--- a/bpf/lib/nodeport.h
+++ b/bpf/lib/nodeport.h
@@ -2882,6 +2882,12 @@ nodeport_rev_dnat_fwd_ipv4(struct __ctx_buff *ctx, struct trace_ctx *trace,
  */
 static __always_inline int rev_nodeport_lb4(struct __ctx_buff *ctx, __s8 *ext_err)
 {
+	struct bpf_fib_lookup_padded fib_params = {
+		.l = {
+			.family		= AF_INET,
+			.ifindex	= ctx_get_ifindex(ctx),
+		},
+	};
 	enum trace_reason __maybe_unused reason = TRACE_REASON_UNKNOWN;
 	int ifindex = 0, ret, l3_off = ETH_HLEN, l4_off;
 	struct ipv4_ct_tuple tuple = {};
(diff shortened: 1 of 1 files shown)


bpf/lib/nodeport.h
```diff
@@ -2535,8 +2548,10 @@ int tail_nodeport_nat_egress_ipv4(struct __ctx_buff *ctx)
 		goto drop_err;
 	}
 
-	ret = fib_redirect_v4(ctx, ETH_HLEN, ip4, true, &ext_err,
-			      ctx_get_ifindex(ctx), &oif);
```
jschwinger233, line 2539: > tail_nodeport_nat_egress_ipv4() potentially adds outer tunnel headers

That's right, just 10 lines above we have a `nodeport_add_tunnel_encap()` followed by `ctx_redirect()`.

> The call to fib_redirect_v4() then decrements outer header's TTL

I have a question for this. In the case of tunneling, we'll not reach here (`fib_redirect_v4()`) because the packet has been `bpf_redirect()`-ed to cilium_vxlan at 2529L, so the outer header has no chance to get TTL decremented in the `fib_redirect_v4()`; sure the inner ip4 doesn't update its TTL either, but do we need to update TTL for a `bpf_redirect()`?

jschwinger233 (commented):
One question for the first commit "fix TTL update for XDP in IPv4 NAT Egress path", thanks :fearful:

bpf/lib/nodeport.h
julianwiedmann: > I have a question for this. In the case of tunneling, we'll not reach here (`fib_redirect_v4()`) because the packet has been `bpf_redirect()`-ed to cilium_vxlan at 2529L, so the outer header has no chance to get TTL decremented in the `fib_redirect_v4()`

For XDP it doesn't get redirected to cilium_vxlan (there's no XDP TX / REDIRECT support in the vxlan driver). Instead ctx_set_encap_info() manually builds the headers and returns ifindex 0, so we do a FIB lookup & redirect based on the new outer headers.

jschwinger233 (approved):

PR #27304, opened 2023-08-06
ingress: Add conformance test for KPR=false

sayboras:
This is to test the combination of kubeProxyReplacement=false, but with nodePort.enabled=true.


viktor-kurchenko (approved):
LGTM.

qmonnet (approved):

PR #27309, opened 2023-08-07
gateway-api: Filter routes based on Section Name and port

sayboras:
### Description

Please refer to individual commit for more details.

### DIFF
--- a/operator/pkg/model/translation/gateway-api/translator.go
+++ b/operator/pkg/model/translation/gateway-api/translator.go
@@ -4,6 +4,7 @@
 package gateway_api
 
 import (
+	"crypto/sha256"
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


operator/pkg/model/translation/gateway-api/translator.go
```diff
@@ -90,7 +90,7 @@ func getService(resource *model.FullyQualifiedResource, allPorts []uint32) *core
 
 	return &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      ciliumGatewayPrefix + resource.Name,
+			Name:      shorten(ciliumGatewayPrefix + resource.Name),
```
aanm, line 94: We need to think about name collisions before shortening this name. Discussed offline, we could some hash to help out in here.

aanm (changes requested):

operator/pkg/model/translation/gateway-api/translator.go
sayboras: thanks :100:

aanm (approved):

mhofstetter (approved):

sayboras:
Reviews from required teams are in, merging this in.

PR #27313, opened 2023-08-07
bpf: avoid calculating L4 offset

julianwiedmann:
Clean up some `ipv6_hdrlen()` calls, which can be replaced by just passing the available L4 offset through.

Each `ipv6_hdrlen()` call is worth around 150 instructions, so this nicely helps with reducing program size.


jschwinger233 (approved):
Maybe PR title isn't 100% accurate, because we also reduced call of ipv4_hdrlen().

julianwiedmann:
> Maybe PR title isn't 100% accurate, because we also reduced call of ipv4_hdrlen().

Ah, true. Only for consistency though, no big savings there unfortunately. I've adjusted the title :+1:.

PR #27319, opened 2023-08-07
ipsec: fixes for key rotation

jrfastab:
Fixups to ensure keys are correctly in place before deleting the old key.

1. Instantiate all xfrm rules when new key shows up instead of waiting for nodeUpdate events through validator to happen. Its not always the case on larger clusters that the nodeUpdate logic will trigger before the key retention logic triggers to delete old keys. In this case we may be missing xfrm rules for new keys for some nodes.

2. Ensure rules are in place before key is updated. Current logic updates datapath key before updating the BPF maps creating smallish window where pkts can be marked for the new key but the BPF maps and datapath wont have the mapping to do this yet resulting in dropped pkts.

Both of these intend to address an increased policy block counter in recent versions.
```release-note
Fix bug that could cause packet drops of type XfrmOutPolBlock while rotating the IPsec key.
```

### DIFF
--- a/pkg/datapath/linux/ipsec/ipsec_linux.go
+++ b/pkg/datapath/linux/ipsec/ipsec_linux.go
@@ -767,7 +767,6 @@ func LoadIPSecKeysFile(path string) (int, uint8, error) {
 func LoadIPSecKeys(r io.Reader) (int, uint8, error) {
 	var spi uint8
 	var keyLen int
-	scopedLog := log
 
 	ipSecLock.Lock()
 	defer ipSecLock.Unlock()
@@ -885,11 +884,17 @@ func LoadIPSecKeys(r io.Reader) (int, uint8, error) {
 		ipSecKeysRemovalTime[oldSpi] = time.Now()
 		ipSecCurrentKeySPI = spi
 	}
+	return keyLen, spi, nil
+}
+
+func SetIPSecSPI(spi uint8) error {
+	scopedLog := log
+
 	if err := encrypt.MapUpdateContext(0, spi); err != nil {
 		scopedLog.WithError(err).Warn("cilium_encrypt_state map updated failed:")
-		return 0, 0, err
+		return err
 	}
-	return keyLen, spi, nil
+	return nil
 }
 
 // DeleteIPsecEncryptRoute removes nodes in main routing table by walking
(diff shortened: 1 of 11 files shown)


pkg/datapath/linux/node.go
```diff
@@ -1905,6 +1905,16 @@ func (n *linuxNodeHandler) NodeValidateImplementation(nodeToValidate nodeTypes.N
 	return n.nodeUpdate(nil, &nodeToValidate, false)
 }
 
+// AllNodeValidateImplementation is called to validate the implementation of the
+// node in the datapath for all existing nodes
+func (n *linuxNodeHandler) AllNodeValidateImplementation() {
+	n.mutex.Lock()
+	defer n.mutex.Unlock()
```
joestringer, line 1913: Are there any subtleties to this new locking pattern? Seems like this is triggered from an independent goroutine responsible for watching the files so it's likely somewhat tolerant of delays, and it could be blocked by other users of the linuxNodeHandler lock. the `n.nodeUpdate()` call below seems like it's more towards the leaf end of processing so not likely to introduce any risk of deadlock or calling back up the call hierarchy right?

joestringer (commented):
I briefly looked over the changes and nothing alarming stood out. I'd prefer someone with more familiarity with this part of the codebase to take a closer look.

pkg/datapath/linux/node.go
jrfastab: My thinking is it is the same pattern (mutex.Lock()/defer) used by nodeAdd() to call nodeUpdate() so needs to be safe to have the mutex lock held while calling nodeUpdate. And then this is called from its own go routine as you note and we are fine to wait for a long while here if needed. As long as its resolved by the time we remove the old keys.

One thought would be maybe we want to add a check for nodeInitialized? But I believe we should initialized the node before calling this so maybe doesn't matter really.

pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -937,7 +937,7 @@ func keyfileWatcher(ctx context.Context, watcher *fswatcher.Watcher, keyfilePath
 			// nodeUpdate(), which is responsible for updating the
 			// IPSec policies and states for all the different EPs
 			// with ipsec.UpsertIPsecEndpoint()
-			nodeHandler.NodeValidateImplementation(*nodediscovery.LocalNode())
+			nodeHandler.AllNodeValidateImplementation()
```
brb, line 945: Small nit: in the comment above s/NodeValidateImplementation/AllNodeValidateImplementation/

brb (approved):
Thanks!

jschwinger233 (approved):

mhofstetter (approved):
Servicemesh related changes (`authmap_gc.go`) look good to me.

(going to remove review request for @meyskens)

thorn3r (approved):
agent changes look good :+1:

jrfastab:
Some tests failed on force push but really all I did was change a comment so seems tests are buggy.

rolinh (approved):

margamanterola:
Marking as backport/author to avoid this getting in the backporter bucket, as it's expected to have conflicts. @jschwinger233 will take care of the backports.

pchaigno:
I've added a release note.

PR #22738, opened 2022-12-14
datapath: fix NodePort to remote hostns backend with tunnel config

julianwiedmann:
When forwarding external service requests to a remote backend in
hostNetwork, the NodePort code in tail_nodeport_nat_egress_ipv4() doesn't
redirect into the tunnel (as it's effectively a host-to-host connection).

Therefore it uses IPV4_DIRECT_ROUTING as src address. So if we're not
populating IPV4_DIRECT_ROUTING, such forwarded requests end up being SNATed
with 0.0.0.0.

Fixes: https://github.com/cilium/cilium/issues/22557
Fixes: https://github.com/cilium/cilium/commit/5283ec938a5f2663126061e3f43bc85c0ded3af1 ("datapath: Introduce DirectRoutingDeviceRequired helper")
Signed-off-by: Julian Wiedmann <jwi@isovalent.com>

PR #27323, opened 2023-08-07
datapath: fix NodePort to remote hostns backend with tunnel config

### LINKED ISSUE #22557: Can't connect to remote NodePort backend in hostNetwork (with tunnel enabled)

### What happened?

In cilium v1.12 with tunneling enabled, it is not possible to send traffic to nodeport services with hostnetwork pods being enabled. I believe the issue stems from this pr #18815. The function `DirectRoutingDeviceRequired` returns false in this case and never sets the macro `IPV4_DIRECT_ROUTING` which is required by nodeport for the SNAT IP for pods which aren't tunneled(i.e host network pods). Thanks for looking at this issue, it took us a bit to trace down but hopefully it's detailed enough. I think the macro still needs to be set in this case for traffic that is not tunneled. 

The datapath code that uses this is here, https://github.com/cilium/cilium/blob/07592906e958f12090830dc8d554774f57e6be32/bpf/lib/nat.h#L599-L613 
I believe that HostNetwork traffic falls into the second part of the conditional which breaks traffic flow. 

### Cilium Version

v1.12.4

### Kernel Version

5.4.0-42-generic

### Kubernetes Version

1.25

### Sysdump

N/A

### Relevant log output

_No response_

### Anything else?

TCP Dump:
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on vxlan0, link-type EN10MB (Ethernet), capture size 262144 bytes
22:22:09.271839 IP 10.200.0.2.47792 > 10.200.0.52.29442: Flags [S], seq 2615184186, win 64390, options [mss 1370,sackOK,TS val 2929822314 ecr 0,nop,wscale 7], length 0
22:22:09.271923 IP 0.0.0.0.47792 > 10.200.0.5.29442: Flags [S], seq 2615184186, win 64390, options [mss 1370,sackOK,TS val 2929822314 ecr 0,nop,wscale 7], length 0

SNAT IP is replaced by 0.0.0.0 since `IPV4_DIRECT_ROUTING` is not set.

### Code of Conduct

michaelasp:
When forwarding external service requests to a remote backend in hostNetwork, the NodePort code in tail_nodeport_nat_egress_ipv4() doesn't redirect into the tunnel (as it's effectively a host-to-host connection).

Therefore it uses IPV4_DIRECT_ROUTING as src address. So if we're not populating IPV4_DIRECT_ROUTING, such forwarded requests end up being SNATed with 0.0.0.0.

This is a copy of #22738 by jwi@isovalent.com, reopening to debug test failures.

Fixes: #22557
Fixes: 5283ec9 ("datapath: Introduce DirectRoutingDeviceRequired helper")

### DIFF
--- a/pkg/datapath/linux/config/config_test.go
+++ b/pkg/datapath/linux/config/config_test.go
@@ -51,6 +51,7 @@ func setup(tb testing.TB) {
 
 	require.NoError(tb, rlimit.RemoveMemlock(), "Failed to remove memory limits")
 
+	option.Config.EnableHostLegacyRouting = true // Disable obtaining direct routing device.
 	node.SetTestLocalNodeStore()
 	node.InitDefaultPrefix("")
 	node.SetInternalIPv4Router(ipv4DummyAddr.AsSlice())
--- a/pkg/datapath/linux/devices_test.go
+++ b/pkg/datapath/linux/devices_test.go
@@ -30,16 +30,17 @@ import (
 )
 
 type DevicesSuite struct {
-	currentNetNS                   netns.NsHandle
-	prevConfigDevices              []string
-	prevConfigDirectRoutingDevice  string
-	prevConfigIPv6MCastDevice      string
-	prevConfigEnableIPv4           bool
-	prevConfigEnableIPv6           bool
-	prevConfigEnableNodePort       bool
-	prevConfigNodePortAcceleration string
-	prevConfigRoutingMode          string
-	prevConfigEnableIPv6NDP        bool
+	currentNetNS                      netns.NsHandle
+	prevConfigDevices                 []string
+	prevConfigDirectRoutingDevice     string
+	prevConfigIPv6MCastDevice         string
+	prevConfigEnableIPv4              bool
+	prevConfigEnableIPv6              bool
+	prevConfigEnableHostLegacyRouting bool
+	prevConfigEnableNodePort          bool
+	prevConfigNodePortAcceleration    string
+	prevConfigRoutingMode             string
+	prevConfigEnableIPv6NDP           bool
 }
 
 var _ = Suite(&DevicesSuite{})
@@ -74,6 +75,7 @@ func (s *DevicesSuite) TearDownTest(c *C) {
 	option.Config.EnableIPv4 = s.prevConfigEnableIPv4
 	option.Config.EnableIPv6 = s.prevConfigEnableIPv6
 	option.Config.EnableNodePort = s.prevConfigEnableNodePort
+	option.Config.EnableHostLegacyRouting = s.prevConfigEnableHostLegacyRouting
 	option.Config.NodePortAcceleration = s.prevConfigNodePortAcceleration
 	option.Config.RoutingMode = s.prevConfigRoutingMode
 	option.Config.EnableIPv6NDP = s.prevConfigEnableIPv6NDP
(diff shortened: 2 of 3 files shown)


michaelasp:
After rebasing on top of https://github.com/cilium/cilium/pull/24677, a nil check introduced in front of device manager allows the tests to pass. Looks like this can be merged with just these changes 😄

michaelasp:
@julianwiedmann these are basically the changes you made in #22738 but updated with some of the newer changes in device manager. Tested the agent tests, and they pass now 🥳

joestringer:
If this is based on #22738, then please ensure that the commit message properly attributes the origins of the source, ie:

```
Co-authored-by: Julian Wiedmann <jwi@isovalent.com>
```

joestringer:
I'm seeing a few CI issues that are due to the PR being based on an older version of `main`. I'd suggest rebasing against the tip of the `main` branch and updating the commit message while you're at it.

michaelasp:
Done!

michaelasp:
Fixed issue in TestDetect, other issues seem unrelated. I reverted my changes just in case and I could repro the TestWriteNodeConfig failures locally.

joestringer:
(Rebasing due to other changes that went into the tree since this PR was rebased, will trigger full CI soon)

derailed (approved):
@michaelasp Thanks for the updates! LGTM

michaelasp:
Thanks @derailed! Could either you or @joestringer start a new run of the CI?

joestringer:
:+1: usually @cilium/tophat will stop by from time to time to trigger CI, but I can also do it. If you're regularly contributing to Cilium, then you may be eligible to [join the Cilium organization](https://github.com/cilium/community/blob/main/CONTRIBUTOR-LADDER.md#organization-member) which will allow you to run the full testsuite without relying on others. Any other [members of the organization](https://github.com/orgs/cilium/teams/Team) can also do this, so for instance if you have colleagues in the organization then that can also be a path to tighten the feedback loop.

michaelasp:
Cilium IPsec upgrade seems like a flake, however interestingly TestWriteNodeConfig is failing in the ci-runtime tests. I wonder why that is. I can reproduce the issue without my change locally well. Let me look into it more.

michaelasp:
Looks like it's passing in other CI runs. Probably related to the change in DirectRoutingDeviceRequired

michaelasp:
All tests are passing 😄 @jibi @ti-mo feel free to take a look! The changes in loader and agent are small, essentially just modifying the tests slightly to not require a direct routing device when unnecessary for the test.

aojea:
@julianwiedmann as previous owner of the PR can you please help us with the reviews and approval process?

pkg/datapath/linux/config/config_test.go
```diff
@@ -51,6 +51,7 @@ func setup(tb testing.TB) {
 
 	require.NoError(tb, rlimit.RemoveMemlock(), "Failed to remove memory limits")
 
+	option.Config.EnableHostLegacyRouting = true // Disable obtaining direct routing device.
```
ti-mo, line 54: Can this be moved closer to where it's used? Does this need to be set to true for all tests in ConfigSuite?

michaelasp: Hey @ti-mo sorry for the long wait on this, was out for the past couple weeks. I believe this needs to be set for practically every test, which was what I did initially before this approach. Let me know what you think is better here, but from what I saw, nearly all tests needed it to pass.

michaelasp:
Hey @lmb sorry for the wait! I just got back to my computer today from a long break, responded 😃.

jibi (commented):
The rationale for the change makes sense to me, but it could have a broader impact than expected as:
* `DirectRoutingDeviceRequired()` can now return true even when tunneling is not enabled
*  because of that `(*DeviceManager) Detect()` will try to auto-detect a direct routing device even in tunneling mode 
* if we fail to detect a device either bpf nodeport gets disabled or the agent returns an error during initialization

so this PR has a potential upgrade impact as it might break some setups that used to work before :thinking:

jibi (approved):
synced internally, I think in practice these changes are fine :+1:

michaelasp:
Awesome! Glad to hear :)

lmb:
Removing ready-to-merge since @ti-mo needs to approve for cilium/loader. Not sure why the automation decided this was ready.

ti-mo (approved):

pkg/datapath/linux/config/config_test.go
michaelasp: Marking resolved since this got approved, @ti-mo feel free to reopen if there is anything you'd like to see done here.

PR #21667, opened 2022-10-11
ipcache: Add ability to override identity via UpsertMetadata

gandro:
The new asynchronous IPCache API tries to determine the identity of a
prefix based on the provided labels. However, in some cases, the sources
providing the IPCache metadata (e.g. CiliumNode, CiliumEndpoint) already
know the numeric identity of the prefix. In that case, we want to give
that predefined global identity preference over any local identity we
might derive from the labels.

All other metadata is still tracked. If the override is ever removed
from the prefix, the metadata will be used to determine the new prefix
identity.

Currently, the override identity is always accepted, regardless whether
the numeric identity is known e.g. to the identity allocator. This
behavior mirrors the current semantics of the legacy IPCache.Upsert
method. In the future, we could consider blocking overrides until they
are actually known to the identity allocator.

Identities added via override do not trigger a policy map update. It is
assumed that the policy map update for those identities is handled
elsewhere. This again mirrors the current behavior of IPCache.Upsert.

PR #27327, opened 2023-08-07
ipcache: Fix refcounting with mix of APIs

### LINKED ISSUE #27169: 1.14.0 - Policy Denies occurring when they should not be

### What happened?

I upgraded my cluster from 1.13-latest to 1.14.0.
I then switched from IPTables to BPF.
In audit mode everything works fine and audits all seem expected.
When audit mode is disabled I see lots of policy denies that I can't seem to explain.

### Cilium Version

1.14.0

### Kernel Version

Linux 5.14.0-284.18.1.el9_2.x86_64 #1 Thu Jun 29 17:06:27 EDT 2023 x86_64 x86_64 x86_64 GNU/Linux

### Kubernetes Version

1.25.8

### Sysdump

_No response_

### Relevant log output
cilium status --verbose:
```
# cilium status --verbose
KVStore:                Ok   Disabled
Kubernetes:             Ok   1.25 (v1.25.8) [linux/amd64]
Kubernetes APIs:        ["EndpointSliceOrEndpoint", "cilium/v2::CiliumClusterwideNetworkPolicy", "cilium/v2::CiliumEndpoint", "cilium/v2::CiliumNetworkPolicy", "cilium/v2::CiliumNode", "cilium/v2alpha1::CiliumCIDRGroup", "core/v1::Namespace", "core/v1::Pods", "core/v1::Service", "networking.k8s.io/v1::NetworkPolicy"]
KubeProxyReplacement:   True      [bond0 REDACTED, private0 172.31.0.90 (Direct Routing)]
Host firewall:          Enabled   [bond0, private0]
CNI Chaining:           none
Cilium:                 Ok   1.14.0 (v1.14.0-b5013e15)
NodeMonitor:            Listening for events on 16 CPUs with 64x4096 of shared memory
Cilium health daemon:   Ok
IPAM:                   IPv4: 3/254 allocated from 10.0.16.0/24,
Allocated addresses:
  10.0.16.167 (kube-system/coredns-588bb58b94-4txvt [restored])
  10.0.16.215 (router)
  10.0.16.50 (health)
IPv4 BIG TCP:           Disabled
IPv6 BIG TCP:           Disabled
BandwidthManager:       Disabled
Host Routing:           BPF
Masquerading:           BPF       [bond0, private0]   10.0.0.0/8 [IPv4: Enabled, IPv6: Disabled]
Clock Source for BPF:   jiffies   [1000 Hz]
Controller Status:      27/27 healthy
  Name                                         Last success   Last error    Count   Message
  bpf-map-sync-cilium_lxc                      8s ago         never         0       no error
  cilium-health-ep                             58s ago        never         0       no error
  dns-garbage-collector-job                    1m2s ago       never         0       no error
  endpoint-189-regeneration-recovery           never          never         0       no error
  endpoint-1924-regeneration-recovery          never          never         0       no error
  endpoint-740-regeneration-recovery           never          never         0       no error
  endpoint-gc                                  1m2s ago       never         0       no error
  ipcache-inject-labels                        59s ago        1h46m0s ago   0       no error
  k8s-heartbeat                                32s ago        never         0       no error
  link-cache                                   14s ago        never         0       no error
  metricsmap-bpf-prom-sync                     6s ago         never         0       no error
  neighbor-table-refresh                       29s ago        never         0       no error
  resolve-identity-189                         58s ago        never         0       no error
  restoring-ep-identity (1924)                 1h45m59s ago   never         0       no error
  restoring-ep-identity (740)                  1h45m59s ago   never         0       no error
  sync-host-ips                                59s ago        never         0       no error
  sync-lb-maps-with-k8s-services               1h45m59s ago   never         0       no error
  sync-policymap-189                           40s ago        never         0       no error
  sync-policymap-1924                          40s ago        never         0       no error
  sync-policymap-740                           40s ago        never         0       no error
  sync-to-k8s-ciliumendpoint (189)             8s ago         never         0       no error
  sync-to-k8s-ciliumendpoint (1924)            9s ago         never         0       no error
  sync-to-k8s-ciliumendpoint (740)             9s ago         never         0       no error
  sync-utime                                   59s ago        never         0       no error
  template-dir-watcher                         never          never         0       no error
  waiting-initial-global-identities-ep (740)   1h45m59s ago   never         0       no error
  write-cni-file                               1h46m2s ago    never         0       no error
Proxy Status:            OK, ip 10.0.16.215, 0 redirects active on ports 10000-20000, Envoy: embedded
Global Identity Range:   min 256, max 65535
Hubble:                  Ok   Current/Max Flows: 4095/4095 (100.00%), Flows/s: 24.28   Metrics: Ok
KubeProxyReplacement Details:
  Status:                 True
  Socket LB:              Enabled
  Socket LB Tracing:      Enabled
  Socket LB Coverage:     Full
  Devices:                bond0 REDACTED, private0 172.31.0.90 (Direct Routing)
  Mode:                   SNAT
  Backend Selection:      Random
  Session Affinity:       Enabled
  Graceful Termination:   Enabled
  NAT46/64 Support:       Disabled
  XDP Acceleration:       Disabled
  Services:
  - ClusterIP:      Enabled
  - NodePort:       Enabled (Range: 30000-32767)
  - LoadBalancer:   Enabled
  - externalIPs:    Enabled
  - HostPort:       Enabled
BPF Maps:   dynamic sizing: on (ratio: 0.002500)
  Name                          Size
  Auth                          524288
  Non-TCP connection tracking   589568
  TCP connection tracking       1179136
  Endpoint policy               65535
  IP cache                      512000
  IPv4 masquerading agent       16384
  IPv6 masquerading agent       16384
  IPv4 fragmentation            8192
  IPv4 service                  65536
  IPv6 service                  65536
  IPv4 service backend          65536
  IPv6 service backend          65536
  IPv4 service reverse NAT      65536
  IPv6 service reverse NAT      65536
  Metrics                       1024
  NAT                           1179136
  Neighbor table                1179136
  Global policy                 16384
  Session affinity              65536
  Sock reverse NAT              589568
  Tunnel                        65536
Encryption:                                      Disabled
Cluster health:                                  51/51 reachable   (2023-08-01T00:03:23Z)
  Name                                           IP                Node          Endpoints
  znsvc1-3.mydomain.net (localhost)   172.31.0.90       reachable     reachable
  ze2101-1.mydomain.net               172.31.0.16       reachable     reachable
  ze2101-10.mydomain.net              172.31.0.25       reachable     reachable
  ze2101-11.mydomain.net              172.31.0.26       reachable     reachable
  ze2101-12.mydomain.net              172.31.0.27       reachable     reachable
  ze2101-13.mydomain.net              172.31.0.28       reachable     reachable
  ze2101-14.mydomain.net              172.31.0.29       reachable     reachable
  ze2101-15.mydomain.net              172.31.0.30       reachable     reachable
  ze2101-16.mydomain.net              172.31.0.31       reachable     reachable
  ze2101-17.mydomain.net              172.31.0.32       reachable     reachable
  ze2101-18.mydomain.net              172.31.0.33       reachable     reachable
  ze2101-19.mydomain.net              172.31.0.34       reachable     reachable
  ze2101-2.mydomain.net               172.31.0.17       reachable     reachable
  ze2101-20.mydomain.net              172.31.0.35       reachable     reachable
  ze2101-21.mydomain.net              172.31.0.36       reachable     reachable
  ze2101-22.mydomain.net              172.31.0.37       reachable     reachable
  ze2101-23.mydomain.net              172.31.0.38       reachable     reachable
  ze2101-24.mydomain.net              172.31.0.39       reachable     reachable
  ze2101-25.mydomain.net              172.31.0.40       reachable     reachable
  ze2101-26.mydomain.net              172.31.0.41       reachable     reachable
  ze2101-27.mydomain.net              172.31.0.42       reachable     reachable
  ze2101-28.mydomain.net              172.31.0.43       reachable     reachable
  ze2101-29.mydomain.net              172.31.0.44       reachable     reachable
  ze2101-3.mydomain.net               172.31.0.18       reachable     reachable
  ze2101-30.mydomain.net              172.31.0.45       reachable     reachable
  ze2101-31.mydomain.net              172.31.0.46       reachable     reachable
  ze2101-32.mydomain.net              172.31.0.47       reachable     reachable
  ze2101-33.mydomain.net              172.31.0.48       reachable     reachable
  ze2101-34.mydomain.net              172.31.0.49       reachable     reachable
  ze2101-35.mydomain.net              172.31.0.50       reachable     reachable
  ze2101-36.mydomain.net              172.31.0.51       reachable     reachable
  ze2101-37.mydomain.net              172.31.0.52       reachable     reachable
  ze2101-38.mydomain.net              172.31.0.53       reachable     reachable
  ze2101-39.mydomain.net              172.31.0.54       reachable     reachable
  ze2101-4.mydomain.net               172.31.0.19       reachable     reachable
  ze2101-40.mydomain.net              172.31.0.55       reachable     reachable
  ze2101-41.mydomain.net              172.31.0.56       reachable     reachable
  ze2101-42.mydomain.net              172.31.0.57       reachable     reachable
  ze2101-43.mydomain.net              172.31.0.58       reachable     reachable
  ze2101-44.mydomain.net              172.31.0.59       reachable     reachable
  ze2101-45.mydomain.net              172.31.0.60       reachable     reachable
  ze2101-46.mydomain.net              172.31.0.61       reachable     reachable
  ze2101-47.mydomain.net              172.31.0.62       reachable     reachable
  ze2101-48.mydomain.net              172.31.0.63       reachable     reachable
  ze2101-5.mydomain.net               172.31.0.20       reachable     reachable
  ze2101-6.mydomain.net               172.31.0.21       reachable     reachable
  ze2101-7.mydomain.net               172.31.0.22       reachable     reachable
  ze2101-8.mydomain.net               172.31.0.23       reachable     reachable
  ze2101-9.mydomain.net               172.31.0.24       reachable     reachable
  znsvc1-4.mydomain.net               172.31.0.91       reachable     reachable
  znsvc1-5.mydomain.net               172.31.0.92       reachable     reachable
```
Here is an example.  
1. policy-verdict output from ~1m on a random node:
```
$ time kubectl -n kube-system exec -it cilium-298mj -- hubble observe flows -t policy-verdict -f | grep AUDITED
Defaulted container "cilium-agent" out of: cilium-agent, config (init), mount-cgroup (init), apply-sysctl-overwrites (init), mount-bpf-fs (init), wait-for-node-init (init), clean-cilium-state (init), install-cni-binaries (init)
Jul 31 22:01:34.453: 162.243.129.7:60929 (world) -> REDACTED:6666 (host) policy-verdict:none INGRESS AUDITED (TCP Flags: SYN)
Jul 31 22:01:38.900: 179.43.167.186:34808 (world) -> REDACTED:2133 (host) policy-verdict:none INGRESS AUDITED (TCP Flags: SYN)
Jul 31 22:01:39.355: 198.235.24.71:56864 (world) -> REDACTED:17516 (host) policy-verdict:none INGRESS AUDITED (TCP Flags: SYN)
Jul 31 22:01:48.727: 185.81.68.102:43562 (world) -> REDACTED:54523 (host) policy-verdict:none INGRESS AUDITED (TCP Flags: SYN)
Jul 31 22:01:49.165: 109.248.6.78:50801 (world) -> REDACTED:1027 (host) policy-verdict:none INGRESS AUDITED (UDP)
Jul 31 22:01:59.260: 89.248.168.235:57596 (world) -> REDACTED:7663 (host) policy-verdict:none INGRESS AUDITED (TCP Flags: SYN)
Jul 31 22:02:04.964: 162.243.142.59:58956 (world) -> REDACTED:3389 (host) policy-verdict:none INGRESS AUDITED (UDP)
Jul 31 22:02:09.086: 45.240.150.10:49146 (world) -> REDACTED:45079 (host) policy-verdict:none INGRESS AUDITED (TCP Flags: SYN)

real    0m41.368s
user    0m0.158s
sys     0m0.080s
```

2. Manually edit cilium-config config-map resource to have audit mode disabled.
3. Restart a single pod
4. Here is the DROPPED log after for ~1m
```
 $ kubectl -n kube-system exec -it cilium-zpkzk -- hubble observe flows -t drop -f
Defaulted container "cilium-agent" out of: cilium-agent, config (init), mount-cgroup (init), apply-sysctl-overwrites (init), mount-bpf-fs (init), wait-for-node-init (init), clean-cilium-state (init), install-cni-binaries (init)
Jul 31 22:06:18.510: 172.31.0.30 (host) <> 172.31.0.50 (remote-node) Unsupported protocol for NAT masquerade DROPPED (ICMPv4 TimeExceeded(TTLExceeded))
Jul 31 22:06:19.279: 172.31.0.30 (host) <> 172.31.0.23 (remote-node) Unsupported protocol for NAT masquerade DROPPED (ICMPv4 TimeExceeded(TTLExceeded))
Jul 31 22:06:19.287: 172.31.0.30 (host) <> 172.31.0.24 (remote-node) Unsupported protocol for NAT masquerade DROPPED (ICMPv4 TimeExceeded(TTLExceeded))
Jul 31 22:06:20.685: 205.210.31.145:57248 (world) <> REDACTED:9002 (host) Policy denied DROPPED (TCP Flags: SYN)
Jul 31 22:06:20.721: 172.31.0.30 (host) <> 172.31.0.27 (remote-node) Unsupported protocol for NAT masquerade DROPPED (ICMPv4 TimeExceeded(TTLExceeded))
Jul 31 22:06:20.786: 172.31.0.30 (host) <> 172.31.0.26 (remote-node) Unsupported protocol for NAT masquerade DROPPED (ICMPv4 TimeExceeded(TTLExceeded))
Jul 31 22:06:20.978: 172.31.0.30 (host) <> 172.31.0.54 (remote-node) Unsupported protocol for NAT masquerade DROPPED (ICMPv4 TimeExceeded(TTLExceeded))
Jul 31 22:06:22.599: 172.31.0.30 (host) <> 172.31.0.33 (remote-node) Unsupported protocol for NAT masquerade DROPPED (ICMPv4 TimeExceeded(TTLExceeded))
Jul 31 22:06:24.711: 172.31.0.30 (host) <> 172.31.0.32 (remote-node) Unsupported protocol for NAT masquerade DROPPED (ICMPv4 TimeExceeded(TTLExceeded))
Jul 31 22:06:26.749: 172.31.0.30 (host) <> 172.31.0.48 (remote-node) Unsupported protocol for NAT masquerade DROPPED (ICMPv4 TimeExceeded(TTLExceeded))
Jul 31 22:06:28.002: 172.31.0.30 (host) <> 172.31.0.22 (remote-node) Unsupported protocol for NAT masquerade DROPPED (ICMPv4 TimeExceeded(TTLExceeded))
Jul 31 22:06:28.113: 172.31.0.30 (host) <> 172.31.0.20 (remote-node) Unsupported protocol for NAT masquerade DROPPED (ICMPv4 TimeExceeded(TTLExceeded))
Jul 31 22:06:31.625: 117.9.1.18:29989 (world) <> REDACTED:23 (host) Policy denied DROPPED (TCP Flags: SYN)
Jul 31 22:06:52.788: 92.63.197.91:45830 (world) <> REDACTED:65000 (host) Policy denied DROPPED (TCP Flags: SYN)
Jul 31 22:06:53.890: 78.128.113.102:44215 (world) <> REDACTED:9060 (host) Policy denied DROPPED (TCP Flags: SYN)
Jul 31 22:10:16.000: kube-system/coredns-588bb58b94-vcx55:39097 (ID:43333) <> 1.1.1.1:53 (ID:16777223) policy-verdict:none EGRESS DENIED (UDP)
Jul 31 22:10:16.000: kube-system/coredns-588bb58b94-vcx55:39097 (ID:43333) <> 1.1.1.1:53 (ID:16777223) Policy denied DROPPED (UDP)
Jul 31 22:10:17.501: kube-system/coredns-588bb58b94-vcx55:43595 (ID:43333) <> 1.1.1.1:53 (ID:16777223) policy-verdict:none EGRESS DENIED (UDP)
Jul 31 22:10:17.501: kube-system/coredns-588bb58b94-vcx55:43595 (ID:43333) <> 1.1.1.1:53 (ID:16777223) Policy denied DROPPED (UDP)
Jul 31 22:10:19.003: kube-system/coredns-588bb58b94-vcx55:32808 (ID:43333) <> 1.1.1.1:53 (ID:16777223) policy-verdict:none EGRESS DENIED (UDP)
Jul 31 22:10:19.003: kube-system/coredns-588bb58b94-vcx55:32808 (ID:43333) <> 1.1.1.1:53 (ID:16777223) Policy denied DROPPED (UDP)
```

5. Policies that would allow the above denied/dropped traffic for coredns:
```
$ kubectl -n kube-system get ciliumnetworkpolicy coredns -o yaml
...
specs:
- description: All coredns pods allowed egress to DNS servers
  egress:
  - toCIDRSet:
    - cidr: 1.1.1.1/32
    - cidr: 8.8.4.4/32
    - cidr: 8.8.8.8/32
    - cidr: 169.254.25.10/32
    toPorts:
    - ports:
      - port: "53"
  endpointSelector:
    matchLabels:
      k8s:io.cilium.k8s.policy.serviceaccount: coredns
      k8s:io.kubernetes.pod.namespace: kube-system
```
### Anything else?

It is very difficult to test and/or get logs because when I leave audit mode an massive amount of DROPs occur.  Cilium does not seem to be respecting the network policies that worked perfectly fine before the move to bpf.

In this image (IPs blacked out) we can see an example of some drops.
image

These drops should not happen due to a `CiliumClusterwideNetworkPolicy`
```
    - name: 'nodes'
      policies:
        - description: "Allow all nodes SSH from MyIP"
          nodeSelector:
            matchLabels: {}
          ingress:
            - fromCIDR:
                - <blacked-out-ip>/32 # MyIP (Redacted)
              toPorts:
                - ports:
                    - port: '22'
                      protocol: TCP
```

When attempting to lookup the ID:
```
Jul 31 19:00:43.148: <my-ip>:53002 (ID:16777231) -> <host-ip>:22 (host) policy-verdict:none INGRESS AUDITED (TCP Flags: SYN)
...
Error: Cannot get identity for given ID 16777231: [GET /identity/{id}][404] getIdentityIdNotFound
```

### Code of Conduct

### LINKED ISSUE #27176: Inconsistency when adding fromCIDR rule under CiliumClusterwideNetworkPolicy

### What happened?

There's an inconsistent behavior when adding `fromCIDR` rule in combination with other rules e.g. entity-based rules.
Since entries from CIDRs are sources that are not managed by Cilium, when such entries are added, Cilium creates Identities/IP list and labels for those sources that it uses internally for policy matching.  

Issue is, when modifying those rules  it goes into inconsistent state where it cannot sync Identities/IPs against all those fromCIDR sources which ultimately halts the whole policy or atleast for those CIDR based rules.  

One way to work around this is by recreating the whole policy so is able to re-create those identities.  
We have been also able to come over this issue by moving the entity-based at the bottom of the policy and let it sync again.

The issue can be seen if ran `cilium identity list`  which will be missing all or some CIDR entries.  
There is also a "404 not found error" seen when running `cilium ip list`  which is basically cilium trying to find identity for one of those CIDR that it has some record stuck but cannot be found.

### Cilium Version

1.14.0

### Kernel Version

4.18.0-372.57.1.el8_6.x86_64

### Kubernetes Version

v1.26.3

### Sysdump

[cilium-sysdump-20230801-104512.zip](https://github.com/cilium/cilium/files/12225955/cilium-sysdump-20230801-104512.zip)

### Relevant log output

```shell
404 error during inconsistent state

root@immense-wasp:/home/cilium# cilium ip list
Error: Cannot get identity for given ID <nil>: [GET /identity/{id}][404] getIdentityIdNotFound
```
```


### Anything else?

_No response_

### Code of Conduct

### LINKED ISSUE #27210: Valid CIDR Identities are getting released upon Cilium Agent restart

### What happened?

I have `CililiumClusterwideNetworkPolicy` deployed and Host Firewall enabled. Among other CCNPs that allow Cilium to function normally, I have CCNP that allows access from trusted subnets (private network, home VPN, etc...) and it looks like this:

````
apiVersion: "cilium.io/v2"
kind: CiliumClusterwideNetworkPolicy
metadata:
  name: "whitelisted-cidrs"
spec:
  description: "Allow access from trusted CIDRs"
  nodeSelector:
    matchLabels: {}
  ingress:
  - fromCIDR:
    - "127.0.0.0/8"
    - "10.0.0.0/8"
    - "172.16.0.0/12"
    - "192.168.0.0/16"
    - "<trusted public ip>:32"
 ```` 

When this policy is applied for the first time everything works as expected. Problem occur 10 minutes after Cilium agent is restarted and is manifested by loosing connectivity to cluster from addresses defined in `whitelisted-cidrs` CCNP.

Since this is huge problem for me because I'm heavily relaying on Cilium Host Firewall feature, I did some investigation and these are my findings:
- Identities associated with CIDRs in `fromCIDR` list are cleared 10 minutes after Cilium agent starts
- [This part of code is responsible for 10 minutes delay](https://github.com/cilium/cilium/blob/b5013e1562dbfe442dd2011a5af46de9095315ff/daemon/cmd/daemon_main.go#L1757-L1771)
- What actually releases Identity is [localIdentityCache release function](https://github.com/cilium/cilium/blob/b5013e1562dbfe442dd2011a5af46de9095315ff/pkg/identity/cache/local.go#L122-L151)
- It releases Identity if refcount for provided Identity is equal to 1
- In Cilium versions prior to v1.14.0 this function is called only once, and it's called from that delayed goroutine from second bullet point, [here](https://github.com/cilium/cilium/blob/b5013e1562dbfe442dd2011a5af46de9095315ff/daemon/cmd/daemon_main.go#L1768)
- This call does not cause Identity to be released because at that point, refcount for given Identity is 2, as it is increased [here](https://github.com/cilium/cilium/blob/b5013e1562dbfe442dd2011a5af46de9095315ff/pkg/identity/cache/local.go#L88-L91) beforehand
- Difference between v1.14.0 and v.1.13.4 is that in v.1.14 release function is called once more when refcount is 1 which causes given Identity to be released which makes cluster unavailable
- Release function is called second time from ipcache.InjectLabels, [here](https://github.com/cilium/cilium/blob/b5013e1562dbfe442dd2011a5af46de9095315ff/pkg/ipcache/metadata.go#L332)
- Reason for this is changed logic inside `releaseIdentity` block, [here](https://github.com/cilium/cilium/blob/b5013e1562dbfe442dd2011a5af46de9095315ff/pkg/ipcache/metadata.go#L266-L284)
- This change is introduced in [this commit](https://github.com/cilium/cilium/commit/c96b9d81a8a679d77f9929e6ed32559a3d54108d)

After I figured out what is happening, I reverted logic in `releaseIdentity` to look like this:
```
	releaseIdentity:
		if entryExists {
			// ...
			if _, ok := idsToAdd[oldID.ID]; !
				previouslyAllocatedIdentities[prefix] = oldID
			}

			// ...
			if prefixInfo == nil && oldID.createdFromMetadata {
				entriesToDelete[prefix] = oldID
			}
		}
```

This change solved the issue, but since I'm new to Cilium code base, I'm not sure will this have some unwanted side effects. If this change looks good, I can submit PR.

### How to reproduce in local Kind cluster
- Checkout Cilium v1.14.0 tag
- Add following snippet to kind-values.yaml: 

````
extraArgs:
  - '--identity-restore-grace-period'
  - '2m'
hostFirewall:
  enabled: "true"
````
- Run `make kind && make kind-image && make kind-install-cilium`
- Apply CCNP with `fromCIDR` rule (like one from above)
- Restart Cilium Agent
- Wait for 2 minutes
- Cluster should become unavailable since fromCIDR identities are released

### Cilium Version

v1.14.0

### Kernel Version

.

### Kubernetes Version

v1.27.3

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

### PR DESCRIPTION
**Pull request by joestringer**
Commit c96b9d81a8a6 ("ipcache: Remove superfluous if condition")                                                                                                                                                                                                                           
introduced a double-free for cases where the ipcache allocates a new                                                                                                                                                                                                                       
(reference to a) security identity and there is already an identity                                                                                                                                                                                                                        
corresponding to those labels which another subsystem injected directly                                                                                                                                                                                                                    
into the ipcache via Upsert() and friends. This commit fixes the issue.                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                           
IPCache has two sets of APIs:                                                                                                                                                                                                                                                              
- Older APIs like Upsert() / UpsertGeneratedIdentities which require the                                                                                                                                                                                                                   
  caller to retain reference counts against each Identity and balance                                                                                                                                                                                                                      
  allocate/release calls.                                                                                                                                                                                                                                                                  
- Newer APIs like UpsertMetadata() / UpsertLabels() / UpsertPrefixes()                                                                                                                                                                                                                     
  which delegate the reference counting responsibility to the ipcache                                                                                                                                                                                                                      
  itself, which will hold one reference to the identity for all callers                                                                                                                                                                                                                    
  of the APIs.                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                           
InjectLabels() is responsible for reference counting the identities                                                                                                                                                                                                                        
associated with ipcache entries that it manages. In order to ensure                                                                                                                                                                                                                        
balanced allocate and release, this function allocates and retains one                                                                                                                                                                                                                     
reference to the identity for itself internally within this function.                                                                                                                                                                                                                      
Under normal operation it will typically release this reference again                                                                                                                                                                                                                      
once it has performed the ipcache update operations. It may be triggered                                                                                                                                                                                                                   
multiple times for the same prefix, in which case it should only ever                                                                                                                                                                                                                      
acquire one reference to the identity, otherwise the references could                                                                                                                                                                                                                      
leak and the identity would not be freed.                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                           
There is one exception to the "always allocate" + "always" release                                                                                                                                                                                                                         
behaviour in InjectLabels(): Given that the ipcache needs to hold                                                                                                                                                                                                                          
exactly one reference to the identity over time to ensure it remains                                                                                                                                                                                                                       
live while in use by the ipcache, the very first time that                                                                                                                                                                                                                                 
InjectLabels() allocates the identity reference, it must hold onto the                                                                                                                                                                                                                     
reference and not immediately release it. In particular, if another                                                                                                                                                                                                                        
subsystem injected an ipcache entry via older APIs, then the first time                                                                                                                                                                                                                    
that InjectLabels() runs, it also needs to acquire and hold a reference                                                                                                                                                                                                                    
to the identity, even though the underlying ipcache already contains an                                                                                                                                                                                                                    
CIDR -> Identity mapping that was injected via the older Upsert() API.                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                           
In commit c96b9d81a8a6 ("ipcache: Remove superfluous if condition"), the                                                                                                                                                                                                                   
check which was ensuring that ipcache entries that transition from being                                                                                                                                                                                                                   
managed by older APIs towards newer APIs would not trigger allocation of                                                                                                                                                                                                                   
a new identity reference to be held by the ipcache itself. The result                                                                                                                                                                                                                      
would be that the resting number of references to the identity would be                                                                                                                                                                                                                    
one less than what is necessary to keep the identity alive. If the other                                                                                                                                                                                                                   
subsystem ever released its reference to the identity, then that would                                                                                                                                                                                                                     
be the last reference and the identity would be freed, despite still                                                                                                                                                                                                                       
being in use.                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                           
This scenario leads to policy recalculation that removes any datapath                                                                                                                                                                                                                      
allow rules for the corresponding CIDRs, ultimately resulting in packet                                                                                                                                                                                                                    
loss for the impacted CIDRs. One such example involves CIDR identity                                                                                                                                                                                                                       
restore startup logic in the daemon. That path allocates identities then                                                                                                                                                                                                                   
injects them into the ipcache using older APIs. If any such CIDRs are                                                                                                                                                                                                                      
used by network policies, then the network policies subsystem will                                                                                                                                                                                                                         
insert the CIDR into the ipcache using newer ipcache APIs. Given this                                                                                                                                                                                                                      
mix of API usage and the bug introduced in the identity reference                                                                                                                                                                                                                          
counting internally in ipcache, this triggers the bug.                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                           
This commit fixes it by checking whether InjectLabels() already assumed                                                                                                                                                                                                                    
ownership over the underlying ipcache entry, which implies that the                                                                                                                                                                                                                        
ipcache already holds a reference to the identity on behalf of all newer                                                                                                                                                                                                                   
API users. During subsequent calls to InjectLabels(), the logic checks                                                                                                                                                                                                                     
that the ipcache entry was installed by this function and accounts for                                                                                                                                                                                                                     
the fact that a prior iteration had allocated an identity reference. It                                                                                                                                                                                                                    
can then safely release references that were acquired during the current                                                                                                                                                                                                                   
execution of the function.                                                                                                                                                                                                                                                                 

Fixes: c96b9d81a8a6 ("ipcache: Remove superfluous if condition")
Reported-by: Boris Petrovic (@carnerito)
Reported-by: Kim-Eirik Karlsen (@kimma)
Reported-by: Jason Witkowski (@jwitko)

Related: https://github.com/cilium/cilium/pull/20116
Related: https://github.com/cilium/cilium/pull/21667
Fixes: #21667
Fixes: #27169
Fixes: #27176
Fixes: #27210

```release-note
Fix bug where startup CIDR restore logic would mishandle reference counting, leading to persistent packet loss to those CIDRs
```

### DIFF
--- a/pkg/ipcache/metadata.go
+++ b/pkg/ipcache/metadata.go
@@ -22,6 +22,7 @@ import (
 	cidrlabels "github.com/cilium/cilium/pkg/labels/cidr"
 	"github.com/cilium/cilium/pkg/lock"
 	"github.com/cilium/cilium/pkg/logging/logfields"
+	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/source"
 )
 
@@ -272,8 +273,28 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
 			// iteration of the loop, then we must balance the
 			// allocation from the prior InjectLabels() call by
 			// releasing the previous reference.
-			previouslyAllocatedIdentities[prefix] = oldID
-
+			entry, entryToBeReplaced := entriesToReplace[prefix]
+			if !oldID.createdFromMetadata && entryToBeReplaced {
+				// If the previous ipcache entry for the prefix
+				// was not managed by this function, then the
+				// previous ipcache user to inject the IPCache
+				// entry retains its own reference to the
+				// Security Identity. Given that this function
+				// is going to assume responsibility for the
+				// IPCache entry now, this path must retain its
+				// own reference to the Security Identity to
+				// ensure that if the other owner ever releases
+				// their reference, this reference stays live.
+				if option.Config.Debug {
+					log.WithFields(logrus.Fields{
+						logfields.Prefix:      prefix,
+						logfields.OldIdentity: oldID.ID,
+						logfields.Identity:    entry.identity.ID,
+					}).Debug("Acquiring Identity reference")
+				}
+			} else {
+				previouslyAllocatedIdentities[prefix] = oldID
+			}
 			// If all associated metadata for this prefix has been removed,
 			// and the existing IPCache entry was never touched by any other
 			// subsystem using the old Upsert API, then we can safely remove
--- a/pkg/ipcache/metadata_test.go
+++ b/pkg/ipcache/metadata_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/cilium/cilium/pkg/identity/cache"
 	"github.com/cilium/cilium/pkg/ipcache/types"
 	"github.com/cilium/cilium/pkg/labels"
+	"github.com/cilium/cilium/pkg/labels/cidr"
 	"github.com/cilium/cilium/pkg/source"
 	testidentity "github.com/cilium/cilium/pkg/testutils/identity"
 )
@@ -171,6 +172,97 @@ func TestInjectExisting(t *testing.T) {
 	id, ok = IPIdentityCache.LookupByIP(prefix.String())
 	assert.True(t, ok)
 	assert.Equal(t, source.KubeAPIServer, id.Source)
+}
+
+// TestInjectWithLegacyAPIOverlap tests that a previously allocated identity
+// will continue to be used in the ipcache even if other users of newer APIs
+// also use the API, and that reference counting is properly balanced for this
+// pattern.This is a common occurrence on startup - and this tests ensures we
+// don't regress the known issue in GH-24502
+//
+// This differs from TestInjectExisting() by reusing the same identity, and by
+// not associating any new labels with the prefix.
+func TestInjectWithLegacyAPIOverlap(t *testing.T) {
+	cancel := setupTest(t)
+	defer cancel()
+
+	// mimic the "restore cidr" logic from daemon.go
+	// for every ip -> identity mapping in the bpf ipcache
+	// - allocate that identity
+	// - insert the cidr=>identity mapping back in to the go ipcache
+	identities := make(map[netip.Prefix]*identity.Identity)
+	prefix := netip.MustParsePrefix("172.19.0.5/32")
+	oldID := identity.NumericIdentity(16777219)
+	_, err := IPIdentityCache.AllocateCIDRs([]netip.Prefix{prefix}, []identity.NumericIdentity{oldID}, identities)
+	assert.NoError(t, err)
+	identityReferences := 1
+
+	IPIdentityCache.UpsertGeneratedIdentities(identities, nil)
+
+	// sanity check: ensure the cidr is correctly in the ipcache
+	id, ok := IPIdentityCache.LookupByIP(prefix.String())
+	assert.True(t, ok)
+	assert.Equal(t, int32(16777219), int32(id.ID))
+
+	// Simulate the first half of UpsertLabels -- insert the labels only in to the metadata cache
+	// This is to "force" a race condition
+	resource := types.NewResourceID(
+		types.ResourceKindCNP, "default", "policy")
+	labels := cidr.GetCIDRLabels(prefix)
+	IPIdentityCache.metadata.upsertLocked(prefix, source.CustomResource, resource, labels)
+
+	// Now, emulate policyAdd(), which calls AllocateCIDRs()
+	_, err = IPIdentityCache.AllocateCIDRs([]netip.Prefix{prefix}, []identity.NumericIdentity{oldID}, nil)
+	assert.NoError(t, err)
+	identityReferences++
+
+	// Now, trigger label injection
+	// This will allocate a new ID for the same /32 since the labels have changed
+	// It should only allocate once, even if we run it multiple times.
+	identityReferences++
+	for i := 0; i < 2; i++ {
+		IPIdentityCache.UpsertLabels(prefix, labels, source.CustomResource, resource)
+		// Need to wait for the label injector to finish; easiest just to remove it
+		IPIdentityCache.controllers.RemoveControllerAndWait(LabelInjectorName)
+	}
+
+	// Ensure the source is now correctly understood in the ipcache
+	id, ok = IPIdentityCache.LookupByIP(prefix.String())
+	assert.True(t, ok)
+	assert.Equal(t, source.CustomResource, id.Source)
+
+	// Release the identity references via the legacy API. As long as the
+	// external subsystems are balancing their references against the
+	// identities, then the remainder of the test will assert that the
+	// ipcache internals will properly reference-count the identities
+	// for users of the newer APIs where ipcache itself is responsible for
+	// reference counting.
+	for i := identityReferences; i > 1; i-- {
+		IPIdentityCache.releaseCIDRIdentities(context.Background(), []netip.Prefix{prefix})
+		identityReferences--
+	}
+
+	// sanity check: ensure the cidr is correctly in the ipcache
+	id, ok = IPIdentityCache.LookupByIP(prefix.String())
+	assert.True(t, ok)
+	assert.Equal(t, oldID.Uint32(), id.ID.Uint32())
+
+	// Check that the corresponding identity in the identity allocator
+	// is still allocated, which implies that it's reference counted
+	// correctly compared to the identityReferences variable in this test.
+	realID := IPIdentityCache.IdentityAllocator.LookupIdentityByID(context.Background(), id.ID)
+	assert.True(t, realID != nil)
+	assert.Equal(t, id.ID.Uint32(), uint32(realID.ID))
+
+	// Remove the identity allocation via newer APIs
+	IPIdentityCache.RemoveLabels(prefix, labels, resource)
+	IPIdentityCache.controllers.RemoveControllerAndWait(LabelInjectorName)
+	identityReferences--
+	assert.Equal(t, identityReferences, 0)
+
+	// Assert that ipcache has released its final reference to the identity
+	realID = IPIdentityCache.IdentityAllocator.LookupIdentityByID(context.Background(), id.ID)
+	assert.True(t, realID == nil)
 }
 
 func TestFilterMetadataByLabels(t *testing.T) {


joestringer:
I've run this three times against https://github.com/cilium/cilium/pull/27253 which was able to reproduce two related other issues during development of these fixes last week. That gives me fair confidence that the fix is addressing the problem it intends to, without introducing any obvious regressions.

christarazi (approved):

squeed (approved):
It took me a bit of thinking, but now I get why this works.

InjectLabels always does a `new := Allocate()` followed by a `Release(existing)`. Most of the time, the identity for a set of labels never changes, so `new == existing` and the reference count is unchanged. In the case where the ID has changed, the existing ID will be properly released and cleaned up.

The issue is when legacy callers do allocation. At that point in time, there is an entry in the ipcache that did not have an ipcache-owned `Allocate()`. Thus, it is **not** safe for the ipcache to call `Release(existing)` since it didn't balance it with its own `Allocate()`

We also cannot just change legacy callers to never `Release()`, since the ipcache will issue at most 0 or 1 spurious `Release()` calls -- and there could be more than one legacy allocation.

pkg/ipcache/metadata.go
```diff
@@ -272,8 +273,28 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
...
+				// If the previous ipcache entry for the prefix
+				// was not managed by this function, then the
+				// previous ipcache user to inject the IPCache
+				// entry retains its own reference to the
+				// Security Identity. Given that this function
+				// is going to assume responsibility for the
+				// IPCache entry now, this path must retain its
+				// own reference to the Security Identity to
+				// ensure that if the other owner ever releases
+				// their reference, this reference stays live.
```
aditighag, line 287: It's not entirely clear to me when the new API would try to allocate a new identity? Your PR description mentions the restore case, but if it's not an upgrade, why wouldn't identities be allocated/injected either by the old or the new API? Could it be that not all subsystems have been migrated to the new api?

aditighag (commented):
Thanks for the detailed comments. However, I lack context about the "new" internal API, so while I can make some sense of the issue based on your and Casey's comments, I don't fully appreciate the fix. Perhaps if there were a diagram explaining the lifecycle management involving allocation/release of identities, that might've helped. Two other reviewers with more context have already approved the change though, so I mainly have comments about testing and need a clarification.

I can see that couple of reporters have tested the fix. 👍 At the same time, it would be nice to have a test to ensure there are no regressions going forward. Maybe a test case can be crafted based on the sequence of events mentioned in Casey's comment?

joestringer:
> Thanks for the detailed comments. However, I lack context about the "new" internal API, so while I can make some sense of the issue based on your and Casey's comments, I don't fully appreciate the fix. Perhaps if there were a diagram explaining the lifecycle management involving allocation/release of identities, that might've helped. Two other reviewers with more context have already approved the change though, so I mainly have comments about testing and need a clarification.

I like the diagram idea, do you have some ideas about how to express lifecycle management as a diagram? Is there a canonical representation?

Broadly speaking the way it works is that there are a range of ipcache users who directly call `ipcache.Upsert()` or `ipcache.UpsertGeneratedIdentities()` with an identity. This means that the caller already discovered the identity and took a reference for that identity to ensure that it stays alive as long as it's used. In that case, the ipcache itself does not take responsibility for reference-counting the identities. Any of the other Upsert() variations like `UpsertMetadata()`, `UpsertPrefixes(), `UpsertLabels()` do not directly reference the identity from other packages. The ipcache API just provides a way for the ipcache itself to allocate and assign an identity to an IP. For these cases, the ipcache is responsible for reference-counting the identities correctly. This PR deals with the interaction between the two cases, which is internally resolved in pkg/ipcache. One of my todo list items is to add comments to deprecate the older APIs to discourage usage and clarify which ones should/shouldn't be used.

> I can see that couple of reporters have tested the fix. +1 At the same time, it would be nice to have a test to ensure there are no regressions going forward. Maybe a test case can be crafted based on the sequence of events mentioned in Casey's comment?

I agree about the testing. I have a PR that I'm working on that was able to point out the problem here: https://github.com/cilium/cilium/pull/27253 . Given that it's a pretty generic change, I'd like to explore how to expand it to cover a range of similar timing-related issues. That will take a bit more time, so I'll continue the work independent of this PR.

pkg/ipcache/metadata.go
joestringer: Sorry, I've been staring at this code for the last week so I forget to explain some of my assumptions :-) 

The new reference is acquired earlier in the function here:

https://github.com/cilium/cilium/blob/2a5f63244e75e4ac10c234567ae5f523c79c56e0/pkg/ipcache/metadata.go#L215

Under the `releaseIdentity` block here, it's added to this `previouslyAllocatedIdentities` map:

https://github.com/cilium/cilium/blob/2a5f63244e75e4ac10c234567ae5f523c79c56e0/pkg/ipcache/metadata.go#L275

Then references are released later in the function using this map:

https://github.com/cilium/cilium/blob/2a5f63244e75e4ac10c234567ae5f523c79c56e0/pkg/ipcache/metadata.go#L332

joestringer: > Could it be that not all subsystems have been migrated to the new api?

Work in progress, unfortunately there's a lot of tech debt to work through in order to introduce the changes in an incremental fashion:

https://github.com/cilium/cilium/issues/21142

christarazi:
@squeed Previously, you mentioned

> InjectLabels always does a new := Allocate() followed by a Release(existing). Most of the time, the identity for a set of labels never changes, so new == existing and the reference count is unchanged. In the case where the ID has changed, the existing ID will be properly released and cleaned up.

which is my understanding as well. 

If we are not confident that's the case, then this is a pretty easy thing to unit test.

joestringer:
> I'm wondering if there's an issue where InjectLabels() is called for a given prefix again. We will again call Allocate(), which will return the same ID, but will not call Release(). In fact, every time we call InjectLabels() we will leak another reference, right?

The new condition says "if the current entry in ipcache was not created by InjectLabels() *and* it's being replaced by one created by InjectLabels(), then we retain the reference". The inverse means that if the entry was already created by InjectLabels(), for instance in the case of repeated InjectLabels() calls, or if InjectLabels will not replace the entry, then the newly allocated reference will be released.

pkg/ipcache/metadata.go
joestringer: I've rewritten the commit + PR descriptions, hopefully that clarifies your questions.

squeed:
> _and_ it's being replaced by one created by InjectLabels(), then we retain the reference

Aha, that's the key. Should be OK then.

pkg/ipcache/metadata.go
aditighag: The missing context was that there is a mixed usage of the older and new APIs; you sort of indirectly confirmed it in your previous comment. Eventually, I suppose the goal is to transition all the internal subsystems to the new API. Thanks for adding this context to the PR description.

aditighag:
The PR was merged before I could take a follow-up look, but it looks like you addressed the main review comments. The updated PR description looks great. 🚀 

>I like the diagram idea, do you have some ideas about how to express lifecycle management as a diagram? Is there a canonical representation?

Something for follow-ups: Listing usages of the old and new APIs annotated with subsystems that are using the old API v/s the new, and the corresponding allocate/release calls can be helpful for future PRs as well knowledge sharing.

joestringer:
@aditighag good suggestion. Following up here: https://github.com/cilium/cilium/pull/21667

PR #27337, opened 2023-08-08
ipcache: propagate cluster ID as part of the key

giorio94:
Insert the cluster ID associated with a given CIDR as part of the ipcache key. No behavioral change is introduced when the cluster ID is 0, which is the default.

### DIFF
--- a/pkg/datapath/ipcache/listener.go
+++ b/pkg/datapath/ipcache/listener.go
@@ -139,7 +139,7 @@ func (l *BPFListener) OnIPIdentityCacheChange(modType ipcache.CacheModification,
 
 	// Update BPF Maps.
 
-	key := ipcacheMap.NewKey(cidr.IP, cidr.Mask, 0)
+	key := ipcacheMap.NewKey(cidr.IP, cidr.Mask, uint8(cidrCluster.ClusterID()))
 
 	switch modType {
 	case ipcache.Upsert:


christarazi (approved):

giorio94:
Codeowners reviews are in, tests are green. Marking as ready to merge.

brb:
@giorio94 Just to double check - in the ClusterMesh we don't have a direct pod to remote cluster pod access, i.e., the access is through a ClusterIP of a K8s service? Otherwise, with this change we disable the former access.

giorio94:
> @giorio94 Just to double check - in the ClusterMesh we don't have a direct pod to remote cluster pod access, i.e., the access is through a ClusterIP of a K8s service? Otherwise, with this change we disable the former access.

Clustermesh does support direct cross cluster pod to pod access, but this change doesn't break it because `cidrCluster.ClusterID()` always returns 0 with the current setup. Indeed, all connectivity checks successfully passed.

brb:
> but this change doesn't break it because cidrCluster.ClusterID() always returns 0

Is it because the `AddClusterFromIP()` instead of the `AddClusterFrom()` used everywhere?

brb:
>> Is it because the AddClusterFromIP() instead of the AddClusterFrom() used everywhere?

> Yes

Any idea why? AFAIK when we receive update about a CiliumEndpoint from a remote cluster we have the CID. So, we could populate it into the IPcache.

PR #23292, opened 2023-01-24
resource: Work around a rare race in initial sync

joamaki:
TestResource_WithFakeClient is sometimes failing due to seeing two Upserts instead of Upsert+Sync. This is due to the fact that we add the initial set of keys from the store to the queue when subscribing in Events() and the fact that adding to the store is happening before our event handler runs. The sequence of events looks like this:

  [informer 1/2] add object A to store
  [resource] subscribe and add object with key A to queue
  [resource] dequeue A and emit Event{Upsert, A}
  [informer 2/2] call AddFunc handler for A and queue A
  [resource] dequeue A and emit Event{Upsert, A}

Work around the issue by allowing the extra Upsert event in the test case.

Proper fix for this would be to either detect a duplicate using ResourceVersion, or implementing a custom (cache.Config) Process function that synchronizes the updating of the store with subscribing (e.g. pkg/k8s/informer/informer.go).

Fixes: #23079
Fixes: 4101e2c768 ("k8s: Add resource package")

PR #27340, opened 2023-08-08
resource: Fix race with event retries of deleted and subsequently recreated objects

joamaki:
See commits for detailed description.

```release-note
resource: Fix race condition in handling of Kubernetes object delete event retrying. In the very rare case when an object was created, deleted and re-created with the same name and the handling of the first deletion failed, the handling of delete event may have been retried even though the object was re-created. Only affected features using the Resource-library (LB IPAM, Mutual Auth and ClusterMesh).
```

### DIFF
--- a/pkg/k8s/resource/resource.go
+++ b/pkg/k8s/resource/resource.go
@@ -50,14 +50,42 @@ import (
 // The resource is lazy, e.g. it will not start the informer until a call
 // has been made to Events() or Store().
 type Resource[T k8sRuntime.Object] interface {
+	// Resource can be observed either via Observe() or via Events(). The observable
+	// is implemented in terms of Events() and same semantics apply.
 	stream.Observable[Event[T]]
 
 	// Events returns a channel of events. Each event must be marked as handled
-	// with a call to Done(), otherwise no new events for this key will be emitted.
+	// with a call to Done() which marks the key processed. No new events for this key
+	// will be emitted before Done() is called.
+	//
+	// A missing Done() will lead to an eventual panic (via finalizer on Event[T]).
+	// Panic on this situation is needed as otherwise no new events would be emitted
+	// and thus this needs to be enforced.
+	//
+	// A stream of Upsert events are emitted first to replay the current state of the
+	// store after which incremental upserts and deletes follow until the underlying
+	// store is synchronized after which a Sync event is emitted and further incremental
+	// updates:
+	//
+	//	(start observing), Upsert, Upsert, Upsert, (done replaying store contents), Upsert, Upsert,
+	//	  (store synchronized with API server), Sync, Upsert, Delete, Upsert, ...
+	//
+	// The emitting of the Sync event does not depend on whether or not Upsert events have
+	// all been marked Done() without an error. The sync event solely signals that the underlying
+	// store has synchronized and that Upsert events for objects in a synchronized store have been
+	// sent to the observer.
 	//
 	// When Done() is called with non-nil error the error handler is invoked, which
-	// can ignore, requeue the event or close the channel. The default error handler
+	// can ignore, requeue the event (by key) or close the channel. The default error handler
 	// will requeue.
+	//
+	// If an Upsert is retried and the object has been deleted, a Delete event will be emitted instead.
+	// Conversely if a Delete event is retried and the object has been recreated with the same key,
+	// an Upsert will be emitted instead.
+	//
+	// If an objects is created and immediately deleted, then a slow observer may not observe this at
+	// all. In all cases a Delete event is only emitted if the observer has seen an Upsert. Whether or
+	// not it had been successfully handled (via Done(nil)) does not affect this property.
 	Events(ctx context.Context, opts ...EventsOpt) <-chan Event[T]
 
 	// Store retrieves the read-only store for the resource. Blocks until
--- a/pkg/k8s/resource/resource.go
+++ b/pkg/k8s/resource/resource.go
@@ -13,13 +13,16 @@ import (
 	"sync"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 	k8sRuntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 
 	"github.com/cilium/cilium/pkg/hive"
 	k8smetrics "github.com/cilium/cilium/pkg/k8s/metrics"
+	"github.com/cilium/cilium/pkg/k8s/watchers/resources"
 	"github.com/cilium/cilium/pkg/lock"
 	"github.com/cilium/cilium/pkg/metrics"
 	"github.com/cilium/cilium/pkg/promise"
@@ -106,9 +109,9 @@ type Resource[T k8sRuntime.Object] interface {
 // See also pkg/k8s/resource/example/main.go for a runnable example.
 func New[T k8sRuntime.Object](lc hive.Lifecycle, lw cache.ListerWatcher, opts ...ResourceOption) Resource[T] {
 	r := &resource[T]{
-		queues: make(map[uint64]*keyQueue),
-		needed: make(chan struct{}, 1),
-		lw:     lw,
+		subscribers: make(map[uint64]*subscriber[T]),
+		needed:      make(chan struct{}, 1),
+		lw:          lw,
 	}
 	r.opts.sourceObj = func() k8sRuntime.Object {
 		var obj T
@@ -191,8 +194,8 @@ type resource[T k8sRuntime.Object] struct {
 
 	needed chan struct{}
 
-	queues map[uint64]*keyQueue
-	subId  uint64
+	subscribers map[uint64]*subscriber[T]
+	subId       uint64
 
 	lw           cache.ListerWatcher
 	synchronized bool // flipped to true when informer has synced.
@@ -256,28 +259,6 @@ func (r *resource[T]) metricEventReceived(action string, valid, equal bool) {
 	metrics.KubernetesEventReceived.WithLabelValues(r.opts.metricScope, action, validStr, equalStr).Inc()
 }
 
-func (r *resource[T]) pushUpdate(key Key) {
-	r.metricEventReceived("update", true, false)
-	r.mu.RLock()
-	for _, queue := range r.queues {
-		queue.AddUpsert(key)
-	}
-	r.mu.RUnlock()
-}
-
-func (r *resource[T]) pushDelete(lastState any) {
-	key := NewKey(lastState)
-	obj := lastState
-	if d, ok := lastState.(cache.DeletedFinalStateUnknown); ok {
-		obj = d.Obj
-	}
-	r.mu.RLock()
-	for _, queue := range r.queues {
-		queue.AddDelete(key, obj)
-	}
-	r.mu.RUnlock()
-}
-
 func (r *resource[T]) Start(hive.HookContext) error {
 	r.wg.Add(1)
 	go r.startWhenNeeded()
@@ -306,18 +287,7 @@ func (r *resource[T]) startWhenNeeded() {
 		return
 	}
 
-	// Construct the informer and run it.
-	handlerFuncs :=
-		cache.ResourceEventHandlerFuncs{
-			AddFunc:    func(obj any) { r.pushUpdate(NewKey(obj)) },
-			UpdateFunc: func(old any, new any) { r.pushUpdate(NewKey(new)) },
-			DeleteFunc: func(obj any) { r.pushDelete(obj) },
-		}
-
-	store, informer := cache.NewTransformingIndexerInformer(
-		r.lw, r.opts.sourceObj(), 0, handlerFuncs,
-		r.opts.indexers, r.opts.transform,
-	)
+	store, informer := r.newInformer()
 	r.storeResolver.Resolve(&typedStore[T]{store})
 
 	r.wg.Add(1)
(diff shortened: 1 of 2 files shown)


pkg/k8s/resource/resource_test.go
```diff
@@ -256,6 +258,112 @@ func TestResource_WithFakeClient(t *testing.T) {
...
+	}()
+
+	var lastVersion uint64
+	var lastNode *corev1.Node
+
+	for ev := range events {
+		if ev.Kind == resource.Delete {
+			// Check that we don't go back in time
+			version, _ := strconv.ParseUint(ev.Object.ObjectMeta.ResourceVersion, 10, 64)
+			require.LessOrEqual(t, lastVersion, version, "expected always increasing ResourceVersion")
```
bimmlerd: `LessOrEqual` vs `always increasing` - is equal ok or is it not? I read always increasing as also not constant, which would be allowed by the check

pkg/k8s/resource/resource_test.go
```diff
@@ -256,6 +258,112 @@ func TestResource_WithFakeClient(t *testing.T) {
...
+			lastVersion = version
+
+			lastNode = nil
+			if rand.Intn(3) == 0 {
+				ev.Done(errors.New("delete failed"))
+			} else {
+				ev.Done(nil)
+			}
+		} else {
+			lastNode = ev.Object
```
bimmlerd: probably deserves a `require non-nil check` here already?

pkg/k8s/resource/resource.go
```diff
@@ -483,21 +491,27 @@ func (r *resource[T]) Events(ctx context.Context, opts ...EventsOpt) <-chan Even
...
-				event.Object = entry.obj.(T)
-			case upsertEntry:
+			case upsertOrDeleteEntry:
 				obj, exists, err := store.GetByKey(entry.key)
-				// If the item didn't exist, then it's been deleted and a delete event will
-				// follow soon.
-				if err != nil || !exists {
-					event.Done(nil)
-					continue loop
+				if !exists || err != nil {
```
bimmlerd: I think this is still racy.

Imagine the sequence `create (C) delete (D) create (C) delete (D)` for the same resource coming from k8s. The end state should be either four, two or no events emitted (CDCD, CD or nothing), but never starting with a D (deletion of not created resource), or ending with C (end state wrong, created instead of deleted). 
However, I claim that the current code can result in `DD`, i.e. double deletion of a resource we have not emitted a create for. The key queue coalesces `upsertOrDeleteEntry{"name"}` as long as we don't call `Get`. Hence, I can choose to let it coalesce the `CD` to just one, with which we enter the loop. Before we get to `store.GetByKey`, however, the third and fourth action `CD` have hit the store, which now does not contains the resource, and `deletedObjs` does. Hence, we emit just `D`, deleting a resource we didn't create.
Since the object in the `deletedObj` doesn't match, in the call to event.Done we do not remove it. The next call to `kq.Get()` thus relivers the `upsertOrDeleteEntry{"name"}` coalesced from the third and fourth action, and we emit another `D` event, since the store doesn't carry the resource, and the deletedObjs still do.

bimmlerd (changes requested):
I think there's a race.

odinuge:
Is there an issue tracking the issues this is causing, and the (potential) consequences of this @joamaki?

Also, do you know what versions of cilium this affects?

pkg/k8s/resource/resource_test.go
joamaki: Must be `LessOrEqual` due to the retries.

pkg/k8s/resource/resource.go
bimmlerd: I've thought about this some more. I don't think there can be a sequence of k8s events in which the end state is 'deleted' in which the resource emits a sequence which ends in 'created'. Formulated differently, the race described above does not exist in the inverted (and plausibly worse) incarnation of missing the deletion of an object.

I think it's invariant that we emit creation events if and only if (iff) the store contains an object for a key. It's a guarantee of the workqueue that it will not coalesce into events which were observed (via `Get`). Further, for the sequence of events emitted by the resource to end with `C`, but the sequence of k8s events to end with `D`, the final `D` event must somehow be missed by the resource. The resource can only 'miss' events if they are coalesced in the key queue, but then we have the invariant that the store must have seen it (as it's the callback of the store to push it into the key queue). Hence, the store can't contain the object and also coalesce the event. Thus, the resource either observes the non-existence of the obj in the store, or gets the delete event. Therefore, the resource doesn't end a sequence of emitted events with `C` if it shouldn't. 

One step further, it seems an acceptable burden to place on the consumer to ignore (from their perspective) spurious deletes. Note that I think the resource shouldn't not emit (i.e. should emit) deletes for objects it hasn't emitted creates for, since consumers may have other (possibly indirect) sources of knowledge about an object and hence might know about it's creation even though this resource didn't provide an event.

joamaki: Agreed on this. I'll look into tracking the emitted creations so deletions will always pair up with the create.

joamaki: Fixed. Creations are now tracked and Resource[T] will only emit a delete event for objects that the subscriber has observed, e.g. if `create X`, `delete X` occurs in quick succession and the subscriber didn't see the `create X` it won't see the `delete X`.

pkg/k8s/resource/resource.go
```diff
@@ -454,6 +454,10 @@ func (r *resource[T]) Events(ctx context.Context, opts ...EventsOpt) <-chan Even
 				debugInfo))
 		}
 
+		// The set of objects emitted to the subscriber. Used to avoid emitting a delete
```
bimmlerd: Can't comment there, but I think the `Events` function doc comment should state the semantics around retries, given it's pretty subtle

pkg/k8s/resource/resource.go
```diff
@@ -464,44 +468,74 @@ func (r *resource[T]) Events(ctx context.Context, opts ...EventsOpt) <-chan Even
...
+							// Subscriber had not seen the creation of this object, ignore the deletion.
+							queue.deletedObjects.CompareAndDelete(entry.key, deletedObj)
+							queue.Done(raw)
+							continue loop
+						}
+						event.Kind = Delete
+						event.Key = entry.key
+						event.Object = deletedObj.(T)
+					} else {
+						// Here we hit a race in which the object was removed from store
```
bimmlerd: We could theoretically have also hit a case where `err != nil` from getting from the store (though I don't think the store can actually error). Nonetheless, if there was a persitent error, I think this would continually requeue the same event

pkg/k8s/resource/resource_test.go
```diff
@@ -256,6 +258,110 @@ func TestResource_WithFakeClient(t *testing.T) {
...
+			lastVersion = version
+
+			if rand.Intn(3) == 0 {
+				ev.Done(errors.New("delete failed"))
+			} else {
+				exists = false
+				ev.Done(nil)
+			}
+		} else if ev.Kind == resource.Upsert {
+			exists = true
```
bimmlerd, line 349: Could also check the version not going backwards here, since we expect to observe the upserts in order

pkg/k8s/resource/resource.go
```diff
@@ -464,44 +468,74 @@ func (r *resource[T]) Events(ctx context.Context, opts ...EventsOpt) <-chan Even
...
-				queue.eventDone(entry, err)
-				r.metricEventProcessed(event.Kind, err == nil)
-			}
-
-			// Add a finalizer to catch forgotten calls to Done().
-			runtime.SetFinalizer(eventDoneSentinel, doneFinalizer)
+			var event Event[T]
 
 			switch entry := entry.(type) {
 			case syncEntry:
```
bimmlerd: Kind of unrelated to the issue we're solving here, but should be documented at least: A consumer will receive a sync event regardless of whether the handling of the upserts was successful. That is, during the replay of the stuff in the store, if you fail to handle an event, you will receive the `Sync` before receiving the retried Upsert.

bimmlerd (approved):
Couple of nits, but LGTM!

pkg/k8s/resource/resource.go
joamaki: Currently the store used here can't error since it's a simple in-memory hash map. I think the idea with having errors here is to be able to use e.g. a store that persists to disk. I'll change this to explicitly handle the unexpected error with a `Fatal`.

pkg/k8s/resource/resource_test.go
joamaki: Added!

pkg/k8s/resource/resource.go
joamaki: Good point. Added documentation about this to `Events()`.

pkg/k8s/resource/resource.go
joamaki: Documented.

joamaki: Documented.

pkg/k8s/resource/resource.go
```diff
@@ -454,6 +454,10 @@ func (r *resource[T]) Events(ctx context.Context, opts ...EventsOpt) <-chan Even
 				debugInfo))
 		}
 
+		// The set of objects emitted to the subscriber. Used to avoid emitting a delete
+		// event for an object the subscriber has not seen.
+		var knownObjects sync.Map
```
aanm: it doesn't need to be sync.Map, does it? From a first glance it looks it's just one go routine accessing it?

pkg/k8s/resource/resource.go
```diff
@@ -536,21 +570,28 @@ func (r *resource[T]) Events(ctx context.Context, opts ...EventsOpt) <-chan Even
...
 	kq.Add(syncEntry{})
 }
 
 func (kq *keyQueue) AddUpsert(key Key) {
+	kq.deletedObjects.Delete(key)
+
 	// The entries must be added by value and not by pointer in order for
 	// them to be compared by value and not by pointer.
-	kq.Add(upsertEntry{key})
+	kq.Add(upsertOrDeleteEntry{key})
```
aanm: I would suggest to leave another comment in here explaining why an "Add" is passing an "upsertOrDelete" entry

aanm (changes requested):

youngnick (approved):
LGTM with @aanm's changes.

pkg/k8s/resource/resource.go
joamaki: It has to be a `sync.Map` (or map + mutex) since it's mutated from the `Done` function that may be called from anywhere by the subscriber. 

And honestly the only reason I really went with `sync.Map` here was to match with `keyQueue.deletedObjects` (which actually needed `sync.Map` for compare-and-swap) and because it was shorter. Could go with `sets.Set[Key]` + `Mutex` instead if it'd be preferred. Let me know what you think. I'll at the very least expand the comment to say that it's accessed concurrently.

pkg/k8s/resource/resource.go
joamaki: `kq` is the workqueue, so we're adding an entry into the queue. It's not `Add` of an object (e.g. not the `T` in `Resource[T]` that is being upserted or deleted), but it's an add of a workqueue entry. Do you still want me to add a comment? What should the comment say?

Or what do you think about switching to `Enqueue` instead? It won't match the workqueue API it's wrapping, but it would be less confusing.

aanm: > Or what do you think about switching to Enqueue instead? It won't match the workqueue API it's wrapping, but it would be less confusing.

Oh, I think this would definitely sound less confusing :)

joamaki: oh wait, that's in client-go's workqueue, I can't change that. I'll make it less confusing by not embedding the `RateLimitingInterface`.

bimmlerd:
Is there a way https://github.com/cilium/cilium/issues/27549 could be triggered by this?

joamaki:
> Is there a way #27549 could be triggered by this?

Don't think so. Probably same as with https://github.com/cilium/cilium/pull/23292. 

We should perhaps look into doing what `pkg/k8s/informer/informer.go` does and implement the 
`Process` function ourselves to make sure we mutate the cache.Store and queue work under the same lock to avoid the situation where a new subscriber will see a key via IterKeys(), process it, and then reprocess the same event due to the ResourceEventHandler queueing it.

pkg/k8s/resource/resource.go
```diff
@@ -522,100 +447,266 @@ func (r *resource[T]) Events(ctx context.Context, opts ...EventsOpt) <-chan Even
...
+					}
+				} else {
+					obj = d.Object
+				}
 
-type deleteEntry struct {
-	key Key
-	obj any
+				// In CI we detect if the objects were modified and panic
+				// this is a no-op in production environments.
```
bimmlerd: Is this actually still true, or just copy pasta?

bimmlerd (approved):
Wow, nice simplification.

pkg/k8s/resource/resource.go
joamaki:
```
test/helpers/kubectl.go:                // "extraEnv[0].name":              "KUBE_CACHE_MUTATION_DETECTOR",
```

https://github.com/kubernetes/client-go/blob/v0.28.1/tools/cache/mutation_detector.go#L36

Heh, looks like we no longer set it anywhere, so no we don't check in CI.  And having it enabled by default is a fairly big hammer. I'll drop this from this fix as thinking about where and how to enable this is not trivial.

aanm (approved):
Can we adapt the pkg/k8s/informer.NewInformerWithStore for the pkg/k8s/resource use cases and use it?

Also, we should keep the mutator cache detector code, it was accidentally removed. I'm going to re-add it.

joamaki:
> Can we adapt the pkg/k8s/informer.NewInformerWithStore for the pkg/k8s/resource use cases and use it?
> 
> Also, we should keep the mutator cache detector code, it was accidentally removed. I'm going to re-add it.

How do I do this with it: https://github.com/cilium/cilium/pull/27340/files#diff-ae3d7e3fabf6dbbdcdd0d826fe38147c28883c23eb2f1ae43f5b80f54f830a7bR699 ?

PR #27341, opened 2023-08-08
ci: set timeout on build images workflows

mhofstetter:
Due to the lack of a global timeout in the build images GHA workflows, some steps might take forever.

Therefore, this commit introduces a global timeout of 45 minutes. (already used in `build-images-base.yaml`)

Example of a hanging workflow run: https://github.com/cilium/cilium/actions/runs/5793333705/job/15700948511


viktor-kurchenko (approved):
LGTM.

PR #27342, opened 2023-08-08
renovate: Don't exclude github.com/{cilium,vishvananda}/netlink anymore

ghost:
The netlink library was excluded from renovate updates in d6fe7de12cd1 ("renovate: exclude github.com/{cilium,vishvananda}/netlink"), because at that point Cilium was using its fork, but then in 9f0be9ecbe73 ("vendor: Update vishvananda/netlink/") it was switched back to the upstream version. Update the renovate configuration accordingly to no longer exclude netlink from updates.

The d6fe7de12cd1 commit message says that it should be switched back, but it seems it hasn't been done yet. I think it needs to be backported to v1.14 too?

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -128,8 +128,6 @@
         // want to update them automatically.
         "go.universe.tf/metallb",
         "github.com/cilium/metallb",
-        "github.com/vishvananda/netlink",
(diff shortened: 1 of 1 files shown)


tklauser (approved):
Thanks. Yeah, let's backport this to v1.14.

PR #27346, opened 2023-08-08
docs: Harmonise references to Cilium Slack

qmonnet:
We link to Cilium Slack at multiple locations in Cilium's documentation. The way we do it is far from consistent:

- We usually provide a direct link to the Slack instance
- ... Which, over time, resulted in at least 13 occurrences of the same URL https://cilium.herokuapp.com
- Sometimes, we create an indirect link to a `_slack` target, pointing to the `gettinghelp.rst` page
- Some other times, we use other indirect links pointing to the glossary.
- At some locations, we also mention that users can get invites to the Slack instance. As far as I know, these invites are no longer necessary.

Let's try to clean this up.

- Create a target link in the RST epilogue in `conf.py`, so we can reference it from any location.
- Call this link everywhere we need, using `` `Cilium Slack`_ `` in a consistent fashion.
- Format most channel names as inline literals.
- Improve formatting (not the content) of some paragraphs that already get updates for Slack references: rewrap, remove trailing spaces, etc.

Marking for backports because it touches the doc framework (`conf.py`, `check-build.sh`).

No change to technical contents from the docs.


nebril (approved):

dylandreimerink (approved):
LGTM

ghost (approved):
awesome

PR #27347, opened 2023-08-08
Update renovate configuration for ginkgo and kindest/node

aanm:
read per commit

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -8,6 +8,7 @@
   // This ensures that the gitAuthor and gitSignOff fields match
   "gitAuthor": "renovate[bot] <bot@renovateapp.com>",
   "includePaths": [
+    ".github/actions/ginkgo/**",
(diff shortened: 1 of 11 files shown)


nebril (approved):

.github/renovate.json5
```diff
@@ -371,11 +374,28 @@
         "on monday"
       ],
     },
+    {
+      "groupName": "all kind-images main",
+      "groupSlug": "all-kind-images-main",
```
joestringer, line 379: @aanm these names seem to imply they only apply to the `main` branch, but this definitely seems to be applying to all branches. For example: 

https://github.com/cilium/cilium/pull/27772

Is that what `main` is supposed to mean here? Should this section have a stanza to restrict the `matchBaseBranches` to `main`?

aanm: @joestringer No, I think the "main" was a bad copy paste.

PR #27351, opened 2023-08-08
Support for batch deletion of endpoints

tklauser:
This PR adds a new endpoint batch delete request to the API which allows the deletion of all endpoints associated with a particular containerID, which simplifies the `CNI DEL` logic for the CNI plugin if more than one endpoint are attached to a particular container.

This is used to support batch deletion in the endpoint deletion queue. The endpoint deletion queue is used to enqueue endpoint deletion requests in `CNI DEL` when cilium-agent is unavailable.
    
Support for batch deletion in the deletion queue is particularly important, because if cilium-agent is unavailable, we cannot query it for endpoints which belong to a particular container.
    
Before this change, a endpoint deletion queue entry just contained a string that represented a unique endpoint identifier. To support more flexible batch deletion requests, we now instead serialize the full HTTP API `EndpointDeleteRequest` as JSON. The new format is more flexible, as it allows us to add more fields in the future. The old format (the endpoint ID as a plain string) is still supported for backwards-compatibility.

See commits for details.

Joint work with @gandro


aanm (approved):

christarazi (approved):

api/v1/openapi.yaml
```diff
@@ -249,6 +249,29 @@ paths:
           description: Endpoints with provided parameters not found
         '429':
           description: Rate-limiting too many requests in the given time frame
+    delete:
```
aditighag, line 252: Can we rename this to `deleteList` to make the intent more clear? There is already a `delete` under `endpoint/id` api - https://github.com/aditighag/cilium/blob/master/api/v1/openapi.yaml#L170.

api/v1/openapi.yaml
```diff
@@ -249,6 +249,29 @@ paths:
...
+      summary: Deletes a list of endpoints
+      description: |
+        Deletes a list of endpoints that have endpoints matching the provided properties
```
aditighag, line 255: How does a user know what the currently supported properties are?

Edit: Adding a reference to the below `EndpointDeleteRequest` field should help.

api/v1/openapi.yaml
```diff
@@ -1348,6 +1377,14 @@ definitions:
         type: boolean
       datapath-configuration:
         "$ref": "#/definitions/EndpointDatapathConfiguration"
+  EndpointDeleteRequest:
```
aditighag: This is a bit confusing as the commit is adding a new API to delete a batch of endpoints. There is already an API to delete endpoints by ID. Can we clearly differentiate between the two?

aditighag (changes requested):
Thanks for articulating the context in the commit description. My comments mainly pertain to differentiating between the existing API to delete an endpoint by id and the new batch api.

ghost (approved):
docs ok

api/v1/openapi.yaml
tklauser: The `delete` here refers to the `DELETE` HTTP request method, so I don't think we can change this.

api/v1/openapi.yaml
tklauser: This is already referenced in the `parameters` section below, ref. https://github.com/cilium/cilium/blob/da64b11ce28ce83f95b6221ca11d50991d635ac4/api/v1/openapi.yaml#L258-L259

It shows up as follows in the Swagger editor documentation:

image

api/v1/openapi.yaml
tklauser: Good point. I've renamed it to `EndpointBatchDeleteRequest` (and the parameter type from `endpoint-delete-request` to `endpoint-batch-delete-request`).

nebril (approved):

tklauser:
Some workflows failed because the images weren't built due to a quay glitch: https://github.com/cilium/cilium/actions/runs/5819382359/attempts/1?pr=27351 I've manually re-triggered the image build and the workflows that timed out waiting for images.

api/v1/openapi.yaml
aditighag: Oh nice! TIL about the online editor.

aditighag (approved):

PR #27366, opened 2023-08-09
bpf: nat: share rewrite logic in RevSNAT path

julianwiedmann:
ICMP_FRAG_NEEDED / ICMPV6_PKT_TOOBIG packets contain the original packet,
which potentially needs to be RevSNATed. We then use the obtained RevSNAT
entry to also rewrite the outer ICMP packet (note that a lookup for the
ICMP tuple would find no revSNAT entry).

For rewriting the outer ICMP packet, return the RevSNAT entry and just
follow the normal packet path. This reduces program size.

### DIFF
--- a/bpf/lib/nat.h
+++ b/bpf/lib/nat.h
@@ -986,15 +986,14 @@ snat_v4_nat(struct __ctx_buff *ctx, struct ipv4_ct_tuple *tuple, int off,
 }
 
 static __always_inline __maybe_unused int
-snat_v4_rev_nat_handle_icmp_frag_needed(struct __ctx_buff *ctx, __u64 off)
+snat_v4_rev_nat_handle_icmp_frag_needed(struct __ctx_buff *ctx, __u64 off,
+					struct ipv4_nat_entry **state)
 {
 	struct ipv4_ct_tuple tuple = {};
-	struct ipv4_nat_entry *state;
 	struct iphdr iphdr;
 	__be16 identifier;
 	__u8 type;
 	__u32 icmpoff = off + sizeof(struct icmphdr);
-	int ret;
 
 	/* According to the RFC 5508, any networking equipment that is
 	 * responding with an ICMP Error packet should embed the original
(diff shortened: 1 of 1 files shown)


bpf/lib/nat.h
```diff
@@ -1042,31 +1041,25 @@ snat_v4_rev_nat_handle_icmp_frag_needed(struct __ctx_buff *ctx, __u64 off)
...
 	 * should point to endpoint that initiated the packet, as-well if
 	 * dest port had been NATed.
 	 */
-	ret = snat_v4_icmp_rewrite_ingress_embedded(ctx, &tuple, state,
-						    off, icmpoff);
-	if (IS_ERR(ret))
-		return ret;
-
-	/* Switch back to the outer header. */
-	tuple.nexthdr = IPPROTO_ICMP;
```
gentoo-root, line 1059: Did this rely on outer saddr == inner daddr and outer daddr == inner saddr?

bpf/lib/nat.h
```diff
@@ -1108,7 +1101,14 @@ snat_v4_rev_nat(struct __ctx_buff *ctx, const struct ipv4_nat_target *target,
...
 			if (icmphdr.code != ICMP_FRAG_NEEDED)
 				return NAT_PUNT_TO_STACK;
-			return snat_v4_rev_nat_handle_icmp_frag_needed(ctx, off);
+
+			ret = snat_v4_rev_nat_handle_icmp_frag_needed(ctx, off, &state);
+			if (IS_ERR(ret))
+				return ret;
+
+			/* Skip L4 rewrite: */
+			tuple.dport = state->to_dport;
```
gentoo-root: Could you elaborate why it's valid to skip L4 rewrite? I'm probably missing something, but that doesn't look equivalent to the old code.

julianwiedmann: The're no port or such in the outer `ICMP_FRAG_NEEDED` header that we need to rewrite. Right now `snat_v4_rewrite_ingress()` would access the ICMP header again, determine the type, and skip the processing (since it's not `ICMP_ECHOREPLY`). So we can also avoid the header access in this case.

But you're right, it's not exactly equivalent. And worth its own patch I suppose. So let me back it out.

bpf/lib/nat.h
julianwiedmann: Yep, it assume that the outer headers match the inner headers.

Which isn't terribly robust - it would be a nice improvement to explicitly check whether we *can* rewrite the outer header with the inner packet's SNAT entry  (and drop otherwise)

bpf/lib/nat.h
gentoo-root: Yeah, I now realize it was a terribly stupid question, thanks for explaining it, somehow I didn't realize it back then :)

> And worth its own patch I suppose. So let me back it out.

Did you move it to some other pull request?

gentoo-root (approved):

bpf/lib/nat.h
julianwiedmann: > Did you move it to some other pull request?

I'll pick it up (or something equivalent) with a later PR, yep. Still plenty of NAT rework to do :grin: .

PR #27367, opened 2023-08-09
endpoint, endpointmanager: Publish max policymap size as metric

### LINKED ISSUE #27332: Implement metric for endpoint policymaps that reports the max policymap per node
From #21508:

>   - Be aware that policymaps are per-endpoint, so if we just export the name directly into prometheus metrics then it will significantly increase cardinality. It is worth thinking about how this metric could be exported across a cluster with 1000s of workloads.
>  - We may want to consider using something like a histogram for policymap pressure per node, so that we can identify the distribution of policymap usage across the endpoints on the node.

christarazi:
This commit reworks the cilium_bpf_map_pressure metric such that the max
policymap size from all endpoints is published. This reduces the
cardinality of the metric because each individual endpoint policymap was
a unique label on the metric. In other words, this commit collapses the
cardinality down to just one.

---

Tasks:

- [x] Manual testing
- [x] Update docs
- [x] Update release note
- [x] Remove unnecessary logs
- [x] Document why 10s for trigger interval
- [x] Decide on label name (i.e. `cilium_policy_*`?)

Fixes: https://github.com/cilium/cilium/issues/27332

```
`cilium_bpf_map_pressure` now has reduced cardinality for endpoint policymaps. Only one label is exposed representing the maximum policymap pressure on the node.
```


pkg/endpointmanager/policymap_pressure.go
```diff
@@ -0,0 +1,66 @@
...
+		return
+	}
+	// mgr.policyMapPressureGauge = metrics.NewBPFMapPressureGauge(policymap.MapName, policymap.PressureMetricThreshold)
+	mgr.policyMapPressureGauge = metrics.NewBPFMapPressureGauge(policymap.MapName, 0.0)
+
+	go mgr.readPolicyMapPressureEvents()
+
+	var err error
+	src, emit, _ := stream.Multicast[endpoint.MapPressureEvent]() // ignoring complete() here as there's no single emitter that can complete it
+	mgr.policyMapPressureEventCh = stream.ToChannel(mgr.ctx, src)
```
christarazi: Buffered channel?

pkg/endpointmanager/policymap_pressure.go
```diff
@@ -0,0 +1,66 @@
...
+			// TODO: Do we need to define a threshold for the comparison?
+			if cur == nil || val > *cur {
+				mgr.currentPolicyMapPressure.Store(&val)
+			}
+			mgr.policyMapPressureTrigger.Trigger()
+		case <-mgr.ctx.Done():
+			log.Infof("EndpointManager policymap event loop terminated")
+			return
+		}
+	}
```
christarazi: If we use buffered channels, then the loop should be written as 

```
for range mgr.policyMapPressureEventCh
```

and when it's closed, the loop terminates. If the context is cancelled, we should drain the channel as well.

However, this is starting to sound like pkg/eventqueue. :thinking:

pkg/endpointmanager/policymap_pressure.go
```diff
@@ -0,0 +1,65 @@
...
+	if !metrics.BPFMapPressure {
+		return
+	}
+	mgr.policyMapPressureGauge = metrics.NewBPFMapPressureGauge(policymap.MapName, 0.0)
+
+	go mgr.readPolicyMapPressureEvents()
+
+	var err error
+	src, emit, _ := stream.Multicast[endpoint.MapPressureEvent]() // ignoring complete() here as there's no single emitter that can complete it
+	mgr.policyMapPressureEventCh = stream.ToChannel(mgr.ctx, src)
```
joamaki: I would just do `src.Observe(mgr.ctx, mgr.onPolicyMapPressureEvent, func(error) {})` here instead of going via a channel. Then one doesn't need to worry about draining the channel etc. concerns.

Since there's only once consumer for the `MapPressureEvent`, do we actually benefit at all from the observable? Essentially "Multicast" gives you a many-to-many "callback". If you've got many-to-one, I'd probably just call a method, e.g. `EndpointManager.UpdatePolicyMapPressure(...)` (or `PolicyMapPressure.Update` implemented by EndpointManager if cyclic imports is a concern).

pkg/endpointmanager/policymap_pressure.go
```diff
@@ -0,0 +1,65 @@
...
+			val := ev.Value
+			log.Infof("EndpointManager policymap received event: %s", val)
+			cur := mgr.currentPolicyMapPressure.Load()
+			// TODO: Do we need to define a threshold for the comparison?
+			if cur == nil || val > *cur {
+				mgr.currentPolicyMapPressure.Store(&val)
+				mgr.policyMapPressureTrigger.Trigger()
+			}
+		case <-mgr.ctx.Done():
+			log.Infof("EndpointManager policymap event loop terminated")
```
joamaki: This won't drain the channel and will block the emitter.

pkg/endpointmanager/policymap_pressure.go
christarazi: Removed channel and usage of `pkg/stream`

pkg/endpointmanager/policymap_pressure.go
christarazi: Removed channel and usage of `pkg/stream`

pkg/endpointmanager/policymap_pressure.go
christarazi: Removed channel and usage of `pkg/stream`

pkg/endpointmanager/policymap_pressure.go
christarazi: Removed channel and usage of `pkg/stream`. Rewrote it such that the call to update the metric is behind an interface that the endpointmanager implements.

pkg/endpointmanager/policymap_pressure.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/cilium/cilium/pkg/trigger"
+)
+
+func (p *policyMapPressure) Update(ev endpoint.PolicyMapPressureEvent) {
+	val := ev.Value
+
+	log.WithField(logfields.Value, val).Debug("EndpointManager policymap received event")
+
+	cur := p.current.Load()
+	// TODO: Do we need to define a threshold for the comparison?
```
qmonnet: Does this need to be addressed before merging?

qmonnet (approved):
Looks good from my side, thanks!

pkg/endpointmanager/policymap_pressure.go
christarazi: Dropped it. I think it's has minimal consequence anyway.

joamaki (approved):

pkg/endpointmanager/policymap_pressure.go
```diff
@@ -0,0 +1,80 @@
...
+	p.trigger, err = trigger.NewTrigger(trigger.Parameters{
+		// It seems like 10s is a small enough window of time where the user
+		// can still reasonably react to a rising BPF map pressure. Keep it
+		// below the default Prometheus scrape interval of 15s anyway.
+		MinInterval: 10 * time.Second,
+		TriggerFunc: func([]string) { p.update() },
+		Name:        "endpointmanager-policymap-max-size-metrics",
+	})
+	if err != nil {
+		log.WithError(err).Panic("Failed to initialize trigger for policymap pressure metric")
```
aditighag, line 47: It's not common to panic in the agent. Why is this an exception?

aditighag (commented):
>The ``cilium_bpf_map_pressure`` for policy maps is now exposed as a single
  label ``cilium_policy_*``, rather than a label per policy map of an endpoint.

There is some context missing in the PR. Why is an overall pressure metric preferred over a per endpoint metric? As a user, it would be beneficial to know the top-k endpoint policy maps under pressure. What if there are multiple endpoint policy maps experiencing similar pressure?
It would also be useful to add a sample snippet for the updated map pressure metric.

pkg/endpointmanager/policymap_pressure.go
christarazi: In this case, the only way this panic gets triggered is if the `TriggerFunc` is nil. Given this code will initialize very early on in the daemon, it acts as a sanity check before proceeding. It is assumed that if this code panics, then it was a developer mistake. There is no risk that this panics in production because it is only called once on dameon init and it will be caught immediately in CI. Furthermore, the endpointmanager is not created conditionally nor is that likely to change in the future; in other words, this init code will always execute meaning if the `TriggerFunc` is ever nil, it will be caught.

christarazi:
@aditighag 

> There is some context missing in the PR. Why is an overall pressure metric preferred over a per endpoint metric? As a user, it would be beneficial to know the top-k endpoint policy maps under pressure. What if there are multiple endpoint policy maps experiencing similar pressure?

Currently, the `cilium_bpf_map_pressure` metric is per map. However, this can be cumbersome from a user's point of view because they will specifically need to write a query to aggregate all endpoint policymaps per node. Users want to know generally if their policymaps are filling up, primarily. To enable this, it brings me to the next point: imagine there are 1000s of endpoints in a cluster. This specific metric would then scale with the number of endpoints. Users will definitely turn it off because observability stacks charge money per metric and a metric with unique labels (cardinality) is considered a whole new metric. This is actually the primary reason for why we've kept the metric disabled until v1.14 (https://github.com/cilium/cilium/pull/24721). While I couldn't find a reference to back up this claim, I do recall it from memory of offline discussions with folks.

Converting the metric into a pre-aggregated max value per node enables users to actually turn on this useful metric. For now, users are disabling it. While I agree, it can be very useful to understand which endpoint is facing policymap issues, it's better to enable users to benefit from the metric to begin with and then we can improve it.

Speaking of improving it, one idea we could consider is to log each time a new policymap becomes the new max. This way users can correlated the metric with the policymap. However, the risk here how much it could spam the logs. Maybe some rate limiting could help. Suggestions welcome :slightly_smiling_face:.

aditighag:
> @aditighag
> 
> > There is some context missing in the PR. Why is an overall pressure metric preferred over a per endpoint metric? As a user, it would be beneficial to know the top-k endpoint policy maps under pressure. What if there are multiple endpoint policy maps experiencing similar pressure?
> 
>  imagine there are 1000s of endpoints in a cluster. This specific metric would then scale with the number of endpoints. Users will definitely turn it off because observability stacks charge money per metric and a metric with unique labels (cardinality) is considered a whole new metric. 

That's a good point! Prometheus has a set of [aggregation operators](https://prometheus.io/docs/prometheus/latest/querying/operators/), but I suppose users would still incur charges for every published metric?
Anyway I wasn't suggesting to publish the metric at a per endpoint level either (see below). 

> Speaking of improving it, one idea we could consider is to log each time a new policymap becomes the new max. This way users can correlated the metric with the policymap. However, the risk here how much it could spam the logs. Maybe some rate limiting could help. Suggestions welcome 🙂.

So what I was insinuating at is for the policy map pressure subsystem to do the aggregation internally, and only publish the top-k results. Maybe some users (as you called out "Users want to know generally if their policymaps are filling up, primarily") would set k to 1. Based on my reading of the metric types documentation, summaries or histograms could provide the aggregated information. It's not clear to me whether endpoint labels could be captured in the aggregated metrics though. With histograms, you would configure buckets like 0.8/0.9/1 for map pressure ratio.

Is that a reasonable middle ground?

tommyp1ckles:
I've been thinking about this a bit more recently.  I think the problem we have is there is a bit of tension between providing metrics that are of reasonable cardinality that can be enabled by default, as well as easy to interpret by users.

One thing I worry about is excessive use of creating special cases such as this, which may ultimately complicate our metrics due to lack of consistency.   As well we're reducing the functionality of these metrics to serve a particular case.

### Re: Cardinality
As @aditighag mentioned, I think there's good use cases for having per policy map pressure.  But per endpoint could be kind of excessive in many use cases.  This seems like using [Prometheus recording rules](url) to aggregate down these metrics (i.e. reduce down map pressures, etc) might be preferable.

### Re: Usability
In my opinion, the map pressure metric is already kind of confusing (it's a percentage, but only above a certain threshold - which is defined separately based on different map types where you have to read source code to know what the threshold is). I worry this might make this metric more confusing.

In this case, if we want a one-stop-shop type metric for this purpose perhaps the better approach would be to provide better examples of queries and recording rule configuration.  Or adding a new metric specific to policy map that can be enabled by default.

pkg/endpointmanager/policymap_pressure.go
joestringer: I feel like we could probably drop the error condition from `NewTrigger()` and just force the caller to pass a function directly in, to ensure it isn't nil. It doesn't really make sense for a caller to initialize a trigger without a function to trigger, since that defeats the whole point of the trigger in the first place. That could help to avoid the need for this sort of awkward error checking + panic logic in arbitrary packages like this.

(I buy your argument in this specific scenario, though I am a bit paranoid in general about `Panic()`, specifically because of the potential for unexpected agent failure)

joestringer: ^^ Doesn't need to be addressed in this PR, just a suggestion for future.

joestringer:
>  I think there's good use cases for having per policy map pressure. But per endpoint could be kind of excessive in many use cases.

These are the same right now yeah? There's one policymap per endpoint. So in k8s scale targets, up to ~100 different datapoints for the metric.

joestringer:
Removing ready-to-merge label while discussion is ongoing.

tommyp1ckles:
> These are the same right now yeah? There's one policymap per endpoint. 

Yeah, I was meaning that we may regret not having that functionality if needed.

> So in k8s scale targets, up to ~100 different datapoints for the metric.

Correct, at least it it wouldn't be per-endpoint * per-node, it would just be proportional to endpoints.

christarazi:
This PR was [discussed](https://docs.google.com/document/d/1Y_4chDk4rznD6UgXPlPvn3Dc7l-ZutGajUv1eF0VDwQ/edit#heading=h.srdx3nmtp0t) during the community meeting today and follow on discussions are taking place on Slack [here](https://cilium.slack.com/archives/C2B917YHE/p1692804722563059).

I will provide a summary of the discussion when it concludes.

christarazi:
Summary of above Slack [discussion](https://cilium.slack.com/archives/C2B917YHE/p1692804722563059) came down to two points:

### Summary

1. There were various proposals about what to do with `cilium_bpf_map_pressure`. Whether it should have high-cardinality labels or rather contain potentially aggregated values which makes it an "indicator". Cardinality could be controlled by an opt-in flag. 
2. If aggregation is chosen, is it the right level to do it at the Cilium level or rather have Prometheus perform aggregation via `PrometheusRule`

For (1), the notion that Cilium should have both "telemetry" and "indicator" type metrics is the right way forward. A "telemetry" method is a very basic metric that essentially represents a raw value of something. No processing has been done on it. Whereas an "indicator" metric is something that has been processed in order to turn it into a signal for something specific. For example, the `cilium_bpf_map_pressure` today is an "indicator" metric because it is a percentage. Ideally, we export the current size and capacity of BPF maps and compute the percentage at the Prometheus level, so we don't lose information by pre-processing. This idea in general makes sense.

However, in the case of policymaps specifically, it doesn't quite make sense to export each endpoint's policymap pressure on a clusterwide level because debugging why the pressure is high can only be done per node, rather than per endpoint. In other words, it is sufficient to only know which node is experiencing policymap pressure. It will become clearer why below. Hence, we don't really need "telemetry" metrics for policymaps or at least, there's no use case that are driving that sort of change right now. This PR aims to make it easier to observe policymap pressure and eventually act on it.

For (2), shipping an aggregation rule via `PrometheusRule`  is similar to alerts, but requires the user to be using prometheus-operator. We don't have any precedence for shipping `PrometheusRule` though in the repo, so we'd need to bootstrap that process. And if a user doesn't use prometheus-operator, there would need to be a separate process to integrate into that system. This sounds like a lot of work to do for a metric that we learned has no value in its current form (high cardinality), and really should just be aggregated and then exported.

### Debugging policymap pressure outlined steps

1. Find the node responsible for the endpoint and exec in.
2. Fetch the identity of the endpoint (needed?)
3. Use cilium policy selectors to get an overview of which selectors are selecting many identities
4. Depending on the type of selector, it tells us what sort of policy could be responsible:
  * CIDR
    * Change policy to use a wider range instead of smaller range?
  * FQDN
    * Look into cilium fqdn cache list. Do I expect to see all of these entries? Could we reduce the TTL, what relevant ToFQDNs options are there to look into, etc?
      * Zombies and CT GC interval?
   * Label
     * Is Cilium configured properly regarding [security-relevant labels](https://docs.cilium.io/en/stable/gettingstarted/terminology/#id4)? Could there be a dynamic label which is triggering many unique identities for what should be "identical" endpoints?
     * Investigate the policy and determine if the selection is too wide (hand wavy).

5. If the previous all checks out, maybe consider bumping the policymap size.

### Conclusion

At this point, I'm going with the current approach in the PR and following up with an expanded version of the troubleshooting steps

aditighag (approved):
Thanks for the summary. It'll be great to see the troubleshooting examples as follow-ups.

christarazi:
@aditighag Here's a PR for the troubleshooting guide: https://github.com/cilium/cilium/pull/27903

PR #27369, opened 2023-08-09
bpf: policy: cleanups to reduce program size

julianwiedmann:
This brings a bit of innocent fine-tuning for the BPF policy code.

For the example of `bpf_host`'s `tail_rev_nodeport_lb6()` (containing HostFW policy code), it shaves off around 100 instructions.

### DIFF
--- a/bpf/lib/policy.h
+++ b/bpf/lib/policy.h
@@ -33,8 +33,9 @@ __account_and_check(struct __ctx_buff *ctx, struct policy_entry *policy,
 
 static __always_inline int
 __policy_can_access(const void *map, struct __ctx_buff *ctx, __u32 local_id,
-		    __u32 remote_id, __u16 dport, __u8 proto, int off __maybe_unused,
-		    int dir, bool is_untracked_fragment, __u8 *match_type, __s8 *ext_err,
+		    __u32 remote_id, __u16 ethertype __maybe_unused, __u16 dport,
+		    __u8 proto, int off __maybe_unused, int dir,
+		    bool is_untracked_fragment, __u8 *match_type, __s8 *ext_err,
 		    __u16 *proxy_port)
 {
 	struct policy_entry *policy;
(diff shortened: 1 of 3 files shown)


gentoo-root (approved):
I'm not sure I fully understand the usefulness of the last patch.. Otherwise looks good!

Is the compiler not smart enough to consolidate these calls itself? What is the benefit of reducing the text size if we are only going through one of the branches?

julianwiedmann:
> I'm not sure I fully understand the usefulness of the last patch.. Otherwise looks good!
> Is the compiler not smart enough to consolidate these calls itself?

Nope - I was a bit surprised too :).

The [baseline](https://github.com/cilium/cilium/actions/runs/5852011226/job/15863639455#step:4:196) is 
`tail_rev_nodeport_lb4: processed 32822 insns (limit 131072), stack depth 216 (2029 unverified insns)`

With the [patch](https://github.com/cilium/cilium/pull/27481) applied this [becomes](https://github.com/cilium/cilium/actions/runs/5853059491/job/15866349797#step:4:182)
`tail_rev_nodeport_lb4: processed 30557 insns (limit 131072), stack depth 216 (1973 unverified insns)`

That's not much of course. But I'll take wins where I can get them, until the 4.19 kernel is gone.

bpf/lib/policy.h
```diff
@@ -48,49 +49,56 @@ __policy_can_access(const void *map, struct __ctx_buff *ctx, __u32 local_id,
...
+# endif
+
+# if defined(ENABLE_ICMP_RULE)
+			/* Convert from unsigned char to unsigned short
+			 * considering byte order(little-endian).
+			 * In the little-endian case, for example, 2byte data "AB"
+			 * convert to "BA".
+			 * Therefore, the "icmp_type" should be shifted not just casting.
+			 */
+			key.dport = (__u16)(icmphdr.type << 8);
```
joestringer, line 76: How can this be correct on all architectures without using some form of `ntohs()`?

joestringer: Hmm, looks like this was originally introduced by @chez-shanpu in https://github.com/cilium/cilium/pull/16516 .

chez-shanpu: Yes, probably this line should use `bpf_htons()` instead of using shift.
I'll fix it and submit a PR.

PR #27370, opened 2023-08-09
Replace metricsmap-bpf-prom-sync with Prometheus Collector pattern

### LINKED ISSUE #27058: Replace Metricsmap-bpf-prom-sync with Prometheus Collector pattern
I was reviewing some work that @christarazi and @asauber were putting together reviewing our controllers/metrics and noticed that we seem to have a controller whose primarily responsibility is to periodically read values from a BPF map and update Prometheus metrics values based on the values in the map.

This approach has a few flaws, in particular:
- Metrics values are always slightly off. Prometheus scrapes periodically already, having another periodic timer means metrics can be weirdly skewed because of these two timers.
- Metrics values might stop getting updates if this controller stops running or doesn't startup successfully

From my previous experience and knowledge of other Go projects that need to deal with a similar problem, I believe there is a better alternative. The alternative is to implement the Prometheus Collector interface. https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#hdr-Custom_Collectors_and_constant_Metrics documents the use-case quite well, but in summary, the idea is that instead of managing counters, whenever Prometheus scrapes your application, it will call Collect() on your collector, and you then get the current value of all your metrics and send them on the channel given to you by Collect(). Basically, this eliminates the extra periodic metrics sync loop happening inside the controller, instead getting metrics values at scrape time. This is an approach taken by Prometheus node-exporter and cadvisor which read this information from /proc and /sys at scrape time.

This eliminates state, avoids metrics value accuracy, and overall is a simpler design that follows existing best practices for metrics collection.

carnerito:
Previously, the 'metricsmap-bpf-prom-sync' controller was responsible for periodically collecting cilium_datapath drop
and forward metrics.
This PR introduces the metricsmapCollector, which implements the Prometheus Collector interface within the
metricsmap package. As a result, the aforementioned controller has been removed.
The metricsmapCollector is registered within the global Prometheus registry during the initialization of metricsmap.
The metrics.Register function has been modified to propagate errors from the registry.Register function instead of
simply overriding it with nil.
The metricsmapCollector comprises two metrics maps: forwardedMetricsMap and droppedMetricsMap.
These maps are populated within a callback function passed to the IterateWithCallback function.
This approach serves two primary purposes:
1. Separation of Map Iteration and Metric Update: By separating the iteration over the BPF map and the updating of
    Prometheus metrics, the implementation ensures that no partial metrics are exposed in case of map iteration failure.
2. Normalization of exposed Metrics: Unlike the statement in the bpf/lib/metrics.h comments, which suggests exposing
    only one reason label for forwarded metrics, the eBPF code exposes multiple reasons. Through testing, it was found
    that reasons 0 and 3 were exposed which triggered this error in prometheus client:
    https://github.com/prometheus/client_golang/issues/242

Furthermore, cilium command has an indirect dependency on metricsmap package. To prevent metrics map initialization
from happening each time cilium command is executed, metricsmap is converted to hive Cell and injected in cilium daemon
Infrastructure module. For details see this comment: https://github.com/cilium/cilium/pull/27370#issuecomment-1710805663.

Fixes: #27058

### DIFF
--- a/pkg/maps/metricsmap/metricsmap.go
+++ b/pkg/maps/metricsmap/metricsmap.go
@@ -4,16 +4,22 @@
 package metricsmap
 
 import (
-	"context"
 	"unsafe"
 
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/cilium/cilium/pkg/ebpf"
+	"github.com/cilium/cilium/pkg/hive/cell"
 	"github.com/cilium/cilium/pkg/logging"
 	"github.com/cilium/cilium/pkg/logging/logfields"
 	"github.com/cilium/cilium/pkg/metrics"
 	monitorAPI "github.com/cilium/cilium/pkg/monitor/api"
+)
+
+var Cell = cell.Module(
+	"metricsmap",
+	"eBPF Metrics Map",
+	cell.Invoke(RegisterCollector),
 )
 
 // IterateCallback represents the signature of the callback function expected by
@@ -33,8 +39,15 @@ type metricsMap struct {
 
 var (
 	// Metrics is the bpf metrics map
-	Metrics metricsMap
-	log     = logging.DefaultLogger.WithField(logfields.LogSubsys, "map-metrics")
+	Metrics = metricsMap{ebpf.NewMap(&ebpf.MapSpec{
+		Name:       MapName,
+		Type:       ebpf.PerCPUHash,
+		KeySize:    uint32(unsafe.Sizeof(Key{})),
+		ValueSize:  uint32(unsafe.Sizeof(Value{})),
+		MaxEntries: MaxEntries,
+		Pinning:    ebpf.PinByName,
+	})}
+	log = logging.DefaultLogger.WithField(logfields.LogSubsys, "map-metrics")
 )
 
 const (
@@ -89,7 +102,6 @@ func (m metricsMap) IterateWithCallback(cb IterateCallback) error {
 	return m.Map.IterateWithCallback(&Key{}, &Values{}, func(k, v interface{}) {
 		key := k.(*Key)
 		values := v.(*Values)
-
 		cb(key, values)
 	})
 }
@@ -137,54 +149,132 @@ func (vs Values) Bytes() uint64 {
 	return b
 }
 
-func updateMetric(getCounter func() (prometheus.Counter, error), newValue float64) {
-	counter, err := getCounter()
+// metricsMapCollector implements Prometheus Collector interface
+type metricsmapCollector struct {
+	droppedCountDesc *prometheus.Desc
+	droppedByteDesc  *prometheus.Desc
+	forwardCountDesc *prometheus.Desc
+	forwardByteDesc  *prometheus.Desc
+
+	// eBPF code seems to expose multiple reasons for forwarded metrics
+	// as opposed to what is stated in bpf/lib/metrics.h comments.
+	// IterateWithCallback iterates through BPF map for each reason and direction.
+	// Since we do not have "reason" label on forwarded metrics, we would end up collecting
+	// same forwarded metric multiple times which is not allowed by prometheus client.
+	// See https://github.com/prometheus/client_golang/issues/242
+	//
+	// promMetrics is a generic map used to sum all values by desired set of labels
+	// for both forwarded and dropped metrics
+	forwardedMetricsMap promMetrics[forwardLabels]
+	droppedMetricsMap   promMetrics[dropLabels]
+}
+
+func newMetricsMapCollector() prometheus.Collector {
+	return &metricsmapCollector{
+		droppedMetricsMap:   make(map[dropLabels]metricValues),
+		forwardedMetricsMap: make(map[forwardLabels]metricValues),
+		droppedByteDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(metrics.Namespace, metrics.SubsystemDatapath, "drop_bytes_total"),
+			"Total dropped bytes, tagged by drop reason and ingress/egress direction",
+			[]string{metrics.LabelDropReason, metrics.LabelDirection}, nil,
+		),
+		droppedCountDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(metrics.Namespace, metrics.SubsystemDatapath, "drop_count_total"),
+			"Total dropped packets, tagged by drop reason and ingress/egress direction",
+			[]string{metrics.LabelDropReason, metrics.LabelDirection}, nil,
+		),
+		forwardCountDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(metrics.Namespace, metrics.SubsystemDatapath, "forward_count_total"),
+			"Total forwarded packets, tagged by ingress/egress direction",
+			[]string{metrics.LabelDirection}, nil,
+		),
+		forwardByteDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(metrics.Namespace, metrics.SubsystemDatapath, "forward_bytes_total"),
+			"Total forwarded bytes, tagged by ingress/egress direction",
+			[]string{metrics.LabelDirection}, nil,
+		),
+	}
+}
+
+type forwardLabels struct {
+	direction string
+}
+
+type dropLabels struct {
+	direction string
+	reason    string
+}
+
+type metricValues struct {
+	bytes float64
+	count float64
+}
+
+type labels comparable
+
+type promMetrics[k labels] map[k]metricValues
+
+func (p promMetrics[k]) upsert(labels k, values *Values) {
+	if v, ok := p[labels]; ok {
+		v.bytes = float64(values.Bytes())
+		v.count = float64(values.Count())
+	}
+	p[labels] = metricValues{
+		bytes: float64(values.Bytes()),
+		count: float64(values.Count()),
+	}
+}
+
+func (mc *metricsmapCollector) Collect(ch chan<- prometheus.Metric) {
+	err := Metrics.IterateWithCallback(func(key *Key, values *Values) {
+		if key.IsDrop() {
+			labelSet := dropLabels{
+				direction: key.Direction(),
+				reason:    key.DropForwardReason(),
+			}
+			mc.droppedMetricsMap.upsert(labelSet, values)
+		} else {
+			labelSet := forwardLabels{
+				direction: key.Direction(),
+			}
+			mc.forwardedMetricsMap.upsert(labelSet, values)
+		}
+	})
 	if err != nil {
-		log.WithError(err).Warn("Failed to update prometheus metrics")
+		log.WithError(err).Warn("Failed to read metrics from BPF map")
+		// Do not update partial metrics
 		return
 	}
 
-	oldValue := metrics.GetCounterValue(counter)
-	if newValue > oldValue {
-		counter.Add(newValue - oldValue)
+	for labels, value := range mc.forwardedMetricsMap {
+		mc.updateCounterMetric(mc.forwardCountDesc, ch, value.count, labels.direction)
+		mc.updateCounterMetric(mc.forwardByteDesc, ch, value.bytes, labels.direction)
+	}
+
+	for labels, value := range mc.droppedMetricsMap {
+		mc.updateCounterMetric(mc.droppedCountDesc, ch, value.count, labels.reason, labels.direction)
+		mc.updateCounterMetric(mc.droppedByteDesc, ch, value.bytes, labels.reason, labels.direction)
 	}
 }
 
-// updatePrometheusMetrics checks the metricsmap key value pair
-// and determines which prometheus metrics along with respective labels
-// need to be updated.
-func updatePrometheusMetrics(key *Key, values *Values) {
-	updateMetric(func() (prometheus.Counter, error) {
-		if key.IsDrop() {
-			return metrics.DropCount.GetMetricWithLabelValues(key.DropForwardReason(), key.Direction())
-		}
-		return metrics.ForwardCount.GetMetricWithLabelValues(key.Direction())
-	}, float64(values.Count()))
-
-	updateMetric(func() (prometheus.Counter, error) {
-		if key.IsDrop() {
-			return metrics.DropBytes.GetMetricWithLabelValues(key.DropForwardReason(), key.Direction())
-		}
-		return metrics.ForwardBytes.GetMetricWithLabelValues(key.Direction())
-	}, float64(values.Bytes()))
+func (mc *metricsmapCollector) updateCounterMetric(desc *prometheus.Desc, metricsChan chan<- prometheus.Metric, value float64, labelValues ...string) {
+	metricsChan <- prometheus.MustNewConstMetric(
+		desc,
+		prometheus.CounterValue,
+		value,
+		labelValues...)
 }
 
-// SyncMetricsMap is called periodically to sync off the metrics map by
-// aggregating it into drops (by drop reason and direction) and
-// forwards (by direction) with the prometheus server.
-func SyncMetricsMap(ctx context.Context) error {
-	return Metrics.IterateWithCallback(func(key *Key, values *Values) {
-		updatePrometheusMetrics(key, values)
-	})
+func (mc *metricsmapCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- mc.forwardByteDesc
+	ch <- mc.forwardCountDesc
+	ch <- mc.droppedCountDesc
+	ch <- mc.droppedByteDesc
 }
 
-func init() {
-	Metrics.Map = ebpf.NewMap(&ebpf.MapSpec{
-		Name:       MapName,
-		Type:       ebpf.PerCPUHash,
-		KeySize:    uint32(unsafe.Sizeof(Key{})),
-		ValueSize:  uint32(unsafe.Sizeof(Value{})),
-		MaxEntries: MaxEntries,
-		Pinning:    ebpf.PinByName,
-	})
+func RegisterCollector() {
+	if err := metrics.Register(newMetricsMapCollector()); err != nil {
+		log.WithError(err).Error("Failed to register metrics map collector to Prometheus registry. " +
+			"cilium_datapath_drop/forward metrics will not be collected")
+	}
 }
--- a/daemon/cmd/cells.go
+++ b/daemon/cmd/cells.go
@@ -34,6 +34,7 @@ import (
 	"github.com/cilium/cilium/pkg/kvstore/store"
 	"github.com/cilium/cilium/pkg/l2announcer"
 	"github.com/cilium/cilium/pkg/logging/logfields"
+	"github.com/cilium/cilium/pkg/maps/metricsmap"
 	"github.com/cilium/cilium/pkg/metrics"
 	"github.com/cilium/cilium/pkg/node"
 	nodeManager "github.com/cilium/cilium/pkg/node/manager"
@@ -79,6 +80,9 @@ var (
 
 		// Provide the modular metrics registry, metric HTTP server and legacy metrics cell.
 		metrics.Cell,
+
+		// Provides cilium_datapath_drop/forward Prometheus metrics.
+		metricsmap.Cell,
 
 		// Provide option.Config via hive so cells can depend on the agent config.
 		cell.Provide(func() *option.DaemonConfig { return option.Config }),
--- a/daemon/cmd/datapath.go
+++ b/daemon/cmd/datapath.go
@@ -14,7 +14,6 @@ import (
 	"github.com/vishvananda/netlink"
 
 	"github.com/cilium/cilium/pkg/cidr"
-	"github.com/cilium/cilium/pkg/controller"
 	datapathIpcache "github.com/cilium/cilium/pkg/datapath/ipcache"
 	"github.com/cilium/cilium/pkg/datapath/linux/ipsec"
 	"github.com/cilium/cilium/pkg/datapath/linux/linux_defaults"
@@ -47,10 +46,7 @@ import (
 	"github.com/cilium/cilium/pkg/node"
 	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/source"
-	"github.com/cilium/cilium/pkg/time"
 )
-
-var metricsmapBPFPromSyncControllerGroup = controller.NewGroup("metricsmap-bpf-prom-sync")
 
 // LocalConfig returns the local configuration of the daemon's nodediscovery.
 func (d *Daemon) LocalConfig() *datapath.LocalNodeConfiguration {
@@ -450,17 +446,6 @@ func (d *Daemon) initMaps() error {
 			}
 		}
 	}
-
-	// Start the controller for periodic sync of the metrics map with
-	// the prometheus server.
-	controller.NewManager().UpdateController(
-		"metricsmap-bpf-prom-sync",
-		controller.ControllerParams{
-			Group:       metricsmapBPFPromSyncControllerGroup,
-			DoFunc:      metricsmap.SyncMetricsMap,
-			RunInterval: 5 * time.Second,
-			Context:     d.ctx,
-		})
 
 	if !option.Config.RestoreState {
 		// If we are not restoring state, all endpoints can be
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -102,6 +102,9 @@ const (
 	// LabelValueOutcomeFail is used as an unsuccessful outcome of an operation
 	LabelValueOutcomeFail = "fail"
 
+	// LabelDropReason is used to describe reason for dropping a packets/bytes
+	LabelDropReason = "reason"
+
 	// LabelEventSourceAPI marks event-related metrics that come from the API
 	LabelEventSourceAPI = "api"
 
@@ -357,22 +360,6 @@ var (
 	ProxyDatapathUpdateTimeout = NoOpCounter
 
 	// L3-L4 statistics
-
-	// DropCount is the total drop requests,
-	// tagged by drop reason and direction(ingress/egress)
-	DropCount = NoOpCounterVec
-
-	// DropBytes is the total dropped bytes,
-	// tagged by drop reason and direction(ingress/egress)
-	DropBytes = NoOpCounterVec
-
-	// ForwardCount is the total forwarded packets,
-	// tagged by ingress/egress direction
-	ForwardCount = NoOpCounterVec
-
-	// ForwardBytes is the total forwarded bytes,
-	// tagged by ingress/egress direction
-	ForwardBytes = NoOpCounterVec
 
 	// Datapath statistics
 
@@ -658,10 +645,6 @@ type LegacyMetrics struct {
 	ProxyPolicyL7Total               metric.Vec[metric.Counter]
 	ProxyUpstreamTime                metric.Vec[metric.Observer]
 	ProxyDatapathUpdateTimeout       metric.Counter
-	DropCount                        metric.Vec[metric.Counter]
-	DropBytes                        metric.Vec[metric.Counter]
-	ForwardCount                     metric.Vec[metric.Counter]
-	ForwardBytes                     metric.Vec[metric.Counter]
 	ConntrackGCRuns                  metric.Vec[metric.Counter]
 	ConntrackGCKeyFallbacks          metric.Vec[metric.Counter]
 	ConntrackGCSize                  metric.Vec[metric.Gauge]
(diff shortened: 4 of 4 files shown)


dylandreimerink (approved):

dylandreimerink:
Hi @carnerito, thank you for this contribution. Could you sign off on your commit? https://docs.cilium.io/en/stable/contributing/development/contributing_guide/#developer-s-certificate-of-origin

carnerito:
@dylandreimerink Can I squash all commits and force-push?

Documentation/bpf/toolchain.rst
```diff
@@ -144,7 +144,8 @@ In order to run through all BPF selftests, the following command is needed:
 
     $ sudo make run_tests
 
-If you see any failures, please contact us on Slack with the full test output.
+If you see any failures, please contact us on `Cilium Slack`_ with the full
+test output.
```
zacharysarah: Who is "us"? I know this is inherited language, but it's an opportunity to be specific and helpful rather than remaining vague.

Documentation/community/governance/commit_access.rst
```diff
@@ -130,11 +130,13 @@ The Maintainers may also choose to delegate working with the CNCF to non-Maintai
 Code of Conduct
 ~~~~~~~~~~~~~~~
 
-`Code of Conduct <https://github.com/cilium/cilium/blob/main/CODE_OF_CONDUCT.md>`__
-violations by community members will be discussed and resolved on the private
-committers Slack channel. If the reported Code of Conduct violator is a Maintainer, the 
-Maintainers will instead designate two Maintainers to work with the
-`CNCF CoC Committee <https://www.cncf.io/conduct/procedures/>`__.  
+`Code of Conduct`_ violations by community members will be discussed and
```
zacharysarah: Use present tense
```suggestion
`Code of Conduct`_ violations by community members are discussed and
```

Documentation/community/governance/commit_access.rst
```diff
@@ -349,8 +351,8 @@ decide an issue. If the maintainers themselves cannot decide an issue, the
 issue will be resolved by voting. The voting process is a simple majority in
 which each committer and each maintainer receives one vote.
 
-Votes are done in the Slack channel #committers using Slack polls. A failure to
-vote is an implicit abstention.
+Votes are done in the Slack channel ``#committers`` using Slack polls. A
+failure to vote is an implicit abstention.
```
zacharysarah: I think we can be gentler.
```suggestion
Votes are done in the Slack channel ``#committers`` using Slack polls.
Declining to vote is an implicit abstention.
```

Documentation/community/roadmap.rst
```diff
@@ -171,15 +171,18 @@ production.
...
-
-Contributions in this area are very welcome. To get started, take a look at the :ref:`guide-to-the-hive`
-documentation and the issues referenced from `modularization meta issue <modularization-issue_>`_.
-If you have any questions or ideas please join us on the #sig-modularization channel on `Cilium Slack <slack_>`_.
+As the project is growing in complexity it is becoming increasingly important
+to be able to divide it into more manageable chunks. To achieve this, we're
+working on modularizing the codebase and going from a tightly coupled design
+(one large initialization and configuration) to a more loosely coupled design
+of mostly self-contained modules. This will make Cilium internals easier to
+comprehend, test and extend.
```
zacharysarah: Avoid "we", use plain, clear language, specific is better than general, future tense is OK here
```suggestion
As the project grows in complexity, it is increasingly important
to be able to divide the project into more manageable chunks. To achieve this, contributors
are modularizing the codebase and going from a tightly coupled design
(one large initialization and configuration) to a more loosely coupled design
of mostly self-contained modules. This will make Cilium internals easier to
comprehend, test, and extend.
```

Documentation/community/roadmap.rst
```diff
@@ -171,15 +171,18 @@ production.
...
+
+Contributions in this area are very welcome. To get started, take a look at the
+:ref:`guide-to-the-hive` documentation and the issues referenced from
+`modularization meta issue <modularization-issue_>`_. If you have any questions
+or ideas please join us on the ``#sig-modularization`` channel on `Cilium
```
zacharysarah: Use plain, clear language (avoid "please")
```suggestion
or ideas, join us on the ``#sig-modularization`` channel on `Cilium
```

Documentation/configuration/argocd-issues.rst
```diff
@@ -83,13 +83,19 @@ Solution
...
-    Note that when using ``helm template``, it is highly recommended you set ``--kube-version`` and ``--api-versions`` with the values matching your target Kubernetes cluster.
-    Helm charts such as Cilium's often conditionally enable certain Kubernetes features based on their availability (beta vs stable) on the target cluster.
-    
-    By specifying ``--api-versions=monitoring.coreos.com/v1`` you should be able to pass validation with ``helm template``.
-
-    
-If you have an issue with Argo CD that's not outlined above, check this `list of Argo CD related issues on GitHub <https://github.com/cilium/cilium/issues?q=is%3Aissue+argocd>`__.
-If you can't find an issue that relates to yours, create one and/or seek help on the :term:`Slack channel`.
+
+    When using ``helm template``, it is highly recommended you set
```
zacharysarah: Use clear, plain language
```suggestion
    When using ``helm template``, set
```

Documentation/configuration/argocd-issues.rst
```diff
@@ -83,13 +83,19 @@ Solution
...
+    ``--kube-version`` and ``--api-versions`` with the values matching your
+    target Kubernetes cluster. Helm charts such as Cilium's often conditionally
+    enable certain Kubernetes features based on their availability (beta vs
+    stable) on the target cluster.
+
+    By specifying ``--api-versions=monitoring.coreos.com/v1`` you should be
+    able to pass validation with ``helm template``.
+
+If you have an issue with Argo CD that's not outlined above, check this `list
```
zacharysarah: accessibility nit
```suggestion
If you have an issue with Argo CD that's not outlined above, review this `list
```

Documentation/glossary.rst
```diff
@@ -12,8 +12,8 @@ Glossary
 Cilium has some terms with special meanings. These should all be covered
 throughout the documentation but for convenience we have also listed some of
 them below with short descriptions. If you need more information, please ask us
-on `Slack <https://cilium.herokuapp.com>`_. Feel free to extend this document
-with words you expected to see here.
+on `Cilium Slack`_. Feel free to extend this document with words you expected
```
zacharysarah: wording nit
```suggestion
on `Cilium Slack`_. Feel free to extend this document with terms you expected
```

Documentation/installation/cni-chaining-generic-veth.rst
```diff
@@ -34,7 +34,8 @@ Validate that the current CNI plugin is using veth
 If the CNI plugin you are chaining with is currently not using veth then the
 ``generic-veth`` plugin is not suitable. In that case, a full CNI chaining
 plugin is required which understands the device model of the underlying plugin.
-Writing such a plugin is trivial, contact us on :ref:`slack` for more details.
+Writing such a plugin is trivial, contact us on `Cilium Slack`_ for more
```
zacharysarah: Who is "us"? I recognize this is inherited language, but it's a chance to be specific and helpful rather than continuing to be vague.

pkg/maps/metricsmap/metricsmap.go
```diff
@@ -137,47 +137,134 @@ func (vs Values) Bytes() uint64 {
...
+// metricsMapCollector implements Prometheus Collector interface
+type metricsmapCollector struct {
+	droppedCountDesc *prometheus.Desc
+	droppedByteDesc  *prometheus.Desc
+	forwardCountDesc *prometheus.Desc
+	forwardByteDesc  *prometheus.Desc
+
+	// eBPF code seems to expose multiple reasons for forwarded metrics
+	// as opposed to what is stated in bpf/lib/metrics.h comments.
+	// IterateWithCallback will iterate through BPF map for each reason and direction.
```
zacharysarah: Verb tense nit
```suggestion
	// IterateWithCallback iterates through BPF map for each reason and direction.
```

zacharysarah (changes requested):
@carnerito Thanks for the updates! Some minor stylistic changes are required, with strong encouragement to fix some of the inherited language touched by this PR.

pkg/maps/metricsmap/metricsmap.go
carnerito: @zacharysarah I'm really sorry for making confusion, but all changes you have proposed are not from my PR. I tried to amend the unsigned commit as suggested by @dylandreimerink and I've pushed someone else's commits to this branch by mistake.
The only thing that you have commented that is part of this PR is regarding the wording in metricsmap.go, and I have fixed that in [#d1988ae](https://github.com/cilium/cilium/pull/27370/commits/d1988ae903d5b7424af189c6bce16bf253c275d6)

Documentation/bpf/toolchain.rst
carnerito: Not part of this PR.

Documentation/community/governance/commit_access.rst
carnerito: Not part of this PR.

Documentation/community/governance/commit_access.rst
carnerito: Not part of this PR.

Documentation/community/roadmap.rst
carnerito: Not part of this PR.

Documentation/community/roadmap.rst
carnerito: Not part of this PR.

Documentation/configuration/argocd-issues.rst
carnerito: Not part of this PR.

Documentation/configuration/argocd-issues.rst
carnerito: Not part of this PR.

Documentation/glossary.rst
carnerito: Not part of this PR.

Documentation/installation/cni-chaining-generic-veth.rst
carnerito: Not part of this PR.

christarazi (commented):
Thanks for the PR! Could you rebase and squash all commits together?

carnerito:
@christarazi Done.

pkg/maps/metricsmap/metricsmap.go
zacharysarah: @carnerito
> all changes you have proposed are not from my PR

~~And yet your changes touch them, creating an opportunity to make docs better. Otherwise--when will docs ever improve?~~

~~I'm disappointed in your choice to refuse uncomplicated improvements, but I'll approve regardless.~~

UPDATE: See https://github.com/cilium/cilium/pull/27370#discussion_r1296458751 and following 👇🏻

zacharysarah (approved):
LGTM

pkg/maps/metricsmap/metricsmap.go
joestringer: It looks to me like there was a misunderstanding here in the way that the changes got pushed into GitHub and the way that GitHub requested reviewers for these changes. From what I can tell, @carnerito never intended to make changes to these files in the PR, but the way that the PR was pushed into GitHub caused GitHub to pull @zacharysarah in for review.

In future it might be worth doing a quick double-check of the PR before marking it "ready for review" to make sure that GitHub presents the set of changes that you expected to present, so that GitHub only pulls in the codeowners for the areas that are being changed.

At this point it looks like @carnerito has been able to resolve this issue, maybe with a fresh push of the branch, and none of these comments are relevant to the changes in the PR any more. Either way it looks like this has been resolved in the end. @carnerito apologies for the confusion around this, thanks again for contributing these changes to Cilium :-)

zacharysarah: @joestringer Thanks for the explanation. @carnerito Apologies for pushing back so strongly; I see the underlying issue now. Thanks for your contributions!

pkg/maps/metricsmap/metricsmap.go
```diff
@@ -137,47 +137,134 @@ func (vs Values) Bytes() uint64 {
...
+		forwardByteDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(metrics.Namespace, metrics.SubsystemDatapath, "forward_bytes_total"),
+			"Total forwarded bytes, tagged by ingress/egress direction",
+			[]string{metrics.LabelDirection}, nil,
+		),
+	}
+}
+
+func (mc *metricsmapCollector) Collect(ch chan<- prometheus.Metric) {
+
```
christarazi, line 202: Nit: unnecessary newline

christarazi (dismissed):
Very cool PR! I was wondering whether we could make use of batch lookups instead of using `IterateWithCallback`, but realized that the metrics map doesn't have very many entries in it, so maybe it's just a nice-to-have thing. With the PR there's already a performance improvement because the controller used to run every 5s and now it's based on the scrape interval which is very likely to be >=15s.

joestringer:
A couple of the CI jobs are failing due to timeouts ([example1](https://github.com/cilium/cilium/actions/runs/5884199682), [example2](https://github.com/cilium/cilium/actions/runs/5884200085) and there are many failures in the [main ginkgo test run](https://github.com/cilium/cilium/actions/runs/5884199995/job/15958487587). It seems like there's some kind of issue causing problems in the CI, but it's not immediately obvious what that issue is.

pkg/maps/metricsmap/metricsmap.go
carnerito: @christarazi Thanks for the review. I didn't know that maps have a BatchLookup method; I'll keep this in mind in the future. As for the newline, I was relying on gomt to do the formatting, but I completely agree that it would be better without it here.

carnerito:
@joestringer This is something that happens from time to time, or do you think it is related to this PR?
Can we try re-running tests?

joestringer:
I haven't seen this before, but I can re-run the CI tests to confirm.

carnerito:
@joestringer Due to lack of experience with GitHub Actions and Cilium CI in general I have no idea how to proceed with this, do you have any suggestions?

It's really strange to me that same tests were passing before I pulled changes from main branch...

carnerito:
@joestringer I've pulled the latest changes from the main branch and tests are not timing out anymore.

pkg/maps/metricsmap/metricsmap.go
```diff
@@ -187,4 +274,14 @@ func init() {
 		MaxEntries: MaxEntries,
 		Pinning:    ebpf.PinByName,
 	})
+
+	// Register metrics map collector only once
+	once.Do(func() {
+		c := newMetricsMapCollector()
```
derailed: nit: no need for this tmp var. Call directly? `metrics.Register(newMetricsCollector())`

pkg/maps/metricsmap/metricsmap.go
```diff
@@ -137,47 +137,134 @@ func (vs Values) Bytes() uint64 {
...
-			return metrics.DropCount.GetMetricWithLabelValues(key.DropForwardReason(), key.Direction())
-		}
-		return metrics.ForwardCount.GetMetricWithLabelValues(key.Direction())
-	}, float64(values.Count()))
+// Helper function that returns metricValues if key is present in map.
+// Otherwise, metricValues for key will be created and returned.
+func lookupOrCreateMetricsMapValues[K comparable](m map[K]metricValues, key K) metricValues {
+	if entry, ok := m[key]; ok {
+		return entry
+	} else {
```
derailed: nit: superfluous else

derailed (dismissed):
@carnerito Nice work Boris! Just a couple small picks...

joestringer:
It's interesting to see that the same ginkgo failure is occurring for another completely different PR. I'm unsure about the cause. https://github.com/cilium/cilium/issues/27688

pkg/maps/metricsmap/metricsmap.go
carnerito: @derailed Thanks for review. I'll make changes after we resolve the issue with CI.

carnerito:
> It's interesting to see that the same ginkgo failure is occurring for another completely different PR. I'm unsure about the cause. #27688

Indeed.

P.S. 
I'm AFK for the next 10 days so I won't be able to actively work on this until then. If you require any modifications to be made in this PR for testing a fix in CI, feel free to reach me out here or on Slack.

joestringer:
We've identified https://github.com/cilium/cilium/issues/27688 as being triggered by other PRs also (and not reliably passing/failing), which suggests that the failure is unrelated to this PR. I think that the next step should be to rebase this PR against tip of main branch and only have the functional changes in the PR, then retrigger CI again. We'll need to pay close attention that CI doesn't point out any other issues with the actual PR, but if CI can pass with these changes then it should be fine to move forward.

joestringer:
@carnerito I'll mark this PR as draft for now. When you get back, could you try to rebase afresh with the actual proposed changes for this PR, then click `Ready for review` at the bottom of the page? We can try afresh with the changes in the PR.

carnerito:
> @carnerito I'll mark this PR as draft for now. When you get back, could you try to rebase afresh with the actual proposed changes for this PR, then click `Ready for review` at the bottom of the page? We can try afresh to test the intended changes in the PR.

I tried to revert to a commit with actual changes, but it didn't go well. As you can see here, this PR reverts some changes made after it was created. Pulling or rebasing onto main shows that everything is up to date, although `git diff metricsmap-collector upstream/main` reveals differences. Since I don't know how to fix this I've opened [#27740](https://github.com/cilium/cilium/pull/27740) with actual changes.

carnerito:
@joestringer With a little help from @christarazi I've managed to fix the issues with this PR. It's ready for review.

christarazi:
@carnerito What was the problem with the code such that the CI tests were failing?

carnerito:
> @carnerito What was the problem with the code such that the CI tests were failing?

I didn't prove it is related to code from this PR. The same CI test failures are happening for other PRs too. @joestringer opened [#27688](https://github.com/cilium/cilium/issues/27688) to track this.

christarazi:
I think that was for a specific failure for when the revert was applied, but this doesn't explain why the rest of CI especially the `ci-ginkgo` suite was timing out on almost all tests when the PR contained the original change. CI timing out was not the case when the revert was applied, so it does indeed seem like the original changes introduced in this PR are somehow causing test failures.

carnerito:
I didn't understand [this comment](https://github.com/cilium/cilium/pull/27370#issuecomment-1693791258) than. Can we mark this PR as draft again until I investigate CI failures?

christarazi:
Marked as draft. There may have been some confusion, but the issue identified (https://github.com/cilium/cilium/issues/27688) has nothing to do with the current failures (e.g. here's the same test case failing for different reasons: https://github.com/cilium/cilium/actions/runs/6004247752/job/16284443998#step:16:630) and is actually the same set of failures I investigated last week. It's a completely different error msg.

carnerito:
@christarazi @joestringer I've figured out why Ginkgo tests were failing. It's a combination of multiple factors.

First:
- Ginkgo tests were failing because the kube-dns deployment was not configured on time -> [View Details](https://github.com/cilium/cilium/actions/runs/6004247752/job/16284443998#step:16:261)
- The kube-dns deployment was not configured on time since the [kube-dns service plumbing check](https://github.com/cilium/cilium/blob/5b0cf2cc3d6abcdf7b1a662691c9620237b31b51/test/helpers/kubectl.go#L2058) was hitting a 10s timeout, as seen in the logs.
- The plumbing check was timing out because the [Cilium commands that are run](https://github.com/cilium/cilium/blob/5b0cf2cc3d6abcdf7b1a662691c9620237b31b51/test/helpers/kubectl.go#L1862-L1884) to fetch Cilium state took more time than expected to finish.
- The delay mentioned in the previous bullet point is introduced by the metricsmap package initialization, which adds approximately 1s latency to any Cilium command (e.g., `cilium service list`) (I'm not sure why metricsmap is initialized since it is not invoked directly by any cilium command, so I guess it's indirect dependency).
- [#e2839ab](https://github.com/cilium/cilium/pull/27370/commits/e2839abbef9875419904e0ff1e535a9e0ee27993) fixes this issue.


Second:
I believe there is a bug in the CiliumExecContext function. [It repeats the same command 5 times without checking if the command has succeeded](https://github.com/cilium/cilium/blob/5b0cf2cc3d6abcdf7b1a662691c9620237b31b51/test/helpers/kubectl.go#L2911-L2920).
If we combine this behavior with the 1-second delay mentioned above and multiply that by the 2 kube-dns pods, we have a reason for exceeding the 10s limit when checking if kube-dns is validating service plumbing.

pkg/maps/metricsmap/metricsmap.go
```diff
@@ -187,4 +272,18 @@ func init() {
...
+	// it is executed in a separate goroutine to prevent introducing delays in cilium command execution
+	once.Do(func() {
+		go func() {
+			err := metrics.Register(newMetricsMapCollector())
+			if err != nil {
+				log.WithError(err).Error("Failed to register metrics map collector to Prometheus registry. " +
+					"cilium_datapath_drop/forward metrics will not be collected")
+			}
+		}()
+	})
```
christarazi: Hm, the fact that this initialization occurs when a Cilium command is run first seems undesirable. Ideally, the daemon initializes all the metrics it needs upfront. The Cilium commands (which are separate binaries from the daemon) should not have to run any code to initialize metrics. 

I think what we could do here is to export `newMetricsMapCollector()` and have to be called within the `pkg/metrics/cell.go` code which is imported by the Agent Cell in `daemon/cmd/cells.go`.

cc @dylandreimerink, does this sound reasonable?

christarazi (commented):
Thanks for the thorough investigation! 

Below is a comment on the metrics initialization code and regarding the bug in Cilium test helper functions, a separate PR would be good for that.

carnerito:
PR that fixes bug in Cilium test helper function [#28131](https://github.com/cilium/cilium/pull/28131).

pkg/maps/metricsmap/metricsmap.go
carnerito: @dylandreimerink @christarazi Can you take a look [at latest commit](https://github.com/cilium/cilium/pull/27370/commits/54b1945191db724ac094c79e9785cf9f1b017d39). I have converted metricsmap to cell.Module which is invoked in `daemon/cmd/cells.go`. Now, only daemon will initialize metricsmap. I've tried to add it to `pkg/metrics/cell.go` as suggested, but than we have circular dependency issue
```
package github.com/cilium/cilium/api/v1/server
        imports github.com/cilium/cilium/pkg/metrics
        imports github.com/cilium/cilium/pkg/maps/metricsmap
        imports github.com/cilium/cilium/pkg/ebpf
        imports github.com/cilium/cilium/pkg/bpf
        imports github.com/cilium/cilium/pkg/controller
        imports github.com/cilium/cilium/pkg/metrics: import cycle not allowed
```
Furthermore, I've tried to add metricsmap collector to metricsmap module, like in Hive examples, with `cell.Metrics(newMetricsmapCollector)`, but I couldn't make it work that way. Even though all metricsmap metrics were registered, metricsmap Collector was not, so Collect method wasn't invoked on the scrape. I know that these metrics were registered because I've spawn separate goroutine that was incrementing counters periodically and I was able to scrape them. 
Is it possible to create custom Collector using metric types defined in `pkg/metrics/metric` package?

christarazi: The structure looks good. I think it's unfortunate that there's an import cycle, which suggests that further refactoring is needed, but let's not tackle that in this PR. It would be good to keep that as a note in a code comment on the metricsmap cell. 

I'll let Dylan get to the finer details here.

lmb:
@christarazi you set the `dont-merge/preview-only` label. Is that still appropriate?

christarazi (commented):
I think you can squash all three commits into one.

carnerito:
Done.

lmb:
Please resolve all conversations (including "outdated" ones) by clicking "Resolve conversation" if the feedback has been addressed. Otherwise merging is not possible.

carnerito:
@lmb Done. All conversations are resolved

lmb:
@carnerito you've squashed the commits but forgot to adjust the commit message.

carnerito:
> @carnerito you've squashed the commits but forgot to adjust the commit message.

@lmb Sorry for that, it's fixed now.

christarazi:
@carnerito There seems to be a legitimate failure in the Ginkgo tests regarding BPF metrics:

```
• Failure [106.658 seconds]
K8sDatapathServicesTest
  Checks N/S loadbalancing
    Supports IPv4 fragments [It]

    Failed to account for INGRESS IPv4 fragments to 10.96.194.76 in BPF metrics
    Expected
        <[]int | len:2, cap:2>: [0, 0]
    To satisfy at least one of these matchers: [%!s(*matchers.EqualMatcher=&{[0 4]}) %!s(*matchers.EqualMatcher=&{[4 0]})]
```

https://github.com/cilium/cilium/actions/runs/6378208180/job/17308444160

carnerito:
@christarazi Can you rerun tests please. The `cilium bpf metrics` commands are currently reliant on the metrics map. I've made a tweak by moving the metrics map initialization to the var block. However, do note that the Prometheus collector is registered only when the metrics map Cell is invoked. I admit it's a bit clunky, but after some digging it seemed to be the best workaround.

christarazi:
@carnerito The easiest way is to run the ginkgo test suite by passing the current kubeconfig that points to a local kind cluster.

pkg/maps/metricsmap/metricsmap.go
```diff
@@ -137,54 +149,133 @@ func (vs Values) Bytes() uint64 {
...
+			prometheus.BuildFQName(metrics.Namespace, metrics.SubsystemDatapath, "forward_bytes_total"),
+			"Total forwarded bytes, tagged by ingress/egress direction",
+			[]string{metrics.LabelDirection}, nil,
+		),
+	}
+}
+
+func (mc *metricsmapCollector) Collect(ch chan<- prometheus.Metric) {
+	err := Metrics.IterateWithCallback(func(key *Key, values *Values) {
+		if key.IsDrop() {
```
derailed, line 230: We could perhaps dry this up a bit by creating a metricsmap generic type and hang methods to upsert or even exporting metrics on the map vs using functions? i.e
```go
type metricsMap[k labels] map[k]metricValues

func (m metricsMap[labels]) upsert(l labels, b, c float64) {
	m[l] = m[l].inc(b, c)
}
// Then...
err := Metrics.IterateWithCallback(func(key *Key, values *Values) {
		if key.IsDrop() {
		  m.droppedMetricMap.upsert(dropLabels{...}, values.Bytes, value.Counts)
		} else {
		  m.forwardedMetricMap.upsert(forwardLabels{...}, values.Bytes, value.Counts)
		}
                ...
}
```

derailed (commented):
@carnerito Thank you Boris for this PR. Nice work!!

pkg/maps/metricsmap/metricsmap.go
carnerito: @derailed Thanks for review! 
I've created a generic `promMetrics` map with upsert() method as suggested. Code in a callback function looks bit cleaner now. 
Regarding exporting, I agree that it would be better if we could have `m.droppedMetricMap.export` and `m.forwardedMetricMap.export` instead those two for loops at the end of callback function, but in that case we would have to associate prometheus.Desc with labels and values. Since we have two values under same key (labels) in metrics map I don't see any elegant way to do this without rewriting most of the logic, if you have any hints, I'm willing to try :)

carnerito: @derailed Hi, do you have any updates on this?

carnerito:
@aanm Cilium build was failing because of [unused import](https://github.com/cilium/cilium/actions/runs/6704188693/job/18216131655?pr=27370#step:21:396). It's fixed now, could you please rerun tests?

aanm:
@carnerito Thanks, can you add the last commit as part of https://github.com/cilium/cilium/pull/27370/commits/8e741dfcb92ab47b666450c957b4a58d61b1713a? We try that each commit is in a "buildable state"

carnerito:
@aanm Just to be sure, do you want me to squash all commits and force-push?

aanm:
> @aanm Just to be sure, do you want me to squash all commits and force-push?

@carnerito I didn't look at all commits but now that you mentioned it, yes please squash all commits together and force-push. Thank you!

carnerito:
@aanm Done

pkg/maps/metricsmap/metricsmap.go
derailed: @carnerito I think what you have now is indeed fine.

derailed (approved):
@carnerito Thank you for this PR and updates Boris!!

joestringer:
Required tests are passing. Seems to be mainly waiting on @cilium/sig-foundations review due to the cells.go change.

joestringer:
Removing my review, I was only involved here as triager trying to keep PRs moving.

christarazi (approved):

dylandreimerink (approved):

pchaigno:
Is it me or this pull request involuntary renamed a bunch of metrics? For example, what used to be called `cilium_drop_count_total` is now called `cilium_datapath_drop_count_total`. Isn't that a breaking change?

carnerito:
@pchaigno Yes, it's fixed in #29226

pchaigno:
Ah, perfect! Thanks! The fix simply didn't reach my PR yet :slightly_smiling_face:

PR #27377, opened 2023-08-09
helm: Use variable for spire server config container image

### LINKED ISSUE #27282: Spire helm chart - static busy box image

### What happened?

Helm chart issue:
The busybox image in both the statefulset and daemonset templates in the spire-agent and spire-server are static (docker.io)
They should be represented in the values file so you can override them with (for instance) a private repository.

### Cilium Version

1.14

### Kernel Version

n/a

### Kubernetes Version

n/a

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

eegseth:
Previously, the busybox image used for configuring spire server was hardcoded into the statefulset.

This patch moves this image reference to the variable authentication.mutual.spire.install.server.spireConfig.image

Fixes: cilium#27282

Signed-off-by: E Egseth <endre@egseth.no>


eegseth:
Please add label needs-backport/1.14

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -3141,6 +3141,9 @@ authentication:
         enabled: true
         # -- SPIRE namespace to install into
         namespace: cilium-spire
+        # -- SPIRE configuration image
+        spireConfig:
```
squeed: nit -- this should be called `init`, since it matches the container name.

@meyskens can we just use spire-agent? does it have a shell?

eegseth: Renamed. Looks like the spire agent doesen't have a shell.

squeed (approved):

joamaki (approved):

meyskens:
Thanks for catching this! While the change looks good seems some git-foo went wrong and this has 67 commits :sweat_smile: do you mind looking into that so we don't pull in extra reviewers and commits via automation? Thanks!

eegseth:
> Thanks for catching this! While the change looks good seems some git-foo went wrong and this has 67 commits 😅 do you mind looking into that so we don't pull in extra reviewers and commits via automation? Thanks!

Working on it 😅 Missed the signoff message in an old commit, and thus managed to fubar the branch...

eegseth:
Reopening a new pr due to fubar branch...

joestringer:
Heads up that in future it should be sufficient to just force-push the branch into github again once you've got it fixed up locally :-)

Issue #27215, opened 2023-08-02
Agent restart causes temporary connection loss by incorrectly removing still alive services


ysksuzuki:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

cilium-agent incorrectly removes alive services on its restart, and it causes temporary connection loss. The agent calls `SyncWithK8sFinished` before calling `UpsertService()` for all alive services, when the service handler takes time processing the events.  The order is not guaranteed if I understand it correctly.
https://github.com/cilium/cilium/blob/1.12.11/pkg/service/service.go#L1111

Extending `k8s-service-cache-size` might mitigate the issue, but I'm wondering if it's possible to ensure calling `UpdateService` for all services before calling `SyncWithK8sFinished`. For example, checking the length of `ServiceCache.Events` channel buffers. 

We saw about 4,000 lines of `Deleting no longer present service` logs on the agent restart.
```
2023-07-31 14:38:15.695	level=warning msg="Deleting no longer present service" l3n4Addr="{IP:10.0.0.36 L4Addr:{Protocol:NONE Port:3306} Scope:0}" serviceID=9986 subsys=service
2023-07-31 14:38:15.695	level=warning msg="Deleting no longer present service" l3n4Addr="{IP:10.0.0.26 L4Addr:{Protocol:NONE Port:3306} Scope:0}" serviceID=1659 subsys=service
2023-07-31 14:38:15.710	level=warning msg="Deleting no longer present service" l3n4Addr="{IP:10.0.0.3 L4Addr:{Protocol:NONE Port:3306} Scope:0}" serviceID=13654 subsys=service
2023-07-31 14:38:15.722	level=warning msg="Deleting no longer present service" l3n4Addr="{IP:10.0.0.5 L4Addr:{Protocol:NONE Port:33060} Scope:0}" serviceID=13025 subsys=service
2023-07-31 14:38:15.734	level=warning msg="Deleting no longer present service" l3n4Addr="{IP:10.0.0.213 L4Addr:{Protocol:NONE Port:31360} Scope:0}" serviceID=12055 subsys=service
2023-07-31 14:38:15.752	level=warning msg="Deleting no longer present service" l3n4Addr="{IP:10.0.0.230 L4Addr:{Protocol:NONE Port:33060} Scope:0}" serviceID=12609 subsys=service
[...]

PR #27378, opened 2023-08-09
k8s: fix incorrect EndpointSlice API version [1.12]

nebril:
This is a backport of https://github.com/cilium/cilium/pull/27277 to 1.12 branch.

This commit fixes a bug that k8s watcher incorrectly chooses the discovery/v1beta1 EndpointSlice in environments where discovery/v1 EndpointSlice is available.
The agent doesn't wait for the discovery/v1 EndpointSlice to be received and ends up removing live services because of this bug.

The issue happens in the following scenario:

1. The agent initializes and runs the informers for Service and EndpointSlice.
2. The reflectors list the resources from kube-apiserver and replace DeltaFIFO.
3. The informers pop items from the DeltaFIFO and call the onAdd handler one by one.
4. If the Service handler is called first, ServiceCache.UpdateService doesn't emit the UpdateService event, because the service is not ready. (No backend)
5. The EndpointSlice handler is called next, ServiceCache.updateEndpoints emits UpdateService event.
6. The agent doesn't wait until the event triggered at step 5 is processed, and calls SyncWithK8sFinished.
7. The agent removes live services because it hasn't heard about it. (It is supposed to hear about those services in the event process triggered at step 5)

fixes: #27215

### DIFF
--- a/pkg/k8s/watchers/watcher.go
+++ b/pkg/k8s/watchers/watcher.go
@@ -419,10 +419,10 @@ func (k *K8sWatcher) resourceGroups() (beforeNodeInitGroups, afterNodeInitGroups
 
 	// To perform the service translation and have the BPF LB datapath
 	// with the right service -> backend (k8s endpoints) translation.
-	if k8s.SupportsEndpointSlice() {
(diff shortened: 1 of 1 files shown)


nebril:
/test-backport-1.12

ysksuzuki (approved):
LGTM. Thanks!

PR #27391, opened 2023-08-09
CI: Rename workflow names

brlbil:
This commit changes the workflow's name fields to make them more human-readable by adding spaces.
Also, some of the names were changed to be more consistent with each other.

### DIFF
--- a/.github/workflows/build-images-base.yaml
+++ b/.github/workflows/build-images-base.yaml
@@ -23,6 +23,7 @@ concurrency:
 
 jobs:
   build-and-push:
+    name: Build and Push Images
     timeout-minutes: 45
     environment: release-base-images
     runs-on: ubuntu-20.04
--- a/.github/workflows/build-images-beta.yaml
+++ b/.github/workflows/build-images-beta.yaml
@@ -20,6 +20,7 @@ permissions:
 jobs:
   build-and-push:
     timeout-minutes: 45
+    name: Build and Push Images
     environment: release-beta-images
     runs-on: ubuntu-20.04
     strategy:
--- a/.github/workflows/build-images-ci.yaml
+++ b/.github/workflows/build-images-ci.yaml
@@ -35,6 +35,7 @@ concurrency:
 jobs:
   build-and-push-prs:
     timeout-minutes: 45
+    name: Build and Push Images
     runs-on: ubuntu-20.04
     strategy:
       matrix:
--- a/.github/workflows/build-images-releases.yaml
+++ b/.github/workflows/build-images-releases.yaml
@@ -17,6 +17,7 @@ permissions:
 jobs:
   build-and-push:
     timeout-minutes: 45
+    name: Build and Push Images
     environment: release
     runs-on: ubuntu-20.04
     strategy:
--- a/.github/workflows/conformance-clustermesh.yaml
+++ b/.github/workflows/conformance-clustermesh.yaml
@@ -1,4 +1,4 @@
-name: ClusterMesh (ci-multicluster)
+name: Conformance Cluster Mesh (ci-clustermesh)
 
 # Any change in triggers needs to be reflected in the concurrency group.
 on:
@@ -67,6 +67,7 @@ env:
 
 jobs:
   commit-status-start:
+    name: Commmit Status Start
     runs-on: ubuntu-latest
     steps:
       - name: Set initial commit status
@@ -75,7 +76,7 @@ jobs:
           sha: ${{ inputs.SHA || github.sha }}
 
   installation-and-connectivity:
-    name: "Installation and Connectivity Test"
+    name: Installation and Connectivity Test
     runs-on: ubuntu-latest
     timeout-minutes: 60
     env:
@@ -475,6 +476,7 @@ jobs:
 
   commit-status-final:
     if: ${{ always() }}
+    name: Commit Status Final
     needs: installation-and-connectivity
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/conformance-e2e.yaml
+++ b/.github/workflows/conformance-e2e.yaml
@@ -1,4 +1,4 @@
-name: Cilium Conformance E2E (ci-e2e)
+name: Conformance E2E (ci-e2e)
 
 # Any change in triggers needs to be reflected in the concurrency group.
 on:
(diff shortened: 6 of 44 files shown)


gentoo-root (approved):

.github/workflows/build-images-base.yaml
```diff
@@ -23,6 +23,7 @@ concurrency:
 
 jobs:
   build-and-push:
+    name: Build and Push Images
```
nbusseneau, line 26: Do you think it's useful to have names on all jobs? I'd advocate for explicitly not setting names on jobs unless absolutely useful, and instead make sure the job IDs correctly represent what the job is doing. I think in general the less we need to worry about additional fields the better.

.github/workflows/conformance-clustermesh.yaml
```diff
@@ -1,4 +1,4 @@
-name: ClusterMesh (ci-multicluster)
+name: Conformance Cluster Mesh (ci-clustermesh)
```
nbusseneau, line 1: If the trigger is changed then `.github/ariane-config.yaml` needs to be updated as well :)

.github/workflows/conformance-e2e.yaml
```diff
@@ -1,4 +1,4 @@
-name: Cilium Conformance E2E (ci-e2e)
+name: Conformance E2E (ci-e2e)
```
nbusseneau, line 1: For any workflow name change, we will need to make sure PR required checks are updated accordingly. Once this PR is finalized and before it is merged, can we prepare the list so as to help maintainers?

nbusseneau (commented):
Thanks, I've added some comments below on general strategy, can do another pass later on specifics.

.github/workflows/build-images-base.yaml
brlbil: The idea behind giving these names is to make them more readable for new report tooling

.github/workflows/conformance-e2e.yaml
brlbil: 👍

.github/workflows/conformance-clustermesh.yaml
brlbil: I am not sure what do you mean with update, `.github/ariane-config.yaml` selects workflows by file name. We do not change the file names here.

nbusseneau: In the change above, the trigger advertised in the workflow name went from `ci-multicluster` to `ci-clustermesh`, so the corresponding trigger needs to be updated in Ariane config.

.github/workflows/build-images-base.yaml
nbusseneau: But is there really a big readability advantage in having jobs named `Build and Push Images` instead of `build-and-push`, as long as they have the same name everywhere? 🤔 I would advocate for not introducing bloat if we can avoid it, if everybody else is convinced this is sound then let's do it but I personally have no problem with having `build-and-push` as job name in the dashboard.

.github/workflows/conformance-clustermesh.yaml
brlbil: I see, I'll fix it.

.github/workflows/build-images-base.yaml
brlbil: I am not sure how much of a readability advantage. I have added a workflow with both jobs with names and without to see the difference. In my option, it is more pleasant if the names are separated by space.

Screenshot

.github/workflows/build-images-beta.yaml
```diff
@@ -20,6 +20,7 @@ permissions:
 jobs:
   build-and-push:
     timeout-minutes: 45
+    name: Build and Push Images
```
nebril, line 23: Should this denote that these are beta images to differentiate from other types of images?

.github/workflows/build-images-ci.yaml
```diff
@@ -35,6 +35,7 @@ concurrency:
 jobs:
   build-and-push-prs:
     timeout-minutes: 45
+    name: Build and Push Images
```
nebril, line 38: Should this denote that these are CI images to differentiate from other types of images?

.github/workflows/build-images-releases.yaml
```diff
@@ -17,6 +17,7 @@ permissions:
 jobs:
   build-and-push:
     timeout-minutes: 45
+    name: Build and Push Images
```
nebril, line 20: Should this denote that these are release images to differentiate from other types of images?

nebril (approved):

.github/workflows/build-images-base.yaml
nbusseneau: I guess the question is "is it pleasant enough to warrant additional bloat in workflows". Personally I would say no, but again I am only one person, would be nice to have more opinions :D

.github/workflows/build-images-beta.yaml
brlbil: This is a `job` name, the `workflow` name is already unique. It describes what the job does.
Same for the others.

.github/workflows/build-images-base.yaml
aanm: @brlbil I don't understand your comment, should there be two pictures?

aanm: Although 'name' is not mandatory, we should still add it to make it explicit and consistent across other workflows.

aanm (approved):

.github/workflows/build-images-base.yaml
nbusseneau: Consistency I agree with -- but if there's already the same ID across workflows, then I'm personally not in favor of additionally copy/pasting the same name. Again, if I'm alone on this hill, let's just go with it :D

brlbil:
Reviews in, failed tests are not related to the changes, setting as ready-to-merge

joestringer:
@brlbil @nbusseneau how do we ensure consistency over time for these? I would like to avoid making rename changes like this as much as possible in future due to the disruption to PRs (renaming => have to rebase and rerun CI for every PR). Every time we create a new CI job there is a risk that we'll deviate again and then later on end up submitting a change again. We are still periodically adding more CI jobs. Is this something that @cilium/ci-structure will now enforce?

nbusseneau:
I think yes, the right way forward would be to be stricter when doing reviews.

PR #27392, opened 2023-08-09
health: only launch /hello after host datapath is ready

lmb:
Delay starting the /hello endpoint until we've loaded the host datapath at least once. This means that the presence of /health can be used to infer not only that the cilium unix socket API is up but also that the datapath can do basic packet processing.

```release-note
Creation of the /hello endpoint is delayed until the host datapath has been initialized.
```

### DIFF
--- a/cilium-health/launch/launcher.go
+++ b/cilium-health/launch/launcher.go
@@ -41,7 +41,7 @@ const (
 )
 
 // Launch starts the cilium-health server and returns a handle to obtain its status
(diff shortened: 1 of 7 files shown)


lmb:
Travis CI hit https://github.com/cilium/cilium/issues/26617

thorn3r (approved):
Changes look good :+1:

dylandreimerink (approved):

PR #27395, opened 2023-08-09
Don't retry one shot jobs during hive shutdown

giorio94:
Stop retrying to execute the function of a one shot job as soon as the job context is canceled, as the hive is already shutting down at that point. Indeed, the expectation is that the function exits immediately when the context expires, and it is most likely to return an error in that condition. Retrying would only delay the hive shutdown (also due to the additional backoff delay between consecutive attempts).

### DIFF
--- a/pkg/hive/job/job.go
+++ b/pkg/hive/job/job.go
@@ -16,6 +16,7 @@ import (
 	"github.com/cilium/cilium/pkg/hive"
 	"github.com/cilium/cilium/pkg/hive/cell"
 	"github.com/cilium/cilium/pkg/hive/internal"
+	"github.com/cilium/cilium/pkg/inctimer"
 	"github.com/cilium/cilium/pkg/lock"
 	"github.com/cilium/cilium/pkg/spanstat"
 	"github.com/cilium/cilium/pkg/stream"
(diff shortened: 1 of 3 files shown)


giorio94:
Converting back to draft while investigating test failures.

giorio94:
conformance-ginkgo hit https://github.com/cilium/cilium/issues/13071. Restarting the affected matrix entry.

pkg/hive/job/job.go
```diff
@@ -277,15 +278,24 @@ func (jos *jobOneShot) start(ctx context.Context, wg *sync.WaitGroup, options op
 
 	stat := &spanstat.SpanStat{}
 
+	timer, cancel := inctimer.New()
+	cancel()
```
rgo3:
```suggestion
	defer cancel()
```

rgo3 (changes requested):

pkg/hive/job/job.go
giorio94: Good catch, thanks!

rgo3 (approved):

rgo3:
Not really sure, deferring the question to @dylandreimerink or @joamaki :grimacing:

PR #27396, opened 2023-08-09
Update kvstore_duration_operation_seconds to not include rate-limiting latency

marseel:
Before, `kvstore_duration_operation_seconds` metric included time spent in rate-limiting.
Now it does not include rate-limited time spent.
This change makes the metric more useful when troubleshooting if propagation delays are caused by slow kvstore or rate-limiting.
Users can still observe delays caused by rate-limiting using metric `api_limiter_wait_duration_seconds`

```release-note
*_kvstore_operations_duration_seconds metrics do not include client-side rate-limiting latency anymore.
```

### DIFF
--- a/Documentation/operations/upgrade.rst
+++ b/Documentation/operations/upgrade.rst
@@ -330,6 +330,16 @@ Helm Options
   were deprecated in Cilium 1.14 in favor of ``tls.ca.cert`` and ``tls.ca.key`` respectively,
   and have been removed. The ```clustermesh-apiserver-ca-cert`` secret is no longer generated.
 
+Changed Metrics
+~~~~~~~~~~~~~~~
+
+* ``cilium_kvstore_operations_duration_seconds``,
+  ``cilium_clustermesh_apiserver_kvstore_operations_duration_seconds``
+  and ``cilium_kvstoremesh_kvstore_operations_duration_seconds``
+  do not include client-side rate-limiting latency anymore.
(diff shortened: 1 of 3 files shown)


Documentation/operations/upgrade.rst
```diff
@@ -330,6 +330,11 @@ Helm Options
   were deprecated in Cilium 1.14 in favor of ``tls.ca.cert`` and ``tls.ca.key`` respectively,
   and have been removed. The ```clustermesh-apiserver-ca-cert`` secret is no longer generated.
 
+Changed Metrics
+~~~~~~~~~~~~~~~
+
+* ``kvstore_operations_duration_seconds`` does not include client-side rate-limiting latency anymore.
```
ghost:
```suggestion
* ``cilium_kvstore_operations_duration_seconds`` does not include client-side rate-limiting latency anymore.
```
I would use full metric names in documentation, to avoid confusion (unfortunately it's not consistent in the upgrade notes for earlier releases).

ghost (changes requested):

giorio94 (approved):
/lgtm

Documentation/operations/upgrade.rst
marseel: I've added also other variants of this metric that are affected + information on what metric to use to observe client-side rate-limiting now.

ghost (approved):

PR #27397, opened 2023-08-09
gha: fix waiting for images in conformance-gingko

giorio94:
Let's add an explicit dependency to the wait-for-images job, so that subsequent jobs are not started until the images are actually available. This also prevents starting the full matrix jobs if the images don't get ready in time.

Example run after the fix (with reduce timeout to trigger failure): https://github.com/cilium/cilium/actions/runs/5812104573/job/15756825851


aanm (approved):
Nice catch!

tklauser (approved):

Issue #27382, opened 2023-08-09
Deadlock in 1.12 ipcache metadata


marseel:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Deadlock in ipcache was introduced by https://github.com/cilium/cilium/pull/26958
The version of Cilium with this change has not been released yet. 

My current understanding of it:
- PR introduced delete in metadata
- This delete needs to acquire metadata mutex
- Metadata delete operation is performed under IPCache lock https://github.com/meyskens/cilium/blob/9be3b287363bbb2c1ef3160aa5e4bb2c236574fa/pkg/ipcache/ipcache.go#L576
- This is in contradiction to comment https://github.com/cilium/cilium/blob/835531431df58070c70699ca4fad5ffc935153a8/pkg/ipcache/metadata.go#L44-L46

### Cilium Version

1.12 branch (unreleased)

### Kernel Version

N/A

### Kubernetes Version

N/A

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

joestringer:
Fixed via https://github.com/cilium/cilium/pull/27406 .

PR #27406, opened 2023-08-09
[v1.12] Remove remote-node labels from ipcache on node delete

joestringer:
Previously the node manager code only ever performed upserts into the
ipcache for remote-node labels on node creation/update events. However,
it never cleaned up these labels upon delete. As a result, if a node was
ever removed, then Cilium would continue to consider traffic towards
that IP as traffic reaching towards a remote node. In extreme cases,
this IP could be reallocated to another host in the network which could
cause connectivity disruption to that peer, particularly if the new peer
is subject to CIDR or Entities based policies.

Commits:
- Revert #26958
- node: Remove labels from ipcache on delete events

Related: https://github.com/cilium/cilium/pull/26958
Fixes: https://github.com/cilium/cilium/issues/27382

Passes the first two test cases here:
https://gist.github.com/joestringer/d7f55da5dd2f33be6148dd52a96a4a54

### DIFF
--- a/pkg/ipcache/metadata.go
+++ b/pkg/ipcache/metadata.go
@@ -79,12 +79,6 @@ func (m *metadata) upsert(prefix string, lbls labels.Labels) {
 	m.Unlock()
 }
 
-func (m *metadata) delete(prefix string) {
-	m.Lock()
-	delete(m.m, prefix)
-	m.Unlock()
-}
-
 // GetIDMetadataByIP returns the associated labels with an IP. The caller must
 // not modifying the returned object as it's a live reference to the underlying
 // map.
--- a/pkg/ipcache/metadata.go
+++ b/pkg/ipcache/metadata.go
@@ -79,6 +79,14 @@ func (m *metadata) upsert(prefix string, lbls labels.Labels) {
 	m.Unlock()
 }
 
+func (ipc *IPCache) RemoveMetadata(prefix string, lbls labels.Labels, src source.Source) {
+	ipc.metadata.Lock()
+	ipc.Lock()
+	_ = ipc.removeLabels(prefix, lbls, src)
+	ipc.Unlock()
+	ipc.metadata.Unlock()
+}
+
 // GetIDMetadataByIP returns the associated labels with an IP. The caller must
 // not modifying the returned object as it's a live reference to the underlying
 // map.
--- a/pkg/node/manager/manager.go
+++ b/pkg/node/manager/manager.go
@@ -55,6 +55,7 @@ type IPCache interface {
 	Delete(IP string, source source.Source) bool
 	TriggerLabelInjection(source source.Source)
 	UpsertMetadata(string, labels.Labels)
+	RemoveMetadata(prefix string, lbls labels.Labels, src source.Source)
 }
 
 // Configuration is the set of configuration options the node manager depends
@@ -526,13 +527,14 @@ func (m *Manager) NodeUpdated(n nodeTypes.Node) {
 			}
 			oldNodeIPAddrs = append(oldNodeIPAddrs, address.IP)
 		}
-		m.deleteIPCache(oldNode.Source, oldNodeIPAddrs, ipsAdded)
+
+		m.deleteIPCache(oldNode.Source, oldNodeIPAddrs, ipsAdded, remoteHostIdentity)
 
 		// Delete the old health IP addresses if they have changed in this node.
-		m.deleteIPCache(oldNode.Source, []net.IP{oldNode.IPv4HealthIP, oldNode.IPv6HealthIP}, healthIPsAdded)
+		m.deleteIPCache(oldNode.Source, []net.IP{oldNode.IPv4HealthIP, oldNode.IPv6HealthIP}, healthIPsAdded, identity.IdentityUnknown)
 
 		// Delete the old ingress IP addresses if they have changed in this node.
-		m.deleteIPCache(oldNode.Source, []net.IP{oldNode.IPv4IngressIP, oldNode.IPv6IngressIP}, ingressIPsAdded)
+		m.deleteIPCache(oldNode.Source, []net.IP{oldNode.IPv4IngressIP, oldNode.IPv6IngressIP}, ingressIPsAdded, identity.IdentityUnknown)
 
 		entry.mutex.Unlock()
 	} else {
@@ -565,9 +567,17 @@ func (m *Manager) upsertIntoIDMD(prefix string, id identity.NumericIdentity) {
 	}
 }
 
+func (m *Manager) removeFromIDMD(prefix string, id identity.NumericIdentity, src source.Source) {
+	if id == identity.ReservedIdentityHost {
+		m.ipcache.RemoveMetadata(prefix, labels.LabelHost, src)
+	} else {
+		m.ipcache.RemoveMetadata(prefix, labels.LabelRemoteNode, src)
+	}
+}
+
 // deleteIPCache deletes the IP addresses from the IPCache with the 'oldSource'
 // if they are not found in the newIPs slice.
-func (m *Manager) deleteIPCache(oldSource source.Source, oldIPs []net.IP, newIPs []string) {
+func (m *Manager) deleteIPCache(oldSource source.Source, oldIPs []net.IP, newIPs []string, remoteID identity.NumericIdentity) {
 	for _, address := range oldIPs {
 		if address == nil {
 			continue
@@ -583,6 +593,9 @@ func (m *Manager) deleteIPCache(oldSource source.Source, oldIPs []net.IP, newIPs
 		// Delete from the IPCache if the node's IP addresses was not
 		// added in this update.
 		if !found {
+			if remoteID != identity.IdentityUnknown {
+				m.removeFromIDMD(addrStr, remoteID, oldSource)
+			}
 			m.ipcache.Delete(addrStr, oldSource)
 		}
 	}
@@ -604,6 +617,16 @@ func (m *Manager) NodeDeleted(n nodeTypes.Node) {
 	if !oldNodeExists {
 		m.mutex.Unlock()
 		return
+	}
+
+	remoteHostIdentity := identity.ReservedIdentityHost
+	if m.conf.RemoteNodeIdentitiesEnabled() {
+		nid := identity.NumericIdentity(n.NodeIdentity)
+		if nid != identity.IdentityUnknown && nid != identity.ReservedIdentityHost {
+			remoteHostIdentity = nid
+		} else if !n.IsLocal() {
+			remoteHostIdentity = identity.ReservedIdentityRemoteNode
+		}
 	}
 
 	// If the source is Kubernetes and the node is the node we are running on
@@ -630,7 +653,9 @@ func (m *Manager) NodeDeleted(n nodeTypes.Node) {
 			continue
 		}
 
-		m.ipcache.Delete(address.IP.String(), n.Source)
+		ip := address.IP.String()
+		m.removeFromIDMD(ip, remoteHostIdentity, n.Source)
+		m.ipcache.Delete(ip, n.Source)
 	}
 
 	for _, address := range []net.IP{
--- a/pkg/ipcache/ipcache.go
+++ b/pkg/ipcache/ipcache.go
@@ -491,10 +491,6 @@ func (ipc *IPCache) deleteLocked(ip string, source source.Source) (namedPortsCha
 	delete(ipc.ipToHostIPCache, ip)
 	delete(ipc.ipToK8sMetadata, ip)
 
-	// this removes the IP from the metadata map
-	// this prevents old assigned labels from being re-used
-	ipc.metadata.delete(ip)
-
 	// Update named ports
 	namedPortsChanged = false
 	if oldK8sMeta != nil && len(oldK8sMeta.NamedPorts) > 0 {
--- a/pkg/node/manager/manager_test.go
+++ b/pkg/node/manager/manager_test.go
@@ -87,6 +87,9 @@ func (i *ipcacheMock) TriggerLabelInjection(s source.Source) {
 func (i *ipcacheMock) UpsertMetadata(string, labels.Labels) {
 }
 
+func (i *ipcacheMock) RemoveMetadata(string, labels.Labels, source.Source) {
+}
+
 type signalNodeHandler struct {
 	EnableNodeAddEvent                    bool
 	NodeAddEvent                          chan nodeTypes.Node


christarazi (commented):
Fix looks good to me.

Did you consider whether we need to also call `RemoveMetadata` in the `NodeUpdated` case which handles an update (old node -> new node) for the old node? The line I'm thinking about is [here](https://github.com/cilium/cilium/blob/58f393f5ab3878eab28c7598004d0e145ae6a780/pkg/node/manager/manager.go#L529).

joestringer:
> Did you consider whether we need to also call RemoveMetadata in the NodeUpdated case which handles an update (old node -> new node) for the old node? The line I'm thinking about is [here](https://github.com/cilium/cilium/blob/58f393f5ab3878eab28c7598004d0e145ae6a780/pkg/node/manager/manager.go#L529).

Hmm, at a glance it looks like we should add it there too.

I have a reproducer via Maartje for the "delete node" case, but I'm not sure I understand how to reassign IPs for existing nodes in order to exercise that path.

christarazi:
> I have a reproducer via Maartje for the "delete node" case, but I'm not sure I understand how to reassign IPs for existing nodes in order to exercise that path.

The only thing that comes to mind is manually doing a `kubectl edit ciliumnode <name>` and changing the node's IP from there.

joestringer:
/test-backport-1.12

pkg/node/manager/manager.go
```diff
@@ -526,7 +527,11 @@ func (m *Manager) NodeUpdated(n nodeTypes.Node) {
 			}
 			oldNodeIPAddrs = append(oldNodeIPAddrs, address.IP)
 		}
+
 		m.deleteIPCache(oldNode.Source, oldNodeIPAddrs, ipsAdded)
+		for _, ip := range oldNodeIPAddrs {
+			m.removeFromIDMD(ip.String(), remoteHostIdentity, n.Source)
+		}
```
christarazi: I think you may need to filter this for loop's iteration to exclude `ipsAdded` from `oldNodeIPAddrs`. Otherwise, all old node IPs which still may be the same IPs in the "new" node are getting wiped from the metadata map.

joestringer: Ah yes, that was a bit of a naive attempt at the solution. I'm trying this as the solution:

```
diff --git a/pkg/node/manager/manager.go b/pkg/node/manager/manager.go
index 7ad1d94f4a13..4366d08f00f1 100644
--- a/pkg/node/manager/manager.go
+++ b/pkg/node/manager/manager.go
@@ -528,16 +528,13 @@ func (m *Manager) NodeUpdated(n nodeTypes.Node) {
                        oldNodeIPAddrs = append(oldNodeIPAddrs, address.IP)
                }
 
-               m.deleteIPCache(oldNode.Source, oldNodeIPAddrs, ipsAdded)
-               for _, ip := range oldNodeIPAddrs {
-                       m.removeFromIDMD(ip.String(), remoteHostIdentity, n.Source)
-               }
+               m.deleteIPCache(oldNode.Source, oldNodeIPAddrs, ipsAdded, remoteHostIdentity)
 
                // Delete the old health IP addresses if they have changed in this node.
-               m.deleteIPCache(oldNode.Source, []net.IP{oldNode.IPv4HealthIP, oldNode.IPv6HealthIP}, healthIPsAdded)
+               m.deleteIPCache(oldNode.Source, []net.IP{oldNode.IPv4HealthIP, oldNode.IPv6HealthIP}, healthIPsAdded, identity.IdentityUnknown)
 
                // Delete the old ingress IP addresses if they have changed in this node.
-               m.deleteIPCache(oldNode.Source, []net.IP{oldNode.IPv4IngressIP, oldNode.IPv6IngressIP}, ingressIPsAdded)
+               m.deleteIPCache(oldNode.Source, []net.IP{oldNode.IPv4IngressIP, oldNode.IPv6IngressIP}, ingressIPsAdded, identity.IdentityUnknown)
 
                entry.mutex.Unlock()
        } else {
@@ -580,7 +577,7 @@ func (m *Manager) removeFromIDMD(prefix string, id identity.NumericIdentity, src
 
 // deleteIPCache deletes the IP addresses from the IPCache with the 'oldSource'
 // if they are not found in the newIPs slice.
-func (m *Manager) deleteIPCache(oldSource source.Source, oldIPs []net.IP, newIPs []string) {
+func (m *Manager) deleteIPCache(oldSource source.Source, oldIPs []net.IP, newIPs []string, remoteID identity.NumericIdentity) {
        for _, address := range oldIPs {
                if address == nil {
                        continue
@@ -597,6 +594,9 @@ func (m *Manager) deleteIPCache(oldSource source.Source, oldIPs []net.IP, newIPs
                // added in this update.
                if !found {
                        m.ipcache.Delete(addrStr, oldSource)
+                       if remoteID != identity.IdentityUnknown {
+                               m.removeFromIDMD(addrStr, remoteID, oldSource)
+                       }
                }
        }
 }
```

joestringer:
/test-backport-1.12

joestringer:
Hit https://github.com/cilium/cilium/issues/27443 , https://github.com/cilium/cilium/issues/24840, and an issue where kubectl gets `signal: killed` during K8sIstioTest. None appear related to the PR.

joestringer:
/test-1.16-4.9

joestringer:
/test-1.21-4.9

pkg/ipcache/metadata.go
```diff
@@ -79,10 +79,12 @@ func (m *metadata) upsert(prefix string, lbls labels.Labels) {
...
 }
 
-func (m *metadata) delete(prefix string) {
-	m.Lock()
-	delete(m.m, prefix)
-	m.Unlock()
+func (ipc *IPCache) RemoveMetadata(prefix string, lbls labels.Labels, src source.Source) {
+	ipc.metadata.Lock()
+	ipc.Lock()
+	_ = ipc.removeLabels(prefix, lbls, src)
```
joestringer, line 85: Heads up that this solves cases where there is a clear disassociation of the node first (eg node is deleted from k8s) before associating the node with another resource like a service, fqdn or kube-apiserver.

In order to solve the other order, we need to handle the resulting labels from this function. I think it's reasonable to review this PR as-is since this at least fixes two bugs (deadlock + deletion of the node followed by association with another resource), then we can follow up on the other case.

nebril (approved):

pkg/node/manager/manager.go
```diff
@@ -565,9 +567,17 @@ func (m *Manager) upsertIntoIDMD(prefix string, id identity.NumericIdentity) {
 	}
 }
 
+func (m *Manager) removeFromIDMD(prefix string, id identity.NumericIdentity, src source.Source) {
+	if id == identity.ReservedIdentityHost {
+		m.ipcache.RemoveMetadata(prefix, labels.LabelHost, src)
+	} else {
+		m.ipcache.RemoveMetadata(prefix, labels.LabelRemoteNode, src)
```
asauber, line 574: Looks like this would be better as a switch statement, as the default here is assuming that the ID is `RemoteNode`

joestringer: Yeah I'd be inclined to agree. I did it this way to match the pattern in the existing `upsertIntoIDMD()`. Moving forward in 1.14 or later this path won't be around so this stylistic change is neither here nor there to me. Might be a little simpler to reason through.

pkg/node/manager/manager.go
```diff
@@ -606,6 +619,16 @@ func (m *Manager) NodeDeleted(n nodeTypes.Node) {
 		return
 	}
 
+	remoteHostIdentity := identity.ReservedIdentityHost
```
asauber, line 622: This block strikes me as too complex. It's at least difficult to read. The name used is `remoteHostIdentity`, but it might end up as a plain numeric identity, remote node, or `host`. 

Can we remove this block and do this instead?

Pass the numeric identity to the call below
```
m.removeFromIDMD(ip, identity.NumericIdentity(n.NodeIdentity), n.Source)
```

Embed this logic of this block in a switch block within that function, possibly with comments explaining the unhandled cases? e.g. "we do not attempt to remove an unknown identity".

asauber (changes requested):

pkg/node/manager/manager.go
asauber: Are there any negative side effects if the label passed here does not exist for the IP?

joestringer: The inner implementation of (`removeIDMD()` -> `RemoveMetadata()` -> ) `removeLabels()` will do a lookup for the prefix and if nothing exists, return nil without complaint. If the prefix has labels but the label does not exist, then `removeLabels()` will use `Labels.Remove()` to remove the label from the set, which is a no-op and that function looks like it handles removing non-existing labels with no trouble. There are a few error check cases where if internal state is a bit inconsistent (for instance, there's no corresponding identity for the target labels for the prefix) then Cilium will emit warnings, which is not very actionable but it won't do anything crazy like crash. This could potentially lead to releasing a reference on an existing identity and removing the entry from the ipcache even though the change in labels should mean this is a no-op, so in that case there is a chance that it could interrupt connectivity for the IP corresponding to the node. So in the general sense this code has not been written defensively to prevent this particular case. At the same time, the caller of this function here will just directly do an `m.ipcache.Delete()` immediately after this, so practically speaking the resulting agent+datapath state in that case is exactly the same whether the label previously existed in the metadata cache or not.

So TL;DR, no there should not be any (additional) negative side effects if there was such a bug.

pkg/node/manager/manager.go
joestringer: Yeah it's complex, but it also matches the `NodeUpdated()` code to appropriately figure out the right host identity depending on how the agent flags for `--enable-remote-node-identity` are configured. We could refactor that logic to make a helper function to hide the complexity if that helps, but my concern with changing this logic is that we could inadvertently introduce another different bug by causing the upsert and delete functions to mismatch.

joestringer: I'll take a look at whether we could refactor / collapse these cases together to improve the readability and code re-use.

ldelossa (approved):

pkg/node/manager/manager.go
joestringer: How does this incremental diff look? Functionally it should be the same but would make the various codepaths a little more consistent:

```
diff --git a/pkg/node/manager/manager.go b/pkg/node/manager/manager.go
index 60a51e464a16..7fa110ac771f 100644
--- a/pkg/node/manager/manager.go
+++ b/pkg/node/manager/manager.go
@@ -376,6 +376,19 @@ func (m *Manager) legacyNodeIpBehavior() bool {
        return true
 }
 
+func (m *Manager) calculateRemoteHostIdentity(n nodeTypes.Node) identity.NumericIdentity {
+       remoteHostIdentity := identity.ReservedIdentityHost
+       if m.conf.RemoteNodeIdentitiesEnabled() {
+               nid := identity.NumericIdentity(n.NodeIdentity)
+               if nid != identity.IdentityUnknown && nid != identity.ReservedIdentityHost {
+                       remoteHostIdentity = nid
+               } else if !n.IsLocal() {
+                       remoteHostIdentity = identity.ReservedIdentityRemoteNode
+               }
+       }
+       return remoteHostIdentity
+}
+
 // NodeUpdated is called after the information of a node has been updated. The
 // node in the manager is added or updated if the source is allowed to update
 // the node. If an update or addition has occurred, NodeUpdate() of the datapath
@@ -387,15 +400,7 @@ func (m *Manager) NodeUpdated(n nodeTypes.Node) {
        dpUpdate := true
        nodeIP := n.GetNodeIP(false)
 
-       remoteHostIdentity := identity.ReservedIdentityHost
-       if m.conf.RemoteNodeIdentitiesEnabled() {
-               nid := identity.NumericIdentity(n.NodeIdentity)
-               if nid != identity.IdentityUnknown && nid != identity.ReservedIdentityHost {
-                       remoteHostIdentity = nid
-               } else if !n.IsLocal() {
-                       remoteHostIdentity = identity.ReservedIdentityRemoteNode
-               }
-       }
+       remoteHostIdentity := m.calculateRemoteHostIdentity(n)
 
        var ipsAdded, healthIPsAdded, ingressIPsAdded []string
 
@@ -560,17 +565,25 @@ func (m *Manager) NodeUpdated(n nodeTypes.Node) {
 // (IDMD) map. The given node identity determines which labels are associated
 // with the CIDR.
 func (m *Manager) upsertIntoIDMD(prefix string, id identity.NumericIdentity) {
-       if id == identity.ReservedIdentityHost {
+       switch id {
+       case identity.IdentityUnknown:
+               // Explicit skip: Health and Ingress IPs aren't injected into
+               // m.ipcache.metadata currently.
+       case identity.ReservedIdentityHost:
                m.ipcache.UpsertMetadata(prefix, labels.LabelHost)
-       } else {
+       default:
                m.ipcache.UpsertMetadata(prefix, labels.LabelRemoteNode)
        }
 }
 
 func (m *Manager) removeFromIDMD(prefix string, id identity.NumericIdentity, src source.Source) {
-       if id == identity.ReservedIdentityHost {
+       switch id {
+       case identity.IdentityUnknown:
+               // Explicit skip: Health and Ingress IPs aren't injected into
+               // m.ipcache.metadata currently, so no need to remove them.
+       case identity.ReservedIdentityHost:
                m.ipcache.RemoveMetadata(prefix, labels.LabelHost, src)
-       } else {
+       default:
                m.ipcache.RemoveMetadata(prefix, labels.LabelRemoteNode, src)
        }
 }
@@ -619,16 +632,6 @@ func (m *Manager) NodeDeleted(n nodeTypes.Node) {
                return
        }
 
-       remoteHostIdentity := identity.ReservedIdentityHost
-       if m.conf.RemoteNodeIdentitiesEnabled() {
-               nid := identity.NumericIdentity(n.NodeIdentity)
-               if nid != identity.IdentityUnknown && nid != identity.ReservedIdentityHost {
-                       remoteHostIdentity = nid
-               } else if !n.IsLocal() {
-                       remoteHostIdentity = identity.ReservedIdentityRemoteNode
-               }
-       }
-
        // If the source is Kubernetes and the node is the node we are running on
        // Kubernetes is giving us a hint it is about to delete our node. Close down
        // the agent gracefully in this case.
@@ -653,6 +656,7 @@ func (m *Manager) NodeDeleted(n nodeTypes.Node) {
                        continue
                }
 
+               remoteHostIdentity := m.calculateRemoteHostIdentity(n)
                ip := address.IP.String()
                m.removeFromIDMD(ip, remoteHostIdentity, n.Source)
                m.ipcache.Delete(ip, n.Source)
```

joestringer: I looked at trying to shuffle this complexity into the removeIDMD() function but that results in making multiple unnecessary calls to the calculate function and it starts to make me nervous whether the refactor may introduce additional unexpected behaviour.

Ultimately though I think that the tidying up of this code would be better targeted at the main branch rather than in a bugfix. I don't think there is a functional difference here that we're exploring, and the code is pretty consistent with existing code (even if that code is a little ugly).

joestringer: Ah for that matter, the whole `upsertIntoIDMD` path is already cleaned up in v1.14/main, so there's no need for improving it there. And the node identity labels function is already refactored.

asauber: I definitely prefer this refactored version

asauber: Saw your most recent two comments after reading and refreshing. I understand where you're coming from and I'll dismiss my review for this one.

asauber (approved):

pkg/node/manager/manager.go
joestringer: We may end up needing some additional refactoring like this sooner or later, depending on how critical the other bug mentioned [here](https://github.com/cilium/cilium/pull/27406#pullrequestreview-1572925607) ends up becoming. I'll store this up my sleeve..

PR #27414, opened 2023-08-10
egressgw: inject datapath config via hive

lmb:
datapath: allow injecting defines into NodeConfig via a map

    The existing functionality to inject defines via a function is necessary for
    cases where the defines are not know at construction time. If they are
    known, it's nicer to be able to specify a plain map. This gives quicker
    feedback if duplicate defines are attempted. It also allows leaving the map
    nil if a feature is disabled, where in the existing implementation you'd
    need a dummy function.

egressgw: inject datapath config via hive

    Remove the hardcoded defines from the HeaderfileWriter and inject them via a
    hive value group instead. This allows adding back EGRESS_POLICY_MAP_SIZE.


julianwiedmann (approved):
lgtm

pkg/maps/egressmap/policy.go
```diff
@@ -66,12 +68,29 @@ type policyMap struct {
 	m *ebpf.Map
 }
 
-func createPolicyMapFromDaemonConfig(daemonConfig *option.DaemonConfig, lc hive.Lifecycle, cfg PolicyConfig) bpf.MapOut[PolicyMap] {
-	if !daemonConfig.EnableIPv4EgressGateway {
-		return bpf.NewMapOut[PolicyMap](nil)
+func createPolicyMapFromDaemonConfig(in struct {
```
dylandreimerink, line 71: nit: I don't love the use of these anonymous structs, wouldn't breaking them out make the code more readable?

dylandreimerink (approved):
The nit is not blocking, just had to say something. Looks good otherwise

pkg/maps/egressmap/policy.go
lmb: Not sure I love them either. Really for this function I'd just like to take / return three arguments, but I can't since hive needs the In and out annotations. Is it better to come up with a type name for each of these? `type policyParams` is ok, `type policyOut` is kinda wonky imo.

How would named types make this more readable? Because the function signature would be more conventional?

PR #27428, opened 2023-08-10
egressgw: test for conflicting IP rules in ENI mode

julianwiedmann:
When the DestCIDR of a policy changes, we first add the new IP rules before
deleting the old ones. But if the new DestCIDR is 0.0.0.0, attempting to
install the matching IP rules results in an -EEXIST error.

To deal with this, addMissingIpRulesAndRoutes() returns a `shouldRetry`
value. If it is set, we retry the installation after
removeUnusedIpRulesAndRoutes() has deleted the unused rules.

Add a test for this scenario.


jibi (approved):

julianwiedmann:
(no-change rebase to get CI healthy)

PR #27430, opened 2023-08-10
Refactor the per-cluster NAT maps manager

giorio94:
Let's simplify the per-cluster NAT maps management logic, removing the dependency on the global variable (which does not play well with the hive framework) and splitting the creation/removal tasks from the retrieval of inner maps.

While being at it, let's also slightly improve the performance around these operations by avoiding to attempt opening the inner maps when not strictly necessary.

Finally, let's also move the fake implementation for testing purposes to a separate package, for better separation.

### DIFF
--- a/cilium/cmd/bpf_nat_list.go
+++ b/cilium/cmd/bpf_nat_list.go
@@ -36,7 +36,7 @@ var bpfNatListCmd = &cobra.Command{
 				cmd.PrintErrf("Invalid ClusterID: %s", err.Error())
 				return
 			}
-			ipv4, ipv6, err := nat.ClusterMaps(uint32(clusterID), true, false)
+			ipv4, ipv6, err := nat.ClusterMaps(uint32(clusterID), true, getIpv6EnableStatus())
 			if err != nil {
 				cmd.PrintErrf("Failed to retrieve cluster maps: %s", err.Error())
 				return
--- a/cilium/cmd/bpf_nat_list.go
+++ b/cilium/cmd/bpf_nat_list.go
@@ -15,7 +15,6 @@ import (
 	"github.com/cilium/cilium/pkg/command"
 	"github.com/cilium/cilium/pkg/common"
 	"github.com/cilium/cilium/pkg/maps/nat"
-	"github.com/cilium/cilium/pkg/option"
 )
 
 // bpfNatListCmd represents the bpf_nat_list command
@@ -32,11 +31,6 @@ var bpfNatListCmd = &cobra.Command{
 			globalMaps[1] = ipv6
 			dumpNat(globalMaps)
 		} else if len(args) == 2 && args[0] == "cluster" {
-			err := nat.InitPerClusterNATMaps(nat.PerClusterNATOuterMapPrefix, true, getIpv6EnableStatus(), option.LimitTableMax)
-			if err != nil {
-				cmd.PrintErrf("Failed to initialize per-cluster SNAT maps: %s\n", err.Error())
-				return
-			}
 			clusterID, err := strconv.ParseUint(args[1], 10, 32)
 			if err != nil {
 				cmd.PrintErrf("Invalid ClusterID: %s", err.Error())
--- a/pkg/maps/nat/fake/per_cluster.go
+++ b/pkg/maps/nat/fake/per_cluster.go
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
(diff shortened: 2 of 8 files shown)


cilium/cmd/bpf_nat_list.go
```diff
@@ -32,17 +31,12 @@ var bpfNatListCmd = &cobra.Command{
...
-			if err != nil {
-				cmd.PrintErrf("Failed to initialize per-cluster SNAT maps: %s\n", err.Error())
-				return
-			}
 			clusterID, err := strconv.ParseUint(args[1], 10, 32)
 			if err != nil {
 				cmd.PrintErrf("Invalid ClusterID: %s", err.Error())
 				return
 			}
-			ipv4, ipv6, err := nat.ClusterMaps(uint32(clusterID), true, false)
```
gentoo-root, line 45: Not sure I understand this change... It looks as a behavioral change, not purely refactoring. Is it really in scope of this commit?

Is it related to this?
>While being at it, let's also slightly improve the performance around
>these operations by avoiding to attempt opening the inner maps when
>not strictly necessary.

But it used to be constant false, not true.

giorio94: Yes, that's a slight behavioral change. The rationale is to align the `cilium bpf nat list cluster ID` command behavior with that of the one used to retrieve global NAT maps, that is showing also IPv6 maps if that address family is enabled. I've moved this change to a separate commit for additional clarity.

gentoo-root (approved):

pkg/maps/nat/types.go
```diff
@@ -12,6 +12,20 @@ import (
 	"github.com/cilium/cilium/pkg/u8proto"
 )
 
+type IPFamily bool
```
derailed, line 16: nit: comment

pkg/maps/nat/types.go
```diff
@@ -12,6 +12,20 @@ import (
...
+
+const (
+	IPv4 = IPFamily(true)
```
derailed, line 20: nit: comment

pkg/maps/nat/per_cluster_nat.go
```diff
@@ -15,49 +14,86 @@ import (
...
-// Global interface to interact with IPv4 and v6 NAT maps. We can choose the
-// implementation of this at startup time by choosing InitPerClusterNATMaps
-// or InitDummyPerClusterNATMaps for initialization.
-var PerClusterNATMaps PerClusterNATMapper
+// ClusterOuterMapName returns the name of the outer per-cluster NAT map
+// for the given IP family. It can be overwritten for testing purposes.
+var ClusterOuterMapName = clusterOuterMapName
 
-// An interface to interact with the global map.
+func clusterOuterMapName(family IPFamily) string {
```
derailed, line 30: Looks a bit cumbersome doing the same thing over and over aka if ipv4 else ...  and xxxV4() xxxV6() parity with for the most part very similar logic?? 
Could be out of scope for this PR but seems leveraging generics would help cleaning this further??

pkg/maps/nat/fake/per_cluster.go
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package nat
```
derailed, line 4: Guessing this is used for testing? If so might make sense to only lasso this in while in test mode and thus use package nat_test instead.

derailed (changes requested):
@giorio94 Nice work Marco! Just a few small picks and comment...

pkg/maps/nat/fake/per_cluster.go
giorio94: Yes, this is the fake implementation provided for testing purposes. AFAIU it is not possible to import the methods defined inside a `something_test.go` file from another package, so that would not work. Still, this PR already moves the fake implementation to a separate package (under the `fake` directory), so that it can be imported separately from the one with the real implementation.

pkg/maps/nat/per_cluster_nat.go
giorio94: Yeah, I agree that there's room for improvement there, especially to reduce code duplication. I would postpone that refactoring though, as this PR is already pretty large, and most occurrences are in the `nat.go` file, which is only touched tangentially by this PR (to replace the ipv4 boolean with the IPFamily constants).

derailed (approved):
@giorio94 Thanks for the updates Marco! LGTM

PR #27433, opened 2023-08-10
Replace use of `strict` to true for kubeProxyReplacement in helm chart

### LINKED ISSUE #27060: `partial` and `strict` are deprecated for helm installation

xtineskim:
- when using the `waitForKubeProxy` flag in helm installation, passing the `--set kubeProxyReplacement=true` during a helm installation of cilium causes an installation error. Replacing the use of `strict` to `true` as strict is deprecated 

Fixes: #27060

### DIFF
--- a/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
+++ b/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
@@ -634,7 +634,7 @@ spec:
         resources:
           {{- toYaml . | trim | nindent 10 }}
         {{- end }}
-      {{- if and .Values.waitForKubeProxy (or (ne $kubeProxyReplacement "strict") (ne $kubeProxyReplacement "true"))  }}
+      {{- if and .Values.waitForKubeProxy (and (ne $kubeProxyReplacement "strict") (ne $kubeProxyReplacement "true"))  }}
       - name: wait-for-kube-proxy
         image: {{ include "cilium.image" .Values.image | quote }}
         imagePullPolicy: {{ .Values.image.pullPolicy }}
--- a/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
+++ b/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
@@ -634,7 +634,7 @@ spec:
         resources:
           {{- toYaml . | trim | nindent 10 }}
         {{- end }}
-      {{- if and .Values.waitForKubeProxy (ne $kubeProxyReplacement "true") }}
+      {{- if and .Values.waitForKubeProxy (or (ne $kubeProxyReplacement "strict") (ne $kubeProxyReplacement "true"))  }}
       - name: wait-for-kube-proxy
         image: {{ include "cilium.image" .Values.image | quote }}
         imagePullPolicy: {{ .Values.image.pullPolicy }}
--- a/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
+++ b/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
@@ -634,7 +634,7 @@ spec:
         resources:
           {{- toYaml . | trim | nindent 10 }}
         {{- end }}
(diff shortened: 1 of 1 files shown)


kaworu (approved):

joamaki (approved):

joestringer:
@xtineskim in general in Cilium we try to retain linear git history, so I've clicked the "rebase" button at the bottom to put this change on top of the tip of `main` and avoid the "merge" commits from earlier. I'll kick off CI and then this should likely be good to merge. Thanks for the submission!

joestringer:
Runtime job hit https://github.com/cilium/cilium/issues/26617 : https://github.com/cilium/cilium/actions/runs/5959529285/job/16165429974

borkmann:
> * when using the `waitForKubeProxy` flag in helm installation, passing the `--set kubeProxyReplacement=true` during a helm installation of cilium causes an installation error. Replacing the use of `strict` to `true` as strict is deprecated

Cc @brb should this be either "true" or "strict"? Meaning, if people still use the "strict" it should not fail either, no?

joestringer:
@xtineskim not sure if there was a conflict or something, but just a heads up that your PRs will almost _always_ be "out of date" compared to the `main` branch. If there's no conflict, then there's usually no reason to rebase. Every time you rebase, we'll need to re-trigger CI.

joestringer (changes requested):
Relating to @borkmann's observation above:

https://github.com/cilium/cilium/blob/6217d01630eb26734f956ac0b801587e05822c13/install/kubernetes/cilium/templates/cilium-configmap.yaml#L66-L69

In the current version, this setting can be set to true, false, disabled, partial or strict. It seems to me like as long as the above helm validation allows "strict", then this check should also cope with that version of the setting.

Either we need to make this an AND statement for not equal to "strict" or "true" so that the daemonset only waits for kube-proxy when kube proxy replacement is disabled, OR we need to change the lines at the top of this message to disallow the "strict" / "partial" / "disabled" settings.

xtineskim:
good catch, and thanks for the feedback. No super strong opinions - I think I'd prefer allowing strict and true (not sure if there is a hard timeline for removing the use of strict/partial)?

joestringer:
(CI passed)

xtineskim:
changed to allow for "strict" or "true"
For enabling w/ "true", users have to run with `--set-string` vs enabling w/ "strict", which just uses `--set`

install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
```diff
@@ -634,7 +634,7 @@ spec:
         resources:
           {{- toYaml . | trim | nindent 10 }}
         {{- end }}
-      {{- if and .Values.waitForKubeProxy (ne $kubeProxyReplacement "strict") }}
+      {{- if and .Values.waitForKubeProxy (or (ne $kubeProxyReplacement "strict") (ne $kubeProxyReplacement "true"))  }}
```
joestringer: Let's say the user configured wait for kubeproxy and kubeproxyreplacement=strict. That will boil down into:

`and true (or (false) (true))`, so the result will be to wait for kube-proxy even though kube-proxy is disabled.

I think that the inner `or` needs to be an `and`...?

```suggestion
      {{- if and .Values.waitForKubeProxy (and (ne $kubeProxyReplacement "strict") (ne $kubeProxyReplacement "true"))  }}
```

joestringer (approved):

brb (approved):
Thanks!

PR #27445, opened 2023-08-11
bpf: install proxy routes using Go, remove init.sh

### LINKED ISSUE #26255: init.sh: create L7 proxy rules in Go
The two bash functions in `init.sh` `move_local_rules_af()` and `setup_proxy_rules()` should be ported to Go.

Investigate if these need to be called from Reinitialize or can move somewhere else.

ti-mo:
```
proxy: implement routing setup in Go

This commit ports the setup of routing table 2004 from init.sh to Go.

It does not remove the old logic from init.sh yet, as it is the last remaining
piece. Removing the script and its surrounding code will follow in a separate
commit.
```

```
init.sh: remove script

bye bye
```

Fixes: #26255

```release-note
bpf: install proxy routes using Go, remove init.sh
```


pkg/proxy/routes.go
```diff
@@ -0,0 +1,102 @@
...
+	tproxyMask = 0xf00
+
+	// Routing table for redirecting packets to the proxy through the local stack.
+	proxyRoutingTable = 2004
+)
+
+var (
+	// Routing rule for traffic to proxy.
+	tproxyRule = route.Rule{
+		Priority: 9,
```
dylandreimerink: I don't know how to feel about this one. I know this is just taken from init.sh. But it is an oddly specific number.

dylandreimerink (approved):

pkg/proxy/routes.go
ti-mo: Would you like to see this assigned to a constant? Or a rationale behind the value 9?

dylandreimerink: I would add a comment at least, if it makes sense inline there or else on a constant. That comment would ideally have a rationale, but if we can't find one it would be nice to know that as well. If its just arbitrary between 0 and 100 at the low side of things then someone doesn't have to be afraid to touch it in fear of some Indiana Jones trap springing at any moment.

markpash (approved):
bye bye init.sh! :champagne:

pkg/proxy/proxy.go
```diff
@@ -423,9 +423,35 @@ func (p *Proxy) SetProxyPort(name string, proxyType ProxyType, port uint16, ingr
...
+	} else {
+		if err := removeRoutesIPv6(); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // ReinstallRules is called by daemon reconfiguration to re-install proxy ports rules that
```
mhofstetter: nit: adapt name in comment

pkg/proxy/routes.go
```diff
@@ -0,0 +1,102 @@
...
+		Device: "lo",
+		Proto:  linux_defaults.RTProto,
+	}
+)
+
+// installRoutesIPv4 configures routes and rules needed to redirect ingress
+// packets to the proxy.
+func installRoutesIPv4() error {
+	if err := route.Upsert(route4); err != nil {
+		return fmt.Errorf("inserting ipv4 proxy route %v: %w", route4, err)
```
mhofstetter, line 61: nit: change all error messages to something like `failed to insert ...`

mhofstetter (approved):
thanks @ti-mo :rocket: only two non-blocking nits from my side

pkg/proxy/routes.go
ti-mo: We usually log these as errors, and error wrapping makes 'failed' a little redundant. Anecdotally, we tend to avoid it in ebpf-go. For example, it comes out as `error="failed to insert ..."` or `x went wrong: failed to insert ...`.

IMO being terse is more readable since there's less reading to be done for unwinding the stack trace: `reticulating splines: shaving yaks: inserting ipv4 proxy route ...`, etc.

pkg/proxy/proxy.go
ti-mo: Thanks, good catch.

pkg/proxy/routes.go
mhofstetter: thanks for the explanation - seems reasonable to me!

tklauser (approved):
Nice work!

pkg/proxy/routes.go
ti-mo: Oh, but that's half the fun. :wink: It's indeed as you described; we move the kernel's default pref 0 rule to pref 100 to clear up 0-99 for our own purposes. I've documented this in the code now.

PR #27447, opened 2023-08-11
[DRAFT] policy: lock the SelectorCache for reading only

bimmlerd:
ConsumeMapChanges and it's call tree below only require read access to the selector cache. Hence, avoid serialising all calls to ConsumeMapChanges by acquiring the full (write) lock.


bimmlerd:
/test-backport-v1.12

bimmlerd:
/test-backport-1.12

PR #27356, opened 2023-08-08
Provide CT/NAT maps GC logic through hive

giorio94:
Please refer to the commit descriptions for additional details.

PR #27448, opened 2023-08-11
Refactor the per-cluster CT maps manager

giorio94:
Let's simplify the per-cluster CT maps management logic, removing the dependency on the global variable (which does not play well with the hive framework) and splitting the creation/removal tasks from the retrieval of inner maps.

While being at it, let's also slightly improve the performance around these operations by avoiding to attempt opening the inner maps when not strictly necessary, as well as keeping the list of known Cluster IDs.

Additionally, let's also move the fake implementation for testing purposes to a separate package, for better separation.

Finally, let's validate the ClusterID in the `bpf ct list cluster <ID>` CLI command, to properly return an error in case it is invalid.

Related: https://github.com/cilium/cilium/pull/27430
~~Depends on: https://github.com/cilium/cilium/pull/27356~~

### DIFF
--- a/pkg/maps/ctmap/per_cluster_ctmap.go
+++ b/pkg/maps/ctmap/per_cluster_ctmap.go
@@ -6,69 +6,90 @@ package ctmap
 import (
 	"errors"
 	"fmt"
-	"io/fs"
 	"reflect"
 	"strconv"
 
-	"golang.org/x/sys/unix"
-
 	"github.com/cilium/ebpf"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/cilium/cilium/pkg/bpf"
 	cmtypes "github.com/cilium/cilium/pkg/clustermesh/types"
 	"github.com/cilium/cilium/pkg/lock"
 )
 
-// Global map that contains all per-cluster CT maps. The actual
-// implementations are either perClusterCTMaps for real deployment
-// and privileged tests or dummyPerClusterCTMaps for unprivileged
-// testing.
-var PerClusterCTMaps PerClusterCTMapper
-
 const (
 	perClusterCTMapMaxEntries = cmtypes.ClusterIDMax + 1
 
-	PerClusterCTOuterMapPrefix   = "cilium_per_cluster_ct_"
-	perClusterTCP4OuterMapSuffix = "tcp4"
-	perClusterANY4OuterMapSuffix = "any4"
-	perClusterTCP6OuterMapSuffix = "tcp6"
-	perClusterANY6OuterMapSuffix = "any6"
+	perClusterCTOuterMapPrefix = "cilium_per_cluster_ct_"
 )
 
-// An interface to interact with all per-cluster CT maps
+// ClusterOuterMapName returns the name of the outer per-cluster CT map
+// for the given type. It can be overwritten for testing purposes.
+var ClusterOuterMapName = clusterOuterMapName
+
+func clusterOuterMapName(typ mapType) string {
+	return perClusterCTOuterMapPrefix + typ.name()
+}
+
+func ClusterOuterMapNameTestOverride(prefix string) {
+	ClusterOuterMapName = func(typ mapType) string {
+		return prefix + "_" + clusterOuterMapName(typ)
+	}
+}
+
+// ClusterInnerMapName returns the name of the inner per-cluster NAT map
+// for the given IP family and cluster ID.
+func ClusterInnerMapName(typ mapType, clusterID uint32) string {
+	return ClusterOuterMapName(typ) + "_" + strconv.FormatUint(uint64(clusterID), 10)
+}
+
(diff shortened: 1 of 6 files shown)


pkg/maps/ctmap/per_cluster_ctmap.go
```diff
@@ -102,379 +123,124 @@ type PerClusterCTMapVal struct {
...
-	return ims, nil
+	return gm.foreach(func(om *PerClusterCTMap) error {
+		return om.cleanup()
+	})
 }
 
-func (gm *dummyPerClusterCTMaps) Cleanup() {
-	gm.RLock()
-	defer gm.RUnlock()
+func (gm *perClusterCTMaps) foreach(fn func(om *PerClusterCTMap) error) error {
```
YutaroHayakawa, line 239: I do not fully understand what happens when the operation against one of the maps is failed. Should we rollback the changes or do you think is it ok to leave it as is?

pkg/maps/ctmap/per_cluster_ctmap.go
```diff
@@ -102,379 +123,124 @@ type PerClusterCTMapVal struct {
...
-	}
-
-	return nil
-}
-
-func (gm *perClusterCTMaps) GetClusterCTMaps(clusterID uint32) []*Map {
-	var (
-		err error
-		im  *Map
+	gm.clusterIDs.Insert(clusterID)
```
YutaroHayakawa, line 173: Should we rollback this Insert when the below map creations fail?

pkg/maps/ctmap/per_cluster_ctmap.go
```diff
@@ -102,379 +123,124 @@ type PerClusterCTMapVal struct {
...
-		gm.any4[clusterID] = struct{}{}
-	}
-
-	if gm.ipv6 {
-		gm.tcp6[clusterID] = struct{}{}
-		gm.any6[clusterID] = struct{}{}
-	}
-
-	return nil
+	gm.clusterIDs.Delete(clusterID)
```
YutaroHayakawa, line 192: Should we rollback this Delete when the below operation failed?

YutaroHayakawa (changes requested):
I have some question around error handling. Otherwise, LGTM 👍

pkg/maps/ctmap/per_cluster_ctmap.go
giorio94: Honestly this is also the aspect I was also doubtful about. At the end, I currently adopted the simplest approach, always performing the given operation on all maps, without attempting any rollback process on the already processed maps in case of errors. 

This is because (a) it is likely that the consumer of these methods will retry again the same operation (e.g., through a controller) in case an error has occurred, (b) errors seem unlikely (as syscall errors), but if they happen we probably ended up in a broken scenario and also the subsequent rollback may also fail. 

All in all, my feeling is that implementing the rollback would make the overall logic more complex and fragile (as it is difficult to actually test the error cases), with only very limited benefit. Additionally, it would also be unclear how to rollback deletion operations, as recreating the map would probably not be the best option. 

As for the "process all and concatenate the errors" vs "abort on first error" approaches, I personally preferred the first because it works better especially in case of deletion/cleanup operations, ensuring for instance that we attempt to delete all maps even if one of the earlier operations failed. This is especially useful as in certain cases we might not want to subsequently retry failed deletion operations.

I'm obviously open to reconsider the above if you feel that roll backing the changes on errors would be beneficial. (Edit: otherwise, if this makes sense, I'll add a comment and/or extend the commit message with the above explanation for additional clarity).

pkg/maps/ctmap/per_cluster_ctmap.go
giorio94: I'm actually unsure about that. The list of ClusterIDs is currently used only as an optimization when retrieving all maps for the GC process; in case of errors it doesn't really matter whether they are returned or not, as if the map does not exist it will just be skipped while attempting to open it. 

(Edit: if this makes sense to you, I'll add a clarification comment above that line)

pkg/maps/ctmap/per_cluster_ctmap.go
giorio94: Similar to the above comment, in case we attempted to delete the given maps, it likely makes sense not to return them anymore, regardless of whether the operation failed or not.

giorio94:
@YutaroHayakawa PTAL

pkg/maps/ctmap/per_cluster_ctmap.go
YutaroHayakawa: Yep, all makes sense to me. Thanks for your explanation. I have no objection to going with the current direction 👍

YutaroHayakawa (approved):
My concerns are addressed. LGTM 👍

giorio94:
The last push introduced the additional clarification comments from the discussion above.

giorio94:
@tklauser ping

tklauser (approved):

PR #27469, opened 2023-08-11
helm: Allow configuration of the install-cni container resources

RenaudWasTaken:

- [N/A] All code is covered by unit and/or runtime tests where feasible
      GitHub issue.
- [N/A] If your commit description contains a `Fixes: <commit-id>` tag, then

The helm chart does not provide a way to specify how much resources this container uses.
This change adds a way to configure this field.

```release-note
Helm: Allow configuration of the install-cni container resources field
```

### DIFF
--- a/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
+++ b/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
@@ -689,10 +689,10 @@ spec:
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         command:
           - "/install-plugin.sh"
+        {{- with .Values.cni.resources }}
         resources:
-          requests:
-            cpu: 100m
-            memory: 10Mi
+          {{- toYaml . | trim | nindent 10 }}
+        {{- end }}
         securityContext:
           {{- if .Values.securityContext.privileged }}
           privileged: true
--- a/install/kubernetes/cilium/values.yaml.tmpl
+++ b/install/kubernetes/cilium/values.yaml.tmpl
@@ -593,6 +593,12 @@ cni:
   # inside the agent pod.
   hostConfDirMountPath: /host/etc/cni/net.d
 
+  # -- Specifies the resources for the cni initContainer
+  resources:
+    requests:
+      cpu: 100m
+      memory: 10Mi
+
 # -- (string) Configure how frequently garbage collection should occur for the datapath
 # connection tracking table.
 # @default -- `"0s"`
--- a/Documentation/helm-values.rst
+++ b/Documentation/helm-values.rst
@@ -756,6 +756,10 @@
      - Configure the log file for CNI logging with retention policy of 7 days. Disable CNI file logging by setting this field to empty explicitly.
      - string
      - ``"/var/run/cilium/cilium-cni.log"``
+   * - :spelling:ignore:`cni.resources`
+     - Specifies the resources for the cni initContainer
+     - object
+     - ``{"requests":{"cpu":"100m","memory":"10Mi"}}``
    * - :spelling:ignore:`cni.uninstall`
      - Remove the CNI configuration and binary files on agent shutdown. Enable this if you're removing Cilium from the cluster. Disable this to prevent the CNI configuration file from being removed during agent upgrade, which can cause nodes to go unmanageable.
      - bool
--- a/install/kubernetes/cilium/README.md
+++ b/install/kubernetes/cilium/README.md
@@ -239,6 +239,7 @@ contributors across the globe, there is almost always someone available to help.
 | cni.hostConfDirMountPath | string | `"/host/etc/cni/net.d"` | Configure the path to where the CNI configuration directory is mounted inside the agent pod. |
 | cni.install | bool | `true` | Install the CNI configuration and binary files into the filesystem. |
 | cni.logFile | string | `"/var/run/cilium/cilium-cni.log"` | Configure the log file for CNI logging with retention policy of 7 days. Disable CNI file logging by setting this field to empty explicitly. |
+| cni.resources | object | `{"requests":{"cpu":"100m","memory":"10Mi"}}` | Specifies the resources for the cni initContainer |
 | cni.uninstall | bool | `false` | Remove the CNI configuration and binary files on agent shutdown. Enable this if you're removing Cilium from the cluster. Disable this to prevent the CNI configuration file from being removed during agent upgrade, which can cause nodes to go unmanageable. |
 | conntrackGCInterval | string | `"0s"` | Configure how frequently garbage collection should occur for the datapath connection tracking table. |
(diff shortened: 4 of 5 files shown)


RenaudWasTaken:
Hello @joamaki and @kaworu!

Gentle bump on this small PR, if you can take a look at it when you have some time and let me know if it looks good :) 

Thanks a lot!

ti-mo:
Added @squeed as a reviewer to hopefully speed things along.
@RenaudWasTaken Could you please include the reason for this change in your commit message and PR description? Does the absence of this cause problems? Also, is 10MiB enough? The `cilium-cni` binary alone weighs 47MiB.

kaworu (approved):
Helm changes LGTM, also interested about [@squeed question](https://github.com/cilium/cilium/pull/27469#issuecomment-1680694209).

install/kubernetes/cilium/values.yaml.tmpl
joestringer, line 1: How did you come to the values here?

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -584,6 +584,12 @@ cni:
   # inside the agent pod.
   hostConfDirMountPath: /host/etc/cni/net.d
 
+  # -- Specifies the resources for the cni container
+  resources:
+    requests:
+      cpu: 100m
+      memory: 10Mi
+
```
joestringer: I'd feel more confident about this PR without setting the defaults.
```suggestion
```

joestringer (changes requested):
Thanks for the PR. In principle I don't see why not to make this configurable if it's important for your deployment case, since this presumably would help to reserve additional required resources on the nodes, and perhaps in constrained environments that can be the difference between being able to reliably redeploy Cilium or not.

At the same time, I'm not sure I understand where the limits came from or how we could keep these accurately up to date over time. I'd suggest dropping the defaults and just making it configurable, then in your environment you can specify whatever limits you find to be viable.

I'm also curious like the others above if there's anything particular about your environments that motivate this change?

joestringer:
It looks like [the smoke tests failed](https://github.com/cilium/cilium/actions/runs/5837500429/job/16127891461?pr=27469), see the error below.

```
please run 'make -C install/kubernetes' and submit your changes
```

RenaudWasTaken:
Hello!

Sorry for the lag! I'll fix the PR early September (I'm out of office until then).

Our use case is that we run a pod at node initialization that runs checks on the node before releasing it for general scheduling.

To do that it's currently configured to take all the resources of the node.

Typically to support this use case we set out system daemonsets such that requests and limits are at 0.

When we deploy cilium with this init container, because of the kubelet accounting strategy, the node does not have it's full resources available.

Which then prevents our pod from scheduling 😅

install/kubernetes/cilium/values.yaml.tmpl
RenaudWasTaken: They were already present in `install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml`, It seemed reasonable to not introduce any changes to the default

install/kubernetes/cilium/values.yaml.tmpl
RenaudWasTaken: So the defaults are already set to these values today in `install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml`, I'm happy to not set them, but that would introduce a behavior change that might or might not break consumers of the chart :)

RenaudWasTaken:
Alright, I've update the PR with the specific changes, but I don't seem to have the ability to mark my PR out of draft
@joestringer could you take another stab and move the PR out of draft?

Thanks a lot!

RenaudWasTaken:
@kaworu, @joestringer and @squeed gentle bump in your inbox :)

joestringer:
Do you see this button near the bottom of the page? This is the one I was talking about.

image

joestringer:
I can run the testsuite to gather that feedback.

joestringer (approved):
Looks to me like nothing is functionally changing for the defaults here, the PR just makes the resources for this particular container configurable :+1:

install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml
```diff
@@ -674,10 +674,10 @@ spec:
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         command:
           - "/install-plugin.sh"
+        {{- with .Values.cni.resources }}
```
joestringer, line 692: One minor comment just for awareness: Because this is dereferencing a new option, we cannot consider backporting this change. It would need to do a null check first if we want to port it to an existing release. As such this PR would first arrive in an upcoming `v1.15-pre.*` snapshot and of course in the final v1.15.x releases.

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -584,6 +584,12 @@ cni:
   # inside the agent pod.
   hostConfDirMountPath: /host/etc/cni/net.d
 
+  # -- Specifies the resources for the cni container
```
squeed: minor nit: please call this the "cni initContainer"

squeed: Another request: the top-level `cni` grouping is for broader CNI settings, rather than individual agent DaemonSet bits. I'd rather see this as part of the agent daemonset configuration.

Unfortunately, we don't have a super clean way of doing that, since the agent settings are top-level, rather than being grouped under something.

So, I'd rather see the key be `initContainerResources.install-cni`.

RenaudWasTaken: Updated to `cni initContainer`, sorry for the lag, I completely missed this

RenaudWasTaken: So you want:
```
cni:
    ....
initContainerResources:
    install-cni:
        enabled: true
        resources:
        ....
```

Is that an accurate description? A few questions:
- Do we break compatibility with the existing options? If not how do we navigate the world with the two options 😬 ?
- Do I also migrate the other init containers?

squeed: (sorry, inbox got a bit ahead of me)

That was my idea. However, if you think the cni resources are causing more harm than good, then I'd be OK with removing them. The only thing the container does is a few `cp`s.

Also, a cleaner hierarchy would probably be

```yaml
initContainers:
  install-cni-binaries:
    resources:
      ...
```

Thoughts?

joamaki (approved):

install/kubernetes/cilium/values.yaml.tmpl
joestringer: @RenaudWasTaken were you able to apply this feedback? Are you waiting on @squeed ?

RenaudWasTaken: Sorry about the lag, the main thought was how do we manage backwards compatibility?

aanm: cc @squeed

squeed (approved):

squeed:
@RenaudWasTaken this should go in, I'm just rebasing on `main` to pick up some test fixes.

RenaudWasTaken:
Thanks for pushing this PR over the line!

PR #27472, opened 2023-08-13
gateway-api: Support all the extended features

sayboras:
### Description

This PR is to support the below extended feature in Gateway API, which is currently missing in profile conformance report https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/reports/v0.7.1/cilium-cilium.yaml.

- [x] HTTPRouteSchemeRedirect
- [x] HTTPRouteHostRewrite
- [x]  HTTPRouteRequestMirror
- [x] HTTPRoutePathRedirect
- [x] HTTPRoutePathRewrite
- [x] HTTPRoutePortRedirect

### DIFF
--- a/operator/pkg/model/ingestion/gateway.go
+++ b/operator/pkg/model/ingestion/gateway.go
@@ -89,31 +89,28 @@ func GatewayAPI(input Input) ([]model.HTTPListener, []model.TLSListener) {
 				var responseHeaderFilter *model.HTTPHeaderFilter
 				var requestRedirectFilter *model.HTTPRequestRedirectFilter
 				var rewriteFilter *model.HTTPURLRewriteFilter
-				if len(rule.Filters) > 0 {
-					for _, f := range rule.Filters {
-						if f.Type == gatewayv1beta1.HTTPRouteFilterRequestHeaderModifier {
-							requestHeaderFilter = &model.HTTPHeaderFilter{
-								HeadersToAdd:    toHTTPHeaders(f.RequestHeaderModifier.Add),
-								HeadersToSet:    toHTTPHeaders(f.RequestHeaderModifier.Set),
-								HeadersToRemove: f.RequestHeaderModifier.Remove,
-							}
-						}
+				var requestMirror *model.HTTPRequestMirror
 
-						if f.Type == gatewayv1beta1.HTTPRouteFilterResponseHeaderModifier {
-							responseHeaderFilter = &model.HTTPHeaderFilter{
-								HeadersToAdd:    toHTTPHeaders(f.ResponseHeaderModifier.Add),
-								HeadersToSet:    toHTTPHeaders(f.ResponseHeaderModifier.Set),
-								HeadersToRemove: f.ResponseHeaderModifier.Remove,
-							}
+				for _, f := range rule.Filters {
+					switch f.Type {
+					case gatewayv1beta1.HTTPRouteFilterRequestHeaderModifier:
+						requestHeaderFilter = &model.HTTPHeaderFilter{
+							HeadersToAdd:    toHTTPHeaders(f.RequestHeaderModifier.Add),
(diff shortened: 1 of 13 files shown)


nbusseneau (approved):
Blind ack for `github-sec` considering the trivial changes.

operator/pkg/model/ingestion/gateway.go
```diff
@@ -120,6 +122,7 @@ func GatewayAPI(input Input) ([]model.HTTPListener, []model.TLSListener) {
 						RequestHeaderFilter:    requestHeaderFilter,
 						ResponseHeaderModifier: responseHeaderFilter,
 						RequestRedirect:        requestRedirectFilter,
+						Rewrite:                rewriteFilter,
```
mhofstetter, line 125: isn't it necessary to add `RequestMirror` in case no matches are defined on the rule too?

operator/pkg/model/translation/envoy_listener.go
```diff
@@ -280,3 +274,20 @@ func newTransportSocket(ciliumSecretNamespace string, tls []model.TLSSecret) (*e
 		},
 	}, nil
 }
+
+func toFilterChainMatch(hostNames []string) *envoy_config_listener.FilterChainMatch {
+	var res *envoy_config_listener.FilterChainMatch
+	// ServerNames must be sorted and unique, however, envoy don't support "*" as a server name
+	serverNames := slices.SortedUnique(hostNames)
+	if len(serverNames) == 0 || (len(serverNames) == 1 && serverNames[0] == "*") {
+		res = &envoy_config_listener.FilterChainMatch{
```
mhofstetter: no need to define the variable `res` and completely set to its need in the respective condition blocks.

I would
* **either** simply handle the "edge case" (no names or `*`) in one if directly returning the FilterChainMatch. And otherwise returning the actual FilterChainMatch.
* **or** initialize the var `res` with the FilterChainMatch with the TransportProtocol set. and in one if statement handle the case where servernames is not empty  and not `*` and then set the serverNames.

operator/pkg/model/translation/envoy_virtual_host.go
```diff
@@ -348,11 +445,18 @@ func getRouteMatch(hostnames []string, hostNameSuffixMatch bool, pathMatch model
 		}
 	}
 	if pathMatch.Prefix != "" {
-		return &envoy_config_route_v3.RouteMatch{
-			PathSpecifier: &envoy_config_route_v3.RouteMatch_SafeRegex{
-				SafeRegex: &envoy_type_matcher_v3.RegexMatcher{
-					Regex: getMatchingPrefixRegex(pathMatch.Prefix),
+		if pathMatch.Prefix == "/" {
```
mhofstetter: instead of introducing the nested `if`-statement I would refactor the whole method from the multiple `ifs` (with early returns) to a switch statement of the following form.

```
switch {
case pathMatch.Exact != "":
  return ...
case pathMatch.Prefix == "/":
  return ...
case pathMatch.Prefix != "":
  return ...
case pathMatch.Regex != "":
  return ...
default:
  return ...
}
```

mhofstetter (approved):
LGTM - thanks a lot @sayboras ! :rocket: Only some non blocking input from my side (see inline comments).

btw: I didn't check all functionality in detail and hope this is covered by Gateway-API conformance tests too.

operator/pkg/model/translation/envoy_listener.go
sayboras: good point :100:

sayboras: I go with option 2, which is pretty nice to read.

operator/pkg/model/translation/envoy_virtual_host.go
sayboras: ah nice, I didn't know that switch {} is allowed :face_exhaling:

PR #27488, opened 2023-08-14
bpf: nodeport: integrate Ingress RevSNAT and RevDNAT paths

julianwiedmann:
The Ingress RevSNAT and RevDNAT handling is currently split into two separate tail-calls. Which means that if the RevSNAT code performs a CT lookup and receives trace information, then we can't easily pass the corresponding `trace_ctx` to the `TO_OVERLAY` trace event in the RevDNAT code.

In practice this means that when in direct-routing mode, the trace events for EgressGW reply traffic don't benefit from CT-driven aggregation.

This PR refactors the relevant code paths, and turns the RevDNAT path into a conditional tail-call. On newer kernels it gets inlined into `tail_nodeport_nat_ingress_ipv4()`, and we can pass through the `trace_ctx` struct all the way to the `send_trace_notify()`.

### DIFF
--- a/bpf/lib/nodeport.h
+++ b/bpf/lib/nodeport.h
@@ -1083,14 +1083,14 @@ int tail_nodeport_nat_ingress_ipv6(struct __ctx_buff *ctx)
 		.min_port = NODEPORT_PORT_MIN_NAT,
 		.max_port = NODEPORT_PORT_MAX_NAT,
 	};
-	struct trace_ctx trace __maybe_unused = {
+	struct trace_ctx trace = {
 		.reason = TRACE_REASON_CT_REPLY,
 		.monitor = TRACE_PAYLOAD_LEN,
 	};
 	__s8 ext_err = 0;
 	int ret;
 
-	ret = snat_v6_rev_nat(ctx, &target, &ext_err);
+	ret = snat_v6_rev_nat(ctx, &target, &trace, &ext_err);
 	if (IS_ERR(ret)) {
 		if (ret == NAT_PUNT_TO_STACK ||
 		    /* DROP_NAT_NO_MAPPING is unwanted behavior in a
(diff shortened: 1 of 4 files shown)


julianwiedmann:
Sorry for the massive churn. But this needed a fair bit of refactoring to slash through all the tech debt :grimacing: .

bpf/lib/nodeport.h
```diff
@@ -2452,16 +2451,17 @@ static __always_inline int rev_nodeport_lb4(struct __ctx_buff *ctx, __s8 *ext_er
...
 	int ret = 0;
 #if defined(ENABLE_HOST_FIREWALL) && defined(IS_BPF_HOST)
 	/* We only enforce the host policies if nodeport.h is included from
 	 * bpf_host.
 	 */
-	struct trace_ctx __maybe_unused trace = {
-		.reason = TRACE_REASON_UNKNOWN,
-		.monitor = 0,
-	};
+
```
aspsk: Extra line

bpf/lib/nodeport.h
```diff
@@ -2544,15 +2567,29 @@ int tail_nodeport_nat_ingress_ipv4(struct __ctx_buff *ctx)
...
+					&trace, &ext_err);
+		if (IS_ERR(ret))
+			goto drop_err;
+
+		/* No redirect needed: */
+		if (ret == CTX_ACT_OK)
+			goto recircle;
+
+		/* Redirected to egress interface: */
+		edt_set_aggregate(ctx, 0);
```
aspsk, line 2584: previously we set it only when !LXC. Can this be called from LXC?

aspsk (commented):
LGTM, left one question and one cosmetic comment

bpf/lib/nodeport.h
julianwiedmann: nice catch, thanks! Fixed.

bpf/lib/nodeport.h
julianwiedmann: I added the following to the patch description, thank you!

>    Note that CILIUM_CALL_IPV*_NODEPORT_NAT_INGRESS is never invoked from  bpf_lxc, so we can remove the #ifdef's around edt_set_aggregate().

aspsk (approved):
LGTM, thanks!

julianwiedmann:
> (one last rebase to pull #27509 in)

this now required a small change for this PR: https://github.com/cilium/cilium/pull/27488/commits/944acfd51aa4280bc21eac5221d69c81df6794c5.

PR #27491, opened 2023-08-14
egressgw: refactor check for conflicting egress IPs

lmb:
The egressgateway code currently doesn't handle some configurations correctly when using ip rules and routes to steer traffic. We've recently added some logic to detect this error. The approach was to only install rules for the first non-conflicting policy.

The problem with this approach is that the egress gateway doesn't process policies in a deterministic fashion. So in the case of a conflict we will choose different "first" policies for each reconciliation. The end result is no better than what we had before, where the "last" policy would win.
The check also can't address the fact that other nodes in the cluster will still send traffic to the "misconfigured" gatewaty.

Refactor the existing check into a separate piece of code, which makes it easier to understand. Also drop the logic which attempts to remove conflicting policies, since it doesn't have the desired effect and complicates behaviour.

Fixes: 76c9822440 ("egressgw: detect conflicting configurations in ENI mode")

### DIFF
--- a/pkg/egressgateway/manager.go
+++ b/pkg/egressgateway/manager.go
@@ -4,6 +4,7 @@
 package egressgateway
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


julianwiedmann:
looks good, thank you!

julianwiedmann (approved):

PR #27492, opened 2023-08-14
bpf: allow overriding Makefile variables

lmb:
Consolidate the important variables at the top of bpf/Makefile and allow injecting custom logic via Makefile.override. This is the same approach used for the root Makefile as well.


dylandreimerink (approved):
:+1:

PR #27509, opened 2023-08-15
bpf: nat: use common set of rewrite helpers

julianwiedmann:
We've grown way too many variants of helpers that rewrite the packet
headers for NAT purposes. They all slightly differ in terms of what
input parameters they expect, and what types of ICMP packets they support.

This PR introduces a generic set of flexible helpers to rule them all, and converts over many of the old paths.

There's a few more places to convert, but I believe this is good enough for a first shot. I'll take on the others as a follow-on.

### DIFF
--- a/bpf/lib/nat.h
+++ b/bpf/lib/nat.h
@@ -552,63 +552,6 @@ snat_v4_icmp_rewrite_egress_embedded(struct __ctx_buff *ctx,
 	return 0;
 }
 
-static __always_inline int snat_v4_rewrite_egress(struct __ctx_buff *ctx,
-						  struct ipv4_ct_tuple *tuple,
-						  struct ipv4_nat_entry *state,
-						  __u32 off, bool has_l4_header)
-{
-	int ret, flags = BPF_F_PSEUDO_HDR;
-	struct csum_offset csum = {};
-	__be32 sum;
-
-	if (state->to_saddr == tuple->saddr &&
-	    state->to_sport == tuple->sport)
-		return 0;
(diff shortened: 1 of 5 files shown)


bpf/lib/nat.h
```diff
@@ -353,6 +353,74 @@ snat_v4_rev_nat_handle_mapping(struct __ctx_buff *ctx,
...
+			case IPPROTO_ICMP:
+				if (l4_store_port(ctx, l4_off, port_off, new_port) < 0)
+					return DROP_WRITE_ERROR;
+
+				from = old_port;
+				to = new_port;
+				l4_sum = csum_diff(&from, 4, &to, 4, 0);
+
+				/* Not initialized by csum_l4_offset_and_flags(): */
+				csum.offset = offsetof(struct icmphdr, checksum);
```
YutaroHayakawa, line 393: Why can't we extend `csum_l4_offset_and_flags`?

YutaroHayakawa (approved):
LGTM with nit.

bpf/lib/nat.h
julianwiedmann: I think we totally can.

My best guess why it hasn't been done so far is that it introduces overhead to all users. I'd like to understand a bit better how much additional complexity these generic helpers are introducing to paths that weren't exposed to ICMP before (eg. `__lb4_rev_nat()`). And fine-tune if needed. Same would apply here - ideally this would a `ipv4_init_csum()` that doesn't care about ICMPv6.

PR #27515, opened 2023-08-15
bpf: compile test ENABLE_EGRESS_GATEWAY_COMMON

lmb:
Ensure that only defining ENABLE_EGRESS_GATEWAY_COMMON actually works. Also fix the issues that this uncovered.

Fixes: 6cad3dc324 ("bpf: gate egressgw datapath on separate defines")


bpf/lib/egress_gateway.h
```diff
@@ -76,6 +78,9 @@ bool egress_gw_request_needs_redirect(struct iphdr *ip4, int ct_status,
 	*tunnel_endpoint = egress_gw_policy->gateway_ip;
 	return true;
 #else
+	(void)ip4;
+	(void)ct_status;
+	(void)tunnel_endpoint;
```
julianwiedmann: This is to suppress the warnings about unused parameters, right? Then let's use `__maybe_unused`.

julianwiedmann (approved):

dylandreimerink (approved):

PR #27517, opened 2023-08-15
bpf: Rename proxy_identity to src_sec_identity

joestringer:
Commit 3f356b0156d8 ("bpf: Split handle_ipv4 in bpf_host.c after
ct_lookup4") used the name proxy_identity for this field, but the source
identity might not be coming from the proxy. Also, commit 2ec713e8e3c9
("bpf: rename remote_endpoint_info.sec_label to sec_identity") resolved
to use sec_identity as a base name scheme for all identities held in
variables in the datapath. To reduce confusion or bad assumptions about
the source of the identity, rename it to src_sec_identity.

### DIFF
--- a/bpf/bpf_host.c
+++ b/bpf/bpf_host.c
@@ -367,15 +367,15 @@ handle_ipv6_cont(struct __ctx_buff *ctx, __u32 secctx, const bool from_host,
 static __always_inline int
 tail_handle_ipv6_cont(struct __ctx_buff *ctx, bool from_host)
 {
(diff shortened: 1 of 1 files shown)


borkmann (approved):

gentoo-root (approved):
> Commit https://github.com/cilium/cilium/commit/3f356b0156d86016695f97a0d8e747ca8f3f2ee0 ("bpf: Split handle_ipv4 in bpf_host.c after
ct_lookup4") used the name proxy_identity for this field

Well, I basically reused the existing name, but in more places.

PR #7060, opened 2019-02-13
cilium, bpf: only account tx for egress direction

borkmann:
... and not for services which is buggy.

Signed-off-by: Daniel Borkmann <daniel@iogearbox.net>

---
This change is [Reviewable](https://reviewable.io/reviews/cilium/cilium/7060)

PR #27527, opened 2023-08-16
add metrics for CT_SERVICE in conntrack

### LINKED ISSUE #27452: CFP:  Metrics for CT_SERVICE
## Feature request

We are using cilium xdp geneve dsr mode as L4 load balancer. The metrics like connections/pkts/bytes of the vip and vip's backends are very important to provide observation for the products.

If cilium can support this, it should a really helpful feature .

We used IPVS before, here are the example statistics that provided by IPVS:

Examples for ipvs:
http://www.austintek.com/LVS/LVS-HOWTO/HOWTO/LVS-HOWTO.monitoring_lvs.html

```
ipvsadm -L -t smtp:smtp --stats 
Prot LocalAddress:Port               Conns   InPkts  OutPkts  InBytes OutBytes   -> RemoteAddress:Port TCP  
smtp.uio.no:smtp              1508879 38457326        0   10461M        0   
-> mail-mx6.uio.no:smtp           374117  9490846        0    2664M        0   
-> mail-mx3.uio.no:smtp           377646  9961956        0    2543M        0   
-> mail-mx2.uio.no:smtp           378502  9288837        0    2707M        0   
-> mail-mx1.uio.no:smtp           378614  9715687        0    2546M        0 

# ipvsadm -L -t smtp:smtp --rate Prot LocalAddress:Port                 CPS    InPPS   OutPPS    InBPS   OutBPS   -> RemoteAddress:Port TCP  smtp.uio.no:smtp                    7       85        0    20480        0   
-> mail-mx6.uio.no:smtp                1       17        0     1126        0   
-> mail-mx3.uio.no:smtp                1       17        0     2023        0   
-> mail-mx2.uio.no:smtp                2       26        0     6681        0   
-> mail-mx1.uio.no:smtp                2       25        0    10650        0
```

wenlxie:
Add metrics for dir=CT_SERVICE

Fixes: [#27452](https://github.com/cilium/cilium/issues/27452)

```release-note
Add metrics count for dir=CT_SERVICE and disable conntrack metrics by default
```

### DIFF
--- a/bpf/lib/conntrack.h
+++ b/bpf/lib/conntrack.h
@@ -292,7 +292,7 @@ __ct_lookup(const void *map, struct __ctx_buff *ctx, const void *tuple,
 		if (dir == CT_INGRESS) {
 			__sync_fetch_and_add(&entry->rx_packets, 1);
 			__sync_fetch_and_add(&entry->rx_bytes, ctx_full_len(ctx));
-		} else if (dir == CT_EGRESS) {
+		} else if (dir == CT_EGRESS || dir == CT_SERVICE) {
 			__sync_fetch_and_add(&entry->tx_packets, 1);
 			__sync_fetch_and_add(&entry->tx_bytes, ctx_full_len(ctx));
 		}
@@ -932,13 +932,15 @@ static __always_inline int ct_create6(const void *map_main, const void *map_rela
 	seen_flags.value |= is_tcp ? TCP_FLAG_SYN : 0;
 	ct_update_timeout(&entry, is_tcp, dir, seen_flags);
 
+#ifdef CONNTRACK_ACCOUNTING
 	if (dir == CT_INGRESS) {
 		entry.rx_packets = 1;
 		entry.rx_bytes = ctx_full_len(ctx);
-	} else if (dir == CT_EGRESS) {
+	} else if (dir == CT_EGRESS || dir == CT_SERVICE) {
 		entry.tx_packets = 1;
 		entry.tx_bytes = ctx_full_len(ctx);
 	}
+#endif
 
 	cilium_dbg3(ctx, DBG_CT_CREATED6, entry.rev_nat_index, ct_state->src_sec_id, 0);
 
@@ -1009,13 +1011,15 @@ static __always_inline int ct_create4(const void *map_main,
 	seen_flags.value |= is_tcp ? TCP_FLAG_SYN : 0;
 	ct_update_timeout(&entry, is_tcp, dir, seen_flags);
 
+#ifdef CONNTRACK_ACCOUNTING
 	if (dir == CT_INGRESS) {
 		entry.rx_packets = 1;
 		entry.rx_bytes = ctx_full_len(ctx);
-	} else if (dir == CT_EGRESS) {
+	} else if (dir == CT_EGRESS || dir == CT_SERVICE) {
 		entry.tx_packets = 1;
 		entry.tx_bytes = ctx_full_len(ctx);
 	}
+#endif
 
 	cilium_dbg3(ctx, DBG_CT_CREATED4, entry.rev_nat_index,
 		    ct_state->src_sec_id, 0);
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -1315,7 +1315,7 @@ func initEnv(vp *viper.Viper) {
 	option.Config.Opts.SetBool(option.TraceNotify, true)
 	option.Config.Opts.SetBool(option.PolicyVerdictNotify, true)
 	option.Config.Opts.SetBool(option.PolicyTracing, option.Config.EnableTracing)
-	option.Config.Opts.SetBool(option.ConntrackAccounting, true)
+	option.Config.Opts.SetBool(option.ConntrackAccounting, false)
 	option.Config.Opts.SetBool(option.ConntrackLocal, false)
 	option.Config.Opts.SetBool(option.PolicyAuditMode, option.Config.PolicyAuditMode)
 	option.Config.Opts.SetBool(option.SourceIPVerification, true)
--- a/test/k8s/service_helpers.go
+++ b/test/k8s/service_helpers.go
@@ -272,6 +272,12 @@ func doFragmentedRequest(kubectl *helpers.Kubectl, srcPod string, srcPort, dstPo
 	ciliumPodK8s2, err := kubectl.GetCiliumPodOnNode(helpers.K8s2)
 	ExpectWithOffset(2, err).Should(BeNil(), "Cannot get cilium pod on k8s2")
 
+	res := kubectl.CiliumExecMustSucceed(context.TODO(), ciliumPodK8s1, "cilium config  get ConntrackAccounting")
+	res.ExpectContains("Enabled")
+
+	res = kubectl.CiliumExecMustSucceed(context.TODO(), ciliumPodK8s2, "cilium config  get ConntrackAccounting")
+	res.ExpectContains("Enabled")
+
 	_, dstPodIPK8s1 := kubectl.GetPodOnNodeLabeledWithOffset(helpers.K8s1, testDS, 1)
 	_, dstPodIPK8s2 := kubectl.GetPodOnNodeLabeledWithOffset(helpers.K8s2, testDS, 1)
 
@@ -287,7 +293,7 @@ func doFragmentedRequest(kubectl *helpers.Kubectl, srcPod string, srcPort, dstPo
 	endpointK8s1 := net.JoinHostPort(dstPodIPK8s1, fmt.Sprintf("%d", dstPodPort))
 	patternInK8s1 := fmt.Sprintf("UDP IN [^:]+:%d -> %s", srcPort, endpointK8s1)
 	cmdInK8s1 := fmt.Sprintf(cmdIn, patternInK8s1)
-	res := kubectl.CiliumExecMustSucceed(context.TODO(), ciliumPodK8s1, cmdInK8s1)
+	res = kubectl.CiliumExecMustSucceed(context.TODO(), ciliumPodK8s1, cmdInK8s1)
 	countInK8s1, _ := strconv.Atoi(strings.TrimSpace(res.Stdout()))
 
 	endpointK8s2 := net.JoinHostPort(dstPodIPK8s2, fmt.Sprintf("%d", dstPodPort))
--- a/test/k8s/services.go
+++ b/test/k8s/services.go
@@ -629,7 +629,12 @@ Secondary Interface %s :: IPv4: (%s, %s), IPv6: (%s, %s)`,
 				options["gke.enabled"] = "false"
 				options["routingMode"] = "native"
 			}
+
 			DeployCiliumOptionsAndDNS(kubectl, ciliumFilename, options)
+
+			cmd := fmt.Sprintf("cilium config %s=%s", helpers.OptionConntrackAccounting, helpers.OptionEnabled)
+			kubectl.CiliumExecMustSucceedOnAll(context.TODO(), cmd, "Unable to enable ConntrackAccounting option")
+			kubectl.CiliumPreFlightCheck()
 			testIPv4FragmentSupport(kubectl, ni)
 		})
 


YutaroHayakawa:
Putting this back to draft since it is WIP. Please push the `Ready for review` button when you are ready.

bpf/lib/conntrack.h
```diff
@@ -256,6 +256,11 @@ __ct_lookup(const void *map, struct __ctx_buff *ctx, const void *tuple,
 		} else if (dir == CT_EGRESS) {
 			__sync_fetch_and_add(&entry->tx_packets, 1);
 			__sync_fetch_and_add(&entry->tx_bytes, ctx_full_len(ctx));
+		} else if (dir == CT_SERVICE) {
+#if defined(ENABLE_DSR)	
+			__sync_fetch_and_add(&entry->rx_packets, 1);
+			__sync_fetch_and_add(&entry->rx_bytes, ctx_full_len(ctx));
+#endif
```
julianwiedmann: I'm a bit surprised here. Why can't we just use the `tx_packets` / `tx_bytes` fields for the `CT_SERVICE` case?

wenlxie: @julianwiedmann  For DSR mode, I'd think tx_pacekts/tx_bytes can be used instead of rx_packets and rx_bytes, but is there any case that tx/rx packets tx/rx bytes are both needed for non DSR mode?

julianwiedmann: Nope, we shouldn't be doing any `CT_SERVICE` lookup from the reverse direction.

Non-DSR replies get handled by this special-purpose CT entry: https://github.com/cilium/cilium/blob/78a5f177693fb443cd946441f45826bf7fa2437a/bpf/lib/nodeport.h#L2673

wenlxie: @julianwiedmann I.C Thanks for point this out. I am new to this and mad a mistake here. It is no need to change the struct ct_entry  here. 

Another issue is I want to expose the metrics like IPVS,  any suggestions for how to expose  them?  
 For more details: 
https://github.com/cilium/cilium/issues/27452
Thanks

bpf/lib/conntrack.h
```diff
@@ -256,6 +256,11 @@ __ct_lookup(const void *map, struct __ctx_buff *ctx, const void *tuple,
 		} else if (dir == CT_EGRESS) {
 			__sync_fetch_and_add(&entry->tx_packets, 1);
 			__sync_fetch_and_add(&entry->tx_bytes, ctx_full_len(ctx));
+		} else if (dir == CT_SERVICE) {
+#if defined(ENABLE_DSR)
+			__sync_fetch_and_add(&entry->tx_packets, 1);
+			__sync_fetch_and_add(&entry->tx_bytes, ctx_full_len(ctx));
+#endif
```
julianwiedmann:
```suggestion
		} else if (dir == CT_EGRESS || dir == CT_SERVICE) {
			__sync_fetch_and_add(&entry->tx_packets, 1);
			__sync_fetch_and_add(&entry->tx_bytes, ctx_full_len(ctx));
		}
```

bpf/lib/conntrack.h
```diff
@@ -963,6 +968,11 @@ static __always_inline int ct_create4(const void *map_main,
 	} else if (dir == CT_EGRESS) {
 		entry.tx_packets = 1;
 		entry.tx_bytes = ctx_full_len(ctx);
+	} else if (dir == CT_SERVICE) {
+#if defined(ENABLE_DSR) 
+		entry.tx_packets = 1;
+		entry.tx_bytes = ctx_full_len(ctx);
+#endif
```
julianwiedmann:
```suggestion
	} else if (dir == CT_EGRESS || dir == CT_SERVICE) {
		entry.tx_packets = 1;
		entry.tx_bytes = ctx_full_len(ctx);
	}
```

julianwiedmann (commented):
Thanks! Some cleanup suggestions below.

I believe there's no reason why we would need to limit this to `ENABLE_DSR`. Non-DSR connections should also benefit from these improved statistics, no?

wenlxie:
> here's no reason why we would need to limit this to `ENABLE_DSR`. Non-DSR connections should also benefit from these improved statistics, no?

Fixed

julianwiedmann:
Imho this looks good :+1:. Let's see if @borkmann still has any context on https://github.com/cilium/cilium/pull/7060#issuecomment-1681677872.

borkmann:
Did anyone benchmark this? :( Two more atomic ops for svc traffic.. could we make all the stats conditional so we could leave the accounting out for users not needing it?

wenlxie:
> Did anyone benchmark this? :( Two more atomic ops for svc traffic.. could we make all the stats conditional so we could leave the accounting out for users not needing it?

Does CONNTRACK_ACCOUNTING not enough?  😄

joestringer:
> Does CONNTRACK_ACCOUNTING not enough?

Looks like this is enabled by default, so adding extra latency to every connection by default is a concern yes.

```
$ git grep -i 'conntrack[ ]\?accounting' daemon
daemon/cmd/daemon_main.go:      option.Config.Opts.SetBool(option.ConntrackAccounting, true)
```

borkmann:
> > Does CONNTRACK_ACCOUNTING not enough?
> 
> Looks like this is enabled by default, so adding extra latency to every connection by default is a concern. The configuration option / flag / macro might not be enough.
> 
> ```
> $ git grep -i 'conntrack[ ]\?accounting' daemon
> daemon/cmd/daemon_main.go:      option.Config.Opts.SetBool(option.ConntrackAccounting, true)
> ```

Is anyone actually using the counters? Perhaps we could disable them by default even to also save on the atomic ops on non-svc traffic. Then it's an opt-in and users who would really want to have them enable it with the constraint to pay the cost.

YutaroHayakawa:
Let me unassign myself as there are already many 👀

julianwiedmann:
@joestringer @borkmann just checking, as this PR has been sitting for a while. What's the next step - should setting `option.ConntrackAccounting` to false happen in this PR, or as a separate change?

wenlxie:
> @joestringer @borkmann just checking, as this PR has been sitting for a while. What's the next step - should setting `option.ConntrackAccounting` to false happen in this PR, or as a separate change?

Add setting option.ConntrackAccounting to false by default change to this PR already.

lmb:
Please resolve https://github.com/cilium/cilium/pull/27527#discussion_r1296686344 if it is no longer relevant.

wenlxie:
> Please resolve [#27527 (comment)](https://github.com/cilium/cilium/pull/27527#discussion_r1296686344) if it is no longer relevant.

It is already resolved @lmb

lmb:
Unfortunately you need to click on the `Resolve conversation` button, otherwise merging is not possible.

@borkmann please take another look at the PR.

pippolo84 (approved):
Agent-related changes LGTM. :+1: 
I suggest to add the reason why we decided to change the default for `option.ConntrackAccounting` in the commit description too.

bpf/lib/conntrack.h
```diff
@@ -884,7 +884,7 @@ static __always_inline int ct_create6(const void *map_main, const void *map_rela
 	if (dir == CT_INGRESS) {
```
borkmann, line 936: I think this should be wrapped also with `#ifdef CONNTRACK_ACCOUNTING` like in the lookup case to make it clear that accounting is enabled/disabled. Did you validate at runtime that by default it's disabled now?

Also, small nit: Please add a proper commit message so that it's not empty.

Otherwise lgtm then.

ti-mo:
@wenlxie Do you need any more input here? Looks like this is mostly good to go with a few adjustments.

wenlxie:
> @wenlxie Do you need any more input here? Looks like this is mostly good to go with a few adjustments.

Just come back from holiday. 
Code had been changed according to the comments. 
Thanks

wenlxie:
@ti-mo  Can this PR be merged? Any other things  that need be fixed?

borkmann (approved):
lgtm, thanks for disabling by default

joestringer:
I'm not sure why so many tests are failing but the next step is to get CI green. Please investigate and/or re-run the tests to validate the functionality of the change.

LeiW000:
When can the PR be merged? It's needed as a partial fix for issue #28487.

joestringer:
@wenlxie did you investigate these issues? If you're hitting the same issues again then there may be a problem related to the changes in the PR.

Example failure from Ginkgo run:
```
• Failure [106.718 seconds]
K8sDatapathServicesTest
  Checks N/S loadbalancing
    Supports IPv4 fragments [It]

    Failed to account for IPv4 fragments to 10.96.248.182 (in)
    Expected
        <[]int | len:2, cap:2>: [0, 0]
    To satisfy at least one of these matchers: [%!s(*matchers.EqualMatcher=&{[0 4]}) %!s(*matchers.EqualMatcher=&{[4 0]})]

    /home/runner/work/cilium/cilium/test/k8s/services.go:630
------------------------------
```

Maybe this is trying to match some of the statistics that this PR is disabling?

wenlxie:
> @wenlxie did you investigate these issues? If you're hitting the same issues again then there may be a problem related to the changes in the PR.
> 
> Example failure from Ginkgo run:
> 
> ```
> • Failure [106.718 seconds]
> K8sDatapathServicesTest
> /home/runner/work/cilium/cilium/test/ginkgo-ext/scopes.go:461
>   Checks N/S loadbalancing
>   /home/runner/work/cilium/cilium/test/ginkgo-ext/scopes.go:461
>     Supports IPv4 fragments [It]
>     /home/runner/work/cilium/cilium/test/ginkgo-ext/scopes.go:515
> 
>     Failed to account for IPv4 fragments to 10.96.248.182 (in)
>     Expected
>         <[]int | len:2, cap:2>: [0, 0]
>     To satisfy at least one of these matchers: [%!s(*matchers.EqualMatcher=&{[0 4]}) %!s(*matchers.EqualMatcher=&{[4 0]})]
> 
>     /home/runner/work/cilium/cilium/test/k8s/services.go:630
> ------------------------------
> ```
> 
> Maybe this is trying to match some of the statistics that this PR is disabling?

Yup, test case need to be enhanced

nebril (approved):

joestringer:
Hi @wenlxie , I'm curious why the PR is being updated constantly and re-running CI. Are you actively developing the PR or are the tests failing?

If the PR is getting out of date with origin/main, then you don't need to rebase unless there is a conflict. Cilium is a fast-moving project, and PRs will always be out-of date. As long as the PR has been recently rebased (last couple of weeks or so) and the CI is passing, that should be good enough.

wenlxie:
> Hi @wenlxie , I'm curious why the PR is being updated constantly and re-running CI. Are you actively developing the PR or are the tests failing?
> 
> If the PR is getting out of date with origin/main, then you don't need to rebase unless there is a conflict. Cilium is a fast-moving project, and PRs will always be out-of date. As long as the PR has been recently rebased (last couple of weeks or so) and the CI is passing, that should be good enough.

@joestringer  I met an issue for the test cases. 

Since I had disable the conntrackaccounting feature by default, so in the test, I need to enable it since that case need to get the stats of IN/OUT packets. 
https://github.com/cilium/cilium/pull/27527/files#diff-1b407c35a1ec3381116cb5950eeadff5381668d04516c8746a6f3efee3b4fa6cR633

But when runs the test for ci-ginkgo, I found that when cilium config ConntrackAccouting=enabled command run success, it still may not really take effect when it try to get the stats afterward. So I'd think it take times to make it work. 
Is there any condition that can indicate that the runtime config take effect in data plane  after change?   Any suggests for how I can check for that? Thanks

joestringer:
Ah okay, never mind my feedback then, makes sense :)

I can't think of any obvious reason why it wouldn't take effect. Typically the tests would configure this via Helm at the start of the test instead of making runtime changes like this, so maybe that interface is slightly less well tested, or there's a typo in the change, or there is a race condition where the config requests cilium-agent to apply the change and then cilium-agent will asynchronously apply the change and maybe that happens afterwards?

LeiW000:
It seems to me that @wenlxie has fixed all the test cases. Is it OK to merge the PR now?

wenlxie:
After add check before test, then the failure case had passed. 
https://github.com/cilium/cilium/pull/27527/files#diff-1b407c35a1ec3381116cb5950eeadff5381668d04516c8746a6f3efee3b4fa6cR634
 
I had run it for more than 3 times and the previous error not exist any more. 

@joestringer

LeiW000:
May I know when the PR can be merged?

lmb:
Waiting on review from @cilium/sig-lb. Seems like GH didn't choose a reviewer?

LeiW000:
Hi Folks, I know that the merge window is open now. May I know when the PR can be merged?

dylandreimerink:
@julianwiedmann do you have particular interest in this PR? Given the self-assigned review. Or can anyone from @cilium/sig-lb complete the review?

julianwiedmann:
> @julianwiedmann do you have particular interest in this PR? Given the self-assigned review. Or can anyone from @cilium/sig-lb complete the review?

Huh, I don't even remember why I re-requested the review ... thanks for the ping. Dropping myself, I'm clearly just blocking progress here.

PR #27532, opened 2023-08-16
doc: Migrate to .readthedocs.yaml v2 configuration and upgrade from Python 3.7 to 3.11

### LINKED ISSUE #26299: Migrate to .readthedocs.yaml configuration file v2
We need to migrate to this configuration file at the risk that all our docs will stop being built https://blog.readthedocs.com/migrate-configuration-v2/

We will likely need to backport this to all branches that we support.

As we change the python version we will also need to update this file:
```patch
diff --git a/Documentation/Dockerfile b/Documentation/Dockerfile
index 30ae02186f..aca9a566a8 100644
--- a/Documentation/Dockerfile
+++ b/Documentation/Dockerfile
@@ -1,5 +1,5 @@
 # Python version should match the one in use in Read The Docs
-FROM docker.io/library/python:3.7.9-alpine3.13 AS docs-base
+FROM docker.io/library/python:3.11-alpine3.17 AS docs-base
 
 LABEL maintainer="maintainer@cilium.io"
 ```

which will unblock this PR https://github.com/cilium/cilium/pull/25689

### PR DESCRIPTION
**Pull request by doniacld**
Migrate to .readthedocs.yaml configuration file v2 before it becomes necessary to render documentation and upgrade Python from 3.7 to 3.11.

Fixes: #26299

```release-note
doc: Migrate to .readthedocs.yaml configuration file v2 
```

Successfully generated documentation locally:
```
Running at http://localhost:9081
docker container run --rm --workdir /src/Documentation --volume /Users/donia/workspace/src/github.com/doniacld/cilium/Documentation/..:/src --env READTHEDOCS_VERSION= --env SKIP_LINT= --user "501:20" \
                --publish 9081:8000 \
                        cilium/docs-builder \
                sphinx-autobuild --open-browser --host 0.0.0.0 "-j=auto" --ignore *.swp -Q . _preview
[sphinx-autobuild] > sphinx-build -j auto -Q /src/Documentation /src/Documentation/_preview
```

### DIFF
--- a/Documentation/.readthedocs.yaml
+++ b/Documentation/.readthedocs.yaml
@@ -1,0 +1,23 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
+# Build documentation with Sphinx
+sphinx:
+  configuration: Documentation/conf.py
+  builder: "dirhtml"
+
+# We recommend specifying your dependencies to enable reproducible builds:
+# https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
+python:
+  install:
+  - requirements: Documentation/requirements.txt
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -57,7 +57,7 @@ jobs:
       - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
         with:
           persist-credentials: false
-      - uses: docker://quay.io/cilium/docs-builder:fdcc2524f647c6e09534f8813aac3141ae478b5b@sha256:54b57f47904af397c1545c04d1e6a87f41f46966d57d04f661c09130aa853c29
+      - uses: docker://quay.io/cilium/docs-builder:d5505e230c0164b51855b002c2a0abeeff78870c@sha256:95940ef277ae14c745769e332c968d8357a43e363eab0b088c5d5efa22089260
         with:
           entrypoint: ./Documentation/check-build.sh
           args: html
--- a/Documentation/Dockerfile
+++ b/Documentation/Dockerfile
@@ -1,5 +1,5 @@
 # Python version should match the one in use in Read The Docs
-FROM docker.io/library/python:3.7.9-alpine3.13 AS docs-base
+FROM docker.io/library/python:3.11-alpine3.17 AS docs-base
 
 LABEL maintainer="maintainer@cilium.io"
 
@@ -31,4 +31,4 @@ ENV MAKE_GIT_REPO_SAFE=1
 ## Workaround odd behaviour of sphinx versionwarning extension. It wants to
 ## write runtime data inside a system directory.
 ## We do rely on this extension, so we cannot just drop it.
-RUN install -m 0777 -d /usr/local/lib/python3.7/site-packages/versionwarning/_static/data
+RUN install -m 0777 -d /usr/local/lib/python3.11/site-packages/versionwarning/_static/data


qmonnet:
Given that the PR touches `Documentation/Dockerfile`, we need to re-generate the `docs-builder` image for the CI. I've triggered the relevant deployment; the first job, to build the image, passed successfully. However, the automation failed to update your PR automatically; this is because your branch is not in the `cilium/cilium` repository.

image


Since we do have a new image, I think it should be enough for you to add a new commit to this PR, with the update for `.github/workflows/documentation.yaml`. The new image reference to  use is the following: `docker://quay.io/cilium/docs-builder:d5505e230c0164b51855b002c2a0abeeff78870c@sha256:95940ef277ae14c745769e332c968d8357a43e363eab0b088c5d5efa22089260`.

qmonnet (changes requested):
Requesting changes, as per my comment above.

Documentation/.readthedocs.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+# Build documentation with Sphinx
+sphinx:
+  configuration: conf.py
+
+# We recommend specifying your dependencies to enable reproducible builds:
+# https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
+python:
+  install:
+  - requirements: requirements.txt
```
aanm: This needs to be `Documentation/requirements.txt`

aanm (changes requested):

Documentation/.readthedocs.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
+# Build documentation with Sphinx
+sphinx:
+  configuration: Documentation/conf.py
```
doniacld, line 16: @aanm Should it be also here the full path?

qmonnet: I think so, and one more question: Isn't this file (`.readthedocs.yaml`) supposed to be at the root of the repo?

qmonnet: https://docs.readthedocs.io/en/stable/config-file/v2.html

> This file is named .readthedocs.yaml and should be placed in the top level of your Git repository.

doniacld: > This file is named .readthedocs.yaml and should be placed in the top level of your Git repository.

Yes!

doniacld: Making more sense now.

qmonnet: Apologies, I was wrong on this one. Logging to RTD interface, I see this is configurable:

image

and it's under `Documentation/`. @aanm did you just configure that? Do we have a preferred location for this file?

aanm: Yes, I configured that to be the same path as this PR so we can test this changes. The previous changes failed the bailed that's why I suggested to change the `requirements:` field

.readthedocs.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
+# Build documentation with Sphinx
+sphinx:
+  configuration: Documentation/conf.py
```
qmonnet: My understanding is that it will use the "standalone HTML" builder by default, when we've been using the HTMLDir

image

```suggestion
  configuration: Documentation/conf.py
  builder: "dirhtml"
```

About this builder:


> This is a subclass of the standard HTML builder. Its output is a directory with HTML files, where each file is called index.html and placed in a subdirectory named like its page name. For example, the document markup/rest.rst will not result in an output file markup/rest.html, but markup/rest/index.html. When generating links between pages, the index.html is omitted, so that the URL would look like markup/rest/.

(https://www.sphinx-doc.org/en/master/usage/builders/index.html#sphinx.builders.html.DirectoryHTMLBuilder)

Documentation/.readthedocs.yaml
doniacld: I just changed it to be at the root of the repository. Should we keep it like this? Or should we move it back to `Documentation`?

qmonnet: Let's keep it like this, I'll update the path on RTD. Thanks!

.readthedocs.yaml
doniacld: Should we also add the following field?
```
  # Fail on all warnings to avoid broken references
  # fail_on_warning: true
 ```

Documentation/.readthedocs.yaml
qmonnet: Can we add an entry into the `CODEOWNERS` file for the new `.readthedocs.yaml`, please? `cilium/docs-structure` should likely be the owner.

doniacld: Sure! It's done.

qmonnet (approved):
Looks good to me, thank you!

tklauser (approved):
LGTM for @cilium/contributing.

nbusseneau (approved):

aanm (approved):

Issue #27080, opened 2023-07-26
CI: Integration tests - github.com/cilium/cilium/operator/pkg/lbipam: `FAIL: TestRemoveRequestedIP`


qmonnet:
## CI failure

Integration tests (Travis):

```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃   FAIL  package: github.com/cilium/cilium/operator/pkg/lbipam   ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛

--- FAIL: TestRemoveRequestedIP (1.10s)

level=info msg=Invoked duration="125.242µs" function="lbipam.mkTestFixture.func7 (lbipam_fixture_test.go:149)" subsys=hive
level=info msg=Invoked duration="17.19µs" function="lbipam.glob..func1 (cell.go:25)" subsys=hive
level=info msg=Starting subsys=hive
level=info msg="Start hook executed" duration="1.206µs" function="*resource.resource[*github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2alpha1.CiliumLoadBalancerIPPool].Start" subsys=hive
level=info msg="Start hook executed" duration=623ns function="*resource.resource[*github.com/cilium/cilium/pkg/k8s/slim/k8s/api/core/v1.Service].Start" subsys=hive
level=info msg="Start hook executed" duration="4.086µs" function="*job.group.Start" subsys=hive
level=info msg="LB-IPAM initializing" subsys=lbipam
level=info msg="LB-IPAM done initializing" subsys=lbipam
    lbipam_test.go:2236: Expected service status update
level=info msg=Stopping subsys=hive
level=info msg="Stop hook executed" duration="38.169µs" function="*job.group.Stop" subsys=hive
level=info msg="Stop hook executed" duration="37.538µs" function="*resource.resource[*github.com/cilium/cilium/pkg/k8s/slim/k8s/api/core/v1.Service].Stop" subsys=hive
level=info msg="Stop hook executed" duration="12.537µs" function="*resource.resource[*github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2alpha1.CiliumLoadBalancerIPPool].Stop" subsys=hive
```

[...]

PR #27541, opened 2023-08-16
bpf: make it easier to figure out which BUILD_PERMUTATION failed

lmb:
The output from "make build_all" is currently super hard to read, since it just spits out a gigantic wall of text. It's not at all clear which permutation failed to build.

Refactor the Makefile to execute the build permutations as individual commands. This uses two somewhat esoteric make features:

* Double colon rules [1]: allows us to "append" rules to a target.
* Conditional eval: allows us to emit rules based on the contents of a variable.

Since double colon rules without prerequisites are always executed we don't need the trick with the fake build dependency to trigger rebuilds.

1: https://www.gnu.org/software/make/manual/html_node/Double_002dColon.html

### DIFF
--- a/bpf/Makefile
+++ b/bpf/Makefile
@@ -192,7 +192,6 @@ all: bpf_all
 bpf_all: $(BPF) subdirs
 
 build_all: force
-	@touch $(BUILD_PERMUTATIONS_DEP)
 	@$(ECHO_CHECK)/*.c BUILD_PERMUTATIONS=1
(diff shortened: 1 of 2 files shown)


lmb:
Side note: seems like `-j N` also works correctly.

dylandreimerink (approved):

lmb:
Travis failure is https://github.com/cilium/cilium/issues/27080

PR #27542, opened 2023-08-16
add lint-go to merge queue check

aanm:
To prevent breaking the main branch after merging two PRs that passed CI but conflicted which each other, we should enable merge queue which should prevent this from ever happening.

More info: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue


nbusseneau (approved):
Ooooh since we can enable it per workflow, then let's try it!

PR #27553, opened 2023-08-17
Refactor hubble redact settings schema [v2]

ChrsMark:
Implementing the refactoring again after https://github.com/cilium/cilium/pull/26989 was merged+reverted (see https://github.com/cilium/cilium/pull/26989#issuecomment-1670013860). In this one `HubbleRedact` is renamed to `HubbleRedactEnabled` as well to avoid the failures in connectivity tests with cilium-cli (https://github.com/cilium/cilium-cli/issues/1896)

This PR implements the refactoring of the `--hubble-redact` (renamed to `hubble-redact-enabled`) option in order to provide more flexibility as per https://github.com/cilium/cilium/issues/23887#issuecomment-1633015416.

Specifically we change 

```yml
hubble:
 redact:
  - http-url-query
  - kafka-api-key
```

to 

```yml
hubble:
  redact:
    enabled: true
    http:
      urlQuery: true
    kafka:
      apiKey: true
```

and 

```console
--hubble-redact=http-url-query,kafka-api-key
```

to

```console
--hubble-redact-enabled=true
--hubble-redact-http-url-query=true
--hubble-redact-kafka-api-key=true
```

With this change we aim to support the option for user provided http-headers list.

This will look like the following (as per https://github.com/cilium/cilium/issues/23887#issuecomment-1632167762):

```yml
hubble:
  redact:
    enabled: false
    http:
      userInfo: true
      urlQuery: true
      headers:
        allow:
          - traceparent
          - tracestate
          - Cache-Control
          - ...
    kafka:
      apiKey: true
```

cc: @ioandr 

### How to test this PR

1. Example input values:
```yml
hubble:
  redact:
    enabled: true
    kafka:
      apiKey: true
```

2. Expected logs:

```console
level=info msg="  --hubble-redact-enabled='true'" subsys=daemon
level=info msg="  --hubble-redact-http-urlquery='false'" subsys=daemon
level=info msg="  --hubble-redact-kafka-apikey='true'" subsys=daemon
level=info msg="configured Hubble with redact options" options="&{CacheSize:10000 RedactHTTPQuery:false RedactKafkaAPIKey:true}" subsys=hubble
```

In addition the validation of at least `http` redact should work as expected.

### DIFF
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -994,8 +994,14 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	)
 	option.BindEnv(vp, option.HubbleMonitorEvents)
 
-	flags.StringSlice(option.HubbleRedact, []string{}, "List of Hubble redact options")
-	option.BindEnv(vp, option.HubbleRedact)
+	flags.Bool(option.HubbleRedactEnabled, defaults.HubbleRedactEnabled, "Hubble redact sensitive information from flows")
+	option.BindEnv(vp, option.HubbleRedactEnabled)
+
+	flags.Bool(option.HubbleRedactHttpURLQuery, defaults.HubbleRedactHttpURLQuery, "Hubble redact http URL query from flows")
+	option.BindEnv(vp, option.HubbleRedactHttpURLQuery)
+
+	flags.Bool(option.HubbleRedactKafkaApiKey, defaults.HubbleRedactKafkaApiKey, "Hubble redact Kafka API key from flows")
+	option.BindEnv(vp, option.HubbleRedactKafkaApiKey)
 
 	flags.StringSlice(option.DisableIptablesFeederRules, []string{}, "Chains to ignore when installing feeder rules.")
 	option.BindEnv(vp, option.DisableIptablesFeederRules)
--- a/daemon/cmd/hubble.go
+++ b/daemon/cmd/hubble.go
@@ -139,8 +139,15 @@ func (d *Daemon) launchHubble() {
 		)
 	}
 
-	if len(option.Config.HubbleRedact) > 0 {
-		parserOpts = append(parserOpts, parserOptions.Redact(logger, option.Config.HubbleRedact))
+	if option.Config.HubbleRedactEnabled {
+		parserOpts = append(
+			parserOpts,
+			parserOptions.Redact(
+				logger,
+				option.Config.HubbleRedactHttpURLQuery,
+				option.Config.HubbleRedactKafkaApiKey,
+			),
+		)
 	}
 
 	d.linkCache = link.NewLinkCache()
--- a/install/kubernetes/cilium/values.yaml.tmpl
+++ b/install/kubernetes/cilium/values.yaml.tmpl
@@ -1034,18 +1034,28 @@ hubble:
   # -- Unix domain socket path to listen to when Hubble is enabled.
   socketPath: /var/run/cilium/hubble.sock
 
-  # -- Configures the list of redact options for Hubble.
+  # -- Configures the redact options for Hubble.
   # Example:
   #
   #   redact:
-  #   - http-url-query
-  #   - kafka-api-key
+  #     enabled: true
+  #     http:
+  #       urlQuery: true
+  #     kafka:
+  #       apiKey: false
   #
-  # You can specify the list of options from the helm CLI:
+  # You can specify the options from the helm CLI:
   #
-  #   --set hubble.redact="{http-url-query,kafka-api-key}"
+  #   --set hubble.redact.enabled="true"
+  #   --set hubble.redact.http.urlQuery="true"
+  #   --set hubble.redact.kafka.apiKey="false"
   #
-  redact: ~
+  redact:
+    enabled: false
+    http:
+        urlQuery: false
+    kafka:
+        apiKey: false
 
   # -- An additional address for Hubble to listen to.
   # Set this field ":4244" if you are enabling Hubble Relay, as it assumes that
--- a/pkg/hubble/parser/seven/http.go
+++ b/pkg/hubble/parser/seven/http.go
@@ -66,7 +66,12 @@ func decodeHTTP(flowType accesslog.FlowType, http *accesslog.LogRecordHTTP, opts
 }
 
 func (p *Parser) httpSummary(flowType accesslog.FlowType, http *accesslog.LogRecordHTTP, flow *flowpb.Flow) string {
-	httpRequest := fmt.Sprintf("%s %s", http.Method, http.URL)
+	uri, _ := url.Parse(http.URL.String())
+	if p.opts.RedactHTTPQuery {
+		uri.RawQuery = ""
+		uri.Fragment = ""
+	}
+	httpRequest := http.Method + " " + uri.String()
 	switch flowType {
 	case accesslog.TypeRequest:
 		return fmt.Sprintf("%s %s", http.Protocol, httpRequest)
--- a/Documentation/cmdref/cilium-agent.md
+++ b/Documentation/cmdref/cilium-agent.md
@@ -193,7 +193,9 @@ cilium-agent [flags]
       --hubble-prefer-ipv6                                        Prefer IPv6 addresses for announcing nodes when both address types are available.
       --hubble-recorder-sink-queue-size int                       Queue size of each Hubble recorder sink (default 1024)
       --hubble-recorder-storage-path string                       Directory in which pcap files created via the Hubble Recorder API are stored (default "/var/run/cilium/pcaps")
-      --hubble-redact strings                                     List of Hubble redact options
+      --hubble-redact-enabled                                     Hubble redact sensitive information from flows
+      --hubble-redact-http-urlquery                               Hubble redact http URL query from flows
+      --hubble-redact-kafka-apikey                                Hubble redact Kafka API key from flows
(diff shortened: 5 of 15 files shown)


ChrsMark:
Something that I have spotted is that we miss to redact the sensitive information from the `Summary` as well. I will tune https://github.com/cilium/cilium/blob/fe07332b5ca73e25032b622276f430d789a56d8a/pkg/hubble/parser/seven/http.go#L69 accordingly.

An example:

```json
{
  "flow": {
    "time": "2023-08-17T15:28:45.940672248Z",
    "uuid": "700f5a3a-263c-4cf0-a059-f6fa2368164c",
    "verdict": "FORWARDED",
    "IP": {
      "source": "10.244.1.75",
      "destination": "10.244.1.29",
      "ipVersion": "IPv4"
    },
    "l4": {
      "TCP": {
        "source_port": 80,
        "destination_port": 35788
      }
    },
    "source": {
      "ID": 257,
      "identity": 56737,
      "namespace": "default",
      "labels": [
        "k8s:app.kubernetes.io/name=deathstar",
        "k8s:class=deathstar",
        "k8s:io.cilium.k8s.policy.cluster=kind-kind",
        "k8s:io.kubernetes.pod.namespace=default",
        "k8s:org=empire"
      ],
      "pod_name": "deathstar-7848d6c4d5-7dwtj",
      "workloads": [
        {
          "name": "deathstar",
          "kind": "Deployment"
        }
      ]
    },
    "destination": {
      "ID": 1973,
      "identity": 27221,
      "namespace": "default",
      "labels": [
        "k8s:app.kubernetes.io/name=tiefighter",
        "k8s:class=tiefighter",
        "k8s:io.cilium.k8s.policy.cluster=kind-kind",
        "k8s:io.kubernetes.pod.namespace=default",
        "k8s:org=empire"
      ],
      "pod_name": "tiefighter"
    },
    "Type": "L7",
    "node_name": "kind-kind/kind-worker",
    "l7": {
      "type": "RESPONSE",
      "latency_ns": "251306",
      "http": {
        "code": 403,
        "method": "POST",
        "url": "http://deathstar.default.svc.cluster.local/v1/request-landing",
        "protocol": "HTTP/1.1",
        "headers": [
          {
            "key": "Content-Length",
            "value": "15"
          },
          {
            "key": "Content-Type",
            "value": "text/plain"
          },
          {
            "key": "X-Request-Id",
            "value": "f9071907-7fdd-4060-9153-37e436941944"
          }
        ]
      }
    },
    "reply": true,
    "event_type": {
      "type": 129
    },
    "is_reply": true,
    "Summary": "HTTP/1.1 403 0ms (POST http://deathstar.default.svc.cluster.local/v1/request-landing?foo=bar)"
  },
  "node_name": "kind-kind/kind-worker",
  "time": "2023-08-17T15:28:45.940672248Z"
}
```

ChrsMark:
[4a51fdb](https://github.com/cilium/cilium/pull/27553/commits/4a51fdb90b26607a79f2295c1781560f85235e97) fixes https://github.com/cilium/cilium/pull/27553#issuecomment-1682502130.

pkg/hubble/parser/seven/http.go
```diff
@@ -66,7 +66,12 @@ func decodeHTTP(flowType accesslog.FlowType, http *accesslog.LogRecordHTTP, opts
 }
 
 func (p *Parser) httpSummary(flowType accesslog.FlowType, http *accesslog.LogRecordHTTP, flow *flowpb.Flow) string {
-	httpRequest := fmt.Sprintf("%s %s", http.Method, http.URL)
+	uri, _ := url.Parse(http.URL.String())
+	if p.opts.RedactHTTPQuery {
+		uri.RawQuery = ""
+		uri.Fragment = ""
+	}
+	httpRequest := fmt.Sprintf("%s %s", http.Method, uri)
```
christarazi: Let's avoid `fmt.Sprintf` in hot paths. This can be a simple string concat instead.

Ref: #19571

christarazi (changes requested):

pkg/hubble/parser/seven/http.go
ChrsMark: Fixed: [4956bd1](https://github.com/cilium/cilium/pull/27553/commits/4956bd17e1b510bd8cb18e6bb43cdf218fbd5bb9)

christarazi: Thanks. Could you squash the commits together?

ChrsMark: Sure! Done :)

christarazi (approved):

ChrsMark:
Hey folks, any chances of moving the review of this one forward ? It would be nice to have this merged soon, so as to unblock the rest of the pending items from https://github.com/cilium/cilium/issues/23887 :).

qmonnet (changes requested):
Looks mostly good from my side. One thing: the renaming of existing agent and Helm options present in released versions may have upgrade impact, so we'll need to document it in the upgrade notes for 1.15 (please add a note in `Documentation/operations/upgrade.rst`).

qmonnet (approved):
OK my bad, I checked whether the initial commit was in 1.14 but didn't realise we had backported the revert after that. All good then, sorry :)

thorn3r (approved):
agent changes look good

daemon/cmd/daemon_main.go
```diff
@@ -994,8 +994,14 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	)
 	option.BindEnv(vp, option.HubbleMonitorEvents)
 
-	flags.StringSlice(option.HubbleRedact, []string{}, "List of Hubble redact options")
-	option.BindEnv(vp, option.HubbleRedact)
+	flags.Bool(option.HubbleRedactEnabled, defaults.HubbleRedactEnabled, "Hubble redact sensitive information from flows")
```
tklauser, line 997: Do we really need a separate CLI option for this? Couldn't we just assume this to be enabled whether either of the following two options (http RUL query and/or kafka API key redaction) is enabled, i.e. replace the single check for `option.Config.HubbleRedactEnabled` by `option.Config.HubbleRedactHttpURLQuery || option.Config.HubbleRedactKafkaApiKey`?

daemon/cmd/hubble.go
```diff
@@ -139,8 +139,15 @@ func (d *Daemon) launchHubble() {
 		)
 	}
 
-	if len(option.Config.HubbleRedact) > 0 {
-		parserOpts = append(parserOpts, parserOptions.Redact(logger, option.Config.HubbleRedact))
+	if option.Config.HubbleRedactEnabled {
```
tklauser, line 142: Per the comment above this would become:

```suggestion
	if option.Config.HubbleRedactHttpURLQuery || option.Config.HubbleRedactKafkaApiKey {
```

or do you see a case where a user would not set `--hubble-redact-enabled` but not one of the other options?

daemon/cmd/daemon_main.go
ChrsMark: That's based on "this approach would provide the best flexibility for the user yet with reasonable default, all the user has to do is to set `hubble.redact.enabled=true`" from https://github.com/cilium/cilium/issues/23887#issuecomment-1632167762.

daemon/cmd/hubble.go
ChrsMark: [The idea](https://github.com/cilium/cilium/issues/23887#issuecomment-1632167762.) is that users can probably just set `--hubble-redact-enabled` and then rely on the defaults. Those that want more "granularity" can define the specific http, kafka, etc settings as well.

Would that make sense?

tklauser: Thanks for the context, that makes sense. Sorry I overlooked it in the PR description.

daemon/cmd/daemon_main.go
tklauser: Makes sense, thanks.

tklauser (approved):

rolinh (commented):
Thanks @ChrsMark ! Looks good to me, I just have a minor suggestion.

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1034,18 +1034,28 @@ hubble:
...
+  #   --set hubble.redact.http.urlQuery="true"
+  #   --set hubble.redact.kafka.apiKey="false"
   #
-  redact: ~
+  redact:
+    enabled: false
+    http:
+        urlQuery: false
+    kafka:
+        apiKey: false
```
rolinh, line 1058: I think these should be `true` by default so that when a user sets `redact.enabled=true` it's all they need to do unless they want to fine tune the behavior.

ChrsMark: Yeap eventually we would like to have this behavior. I remember this was proposed at https://github.com/cilium/cilium/issues/23887#issuecomment-1614840591 to be fixed in a standalone PR. So how about dealing with those in a follow-up PR as proposed at https://github.com/cilium/cilium/issues/23887#issuecomment-1614840591? 

(we would need to maybe to discuss separately what the reasonable defaults would be?) 

Otherwise I can definitely apply the changes here :).

joamaki (approved):

install/kubernetes/cilium/values.yaml.tmpl
rolinh: Works for me :)

rolinh (approved):

PR #27561, opened 2023-08-17
cgroups: Fix race to load cgroup.hostRoot option

kvaps:
While loading, the package `pkg/cgroups/manager` sets the global variable `cgroupRoot` from the `pkg/cgroups` package.
This occurs before the configuration is fully loaded and the correct path is set in the `pkg/cgroups` package.

This variable is later used by the `validateCgroupPath` function.
As a result, the `validateCgroupPath` function always works with the default value instead of the user-defined one.

While using `--set cgroup.autoMount.enabled=false` and `--set cgroup.hostRoot=/sys/fs/cgroup` options, this might lead strange errors like:

```
level=info msg="  --cgroup-root='/sys/fs/cgroup'" subsys=daemon
level=info msg="Mounted cgroupv2 filesystem at /sys/fs/cgroup" subsys=cgroups
level=warning msg="No valid cgroup base path found: socket load-balancing tracing with Hubble will not work.See the kubeproxy-free guide for more details." subsys=cgroup-manager
```

And path `/sys/fs/cgroup/kubepods` correctly existing.

```release-note
cgroups: Fix race to load cgroup.hostRoot option
```

### DIFF
--- a/pkg/cgroups/manager/provider.go
+++ b/pkg/cgroups/manager/provider.go
@@ -14,7 +14,6 @@ import (
 )
 
 var (
-	cgroupRoot = cgroups.GetCgroupRoot()
 	// example default cgroup path in kubernetes environments
 	// /kubepods/burstable/pod1858680e-b044-4fd5-9dd4-f137e30e2180/e275d1a37782ab30008aa3ae6666cccefe53b3a14a2ab5a8dc459939107c8c0
 	defaultCgroupBasePath = "/kubepods"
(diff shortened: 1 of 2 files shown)


aspsk:
Thanks @kvaps, the change looks great! However, it breaks the `pkg/cgroups/manager/manager_test.go` and the `pkg/cgroups/manager/provider_test.go` tests, as they use the variable you've removed. Could you please patch the tests as well?

kvaps:
Yeah, sorry, added it to `provider_test.go`

aspsk:
@kvaps thanks. Could you also please rebase on top of `main`?

kvaps:
Sure, done

aspsk (approved):
Thanks!

PR #27567, opened 2023-08-17
CI: Add conn-disrupt-test action for reuse

jschwinger233:
This commit makes conn-disrupt-test a github action, so upgrade test and IPsec key rotation test don't have to copy and paste everywhere.

The idea is to allow caller workflow to specify the commands to execute, then this action will follow the steps:
1. Run "cilium-cli connectivity test --conn-disrupt-test-setup";
2. Run whatever caller workflow passes: could be upgrade operation or IPsec key rotation;
3. Run "cilium-cli connectivity test --include-conn-disrupt-test";

### DIFF
--- a/.github/actions/conn-disrupt-test/action.yaml
+++ b/.github/actions/conn-disrupt-test/action.yaml
@@ -1,0 +1,49 @@
+name: Cilium Conn Disrupt Test
+description: Setup and test cilium connectivity and connection disruption status after a caller-provided risky operation, such as upgrade or IPsec key rotation.
+
+inputs:
+  job-name:
+    required: true
+    type: string
+  operation-cmd:
+    required: true
+    type: string
+  extra-connectivity-test-flags:
+    required: false
+    type: string
+
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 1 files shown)


bimmlerd (approved):

joestringer:
@jschwinger233 @brb it looks like @brb was expecting to give a review, but you've set `ready-to-merge`. I have two concerns here: One, the baseline expectation to merge PRs is run CI plus codeowner reviews are in. CI is missing, though I understand this is a new github workflow. At the same time, the new CI should properly skip all unrelated jobs so we should get in the habit of always running CI and not trying to bypass or ignore it. Two, if @brb has plans to review then we should give him the opportunity to review it.

brb (approved):
Thanks!

pchaigno (approved):

pchaigno:
I created https://github.com/cilium/cilium/issues/27647 for the flake affecting ConformanceGatewayAPI. It's clearly unrelated as the present PR only introduces a new GitHub Action (not even workflow) which isn't used anywhere yet. Marking ready to merge.

PR #27568, opened 2023-08-17
bgpv1: Consolidate reconciler-specific maps into generic ReconcilerMetadata

rastislavs:
This PR finishes the efforts of making the BGP reconcilers independent and plugable / separateable by removing reconciler-specific maps from the BGP Manager's `ServerWithConfig` struct. Instead of that, all reconcilers can keep their metadata in a common `ReconcilerMetadata` map, keyed by the reconciler name.

As a side effect, the reconciler interface gets a new `Name()` method that can be used to retrieve reconciler's name.


harsimran-pabla (approved):

rastislavs:
approved & all green, marking as ready to merge

PR #27572, opened 2023-08-17
fqdn: fix broken sorting algorithm for DNSZombies

bimmlerd:
The sorting function for DNSZombies was subtly broken, and didn't do what it advertised. Write some tests to confirm the suspicion and fix the function.

Reported-by: Jussi Maki <joamaki@isovalent.com>

```release-note
Prioritization of which DNS mappings to keep was suboptimal, leading to evictions of mappings related to alive connections, worsening performance of fqdn policies and causing spurious logging.
```

### DIFF
--- a/pkg/fqdn/cache_test.go
+++ b/pkg/fqdn/cache_test.go
@@ -11,6 +11,7 @@ import (
 	"net/netip"
 	"regexp"
 	"sort"
+	"testing"
 	"time"
 
 	. "github.com/cilium/checkmate"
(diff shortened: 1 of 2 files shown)


bimmlerd:
Given this is a bug, I guess we backport it to at least 1.14. I assess the risk of this breaking something as pretty minimal, so might be worth backporting even further, as it's not great to throw away these mappings for live connections :sweat:.

pkg/fqdn/cache_test.go
```diff
@@ -1109,3 +1110,171 @@ func (ds *DNSCacheTestSuite) TestOverlimitPreferNewerEntries(c *C) {
...
+	}
+
+	// Generate zombie mappings which cover all cases of the two times
+	// being either moment 0, 1 or 2, as well as with 0, 1 or 2 names.
+	names := []string{"example.org", "test.com"}
+	nMoments := len(moments)
+	allMappings := make([]*DNSZombieMapping, 0, nMoments*nMoments*nMoments)
+	for _, mi := range moments {
+		for _, mj := range moments {
+			for k := range moments {
```
giorio94: Nit. This `for` should iterate on the number of `names`, and not on the `moments` array (which works only because the length matches).

pkg/fqdn/cache_test.go
```diff
@@ -1109,3 +1110,171 @@ func (ds *DNSCacheTestSuite) TestOverlimitPreferNewerEntries(c *C) {
...
+		})
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ol := len(tt.args.zombies)
+			sortZombieMappingSlice(tt.args.zombies)
+			if len(tt.args.zombies) != ol {
+				t.Fatalf("length of slice changed by sorting")
+			}
```
giorio94, line 1276: Nit. To be super picky, one could also check that the elements match, in addition to the slice length (e.g., through `assert.ElementsMatch`). Feel free to ignore this comment though, as with the current implementation it would test that the go sort method works.

giorio94 (approved):
/lgtm, nice catch!

jrajahalme (approved):
Thanks for the bug fix! Could fix the nit by @giorio94 in the test.

pkg/fqdn/cache_test.go
bimmlerd: True, but yeah as you said we'd just be testing Go's sort - I'll leave it.

joamaki:
When over `tofqdns-endpoint-max-ip-per-hostname` IPs exist for a name, then this sorting kicks in and tries to evict the oldest entries. Due to the sorting being broken, it may have evicted alive IPs instead leading to connection drops. This was made worse by the conntrack GC interval becoming too long, possibly many hours (this is what marks the aliveness for these entries), leading to this being the only mechanism via which these entries are evicted. 

https://github.com/cilium/cilium/pull/27870 adds the option of setting the maximum interval for conntrack GC, which should allow configuring it in a way to allow GCing before the `max-ips-per-hostname` is hit and thus lower the peak amount of allocated identities.

PR #27574, opened 2023-08-17
egressgw: small test fixes

lmb:
bpf, egressgw: gate use of EGRESS_POLICY_MAP on define

    EGRESS_POLICY_MAP is only defined if the egress gateway is enabled. Allow
    including egressgw.h even if that is not the case, so that the helpers can
    be reused.

bpf, egressgw: make it easier to correlate test cases with debug msg

    When an egressgw BPF unit test fails it helpfully dumps any cilium debug 
    messages with the test. The messages often contain port numbers which can be
    used to infer which subtest generated them. Make it easier to go from port
    number to subtest by accounting for endiannness in client_port().

bpf, egressgw: fix tc_egressgw_skip_excluded_cidr_redirect case

    Use the correct test enum in tc_egressgw_skip_excluded_cidr_redirect to
    avoid tests interfering with each other.

bpf, egressgw: always clear excluded cidr at end of snat test

    Always clear the exclude CIDR rule, regardless of whether the test succeeds
    or not. Otherwise subsequent tests will fail due to the presence of the
    exclude rule.

### DIFF
--- a/bpf/tests/lib/egressgw.h
+++ b/bpf/tests/lib/egressgw.h
@@ -28,16 +28,16 @@ enum egressgw_test {
 };
 
 struct egressgw_test_ctx {
-	enum egressgw_test test;
+	__u16 test;
 	enum ct_dir dir;
 	__u64 tx_packets;
 	__u64 rx_packets;
 	__u32 status_code;
 };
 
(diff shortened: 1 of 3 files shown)


aspsk (approved):
LGTM. Also thanks for the great commit descriptions.

PR #27575, opened 2023-08-17
Fix possible cross-cluster connection drops on agents restart when clustermesh is enabled

### LINKED ISSUE #26462: clustermesh: downtime/dropped packets due to network policy on upgrade

### What happened?

We are currently running Cilium on EKS in chaining mode with aws-cni. We are utilizing clustermesh across 6 clusters and the configuration and installation is managed by FluxCD. Our application has an auth service that lives in one of the six clusters and proxy service that lives in all clusters. The proxies connect to each other in a mesh as well as initiate a connection to the auth service. The proxies rely on a global service to connect to auth, they dial each other directly via a different discovery mechanism.

Single cluster upgrades don't appear to be an issue but if multiple clusters upgrade close enough together we observe connection issues between proxies in those clusters. Specifically we are looking at the metric `hubble_drop_total{reason="Policy denied"}` as well as metric on our application that indicates the number of other proxies a proxy is connected to. Looking at the logs of the cilium agent of the node where affected proxies were running it seems like it isn't uncommon to see a connection to etcd, disconnect, and reconnect. It is fairly reproducible by running `kubectl rollout restart deployment clustermesh-apiserver` in cluster1 and `kubectl rollout restart ds cilium` in cluster2.

I considered trying to split the cilium install and the clustermesh install into separate helm installs as that would give us control when they would restart and not happen at the same time. Unfortunately, the `cilium-ca` secret is expected to be created every time and would cause a conflict (there may be other but this was the first roadblock I hit, could potentially get around this by allowing users to specify this).

We are hoping to better understand the following:
1. Expectations of upgrades as the upgrade guide states `Minimal to None` affect on L3/L4 and if we are hitting that or if we are hitting something else.
2. Is it expected that Cilium upgrades in clustermesh happen one cluster at a time? Our FluxCD reconciles every 10 min and currently have no coordination that at the moment but are looking at options for this.
3. Would we be better off configuring our cilium install to be backed by etcd? That seems like it would remove the possibility of etcd connections being broken on cilium startup since there wouldn't be a `clustermesh-apiserver` pod being recreated. Is there documentation for this by chance?
4. Possibly separate the `etcd` container (and give it a PV) from the `apiserver` container so that clustermesh isn't rolled at the same time? Any gotchas with this approach?

Thanks for all the work on this awesome CNI!

### Cilium Version

1.12.3

### Kernel Version

Linux ip-10-16-112-106.us-west-2.compute.internal 5.4.242-155.348.amzn2.x86_64 #1 SMP Mon May 8 12:52:40 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

### Kubernetes Version

Server Version: v1.23.17-eks-0a21954

### Sysdump

Sysdump file was too large to upload

### Relevant log output

```shell
level=info msg="Established connection to remote etcd" clusterName=redact config=/var/lib/cilium/clustermesh/redact kvstoreErr="<nil>" kvstoreStatus="etcd: 1/1 connected, lease-ID=0, lock lease-ID=0, has-quorum=timeout while waiting for initial connection, consecutive-errors=1: https://redact:2379 - 3.5.4 (Leader)" subsys=clustermesh
```

### Anything else?

_No response_

### Code of Conduct

giorio94:
This PR modifies the endpoint regeneration logic to additional wait for remote clusters ipcache synchronization before performing the datapath regeneration, to make sure that the new ipcache map starts to be used only when it has been completely populated. 

A user-configurable timeout (defaulting to one minute) limits the maximum wait interval, to prevent blocking the local endpoints regeneration process for an extended period of time if a remote cluster is not ready (which could also lead to checken-and-egg dependencies). A special case is IPSec enabled, as this currently requires a synchronous regeneration of the host endpoint: here, we limit the timeout to at most 5 seconds, to not block the agent startup process.

In addition, when WireGuard is enabled, the deletion of obsolete peers is now performed after listing all nodes from remote clusters, again to prevent possible connectivity drops on agent restart.

Please refer to the individual commit messages for additional details.

Tested with https://github.com/cilium/cilium/pull/27232: https://github.com/cilium/cilium/actions/runs/5949436519
The downgrade test was based on the current tip of v1.14 with these changes on top; hence, I'm marking this PR for backport to v1.14 (and backport/author given that I've already the PR for that). 

Fixes: https://github.com/cilium/cilium/issues/26462

```release-note
Fix possible cross-cluster connection drops on agents restart when clustermesh is enabled
```

### DIFF
--- a/pkg/clustermesh/clustermesh.go
+++ b/pkg/clustermesh/clustermesh.go
@@ -157,6 +157,7 @@ func (cm *ClusterMesh) NewRemoteCluster(name string, status common.StatusFunc) c
 		name,
 		cm.conf.NodeKeyCreator,
 		cm.conf.NodeObserver,
+		store.RWSWithOnSyncCallback(func(ctx context.Context) { close(rc.synced.nodes) }),
 		store.RWSWithEntriesMetric(cm.conf.Metrics.TotalNodes.WithLabelValues(cm.conf.ClusterName, cm.nodeName, rc.name)),
 	)
 
@@ -185,6 +186,12 @@ func (cm *ClusterMesh) NumReadyClusters() int {
 // SyncedWaitFn is the type of a function to wait for the initial synchronization
 // of a given resource type from all remote clusters.
 type SyncedWaitFn func(ctx context.Context) error
+
+// NodesSynced returns after that the initial list of nodes has been received
+// from all remote clusters, and synchronized with the different subscribers.
+func (cm *ClusterMesh) NodesSynced(ctx context.Context) error {
+	return cm.synced(ctx, func(rc *remoteCluster) SyncedWaitFn { return rc.synced.Nodes })
+}
 
 // ServicesSynced returns after that the initial list of shared services has been
 // received from all remote clusters, and synchronized with the BPF datapath.
--- a/pkg/clustermesh/clustermesh.go
+++ b/pkg/clustermesh/clustermesh.go
@@ -5,6 +5,7 @@ package clustermesh
 
 import (
 	"context"
+	"errors"
 
 	"github.com/cilium/cilium/api/v1/models"
 	"github.com/cilium/cilium/pkg/allocator"
@@ -17,7 +18,6 @@ import (
 	"github.com/cilium/cilium/pkg/k8s"
 	"github.com/cilium/cilium/pkg/kvstore"
 	"github.com/cilium/cilium/pkg/kvstore/store"
-	"github.com/cilium/cilium/pkg/lock"
 	"github.com/cilium/cilium/pkg/logging"
 	"github.com/cilium/cilium/pkg/logging/logfields"
 	nodeTypes "github.com/cilium/cilium/pkg/node/types"
@@ -150,8 +150,8 @@ func (cm *ClusterMesh) NewRemoteCluster(name string, status common.StatusFunc) c
 		mesh:         cm,
 		usedIDs:      cm.conf.ClusterIDsManager,
 		status:       status,
-		swg:          lock.NewStoppableWaitGroup(),
 		storeFactory: cm.conf.StoreFactory,
+		synced:       newSynced(),
 	}
 	rc.remoteNodes = cm.conf.StoreFactory.NewWatchStore(
 		name,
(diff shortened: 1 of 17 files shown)


giorio94:
Converting back to draft, as the remaining part of the fix for https://github.com/cilium/cilium/issues/26462 (when endpoints are selected by netpols) should be simple enough to be included here.

ti-mo (approved):

giorio94:
Checks are green. Marking again ready for review.

thorn3r (approved):
excellent work and beautiful commit structure/messages :heart_eyes:

pkg/endpoint/regenerator.go
```diff
@@ -0,0 +1,110 @@
...
+	wctx, cancel := context.WithTimeout(ctx, r.ClusterMeshIPIdentitiesSyncTimeout)
+	defer cancel()
+	err := r.cmWaitFn(wctx)
+
+	switch {
+	case ctx.Err() != nil:
+		// The context associated with the endpoint has been canceled.
+		return ErrNotAlive
+	case err != nil:
+		// We don't return an error in case the wait operation timeout out, as we can
```
tklauser: nit (only worth addressing in case there are other change requests):

```suggestion
		// We don't return an error in case the wait operation timed out, as we can
```

tklauser (approved):

giorio94:
Rebased onto main to pick CI changes, as workflow names changes.

jrajahalme (approved):

giorio94:
Rebased onto main to fix conflicts.

pkg/clustermesh/clustermesh.go
```diff
@@ -179,21 +179,31 @@ func (cm *ClusterMesh) NumReadyClusters() int {
...
-	for _, swg := range swgs {
-		select {
-		case <-swg.WaitChannel():
-		case <-ctx.Done():
-			return ctx.Err()
+	for _, wait := range waiters {
+		err := wait(ctx)
+
+		// Ignore the error in case the given cluster was disconnected in
+		// the meanwhile, as we do not longer care about it.
```
brb, line 221: Nit: maybe an info log would be helpful for future investigations to answer "why I don't see services from cluster X"?

brb (approved):
Thanks! Small nit. Keen to see the upcoming CM upgrade tests.

pkg/clustermesh/clustermesh.go
giorio94: I didn't add it because we already have a similar log output on cluster disconnection (which in this context refers to a user manually removing the cluster configuration, not the disconnection from etcd): https://github.com/cilium/cilium/blob/8e55b7ec3d6e9d1eb692c619c71e42ec5412d33a/pkg/clustermesh/common/remote_cluster.go#L392-L397

giorio94:
Rebased onto main to pick CI changes

michi-covalent:
temporarily moving this back to `needs/backport/1.14` to make the release script happy.

PR #27576, opened 2023-08-17
ipcache: Deprecate old API

joestringer:
The following commits have introduced a new API for interacting with the
ipcache, which is now preferred for all future interactions:

- Commit e63781492e1d ("ipcache: Introduce new asynchronous API")
- Commit b6cef9292b94 ("ipcache: Add new {Upsert,Remove}Prefixes() APIs")
- Commit f9173b7191fe ("ipcache: Add ability to override identity via UpsertIdentity")

Mark the older APIs as deprecated in order to help to slowly phase out
future usage of those APIs. For more details, see GitHub issue #21142.

Note that there are still various paths using the older APIs to insert
ipcache entries. If an entry is inserted using the older API, then the
corresponding delete event should likewise use the same older API to
delete the entry. Future users should both insert and delete using the
newer API. Do not mix and match the upserts and deletes from deprecated
APIs and non-deprecated APIs for the same event type / resource update
into ipcache.

PR https://github.com/cilium/cilium/pull/27327 recently fixed bugs in
the case of fixed APIs and added unit tests for those cases, but still
it's better for all future users to focus on using the newer APIs for
better internal feature compatibility rather than mixing and matching.

Suggested-by: Aditi Ghag <aditi@cilium.io>
Signed-off-by: Joe Stringer <joe@cilium.io>

### DIFF
--- a/pkg/ipcache/cidr.go
+++ b/pkg/ipcache/cidr.go
@@ -34,6 +34,8 @@ import (
 //
 // Upon success, the caller must also arrange for the resulting identities to
 // be released via a subsequent call to ReleaseCIDRIdentitiesByCIDR().
+//
+// Deprecated: Prefer UpsertLabels() instead.
 func (ipc *IPCache) AllocateCIDRs(
 	prefixes []netip.Prefix, oldNIDs []identity.NumericIdentity, newlyAllocatedIdentities map[netip.Prefix]*identity.Identity,
 ) ([]*identity.Identity, error) {
@@ -96,6 +98,8 @@ func (ipc *IPCache) AllocateCIDRs(
 //
 // Upon success, the caller must also arrange for the resulting identities to
 // be released via a subsequent call to ReleaseCIDRIdentitiesByID().
+//
+// Deprecated: Prefer UpsertLabels() instead.
 func (ipc *IPCache) AllocateCIDRsForIPs(
 	prefixes []net.IP, newlyAllocatedIdentities map[netip.Prefix]*identity.Identity,
 ) ([]*identity.Identity, error) {
@@ -128,6 +132,8 @@ func cidrLabelToPrefix(id *identity.Identity) (prefix netip.Prefix, ok bool) {
 // that were not already upserted. If any 'usedIdentities' are upserted, these
 // are counted separately as they may provide an indication of another logic
 // error elsewhere in the codebase that is causing premature ipcache deletions.
+//
+// Deprecated: Prefer UpsertLabels() instead.
 func (ipc *IPCache) UpsertGeneratedIdentities(newlyAllocatedIdentities map[netip.Prefix]*identity.Identity, usedIdentities []*identity.Identity) {
 	for prefix, id := range newlyAllocatedIdentities {
 		ipc.Upsert(prefix.String(), nil, 0, nil, Identity{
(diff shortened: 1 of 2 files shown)


pkg/ipcache/cidr.go
```diff
@@ -34,6 +34,8 @@ import (
 //
 // Upon success, the caller must also arrange for the resulting identities to
 // be released via a subsequent call to ReleaseCIDRIdentitiesByCIDR().
+//
+// DEPRECATED. Prefer UpsertLabels() or OverrideIdentity() instead.
```
christarazi: How come we are recommending `OverrideIdentity()`? I understand that it's part of the new API, but I think we should be careful of any usages of `OverrideIdentity` because in a way, it's similar to the old API's semantics and delegation to each caller.

I think I'd prefer that we leave it unmentioned or at least document `OverrideIdentity` that it should only be used if absolutely necessary.

pkg/ipcache/cidr.go
```diff
@@ -128,6 +132,8 @@ func cidrLabelToPrefix(id *identity.Identity) (prefix netip.Prefix, ok bool) {
 // that were not already upserted. If any 'usedIdentities' are upserted, these
 // are counted separately as they may provide an indication of another logic
 // error elsewhere in the codebase that is causing premature ipcache deletions.
+//
+// DEPRECATED. Prefer UpsertLabels() or OverrideIdentity() instead.
 func (ipc *IPCache) UpsertGeneratedIdentities(newlyAllocatedIdentities map[netip.Prefix]*identity.Identity, usedIdentities []*identity.Identity) {
```
christarazi, line 137: I guess the thing to note here is that the incremental policy update path (FQDN) still requires this function until we implement an equivalent mechanism, but that's tracked by #21142.

joestringer: Oh yeah, this is deprecation, not deletion. There's still a whole bunch of users of these existing APIs. The goal here is to (a) discourage using older APIs for new code and (b) encourage discussion around how to remove usage of the older APIs via #21142.

pkg/ipcache/cidr.go
joestringer: Good point yeah, I'll remove this and add a note to `OverrideIdentity` to explain that using that API could cause feature conflicts with any users of UpsertLabels, UpsertMetadata, etc., so should be avoided wherever possible if any other subsystems may be using the same IPs.

pkg/ipcache/ipcache.go
```diff
@@ -254,6 +254,11 @@ func (ipc *IPCache) getK8sMetadata(ip string) *K8sMetadata {
 // given IP. It is optional (may be nil) and is propagated to the listeners.
 // k8sMeta contains Kubernetes-specific metadata such as pod namespace and pod
 // name belonging to the IP (may be nil).
+//
+// When deleting ipcache entries that were previously inserted via this
+// function, ensure that the corresponding delete occurs via Delete().
+//
+// DEPRECATED. Prefer UpsertLabels() or OverrideIdentity() instead.
```
aditighag: I think this should be: `Deprecated` per Go documentation.

aditighag (commented):
Thanks for following up on this.

>Future users should both insert and delete using the
newer API. Do not mix and match the upserts and deletes from deprecated
APIs and non-deprecated APIs for the same event type / resource update
into ipcache.

Sorry, I lost track of the previous PR where we had a brief discussion around testing. 😓 But this mix and match can be easily unit tested such that future usages can be caught via tests? At a high level, just an example - Calling `ReleaseCIDRIdentitiesByCIDR` and `UpsertLabels` and so on...  If you've already added such unit tests, please ignore my comment (although it doesn't hurt to link them in the PR description. :)). 

FWIW, there seems to be a Go convention to deprecate an API - https://github.com/golang/go/wiki/Deprecated. The deprecated APIs need to have a line with "`Deprecated:` ...", as opposed to `DEPRECATED`.

christarazi (approved):

aditighag (approved):

PR #27582, opened 2023-08-18
go.mod, vendor: use github.com/cilium/dns fork directly

tklauser:
We have been maintaining and using that fork for a long time and it looks like the custom changes won't make it upstream any time soon.

There are no other vendored dependencies using miekg/dns, so switching to the cilium/dns fork shouldn't have any side effects. The fork's module name was changed to match its import path in https://github.com/cilium/dns/pull/4.

Let's replace the `github.com/miekg/dns` import path by `github.com/cilium/dns` to get rid of another `replace` directive in `go.mod` and thus make life a tiny bit easier for downstream packages importing `github.com/cilium/cilium`.


rolinh (approved):

jrajahalme (approved):

tklauser:
Rebased to resolve merge conflict.

borkmann (approved):

aanm (approved):

Issue #27320, opened 2023-08-07
Replace conditional compilation based on `#ifdef` with regular if statements


dylandreimerink:
Currently, Ciliums BPF code is riddled with `#ifdef` pre-processor macros which are used to conditionally compile certain features into the datapath. There are good reasons for doing so, it means that Cilium users who disabled the a given feature do not "pay" for any delays in the code related to checking at runtime if a feature is enabled or not. Loading less bytecode into the kernel can also be critical for complexity reasons. That way users of older kernels could pick which features they want but could not enable to many at once.

There are, downsides to this approach. It makes the code hard to understand, both due to the mixing of preprocessor and normal C statements. Another downside is that we need to ship the clang/LLVM toolchain with our application and recompile our BPF programs if feature settings change. Shipping the toolchain adds significant size to the docker image and compiling can take a while which is not idea. Lastly, dynamic compilation like this makes any attempts at signing BPF code (which is not yet a thing) practically impossible.

The idea proposes in this issue to to replace these pre-processor macros with regular C if statements which check the value of a global variable. 

Going from:
```c
void some_func() 
{
  // [...]
#ifdef IPV4_ENABLED
  return process_x_ipv4(...);
#endif

#ifdef IPV6_ENABLED
  return process_x_ipv6(...);
#endif
}
```

To:
```c
void some_func() 
{
  // [...]
  if (feature_ipv4)
    return process_x_ipv4(...);

  if (feature_ipv6)
    return process_x_ipv6(...);
}
```

The effect of this is that we always compile all of the code available, thus detecting any compilation issues in CI or at the time of cutting a release. By setting the value of these `feature_ipv4` and `feature_ipv6` at runtime (in the loader) we can pick which if a branch should be taken or not. This way we can disable a branch and the verifier will not be able to reach it and thus also not count it towards the complexity.

Since kernel version [v4.15](https://github.com/torvalds/linux/commit/c131187db2d3fa2f8bf32fdf4e9a4ef805168467) the kernel will replace dynamically dead code with NOP instructions for security reasons. The conditional branching instruction is still left in place.

[...]

PR #27589, opened 2023-08-18
bpf: static data: use inline asm to access static data

ti-mo:
The new static data approach is fully documented in the code. Using a volatile inline asm block makes for a more predictable outcome in the bytecode, meaning we can simplify the substitution logic on the Go side and start using runtime-provided constants for user space dead code elimination.

Part of: https://github.com/cilium/cilium/issues/27320

```release-note
bpf: static data: use inline asm to access static data
```

### DIFF
--- a/bpf/include/bpf/compiler.h
+++ b/bpf/include/bpf/compiler.h
@@ -53,10 +53,6 @@
 # define __stringify(X)		#X
 #endif
 
-#ifndef __fetch
-# define __fetch(X)		(__u64)(&(X))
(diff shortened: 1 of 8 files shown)


dylandreimerink (approved):

aojea:
@yasz24  can you PTAL and see the areas where you can help in follow ups?

ti-mo:
This will conflict with https://github.com/cilium/cilium/pull/28417, will wait for that to land first.

PR #27590, opened 2023-08-18
Fix a bug where cilium host IP is not read from k8s node annotations

### LINKED ISSUE #27439: CiliumInternalIP removed & re-added on cilium-agent restart

### What happened?

tl;dr On cilium-agent start-up, the CiliumNode object is updated to remove the `CiliumInternalIP`'s from `spec.addresses` and then immediately added back. 

Hello! I'm down a huge rabbit hole but believe I've finally found a reproducible bug present in Cilium 1.11.18, 1.12.11, and 1.14 while not occurring on Cilium 1.10.13. What got me here is a saga but, the issue is that whenever a cilium-agent pod is deleted (noticeable during daemonset rollouts), every cilium-agent is `update`'ing it's CiliumNode object to remove the `CiliumInternalIP` entires and then immediately adding them back. 

This is an issue as the downstream impacts cause all other nodes in the cluster to receive these updates (through kvstore in our case, appears present in CRD mode too). When these updates are processed in 1.11 and 1.12, they go through a very expensive `TriggerLabelInjection` method https://github.com/cilium/cilium/blob/v1.11/pkg/ipcache/metadata.go#L519-L567. From my profiling of the agent during this daemonset rollouts, this can yield +30% more time getting spent in that code path as a result. 

The CPU usage issue likely isn't an issue in 1.13+ as https://github.com/cilium/cilium/pull/19765 has re-worked most of it. However, while trying to upgrade to 1.12 this is becoming quite the issue as agents in larger clusters are getting overwhelmed with CiliumNode updates causing significant (i.e. <100mcores to >1.9 cores) CPU usage increases.

I've found this to be the easiest way to reproduce the superflous updates to the CiliumNode object:
1. In one terminal run: `kubectl get ciliumnode CILIUM_NODE_HERE -w -o json | jq '.spec.addresses'`
2. In another, `kubectl delete` the cilium-agent pod on that node 
3. Wait a bit, and you'll notice that 1 event will happen without `CiliumInternalIP`'s listed followed by another adding them back. 

I'll update this issue as I discover more relevent information. 

### Cilium Version

I've confirmed this behaviour on v1.11.18, 1.12.11, and 1.14

### Kernel Version

5.15 linux-aws

### Kubernetes Version

1.24

### Sysdump

Unable 

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

hemanthmalla:
After https://github.com/cilium/cilium/commit/0696874a932a07f9a5ed2b7a5f7aeb2db0757379 refactored the logic to read annotations and change the default behavior, addrs array was never assigned to newNode.IPAddresses after it was populated.

Fixes: #27439 partly,  there seem to be a race between when `n.localNode.IPAddresses` is populated and when the first ciliumnode(CN) update is made after an agent restart. In our fork we temporarily [added a behavior to retain host IP](https://github.com/DataDog/cilium/pull/517) from ciliumnode when localNode hasn't yet been populated to avoid the double writes. This isn't perfect because in the rare event where `cilium_host`'s router IP has been removed, it would not be updated on the CN.

Open Questions : 
* The daemon initialization process already seems to have strict requirements in what order restoration should happen. Can the order be improved so that we don't update CN before localNode.IPAddresses are populated ?

Thanks to @jaredledvina  for identifying the duplicate writes

### DIFF
--- a/pkg/k8s/node.go
+++ b/pkg/k8s/node.go
@@ -161,6 +161,7 @@ func ParseNode(k8sNode *slim_corev1.Node, source source.Source) *nodeTypes.Node 
 
 	k8sNodeAddHostIP(annotation.CiliumHostIP, annotation.CiliumHostIPAlias)
 	k8sNodeAddHostIP(annotation.CiliumHostIPv6, annotation.CiliumHostIPv6Alias)
+	newNode.IPAddresses = addrs
 
 	if key, ok := annotation.Get(k8sNode, annotation.CiliumEncryptionKey, annotation.CiliumEncryptionKeyAlias); ok {
 		if u, err := strconv.ParseUint(key, 10, 8); err == nil {
--- a/pkg/k8s/node_test.go
+++ b/pkg/k8s/node_test.go
@@ -29,8 +29,10 @@ func (s *K8sSuite) TestParseNode(c *C) {
 		ObjectMeta: slim_metav1.ObjectMeta{
 			Name: "node1",
 			Annotations: map[string]string{
-				annotation.V4CIDRName: "10.254.0.0/16",
-				annotation.V6CIDRName: "f00d:aaaa:bbbb:cccc:dddd:eeee::/112",
+				annotation.V4CIDRName:     "10.254.0.0/16",
+				annotation.V6CIDRName:     "f00d:aaaa:bbbb:cccc:dddd:eeee::/112",
+				annotation.CiliumHostIP:   "10.254.9.9",
+				annotation.CiliumHostIPv6: "fd00:10:244:1::8ace",
 			},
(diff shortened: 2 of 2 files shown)


hemanthmalla:
cc @christarazi / @aanm

christarazi:
> The daemon initialization process already seems to have strict requirements in what order restoration should happen. Can the order be improved so that we don't update CN before localNode.IPAddresses are populated ?

I think this sound reasonable to me. I am a bit surprised that's not the case today actually.

christarazi (approved):
Nice find!

aanm (approved):
Nice catch! Thank you

PR #27593, opened 2023-08-18
Resiliency: Add retry logic to stale ceps checks

derailed:

      GitHub issue.

Resiliency: Add retry logic to stale CEPS checks

> NOTE: Depends on PR: #27614 

When the cilium agent starts it attempts to reconcile CEPS from loaded state. 
If a failure occurs no further attempts will take place, thus stale CEPS could remain on the cluster.

- Add retry logic to allow retries to take place should errors occur during CEPS reconciliation.

Fixes: #issue-number

```release-note
During startup, the agent attempts to clear out any obsolete CiliumEndpoints. 
Add retry logic to ensure this process is attempted more than once should errors occur during reconciliation.
```

### DIFF
--- a/daemon/cmd/ciliumendpoints.go
+++ b/daemon/cmd/ciliumendpoints.go
@@ -83,6 +83,11 @@ func (d *Daemon) deleteCiliumEndpoint(
 	ciliumClient ciliumv2.CiliumV2Interface,
 	eps localEndpointCache,
 	endpointSliceEnabled bool) error {
+	logwf := log.WithFields(logrus.Fields{
+		logfields.CEPName:      cepName,
+		logfields.K8sNamespace: cepNamespace,
+	})
+
 	// To avoid having to store CEP UIDs in CES Endpoints array, we have to get the latest
 	// referenced CEP from apiserver to verify that it still references this node.
 	// To avoid excessive api calls, we only do this if CES is enabled and the CEP
@@ -91,17 +96,14 @@ func (d *Daemon) deleteCiliumEndpoint(
 		cep, err := ciliumClient.CiliumEndpoints(cepNamespace).Get(ctx, cepName, metav1.GetOptions{})
 		if err != nil {
 			if k8serrors.IsNotFound(err) {
-				log.WithError(err).WithFields(logrus.Fields{logfields.CEPName: cepName, logfields.K8sNamespace: cepNamespace}).
-					Info("CEP no longer exists, skipping staleness check")
+				logwf.WithError(err).Info("CEP no longer exists, skipping staleness check")
 				return nil
 			}
-			log.WithError(err).WithFields(logrus.Fields{logfields.CEPName: cepName, logfields.K8sNamespace: cepNamespace}).
-				Error("Failed to get possibly stale ciliumendpoints from apiserver")
+			logwf.WithError(err).Error("Failed to get possibly stale ciliumendpoints from apiserver")
 			return resiliency.NewRetryableErr(err)
 		}
 		if cep.Status.Networking.NodeIP != node.GetCiliumEndpointNodeIP() {
-			log.WithError(err).WithFields(logrus.Fields{logfields.CEPName: cepName, logfields.K8sNamespace: cepNamespace}).
-				Debug("Stale CEP fetched apiserver no longer references this Node, skipping.")
+			logwf.WithError(err).Debug("Stale CEP fetched apiserver no longer references this Node, skipping.")
 			return nil
 		}
 		cepUID = &cep.ObjectMeta.UID
@@ -112,23 +114,19 @@ func (d *Daemon) deleteCiliumEndpoint(
 	// This may occur for various reasons:
 	// * Pod was restarted while Cilium was not running (likely prior to CNI conf being installed).
 	// * Local endpoint was deleted (i.e. due to reboot + temporary filesystem) and Cilium or the Pod where restarted.
-	log.WithFields(logrus.Fields{
-		logfields.CEPName:      cepName,
-		logfields.K8sNamespace: cepNamespace,
-	}).Info("Found stale ciliumendpoint for local pod that is not being managed, deleting.")
+	logwf.Info("Found stale ciliumendpoint for local pod that is not being managed, deleting.")
 	if err := ciliumClient.CiliumEndpoints(cepNamespace).Delete(ctx, cepName, metav1.DeleteOptions{
 		Preconditions: &metav1.Preconditions{
 			UID: cepUID,
 		},
 	}); err != nil {
-		logger := log.WithError(err).WithFields(logrus.Fields{logfields.CEPName: cepName, logfields.K8sNamespace: cepNamespace})
 		if k8serrors.IsNotFound(err) {
 			// CEP not found, likely already deleted. Do not log as an error as that
 			// will fail CI runs.
-			logger.Debug("Could not delete stale CEP")
+			logwf.Debug("Could not delete stale CEP")
 			return nil
 		}
-		logger.Error("Could not delete stale CEP")
+		logwf.Error("Could not delete stale CEP")
 		return resiliency.NewRetryableErr(err)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/resiliency/types.go
christarazi, line 1: Could you split the addition of this package to a separate PR? It will make it easier to review and iterate on the package's API plus give us the flexibility in backporting. The usage code can still stay in this PR and it can be based off of the new PR.

derailed: Good idea! will do...

derailed: That said, I think some of these resiliency additions are likely going to be very invasive and likely not easy to back port...

daemon/cmd/daemon_main.go
```diff
@@ -1758,11 +1759,30 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
...
-			if err := d.cleanStaleCEPs(context.Background(), d.endpointManager, params.Clientset.CiliumV2(), option.Config.EnableCiliumEndpointSlice); err != nil {
-				log.WithError(err).Error("Failed to clean up stale CEPs")
+			bo := wait.Backoff{
+				Duration: 500 * time.Millisecond,
+				Factor:   1,
+				Jitter:   0.1,
+				Steps:    5,
+				Cap:      0,
+			}
+			err := wait.ExponentialBackoffWithContext(context.Background(), bo, func(ctx context.Context) (done bool, err error) {
```
tommyp1ckles, line 1789: One thing I'm wondering here is how we decide between using this approach vs pkg/controller (i.e. having a controller that runs until its error returns nil).

Also, another thought is that perhaps this might be a good case for the new "Job" type which supports retry until completion for OneShot jobs: https://github.com/cilium/cilium/blob/main/pkg/hive/job/job.go#L204

daemon/cmd/ciliumendpoints.go
```diff
@@ -46,8 +50,9 @@ func (d *Daemon) cleanStaleCEPs(ctx context.Context, eps localEndpointCache, cil
 			}
 			for _, cep := range ces.Endpoints {
 				if cep.Networking.NodeIP == node.GetCiliumEndpointNodeIP() && eps.LookupCEPName(ces.Namespace+"/"+cep.Name) == nil {
-					d.deleteCiliumEndpoint(ctx, ces.Namespace, cep.Name, nil, ciliumClient, eps,
-						enableCiliumEndpointSlice)
+					if err := d.deleteCiliumEndpoint(ctx, ces.Namespace, cep.Name, nil, ciliumClient, eps, enableCiliumEndpointSlice); err != nil {
+						errs = append(errs, err)
```
tommyp1ckles: nice, I like the idea of still trying to attempt cleanup of other CEPs even if one fails.

daemon/cmd/daemon_main.go
christarazi: One part of the criteria is that older stable branches won't have "Job" type.

christarazi: Controller pattern is the typical pattern in the Cilium code base. It already has logic to retry exponentially. It seems like it integrates well here in this case, so we might as well utilize it. 

With the controller pattern, it becomes easier to hold a queue of items that need to be processed and go through them one by one. If one fails, that item gets put back onto the queue. The controller will exit when the queue is empty. In this case, the item in the queue here is a stale CEP.

tommyp1ckles: > One part of the criteria is that older stable branches won't have "Job" type.

Yup, good point 👍

joestringer:
Is this a fix or a regular change? Should it be backported per the [backport criteria](https://docs.cilium.io/en/stable/contributing/release/backports/#id1)? Please also fill out the PR description so that reviewers can more easily understand the state, intention of the PR. Also please link any and all related issues/other PRs and the relationship / dependencies to make the intent and next steps clear. 

I think I was pulled in for review by @cilium/tophat but I feel like there was another PR around with the `pkg/resiliency` changes, so I'll defer on providing approval for now in case that gets resolved by the other PR getting merged + this one getting rebased.

derailed:
@joestringer Thank you! I've updated the description per your comment.
I think some of these will be slated for back-porting if deemed of necessary value/effort. 
I'll delegate to @christarazi here for advise.

aditighag (approved):
LGTM for my code owners (CODEOWNERS) change.

daemon/cmd/daemon_main.go
thorn3r: agreed, controller seems like the right fit here

christarazi: In case you weren't aware @derailed, controller here means the infrastructure provided by `pkg/controller`.

derailed: Somehow using a controller for this use case feels a bit heavy, the current impl already handles exp retries + retry limit. Possible with a controller of course but does not seems like the right use case for it in this instance imho.

christarazi: While the difference between a controller and the current state of the PR is minimal in terms of an actual mechanism for retries, controllers have a bit more structure around them. Controllers have a status and that is aggregated and reported to (1) metrics and (2) `cilium status` output. If this controller is failing, the user could easily be informed by the controller infrastructure, while the current PR today would have to add code to come to parity. In this case, there's no added benefit to duplicating the code for the PR to come to parity with what `pkg/controller` already provides. Does this make sense?

derailed: After discussing this with @christarazi, the current implementation is a better fit for this particular use case

thorn3r (approved):

daemon/cmd/daemon_main.go
```diff
@@ -1775,11 +1776,30 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
...
+					if resiliency.IsRetryable(err) {
+						return false, err
+					}
+					return true, err
+				}
+
+				return true, nil
+			})
+			if err != nil {
+				log.WithError(err).Error("Failed to clean up stale CEPs after retry")
```
christarazi:
```suggestion
					log.WithError(err).WithField(logfields.Attempt, x).Error("Failed to clean up stale CEPs")
					if resiliency.IsRetryable(err) {
						return false, err
					}
					return true, err
				}

				return true, nil
			})
			if err != nil {
				log.WithError(err).Error("Failed to clean up stale CEPs after multiple attempts")
```

I think we can add `attempt` here in the `WithFields` for extra context.

daemon/cmd/ciliumendpoints.go
```diff
@@ -87,16 +93,16 @@ func (d *Daemon) deleteCiliumEndpoint(
 			if k8serrors.IsNotFound(err) {
 				log.WithError(err).WithFields(logrus.Fields{logfields.CEPName: cepName, logfields.K8sNamespace: cepNamespace}).
 					Info("CEP no longer exists, skipping staleness check")
-			} else {
-				log.WithError(err).WithFields(logrus.Fields{logfields.CEPName: cepName, logfields.K8sNamespace: cepNamespace}).
-					Error("Failed to get possibly stale ciliumendpoints from apiserver, skipping.")
+				return nil
```
christarazi: We could so a bit of refactoring here and pull up the 

```
log.WithFields(logrus.Fields{
		logfields.CEPName:      cepName,
		logfields.K8sNamespace: cepNamespace,
```

a bit earlier and refactor the log statements which aren't DRY to use the logger fields from above.

derailed: Thanks Chris! I saw that but figured it might be a bit out of scope for this pr??
I'll add the refactor in a separate commit.

daemon/cmd/daemon_main.go
derailed: Good point! I'll update accordingly.

christarazi (approved):
LGTM. I think we need to drop the first commit about the resiliency package and it's good to go from my side

PR #27602, opened 2023-08-21
bpf: lxc: defer CT_INGRESS entry creation for loopback connections

julianwiedmann:
Currently the CT_INGRESS entry for a loopback connection is already created when the first request leaves the client, as from-container creates the CT_EGRESS entry (see the loopback handling in ct_create4() for details).

This is unusual - we would normally just create the CT_INGRESS entry as the first packet passes through to-container into the backend. But for loopback connections it is needed, so that
1.) to-container finds a CT entry with .loopback set, and thus skips
    network policy enforcement even for the first packet, and
2.) the CT entry has its rev_nat_index field populated, and thus can
    RevNAT replies in from-container.

This approach conflicts with the fact that loopback replies skip the client's to-container path (to avoid network policy enforcement).

Once the loopback connection is closed, the backend's from-container path observes the FIN / RST, and __ct_lookup4() updates the CT_INGRESS entry's lifetime to CT_CLOSE_TIMEOUT. But the client's to-container path will not observe the FIN / RST, and thus the CT_EGRESS entry's lifetime remains as CT_CONNECTION_LIFETIME_TCP. Therefore the CT_INGRESS entry will expire earlier, and potentially gets garbage-collected while the CT_EGRESS entry stays in place.

If the same loopback connection is subsequently re-opened, the client's from-container path finds the CT_EGRESS entry and thus will *not* call ct_create4(). Consequently the CT_INGRESS entry is not created, and the backend will not apply the loopback-specific handling described above. Inbound requests are potentially dropped (due to network policy), and/or replies are not RevNATed.

Fix this up by letting the backend path create its CT_INGRESS entry as usual. It just needs a bit of detection code in its CT_NEW handling to understand that the first packet belongs to a .loopback connection, and populate its own CT_INGRESS entry accordingly.

```release-note
Fix connectivity issues caused by missing conntrack entry when service pod connects to itself via clusterIP.
```

### DIFF
--- a/bpf/bpf_lxc.c
+++ b/bpf/bpf_lxc.c
@@ -1022,6 +1022,16 @@ ct_recreate4:
 	 */
 	if (is_defined(ENABLE_ROUTING) || hairpin_flow ||
 	    is_defined(ENABLE_HOST_ROUTING)) {
+		/* Hairpin requests need to pass through the backend's to-container
+		 * path, to create a CT_INGRESS entry with .lb_loopback set. This
+		 * drives RevNAT in the backend's from-container path.
+		 *
+		 * Hairpin replies are fully RevNATed in the backend's from-container
+		 * path. Thus they don't match the CT_EGRESS entry, and we can't rely
+		 * on a CT_REPLY result that would provide bypass of ingress policy.
+		 * Thus manually skip the ingress policy path.
+		 */
+		bool bypass_ingress_policy = hairpin_flow && ct_status == CT_REPLY;
 		struct endpoint_info *ep;
 
 		/* Lookup IPv4 address, this will return a match if:
@@ -1047,8 +1057,8 @@ ct_recreate4:
 			policy_clear_mark(ctx);
 			/* If the packet is from L7 LB it is coming from the host */
 			return ipv4_local_delivery(ctx, ETH_HLEN, SECLABEL_IPV4, ip4,
-						   ep, METRIC_EGRESS, from_l7lb, hairpin_flow,
-						   false, 0);
+						   ep, METRIC_EGRESS, from_l7lb,
+						   bypass_ingress_policy, false, 0);
 		}
 	}
 
@@ -1793,7 +1803,19 @@ ipv4_policy(struct __ctx_buff *ctx, int ifindex, __u32 src_label,
 	 * define policy rules to allow pods to talk to themselves. We still
 	 * want to execute the conntrack logic so that replies can be correctly
 	 * matched.
+	 *
+	 * If ip4.saddr is IPV4_LOOPBACK, this is almost certainly a loopback
+	 * connection. Populate
+	 * - .loopback, so that policy enforcement is bypassed, and
+	 * - .rev_nat_index, so that replies can be RevNATed.
 	 */
+	if (ret == CT_NEW && ip4->saddr == IPV4_LOOPBACK &&
+	    ct_has_loopback_egress_entry4(get_ct_map4(tuple), tuple,
+					  &ct_state_new.rev_nat_index)) {
+		ct_state_new.loopback = true;
+		goto skip_policy_enforcement;
+	}
+
 	if (unlikely(ct_state->loopback))
 		goto skip_policy_enforcement;
 #endif /* ENABLE_PER_PACKET_LB && !DISABLE_LOOPBACK_LB */
--- a/bpf/lib/conntrack.h
+++ b/bpf/lib/conntrack.h
@@ -977,33 +977,6 @@ static __always_inline int ct_create4(const void *map_main,
 	if (unlikely(err < 0))
 		goto err_ct_fill_up;
 
-#ifndef DISABLE_LOOPBACK_LB
-	if (dir == CT_EGRESS && ct_state->addr && ct_state->loopback) {
-		__u8 flags = tuple->flags;
-		__be32 saddr, daddr;
-
-		saddr = tuple->saddr;
-		daddr = tuple->daddr;
-
-		/* We are looping back into the origin endpoint through a
-		 * service. Set up a conntrack tuple for the reply to ensure we
-		 * do rev NAT before attempting to route the destination
-		 * address which will not point back to the right source.
-		 */
-		tuple->flags = TUPLE_F_IN;
-		tuple->saddr = ct_state->svc_addr;
-		tuple->daddr = ct_state->addr;
-
-		err = map_update_elem(map_main, tuple, &entry, 0);
-		if (unlikely(err < 0))
-			goto err_ct_fill_up;
-
-		tuple->saddr = saddr;
-		tuple->daddr = daddr;
-		tuple->flags = flags;
-	}
-#endif
-
 	if (map_related != NULL) {
 		/* Create an ICMP entry to relate errors */
 		struct ipv4_ct_tuple icmp_tuple = {
@@ -1031,6 +1004,27 @@ err_ct_fill_up:
 	send_signal_ct_fill_up(ctx, SIGNAL_PROTO_V4);
 	return DROP_CT_CREATE_FAILED;
 }
+
+#ifndef DISABLE_LOOPBACK_LB
+static __always_inline bool
+ct_has_loopback_egress_entry4(const void *map, struct ipv4_ct_tuple *tuple,
+			      __u16 *rev_nat_index)
+{
+	__u8 flags = tuple->flags;
+	struct ct_entry *entry;
+
+	tuple->flags = TUPLE_F_OUT;
+	entry = map_lookup_elem(map, tuple);
+	tuple->flags = flags;
+
+	if (entry && entry->lb_loopback) {
+		*rev_nat_index = entry->rev_nat_index;
+		return true;
+	}
+
+	return false;
+}
+#endif
 
 static __always_inline bool
 __ct_has_nodeport_egress_entry(const struct ct_entry *entry,
--- a/bpf/tests/hairpin_sctp_flow.c
+++ b/bpf/tests/hairpin_sctp_flow.c
@@ -46,6 +46,7 @@ struct {
 } entry_call_map __section(".maps") = {
 	.values = {
 		[0] = &cil_from_container,
+		[1] = &cil_to_container,
 	},
 };
 
@@ -165,8 +166,98 @@ int hairpin_flow_forward_check(__maybe_unused const struct __ctx_buff *ctx)
 	test_finish();
 }
 
+/* Let backend's ingress path create its CT own entry: */
+PKTGEN("tc", "hairpin_sctp_flow_2_forward_ingress_v4")
+int hairpin_flow_forward_ingress_pktgen(struct __ctx_buff *ctx)
+{
+	struct pktgen builder;
+	volatile const __u8 *src = mac_one;
+	volatile const __u8 *dst = mac_two;
+	struct iphdr *l3;
+	struct sctphdr *l4;
+
+	/* Init packet builder */
+	pktgen__init(&builder, ctx);
+
+	l3 = pktgen__push_ipv4_packet(&builder, (__u8 *)src, (__u8 *)dst,
+				      IPV4_LOOPBACK, v4_pod_one);
+	if (!l3)
+		return TEST_ERROR;
+
+	/* Push SCTP header */
+	l4 = pktgen__push_sctphdr(&builder);
+
+	if (!l4)
+		return TEST_ERROR;
+	if ((void *)l4 + sizeof(struct sctphdr) > ctx_data_end(ctx))
+		return TEST_ERROR;
+
+	l4->source = tcp_src_one;
+	l4->dest = tcp_svc_one;
+
+	/* Calc lengths, set protocol fields and calc checksums */
+	pktgen__finish(&builder);
+
+	return 0;
+}
+
+SETUP("tc", "hairpin_sctp_flow_2_forward_ingress_v4")
+int hairpin_flow_forward_ingress_setup(struct __ctx_buff *ctx)
+{
+	/* Jump into the entrypoint */
+	tail_call_static(ctx, &entry_call_map, 1);
+	/* Fail if we didn't jump */
+	return TEST_ERROR;
+}
+
+CHECK("tc", "hairpin_sctp_flow_2_forward_ingress_v4")
+int hairpin_flow_forward_ingress_check(__maybe_unused const struct __ctx_buff *ctx)
+{
+	void *data;
+	void *data_end;
+	__u32 *status_code;
+	struct iphdr *l3;
+	struct sctphdr *l4;
+
+	test_init();
+
+	data = (void *)(long)ctx->data;
+	data_end = (void *)(long)ctx->data_end;
+
+	if (data + sizeof(__u32) > data_end)
+		test_fatal("status code out of bounds");
+
+	status_code = data;
+
+	assert(*status_code == TC_ACT_OK);
+
+	l3 = data + sizeof(__u32) + sizeof(struct ethhdr);
+
+	if ((void *)l3 + sizeof(struct iphdr) > data_end)
+		test_fatal("l3 out of bounds");
+
+	if (l3->saddr != IPV4_LOOPBACK)
+		test_fatal("src IP changed");
+
+	if (l3->daddr != v4_pod_one)
+		test_fatal("dest IP changed");
+
+	l4 = (void *)l3 + sizeof(struct iphdr);
+
+	if ((void *)l4 + sizeof(struct sctphdr) > data_end)
+		test_fatal("l4 out of bounds");
+
+	if (l4->source != tcp_src_one)
+		test_fatal("src SCTP port changed");
+
+	if (l4->dest != tcp_svc_one)
+		test_fatal("dst SCTP port changed");
+
+	test_finish();
+}
+
 /* Test that a packet in the reverse direction gets translated back. */
-SETUP("tc", "hairpin_sctp_flow_2_reverse_v4")
+SETUP("tc", "hairpin_sctp_flow_3_reverse_v4")
 int hairpin_flow_rev_setup(struct __ctx_buff *ctx)
 {
 	struct pktgen builder;
@@ -214,7 +305,7 @@ int hairpin_flow_rev_setup(struct __ctx_buff *ctx)
 	return TEST_ERROR;
 }
 
-CHECK("tc", "hairpin_sctp_flow_2_reverse_v4")
+CHECK("tc", "hairpin_sctp_flow_3_reverse_v4")
 int hairpin_flow_rev_check(__maybe_unused const struct __ctx_buff *ctx)
 {
 	void *data;
--- a/bpf/tests/pktgen.h
+++ b/bpf/tests/pktgen.h
@@ -79,9 +79,13 @@ static volatile const __u8 v6_node_three[] = {0xfd, 0x07, 0, 0, 0, 0, 0, 0,
 					   0, 0, 0, 0, 0, 0, 0, 3};
 
 /* Source port to be used by a client */
-#define tcp_src_one	__bpf_htons(22334)
-#define tcp_src_two	__bpf_htons(33445)
-#define tcp_src_three	__bpf_htons(44556)
+#define tcp_src_one	__bpf_htons(22330)
+#define tcp_src_two	__bpf_htons(33440)
+#define tcp_src_three	__bpf_htons(44550)
+
+#define tcp_dst_one	__bpf_htons(22331)
+#define tcp_dst_two	__bpf_htons(33441)
+#define tcp_dst_three	__bpf_htons(44551)
 
 #define tcp_svc_one	__bpf_htons(80)
 #define tcp_svc_two	__bpf_htons(443)
--- a/bpf/tests/tc_nodeport_test.c
+++ b/bpf/tests/tc_nodeport_test.c
@@ -103,7 +103,7 @@ int hairpin_flow_forward_setup(struct __ctx_buff *ctx)
 
 	lb_v4_add_service(v4_svc_one, tcp_svc_one, 1, revnat_id);
 	lb_v4_add_backend(v4_svc_one, tcp_svc_one, 1, 124,
-			  v4_pod_one, tcp_svc_one, IPPROTO_TCP, 0);
+			  v4_pod_one, tcp_dst_one, IPPROTO_TCP, 0);
 
 	/* Add an IPCache entry for pod 1 */
 	ipcache_v4_add_entry(v4_pod_one, 0, 112233, 0, 0);
@@ -156,7 +156,7 @@ int hairpin_flow_forward_check(__maybe_unused const struct __ctx_buff *ctx)
 	if (l4->source != tcp_src_one)
 		test_fatal("src TCP port was changed");
 
-	if (l4->dest != tcp_svc_one)
+	if (l4->dest != tcp_dst_one)
 		test_fatal("dst TCP port incorrect");
 
 	struct ipv4_ct_tuple tuple = {};
@@ -183,7 +183,7 @@ int hairpin_flow_forward_check(__maybe_unused const struct __ctx_buff *ctx)
 	tuple.saddr = IPV4_LOOPBACK;
 	tuple.sport = tcp_src_one;
 	tuple.daddr = v4_pod_one;
-	tuple.dport = tcp_svc_one;
+	tuple.dport = tcp_dst_one;
 
 	/* Addrs are stored in reverse order: */
 	ipv4_ct_tuple_swap_addrs(&tuple);
@@ -213,7 +213,7 @@ int hairpin_flow_forward_ingress_pktgen(struct __ctx_buff *ctx)
 	l4 = pktgen__push_ipv4_tcp_packet(&builder,
 					  (__u8 *)src, (__u8 *)dst,
 					  IPV4_LOOPBACK, v4_pod_one,
-					  tcp_src_one, tcp_svc_one);
+					  tcp_src_one, tcp_dst_one);
 	if (!l4)
 		return TEST_ERROR;
 
@@ -278,7 +278,7 @@ int hairpin_flow_forward_ingress_check(__maybe_unused const struct __ctx_buff *c
 	if (l4->source != tcp_src_one)
 		test_fatal("src TCP port changed");
 
-	if (l4->dest != tcp_svc_one)
+	if (l4->dest != tcp_dst_one)
 		test_fatal("dst TCP port changed");
 
 	struct ipv4_ct_tuple tuple = {};
@@ -290,7 +290,7 @@ int hairpin_flow_forward_ingress_check(__maybe_unused const struct __ctx_buff *c
 	tuple.saddr = IPV4_LOOPBACK;
 	tuple.sport = tcp_src_one;
 	tuple.daddr = v4_pod_one;
-	tuple.dport = tcp_svc_one;
+	tuple.dport = tcp_dst_one;
 
 	/* Addrs are stored in reverse order: */
 	ipv4_ct_tuple_swap_addrs(&tuple);
@@ -319,7 +319,7 @@ int hairpin_flow_reverse_pktgen(struct __ctx_buff *ctx)
 	l4 = pktgen__push_ipv4_tcp_packet(&builder,
 					  (__u8 *)src, (__u8 *)dst,
 					  v4_pod_one, IPV4_LOOPBACK,
-					  tcp_svc_one, tcp_src_one);
+					  tcp_dst_one, tcp_src_one);
 	if (!l4)
 		return TEST_ERROR;
 
--- a/bpf/tests/tc_nodeport_test.c
+++ b/bpf/tests/tc_nodeport_test.c
@@ -46,6 +46,7 @@ struct {
 } entry_call_map __section(".maps") = {
 	.values = {
 		[0] = &cil_from_container,
+		[1] = &cil_to_container,
 	},
 };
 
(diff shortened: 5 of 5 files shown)


aojea:
can you please add me as reviewer, I will make some time during this week to review and test this

aojea:
surprisingly I can not reproduce it with and without this patch, however, the CT entries are different

without the patch
```
cilium bpf ct list global -d | grep  41746
TCP IN 169.254.42.1:41746 -> 10.244.0.70:80 expires=1600072 (remaining: 21442 sec(s)) RxPackets=0 RxBytes=0 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=21 TxBytes=2109 TxFlagsSeen=0x1b LastTxReport=1578472 Flags=0x001a [ TxClosing LBLoopback SeenNonSyn ] RevNAT=6 SourceSecurityID=48450 IfIndex=0
TCP OUT 169.254.42.1:41746 -> 10.244.0.70:80 expires=1600072 (remaining: 21442 sec(s)) RxPackets=0 RxBytes=0 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=30 TxBytes=2445 TxFlagsSeen=0x1b LastTxReport=1578472 Flags=0x001a [ TxClosing LBLoopback SeenNonSyn ] RevNAT=6 SourceSecurityID=48450 IfIndex=0
TCP OUT 10.96.234.161:80 -> 10.244.0.70:41746 service expires=1600072 (remaining: 21442 sec(s)) RxPackets=0 RxBytes=9 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=0 TxBytes=0 TxFlagsSeen=0x1b LastTxReport=1578472 Flags=0x0012 [ TxClosing SeenNonSyn ] RevNAT=6 SourceSecurityID=0 IfIndex=0
```

with this patch
```
TCP IN 169.254.42.1:41746 -> 10.244.0.70:80 expires=1580663 (remaining: -10 sec(s)) RxPackets=6 RxBytes=489 RxFlagsSeen=0x1b LastRxReport=1580653 TxPackets=4 TxBytes=407 TxFlagsSeen=0x1b LastTxReport=1580653 Flags=0x001b [ RxClosing TxClosing LBLoopback SeenNonSyn ] RevNAT=6 SourceSecurityID=48450 IfIndex=0
TCP OUT 169.254.42.1:41746 -> 10.244.0.70:80 expires=1602253 (remaining: 21580 sec(s)) RxPackets=0 RxBytes=0 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=6 TxBytes=489 TxFlagsSeen=0x1b LastTxReport=1580653 Flags=0x001a [ TxClosing LBLoopback SeenNonSyn ] RevNAT=6 SourceSecurityID=48450 IfIndex=0
TCP OUT 10.96.234.161:80 -> 10.244.0.70:41746 service expires=1602253 (remaining: 21580 sec(s)) RxPackets=0 RxBytes=9 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=0 TxBytes=0 TxFlagsSeen=0x1b LastTxReport=1580653 Flags=0x0012 [ TxClosing SeenNonSyn ] RevNAT=6 SourceSecurityID=0 IfIndex=0
```

bpf/bpf_lxc.c
```diff
@@ -1043,6 +1043,7 @@ static __always_inline int handle_ipv4_from_lxc(struct __ctx_buff *ctx, __u32 *d
 	 */
 	if (is_defined(ENABLE_ROUTING) || hairpin_flow ||
 	    is_defined(ENABLE_HOST_ROUTING)) {
+		bool bypass = hairpin_flow && ct_status == CT_REPLY;
```
aojea: so, this is so `hairpin_flow` are only considered for `ipv4_local_delivery(` if they are part of a REPLY request?

aojea:
ok, @julianwiedmann I decipher the problem I was having to reproduce it. 

I need to start from clean state each time I try a new version, this means flushing conntrack table and install a new pod (I do not know how to force the bpf recompilation :/)

With this patch I observe the "wrong" behavior, asymmetry on the timers (though RxClosing is set correctly) it sets the remaining time of TCP_IN to 5s, but the TCP_OUT to 21595s

```
cilium bpf ct list global -d | grep 43488
TCP OUT 10.96.77.183:80 -> 10.244.1.160:43488 service expires=1700265 (remaining: 21595 sec(s)) RxPackets=0 RxBytes=13 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=0 TxBytes=0 TxFlagsSeen=0x1b LastTxReport=1678662 Flags=0x0012 [ TxClosing SeenNonSyn ] RevNAT=5 SourceSecurityID=0 IfIndex=0
TCP IN 169.254.42.1:43488 -> 10.244.1.160:80 expires=1678675 (remaining: 5 sec(s)) RxPackets=15 RxBytes=1254 RxFlagsSeen=0x1b LastRxReport=1678662 TxPackets=15 TxBytes=1881 TxFlagsSeen=0x1b LastTxReport=1678662 Flags=0x001b [ RxClosing TxClosing LBLoopback SeenNonSyn ] RevNAT=5 SourceSecurityID=48450 IfIndex=0
TCP OUT 169.254.42.1:43488 -> 10.244.1.160:80 expires=1700265 (remaining: 21595 sec(s)) RxPackets=0 RxBytes=0 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=15 TxBytes=1254 TxFlagsSeen=0x1b LastTxReport=1678662 Flags=0x001a [ TxClosing LBLoopback SeenNonSyn ] RevNAT=5 SourceSecurityID=48450 IfIndex=0
```

with versions v1.14.0 and v1.14.1 the timers look ok, but the RxClosing state is not present
```
cilium bpf ct list global -d | grep 43488
TCP OUT 10.96.239.241:80 -> 10.244.1.167:43488 service expires=1699206 (remaining: 21596 sec(s)) RxPackets=0 RxBytes=11 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=0 TxBytes=0 TxFlagsSeen=0x1b LastTxReport=1677602 Flags=0x0012 [ TxClosing SeenNonSyn ] RevNAT=6 SourceSecurityID=0 IfIndex=0
TCP OUT 169.254.42.1:43488 -> 10.244.1.167:80 expires=1699206 (remaining: 21596 sec(s)) RxPackets=0 RxBytes=0 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=15 TxBytes=1257 TxFlagsSeen=0x1b LastTxReport=1677602 Flags=0x001a [ TxClosing LBLoopback SeenNonSyn ] RevNAT=6 SourceSecurityID=48450 IfIndex=0
TCP IN 169.254.42.1:43488 -> 10.244.1.167:80 expires=1699206 (remaining: 21596 sec(s)) RxPackets=0 RxBytes=0 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=16 TxBytes=1955 TxFlagsSeen=0x1b LastTxReport=1677602 Flags=0x001a [ TxClosing LBLoopback SeenNonSyn ] RevNAT=6 SourceSecurityID=48450 IfIndex=0
```

This is how can you reproduce the issue, the problem is that this is not easy to automate as it depends on the timers to be incorrect and eventually get one of these entries that has expired in one direction and not in the other, that make take some time O(min)

1. Install cilium with kind `KUBEPROXY_MODE="none" make kind && make kind-image`

2. Install cilium now , I use this one but for this case I assume any with kube-proxy strict should be enough
````
helm upgrade -i cilium ./install/kubernetes/cilium \
    --wait \
    --namespace kube-system \
    --set k8sServiceHost="kind-control-plane" \
    --set k8sServicePort="6443" \
    --set debug.enabled=true \
    --set pprof.enabled=true \
    --set enableIPv4Masquerade=false \
    --set enableIPv6Masquerade=false \
    --set enableIPv4Masquerade=false \
    --set hostFirewall.enabled=false \
    --set socketLB.enabled=true \
    --set socketLB.hostNamespaceOnly=true \
    --set nodeinit.enabled=true \
    --set ipam.mode=kubernetes \
    --set routingMode=native \
    --set ipv4.enabled=true \
    --set ipv4NativeRoutingCIDR=10.244.0.0/16 \
    --set ipv6.enabled=false \
    --set image.override="localhost:5000/cilium/cilium-dev:local" \
    --set image.pullPolicy=Never \
    --set operator.image.override="localhost:5000/cilium/operator-generic:local" \
    --set operator.image.pullPolicy=Never \
    --set operator.image.suffix="" \
    --set securityContext.privileged=true \
    --set gatewayAPI.enabled=false
```

Now these are the steps to reproduce, everything you install a new cilium versions you have to start from here by doing
`make kind-image` to build a cilium image for current checkout commit
`kubectl -n kube-system rollout restart ds/cilium` to reload the cilium agent

Install a Pod with a Services that references itself and it runs a web server

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: server-deployment
  labels:
    app: MyApp
spec:
  replicas: 1
  selector:
    matchLabels:
      app: MyApp
  template:
    metadata:
      labels:
        app: MyApp
    spec:
      nodeName: kind-worker
      containers:
      - name: agnhost
        image: httpd:2
        ports:
        - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: myservice
spec:
  type: ClusterIP
  selector:
    app: MyApp
  clusterIP: 10.96.7.7
  ports:
    - protocol: TCP
      port: 80
      targetPort: 80
```
The pod will always run in the node `kind-worker`for simplicity and the Service has the ClusterIP hardcoded to 10.96.7.7, I run 100000 request to be sure all ephemera ports are exercised and I grep for one of them
```
kubectl get pods
NAME                                 READY   STATUS    RESTARTS   AGE
server-deployment-5d9f6fdf8b-ggbvm   1/1     Running   0          12s
```
```
kubectl exec -it server-deployment-5d9f6fdf8b-ggbvm -- ab -n 60000 -c 100 http://10.96.7.7/
This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 10.96.7.7 (be patient)
Completed 6000 requests
Completed 12000 requests
Completed 18000 requests
Completed 24000 requests
Completed 30000 requests
```

In other console log into the corresponding cilium agent to get the conntrack entries
```
docker exec -it kind-worker bash
crictl exec -it $(crictl ps | grep cilium-agent | awk '{print $1}') bash
 cilium bpf ct list global -d | grep 43488
TCP OUT 10.96.77.183:80 -> 10.244.1.160:43488 service expires=1700265 (remaining: 21595 sec(s)) RxPackets=0 RxBytes=13 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=0 TxBytes=0 TxFlagsSeen=0x1b LastTxReport=1678662 Flags=0x0012 [ TxClosing SeenNonSyn ] RevNAT=5 SourceSecurityID=0 IfIndex=0
TCP IN 169.254.42.1:43488 -> 10.244.1.160:80 expires=1678675 (remaining: 5 sec(s)) RxPackets=15 RxBytes=1254 RxFlagsSeen=0x1b LastRxReport=1678662 TxPackets=15 TxBytes=1881 TxFlagsSeen=0x1b LastTxReport=1678662 Flags=0x001b [ RxClosing TxClosing LBLoopback SeenNonSyn ] RevNAT=5 SourceSecurityID=48450 IfIndex=0
TCP OUT 169.254.42.1:43488 -> 10.244.1.160:80 expires=1700265 (remaining: 21595 sec(s)) RxPackets=0 RxBytes=0 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=15 TxBytes=1254 TxFlagsSeen=0x1b LastTxReport=1678662 Flags=0x001a [ TxClosing LBLoopback SeenNonSyn ] RevNAT=5 SourceSecurityID=48450 IfIndex=0

```

aojea:
I talked with @sypakine and he still sees the issue in v1.14.0, it turns out that the behavior exhibited depends on the flags, with this config in v1.14.0 still shows assymetrics values

```
KUBEPROXY_MODE="none" make kind && \
make kind-image && \
helm upgrade -i cilium ./install/kubernetes/cilium \
  --wait \
  --namespace kube-system \
  --set k8sServiceHost="kind-control-plane" \
  --set k8sServicePort="6443" \
  --set debug.enabled=true \
  --set pprof.enabled=true \
  --set enableIPv4Masquerade=false \
  --set enableIPv6Masquerade=false \
  --set enableIPv4Masquerade=false \
  --set hostFirewall.enabled=false \
  --set socketLB.hostNamespaceOnly=true \
  --set nodeinit.enabled=true \
  --set ipam.mode=kubernetes \
  --set ipv4.enabled=true \
  --set ipv4NativeRoutingCIDR=10.244.0.0/16 \
  --set ipv6.enabled=false \
  --set image.override="localhost:5000/cilium/cilium-dev:local" \
  --set image.pullPolicy=Never \
  --set operator.image.override="localhost:5000/cilium/operator-generic:local" \
  --set operator.image.pullPolicy=Never \
  --set operator.image.suffix="" \
  --set securityContext.privileged=true \
  --set gatewayAPI.enabled=false \
  --set socketLB.enabled=false \
  --set bpf.hostLegacyRouting=true \
  --set endpointRoutes.enabled=true
```

julianwiedmann:
> With this patch I observe the "wrong" behavior, asymmetry on the timers (though RxClosing is set correctly) it sets the remaining time of TCP_IN to 5s, but the TCP_OUT to 21595s

Yep, not trying to fix the timers :grin:. That would require fixing up all the path symmetry for loopback traffic. Bigger topic.

But if the TCP_IN entry expires & gets GC-collected (while the TCP_OUT stays around), you now should see that a new connection re-creates the required TCP_IN entry.

> it turns out that the behavior exhibited depends on the flags, with this config in v1.14.0 still shows assymetrics values

ack, `endpointRoutes.enabled=true` is not a good combination for loopback connections right now. It doesn't play well with the bypass logic.

bpf/bpf_lxc.c
julianwiedmann: Yeah that's a dirty hack atm to get it working (primarily why this PR is still WIP / in draft). If we want the backend's `to-container` path to create the TCP_IN entry, then we can't bypass it.

Ideally we wouldn't need to bypass the client's `to-container` either, but that requires bigger changes (shifting around where the Service RevDNAT happens for loopback, so that the client's `to-container` can associate the replies with its TCP_OUT entry).

aojea:
> Yep, not trying to fix the timers 😁. That would require fixing up all the path symmetry for loopback traffic. Bigger topic.
> But if the TCP_IN entry expires & gets GC-collected (while the TCP_OUT stays around), you now should see that a new connection re-creates the required TCP_IN entry.

understood , cc @sypakine , I'll fall back to the `while true; do curl http://serviceip ; done` that eventually triggered this problem

bpf/bpf_lxc.c
aojea: this question is for my own education 😄

bpf/bpf_lxc.c
```diff
@@ -1803,6 +1804,19 @@ ipv4_policy(struct __ctx_buff *ctx, int ifindex, __u32 src_label,
...
+	/* If ip4.saddr is IPV4_LOOPBACK, this is almost certainly a loopback
+	 * connection. Populate
+	 * - .loopback, so that policy enforcement is bypassed, and
+	 * - .rev_nat_index, so that replies can be RevNATed.
+	 */
+	if (ret == CT_NEW && ip4->saddr == IPV4_LOOPBACK &&
+	    ct_has_loopback_egress_entry4(get_ct_map4(tuple), tuple,
+					  &ct_state_new.rev_nat_index)) {
+		ct_state_new.loopback = true;
+		goto skip_policy_enforcement;
```
aojea, line 1816: we have another issue with hairpin traffic and network policies , @sypakine  knows more about it.
I think that if the network policies matches the pod it actually drops the traffic, and before in 1.11 it was not doing that.

Just using this comment to not forget about this

sypakine: I'll file a new issue for it. I think we can treat it as orthogonal to this fix.

sypakine:
> > With this patch I observe the "wrong" behavior, asymmetry on the timers (though RxClosing is set correctly) it sets the remaining time of TCP_IN to 5s, but the TCP_OUT to 21595s
> 
> Yep, not trying to fix the timers 😁. That would require fixing up all the path symmetry for loopback traffic. Bigger topic.
> 
> But if the TCP_IN entry expires & gets GC-collected (while the TCP_OUT stays around), you now should see that a new connection re-creates the required TCP_IN entry.
> 
> > it turns out that the behavior exhibited depends on the flags, with this config in v1.14.0 still shows assymetrics values
> 
> ack, `endpointRoutes.enabled=true` is not a good combination for loopback connections right now. It doesn't play well with the bypass logic.

I verified this fix. It does re-create the appropriate TCP_IN entry.

Repro: 

```
gh repo clone cilium/ciium cilium-oss && cd !$ && gh pr checkout 27602
KUBEPROXY_MODE="none" make kind && \
make kind-image && \
helm upgrade -i cilium ./install/kubernetes/cilium \
  --wait \
  --namespace kube-system \
  --set k8sServiceHost="kind-control-plane" \
  --set k8sServicePort="6443" \
  --set debug.enabled=true \
  --set pprof.enabled=true \
  --set enableIPv4Masquerade=false \
  --set enableIPv6Masquerade=false \
  --set enableIPv4Masquerade=false \
  --set hostFirewall.enabled=false \
  --set socketLB.hostNamespaceOnly=true \
  --set nodeinit.enabled=true \
  --set ipam.mode=kubernetes \
  --set ipv4.enabled=true \
  --set ipv4NativeRoutingCIDR=10.244.0.0/16 \
  --set ipv6.enabled=false \
  --set image.override="localhost:5000/cilium/cilium-dev:local" \
  --set image.pullPolicy=Never \
  --set operator.image.override="localhost:5000/cilium/operator-generic:local" \
  --set operator.image.pullPolicy=Never \
  --set operator.image.suffix="" \
  --set securityContext.privileged=true \
  --set gatewayAPI.enabled=false \
  --set socketLB.enabled=false \
  --set bpf.hostLegacyRouting=true \
  --set endpointRoutes.enabled=true

alias kk="kubectl --context kind-kind"

kk --context kind-kind apply -f - <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: server
  namespace: default
  labels:
    app: server
spec:
  containers:
  - name: agnhost
    image: k8s.gcr.io/e2e-test-images/agnhost:2.39
    args:
    - netexec
    - --http-port=80
    ports:
    - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: server-svc
spec:
  type: ClusterIP
  selector:
    app: server
  ports:
    - protocol: TCP
      port: 80
      targetPort: 80
EOF

time while ! [[ "$( kk get svc server-svc -ojsonpath='{.spec.clusterIP}' | xargs )" != "" ]]; do echo -n "."; sleep 1; done
clusterip=$( kk get svc server-svc -ojsonpath='{.spec.clusterIP}' )

time while ! [[ "$( kk get pod server -ojsonpath='{.status.conditions[?(@.type=="Ready")].status}' )" =~ "True" ]]; do echo -n "."; sleep 1; done

host=$( kk get pod server \
    -ojsonpath='{ .spec.nodeName }' )
agent=$( kk get pod -n kube-system \
    -l k8s-app=cilium \
    --field-selector=spec.nodeName=${host} \
    -ojsonpath='{.items[].metadata.name}' )

watch "kubectl --context kind-kind exec -it -n kube-system ${agent} -c cilium-agent -- cilium bpf ct list global -d | grep ':55555'"

# new terminal
kk exec server -- curl --no-keepalive --connect-timeout 5 --local-port 55555 -v ${clusterip}:80/hostname

# wait for TCP_IN entry to timeout
kk exec server -- curl --no-keepalive --connect-timeout 5 --local-port 55555 -v ${clusterip}:80/hostname
```

julianwiedmann:
(pushed a slightly updated version with more comments, and tests for the expected CT entries)

bpf/bpf_lxc.c
julianwiedmann: Added a comment that explains what's going on here :+1:.

aojea:
Ok, now I have a consistent way to reproduce it and verify this patch works, I don't know how can be easily automated since depend on lowering the value of `--conntrack-gc-interval` so it reaps the dead entry.

The steps to repro are exactly the same as https://github.com/cilium/cilium/pull/27602#issuecomment-1692079927, in this case I just patch the cilium config-map to set a value of 60 seconds and restart the agents to pick up the new config
```
 kubectl -n kube-system get cm cilium-config -o yaml | grep conntrack-gc-interval
  conntrack-gc-interval: 1m0s
```

So, once you create the container you do the request against the service and it hairpins creating the conntrack entries that by default I think it has 15 seconds (I use the same trick as @sypakine so we have a well known port to grep `curl --no-keepalive --connect-timeout 5 --local-port 55555 -v ${clusterip}:80/hostname` )

```
bpf ct list global -d | grep 169.254.42.1:55555
TCP IN 169.254.42.1:55555 -> 10.244.1.70:80 expires=1851690 (remaining: 8 sec(s)) RxPackets=6 RxBytes=477 RxFlagsSeen=0x1b LastRxReport=1851680 TxPackets=4 TxBytes=452 TxFlagsSeen=0x1b LastTxReport=1851680 Flags=0x001b [ RxClosing TxClosing LBLoopback SeenNonSyn ] RevNAT=6 SourceSecurityID=27633 IfIndex=0
TCP OUT 169.254.42.1:55555 -> 10.244.1.70:80 expires=1873280 (remaining: 21598 sec(s)) RxPackets=0 RxBytes=0 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=6 TxBytes=477 TxFlagsSeen=0x1b LastTxReport=1851680 Flags=0x001a [ TxClosing LBLoopback SeenNonSyn ] RevNAT=6 SourceSecurityID=27633 IfIndex=0

```

Now, after 60 seconds, the GC will delete that entry
```
# cilium bpf ct list global -d | grep 169.254.42.1:55555
TCP OUT 169.254.42.1:55555 -> 10.244.1.70:80 expires=1873280 (remaining: 21535 sec(s)) RxPackets=0 RxBytes=0 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=6 TxBytes=477 TxFlagsSeen=0x1b LastTxReport=1851680 Flags=0x001a [ TxClosing LBLoopback SeenNonSyn ] RevNAT=6 SourceSecurityID=27633 IfIndex=0
```


 without this patch the connection times out
```
bash-5.0# curl --no-keepalive --connect-timeout 5 --local-port 55555 10.96.7.7
NOW: 2023-08-25 13:28:16.025963709 +0000 UTC m=+79359.922380945bash-5.0# curl --no-keepalive --connect-timeout 5 --local-port 55555 10.96.7.7
curl: (28) Connection timeout after 5001 ms
```

with the patch a new entry is created:
```
cilium bpf ct list global -d | grep 169.254.42.1:55555
TCP IN 169.254.42.1:55555 -> 10.244.1.70:80 expires=1851771 (remaining: 8 sec(s)) RxPackets=6 RxBytes=477 RxFlagsSeen=0x1b LastRxReport=1851761 TxPackets=4 TxBytes=452 TxFlagsSeen=0x1b LastTxReport=1851761 Flags=0x001b [ RxClosing TxClosing LBLoopback SeenNonSyn ] RevNAT=6 SourceSecurityID=27633 IfIndex=0
TCP OUT 169.254.42.1:55555 -> 10.244.1.70:80 expires=1873361 (remaining: 21598 sec(s)) RxPackets=0 RxBytes=0 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=12 TxBytes=954 TxFlagsSeen=0x1b LastTxReport=1851761 Flags=0x001a [ TxClosing LBLoopback SeenNonSyn ] RevNAT=6 SourceSecurityID=27633 IfIndex=0

```


I wish I knew how to create some test for this, ideally we should have some way of forcing the garbage collection of the conntrack entries, the alternative is to go to the bpf map and forcefully delete the entry 

/lgtm

bpf/bpf_lxc.c
aojea: yeah, please do, we should not conflate both problems, Thanks

bpf/lib/conntrack.h
```diff
@@ -1032,6 +1005,27 @@ static __always_inline int ct_create4(const void *map_main,
 	return DROP_CT_CREATE_FAILED;
 }
 
+#ifndef DISABLE_LOOPBACK_LB
+static __always_inline bool
+ct_has_loopback_egress_entry4(const void *map, struct ipv4_ct_tuple *tuple,
```
aojea, line 1010: naive question, what are the naming conventions here?
asking because the function mutates the input setting the `rev_nat_index` , right?

bpf/bpf_lxc.c
sypakine: filed https://github.com/cilium/cilium/issues/27709

bpf/bpf_lxc.c
```diff
@@ -1022,6 +1022,16 @@ static __always_inline int handle_ipv4_from_lxc(struct __ctx_buff *ctx, __u32 *d
...
+		/* Hairpin requests need to pass through the backend's to-container
+		 * path, to create a CT_INGRESS entry with .lb_loopback set. This
+		 * drives RevNAT in the backend's from-container path.
+		 *
+		 * Hairpin replies are fully RevNATed in the backend's from-container
+		 * path. Thus they don't match the CT_EGRESS entry, and we can't rely
+		 * on a CT_REPLY result that would provide bypass of ingress policy.
+		 * Thus manually skip the ingress policy path.
+		 */
+		bool bypass_ingress_policy = hairpin_flow && ct_status == CT_REPLY;
```
aditighag, line 1034: This means we defer bypassing policy enforcement for requests only (TCP syns), replies bypass policy enforcement straightaway. 

Thinking over this a bit more...
I wonder when packets are not passed through stack (e.g., endpoint routes), can we store some context in the packet scratch buffer.

aditighag (commented):
Thanks for the detailed description, and making an effort to detangle the loopback datapath. I made an initial pass, still mulling over the fix. Curios to know your thoughts about my alternate approach. 

In a nutshell, this PR updates the egress logic that previously also created an ingress entry for the loopback case, and instead let the ingress create a ct entry if the corresponding egress entry existed. The main issue where is that the datapath can't update the ingress entry gc timeout when it updates the egress entry, so there is inherently an asymmetry here. 
From that perspective, the change makes sense to me, but it's not ideal that ct logic is so tightly coupled with policy enforcement. 

>If the same loopback connection is subsequently re-opened, the client's from-container path finds the CT_EGRESS entry and thus will not call ct_create4(). 

Just for the sake of discussion, can it check for an ingress entry, and create one if doesn't exist? On a related note, we have logic to select a new backend after some pre-configured time has passed (see `ct_entry_expired_rebalance`). The same logic should apply for loopback connections as well. Wdyt?

bpf/bpf_lxc.c
julianwiedmann: >I wonder when packets are not passed through stack (e.g., endpoint routes), can we store some context in the packet scratch buffer.

As optimization to avoid the `ct_has_loopback_egress_entry4()` call you mean? Sure, that might work. Not something I would want to backport though.

I also believe we're still at a stage where we need to streamline the code paths, and produce something that behaves (close to) identically with/without per-EP routes. Fine-tuning can come afterwards.

bpf/lib/conntrack.h
julianwiedmann: >naive question, what are the naming conventions here?

These "raw" CT lookups are so low-level that I wouldn't strive for perfect naming / documentation. If you need to use them, you've already left safe territory and better know what you're doing.

> asking because the function mutates the input setting the rev_nat_index , right?

yep, that's needed to make the RevNAT work for loopback connections. Previously the CT_INGRESS entry would just implicitly inherit the `rev_nat_index` from the CT_EGRESS entry during creation time.

julianwiedmann:
>From that perspective, the change makes sense to me, but it's not ideal that ct logic is so tightly coupled with policy enforcement.

I think that's beyond the scope of this PR :). The fact that CT drives policy enforcement/bypass is baked into Cilium pretty deep.

> Just for the sake of discussion, can it check for an ingress entry, and create one if doesn't exist?

I suppose it could. I don't see the advantage of doing so - it would just add even more special logic to an already confusing code path. Also merely checking for its existence isn't sufficient for robustness - GC could still kick in 1 nsec later and clear up the CT_INGRESS entry. Beyond that the current CT_INGRESS entry isn't even created properly - note how it accidentally inherits the CT_EGRESS entry's TX statistics and `tx_flags_seen` / `last_tx_report` values. It's just not a robust design :shrug:.

I've mulled over this quite a bit, imho letting the to-container path handle the CT_INGRESS creation "as usual" is the cleanest & robust solution, with the least potential for surprises.


> On a related note, we have logic to select a new backend after some pre-configured time has passed (see ct_entry_expired_rebalance). The same logic should apply for loopback connections as well. Wdyt?

I would expect that it does? The service lookup in `bpf_lxc` goes through the same `lb4_local()` code.

julianwiedmann:
Targeting backports at 1.14 for now. Let's see how bad that is before looking at 1.12 / 1.13.

bpf/bpf_lxc.c
aditighag: Well, that too. But, mainly so that we could have a consistent behavior for all packets in a flow.

aditighag:
> > From that perspective, the change makes sense to me, but it's not ideal that ct logic is so tightly coupled with policy enforcement.
> 
> I think that's beyond the scope of this PR :). The fact that CT drives policy enforcement/bypass is baked into Cilium pretty deep.
> 

Yes, agreed.

> > Just for the sake of discussion, can it check for an ingress entry, and create one if doesn't exist?
> 
> I suppose it could. I don't see the advantage of doing so - it would just add even more special logic to an already confusing code path. Also merely checking for its existence isn't sufficient for robustness - GC could still kick in 1 nsec later and clear up the CT_INGRESS entry. Beyond that the current CT_INGRESS entry isn't even created properly - note how it accidentally inherits the CT_EGRESS entry's TX statistics and `tx_flags_seen` / `last_tx_report` values. It's just not a robust design 🤷.
> 
> I've mulled over this quite a bit, imho letting the to-container path handle the CT_INGRESS creation "as usual" is the cleanest & robust solution, with the least potential for surprises.
> 

I agree that the more aligned the loopback datapath with regular flow processing is, the better. However, I'm not sure if deferring policy enforcement just for request packets in a flow versus other reply packets would cause any other issues down the road. 

I suppose the fix is good for now. 👍 

Nit: the release note can be improved from end user's perspective. Maybe something along the lines "Fix connectivity issues  caused by missing conntrack entry when service pod connects to itself via clusterIP."

aditighag (approved):

aditighag:
🚀 🚢

bpf/bpf_lxc.c
julianwiedmann: Ah yes, agreed. I'm hopeful that we can also sort out the bypass for CT_REPLY during the 1.15 cycle :crossed_fingers:.

julianwiedmann:
> > > From that perspective, the change makes sense to me, but it's not ideal that ct logic is so tightly coupled with policy enforcement.
> > 
> > 
> > I think that's beyond the scope of this PR :). The fact that CT drives policy enforcement/bypass is baked into Cilium pretty deep.
> 
> Yes, agreed.
> 
> > > Just for the sake of discussion, can it check for an ingress entry, and create one if doesn't exist?
> > 
> > 
> > I suppose it could. I don't see the advantage of doing so - it would just add even more special logic to an already confusing code path. Also merely checking for its existence isn't sufficient for robustness - GC could still kick in 1 nsec later and clear up the CT_INGRESS entry. Beyond that the current CT_INGRESS entry isn't even created properly - note how it accidentally inherits the CT_EGRESS entry's TX statistics and `tx_flags_seen` / `last_tx_report` values. It's just not a robust design 🤷.
> > I've mulled over this quite a bit, imho letting the to-container path handle the CT_INGRESS creation "as usual" is the cleanest & robust solution, with the least potential for surprises.
> 
> I agree that the more aligned the loopback datapath with regular flow processing is, the better. However, I'm not sure if deferring policy enforcement just for request packets in a flow versus other reply packets would cause any other issues down the road.
> 

Agreed, getting this cleaned up completely (to a point where *all* loopback traffic passes through *all* 4 levels of `from-/to-container`) would make me much happier as well.

> 
> Nit: the release note can be improved from end user's perspective. Maybe something along the lines "Fix connectivity issues caused by missing conntrack entry when service pod connects to itself via clusterIP."

That's much better than my attempt, thank you! :+1:

PR #27604, opened 2023-08-21
bpf: lxc: clarify kube-proxy workaround in to-container path

julianwiedmann:
`ipv*_policy()` (the pod ingress policy code) can be called in two ways:
1. as an EP-specific tail-call via `CILIUM_MAP_POLICY` (here `tail_ipv*_policy()` calls the policy code), or
2. when `cil_to_container()` is attached to the veth interface (here `tail_ipv*_to_endpoint()` is the caller)

Lift the kube-proxy workaround into the callers, so we can fine-tune its behaviour. As `cil_to_container()` calls `bpf_clear_meta()`, we can trust that `CB_IFINDEX` is 0 and don't need to include the call to `redirect_ep()` from this path.

This makes it perfectly clear that the `cil_to_container()` path simply lets the packet pass through to the veth peer.

Note: it's *very* likely that we could also remove the `PACKET_HOST` part, but let's not change behaviour for now.

### DIFF
--- a/bpf/bpf_lxc.c
+++ b/bpf/bpf_lxc.c
@@ -1406,8 +1406,7 @@ out:
 #ifdef ENABLE_IPV6
 static __always_inline int
 ipv6_policy(struct __ctx_buff *ctx, int ifindex, __u32 src_label,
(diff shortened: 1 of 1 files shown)


aspsk (approved):
Yes, looks better

PR #27605, opened 2023-08-21
bpf,tests: Change makefile cache to rebuild on header changes

dylandreimerink:
During BPF test development, it is common to make changes in dependencies of the test C file and then re-test. However, because we don't explicitly list header files and don't use them in the traditional way of only containing forward declarations and using other C files for the contents, the makefile doesn't understand it needs to rebuild.

This commit makes clang emit a dependency file for each C file, this dependency file is then fed into `make` so it is aware of all files that contribute to the test C file, this makes it so any changes to any dependent file including .h files will invalidate the `make` cache.

```release-note
Change makefile cache to rebuild on header changes
```

### DIFF
--- a/bpf/tests/Makefile
+++ b/bpf/tests/Makefile
@@ -19,6 +19,8 @@ CLANG_FLAGS += -Wno-unknown-warning-option
 CLANG_FLAGS += -Wno-gnu-variable-sized-type-not-at-end
 CLANG_FLAGS += -Wimplicit-int-conversion -Wenum-conversion
 CLANG_FLAGS += -Wimplicit-fallthrough
+# Create dependency files for each .o file.
+CLANG_FLAGS += -MD
 LLC_FLAGS := -march=bpf
 # Mimics the mcpu values set by cilium-agent. See GetBPFCPU().
 ifeq ("$(KERNEL)","netnext")
@@ -29,10 +31,12 @@ endif
 
 .PHONY: all clean
 
-TEST_OBJECTS = $(patsubst %.c, %.o, $(wildcard *.c)) $(patsubst %c, %i, $(wildcard *.c))
+TEST_OBJECTS = $(patsubst %.c, %.o, $(wildcard *.c))
 
-%.o: %.ll $(LIB)
+%.o: %.ll %.i $(LIB)
 	$(ECHO_CC)
+	# Remove the .o file to force recompilation, only rely on make's caching, not clangs
+	rm -f $@
 	$(QUIET) ${LLC} ${LLC_FLAGS} -filetype=obj -o $@ $<
 
 %.ll: %.c
@@ -49,3 +53,6 @@ clean:
 	rm -f $(wildcard *.ll)
 	rm -f $(wildcard *.o)
 	rm -f $(wildcard *.i)
+	rm -f $(wildcard *.d)
+
+-include $(TEST_OBJECTS:.o=.d)
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.a
 *.so
 *.so.*
+*.d
 
 # LLVM IR files
 *.ll


nebril (approved):

dylandreimerink:
Build team review is in, CI green, marking as ready-to-merge

joestringer (approved):

bpf/tests/Makefile
```diff
@@ -49,3 +53,6 @@ clean:
 	rm -f $(wildcard *.ll)
 	rm -f $(wildcard *.o)
 	rm -f $(wildcard *.i)
+	rm -f $(wildcard *.d)
+
+-include $(TEST_OBJECTS:.o=.d)
```
joestringer, line 58: What is this include statement doing? What's the problem with the previous version of the definition?

dylandreimerink: The `.d` files which is generated by clang contains a list of all files which were used during the previous compilation of the .o file including all header file (.h). They look something like this:

```
some-test.o: some-test.c lib/ip.h bpf/include/something.h
```

So this line then takes all files from `TEST_OBJECTS` and replaces the .o with a .d. And we then include all of these .d files as makefiles into this one.

The effect this has is that it makes `make` aware of all of the .h files that the root .c includes without having to explicitly list all of them in the makefile manually. So any time a .h file is edited calling `make` afterwards will trigger recompilation of all .c files which include it.

> What's the problem with the previous version of the definition?

Before this change, modifying the .h would result in make saying `make: Nothing to be done for 'all'`. You would have to call `make clean` to force recompilation or make a bogus change to the .c file so it would recognize something changed. Leading to everyone being confused why changes in .h files did not change the outcome of tests, until they realize this and then have to manually rebuild every time.

Based on this guide: https://make.mad-scientist.net/papers/advanced-auto-dependency-generation/

joestringer: I see, thanks for the explanation. Does the `include` line need to be prefixed with `-`? This ensures that nonzero exit statuses are ignored, right? But it's not clear to me that this is required, and I don't see it in the guide.

dylandreimerink: Yes, the `-` ignores errors about missing files. The .d files are generated after the initial call to clang, so when compiling the first time no .d files exist yet. In the guide they use the `wildcard` function to avoid this issue, but this works as well.

joestringer:
This seems to me like it just disables all caching. No changes between running the make commands:

```
$ time sh -c 'make -C bpf'
make: Entering directory '/home/joe/git/cilium/bpf'
  CC     bpf/bpf_lxc.ll
  CC     bpf/bpf_lxc.o
  CC     bpf/bpf_overlay.ll
  CC     bpf/bpf_overlay.o
  CC     bpf/bpf_sock.ll
  CC     bpf/bpf_sock.o
  CC     bpf/bpf_host.ll
  CC     bpf/bpf_host.o
  CC     bpf/bpf_xdp.ll
  CC     bpf/bpf_xdp.o
  CC     bpf/bpf_network.ll
  CC     bpf/bpf_alignchecker.ll
  CC     bpf/bpf_network.o
  CC     bpf/bpf_alignchecker.o
make[1]: Entering directory '/home/joe/git/cilium/bpf/custom'
  CC     bpf/custom/bytecount.ll
  CC     bpf/custom/bytecount.o
make[1]: Leaving directory '/home/joe/git/cilium/bpf/custom'
make: Leaving directory '/home/joe/git/cilium/bpf'

real    0m9.682s
user    0m9.167s
sys     0m0.544s
$ time sh -c 'make -C bpf'
make: Entering directory '/home/joe/git/cilium/bpf'
  CC     bpf/bpf_lxc.ll
  CC     bpf/bpf_lxc.o
  CC     bpf/bpf_overlay.ll
  CC     bpf/bpf_overlay.o
  CC     bpf/bpf_sock.ll
  CC     bpf/bpf_sock.o
  CC     bpf/bpf_host.ll
  CC     bpf/bpf_host.o
  CC     bpf/bpf_xdp.ll
  CC     bpf/bpf_xdp.o
make[1]: Entering directory '/home/joe/git/cilium/bpf/custom'
  CC     bpf/custom/bytecount.ll
  CC     bpf/custom/bytecount.o
make[1]: Leaving directory '/home/joe/git/cilium/bpf/custom'
make: Leaving directory '/home/joe/git/cilium/bpf'

real    0m9.409s
user    0m9.004s
sys     0m0.426s
```

bpf/tests/Makefile
joestringer: Hmm, when I try this locally I don't see any `*.d` files.

dylandreimerink:
> This seems to me like it just disables all caching.

You are calling make in the main /bpf dir, not the /bpf/tests dir where I made the change. I did not add this logic to the makefile we use for the normal BPF programs

bpf/tests/Makefile
joestringer: In case it makes a difference:
```
$ clang --version
Ubuntu clang version 14.0.0-1ubuntu1.1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

joestringer:
Ah missed that :facepalm: . I thought `make -C bpf` would compile everything there + subdirs, but evidently not.

I did still hit a weird problem where the caching didn't kick in until the third compile, ie the first two times it took the full time, then it was only cached for the third build. But I guess that's better than what we have today :)

bpf/tests/Makefile
joestringer: ^ PEBKAC, wasn't running make for the right dir.

PR #27607, opened 2023-08-21
bpf: nodeport: constrain CT lookups to relevant entry types

julianwiedmann:
The nodeport code typically only cares about specific types of CT entries
(those that have the .nodeport and/or .dsr flags set). It carries all sorts
of cruft to deal with aliased / stale entries.

As we can't bake the .nodeport / .dsr flags into the actual CT map key
(it's not flexible enough, and would break compatibility), do the next
best thing and teach the low-level CT lookup code about such constraints.
If the caller specifies a set of CT_ENTRY_* types, then the lookup code
will validate the matched entry accordingly.

If the matched entry doesn't match any of the types, it's treated as
CT_NEW. This for instance avoids any unwanted update to the entry's
statistics.

Note that ideally the "generic" CT lookups (bpf_lxc, HostFW, SNAT) would have the opposite behaviour, and exclude the nodeport-specific entry types from their lookup. Future work.

### DIFF
--- a/bpf/lib/conntrack.h
+++ b/bpf/lib/conntrack.h
@@ -29,6 +29,12 @@ enum ct_scope {
 	SCOPE_BIDIR,
 };
 
+enum ct_entry_type {
+	CT_ENTRY_ANY		= 0,
+	CT_ENTRY_NODEPORT	= (1 << 0),
+	CT_ENTRY_DSR		= (1 << 1),
+};
+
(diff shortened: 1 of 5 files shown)


julianwiedmann:
@gentoo-root :wave: I vaguely remember discussing this a while ago. Here is your chance to say that you hate it :grin: .

aspsk (approved):
Looks awesome!

PR #27614, opened 2023-08-21
Introduce resiliency package

derailed:

      GitHub issue.

This package introduces the concept of a retryable error. 
This provides call sites with a hint to potentially retry a failed call if need be.

### DIFF
--- a/pkg/resiliency/error.go
+++ b/pkg/resiliency/error.go
@@ -1,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package resiliency
+
+// RetryableErr tracks errors that could be retried.
+type RetryableErr struct {
+	error
+}
+
+// NewRetryableErr returns a new instance.
+func NewRetryableErr(e error) RetryableErr {
+	return RetryableErr{error: e}
+}
--- a/pkg/resiliency/retry.go
+++ b/pkg/resiliency/retry.go
@@ -1,0 +1,13 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package resiliency
+
+import (
+	"errors"
+)
+
+// IsRetryable checks if an error can be retried.
+func IsRetryable(e error) bool {
+	return errors.As(e, new(RetryableErr))
+}
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -481,6 +481,7 @@ Makefile* @cilium/build
 /pkg/rate/metrics @cilium/metrics
 /pkg/recorder @cilium/sig-datapath
 /pkg/redirectpolicy @cilium/sig-lb
+/pkg/resiliency @cilium/sig-agent
 /pkg/safeio @cilium/sig-agent
 /pkg/serializer @cilium/sig-agent
 /pkg/service @cilium/sig-lb


pkg/resiliency/wrap.go
```diff
@@ -0,0 +1,68 @@
...
+
+// RetryableErr represents an error that can possibly be retried.
+type RetryableErr struct {
+	error
+
+	Kind Category
+}
+
+// Retry checks whether error could be retried.
+// TODO will need m'o tuning..
```
joestringer: What is the intention regarding this TODO?

pkg/resiliency/types.go
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package resiliency
```
joestringer: This new package should probably be owned by some team in `CODEOWNERS`, maybe @cilium/sig-agent?

christarazi:
Just a heads up, there's no need to rebase PRs unless there are conflicting files. GH will warn on the latter. Otherwise, when a PR is merged, it is automatically rebased on top of main. So no need to explicitly rebase PRs periodically unless absolutely necessary.

pkg/resiliency/error.go
```diff
@@ -0,0 +1,73 @@
...
+// NewErr returns a new instance.
+func NewErr(e error, t Tag) taggedErr {
+	return taggedErr{
+		error: e,
+		tag:   t,
+	}
+}
+
+// Error returns a human representation.
+// TODO do we want to decorate at all??
```
joestringer: To kick off the discussion: What will the user do with this information?

pkg/resiliency/error.go
```diff
@@ -0,0 +1,73 @@
...
+func (r taggedErr) Error() string {
+	if r.tag == none {
+		return r.error.Error()
+	}
+
+	return fmt.Sprintf("[%s] %s", r.tag, r.error.Error())
```
joestringer: I wonder if this should actually be a separate attribute in structured logging rather than embedded in the message. I could see a way for that to be filtered by users to then evaluate impact and prioritize which types of errors to do something about.

derailed: Thank you for the input @joestringer!

Not sure?? I was thinking that this would help categorizing errors using tags in conjunction with error codes (later??) 
ie `[resExt: 1001] K8s api server unreachable.`   
We could then document codes with further description, potential resolves, ... to help user get to root cause based on these tag/code combo??
 
It seems best if this tag:code goes along with the error at the point its raises. More often than not, we are going to have multiple errors bubble up to the call site. The current thought is to leverage errors.Join to compile the list (which basically dumps the errors separated with new lines). Hence a single log line could reveal several errors with different tags/codes. Does this make sense?

pkg/resiliency/wrap.go
derailed: this has been removed

pkg/resiliency/types.go
derailed: Good catch!! Tx Joe!

pkg/resiliency/error.go
joestringer: I see, for that style use case I can see why integrating it into the message could be helpful. What I was pondering is if a user might want to filter all errors of a specific type like `resExt` in some external system like splunk, elastic, etc.; having a dedicated field would help in that case. But I guess both points are moot until we widely use this form to present errors to the user.

pkg/resiliency/retry.go
```diff
@@ -0,0 +1,28 @@
...
+func IsRetryable(e error) bool {
+	re := new(RetryableErr)
+
+	errs, ok := e.(interface {
+		Unwrap() []error
+	})
+	if !ok {
+		return errors.As(e, re)
+	}
+	for _, err := range errs.Unwrap() {
```
christarazi: I wonder if we need this at all for a first cut of the package. I'm intending towards a usage pattern where the developer is writing code that errors and they would use `NewRetryableErr` and this error would be returned to the caller which would presumably be responsible for retrying, and if not, then delegating further up the stack by re-wrapping the error again with `NewRetryableErr`.  Eventually, some calling code will kick off the retry process. 

In that pattern, we don't necessarily need `Unwrap` and `errors.As`, but rather simply `errors.Is`.

The reason I'm thinking this is that I can't really imagine a scenario where there's a retryable error thrown deep in the call stack and every layer above it doesn't consider the error retryable.

Posting this to see if it also aligns with others' thinking.

pkg/resiliency/error.go
derailed: Removed!

pkg/resiliency/error.go
derailed: I think this is still a good point as we should make it easier for users to assess the system condition/degradation albeit via logs or perhaps health/metrics for the poor souls that can't afford to store/parse large volumes of logs. 
The original idea, prior to this pkg `complete lobotomy` was to leverage error behaviors vs types to clue the call sites how to best handle the conditions based on `behaviors` ie retry, alert, report, health, log, etc... 

Imho, health/condition signals should intrinsic to our system and not externally derived.

pkg/resiliency/retry.go
derailed: Tx Chris! Good point!! This call is now worthless as behaviors were axed.

christarazi (approved):
The previous revision had a few unit tests which would still be helpful. I think a very simple sanity check unit test is still good to have in the tree. It doesn't need to be complicated because we've removed the need for the `Unwrap`.

derailed:
@christarazi Right I've axe them since I don't think we need to test out the golang apis?

joestringer (approved):

pkg/resiliency/error.go
joestringer: I see. I'm not super strongly opinionated in the above, but thought it's worth raising a discussion either way. At this point the PR doesn't have the related changes, so I think we can resolve out this conversation. If/when we revisit the proposal from this conversation thread, we can continue the discussion.

joestringer:
Looks like this needs a rebase + re-run CI, then it should be good to merge.

PR #27617, opened 2023-08-22
envoy: fix init order between accesslog and xDS server

mhofstetter:
Currently, the envoy accesslog server depends on the xDS server to get information about local endpoints. This leads to the xDS server being initialized and started before the accesslog server. Therefore, the accesslog server might not be ready when Envoy is receiving the xDS resources and starts to initialize Cilium components in envoy. This results in silent errors of the form `Cilium filter can not open access log socket` (e.g. [Cilium L7 policy filter](https://github.com/cilium/proxy/blob/f4eb004fd81e38ad05d4014fd5f5ff8e9bbf6fe8/cilium/l7policy.cc#L57)).

This commit introduces the LocalEndPointStore which contains this information and is shared between the two components. The direct dependency between the accesslog server and xDS server can be replaced - and the components start initializing at the same time.

In addition, the dependency from the xDS server to the accesslog server gets introduced.

This dependency only serves the purpose of enforcing the accesslog server being initialized first before the xDS server is starting up.

### DIFF
--- a/pkg/envoy/accesslog_server.go
+++ b/pkg/envoy/accesslog_server.go
@@ -27,15 +27,15 @@ import (
 )
 
 type AccessLogServer struct {
-	socketPath string
-	xdsServer  XDSServer
-	stopCh     chan struct{}
+	socketPath         string
+	localEndpointStore *LocalEndpointStore
+	stopCh             chan struct{}
 }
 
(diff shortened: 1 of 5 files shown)


jrajahalme (approved):
Thank you for this fix :-)

PR #27618, opened 2023-08-22
cilium: Add option to masq to source route

borkmann:
See commit msgs.

### DIFF
--- a/Documentation/network/concepts/masquerading.rst
+++ b/Documentation/network/concepts/masquerading.rst
@@ -166,3 +166,11 @@ option ``egress-masquerade-interfaces: eth0`` can be used.
    It is possible to specify an interface prefix as well, by specifying
    ``eth+``, all interfaces matching the prefix ``eth`` will be used for
    masquerading.
+
+For the advanced case where the routing layer would select different source addresses
+depending on the destination CIDR, the option ``enable-masquerade-to-route-source: "true"``
+can be used in order to masquerade to the source addresses rather than to the
+primary interface address. The latter is then only considered as a catch-all
+fallback, and for the default routes. For these advanced cases the user needs
+to ensure that there are no overlapping destination CIDRs as routes on the
+relevant masquerading interfaces.
--- a/pkg/datapath/iptables/iptables.go
+++ b/pkg/datapath/iptables/iptables.go
@@ -1198,11 +1198,11 @@ func (m *IptablesManager) installMasqueradeRules(prog iptablesInterface, ifName,
 			"-A", ciliumPostNatChain,
 		}
 
-		// If EgressMasqueradeInterfaces is set, we need to mirror base condition
-		// of the "cilium masquerade non-cluster" rule below, as the allocRange might
-		// not be valid in such setups (e.g. in ENI mode).
-		if option.Config.EgressMasqueradeInterfaces != "" {
-			progArgs = append(progArgs, "-o", option.Config.EgressMasqueradeInterfaces)
+		// If MasqueradeInterfaces is set, we need to mirror base condition of the
+		// "cilium masquerade non-cluster" rule below, as the allocRange might not
+		// be valid in such setups (e.g. in ENI mode).
+		if len(option.Config.MasqueradeInterfaces) > 0 {
+			progArgs = append(progArgs, "-o", strings.Join(option.Config.MasqueradeInterfaces, ","))
 		} else {
 			progArgs = append(progArgs, "-s", allocRange)
 		}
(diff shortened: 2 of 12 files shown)


kaworu (approved):
Helm changes LGTM.

Documentation/network/concepts/masquerading.rst
```diff
@@ -166,3 +166,11 @@ option ``egress-masquerade-interfaces: eth0`` can be used.
...
    ``eth+``, all interfaces matching the prefix ``eth`` will be used for
    masquerading.
+
+For the advanced case where the routing layer would select different source addresses
+depending on the destination CIDR, the option ``enable-masquerade-to-route-source: "true"``
+can be used in order to masquerade to the source addresses rather than to the
+primary interface address. The latter is then only considered as a catch-all
+fallback, and for the default routes. For these advanced cases the user needs
+to ensure that there are no overlapping destination CIDRs as routes on the
+relevant masquerading interfaces.
```
brb, line 176: Nit: It'd be nice to add example (from the commit msg) to illustrate the case.

pkg/datapath/iptables/iptables.go
```diff
@@ -1217,7 +1219,77 @@ func (m *IptablesManager) installMasqueradeRules(prog iptablesInterface, ifName,
...
 	}
 
-	// Masquerade all egress traffic leaving the node
+	// Masquerade egress traffic leaving the node based on source routing
+	//
+	// If this option is enabled, then it takes precedence over the catch-all
+	// MASQUERADE further below.
+	if option.Config.EnableMasqueradeRouteSource {
+		devices := option.Config.GetDevices()
+		if option.Config.EgressMasqueradeInterfaces != "" {
```
brb: Maybe we could change the `EgressMasqueradeInterfaces` flag type to `StringSlice`? Then the split below would not be needed.

pkg/datapath/iptables/iptables.go
```diff
@@ -1217,7 +1219,77 @@ func (m *IptablesManager) installMasqueradeRules(prog iptablesInterface, ifName,
...
+			devices = strings.Split(option.Config.EgressMasqueradeInterfaces, ",")
+		}
+		family := netlink.FAMILY_V4
+		if prog == ip6tables {
+			family = netlink.FAMILY_V6
+		}
+		if option.Config.EnableIPv4 && prog == ip4tables ||
```
brb: I think we don't need to check for `EnableIPv{4,6}` here, as the callers of the `installMasqueradeRules()` already do the check.

pkg/datapath/iptables/iptables.go
```diff
@@ -1217,7 +1219,77 @@ func (m *IptablesManager) installMasqueradeRules(prog iptablesInterface, ifName,
 		}
 	}
 
-	// Masquerade all egress traffic leaving the node
+	// Masquerade egress traffic leaving the node based on source routing
+	//
+	// If this option is enabled, then it takes precedence over the catch-all
+	// MASQUERADE further below.
+	if option.Config.EnableMasqueradeRouteSource {
+		devices := option.Config.GetDevices()
```
brb, line 1227: Do we actually want to fallback to `--devices` if `--egress-masquerade-interfaces` is not set? Asking, as AFAIK `--devices` are only used to control the NodePort BPF and the BPF masq, while previously  the iptables-based MASQ didn't consider them.

brb (commented):
Nice work! A few comments.

giorio94 (approved):
sig-agent changes LGTM

pkg/datapath/iptables/iptables.go
borkmann: good point, agree. fixed

pkg/datapath/iptables/iptables.go
borkmann: Makes sense, in prior cases we just use the plain EgressMasqueradeInterfaces string, but I guess it makes sense to switch to slice and have it the other way around.

pkg/datapath/iptables/iptables.go
borkmann: The source selection can be tied to devices unless next hop groups are used, so both yes & no with regards to their need. I could bail out with error if r.LinkIndex > 0 and len(devices) is 0.

pkg/datapath/iptables/iptables.go
borkmann: Ok, done.

borkmann: Looks like I'm running into https://github.com/cilium/cilium-cli/issues/1896 where CI breaks given cilium-cli vendors in cilium itself :/

pkg/datapath/iptables/iptables.go
borkmann: Implemented the above now.

pkg/datapath/iptables/iptables.go
borkmann: Tobias gave some hints how to work-around this by keeping the old EnableMasqueradeRouteSource alive as string.

Issue #27318, opened 2023-08-07
CI: integration-test: Test/IPAMSuite/TestIpamManyNodes


learnitall:
## CI failure

Hit on #26327: https://github.com/cilium/cilium/actions/runs/5764670061/job/15629127498

```
    ipam_test.go:[346](https://github.com/cilium/cilium/actions/runs/5764670061/job/15629127498#step:8:347): 
        ... obtained int = 960
        ... expected int = 1000
```

PR #27621, opened 2023-08-22
chart: optimize spire image for offline environment

weizhoublue:
optimize the chart values for spire

* align the image definition map with other images, such as cilium operator and hubble, which decouple the repository and tag, and could better deploy in an offline environment

* prevent hardcode the busybox image in the yaml, which could help deploy cilium in an offline K8S cluster 

```release-note
Change the Helm values configuration for SPIRE to match other images in the Helm charts
```


joestringer:
Heads up @mhofstetter this will have user-facing impact due to the change in the format for Helm values. Hence I've updated `release-note/minor`. This change should also be documented in the upgrade guide.

weizhoublue:
> Heads up @mhofstetter this will have user-facing impact due to the change in the format for Helm values. Hence I've updated `release-note/minor`. This change should also be documented in the upgrade guide.

OK, I will add 'Upgrade Notes' in the upgrade guide ,  it is documentation/operations/upgrade.rst,  right ?

Documentation/operations/upgrade.rst
```diff
@@ -340,6 +340,9 @@ Helm Options
   were deprecated in Cilium 1.14 in favor of ``tls.ca.cert`` and ``tls.ca.key`` respectively,
   and have been removed. The ```clustermesh-apiserver-ca-cert`` secret is no longer generated.
 
+* Values ``authentication.mutual.spire.install.agent.image`` and ``authentication.mutual.spire.install.server.image``
+  in Cilium 1.14 is changed to decouple the repository and tag, which could better deploy in an offline environment.
```
mhofstetter: these release notes already target v1.15 of cilium - and no longer 1.14.

I wouldn't mention the version at all - because it should already be clear from the context. I would rather focus on describing the actual type-change of the values

-> `Values ``authentication.mutual.spire.install.agent.image`` and ``authentication.mutual.spire.install.server.image``
  changed their type from a string to a structured definition that decouples repository and tag. This improves the usage in offline environments.`

@zacharysarah wdyt?

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -3150,10 +3150,24 @@ authentication:
...
+          repository: "docker.io/library/busybox"
+          tag: "1.35.0"
+          digest: "sha256:223ae047b1065bd069aac01ae3ac8088b3ca4a527827e283b85112f29385fb1b"
+          useDigest: true
+          pullPolicy: "Always"
         # SPIRE agent configuration
         agent:
           # -- SPIRE agent image
-          image: ghcr.io/spiffe/spire-agent:1.6.3@sha256:8eef9857bf223181ecef10d9bbcd2f7838f3689e9bd2445bede35066a732e823
+          image:
```
mhofstetter, line 3177: Changing the type and re-using the name of the existing values is a breaking change as we do not give users a deprecation period.

Deprecating the old value and introducing a new one (with a different name) is what we normally do. But i agree that we should use the property `image` to be consistent with other definitions of images.

@kaworu what's the recommended approach in such cases from a sig-helm perspective? Would the breaking change be ok for v1.15 (without backporting it to v1.14)?

mhofstetter (changes requested):
@weizhoublue thanks for the contribution! The change request itself looks reasonable. 

But as@joestringer already pointed out - i have some questions about the breaking change (vs deprecation period) and the respective communication in the upgrade notes for the respective SIGs.

zacharysarah (approved):
LGTM

install/kubernetes/cilium/values.yaml.tmpl
weizhoublue: indeed, I also hesitated to change a new name inconsistent with other definitions of images

Documentation/operations/upgrade.rst
weizhoublue: agree, and change to your suggestion

install/kubernetes/cilium/values.yaml.tmpl
kaworu: > what's the recommended approach in such cases from a sig-helm perspective?

Generally speaking, when possible: go through a deprecation stage introducing the alternative for the next version n and document in upgrade.rst, then remove the deprecated option in n+1.

Here it's tricky because we want both the new and old values under the same `image:` key, but we could check in Helm the `typeOf` `image` and use the "old" or "new" behaviour when it's a string or an object, respectively.

> Would the breaking change be ok for v1.15 (without backporting it to v1.14)?

Since this block is about Mutual Authentication configuration, and given [Mutual Authentication is in Beta maturity stage](https://docs.cilium.io/en/latest/network/servicemesh/mutual-authentication/mutual-authentication/), I think it's fine to break it for v1.15.

mhofstetter: @kaworu Nice - i wasn't aware of the `typeOf` function in Helm. Thanks!

@weizhoublue Given this knowledge, I'd recommend to introduce this change backwards-compatible.

Something like this should work.
```
{{- if eq (typeOf .Values.authentication.mutual.spire.install.agent.image) "string" }}
image: {{ .Values.authentication.mutual.spire.install.agent.image }}
{{- else }}
image: {{ include "cilium.image" .Values.authentication.mutual.spire.install.agent.image | quote }}
imagePullPolicy: {{ .Values.authentication.mutual.spire.install.agent.image.pullPolicy }}
{{- end }}
```

At least a quick local test seems to work. The only thing that irritates me a little bit is the output when "overriding" the value with type `string`

```
coalesce.go:220: warning: cannot overwrite table with non table for cilium.authentication.mutual.spire.install.agent.image
```

at least when executing the Helm template command like this (on the PR branch)

```
helm template install/kubernetes/cilium --set authentication.mutual.spire.enabled=true --set authentication.mutual.spire.install.enabled=true --set authentication.mutual.spire.install.agent.image=ghcr.io/spiffe/spire-agent:1.6.3@sha256:8eef9857bf223181ecef10d9bbcd2f7838f3689e9bd2445bede35066a732e823
```

:thinking:

Documentation/operations/upgrade.rst
mhofstetter: @weizhoublue Thanks - looks like you missed the last dot in the sentence.

weizhoublue: thanks

install/kubernetes/cilium/values.yaml.tmpl
weizhoublue: OK, it would be a graceful manner for backwards-compatible

mhofstetter (approved):
LGTM - thanks again!

kaworu (approved):

install/kubernetes/cilium/templates/spire/server/statefulset.yaml
```diff
@@ -36,14 +36,20 @@ spec:
...
         command:
           - /bin/sh
           - -c
           - |
             {{- tpl (.Files.Get "files/spire/init.bash") . | nindent 12 }}
       - name: spire-server
+        {{- if eq (typeOf .Values.authentication.mutual.spire.install.server.image) "string" }}
         image: {{ .Values.authentication.mutual.spire.install.server.image }}
+        {{- else }}
+        image: {{ include "cilium.image" .Values.authentication.mutual.spire.install.server.image | quote }}
```
joestringer, line 50: What does the "cilium.image" include bit do here exactly?

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -3150,10 +3150,24 @@ authentication:
...
+            override: ~
+            repository: "ghcr.io/spiffe/spire-agent"
+            tag: "1.6.3"
+            digest: "sha256:8eef9857bf223181ecef10d9bbcd2f7838f3689e9bd2445bede35066a732e823"
```
joestringer: Usually for these external dependencies, we would define them as exports in `install/kubernetes/Makefile.values` in order to simplify the updates, then in the `install/kubernetes/cilium/values.yaml.tmpl` file we have variables. You can scan up and look at examples around `HUBBLE_UI_BACKEND_REPO` (plus grep for this symbol in the codebase) to get a sense for how these are plumbed through.

I'd suggest using variable prefixes like the following and use the exact same pattern as other code in the template file uses:

- SPIRE_INIT_* for the busybox images,
- SPIRE_AGENT_* for the agent images,
- SPIRE_SERVER_* for the server images.

joestringer (changes requested):
Thanks for the submission! I had one question on the templates for something I don't understand, plus a request to make the image variables definitions to be lined up the same way as other external dependencies in the tree, using `install/kubernetes/Makefile.values`.

install/kubernetes/cilium/templates/spire/agent/daemonset.yaml
```diff
@@ -29,15 +29,21 @@ spec:
...
+          image: {{ include "cilium.image" .Values.authentication.mutual.spire.install.initImage | quote }}
+          imagePullPolicy: {{ .Values.authentication.mutual.spire.install.initImage.pullPolicy }}
           command:
             - /bin/sh
             - -c
             - |
               {{- tpl (.Files.Get "files/spire/wait-for-spire.bash") . | nindent 14 }}
       containers:
         - name: spire-agent
+          {{- if eq (typeOf .Values.authentication.mutual.spire.install.agent.image) "string" }}
```
tommyp1ckles, line 41: The PR descriptions making this an "image" and "tag" value, similar to other components?

install/kubernetes/cilium/values.yaml.tmpl
weizhoublue: sure, thanks

install/kubernetes/cilium/templates/spire/agent/daemonset.yaml
weizhoublue: yes
the common way of cilium images use "tag" and "repository" things to define the image
I think it conforms to best practices, especially for offline environment, the administrator could just set helm option --xx. repository=offline.io/cilium , and is not necessary to care about the "tag" for cilium v1.4 or v1.5, and just keep the best matched and recommended tag in the chart

Therefore, the original definition of spire image does not conform what other cilium image does, and that is what this PR want to optimize

install/kubernetes/cilium/templates/spire/server/statefulset.yaml
mhofstetter: Calling / Including the Helm template [cilium.image](https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/templates/_helpers.tpl#L8-L27). 

> Render full image name from given values, e.g:
> ```
> image:
>   repository: quay.io/cilium/cilium
>   tag: v1.10.1
>   useDigest: true
>   digest: abcdefgh
> ```
> then `include "cilium.image" .Values.image`
> will return `quay.io/cilium/cilium:v1.10.1@abcdefgh`

install/kubernetes/cilium/templates/spire/agent/daemonset.yaml
weizhoublue: and why it is compatible with the string type here , is owing to [comment](https://github.com/cilium/cilium/pull/27621#discussion_r1303870784)

install/kubernetes/cilium/templates/spire/server/statefulset.yaml
joestringer: I'm a bit confused by that. Why would someone use the Cilium image for the SPIRE server statefulset?

install/kubernetes/cilium/templates/spire/agent/daemonset.yaml
joestringer: Sorry, some of the confusion may be because I added the release note. I will say though that these settings should conform to the same configuration format that is exposed for components like hubble-ui.

install/kubernetes/cilium/templates/spire/server/statefulset.yaml
weizhoublue: I guess your point is why it is not 'spire.image' or 'xxx.image' 

actually, the `cilium.image` is suitable for all images with following statement, it mainly care about the structure, but does not care about what project the image belongs to. I think it means how to generate an image string by cilium manner, but does not mean generating cilium's image. Currently in the chart code, it is also used to render all images including etcd and envoy image which is not belonged to cilium.

```
{{/*
Render full image name from given values, e.g:
image:
  repository: quay.io/cilium/cilium
  tag: v1.10.1
  useDigest: true
  digest: abcdefgh
then `include "cilium.image" .Values.image`
will return `quay.io/cilium/cilium:v1.10.1@abcdefgh`
*/}}
{{- define "cilium.image" -}}
{{- $digest := (.useDigest | default false) | ternary (printf "@%s" .digest) "" -}}
{{- if .override -}}
{{- printf "%s" .override -}}
{{- else -}}
{{- printf "%s:%s%s" .repository .tag $digest -}}
{{- end -}}
{{- end -}}
```

My idea is ok to use 'cilium.image' here. And it is also ok to define 'xxx.image' with same function code, or change the 'cilium.image' to a common name

joestringer: Oh I see. `cilium.image` at face value sounds like an actual string instance of the Cilium image and not the template. We can probably improve the naming here to make it clear that this is an image template evaluation statement, but that's out-of-scope for this PR.

joestringer (approved):
Docs look good and my earlier comments were resolved. I didn't take a fresh look at the Helm format, delegating review to @cilium/helm for that.

install/kubernetes/cilium/templates/spire/server/statefulset.yaml
weizhoublue: yes, I would like to help rename `cilium.image` to be not misunderstood after the PR

weizhoublue:
the CI failed owing to some [connectivity](https://github.com/cilium/cilium/actions/runs/5986189734/job/16239052309) and [ipam](https://github.com/cilium/cilium/actions/runs/5986189895/job/16239075520) tests, which seem to have no bussiness with PR.
Should I retry to rebase main branch

joestringer:
> the CI failed owing to some [connectivity](https://github.com/cilium/cilium/actions/runs/5986189734/job/16239052309) and [ipam](https://github.com/cilium/cilium/actions/runs/5986189895/job/16239075520) tests, which seems to have no bussiness with PR.

Known failures #27672 and #27318.

aditighag:
@weizhoublue Hi, Could you rebase the PR as there is a conflict? Thanks!

weizhoublue:
> @weizhoublue Hi, Could you rebase the PR as there is a conflict? Thanks!

sure, thanks for reminding

PhilipSchmid:
Can we please get this one backported to 1.14?

PR #27622, opened 2023-08-22
cilium: Fix 16bit ifindex limitation

### LINKED ISSUE #16260: datapath: Support devices for bpf_host which ifindex > max(u16)
Currently, we disallow devices which ifindex cannot fit in u16. This is because we store ifindex in the CT entry struct which only had the u16 spare field left (https://github.com/cilium/cilium/pull/11267/commits/7282119184a4221c2a61f5b93d6889e8558705be#diff-78122dbc1bf6a53a08f02c035e071f869eaa123b52822d789b79c04ff1cbdf69R550).

To overcome the limitation, we can introduce an intermediate mapping for u32(ifindex) => u16(surrogate_ifindex). In the CT entry we would store surrogate_ifindex, and in the datapath we would use a generated macro function by the agent to derive the real ifindex (similar as we do today with e.g. `NATIVE_DEV_MAC_BY_IFINDEX`).

Extra care should be taken for considering upgrade / downgrade.

borkmann:
See details in commit desc.

Fixes: #16260


julianwiedmann:
note that you will need a rebase for picking up https://github.com/cilium/cilium/pull/27592 to fix CI

bpf/lib/conntrack.h
```diff
@@ -246,7 +246,8 @@ __ct_lookup(const void *map, struct __ctx_buff *ctx, const void *tuple,
 			ct_state->proxy_redirect = entry->proxy_redirect;
 			ct_state->from_l7lb = entry->from_l7lb;
 			ct_state->from_tunnel = entry->from_tunnel;
-			ct_state->ifindex = entry->ifindex;
+			if (!is_defined(HAVE_FIB_IFINDEX))
+				ct_state->ifindex = entry->ifindex;
```
julianwiedmann: Let's also wrap the  `ct_state->ifindex` users in `nodeport_rev_dnat_ingress_ipv*()` ? Otherwise this looks like the `ifindex` wasn't even used anymore :).

And maybe also the `ifindex` field in the `ct_state` itself, similar as for `DISABLE_LOOPBACK_LB`.

julianwiedmann (changes requested):
lgtm, thank you! Needs just one small change I think (see review below).

As follow-on, let's also add `HAVE_FIB_IFINDEX` to the relevant complexity tests to reflect the real world.

> Apart from fixing the bleed, this can be further refined by not relying
on the asm.FnRedirectPeer helper presence but by actually doing a runtime
BPF program probe so that stable kernels can even be covered.

We can further also check for `ENABLE_SKIP_FIB`.

julianwiedmann:
> We can further also check for `ENABLE_SKIP_FIB`.

ah, maybe not :thinking: . That's a runtime thing - so if one changes their config to remove `ENABLE_SKIP_FIB`, they suddenly require the `ifindex` of existing CT entries to be populated.

borkmann:
> As follow-on, let's also add `HAVE_FIB_IFINDEX` to the relevant complexity tests to reflect the real world.

Ah yes, I'll add it.

borkmann:
> > We can further also check for `ENABLE_SKIP_FIB`.
> 
> ah, maybe not 🤔 . That's a runtime thing - so if one changes their config to remove `ENABLE_SKIP_FIB`, they suddenly require the `ifindex` of existing CT entries to be populated.

Yeah, I'd avoid combining the two.

bpf/lib/conntrack.h
borkmann: Agree, good point. ct_state is not part of map, so this is doable.

borkmann:
(checkpatch complaint unrelated given this refers to kernel commit not cilium one)

julianwiedmann (approved):
looks good, ty!

ti-mo (approved):
Thanks!

tklauser:
@dborkmann I've tried to backport this PR to `v1.14` as part of the backporter rotation but hit some non-trivial merge conflicts in `bpf/`. Would you be ok to backport this yourself given you're more familiar with the code in question? Thanks

joestringer:
v1.14 backport makes sense, does this also affect older releases?

borkmann:
> v1.14 backport makes sense, does this also affect older releases?

It does, there it would have to be even more custom patch for older stable releases given they don't have https://github.com/cilium/cilium/pull/23884 (only 1.14).

julianwiedmann:
FYI, I opened https://github.com/cilium/cilium/issues/27642 to cover the fine-tuning of the feature probe.

PR #27623, opened 2023-08-22
vendor: downgrade github.com/shirou/gopsutil/v3 to v3.23.2

aanm:
Until github.com/shoenig/go-m1cpu is added as an exception in the list of libraries that we can use, we need to downgrade this dependency down to the version that it didn't have the "github.com/shoenig/go-m1cpu" as part of it.


borkmann (approved):

tklauser (changes requested):
I think we'll need to downgrade gops here too: https://github.com/cilium/cilium/blob/e6a7df47243f03144124c4f49537bd6f7d18afc1/images/runtime/build-gops.sh#L12

Otherwise the version will mismatch between `gops` CLI and the gops agent in Cilium binaries.

tklauser (approved):

sayboras (approved):

tgraf:
After further investigation, the conclusion was reached that the whole use of `shirou/gopsutil/v3` will be replaced with a smaller dependency or native code to retrieve load, CPU, and memory information on the node. This is only used for debugging perspective and the dependency chain it involves is not worth the value derived from it.

Issue to track replacement:
https://github.com/cilium/cilium/issues/27635

PR #27627, opened 2023-08-22
typo in the debug document

weizhoublue:
there is no secret named 'clustermesh-secrets' , it should be  cilium-clustermesh

```
[root@master10 ~]# kubectl get secrets -n kube-system cilium-clustermesh
NAME                 TYPE     DATA   AGE
cilium-clustermesh   Opaque   1      39m

[root@master10 ~]# kubectl get secrets -n kube-system clustermesh-secrets
Error from server (NotFound): secrets "clustermesh-secrets" not found

```

actually 'clustermesh-secrets' is the name of the volume mounting the secret
```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: cilium
    .......
        - name: clustermesh-secrets
          mountPath: /var/lib/cilium/clustermesh
          readOnly: true
      ........
        # To read the clustermesh configuration
      - name: clustermesh-secrets
        projected:
          # note: the leading zero means this number is in octal representation: do not remove it
          defaultMode: 0400
          sources:
          - secret:
              name: cilium-clustermesh
              optional: true
              # note: items are not explicitly listed here, since the entries of this secret
              # depend on the peers configured, and that would cause a restart of all agents
              # at every addition/removal. Leaving the field empty makes each secret entry
              # to be automatically projected into the volume as a file whose name is the key.
```

### DIFF
--- a/Documentation/operations/troubleshooting_clustermesh.rst
+++ b/Documentation/operations/troubleshooting_clustermesh.rst
@@ -76,7 +76,7 @@ Manual Verification of Setup
 
     If the configuration is not found, check the following:
 
-    * The Kubernetes secret ``clustermesh-secrets`` is imported correctly.
+    * The Kubernetes secret ``cilium-clustermesh`` is imported correctly.
 
     * The secret contains a file for each remote cluster with the filename
       matching the name of the remote cluster.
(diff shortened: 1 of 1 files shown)


zacharysarah (approved):
@weizhoublue Nice fix. 🌟 LGTM

PR #27628, opened 2023-08-22
Replace StateDB

dylandreimerink:
This PR replaces the usage of StateDB with StateDB2 in L2 announcement and device controller / device manager.

```release-note
Replace StateDB with StateDB2
```


pkg/datapath/l2responder/l2responder.go
```diff
@@ -83,136 +82,100 @@ func (p *l2ResponderReconciler) run(ctx context.Context) error {
...
 	ticker := time.NewTicker(5 * time.Minute)
 
+	tbl := p.params.L2AnnouncementTable
+	txn := p.params.StateDB.WriteTxn(tbl)
+	tracker, err := tbl.DeleteTracker(txn, "l2-responder-reconciler")
+	if err != nil {
+		txn.Abort()
+		return fmt.Errorf("delete tracker: %w", err)
+	}
+	txn.Commit()
```
giorio94, line 92: Possible follow-up item. This logic looks common enough that could be abstracted in the statedb package.

pkg/l2announcer/l2announcer.go
```diff
@@ -835,22 +835,18 @@ func (l2a *L2Announcer) processLeaderEvent(event leaderElectionEvent) error {
 }
 
 func (l2a *L2Announcer) recalculateL2EntriesTableEntries(ss *selectedService) error {
-	txn := l2a.params.StateDB.WriteTxn()
-	w := l2a.params.L2AnnounceTable.Writer(txn)
+	tbl := l2a.params.L2AnnounceTable
+	txn := l2a.params.StateDB.WriteTxn(tbl)
+	defer txn.Abort()
```
giorio94, line 840: IIUC abort will cause a metric update even in case a commit already happened (while the same metric is not updated on `Commit()`.

giorio94 (approved):
I don't have much knowledge about the l2responder logic, but sig-agent changes look sensible to me.

pkg/l2announcer/l2announcer.go
joamaki: Fixed this one!

pkg/l2announcer/l2announcer_test.go
```diff
@@ -474,119 +467,17 @@ func TestPolicyRedundancy(t *testing.T) {
...
-	})
-
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-	fix.announcer.jobgroup.Stop(ctx)
-	cancel()
-}
-
-// Test that when the last policy selecting a service is deleted, the selected service and proxy neighbor entries go
-// away.
-func TestPolicySoftDeleteService(t *testing.T) {
```
joestringer, line 497: Do we need to revisit this test after this PR?

pkg/l2announcer/l2announcer_test.go
```diff
@@ -338,19 +335,16 @@ func TestHappyPathPermutations(t *testing.T) {
...
-					IP:               net.ParseIP(blueService().Spec.ExternalIPs[0]),
-					NetworkInterface: bluePolicy().Spec.Interfaces[0],
-					Origins:          []resource.Key{serviceKey(blueService())},
-					Deleted:          false,
-					Revision:         1,
+					L2AnnounceKey: tables.L2AnnounceKey{
+						IP:               netip.MustParseAddr(blueService().Spec.ExternalIPs[0]),
+						NetworkInterface: bluePolicy().Spec.Interfaces[0],
+					},
+					Origins: []resource.Key{serviceKey(blueService())},
```
joestringer, line 347: We may want to have a bit of a discussion around this style of testing if changing the structure of data requires changing tests. I realize it's somewhat in the vein of "testing the translation from one table to another", but I think it's usually a weakness in the tests if changing the target code requires reworking the various structure declarations in the test code.

joestringer (approved):
I briefly perused everything in my codeowners except `l2responder`. Approving for expedient removal of MPL-licensed code.

pkg/l2announcer/l2announcer_test.go
dylandreimerink: We shouldn't have to. The old statedb had no way of tracking deletes so we had a custom soft delete mechanism which is just gone now, so we don't specifically have to test it anymore.

pkg/l2announcer/l2announcer_test.go
dylandreimerink: I agree with the shortcomings of this method, but not really sure how I would change things. Do you suggest some sort of constructor or is there some other technique? Looking forward to that discussion.

PR #27636, opened 2023-08-22
feat: Add annotation propagation from Gateway to Load Balancer service

### LINKED ISSUE #25357: Support Use of Annotations on Cilium Gateway API Load Balancer Services

### What happened?

1. using Linode's infrastrucuture
2. create a Kubernetes cluster with cilium as CNI & Gateway Class
3. Create HTTP Routes & add Cert Manager
4. Traffic needed to validate the HTTP Cert Request fail / timeout
```Failed to connect to hostname.example.com port 80 after 4 ms: Couldn't connect to server```
5. this is happening because the load-balencer service needs the following annotation for linode LB / pod routing
```service.beta.kubernetes.io/linode-loadbalancer-hostname-only-ingress: "true"```
6. If I manually edit the `cilium-gateway-service` which generates the LB - the request will work for a moment and then fail as the gateway controller removes the annotations.

is it possible to have the controller to not remove the annotations

- and/or -

the ability to supply the needed service annotations :)

### Cilium Version

v1.13.0

### Kernel Version

UBUNTU 22.04
Linux dev-master-0 5.15.0-60-generic #66-Ubuntu SMP Fri Jan 20 14:29:49 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

### Kubernetes Version

v1.26.3

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

### PR DESCRIPTION
**Pull request by sergeyshevch**
Now annotations from Gateway prefixed by prefixes defined in config will be propagated to Service. I followed actual ingress controller implementation for such feature and made it similar way

      GitHub issue.
- [X] Are you a user of Cilium? Please add yourself to the [Users doc](https://github.com/cilium/cilium/blob/main/USERS.md)

Fixes: #25357

```release-note
Annotations prefixed by `gateway-lb-annotation-prefixes` flag value will be propagated to underlying Load Balancer Service
```

### DIFF
--- a/install/kubernetes/cilium/templates/cilium-configmap.yaml
+++ b/install/kubernetes/cilium/templates/cilium-configmap.yaml
@@ -256,6 +256,8 @@ data:
   enable-gateway-api: "true"
   enable-gateway-api-secrets-sync: {{ .Values.gatewayAPI.secretsNamespace.sync | quote }}
   gateway-api-secrets-namespace: {{ .Values.gatewayAPI.secretsNamespace.name | quote }}
+  ingress-lb-annotation-prefixes: {{ .Values.gatewayAPI.gatewayLBAnnotationPrefixes | join " " | quote }}
+
 {{- end }}
 
 {{- if hasKey .Values "loadBalancer" }}
--- a/install/kubernetes/cilium/values.yaml
+++ b/install/kubernetes/cilium/values.yaml
@@ -768,6 +768,10 @@ gatewayAPI:
     # If disabled, TLS secrets must be maintained externally.
     sync: true
 
+  # -- GatewayLBAnnotations are the annotation prefixes, which are used to filter annotations to propagate
+  # from Gateway to the Load Balancer service
+  gatewayLBAnnotationPrefixes: [ 'service.beta.kubernetes.io', 'service.kubernetes.io', 'cloud.google.com' ]
+
 # -- Enables the fallback compatibility solution for when the xt_socket kernel
 # module is missing and it is needed for the datapath L7 redirection to work
 # properly. See documentation for details on when this can be disabled:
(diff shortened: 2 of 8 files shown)


sergeyshevch:
I'm not sure where to update documentation about annotation propagation. I guess I need to reflect it in similar way as in Ingress controller annotations page


```
Additionally, cloud-provider specific annotations for the LoadBalancer service
are supported.

By default, annotations with values beginning with:

* ``service.beta.kubernetes.io``
* ``service.kubernetes.io``
* ``cloud.google.com``

will be copied from an Ingress object to the generated LoadBalancer service objects.

This setting is controlled by the Cilium Operator's ``gateway-lb-annotation-prefixes``
config flag, and can be configured in Cilium's Helm ``values.yaml``
using the ``gatewayAPI.gatewayLBAnnotationPrefixes`` setting.

Please refer to the `Kubernetes documentation <https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer>`_
for more details.
```

youngnick:
Thanks for this PR, @sergeyshevch, but I'm afraid that we can't accept it - if we add annotation copying for Gateway now, it will stick around forever, even once we build something better. For this sort of thing, annotation copying is a hack that was necessary for Ingress, but now that we have structured config available in Gateway API, we can do this better.

I'm literally working upstream at the moment to try and get this finished, see https://github.com/kubernetes-sigs/gateway-api/pull/1757 for the GEP work.

youngnick:
As a result, I'm going to close this one. Thanks again for the effort.

sergeyshevch:
That's quite sad because the lack of annotation propagations produces ugly hacks with mutation policies to set up the gateway correctly.

On the other hand, I totally support you in the upstream work and it's a better approach to make this feature.

Maybe we can implement it in any way now while looking forward to upstream Gateway API spec and later switch our implementation to structured config

youngnick:
I'm working on getting it to implementable before Gateway API GA - which will be in late October. I understand having to wait sucks, but the alternative is that we do something short-term now, and then spend years moving everyone to the better solution, which also sucks.

PR #27638, opened 2023-08-22
pkg/cidr: Move linux specific variable references from netlink

aditighag:
Commit e731cbe27e added references to linux specific consts from the netlink package, and it broke the ginkgo compilation on darwin targets - https://github.com/cilium/cilium/actions/runs/5944152105/job/16120804454?pr=27520. 

/cc @borkmann 

Reported-by: Tim Horner <timothy.horner@isovalent.com>
Fixes: e731cbe27e "cilium, iptables: Support option to masquerade source IP from routing layer"

### DIFF
--- a/.github/workflows/lint-build-commits.yaml
+++ b/.github/workflows/lint-build-commits.yaml
@@ -96,6 +96,7 @@ jobs:
         with:
           filters: |
             src:
+              - 'pkg/**'
               - 'test/**'
 
       # Runs only if code under test/ is changed.
--- a/pkg/cidr/cidr_unspecified.go
+++ b/pkg/cidr/cidr_unspecified.go
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+//go:build !linux
+
+package cidr
+
+// Dummy values on non-linux platform
+const (
+	FAMILY_V4 = iota
+	FAMILY_V6
+)
--- a/pkg/cidr/cidr.go
+++ b/pkg/cidr/cidr.go
@@ -7,8 +7,6 @@ import (
 	"bytes"
 	"fmt"
 	"net"
-
-	"github.com/vishvananda/netlink"
 )
 
 // NewCIDR returns a new CIDR using a net.IPNet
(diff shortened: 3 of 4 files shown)


.github/workflows/lint-build-commits.yaml
```diff
@@ -100,7 +100,7 @@ jobs:
 
       # Runs only if code under test/ is changed.
       - name: Check if ginkgo test suite build works for every commit
-        if: steps.test-tree.outputs.src == 'true'
+#        if: steps.test-tree.outputs.src == 'true'
```
joestringer: Need to drop this before merging.

joestringer (approved):
Changes LGTM. I wonder if we also need something like this since we clearly have test code depending on `pkg/`:

```
diff --git a/.github/workflows/lint-build-commits.yaml b/.github/workflows/lint-build-commits.yaml
index 32d9fe21254d..c27c724f1ed3 100644
--- a/.github/workflows/lint-build-commits.yaml
+++ b/.github/workflows/lint-build-commits.yaml
@@ -96,6 +96,7 @@ jobs:
         with:
           filters: |
             src:
+              - 'pkg/**'
               - 'test/**'
 
       # Runs only if code under test/ is changed.
```

pkg/cidr/cidr_unspecified.go
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+//go:build !linux
+
+package cidr
+
+// Dummy values on non-linux platform
+const (
+	FAMILY_V4 = 0
```
joestringer:
```suggestion
	FAMILY_V4 = iota
```

thorn3r (approved):
thanks for the quick fix!

aditighag:
Ginkgo GHA ran successfully - https://github.com/cilium/cilium/actions/runs/5945229937/job/16123944232?pr=27638. 
Dropping the commit.

aditighag:
> Changes LGTM. I wonder if we also need something like this since we clearly have test code depending on `pkg/`:
> 
> ```
> diff --git a/.github/workflows/lint-build-commits.yaml b/.github/workflows/lint-build-commits.yaml
> index 32d9fe21254d..c27c724f1ed3 100644
> --- a/.github/workflows/lint-build-commits.yaml
> +++ b/.github/workflows/lint-build-commits.yaml
> @@ -96,6 +96,7 @@ jobs:
>          with:
>            filters: |
>              src:
> +              - 'pkg/**'
>                - 'test/**'
>  
>        # Runs only if code under test/ is changed.
> ```

Yes, I was going to suggest running the GHA unconditionally, but your suggestion is better. I can push the commit to this PR.

joestringer (approved):

borkmann (commented):
Awesome, thanks for the fix!

PR #27639, opened 2023-08-23
bgpv1: BGP speaker OnFibEvent

ldelossa:
This commit adds a `OnFibEvent` function to types.ServerParameters.

This function is called when the instantiated BGP speaker experiences a
fib event.

When a FIB even occurs BGP-CP will be triggered for reconciliation.

This will be useful in future commits when we also learn routes from the
upstream peer.

Signed-off-by: Louis DeLosSantos <louis.delos@isovalent.com>

```release-note
bgpv1: add controller reconciliation on FIB events
```

### DIFF
--- a/pkg/bgpv1/manager/reconcile.go
+++ b/pkg/bgpv1/manager/reconcile.go
@@ -9,6 +9,7 @@ import (
 	"net/netip"
 
 	"github.com/cilium/cilium/pkg/bgpv1/agent"
+	"github.com/cilium/cilium/pkg/bgpv1/agent/signaler"
 	"github.com/cilium/cilium/pkg/bgpv1/types"
 	"github.com/cilium/cilium/pkg/hive/cell"
 	"github.com/cilium/cilium/pkg/k8s"
@@ -49,6 +50,11 @@ var ConfigReconcilers = cell.ProvidePrivate(
 	NewLBServiceReconciler,
 )
 
+type PreflightReconcilerParams struct {
+	cell.In
+	BGPCPSignaler *signaler.BGPCPSignaler
+}
+
 type PreflightReconcilerOut struct {
 	cell.Out
 
@@ -66,11 +72,15 @@ type PreflightReconcilerOut struct {
 //
 // permanent configurations for BgpServers (ones that cannot be changed after creation)
 // are router ID and local listening port.
-type PreflightReconciler struct{}
+type PreflightReconciler struct {
+	signaler *signaler.BGPCPSignaler
+}
 
-func NewPreflightReconciler() PreflightReconcilerOut {
+func NewPreflightReconciler(params PreflightReconcilerParams) PreflightReconcilerOut {
 	return PreflightReconcilerOut{
-		Reconciler: &PreflightReconciler{},
+		Reconciler: &PreflightReconciler{
+			signaler: params.BGPCPSignaler,
+		},
 	}
 }
 
@@ -150,6 +160,11 @@ func (r *PreflightReconciler) Reconcile(ctx context.Context, p ReconcileParams) 
 			RouteSelectionOptions: &types.RouteSelectionOptions{
 				AdvertiseInactiveRoutes: true,
 			},
+		},
+		OnFIBEvent: func() {
+			if r.signaler != nil {
+				r.signaler.Event(struct{}{})
+			}
 		},
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/bgpv1/manager/reconcile.go
```diff
@@ -151,6 +161,11 @@ func (r *PreflightReconciler) Reconcile(ctx context.Context, p ReconcileParams)
 				AdvertiseInactiveRoutes: true,
 			},
 		},
+		OnFIBEvent: func() {
+			if r.signaler != nil {
```
YutaroHayakawa, line 165: Maybe `r.signaler == nil` is impossible? We set `r.signaler` on initialization.

pkg/bgpv1/types/bgp.go
```diff
@@ -69,7 +69,8 @@ type GetBGPResponse struct {
 
 // ServerParameters contains information for underlying bgp implementation layer to initializing BGP process.
 type ServerParameters struct {
-	Global BGPGlobal
+	Global     BGPGlobal
+	OnFIBEvent func()
```
YutaroHayakawa, line 73: Why do we need this?

YutaroHayakawa (changes requested):

pkg/bgpv1/manager/reconcile.go
ldelossa: This is nice for unit testing, as the unit tests won't require a Signaler to be constructed. Thoughts? I'm okay to change this if you feel strongly about it.

pkg/bgpv1/manager/reconcile.go
```diff
@@ -151,6 +161,11 @@ func (r *PreflightReconciler) Reconcile(ctx context.Context, p ReconcileParams)
...
+				r.signaler.Event(struct{}{})
```
rastislavs, line 166: I wonder if we shouldn't have some option for this instead of signaling on each event unconditionally? Too many neighbors and/or a flapping neighbor may trigger too many unnecessary resyncs.

Alternatively (just brain-storming in my head), maybe it would make sense to pass the actual `Path` that triggered the event and define a `ConfigReconciler` API method that would be called on FIB event on all reconcilers - and they could decide whether the event is worth triggering a reconciliation?

ldelossa: The latter approach sounds the same as just triggering reconciliation at the high level, and each reconciler determining that they need to do no work. 

I think we need to design all reconcilers to do minimal work to understand if they can just exit, which i think is the case today. 

The former question... its possible we could add a method to a Reconciler like:
'NeedsReconcilationOnFIBEvent bool', Allowing the BGP-CP to recognize if it should reconcile on FIB events at all.

We would then need a way to ask the BGP-CP, are we running with any ConfigReconcilers which need reconciliation on fib events. hmm. Something like this?

rastislavs: > We would then need a way to ask the BGP-CP, are we running with any ConfigReconcilers which need reconciliation on fib events. hmm. Something like this?

yeah, something like that could work as well?

My original idea was something like:

```go
type ConfigReconciler interface {
	Priority() int
	Reconcile(ctx context.Context, params ReconcileParams) error
	FIBEvent(p *types.Path)
}
```

Then I guess most reconcilers would just return in `FIBEvent` and some could do something like:

```go
func (r *AwesomeReconciler) FIBEvent(p *types.Path) {
  if p.NLRI.AFI() == x &&  p.NLRI.SAAFI() == y {
     signal_to_trigger_reconciliation()
  }
}
```

That way we could trigger reconciliation of all reconcilers only upon relevant FIB events..

pkg/bgpv1/types/bgp.go
harsimran-pabla, line 73: nit: FIB is slightly misleading term. Route learned from BGP is not really gone into FIB ( linux routing table ). I'd suggest `OnRouteEvent`.

harsimran-pabla, line 73: Do we want to add `OnNeighborEvent`callback as well. This would help in writing metrics for neighbor specific changes.

pkg/bgpv1/manager/reconcile.go
YutaroHayakawa: How about injecting a dummy signaler on testing then?

PR #27643, opened 2023-08-23
statedb: Rename statedb2 to statedb

joamaki:
Now that old pkg/statedb has been removed, rename
pkg/statedb2 to pkg/statedb.


bimmlerd (approved):

joestringer:
Will this cause backporting issues for statedb features in v1.14?

christarazi (approved):

joamaki:
> Will this cause backporting issues for statedb features in v1.14?

Would only cause it for L2 announcements, so that should be manageable.

joestringer (approved):
Looks like tophat got pulled in because of pkg/statedb2. I didn't review in detail but it looks like a straightforward rename. I assume others have looked more closely at the details.

PR #27645, opened 2023-08-23
Set kvstoremesh image when pushing the development helm chart

giorio94:
Configure the kvstoremesh repository and tag to match the one of the other cilium components. Additionally, explicitly disable the digest, which is otherwise configured in stable branches.

Outcome with this change:

```
$ helm show values -n kube-system oci://quay.io/cilium-charts-dev/cilium \
      --version 1.15.0-dev-dev.725-pr-giorio94-main-gha-push-chart-kvstoremesh-3a6c2670f8 | \
      grep 'kvstoremesh:' -A12 -m 1

kvstoremesh:
  # -- Enable KVStoreMesh. KVStoreMesh caches the information retrieved
  # from the remote clusters in the local etcd instance.
  enabled: false
  # -- KVStoreMesh image.
  image:
    override: ~
    repository: "quay.io/cilium/kvstoremesh-ci"
    tag: "3a6c2670f89bbc8b9de6ebf69cae1aab4f9f2718"
    # kvstoremesh-digest
    digest: ""
    useDigest: false
    pullPolicy: "Always"

```

### DIFF
--- a/.github/workflows/push-chart-ci.yaml
+++ b/.github/workflows/push-chart-ci.yaml
@@ -120,7 +120,11 @@ jobs:
             "hubble.relay.image.repository": "quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/hubble-relay-ci",
             "hubble.relay.image.tag": "${{ steps.get-ref.outputs.sha }}",
             "clustermesh.apiserver.image.repository": "quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/clustermesh-apiserver-ci",
-            "clustermesh.apiserver.image.tag": "${{ steps.get-ref.outputs.sha }}"
+            "clustermesh.apiserver.image.tag": "${{ steps.get-ref.outputs.sha }}",
+            "clustermesh.apiserver.kvstoremesh.image.repository": "quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/kvstoremesh-ci",
+            "clustermesh.apiserver.kvstoremesh.image.tag": "${{ steps.get-ref.outputs.sha }}",
+            "clustermesh.apiserver.kvstoremesh.image.digest": "",
+            "clustermesh.apiserver.kvstoremesh.image.useDigest": false
           }
         registry: quay.io
         registry_namespace: ${{ env.QUAY_CHARTS_ORGANIZATION_DEV }}


nbusseneau (approved):

giorio94:
Last push was to fix a missing open quote in the useDigest parameter.

giorio94:
I did one additional test after the fix, and the chart was rendered and installed correctly. I'm marking this PR ready to merge given that the review is in, and it only touches a GHA workflow.

.github/workflows/push-chart-ci.yaml
```diff
@@ -120,7 +120,11 @@ jobs:
             "hubble.relay.image.repository": "quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/hubble-relay-ci",
             "hubble.relay.image.tag": "${{ steps.get-ref.outputs.sha }}",
             "clustermesh.apiserver.image.repository": "quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/clustermesh-apiserver-ci",
-            "clustermesh.apiserver.image.tag": "${{ steps.get-ref.outputs.sha }}"
+            "clustermesh.apiserver.image.tag": "${{ steps.get-ref.outputs.sha }}",
+            "clustermesh.apiserver.kvstoremesh.image.repository": "quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/kvstoremesh-ci",
+            "clustermesh.apiserver.kvstoremesh.image.tag": "${{ steps.get-ref.outputs.sha }}",
+            "clustermesh.apiserver.kvstoremesh.image.digest": "",
+            "clustermesh.apiserver.kvstoremesh.image.useDigest": false
```
joestringer, line 127: Why do these changes need to be put here, but the other images don't specify digest/useDigest?

joestringer: > Additionally, explicitly disable the digest, which is otherwise configured in stable branches.

It's configured that way so that if you pull the branch and install from Helm on that branch, it'll install the latest version from that branch. If we want to change this behaviour, we should change it globally, not for one specific image.

joestringer (changes requested):
Let's resolve the discussion above first. Whether or not it makes sense to override the digest pieces, it should be consistent for all images, not different for the kvstoremesh.

.github/workflows/push-chart-ci.yaml
giorio94: Yeah, the other images are modified [here](https://github.com/cilium/cilium/pull/27646) (it's a different PR because it needs to be backported to different branches). 

The issue is that this workflow overrides the tag to build a chart using the images from the current commit, but given that it does not remove the digest it will not work on stable branches (as the digest would be that of the latest release for that branch).

giorio94: With the ocurrent action, I had to manually disable the digest when using the development chart from a PR targeting v1.14 in https://github.com/cilium/cilium/pull/27232/commits/978071bdd353aa6add05c3c6318eca1b00e9bc03

joestringer (approved):

joestringer:
Filed https://github.com/cilium/cilium/issues/27675 for the test failure, it's clearly unrelated and seems like a bug in the metrics CI test.

PR #27656, opened 2023-08-23
Fix computation of IPsec max. sequence number

pchaigno:
See commits for details. First commit refactors to allow for subsequent unit test; second fixes the bug. third adds a unit test.

```release-note
Fix a bug that could cause an incorrect max. sequence number to be reported by `cilium encrypt status` when IPsec is enabled.
```

### DIFF
--- a/cilium/cmd/encrypt_status.go
+++ b/cilium/cmd/encrypt_status.go
@@ -89,14 +89,9 @@ func countUniqueIPsecKeys() int {
 	return len(keys)
 }
 
-func maxSequenceNumber() string {
+func extractMaxSequenceNumber(ipOutput string) string {
 	maxSeqNum := "0"
-	out, err := exec.Command("ip", "xfrm", "state", "list", "reqid", ciliumReqId).Output()
-	if err != nil {
-		Fatalf("Cannot get xfrm states: %s", err)
-	}
-	commandOutput := string(out)
-	lines := strings.Split(commandOutput, "\n")
+	lines := strings.Split(ipOutput, "\n")
 	for _, line := range lines {
 		matched := regex.FindStringSubmatchIndex(line)
 		if matched != nil {
@@ -106,6 +101,16 @@ func maxSequenceNumber() string {
 			}
 		}
 	}
+	return maxSeqNum
+}
+
+func maxSequenceNumber() string {
+	out, err := exec.Command("ip", "xfrm", "state", "list", "reqid", ciliumReqId).Output()
+	if err != nil {
+		Fatalf("Cannot get xfrm states: %s", err)
+	}
+	commandOutput := string(out)
+	maxSeqNum := extractMaxSequenceNumber(commandOutput)
 	if maxSeqNum == "0" {
 		return "N/A"
 	}
--- a/cilium/cmd/encrypt_status.go
+++ b/cilium/cmd/encrypt_status.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/prometheus/procfs"
@@ -30,7 +31,7 @@ type void struct{}
 var (
 	voidType    void
 	countErrors int
-	regex       = regexp.MustCompile("oseq[[:blank:]](0[xX][[:xdigit:]]+)?")
+	regex       = regexp.MustCompile("oseq[[:blank:]]0[xX]([[:xdigit:]]+)")
 )
 
 var encryptStatusCmd = &cobra.Command{
@@ -89,13 +90,17 @@ func countUniqueIPsecKeys() int {
 	return len(keys)
 }
 
-func extractMaxSequenceNumber(ipOutput string) string {
-	maxSeqNum := "0"
+func extractMaxSequenceNumber(ipOutput string) int64 {
+	maxSeqNum := int64(0)
 	lines := strings.Split(ipOutput, "\n")
 	for _, line := range lines {
 		matched := regex.FindStringSubmatchIndex(line)
 		if matched != nil {
-			oseq := line[matched[2]:matched[3]]
+			oseq, err := strconv.ParseInt(line[matched[2]:matched[3]], 16, 64)
+			if err != nil {
+				Fatalf("Failed to parse sequence number '%s': %s",
+					line[matched[2]:matched[3]], err)
+			}
 			if oseq > maxSeqNum {
 				maxSeqNum = oseq
 			}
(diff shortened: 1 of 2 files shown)


cilium/cmd/encrypt_status.go
```diff
@@ -89,13 +90,17 @@ func countUniqueIPsecKeys() int {
...
 	for _, line := range lines {
 		matched := regex.FindStringSubmatchIndex(line)
 		if matched != nil {
-			oseq := line[matched[2]:matched[3]]
+			oseq, err := strconv.ParseInt(line[matched[2]:matched[3]], 16, 64)
+			if err != nil {
+				Fatalf("Failed to parse sequence number '%s': %s",
+					line[matched[2]:matched[3]], err)
+			}
 			if oseq > maxSeqNum {
```
gentoo-root, line 104: When I looked at the first commit, I was about to comment this is buggy, until I realized that your second commit fixes exactly this :D

gentoo-root (approved):

cilium/cmd/encrypt_status.go
pchaigno: Then you've got better eyes than me because I think I was the original reviewer :grimacing:

cilium/cmd/encrypt_status.go
```diff
@@ -89,13 +90,17 @@ func countUniqueIPsecKeys() int {
 	return len(keys)
 }
 
-func extractMaxSequenceNumber(ipOutput string) string {
-	maxSeqNum := "0"
+func extractMaxSequenceNumber(ipOutput string) int64 {
+	maxSeqNum := int64(0)
```
derailed, line 94: nit: use the zero value i.e var maxSeqNum int64

cilium/cmd/encrypt_status.go
```diff
@@ -89,13 +90,17 @@ func countUniqueIPsecKeys() int {
...
 	lines := strings.Split(ipOutput, "\n")
 	for _, line := range lines {
 		matched := regex.FindStringSubmatchIndex(line)
 		if matched != nil {
```
derailed, line 98: nit: perhaps better to reverse the logic here for line of sight ie if no match continue.

Also not keen on bombing on a utility func. Might be better to return the error and let the call site decide on how to proceed from there.

cilium/cmd/encrypt_status_test.go
```diff
@@ -119,3 +119,30 @@ func (s *EncryptStatusSuite) TestGetXfrmStats(c *C) {
...
+	sel src 0.0.0.0/0 dst 0.0.0.0/0
+src 10.84.1.32 dst 10.84.2.145
+	proto esp spi 0x00000003 reqid 1 mode tunnel
+	replay-window 0
+	mark 0x7e63e00/0xffffff00 output-mark 0xe00/0xf00
+	aead rfc4106(gcm(aes)) 0x64ad37a9d8a8f20fb2e74ef6000f9d580898719f 128
+	anti-replay context: seq 0x0, oseq 0x13e0, bitmap 0x00000000
+	sel src 0.0.0.0/0 dst 0.0.0.0/0`
+
+	maxSeqNumber := extractMaxSequenceNumber(ipOutput)
```
derailed, line 146: nit: Per the above we could test call boundaries without bombing vs just happy path

derailed (changes requested):
@pchaigno Great catch! Nice work. Just a few small picks...

cilium/cmd/encrypt_status.go
pchaigno: Why?

cilium/cmd/encrypt_status.go
pchaigno: Do you mean the fatal by "bombing"? This isn't really new. All our commands currently behave this way.

cilium/cmd/encrypt_status_test.go
pchaigno: Yep, that's a good point. I'm however unsure of what a valid input that would cause a fatal would be. Any idea?

cilium/cmd/encrypt_status.go
derailed: imho easier to read and leverages the language zero value mechanic. Since we need to return int64 might as well also declare the variable as so.

cilium/cmd/encrypt_status.go
derailed: Right! yes I see the pattern and I am personally not fond of it ;( 
I get that this is a limited use case but since we are testing the call might as well test the boundaries without potentially causing the test to bomb too. To boot we could leverage the return error to also tell the caller no max seq number can be found vs making up a value aka "n/a"??

cilium/cmd/encrypt_status.go
gentoo-root: For me (as a non-native Go speaker :D) `maxSeqNum := int64(0)` looks more straightforward and explicit than `var maxSeqNum int64`. I.e. with the former someone who reads the code doesn't need to know Go (and its zero value mechanic) to understand that the variable is zero-initialized. Is it some sort of a well-known guideline?

cilium/cmd/encrypt_status_test.go
derailed: Tx Paul! I am just neat picking here. So please feel free to drop if too much noise...
Me thinks "anti-replay context: seq 0x0, oseq 0x, bitmap 0x00000000" will cause us to bomb ;(

cilium/cmd/encrypt_status.go
derailed: Thanks Max! I totally get this, but when in Rome... Go proverbs: "Make the zero value useful"

pchaigno: This should be covered by a linter if we want it consistent one way or the other. In the absence of such a linter, I don't think we should be rejecting PRs based on that.

cilium/cmd/encrypt_status_test.go
pchaigno: > Tx Paul! I am just neat picking here. So please feel free to drop if too much noise...

Well... you did reject the PR based on those three nitpicks ;)

> Me thinks "anti-replay context: seq 0x0, oseq 0x, bitmap 0x00000000" will cause us to bomb ;(

But that's not something ip x states can return.

derailed: Well you did ask for a scenario that would cause this call to bomb.
At the very least, we should check the returned slice bounds and/or fix the regex since it allowed it in the first place?

pchaigno: But what you call bomb is erroring out from the CLI with a proper error message. Or am I missing something?

derailed: You are right! My feel on these kind of calls is we should just return an error and let the call site decide. Best if fatal only occur at the top level. Also easier to test boundaries.
But that's only my opinion... You can mark it as resolved.

derailed (approved):
@pchaigno Thanks Paul for the comments. LGTM

cilium/cmd/encrypt_status_test.go
pchaigno: I found one (I *think* realistic) case that I thought could cause a fatal (interrupted output) so I added it. Turns out it's correctly handled, but probably worth having just in case.

PR #27663, opened 2023-08-23
Delete auth map entries of deleted SVID

meyskens:
This change adds a watcher on the SPIRE updates for deleted SVIDs. If one is deleted it will be signaled over the rotation channel. Should the ID be present in the auth map it will be immediately removed.

```release-note
Delete auth map entries for removed Security IDs in SPIRE
```


youngnick (approved):

joestringer:
Is this really `release-note/bug`?

PR #27665, opened 2023-08-23
config: Update TCP conntrack entry timeout

aditighag:
Update the TCP conntrack entry timeout previously set to a large value (6 hours).

In some cases, a large timeout leads to "sticky" CT entries, and this can lead to skewed backend selection. Examples of such cases: ephemeral ports exhaustion, or when backend pods are scaled up. At the same time, we don't want to prematurely garbage collect a CT entry, as that may reset long-lived TCP sessions. Update the timeout for TCP, and use the keep alive timeouts used by the Linux kernel as a reference (suggested by Daniel Borkmann).

Signed-off-by: Aditi Ghag <aditi@cilium.io>

```release-note
Update the TCP conntrack entry timeouts to a lower value, so that closed entries are garbage collected earlier, thus freeing up the conntrack map.  
```


daemon/cmd/daemon_main.go
```diff
@@ -802,13 +802,13 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	flags.Int(option.CTMapEntriesGlobalAnyName, option.CTMapEntriesGlobalAnyDefault, "Maximum number of entries in non-TCP CT table")
 	option.BindEnvWithLegacyEnvFallback(vp, option.CTMapEntriesGlobalAnyName, "CILIUM_GLOBAL_CT_MAX_ANY")
 
-	flags.Duration(option.CTMapEntriesTimeoutTCPName, 21600*time.Second, "Timeout for established entries in TCP CT table")
+	flags.Duration(option.CTMapEntriesTimeoutTCPName, 7200*time.Second, "Timeout for established entries in TCP CT table")
```
julianwiedmann: Agree this change also makes sense for non-Service connections. 2h is still more than plenty.

giorio94 (approved):
One question out of curiosity, the commit message mentions to align this timeout to the default value of `tcp_keepalive_time`. Would that mean that even if the CT entry gets GCed by Cilium, then it would be restored automatically due to the keepalive, or that would instead cause the failure of the connection? In the latter case, I wonder if it could make sense to slightly increase this timeout so that it also includes the time required for the keepalive (kernel docs mention ~11min at most).

Approving to cover the code-owners, but I'd defer the final approval to @cilium/sig-datapath (I see Julian is already marked as a reviewer), who knows better about the specificity of connection tracking and the interactions with this timeout.

borkmann:
> One question out of curiosity, the commit message mentions to align this timeout to the default value of `tcp_keepalive_time`. Would that mean that even if the CT entry gets GCed by Cilium, then it would be restored automatically due to the keepalive, or that would instead cause the failure of the connection? In the latter case, I wonder if it could make sense to slightly increase this timeout so that it also includes the time required for the keepalive (kernel docs mention ~11min at most).

Was thinking the same, we should slightly make this larger e.g. 7800 secs.

aditighag:
>I wonder if it could make sense to slightly increase this timeout so that it also includes the time required for the keepalive (kernel docs mention ~11min at most).

I was wondering the same, as the kernel documentation is a bit confusing. 

>tcp_keepalive_time - INTEGER
	How often TCP sends out keepalive messages when keepalive is enabled.
	Default: 2hours.

>tcp_keepalive_intvl - INTEGER
	How frequently the probes are send out. Multiplied by
	tcp_keepalive_probes it is time to kill not responding connection,
	after probes started. Default value: 75sec i.e. connection
	will be aborted after ~11 minutes of retries.

I interpreted this as 2 hours is the total time the probes are sent for, and that it subsumes the 11 minutes. Julian also mentioned that 2h is still plenty. 

Anyway it's easy enough to bump up the time, but is it really required?

julianwiedmann (approved):

borkmann:
> I interpreted this as 2 hours is the total time the probes are sent for, and that it subsumes the 11 minutes. Julian also mentioned that 2h is still plenty.
> 
> Anyway it's easy enough to bump up the time, but is it really required?
> 
> Edit: I read the kernel code a bit, and based on my understanding `tcp_keepalive_time` also accounts for TCP retransmission timeouts, as expected. This documentation clarifies this further - https://tldp.org/HOWTO/TCP-Keepalive-HOWTO/usingkeepalive.html. `This means that the keepalive routines wait for two hours (7200 secs) before sending the first keepalive probe, and then resend it every 75 seconds. If no ACK response is received for nine consecutive times, the connection is marked as broken.`

Where do you see it in the kernel code? I'm reading that the keepalive_time_when()  aka ipv4.sysctl_tcp_keepalive_time is the timeout we charge the TCP timer to come alive. See tcp_keepalive_timer() and the test for `elapsed >= keepalive_time_when(tp)`. So only after 2h it kicks in, therefore I still think we need to slightly bump it. Let me know if you read this differently.

borkmann (approved):

aditighag:
> > I interpreted this as 2 hours is the total time the probes are sent for, and that it subsumes the 11 minutes. Julian also mentioned that 2h is still plenty.
> > Anyway it's easy enough to bump up the time, but is it really required?
> > Edit: I read the kernel code a bit, and based on my understanding `tcp_keepalive_time` also accounts for TCP retransmission timeouts, as expected. This documentation clarifies this further - https://tldp.org/HOWTO/TCP-Keepalive-HOWTO/usingkeepalive.html. `This means that the keepalive routines wait for two hours (7200 secs) before sending the first keepalive probe, and then resend it every 75 seconds. If no ACK response is received for nine consecutive times, the connection is marked as broken.`
> 
> Where do you see it in the kernel code? I'm reading that the keepalive_time_when() aka ipv4.sysctl_tcp_keepalive_time is the timeout we charge the TCP timer to come alive. See tcp_keepalive_timer() and the test for `elapsed >= keepalive_time_when(tp)`. So only after 2h it kicks in, therefore I still think we need to slightly bump it. Let me know if you read this differently.

Yup, I saw the same code, and that's why I already bumped up the timeout.

PR #27669, opened 2023-08-23

jrfastab:
took some longer time. For example I ran `ip x s` in a pod about
received are dropped.

traffic with the new IPs.

XfrmIn*Policy* errors I've seen.

### DIFF
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -941,6 +941,16 @@ func NewDaemon(ctx context.Context, cancel context.CancelFunc, epMgr *endpointma
 		return nil, nil, err
 	}
 
+	// allocateIPs got us the routerIP so now we can create ipsec endpoint
+	// we must do this before publishing the router IP otherwise remote
+	// nodes could pick up the IP and send us outer headers we do not yet
+	// have xfrm rules for.
+	if option.Config.EnableIPSec {
+		if err := ipsec.Init(); err != nil {
+			log.WithError(err).Error("IPSec init failed")
+		}
+	}
+
 	// Must occur after d.allocateIPs(), see GH-14245 and its fix.
 	d.nodeDiscovery.StartDiscovery()
 
--- a/pkg/datapath/linux/ipsec/ipsec_linux.go
+++ b/pkg/datapath/linux/ipsec/ipsec_linux.go
@@ -1081,3 +1081,32 @@ func StartStaleKeysReclaimer(ctx context.Context) {
 		}
 	}()
 }
+
+// We need to install xfrm state for the local router (cilium_host) early
+// in daemon init path. This is to ensure that we have the xfrm state in
+// place before we advertise the routerIP where other nodes may potentially
+// pick it up and start sending traffic to us. This was previously racing
+// and creating XfrmInNoState errors because other nodes picked up node
+// update before Xfrm config logic was in place. So special case init the
+// rule we need early in init flow.
+func Init() error {
+	outerLocalIP := node.GetInternalIPv4Router()
+	wildcardIP := net.ParseIP("0.0.0.0")
+	localCIDR := node.GetIPv4AllocRange().IPNet
+	localWildcardIP := &net.IPNet{IP: wildcardIP, Mask: net.IPv4Mask(0, 0, 0, 0)}
+
+	if _, err := ipSecReplaceStateIn(outerLocalIP, wildcardIP, false); err != nil {
+		return fmt.Errorf("unable to replace local state: %s", err)
+	}
+	if err := ipSecReplacePolicyIn(localWildcardIP, localCIDR, wildcardIP, outerLocalIP); err != nil {
+		if !os.IsExist(err) {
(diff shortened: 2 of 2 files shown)


pkg/nodediscovery/nodediscovery.go
```diff
@@ -174,7 +174,8 @@ func (n *NodeDiscovery) StartDiscovery() {
 	defer n.localNodeLock.Unlock()
 
 	n.fillLocalNode()
-
+	n.Manager.NodeUpdated(n.localNode)
+	// Ensure local nodeUpdates are completed before we add local node to cluster.
```
zacharysarah: Optional docs improvement
```suggestion
	// Ensure local nodeUpdates are completed before adding the local node to the cluster.
```

zacharysarah (dismissed):
Optional fix, otherwise LGTM for docs

joestringer:
@zacharysarah heads up if something is in draft, we typically wouldn't bother reviewing it. I set this to draft because initially it pulled in half the Cilium organization unintentionally because of the wrong base branch. cc also @jrfastab why to start with draft PRs then only mark "ready for review" when it's ready. We should probably drop all of the random reviewers from this PR too.

zacharysarah:
@joestringer Ahhh sorry, I was on review autopilot and didn't notice the draft status. 😳

jrfastab:
oh yeah missed setting draft and pushed to wrong branch all sorts of typos. Sorry for the noise.

jrfastab:
this doesn't work fwiw it calling d.init() here appears to lock up agent startup.

daemon/cmd/daemon.go
```diff
@@ -941,6 +941,14 @@ func NewDaemon(ctx context.Context, cancel context.CancelFunc, epMgr *endpointma
 		return nil, nil, err
 	}
 
+	// allocateIPs got us the routerIP so now we can create ipsec endpoint
+	// we must do this before publishing the router IP otherwise remote
+	// nodes could pick up the IP and send us outer headers we do not yet
+	// have xfrm rules for.
```
joestringer, line 947: nit: I'd probably add a comment to the effect of this to ensure later refactors can take into account the dependencies for this new logic:
```suggestion
	// have xfrm rules for.
	//
	// This must occur strictly after `setupIPSec()` and `d.allocateIPs()`.
```

pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -1081,3 +1081,16 @@ func StartStaleKeysReclaimer(ctx context.Context) {
...
+// in daemon init path. This is to ensure that we have the xfrm state in
+// place before we advertise the routerIP where other nodes may potentially
+// pick it up and start sending traffic to us. This was previously racing
+// and creating XfrmInNoState errors because other nodes picked up node
+// update before Xfrm config logic was in place. So special case init the
+// rule we need early in init flow.
+func Init() {
+	outerLocalIP := node.GetInternalIPv4Router()
+	wildcardIP := net.ParseIP("0.0.0.0")
+	ipSecReplaceStateIn(outerLocalIP, wildcardIP, false)
```
joestringer: This seems fine iff this will only _add_ state for the target IP and not _replace_, which I guess is implied by the target use case here where the target IP will be different? If for some reason the keys on the filesystem change while Cilium is down and this IP is *not* new, then I suspect this may replace existing state and possibly break existing connections...?

jrfastab: It can replace but only if the tuple matches, but tuple in xfrm context also includes SPI so if the cilium_host IP didn't change and keys rolled while we are off restarting we still wont replace because a key roll increments SPI.

daemon/cmd/daemon.go
jrfastab: will do.

jrfastab:
I guess I should actually keep this marked draft. I know folks prefer fixes to go through main first.

jrfastab:
Reviewing code I believe having two patches is redundant and we can have just 2nd patch. Testing now.

daemon/cmd/daemon.go
```diff
@@ -994,15 +1003,6 @@ func NewDaemon(ctx context.Context, cancel context.CancelFunc, epMgr *endpointma
 		d.bootstrapClusterMesh(nodeMngr)
 	}
 
-	// Must be done at least after initializing BPF LB-related maps
-	// (lbmap.Init()).
```
ti-mo: Shouldn't this comment be preserved in addition to the new one you added? This context is useful to have if it's still relevant.

jrfastab:
@pchaigno @joestringer question, why is patch 3 required to fix the errors? I thought patch 1 should resolve the problem.

jrfastab:
The reason patch 3 was still needed is because I only setup the States in the ipsec.Init() but we really need to configure the state and policy otherwise we solve the noState errors and instead get Policy lookup errors.

With current patch I've been doing upgrades without drops consistently on vxlan mode (where I originally was able to reproduce the error) I'll test ENI modes now as well.

PR #27673, opened 2023-08-23
Resiliency: Add retry logic to attempt to clear out stale hostip

derailed:

      GitHub issue.

Resiliency: Add retry logic to attempt to clear out stale hostip

NOTE: Depends on PR: https://github.com/cilium/cilium/pull/27614

When the daemon initializes it attempts to clean out stale host ips. 
This process only occurs one time irregardless of errors occurring during that reconciliation.

- Add retry logic on the daemon to attempt to clear out stale host ips should errors occurred during that process.
    
> NOTE: Not super sure on the best way to handle this use case??
This will potentially put the brakes on while constructing the daemon.
This ctor code is currently very procedural (not to mention a mile long!) 
as each step depends on the resolution of the previous one?
    
Fixes: #issue-number

### DIFF
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -87,6 +87,7 @@ import (
 	"github.com/cilium/cilium/pkg/rate"
 	"github.com/cilium/cilium/pkg/recorder"
 	"github.com/cilium/cilium/pkg/redirectpolicy"
+	"github.com/cilium/cilium/pkg/resiliency"
 	"github.com/cilium/cilium/pkg/service"
 	serviceStore "github.com/cilium/cilium/pkg/service/store"
 	"github.com/cilium/cilium/pkg/source"
@@ -312,7 +313,7 @@ func createPrefixLengthCounter() *counter.PrefixLengthCounter {
 // that the most up-to-date information has been retrieved. At this point, the
 // daemon is aware of all the necessary information to restore the appropriate
 // IP.
-func (d *Daemon) restoreCiliumHostIPs(ipv6 bool, fromK8s net.IP) {
+func (d *Daemon) restoreCiliumHostIPs(ipv6 bool, fromK8s net.IP) error {
 	var (
 		cidrs  []*cidr.CIDR
 		fromFS net.IP
@@ -343,12 +344,7 @@ func (d *Daemon) restoreCiliumHostIPs(ipv6 bool, fromK8s net.IP) {
 	}
 
 	restoredIP := node.RestoreHostIPs(ipv6, fromK8s, fromFS, cidrs)
-	if err := removeOldRouterState(ipv6, restoredIP); err != nil {
-		log.WithError(err).Warnf(
-			"Failed to remove old router IPs (restored IP: %s) from cilium_host. Manual intervention is required to remove all other old IPs.",
-			restoredIP,
-		)
-	}
+	return removeOldRouterState(ipv6, restoredIP)
 }
 
 // removeOldRouterState will try to ensure that the only IP assigned to the
@@ -360,8 +356,9 @@ func removeOldRouterState(ipv6 bool, restoredIP net.IP) error {
 		// There's no old state remove as the host device doesn't exist and
 		// there's no restored IP anyway.
 		return nil
-	} else if err != nil {
-		return err
+	}
+	if err != nil {
+		return resiliency.Retryable(err)
 	}
 
 	family := netlink.FAMILY_V4
@@ -370,7 +367,7 @@ func removeOldRouterState(ipv6 bool, restoredIP net.IP) error {
 	}
 	addrs, err := netlink.AddrList(l, family)
 	if err != nil {
-		return err
+		return resiliency.Retryable(err)
 	}
 
 	isRestoredIP := func(a netlink.Addr) bool {
@@ -382,20 +379,16 @@ func removeOldRouterState(ipv6 bool, restoredIP net.IP) error {
 
 	log.Info("More than one stale router IP was found on the cilium_host device after restoration, cleaning up old router IPs.")
 
-	var errs []error
 	for _, a := range addrs {
 		if isRestoredIP(a) {
 			continue
 		}
-		if err := netlink.AddrDel(l, &a); err != nil {
-			errs = append(errs, fmt.Errorf("failed to remove IP %s: %w", a.IP, err))
+		if e := netlink.AddrDel(l, &a); e != nil {
+			err = errors.Join(err, resiliency.Retryable(fmt.Errorf("failed to remove IP %s: %w", a.IP, e)))
 		}
 	}
-	if len(errs) > 0 {
-		return fmt.Errorf("failed to remove all old router IPs: %v", errs)
-	}
 
-	return nil
+	return err
 }
 
 // newDaemon creates and returns a new Daemon with the parameters set in c.
@@ -1068,11 +1061,22 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 	// router IPs from the K8s resources if we weren't able to restore them
 	// from the fs. We must do this after IPAM because we must wait until the
 	// K8s resources have been synced. Part 2/2 of restoration.
-	if option.Config.EnableIPv4 {
-		d.restoreCiliumHostIPs(false, router4FromK8s)
+	gcHostIPsFn := func(ctx context.Context, retries int) (done bool, err error) {
+		var errs error
+		if option.Config.EnableIPv4 {
+			errs = errors.Join(errs, d.restoreCiliumHostIPs(false, router4FromK8s))
+		}
+		if option.Config.EnableIPv6 {
+			errs = errors.Join(errs, d.restoreCiliumHostIPs(true, router6FromK8s))
+		}
+		if resiliency.IsRetryable(errs) {
+			log.WithField(logfields.Attempt, retries).WithError(errs).Warnf("Failed to remove old router IPs from cilium_host.")
+			return false, nil
+		}
+		return true, errs
 	}
-	if option.Config.EnableIPv6 {
-		d.restoreCiliumHostIPs(true, router6FromK8s)
+	if err := resiliency.Retry(ctx, 100*time.Millisecond, 3, gcHostIPsFn); err != nil {
+		log.WithError(err).Error("Restore of the cilium_host ips failed. Manual intervention is required to remove all other old IPs.")
 	}
 
 	// restore endpoints before any IPs are allocated to avoid eventual IP
--- a/daemon/cmd/ciliumendpoints.go
+++ b/daemon/cmd/ciliumendpoints.go
@@ -100,7 +100,7 @@ func (d *Daemon) deleteCiliumEndpoint(
 				return nil
 			}
 			logwf.WithError(err).Error("Failed to get possibly stale ciliumendpoints from apiserver")
-			return resiliency.NewRetryableErr(err)
+			return resiliency.Retryable(err)
 		}
 		if cep.Status.Networking.NodeIP != node.GetCiliumEndpointNodeIP() {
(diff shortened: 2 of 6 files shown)


daemon/cmd/daemon.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
 	"golang.org/x/sync/semaphore"
+	"k8s.io/apimachinery/pkg/util/wait"
```
bimmlerd: I think we have `pkg/backoff` in cilium for exponential backoffs, did you consider using that instead of k8s?

daemon/cmd/daemon.go
```diff
@@ -1065,13 +1059,30 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
...
-		d.restoreCiliumHostIPs(false, router4FromK8s)
-	}
+	ipv6, router := false, router4FromK8s
 	if option.Config.EnableIPv6 {
-		d.restoreCiliumHostIPs(true, router6FromK8s)
+		ipv6, router = true, router6FromK8s
 	}
 
+	bo := wait.Backoff{
+		Duration: 500 * time.Millisecond,
```
bimmlerd: I think that's a long time to wait before retrying given this is on the critical path for daemon startup, I'd shorten this.

bimmlerd (changes requested):
Given we've just logged a warning before, I think we should delay startup by as much. generally lgtm, though I have no idea whether these errors should be considered retryable

daemon/cmd/daemon.go
derailed: @bimmlerd  Thanks for your input David! Best I can tell `pkg/backoff` expects you to manage the retry loop, thus opted for using k8s retry built-in instead which felt more appropriate for this use case.

daemon/cmd/daemon.go
derailed: Good point! I am not versed with netlink hence not sure what would be a reasonable delay to retry the service? Lowering it down to 100ms. Please pipe in if you think otherwise. Thank you!

bimmlerd:
Yeah I don't know enough about netlink either, I'll defer my review to @ti-mo.

daemon/cmd/daemon.go
```diff
@@ -1065,13 +1059,30 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 	// router IPs from the K8s resources if we weren't able to restore them
 	// from the fs. We must do this after IPAM because we must wait until the
 	// K8s resources have been synced. Part 2/2 of restoration.
-	if option.Config.EnableIPv4 {
-		d.restoreCiliumHostIPs(false, router4FromK8s)
-	}
+	ipv6, router := false, router4FromK8s
 	if option.Config.EnableIPv6 {
-		d.restoreCiliumHostIPs(true, router6FromK8s)
```
jibi, line 1075: I think this drifts from the original original behavior: in case both ipv4 and ipv6 are enabled, only v6 addresses will be restored, while before both v4 and v6 would have been restored

daemon/cmd/daemon.go
jibi: I would have also suggested to use `pkg/backoff`, mostly for consistency with the rest of the codebase, although I do agree that `wait` seems more handy to deal with retries :+1:

jibi (changes requested):

daemon/cmd/daemon.go
derailed: Rats! Great catch. Thank you Gilberto!

daemon/cmd/daemon.go
```diff
@@ -343,25 +345,21 @@ func (d *Daemon) restoreCiliumHostIPs(ipv6 bool, fromK8s net.IP) {
...
 func removeOldRouterState(ipv6 bool, restoredIP net.IP) error {
 	l, err := netlink.LinkByName(defaults.HostDevice)
-	if errors.As(err, &netlink.LinkNotFoundError{}) && restoredIP == nil {
-		// There's no old state remove as the host device doesn't exist and
-		// there's no restored IP anyway.
-		return nil
-	} else if err != nil {
-		return err
+	if err != nil {
+		if errors.As(err, &netlink.LinkNotFoundError{}) && restoredIP == nil {
```
ti-mo: Nit: `if errors.As() ... { return nil }` can be lifted out of the `if err != nil {}` block, it already does its own nil checking. If the `As` doesn't match, execution will fall through to the regular `if err != nil {}`, where the retry is issued.

Less indentation reads a bit easier.

daemon/cmd/daemon.go
```diff
@@ -343,25 +345,21 @@ func (d *Daemon) restoreCiliumHostIPs(ipv6 bool, fromK8s net.IP) {
...
+			// There's no old state remove as the host device doesn't exist and
+			// there's no restored IP anyway.
+			return nil
+		}
+		return resiliency.NewRetryableErr(err)
```
ti-mo: Could this be named `resiliency.Retry()` instead? It makes sense to use the imperative 'retry' here, since we want to signal an underlying layer to retry the operation. It also saves some typing and reading. :) I'd name the wrapper type `type Retryable` for brevity as well. Perhaps at some point it will hold more than just an error.

daemon/cmd/daemon.go
```diff
@@ -1065,13 +1059,29 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
...
-	}
-	if option.Config.EnableIPv6 {
-		d.restoreCiliumHostIPs(true, router6FromK8s)
+	bo := wait.Backoff{
+		Duration: 100 * time.Millisecond,
+		Factor:   1,
+		Jitter:   0.1,
+		Steps:    3,
+	}
+	err = wait.ExponentialBackoffWithContext(ctx, bo, func(ctx context.Context) (done bool, err error) {
```
ti-mo: Will this retry forever? At which point do we give up and crash the agent? (or drop the operation?)

daemon/cmd/daemon.go
derailed: Good point Timo! I am open to a rename but I think for the time being the naming make sense. 
`Retry()` feels a bit more obscure since we are not performing an actual retry here but wrapping the error to potentially trap it further up the chain. 
In the future, should we opt that additional context around the condition would be deemed useful we could rename accordingly?

daemon/cmd/daemon.go
derailed: This will retry at most 3 times before we give up.

daemon/cmd/daemon.go
derailed: Indeed. Great catch!

daemon/cmd/daemon.go
```diff
@@ -343,12 +345,7 @@ func (d *Daemon) restoreCiliumHostIPs(ipv6 bool, fromK8s net.IP) {
 	}
 
 	restoredIP := node.RestoreHostIPs(ipv6, fromK8s, fromFS, cidrs)
-	if err := removeOldRouterState(ipv6, restoredIP); err != nil {
-		log.WithError(err).Warnf(
-			"Failed to remove old router IPs (restored IP: %s) from cilium_host. Manual intervention is required to remove all other old IPs.",
-			restoredIP,
```
christarazi, line 349: I think it can still be nice to know what the restored IP is in the log msg so that the user can know which old IPs to remove

daemon/cmd/daemon.go
```diff
@@ -1065,13 +1059,29 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
...
+			errs = errors.Join(errs, d.restoreCiliumHostIPs(true, router6FromK8s))
+		}
+		if resiliency.IsRetryable(errs) {
+			log.WithError(errs).Warnf("Failed to remove old router IPs from cilium_host. Retrying...")
+			return false, nil
+		}
+		return true, errs
+	})
+	if err != nil {
+		log.WithError(err).Error("Restore host ips failed. Manual intervention is required to remove all other old IPs.")
```
christarazi:
```suggestion
		log.WithError(err).Error("Restore of the cilium_host IP failed. Manual intervention is required to remove all other old IPs.")
```

Just to not differ too much from the original log msg.

daemon/cmd/daemon.go
christarazi: I feel like this boilerplate should ideally be collapsed down into a helper function, so that the daemon code doesn't just increasingly become longer each time we add resiliency to the restore paths.

christarazi (changes requested):

daemon/cmd/daemon.go
derailed: No loss here, as we moved the log message further up the stack. 
Hence failed deletions will still log the respective host IPs that could not be deleted.

daemon/cmd/daemon.go
derailed: Good catch! Tx Chris!!

daemon/cmd/daemon.go
derailed: Indeed! I'll float an idea here and we can backfill the other prs if the proposal is deemed acceptable...

christarazi (approved):
Very clean now with the helper!

bimmlerd (approved):

daemon/cmd/daemon.go
ti-mo: Sure, then what about popping the `New` and/or `Err` parts off of the function name? We'd end up with `err = errors.Join(err, resiliency.Retryable(fmt.Errorf("failed to remove IP %s: %w", a.IP, e)))`. That pairs up nicely with `if resiliency.IsRetryable(errs) { ... }`. (Retryable() in the callee caught by IsRetryable in the caller)

And since we're exposing `IsRetryable()` as the canonical way of checking retryable errors, we should also unexport `type RetryableErr`.

> In the future, should we opt that additional context around the condition would be deemed useful we could rename accordingly?

By that time, there will be so many users that this turns into a huge change, meaning increased pain during backports. Let's just get this right from the beginning. :slightly_smiling_face:

derailed: Ok I'll make it so. Tx Timo!

christarazi:
It looks like the "build commits" job is failing because one of the commits is failing to build:

```
# github.com/cilium/cilium/daemon/cmd
Error: cmd/ciliumendpoints.go:103:22: undefined: resiliency.NewRetryableErr
Error: cmd/ciliumendpoints.go:130:21: undefined: resiliency.NewRetryableErr
```

Basically, each commit should be able to build independently.

daemon/cmd/daemon.go
```diff
@@ -343,12 +344,7 @@ func (d *Daemon) restoreCiliumHostIPs(ipv6 bool, fromK8s net.IP) {
 	}
 
 	restoredIP := node.RestoreHostIPs(ipv6, fromK8s, fromFS, cidrs)
-	if err := removeOldRouterState(ipv6, restoredIP); err != nil {
-		log.WithError(err).Warnf(
-			"Failed to remove old router IPs (restored IP: %s) from cilium_host. Manual intervention is required to remove all other old IPs.",
-			restoredIP,
-		)
-	}
+	return removeOldRouterState(ipv6, restoredIP)
```
jibi, line 347: nit: since all errors returned by `removeOldRouterState` are retryable, can we just decorate the eventual returned error here to make this more explicit (and avoid all the changes to the function)?

jibi (approved):

lmb:
Please resolve all conversations to get this mergeable.

daemon/cmd/daemon.go
derailed: @jibi - not keen on this idea as it should be a `explicit` developer choice here vs assuming all errors return by this call will be deemed retryable. If we were to return a non retryable error later on, devs might not be aware that error will be retried.

PR #27680, opened 2023-08-24
Fix KVStoreMesh metric's ConfigName

marseel:
Description in commit messages.

```release-note
Fixes name used for disabling KVStoreMesh metrics.
```

### DIFF
--- a/pkg/metrics/metric/metric.go
+++ b/pkg/metrics/metric/metric.go
@@ -189,3 +189,10 @@ type Opts struct {
 	// If true, the metric has to be explicitly enabled via config or flags
 	Disabled bool
 }
+
+func (b Opts) GetConfigName() string {
+	if b.ConfigName == "" {
+		return prometheus.BuildFQName(b.Namespace, b.Subsystem, b.Name)
+	}
+	return b.ConfigName
+}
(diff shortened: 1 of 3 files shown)


pkg/metrics/metric/metric.go
```diff
@@ -189,3 +189,10 @@ type Opts struct {
 	// If true, the metric has to be explicitly enabled via config or flags
 	Disabled bool
 }
+
+func (b Opts) GetConfigName() string {
+	if b.ConfigName == "" {
+		return prometheus.BuildFQName(b.Namespace, b.Subsystem, b.Name)
```
tommyp1ckles, line 195: I wonder if we should enforce that these are never empty

tommyp1ckles (approved):

pkg/metrics/metric/metric.go
marseel: So the only hard requirement is that `Name` is non-empty. That's why I used `BuildFQName` instead of regular `strings.Join`

giorio94 (approved):
Thanks!

jrajahalme (approved):

Issue #27417, opened 2023-08-10
CI: multi-pool-ipam-conformance-test - cilium agent does not become ready


giorio94:
## CI failure

Hit on https://github.com/cilium/cilium/pull/27242: https://github.com/cilium/cilium/actions/runs/5819064331/job/15776739547?pr=27242
Sysdump: [cilium-sysdump-out.zip](https://github.com/cilium/cilium/files/12311200/cilium-sysdump-out.zip)

```
     /¯¯\
 /¯¯\__/¯¯\    Cilium:             2 errors
 \__/¯¯\__/    Operator:           OK
 /¯¯\__/¯¯\    Envoy DaemonSet:    disabled (using embedded mode)
 \__/¯¯\__/    Hubble Relay:       OK
    \__/       ClusterMesh:        disabled

Deployment             hubble-relay       Desired: 1, Ready: 1/1, Available: 1/1
Deployment             cilium-operator    Desired: 1, Ready: 1/1, Available: 1/1
DaemonSet              cilium             Desired: 2, Ready: 1/2, Available: 1/2, Unavailable: 1/2
Containers:            cilium             Running: 2
                       hubble-relay       Running: 1
                       cilium-operator    Running: 1
Cluster Pods:          4/4 managed by Cilium
Helm chart version:    1.15.0-dev
Image versions         cilium             quay.io/cilium/cilium-ci:63f7ee533ca79edac909e20f9ff17f04961d688c: 2
                       hubble-relay       quay.io/cilium/hubble-relay-ci:63f7ee533ca79edac909e20f9ff17f04961d688c: 1
                       cilium-operator    quay.io/cilium/operator-generic-ci:63f7ee533ca79edac909e20f9ff17f04961d688c: 1
Errors:                cilium             cilium          1 pods of DaemonSet cilium are not ready
                       cilium             cilium-8crjq    unable to retrieve cilium status: command terminated with exit code 1
```

```
NAMESPACE            NAME                                                  READY   STATUS             RESTARTS      AGE     IP            NODE                          NOMINATED NODE   READINESS GATES
kube-system          cilium-8crjq                                          0/1     Running            1 (69s ago)   5m2s    172.18.0.2    chart-testing-control-plane   <none>           <none>
kube-system          cilium-operator-7dc59c9b8b-f79f2                      1/1     Running            0             5m2s    172.18.0.3    chart-testing-worker          <none>           <none>
[...]

PR #27684, opened 2023-08-24
daemon: don't wait for presence of unused CNC CRD

### LINKED ISSUE #27683: Agent shouldn't wait for `CiliumNodeConfig` CRD
Currently, cilium-agent waits for the `CiliumNodeConfig` CRD to be present in the cluster because it's included as one of the agent CRDs, but the agent doesn't actually operate on this CRD, it's only used in the `config` init container: https://github.com/cilium/cilium/blob/dcef00d653e1f9685e5bdadb1304d9b8679e44e6/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml#L402-L408

and the `build-config` CLI command already tolerates the CRD not existing: https://github.com/cilium/cilium/blob/dcef00d653e1f9685e5bdadb1304d9b8679e44e6/pkg/option/resolver/resolver.go#L238-L240

Can we remove the wait so that the agent can still function if the CRD is not present in the cluster? From an operator's perspective, I think there are several good reasons to not install this CRD:
- I don't want users to be able to override cilium config easily
- It's an alpha API, and I don't want to risk upgrades
- I want to reduce the API surface for its own sake

I don't think this requires any new flags. Just need to remove this line: https://github.com/cilium/cilium/blob/dcef00d653e1f9685e5bdadb1304d9b8679e44e6/pkg/k8s/synced/crd.go#L46

akhilles:
Currently, cilium-agent waits for the presence of the CiliumNodeConfig CRD but it doesn't need to. The commit allows for the agent to function even if the CRD is not present in the cluster.

Fixes: #27683

```release-note
daemon: don't wait for presence of unused CiliumNodeConfig CRD
```

### DIFF
--- a/pkg/k8s/synced/crd.go
+++ b/pkg/k8s/synced/crd.go
@@ -42,7 +42,6 @@ func agentCRDResourceNames() []string {
 		CRDResourceName(v2.CCNPName),
 		CRDResourceName(v2.CNName),
 		CRDResourceName(v2.CIDName),
-		CRDResourceName(v2alpha1.CNCName),
 		CRDResourceName(v2alpha1.CCGName),
 		CRDResourceName(v2alpha1.CPIPName),
 	}
@@ -100,6 +99,7 @@ func AllCiliumCRDResourceNames() []string {
 	return append(
 		AgentCRDResourceNames(),
 		CRDResourceName(v2.CEWName),
+		CRDResourceName(v2alpha1.CNCName),
 	)
 }
 


youngnick (approved):

akhilles:
The multi-pool-ipam-conformance-test test failure seems to be a known flake: https://github.com/cilium/cilium/issues/27417. @youngnick could you please re-run the tests?

Also, I'm not sure why this change is blocked by "needs-release-note-label". I did add a release note to the PR description.

julianwiedmann:
@joestringer this PR has been sitting for a while. Do you think someone besides @squeed could provide the needed review? Or should we proceed to get CI green and :ship: ?

akhilles:
Triage of test failures:

[Conformance E2E (ci-e2e)](https://github.com/cilium/cilium/actions/runs/6198047511/job/16827764982):
```
Test [echo-ingress-mutual-auth-spiffe]:
  ❌ echo-ingress-mutual-auth-spiffe/pod-to-pod/curl-ipv4-3: cilium-test/client2-646b88fb9b-srtq2 (10.244.3.242) -> cilium-test/echo-same-node-775456cfcf-4rxjx (10.244.3.208:8080)
```
This looks like https://github.com/cilium/cilium/issues/26351.

[Conformance AKS (ci-aks)](https://github.com/cilium/cilium/actions/runs/6198047379/job/16827803448), not marked as required:
```
connectivity test failed: failed to fetch cilium runtime config: error dialing backend: read unix @->/tunnel-uds/proxysocket: read: connection reset by peer
```

This looks like https://github.com/cilium/cilium/issues/27989.

joestringer:
> @joestringer this PR has been sitting for a while. Do you think someone besides @squeed could provide the needed review? Or should we proceed to get CI green and 🚢 ?

The key question I don't know the answer to is this: If the daemon doesn't wait for CNC, does that introduce some race condition or other issue with applying that configuration? Or was there never a reason for the agent to wait for this resource in the first place?

squeed:
A fair question. No, we do not need to wait for the CiliumNodeConfig. @youngnick, how did you find yourself in this situation?

The command `cilium build-config`, which is what actually consumes these resources, is tolerant of the CRD not existing.

squeed (approved):

squeed:
(does this change somehow prevent the CRD from being created?)

akhilles:
> Does this change somehow prevent the CRD from being created? I forget the agent's flow for CRD creation.

I've tested this and it doesn't affect CRD creation because the operator uses `AllCiliumCRDResourceNames` in [`CreateCustomResourceDefinitions`](https://github.com/cilium/cilium/blob/9a9c9aae5ec033456612e499c8adc37766701fd7/pkg/k8s/apis/cilium.io/client/register.go#L168). I believe the split of "agent" CRDs and "all" CRDs was initially done to support this use case: https://github.com/cilium/cilium/pull/13820.

akhilles:
Could the `dont-merge` be removed since we have some approvals and the tests are passing? :)

akhilles:
Anything I can do to help get this merged in?

squeed:
Just a poke was all that was needed :-)

PR #27687, opened 2023-08-24
ci: add scheduled runs for Ariane workflows

nbusseneau:
In 9949c5a1891aff8982bfc19e7fc195e7ecc2abf1, we introduced Ariane to handle CI workflows via `workflow_dispatch` events. One of the nice consequences was that we could finally move CI workflows for stable branches to the appropriate stable branch, instead of having to keep them in `main`.

However, this had the side effect of disabling running these workflows on a schedule, since the `schedule` events are only triggered for files present in the `main` branch.

In this commit, we introduce a `schedule`-based workflow on `main` whose sole purpose is to trigger these missing tests. It automatically detects which workflows to run based on the current Ariane configuration in that branch, and thus the only need in terms of maintenance for us should be to add / remove stable branches in the matrix as appropriate.

### DIFF
--- a/.github/workflows/ariane-scheduled.yaml
+++ b/.github/workflows/ariane-scheduled.yaml
@@ -1,0 +1,46 @@
+name: Ariane scheduled workflows
+
+on:
+  # Run every 6 hours
+  schedule:
+    - cron: '0 */6 * * *'
+
+permissions:
+  # To be able to access the repository with actions/checkout
+  contents: read
+  # To trigger workflows via workflow_dispatch
+  actions: write
+
+jobs:
+  ariane-scheduled:
+    strategy:
+      fail-fast: false
+      matrix:
+        branch: 
+          - "1.12"
+          - "1.13"
+          - "1.14"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout branch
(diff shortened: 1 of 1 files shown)


christarazi (approved):

.github/workflows/ariane-scheduled.yaml
```diff
@@ -0,0 +1,47 @@
...
+  # To trigger workflows via workflow_dispatch
+  actions: write
+
+jobs:
+  ariane-scheduled:
+    strategy:
+      fail-fast: false
+      matrix:
+        branch: 
+          - "1.11"
```
aanm: we can drop 1.11

aanm (approved):

.github/workflows/ariane-scheduled.yaml
nbusseneau: Nice! Dropped!

nbusseneau:
This does not need to run any CI, reviews are in, marking `ready-to-merge`.

joestringer:
@nbusseneau I'm seeing these scheduled workflows failing on forks, and it's sending me emails. Can we limit this to the cilium/cilium repo?

PR #27692, opened 2023-08-25
bpf: ct: document some unused fields in ct_entry struct

julianwiedmann:
Update the `ct_entry` struct to point out some (potentially) unused fields.

### DIFF
--- a/bpf/lib/common.h
+++ b/bpf/lib/common.h
@@ -934,7 +934,7 @@ struct ct_entry {
 	__u32 lifetime;
 	__u16 rx_closing:1,
 	      tx_closing:1,
-	      nat46:1,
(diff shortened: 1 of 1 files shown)


bpf/lib/common.h
```diff
@@ -927,14 +927,17 @@ struct ct_entry {
 	 */
 	union {
 		__u64 rx_bytes;
-		__u64 backend_id;
+		struct {
+			__u32 unused;
+			__u32 backend_id;
+		};
```
julianwiedmann: wait - there's an endianness aspect here, right? :disappointed:

julianwiedmann: yeah dropped this for now, need to have another look

borkmann (approved):

PR #27693, opened 2023-08-25
endpoint: Only perform full synchronization periodically

joamaki:
Performing a full dump of the policy map on every policy map synchronization
is expensive and is only necessary to catch rare cases where the agent's view
of the policy map has diverged from the kernel's view which should only happen
either due to kernel or other bugs or some other application modifying the
endpoint policy map.

To reduce the cost of synchronizing large endpoint policy maps perform the
full synchronization only periodically. The full synchronization is defaults
to 15 minutes. Configurable with the hidden option
"--policy-map-full-reconciliation-interval".

Fixes: 9dc1350fdb ("endpoint: Enhance policy map sync")

```release-note
endpoint: Only perform the full policy map synchronization periodically (every 15 minutes) to reduce overhead with large endpoint policy maps
```

### DIFF
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -840,6 +840,10 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	flags.Int(option.PolicyMapEntriesName, policymap.MaxEntries, "Maximum number of entries in endpoint policy map (per endpoint)")
 	option.BindEnv(vp, option.PolicyMapEntriesName)
 
+	flags.Duration(option.PolicyMapFullReconciliationIntervalName, 15*time.Minute, "Interval for full reconciliation of endpoint policy map")
+	option.BindEnv(vp, option.PolicyMapFullReconciliationIntervalName)
+	flags.MarkHidden(option.PolicyMapFullReconciliationIntervalName)
+
 	flags.Int(option.SockRevNatEntriesName, option.SockRevNATMapEntriesDefault, "Maximum number of entries for the SockRevNAT BPF map")
 	option.BindEnv(vp, option.SockRevNatEntriesName)
 
--- a/pkg/controller/manager.go
+++ b/pkg/controller/manager.go
@@ -56,6 +56,7 @@ func (m *Manager) updateController(name string, params ControllerParams) *manage
 	start := time.Now()
 
 	m.mutex.Lock()
+	defer m.mutex.Unlock()
 
 	if m.controllers == nil {
 		m.controllers = controllerMap{}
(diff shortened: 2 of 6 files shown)


pkg/endpoint/bpf.go
```diff
@@ -1374,25 +1380,33 @@ func (e *Endpoint) syncPolicyMapWithDump() error {
...
-			DoFunc: func(ctx context.Context) (reterr error) {
-				// Failure to lock is not an error, it means
-				// that the endpoint was disconnected and we
-				// should exit gracefully.
-				if err := e.lockAlive(); err != nil {
-					return controller.NewExitReason("Endpoint disappeared")
-				}
-				defer e.unlock()
-				return e.syncPolicyMapWithDump()
+	e.startSyncOnce.Do(func() {
```
joamaki: Opted to do this via `sync.Once` instead of changing where `syncPolicyMapController` is invoked from as there were multiple places allocating an endpoint (createEndpoint, parseEndpoint, ...) and no definite "start the endpoint" method. So seemed like keeping the original structure was safest.

jrajahalme (approved):
Thanks for fixing this, I had no idea we were issuing a full dump on each endpoint (policy) regeneration!

pkg/endpoint/bpf.go
```diff
@@ -1312,20 +1312,26 @@ func (e *Endpoint) dumpPolicyMapToMapState() (policy.MapState, error) {
...
-// It checks the current contents of the endpoint's PolicyMap and deletes any
-// PolicyKeys that are not present in the endpoint's desiredMapState. It then
-// adds any keys that are not present in the map. When a key from desiredMapState
-// is inserted successfully to the endpoint's BPF PolicyMap, it is added to the
-// endpoint's realizedMapState field. Returns an error if the endpoint's BPF
-// PolicyMap is unable to be dumped, or any update operation to the map fails.
+// syncPolicyMapWithDump is invoked periodically to perform a full reconciliation
+// of the endpoint's PolicyMap against the BPF maps to catch cases where either
+// due to kernel issue or user intervention the agent's view of the PolicyMap
+// state has converged from the kernel. A warning is logged if this method finds
```
derailed: nit: diverged?

daemon/cmd/daemon_main.go
```diff
@@ -838,6 +838,10 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	flags.Int(option.PolicyMapEntriesName, policymap.MaxEntries, "Maximum number of entries in endpoint policy map (per endpoint)")
 	option.BindEnv(vp, option.PolicyMapEntriesName)
 
+	flags.Duration(option.PolicyMapFullReconciliationIntervalName, 15*time.Minute, "Interval for full reconciliation of endpoint policy map")
```
derailed, line 843: nit: should the default be defined as a const?

pkg/endpoint/bpf.go
```diff
@@ -1374,25 +1380,33 @@ func (e *Endpoint) syncPolicyMapWithDump() error {
...
-					return controller.NewExitReason("Endpoint disappeared")
-				}
-				defer e.unlock()
-				return e.syncPolicyMapWithDump()
+	e.startSyncOnce.Do(func() {
+		ctrlName := fmt.Sprintf("sync-policymap-%d", e.ID)
+		e.controllers.UpdateController(ctrlName,
+			controller.ControllerParams{
+				Group: syncPolicymapControllerGroup,
+				DoFunc: func(ctx context.Context) (reterr error) {
```
derailed: Do we need a named return here?

derailed (approved):
@joamaki Nice work! just a couple comments...

pkg/endpoint/bpf.go
christarazi: This kind of invocation pattern is what `pkg/trigger` provides, but what it misses is the retrying ability of `pkg/controller`. A bit unfortunate that there's no mechanism to have both, but `sync.Once` works for now.

christarazi (commented):
The fix makes sense to me. How do we know that Cilium wasn't relying on this controller running frequently for correct operation?

joamaki:
We know Cilium wasn’t relying on this because we were not seeing the warnings that would be logged if full sync did catch something. Furthermore, @jrajahalme who wrote this originally did not intend this as he expected UpdateController to be a no-op after the first call and not a trigger to run it immediately.

daemon/cmd/daemon_main.go
joamaki: Good question. Half of the flags defined here seem to set the default directly and other half have it in `defaults` package. My personal preference is not to put it to `pkg/defaults` when it's used only here and only have things in `pkg/defaults` that are e.g. shared with operator and thus repeated. 

Don't mind either way. If you think it'd be nicer to have it in `pkg/defaults` I can move it there.

pkg/endpoint/bpf.go
joamaki: Ah yeah, might as well clean that old code up.

pkg/endpoint/bpf.go
joamaki: Thanks good catch.

pkg/endpoint/bpf.go
joamaki: This made me think we should just add this ability to `pkg/controller`. I added a `CreateController` method that creates the controller if it doesn't exist and does nothing if it does. This way we don't need the `sync.Once` etc.
WDYT?

pkg/controller/manager.go
```diff
@@ -81,33 +81,56 @@ func (m *Manager) updateController(name string, params ControllerParams) *manage
...
+	m.mutex.Lock()
+	if m.controllers != nil {
+		if _, exists := m.controllers[name]; exists {
+			m.mutex.Unlock()
+			return false
+		}
+	} else {
+		m.controllers = controllerMap{}
+	}
+	m.createControllerAndUnlock(name, params)
```
christarazi: Nit: I think I would prefer a `createControllerLocked` and just do the unlocking of the mutex at the caller level, rather than mixing responsibilities.

christarazi (approved):

pkg/controller/manager.go
joamaki: Didn't want to change the locking semantics (e.g. to hold both this manager and the globalStatus manager locks). But looks like we do that in e.g. `RemoveController` anyway. I'll change it to `createControllerLocked`.

pkg/endpoint/bpf.go
```diff
@@ -1339,12 +1345,18 @@ func (e *Endpoint) syncPolicyMapWithDump() error {
...
+const (
+	// syncPolicyMapInterval is the interval for periodic full reconciliation of
+	// the policy map to catch unexpected discrepancies with agent and kernel state.
+	syncPolicyMapInterval = 15 * time.Minute
+)
+
+func (e *Endpoint) startSyncPolicyMapController() {
 	ctrlName := fmt.Sprintf("sync-policymap-%d", e.ID)
-	e.controllers.UpdateController(ctrlName,
+	e.controllers.CreateController(ctrlName,
```
joestringer, line 1356: Super minor nit but it's odd to return a boolean and ignore it. Looking at the line of code here it's not obvious that the result is being ignored. Ideally the underlying function wouldn't return a boolean if it's unused in production code (I see it's used in tests), but this could at least signal that the result is deliberately ignored with a `_ = e.controllers.CreateController(...)`.

joestringer:
Huh, I was pondering why the policy trigger doesn't ratelimit these and then realized that's because the policy trigger is for triggering re-evaluation of the policy, not attempting a fresh synchronization of the synchronized policy to the BPF policymaps. So fair :+1:

gandro:
I managed to resolve the conflicts for v1.14, but the conflicts for v1.13 and v1.12 are more complex (i.e. there is no managedController type). Marking as `backport/author`.

PR #27697, opened 2023-08-25
Auth map GC on Endpoint delete

meyskens:
This change will trigger a endpoint list based auth map garbage collection on a delete call from the endpoint manager. This ensures instant deletion of an auth map entry if an endpoint gets deleted that was the last of a given security id.

Some of this is already done with the policy map GC. This change adds an aditional safeguard to it and does it on a delete event where the policy map only runs on scheduled  intervals.

```release-note
Auth map garbage collection will trigger if last local endpoint of a security identity was removed
```

### DIFF
--- a/pkg/auth/authmap_gc.go
+++ b/pkg/auth/authmap_gc.go
@@ -11,6 +11,8 @@ import (
 	"github.com/sirupsen/logrus"
 
 	datapathTypes "github.com/cilium/cilium/pkg/datapath/types"
+	"github.com/cilium/cilium/pkg/endpoint"
+	"github.com/cilium/cilium/pkg/endpointmanager"
 	"github.com/cilium/cilium/pkg/identity"
 	"github.com/cilium/cilium/pkg/identity/cache"
 	"github.com/cilium/cilium/pkg/lock"
@@ -36,6 +38,8 @@ type authMapGarbageCollector struct {
 	ciliumIdentitiesDiscovered map[identity.NumericIdentity]struct{}
 	ciliumIdentitiesSynced     bool
 	ciliumIdentitiesDeleted    map[identity.NumericIdentity]struct{}
+
+	endpointRepo endpointRepository
 }
 
 func (r *authMapGarbageCollector) Name() string {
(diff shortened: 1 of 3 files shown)


pkg/auth/authmap_gc_test.go
```diff
@@ -289,12 +292,54 @@ func Test_authMapGarbageCollector_cleanup(t *testing.T) {
...
+		1: {},
+		2: {},
+		3: {},
+	}
+	gc.ciliumIdentitiesSynced = true
+
+	err := gc.cleanupEndpoints(ctx)
+	assert.NoError(t, err)
+	assert.Len(t, authMap.entries, 1, "GC runs should delete all entries where the secrity ID no longer is in the endpoint map")
+	assert.Contains(t, authMap.entries, authKey{localIdentity: 3, remoteIdentity: 1, remoteNodeID: 100, authType: policy.AuthTypeSpire})
```
youngnick, line 334: Maybe also a test of the noop case - that if you run `cleanupEndpoints` and all the relevant identities are still present in the endpoint map, nothing is deleted?

pkg/auth/authmap_gc.go
```diff
@@ -387,3 +400,55 @@ func (r *authMapGarbageCollector) cleanupExpiredEntries(_ context.Context) error
...
+func (r *authMapGarbageCollector) subscribeToEndpointEvents(endpointManager endpointmanager.EndpointManager) {
+	endpointManager.Subscribe(r)
+}
+
+func (r *authMapGarbageCollector) EndpointCreated(ep *endpoint.Endpoint) {
+	// noop function to satisfy interface
+}
+
+func (r *authMapGarbageCollector) EndpointDeleted(ep *endpoint.Endpoint, conf endpoint.DeleteConfig) {
+	// when and endpoint got removed clean the authmap entries
```
youngnick:
```suggestion
	// when an endpoint got removed clean the authmap entries
```

youngnick (approved):
The tiniest of non-blocking format nits, but aside from that, LGTM. Nice work.

PR #27698, opened 2023-08-25
helm: fix envoy daemonset loglevel with multiple verbose debug groups

mhofstetter:
Currently, the Envoy DaemonSet loglevel is not set correctly if multiple verbose debug groups are passed as helm values. 

-> `Unknown verbose debug group: <...>` in Cilium Agent log

The problem is that the comma is used as group separator instead of a space. (Got confused because Go Viper flags are separated by a comma.)

This commit fixes this by changing the separator.


youngnick (approved):
Nice one.

mhofstetter:
rebased to `main` to pull in naming changes of github actions

nebril (approved):

PR #27705, opened 2023-08-25
bgpv1: Implement `AdvertisedPathAttributes` for `CiliumBGPNeighbor` using BGP routing policies

rastislavs:
This PR adds a new field `AdvertisedPathAttributes` into the `CiliumBGPNeighbor` part of the `CiliumBGPPeeringPolicy` CRD, which can be used to apply additional NLRI attributes (communities / local preference) to matched routes when advertising them to a BGP peer.

The implementation relies on the [GoBGP routing policies](https://github.com/osrg/gobgp/blob/master/docs/sources/policy.md), which are internally abstracted to a new `RoutePolicy` type and  new `Router` interface methods `AddRoutePolicy` + `RemoveRoutePolicy`.

For now, we support this for routes advertised for k8s LB services (matching by `CiliumLoadBalancerIPPool`s) and Pod CIDRs (matching by `CiliumNode`s).

For more details, please see the individual commits.

### Example:

`CiliumBGPPeeringPolicy` configuration:

```yaml
kind: CiliumBGPPeeringPolicy
spec:
  virtualRouters: 
  - localASN: 65000
    exportPodCIDR: true
    serviceSelector:
      matchLabels:
        advertise: bgp
    neighbors:
    - peerASN: 65000
      peerAddress: 172.0.0.1/32
      advertisedPathAttributes:
      # add community attribute to all routes advertised for the matched CiliumLoadBalancerIPPool
      - selectorType: CiliumLoadBalancerIPPool
        selector:
          matchLabels:
            environment: production
        communities:
          standard:
          - 65001:100
      # add local preference & community attributes to Pod CIDR routes of all nodes
      - selectorType: PodCIDR
        localPreference: 150
        communities:
          standard:
          - 65001:150
```

Matching a LB pool:
```yaml
apiVersion: "cilium.io/v2alpha1"
kind: CiliumLoadBalancerIPPool
metadata:
  name: "cilium-pool"
  labels:
    environment: production
spec:
  cidrs:
  - cidr: "192.168.100.0/24"
```

Causes adding a Community and setting appropriate Local Preference on advertised routes (`10.244.0.0/24` is a pod CIDR and `192.168.100.190/32` is a k8s service VIP):
```
$ cilium bgp routes advertised ipv4 unicast peer 172.0.0.1

VRouter   Prefix               NextHop     Age     Attrs
65000     10.244.0.0/24        172.0.0.2   3m31s   [{Origin: i} {LocalPref: 150} {Communities: 65001:150}] {Nexthop: 172.0.0.2}
65000     192.168.100.190/32   172.0.0.2   3m32s   [{Origin: i} {LocalPref: 100} {Communities: 65001:100}] {Nexthop: 172.0.0.2} 
```

(Note that Local Preference attribute is sent only to iBGP peers and defaults to `100`)

On the peered router:

```
tor# sh bgp ipv4 10.244.0.0/24
BGP routing table entry for 10.244.0.0/24, version 14
Paths: (1 available, best #1, table default)
  Not advertised to any peer
  Local
    172.0.0.2 from 172.0.0.2 (172.0.0.2)
      Origin IGP, localpref 150, valid, internal, best (First path received)
      Community: 65001:150
      Last update: Mon Aug 28 09:13:16 2023

tor# sh bgp ipv4 192.168.100.190/32
BGP routing table entry for 192.168.100.190/32, version 13
Paths: (1 available, best #1, table default)
  Not advertised to any peer
  Local
    172.0.0.2 from 172.0.0.2 (172.0.0.2)
      Origin IGP, localpref 100, valid, internal, best (First path received)
      Community: 65001:100
      Last update: Mon Aug 28 09:13:16 2023
```

### Release Note:
```release-note
Implement `AdvertisedPathAttributes` for `CiliumBGPNeighbor` in the `CiliumBGPPeeringPolicy` CRD to allow setting BGP Community and Local Preference path attributes for advertised BGP routes.
```


pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -101,6 +101,46 @@ type CiliumBGPNeighborGracefulRestart struct {
...
+// as a 4-byte decimal number or two 2-byte decimal numbers separated by a colon,
+// or a BGP Large Community Attribute value (RFC 8092) as three 4-byte decimal numbers separated by a colon.
+// +kubebuilder:validation:Pattern=`^\d+(:\d+)?(:\d+)?$`
+type BGPCommunityString string
+
+// CiliumBGPRouteAttributes can be used to apply additional path attributes
+// to matched routes when advertising them to a BGP peer.
+type CiliumBGPRouteAttributes struct {
+	// SelectorType defines type of the object on which the Selector applies.
+	// - For "Node" the Selector matches k8s node resources (resulting into PodCIDR routes).
```
ldelossa: Will this always just select PodCIDR routers? If so, why not just call this PodCIDR for selector type? 

Will Node express different type of "Types" in the future? If so, I don't see anyway to suggest a 'sub-type' or anything. 

Or is the reason this is 'Node' to better align with the 'Selector' as you'd put the label on the node?

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -101,6 +101,46 @@ type CiliumBGPNeighborGracefulRestart struct {
...
 
+// BGPCommunityString holds a string containing a BGP Community Attribute value (RFC 1997)
+// as a 4-byte decimal number or two 2-byte decimal numbers separated by a colon,
+// or a BGP Large Community Attribute value (RFC 8092) as three 4-byte decimal numbers separated by a colon.
+// +kubebuilder:validation:Pattern=`^\d+(:\d+)?(:\d+)?$`
+type BGPCommunityString string
+
+// CiliumBGPRouteAttributes can be used to apply additional path attributes
+// to matched routes when advertising them to a BGP peer.
+type CiliumBGPRouteAttributes struct {
```
ldelossa: Per RFC shouldn't this be called 'CiliumBGPPathAttributes'?

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -101,6 +101,46 @@ type CiliumBGPNeighborGracefulRestart struct {
...
+	// +kubebuilder:validation:Enum=CiliumNode;CiliumLoadBalancerIPPool
+	// +kubebuilder:validation:Required
+	SelectorType string `json:"selectorType"`
+
+	// Selector selects a group of objects of the MatchType
+	// resulting into routes that will be announced with the configured Attributes.
+	// If nil / not set, all objects of the SelectorType are selected.
+	//
+	// +kubebuilder:validation:Optional
+	Selector *slimv1.LabelSelector `json:"selector,omitempty"`
```
ldelossa, line 151: Label vs annotations, why'd you go with labels? Currently we use Node annotations for BGP-CP configuration. I don't feel strongly either way but it may align more to use one and stick with it.

pkg/bgpv1/types/bgp.go
```diff
@@ -57,6 +57,102 @@ type PathResponse struct {
...
+)
+
+// Policy represents a BGP routing policy, also called "route map" in some BGP implementations.
+// It can contain multiple Statements that are evaluated in the given order. Each Statement
+// contains conditions for matching a route and actions taken if a route matches the conditions.
+// Whenever a Statement matches a route and the action taken on it is to either accept or reject the route,
+// the policy evaluation for the given route stops, and no further Statements nor other Policies are evaluated.
+//
+// +deepequal-gen=true
+type Policy struct {
```
ldelossa: Consider naming this 'RoutePolicy'? Since you seem to refer to it by this name in most comment strings.

pkg/bgpv1/manager/signaled_store.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/cilium/cilium/pkg/hive"
+	"github.com/cilium/cilium/pkg/hive/cell"
+	"github.com/cilium/cilium/pkg/k8s/resource"
+
+	k8sRuntime "k8s.io/apimachinery/pkg/runtime"
+)
+
+// SignaledStore is a super set of the resource.Store for the BGP Control Plane reconcilers usage.
+// It automatically signals the BGP Control Plane whenever an event happens on the resource.
+type SignaledStore[T k8sRuntime.Object] interface {
```
ldelossa: I don't quite understand the naming of this? Can you explore or can we consider a better name?

ldelossa: Can reconcile.go import both the 'BGPCPSignaler' and this new 'SignaledStore' without causing an import cycle? 

If not, then we may want to reconsider where this lives, as the BGP-CP signaler may wind up in the preflight handler when constructing BGP servers.

:lotus_position:

ldelossa (changes requested):
Some questions to clarify, but overall code and structure looks good.

pkg/bgpv1/manager/signaled_store.go
rastislavs: Yeah, I'm not super happy about this naming either, so any suggestions are welcome :)

My thinking was: it is a `resource.Store` which automatically signals the BGP CP upon object events.
Then we already have a similar `DiffStore`, so wanted to keep the name similar to it as well.

rastislavs: I don't see an issue with that ATM, but let me double-check what would happen if a reconciler ended up in a different package than the `SignaledStore`.

pkg/bgpv1/types/bgp.go
rastislavs: yeah makes sense 👌

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: Hmm maybe it should?.. but what about the `OutgoingRouteAttributes` field name in the `CiliumBGPNeighbor` then? Should it also become `OutgoingPathAttributes`?

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: In k8s, labels are most-often used to select objects/resources of some other type, e.g. k8s Deployment selects Pods by labels. Note that also in BGP CP, we have `ServiceSelector` which selects services by labels, and `NodeSelector` which selects nodes by labels.

The node annotations are a bit different, as they are not selecting other resources, but passing some arbitrary configuration values. For that annotations work better.

This is the former case, so I think we are well aligned here.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: > Will this always just select PodCIDR routers? 

I was a bit forward-thinking about [Multi-Pool IPAM](https://docs.cilium.io/en/v1.14/network/kubernetes/ipam-multi-pool/), but for that it may be a better idea to select `CiliumPodIPPool` resources instead. So yeah, could be just PodCIDR after all.

> Or is the reason this is 'Node' to better align with the 'Selector' as you'd put the label on the node?

Right, as this is called SelectorType, I think it should suggest which type the `Selector` is actually selecting (which BTW is incorrect ATM in the description, it should be `CiliumNode` - I have already renamed this one myself few times :joy: ).

I'm not super-happy abut this name myself TBH. Maybe `PodCIDRNode`? But I am also fine with just `PodCIDR`.

ldelossa: Tough one, I can see the usefulness of both when mixed with NodeSelector, but in the sake of "calling things what they are" PodCIDR may make the most sense. Your call tho.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
ldelossa: I would think so, I'm not sure what a 'Route Attribute' is, I've only heard of 'Path Attribute' but I could just be missing some context also. But if we want to align to RFC I'd call these things 'PathAttributes'.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
ldelossa: Yup, totally see your point. resolving.

pkg/bgpv1/manager/signaled_store.go
ldelossa: Yup just a double check would ease my suspicions, I didn't hack the code to try myself.

ldelossa: Would you consider this object an 'EventSource' for BGP-CP. If its the thing that's generating "events" for the BGP-CP then it could be something like 'BGPControlPlaneEventSource"?

ldelossa: Would you consider this object an 'EventSource' for BGP-CP. If its the thing that's generating "events" for the BGP-CP then it could be something like 'BGPControlPlaneEventSource"?

pkg/bgpv1/gobgp/conversions.go
```diff
@@ -94,3 +94,113 @@ func ToAgentPaths(paths []*gobgp.Path) ([]*types.Path, error) {
...
+	// community actions
+	if len(apiStatement.Actions.AddCommunities) > 0 {
+		s.Actions.Community = &gobgp.CommunityAction{
+			Type:        gobgp.CommunityAction_ADD,
+			Communities: apiStatement.Actions.AddCommunities,
+		}
+	}
+
+	// local pref actions
+	if apiStatement.Actions.SetLocalPreference != nil {
```
YutaroHayakawa, line 174: Here we unconditionally apply local preference, but according to the RFC, we must not include local preference to eBGP peer. Does GoBGP automatically take care of checking that? If so, does it return an error or just silently remove the attribute?

https://datatracker.ietf.org/doc/html/rfc4271#section-5.1.5

pkg/bgpv1/gobgp/server.go
```diff
@@ -379,6 +383,105 @@ func (g *GoBGPServer) WithdrawPath(ctx context.Context, p types.PathRequest) err
...
+}
+
+func (g *GoBGPServer) getGlobalPolicyAssignment(policy *gobgp.Policy, policyType types.PolicyType) *gobgp.PolicyAssignment {
+	return &gobgp.PolicyAssignment{
+		Name:          globalPolicyAssignmentName,
+		Direction:     toGoBGPPolicyDirection(policyType),
+		DefaultAction: gobgp.RouteAction_NONE, // no change to the default action
+		Policies:      []*gobgp.Policy{policy},
+	}
+}
```
YutaroHayakawa, line 482: Why do we use global policy instead of per-neighbor policy? Don't we end up with the linear policy evaluation of # of neighbors within GoBGP in this case?

pkg/bgpv1/manager/policy_reconciler.go
```diff
@@ -0,0 +1,236 @@
...
+		communities = append(communities, string(c))
+	}
+	sort.Strings(communities)
+
+	if len(v4Prefixes) > 0 {
+		policy.Statements = append(policy.Statements, policyStatement(neighborAddress, v4Prefixes, attrs.LocalPreference, communities))
+	}
+	if len(v6Prefixes) > 0 {
+		policy.Statements = append(policy.Statements, policyStatement(neighborAddress, v6Prefixes, attrs.LocalPreference, communities))
+	}
```
YutaroHayakawa: If we merge all rules into single Statements list, do we need to differentiate IPv4 and IPv6?

pkg/bgpv1/manager/policy_reconciler.go
```diff
@@ -0,0 +1,236 @@
...
+	return policy, nil
+}
+
+// routeAttributesPolicyName returns a hash of (NeighborAddress, SelectorType, Selector) to be used
```
YutaroHayakawa: Can't we just concatenate these three? I feel it is more human-readable and easy to debug. Do we have character length limit on GoBGP-side or something?

YutaroHayakawa (changes requested):
I did an initial look. Thanks! This is cool! I have several questions.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: What about `CiliumBGPPathAttributes` for the struct name + `AdvertisedRoutesPathAttributes` for the field name?
To also make the later matching the CLI "bgp routes advertised".

pkg/bgpv1/manager/signaled_store.go
rastislavs: Hmm, it's not only the source of events, it also provides api to list resources etc.
What about `BGPControlPlaneResourceStore` ? It is really that - BGP CP-specific resource store (with automatic BGP CP event generation).

pkg/bgpv1/gobgp/conversions.go
rastislavs: In this case, GoBGP is "smart enough" to not include local preference path attribute on Paths advertised to eBGP peers even if a matched policy has this action configured. I think eventually it is good, as we can (if needed later, not in this PR) reuse the same policy for multiple peers, no matter whether they are iBGP or eBGP.

pkg/bgpv1/gobgp/server.go
rastislavs: Unfortunately, this is how it s implemented in GoBGP:
-  [per-neighbor policy](https://github.com/osrg/gobgp/blob/master/docs/sources/policy.md#route-server-policy-model) can be used only in the route server mode.
- for non-route-server mode, only [global import/export policy](https://github.com/osrg/gobgp/blob/master/docs/sources/policy.md#policy-model) can be used, and if it is neighbor-specific, it needs to be part of the rules (a per-neighbor policy would not be accepted by GoBGP).

As we now do not use the route-server mode, this is the only option that works for us. I will put a comment into the code clarifying this.

Have not confirmed that with the code, but I guess you are right, policy evaluation in the non-route-server mode will most likely be linear.

pkg/bgpv1/manager/policy_reconciler.go
rastislavs: Yeah we need to differentiate here - families can not be mixed in a single Statement, but it is possible to have multiple Statements with different families in a single policy. Will add a comment clarifying this.

pkg/bgpv1/manager/policy_reconciler.go
rastislavs: Nope there is no limit. I just thought that having a string representation of a label Selector in the name may be a bit strange, but so is hash :) so yeah we can go with that as well. Or only hash the Selector and use plain-text NeighborAddress and SelectorType.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: "calling things what they are" makes sense, will change it to `PodCIDR`

rastislavs: renamed to `PodCIDR`

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: Renamed to `CiliumBGPPathAttributes` (type name) + `AdvertisedPathAttributes` (field name)

pkg/bgpv1/types/bgp.go
rastislavs: Renamed to `RoutePolicy`

pkg/bgpv1/manager/signaled_store.go
rastislavs: Eventually I went with the name `BGPCPResourceStore`

pkg/bgpv1/gobgp/conversions.go
rastislavs: Left a comment about the behavior in the code

pkg/bgpv1/gobgp/server.go
rastislavs: Left comments about this in the `AddRoutePolicy` method

pkg/bgpv1/manager/policy_reconciler.go
rastislavs: Left a comment about this in the code.

pkg/bgpv1/manager/policy_reconciler.go
rastislavs: Eventually I went with human-readable NeighborAddress & SelectorType and hashed Selector. That renders the most meaningful policy names IMO, for example:
```
172.0.0.1/32-CiliumLoadBalancerIPPool-325f8fc93e320b4bb1b1ea186c0f2d53e787175a3b5725c4be96cfd93fcf8f2e
```

Note that without any hashing it would be something like this:
```
172.0.0.1/32-CiliumLoadBalancerIPPool-&LabelSelector{MatchLabels:map[string]string{environment: production,},MatchExpressions:[]LabelSelectorRequirement{},}
```

pkg/bgpv1/manager/signaled_store.go
rastislavs: Tried using the `BGPCPResourceStore` from a reconciler outside of the `pkg/bgpv1/manager` 🧘
No issues with the import cycles.

pkg/bgpv1/gobgp/conversions.go
YutaroHayakawa: Nice 👍 Resolving.

YutaroHayakawa (approved):
My concerns are addressed. LGTM 👍

ldelossa (approved):

harsimran-pabla (approved):
Great work!!

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -29,6 +29,10 @@ const (
 	DefaultBGPGRRestartTimeSeconds = 120
 	// BGPLoadBalancerClass defines the BGP Control Plane load balancer class for Services.
 	BGPLoadBalancerClass = "io.cilium/bgp-control-plane"
+	// PodCIDRSelectorName defines the name for a selector matching Pod CIDRs.
```
danehans: This is a selector for IPAM pod CIDRs, correct? Might not be a bad idea to provide a bit more clarity to not confuse it with multi-pool IPAM pod CIDRs.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -101,6 +105,46 @@ type CiliumBGPNeighborGracefulRestart struct {
 	RestartTimeSeconds *int32 `json:"restartTimeSeconds,omitempty"`
 }
 
+// BGPCommunityString holds a string containing a BGP Community Attribute value (RFC 1997)
+// as a 4-byte decimal number or two 2-byte decimal numbers separated by a colon,
+// or a BGP Large Community Attribute value (RFC 8092) as three 4-byte decimal numbers separated by a colon.
+// +kubebuilder:validation:Pattern=`^\d+(:\d+)?(:\d+)?$`
```
danehans: This regex is a bit more permissive than what the referenced RFCs define. For example:

1. It does not limit the number of digits, so excessively large numbers that exceed 4-byte or 2-byte sizes will be allowed.

2. Although it permits a single number, a number followed by one colon and another number, or a sequence of three numbers separated by colons, it will not differentiate between what constitutes a BGP Community Attribute and what constitutes a BGP Large Community Attribute.

WDYT about using:
```go
// +kubebuilder:validation:Pattern:=^([0-9]{1,10}:[0-9]{1,10})|([0-9]{1,10})|([0-9]{1,10}:[0-9]{1,10}:[0-9]{1,10})$

```

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -101,6 +105,46 @@ type CiliumBGPNeighborGracefulRestart struct {
...
+type BGPCommunityString string
+
+// CiliumBGPPathAttributes can be used to apply additional path attributes
+// to matched routes when advertising them to a BGP peer.
+type CiliumBGPPathAttributes struct {
+	// SelectorType defines type of the object on which the Selector applies.
```
danehans: nit: s/type of the object/the object type/

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -101,6 +105,46 @@ type CiliumBGPNeighborGracefulRestart struct {
...
+	// - For "PodCIDR" the Selector matches k8s CiliumNode resources (resulting into PodCIDR routes).
```
danehans: Not all k8s CiliumNode resources but just for the ones that this CiliumBGPPeeringPolicy applies to, correct? This is for the IPAM pod CIDRs, correct?

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -101,6 +105,46 @@ type CiliumBGPNeighborGracefulRestart struct {
...
+	// - For "CiliumLoadBalancerIPPool" the Selector matches CiliumLoadBalancerIPPool custom resources
+	//   (resulting into IP pool's CIDRs routes).
+	//
+	// +kubebuilder:validation:Enum=PodCIDR;CiliumLoadBalancerIPPool
+	// +kubebuilder:validation:Required
+	SelectorType string `json:"selectorType"`
+
+	// Selector selects a group of objects of the MatchType
```
danehans: Should this be SelectorType instead of MatchType? Selector defines a label selector for matching objects of SelectorType type, correct?

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -101,6 +105,46 @@ type CiliumBGPNeighborGracefulRestart struct {
...
+	// If nil / not set, all objects of the SelectorType are selected.
+	//
+	// +kubebuilder:validation:Optional
+	Selector *slimv1.LabelSelector `json:"selector,omitempty"`
+
+	// Communities defines a set of community values advertised
+	// in the BGP Communities path attribute.
+	//
+	// +kubebuilder:validation:Optional
+	Communities []BGPCommunityString `json:"communities,omitempty"`
```
danehans: Does the RFC define a maximum number of communities that can be advertised in the BGP Communities path attribute? If so, consider adding the `// +kubebuilder:validation:MaxItems
<int>` marker.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -101,6 +105,46 @@ type CiliumBGPNeighborGracefulRestart struct {
...
+
+	// LocalPreference defines the preference value advertised
+	// in the BGP Local Preference path attribute.
+	//
```
danehans, line 164: nit: Although the `+Optional` marker is used in this and `Communities` fields, it's nice to have a line that defines the behavior if the field is unspecified (like what you did for the `Selector` field).

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -182,6 +226,12 @@ type CiliumBGPNeighbor struct {
 	//
 	// +kubebuilder:validation:Optional
 	Families []CiliumBGPFamily `json:"families"`
+	// AdvertisedPathAttributes can be used to apply additional path attributes
+	// to selected routes when advertising them to the peer.
+	// If empty / nil, no additional path attributes are advertised.
+	//
+	// +kubebuilder:validation:Optional
+	AdvertisedPathAttributes []CiliumBGPPathAttributes `json:"advertisedPathAttributes,omitempty"`
```
danehans, line 257: Same as https://github.com/cilium/cilium/pull/27705/files#r1318871832

pkg/bgpv1/gobgp/server.go
```diff
@@ -342,6 +346,28 @@ func (g *GoBGPServer) RemoveNeighbor(ctx context.Context, n types.NeighborReques
 	return nil
 }
 
+// ResetNeighbor resets BGP peering with the provided neighbor address.
+func (g *GoBGPServer) ResetNeighbor(ctx context.Context, r types.ResetNeighborRequest) error {
+	// for this request we need a peer address without prefix
+	peerAddr := r.PeerAddress
```
danehans, line 352: Should `PeerAddress` be `PeerPrefix` since `ResetNeighborRequest` is storing a prefix? Otherwise, maybe L353-354 should be called when storing values of type `PeerAddress`.

pkg/bgpv1/manager/route_policy_reconciler.go
```diff
@@ -0,0 +1,269 @@
...
+			RouteAction:        types.RoutePolicyActionNone, // continue with the processing of the next statements / policies
+			SetLocalPreference: localPref,
+			AddCommunities:     communities,
+		},
+	}
+}
+
+// peerAddressFromPolicy returns the first neighbor address found in a routing policy.
+func peerAddressFromPolicy(p *types.RoutePolicy) string {
+	for _, s := range p.Statements {
```
danehans, line 273: Check if `p` is `nil` and return `""`.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: Right, this is only for standard IPAM, not multi-pool. We may add multi-pool support by selecting `CiliumPodIPPool` custom resources later, after https://github.com/cilium/cilium/pull/27100 is in.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: Your pattern does not work as expected, for example `45:whatever` matches it.

But I get the point with limiting the digits, we can go with something like: 

`^[0-9]{1,10}(:[0-9]{1,10})?(:[0-9]{1,10})?$`

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: There is no fixed maximum number specified in the RFC (`optional transitive attribute of variable length`)

pkg/bgpv1/gobgp/server.go
rastislavs: Not sure, as out CRD API - `CiliumBGPNeighbor.PeerAddress` is actually a prefix and that is how we pass it to `AddNeighbor` / `UpdateNeighbor`, so I think we should be consistent here.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: I am not aware of any fixed limitation on the amount of path attributes. There is a limitation on maximum message size, but it is hard to derive the maximum path attributes from it as they are variable length.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: Correct, this config only affects routes that are already being advertised based on `CiliumBGPPeeringPolicy`, does not cause advertising any new routes. Clarified it a bit more in the description.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
harsimran-pabla: For standard community, this would be the regex 

`([0-9]|[1-9][0-9]{1,3}|[1-5][0-9]{4}|6[0-5][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5]):([0-9]|[1-9][0-9]{1,3}|[1-5][0-9]{4}|6[0-5][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])`

Regex for other types is also defined in this : 
https://datatracker.ietf.org/doc/draft-ietf-idr-bgp-model/ 

FYI, double-check regex - the above one was not correct in the RFC.

rastislavs: Hmm, GoBGP also allows single decimal uint32 number, do we want to allow it or we want to stick with formats defined in the YANG model RFC?

rastislavs: So based on the YANG RFC, this would be something like:


`^(([0-9]|[1-9][0-9]{1,3}|[1-5][0-9]{4}|6[0-5][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5]):([0-9]|[1-9][0-9]{1,3}|[1-5][0-9]{4}|6[0-5][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5]))$|^((4[0-2][0-9][0-4][0-9][0-6][0-7][0-2][0-9][0-6]|[1-3][0-9]{9}|[1-9]([0-9]{1,7})?[0-9]|[0-9]):(4[0-2][0-9][0-4][0-9][0-6][0-7][0-2][0-9][0-6]|[1-3][0-9]{9}|[1-9]([0-9]{1,7})?[0-9]|[0-9]):(4[0-2][0-9][0-4][0-9][0-6][0-7][0-2][0-9][0-6]|[1-3][0-9]{9}|[1-9]([0-9]{1,7})?[0-9]|[0-9]))$
`

Is this what we want here? I am just afraid that there might be some corner case that would not be obvious right now (especially as @harsimran-pabla already found a bug in the RFC :)) And it is not very helpful for human readers.

rastislavs: Eventually I decided to split the Standard and Large communities into two different types, so that the validation patterns can be sane, and we have better distinction between the two in the code. BTW, the pattern from the aforementioned YANG model was wrong also for Large community. I have tested those in the code with a regexp validator and seem to be working fine for me, but feel free give it another look & test.

pkg/bgpv1/types/bgp.go
```diff
@@ -46,6 +46,23 @@ type NeighborRequest struct {
...
+	SoftResetDirectionIn SoftResetDirection = iota
+	SoftResetDirectionOut
+	SoftResetDirectionBoth
+)
+
+// ResetNeighborRequest contains parameters used when resetting a BGP peer
+type ResetNeighborRequest struct {
+	PeerAddress        string
+	Soft               bool
+	SoftResetDirection SoftResetDirection
```
danehans, line 62: Instead of exposing a `bool`, could `SoftResetDirection` be a pointer, and if `nil`, soft reset would be considered `false`? This can be resolved in a follow-up PR if you agree.

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
```diff
@@ -101,6 +106,68 @@ type CiliumBGPNeighborGracefulRestart struct {
...
+	// If nil / not set, all objects of the SelectorType are selected.
+	//
+	// +kubebuilder:validation:Optional
+	Selector *slimv1.LabelSelector `json:"selector,omitempty"`
+
+	// Communities defines a set of community values advertised in the supported BGP Communities path attributes.
+	// If nil / not set, no BGP Communities path attribute will be advertised.
+	//
+	// +kubebuilder:validation:Optional
+	Communities *BGPCommunities `json:"communities,omitempty"`
```
danehans, line 157: What is the expected behavior if `Communities` is defined but `Standard` and `Large` is `nil` since they're both optional? If this is unexpected, add some validation logic to the reconcile loop. I think this can be done in a follow-on PR if you agree.

danehans (approved):
This is a really nice feature. I have a few nits that can be resolved in a follow-up PR.

nathanjsweet (approved):

pkg/k8s/apis/cilium.io/v2alpha1/bgpp_types.go
rastislavs: That would be fine IMO, the behavior would be: nil `Standard` and `Large` in non-nil `Communities` "container" would produce no community path attribute.

pkg/bgpv1/types/bgp.go
rastislavs: I am not sure here, I feel that being explicit sometimes may be better for readability - and pointer type requires an assignment to a var upon initialization to use a constant value, so the API is not that easy to use... but yeah, if you feel like it would be better, we can try that in a separate PR.

PR #27708, opened 2023-08-25
Setup Renovate for SPIRE deployment

meyskens:
This adds a check for the spire-server and spire-agent Docker images for Renovate.
I also fixes the busybox image to be allowed newer than 1.35.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -211,6 +211,34 @@
       ]
     },
     {
+      "groupName": "spire-images",
+      "matchFiles": [
+        "install/kubernetes/cilium/values.yaml.tmpl"
+      ],
+      "matchPackageNames": [
+        "ghcr.io/spiffe/spire-agent",
+        "ghcr.io/spiffe/spire-server"
+      ],
+      "matchBaseBranches": [
+        "main",
+      ],
+      "allowedVersions": ">1.6",
+    },
+    {
+      "groupName": "spire-images",
+      "matchFiles": [
+        "install/kubernetes/cilium/values.yaml.tmpl"
+      ],
+      "matchPackageNames": [
+        "ghcr.io/spiffe/spire-agent",
+        "ghcr.io/spiffe/spire-server"
+      ],
+      "matchBaseBranches": [
+        "v1.14",
+      ],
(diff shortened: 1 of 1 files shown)


bimmlerd (approved):

.github/renovate.json5
```diff
@@ -201,6 +201,21 @@
...
+      ],
+      "matchPackageNames": [
+        "ghcr.io/spiffe/spire-agent",
+        "ghcr.io/spiffe/spire-server"
+      ],
+      "matchBaseBranches": [
+        "main",
+        "v1.14",
+      ],
+      "allowedVersions": ">1.6",
```
joestringer, line 225: How do we want to bound the updates here?

As-is, this will happily update SPIRE to any arbitrary future version, whether compatible or not. I guess CI will catch incompatibilites, so not such a big deal.

As-is, for the next ~15 months we'll continually update v1.14 branch to any latest SPIRE version. This could mean that when upgrading Cilium 1.14.10 to 1.14.11, it tries to update SPIRE 1.6 to 1.7 or some later version. I'm not sure that's what we want.

joestringer: We don't have any `>` checks in this file today, and I think it's because we don't want to update minor or major versions of dependencies in case that causes breakage. We mostly assume that the versions should stick within the current minor release to pick up bugfixes & security updates... then if it turns out that the version goes EOL we would need to revisit and consciously bump, rather than automatically introducing significant version changes at some point when users expect the default versions to be relatively stable.

joestringer: ^ For `main`, I think that in principle I'm probably okay with `>` being used, but I'm more concerned for older branches. I'd be curious if other Cilium org members have other insights or opinions to bring on this topic.

aditighag:
@joestringer Please remove the ready-to-merge label added by MLH if there is still discussion ongoing on the PR.

.github/renovate.json5
meyskens: Good point on 1.14 branch. I have changed it so 1.14 stays on 1.6

joestringer (approved):

PR #27710, opened 2023-08-25
ipam, metrics: Add new capacity metric

### LINKED ISSUE #27166: Simplify and improve IPAM metrics agnostic of IPAM mode
We currently have two different sets of IPAM metrics:

* Cloud-based IPAM: https://github.com/cilium/cilium/blob/6ba173d9ba392e05ff58930b482d697e00c70805/pkg/ipam/metrics/metrics.go#L47
* Non-cloud-based IPAM: https://github.com/cilium/cilium/blob/6ba173d9ba392e05ff58930b482d697e00c70805/pkg/metrics/metrics.go#L1042

If you are running on ENI, Azure, or Alibaba, the first set of metrics are used to monitor IPAM. Anything else like cluster-pool or kubernetes IPAM mode, the second set of metrics are used. However, the problem with the latter is that there's no way to monitor the capacity from the metrics.

christarazi:
- ipam, metrics: Add new capacity metric
- ipam, metrics: Rename metric for consistency

Followups following the merge of this PR: https://github.com/cilium/cilium/issues/27716

--- 

Most important commit for convenience:  

> ipam, metrics: Add new capacity metric
>     
> This commit implements a new metric which tracks the total IPAM
> capacity, which to be clear is not the current capacity, but rather size
> of the IPAM pool.
>     
> Users can now track the currently allocated count of IPs across all IPAM
> modes by taking this metric and subtracting by number of IPs allocated.
>     
> Previously, this information was only visible from `cilium status`.
>     
> In Prometheus, the query would be:
>     
> ```
> cilium_ipam_capacity{family=~"*"} - cilium_ip_addresses{family=~"*"}
> ```
>     

---

Fixes: https://github.com/cilium/cilium/issues/27166

### DIFF
--- a/Documentation/observability/metrics.rst
+++ b/Documentation/observability/metrics.rst
@@ -477,6 +477,7 @@ IPAM
 ======================================== ============================================ ========== ========================================================
 Name                                     Labels                                       Default    Description
 ======================================== ============================================ ========== ========================================================
+``ipam_capacity``                        ``family``                                   Enabled    Total number of IPs in the IPAM pool labeled by family
 ``ipam_events_total``                                                                 Enabled    Number of IPAM events received labeled by action and datapath family type
 ``ip_addresses``                         ``family``                                   Enabled    Number of allocated IP addresses
 ======================================== ============================================ ========== ========================================================
--- a/Documentation/operations/upgrade.rst
+++ b/Documentation/operations/upgrade.rst
@@ -315,7 +315,7 @@ Annotations:
   Upgrading from Cilium 1.14.x or earlier to 1.15.y or later does not
   trigger this problem. Downgrading from Cilium 1.15.y or later to Cilium
   1.14.x or earlier may trigger this problem.
-  
+
 .. _upgrade_cilium_cli_helm_mode:
 
 Cilium CLI
@@ -339,6 +339,11 @@ Helm Options
 * Values ``clustermesh.apiserver.tls.ca.cert`` and ``clustermesh.apiserver.tls.ca.key``
   were deprecated in Cilium 1.14 in favor of ``tls.ca.cert`` and ``tls.ca.key`` respectively,
   and have been removed. The ```clustermesh-apiserver-ca-cert`` secret is no longer generated.
+
+Added Metrics
+~~~~~~~~~~~~~
+
+* ``cilium_ipam_capacity``
 
 Changed Metrics
 ~~~~~~~~~~~~~~~
--- a/pkg/ipam/multipool.go
+++ b/pkg/ipam/multipool.go
@@ -718,6 +718,24 @@ func (c *multiPoolAllocator) Dump() (map[string]string, string) {
 	return c.manager.dump(c.family)
 }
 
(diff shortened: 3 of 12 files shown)


pkg/ipam/multipool.go
```diff
@@ -718,6 +718,25 @@ func (c *multiPoolAllocator) Dump() (map[string]string, string) {
...
+		switch c.family {
+		case IPv4:
+			p = pool.v4
+		case IPv6:
+			p = pool.v6
+		}
+		if p == nil {
+			continue
+		}
+		_, _, c, _, _ := p.dump()
```
tommyp1ckles: dump seems to iterate all the ips using `ForEach`.  Might be better to have a seperate function to compute just the utilization.

Documentation/observability/metrics.rst
```diff
@@ -477,6 +477,7 @@ IPAM
 ======================================== ============================================ ========== ========================================================
 Name                                     Labels                                       Default    Description
 ======================================== ============================================ ========== ========================================================
+``ipam_capacity``                        ``family``                                   Enabled    Total number of IPs in the IPAM pool labeled by family
```
tommyp1ckles, line 480: Is ipam_capacity the right name here?

pkg/ipam/multipool.go
christarazi: Done

Documentation/observability/metrics.rst
christarazi: Open to suggestions.

Documentation/operations/upgrade.rst
```diff
@@ -315,7 +315,12 @@ Annotations:
   Upgrading from Cilium 1.14.x or earlier to 1.15.y or later does not
   trigger this problem. Downgrading from Cilium 1.15.y or later to Cilium
   1.14.x or earlier may trigger this problem.
-  
+
+Added Metrics
```
ghost: There is a "Changed Metrics" section in the upgrade notes, can you move the "Added Metrics" section to directly precede it?

ghost (changes requested):
One suggestion for the docs structure.

Also I got confused by the promql in the commit message:
```
cilium_ipam_capacity{family=~"*"} - cilium_ip_addresses{family=~"*"}
```
This query used as is won't work because `*` is not a correct regex. Did you mean to use it as a placeholder? Nothing critical, but I would rather write something like
```
cilium_ipam_capacity{family="ipv4"} - cilium_ip_addresses{family="ipv4"}
```
as an example.

Documentation/operations/upgrade.rst
christarazi: Ah, didn't see that other section. Moved.

ghost (approved):
looks good

aditighag (approved):

Documentation/observability/metrics.rst
tommyp1ckles: nevermind I think this makes sense

tommyp1ckles (approved):

PR #27713, opened 2023-08-25
ipam: when a CiliumNode is removed, delete node label from metrics.

tommyp1ckles:
Affects:

* operator_ipam_available_ips
* operator_ipam_used_ips
* operator_ipam_needed_ips

Which have the label "target_name", previously when a Node was deleted the metric continued to be emitted by the Prometheus exporter, leading to confusing sum() values across a cluster.


christarazi (approved):
TIL about `DeleteLabelValues`. I thought it wasn't possible. Thanks for the fix. 

A few CI failures around build. Also, it would be good to put a `Fixes` tag for the commit that this PR is fixing so we know where to backport this to. I'm pretty sure it's only for 1.14 so I've added that already.

PR #27714, opened 2023-08-25
Resiliency: Node manager reconciliation path yields unchecked errors

derailed:

      GitHub issue.

> NOTE: Initial pass!

Found several places in the node manager control loop where we are ignoring potential errors and either not logging them or worth not bubbling up to their respective call sites.

> NOTE! Please go thru this PR with a fine comb as I don't quite have the necessary chops on whether some of these should be hard errors or not. 
Thank you!!

Fixes: #issue-number

### DIFF
--- a/pkg/datapath/linux/ipsec/ipsec_linux.go
+++ b/pkg/datapath/linux/ipsec/ipsec_linux.go
@@ -34,6 +34,7 @@ import (
 	"github.com/cilium/cilium/pkg/node"
 	"github.com/cilium/cilium/pkg/nodediscovery"
 	"github.com/cilium/cilium/pkg/option"
+	"github.com/cilium/cilium/pkg/resiliency"
 )
 
 type IPSecDir string
@@ -205,10 +206,13 @@ func xfrmStateReplace(new *netlink.XfrmState) error {
 		}
 	}
 
-	oldXFRMMark := &netlink.XfrmMark{
-		Value: ipSecXfrmMarkSetSPI(linux_defaults.RouteMarkEncrypt, uint8(new.Spi)),
-		Mask:  linux_defaults.IPsecOldMarkMaskOut,
-	}
+	var (
+		oldXFRMMark = &netlink.XfrmMark{
+			Value: ipSecXfrmMarkSetSPI(linux_defaults.RouteMarkEncrypt, uint8(new.Spi)),
+			Mask:  linux_defaults.IPsecOldMarkMaskOut,
+		}
+		errs = resiliency.NewErrorSet("failed to delete old xfrm states", len(states))
+	)
 	for _, s := range states {
 		// This is the XFRM OUT state from a previous Cilium version.
 		// Because its mark matches the new mark (0xXXXX3e00/0xffffff00 ∈
@@ -228,7 +232,7 @@ func xfrmStateReplace(new *netlink.XfrmState) error {
 			xfrmMarkEqual(s.Mark, oldXFRMMark) && s.Spi == new.Spi {
 			err := netlink.XfrmStateDel(&s)
 			if err != nil {
-				scopedLog.WithError(err).Error("Failed to remove old XFRM state")
+				errs.Add(fmt.Errorf("failed to remove old xfrm state %s: %w", s.String(), err))
 			} else {
 				scopedLog.Infof("Temporarily removed old XFRM state")
 				defer func(oldXFRMState netlink.XfrmState) {
@@ -238,6 +242,10 @@ func xfrmStateReplace(new *netlink.XfrmState) error {
 				}(s)
 			}
 		}
+	}
+	if err := errs.Error(); err != nil {
+		scopedLog.WithError(err).Error("Failed to clean up old XFRM state")
+		return err
 	}
 
 	// It doesn't exist so let's attempt to add it.
@@ -267,25 +275,27 @@ func xfrmStateReplace(new *netlink.XfrmState) error {
 // the kernel used to reject our check with EEXIST. That logic is upstream in
 // __xfrm_state_lookup.
 func xfrmDeleteConflictingState(states []netlink.XfrmState, new *netlink.XfrmState) (bool, error) {
-	deletedSomething := false
+	var (
+		deletedSomething bool
+		errs             = resiliency.NewErrorSet("failed to delete conflicting XFRM states", len(states))
+	)
 	for _, s := range states {
 		if new.Spi == s.Spi && (new.Mark == nil) == (s.Mark == nil) &&
 			(new.Mark == nil || new.Mark.Value&new.Mark.Mask&s.Mark.Mask == s.Mark.Value) &&
 			xfrmIPEqual(new.Src, s.Src) && xfrmIPEqual(new.Dst, s.Dst) {
-			err := netlink.XfrmStateDel(&s)
-			if err == nil {
-				deletedSomething = true
-				log.WithFields(logrus.Fields{
-					logfields.SPI:           s.Spi,
-					logfields.SourceIP:      s.Src,
-					logfields.DestinationIP: s.Dst,
-				}).Info("Removed a conflicting XFRM state")
-			} else {
-				return deletedSomething, fmt.Errorf("Failed to remove a conflicting XFRM state: %w", err)
+			if err := netlink.XfrmStateDel(&s); err != nil {
+				errs.Add(err)
+				continue
 			}
+			deletedSomething = true
+			log.WithFields(logrus.Fields{
+				logfields.SPI:           s.Spi,
+				logfields.SourceIP:      s.Src,
+				logfields.DestinationIP: s.Dst,
+			}).Info("Removed a conflicting XFRM state")
 		}
 	}
-	return deletedSomething, nil
+	return deletedSomething, errs.Error()
 }
 
 // This function compares two IP addresses and returns true if they are equal.
@@ -538,7 +548,6 @@ func matchesOnNodeID(mark *netlink.XfrmMark) bool {
 }
 
 func ipsecDeleteXfrmState(nodeID uint16) error {
-	var errs error
 	scopedLog := log.WithFields(logrus.Fields{
 		logfields.NodeID: nodeID,
 	})
@@ -548,15 +557,17 @@ func ipsecDeleteXfrmState(nodeID uint16) error {
 		scopedLog.WithError(err).Warning("Failed to list XFRM states for deletion")
 		return err
 	}
+
+	errs := resiliency.NewErrorSet(fmt.Sprintf("failed to delete node (%d) xfrm states", nodeID), len(xfrmStateList))
 	for _, s := range xfrmStateList {
 		if matchesOnNodeID(s.Mark) && getNodeIDFromXfrmMark(s.Mark) == nodeID {
 			if err := netlink.XfrmStateDel(&s); err != nil {
-				scopedLog.WithError(err).Warning("Failed to delete XFRM state")
-				errs = errors.Join(errs, fmt.Errorf("failed to delete xfrm state (%s): %w", s.String(), err))
+				errs.Add(fmt.Errorf("failed to delete xfrm state (%s): %w", s.String(), err))
 			}
 		}
 	}
-	return errs
+
+	return errs.Error()
 }
 
 func ipsecDeleteXfrmPolicy(nodeID uint16) error {
@@ -569,16 +580,20 @@ func ipsecDeleteXfrmPolicy(nodeID uint16) error {
 		scopedLog.WithError(err).Warning("Failed to list XFRM policies for deletion")
 		return fmt.Errorf("failed to list xfrm policies: %w", err)
 	}
-	var errs error
+	errs := resiliency.NewErrorSet("failed to delete xfrm policies", len(xfrmPolicyList))
 	for _, p := range xfrmPolicyList {
 		if matchesOnNodeID(p.Mark) && getNodeIDFromXfrmMark(p.Mark) == nodeID {
 			if err := netlink.XfrmPolicyDel(&p); err != nil {
-				scopedLog.WithError(err).Warning("Failed to delete XFRM policy")
-				errs = errors.Join(errs, fmt.Errorf("failed to delete xfrm policy (%s): %w", p.String(), err))
+				errs.Add(fmt.Errorf("unable to delete xfrm policy %s: %w", p.String(), err))
 			}
 		}
 	}
-	return errs
+	if err := errs.Error(); err != nil {
+		scopedLog.WithError(err).Warning("Failed to delete XFRM policy")
+		return err
+	}
+
+	return nil
 }
 
 /* UpsertIPsecEndpoint updates the IPSec context for a new endpoint inserted in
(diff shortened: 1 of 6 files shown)


pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -281,11 +284,11 @@ func xfrmDeleteConflictingState(states []netlink.XfrmState, new *netlink.XfrmSta
 					logfields.DestinationIP: s.Dst,
 				}).Info("Removed a conflicting XFRM state")
 			} else {
-				return deletedSomething, fmt.Errorf("Failed to remove a conflicting XFRM state: %w", err)
+				errs = errors.Join(errs, fmt.Errorf("Failed to remove a conflicting XFRM state: %w", err))
 			}
 		}
 	}
-	return deletedSomething, nil
+	return deletedSomething, errs
```
christarazi: I think what might be a better alternative here is to aggregate the msg and de-dup them a bit because then the error string might get too long given we are at a "leaf" function in the reconciliation loop. 

This can be done by counting the number of failures out of the number of states to delete, i.e. "5/10 failed", and keeping a set of the unique underlying errors that occurred. Given that we are interacting with netlink here, the error returned from those calls are going to be syscall-style errno error types such as EINVAL, ENOMEM, etc.

I would do something like this pseudocode:

```go
var (
	deletedSomething bool
	errorTypes       set.Set[error]
	failed           int
)
...
if err != nil {
	failed++
	errorTypes.Insert(err)
}
...
if failed > 0 {
	return deletedSomething, fmt.Errorf("failed to remove a conflicting XFRM states (%d/%d): %w", failed, len(states), errorTypes.Get()...)
}     
```

pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -727,12 +730,14 @@ func DeleteXfrm() {
...
-					log.WithError(err).Warning("deleting old xfrm state failed")
-				}
+	if err != nil {
+		log.WithError(err).Warning("unable to fetch xfrm state list")
+		return
+	}
+	for _, s := range xfrmStateList {
+		if isXfrmStateCilium(s) {
+			if err := netlink.XfrmStateDel(&s); err != nil {
+				log.WithError(err).Warning("deleting old xfrm state failed")
```
christarazi: This is a good step, but I'd go further: let's make this function return an error and return an aggregated error of how many resources failed to be deleted and a record of the error types observed similar to my above comment. Then we can move the logging of the error to the caller of this function, rather than on each failure to delete an individual resource. I think this makes sense given the role of this function as a "leaf" in terms of XFRM logic and interaction with netlink.

pkg/datapath/linux/ipsec/probe_linux.go
```diff
@@ -72,5 +74,5 @@ func ProbeXfrmStateOutputMask() error {
 	if probedState.OutputMark.Mask != linux_defaults.RouteMarkMask {
 		return errors.New("incorrect value for probed IPSec output mask attribute")
 	}
-	return nil
+	return
```
christarazi, line 77: I'm confused by this change. What will be the behavior if the code failed on line 75 and returns the error `incorrect value for probed IPSec output mask attribute` if the deferred function will also execute and return `e`?

pkg/datapath/linux/node.go
```diff
@@ -566,15 +566,14 @@ func upsertIPsecLog(err error, spec string, loc, rem *net.IPNet, spi uint8) erro
 
 func (n *linuxNodeHandler) registerIpsecMetricOnce() {
 	n.ipsecMetricOnce.Do(func() {
-		metrics.Register(n.ipsecMetricCollector)
+		if err := metrics.Register(n.ipsecMetricCollector); err != nil {
+			log.WithError(err).Errorf("IPSec metrics registration failed. No metrics will be reported!")
+		}
```
christarazi: This metrics registration code is a bit interesting; I don't think it is possible for it to return errors, **currently**. Looking at the [code](https://github.com/cilium/cilium/blob/d06140fc9e604a39c6cf71ee596cbb0e27691ef3/pkg/metrics/metrics.go#L1432-L1438), the error is never captured. We should probably fix this, cc @dylandreimerink as last author of the code. Once it's captured, then the changes in the PR would actually be utilized. 

I'm also wondering what sort of errors metrics registration could return, but I digress...

For this PR, my only comment is we don't need the dynanmic formatting:

```suggestion
		if err := metrics.Register(n.ipsecMetricCollector); err != nil {
			log.WithError(err).Errorf("IPsec metrics registration failed. No metrics will be reported!")
		}
```

pkg/datapath/linux/node.go
```diff
@@ -1828,7 +1830,9 @@ func (n *linuxNodeHandler) NodeConfigurationChanged(newConfig datapath.LocalNode
 		}
 	}
 
-	n.updateOrRemoveNodeRoutes(prevConfig.AuxiliaryPrefixes, newConfig.AuxiliaryPrefixes, true)
+	if err := n.updateOrRemoveNodeRoutes(prevConfig.AuxiliaryPrefixes, newConfig.AuxiliaryPrefixes, true); err != nil {
+		return err
+	}
```
christarazi: Just to provide a bit more context on the error:

```suggestion
	if err := n.updateOrRemoveNodeRoutes(prevConfig.AuxiliaryPrefixes, newConfig.AuxiliaryPrefixes, true); err != nil {
		return fmt.Errorf("failed to update or remove node routes: %w", err)
	}
```

pkg/datapath/linux/node.go
```diff
@@ -1857,25 +1861,36 @@ func (n *linuxNodeHandler) NodeConfigurationChanged(newConfig datapath.LocalNode
...
+	var errs error
 	if newConfig.UseSingleClusterRoute {
-		n.updateOrRemoveClusterRoute(n.nodeAddressing.IPv4(), newConfig.EnableIPv4)
-		n.updateOrRemoveClusterRoute(n.nodeAddressing.IPv6(), newConfig.EnableIPv6)
+		if err := n.updateOrRemoveClusterRoute(n.nodeAddressing.IPv4(), newConfig.EnableIPv4); err != nil {
+			errs = errors.Join(errs, err)
+		}
+		if err := n.updateOrRemoveClusterRoute(n.nodeAddressing.IPv6(), newConfig.EnableIPv6); err != nil {
+			errs = errors.Join(errs, err)
+		}
```
christarazi: I would also add some context to the errors here as well:

```suggestion
		if err := n.updateOrRemoveClusterRoute(n.nodeAddressing.IPv4(), newConfig.EnableIPv4); err != nil {
			errs = errors.Join(errs, fmt.Errorf("failed to update or remove IPv4 cluster router: %w", err))
		}
		if err := n.updateOrRemoveClusterRoute(n.nodeAddressing.IPv6(), newConfig.EnableIPv6); err != nil {
			errs = errors.Join(errs, fmt.Errorf("failed to update or remove IPv6 cluster router: %w", err))
		}
```

pkg/datapath/linux/node.go
```diff
@@ -1857,25 +1861,36 @@ func (n *linuxNodeHandler) NodeConfigurationChanged(newConfig datapath.LocalNode
...
 	} else if prevConfig.UseSingleClusterRoute {
 		// single cluster route has been disabled, remove route
-		n.deleteNodeRoute(n.nodeAddressing.IPv4().AllocationCIDR(), false)
-		n.deleteNodeRoute(n.nodeAddressing.IPv6().AllocationCIDR(), false)
+		if err := n.deleteNodeRoute(n.nodeAddressing.IPv4().AllocationCIDR(), false); err != nil {
+			errs = errors.Join(errs, err)
+		}
+		if err := n.deleteNodeRoute(n.nodeAddressing.IPv6().AllocationCIDR(), false); err != nil {
+			errs = errors.Join(errs, err)
+		}
```
christarazi, line 1889: I would add more context here as well:

```suggestion
		if err := n.deleteNodeRoute(n.nodeAddressing.IPv4().AllocationCIDR(), false); err != nil {
			errs = errors.Join(errs, fmt.Errorf("failed to remove old IPv4 cluster route: %w", err))
		}
		if err := n.deleteNodeRoute(n.nodeAddressing.IPv6().AllocationCIDR(), false); err != nil {
			errs = errors.Join(errs, fmt.Errorf("failed to remove old IPv6 cluster route: %w", err))
		}
```

Additionally, when I think about this code path a bit more, it's a bit worrying if we couldn't delete a previous node route for the cluster when the node changed configuration. Failing to do this could override the desired routing configuration for the node and potentially cause connectivity outages. At the same time, I'm not sure how this could be possible. `UseSingleClusterRoute` comes from the Agent configuration, so if it changes, then a restart is required for it to take effect.

Let's see if @cilium/sig-datapath could explain a bit more about this path and what sort of error visibility makes sense.

christarazi (changes requested):
Thanks for the PR!

A few comments to improve the error paths.

pkg/datapath/linux/ipsec/ipsec_linux.go
derailed: Thanks Chris! Good point!! I'll lasso in something along those lines...

pkg/datapath/linux/ipsec/ipsec_linux.go
derailed: Hum... I was not sure about this part. The original code just logged and did not propagate. I am guessing because we want to add scope to the logs which will be lost unless we wrap the error further?? But would kill the log indexer if that is the intent here??

pkg/datapath/linux/ipsec/probe_linux.go
derailed: I've tried to keep this code on par with the original intent which would try to delete the state no matter what happens after the defer. You are correct tho the current update would blow the previous err out (if any). I'll fix that.
Not keen on named returned, nor the defer here but thinking there is a combo of hard errors here (ie no state) or `softer` err ie the state is not in the shape we want. I would think the defer needs to move down as there no point in deleting when the state does not exist in the first place?

pkg/datapath/linux/node.go
derailed: Right noticed that too. But figure I'll flag it. Since we swallow the error too in the wrapper??

pkg/datapath/linux/node.go
derailed: Right! This is what tracing is for... where does this end if you are in a deep callstack?? net effect errors are a mile long ;( 
This is where I struggled with this pr. At times we add context by wrapping, others we reach out for scoped loggers and hence don't propagate...

pkg/datapath/linux/node.go
pchaigno: Maybe ping the original author? It's unlikely the whole datapath group (28 people) is going to know about this.

pkg/datapath/linux/node.go
dylandreimerink: Agreed, that just ignoring the errors is not the correct way to deal with that. I don't exactly know why I made that decision. I think I did so because ultimately the goal is to migrate all metrics registration over to using hive instead of these global methods, which would move any error reporting burden to the registry anyway. But until then, we should at least be aware of registration errors.

I think the code is good as is, given the function signature has an error, it should always be checked in case the function starts returning actual errors.

pkg/datapath/linux/node.go
christarazi: The original author is Thomas Graf :wink:

pchaigno: Then nobody knows :sweat_smile:

pkg/datapath/linux/ipsec/ipsec_linux.go
christarazi: Yeah, my suggestion would be to try to prevent as much of this pattern where the leaf functions are logging and instead kick the error back up to the caller. Then the caller can decide to log or not. There can be exceptions to this rule depending on the leaf function, but in this case, it seems like it could do with the same thing we implemented in `xfrmDeleteConflictingState()`

pkg/datapath/linux/ipsec/probe_linux.go
christarazi: If we are relying on the defer to do something specific and we need to know whether it succeeded or not, then it should not be `defer`red. It should just be a normal function call, because the `defer` just complicates the control flow for no reason.

pkg/datapath/linux/ipsec/ipsec_linux.go
derailed: Makes sense. I've changes the call to now return an error.

christarazi (approved):

dylandreimerink (approved):

pkg/datapath/linux/ipsec/errorset.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func newSetError(msg string, c int) *setError {
+	return &setError{
+		msg:   msg,
+		total: c,
+	}
+}
+
+func (e *setError) add(errs ...error) {
```
gentoo-root: This operation is O(n^2). How many items can be in an error set in the worst case?

pkg/datapath/linux/ipsec/errorset.go
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package ipsec
```
gentoo-root: This structure doesn't look IPsec-specific. Maybe there is a generic location more fitting for it?

pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -715,27 +717,39 @@ func isXfrmStateCilium(state netlink.XfrmState) bool {
...
-				}
+	if err != nil {
+		return err
+	}
+	var errs error
+	for _, p := range xfrmPolicyList {
+		if isXfrmPolicyCilium(p) {
+			if err := netlink.XfrmPolicyDel(&p); err != nil {
+				log.WithError(err).Warning("deleting xfrm policy failed")
+				errs = errors.Join(errs, err)
```
gentoo-root: Why is setError not used here, unlike the previous function?

pkg/datapath/linux/ipsec/errorset.go
derailed: Great point. Thank you Maxim! Was not sure where to park it.

pkg/datapath/linux/ipsec/errorset.go
derailed: Indeed. Great catch! 
I am not sure how often nor how many errors could be reported here. 
Changed the implementation to leverage a map instead.

pkg/datapath/linux/ipsec/ipsec_linux.go
derailed: Great observation! Switched the implementation

gentoo-root: Did you? It's still the same.

pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -715,27 +718,42 @@ func isXfrmStateCilium(state netlink.XfrmState) bool {
...
+	for _, p := range xfrmPolicyList {
+		if isXfrmPolicyCilium(p) {
+			if err := netlink.XfrmPolicyDel(&p); err != nil {
+				log.WithError(err).Warning("deleting xfrm policy failed")
+				ee.Add(err)
 			}
 		}
 	}
+	if errs := ee.Errors(); errs != nil {
+		return errors.Join(errs...)
```
gentoo-root: Why not use `ee.Error()`? Same below.

derailed: Indeed! Now the initial impetus was to dedup state errors only. Not sure if policies can be deduped??
So per your comment, I've updated this PR to dedup both policies and states. Also noticed that there are several places where this actually occurs.
Please double check me on this! Thank you.

sayboras (approved):
LGTM :heavy_check_mark:

lmb:
This is still missing approval from cilium/ipsec, sorry.

pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -239,6 +243,11 @@ func xfrmStateReplace(new *netlink.XfrmState) error {
 			}
 		}
 	}
+	if err := errs.Error(); err != nil {
+		// TODO: Should this be a hard error and return instead??
```
gentoo-root: If we return, neither the old state nor the new state will exist.

If we go on, though, they will conflict, and xfrmDeleteConflictingState will attempt to make one more attempt to remove the old state, which is likely to fail with the same error. So we'll end up in the same state, but we'll lose the original error. I guess it makes sense to return right here while we still have the proper error message.

pkg/datapath/linux/ipsec/ipsec_linux.go
gentoo-root: >Not sure if policies can be deduped??

Me neither, but I don't see why not.

pkg/datapath/linux/ipsec/ipsec_linux.go
derailed: Thank you Max! Makes perfect sense

gentoo-root (approved):

pchaigno:
This pull request is creating quite a big mess for IPsec backports :disappointed: 

I would have preferred if it had limited itself to fixing the couple cases where failures where not reported as errors instead of also aggregating tons of error messages.

PR #27724, opened 2023-08-28

jrfastab:
took some longer time. For example I ran `ip x s` in a pod about
received are dropped.

traffic with the new IPs.

XfrmIn*Policy* errors I've seen.

### DIFF
--- a/pkg/datapath/linux/ipsec/ipsec_linux.go
+++ b/pkg/datapath/linux/ipsec/ipsec_linux.go
@@ -1121,3 +1121,32 @@ func StartStaleKeysReclaimer(ctx context.Context) {
 		}
 	}()
 }
+
+// We need to install xfrm state for the local router (cilium_host) early
+// in daemon init path. This is to ensure that we have the xfrm state in
+// place before we advertise the routerIP where other nodes may potentially
+// pick it up and start sending traffic to us. This was previously racing
+// and creating XfrmInNoState errors because other nodes picked up node
+// update before Xfrm config logic was in place. So special case init the
+// rule we need early in init flow.
+func Init() error {
+	outerLocalIP := node.GetInternalIPv4Router()
+	wildcardIP := net.ParseIP("0.0.0.0")
+	localCIDR := node.GetIPv4AllocRange().IPNet
+	localWildcardIP := &net.IPNet{IP: wildcardIP, Mask: net.IPv4Mask(0, 0, 0, 0)}
+
+	if _, err := ipSecReplaceStateIn(outerLocalIP, wildcardIP, false); err != nil {
+		return fmt.Errorf("unable to replace local state: %s", err)
+	}
+	if err := ipSecReplacePolicyIn(localWildcardIP, localCIDR, wildcardIP, outerLocalIP); err != nil {
+		if !os.IsExist(err) {
+			return fmt.Errorf("unable to replace policy in: %s", err)
+		}
+	}
+	if err := IpSecReplacePolicyFwd(localCIDR, outerLocalIP); err != nil {
+		if !os.IsExist(err) {
+			return fmt.Errorf("unable to replace policy fwd: %s", err)
+		}
+	}
+	return nil
+}
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -1085,6 +1085,16 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 		return nil, nil, err
 	}
 
+	// allocateIPs got us the routerIP so now we can create ipsec endpoint
+	// we must do this before publishing the router IP otherwise remote
+	// nodes could pick up the IP and send us outer headers we do not yet
+	// have xfrm rules for.
+	if option.Config.EnableIPSec {
+		if err := ipsec.Init(); err != nil {
+			log.WithError(err).Error("IPSec init failed")
+		}
+	}
+
 	// Must occur after d.allocateIPs(), see GH-14245 and its fix.
 	d.nodeDiscovery.StartDiscovery()
 


pchaigno (changes requested):
I've read the code and commit description multiple times, but I am unable to understand any of this :disappointed: 

What is the impact of this change? I am guessing connectivity drops; how does the code change described here cause those drops?
What is the relation between node annotations and IPsec?

> Interesting side effect is to know if all subsystems can have two configs in place at the same time.

What kind of config is this referring to? XFRM config?

> IPSec is OK to manage the two states which appears to be requirement to not drop traffic on update while multiple IPs are in the network.

Does that mean we already have logic in place to support having the two "configs" in place at the same time?

jrfastab:
@pchaigno I've rewritten this patch after more testing to simplify it I hope. And hopefully improved commit message. I've tested on v1.11 backport heavily as well.

> I've read the code and commit description multiple times, but I am unable to understand any of this 😞
> 
> What is the impact of this change? I am guessing connectivity drops; how does the code change described here cause those drops? What is the relation between node annotations and IPsec?
> 

Correct without this I see dropped packets on upgrade and cilium agent restart in vxlan environment. The K8s node object is relevant because its where we get the IPAddresses[] from which has the CiliumInternalIP address used when configuring IPSec. Its the IP address of the cilium_host and the IP Address used for IPSec outter IP. We need to ensure that the local datapath is configured before these IP addresses are advertised to the cluster otherwise we have a race. Because other remote nodes may learn the new IPAddress and start sending IPSec encrypted traffic with it before we have our datapath setup we get Xfrm InNoState errors.

Making things more prone to racing we learn IPAddresses[] locally the same way a remote does through the k8s watcher. So its actually fairly common to get a case with a busy node (nginx server under stress test) to drop packets with the XfrmInNoState error.

We fix the situation here by initializing the ipsec xfrm in rules before advertising the new IPAddresses[] to other nodes. This removes the race condition. The previous patches had a missing part I only setup the Xfrm state rules at init not the policy as well which resulted in moving the NoState errors to PolicyBlock errors.
 
> > Interesting side effect is to know if all subsystems can have two configs in place at the same time.
> 
> What kind of config is this referring to? XFRM config?
> 

So we care about xfrm config here. But this was mostly a general comment. I realized when doing this that I can fix the xfrm case but really any Cilium feature that wants a no drop on upgrade or restart needs to ensure its config is in place before advertising the IPAddresses[] to the cluster. I did not audit all the code to ensure this is true for other features.

> > IPSec is OK to manage the two states which appears to be requirement to not drop traffic on update while multiple IPs are in the network.
> 
> Does that mean we already have logic in place to support having the two "configs" in place at the same time?

We currently support this yet. If you roll the cilium_agent you can get a new set of IPAddresses[]. IPSec stack will keep two sets of xfrm rules in place. One for the old IPAddresses and one for the new IPAddresses. For vxlan tunnel envs this is easy to reproduce simply restart an agent. Assuming it gets a new cilium_host IP (which is common) there will now be two sets of rules.

A follow up patch would be nice to go through and cleanup the old IPs after a timeout. At the moment they will remain until a key rotation or node is torn down.

pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -1121,3 +1121,32 @@ func StartStaleKeysReclaimer(ctx context.Context) {
...
+	if err := ipSecReplacePolicyIn(localWildcardIP, localCIDR, wildcardIP, outerLocalIP); err != nil {
+		if !os.IsExist(err) {
+			return fmt.Errorf("unable to replace policy in: %s", err)
+		}
+	}
+	if err := IpSecReplacePolicyFwd(localCIDR, outerLocalIP); err != nil {
+		if !os.IsExist(err) {
+			return fmt.Errorf("unable to replace policy fwd: %s", err)
+		}
+	}
```
pchaigno, line 1150: There's a lot of code duplication between this here and `enableIPsecIPv4`/`UpsertIPsecEndpoint`. These XFRM configs also don't exactly match those installed by `enableIPsecIPv4` in that the latter uses `n.nodeConfig.IPv4PodSubnets` as the destination CIDRs for XFRM IN policies. Finally, the IPv6 counterpart is missing.

Maybe to fix all of the above we could move the `newNode.IsLocal()` case of `enableIPsecIPv{4,6}` to this `Init` function instead of duplicating?

pchaigno (changes requested):
Thanks for the explanation! The commit description also looks a lot better :slightly_smiling_face: 

I think we could however avoid the code duplication with the existing XFRM IN config installations. The two code snippets should do the same thing, but they have already started diverging.

pkg/datapath/linux/ipsec/ipsec_linux.go
jrfastab: One way to refactor would be to make it this?
```
func Init() error {
        outerLocalIP := node.GetInternalIPv4Router()                                                                                                                                   
        wildcardIP := net.ParseIP("0.0.0.0")                                                                                                                                                                                           
        localCIDR := node.GetIPv4AllocRange().IPNet                                                                                                                
        localWildcardIP := &net.IPNet{IP: wildcardIP, Mask: net.IPv4Mask(0, 0, 0, 0)}                                                                      

         _, err := UpsertIPsecEndpoint(localCIDR, localWildcardIP, outerLocalIP, wildcardIP, 0, IPDitIn, gsldr)
        return err;
}
```
I believe that is equivalent would probably want to double check. I prefer that to reaching into pkg/datapath/linux/node.go and start refactoring node code. I think it should probably be done but not in the context of a fix.

joestringer:
Superseded by https://github.com/cilium/cilium/pull/28012 . I'll mark draft.

thorn3r:
removing the `needs-backport` labels since this is closed.

PR #27725, opened 2023-08-28
chart: define the envoy image variable in the makefile

weizhoublue:
As @joestringer mentioned in the [comment](https://github.com/cilium/cilium/pull/27621#discussion_r1304767975), usually for these external dependencies, cilium would define them as exports in install/kubernetes/Makefile.values in order to simplify the updates, then in the install/kubernetes/cilium/values.yaml.tmpl file we have variables


youngnick (approved):
LGTM but need a check from @mhofstetter that I haven't missed something due to special handling of this image.

images/scripts/check-cilium-envoy-image.sh
```diff
@@ -12,10 +12,12 @@ root_dir="$(git rev-parse --show-toplevel)"
...
 cd "${root_dir}"
 
-image="quay.io/cilium/cilium-envoy"
-image_tag="$(yq '.envoy.image.tag' ./install/kubernetes/cilium/values.yaml.tmpl)"
-image_sha256="$(yq '.envoy.image.digest' ./install/kubernetes/cilium/values.yaml.tmpl)"
+image="$(yq '.envoy.image.repository' ./install/kubernetes/cilium/values.yaml)"
+image_tag="$(yq '.envoy.image.tag' ./install/kubernetes/cilium/values.yaml)"
+image_sha256="$(yq '.envoy.image.digest' ./install/kubernetes/cilium/values.yaml)"
 
+# in case that the sed fails to replace with the matched regular expression
```
mhofstetter: the comment might be a little bit missleading. isn't is more something like a "pre-check" that the regex at least matches _something_ in `images/cilium/Dockerfile` before executing the `sed` command?

install/kubernetes/Makefile.values
```diff
@@ -42,6 +42,10 @@ export CILIUM_ETCD_OPERATOR_DIGEST:=sha256:04b8327f7f992693c2cb483b999041ed8f92e
 export CILIUM_NODEINIT_REPO:=quay.io/cilium/startup-script
 export CILIUM_NODEINIT_VERSION:=62093c5c233ea914bfa26a10ba41f8780d9b737f
 
+export CILIUM_ENVOY_REPO:=quay.io/cilium/cilium-envoy
```
mhofstetter, line 45: by moving the values from `values.yaml.tmpl` to `Makefile.values` we also have to adapt the [update-cilium-envoy-image.sh](https://github.com/cilium/cilium/blob/main/images/scripts/update-cilium-envoy-image.sh#L36-L44) script.

It should make the script a little bit simpler because it's way easier to match in the `Makefile.values` than it was the `values.yaml.tmpl`

mhofstetter (changes requested):
thanks for the additional follow up contribution @weizhoublue.

We need to adapt another script before we're good to go - see inline comments.

images/scripts/check-cilium-envoy-image.sh
weizhoublue: yes
for example, when someone modifies ./images/cilium/Dockerfile as following 
```
"FROM quay.io/cilium/cilium-envoy**** as cilium-envoy" 
to 
"FROM quay.io/cilium/cilium-envoy**** not as cilium-envoy"
```
that will make sed command here fails to replace the string, as we know, the sed command still return exit code 0, and no modification happen here, which make this scrip fail to detect the change

I would optimize comment words for easily understadning

mhofstetter: :+1: i get the use-case - my "suggestion" is to rephrase the comment that this is a pre-check.

install/kubernetes/Makefile.values
weizhoublue: thanks for reminding, I will adapt that file

weizhoublue: BTW, should update-cilium-envoy-image.sh also update ./install/kubernetes/cilium/values.yaml ?

images/scripts/check-cilium-envoy-image.sh
```diff
@@ -12,10 +12,12 @@ root_dir="$(git rev-parse --show-toplevel)"
...
 
-image="quay.io/cilium/cilium-envoy"
-image_tag="$(yq '.envoy.image.tag' ./install/kubernetes/cilium/values.yaml.tmpl)"
-image_sha256="$(yq '.envoy.image.digest' ./install/kubernetes/cilium/values.yaml.tmpl)"
+image="$(yq '.envoy.image.repository' ./install/kubernetes/cilium/values.yaml)"
+image_tag="$(yq '.envoy.image.tag' ./install/kubernetes/cilium/values.yaml)"
+image_sha256="$(yq '.envoy.image.digest' ./install/kubernetes/cilium/values.yaml)"
 
+# pre-check for sed, in case that this script may fail to detect change when the `sed` command fails to replace the string and return code 0
+grep -E "(FROM ${image}:)(.*)(@sha256:[0-9a-z]*)( as cilium-envoy)" ./images/cilium/Dockerfile &>/dev/null || exit 1
```
nebril: nitpick: would be nice to put the regex in a variable to not have it repeated in these two lines.

images/scripts/update-cilium-envoy-image.sh
```diff
@@ -18,6 +18,10 @@ github_branch="main"
 image="quay.io/cilium/cilium-envoy"
 
 latest_commit_sha="$(curl -s https://api.github.com/repos/${github_repo}/commits/${github_branch} | jq -r '.sha')"
+if [ "${latest_commit_sha}" == "null" ] ; then
```
nebril: We have `errexit` and `pipefail` set at the beginning of the script, so we might also add `--fail` flag to curl in the line above.

nebril (approved):
Thanks for the PR! LGTM overall, two non-blocking nitpicks left inline.

images/scripts/update-cilium-envoy-image.sh
```diff
@@ -30,17 +34,18 @@ fi
...
-# yq would change formatting: https://github.com/mikefarah/yq/issues/465
-# use of envoy.image.override (which would allow match in one line) isn't optimal either
-< ./install/kubernetes/cilium/values.yaml.tmpl tr '\n' '\f' |
-  sed -E "s|(# -- Envoy container image\..*tag: \")(v[0-9a-zA-Z\.-]*)(\")|\1${image_tag}\3|" |
-  sed -E "s|(# -- Envoy container image\..*digest: \")(sha256:[0-9a-z]*)(\")|\1${image_sha256}\3|" |
-  tr '\f' '\n' > ./install/kubernetes/cilium/values.yaml.tmpl_tmp &&
-  mv ./install/kubernetes/cilium/values.yaml.tmpl_tmp ./install/kubernetes/cilium/values.yaml.tmpl
+MAKEFILEPATH="./install/kubernetes/Makefile.values"
+echo "Updating image in ${MAKEFILEPATH}"
+sed -i -E "s|export[[:space:]]+CILIUM_ENVOY_VERSION:=.*|export CILIUM_ENVOY_VERSION:=${image_tag}|"   ${MAKEFILEPATH}
```
mhofstetter: looks like there are some unnecessary whitespaces in front of the `${MAKEFILEPATH}` (on both lines)

images/scripts/update-cilium-envoy-image.sh
weizhoublue: Actually, owing to my network issue, I encountered following issue at the first begging. I could add --exit-status for jq

```
++ curl -s https://api.github.com/repos/cilium/proxy/commits/main
++ jq -r .sha
+ latest_commit_sha=null
++ curl -s https://raw.githubusercontent.com/cilium/proxy/null/ENVOY_VERSION
+ envoy_version='404: Not Found'
+ image_tag='404: Not Found-null'
+ image_full='quay.io/cilium/cilium-envoy:404: Not Found-null'
```

install/kubernetes/Makefile.values
mhofstetter: thanks

mhofstetter (approved):

PR #27729, opened 2023-08-28
.github: use kindest/node instead of quay.io/cilium/kindest-node

aanm:
With the new minor k8s version being released in kindest/node, we don't need to keep using quay.io/cilium/kindest-node as it was primarily used to test RCs.


viktor-kurchenko (approved):
LGTM.

PR #27738, opened 2023-08-28
ci-e2e: Add secondary network NodePort tests

brb:
Successful run of new tests - https://github.com/cilium/cilium/actions/runs/6046909722/job/16409351135


borkmann (approved):

aanm (approved):

nbusseneau (approved):

julianwiedmann:
@brb curious, do you remember what's missing to also enable this test for KPR=false? Doing github.com/cilium/cilium/pull/39718 felt a bit clunky, would be cool if we could just run this by default.

brb:
@julianwiedmann IIRC it should work out of the box. Back then I did not add it, as my focus was on testing multi-homing in nodeport.h.

PR #27740, opened 2023-08-28
Replace metricsmap-bpf-prom-sync with Prometheus Collector pattern

### LINKED ISSUE #27058: Replace Metricsmap-bpf-prom-sync with Prometheus Collector pattern

carnerito:
Previously, the 'metricsmap-bpf-prom-sync' controller was responsible for periodically collecting cilium_datapath drop and forward metrics.
This commit introduces the metricsmapCollector, which implements the Prometheus Collector interface within the metricsmap package. As a result, the aforementioned controller has been removed. The metricsmapCollector is registered within the global Prometheus registry during the initialization of metricsmap. The metrics.Register function has been modified to propagate errors from the registry.Register function instead of simply overriding it with nil.
The metricsmapCollector comprises two metrics maps: forwardedMetricsMap and droppedMetricsMap. These maps are populated within a callback function passed to the IterateWithCallback function. This approach serves two primary purposes:
1. Separation of Map Iteration and Metric Update: By separating the iteration over the BPF map and the updating of Prometheus metrics, the implementation ensures that no partial metrics are exposed in case of map iteration failure.
2. Normalization of exposed Metrics: Unlike the statement in the bpf/lib/metrics.h comments, which suggests exposing only one reason label for forwarded metrics, the eBPF code exposes multiple reasons. Through testing, it was found that reasons 0 and 3 were exposed which triggered this error in prometheus client: https://github.com/prometheus/client_golang/issues/242

Fixes: #27058

### DIFF
--- a/daemon/cmd/datapath.go
+++ b/daemon/cmd/datapath.go
@@ -9,13 +9,11 @@ import (
 	"net/netip"
 	"os"
 	"strings"
-	"time"
 
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
 
 	"github.com/cilium/cilium/pkg/cidr"
-	"github.com/cilium/cilium/pkg/controller"
 	datapathIpcache "github.com/cilium/cilium/pkg/datapath/ipcache"
 	"github.com/cilium/cilium/pkg/datapath/linux/ipsec"
 	"github.com/cilium/cilium/pkg/datapath/linux/linux_defaults"
@@ -48,8 +46,6 @@ import (
 	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/source"
 )
-
-var metricsmapBPFPromSyncControllerGroup = controller.NewGroup("metricsmap-bpf-prom-sync")
 
 // LocalConfig returns the local configuration of the daemon's nodediscovery.
 func (d *Daemon) LocalConfig() *datapath.LocalNodeConfiguration {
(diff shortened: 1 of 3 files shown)


christarazi:
Seems like these round of changes are suffering from the same problems as https://github.com/cilium/cilium/pull/27370.

I'd also try to suggest keeping to one PR, to avoid one PR per iterative change. One way to restart is to reset the original branch make to `origin/main` and then cherry pick the commits on top, and repush back to the original PR.

carnerito:
@christarazi Thanks a lot. Reset & cherry-pick did the trick. I'm closing this PR since [#27370](https://github.com/cilium/cilium/pull/27370) is fixed.

Issue #26669, opened 2023-07-06
CI: Runtime Test (privileged): Test/linuxPrivilegedIPv6OnlyTestSuite/TestArpPingHandlingForMultiDevice


pippolo84:
## CI failure

Seen [here](https://github.com/cilium/cilium/actions/runs/5475179911/attempts/1)

Output:

```
level=info msg="Detected devices" devices="[]" subsys=linux-datapath
level=info msg="Detected devices" devices="[dummy0]" subsys=linux-datapath
level=info msg="Detected devices" devices="[dummy0]" subsys=linux-datapath
level=info msg="Direct routing device detected" direct-routing-device=dummy1 subsys=linux-datapath
level=info msg="Detected devices" devices="[dummy0 dummy1 dummy2]" subsys=linux-datapath
level=info msg="Direct routing device detected" direct-routing-device=dummy1 subsys=linux-datapath
level=info msg="Detected devices" devices="[dummy0 dummy1 dummy2]" subsys=linux-datapath
level=info msg="Direct routing device detected" direct-routing-device=cilium_foo subsys=linux-datapath
level=info msg="IPv6 multicast device detected" ipv6-mcast-device=cilium_foo subsys=linux-datapath
level=info msg="Detected devices" devices="[cilium_foo dummy0 dummy1 dummy2]" subsys=linux-datapath
level=info msg="Direct routing device detected" direct-routing-device=cilium_foo subsys=linux-datapath
level=info msg="Detected devices" devices="[cilium_foo dummy0 dummy1 dummy2]" subsys=linux-datapath
level=info msg="Detected devices" devices="[cilium_foo dummy0 dummy1 dummy2 veth0]" subsys=linux-datapath
level=info msg="Detected devices" devices="[cilium_foo dummy0 dummy1 dummy2 dummy3 veth0]" subsys=linux-datapath
level=info msg="Detected devices" devices="[cilium_foo dummy0 dummy1 dummy2 dummy3 veth0]" subsys=linux-datapath
level=info msg="Detected devices" devices="[cilium_foo dummy0 dummy1 dummy2 veth0]" subsys=linux-datapath
level=info msg="Detected devices" devices="[bond0 cilium_foo dummy0 dummy1 veth0]" subsys=linux-datapath
level=info msg="Detected devices" devices="[dummy0 dummy1]" subsys=linux-datapath
level=info msg="Devices changed" devices="[dummy0]" subsys=linux-datapath
[...]

PR #27747, opened 2023-08-28
daemon: Do not require native routing CIDR if ipmasq-agent is enabled

gandro:
When running with ipmasq-agent-based eBPF masquerading, there should be no need to mandatory require a native routing CIDR, since all CIDRs can be configured via the BPF map instead.

This PR therefore does not set the `IPV4_SNAT_EXCLUSION_DST_CIDR` field if no native routing CIDR is set. Notably, it also does not fall back on the local pod CIDR like we do in regular BPF masquearding, since in many IPAM modes there is no local pod CIDR to begin with.

Note that this change is backwards-compatible, because we have a check that ensures that the native routing CIDR is always set if BPF masquerading is enabled. So all existing users of ip-masq-agent already have a native routing CIDR set. This PR also removes that check, thus allowing future uses to not set a native routing CIDR when running with ipmasq-agent based masquerading.

### DIFF
--- a/pkg/option/config.go
+++ b/pkg/option/config.go
@@ -3772,29 +3772,52 @@ func (c *DaemonConfig) checkMapSizeLimits() error {
 }
 
 func (c *DaemonConfig) checkIPv4NativeRoutingCIDR() error {
-	if c.GetIPv4NativeRoutingCIDR() == nil && c.EnableIPv4Masquerade && !c.TunnelingEnabled() &&
-		c.IPAMMode() != ipamOption.IPAMENI && c.EnableIPv4 && c.IPAMMode() != ipamOption.IPAMAlibabaCloud {
-		return fmt.Errorf(
-			"native routing cidr must be configured with option --%s "+
-				"in combination with --%s --%s=%s --%s=%s --%s=true",
-			IPv4NativeRoutingCIDR, EnableIPv4Masquerade, RoutingMode, RoutingModeNative,
-			IPAM, c.IPAMMode(), EnableIPv4Name)
+	if c.GetIPv4NativeRoutingCIDR() != nil {
+		return nil
+	}
+	if !c.EnableIPv4 || !c.EnableIPv4Masquerade {
+		return nil
+	}
+	if c.EnableIPMasqAgent {
+		return nil
+	}
+	if c.TunnelingEnabled() {
+		return nil
+	}
+	if c.IPAMMode() == ipamOption.IPAMENI || c.IPAMMode() == ipamOption.IPAMAlibabaCloud {
+		return nil
 	}
 
-	return nil
+	return fmt.Errorf(
+		"native routing cidr must be configured with option --%s "+
+			"in combination with --%s=true --%s=true --%s=false --%s=%s --%s=%s",
+		IPv4NativeRoutingCIDR,
+		EnableIPv4Name, EnableIPv4Masquerade,
+		EnableIPMasqAgent,
+		RoutingMode, RoutingModeNative,
+		IPAM, c.IPAMMode())
 }
 
 func (c *DaemonConfig) checkIPv6NativeRoutingCIDR() error {
-	if c.GetIPv6NativeRoutingCIDR() == nil && c.EnableIPv6Masquerade && !c.TunnelingEnabled() &&
-		c.EnableIPv6 {
-		return fmt.Errorf(
-			"native routing cidr must be configured with option --%s "+
-				"in combination with --%s --%s=%s --%s=true",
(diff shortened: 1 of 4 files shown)


christarazi (approved):

aanm (approved):

gandro:
There seems to have been some infra related flakes:

- ci-ipsec-upgrade https://github.com/cilium/cilium/actions/runs/6002359913/job/16278558972 failed downloading the helm charts
- ci-clustermesh https://github.com/cilium/cilium/actions/runs/6002358831/job/16278555174 failed reaching 1.1.1.1,
- ci-e2e https://github.com/cilium/cilium/actions/runs/6002358836/job/16278553484 failed reaching 1.1.1.1
- ci-ipsec-e2e https://github.com/cilium/cilium/actions/runs/6002359028/job/16278554830  failed reaching one.one.one.one
- runtime https://github.com/cilium/cilium/actions/runs/6002359548/job/16278594058 hit https://github.com/cilium/cilium/issues/26669

pkg/option/config.go
```diff
@@ -3772,26 +3772,36 @@ func (c *DaemonConfig) checkMapSizeLimits() error {
...
-			IPAM, c.IPAMMode(), EnableIPv4Name)
+				"in combination with --%s=true --%s=true --%s=false --%s=%s --%s=%s",
+			IPv4NativeRoutingCIDR,
+			EnableIPv4Name, EnableIPv4Masquerade,
+			EnableIPMasqAgent,
+			RoutingMode, RoutingModeNative,
+			IPAM, c.IPAMMode())
 	}
 
 	return nil
```
pippolo84: Given that this helper is now pretty complicated, I'm wondering if reversing the conditions and using early return might improve readability:

```
	if c.GetIPv4NativeRoutingCIDR() != nil {
		return nil
	}
	if !c.EnableIPv4 || !c.EnableIPv4Masquerade {
		return nil
	}
	if c.EnableIPMasqAgent {
		return nil
	}
	if c.TunnelingEnabled() {
		return nil
	}
	if c.IPAMMode() == ipamOption.IPAMENI || c.IPAMMode() == ipamOption.IPAMAlibabaCloud {
		return nil
	}

	return fmt.Errorf(
		"native routing cidr must be configured with option --%s "+
			"in combination with --%s=true --%s=true --%s=false --%s=%s --%s=%s",
		IPv4NativeRoutingCIDR,
		EnableIPv4Name, EnableIPv4Masquerade,
		EnableIPMasqAgent,
		RoutingMode, RoutingModeNative,
		IPAM, c.IPAMMode())
```

pkg/option/config.go
```diff
@@ -3772,26 +3772,36 @@ func (c *DaemonConfig) checkMapSizeLimits() error {
...
-			IPv6NativeRoutingCIDR, EnableIPv6Masquerade, RoutingMode, RoutingModeNative,
-			EnableIPv6Name)
+				"in combination with --%s=true --%s=true --%s=false --%s=%s",
+			IPv6NativeRoutingCIDR,
+			EnableIPv6Name, EnableIPv6Masquerade,
+			EnableIPMasqAgent,
+			RoutingMode, RoutingModeNative)
 	}
 
 	return nil
```
pippolo84: Same here with:
```
	if c.GetIPv6NativeRoutingCIDR() != nil {
		return nil
	}
	if !c.EnableIPv6 || !c.EnableIPv6Masquerade {
		return nil
	}
	if c.EnableIPMasqAgent {
		return nil
	}
	if c.TunnelingEnabled() {
		return nil
	}
	return fmt.Errorf(
		"native routing cidr must be configured with option --%s "+
			"in combination with --%s=true --%s=true --%s=false --%s=%s",
		IPv6NativeRoutingCIDR,
		EnableIPv6Name, EnableIPv6Masquerade,
		EnableIPMasqAgent,
		RoutingMode, RoutingModeNative)
```

pippolo84 (approved):
LGTM, thanks!
I've left a non-blocking suggestion inline, feel free to ignore that if you think the current version is more readable.

pkg/option/config.go
gandro: Yeah good point! I'll change it

qmonnet (approved):

ti-mo (approved):

derailed (approved):
@gandro LGTM Nice work Sebastian! 
Note: seeing a lot of parallels and similar logic in our codebase related to ipv4/6 handling. Wondering if we could apply generics for these use cases to dry up our code??

gandro:
> @gandro LGTM Nice work Sebastian! Note: seeing a lot of parallels and similar logic in our codebase related to ipv4/6 handling. Wondering if we could apply generics for these use cases to dry up our code??

Good point. I think there is some experimentation needed first though

Issue #26621, opened 2023-07-04
Failure using non-default ip pool when combining multi-pool with etcd


MaienM:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

I've been playing around with multi-pool mode. I was using the following options + test resources, and it was all working as expected until I tried to use etcd.

Install:

```
helm install cilium cilium/cilium \
	--version=1.14.0-snapshot.4 \
	--namespace=cilium \
	--create-namespace \
	--set k8sServiceHost=192.168.50.3 \
	--set k8sServicePort=6443 \
	--set cni.install=false \
	--set-string extraConfig.write-cni-conf-when-ready= \
	--set ipam.mode=multi-pool \
	--set tunnel=disabled \
	--set autoDirectNodeRoutes=true \
	--set ipv4NativeRoutingCIDR=10.200.0.0/8 \
	--set endpointRoutes.enabled=true \
	--set-string extraConfig.enable-local-node-route=false \
	--set kubeProxyReplacement=strict \
	--set bpf.masquerade=true \
	--set ipam.operator.autoCreateCiliumPodIPPools.default.ipv4.cidrs='{10.200.0.0/16}' \
	--set ipam.operator.autoCreateCiliumPodIPPools.default.ipv4.maskSize=24 \
	--set ipam.operator.autoCreateCiliumPodIPPools.iot.ipv4.cidrs='{192.168.3.16/28}' \
	--set ipam.operator.autoCreateCiliumPodIPPools.iot.ipv4.maskSize=30
```

Resources:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: default-test
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      name: default-test
  template:
    metadata:
      labels:
        name: default-test
      annotations:
        ipam.cilium.io/ip-pool: default
    spec:
      containers:
        - name: toolbox
          args: ["tail", "-F", "/dev/null"]
          image: docker.io/wbitt/network-multitool
          imagePullPolicy: IfNotPresent
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: iot-test
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      name: iot-test
  template:
    metadata:
      labels:
        name: iot-test
      annotations:
        ipam.cilium.io/ip-pool: iot
    spec:
      containers:
        - name: toolbox
          args: ["tail", "-F", "/dev/null"]
[...]

Issue #24588, opened 2023-03-27
Cilium in `cluster-pool-v2beta` IPAM mode creates duplicate "ghost" node on every agent


eyanulis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

When running in `cluster-pool-v2beta` IPAM mode, each Cilium agent shows a duplicate "ghost" node in `cilium node list` which has no name and which has the same endpoint CIDR as the node where that agent is running. **This behavior is NOT observed in v1 cluster-pool mode.**

I have reproduced this in my own home cluster, and have minimally reproduced it in the development Vagrant environment as well.

Home cluster:

```
# node 1
root@kube-storage-01-lng1:/home/cilium# cilium node list
Name                   IPv4 Address   Endpoint CIDR      IPv6 Address   Endpoint CIDR
                                      198.18.16.0/26 <---------- ghost node                    
kube-compute-01-lng1   198.18.3.2     198.18.16.128/26                  
kube-compute-02-lng1   198.18.3.3     198.18.16.64/26                   
kube-storage-01-lng1   198.18.3.4     198.18.16.0/26                    

# node 2
root@kube-compute-01-lng1:/home/cilium# cilium node list
Name                   IPv4 Address   Endpoint CIDR      IPv6 Address   Endpoint CIDR
                                      198.18.16.128/26 <---------- ghost node                  
kube-compute-01-lng1   198.18.3.2     198.18.16.128/26                  
kube-compute-02-lng1   198.18.3.3     198.18.16.64/26                   
kube-storage-01-lng1   198.18.3.4     198.18.16.0/26                    

# node 3
root@kube-compute-02-lng1:/home/cilium# cilium node list
Name                   IPv4 Address   Endpoint CIDR      IPv6 Address   Endpoint CIDR
                                      198.18.16.64/26 <---------- ghost node                   
kube-compute-01-lng1   198.18.3.2     198.18.16.128/26                  
kube-compute-02-lng1   198.18.3.3     198.18.16.64/26                   
kube-storage-01-lng1   198.18.3.4     198.18.16.0/26                    
```

Vagrant development VM:
```
vagrant@k8s1:~/go/src/github.com/cilium/cilium$ cilium node list
Name   IPv4 Address    Endpoint CIDR   IPv6 Address   Endpoint CIDR
                       10.11.0.0/24 <---------- ghost node                   
k8s1   192.168.60.11   10.11.0.0/24                   
```

[...]

PR #27753, opened 2023-08-28
ipam: Remove cluster-pool-v2beta code

gandro:
This IPAM mode has been deprecated in Cilium v1.14 and has been superseeded by multi-pool IPAM instead. This commit removes all code related to cluster-pool-v2beta from Cilium.


pkg/ipam/pool.go
```diff
@@ -222,89 +193,6 @@ func (p *podCIDRPool) releaseExcessCIDRsMultiPool(neededIPs int) {
...
-func (p *podCIDRPool) clusterPoolV2Beta1Status(allocationThreshold, releaseThreshold int) types.PodCIDRMap {
-	p.mutex.Lock()
-	defer p.mutex.Unlock()
-
-	if allocationThreshold <= 0 {
-		allocationThreshold = defaults.IPAMPodCIDRAllocationThreshold
-	}
-
-	if releaseThreshold <= 0 {
-		releaseThreshold = defaults.IPAMPodCIDRReleaseThreshold
```
tklauser, line 276: `defaults.IPAMPodCIDRAllocationThreshold` and `defaults.IPAMPodCIDRReleaseThreshold` now seem unused, probably they can be deleted as well?

tklauser (approved):
Thanks for nicely splitting these up into commits which made it a lot easier to review and verify!

One non-blocking minor nit inline.

pkg/ipam/pool.go
gandro: Oh yeah, good catch! Those need to be removed as well!

gandro:
Rebased on main to resolve conflicts

kaworu (approved):
Helm changes LGTM

gandro:
Had to re-trigger some tests manually due to changed workflow names.

zacharysarah (approved):
LGTM for docs

gandro:
@tommyp1ckles Ping for review

tommyp1ckles (approved):
k8s changes look good

gandro:
CI is stuck because the PR has been open for a while and new tests have been introduce. I'll rebase to be able to re-run CI

log1cb0mb:
Can we expect a proper upgrade/migration path from `cluster-pool-v2beta` to `multi-pool` with v1.15? 
It would have been lot better if this: https://github.com/cilium/cilium/pull/27100 was included in v1.15 first so that easier to do migration before removing the whole thing.
OR BGP CP multi-pool support should be backported to v1.14?

gandro:
Unfortunately, [BGP for multi-pool](https://github.com/cilium/cilium/pull/27100) cannot be backported to v1.14, due to some other larger refactoring PRs that depend on it. `clusterpool-v2beta` will only removed in v1.15 -  In v1.15 `multi-pool` does have feature parity the removed clusterpool-v2beta, including BGP support. 

Generally speaking, there is no migration guide planned, since clusterpool-v2beta was never recommended for production. However, having said that, I think we would accept contribution which adds some migration documentation and potentially also some helper logic in the agent for example which reads the old `clusterpool-v2beta` podCIDRs and writes them into the `multi-pool` spec.

log1cb0mb:
In that case, can we consider not dropping `clusterpool-v2beta` from v1.15 so that users have just one thing to deal with. Assuming there is a way to co-exist both IPAM modes?

gandro:
> In that case, can we consider not dropping `clusterpool-v2beta` from v1.15 so that users have just one thing to deal with. Assuming there is a way to co-exist both IPAM modes?

I don't think simply reverting this PR is a good way forward. The `clusterpool-v2beta` code had a few of servere bugs (e.g. https://github.com/cilium/cilium/issues/24588 or https://github.com/cilium/cilium/issues/26621 and has been blocking some other tech debt removal work. So even reverting this PR will not be an easy task, since we already started follow-up cleanups.

I also don't think it's possible to run in two IPAM modes: There is only one operator, and it only runs one IPAM allocator at a time. It would 

To find the best solution, I think it would be useful to get a bit more context on what deployments you have and what problems you want to solve. I'm not opposed to adding code or logic that helps the migration both in v1.14 or v1.15. I can totally envision a migration mode where we make the multi-pool logic backwards-compatible with the clusterpool-v2beta fields, and I think such a solution would be less work and less overall risk than backporting BGP to v1.14 or reverting this PR.

log1cb0mb:
That makes sense and originally I was looking for "best way" to migrate from `v2beta` to `multi-pool`. I understand former was in beta state and now its being deprecated but our use case required its usage.

We are heavily relying on BGP CP so now the problem is in order to migrate to `multi-pool` which every user must, we need to also consider migration to `multi-pool` support in BGP CP. I was hoping that we could first take care of migrating to `multi-pool` within v1.14 provided that BGP CP supports it before we move to v1.15 where `v2-beta` is removed completely.

Now thinking about it, BGP CP support for `multi-pool` in v.14 wouldnt have really helped us much as we would have to naturally migrate to using both at the same time regardless.

So I suppose what I am looking for is how do we perform this migration in best way possible without effecting any workload. I am thinking we are going to have to run different versions at the same time where each node (running say v1.15) is migrated to new pool one at a time. Let's say we manage to rollout in this fashion so my concern is if cilium global config is changed based on v1.15 will it effect existing v1.14 agents and not just IPAM functionality but also BGP CP?

gandro:
Thanks a lot for the clarification!

> Let's say we manage to rollout in this fashion so my concern is if cilium global config is changed based on v1.15 will it effect existing v1.14 agents and not just IPAM functionality but also BGP CP?

So yeah, unless Cilium agents are restarted, they will not start to use the new configuration. Only once they are restarted, will they read the newer `ConfigMap`. If you want to keep certain nodes on an different configuration regardless of restarts, you can use the [CiliumNodeConfig](https://docs.cilium.io/en/stable/configuration/per-node-config/#ciliumnodeconfig-objects), which might be of use here.

I think the bigger challenge however is the fact that `clusterpool-v2beta` and `multi-pool` are using different fields for storing the per-node podCIDRs. Basically, as is at the moment, once a node switches over to the new IPAM mode, it will no longer observe the pod CIDRs that it previously had and instead will attempt allocate fresh ones. This will cause disruption in the workload, because existing workloads have not released they IP yet. 

The good news however is that I think this should be fixable with some migration logic from v1.14 with `clusterpool-v2beta` to v1.15 with `multi-pool`: 

Step 1: Prepare a v1.15 configuration, that uses `multi-pool` IPAM with a `default` IPAM pool to use the same CIDRs as the `--cluster-pool-ipv4/6-{cidr,mask}` configuration (Helm: `cilium.ipam.operator.clusterPoolIPv4/6{PodCIDRList/MaskSize}`) used by `clusterpool-v2beta`.

Step 2: Mirror all per-node pod CIDRs from all CiliumNode `spec.ipam.podCIDRs` into the `spec.ipam.pools["default"].allocated` field. This will make sure that the `multi-pool` IPAM operator code is aware of the pod CIDRs previously allocated to a node.

Step 3: Update the Operator to Cilium v1.15 with multi-pool IPAM mode. Thanks to the mirrored pod CIDRs, it will make sure that any already allocated pod CIDRs are retained and not re-allocated to any new nodes. At this point, any v1.14 agent running in `clusterpool-v2beta` mode will continue running, but it will not be able to allocate any new CIDRs or release old ones. While this might hinder new pods from being scheduled if the node is running low on IPs, this will not interrupt any running pods.

Step 4: Update Cilium agents one by one from v1.14 with `clusterpool-v2beta` to v1.15 with `multi-pool`. Thanks to the mirrored pod CIDRs, they will remember their old pod CIDRs and restore all endpoints with the old pod CIDR. All this logic is already present in the code.

(Optional) Step 5: If migration is successful and no downgrade is needed, go ahead and clear any `spec.ipam.podCIDRs` from the `CiliumNode`. This step is optional, as the field will simply be ignored by all agents and the operator once running in `multi-pool` IPAM mode.


The only code that needs to be created for this migration plan to work is the code in step 2. This could either be done by a custom shell or python script, or we could add this logic to the Cilium operator. I would accept such a PR.

log1cb0mb:
Good stuff! Thank you for all the details. 

Isnt there a shortcut to all of this, you drain/delete the node and rejoin or bring the node up with no previous CNI config and with v1.15?

This obviously assumes that somehow we can control v1.15 is deployed onto cluster however agent is not rolled out on every node except new joining ones. Use this approach along with `CiliumNodeConfig` CRD?

log1cb0mb:
Yes, restarting workload on one node at a time is not big of an issue as there is redundancy.

Overlapping/Copying per-node CIDR over is a good point, I had not considered that. iirc, cilium operator does that where it allocates ipam block per node from the start of podCIDR subnet.

Let me give this a try, will test this out in dev environment and see how it goes.

PR #27755, opened 2023-08-28
add Twilio to Users list

michaelsaah:


xmulligan (approved):

xmulligan:
Thanks for the addition @michaelsaah ! Would you be interested in doing a CNCF case study about your use of Cilium? You can see other examples here https://www.cncf.io/case-studies/?_sft_lf-project=cilium. We would just need 30 minutes of your time and then write it up for you

michaelsaah:
hi @xmulligan, you can put us down for one in the future. at the moment we're still fairly early in our adoption so we wouldn't have all that much to talk about yet.

xmulligan:
Sounds good, I'll circle back on this!

PR #27757, opened 2023-08-28
docs-builder image: Only build from cilium/cilium

michi-covalent:
Most forks don't need to run this workflow. Check the repository owner and only build from cilium/cilium.

### DIFF
--- a/.github/workflows/build-images-docs-builder.yaml
+++ b/.github/workflows/build-images-docs-builder.yaml
@@ -22,6 +22,7 @@ concurrency:
 jobs:
   build-and-push:
     name: Build and Push Image
+    if: github.repository_owner == 'cilium'
     runs-on: ubuntu-22.04
     timeout-minutes: 30
     environment: docs-builder


michi-covalent:
@learnitall could you approve https://github.com/cilium/cilium/actions/runs/6005524833/job/16288411316?pr=27757 🙏

.github/workflows/build-images-docs-builder.yaml
```diff
@@ -90,7 +91,7 @@ jobs:
   update-pr:
     name: Update Pull Request with new image reference
     needs: build-and-push
-    if: needs.build-and-push.outputs.digest
+    if: github.repository_owner == 'cilium' && needs.build-and-push.outputs.digest
```
qmonnet: Are this one, and the one below, necessary? If the owner is not `cilium`, then `build-and-push` will not run and `needs.build-and-push.outputs.digest` shouldn't resolve as `true`?

Looks good otherwise for adding the restriction. The workflow won't work on forks anyway, unless the owner has bothered setting up the relevant environment and secrets.

michi-covalent: oh! that's a good point. skipping `build-and-push` should be enough. updating the PR now...

michi-covalent: could you approve https://github.com/cilium/cilium/actions/runs/6007268461/job/16293139576?pr=27757 just to make sure it still works 🙏

qmonnet (approved):
Ack for the first commit, thanks!

joestringer (approved):

learnitall (approved):
Great!

nbusseneau (changes requested):
We had guards in the past and since moved on to not guarding workflows and instead recommending in the documentation that forks disable Actions. Unless we have a strong rationale for adding guards again, I would advise against doing so, because it's explicitly useful for forks to be able to run Actions when they need to (e.g. a developer working on a specific workflow from their fork instead of in upstream).

michi-covalent:
yeah that makes sense, somehow it didn't occur to me that downstreams can just disable this action. closing 💥

PR #27362, opened 2023-08-08
Read FQDNRejectResponseCode from config

ayuspin:
Currently dnsProxy.dnsRejectResponseCode helm value is ignored because FQDNRejectResponseCode is not populated from viper

PR #27794, opened 2023-08-29
config: Use String instead of StringVar method

pippolo84:
Commit 87518a39a9 ("Read FQDNRejectResponseCode from config") fixed the FQDNRejectResponseCode option handling, adding the missing code to read the value from the ConfigMap.  Since the usage of StringVar does not correctly handle that case by itself, using it is pointless.

This commit changes it to use String as we already do for every other string agent option.

Relates: #27362


gandro (approved):
Thanks!

pippolo84:
Conformance E2E hit https://github.com/cilium/cilium/issues/27762

PR #27796, opened 2023-08-29
identity/cache: only call SortedList for release

bimmlerd:
First commit introduces a specific benchmark, the next two commits are optimizations. Commit msgs should be descriptive. The first optimization commit does not improve benchmark time as the benchmark doesn't exercise that path, but the second optimization clearly improves benchmark time.

Below the metrics of a kind cluster which runs 10 pods which send a single UDP packet to `host1.test.org.`, while having a custom DNS server which responds with 5 random IPs for `host1.test.org.` with a TTL of 5 seconds. This mimicks S3 behaviour.

In the graph, you can see that there's a doubling of identities, that's the identity restoration grace period of 10 minutes; it shows when the pods were restarted with the two commits of this PR applied. They show an approximate reduction of 5MiB/s and 50KHz allocations.
   
image

### DIFF
--- a/pkg/identity/cache/local.go
+++ b/pkg/identity/cache/local.go
@@ -81,11 +81,15 @@ func (l *localIdentityCache) getNextFreeNumericIdentity(idCandidate identity.Num
 // in as the 'oldNID' parameter; identity.InvalidIdentity must be passed if no
 // previous numeric identity exists. 'oldNID' will be reallocated if available.
 func (l *localIdentityCache) lookupOrCreate(lbls labels.Labels, oldNID identity.NumericIdentity) (*identity.Identity, bool, error) {
+	// Not converting to string saves an allocation, as byte key lookups into
+	// string maps are optimized by the compiler, see
+	// https://github.com/golang/go/issues/3512.
+	repr := lbls.SortedList()
+
 	l.mutex.Lock()
 	defer l.mutex.Unlock()
 
-	stringRepresentation := string(lbls.SortedList())
-	if id, ok := l.identitiesByLabels[stringRepresentation]; ok {
+	if id, ok := l.identitiesByLabels[string(repr)]; ok {
 		id.ReferenceCount++
 		return id, false, nil
 	}
@@ -102,7 +106,7 @@ func (l *localIdentityCache) lookupOrCreate(lbls labels.Labels, oldNID identity.
 		ReferenceCount: 1,
 	}
 
-	l.identitiesByLabels[stringRepresentation] = id
+	l.identitiesByLabels[string(repr)] = id
 	l.identitiesByID[numericIdentity] = id
 
 	if l.events != nil {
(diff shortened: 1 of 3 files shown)


aanm (approved):
LGTM but would be nice to have some bechmarks as well to compare before/after.

bimmlerd:
```
goos: linux
goarch: arm64
pkg: github.com/cilium/cilium/pkg/labels
                     │     old     │                 opt                 │
                     │   sec/op    │   sec/op     vs base                │
Labels_SortedList-10   3.279µ ± 6%   2.209µ ± 6%  -32.65% (p=0.000 n=10)

                     │    old     │                opt                 │
                     │    B/op    │    B/op     vs base                │
Labels_SortedList-10   504.0 ± 0%   360.0 ± 0%  -28.57% (p=0.000 n=10)

                     │     old     │                opt                 │
                     │  allocs/op  │ allocs/op   vs base                │
Labels_SortedList-10   13.000 ± 0%   3.000 ± 0%  -76.92% (p=0.000 n=10)
```
for the existing benchmark - I'll write another one which focusses on a the relevant CIDR identity case.

aditighag (approved):

pkg/identity/cache/local.go
```diff
@@ -84,8 +84,11 @@ func (l *localIdentityCache) lookupOrCreate(lbls labels.Labels, oldNID identity.
 	l.mutex.Lock()
 	defer l.mutex.Unlock()
 
-	stringRepresentation := string(lbls.SortedList())
-	if id, ok := l.identitiesByLabels[stringRepresentation]; ok {
+	repr := lbls.SortedList()
+	// Not converting to string saves an allocation, as byte key lookups into
+	// string maps are optimized by the compiler, see
+	// https://github.com/golang/go/issues/3512.
+	if id, ok := l.identitiesByLabels[string(repr)]; ok {
```
schlosna: Not directly related to the byte/string conversion, but I'm curious if the locking should be deferred until after converting labels to bytes to reduce possible contention by shrinking the critical section.

```suggestion
	repr := lbls.SortedList()
	// Not converting to string saves an allocation, as byte key lookups into
	// string maps are optimized by the compiler, see
	// https://github.com/golang/go/issues/3512.
	l.mutex.Lock()
	defer l.mutex.Unlock()
	if id, ok := l.identitiesByLabels[string(repr)]; ok {
```

bimmlerd: Yep, that's a good idea, thanks.

bimmlerd:
I've pushed the above-suggested change to pull the call to `SortedList` out of the critical section of `lookupOrCreate`. Please take another look.

bimmlerd:
CI triage:
- Conformance Runtime (ci-runtime) seems to have hit #26479, rerunning.

aanm (approved):

joamaki (approved):

aditighag (approved):

pkg/identity/cache/local.go
```diff
@@ -81,11 +81,15 @@ func (l *localIdentityCache) getNextFreeNumericIdentity(idCandidate identity.Num
 // in as the 'oldNID' parameter; identity.InvalidIdentity must be passed if no
 // previous numeric identity exists. 'oldNID' will be reallocated if available.
 func (l *localIdentityCache) lookupOrCreate(lbls labels.Labels, oldNID identity.NumericIdentity) (*identity.Identity, bool, error) {
+	// Not converting to string saves an allocation, as byte key lookups into
+	// string maps are optimized by the compiler, see
+	// https://github.com/golang/go/issues/3512.
```
aditighag, line 86: TIL! 🚀

PR #27797, opened 2023-08-29
ci: add conformance-ginkgo-race-detection

### LINKED ISSUE #26170: Add race-detector tests back to the CI pipeline
The race-detection CI jenkins jobs have been broken since it requires a new ubuntu version built into the vagrant VM images.

With the introduction of ginkgo-based GH jobs we should be able to re-enable the race-detector pipelines back into our CI.

markpash:
Duplicate the ginkgo workflow but with the race detection ci images.

Fixes: #26170

### DIFF
--- a/.github/workflows/conformance-ginkgo-race-detection.yaml
+++ b/.github/workflows/conformance-ginkgo-race-detection.yaml
@@ -1,0 +1,485 @@
+name: Conformance Ginkgo Race Detection (ci-ginkgo-race-detection)
+
+# Any change in triggers needs to be reflected in the concurrency group.
+on:
+  workflow_dispatch:
+    inputs:
+      PR-number:
+        description: "Pull request number."
+        required: true
+      context-ref:
+        description: "Context in which the workflow runs. If PR is from a fork, will be the PR target branch (general case). If PR is NOT from a fork, will be the PR branch itself (this allows committers to test changes to workflows directly from PRs)."
+        required: true
+      SHA:
+        description: "SHA under test (head of the PR branch)."
+        required: true
+      extra-args:
+        description: "[JSON object] Arbitrary arguments passed from the trigger comment via regex capture group. Parse with 'fromJson(inputs.extra-args).argName' in workflow."
+        required: false
+        default: '{}'
+  # Run every 6 hours
+  schedule:
+    - cron:  '0 1/6 * * *'
+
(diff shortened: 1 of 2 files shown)


markpash:
I've duplicated the workflow, but not sure exactly how to test that this works as it should. I'd love some advice on how I can test this.

.github/workflows/conformance-ginkgo-race-detection.yaml
```diff
@@ -0,0 +1,485 @@
...
+    name: Build Ginkgo E2E
+    timeout-minutes: 30
+    steps:
+      # If any of these steps are modified, please update the copy of these
+      # steps further down under the 'setup-and-test' jobs.
+
+      # Warning: since this is a privileged workflow, subsequent workflow job
+      # steps must take care not to execute untrusted code.
+      - name: Checkout pull request branch (NOT TRUSTED)
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
```
qmonnet, line 105: This version is outdated w.r.t. the current ginkgo workflow in the `main` branch.

Same for all other uses of `actions/checkout`, and for `setup-go`.

qmonnet (commented):
How will people remember to update the two workflows at the same time?

Given that the triggers are the same as the Ginkgo workflow, would it make sense instead to extend the Matrix to run with the race images as well, rather than duplicating the workflow entirely?

> I'd love some advice on how I can test this.

Not sure if it will work, but from your changes in the `ariane-config.yaml`, let's try `/ci-gingko`?

nbusseneau:
It will indeed not run from a fork, you will need to re-open your PR with a branch located on upstream :) I would suggest using a separate trigger in Ariane config to be able to trigger only that workflow, and there is a special trick you need to do to get your new workflow to register when you open the PR: add a temporary `pull_request: {}` trigger to the workflow, just to register the workflow on the repository when the PR opens. After it has been registered, you will be able to run the workflow with the Ariane trigger no problem, and can remove the `pull_request: {}` trigger.

As for structure: for testing let's just do it copy/pasted for now, when we are happy with how it works I would suggest we look into merging the changes in the main Ginkgo workflow file, so that only one file needs to be maintained. One idea I have is to leverage the `extra-args` parameters that Ariane allows, so we could trigger race pipelines with a trigger like this: `/ci-ginkgo race`

PR #27798, opened 2023-08-29
bpf: lxc: support Pod->Service->Pod hairpinning with endpoint routes

### LINKED ISSUE #27709: IPv4 Service hairpin return-traffic denied by ingress policy

### What happened?

When traffic to a service hairpins to the same pod (pod is both the source and destination of the request) -- and egress packet is SNATed to LBLoopback IP -- the return traffic is not matched for ingress network policy and the packet is denied.

To use the LBLoopback IP, set ` --set bpf.hostLegacyRouting=true  --set endpointRoutes.enabled=true` (see full reproduction below).

Egress traffic is not affected by egress policy; this is addressed by https://github.com/cilium/cilium/commit/52cd6da139c1ac5d67de65a821f953c936034f2e.

Note: This may be related to #23911 

---

Two things are happening:
1. There is no conntrack entry for the return-traffic (i.e. `ct_status=CT_NEW`) which bypasses the first `skip_policy_enforcement` jump 
   > `CPU 15: MARK 0xcf5e125b FROM 939 DEBUG: CT verdict: New, revnat=0`

   https://github.com/cilium/cilium/blob/c191ef6f9b5b1dd87bb4f513bfc67641b421dad0/bpf/bpf_lxc.c#L1776-L1807

2. the created (post-revnat) `TCP_OUT` CT entry is not marked as `LBLoopback`, which bypasses the second `skip_policy_enforcement` jump 
   > `TCP OUT 10.96.184.106:80 -> 10.244.1.183:33333 service expires=858932 (remaining: 37 sec(s)) RxPackets=0 RxBytes=8 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=0 TxBytes=0 TxFlagsSeen=0x02 LastTxReport=858869 Flags=0x0000 [ ] RevNAT=5 SourceSecurityID=0 IfIndex=0`

   https://github.com/cilium/cilium/blob/c191ef6f9b5b1dd87bb4f513bfc67641b421dad0/bpf/bpf_lxc.c#L1812-L1820

---

It seems that a CT entry should be created so that the return-traffic is correctly identified as `CT_REPLY`, thereby allowing the return traffic to skip enforcement.

---

context on hairpin scenario:
- https://github.com/cilium/cilium/pull/27602
- https://github.com/cilium/cilium/issues/23913

### Cilium Version

```
$ cilium version
cilium-cli: v0.13.0 compiled with go1.20 on linux/amd64
cilium image (default): v1.13.0
cilium image (stable): v1.14.1
cilium image (running): -dev:local

$ git branch --show-current
v1.14
```

### Kernel Version

`Linux kind-worker 6.3.11-1rodete1-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.3.11-1rodete1 (2023-07-18) x86_64 x86_64 x86_64 GNU/Linux`

### Kubernetes Version

```
$ kubectl version
Client Version: v1.28.0
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.27.1
```

### Sysdump

See reproduction below.

### Relevant log output

```shell
# from reproduction below
# comments are (obviously) my own
$ kk exec server -- curl --no-keepalive --connect-timeout 5 --local-port 33333 -v ${clusterip}:80/hostname

CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Conntrack lookup 1/2: src=10.244.1.183:33333 dst=10.96.184.106:80
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Conntrack lookup 2/2: nexthdr=6 flags=4 dir=2 scope=1
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: CT verdict: New, revnat=0
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Service backend slot lookup: slot=1, dport=80
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Conntrack create: proxy-port=0 revnat=5 src-identity=0 lb=0.0.0.0
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Loopback SNAT from=10.244.1.183 to=169.254.42.1
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Conntrack lookup 1/2: src=169.254.42.1:33333 dst=10.244.1.183:80
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Conntrack lookup 2/2: nexthdr=6 flags=1 dir=0 scope=2
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: CT verdict: New, revnat=0
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Successfully mapped addr=10.244.1.183 to identity=25928
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Conntrack create: proxy-port=0 revnat=5 src-identity=25928 lb=0.0.0.0
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Attempting local delivery for container id 939 from seclabel 25928
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Inheriting identity=2 from stack
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Conntrack lookup 1/2: src=169.254.42.1:33333 dst=10.244.1.183:80
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Conntrack lookup 2/2: nexthdr=6 flags=0 dir=1 scope=2
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: CT entry found lifetime=857692, revnat=5
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: CT verdict: Established, revnat=5
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Successfully mapped addr=169.254.42.1 to identity=2
CPU 15: MARK 0x30a936bd FROM 939 DEBUG: Attempting local delivery for container id 939 from seclabel 25928
-> endpoint 939 flow 0x30a936bd , identity world->25928 state established ifindex 0 orig-ip 169.254.42.1: 169.254.42.1:33333 -> 10.244.1.183:80 tcp SYN
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: Conntrack lookup 1/2: src=10.244.1.183:80 dst=169.254.42.1:33333
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: Conntrack lookup 2/2: nexthdr=6 flags=1 dir=0 scope=2
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: CT entry found lifetime=857692, revnat=5
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: CT verdict: Reply, revnat=5
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: Successfully mapped addr=169.254.42.1 to identity=2
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: Reverse NAT lookup, index=5
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: Performing reverse NAT, address=10.96.184.106 port=80
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: Loopback reverse SNAT from=169.254.42.1 to=10.244.1.183
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: Attempting local delivery for container id 939 from seclabel 25928
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: Inheriting identity=2 from stack
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: Conntrack lookup 1/2: src=10.96.184.106:80 dst=10.244.1.183:33333
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: Conntrack lookup 2/2: nexthdr=6 flags=0 dir=1 scope=2
# CT entry is new, so does not skip enforcement; see conntrack entries below monitor output
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: CT verdict: New, revnat=0
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: Successfully mapped addr=10.96.184.106 to identity=2
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: Attempting local delivery for container id 939 from seclabel 25928
CPU 15: MARK 0xec59bf30 FROM 939 DEBUG: Policy evaluation would deny packet from 2 to 25928
Policy verdict log: flow 0xec59bf30 local EP ID 939, remote ID world, proto 6, ingress, action deny, auth: disabled, match none, 10.96.184.106:80 -> 10.244.1.183:33333 tcp SYN, ACK
xx drop (Policy denied) flow 0xec59bf30 to endpoint 939, ifindex 10, file bpf_lxc.c:2021, , identity world->25928: 10.96.184.106:80 -> 10.244.1.183:33333 tcp SYN, ACK


TCP IN 169.254.42.1:33333 -> 10.244.1.183:80 expires=857844 (remaining: 60 sec(s)) RxPackets=6 RxBytes=444 RxFlagsSeen=0x02 LastRxReport=857776 TxPackets=15 TxBytes=1110 TxFlagsSeen=0x12 LastTxReport=857784 Flags=0x0008 [ LBLoopback ] RevN
AT=5 SourceSecurityID=25928 IfIndex=0
TCP OUT 169.254.42.1:33333 -> 10.244.1.183:80 expires=857839 (remaining: 55 sec(s)) RxPackets=0 RxBytes=0 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=6 TxBytes=444 TxFlagsSeen=0x02 LastTxReport=857776 Flags=0x0008 [ LBLoopback ] RevNAT=5 Sou
rceSecurityID=25928 IfIndex=0
# CT entry is not marked as LBLoopback, so does not skip enforcement
TCP OUT 10.96.184.106:80 -> 10.244.1.183:33333 service expires=857839 (remaining: 55 sec(s)) RxPackets=0 RxBytes=8 RxFlagsSeen=0x00 LastRxReport=0 TxPackets=0 TxBytes=0 TxFlagsSeen=0x02 LastTxReport=857776 Flags=0x0000 [ ] RevNAT=5 SourceS
ecurityID=0 IfIndex=0
```

### Anything else?

Reproduction on `main`:

```
gh repo clone cilium/ciium cilium-oss && cd !$
KUBEPROXY_MODE="none" make kind && \
make kind-image && \
helm upgrade -i cilium ./install/kubernetes/cilium \
  --wait \
  --namespace kube-system \
  --set k8sServiceHost="kind-control-plane" \
  --set k8sServicePort="6443" \
  --set debug.enabled=true \
  --set pprof.enabled=true \
  --set enableIPv4Masquerade=false \
  --set enableIPv6Masquerade=false \
  --set enableIPv4Masquerade=false \
  --set hostFirewall.enabled=false \
  --set socketLB.hostNamespaceOnly=true \
  --set nodeinit.enabled=true \
  --set ipam.mode=kubernetes \
  --set ipv4.enabled=true \
  --set ipv4NativeRoutingCIDR=10.244.0.0/16 \
  --set ipv6.enabled=false \
  --set image.override="localhost:5000/cilium/cilium-dev:local" \
  --set image.pullPolicy=Never \
  --set operator.image.override="localhost:5000/cilium/operator-generic:local" \
  --set operator.image.pullPolicy=Never \
  --set operator.image.suffix="" \
  --set securityContext.privileged=true \
  --set gatewayAPI.enabled=false \
  --set socketLB.enabled=false \
  --set bpf.hostLegacyRouting=true \
  --set endpointRoutes.enabled=true

alias kk="kubectl --context kind-kind"
kk patch cm -n kube-system cilium-config --patch '
data:
  debug: "true"
  debug-verbose: "datapath"
' && \
kk rollout restart ds -n kube-system cilium && kk rollout status ds -n kube-system cilium

kk --context kind-kind apply -f - <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: server
  namespace: default
  labels:
    app: server
spec:
  containers:
  - name: agnhost
    image: k8s.gcr.io/e2e-test-images/agnhost:2.39
    args:
    - netexec
    - --http-port=80
    ports:
    - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: server-svc
spec:
  type: ClusterIP
  selector:
    app: server
  ports:
    - protocol: TCP
      port: 80
      targetPort: 80
EOF

time while ! [[ "$( kk get svc server-svc -ojsonpath='{.spec.clusterIP}' | xargs )" != "" ]]; do echo -n "."; sleep 1; done
clusterip=$( kk get svc server-svc -ojsonpath='{.spec.clusterIP}' )

time while ! [[ "$( kk get pod server -ojsonpath='{.status.conditions[?(@.type=="Ready")].status}' )" =~ "True" ]]; do echo -n "."; sleep 1; done

host=$( kk get pod server \
    -ojsonpath='{ .spec.nodeName }' )
agent=$( kk get pod -n kube-system \
    -l k8s-app=cilium \
    --field-selector=spec.nodeName=${host} \
    -ojsonpath='{.items[].metadata.name}' )

# success
kk exec server -- curl --no-keepalive --connect-timeout 5 -v ${clusterip}:80/hostname

kk apply -f - <<EOF
kind: NetworkPolicy
metadata:
  name: allow-egress-to-self
  namespace: default
spec:
  podSelector:
    matchLabels:
      app: server
  egress:
  - to:
    - podSelector:
        matchLabels:
          app: server
  policyTypes:
  - Egress
EOF

# success
kk exec server -- curl --no-keepalive --connect-timeout 5 -v ${clusterip}:80/hostname

kk apply -f - <<EOF
kind: NetworkPolicy
metadata:
  name: allow-ingress-to-self
  namespace: default
spec:
  podSelector:
    matchLabels:
      app: server
  ingress:
  - from:
    - podSelector:
        matchLabels:
          app: server
  policyTypes:
  - Ingress
EOF

# fail
kk exec server -- curl --no-keepalive --connect-timeout 5 --local-port 33333 -v ${clusterip}:80/hostname

# use to observe datapath debug logging
kk exec -it -n kube-system ${agent} -- cilium monitor -D -v --related-to $( kk get cep server -ojsonpath='{ .status.id }' )
```

---

Workaround

Adding the cluster IP to the ingress policy resolves the issue:

```
kk apply -f - <<EOF
kind: NetworkPolicy
metadata:
  name: allow-to-self
  namespace: default
spec:
  podSelector:
    matchLabels:
      app: server
  ingress:
  - from:
    - podSelector:
        matchLabels:
          app: server
    - ipBlock:
        cidr: "${clusterip}/32"
  policyTypes:
  - Ingress
EOF

# success
kk exec server -- curl --no-keepalive --connect-timeout 5 -v ${clusterip}:80/hostname
```

### Code of Conduct

ti-mo:
This patch adds a check to determine whether a Pod is sending traffic to itself through a Service. With endpoint routes enabled, this traffic is subject to policy, and we want to avoid the user to have to explicitly allow traffic to+from itself.

Fixes: https://github.com/cilium/cilium/issues/27709
Strictly depends on https://github.com/cilium/cilium/pull/27602 (for backports)

```release-note
bpf: lxc: support Pod->Service->Pod hairpinning with endpoint routes
```

### DIFF
--- a/bpf/bpf_lxc.c
+++ b/bpf/bpf_lxc.c
@@ -1969,6 +1969,76 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
 	return ret;
 }
 
+static __always_inline bool
+ipv4_to_endpoint_is_hairpin_flow(struct __ctx_buff *ctx, struct iphdr *ip4)
+{
+	__be16 client_port, backend_port, service_port;
+	struct ipv4_ct_tuple tuple = {};
+	struct lb4_backend *backend;
+	__be32 pod_ip, service_ip;
+	struct ct_entry *entry;
+	struct ct_map *map;
+	int err, l4_off;
+
+	/* Extract the tuple from the packet so we can freely access addrs and ports.
+	 * All values are in network byte order.
+	 */
+	err = lb4_extract_tuple(ctx, ip4, ETH_HLEN, &l4_off, &tuple);
+	if (IS_ERR(err))
+		return false;
+
+	/* If the packet originates from a regular, non-loopback address, it will look
+	 * like service_ip:client_port -> pod_ip:service_port.
+	 *
+	 * In order to determine whether the packet has been hairpinned, we need to
+	 * obtain the backend (listen) port first, requiring a CT lookup with the
+	 * TUPLE_F_SERVICE flag, followed by a backend lookup. After this, the regular
+	 * CT TUPLE_F_OUT lookup can proceed.
+	 */
+	service_ip = tuple.saddr;
+	pod_ip = tuple.daddr;
+	client_port = tuple.sport;
+	service_port = tuple.dport;
+
+	tuple.daddr = service_ip;
+	tuple.saddr = pod_ip;
+	tuple.dport = client_port;
+	tuple.sport = service_port;
+
+	tuple.flags = TUPLE_F_SERVICE;
+
+	map = get_ct_map4(&tuple);
+	entry = map_lookup_elem(map, &tuple);
+	if (!entry)
+		return false;
+
+	backend = lb4_lookup_backend(ctx, entry->backend_id);
+	if (!backend)
+		return false;
+
+	backend_port = backend->port;
+
+	/* Now the backend (listen) port inside the container is known, an egress CT
+	 * lookup can be performed.
+	 */
+	tuple.daddr = IPV4_LOOPBACK;
+	tuple.saddr = pod_ip;
+	tuple.dport = backend_port;
+	tuple.sport = client_port;
+
+	tuple.flags = TUPLE_F_OUT;
+
+	map = get_ct_map4(&tuple);
+	entry = map_lookup_elem(map, &tuple);
+	if (entry)
+		/* The packet is considered hairpinned if its egress CT entry has the
+		 * loopback flag set.
+		 */
+		return entry->lb_loopback == 1;
+
+	return false;
+}
+
 __section_tail(CILIUM_MAP_CALLS, CILIUM_CALL_IPV4_TO_ENDPOINT)
 int tail_ipv4_to_endpoint(struct __ctx_buff *ctx)
 {
@@ -2016,6 +2086,23 @@ int tail_ipv4_to_endpoint(struct __ctx_buff *ctx)
 	update_metrics(ctx_full_len(ctx), METRIC_INGRESS, REASON_FORWARDED);
 #endif
 	ctx_store_meta(ctx, CB_SRC_LABEL, 0);
+
+	/* Check if packet is locally hairpinned (pod reaching itself through a
+	 * service) and skip the policy check if that is the case. Otherwise, pods may
+	 * need to explicitly allow traffic to themselves in some network
+	 * configurations.
+	 */
+	if (ipv4_to_endpoint_is_hairpin_flow(ctx, ip4)) {
+		send_trace_notify4(ctx, TRACE_TO_LXC,
+				   ctx_load_meta(ctx, CB_SRC_LABEL),
+				   SECLABEL, ip4->saddr, LXC_ID,
+				   ctx->ingress_ifindex,
+				   TRACE_REASON_UNKNOWN, 0);
+
+		/* Skip policy check for hairpinned flow */
+		ret = CTX_ACT_OK;
+		goto out;
+	}
 
 	ret = ipv4_policy(ctx, 0, src_sec_identity, NULL, &ext_err, &proxy_port,
 			  false);


bpf/bpf_lxc.c
```diff
@@ -1941,6 +1941,119 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
 	return ret;
 }
 
+static __always_inline bool
+ipv4_to_endpoint_is_hairpin_flow(struct __ctx_buff *ctx, struct iphdr *ip4, __u32 __maybe_unused src_sec_identity)
+{
+	struct ipv4_ct_tuple tuple = {};
+	struct lb4_backend *backend __maybe_unused;
```
qmonnet:
```suggestion
	struct lb4_backend *backend;
```

I think I added the attribute in my version because I commented out code in the middle to focus on the second lookup. But it's not unused

bpf/bpf_lxc.c
```diff
@@ -1941,6 +1941,119 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
 	return ret;
 }
 
+static __always_inline bool
+ipv4_to_endpoint_is_hairpin_flow(struct __ctx_buff *ctx, struct iphdr *ip4, __u32 __maybe_unused src_sec_identity)
```
qmonnet:
```suggestion
ipv4_to_endpoint_is_hairpin_flow(struct __ctx_buff *ctx, struct iphdr *ip4, __u32 src_sec_identity)
```

Doesn't seem unused

bpf/bpf_lxc.c
```diff
@@ -1941,6 +1941,119 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
...
+		tuple.dport = client_port;
+		tuple.sport = service_port;
+
+		tuple.flags = TUPLE_F_SERVICE;
+
+		map = select_ct_map4(ctx, CT_SERVICE, &tuple);
+		entry = map_lookup_elem(map, &tuple);
+		if (!entry) {
+			return false;
+		}
```
qmonnet:
```suggestion
		if (!entry)
			return false;
```

bpf/bpf_lxc.c
```diff
@@ -1941,6 +1941,119 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
...
+
+		backend = lb4_lookup_backend(ctx, entry->backend_id);
+		if (!backend) {
+			return false;
+		}
```
qmonnet:
```suggestion
		if (!backend)
			return false;
```

bpf/bpf_lxc.c
```diff
@@ -1941,6 +1941,119 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
...
+	tuple.dport = backend_port;
+	tuple.sport = client_port;
+
+	tuple.flags = TUPLE_F_OUT;
+
+	map = select_ct_map4(ctx, CT_EGRESS, &tuple);
+	entry = map_lookup_elem(map, &tuple);
+	if (entry) {
+		return entry->lb_loopback == 1;
+	}
```
qmonnet:
```suggestion
	if (entry)
		return entry->lb_loopback == 1;
```

... Or should we insert some `cilium_dbg()` here?

bpf/bpf_lxc.c
```diff
@@ -1941,6 +1941,119 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
...
+
+	/* Two different types of packets need to be handled here. client->server and
+	 * server->client packets both pass through the to-endpoint path; presumably
+	 * they are NATed differently depending on the direction. We can observe both
+	 * service_ip:client_port -> pod_ip:service_port and
+	 * IPV4_LOOPBACK:backend_port -> pod_ip:client_port.
+	 */
+	if (tuple.saddr != IPV4_LOOPBACK) {
+		/* 
+		 * If the packet originates from a non-loopback address, it will look like:
```
qmonnet: Nit: Comment style, here and elsewhere

```suggestion
		/* If the packet originates from a non-loopback address, it will look like:
```

bpf/bpf_lxc.c
```diff
@@ -1941,6 +1941,119 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
...
+		 * TUPLE_F_SERVICE flag, followed by a backend lookup. After this, the regular
+		 * CT TUPLE_F_OUT lookup can proceed.
+		 *
+		 * We're looking for the following CT entry:
+		 * daddr 10.96.211.200 (service ip)
+		 * saddr 10.244.1.101 (pod ip)
+		 * dport 12345 (client port)
+		 * sport 8080 (service port)
+		 * nexthdr 06 (TCP)
+		 * flags 04 (TUPLE_F_SERVICE)
```
qmonnet: We can maybe drop this part from the comments. The numeric values don't tell much, given that we haven't fully introduced the setup and explained which IP goes to which entity; and your assignments below with the dedicated variables make it very explicit how we build the tuple for the lookup.

bpf/bpf_lxc.c
```diff
@@ -1941,6 +1941,119 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
...
+	 * server->client packets both pass through the to-endpoint path; presumably
+	 * they are NATed differently depending on the direction. We can observe both
+	 * service_ip:client_port -> pod_ip:service_port and
+	 * IPV4_LOOPBACK:backend_port -> pod_ip:client_port.
+	 */
+	if (tuple.saddr != IPV4_LOOPBACK) {
+		/* 
+		 * If the packet originates from a non-loopback address, it will look like:
+		 * 10.96.211.200:12345 -> 10.244.1.101:8080 or service_ip:client_port ->
+		 * pod_ip:service_port.
```
qmonnet:
```suggestion
		 * service_ip:client_port -> pod_ip:service_port.
```

bpf/bpf_lxc.c
```diff
@@ -1941,6 +1941,119 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
...
+		/* 
+		* If the packet originates from the loopback address, it will look like:
+		* 169.254.42.1:80 -> 10.244.1.101:12345 or IPV4_LOOPBACK:backend_port ->
+		* pod_ip:client_port.
+		*
+		* This gives us enough information to directly look up the egress CT entry
+		* (TUPLE_F_OUT) with the loopback flag, meaning the packet was locally
+		* hairpinned.
+		*/
+
```
qmonnet:
```suggestion
```

bpf/bpf_lxc.c
qmonnet: Note: I haven't given much thought into the naming and location (in the source) for `ipv4_to_endpoint_is_hairpin_flow()`. I guess it's good enough that we can keep it like this, but if you have a better idea don't hesitate to rename it.

qmonnet (changes requested):
Please find some minor comments below (even though I understand this is still a draft)

bpf/bpf_lxc.c
```diff
@@ -1941,6 +1941,119 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
...
+	struct ct_map *map;
+	int err, l4_off;
+	__be32 pod_ip, service_ip;
+	__be16 client_port, backend_port, service_port;
+
+	/* For reasons unknown, hairpinned packets don't have their IDs resolved and
+	 * are always seen as originating from world.
+	 */
+	if (src_sec_identity != WORLD_ID)
+		return false;
```
sypakine: This reads as if the result is unintended; an integration test will be necessary to prevent a (possible) fix for ID resolution from rendering this function moot.

bpf/bpf_lxc.c
```diff
@@ -2011,6 +2081,24 @@ int tail_ipv4_to_endpoint(struct __ctx_buff *ctx)
 #endif
 	ctx_store_meta(ctx, CB_SRC_LABEL, 0);
 
+	/* Check if packet is locally hairpinned (pod reaching itself through a
+	 * service) and skip the policy check if that is the case. Otherwise, pods may
+	 * need to explicitly allow traffic to themselves in some network
+	 * configurations.
+	 */
+	if (ip4->protocol == IPPROTO_TCP &&
```
julianwiedmann: Note that UDP and SCTP are valid choices here too I believe. So you could just trust `lb4_extract_tuple()` to filter that for you.

bpf/bpf_lxc.c
ti-mo: I think it's fine!

bpf/bpf_lxc.c
ti-mo: Agreed, in fact, after merging https://github.com/cilium/cilium/pull/27602 and rebasing, we discovered that packets in the reply direction suddenly _did_ receive proper secID and succeeded the policy check. This means a skip was no longer necessary, so no point in the hairpin check to begin with, but you're right in that it's not the most robust solution. I've removed the check to let the CT lookups do the work.

bpf/bpf_lxc.c
ti-mo: Good suggestion, thanks!

bpf/bpf_lxc.c
ti-mo: Yes, was considering that too, will have to look into how I can introduce one.

ti-mo: Since the goal is to make this easy to backport, I'll hold off on this. I'd have to add a new entry to the DBG_ enum, which means protobuf needs to be regenerated, which already complicates backporting a bit. Also, it looks like protobuf wasn't regenerated after adding `DBG_L7_LB` about 2 years ago, so this would need to be picked up in this PR. Looks like we haven't added any DBG_ entries in a while, and this one doesn't seem to fit any existing ones at first glance.

qmonnet: Sounds good to me, we can always add something to `main` as a follow-up if we want to.

bpf/bpf_lxc.c
```diff
@@ -1963,6 +1963,79 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
...
+	tuple.daddr = IPV4_LOOPBACK;
+	tuple.saddr = pod_ip;
+	tuple.dport = backend_port;
+	tuple.sport = client_port;
+
+	tuple.flags = TUPLE_F_OUT;
+
+	map = select_ct_map4(ctx, CT_EGRESS, &tuple);
+	if (!map)
+		return false;
```
qmonnet: I missed this check, good catch. We should also add it to the first call to `select_ct_map4()`, above in the function.

ti-mo: Right, of course. It's interesting that the verifier only complains about the second `select_ct_map4()` missing a null check and not the first one.

bpf/bpf_lxc.c
ti-mo: Created https://github.com/cilium/cilium/issues/27826 for following up on this.

qmonnet (approved):
Looks good, thanks for making this work!

bpf/bpf_lxc.c
```diff
@@ -1963,6 +1963,82 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
...
+	service_port = tuple.dport;
+
+	tuple.daddr = service_ip;
+	tuple.saddr = pod_ip;
+	tuple.dport = client_port;
+	tuple.sport = service_port;
+
+	tuple.flags = TUPLE_F_SERVICE;
+
+	map = select_ct_map4(ctx, CT_SERVICE, &tuple);
```
julianwiedmann: Pretty sure this should be `get_ct_map4()`. There's no chance that Cluster-IDs could play into the mix here - not in theory, and not in practice (as `cil_to_container()` clears the `skb->cb` and we thus know it doesn't contain a `cluster_id`).

(same for the second user below)

bpf/bpf_lxc.c
```diff
@@ -2011,6 +2087,23 @@ int tail_ipv4_to_endpoint(struct __ctx_buff *ctx)
...
+		send_trace_notify4(ctx, TRACE_TO_LXC,
+				   ctx_load_meta(ctx, CB_SRC_LABEL),
+				   SECLABEL, ip4->saddr, LXC_ID,
+				   ctx->ingress_ifindex,
+				   TRACE_REASON_UNKNOWN, 0);
+
+		/* Skip policy check for hairpinned flow */
+		ret = CTX_ACT_OK;
+		goto out;
+	}
```
julianwiedmann, line 2105: Just to note, as I was stuck here for a moment.

Ideally we would traverse `ipv4_policy()`, do CT updates etc and only skip the network policy lookup.
But the packet won't find any CT entry (that's the whole problem). so that would be pointless, and completely skipping here should be fine :+1:.

julianwiedmann (changes requested):
One small change in the CT logic, otherwise looks good! :pray:

bpf/bpf_lxc.c
qmonnet: Good catch, thank you! Now addressed.

julianwiedmann (approved):

bpf/bpf_lxc.c
```diff
@@ -1963,6 +1963,76 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
...
+
+	tuple.daddr = service_ip;
+	tuple.saddr = pod_ip;
+	tuple.dport = client_port;
+	tuple.sport = service_port;
+
+	tuple.flags = TUPLE_F_SERVICE;
+
+	map = get_ct_map4(&tuple);
+	entry = map_lookup_elem(map, &tuple);
```
aspsk, line 2011: Is verifier ok with `map` maybe being NULL?

bpf/bpf_lxc.c
```diff
@@ -1963,6 +1963,76 @@ int tail_ipv4_policy(struct __ctx_buff *ctx)
...
+static __always_inline bool
+ipv4_to_endpoint_is_hairpin_flow(struct __ctx_buff *ctx, struct iphdr *ip4)
+{
+	struct ipv4_ct_tuple tuple = {};
+	struct lb4_backend *backend;
+	struct ct_entry *entry;
+	struct ct_map *map;
+	int err, l4_off;
+	__be32 pod_ip, service_ip;
+	__be16 client_port, backend_port, service_port;
```
aspsk: Reverse x-mass tree?

aspsk (commented):
LGTM, only one comment and one style comment

aditighag:
@qmonnet @ti-mo We have an [existing test](https://github.com/cilium/cilium/blob/main/test/k8s/services.go#L137-L137) to validate the hairpin connectivity. Please extend the configuration for per-ep routes.
If this PR is close to getting merged, you can do it as a follow-up.

brb:
> We have an [existing test](https://github.com/cilium/cilium/blob/main/test/k8s/services.go#L137-L137) to validate the hairpin connectivity. Please extend the configuration for per-ep routes.

Let's better add it as the CLI connectivity test, as the ginkgo is deprecated.

bpf/bpf_lxc.c
qmonnet: We had a check on the `map` being NULL before using `get_ct_map4()`. Now that we use this function instead, we can't get a NULL value, see how it's implemented:

```c
static __always_inline void *
get_ct_map4(const struct ipv4_ct_tuple *tuple)
{
	if (tuple->nexthdr == IPPROTO_TCP)
		return &CT_MAP_TCP4;

	return &CT_MAP_ANY4;
}
```

bpf/bpf_lxc.c
qmonnet: Sure I can change that.

qmonnet: Now done, thanks for the review!

qmonnet:
Yes I'll look into the related CI test, preferably as a follow-up given that this PR is in a good shape already

aspsk (approved):

gandro:
@qmo This also needs a custom backport to v1.14 due to conflicts unfortunately

qmonnet:
OK thank you I will look into it

PR #27802, opened 2023-08-29
egressgw: policy: ensure egressGateway field is not nil

jibi:


pkg/egressgateway/policy.go
```diff
@@ -261,6 +261,10 @@ func ParseCEGP(cegp *v2.CiliumEgressGatewayPolicy) (*PolicyConfig, error) {
 	}
 
 	egressGateway := cegp.Spec.EgressGateway
+	if egressGateway == nil {
+		return nil, fmt.Errorf("CiliumEgressGatewayPolicy's egressGateway can't be empty")
+	}
+
```
julianwiedmann: drive-by - in some of the error paths further below, we're already producing a formatted error with the resource name. Let's clean that up at some point, it would help users to understand which CEGP is the culprit.

julianwiedmann (approved):
thanks!

pkg/egressgateway/policy.go
jibi: good shout, after the switch to resource[T] we are not actually logging any of these errors :grimacing:  so I'll refactor this function a bit and make the caller log the error together with the CEGP name

PR #27803, opened 2023-08-29
cilium/cmd: make output of 'cilium policy selectors' sorted.

tommyp1ckles:
Previously, the output would be sent in no particular order, thus when running the 'policy selectors' command output would often be printed in different order. This makes it hard to compare subsequent outputs.

This PR simply sorts output by selector string before it is printed to make the ordering stable.

### DIFF
--- a/cilium/cmd/policy_selectors.go
+++ b/cilium/cmd/policy_selectors.go
@@ -6,6 +6,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"sort"
 	"text/tabwriter"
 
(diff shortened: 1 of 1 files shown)


joestringer:
Minor nit: I'd consider this API unstable, but I think you're communicating that the output should be sorted. I agree with that :-)

tommyp1ckles:
@joestringer updated commit so people don't get the wrong idea

derailed (approved):
@tommyp1ckles Nice!

Issue #14157, opened 2020-11-24
Consider semver v2 for the master branch development version


joestringer:
We had a discussion on Slack regarding shifting the `VERSION` file for the master branch from the `1.X-1.90` style for development versions towards `1.X.0-dev`. For example, rather than using `1.9.90` version to denote a pre-v1.10 development version, instead use `1.10.0-dev`. This would be the version number tagged on "nightly" type images, such as `latest` docker images and pre-release helm charts.

## Positives

* Mitigate confusion for projects that vendor cilium (eg hubble-cli)
  * Eg. right now hubble-cli has this in the go.mod:
    `github.com/cilium/cilium v1.9.0-rc1.0.20201120102040-c0b68419e6d8`
    This looks like a pre-1.9 version, but it's actually just the latest tag in the master branch and the actual code is a 1.10 development version.
* Reduce likelihood that someone could deploy a latest version and confuse "1.9.9" vs. 1.9.90"
  * Not so likely since the user would have to install `cilium/cilium:latest` which users are usually steered away from in the docs.
* Easier to generate helm charts for the master branch
  * Rather than modifying the `VERSION` file in the repo, it would already have the correct version.
  * Note, if we want to update older branch pre-release helm charts, we still need to manually modify the version in the tree.

## Concerns

* Possibility to break something unintended in the tree
  * Related:
  ```
  $ git grep "\.90"
  Documentation/concepts/kubernetes/troubleshooting.rst:        INFO Cilium 0.8.90 f022e2f Thu, 27 Apr 2017 23:17:56 -0700 go version go1.7.5 linux/amd64
  Documentation/contributing/release/feature.rst:#. Update the ``VERSION`` file to contain ``v1.2.90``
  Documentation/gettingstarted/aws.rst:          lastUpdated: 2018-11-30T11:28:03.907678888Z
  Binary file Documentation/gettingstarted/images/cilium_k8s_demo_l7-policy-230817.png matches
  Binary file Documentation/images/identity_store.png matches
  VERSION:1.9.90
  clustermesh-apiserver/test/mock.json:{"kind": "Endpoints", "metadata": {"name": "foo", "namespace": "default"}, "subsets": [{"addresses": [{"ip": "34.90.138.138"}], "ports": [{"name": "https", "port": 443, "protocol": "TCP"}]}]}
  contrib/packaging/deb/debian/changelog:cilium (0.9.90) UNRELEASED; urgency=medium
[...]

PR #27807, opened 2023-08-29
contrib: Add support for X.Y.Z-pre.N releases

joestringer:
Previously with "vX.Y.Z-snapshot.N" release tagging for monthly
snapshots, we found that this caused an ordering problem once we started
releasing "vX.Y.Z-rc.N" versions. This is because -snapshot is a higher
version than -rc. When other projects attempted to update their Go mod
dependencies for Cilium, it would select the previous month's snapshot
release instead of the new RC.

This commit therefore removes support for -snapshot and adds -pre
instead for periodic (monthly) snapshots of Cilium, starting during the
v1.15.0-dev development cycle. Commands that list previous tags need to
retain the "-snapshot" matches, but the input validation will no longer
allow this version format.

Related: https://github.com/cilium/cilium/issues/14157

### DIFF
--- a/contrib/backporting/common.sh
+++ b/contrib/backporting/common.sh
@@ -4,7 +4,8 @@
 
 set -e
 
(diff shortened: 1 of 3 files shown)


aanm (approved):
Thank you!

PR #27812, opened 2023-08-30
bpf: lxc: transfer sec identity for per-EP loopback in reply direction

julianwiedmann:
l3_local_delivery() contains a special case to bypass ingress policy for loopback traffic. But when redirecting loopback requests / replies in a per-EP config, they don't actually bypass the policy program. The redirect_ep() delivers them to the host-side veth interface, and ingress policy is applied as the packet passes to the pod-side veth peer.

https://github.com/cilium/cilium/pull/27602 recently changed the logic so that `hairpin_flow` is only set for replies. A subtle side-effect of that change is that in a per-EP config, l3_local_delivery() will now copy the the src sec identity into the skb->mark. cil_to_container() then extracts it, and tail_ipv4_to_endpoint() no longer performs a ipcache lookup for ip4->saddr (which is IPV4_LOOPBACK, and would result in WORLD_ID).

Right now this is mostly a cosmetic improvement, as the identity is not used for policy enforcement of loopback traffic. But there are still users (eg. drop notifications) in tail_ipv4_to_endpoint() that benefit from having the correct security identity in place.

Therefore make the same change in reply direction, so that we also get the correct security identity there (instead of an ipcache lookup for the Service IP, which would return WORLD_ID). For this push the loopback check into the tail-call specific path of l3_local_delivery(), while the per-EP path inserts the identity into skb->mark and then just calls its "normal" redirect_ep().

Reported-by: Timo Beckers <timo@isovalent.com>

### DIFF
--- a/bpf/lib/l3.h
+++ b/bpf/lib/l3.h
@@ -71,7 +71,8 @@ static __always_inline int ipv4_l3(struct __ctx_buff *ctx, int l3_off,
 #ifndef SKIP_POLICY_MAP
 static __always_inline int
 l3_local_delivery(struct __ctx_buff *ctx, __u32 seclabel,
-		  const struct endpoint_info *ep, __u8 direction __maybe_unused,
+		  const struct endpoint_info *ep __maybe_unused,
+		  __u8 direction __maybe_unused,
 		  bool from_host __maybe_unused, bool hairpin_flow __maybe_unused,
 		  bool from_tunnel __maybe_unused, __u32 cluster_id __maybe_unused)
 {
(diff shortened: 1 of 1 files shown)


aditighag (commented):
I don't see any good reason to pass the identity if the goal is to just bypass ingress policy for loopback traffic anyway. The loopback snat ip has a special meaning. So if we were to assign a special identity, and use that to skip the additional conntrack lookups in the per ep policy enforcement BPF program, it would make sense to pass the correct identity. (I don't remember when the reverse snat happens though). There are couple of considerations to this approach though: do we plumb a special ipcache entry via userspace, or keep it all in the datapath since packet context carries the identity thereby skipping ipcache lookups? Any thoughts?

julianwiedmann:
> I don't see any good reason to pass the identity if the goal is to just bypass ingress policy for loopback traffic anyway.

As described, there's other users of the security identity in the `to-container` code. And pushing the loopback handling in `l3_local_delivery()` down a bit further seems like a nice win in itself. Maybe one day it's gone completely :shrug:.

> The loopback snat ip has a special meaning. So if we were to assign a special identity, and use that to skip the additional conntrack lookups in the per ep policy enforcement BPF program, it would make sense to pass the correct identity. (I don't remember when the reverse snat happens though).

As mentioned in the patch description, at this point we're dealing with `ip4->saddr = Service_IP`.

> There are couple of considerations to this approach though: do we plumb a special ipcache entry via userspace, or keep it all in the datapath since packet context carries the identity thereby skipping ipcache lookups? Any thoughts?

There's nothing special about passing the "actual" identity via metadata, even when it doesn't match the current packet content. I don't think we suddenly need to start populating the ipcache for these kind of cases.

aditighag (approved):

aditighag:
>As mentioned in the patch description, at this point we're dealing with ip4->saddr = Service_IP.

Sorry, I missed that.

PR #27818, opened 2023-08-30
ingress: fix panic on ingress rule without HTTPIngressRule

mhofstetter:
Currently, defining an `Ingress` without an `HTTPIngressRule` (e.g. only Host set) results in a panic in the Cilium Operator.

Therefore, this commit changes the ingress ingestion to process the HTTP paths only if the HTTPIngressRule is set on the rule.

### DIFF
--- a/operator/pkg/model/ingestion/ingress.go
+++ b/operator/pkg/model/ingestion/ingress.go
@@ -90,6 +90,12 @@ func Ingress(ing slim_networkingv1.Ingress, defaultSecretNamespace, defaultSecre
 		}
 
 		l.Hostname = host
+		if rule.HTTP == nil {
+			log.WithField(logfields.Ingress, ing.Namespace+"/"+ing.Name).
+				Warn("Invalid Ingress rule without spec.rules.HTTP defined, skipping rule")
+			continue
+		}
+
 		for _, path := range rule.HTTP.Paths {
 
 			route := model.HTTPRoute{}
--- a/operator/pkg/model/ingestion/ingress_test.go
+++ b/operator/pkg/model/ingestion/ingress_test.go
@@ -1270,6 +1270,21 @@ func useDefaultListenersTLSsecret(listeners []model.HTTPListener) []model.HTTPLi
 	return ret
 }
 
+func removeIngressHTTPRuleValues(ing slim_networkingv1.Ingress) slim_networkingv1.Ingress {
+	var rules []slim_networkingv1.IngressRule
+
+	for _, r := range ing.Spec.Rules {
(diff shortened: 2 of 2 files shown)


operator/pkg/model/ingestion/ingress.go
```diff
@@ -90,35 +90,37 @@ func Ingress(ing slim_networkingv1.Ingress, defaultSecretNamespace, defaultSecre
...
-
-			switch *path.PathType {
-			case slim_networkingv1.PathTypeExact:
-				route.PathMatch.Exact = path.Path
-			case slim_networkingv1.PathTypePrefix:
-				route.PathMatch.Prefix = path.Path
-			case slim_networkingv1.PathTypeImplementationSpecific:
-				route.PathMatch.Regex = path.Path
-			}
+		if rule.HTTP != nil {
```
margamanterola: How about inverting the logic here to avoid the extra nesting? I do note that the code is updating the map entry at the end which would be skipped if we do a `continue` here, does it really need to be updated even when there were no http entries in the rule? If so, then I guess it means a duplicated line before the `continue`, but I think the code would still be more readable without the extra nesting.

mhofstetter: thanks - good point and definitely way more readable. i'd like to keep updating the map to ensure that the other properties are properly reflected in the envoy config.

update: introduced logging a warning and skipping complete rule

operator/pkg/model/ingestion/ingress.go
```diff
@@ -90,6 +90,11 @@ func Ingress(ing slim_networkingv1.Ingress, defaultSecretNamespace, defaultSecre
 		}
 
 		l.Hostname = host
+		if rule.HTTP == nil {
```
mhofstetter, line 93: An additional aspect to this problem is that we actually don't validate (and potentially reject) the CRD.

According to the Ingres spec `HTTP` needs to be set anyway.

```
// IngressRuleValue represents a rule to apply against incoming requests. If the
// rule is satisfied, the request is routed to the specified backend. Currently
// mixing different types of rules in a single Ingress is disallowed, so exactly
// one of the following must be set.
type IngressRuleValue struct {
	// +optional
	HTTP *HTTPIngressRuleValue `json:"http,omitempty" protobuf:"bytes,1,opt,name=http"`
}
```

mhofstetter: Update: In case an Ingress rule doesn't define the HTTP rules, a warning gets logged

meyskens (approved):
LGTM for me

PR #27820, opened 2023-08-30
go.mod, renovate: specify and update Go toolchain version

tklauser:
According to [1] as of Go 1.21 we either need to specify the full toolchain version in the `go` directive or add a `toolchain` directive with the concrete toolchain version. Opt for the former and make sure it's kept up to date by renovate bot.

[1] https://github.com/golang/go/issues/62278#issuecomment-1693538776

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -482,6 +482,14 @@
       "matchStrings": [
         "# renovate: datasource=(?<datasource>.*?) depName=(?<depName>.*?)\\s+.+_VERSION=\"(?<currentValue>.*)\""
       ]
+    },
+    {
+      "fileMatch": [
+        "^go.mod$",
+      ],
+      "matchStrings": [
+        "// renovate: datasource=(?<datasource>.*?) depName=(?<depName>.*?)\\s+.+go (?<currentValue>.*)"
+      ]
     }
   ]
 }
--- a/.github/workflows/conformance-runtime.yaml
+++ b/.github/workflows/conformance-runtime.yaml
@@ -53,6 +53,10 @@ concurrency:
     }}
   cancel-in-progress: true
 
+env:
+  # renovate: datasource=golang-version depName=go
+  go-version: 1.21.0
+
 jobs:
   commit-status-start:
     name: Commit Status Start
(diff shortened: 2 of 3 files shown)


rolinh (approved):

tklauser:
Pushed an additional commit to run the privileged runtime tests with the proper Go toolchain version. See commit message of the 1st commit for details.

qmonnet (approved):
Looks good as far as I can tell

brlbil (approved):
LGTM

joestringer:
I'm getting errors in my editor like this after merging this PR:

```
Error loading workspace /home/joe/git/cilium/go.mod:4: invalid go version '1.21.0': must match format 1.23 
```

I've updated my local Go binary to v1.21.0 but that doesn't seem to help.

tklauser:
> I'm getting errors in my editor like this after merging this PR:
> 
> ```
> Error loading workspace /home/joe/git/cilium/go.mod:4: invalid go version '1.21.0': must match format 1.23 
> ```
> 
> I've updated my local Go binary to v1.21.0 but that doesn't seem to help.

Without knowing your exact editor setup it's hard to tell what's wrong. I assume you're using some sort of Go integration like go-vim or similar? These extensions usually rely on various supporting tool binaries (e.g. gopls, linters etc). Could you try rebuilding these using Go 1.21 as well? Also cleaning all Go caches using `go clean -cache -modcache -testcache -fuzzcache` might help.

joestringer:
:+1: thanks for the tips. `go clean` didn't help, but `PlugUpdate` in vim to update my plugins did resolve the issue.

joestringer:
@tklauser why do we need to keep the patch version here in sync with upstream? Is this how we derive the Go version to install and build with in CI? Put another way, is it possible to keep the go.mod at 1.21.0 until we migrate to 1.22?

tklauser:
> @tklauser why do we need to keep the patch version here in sync with upstream? Is this how we derive the Go version to install and build with in CI? Put another way, is it possible to keep the go.mod at 1.21.0 until we migrate to 1.22?

@joestringer the idea here was to keep up with the latest patch version in case anything (e.g. CI) or anyone is relying on `go.mod` as a source of truth for the Go version. I agree that this might cause friction and from quickly glancing through our CI workflows it doesn't look like any of them is using the Go version in `go.mod` as a source of truth, so SGTM to keep `go.mod` at 1.x.0 until we switch to 1.(x+1).0.

PR #27823, opened 2023-08-30
Register cluster-id and cluster-name flags through hive

giorio94:
Currently, the clustermesh package already leverages a ClusterIDName struct to wrap the ClusterID and ClusterName information. Let's use it also to register the associated flags, so that we can directly depend on it on the different components, and reuse it for the kvstoremesh and clustermesh-apiserver binaries uniformly.

To avoid the need for modifying all usages, we preserve the current fields inside DaemonConfig, which continue to be populated as before.

### DIFF
--- a/clustermesh-apiserver/users_mgmt.go
+++ b/clustermesh-apiserver/users_mgmt.go
@@ -12,6 +12,7 @@ import (
 	"github.com/spf13/pflag"
 	"gopkg.in/yaml.v3"
 
+	cmtypes "github.com/cilium/cilium/pkg/clustermesh/types"
 	"github.com/cilium/cilium/pkg/controller"
 	"github.com/cilium/cilium/pkg/fswatcher"
 	"github.com/cilium/cilium/pkg/hive"
@@ -57,6 +58,7 @@ type usersConfigFile struct {
 
 type usersManager struct {
 	UsersManagementConfig
+	clusterInfo cmtypes.ClusterInfo
 
 	client        kvstore.BackendOperationsUserMgmt
 	clientPromise promise.Promise[kvstore.BackendOperationsUserMgmt]
@@ -68,7 +70,12 @@ type usersManager struct {
 	wg   sync.WaitGroup
 }
 
-func registerUsersManager(lc hive.Lifecycle, cfg UsersManagementConfig, clientPromise promise.Promise[kvstore.BackendOperationsUserMgmt]) error {
+func registerUsersManager(
+	lc hive.Lifecycle,
+	cfg UsersManagementConfig,
+	cinfo cmtypes.ClusterInfo,
+	clientPromise promise.Promise[kvstore.BackendOperationsUserMgmt],
+) error {
 	if !cfg.ClusterUsersEnabled {
 		log.Info("etcd users management disabled")
 		return nil
@@ -167,7 +174,7 @@ func (us *usersManager) sync(ctx context.Context) error {
 	}
 
 	for _, user := range users.Users {
-		if user.Name == cfg.clusterName {
+		if user.Name == us.clusterInfo.Name {
 			continue
 		}
 
--- a/clustermesh-apiserver/vmmanager.go
+++ b/clustermesh-apiserver/vmmanager.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/cilium/cilium/api/v1/models"
 	"github.com/cilium/cilium/pkg/cidr"
+	cmtypes "github.com/cilium/cilium/pkg/clustermesh/types"
 	"github.com/cilium/cilium/pkg/identity"
 	identityCache "github.com/cilium/cilium/pkg/identity/cache"
 	identitymodel "github.com/cilium/cilium/pkg/identity/model"
@@ -38,6 +39,8 @@ import (
 )
 
 type VMManager struct {
+	clusterInfo cmtypes.ClusterInfo
+
 	ciliumClient      clientset.Interface
 	identityAllocator *identityCache.CachingIdentityAllocator
 
@@ -47,15 +50,16 @@ type VMManager struct {
 	backend kvstore.BackendOperations
 }
 
-func NewVMManager(clientset k8sClient.Clientset, backend kvstore.BackendOperations) *VMManager {
+func NewVMManager(cinfo cmtypes.ClusterInfo, clientset k8sClient.Clientset, backend kvstore.BackendOperations) *VMManager {
 	m := &VMManager{
+		clusterInfo:  cinfo,
 		ciliumClient: clientset,
 		backend:      backend,
 	}
 	m.identityAllocator = identityCache.NewCachingIdentityAllocator(m)
 
 	if option.Config.EnableWellKnownIdentities {
-		identity.InitWellKnownIdentities(option.Config)
+		identity.InitWellKnownIdentities(option.Config, cinfo)
 	}
 	m.identityAllocator.InitIdentityAllocator(clientset)
 	m.startCiliumExternalWorkloadWatcher(clientset)
@@ -107,7 +111,7 @@ func (m *VMManager) GetNodeSuffix() string {
 	return "vm-allocator"
 }
 
-func nodeOverrideFromCEW(n *nodeTypes.RegisterNode, cew *ciliumv2.CiliumExternalWorkload) *nodeTypes.RegisterNode {
+func (m *VMManager) nodeOverrideFromCEW(n *nodeTypes.RegisterNode, cew *ciliumv2.CiliumExternalWorkload) *nodeTypes.RegisterNode {
 	nk := n.DeepCopy()
 
 	nk.Labels = make(map[string]string, len(cew.Labels))
@@ -124,9 +128,9 @@ func nodeOverrideFromCEW(n *nodeTypes.RegisterNode, cew *ciliumv2.CiliumExternal
 	}
 
 	// Override cluster
-	nk.Cluster = cfg.clusterName
-	nk.ClusterID = cfg.clusterID
-	nk.Labels[k8sConst.PolicyLabelCluster] = cfg.clusterName
+	nk.Cluster = m.clusterInfo.Name
+	nk.ClusterID = m.clusterInfo.ID
+	nk.Labels[k8sConst.PolicyLabelCluster] = m.clusterInfo.Name
 
 	// Override CIDRs if defined
 	if cew.Spec.IPv4AllocCIDR != "" {
@@ -174,7 +178,7 @@ func (m *VMManager) OnUpdate(k store.Key) {
 
 		if n.NodeIdentity == 0 {
 			// Phase 1: Initial registration with zero ID, return configuration
-			nk := nodeOverrideFromCEW(n, cew)
+			nk := m.nodeOverrideFromCEW(n, cew)
 
 			log.Debugf("CEW: VM Cilium Node updated: %v -> %v", n, nk)
 			// FIXME: GH-17909 Balance this call with a call to release the identity.
@@ -197,7 +201,7 @@ func (m *VMManager) OnUpdate(k store.Key) {
 			// Phase 2: non-zero ID registration with addresses
 
 			// Override again, just in case the external node is misbehaving
-			nk := nodeOverrideFromCEW(n, cew)
+			nk := m.nodeOverrideFromCEW(n, cew)
 
 			id := m.LookupNodeIdentity(nk)
 			if id == nil || id.ID.Uint32() != nk.NodeIdentity {
--- a/operator/cmd/root.go
+++ b/operator/cmd/root.go
@@ -35,6 +35,7 @@ import (
 	"github.com/cilium/cilium/operator/pkg/ingress"
 	"github.com/cilium/cilium/operator/pkg/lbipam"
 	operatorWatchers "github.com/cilium/cilium/operator/watchers"
+	cmtypes "github.com/cilium/cilium/pkg/clustermesh/types"
 	"github.com/cilium/cilium/pkg/components"
 	"github.com/cilium/cilium/pkg/controller"
 	"github.com/cilium/cilium/pkg/defaults"
@@ -100,6 +101,9 @@ var (
 		"operator-controlplane",
 		"Operator Control Plane",
 
+		cell.Config(cmtypes.DefaultClusterInfo),
+		cell.Invoke(func(cinfo cmtypes.ClusterInfo) error { return cinfo.Validate() }),
+
 		cell.Invoke(
 			registerOperatorHooks,
 		),
@@ -119,9 +123,7 @@ var (
 			return identitygc.SharedConfig{
 				IdentityAllocationMode: daemonCfg.IdentityAllocationMode,
 				EnableMetrics:          operatorCfg.EnableMetrics,
-				ClusterName:            daemonCfg.LocalClusterName(),
 				K8sNamespace:           daemonCfg.CiliumNamespaceName(),
-				ClusterID:              daemonCfg.LocalClusterID(),
 			}
 		}),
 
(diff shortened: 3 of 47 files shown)


giorio94:
Rebased onto main to pick CI changes

gandro (approved):
Code-wise looks good to me. 

I'd be interested to know more about the motivation behind this change, and what the plan is going forward. The cluster name is something that is needed by many subsystems to namespace global resources: Node names, EC2 network interface tags in ENI IPAM, service labels, identity labels.

Is the goal that all those subsystems (node discovery, IPAM, service, identity allocation, IPCache) eventually rely on the clustermesh cell? Will that not cause potentially circular cell dependencies? 

While I agree we should reduce global configuration state, the cluster name is something that is inherently global and relevant to a lot of subsystems. If we want to make the clustername and cluster id a dependency to be injected into other subsystems, it should probably be a leaf dependency that depends on nothing else, so I'm not sure if the clustermesh cell is the right place (since it itself has lots of dependencies already). Did you consider introducing a separate cluster-id cell instead?

giorio94:
> Code-wise looks good to me.
> 
> I'd be interested to know more about the motivation behind this change, and what the plan is going forward. The cluster name is something that is needed by many subsystems to namespace global resources: Node names, EC2 network interface tags in ENI IPAM, service labels, identity labels.
> 

Essentially this came up while refactoring the clustermesh-apiserver and kvstoremesh in preparation to merging them into a single binary (as part of a larger effort to unify all clustermesh-apiserver components into a single image for simplicity: https://github.com/cilium/cilium/issues/23770). 

The idea is to extract these two flags into a separate object injected through Hive, so that we can reuse them uniformly in all the different components, as well as remove the hacky conversions from the global DaemonConfig, which were for instance used before in the clustermesh cell. Over time, it would be great if also the other subsystems requiring these parameter could depend directly on the `ClusterIDName` struct, rather than reading the global DaemonConfig.

> Is the goal that all those subsystems (node discovery, IPAM, service, identity allocation, IPCache) eventually rely on the clustermesh cell? Will that not cause potentially circular cell dependencies?
> 
> While I agree we should reduce global configuration state, the cluster name is something that is inherently global and relevant to a lot of subsystems. If we want to make the clustername and cluster id a dependency to be injected into other subsystems, it should probably be a leaf dependency that depends on nothing else, so I'm not sure if the clustermesh cell is the right place (since it itself has lots of dependencies already). Did you consider introducing a separate cluster-id cell instead?

The different subsystems can depend on the `ClusterIDName` struct alone, without needing to depend on the whole ClusterMesh object (which would pose circular dependency issues as you mentioned). For the same reason, the struct is defined in a separate subpackage, and registered at the top of the cell hierarchy, not as part of the clustermesh cell (which only depends on it).

gandro:
>  For the same reason, the struct is defined in a separate subpackage, and registered at the top of the cell hierarchy, not as part of the clustermesh cell (which only depends on it).

Awesome, thanks for clarifying. I'm still not super familiar with the hive framwork, so I didn't know one could depend on struct alone this way. Sounds good to me!

pkg/option/config.go
```diff
@@ -2888,7 +2882,7 @@ func (c *DaemonConfig) Validate(vp *viper.Viper) error {
 	if c.ClusterID != 0 {
```
derailed: Since we have a new type here we could perhaps encapsulate validation to ensure the values are legit vs externalizing it?

pkg/option/config.go
```diff
@@ -3036,8 +3030,8 @@ func (c *DaemonConfig) Populate(vp *viper.Viper) {
 	c.AutoCreateCiliumNodeResource = vp.GetBool(AutoCreateCiliumNodeResource)
 	c.BPFRoot = vp.GetString(BPFRoot)
 	c.CGroupRoot = vp.GetString(CGroupRoot)
-	c.ClusterID = vp.GetUint32(ClusterIDName)
-	c.ClusterName = vp.GetString(ClusterName)
+	c.ClusterID = vp.GetUint32(clustermeshTypes.OptClusterID)
```
derailed, line 3019: Could we use the new type here ie c.ClusterIDName = NewClusterIDName(...)?

pkg/clustermesh/types/option.go
```diff
@@ -0,0 +1,29 @@
...
+const (
+	// OptClusterName is the name of the OptClusterName option
+	OptClusterName = "cluster-name"
+
+	// OptClusterID is the name of the OptClusterID option
+	OptClusterID = "cluster-id"
+)
+
+// ClusterIDName groups together the ClusterID and the ClusterName
+type ClusterIDName struct {
```
derailed: nit: No keen on the naming here for 3 reasons:

1. Naming the struct for what it contains vs what it provides i.e ClusterInfo or ClusterMeta.
2. We are already in the context of a cluster thus we could dry the members names aka ID, Name if we weren't using anonymous aggregation that is...
3. If we were to add more fields later on the current name would be a bit of a dud.

pkg/clustermesh/types/option.go
```diff
@@ -0,0 +1,29 @@
...
+	ClusterID   uint32
+	ClusterName string
+}
+
+func (ClusterIDName) Flags(flags *pflag.FlagSet) {
+	flags.Uint32(OptClusterID, 0, "Unique identifier of the cluster")
```
derailed: Should we use a default here for DefaultClusterID to parallel the clustername default?

pkg/option/config.go
derailed: Perhaps leverage DefaultClusterID if defined??

operator/watchers/k8s_service_sync.go
```diff
@@ -136,7 +125,7 @@ func StartSynchronizingServices(ctx context.Context, wg *sync.WaitGroup, cfg Ser
 		<-kvstoreReady
 
 		log.Info("Starting to synchronize Kubernetes services to kvstore")
-		k8sServiceHandler(ctx, cfg.LocalClusterName(), cfg.SharedOnly, cfg.LocalClusterID())
+		k8sServiceHandler(ctx, cfg.ClusterName, cfg.SharedOnly, cfg.ClusterID)
```
derailed: Perhaps we could streamline args in these calls and pass in ClusterIDName instead? if considered in scope with this PR or later refactor...

clustermesh-apiserver/vmmanager.go
```diff
@@ -124,9 +128,9 @@ func nodeOverrideFromCEW(n *nodeTypes.RegisterNode, cew *ciliumv2.CiliumExternal
 	}
 
 	// Override cluster
-	nk.Cluster = cfg.clusterName
-	nk.ClusterID = cfg.clusterID
-	nk.Labels[k8sConst.PolicyLabelCluster] = cfg.clusterName
+	nk.Cluster = m.ClusterName
```
derailed: Not sure if applicable here but if RegisterNode used CluserIDName vs separates might clean up our structs a bit more?

clustermesh-apiserver/users_mgmt.go
```diff
@@ -57,6 +58,7 @@ type usersConfigFile struct {
 
 type usersManager struct {
 	UsersManagementConfig
+	cmtypes.ClusterIDName
```
derailed: Does this have to be anonymous? If not we could dry up the naming...

derailed (commented):
@giorio94 Very nice work Marco and nice clean up! I think there might be an opportunity to further improve, but I am a bit out of context here, so feel free to disregard..

clustermesh-apiserver/vmmanager.go
giorio94: I wouldn't personally push this refactoring that far, because of the number of consequential changes. The idea of this PR is mainly to uniform the flags registration, without touching the rest of the usages as much as possible.

pkg/option/config.go
giorio94: Makes sense, I've moved the validation logic as methods of the `ClusterInfo` struct.

pkg/option/config.go
giorio94: This is tricky, because the only way to keep the current names of the fields (to avoid codebase-wide changes) would be to embed the `ClusterInfo` struct into `DaemonConfig`, which feels weird as it would also inherit the associated methods. Although suboptimal, I'd keep the current approach.

pkg/clustermesh/types/option.go
giorio94: Yeah, I agree that this naming was suboptimal. For context, when I introduced this `ClusterIDName` struct a while ago, it was essentially meant to be embedded inside other structs, which is the reason for the more verbose `ClusterID`/`ClusterName` field names. I've now renamed the struct to `ClusterInfo`, and the fields to `ID` and `Name`.

pkg/clustermesh/types/option.go
giorio94: There's currently no `DefaultClusterID` constant defined anywhere, and we typically just compare against 0. I would not introduce such constant though, both for consistency with the existing code, and because depending on the scope a `DefaultClusterID` might not make sense as a thing (essentially, ClusterID=0 disables clustermesh functionalities). I've defined a `DefaultClusterInfo` variable though, to avoid hardcoding the defaults inside the flags definitions.

clustermesh-apiserver/users_mgmt.go
giorio94: I've removed all anonymous usages of this struct to dry up the naming.

pkg/clustermesh/types/option.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// OptClusterID is the name of the OptClusterID option
+	OptClusterID = "cluster-id"
+)
+
+// ClusterInfo groups together the ClusterID and the ClusterName
+type ClusterInfo struct {
+	ID   uint32
+	Name string
+}
```
giorio94: @thorn3r I wonder at this point if it could make sense to add the `MaxClusters` inside this struct, and adapt the validation based on that. The only caveat is that it will be also inherited by kvstoremesh at that point, but that's not a big issue since it comes for free. More in general, do you think that the changes introduced here would play well with https://github.com/cilium/cilium/pull/27520?

pkg/clustermesh/types/option.go
giorio94: So... This change ended up breaking everything, because the name of the flag must match that of the target struct field. Let me rework this differently.

giorio94: It turned out to be easier than expected, just needed to add the tags.

giorio94:
@derailed PTAL

pkg/clustermesh/types/option.go
thorn3r: yes I like that idea. I was already having some difficulty with naming a new type that essentially just has an embedded integer :laughing:. Additionally, with the changes in https://github.com/cilium/cilium/pull/27520 the `MaxConnectedClusters` becomes essential to more than just ClusterMesh (it affects identity allocation), so in the absence of refactoring pkg/identity into a hive cell, this seems like the best option :+1:

thorn3r (approved):
excellent work :rocket: This will help simplify the refactor I've been working on for https://github.com/cilium/cilium/pull/27520 :pray:

pkg/clustermesh/types/option.go
```diff
@@ -0,0 +1,68 @@
...
+func (c ClusterInfo) ValidateStrict() error {
+	if err := ValidateClusterID(c.ID); err != nil {
+		return err
+	}
+
+	return c.validateName()
+}
+
+func (c ClusterInfo) validateName() error {
+	if c.ID != 0 {
```
derailed: nit: might as well return early if c.ID == 0

pkg/clustermesh/types/option.go
```diff
@@ -0,0 +1,68 @@
...
+	ID   uint32 `mapstructure:"cluster-id"`
+	Name string `mapstructure:"cluster-name"`
+}
+
+var DefaultClusterInfo = ClusterInfo{
+	ID:   0,
+	Name: defaults.ClusterName,
+}
+
+func (def ClusterInfo) Flags(flags *pflag.FlagSet) {
```
derailed, line 35: nit: comment

pkg/clustermesh/types/option.go
```diff
@@ -0,0 +1,68 @@
...
+	OptClusterID = "cluster-id"
+)
+
+// ClusterInfo groups together the ClusterID and the ClusterName
+type ClusterInfo struct {
+	ID   uint32 `mapstructure:"cluster-id"`
+	Name string `mapstructure:"cluster-name"`
+}
+
+var DefaultClusterInfo = ClusterInfo{
```
derailed, line 29: nit: comment

pkg/option/config.go
derailed: Thanks Marco! I see what you are saying. Figured it might be an opportunity to begin cleaning this insanely large struct in favor of segmenting based on config domain. Using named association would scope the delegation vs `inheriting` the behaviors...  But in this case clusterInfo validation when it comes to configuration would be a good thing anyway. But we would need to establish pattens across the board which is a big undertaking at this stage ;(

derailed (commented):
@giorio94 Thanks for the updates! Just a couple more picks...

pkg/clustermesh/types/option.go
giorio94: I've merged the two if statements into one.

pkg/option/config.go
giorio94: I definitely second the idea of removing configurations from the DaemonConfig struct. I took another pass at removing a few more usages of the ClusterID/ClusterName fields when possible (mainly the ones which would also be useful in the context of https://github.com/cilium/cilium/pull/27520, but I wouldn't go further as not to excessively increase the scope of this PR.

giorio94:
Removed CI-related reviewers because that commit has been dropped.

giorio94:
I've reverted the changes affecting external workloads, as the scope of this PR was increasing too much. Let's see if now it passes all tests.

giorio94:
@derailed PTAL

pkg/endpoint/endpoint_test.go
```diff
@@ -141,7 +140,6 @@ func NewCachingIdentityAllocator(owner cache.IdentityAllocatorOwner) fakeIdentit
 func (s *EndpointSuite) SetUpTest(c *C) {
 	/* Required to test endpoint CEP policy model */
 	kvstore.SetupDummy(c, "etcd")
-	identity.InitWellKnownIdentities(&fakeConfig.Config{})
```
tommyp1ckles, line 144: Why do we  not need these here anymore?

tommyp1ckles (approved):

pkg/endpoint/endpoint_test.go
giorio94: WellKnownIdentities are used only in very specific cases, mainly to bootstrap Cilium with etcd running in pod network (to break the chicken egg dependency). Given that they were not necessary here (as well as in the clustermesh case), I just removed the line altogether instead of adapting the parameters.

operator/cmd/root.go
```diff
@@ -100,6 +101,9 @@ var (
 		"operator-controlplane",
 		"Operator Control Plane",
 
+		cell.Config(cmtypes.DefaultClusterInfo),
+		cell.Invoke(func(cinfo cmtypes.ClusterInfo) error { return cinfo.Validate() }),
```
joamaki, line 105: I think you don't need to wrap it, but rather you can just do: `cell.Invoke(cmtypes.ClusterInfo.Validate)`. Debatable whether this is less or more confusing, but it is shorter.. :D

Why do we need `Validate` here, but not in `daemon/cmd/cells.go`? If it's missing from there, how about doing `cmtypes.ConfigCell` that's `cell.Group(cell.Config(DefaultClusterInfo), cell.Invoke(ClusterInfo.Validate))`? Preferably this config would be part of some module and not referenced naked here like this, but perhaps that's to come in the future?

pkg/option/config.go
joamaki: How about instead of pulling them from viper like this, we would instead fill them into `DaemonConfig` from an `Invoke` that is part of the config cell (see my earlier comment)? Or alternatively leave it as-is, if these are going away soon anyway.

joamaki (approved):

derailed (approved):
@giorio94 LGTM

nathanjsweet (approved):

operator/cmd/root.go
giorio94: > I think you don't need to wrap it, but rather you can just do: `cell.Invoke(cmtypes.ClusterInfo.Validate)`. Debatable whether this is less or more confusing, but it is shorter.. :D

Uh, that's right. I didn't realize it.

> Why do we need Validate here, but not in daemon/cmd/cells.go? If it's missing from there, how about doing cmtypes.ConfigCell that's cell.Group(cell.Config(DefaultClusterInfo), cell.Invoke(ClusterInfo.Validate))? Preferably this config would be part of some module and not referenced naked here like this, but perhaps that's to come in the future?

So, the thing is that depending on the component we need to perform a slightly different validation, mainly because for kvstoremesh we don't allow the cluster ID to be zero (which also means that the check needs to be performed from a start hook, as it is the default value). The validation in the agent is currently still part of `Config.Validate` mainly because that is performed earlier to prevent possible issues for components which do not directly depend on the `ClusterInfo` struct.

pkg/option/config.go
giorio94: The main issue of pulling it from an invoke hook is that we would postpone a bit the initialization, which would be an issue if any (legacy) component not directly depending on it would try to read these values. The current solution seems a bit less error prone to me from that point of view.

PR #27835, opened 2023-08-30
metric: provide way to declare labels.

tommyp1ckles:
metric: provide way to declare labels.
    
    Adds *WithLabels Vec[T] constructors for histogram/gauge/counter.
    This provides a mechanism to declare a metrics possible range of label values up front.
    The xxxWithLabels functions will automatically initialize such metrics
    such that they will be exported with zero values by the prometheus
    endpoint.
    
    One of the stated best practices for Prometheus instrumentation is
    avoiding missing metrics [1].
    
    This change will make it easier to provide metrics that are initialized
    correctly to produce more reliable metrics.
    
    [1] https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics

```release-note
Pre-initialize several known metric vectors to avoid empty metrics (specifically: endpoint_regenerations_total, policy_change_total,  policy_implementation_delay, policy_l7_total and kubernetes_events metrics). 
```

### DIFF
--- a/pkg/metrics/metric/collections/product.go
+++ b/pkg/metrics/metric/collections/product.go
@@ -1,0 +1,74 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package collections
+
+// CartesianProduct returns the cartesian product of the input vectors as
+// a vector of vectors, each with length the same as the number of input vectors.
+func CartesianProduct[T any](vs ...[]T) [][]T {
+	if len(vs) == 0 {
+		return [][]T{}
+	}
+
+	dimension := len(vs) // Each output will be a vector of this length.
+	// Iterate to find out the number of output vectors.
+	size := len(vs[0])
+	for i := 1; i < len(vs); i++ {
+		size *= len(vs[i])
+	}
+
+	// Allocate the output vectors.
+	dst := make([][]T, size)
+	for i := range dst {
+		dst[i] = make([]T, dimension)
+	}
+
+	lastm := 1
+	for i := 0; i < dimension; i++ {
+		permuteColumn[T](dst, i, lastm, vs[i])
+		lastm = lastm * len(vs[i])
+	}
+	return dst
+}
+
+// permuteColumn fills in the nth column of the output vectors of the cartesian
+// product of the input vectors.
+//
+// leftPermSize is the number of vectors as a result of permuting 0,..,col-1 columns.
+// That is, this is the block size upon which we will repeat the values of v0 such that
+// every previous permutation is again permuted with each value of v0.
+//
+// For ex.
+// CartesianProduct[string]({"a", "b"}, {"x", "y", "z"})
+//
+// Iteration (i.e. col, leftPermSize=1) 1:
+//
+// dst = [
+// ["a"],
+// ["b"],
+// ["a"]
+// ["b"]
+// ["a"]
+// ["b"]
+// ]
+//
+// Iteration (leftPermSize=2):
+//
+// dst = [
+// ["a", "x"], // <- each elem of vec is repeated leftPermSize times.
+// ["b", "x"],
+// ["a", "y"]
+// ["b", "y"]
+// ["a", "z"]
+// ["b", "z"]
+// ]
+func permuteColumn[T any](dst [][]T, col int, leftPermSize int, vec []T) {
+	// Go down the column with the current lhs.
+	// You want to skip along, lastm elements at a time.
+	for i := 0; i < len(dst); i += leftPermSize { // So we're skipping n rows at a time,
+		vi := (i / leftPermSize) % len(vec)
+		for off := 0; off < leftPermSize; off++ { // this is a repeat
+			dst[i+off][col] = vec[vi]
+		}
+	}
+}
--- a/pkg/metrics/metric/collections/product_test.go
+++ b/pkg/metrics/metric/collections/product_test.go
@@ -1,0 +1,63 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package collections
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIteratorProduct(t *testing.T) {
+	assert := assert.New(t)
+	p := CartesianProduct(
+		[]string{"CiliumNetworkPolicy", "CiliumClusterwideNetworkPolicy", "NetworkPolicyName"},
+		[]string{"update", "delete"},
+		[]string{"true", "false"},
+		[]string{"0", "1"},
+	)
+	assert.Len(p, 24, "Should be 3 * 2 * 2 * 2 = 24 permutations")
+	vs := map[string]any{}
+	for _, v := range p {
+		vs[fmt.Sprintf("%v", v)] = v
+	}
+	assert.Len(vs, 24, "Elements should be unique")
+}
+
+func TestIteratorProductElements(t *testing.T) {
+	assert := assert.New(t)
+	p := CartesianProduct(
+		[]string{"true", "false"},
+		[]string{"0", "1"},
+		[]string{"foo"},
+	)
+	assert.Equal(4, len(p), "Should be 2 * 2 * 1 = 4 permutations")
+	assert.Equal(len(p[0]), 3)
+	vs := map[string]any{}
+	for _, v := range p {
+		vs[fmt.Sprintf("%v", v)] = v
+	}
+	assert.Contains(vs, "[true 0 foo]")
+	assert.Contains(vs, "[true 1 foo]")
+	assert.Contains(vs, "[false 0 foo]")
+	assert.Contains(vs, "[false 1 foo]")
+}
+
+func TestIteratorProductEmpty(t *testing.T) {
+	assert := assert.New(t)
+	p := CartesianProduct(
+		[]string{"CiliumNetworkPolicy", "CiliumClusterwideNetworkPolicy", "NetworkPolicyName"},
+		[]string{},
+	)
+	assert.Empty(p)
+
+	p = CartesianProduct[string]()
+	assert.Empty(p)
+
+	assert.Empty(CartesianProduct([]string{}, []string{}, []string{}))
+	assert.Empty(0, CartesianProduct[int]())
+	assert.Len(CartesianProduct([]string{""}, []string{""}, []string{""}), 1)
+	CartesianProduct[interface{}](nil, nil) // Test some weird cases.
+}
--- a/pkg/metrics/metric/counter.go
+++ b/pkg/metrics/metric/counter.go
@@ -60,7 +60,9 @@ func (c *counter) Add(val float64) {
 	}
 }
 
-func NewCounterVec(opts CounterOpts, labelNames []string) DeletableVec[Counter] {
+// NewCounterVec creates a new DeletableVec[Counter] based on the provided CounterOpts and
+// partitioned by the given label names.
+func NewCounterVec(opts CounterOpts, labelNames []string) *counterVec {
 	return &counterVec{
 		CounterVec: prometheus.NewCounterVec(opts.toPrometheus(), labelNames),
 		metric: metric{
@@ -70,12 +72,47 @@ func NewCounterVec(opts CounterOpts, labelNames []string) DeletableVec[Counter] 
 	}
 }
 
+// NewCounterVecWithLabels creates a new DeletableVec[Counter] based on the provided CounterOpts and
+// partitioned by the given labels.
+// This will also initialize the labels with the provided values so that metrics with known label value
+// ranges can be pre-initialized to zero upon init.
+//
+// This should only be used when all label values are known at init, otherwise use of the
+// metric vector with uninitialized labels will result in warnings.
+//
+// Note: Disabled metrics will not have their label values initialized.
+//
+// For example:
+//
+//	NewCounterVecWithLabels(CounterOpts{
+//		Namespace: "cilium",
+//		Subsystem: "subsystem",
+//		Name:      "cilium_test",
+//		Disabled:  false,
+//	}, Labels{
+//		{Name: "foo", Values: NewValues("0", "1")},
+//		{Name: "bar", Values: NewValues("a", "b")},
+//	})
+//
+// Will initialize the following metrics to:
+//
+//	cilium_subsystem_cilium_test{foo="0", bar="a"} 0
+//	cilium_subsystem_cilium_test{foo="0", bar="b"} 0
+//	cilium_subsystem_cilium_test{foo="1", bar="a"} 0
+//	cilium_subsystem_cilium_test{foo="1", bar="b"} 0
+func NewCounterVecWithLabels(opts CounterOpts, labels Labels) *counterVec {
+	cv := NewCounterVec(opts, labels.labelNames())
+	initLabels[Counter](&cv.metric, labels, cv, opts.Disabled)
+	return cv
+}
+
 type counterVec struct {
 	*prometheus.CounterVec
 	metric
 }
 
 func (cv *counterVec) CurryWith(labels prometheus.Labels) (Vec[Counter], error) {
+	cv.checkLabels(labels)
 	vec, err := cv.CounterVec.CurryWith(labels)
 	if err == nil {
 		return &counterVec{CounterVec: vec, metric: cv.metric}, nil
@@ -118,6 +155,7 @@ func (cv *counterVec) GetMetricWithLabelValues(lvs ...string) (Counter, error) {
 }
 
 func (cv *counterVec) With(labels prometheus.Labels) Counter {
+	cv.checkLabels(labels)
 	if !cv.enabled {
 		return &counter{
 			metric: metric{enabled: false},
@@ -132,6 +170,7 @@ func (cv *counterVec) With(labels prometheus.Labels) Counter {
 }
 
 func (cv *counterVec) WithLabelValues(lvs ...string) Counter {
+	cv.checkLabelValues(lvs...)
 	if !cv.enabled {
 		return &counter{
 			metric: metric{enabled: false},
--- a/pkg/metrics/metric/gauge.go
+++ b/pkg/metrics/metric/gauge.go
@@ -95,14 +95,51 @@ func (g *gauge) SetToCurrentTime() {
 	}
 }
 
-func NewGaugeVec(opts GaugeOpts, labelNames []string) DeletableVec[Gauge] {
-	return &gaugeVec{
+// NewGaugeVec creates a new DeletableVec[Gauge] based on the provided GaugeOpts and
+// partitioned by the given label names.
+func NewGaugeVec(opts GaugeOpts, labelNames []string) *gaugeVec {
+	gv := &gaugeVec{
 		GaugeVec: prometheus.NewGaugeVec(opts.toPrometheus(), labelNames),
 		metric: metric{
 			enabled: !opts.Disabled,
 			opts:    Opts(opts),
 		},
 	}
+	return gv
+}
+
+// NewGaugeVecWithLabels creates a new DeletableVec[Gauge] based on the provided CounterOpts and
+// partitioned by the given labels.
+// This will also initialize the labels with the provided values so that metrics with known label value
+// ranges can be pre-initialized to zero upon init.
+//
+// This should only be used when all label values are known at init, otherwise use of the
+// metric vector with uninitialized labels will result in warnings.
+//
+// Note: Disabled metrics will not have their label values initialized.
+//
+// For example:
+//
+//	NewGaugeVecWithLabels(GaugeOpts{
+//		Namespace: "cilium",
+//		Subsystem: "subsystem",
+//		Name:      "cilium_test",
+//		Disabled:  false,
+//	}, Labels{
+//		{Name: "foo", Values: NewValues("0", "1")},
+//		{Name: "bar", Values: NewValues("a", "b")},
+//	})
+//
+// Will initialize the following metrics to:
+//
+//	cilium_subsystem_cilium_test{foo="0", bar="a"} 0
+//	cilium_subsystem_cilium_test{foo="0", bar="b"} 0
+//	cilium_subsystem_cilium_test{foo="1", bar="a"} 0
+//	cilium_subsystem_cilium_test{foo="1", bar="b"} 0
+func NewGaugeVecWithLabels(opts GaugeOpts, labels Labels) *gaugeVec {
+	gv := NewGaugeVec(opts, labels.labelNames())
+	initLabels[Gauge](&gv.metric, labels, gv, opts.Disabled)
+	return gv
 }
 
 type gaugeVec struct {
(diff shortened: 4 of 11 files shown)


pkg/metrics/metric/metric.go
```diff
@@ -196,3 +241,84 @@ func (b Opts) GetConfigName() string {
...
+
+func NewValues(vs ...string) Values {
+	vals := Values{}
+	for _, v := range vs {
+		vals[v] = struct{}{}
+	}
+	return vals
+}
+
+func NewLabel(name string, lvls ...string) Label {
```
chancez: I'm not sure I love the function signature because at a the call site, it's not so obvious the first value is the label, and the remaining are the values. Could we remove `NewLabel` and use composite literals instead? The `NewValues` could still exist as a helper to construct it.

At the call site it would then look like:
```
metric.Labels{
  {
    Name: "outcome", 
    Values: metric.NewValues(LabelValueOutcomeSuccess, LabelValueOutcomeFail),
  },
}
```

It might even be nice to just use a `[]string` instead of `Values`, and handle the duplication check somewhere else, then the call site is very simple and easy to understand.

tommyp1ckles: Agreed, moved to using the composite literal approach, seems much cleaner now.

chancez (approved):
Overall approach looks good.

dylandreimerink (approved):
LGTM!

bimmlerd (approved):

pkg/iterator/iterator.go
```diff
@@ -0,0 +1,73 @@
+// SPDX-License-Identifier: Apache-2.0
```
squeed: non-blocking suggestion? call this `pkg/collections` instead?

pkg/iterator/iterator.go
```diff
@@ -0,0 +1,73 @@
...
+// Copyright Authors of Cilium
+
+package iterator
+
+// VecIterator is an iterator that returns a vector of elements of type T for
+// each iteration.
+type VecIterator[T any] interface {
```
squeed: VecIterator seems like an internal implementation detail for CartesianProduct. Is exposing this type really necessary?

pkg/iterator/product.go
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package iterator
+
+// CartesianProduct computes the cartesian product of the given iterators, going
+// from left to right.
+// The returned iterator provides the final computed vectors.
+func CartesianProduct[T any](vs ...VecIterator[T]) VecIterator[T] {
```
squeed: Why not write this as ```func CartesianProduct[T any](vs [][]T) []T``` and remove the need for the wrapper type?

pkg/iterator/iterator.go
derailed: I am also unclear on why this is in a separate package why not roll this in the metrics/metric package?

pkg/iterator/product_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+package iterator
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIteratorProduct(t *testing.T) {
```
derailed: This only tests happy path...

pkg/metrics/metric/counter.go
```diff
@@ -70,12 +70,28 @@ func NewCounterVec(opts CounterOpts, labelNames []string) DeletableVec[Counter]
 	}
 }
 
+func NewCounterVecWithLabels(opts CounterOpts, labels Labels) DeletableVec[Counter] {
```
derailed: Seems we have a lot of repetitions here. Perhaps we could compose here aka WithLabels() to decorate the metric vs having to repeat for every metric types?

pkg/metrics/metric/gauge.go
```diff
@@ -96,13 +96,32 @@ func (g *gauge) SetToCurrentTime() {
...
 			opts:    Opts(opts),
 		},
 	}
+	gv.forEachLabelVector(func(vs []string) {
+		gv.WithLabelValues(vs...)
+	})
+	return gv
+}
+
+func NewGaugeVecWithLabels(opts GaugeOpts, labels Labels) DeletableVec[Gauge] {
```
derailed: nit: needs docs. Same for exported defs below.

pkg/iterator/product.go
tommyp1ckles: The idea was to use a flat vector as the shape of the underlying data - the VecIterator is just a two dimensional iterator over a one dimensional array.

pkg/metrics/metric/metric.go
```diff
@@ -196,3 +239,72 @@ func (b Opts) GetConfigName() string {
...
+type labelSet struct {
+	lbls Labels
+	m    map[string]map[string]struct{}
+}
+
+func (l labelSet) namesToValues() map[string]map[string]struct{} {
+	if l.m != nil {
+		return l.m
+	}
+	m := make(map[string]map[string]struct{})
```
derailed: Shouldn't l.m be assigned here?

pkg/metrics/metric/metric.go
```diff
@@ -196,3 +239,72 @@ func (b Opts) GetConfigName() string {
 	}
 	return b.ConfigName
 }
+
+type Label struct {
```
derailed, line 250: Might be a good idea to add tests for some of the calls below...

pkg/metrics/metrics.go
```diff
@@ -718,13 +719,18 @@ func NewLegacyMetrics() *LegacyMetrics {
...
+		EndpointRegenerationTotal: metric.NewCounterVecWithLabels(metric.CounterOpts{
 			ConfigName: Namespace + "_endpoint_regenerations_total",
 
 			Namespace: Namespace,
 			Name:      "endpoint_regenerations_total",
 			Help:      "Count of all endpoint regenerations that have completed, tagged by outcome",
-		}, []string{"outcome"}),
+		}, metric.Labels{
+			{
+				Name:   "outcome",
```
derailed: Perhaps we could define consts for common labels?

pkg/metrics/metrics.go
```diff
@@ -771,13 +777,19 @@ func NewLegacyMetrics() *LegacyMetrics {
...
 
 			Namespace: Namespace,
 			Name:      "policy_change_total",
 			Help:      "Number of policy changes by outcome",
-		}, []string{"outcome"}),
+		}, metric.Labels{
+			{
+				Name: "outcome",
+				// TODO: Use common label values for success/fail(ure) across all metrics.
+				Values: metric.NewValues("success", "failure"),
```
derailed: Perhaps define SuccessValues, TruthValues, UpdateValues, etc... to normalize  and reuse?

derailed (changes requested):
@tommyp1ckles Nice work! Hopefully we can dry up the impl a bit more...

pkg/metrics/metrics.go
tommyp1ckles: Success appears to already exist, so I'll use that.  "failure" is particular to this metric so I'll put it in a const but make a note that this is for legacy use.

In the future we should encourage the use of just "fail" and we should eventually follow up with cleaning up these labels and documenting the changes, but that's likely out of scope of this PR.

pkg/metrics/metric/metric.go
tommyp1ckles: good idea

pkg/metrics/metric/counter.go
tommyp1ckles: I tried a few approaches here, but the main thing I wanted was to avoid having to declare the labels twice, so any approach besides just passing the labels seemed to yield awkward results.

I did DRY up some of the common code however.

pkg/iterator/product.go
tommyp1ckles: However I've made the exported function just pass a array.

tommyp1ckles:
@derailed @squeed addressed/responded to review points, as well I moved the collections package into being a part of pkg/metrics/metric 😄

pkg/metrics/metric/collections/product_test.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func TestIteratorProductElements(t *testing.T) {
+	assert := assert.New(t)
+	p := CartesianProduct(
+		[]string{"true", "false"},
+		[]string{"0", "1"},
+		[]string{"foo"},
+	)
+	assert.Equal(4, p.Size(), "Should be 3 x 2 x 2 x 2 = 24 permutations")
```
derailed: Incorrect assert msg?

pkg/metrics/metric/collections/product_test.go
```diff
@@ -0,0 +1,61 @@
...
+	)
+	assert.Equal(0, p.Size())
+
+	p = CartesianProduct[string]()
+	assert.Equal(0, p.Size())
+
+	assert.Equal(0, CartesianProduct([]string{}, []string{}, []string{}).Size())
+	assert.Equal(0, CartesianProduct[int]().Size())
+	assert.Equal(1, CartesianProduct([]string{""}, []string{""}, []string{""}).Size())
+	CartesianProduct[interface{}](nil, nil) // Test some weird cases.
```
derailed, line 62: Does this actually test anything?

pkg/metrics/metric/collections/product.go
```diff
@@ -0,0 +1,76 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package collections
+
+func CartesianProduct[T any](vs ...[]T) VecIterator[T] {
```
derailed: Not sure why we are using the interface here vs the actual type?

nit: Docs?

pkg/metrics/metric/collections/product.go
```diff
@@ -0,0 +1,76 @@
...
+	for i := 1; i < len(vs); i++ {
+		rhs := vs[i]
+		lhs = product(lhs, rhs)
+	}
+
+	return lhs
+}
+
+// So if you pass a slice view, you can have this mutate that.
+func product[T any](a VecIterator[T], b VecIterator[T]) VecIterator[T] {
```
derailed: nit: Dry a, b xxx

pkg/metrics/metric/counter.go
```diff
@@ -60,7 +60,7 @@ func (c *counter) Add(val float64) {
 	}
 }
 
-func NewCounterVec(opts CounterOpts, labelNames []string) DeletableVec[Counter] {
```
derailed, line 63: nit: please doc exported calls here and below...

pkg/metrics/metric/counter.go
```diff
@@ -70,12 +70,53 @@ func NewCounterVec(opts CounterOpts, labelNames []string) DeletableVec[Counter]
 	}
 }
 
+// NewCounterVec creates a new DeletableVec[Counter] based on the provided CounterOpts and
+// partitioned by the given label names.
+func NewCounterVec(opts CounterOpts, labelNames []string) DeletableVec[Counter] {
```
derailed: Don't think its necessary to wrap the unexported constructor. We should just return the instance of the type.
Returning the interface here says that's all this counter will ever be.

pkg/metrics/metric/collections/iterator.go
```diff
@@ -0,0 +1,83 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package collections
+
+// VecIterator is an iterator that returns a vector of elements of type T for
+// each iteration.
+type VecIterator[T any] interface {
```
derailed: Is this used anywhere else? Seems we could do away with it??

pkg/metrics/metric/metric.go
```diff
@@ -196,3 +239,85 @@ func (b Opts) GetConfigName() string {
...
+		return l.m
+	}
+	l.m = make(map[string]map[string]struct{})
+	for _, label := range l.lbls {
+		l.m[label.Name] = label.Values
+	}
+	return l.m
+}
+
+func (l labelSet) checkLabels(labels prometheus.Labels) error {
```
derailed: Why are we changing semantics here? ie value vs ptr receiver??

pkg/metrics/metric/metric.go
```diff
@@ -196,3 +239,85 @@ func (b Opts) GetConfigName() string {
...
+				return fmt.Errorf("value %s not allowed for label %s (should be one of %v)", value, name, lvs)
+			}
+		} else {
+			return fmt.Errorf("invalid label name: %s", name)
+		}
+	}
+	return nil
+}
+
+func (l labelSet) checkLabelValues(lvs []string) error {
```
derailed: Likewise mixing semantics

pkg/metrics/metric/metric.go
```diff
@@ -196,3 +239,85 @@ func (b Opts) GetConfigName() string {
...
+
+func NewValues(vs ...string) Values {
+	vals := Values{}
+	for _, v := range vs {
+		vals[v] = struct{}{}
+	}
+	return vals
+}
+
+type Labels []Label
```
derailed, line 270: nit: docs

pkg/metrics/metric/metric.go
```diff
@@ -196,3 +239,85 @@ func (b Opts) GetConfigName() string {
...
 	return b.ConfigName
 }
+
+type Label struct {
+	Name string
+	// If defined, only these values are allowed.
+	Values Values
+}
+
+type Values map[string]struct{}
```
derailed, line 257: nit: docs

pkg/metrics/metric/metric.go
```diff
@@ -20,6 +29,40 @@ type WithMetadata interface {
...
+	}
+	collections.CartesianProduct(labelValues...).ForEach(fn)
+}
+
+// checkLabelValues checks that the provided label values are within the range
+// of provided label values, if labels where defined using the Labels type.
+// Violations are logged as errors for detection, but metrics should still
+// be collected as is.
+func (b *metric) checkLabelValues(lvs ...string) {
+	if b.labels != nil {
```
derailed: nit: if we don't have labels might as well return!

pkg/metrics/metric/metric.go
```diff
@@ -20,6 +29,40 @@ type WithMetadata interface {
...
+		if err := b.labels.checkLabelValues(lvs); err != nil {
+			logger.WithError(err).Error("metric label constraints violated")
+		}
+	}
+}
+
+func (b *metric) checkLabels(labels prometheus.Labels) {
+	if b.labels != nil {
```
derailed: nit: same as above^

pkg/metrics/metric/metric.go
```diff
@@ -196,3 +239,85 @@ func (b Opts) GetConfigName() string {
 	}
 	return b.ConfigName
 }
+
+type Label struct {
```
derailed, line 250: nit: docs

pkg/metrics/metric/metric.go
```diff
@@ -196,3 +239,85 @@ func (b Opts) GetConfigName() string {
...
+	Name string
+	// If defined, only these values are allowed.
+	Values Values
+}
+
+type Values map[string]struct{}
+
+func NewValues(vs ...string) Values {
```
derailed, line 260: nit: docs

pkg/metrics/metric/metric.go
```diff
@@ -196,3 +239,85 @@ func (b Opts) GetConfigName() string {
...
+	}
+	return nil
+}
+
+func (l labelSet) checkLabelValues(lvs []string) error {
+	if len(l.lbls) != len(lvs) {
+		return fmt.Errorf("invalid labels")
+	}
+	for i, label := range l.lbls {
+		if label.Values != nil {
```
derailed: nit: Don't think we need to check for nil map here?

pkg/metrics/metrics.go
```diff
@@ -771,13 +781,18 @@ func NewLegacyMetrics() *LegacyMetrics {
...
+		PolicyChangeTotal: metric.NewCounterVecWithLabels(metric.CounterOpts{
 			ConfigName: Namespace + "_policy_change_total",
 
 			Namespace: Namespace,
 			Name:      "policy_change_total",
 			Help:      "Number of policy changes by outcome",
-		}, []string{"outcome"}),
+		}, metric.Labels{
+			{
+				Name:   "outcome",
```
derailed: nit: use const?

pkg/metrics/metrics.go
```diff
@@ -965,19 +994,51 @@ func NewLegacyMetrics() *LegacyMetrics {
...
+				Name:   LabelScope,
+				Values: metric.NewValues("CiliumNetworkPolicy", "CiliumClusterwideNetworkPolicy", "NetworkPolicy"),
+			},
+			{
+				Name:   LabelAction,
+				Values: metric.NewValues("update", "delete"),
+			},
+			{
+				Name:   "valid",
+				Values: metric.NewValues("true", "false"),
```
derailed: nit: could dry this up ie metric.TruthValues metric.ActionValues??

derailed (changes requested):
@tommyp1ckles Tx for the updates! a few more picks...

pkg/metrics/metric/collections/product_test.go
tommyp1ckles: Just testing this doesn't crash when you pass nil values.

pkg/metrics/metric/collections/product.go
tommyp1ckles: The returned type is a contiguous slice, hence the VecIterator type allows you to access/iterate the vector elements.

pkg/metrics/metric/metric.go
tommyp1ckles: wanted to avoid crashing in case someone doesn't set this value.

pkg/metrics/metric/collections/product.go
tommyp1ckles: Since we probably don't actually need to worry about memory access patterns since this is no longer a general library, I've simplified this down to just take and return a 2d vector [][]T.

pkg/metrics/metric/gauge.go
```diff
@@ -95,14 +95,51 @@ func (g *gauge) SetToCurrentTime() {
...
+// ranges can be pre-initialized to zero upon init.
+//
+// This should only be used when all label values are known at init, otherwise use of the
+// metric vector with uninitialized labels will result in warnings.
+//
+// Note: Disabled metrics will not have their label values initialized.
+//
+// For example:
+//
+//	NewGaugeVec(GaugeOpts{
```
derailed: nit: NewGaugeVecWithLabels

pkg/metrics/metric/counter.go
```diff
@@ -70,12 +72,47 @@ func NewCounterVec(opts CounterOpts, labelNames []string) DeletableVec[Counter]
...
+//		{Name: "bar", Values: NewValues("a", "b")},
+//	})
+//
+// Will initialize the following metrics to:
+//
+//	cilium_subsystem_cilium_test{foo="0", bar="a"} 0
+//	cilium_subsystem_cilium_test{foo="0", bar="b"} 0
+//	cilium_subsystem_cilium_test{foo="1", bar="a"} 0
+//	cilium_subsystem_cilium_test{foo="1", bar="b"} 0
+func NewCounterVecWithLabels(opts CounterOpts, labels Labels) DeletableVec[Counter] {
```
derailed: Couldn't we return the instance vs the interface?

pkg/metrics/metric/gauge.go
```diff
@@ -95,14 +95,51 @@ func (g *gauge) SetToCurrentTime() {
...
+//		{Name: "bar", Values: NewValues("a", "b")},
+//	})
+//
+// Will initialize the following metrics to:
+//
+//	cilium_subsystem_cilium_test{foo="0", bar="a"} 0
+//	cilium_subsystem_cilium_test{foo="0", bar="b"} 0
+//	cilium_subsystem_cilium_test{foo="1", bar="a"} 0
+//	cilium_subsystem_cilium_test{foo="1", bar="b"} 0
+func NewGaugeVecWithLabels(opts GaugeOpts, labels Labels) DeletableVec[Gauge] {
```
derailed: Could we return the instance here vs the interface?

pkg/metrics/metric/histogram.go
```diff
@@ -81,12 +83,28 @@ func NewHistogramVec(opts HistogramOpts, labelNames []string) Vec[Observer] {
...
+// NewHistogramVec creates a new Vec[Observer] based on the provided CounterOpts and
+// partitioned by the given labels.
+// This will also initialize the labels with the provided values so that metrics with known label value
+// ranges can be pre-initialized to zero upon init.
+//
+// This should only be used when all label values are known at init, otherwise use of the
+// metric vector with uninitialized labels will result in warnings.
+//
+// Note: Disabled metrics will not have their label values initialized.
+func NewHistogramVecWithLabels(opts HistogramOpts, labels Labels) Vec[Observer] {
```
derailed: Return the instance vs interface?

pkg/metrics/metric/metric.go
derailed: I think you can index into a nil map

derailed (changes requested):
@tommyp1ckles LG!

pkg/metrics/metric/counter.go
tommyp1ckles: Looking at this more, I'm a bit more apprehensive to make this change since that will make the rest of the package inconsistent, I'd like to understand the initial decision to return the generic interface types (cc: @dylandreimerink )

Avoiding returning interfaces types is considered good practice, but in this case I worry we might be making the code harder to understand.  The whole intent of the metric library is provide explicitly everything that metrics provide.

squeed (approved):

pkg/metrics/metric/counter.go
dylandreimerink: > I'd like to understand the initial decision to return the generic interface types 

I modeled the API after the prometheus API itself which does the exact same thing. My goal was to keep things mostly the same, and just append the features we needed (metadata + enable/disable capabilities).

tommyp1ckles: Interestingly, [NewCounterVec](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#NewCounterVec) still returns a struct compared to everything else (including NewCounter, ... etc).

tommyp1ckles: Going to switch it to *gaugeVec, *counterVec, *histogramVec, since that actually seems consistent with the prometheus library.

pkg/metrics/metric/gauge.go
```diff
@@ -95,14 +95,51 @@ func (g *gauge) SetToCurrentTime() {
 	}
 }
 
-func NewGaugeVec(opts GaugeOpts, labelNames []string) DeletableVec[Gauge] {
-	return &gaugeVec{
+// NewGaugeVec creates a new DeletableVec[Gauge] based on the provided GaugeOpts and
+// partitioned by the given label names.
+func NewGaugeVec(opts GaugeOpts, labelNames []string) *gaugeVec {
+	gv := &gaugeVec{
```
derailed, line 101: nit: Don't need to declare var here.

derailed (approved):
@tommyp1ckles Tx for the updates!

Issue #27804, opened 2023-08-29
Document troubleshooting steps / guideline for policymap pressure 


christarazi:
See https://github.com/cilium/cilium/pull/27367#issuecomment-1696584421.

PR #27838, opened 2023-08-31
api, cmd, policy: Show rule labels in `policy selectors` output

christarazi:
- policy: Plumb rule labels through selector cache
- api, cmd, policy: Show rule labels in `policy selectors` output

---

Commits for convenience of review:

> policy: Plumb rule labels through selector cache
>
> In order to make investigating the output of `cilium policy selectors`
> easier to understand, plumb the rule labels from KNP, CNP, and CCNPs.
> This makes it so that the user can very easily see which policy the
> selector came from.
>
> Benchmark results before and after given that selectorcache can be a hot
> path in the policy engine:
>
> ```
> $ go test -v ./pkg/policy -run '^$' -bench 'BenchmarkRegenerateL3EgressPolicyRules' -test.benchtime 100x -test.benchmem -test.count 10
> ...
> $ benchstat old.txt new.txt
> name                              old time/op    new time/op    delta
> RegenerateL3EgressPolicyRules-16    8.13ms ±13%    8.01ms ±11%     ~     (p=0.912 n=10+10)
>
> name                              old alloc/op   new alloc/op   delta
> RegenerateL3EgressPolicyRules-16    1.75MB ± 0%    1.94MB ± 0%  +10.78%  (p=0.000 n=10+9)
>
> name                              old allocs/op  new allocs/op  delta
> RegenerateL3EgressPolicyRules-16     24.1k ± 0%     25.1k ± 0%   +4.15%  (p=0.000 n=10+10)
> ```

> api, cmd, policy: Show rule labels in `policy selectors` output
>
> Use the functionality implemented in previous commits to output the rule
> labels in the selector output. This is useful for understanding which
> policy the selector comes from, which makes debugging issues much
> easier.
>
> Example output:
>
> ```
> root@kind-worker:/home/cilium# cilium policy selectors
> SELECTOR                                                                                                                                                            LABELS                          USERS   IDENTITIES
> &LabelSelector{MatchLabels:map[string]string{k8s.io.kubernetes.pod.namespace: kube-system,k8s.k8s-app: kube-dns,},MatchExpressions:[]LabelSelectorRequirement{},}   default/tofqdn-dns-visibility   1       16500
> &LabelSelector{MatchLabels:map[string]string{reserved.none: ,},MatchExpressions:[]LabelSelectorRequirement{},}                                                      default/tofqdn-dns-visibility   1
> MatchName: , MatchPattern: *                                                                                                                                        default/tofqdn-dns-visibility   1       16777217
>                                                                                                                                                                                                             16777218
>                                                                                                                                                                                                             16777219
> ```

```release-note
Improve the usability of the `cilium policy selectors` command by including the policy name and namespace in order to easily understand which selector comes from what policy
```

---

Related: https://github.com/cilium/cilium/issues/27804

### DIFF
--- a/cilium/cmd/policy_selectors.go
+++ b/cilium/cmd/policy_selectors.go
@@ -12,7 +12,11 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/cilium/cilium/pkg/command"
+	k8sconst "github.com/cilium/cilium/pkg/k8s/apis/cilium.io"
+	"github.com/cilium/cilium/pkg/labels"
 )
+
+var verbosePolicySelectors bool
 
 // policyCacheGetCmd represents the policy selectors command
 var policyCacheGetCmd = &cobra.Command{
@@ -31,10 +35,22 @@ var policyCacheGetCmd = &cobra.Command{
 			sort.Slice(resp, func(i, j int) bool {
 				return resp[i].Selector < resp[j].Selector
 			})
-			fmt.Fprintf(w, "SELECTOR\tUSERS\tIDENTITIES\n")
+			fmt.Fprintf(w, "SELECTOR\tLABELS\tUSERS\tIDENTITIES\n")
+
 			for _, mapping := range resp {
+				lbls := constructLabelsArrayFromAPIType(mapping.Labels)
+
 				first := true
 				fmt.Fprintf(w, "%s", mapping.Selector)
+				if verbosePolicySelectors {
+					var lstr string
+					if len(lbls) != 0 {
+						lstr = lbls.Sort().String()
+					}
+					fmt.Fprintf(w, "\t%s", lstr)
+				} else {
+					fmt.Fprintf(w, "\t%s", getNameAndNamespaceFromLabels(lbls))
+				}
 				fmt.Fprintf(w, "\t%d", mapping.Users)
 				if len(mapping.Identities) == 0 {
 					fmt.Fprintf(w, "\t\n")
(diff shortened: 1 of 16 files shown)


pkg/policy/selectorcache.go
```diff
@@ -33,6 +33,11 @@ type CachedSelector interface {
 	// be modified, as it is shared among multiple users.
 	GetSelections() []identity.NumericIdentity
 
+	// GetMetadataLabels returns metadata labels for additional context
+	// surrounding the selector. These are typically the labels associated with
+	// Cilium rules.
+	GetMetadataLabels() labels.LabelArray
```
tommyp1ckles, line 39: Thought: I notice that the LabelArray type is being printed as is, it might be nice to implement `fmt.Stringer` on this so that we can provide an easier to read representation of the array.

cilium/cmd/policy_selectors.go
```diff
@@ -31,10 +35,22 @@ var policyCacheGetCmd = &cobra.Command{
 			sort.Slice(resp, func(i, j int) bool {
 				return resp[i].Selector < resp[j].Selector
 			})
-			fmt.Fprintf(w, "SELECTOR\tUSERS\tIDENTITIES\n")
+			fmt.Fprintf(w, "SELECTOR\tLABELS\tUSERS\tIDENTITIES\n")
```
tommyp1ckles, line 38: Do we want to print this header if `verbosePolicySelectors` is false?

christarazi: I think it's still useful to have the header if the full labels (verbose) are not shown. In fact, I think the output would be very hard to parse without the headers describing what is what.

pkg/policy/selectorcache.go
christarazi: It's hard to predict what the desired representation should be because it's an array. There are other `String()` methods on the related label types so it's not obvious if we need yet another `Stringer`.

tommyp1ckles (approved):
Looks good, this is a really useful change 🚀

nathanjsweet (approved):

PR #27840, opened 2023-08-31
Mutual Auth: Perform policy check on mutual auth mTLS handshake

meyskens:
This adds a policy allowed check on the receiving end of the mTLS handshake.
The server will check in the local policy map if the two identities
have their auth type set to SPIRE.
If this isn't the case it will deny the handshake.

```release-note
Perform policy check on receiving side of Mutual Auth mTLS handshake
```


meyskens:
This PR depends on https://github.com/cilium/cilium/pull/27682 and will be re-based and marked as ready for review once that one is merged

meyskens:
Seems some integration tests are not yet up to date, fixing that

PR #27841, opened 2023-08-31
Change routing-mode and tunnel-protocol based on .Values.tunnel and .Values.routingMode

### LINKED ISSUE #27756: helm: apparent incompatible values when using routingMode with tunnel unset

### What happened?

We run cilium chained to aws-vpc-cni in EKS for clustermesh and network policy enforcement.

Going through the upgrade docs I see that `tunnel` is [deprecated](https://docs.cilium.io/en/stable/operations/upgrade/#deprecated-options) in favor of `routing-mode` and `tunnel-protocol`. In my helm values, I removed `tunnel` and set `routingMode: native`. This generated a configmap with the following section:
```yaml
  # Encapsulation mode for communication between nodes
  # Possible values:
  #   - disabled
  #   - vxlan (default)
  #   - geneve
  # Default case
  routing-mode: "tunnel"
  tunnel-protocol: "vxlan"
  routing-mode: "native"
```

Upon application you end up with configmap with the following values:
```yaml
  tunnel-protocol: "vxlan"
  routing-mode: "native"
```

I'm not sure this is a valid configuration. We manage our application using fluxcd and the multiple definitions in the configmap for `routing-mode` cause it to fail with an error. The short term work around is to keep `tunnel: disabled` and not set `routingMode` but I don't think the helm chart behavior is correct for this.

### Cilium Version

v1.14.1

### Kernel Version

Linux ip-10-16-106-43.us-west-2.compute.internal 5.10.186-179.751.amzn2.x86_64 #1 SMP Tue Aug 1 20:51:38 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

### Kubernetes Version

v1.25.12

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

macmiranda:
Fixes https://github.com/cilium/cilium/issues/27756

### DIFF
--- a/install/kubernetes/cilium/templates/cilium-configmap.yaml
+++ b/install/kubernetes/cilium/templates/cilium-configmap.yaml
@@ -440,7 +440,7 @@ data:
   routing-mode: "tunnel"
   tunnel-protocol: "vxlan"
 {{- else if .Values.routingMode }}
-  routing-mode: "{{ .Values.routingMode }}"
+  routing-mode: {{ .Values.routingMode | quote }}
 {{- else }}
   {{- if eq .Values.tunnel "disabled" }}
   routing-mode: "native"
@@ -458,15 +458,15 @@ data:
 {{- end }}
 
 {{- if .Values.tunnelProtocol }}
-  tunnel-protocol: "{{ .Values.tunnelProtocol }}"
+  tunnel-protocol: {{ .Values.tunnelProtocol | quote }}
 {{- end }}
 
 {{- if .Values.tunnelPort }}
-  tunnel-port: "{{ .Values.tunnelPort }}"
+  tunnel-port: {{ .Values.tunnelPort | quote }}
 {{- end }}
 
 {{- if .Values.MTU }}
-  mtu: "{{ .Values.MTU }}"
+  mtu: {{ .Values.MTU | quote }}
 {{- end }}
 
 {{- if .Values.eni.enabled }}
--- a/install/kubernetes/cilium/templates/cilium-configmap.yaml
+++ b/install/kubernetes/cilium/templates/cilium-configmap.yaml
@@ -439,24 +439,22 @@ data:
 {{- else if .Values.aksbyocni.enabled }}
   routing-mode: "tunnel"
   tunnel-protocol: "vxlan"
-{{- end }}
-
-{{- if eq .Values.tunnel "disabled" }}
+{{- else if .Values.routingMode }}
+  routing-mode: "{{ .Values.routingMode }}"
+{{- else }}
+  {{- if eq .Values.tunnel "disabled" }}
   routing-mode: "native"
-{{- else if eq .Values.tunnel "vxlan" }}
+  {{- else if eq .Values.tunnel "vxlan" }}
   routing-mode: "tunnel"
   tunnel-protocol: "vxlan"
-{{- else if eq .Values.tunnel "geneve" }}
+  {{- else if eq .Values.tunnel "geneve" }}
   routing-mode: "tunnel"
   tunnel-protocol: "geneve"
-{{- else if not (or .Values.gke.enabled .Values.aksbyocni.enabled) }}
+  {{- else }}
   # Default case
   routing-mode: "tunnel"
   tunnel-protocol: "vxlan"
-{{- end }}
-
-{{- if .Values.routingMode }}
-  routing-mode: "{{ .Values.routingMode }}"
+  {{- end }}
 {{- end }}
 
 {{- if .Values.tunnelProtocol }}


gandro (changes requested):
Thanks for the PR! Overall looks good, and most combinations work, but I've found that specifying *both* a routing mode and a tunnel protocol did not work as expected, e.g. `--set tunnel=geneve --set routingMode=tunnel` did not set the `tunnel-protocol` correctly, which I think it should also do.

Also, the commit subject is overly long, thus causing `checkpatch.pl` to fail in CI. Could you shorten it a bit and instead put the details in the commit message description? Thanks

gandro:
Marking this for `needs-backport` since the tunnel flag will be removed in v1.15 according to the upgrade docs

macmiranda:
The way I see it, `tunnel` should only be taken into consideration if `routingMode` isn't set, wouldn't you agree?

What if someone sets `tunnel` and `tunnelProtocol` to different things? That would unnecessarily complicate things. 

Since `tunnel` is deprecated, it should be ignored when (the new) `routingMode` and `tunnelProtocol` are set.

The only case in which `tunnel` should be honored is if it were already set and none of the new values were explicitly given. (This would only happen if you upgraded Cilium without reading the release notes or ignored the deprecation notice).

macmiranda:
> Also, the commit subject is overly long, thus causing checkpatch.pl to fail in CI. Could you shorten it a bit and instead put the details in the commit message description? Thanks

Will fix it tomorrow.

gandro (approved):
> Since tunnel is deprecated, it should be ignored when (the new) routingMode and tunnelProtocol are set.

Yeah I think you are right, I missed the fact that `tunnel` has been replaced by `tunnelProtocol` in Helm too. 

I wonder if we should introduce some additional validations that disallow `tunnelProtocol` without a routing mode, or `tunnel` with `routingMode`. 

At the moment, if I specify `--set tunnelProtocol=geneve` without `routingMode`, it generates the following for some reason:

```
  # Encapsulation mode for communication between nodes
  # Possible values:
  #   - disabled
  #   - vxlan (default)
  #   - geneve
  # Default case
  routing-mode: "tunnel"
  tunnel-protocol: "vxlan"
  tunnel-protocol: "geneve"
```

gandro (approved):
Awesome, thanks for the follow-up changes!

gandro:
Interesting, so with the new validation, some of our CI is now failing. For example:

https://github.com/cilium/cilium/actions/runs/6084377949/job/16513347947?pr=27841

```
Error: Unable to install Cilium: execution error at (cilium/templates/cilium-configmap.yaml:82:6): routingMode and tunnel are mutually exclusive. Please set only one of them.
```

We might have to change the CI workflows before we can merge this PR.

macmiranda:
Hi @gandro,

Yes, that one was easy to fix by removing `--helm-set=tunnel=disabled` from the install command.

https://github.com/cilium/cilium/actions/runs/6084377962/job/16513356982 is a bit trickier since I can't find where the `tunnel` value is coming from (assuming there is nothing wrong with the validation).

macmiranda:
I'd say it's a bug in `cilium-cli`:

```
❯ cilium install --dry-run-helm-values --helm-set=routingMode=native
cluster:
  name: colima-docker
operator:
  replicas: 1
routingMode: native
serviceAccounts:
  cilium:
    name: cilium
  operator:
    name: cilium-operator
tunnel: vxlan
```
Note I'm passing `routingMode` but it still generates `tunnel` in the output. If I force `--datapath-mode=native`, it works as expected.

Do you want me to add the workaround to the worflows?

macmiranda:
[Here](https://github.com/cilium/cilium-cli/blob/19a57012227e1e842db4a124c937a870c1c2e536/install/autodetect.go#L83-L84) the `datapath-mode` is set to `tunnel` because we're running on Kind.

And [this](https://github.com/cilium/cilium-cli/blob/19a57012227e1e842db4a124c937a870c1c2e536/install/helm.go#L198-L199) is still using the old syntax. It should set `tunnelProtocol` instead.

macmiranda:
**Up for discussion**: Should we actually force `routingMode=native` in the test workflows if `cilium-cli` data-path autodetect feature uses `tunnel` under similar circumstances?

gandro:
> **Up for discussion**: Should we actually force `routingMode=native` in the test workflows if `cilium-cli` data-path autodetect feature uses `tunnel` under similar circumstances?

Thanks for looking into it. I think we want to fix the CLI first. Especially if we backport this to v1.14, it might cause frictions to CLI users.

Aslo note that commit https://github.com/cilium/cilium/pull/27841/commits/30f3cf67808e189f25a15414e12d7abcd9d9c8b8 is likely not working because we don't run CI workflow from forks for security reasons.

Maybe a way forward would be to merge your PR _without_ the validation, backport it to v1.14, In separate PRs, we can then fix CLI and CI for `main` branches, and then introduce the validation once all the tooling and CI has caught up. What do you think?

macmiranda:
> Maybe a way forward would be to merge your PR _without_ the validation, backport it to v1.14, In separate PRs, we can then fix CLI and CI for `main` branches, and then introduce the validation once all the tooling and CI has caught up. What do you think?

Seems reasonable, though I have no say on most of what's going to happen outside this PR. I'd gladly revert it to a CI-passing state so that it can be merged. Everything else would depend more on Cilium committers than on me. If you're ok with it, just let me know when I can create the 2nd PR to add the validation.

Shall we proceed?

gandro:
> 

Sounds like a plan! Yeah I'm happy to drive the CI fixes forward. If you want, you could of course also attempt a PR for the CLI against `cilium/cilium-cli` to remove it's use of the `tunnel` flag on v1.14 and newer, but I can also take that over.

macmiranda:
@gandro anything else?

gandro:
> @gandro anything else?

No, this looks good to me know! Thanks again, let's wait for the other CODEOWNERs to review.

nathanjsweet (approved):

install/kubernetes/cilium/templates/cilium-configmap.yaml
```diff
@@ -439,36 +439,34 @@ data:
...
   routing-mode: "native"
-{{- else if eq .Values.tunnel "vxlan" }}
+  {{- else if eq .Values.tunnel "vxlan" }}
   routing-mode: "tunnel"
   tunnel-protocol: "vxlan"
-{{- else if eq .Values.tunnel "geneve" }}
+  {{- else if eq .Values.tunnel "geneve" }}
   routing-mode: "tunnel"
   tunnel-protocol: "geneve"
-{{- else if not (or .Values.gke.enabled .Values.aksbyocni.enabled) }}
```
michi-covalent, line 452: i'm a bit worried about backporting this change to v1.14 branch. it changes the routing mode between patch versions v1.14.2 and v1.14.3 as @giorio94 pointed out in https://github.com/cilium/cilium-cli/issues/2070#issuecomment-1791113289. trying to figure out how to document this change in the upgrade guide 💭

giorio94: My gut feeling is that we need to fix this logic, as with the current patch it seems that it is no longer possible to override the routing mode when `gke.enabled` is true.

gandro: IMO, enabling tunnel mode with `gke.enabled` was never supported. The docs state pretty clearly that `gke.enabled` will run direct routing mode: https://docs.cilium.io/en/stable/network/concepts/routing/#id6 and that BYOCNI runs with encapsulation: [AKS tab](https://docs.cilium.io/en/stable/gettingstarted/k8s-install-default/#install-cilium)

michi-covalent: [the external workload page](https://docs.cilium.io/en/stable/network/external-workloads/#enable-support-for-external-workloads) says:

> So far this functionality is only tested with the vxlan tunneling datapath mode (default for most installations).

and the external workload test is failing at `Verify cluster DNS on VM` step in cilium-cli repo when the routing mode is set to `native` like this https://github.com/cilium/cilium-cli/actions/runs/6911454905/job/18806005510?pr=2079

kind of tempted to just set the routing mode to `tunnel` using `--datapath-mode` flag like we do in cilium/cilium repo. cc @jrajahalme i guess external workload is not supported with the native routing mode? 💭💭💭

giorio94: > kind of tempted to just set the routing mode to tunnel using --datapath-mode flag like we do in cilium/cilium repo

Makes sense to me

michi-covalent: 🚀🙏 https://github.com/cilium/cilium-cli/pull/2079

PR #27846, opened 2023-08-31
chore: Add deezer as cilium user

zwindler:
Hello,

[deezer](https://www.deezer.com/fr/) is now using Cilium as it's CNI for all Kubernetes clusters, can we be added to the list?

Regards
Denis


xmulligan:
Thanks for the addition @zwindler !

xmulligan (approved):

PR #27849, opened 2023-08-31
bpf: ct: reuse get_ct_map*() in get_cluster_ct_map*()

julianwiedmann:
Don't open-code the map selection, but use the standard helper. This avoids that the two code paths drift apart.


ldelossa (approved):
:+1:

PR #27851, opened 2023-08-31
bpf: egressgw: pass IPv4 tuple to egress_gw_request_needs_redirect 

jibi:
pass the IPv4 tuple to egress_gw_request_needs_redirect instead of the
full IP header as only the source and destination addresses are needed
to match an egress gateway policy

### DIFF
--- a/bpf/lib/egress_gateway.h
+++ b/bpf/lib/egress_gateway.h
@@ -36,7 +36,7 @@ struct egress_gw_policy_entry *lookup_ip4_egress_gw_policy(__be32 saddr, __be32 
 #endif /* ENABLE_EGRESS_GATEWAY */
 
 static __always_inline
-bool egress_gw_request_needs_redirect(struct iphdr *ip4 __maybe_unused,
+bool egress_gw_request_needs_redirect(struct ipv4_ct_tuple *rtuple __maybe_unused,
 				      int ct_status __maybe_unused,
 				      __u32 *tunnel_endpoint __maybe_unused)
 {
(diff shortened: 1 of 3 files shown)


bpf/lib/egress_gateway.h
```diff
@@ -52,7 +52,8 @@ bool egress_gw_request_needs_redirect(struct iphdr *ip4 __maybe_unused,
 	if (ct_status == CT_REPLY || ct_status == CT_RELATED)
 		return false;
 
-	egress_gw_policy = lookup_ip4_egress_gw_policy(ip4->saddr, ip4->daddr);
+	egress_gw_policy = lookup_ip4_egress_gw_policy(ipv4_ct_reverse_tuple_saddr(rtuple),
+						       ipv4_ct_reverse_tuple_daddr(rtuple));
```
julianwiedmann, line 56: Ah nice, with the preceeding checks we can even be certain that the tuple is in forward-direction layout.

julianwiedmann (approved):
looks good, thanks!

bpf/lib/egress_gateway.h
jibi: I was thinking about making these helpers work with both directions and dynamically return the correct or swapped address based on the `TUPLE_F_{IN,OUT}` value of `flags`, but as I was not sure about complexity implications I decided to go with this simpler thing for now that just helps with readability (and experiment with that in a separate pr)

julianwiedmann: Yep, experimenting sounds good :+1:.

We've started using specific terminology when reviewing CT PRs recently ("NAT layout" for tuples that match the packet headers, and "CT layout" for tuples that have flipped ports). Maybe it's time to bake this into actual helper names etc..

PR #27854, opened 2023-08-31
Correlate flows with CiliumNetworkPolicies

### LINKED ISSUE #26438: CFP: Correlate policy verdicts to network policies
## Cilium Feature Proposal

**Desired Feature**

Policy verdict correlation

When a policy verdict event is produced for a flow, correlate the set of network policies that result in that verdict.

Example output for a flow:

```
{
  "flow": {
    "time": "2023-06-22T17:23:02.204172941Z",
    "verdict": "FORWARDED",
    "ethernet": {
      "source": "9e:b2:fc:ca:4b:7a",
      "destination": "a6:35:14:0d:02:64"
    },
    "IP": {
      "source": "10.128.0.22",
      "destination": "10.48.1.17",
      "ipVersion": "IPv4"
    },
    "l4": {
      "TCP": {
        "source_port": 80,
        "destination_port": 55555,
        "flags": {
          "SYN": true,
          "ACK": true
        }
      }
    },
    "source": {
      "identity": 2,
      "labels": [
        "reserved:world"
      ]
    },
    "destination": {
      "ID": 3798,
      "identity": 31051,
      "namespace": "default",
      "labels": [
        "k8s:io.cilium.k8s.policy.cluster=default",
        "k8s:io.kubernetes.pod.namespace=default"
      ],
      "pod_name": "client-7494b5866d-w69gw",
      "workloads": [
        {
          "name": "client",
          "kind": "Deployment"
        }
      ]
    },
    "Type": "L3_L4",
    "node_name": "gke-cluster-default-pool-025d71ba-q92f",
    "event_type": {
      "type": 5
    },
    "source_service": {
      "name": "service",
      "namespace": "default"
    },
    "policy_match_type": 4,
    "policies":[                                 #<--- new field
      {
         "kind":"NetworkPolicy",
         "name":"allow-all",
         "namespace":"default"
      },
      {
         "kind":"CiliumClusterwideNetworkPolicy",
         "name":"allow-all",
      },
    ], 
    "is_reply": false,
    "Summary": "TCP Flags: SYN, ACK"
  },
  "node_name": "gke-cluster-default-pool-025d71ba-q92f",
  "time": "2023-06-22T17:23:02.204172941Z"
}
```

**Proposed Solution**

Proposal is outlined in the following Google Doc:

https://docs.google.com/document/d/1puY1kAwZ7QK4Kfpn3kDjo3tCh5OzgOIkxL3rQzosVFI/edit#

chancez:
Using GetRealizedPolicyRuleLabelsForKey, lookup policy labels for a flow using the: direction, endpointIP, remoteIdentity, protocool and destination port. After getting the labels, reconstruct the policy name and namespace using the well-known policy labels.

Relates to https://github.com/cilium/hubble/issues/1100
Closes https://github.com/cilium/cilium/issues/26438.

Reviewers note: The protobuf field numbers may look arbitrary, however, this is a port of a feature we maintained internally, so for compatibility, it would be ideal to reuse the existing protobuf field names and numbers.

```release-note
Correlate flows with CiliumNetworkPolicies
```


pkg/hubble/parser/seven/parser.go
```diff
@@ -150,6 +151,11 @@ func (p *Parser) Decode(r *accesslog.LogRecord, decoded *flowpb.Flow) error {
 	decoded.TraceContext = p.getTraceContext(r)
 	decoded.Summary = p.getSummary(r, decoded)
 
+	err = correlation.CorrelatePolicy(p.endpointGetter, decoded)
+	if err != nil {
+		p.log.WithError(err).Error("error correlating policy with flow")
+	}
+
```
michi-covalent: maybe:
```suggestion
	if p.endpointGetter != nil {
		err = correlation.CorrelatePolicy(p.endpointGetter, decoded)
		if err != nil {
			p.log.WithError(err).Error("error correlating policy with flow")
		}
	}
```
to be consistent with pkg/hubble/parser/threefour/parser.go?

also a bit worried about logging an error here. when could `CorrelatePolicy` fail, and what can users do about it?

michi-covalent (approved):
lgtm for the files owned by hubble team, just some minor suggestions / questions.

i'll add do-not-merge label. i'd like to get this reviewed by @gandro 🙏

pkg/hubble/parser/seven/parser.go
chancez: Ah, the function actually doesn't have any cases where it returns any errors, so we could remove it. That being said the function logs errors in cases where it cannot correlate to a policy. For example: if it cannot lookup the endpoint, or if it cannot find a policy. I pushed a new commit to change the return type. Let me know what you think.

pkg/hubble/parser/seven/parser.go
```diff
@@ -150,6 +151,11 @@ func (p *Parser) Decode(r *accesslog.LogRecord, decoded *flowpb.Flow) error {
 	decoded.TraceContext = p.getTraceContext(r)
 	decoded.Summary = p.getSummary(r, decoded)
 
+	err = correlation.CorrelatePolicy(p.endpointGetter, decoded)
```
gandro: I don't think it makes sense to call this from the L7 parser, since we only support policy-verdict events at the moment. Maybe in the future we'll also support L7, but I'd hold off calling it here for now.

pkg/hubble/parser/seven/parser.go
```diff
@@ -151,9 +151,8 @@ func (p *Parser) Decode(r *accesslog.LogRecord, decoded *flowpb.Flow) error {
 	decoded.TraceContext = p.getTraceContext(r)
 	decoded.Summary = p.getSummary(r, decoded)
 
-	err = correlation.CorrelatePolicy(p.endpointGetter, decoded)
-	if err != nil {
-		p.log.WithError(err).Error("error correlating policy with flow")
+	if p.endpointGetter != nil {
```
gandro: This commit is not yet squashed into the previous one. Is that intentional?

gandro (changes requested):
Overall looks good to me, thanks! One note regarding the L7 parser invocation

pkg/hubble/parser/seven/parser.go
chancez: Yeah, for ease of review. I'll squash after fixing your other comment.

gandro (approved):
Neat, thanks!

rolinh (approved):

api/v1/flow/flow.proto
```diff
@@ -120,6 +120,11 @@ message Flow {
     // This can be used to extend functionality for other Hubble compatible
     // APIs, or experiment with new functionality without needing to change the public API.
     google.protobuf.Any extensions = 150000;
+
+    // The CiliumNetworkPolicies allowing the egress of the flow.
+    repeated Policy egress_allowed_by = 21001;
```
nathanjsweet, line 125: What about denials?

pkg/policy/correlation/correlation.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func lookupPolicyForKey(ep v1.EndpointInfo, key policy.Key) (derivedFrom labels.LabelArrayList, rev uint64, ok bool) {
+	// Check for L4 policy rules
+	derivedFrom, rev, ok = ep.GetRealizedPolicyRuleLabelsForKey(key)
+
+	// TODO: We should inspect f.GetPolicyMatchType() to reduce the number of lookups here
+	if !ok {
+		// Check for L3 policy rules
+		derivedFrom, rev, ok = ep.GetRealizedPolicyRuleLabelsForKey(policy.Key{
```
nathanjsweet, line 119: Make sure to check that `DestPort` and `NextHdr` aren't already `0`, as it is a possibility in the `extractFlowKey` method.

nathanjsweet (changes requested):
A couple of nits, but otherwise this looks good.

api/v1/flow/flow.proto
nathanjsweet: I think you should name this `egress_verdict_by`.

chancez: Was hoping to keep the existing field names since we were using this in a fork of Cilium using this patch and keeping field names (for JSON export) would let us be compatible with our fork.

pkg/policy/correlation/correlation.go
chancez: Updated.

nathanjsweet (approved):

PR #27861, opened 2023-08-31
contrib: Move github draft release to post-release

joestringer:
There's a few suboptimal things with the current release process:
- If one contributor prepares part of the release but another person
  tags the release, then later steps in the process can fail because
  early scripts create local files that are required to complete the
  release. Those files won't be present on the second person's machine.
- Despite programmatically creating the release text, the workflow
  expects that a release preparer will manually pull the docker digests
  and insert it into the release text.

We can fix this by deferring the github release creation until the
post-release phase. At this point, the digests are known for the target
releases, so we can pull those from the specified target URL. At this
point, the changelog also includes all of the remaining details that we
need to prepare the release text.

As a side effect, this allows us to align the processes between the
patch versions and prereleases more closely.

Related: https://github.com/cilium/release/pull/141

### DIFF
--- a/contrib/release/post-release.sh
+++ b/contrib/release/post-release.sh
@@ -44,9 +44,9 @@ handle_args() {
         exit 1
     fi
 
-    if [ ! -z "$2" ] && ! echo "$2" | grep -q "[0-9]\+\.[0-9]\+\.[0-9]\+"; then
+    if [ ! -z "$2" ] && ! echo "$2" | grep -q "$RELEASE_REGEX"; then
         usage 2>&1
-        common::exit 1 "Invalid VERSION ARG \"$2\"; Expected X.Y.Z"
+        common::exit 1 "Invalid VERSION ARG \"$2\"; $RELEASE_FORMAT_MSG"
     fi
 
     if [ -z "${GITHUB_TOKEN}" ]; then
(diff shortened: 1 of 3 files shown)


contrib/release/post-release.sh
```diff
@@ -65,6 +65,21 @@ main() {
...
+    tail -n+4 CHANGELOG.md | sed '/^## v.*$/q' >> $version-release-summary.txt
+    tail -n+2 digest-$version.txt >> $version-release-summary.txt
+    logecho "Creating Github draft release"
+    logrun hub release create -d -F $version-release-summary.txt $version
+    logecho "Browse to $RELEASES_URL to see the draft release"
+
+    if version_is_prerelease "$version"; then
+        # No digest updates for main branch for prereleases.
+        return
+    fi
```
joestringer, line 81: There's a subtlety here: This will *not* submit digest PRs for `*-pre*` releases. This is because we expect these pre release versions to always be tagged on `main`, and we don't need to update the digests there.

However, it *will* submit digest changes PRs for `*-rc*` releases, assuming that those are made on branches. 

The biggest impact would be on the `vX.Y.0-rc.0` release, since it would attempt to prepare the digest changes for that release. If RC0 is prepared on main instead of the branch, then maybe the script will attempt to submit the digests PR to the v1.15 branch before it exists, and fail out...?

This matches the release workflows for main branch in https://github.com/cilium/release/pull/141 .

contrib/release/post-release.sh
```diff
@@ -65,6 +65,21 @@ main() {
 
     git checkout -b pr/$version-digests $version
     ${DIR}/pull-docker-manifests.sh "$@"
+
+    echo -e "$ersion\n" > $version-release-summary.txt
+    # Grab the release notes for the current release, stop before the next
+    # release. The start of line `## vX.Y.Z` lines will match in command.
+    tail -n+4 CHANGELOG.md | sed '/^## v.*$/q' >> $version-release-summary.txt
```
joestringer, line 72: I tested this sed expression with both stable branch changelogs as well as prereleases, and it seems to correctly pick up the most recent changelog correctly. This relies on `start-release.sh` calling `prep-changelog.sh` and commiting the `CHANGELOG.md` into the release commit.

contrib/release/post-release.sh
joestringer: Ah well actually the script below asks the release manager to confirm if that's what we want. So I think both cases are covered :+1:

aanm (approved):

PR #27868, opened 2023-09-01
devices: Do not ignore link local addresses

joamaki:
The devices controller ignored link local addresses which usually are the automatically assigned fe80:: and 169.254.0.0/16. In some setups however link local addresses are used for K8s purposes and there is no good rational for ignoring them. The likely reason why they were initially ignored was to make the tests predictable by not having automatically assigned addresses on the test network devices.

This only affected the unreleased v1.15. 

```release-note
Do not ignore link local addresses when detecting network devices. This fixes a problem in setups where network devices that only had link local addresses were ignored.
```

### DIFF
--- a/pkg/datapath/linux/devices_controller_test.go
+++ b/pkg/datapath/linux/devices_controller_test.go
@@ -60,6 +60,17 @@ func devicesControllerTestSetup(t *testing.T) {
 	})
 }
 
+func containsAddress(dev *tables.Device, addrStr string) bool {
+	addr := netip.MustParseAddr(addrStr)
+	for _, a := range dev.Addrs {
+		if a.Addr == addr {
+			return true
+		}
+	}
+	return false
+
+}
+
 func TestDevicesController(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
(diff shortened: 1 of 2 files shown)


joamaki:
Hey, how should we mark bugs that were in an unreleased version (this fixed a bug in recently merged PR which did not make it into v1.14)?

pkg/datapath/linux/devices_controller_test.go
```diff
@@ -60,6 +60,17 @@ func devicesControllerTestSetup(t *testing.T) {
...
 
+func containsAddress(dev *tables.Device, addrStr string) bool {
+	addr := netip.MustParseAddr(addrStr)
+	for _, a := range dev.Addrs {
+		if a.Addr == addr {
+			return true
+		}
+	}
+	return false
+
```
aspsk, line 71: extra empty line

aspsk (approved):
LGTM, one style change

julianwiedmann:
> Hey, how should we mark bugs that were in an unreleased version (this fixed a bug in recently merged PR which did not make it into v1.14)?

It's `kind/bug`, but `release-note/misc` (as users were never exposed to it).

pkg/datapath/linux/devices_controller_test.go
joamaki: Since the tests passed and it's in a `_test.go` file I think I'll leave it. I'm definitely going to go back to these tests in the near future anyway.

PR #27857, opened 2023-08-31
ci: run verifier tests with proper Go toolchain version

tklauser:
Currently, the verifier tests are built and run with whatever Go toolchain version the LVH images ship. To make sure these tests are run with the same Go toolchain version as the rest of the tests install a copy of that Go toolchain before running the test.

This wasn't caught in CI on https://github.com/cilium/cilium/pull/27820 because the verifier workflow wasn't run passed on the ariane exclusion list.

Reported-by: @julianwiedmann

PR #27869, opened 2023-09-01
ci: disable cgo when installing Go toolchain

tklauser:
The LVH images don't have GCC, so we need to explicitly disable cgo. This fixes the following failure:

	# runtime/cgo
	_cgo_export.c:3:10: fatal error: stdlib.h: No such file or directory
	    3 | #include <stdlib.h>
	      |          ^~~~~~~~~~
	compilation terminated.

Fixes: bb3eec41c750 ("ci: run verifier tests with proper Go toolchain version")


aanm (approved):

giorio94:
Marked for backport to all stable branches as it fixes https://github.com/cilium/cilium/pull/27857, which is being backported as well.

PR #27870, opened 2023-09-01
daemon: Add option conntrackGCMaxInterval

joamaki:
When ToFQDN policies are in use (CIDR) identities are created for each IP in a DNS response that matches a ToFQDN policy. These identities are garbage collected when

  1) all endpoints with ToFQDN policies are removed
  2) the maximum number of IPs per host is reached (tofqdn-max-ips-per-host)
  3) when the identity has been unused and not refreshed by conntrack map GC

Problems arise when he conntrack GC interval becomes very long. If it's backed by a LRU BPF map, the maximum is set to 12 hours (defaults.ConntrackGCMaxLRUInterval), meaning it will take 12 hours before unused identities are marked dead and collected (unless tofqdn-max-ips-per-host limit is reached for the FQDN entry).

To allow user to have more control over this add the conntrackGCMaxInterval option that will allow limiting the maximum interval to something less than the 12 hours.

```release-note
Add option conntrackGCMaxInterval to allow limiting the maximum connection tracking GC interval. By default the automatic interval calculation may increase the interval up to 12 hours, which may incur an unreasonable delay to releasing of CIDR identities created from ToFQDN policies. Setting this option will limit the interval and ensure such identities are marked unused earlier and removed.
```


install/kubernetes/Makefile
```diff
@@ -17,7 +17,7 @@ MIN_K8S_VERSION := "v$(MIN_K8S_MAJOR).$(MIN_K8S_MINOR).0"
 CILIUM_CHARTS := $(ROOT_DIR)/$(RELATIVE_DIR)/cilium
 CHART_FILE := "$(CILIUM_CHARTS)/Chart.yaml"
 
-VALUES_TMPL_FILE_NAME ?= *values.yaml.tmpl*
+VALUES_TMPL_FILE_NAME ?= *values.yaml.tmpl
```
joamaki, line 20: Removing the `*` as this was also match files left behind by rebase/merge conflicts (e.g. values.yaml.tmpl.orig). Hopefully that'll save the next person some time and confusion :D

gandro: I believe these wildcards are there on purpose. Some distributions of Cilium use it to inject additional value templates during the build.

Edit: Ah, I missed the fact that it's still a suffix-wildcard. That should be fine then, nevermind.

gandro (approved):
Thanks!

daemon/cmd/daemon_main.go
```diff
@@ -210,6 +210,9 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	flags.Duration(option.ConntrackGCInterval, time.Duration(0), "Overwrite the connection-tracking garbage collection interval")
 	option.BindEnv(vp, option.ConntrackGCInterval)
 
+	flags.Duration(option.ConntrackGCMaxInterval, time.Duration(0), "Set the maximum for the automatic connection-tracking garbage collection interval calculation")
```
ldelossa: Are we interpreting this flag in seconds? If so, do we want to state that in the desc here? 

If not, do we want to have an implicit or explicit 'min' value as well?

ldelossa (changes requested):

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -589,6 +589,13 @@ cni:
 # @default -- `"0s"`
 conntrackGCInterval: ""
 
+# -- (string) Configure how the maximum frequency for the garbage collection of the
```
learnitall: Grammar nit:
```suggestion
# -- (string) Configure the maximum frequency for the garbage collection of the
```

learnitall (changes requested):
Small grammar nit, but otherwise LGTM.

One suggestion I do have, which isn't necessary but would be nice, is somehow changing the value for the `conntrackGCMaxInterval` to the actual default that Cilium will use if it isn't set. Having a default value for `0s` is a bit confusing.

daemon/cmd/daemon_main.go
joamaki: It's the same type of flag as the `ConntrackGCInterval`. It's a `time.Duration` and Go can parse it in different forms, e.g. `10ms` is 10 milliseconds, `30m` is 30 minutes and so on. See https://pkg.go.dev/time#ParseDuration. I don't think minimum is required. Users should be smart enough to use sensible durations here.

joamaki:
> Small grammar nit, but otherwise LGTM.
> 
> One suggestion I do have, which isn't necessary but would be nice, is somehow changing the value for the `conntrackGCMaxInterval` to the actual default that Cilium will use if it isn't set. Having a default value for `0s` is a bit confusing.

The actual default depends on the kernel, e.g. whether or not LRU maps are used, so we cannot up-front tell what it is.

install/kubernetes/cilium/values.yaml.tmpl
joamaki: Thanks, fixed!

ldelossa (approved):

Documentation/helm-values.rst
```diff
@@ -736,6 +736,10 @@
      - Configure how frequently garbage collection should occur for the datapath connection tracking table.
      - string
      - ``"0s"``
+   * - :spelling:ignore:`conntrackGCMaxInterval`
+     - Configure how the maximum frequency for the garbage collection of the connection tracking table. Only affects the automatic computation for the frequency and has no effect when 'conntrackGCInterval' is set. This can be set to more frequently clean up unused identities created from ToFQDN policies.
+     - string
+     - ``"0s"``
```
learnitall: Looks like this description is outdated. Can you update this file? (ie `cd Documentation && make update-helm-values`)

learnitall (dismissed):

learnitall:
> The actual default "maximum" depends on the kernel, e.g. whether or not LRU maps are used, so we cannot up-front tell what it is.

> I've dropped the @default from the helm values.yaml.tmpl as it's not useful information nor does this flag need to be always set.

Ah ok that makes sense. Dropping the `@default` is a good middle ground, thanks!

Documentation/helm-values.rst
joamaki: Oops, always forget one of these steps. Fixed.

learnitall (approved):

asauber (approved):
LGTM, and the new logic looks correct

tommyp1ckles (approved):

daemon/cmd/daemon_main.go
```diff
@@ -205,6 +205,9 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
...
```
joamaki: Thanks, reworded it. EDIT: And forgot cmdref regeneration again doh.

PR #27873, opened 2023-09-01
Fix bug when reusing the same cell in multiple hives

giorio94:
Currently, the "Invoke" cell internally stores a reference to the container when `Apply()` gets called. Yet, this causes issues if the same cell is registered as part of multiple hives (e.g., constructing multiple subcommands), as that reference gets overwritten by new `hive.New()` executions.

Let's remove this explicit reference, capturing instead the container parameter through an anonymous function. Additionally, let's craft a unit test to prevent regressions.

```release-note
Fix bug when reusing the same cell in multiple hives
```

### DIFF
--- a/pkg/hive/cell/invoke.go
+++ b/pkg/hive/cell/invoke.go
@@ -10,7 +10,6 @@ import (
 )
 
 type invoker struct {
-	cont  container
 	funcs []namedFunc
 }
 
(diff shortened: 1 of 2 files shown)


joamaki (approved):
Oopsie. Thanks for the fix, LGTM!

joamaki:
Even though there was nothing shipped in v1.14 that was affected by this bug, lets backport this before we backport code that could be affected.

giorio94:
Rebased onto main

PR #27875, opened 2023-09-01
ci: Add a workflow to update labels of backported PRs

pippolo84:
Add a reusable workflow to be called from stable branches whenever a backport PR is merged. The workflow scans the body of the backport PR (that is, a PR like [this one](https://github.com/cilium/cilium/pull/27739)) to get the list of the original PRs, and updates the label `backport-pending/<version>` to `backport-done/<version>` for each of them.

The list of PRs to be updated is fetched from the Backport PR body section labeled `upstream-prs`.

Fixes: https://github.com/cilium/github-actions/issues/12

PR #27876, opened 2023-09-01
[v1.14] ci: Add a call to the update label backport action

pippolo84:
Add an action to call the workflow that update the labels of backported PRs in v1.14 stable branch.

Depends on: #27875

### DIFF
--- a/.github/workflows/call-backport-label-updater.yaml
+++ b/.github/workflows/call-backport-label-updater.yaml
@@ -1,0 +1,19 @@
+---
+    name: Call Backport Label Updater
+    on:
+      pull_request:
+        types:
+          - closed
+        branches:
+          - v1.14
+    jobs:
+      call-backport-label-updater:
+        if: |
+            github.event.pull_request.merged == true &&
+            contains(github.event.pull_request.body, 'upstream-prs') &&
+            contains(github.event.pull_request.labels.*.name, 'backport/1.14')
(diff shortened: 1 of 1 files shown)


.github/workflows/call-backport-label-updater.yaml
```diff
@@ -0,0 +1,19 @@
...
+      pull_request:
+        types:
+          - closed
+        branches:
+          - v1.14
+    jobs:
+      call-backport-label-updater:
+        if: |
+            github.event.pull_request.merged == true &&
+            contains(github.event.pull_request.title, 'Backports') &&
```
aanm: Maybe we should have this instead
```
contains(github.event.pull_request.body, 'upstream-prs')
```

which will work for all cases that we want automation.

aanm (changes requested):

.github/workflows/call-backport-label-updater.yaml
pippolo84: Fixed. I've updated also https://github.com/cilium/cilium/pull/27877 and https://github.com/cilium/cilium/pull/27879

aanm (approved):

brlbil (approved):

pippolo84:
This PR just adds a call to a workflow, so running the full CI is pointless. Marking this ready-to-merge

PR #27881, opened 2023-09-01
egressgw: switch unit tests to reconciliationEventsCount

jibi:
instead of just retrying to assert that egressgw bpf entries and IP
rules/routes are consistent with the expected state, use the newly
introduced reconciliationEventsCount manager's variable to ensure that
the manager has actually gone through a reconciliation round

### DIFF
--- a/pkg/egressgateway/manager_privileged_test.go
+++ b/pkg/egressgateway/manager_privileged_test.go
@@ -157,11 +157,15 @@ func (k *EgressGatewayTestSuite) TestEgressGatewayManager(c *C) {
 	close(k.cacheStatus)
 	k.policies.sync(c)
 
+	reconciliationEventsCount := egressGatewayManager.reconciliationEventsCount
+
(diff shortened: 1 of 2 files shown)


pkg/egressgateway/manager_privileged_test.go
```diff
@@ -558,14 +592,7 @@ func parseIPRule(sourceIP, destCIDR, egressIP string, ifaceIndex int) parsedIPRu
...
 func assertIPRules(c *C, rules []ipRule) {
-	var err error
-	for i := 0; i < 10; i++ {
-		if err = tryAssertIPRules(rules); err == nil {
-			break
-		}
-		time.Sleep(10 * time.Millisecond)
-	}
-
+	err := tryAssertIPRules(rules)
```
jibi, line 595: I'm keeping this assert/tryAssert split as it maintains the code more readable imo, as tryAssert just needs to return an error rather than calling c.Assert(err, IsNil) each time

jschwinger233 (approved):
LGTM

PR #27884, opened 2023-09-01
Add generic reconciler and replace L2 responder reconciler

dylandreimerink:
This PR adds generic map reconciliation logic. The idea behind this is that in order to add resilience, we want to move most maps to a pattern where the desired state lives in a stateDB table and reconciliation logic will keep the BPF map in sync with this desired state.

Since most BPF maps have the same requirements, we can reuse the same reconciliation logic. The only map to currently use a similar reconciler is the L2 responder map, so this PR also changes it out for this generic reconciler.

Lastly, this PR changes the L2 announcer to use the devices table instead of the legacy method of getting devices info and the GARP capability is now provided via hive so it can be mocked out. See commit messages for details.

```release-note
Add generic reconciler and replace L2 responder reconciler
```


pkg/mapreconciler/reconciler.go
```diff
@@ -0,0 +1,276 @@
...
+		default:
+		}
+
+		// This will block until an item is available or the queue is shutdown.
+		// Shutdown happens when Stop() is called so this should not block on cell shutdown.
+		item, shutdown := m.retryQueue.Get()
+		if shutdown {
+			return nil
+		}
+
```
joamaki, line 236: Where is `m.retryQueue.Done` called? `Forget` does nothing except clear the rate-limiting data. Without call to `Done(key)` the `key` will never be processed again. Tests probably lacking coverage here?

pkg/mapreconciler/reconciler.go
```diff
@@ -0,0 +1,276 @@
...
+// partialReconciliation processes the changes in the delete tracker and applies them to the BPF map.
+// It doesn't reconcile values that have not changed and doesn't check the map for rouge values, this makes it fast.
+func (m *mapReconciler[E, K, V]) partialReconciliation(ctx context.Context, tracker *statedb.DeleteTracker[E], revIn uint64) (uint64, <-chan struct{}) {
+	rxn := m.db.ReadTxn()
+	rev, invalidate, err := tracker.Process(rxn, revIn, func(entry E, deleted bool, _ uint64) error {
+		if deleted {
+			err := m.m.Delete(entry.Key())
+			// If we error, retry this key later, don't hold up the delete tracker.
+			if err != nil && !errors.Is(err, ebpf.ErrKeyNotExist) {
+				m.retryQueue.AddRateLimited(entry.Key())
```
joamaki, line 123: Doesn't this lead to races? E.g. it might be we'll retry the delete while it was added back? `Forget` doesn't remove anything from the queue. It would probably be better to not use `workqueue` here at all, but rather implement it directly and drive it from the same loop.

pkg/mapreconciler/reconciler.go
dylandreimerink: Hmm, that is unfortunate naming, but then again I did not check the API docs. You are likely right about the tests. This is challenging since the retry logic is all time based which makes for really flaky tests.

pkg/mapreconciler/reconciler.go
dylandreimerink: The idea is that we only store the key for retries, just like Resource[T] so it doesn't matter if something happened in the table between us adding the key and the actual retry action. Not sure this warrants a custom retry queue, although it might make testing and verification easier :thinking:

joamaki: It's more that the snapshot that the retry gets might be older or newer than what we have here, e.g. that's why it might travel back in time.

pkg/mapreconciler/reconciler.go
```diff
@@ -0,0 +1,276 @@
...
+		}
+
+		// clear the key from the retry queue (if present) since we just performed an update without error
+		m.retryQueue.Forget(k)
+	}
+
+	return rev
+}
+
+func (m *mapReconciler[E, K, V]) Retirer(ctx context.Context) error {
```
joamaki, line 220: Should be `Retrier`? Is there a reason this is exported? 
BTW, would `retryLoop` and `reconcileLoop` work better as names?

PR #27886, opened 2023-09-01
external-workloads: do not automatically infer ClusterID/Name

giorio94:
Currently, when the Cilium agent running on a VM joins an existing cluster, it automatically infers the corresponding ClusterID and ClusterName. Yet, this introduces a dependency order, because all consumers of that field need to be initialized after the connection to the remote cluster, which is especially problematic as more components get converted to cells. Additionally, this may also lead to possible user confusion, as the corresponding parameter is reported as not initialized in the logs.

Hence, let's require these parameters to be explicitly set, reporting a failure in case they don't match those automatically discovered.

Depends on https://github.com/cilium/cilium-cli/pull/1948

```release-note
Don't automatically infer ClusterID and ClusterName for external workloads.
```

### DIFF
--- a/Documentation/operations/upgrade.rst
+++ b/Documentation/operations/upgrade.rst
@@ -320,6 +320,12 @@ Annotations:
   labels ``reserved:init``, these policies must be converted to
   ``CiliumClusterwideNetworkPolicy`` by changing the resource type for the
   policy.
+* Cluster name and ID are no longer automatically inferred by Cilium agents
+  running on :ref:`external workloads <external_workloads>`.
+  If the cluster name and ID are different from the default values, you must
+  specify them as parameters.
+  Generate the installation script using Cilium CLI >=v0.15.8 to automatically
+  include these parameters.
 
 .. _upgrade_cilium_cli_helm_mode:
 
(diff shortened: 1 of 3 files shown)


gandro (commented):
Thanks! Does this need to be documented in the upgrade notes as a breaking change or is this not a configuration we expect users to have?

giorio94:
> Does this need to be documented in the upgrade notes as a breaking change or is this not a configuration we expect users to have?

External workloads are marked as a beta feature, but I agree that it makes sense to call this out as a breaking change. I've updated the upgrade notes to document it.

gandro (approved):

Documentation/operations/upgrade.rst
```diff
@@ -320,6 +320,11 @@ Annotations:
   labels ``reserved:init``, these policies must be converted to
   ``CiliumClusterwideNetworkPolicy`` by changing the resource type for the
   policy.
+* Cluster name and ID are no longer automatically inferred by Cilium agents
+  running on :ref:`external workloads <external_workloads>`, and need to be
+  explicitly specified as parameters (if different from the default values).
```
zacharysarah: Small edits for clarity
```suggestion
  running on :ref:`external workloads <external_workloads>`. If the cluster name
  and ID are different from the default values, you must specify them 
  as parameters.
```

zacharysarah (changes requested):
@giorio94 Thanks for your patience. This is well-written and needs only some small tweaks for approval.

giorio94:
> @giorio94 Thanks for your patience. This is well-written and needs only some small tweaks for approval.

Thanks @zacharysarah for the feedback, I've applied your suggestion.

zacharysarah (approved):
LGTM

PR #27889, opened 2023-09-01
hubble: Use protobuf GetType() helper in v1.FlowProtocol() to avoid possible panic

chancez:
If the event type is unset, or the flow is nil, it's possible that FlowProtocol could panic. To avoid this, use the GetType() helper method which is nil safe.

```release-note
hubble: Use protobuf GetType() helper in v1.FlowProtocol() to avoid possible panic
```

### DIFF
--- a/pkg/hubble/api/v1/flow.go
+++ b/pkg/hubble/api/v1/flow.go
@@ -11,7 +11,7 @@ import (
 // FlowProtocol returns the protocol best describing the flow. If available,
 // this is the L7 protocol name, then the L4 protocol name.
 func FlowProtocol(flow *pb.Flow) string {
-	switch flow.GetEventType().Type {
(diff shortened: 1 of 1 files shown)


michi-covalent (approved):
😇

rolinh (approved):

rolinh:
I guess this affects v1.14 and should be backported, correct?

chancez:
@rolinh Sort of. It's not an actual bug in practice, just a possible bug, so I don't know if we need to backport it. I only noticed when I was writing some new tests in another PR, and the tests were triggering a panic here because the code used this function.

ghost (approved):

chancez:
Rebased, and then I shortened the commit message summary to satisfy `checkpatch`.

PR #27892, opened 2023-09-01
Improve bump-readme.sh

joestringer:
This PR has two motivations:
- Add support for automatically bumping the latest "prerelease" version in the
  README.md so that prereleases can follow the same process as patch releases
- Add guardrails to prevent malformed tables from breaking the representation
  of the latest versions in the readme.

One of the problems we have occasionally is when a new version changes the
length of the line in the table in the README. This breaks the format of the
table, which then causes GitHub to stop rendering the table altogether. Ideally
we'd make the bump readme script smart enough to properly format the table, but
we should probably reimplement the whole thing in a real language like Go to
be able to reliably mutate the readme for this case. One step short of that is
to at least detect a malformed table and report an error to the user to then
fix it. So, go with that approach for now.

This PR also adds some default bash error checking options to properly fail out
when part of the process fails.

### DIFF
--- a/contrib/release/bump-readme.sh
+++ b/contrib/release/bump-readme.sh
@@ -112,7 +112,7 @@ check_table "commits/v1"
 
 git add README.rst stable.txt Documentation/_static/stable-version.json $ACTS_YAML
 if ! git diff-index --quiet HEAD -- README.rst stable.txt Documentation/_static/stable-version.json $ACTS_YAML; then
-    git commit -s -m "Update stable releases"
+    git commit -s -m "README: Update releases"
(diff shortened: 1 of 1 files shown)


aditighag (commented):
Looks like useful improvements. I can rubber-stamp this if you've already used the script. Or maybe we could request for a review from someone else who would be using the script?

joestringer:
I used this version to generate https://github.com/cilium/cilium/pull/27864 and double-check the table formatting, so I think it should be good to go.

aanm (approved):

Issue #15283, opened 2021-03-09
Debug message "stale identity observed" is frequently printed in logs


joestringer:
Cilium latest, during v1.10 dev cycle.

Seen in a failed PR, this message is frequently printed to the log:

```
2021-03-09T21:05:11.508381379Z level=debug msg="stale identity observed" identity=6 ipAddr=10.0.0.112 oldIdentity=1 subsys=hubble
```

It's just a debug message, but given how frequently it's being logged it seems like it's trying to draw attention to something being out-of-place compared to how the code was written. Just filing this to follow up and see if it's surprising to anyone. CC @cilium/hubble 

Example CI job:
https://jenkins.cilium.io/job/Cilium-PR-K8s-1.14-kernel-4.9/113/testReport/junit/Suite-k8s-1/14/K8sConformance_Portmap_Chaining_Check_connectivity_check_compliance_with_portmap_chaining/

[71a2aa28_K8sConformance_Portmap_Chaining_Check_connectivity-check_compliance_with_portmap_chaining.zip](https://github.com/cilium/cilium/files/6111782/71a2aa28_K8sConformance_Portmap_Chaining_Check_connectivity-check_compliance_with_portmap_chaining.zip)

pchaigno:
I believe bug https://github.com/cilium/cilium/issues/21844 has caused an increase of these debug messages, specifically for former remote-node IP addresses:
```
2022-10-21T14:51:01.983100689Z level=debug msg="stale identity observed" identity=2 ipAddr=10.164.0.29 oldIdentity=6 subsys=hubble
```

https://github.com/cilium/cilium/issues/21844 was introduced by 454df0c065bd94aa30a9d703732d64a24d3b6073 (Jun. 2022), so it doesn't explain the first report of this debug message (Mar. 2021).

gandro:
Unrelated to Paul's comment: I agree with the original sentiment, that the message can be quite spammy. I think the biggest offender is due to a conceptual mismatch between Hubble and the datapath: https://github.com/cilium/cilium/issues/14427

I wonder if we somehow could filter out messages caused by #14427 and leave the rest in place. It should then only be emitted temporarily when an identity of a prefix changes, or consistently if we have an actual bug (which I think is useful to log).

giorio94:
It looks like that something changed recently and made this even worse. The following entries were printed +16K and +1K times respectively in less than 1h (latest on master):

```
[...]

Issue #14427, opened 2020-12-16
Hubble's identity lookup seems inconsistent with datapath


gandro:
PR #14090 introduced a check in the Hubble parser to emit a debug log message when the numeric identity of the datapath is inconsistent with the identity Hubble mapped to the observed IP. 

I have found that this message is hit very often in current master, indicating a bug in Hubble's mapping from IP to identity. The following output is from the connectivity-check.yaml.

``` 
Dec 16 10:43:53 k8s1 cilium-agent[6429]: level=debug msg="stale identity observed" identity=16777218 ipAddr=172.217.168.36 oldIdentity=2 subsys=hubble
Dec 16 10:43:53 k8s1 cilium-agent[6429]: level=debug msg="stale identity observed" identity=12982 ipAddr=10.11.0.40 oldIdentity=1 subsys=hubble
Dec 16 10:43:53 k8s1 cilium-agent[6429]: level=debug msg="stale identity observed" identity=18788 ipAddr=192.168.33.11 oldIdentity=1 subsys=hubble
```

These messages are expected to show up after an identity change where the IPCache has not yet been synced with the BPF maps - however the above messages persist, indicating that datapath performs the lookup is inconsistent with the Hubble one.

"identity" refers to the identity used in the datapath, "oldIdentity" refers to the user-space mapping found by Hubble.

```
$ cilium identity get 18788
ID      LABELS
18788   k8s:io.cilium.k8s.policy.cluster=default
        k8s:io.cilium.k8s.policy.serviceaccount=default
        k8s:io.kubernetes.pod.namespace=cilium-test
        k8s:name=echo-b
$ cilium identity get 12982
ID      LABELS
12982   k8s:io.cilium.k8s.policy.cluster=default
        k8s:io.cilium.k8s.policy.serviceaccount=default
        k8s:io.kubernetes.pod.namespace=cilium-test
        k8s:name=pod-to-b-intra-node-nodeport
cilium identity get 16777218
ID         LABELS
16777218   cidr:172.217.168.36/32
           reserved:world
```

It seems likely that the messages have different root causes:

- ~The `16777218` identity is coming from an FQDN policy. Hubble for some reason in unable to look it up in the IPCache. That clearly is a bug in Hubble's lookup mechanism.~ Fixed in #14477 

[...]

PR #27894, opened 2023-09-02
Reduce "stale identity observed" warnings

leblowl:

      GitHub issue.

## Overview

Hi, I've been looking into the "stale identity observed" issues (https://github.com/cilium/cilium/issues/14427 and https://github.com/cilium/cilium/issues/15283) and documented some of the situations where I've seen these log messages in my setup. Ultimately, I think the log messages arise from a misunderstanding between the datapath and Hubble. In the cases I studied, the datapath looks like it's using an appropriate identity. However Hubble has a limited amount of information and isn't really aware of what the datapath is doing in these situations. It just recieves an IP and security identity from the datapath and warns if the security identity doesn't match the IP cache security identity. However, in some cases, the datapath doesn't use the IP cache security identity, or it sends trace notify events before resolving a security identity via the IP cache, or hardcodes an identity when tunneling a packet, or when a proxy is involved, the datapath doesn't know the original IP for a given security identity.

Currently, this warning is not very helpful. There are so many warnings that it's easy to ignore them.

One solution is to see if we can create various conditions on the Hubble side so that it can tell when to send a warning, such that the warnings are actually meaningful. That's what I've explored in this PR.

### Case 1

When encapsulating a packet for sending via the overlay network, if the source seclabel = HOST_ID, then we reassign seclabel with LOCAL_NODE_ID and then send a trace notify: https://github.com/cilium/cilium/blob/c7bf490b8e2feb141cc01e98763dbda313af0288/bpf/lib/encap.h#L39

In this case any packets with identity HOST_ID, are going to result in warnings from Hubble.

A simple solution is to ignore stale TRACE_TO_OVERLAY events when datapath security ID = remote_node && userspace security ID = host.

Here is an example:
```
2023-08-30T19:59:00.281808717Z stderr F level=debug msg="&{{4 4 4 3848637028 168 128 1 remote-node health 0 5 0 4} ::} 10.244.1.40 10.244.0.61 TO_OVERLAY" subsys=hubble
2023-08-30T19:59:00.281811808Z stderr F level=debug msg="stale identity observed" identity=6 ipAddr=10.244.1.40 oldIdentity=1 subsys=hubble
```

In this case 10.244.1.40 is the IP address of cilium_host.

### Case 2

Sometimes packets from endpoint link-local addresses are intercepted by cil_from_container. Because link-local addresses are not stored in the IP cache, Hubble assigns them ID 2 (WORLD_ID) (edit: now 9 or 10 for WORLD_IPV4/IPV6).

Here is an example:
```
2023-08-30T20:15:14.102227401Z stderr F level=debug msg="&{{4 5 1189 0 70 70 1 health unknown 0 5 0 0} ::} fe80::c809:60ff:fecd:ab08 ff02::2 FROM_ENDPOINT" subsys=hubble
2023-08-30T20:15:14.102232825Z stderr F level=debug msg="stale identity observed" identity=4 ipAddr="fe80::c809:60ff:fecd:ab08" oldIdentity=2 subsys=hubble
```
fe80::c809:60ff:fecd:ab08 is the link-local address of the cilium-health-responder namespace. I've also observed endpoint link-local addresses.

A simple solution is to ignore stale TRACE_FROM_ENDPOINT events when datapath source security ID = {4 | !reserved} && userspace security ID is a world ID. Alternatively, in this case, I wonder if we can just add link-local addresses to the IP cache or prevent endpoints from sending these packets.

### Case 3

When a pod sends a packet to the Kubernetes API its IP is masqueraded and then when it receives a response and the masquerade is reversed, cil_from_host determines that the source ID is 2 (edit: now 2, 9 or 10) because there is no packet mark.

Here is an example:
```
2023-08-31T23:21:50.489380414Z stderr F level=debug msg="&{{4 5 742 2245691099 201 128 1 60454 unknown 0 5 0 0} ::} 10.244.1.242 10.96.0.1 FROM_ENDPOINT" subsys=hubble
2023-08-31T23:21:50.489451772Z stderr F level=debug msg="&{{4 3 742 2245691099 201 128 1 60454 kube-apiserver 0 1 0 0} ::} 10.244.1.242 172.20.0.3 TO_STACK" subsys=hubble
2023-08-31T23:21:50.490396351Z stderr F level=debug msg="&{{4 10 4 1883766498 155 128 1 unknown unknown 0 5 0 6} ::} 172.20.0.3 172.20.0.2 FROM_NETWORK" subsys=hubble
2023-08-31T23:21:50.490422485Z stderr F level=debug msg="&{{4 7 4 1883766498 155 128 1 world unknown 0 5 0 6} ::} 172.20.0.3 10.244.1.242 FROM_HOST" subsys=hubble
2023-08-31T23:21:50.490448558Z stderr F level=debug msg="stale identity observed" identity=2 ipAddr=172.20.0.3 oldIdentity=7 subsys=hubble
2023-08-31T23:21:50.490659919Z stderr F level=debug msg="&{{4 0 742 1883766498 155 128 1 kube-apiserver 60454 742 2 0 9} ac14:3::} 172.20.0.3 10.244.1.242 TO_ENDPOINT" subsys=hubble
```
In this case, we can see 10.244.1.242 (a CoreDNS instance) is reaching out to 10.96.0.1 (the ClusterIP for the Kubernetes API). 10.96.0.1 get's resolved to 172.20.0.3. And then 10.244.1.242 get's masqueraded to 172.20.0.2 when it's being sent out via eth0. When the response is recieved, 172.20.0.2 is reversed to 10.244.1.242 and then it's routed to cilium_host. It's at this point that the datapath considers the ID as 2 because there isn't any mark on the packet.

A simple solution is to ignore stale TRACE_FROM_HOST events when datapath source security ID = 2 (edit: now 2, 9 or 10) and userspace security ID = 7. In general, if packets are picked up by cil_from_host and they do not have a packet mark, the TRACE_FROM_HOST event will contain ID = 2 (edit: now 2, 9 or 10), as that's the default in inherit_identity_from_host, and the trace happens before any IP cache resolution. Should inherit_identity_from_host return unknown by default?

### Case 4

When proxied packets (via Cilium DNS proxy) are sent from the host their source IP is that of the host, yet their security identity is retained from the original source pod.

Here is an example:
```
2023-09-01T01:51:37.455236212Z stderr F level=debug msg="&{{4 5 1825 2275186770 110 110 1 47432 unknown 0 5 0 0} ::} 10.244.1.11 10.96.0.10 FROM_ENDPOINT" subsys=hubble
2023-09-01T01:51:37.455330139Z stderr F level=debug msg="&{{4 1 1825 2275186770 110 110 1 47432 unknown 36245 0 0 0} ::} 10.244.1.11 10.244.0.127 TO_PROXY" subsys=hubble
2023-09-01T01:53:04.360088799Z stderr F level=debug msg="&{{4 7 4 3946258847 110 110 1 47432 unknown 0 5 0 0} ::} 10.244.1.40 10.244.0.127 FROM_HOST" subsys=hubble
2023-09-01T01:53:04.360106623Z stderr F level=debug msg="stale identity observed" identity=47432 ipAddr=10.244.1.40 oldIdentity=1 subsys=hubble
2023-09-01T01:53:04.360129059Z stderr F level=debug msg="&{{4 4 4 3946258847 110 110 1 47432 60454 0 5 0 4} ::} 10.244.1.40 10.244.0.127 TO_OVERLAY" subsys=hubble
2023-09-01T01:53:04.360136765Z stderr F level=debug msg="stale identity observed" identity=47432 ipAddr=10.244.1.40 oldIdentity=1 subsys=hubble
```
In this case, 10.244.1.11 sends a DNS request to 10.96.0.10, which get's resolved to 10.244.0.127. The Cilium DNS proxy recieves the packet and sends it back out via cilium_host with address 10.244.1.40. Since the packet retains it's original source security ID, we recieve a warning from both TRACE_FROM_HOST and TRACE_TO_OVERLAY events.

A simple solution is to ignore stale TRACE_FROM_HOST/TRACE_TO_OVERLAY events when datapath source security ID is a local endpoint ID and userspace security ID = 1.

### Case 5

When proxied packets (via Cilium DNS proxy) are recieved by the destination host their source IP is that of the proxy, yet their security identity is retained from the original source pod. This is a similar case to #4, but on the receiving side.

In one case, the packet is sent to a DNS instance on the same host:
```
2023-09-01T02:32:11.417310878Z stderr F level=debug msg="&{{4 0 742 2061866022 110 110 1 47432 60454 742 0 0 9} af4:128::} 10.244.1.40 10.244.1.242 TO_ENDPOINT" subsys=hubble
2023-09-01T02:32:11.417338717Z stderr F level=debug msg="stale identity observed" identity=47432 ipAddr=10.244.1.40 oldIdentity=1 subsys=hubble
```
And alternatively, to a separate host:
```
2023-09-01T02:32:12.267932053Z stderr F level=debug msg="&{{4 0 312 3008716175 110 110 1 47432 60454 312 0 0 11} af4:128::} 10.244.1.40 10.244.0.127 TO_ENDPOINT" subsys=hubble
2023-09-01T02:32:12.267936768Z stderr F level=debug msg="stale identity observed" identity=47432 ipAddr=10.244.1.40 oldIdentity=6 subsys=hubble
```
This results in a stale identity warning from TO_ENDPOINT when delivering to DNS servers on the same host and remote hosts.

A simple solution is to ignore stale TRACE_TO_ENDPOINT events where the datapath source security ID isn't reserved (must be an endpoint) and the userspace security ID is either 1 (host) or 6 (remote-node).

### Case 6

When reinstalling Cilium and the agent is just starting, packets arrive via tunnel for the health responder, for some reason instead of going directly to the health responder endpoint, they are passed to the stack.
```
2023-09-01T18:07:02.393926998Z stderr F level=debug msg="{{4 10 4 2383703504 124 124 1 unknown unknown 0 5 0 6} ::} 172.20.0.3 172.20.0.2 57885 8472 FROM_NETWORK" subsys=hubb
le
2023-09-01T18:07:02.393939748Z stderr F level=debug msg="{{4 9 0 2383703504 74 74 1 unknown unknown 0 5 0 4} ::} 10.244.0.186 10.244.1.104 35416 4240 FROM_OVERLAY" subsys=hub
ble
2023-09-01T18:07:02.393963115Z stderr F level=debug msg="{{4 3 4 2383703504 74 74 1 unknown unknown 0 5 0 3} ::} 10.244.0.186 10.244.1.104 35416 4240 TO_STACK" subsys=hubble
2023-09-01T18:07:02.393980895Z stderr F level=debug msg="{{4 7 4 2383703504 74 74 1 world unknown 0 5 0 3} ::} 10.244.0.186 10.244.1.104 35416 4240 FROM_HOST" subsys=hubble
2023-09-01T18:07:02.393993862Z stderr F level=debug msg="stale identity observed" identity=2 ipAddr=10.244.0.186 oldIdentity=6 subsys=hubble
```
In this case 10.244.1.104 is the health endpoint.

Perhaps because the endpoint map is not updated quickly enough, these packets get routed to the host and appear to enter a routing loop.
```
2023-09-01T18:07:02.394106942Z stderr F level=debug msg="{{4 3 4 2383703504 74 74 1 unknown unknown 0 5 0 3} ::} 10.244.0.186 10.244.1.104 35416 4240 TO_STACK" subsys=hubble
2023-09-01T18:07:02.394118785Z stderr F level=debug msg="{{4 7 4 2383703504 74 74 1 world unknown 0 5 0 2} ::} 10.244.0.186 10.244.1.104 35416 4240 FROM_HOST" subsys=hubble
2023-09-01T18:07:02.394130921Z stderr F level=debug msg="stale identity observed" identity=2 ipAddr=10.244.0.186 oldIdentity=6 subsys=hubble
2023-09-01T18:07:02.394144146Z stderr F level=debug msg="{{4 3 4 2383703504 74 74 1 unknown unknown 0 5 0 3} ::} 10.244.0.186 10.244.1.104 35416 4240 TO_STACK" subsys=hubble
2023-09-01T18:07:02.394160199Z stderr F level=debug msg="{{4 7 4 2383703504 74 74 1 world unknown 0 5 0 2} ::} 10.244.0.186 10.244.1.104 35416 4240 FROM_HOST" subsys=hubble
2023-09-01T18:07:02.394167603Z stderr F level=debug msg="stale identity observed" identity=2 ipAddr=10.244.0.186 oldIdentity=6 subsys=hubble
2023-09-01T18:07:02.394179001Z stderr F level=debug msg="{{4 3 4 2383703504 74 74 1 unknown unknown 0 5 0 3} ::} 10.244.0.186 10.244.1.104 35416 4240 TO_STACK" subsys=hubble
2023-09-01T18:07:02.394188292Z stderr F level=debug msg="{{4 7 4 2383703504 74 74 1 world unknown 0 5 0 2} ::} 10.244.0.186 10.244.1.104 35416 4240 FROM_HOST" subsys=hubble
2023-09-01T18:07:02.394190751Z stderr F level=debug msg="stale identity observed" identity=2 ipAddr=10.244.0.186 oldIdentity=6 subsys=hubble
```
The reason we get a warning is because at the beginning of cil_from_host, it only checks the packet mark for a security ID and then immediately sends a trace notify. These packets are coming from another host so they don't have a packet mark. After cilium get's up and going these warnings go away. In this instance I think we have an actual issue, due to a race condition at startup, though I'm not entirely sure how it could be solved or if it needs to be.

### Case 7

Link-local packets appear to be dropped and drop-notify events also cause stale identity (due to them missing from the ipcache). I haven't addressed this case yet.

## Alternatives

This PR does limit the warnings, but it seems like there is also a possibility that it limits legitimate warnings. I'm not sure I know enough about the various datapath cases to tell for sure. But, ignoring certain messages based on assumptions of how the datapath is working seems fragile. Just during the time I was preparing this PR, we added new identities for WORLD_IPV4 and WORLD_IPV6 which broke the logic. Perhaps a better solution would be to use different events besides trace/drop notify for these warnings. That way we can emit events from specific points in the datapath and warn on all of these events if they contain stale/incorrect information. It seems like another solution would be to potentially change some of the trace notify calls in the datapath, so they emit information that's more in sync with what userspace expects. But that seems like it would obscure the actual workings of the datapath.

### DIFF
--- a/pkg/hubble/parser/common/endpoint.go
+++ b/pkg/hubble/parser/common/endpoint.go
@@ -15,6 +15,14 @@ import (
 	"github.com/cilium/cilium/pkg/logging/logfields"
 )
 
+type DatapathContext struct {
+	SrcIP                 netip.Addr
+	SrcLabelID            uint32
+	DstIP                 netip.Addr
+	DstLabelID            uint32
+	TraceObservationPoint pb.TraceObservationPoint
+}
+
 type EndpointResolver struct {
 	log            logrus.FieldLogger
 	endpointGetter getters.EndpointGetter
@@ -36,7 +44,7 @@ func NewEndpointResolver(
 	}
 }
 
-func (r *EndpointResolver) ResolveEndpoint(ip netip.Addr, datapathSecurityIdentity uint32) *pb.Endpoint {
+func (r *EndpointResolver) ResolveEndpoint(ip netip.Addr, datapathSecurityIdentity uint32, context DatapathContext) *pb.Endpoint {
 	// The datapathSecurityIdentity parameter is the numeric security identity
 	// obtained from the datapath.
 	// The numeric identity from the datapath can differ from the one we obtain
@@ -45,29 +53,76 @@ func (r *EndpointResolver) ResolveEndpoint(ip netip.Addr, datapathSecurityIdenti
 	// created and the time the event reaches the Hubble parser.
 	// To aid in troubleshooting, we want to preserve what the datapath observed
 	// when it made the policy decision.
-	resolveIdentityConflict := func(identity identity.NumericIdentity) uint32 {
+	resolveIdentityConflict := func(userspaceID identity.NumericIdentity, isLocalEndpoint bool) uint32 {
 		// if the datapath did not provide an identity (e.g. FROM_LXC trace
 		// points), use what we have in the user-space cache
-		userspaceSecurityIdentity := identity.Uint32()
-		if datapathSecurityIdentity == 0 {
-			return userspaceSecurityIdentity
+		datapathID := identity.NumericIdentity(datapathSecurityIdentity)
+		if datapathID == identity.IdentityUnknown {
+			return userspaceID.Uint32()
 		}
 
-		if datapathSecurityIdentity != userspaceSecurityIdentity {
-			r.log.WithFields(logrus.Fields{
-				logfields.Identity:    datapathSecurityIdentity,
-				logfields.OldIdentity: userspaceSecurityIdentity,
-				logfields.IPAddr:      ip,
-			}).Debugf("stale identity observed")
+		if datapathID != userspaceID {
+			if context.TraceObservationPoint == pb.TraceObservationPoint_TO_OVERLAY &&
+				ip == context.SrcIP && datapathID.Uint32() == context.SrcLabelID &&
+				datapathID == identity.ReservedIdentityRemoteNode &&
+				userspaceID == identity.ReservedIdentityHost {
+				// Ignore
+				//
+				// When encapsulating a packet for sending via the overlay network, if the source
+				// seclabel = HOST_ID, then we reassign seclabel with LOCAL_NODE_ID and then send
+				// a trace notify.
+			} else if context.TraceObservationPoint == pb.TraceObservationPoint_FROM_ENDPOINT &&
+				ip == context.SrcIP && datapathID.Uint32() == context.SrcLabelID &&
+				(datapathID == identity.ReservedIdentityHealth || !datapathID.IsReservedIdentity()) &&
+				userspaceID.IsWorld() {
+				// Ignore
+				//
+				// Sometimes packets from endpoint link-local addresses are intercepted by
+				// cil_from_container. Because link-local addresses are not stored in the IP cache,
+				// Hubble assigns them WORLD_ID.
+			} else if context.TraceObservationPoint == pb.TraceObservationPoint_FROM_HOST &&
+				ip == context.SrcIP && datapathID.Uint32() == context.SrcLabelID &&
+				datapathID.IsWorld() && userspaceID == identity.ReservedIdentityKubeAPIServer {
+				// Ignore
+				//
+				// When a pod sends a packet to the Kubernetes API, its IP is masqueraded and then
+				// when it receives a response and the masquerade is reversed, cil_from_host
+				// determines that the source ID is WORLD_ID because there is no packet mark.
+			} else if (context.TraceObservationPoint == pb.TraceObservationPoint_FROM_HOST ||
+				context.TraceObservationPoint == pb.TraceObservationPoint_TO_OVERLAY) &&
+				ip == context.SrcIP && datapathID.Uint32() == context.SrcLabelID &&
+				isLocalEndpoint && userspaceID == identity.ReservedIdentityHost {
+				// Ignore
+				//
+				// When proxied packets (via Cilium DNS proxy) are sent from the host their source
+				// IP is that of the host, yet their security identity is retained from the
+				// original source pod.
+			} else if context.TraceObservationPoint == pb.TraceObservationPoint_TO_ENDPOINT &&
+				ip == context.SrcIP && datapathID.Uint32() == context.SrcLabelID &&
+				!datapathID.IsReservedIdentity() &&
+				(userspaceID == identity.ReservedIdentityHost || userspaceID == identity.ReservedIdentityRemoteNode) {
+				// Ignore
+				//
+				// When proxied packets (via Cilium DNS proxy) are received by the destination
+				// host their source IP is that of the proxy, yet their security identity is
+				// retained from the original source pod. This is a similar case to #4, but on the
+				// receiving side.
+			} else {
+				r.log.WithFields(logrus.Fields{
+					logfields.Identity:    datapathID.Uint32(),
+					logfields.OldIdentity: userspaceID.Uint32(),
+					logfields.IPAddr:      ip,
+				}).Debugf("stale identity observed")
+			}
 		}
 
-		return datapathSecurityIdentity
+		return datapathID.Uint32()
 	}
 
 	// for local endpoints, use the available endpoint information
 	if r.endpointGetter != nil {
 		if ep, ok := r.endpointGetter.GetEndpointInfo(ip); ok {
-			epIdentity := resolveIdentityConflict(ep.GetIdentity())
+			epIdentity := resolveIdentityConflict(ep.GetIdentity(), true)
 			e := &pb.Endpoint{
 				ID:        uint32(ep.GetID()),
 				Identity:  epIdentity,
@@ -90,7 +145,7 @@ func (r *EndpointResolver) ResolveEndpoint(ip netip.Addr, datapathSecurityIdenti
 	var namespace, podName string
 	if r.ipGetter != nil {
 		if ipIdentity, ok := r.ipGetter.LookupSecIDByIP(ip); ok {
-			numericIdentity = resolveIdentityConflict(ipIdentity.ID)
+			numericIdentity = resolveIdentityConflict(ipIdentity.ID, false)
 		}
 		if meta := r.ipGetter.GetK8sMetadata(ip); meta != nil {
 			namespace, podName = meta.Namespace, meta.PodName
--- a/pkg/hubble/parser/sock/parser.go
+++ b/pkg/hubble/parser/sock/parser.go
@@ -91,8 +91,14 @@ func (p *Parser) Decode(data []byte, decoded *flowpb.Flow) error {
 	srcIP, _ := ippkg.AddrFromIP(epIP)
 	srcPort := uint16(0) // source port is not known for TraceSock events
 
-	srcEndpoint := p.epResolver.ResolveEndpoint(srcIP, 0)
-	dstEndpoint := p.epResolver.ResolveEndpoint(dstIP, 0)
+	datapathContext := common.DatapathContext{
+		SrcIP:      srcIP,
+		SrcLabelID: 0,
+		DstIP:      dstIP,
+		DstLabelID: 0,
+	}
+	srcEndpoint := p.epResolver.ResolveEndpoint(srcIP, 0, datapathContext)
+	dstEndpoint := p.epResolver.ResolveEndpoint(dstIP, 0, datapathContext)
 
 	// On the reverse path, source and destination IP of the packet are reversed
 	isRevNat := decodeRevNat(sock.XlatePoint)
--- a/pkg/hubble/parser/threefour/parser.go
+++ b/pkg/hubble/parser/threefour/parser.go
@@ -186,8 +186,15 @@ func (p *Parser) Decode(data []byte, decoded *pb.Flow) error {
 	}
 
 	srcLabelID, dstLabelID := decodeSecurityIdentities(dn, tn, pvn)
-	srcEndpoint := p.epResolver.ResolveEndpoint(srcIP, srcLabelID)
-	dstEndpoint := p.epResolver.ResolveEndpoint(dstIP, dstLabelID)
+	datapathContext := common.DatapathContext{
+		SrcIP:                 srcIP,
+		SrcLabelID:            srcLabelID,
+		DstIP:                 dstIP,
+		DstLabelID:            dstLabelID,
+		TraceObservationPoint: decoded.TraceObservationPoint,
+	}
+	srcEndpoint := p.epResolver.ResolveEndpoint(srcIP, srcLabelID, datapathContext)
+	dstEndpoint := p.epResolver.ResolveEndpoint(dstIP, dstLabelID, datapathContext)
 	var sourceService, destinationService *pb.Service
 	if p.serviceGetter != nil {
 		sourceService = p.serviceGetter.GetServiceByAddr(srcIP, srcPort)


pkg/hubble/parser/common/endpoint.go
```diff
@@ -45,23 +53,81 @@ func (r *EndpointResolver) ResolveEndpoint(ip netip.Addr, datapathSecurityIdenti
...
+				// When proxied packets (via Cilium DNS proxy) are sent from the host their source
+				// IP is that of the host, yet their security identity is retained from the
+				// original source pod.
+			} else if context.TraceObservationPoint == pb.TraceObservationPoint_TO_ENDPOINT &&
+				ip == context.SrcIP && datapathID.Uint32() == context.SrcLabelID &&
+				!datapathID.IsReservedIdentity() &&
+				(userspaceID == identity.ReservedIdentityHost || userspaceID == identity.ReservedIdentityRemoteNode) {
+				// Ignore
+				//
+				// When proxied packets (via Cilium DNS proxy) are recieved by the destination
```
gandro: Linter found a typo here: "`recieved` is a misspelling of `received` (misspell)"

pkg/hubble/parser/common/endpoint.go
```diff
@@ -45,23 +53,81 @@ func (r *EndpointResolver) ResolveEndpoint(ip netip.Addr, datapathSecurityIdenti
...
+		if datapathID != userspaceID {
+
+			localEndpoint := false
+			eps := r.endpointGetter.GetEndpoints()
+			for ep := range eps {
+				id, _ := ep.GetSecurityIdentity()
+				if id.ID == datapathID {
+					localEndpoint = true
+				}
+			}
```
gandro: This is a relatively expensive operation to invoke on each flow "just" for a debug log. I'm also not sure it's correct, a local endpoint can share a security identity with a remote endpoint after all (e.g. two replicas of the same pod on different nodes), in which case this would yield false positives.

Typically, we perform a lookup based on the IP address in the endpoint manager to determine if a IP belongs to a local endpoint. Though I'm not sure if this would work here for the link local case you mentioned.

gandro (commented):
This is an amazing piece of work, thanks so much for the investigation and detailed description.

Overall, I think the cases you found all sound reasonable. But I do agree that there is some fragility introduced here, the code is very likely to desync with the datapath over time.

Maybe another option to explore would be if we can find some simple heuristic that works "well enough" to cover most cases. Looking at all the different cases you mentioned, all of the either involve a reserved identity or a world identity.

In the end, assuming there will never be perfect parity due to the information asymmetry between datapath and userspace, I think we need to decide if we want to live with false positives or false negatives.

At the moment, we have lots of false positives (i.e. "stale" log lines for entries that are not actually stale), but no false negatives, because we just log everything.

The current solution in this PR at the removes most of the false positives, while also probably introducing some rare false negatives. It's clearly an improvement, but it has the downside that it's a complex heuristic that can become obsolete if the datapath changes.

A extreme on the other side could be to only log  it if both identities are neither reserved nor world, since it seems that this would cover most of the cases here too. In such an implementation, we would have as few false positives as your current solution, at the cost of some false negatives (i.e. we might actually miss some stale IPCache entries if we do that).

I don't really where we should land on the false positive vs false negative tradeoff. I'm tempted to accept a decent amount of false negatives if the false positive rate low. In the end, any logging with lots of false positives will not provide much value, because it just comes too noisy.

I'll also bring this up with some other @cilium/sig-hubble members.

pkg/hubble/parser/common/endpoint.go
chancez: If the expensiveness is a concern, we _could_ check the log level and only do the check if it's debug maybe?

leblowl: I agree, I don't think this actually works

leblowl:
Thanks for the review! I agree, I think a simpler and more general heuristic would be easier to maintain and hopefully work well enough. It got me thinking: what are the cases that we are trying to notice with this log warning? I think there is perhaps one case where the BPF map is not updated quickly enough and userspace has a different view of the ip/identity mapping than the datapath has. Is that the only case? If it is, another option might be putting a special trace/debug event after each call to `lookup_ip{4,6}_remote_endpoint` and only warning about stale identities from those events. I'm not sure if this warning is important enough to justify an additional trace/debug event.

gandro:
> It got me thinking: what are the cases that we are trying to notice with this log warning?

The main motivation was to collect some data to understand why datapath and Hubble would diverge. Unfortunately, no one ever followed up on that until you did, so thanks a lot for that!

> I think there is perhaps one case where the BPF map is not updated quickly enough and userspace has a different view of the ip/identity mapping than the datapath has

Yeah, I think besides the logical discrepancies that you documented, this map sync lag the only major reason why we would observe "stale identities".

>  If it is, another option might be putting a special trace/debug event after each call to lookup_ip{4,6}_remote_endpoint and only warning about stale identities from those events.

I'm not sure what you mean by this? Could you elaborate a bit?


On a more high-level note: We've discussed this PR with some members of the Hubble team and we think the approach PR strikes a good balance between false positives and negatives. It provides a lot of value to have those cases documented in the code, so having them as separate cases is probably more meaningful than having a heuristic. So we're fine to merge this as is, assuming the issue with the endpoint manager is fixed.

lmb:
I'm converting this to Draft since there is outstanding feedback to be addressed. Please click on "Ready for review" once you've made the necessary changes.

leblowl:
Thanks!

leblowl:
@gandro

Sorry for the delay.

> I'm not sure what you mean by this? Could you elaborate a bit?

The idea is similar to this section of code: https://github.com/cilium/cilium/blob/9aa5068283c29864b625a7f7b983a379bc33fc8c/bpf/bpf_host.c#L486-L504

Whenever we call `lookup_ip4_remote_endpoint` or `lookup_ip6_remote_endpoint` in the BPF code, we can emit debug messages like `DBG_IP_ID_MAP_SUCCEED4` with the IP cache security ID. These debug messages could come directly after `lookup_ip{4,6}_remote_endpoint` calls and so they should always tell us if the IP cache BPF map is stale and thus wouldn't require special conditions in the Hubble code. However, these debug messages wouldn't track other parts of the datapath where we get the security ID from the packet mark for example.

I can fix up the local endpoint check and address the test failures.

gandro:
I see, thanks for the clarification! I think it becomes more clear to me now that the log messages could be understood to serve two use-cases:

1. Serve as a indicator that there was an actual mismatch between datapath's and userspace's view of the IPCache. I think this use-case is already pretty well covered by `DBG_IP_ID_MAP_SUCCEED4` debug message and the `cilium ip list` (showing userspace state) and `cilium bpf ipcache list` (showing datapath state) commands.
2. Serve as an indicator that the way Hubble determines the security identity of an IP does not align with what the datapath does. This is particularly relevant when annotating trace events where the datapath does not perform an IPCache lookup, such as e.g. in the FROM_LXC trace point. In that case, the metadata annotated by Hubble might be misleading.

As outlined in the amazing PR description, I think you've identified quite a few gaps where Hubble annotates potentially misleading data. Therefore one could argue that the log message has now mostly outlived it's purpose at this point. Instead, the it might be more useful to understand how can we fix the identified mismatches such that Hubble correctly annotates flows going forward.

Having said that, I'm more than happy to accept the overall PR (assuming fixed endpoint lookup and tests). Even if I mentioned that I believe the log message kind of became obsolete thanks to the fact that we now know where the discrepancies are coming from, there might still be more cases out there in the wild that we haven't discovered yet.

For the Hubble project in general, I think the plan forward here should be:

1. Fix the outstanding technical issues (i.e. endpoint check) with the PR, but otherwise merge it more or less as is
2. Document the discovered mismatches as "known bugs", e.g. in an separate issue
3. Investigate to what degree these known mismatches are fixable. If they are, potentially fix them. If they are not easily fixable, we have at least documented them

leblowl:
Thanks! That makes sense. I think I've fixed up the PR.

gandro (approved):
Awesome work, this looks good now! Thanks a lot

joestringer:
(Awaiting two unresolved conversations to be resolved before merging)

gandro:
Mine has been resolved. cc @chancez if you want to take a look

chancez:
To be honest, most of it's incomprehensible to me, so I'll just remove myself as a reviewer 😅

chancez (approved):
I guess I cannot remove myself, so ✅

joestringer:
@chancez your review was requested specifically on behalf of hubble, so I think the minimum is just the mechanical aspects of changes to those specific files. It looks like @gandro already engaged from a broader perspective so we're probably good on the general feedback side. Thanks for the review!

pkg/hubble/parser/common/endpoint.go
joestringer: @gandro @chancez I guess this was the remaining item that was not yet resolved. The feedback is marked as "outdated", is it still applicable?

gandro: This was resolved, yeah!

gandro:
Let's backport this in order to improve the situation in CI https://github.com/cilium/cilium/issues/15283

PR #27918, opened 2023-09-04
docs: egressgw: document incompatibility with Clustermesh

julianwiedmann:
The current way of subscribing to node events is not compatible with Clustermesh (see https://github.com/cilium/cilium/pull/25794). Reflect this in the documentation.

### DIFF
--- a/Documentation/network/egress-gateway.rst
+++ b/Documentation/network/egress-gateway.rst
@@ -76,6 +76,9 @@ Because egress gateway isn't compatible with identity allocation mode ``kvstore`
 you must use Kubernetes as Cilium's identity store (``identityAllocationMode``
 set to ``crd``). This is the default setting for new installations.
 
+Egress gateway is not compatible with the Cluster Mesh feature. The gateway selected
+by an egress gateway policy must be in the same cluster as the selected pods.
+
 Egress gateway is not supported for IPv6 traffic.
 
 Enable egress gateway


julianwiedmann:
@YutaroHayakawa do you know if there is any way of determining whether a node is set up for Clustermesh? So that we could at least give users a warning similar to https://github.com/cilium/cilium/blob/0da3f7e29e9bfaa860d652a7be7598a556cf9647/pkg/egressgateway/manager.go#L212-L215

bimmlerd (approved):

Documentation/network/egress-gateway.rst
```diff
@@ -76,6 +76,9 @@ Because egress gateway isn't compatible with identity allocation mode ``kvstore`
 you must use Kubernetes as Cilium's identity store (``identityAllocationMode``
 set to ``crd``). This is the default setting for new installations.
 
+Egress gateway is not compatible with Clustermesh. The gateway selected by an
```
learnitall: We refer to clustermesh a couple of different ways in the docs, but I believe the most common is `Cluster Mesh`.

```suggestion
Egress gateway is not compatible with Cluster Mesh. The gateway selected by an
```

learnitall (changes requested):
One nit, but otherwise LGTM!

YutaroHayakawa:
> @YutaroHayakawa do you know if there is any way of determining whether a node is set up for Clustermesh? So that we could at least give users a warning similar to

Umm, I guess we can depend on `ClusterMesh` object as optional and check if it is `nil` or not. Like

```
type EgressGatewayParams struct {
  cell.In

  cm *ClusterMesh `optional:"true"`
}

func NewEgressGateway(params EgressGatewayParams) (EgressGateway, error) {
  if params.cm != nil {
    return nil, fmt.Errorf("Egress Gateway is incompatible with ClusterMesh")
  }
  ...
}
```

@giorio94 Do you have any other good ideas?

YutaroHayakawa (approved):

jibi (approved):

Documentation/network/egress-gateway.rst
julianwiedmann: :+1: , fixed up.

@cilium/sig-clustermesh maybe y'all want to harmonize the remaining occurrences in the Docs?

learnitall (approved):
LGTM, thank you!

julianwiedmann:
@youngnick I think we don't need the full CI here, but feel free to disagree :+1:

youngnick:
I agree, merging.

giorio94:
> > @YutaroHayakawa do you know if there is any way of determining whether a node is set up for Clustermesh? So that we could at least give users a warning similar to
> 
> Umm, I guess we can depend on `ClusterMesh` object as optional and check if it is `nil` or not. Like
> 
> ```
> type EgressGatewayParams struct {
>   cell.In
> 
>   cm *ClusterMesh `optional:"true"`
> }
> 
> func NewEgressGateway(params EgressGatewayParams) (EgressGateway, error) {
>   if params.cm != nil {
>     return nil, fmt.Errorf("Egress Gateway is incompatible with ClusterMesh")
>   }
>   ...
> }
> ```
> 
> @giorio94 Do you have any other good ideas?

Your proposal makes sense to me. The optional tag is not strictly required, as the ClusterMesh object is always provided, although it is nil if not initialized.

PR #27923, opened 2023-09-04
envoy: fix lb backend endpoint calculation

mhofstetter:
Currently, mapping loadbalancing backends to Envoy endpoints contains a bug that the LbEndpoints are kept/appended over the
whole backendMap. Therefore, later endpoints will contain the LBEndpoints of all previous backends.

This commit fixes this by putting the variable `lbEndpoints` into the right scope (per port).

### DIFF
--- a/pkg/envoy/ciliumenvoyconfig_test.go
+++ b/pkg/envoy/ciliumenvoyconfig_test.go
@@ -7,18 +7,25 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/netip"
+	"testing"
 
 	cilium "github.com/cilium/proxy/go/cilium/api"
 	envoy_config_cluster "github.com/cilium/proxy/go/envoy/config/cluster/v3"
 	envoy_config_core "github.com/cilium/proxy/go/envoy/config/core/v3"
+	endpointv3 "github.com/cilium/proxy/go/envoy/config/endpoint/v3"
 	envoy_config_listener "github.com/cilium/proxy/go/envoy/config/listener/v3"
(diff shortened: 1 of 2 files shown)


pkg/envoy/ciliumenvoyconfig_test.go
```diff
@@ -892,3 +899,48 @@ func (s *JSONSuite) TestListenersAddedOrDeleted(c *C) {
...
+				L3n4Addr: *loadbalancer.NewL3n4Addr(loadbalancer.TCP, types.AddrClusterFrom(testAddr, 0), 15000, 3),
+			},
+		},
+	}
+
+	endpoints := getEndpointsForLBBackends(serviceName, backends)
+	assert.Len(t, endpoints, 4)
+
+	var allClusterNames []string
+	for i := 0; i < len(endpoints); i++ {
```
tommyp1ckles: nit: could just be `for i := range endpoints`

tommyp1ckles:
Looks good, sorry for the delay, was there an issue for this bug? I'm curious what the effects where?

tommyp1ckles (approved):

pkg/envoy/ciliumenvoyconfig_test.go
mhofstetter: changed

mhofstetter:
rebased to `main` without any further changes

mhofstetter:
> Looks good, sorry for the delay, was there an issue for this bug?

no there isn't.

tommyp1ckles:
@mhofstetter This probably needs backports right? (looks like the same issue might exist back to at least v1.13).

PR #27931, opened 2023-09-04
docs: Add Makefile and documentation for "fast" development targets

aanm:
This commit introduces documentation for "fast" development targets in Cilium. The "fast" targets allow for quicker development workflows by building Cilium locally and deploying it on an existing kind cluster. This approach eliminates the need to create and deploy Docker images each time a source code change is made.

Additionally, the documentation has been updated to include these new Makefile targets and provides a quick start guide with essential information for working with Cilium.

When a cluster is already running with a Cilium container, we can simply volume mount the binaries into the container and restart the pod to pick up the changes. A similar approach can be used for C code, with the caveat that the bpf/lib directory needs write permissions by Cilium to generate the C Header features file, making it impossible to use the volume as read-only. Since all nodes share the same kernel features, there won't be any conflicts when multiple nodes write to the same volume mount.

### DIFF
--- a/Documentation/contributing/development/dev_setup.rst
+++ b/Documentation/contributing/development/dev_setup.rst
@@ -10,35 +10,52 @@
 Development Setup
 =================
 
-Dev Container
-~~~~~~~~~~~~~
+This page provides an overview of different methods for efficient
+development on Cilium. Depending on your needs, you can choose the most
+suitable method.
 
-Cilium provides `Dev Container <https://code.visualstudio.com/docs/devcontainers/containers>`_ configuration for Visual Studio Code Remote Containers
-and `Github Codespaces <https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/introduction-to-dev-containers>`_.
-This allows you to use a preconfigured development environment in the cloud or locally.
-The container is based on the official Cilium builder image and provides all the dependencies
-required to build Cilium.
+Quick Start
+-----------
 
-.. note::
+If you're in a hurry, here are the essential steps to get started:
 
-    The current Dev Container is running as root. Non-root user support requires non-root
-    user in Cilium builder image, which is related to :gh-issue:`23217`.
+On Linux:
+
+1. ``make kind`` - Provisions a Kind cluster.
+2. ``make kind-install-cilium-fast`` - Installs Cilium on the Kind cluster.
+3. ``make kind-image-fast`` - Builds Cilium and deploys it.
+
+On any OS:
+
+1. ``make kind`` - Provisions a Kind cluster.
+2. ``make kind-image`` - Builds Docker images.
+3. ``make kind-install-cilium`` - Installs Cilium on the Kind cluster.
+
+Detailed Instructions
+---------------------
+
+Depending on your specific development environment and requirements, you
+can follow the detailed instructions below.
 
 Verifying Your Development Setup
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Assuming you have Go installed, you can quickly verify many elements of your
-development setup by running:
+development setup by running the following command:
 
 .. code-block:: shell-session
 
     $ make dev-doctor
 
-Requirements
-~~~~~~~~~~~~
+Depending on your end-goal, not all dependencies listed are required to develop
+on Cilium. For example, "Ginkgo" is not required if you want to improve our
+documentation. Thus, do not consider that you need to have all tools installed.
 
-You need to have the following tools available in order to effectively
-contribute to Cilium:
+Version Requirements
+~~~~~~~~~~~~~~~~~~~~
+
+If using these tools, you need to have the following versions from them
+in order to effectively contribute to Cilium:
 
 +--------------------------------------------------------------+------------------------------+-----------------------------------------------------------------+
 | Dependency                                                   | Version / Commit ID          | Download Command                                                |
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -563,6 +563,61 @@ kind-image: ## Build cilium and operator images and import them into kind.
...
+	# https://github.com/cilium/cilium-cli/issues/1070
+	$(CILIUM_CLI) install \
+		--chart-directory=$(ROOT_DIR)/install/kubernetes/cilium \
+		--helm-values=$(ROOT_DIR)/contrib/testing/kind-common.yaml \
+		--helm-values=$(ROOT_DIR)/contrib/testing/kind-fast.yaml \
+		--version= \
+		>/dev/null 2>&1 &
+
+.PHONY: kind-image-fast
+kind-image-fast: ## Build cilium, operator, clustermesh-apiserver and mount them into a volume
```
qmonnet:
```suggestion
kind-image-fast: ## Build cilium, operator, clustermesh-apiserver and mount them into a volume for all available Kind clusters.
```

Makefile
```diff
@@ -563,6 +563,61 @@ kind-image: ## Build cilium and operator images and import them into kind.
 	$(MAKE) kind-image-agent
 	$(MAKE) kind-image-operator
 
+.PHONY: kind-install-cilium-fast
+kind-install-cilium-fast: kind-ready ## Install a local Cilium version into the cluster.
```
qmonnet, line 567: We could maybe refactor this with `kind-install-cilium` by making a function and pass the Helm value file to use as argument. Although, I'm also fine with the current version.

nebril (approved):

Documentation/contributing/development/dev_setup.rst
```diff
@@ -90,16 +107,25 @@ You can find the setup for a `kind <https://kind.sigs.k8s.io/>`_ environment in
...
 * ``make kind``: Creates a kind cluster based on the configuration passed in.
-  For more information, see _`Configuration for clusters`.
+  For more information, see `configurations_for_clusters`.
+* ``make kind-down``: Tears down and deletes the cluster.
+
+Depending on your environment you can build Cilium by using the following
+makefile targets:
+
+For Linux and Mac OS
+^^^^^^^^^^^^^^^^^^^^
```
qmonnet, line 120: I realise our theme does not make a visual distinction between level 3 and level 4 headings :grimacing:

Makefile
```diff
@@ -563,6 +563,61 @@ kind-image: ## Build cilium and operator images and import them into kind.
...
+
+.PHONY: kind-image-fast
+kind-image-fast: ## Build cilium, operator, clustermesh-apiserver and mount them into a volume
+	$(eval dst:=/cilium-binaries)
+	$(MAKE) -C cilium
+	$(MAKE) -C daemon
+	$(MAKE) -C operator cilium-operator-generic
+	$(MAKE) -C clustermesh-apiserver
+
+	for cluster_name in $(shell kind get clusters); do \
```
qmonnet: How about allowing users to specify the clusters to use, just in case someone has another unrelated Kind cluster running? Something like:

```suggestion
	for cluster_name in $${KIND_CLUSTERS:-$(shell kind get clusters)}; do \
```

Documentation/contributing/development/dev_setup.rst
```diff
@@ -10,35 +10,52 @@
...
 
 .. code-block:: shell-session
 
     $ make dev-doctor
 
-Requirements
-~~~~~~~~~~~~
+Depending on your end-goal, not all dependencies listed are required to develop
+on Cilium. For example, "Ginkgo" is not required if you want to improve our
+documentation. Thus, do not consider that you need to have all tools installed.
```
qmonnet, line 52: Ideally, the follow-up to this would be to add details on what tasks the deps relate to :innocent:.

qmonnet (approved):
Looks all good. Some suggestions only, feel free to discard.

Thanks for this!

qmonnet:
Is it something we could backport, so we could use it when troubleshooting issues on the older branches, as well?

Documentation/contributing/development/dev_setup.rst
```diff
@@ -108,11 +134,26 @@ Cilium images:
   Use if only the agent image needs to be rebuilt for faster iteration.
 * ``make kind-install-cilium``: Installs Cilium into the cluster using the
   Cilium CLI.
-* ``make kind-down``: Tears down and deletes the cluster.
 
-The preceding list includes the most used commands for convenience. For more
+The preceding list includes the most used commands for **convenience**. For more
 targets, see the ``Makefile``.
```
qmonnet:
```suggestion
The preceding list includes the most used commands for **convenience**. For more
targets, see the ``Makefile`` (or simply run ``make help``).
```

Documentation/contributing/development/dev_setup.rst
aanm: I agree :sweat_smile: :innocent:

Documentation/contributing/development/dev_setup.rst
aanm: It doesn't? It looked to me that the level 4 had a smaller font size.

Makefile
aanm: Yes, I thought about it but it sounded more messy.

Documentation/contributing/development/dev_setup.rst
qmonnet: Alright, 18.75px vs. 17.25px. The difference is thin but it exists, I had missed it. Not something for this PR anyway.

qmonnet (approved):
Looks good, thank you!

PR #27933, opened 2023-09-05
can create the directory for the customized cni conf and remove the cni conf file in cleanup command

sofat1989:
Usually we will write the customized cni conf file and write it into the directory `/etc/cni/net.d` in the host. For example, 
```
- mountPath: /host/etc/cni/net.d
          name: etc-cni-netd

- hostPath:
          path: /etc/cni/net.d
          type: DirectoryOrCreate
        name: etc-cni-netd
```
So the value of write-cni-conf-when-ready is `/host/etc/cni/net.d/xxx-cilium.conf`.
In the cleanup command, it will not remove this file. 

And sometimes we need to specify  write-cni-conf-when-ready  to any pathes, and the directory may not exist. So the writing cni file will fail. 

The commit is to solve these two issues above.

```release-note
can create the directory for the customized cni conf and remove the cni conf file in cleanup command
```

### DIFF
--- a/cilium/cmd/cleanup.go
+++ b/cilium/cmd/cleanup.go
@@ -41,9 +41,10 @@ working condition after uninstalling the Cilium agent.`,
 }
 
 var (
-	cleanAll bool
-	cleanBPF bool
-	force    bool
+	cleanAll      bool
+	cleanBPF      bool
+	force         bool
+	customCNIConf string
 )
 
 const (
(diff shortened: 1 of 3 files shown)


gandro:
cc @squeed as our resident CNI configuration file expert

asauber:
The general idea here LGTM, and this does look necessary, given that we have a config option for the path.

A better release note would be:

```
create directory for custom CNI config path on startup
remove custom CNI config file with cilium cleanup command
```

asauber (approved):

youngnick (approved):

gandro (approved):
Helm looks good to me. I'll defer the logic changes to the other reviewers

Issue #17259, opened 2021-08-30
Cilium refuses to start with a network interface ifindex that exceeds max(uint16)


clrxbl:
## Bug report

If Cilium is started with the `--devices` parameter set to a specific network interface that has an ifindex higher than max(uint16), Cilium refuses to start by crashing with a fatal error message. Temporary workaround for this is to restart the server where the cilium-agent is trying to run on, but this is obviously not acceptable in the long run.

**General Information**

- Cilium version:
```
Client: 1.10.3 4145278 2021-07-15T16:11:03+02:00 go version go1.16.6 linux/amd64
Daemon: 1.10.3 4145278 2021-07-15T16:11:03+02:00 go version go1.16.6 linux/amd64
```
- Kernel version
```
Linux dev1 5.8.0-59-generic #66~20.04.1-Ubuntu SMP Thu Jun 17 11:14:10 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
Linux dev2 5.8.0-59-generic #66~20.04.1-Ubuntu SMP Thu Jun 17 11:14:10 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
Linux dev3 5.11.0-27-generic #29~20.04.1-Ubuntu SMP Wed Aug 11 15:58:17 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```
- Orchestration system version in use
```
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.1", GitCommit:"632ed300f2c34f6d6d15ca4cef3d3c7073412212", GitTreeState:"archive", BuildDate:"2021-08-19T21:28:12Z", GoVersion:"go1.17", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2+k3s1", GitCommit:"5a67e8dc473f8945e8e181f6f0b0dbbc387f6fca", GitTreeState:"clean", BuildDate:"2021-06-21T20:52:44Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
```
- Link to relevant artifacts (policies, deployments scripts, ...): Not applicable.
- Generate and upload a system zip: Not applicable.

**How to reproduce the issue**

1. You start Cilium with the `--devices` parameter set to a specific network interface that has an ifindex that exceeds max(uint16)
2. Cilium crashes with the following message:

```
[...]

PR #27953, opened 2023-09-05
metrics: add a metric for max observed endpoint ifindex

asauber:
Adds a metric `endpoint_max_ifindex`, which is the current maximum
interface index for all endpoints. The metric is updated during the
periodic invocation of `syncHostIPs`.

This metric can be used to determine if the interface index for the next
Pod will exceed Cilium's limit of max(uint16) on older kernels. On
kernels which do not provide ifindex via the FIB, Cilium must store the
ifindex in the CT map, and 16 bits are reserved per entry for this
purpose. This presents a limit of 65535 lifetime interfaces, which can
be approached quickly with sufficient pod churn.

Users who are subject to this limitation (typically a kernel version
less than 5.10), are advised to reboot or roll the host in this case.

Its default-enabled status is dynamic. On kernels which provide ifindex
via the FIB, the metric is disabled (since Cilium's ifindex limits are
not subject to max(uint16) in this case).

It can be enabled with the following Helm values:

```yaml
prometheus:
  enabled: true
  metrics:
    - +cilium_endpoint_max_ifindex
```

On kernels which do not provide ifindex via the FIB, the metric is enabled by default.

### Dynamic default based on Kernel version

On Linux 4.19 without metric explicitly enabled

```
$ cat contrib/testing/kind-values.yaml | grep -A 1 prom
prometheus:
  enabled: true

$ curl 2>/dev/null 172.18.0.4:9962/metrics | grep ifindex
# HELP cilium_endpoint_max_ifindex Maximum interface index observed for existing endpoints
# TYPE cilium_endpoint_max_ifindex gauge
cilium_endpoint_max_ifindex 7

$ uname -r
kind-4.19
```

On Linux 6.4 without metric explicitly enabled

```
$ cat contrib/testing/kind-values.yaml | grep -A 1 prom
prometheus:
  enabled: true
  
$ curl 2>/dev/null 172.19.0.3:9962/metrics | grep ifindex

$ 
```

On Linux 6.4 with metric explicitly enabled

```
$ cat contrib/testing/kind-values.yaml | grep -A 3 prom
prometheus:
  enabled: true
  metrics:
    - +cilium_endpoint_max_ifindex

$ curl 2>/dev/null 172.19.0.3:9962/metrics | grep ifindex
# HELP cilium_endpoint_max_ifindex Maximum interface index observed for existing endpoints
# TYPE cilium_endpoint_max_ifindex gauge
cilium_endpoint_max_ifindex 11

$ uname -r
6.4.12-arch1-1
```

### Longer-running example

Roll a pod or two

```
$ k delete pod -n kube-system          coredns-5d78c9869d-5w8pc
pod "coredns-5d78c9869d-5w8pc" deleted
```

and the `ifindex` increases

```
$ curl 172.19.0.2:9962/metrics 2>/dev/null | grep ifindex
# HELP cilium_endpoint_max_ifindex Maximum interface index observed for existing endpoints
# TYPE cilium_endpoint_max_ifindex gauge
cilium_endpoint_max_ifindex 43
```

Addresses: https://github.com/cilium/cilium/issues/17259
Addresses: https://github.com/cilium/cilium/issues/16260

```release-note
metrics: add a metric for max observed endpoint ifindex
```

### DIFF
--- a/Documentation/observability/metrics.rst
+++ b/Documentation/observability/metrics.rst
@@ -270,10 +270,19 @@ Endpoint
 Name                                         Labels                                             Default    Description
 ============================================ ================================================== ========== ========================================================
 ``endpoint``                                                                                    Enabled    Number of endpoints managed by this agent
+``endpoint_max_ifindex``                                                                        Disabled   Maximum interface index observed for existing endpoints
 ``endpoint_regenerations_total``             ``outcome``                                        Enabled    Count of all endpoint regenerations that have completed
 ``endpoint_regeneration_time_stats_seconds`` ``scope``                                          Enabled    Endpoint regeneration time stats
 ``endpoint_state``                           ``state``                                          Enabled    Count of all endpoints
 ============================================ ================================================== ========== ========================================================
+
+The default enabled status of ``endpoint_max_ifindex`` is dynamic. On earlier
+kernels (typically with version lower than 5.10), Cilium must store the
+interface index for each endpoint in the conntrack map, which reserves 16 bits
+for this field. If Cilium is running on such a kernel, this metric will be
+enabled by default. It can be used to implement an alert if the ifindex is
+approaching the limit of 65535. This may be the case in instances of
+significant Endpoint churn.
 
 Services
 ~~~~~~~~
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -19,6 +19,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/cilium/cilium/api/v1/models"
+	"github.com/cilium/cilium/pkg/datapath/linux/probes"
 	"github.com/cilium/cilium/pkg/metrics/metric"
 	"github.com/cilium/cilium/pkg/promise"
 	"github.com/cilium/cilium/pkg/version"
@@ -257,6 +258,9 @@ var (
 	// Endpoint is a function used to collect this metric.
 	// It must be thread-safe.
 	Endpoint metric.GaugeFunc
+
+	// EndpointMaxIfindex is the maximum observed interface index for existing endpoints
+	EndpointMaxIfindex = NoOpGauge
 
 	// EndpointRegenerationTotal is a count of the number of times any endpoint
 	// has been regenerated and success/fail outcome
@@ -570,6 +574,7 @@ type LegacyMetrics struct {
 	NodeConnectivityStatus           metric.Vec[metric.Gauge]
 	NodeConnectivityLatency          metric.Vec[metric.Gauge]
 	Endpoint                         metric.GaugeFunc
+	EndpointMaxIfindex               metric.Gauge
 	EndpointRegenerationTotal        metric.Vec[metric.Counter]
 	EndpointStateCount               metric.Vec[metric.Gauge]
 	EndpointRegenerationTimeStats    metric.Vec[metric.Observer]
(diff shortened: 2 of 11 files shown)


asauber:
The metric was tested on Linux 4.19 interactively using an `lvh` VM:
https://docs.cilium.io/en/stable/contributing/testing/e2e/#running-tests-in-vm

After bringing up the VM:

#### On the local machine

```
# build new container images for kind
make kind-image

# get our recently built cilium image onto the VM
docker save -o ciliumdev.tar localhost:5000/cilium/cilium-dev:local
scp -P 2224 ciliumdev.tar root@localhost:/host/cilium

# get our recently built cilium operator image onto the VM
docker save -o ciliumop.tar localhost:5000/cilium/operator-generic:local

# get cilium CLI on to the VM
scp -P 2224 $(which cilium) root@localhost:/host/ciliumcli
```

#### On the VM

```
alias cilium=/host/ciliumcli
alias k=kubectl

# load Cilium image
docker load -i ciliumdev.tar 
kind load docker-image localhost:5000/cilium/cilium-dev:local

# load Cilium operator image
docker load -i ciliumop.tar
kind load docker-image localhost:5000/cilium/operator-generic:local

cilium uninstall
cilium install \
	--chart-directory=/host/cilium/install/kubernetes/cilium \
	--helm-values=/host/cilium/contrib/testing/kind-values.yaml \
	--version=

k get pods -A -owide

**Comment by asauber** [comment]
Upon the first run of CI for this PR, an import cycle was detected. This commit fixes that import cycle, with details in the commit message:

https://github.com/cilium/cilium/pull/27953/commits/9c4a4318bf0804bc6ab31bc267aeb8aa445c8f00

**Review by nebril — APPROVED** [review]

**Review by borkmann — APPROVED** [review]

**Comment by borkmann** [comment]
@asauber could we add backport 1.14 label given the ifindex fix went also in there?

**Review by derailed — APPROVED** [review]
@asauber Nice work! LG. 
For my info: I am not sure what would this new metrics indicates if a system is nearing the max? and what can a user do about it?

**Comment by asauber** [comment]
@derailed yes, this can be used to implement an alert to determine if the cluster is nearing the max ifindex of 65535 (if on an affected kernel version)

The current suggested mitigation in that case is to roll or reboot the Node.

File: pkg/metrics/metrics.go [hunk 1]
```diff
@@ -570,6 +573,7 @@ type LegacyMetrics struct {
 	NodeConnectivityStatus           metric.Vec[metric.Gauge]
 	NodeConnectivityLatency          metric.Vec[metric.Gauge]
 	Endpoint                         metric.GaugeFunc
+	EndpointMaxIfindex               metric.Gauge
```
christarazi, line 577: Why isn't this `metrics.Vec[metric.Gauge]`? I presume the answer is that we don't have labels for this metric which is fine, but would that eliminate the possibility of using `cluster`, `namespace`, `instance` labels that are added automatically? (at least I think they are).

Documentation/observability/metrics.rst
christarazi, line 1: Can you add an upgrade note as well about the newly added metric?

pkg/metrics/metrics.go
christarazi, line 1: This commit can be squashed with the previous commit on this file. No need for separate commits that have diffs between them.

christarazi (changes requested):
Thanks for the PR!

A few comments below. One other thing is that I don't see any commit descriptions for the commits and all the context is instead in the PR. As a project, we are not opposed to a PR description, but we state in our [contrib guide](https://docs.cilium.io/en/stable/contributing/development/contributing_guide/#submitting-a-pull-request) that PRs should contain commits with a description. Plus, this is beneficial when bisecting / git blame.

pkg/metrics/metrics.go
asauber: I'll confirm whether or not those are available with some local testing.

Documentation/observability/metrics.rst
asauber: Yes, I like that idea.

learnitall (approved):
Looks great on docs! Thank you for adding the description as to the dynamic default status, awesome stuff.

pkg/metrics/metrics.go
asauber: Using `metric.Vec` isn't necessary in order to get those additional labels. They come from the prometheus scraper config. As an example, you will find config similar to the following for all metrics endpoints configured using the Prometheus operator:

```yaml
- job_name: serviceMonitor/kube-system/cilium-agent/0
  ...
  scrape_interval: 10s
  metrics_path: /metrics
  - source_labels:
    - __meta_kubernetes_namespace
    target_label: namespace
```

With a Grafana instance hooked up, we can see that many labels are added to this plain `metric.Gauge` with a typical configuration:
 
Screenshot_20230911_142730-1

nathanjsweet (approved):

christarazi (approved):

rgo3 (approved):

PR #27957, opened 2023-09-06
ci: trigger multi-pool conformance workflow using ariane

tklauser:
Convert the multi-pool conformance workflow to be triggered by Ariane using the `/ci-multi-pool` trigger phrase.

### DIFF
--- a/.github/workflows/conformance-multi-pool.yaml
+++ b/.github/workflows/conformance-multi-pool.yaml
@@ -1,23 +1,53 @@
-name: Conformance Multi Pool IPAM
+name: Conformance Multi Pool IPAM (ci-multi-pool)
 
 # Any change in triggers needs to be reflected in the concurrency group.
 on:
-  pull_request:
-    paths-ignore:
-      - 'Documentation/**'
-      - 'test/**'
-  push:
-    branches:
-      - main
-      - ft/main/**
-    paths-ignore:
-      - 'Documentation/**'
-      - 'test/**'
+  workflow_dispatch:
+    inputs:
+      PR-number:
+        description: "Pull request number."
+        required: true
+      context-ref:
+        description: "Context in which the workflow runs. If PR is from a fork, will be the PR target branch (general case). If PR is NOT from a fork, will be the PR branch itself (this allows committers to test changes to workflows directly from PRs)."
+        required: true
+      SHA:
+        description: "SHA under test (head of the PR branch)."
+        required: true
+      extra-args:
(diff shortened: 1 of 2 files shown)


tklauser:
Successful run: https://github.com/cilium/cilium/actions/runs/6095123851

.github/workflows/conformance-multi-pool.yaml
```diff
@@ -35,11 +65,12 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 120
     steps:
-      - name: Checkout main branch to access local actions
-        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+      - name: Checkout context ref (trusted)
+        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
         with:
-          ref: ${{ github.event.repository.default_branch }}
+          ref: ${{ inputs.context-ref || github.sha }}
```
bimmlerd, line 71: what does the `|| github.sha` mean in this case? If the `context-ref` is empty (can it be?) we fall back to checking out the SHA of the PR? Is that potentially unsafe if one somehow can force context-ref to be empty?

bimmlerd (commented):
generally LGTM, one question

.github/workflows/conformance-multi-pool.yaml
tklauser: `context-ref` is empty if the workflow isn't triggered on a PR but by a scheduled run, ref. https://github.com/cilium/cilium/pull/27957/files#diff-083aeca09a7733ba0341688cafa9223cc3a723c9894795d5e9a8371fc606707eR20-R22 `github.sha` will always point to a SHA on `main` in this case.

On PR runs `context-ref` is always required to be set: https://github.com/cilium/cilium/pull/27957/files#diff-083aeca09a7733ba0341688cafa9223cc3a723c9894795d5e9a8371fc606707eR10-R12

bimmlerd: Ah, makes sense. Thanks for the explanation!

bimmlerd (approved):

gandro (approved):

viktor-kurchenko (approved):

tklauser:
Looks like I forgot to add the `commit-status-start` job which leads to the check not showing up on PRs until the workflow has finished (e.g. visible on https://github.com/cilium/cilium/pull/27968). Should be fixed by https://github.com/cilium/cilium/pull/27969.

PR #27960, opened 2023-09-06
bgpv1: Introduce ResetServer method for BGP reconcilers to avoid shared metadata

rastislavs:
In order to make BGP reconcilers completely generic and decoupled from the BGP manager, each reconciler keeps its state metadata privately within its own datastructure instead of the common `ServerWithConfig` struct.

As this data needs to be reset whenever the BGP server is recreated, this change introduces a new `ResetServer()` reconciler method, which is called whenever the reconcilers should reset their cached state.

This is an alternative to the original idea drafted in https://github.com/cilium/cilium/pull/27568

### DIFF
--- a/pkg/bgpv1/manager/instance.go
+++ b/pkg/bgpv1/manager/instance.go
@@ -9,7 +9,6 @@ import (
 	"github.com/cilium/cilium/pkg/bgpv1/gobgp"
 	"github.com/cilium/cilium/pkg/bgpv1/types"
 	v2alpha1api "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2alpha1"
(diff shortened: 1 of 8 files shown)


harsimran-pabla:
I am not sure if this is still under consideration. Couple of issues with this approach
- State should be stored per ASN.
- State sharing between reconcilers is not possible anymore. This is a minor inconvenience, a possible workaround would be reconcilers referencing other reconcilers.

rastislavs:
@harsimran-pabla 
> I am not sure if this is still under consideration. Couple of issues with this approach

Did not have time to follow up on this yet, but I have to admit that I am starting to see more issues with this approach then when I was first thinking about it.

rastislavs:
closing this favor of https://github.com/cilium/cilium/pull/27568

PR #27963, opened 2023-09-06
ipam/multipool: Fix bug where allocator was unable to update CiliumNode

gandro:
This commit fixes an issue where the multi-pool allocator was unable to
update a CiliumNode resource because of concurrent writes. This
manifested in the following error being emitted repeatedly:

```
level=debug msg="Controller run failed" consecutiveErrors=48
error="failed to update spec: Operation cannot be fulfilled on
ciliumnodes.cilium.io \"kind-worker\": the object has been modified;
please apply your changes to the latest version and try again"
name=ipam-multi-pool-sync-kind-worker subsys=controller
uuid=12ba9a52-d36f-48fe-a7b7-3cf97c2cdb26
```

This would happen because the operator CiliumNode watcher does not call
the `Upsert` function if only the metadata (e.g. resource version,
labels, annotations, etc) of a node changes. This meant that the
allocator was working with a stale `resourceVersion` of the CiliumNode
object, causing any updates to fail until `Upsert` would be called again
because some non-metadata field changed.

This commit fixes that issue by having the controller fetch the most
recent version of the `CiliumNode` if the Kubernetes API reports that
there have been concurrent changes. This behavior matches the behavior
of the cluster-pool and ENI/Azure/AlibabaCloud implementation, which
already correctly fetched the resource again upon conflicts.

In addition, this commit also adds a unit test to test this new
behavior.

### DIFF
--- a/pkg/ipam/allocator/multipool/node_handler.go
+++ b/pkg/ipam/allocator/multipool/node_handler.go
@@ -82,7 +82,9 @@ func (n *NodeHandler) upsertLocked(resource *v2.CiliumNode) {
 	if !n.restoreFinished {
 		n.nodesPendingAllocation[resource.Name] = resource
 		_ = n.poolManager.AllocateToNode(resource)
+		return
 	}
+
 	n.createUpsertController(resource)
 }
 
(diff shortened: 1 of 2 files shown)


tklauser (approved):
Nice find. Thanks for adding a unit test as well!

tommyp1ckles (approved):
nice fix, changes look good on my end

PR #27964, opened 2023-09-06
Fix k8s code generation

aanm:
Read per commit basis

### DIFF
--- a/contrib/scripts/check-api-code-gen.sh
+++ b/contrib/scripts/check-api-code-gen.sh
@@ -39,10 +39,12 @@ git add --intent-to-add .
 
 # Check for diff
 diff="$(git diff)"
+diff_staged="$(git diff --staged)"
 
-if [ -n "$diff" ]; then
+if [ -n "$diff" ] || [ -n "$diff_staged" ]; then
 	echo "Ungenerated api source code:"
 	echo "$diff"
+	echo "$diff_staged"
 	echo "Please run 'make generate-api generate-health-api generate-hubble-api generate-operator-api' and submit your changes"
 	exit 1
 fi
--- a/contrib/scripts/check-k8s-code-gen.sh
+++ b/contrib/scripts/check-k8s-code-gen.sh
@@ -23,10 +23,12 @@ git add --intent-to-add .
 
 # Check for diff
 diff="$(git diff)"
+diff_staged="$(git diff --staged)"
 
-if [ -n "$diff" ]; then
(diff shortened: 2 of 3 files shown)


contrib/scripts/check-k8s-code-gen.sh
```diff
@@ -22,7 +22,7 @@ make generate-k8s-api manifests
 git add --intent-to-add .
 
 # Check for diff
-diff="$(git diff)"
+diff="$(git diff --staged)"
```
mhofstetter: we should add this here too: https://github.com/cilium/cilium/blob/main/contrib/scripts/check-api-code-gen.sh#L41

mhofstetter (approved):

contrib/scripts/check-api-code-gen.sh
```diff
@@ -38,7 +38,7 @@ make generate-hubble-api
 git add --intent-to-add .
 
 # Check for diff
-diff="$(git diff)"
+diff="$(git diff --staged)"
```
bimmlerd: I think this breaks the detection of new files, again, though. Quick test repo, `existing` is a committed file, I've run `rm existing; echo "blub2" > new` and then `git add --intent-to-add .` . `git diff` shows the diff in `new`, `git diff --staged` shows the diff in `existing`, `git diff` shows the diff in `new`...

```
david@lynx:~/tmprepo:main -$ git add --intent-to-add .
david@lynx:~/tmprepo:main -$ git diff --staged        
diff --git a/existing b/existing
deleted file mode 100644
index 40142d0..0000000
--- a/existing
+++ /dev/null
@@ -1 +0,0 @@
-blub
david@lynx:~/tmprepo:main -$ git diff         
diff --git a/new b/new
new file mode 100644
index 0000000..4c30ff2
--- /dev/null
+++ b/new
@@ -0,0 +1 @@
+blub2
```

bimmlerd (changes requested):

contrib/scripts/check-api-code-gen.sh
mhofstetter: why using `--intent-to-add` at all? (only tested it with a normal `git add .` :man_facepalming: - then it works with `git diff --staged` for new files too (at least on my machine :P ))

bimmlerd (approved):
LGTM as is - as Marco H says, I'm not entirely sure what `intent-to-add` gives us, but Marco I will have had his reasons :)

PR #27968, opened 2023-09-06
ipam: report IP owner of non-default pool IPs in multi-pool IPAM

tklauser:
Before this change, only IPs from the default pool are reported in `cilium status --verbose` in the "Allocated addresses" section when using multi-pool IPAM mode. With this change, all IPs from non-default pools are reported as well by making use of the fact that the pool name is prepended to the IP in `(*multiPoolManager).dump`.

Example:

```
% ks exec -it cilium-n8xw7 -- cilium status --verbose
[...]
IPAM:                   IPv4: 2 IPAM pool(s) available,
Allocated addresses:
  10.10.0.46 (router)
  10.10.0.48 (kube-system/coredns-5d78c9869d-d4nts [restored])
  10.10.0.61 (kube-system/hubble-relay-597b7bdff8-pwl69 [restored])
  10.10.0.62 (health)
  jupiter/192.168.0.15 (default/nginx-jupiter)
[...]
```

Note that IPs from non-default pools (in this case the `jupiter` pool) are reported with the pool name prefixed.

For #25470

### DIFF
--- a/pkg/ipam/allocator.go
+++ b/pkg/ipam/allocator.go
@@ -312,6 +312,14 @@ func (ipam *IPAM) ReleaseIP(ip net.IP, pool Pool) error {
 	return ipam.releaseIPLocked(ip, pool)
 }
 
+func splitIPAndPool(ip string) (string, Pool) {
+	// IPs from non-default pools are prefixed with <pool name>/
+	if i := strings.LastIndexByte(ip, '/'); i >= 0 {
+		return ip[i+1:], Pool(ip[:i])
+	}
+	return ip, PoolDefault
+}
+
 // Dump dumps the list of allocated IP addresses
 func (ipam *IPAM) Dump() (allocv4 map[string]string, allocv6 map[string]string, status string) {
 	var st4, st6 string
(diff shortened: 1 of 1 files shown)


gandro (approved):
Thanks for fixing this. Have you also considered changing the type of the `IPv4Allocator.Dump()` interface to store the pool name separately? To my knowledge, the `Allocator` interface (where `Dump` is a part of) while exported is actually not used outside the IPAM package, so I think we should consider changing the signature of `ipam.Allocator.Dump` to contain the pool name if possible.

The `ipam.IPAM.Dump` signature can remain as is, I would only change  `ipam.Allocator.Dump` so we can transport the information without having to split strings.

tklauser:
> Thanks for fixing this. Have you also considered changing the type of the `IPv4Allocator.Dump()` interface to store the pool name separately? To my knowledge, the `Allocator` interface (where `Dump` is a part of) while exported is actually not used outside the IPAM package, so I think we should consider changing the signature of `ipam.Allocator.Dump` to contain the pool name if possible.
> 
> The `ipam.IPAM.Dump` signature can remain as is, I would only change `ipam.Allocator.Dump` so we can transport the information without having to split strings.

Thanks for the suggestion. I haven't considered this so far but it would make the interface a lot cleaner. I'd probably defer that change to a follow-up PR, so this rather small PR can be backported (given it adds additional debugging/tourbleshooting information for multi-pool IPAM) without much internal API refactoring.

Issue #25676, opened 2023-05-25
Cross-node Hubble flows don't contain Kubernetes workload name


ghost:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Cross-node Hubble flows are missing Kubernetes workload name for either source or destination pod (depending on whether it's an egress on ingress flow).

This also affects Hubble metrics. With metrics configuration like:
```labelsContext=source_workload,destination_workload;sourceContext=workload-name|pod-name;destinationContext=workload-name|pod-name```
`source` and `destination` labels are inconsistent - they will have pod name or workload name, depending on where the source and destination pods are running. Also `source_workload` and `destination_workload` will be correct for some metrics and empty for others.

### Cilium Version

1.13.2

### Kernel Version

?

### Kubernetes Version

1.24.11-gke.1000

### Anything else?

The function to retrieve the workload name is called [here](https://github.com/cilium/cilium/blob/main/pkg/hubble/parser/common/endpoint.go#LL79C5). It seems that it's called only for local endpoints, not remote ones.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

michi-covalent:
extra credit for adding a test coverage using `cilium connectivity test` ✅

mritunjaysharma394:
Hi @lambdanis , I think I can try my hands fixing it, let me know if its still open to work and no one else has already begun the work on it, thanks :))

lambdanis:
@mritunjaysharma394 Would be great, thank you!

mhulscher:
Is this being worked on? With this bug, hubble L7 metrics are near useless for observing service to service traffic.

michi-covalent:
@mritunjaysharma394 let me know if you are planning to work on this. if not, i'll unassign you from this issue so that somebody else can pick it up 🙏

mritunjaysharma394:
Hi, I am sorry, this got lost in the way as just next week to this I got a new job and it has been little hectic, please feel free to unassign me, absolutely sorry for the inconvenience caused! 🙏

sujoshua:
I'd be glad to fix this issue if nobody else is working on it.

lambdanis:
@sujoshua please feel free to open a PR :bowing_woman: I assigned you.

sujoshua:
[...]

PR #27974, opened 2023-09-06
workload struct added and solved hubble metrics workload missing problem

sujoshua:
This pr is related to #25676.

This issue cannot be easily resolved due to design problems in the initial phase. In the previous code, each agent only stored the workloads of pods on the node where the agent resided.

To address this, the code was modified by adding a `workload` struct and including this field in the `K8sMetadata` and `endpoint` structs of the `ipcache`. This modification enabled parsing the workload information from the pod whenever a pod creation or update was observed through the K8s watcher, and storing it in the `endpoint` struct.

For environments where the `ciliumendpoint` CRD is available and etcd is not enabled, the `RunK8sCiliumEndpointSync` function was introduced to marshal the workload information maintained by each node for its pods into JSON format and append it as an annotation to the `ciliumendpoint` CRD, ensuring synchronization across all nodes.

In environments with etcd enabled, the `workload` field was added to the `IPIdentityPair` struct to achieve workload information synchronization.

Furthermore, compared to the previous implementation, which required extracting the workload from the pod every time data was collected by Hubble, this pull request adopts a space-for-time approach by directly storing the workload, resulting in time savings.

Note: Contrary to what was mentioned in the issue, this update is indeed compatible and does not require any changes. It can be directly used. This is because I eventually discovered that when endpoint  is `restore` it still interacts with the K8s API server.

cheers！

### DIFF
--- a/.github/workflows/conformance-gateway-api.yaml
+++ b/.github/workflows/conformance-gateway-api.yaml
@@ -113,7 +113,7 @@ jobs:
           fi
           echo sha=${SHA} >> $GITHUB_OUTPUT
           
-          SUPPORTED_FEATURES="ReferenceGrant,HTTPRoute,TLSRoute,HTTPRouteQueryParamMatching,HTTPRouteMethodMatching,GatewayClassObservedGenerationBump,HTTPRouteHostRewrite,HTTPRoutePathRewrite,HTTPRouteSchemeRedirect,HTTPRoutePathRedirect,HTTPRoutePortRedirect,HTTPRouteRequestMirror"
+          SUPPORTED_FEATURES="ReferenceGrant,HTTPRoute,TLSRoute,HTTPRouteQueryParamMatching,HTTPRouteMethodMatching,GatewayClassObservedGenerationBump,HTTPRouteHostRewrite,HTTPRoutePathRewrite,HTTPRouteSchemeRedirect,HTTPRoutePathRedirect,HTTPRoutePortRedirect,HTTPRouteRequestMirror,HTTPRouteRequestMultipleMirrors"
           if [ ${{ matrix.crd-channel }} == "experimental" ]; then
             SUPPORTED_FEATURES+=",HTTPResponseHeaderModification,RouteDestinationPortMatching"
           fi
--- a/operator/pkg/model/ingestion/gateway.go
+++ b/operator/pkg/model/ingestion/gateway.go
@@ -93,7 +93,7 @@ func GatewayAPI(input Input) ([]model.HTTPListener, []model.TLSListener) {
 				var responseHeaderFilter *model.HTTPHeaderFilter
 				var requestRedirectFilter *model.HTTPRequestRedirectFilter
 				var rewriteFilter *model.HTTPURLRewriteFilter
-				var requestMirror *model.HTTPRequestMirror
+				var requestMirrors []*model.HTTPRequestMirror
 
 				for _, f := range rule.Filters {
 					switch f.Type {
@@ -114,7 +114,7 @@ func GatewayAPI(input Input) ([]model.HTTPListener, []model.TLSListener) {
 					case gatewayv1beta1.HTTPRouteFilterURLRewrite:
 						rewriteFilter = toHTTPRewriteFilter(f.URLRewrite)
 					case gatewayv1beta1.HTTPRouteFilterRequestMirror:
-						requestMirror = toHTTPRequestMirror(f.RequestMirror, r.Namespace)
+						requestMirrors = append(requestMirrors, toHTTPRequestMirror(f.RequestMirror, r.Namespace))
 					}
 				}
 
@@ -127,7 +127,7 @@ func GatewayAPI(input Input) ([]model.HTTPListener, []model.TLSListener) {
 						ResponseHeaderModifier: responseHeaderFilter,
 						RequestRedirect:        requestRedirectFilter,
 						Rewrite:                rewriteFilter,
-						RequestMirror:          requestMirror,
+						RequestMirrors:         requestMirrors,
 					})
 				}
 
@@ -144,7 +144,7 @@ func GatewayAPI(input Input) ([]model.HTTPListener, []model.TLSListener) {
 						ResponseHeaderModifier: responseHeaderFilter,
 						RequestRedirect:        requestRedirectFilter,
 						Rewrite:                rewriteFilter,
-						RequestMirror:          requestMirror,
+						RequestMirrors:         requestMirrors,
 					})
 				}
 			}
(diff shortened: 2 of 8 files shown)


chancez:
The overall implementation looks fairly solid to me. Is there a reason to use an annotation vs adding new fields to the endpoint itself?

chancez:
> > The overall implementation looks fairly solid to me. Is there a reason to use an annotation vs adding new fields to the endpoint itself?
> 
> update a crd is expensive but add an annotation is cheap

Expensive in what way?

sujoshua:
> Expensive in what way?

Because this involves updating the existing version of the ciliumendpoint CRD, there are many places in the code that need to be changed. I'm not sure if the community is willing to update it.

chancez:
> Because this involves updating the existing version of the ciliumendpoint CRD, there are many places in the code that need to be changed. I'm not sure if the community is willing to update it.

Besides actually updating the CRD itself, updating it to add a new field shouldn't require updating it in any more places than the places you're currently checking for the annotation, and would provide a proper schema we can validate, etc.

sujoshua:
> Besides actually updating the CRD itself, updating it to add a new field shouldn't require updating it in any more places than the places you're currently checking for the annotation, and would provide a proper schema we can validate, etc.

Do we need to update the version identifier of the current CRD from v2 to v3?

chancez:
> > Besides actually updating the CRD itself, updating it to add a new field shouldn't require updating it in any more places than the places you're currently checking for the annotation, and would provide a proper schema we can validate, etc.
> 
> Do we need to update the version identifier of the current CRD from v2 to v3?

When adding a new field? No. You only have to modify the version of a CRD when making backwards incompatible changes. Adding a new field is generally backwards compatible.

youngnick:
I definitely agree that this would be much better as a field on a CRD rather than an annotation - as @chancez says, that makes the data structured instead of free-text.

Adding a new field is a backward-compatible change as long as the behavior for an empty object isn't changed - as long as the default on the new field is either a nil pointer or an empty struct, it's all good.

sujoshua:
Add a field `status.workload` in  CiliumEndpoint CRD now. Like this:
```
    .......
    state: ready
    visibility-policy-status: <status disabled>
    workload:
      kind: Deployment
      name: hubble-ui
      namespace: kube-system
```

pkg/hubble/parser/common/endpoint.go
```diff
@@ -75,41 +74,36 @@ func (r *EndpointResolver) ResolveEndpoint(ip netip.Addr, datapathSecurityIdenti
...
+				e.Workloads = []*pb.Workload{{Kind: workload.Kind, Name: workload.Name}}
 			}
 			return e
 		}
 	}
 
 	// for remote endpoints, assemble the information via ip and identity
-	numericIdentity := datapathSecurityIdentity
-	var namespace, podName string
+	e := pb.Endpoint{
```
chancez: Is there a reason to create the endpoint object up here and assigning to the fields, rather than the old method which just returned the `pb.Endpoint` with each field set? I feel like this approach makes it easier to miss setting fields, since it's broken across multiple conditions and lines.

sujoshua: Taking performance into account, as I minimize variable creation. If you consider the improvement to be marginal and not worth sacrificing code readability, I can revert back to the original approach. :-)

chancez: More variables doesn't necessarily mean worse performance. The compiler could be optimizing away any of the temporary variables anyways.

pkg/k8s/utils/utils.go
```diff
@@ -219,3 +220,15 @@ func GetClusterIPByFamily(ipFamily slim_corev1.IPFamily, service *slim_corev1.Se
...
+
+// EscapeJsonPointer escapes string to be used as a JSON pointer.
+//
+// This is performed by first transforming any
+//
+//	occurrence of the sequence '~' to '~0', and then transforming any
+//	occurrence of the sequence '/' to '~1'.
+//
+// Details - https://datatracker.ietf.org/doc/html/rfc6901#section-4
+func EscapeJsonPointer(s string) string {
```
chancez: I'm not seeing this used anywhere.

pkg/k8s/watchers/pod.go
```diff
@@ -851,6 +851,31 @@ func (k *K8sWatcher) updatePodHostData(oldPod, newPod *slim_corev1.Pod, oldPodIP
 	specEqual := oldPod != nil && newPod.Spec.DeepEqual(&oldPod.Spec)
 	hostIPEqual := oldPod != nil && newPod.Status.HostIP != oldPod.Status.HostIP
 
+	ownerReferenceEqual := false
```
chancez: Is this strictly necessary? I don't think ownerReferences generally change on existing resources. It's possible, but pretty uncommon (it would generally only happen if the user manually changed it). I'm not opposed to having the logic, but curious if there's an actual real necessity for it.

pkg/types/workload.go
```diff
@@ -0,0 +1,125 @@
...
+// Below is the original comment in k8s/utils/workload.go:
+// GetDeployMetaFromPod has been renamed to GetWorkloadMetaFromPod and has
+// been updated to use the cilium slim API types.
+// We do not store the APIVersion of the owning workload in the TypeMeta
+// either, because it isn't needed for our purposes, and our slim types do not
+// have this field.
+// We fallback to the pod's ownerReference if we cannot find a more suitable
+// workload based on heuristics, whereas the original code defaulted to the
+// pod's name. This may be the case when using ReplicaSets without a Deployment.
+func GetWorkloadDataFromPod(pod *slim_corev1.Pod) (*Workload, error) {
```
chancez: This mostly seems to be a copy of `utils.GetWorkloadMetaFromPod`. Can you reuse that function, or adapt it so the logic isn't duplicated between these two functions?

pkg/k8s/utils/utils.go
sujoshua: I assume that someone may use it oneday. Ok then, deleted.

pkg/k8s/watchers/pod.go
sujoshua: I acknowledge that this is a rare situation, but since Kubernetes is an open platform, we should not have high expectations regarding user usage, especially in an era where this operator is prevalent.

pkg/types/workload.go
sujoshua: Since original function `utils.GetWorkloadMetaFromPod` is only used in one place now. so I adapt that place  to use the new function and delete the function  `utils.GetWorkloadMetaFromPod`.

chancez:
Latest changes LGTM. Can you rebase?

PR #27976, opened 2023-09-06
pkg/l2announcer: Adds Service LoadBalancerClass Support

### LINKED ISSUE #26556: L2-announcements and BGP compatibility

danehans:
- Adds `L2AnnounceLoadBalancerClass` constant to define the load balancer class supported by the L2 Announcer.
- Updates `upsertSvc()` to check the load balancer class of the provided service.
- Updates the `ServiceSelector` godocs to reflect the new API behavior.
- Updates operator LB IPAM pkg to support L2 Announcer LB class.
- Adds unit tests.

Fixes: #26556

```release-note
CiliumL2AnnouncementPolicy will only select Services that do not specify a LoadBalancerClass or specify a LoadBalancerClass of "io.cilium/l2-announcer".
```

### DIFF
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -899,7 +899,9 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 	// which can be modified after the device detection.
 	var devices []string
 	if d.deviceManager != nil {
-		if _, err := d.deviceManager.Detect(params.Clientset.IsEnabled()); err != nil {
+		if detected, err := d.deviceManager.Detect(params.Clientset.IsEnabled()); err == nil {
+			devices = append(devices, detected...)
+		} else {
 			if option.Config.AreDevicesRequired() {
 				// Fail hard if devices are required to function.
 				return nil, nil, fmt.Errorf("failed to detect devices: %w", err)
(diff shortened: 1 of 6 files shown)


dylandreimerink (approved):
Thank you for working on this. Looks good to me

nathanjsweet (approved):

danehans:
CI is passing, @dylandreimerink @nathanjsweet do you have any recommendations for other maintainers to review and merge?

PR #27977, opened 2023-09-06
Clean up tests-ipsec-upgrade workflow

michi-covalent:
- Add a script to get the Cilium version to downgrade to instead of
  hardcoding it in the workflow file. The script uses the top-level
  VERSION file to infer the previous version.
- Check out the git branch to get the Helm chart instead of doing a wget
  to download the source archive.

Signed-off-by: Michi Mutsuzaki <michi@isovalent.com>

### DIFF
--- a/contrib/scripts/print-downgrade-version.sh
+++ b/contrib/scripts/print-downgrade-version.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+#
+# A utility script to print the branch name of the previous stable release.
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+VERSION="$(cat "$SCRIPT_DIR/../../VERSION")"
+if [[ $VERSION =~ ([0-9^]+)\.([0-9^]+)\..* ]] ; then
+    major=${BASH_REMATCH[1]}
+    minor=${BASH_REMATCH[2]}
+    ((minor--))
+    echo "v${major}.${minor}${BRANCH_SUFFIX:-}"
+else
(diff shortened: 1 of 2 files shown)


jschwinger233 (approved):

contrib/scripts/print-downgrade-version.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+#
+# A utility script to print the branch name of the previous stable release.
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+VERSION="$(cat "$SCRIPT_DIR/../../VERSION")"
+if [[ $VERSION =~ ([0-9^]+)\.([0-9^]+)\..* ]] ; then
+    major=${BASH_REMATCH[1]}
+    minor=${BASH_REMATCH[2]}
+    ((minor--))
```
viktor-kurchenko, line 10: Is it possible that `minor` will be zero?

michi-covalent: yes it is possible when cilium releases v2.0.

viktor-kurchenko: > yes it is possible when cilium releases v2.0.

So, then the downgraded version will be: `v2.-1.` right?

brb (approved):
:tomato:

nebril (approved):

contrib/scripts/print-downgrade-version.sh
michi-covalent: yeah this thing will fail when we get to cilium v2.0. let's fix the script when that happens 🥰

PR #27981, opened 2023-09-06
Resiliency: Add checks to ensure endpoint BPF programs remain loaded

derailed:

      GitHub issue.

Resiliency: Add checks to ensure endpoints related BPF programs remain loaded
    
We've seen instance where BPF programs are deleted by third party
software (aka Ddog). When this occurrs either agents will fail to
restart or connectivity may be compromised.
    
This PR attempts to reconciliate BPF program state with their associated
endpoints by ensuring endpoints programs are reloaded when necessary.
        

### DIFF
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -1276,6 +1276,8 @@ func (d *Daemon) Close() {
 	}
 	identitymanager.RemoveAll()
 	d.cgroupManager.Close()
+
+	// Ensures all controllers are stopped!
 	d.controllers.RemoveAllAndWait()
 }
 
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -1627,6 +1627,7 @@ var daemonCell = cell.Module(
 	// by the debuginfo API
 	cell.ProvidePrivate(daemonSettings),
 	cell.Invoke(func(promise.Promise[*Daemon]) {}), // Force instantiation.
+	endpointBPFrogWatchdogCell,
 )
 
 type daemonParams struct {
--- a/daemon/cmd/watchdogs.go
+++ b/daemon/cmd/watchdogs.go
@@ -1,0 +1,136 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package cmd
+
+import (
+	"context"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/pflag"
+
+	"github.com/cilium/cilium/pkg/controller"
+	"github.com/cilium/cilium/pkg/datapath/loader"
+	"github.com/cilium/cilium/pkg/endpoint"
+	"github.com/cilium/cilium/pkg/hive"
+	"github.com/cilium/cilium/pkg/hive/cell"
+	"github.com/cilium/cilium/pkg/logging/logfields"
+	"github.com/cilium/cilium/pkg/promise"
+)
+
+const epBPFProgWatchdog = "ep-bpf-prog-watchdog"
+
+type epBPFProgWatchdogConfig struct {
+	EndpointBPFProgWatchdogInterval time.Duration
+}
+
+func (c epBPFProgWatchdogConfig) Flags(flags *pflag.FlagSet) {
+	flags.Duration("endpoint-bpf-prog-watchdog-interval", c.EndpointBPFProgWatchdogInterval,
+		"Interval to trigger endpoint BPF programs load check watchdog")
+}
+
+type epBPFProgWatchdogParams struct {
+	cell.In
+
+	Config        epBPFProgWatchdogConfig
+	Logger        logrus.FieldLogger
+	Lifecycle     hive.Lifecycle
+	DaemonPromise promise.Promise[*Daemon]
+}
+
+var (
+	// endpointBPFrogWatchdogCell triggers a job to ensure device tc programs remain loaded.
+	endpointBPFrogWatchdogCell = cell.Module(
+		epBPFProgWatchdog,
+		"Periodically checks that endpoint BPF programs remain loaded",
+
+		cell.Config(epBPFProgWatchdogConfigDefault),
+		cell.Invoke(registerEndpointBPFProgWatchdog),
+	)
+
+	epBPFProgWatchdogConfigDefault = epBPFProgWatchdogConfig{
+		EndpointBPFProgWatchdogInterval: 30 * time.Second,
+	}
+)
+
+func registerEndpointBPFProgWatchdog(p epBPFProgWatchdogParams) {
+	if p.Config.EndpointBPFProgWatchdogInterval == 0 {
+		return
+	}
+
+	var (
+		ctx, cancel = context.WithCancel(context.Background())
+		mgr         = controller.NewManager()
+	)
+	p.Lifecycle.Append(hive.Hook{
+		OnStart: func(hive.HookContext) error {
+			mgr.UpdateController(
+				epBPFProgWatchdog,
+				controller.ControllerParams{
+					Group: controller.NewGroup(epBPFProgWatchdog),
+					DoFunc: func(ctx context.Context) error {
+						d, err := p.DaemonPromise.Await(ctx)
+						if err != nil {
+							return err
+						}
+						return d.checkEndpointBPFPrograms(ctx, p)
+					},
+					RunInterval: p.Config.EndpointBPFProgWatchdogInterval,
+					Context:     ctx,
+				},
+			)
+
+			return nil
+		},
+		OnStop: func(hive.HookContext) error {
+			cancel()
+			mgr.RemoveAllAndWait()
+			return nil
+		},
+	},
+	)
+}
+
+func (d *Daemon) checkEndpointBPFPrograms(ctx context.Context, p epBPFProgWatchdogParams) error {
+	var (
+		loaded = true
+		err    error
+		eps    = d.endpointManager.GetEndpoints()
+	)
+	for _, ep := range eps {
+		if ep.GetState() != endpoint.StateReady {
+			continue
+		}
+		loaded, err = loader.DeviceHasTCProgramLoaded(ep.HostInterface(), ep.RequireEgressProg())
+		if err != nil {
+			log.WithField(logfields.Endpoint, ep.HostInterface()).
+				WithError(err).
+				Error("Unable to assert if endpoint BPF programs need to be reloaded")
+			return err
+		}
+		// We've detected missing bpf progs for this endpoint.
+		// Break and trigger bpf progs reload.
+		if !loaded {
+			break
+		}
+	}
+	if loaded {
+		return nil
+	}
+
+	log.WithField(logfields.Count, len(eps)).
+		Warn(
+			"Detected unexpected endpoint BPF program removal. " +
+				"Consider investigating whether other software running on this machine is removing Cilium's endpoint BPF programs. " +
+				"If endpoint BPF programs are removed, the associated pods will lose connectivity and only reinstating the programs will restore connectivity.",
+		)
+	wg, err := d.TriggerReloadWithoutCompile(epBPFProgWatchdog)
+	if err != nil {
+		log.WithError(err).Error("Failed to reload Cilium endpoints BPF programs")
+	} else {
+		wg.Wait()
+	}
+
+	return err
+}
--- a/pkg/datapath/loader/netlink.go
+++ b/pkg/datapath/loader/netlink.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/vishvananda/netlink"
@@ -24,6 +25,8 @@ import (
 	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/sysctl"
 )
+
+const qdiscClsact = "clsact"
 
 func directionToParent(dir string) uint32 {
 	switch dir {
@@ -44,7 +47,7 @@ func replaceQdisc(link netlink.Link) error {
 
 	qdisc := &netlink.GenericQdisc{
 		QdiscAttrs: attrs,
-		QdiscType:  "clsact",
+		QdiscType:  qdiscClsact,
 	}
 
 	return netlink.QdiscReplace(qdisc)
@@ -691,4 +694,62 @@ func renameDevice(from, to string) error {
 	}
 
 	return nil
+}
+
+// DeviceHasTCProgramLoaded checks whether a given device has tc filter/qdisc progs attached.
+func DeviceHasTCProgramLoaded(hostInterface string, checkEgress bool) (bool, error) {
+	const bpfProgPrefix = "cil_"
+
+	l, err := netlink.LinkByName(hostInterface)
+	if err != nil {
+		return false, fmt.Errorf("unable to find endpoint link by name: %w", err)
+	}
+
+	dd, err := netlink.QdiscList(l)
+	if err != nil {
+		return false, fmt.Errorf("unable to fetch qdisc list for endpoint: %w", err)
+	}
+	var found bool
+	for _, d := range dd {
+		if d.Type() == qdiscClsact {
+			found = true
+			break
+		}
+	}
+	if !found {
+		return false, nil
+	}
+
+	ff, err := netlink.FilterList(l, netlink.HANDLE_MIN_INGRESS)
+	if err != nil {
+		return false, fmt.Errorf("unable to fetch ingress filter list: %w", err)
+	}
+	var filtersCount int
+	for _, f := range ff {
+		if filter, ok := f.(*netlink.BpfFilter); ok {
+			if strings.HasPrefix(filter.Name, bpfProgPrefix) {
+				filtersCount++
+			}
+		}
+	}
+	if filtersCount == 0 {
+		return false, nil
+	}
+	if !checkEgress {
+		return true, nil
+	}
+
+	ff, err = netlink.FilterList(l, netlink.HANDLE_MIN_EGRESS)
+	if err != nil {
+		return false, fmt.Errorf("unable to fetch egress filter list: %w", err)
+	}
+	filtersCount = 0
+	for _, f := range ff {
+		if filter, ok := f.(*netlink.BpfFilter); ok {
+			if strings.HasPrefix(filter.Name, bpfProgPrefix) {
+				filtersCount++
+			}
+		}
+	}
+	return len(ff) > 0 && filtersCount > 0, nil
 }
--- a/Documentation/cmdref/cilium-agent.md
+++ b/Documentation/cmdref/cilium-agent.md
@@ -162,6 +162,7 @@ cilium-agent [flags]
       --encrypt-node                                              Enables encrypting traffic from non-Cilium pods and host networking (only supported with WireGuard, beta)
       --encryption-strict-mode-allow-remote-node-identities       Allows unencrypted traffic from pods to remote node identities within the strict mode CIDR. This is required when tunneling is used or direct routing is used and the node CIDR and pod CIDR overlap.
       --encryption-strict-mode-cidr string                        In strict-mode encryption, all unencrypted traffic coming from this CIDR and going to this same CIDR will be dropped
+      --endpoint-bpf-prog-watchdog-interval duration              Interval to trigger endpoint BPF programs load check watchdog (default 30s)
       --endpoint-queue-size int                                   Size of EventQueue per-endpoint (default 25)
       --endpoint-status strings                                   Enable additional CiliumEndpoint status features (controllers,health,log,policy,state)
       --envoy-config-timeout duration                             Timeout duration for Envoy Config acknowledgements (default 2m0s)
--- a/Documentation/cmdref/cilium-agent_hive.md
+++ b/Documentation/cmdref/cilium-agent_hive.md
@@ -32,6 +32,7 @@ cilium-agent hive [flags]
       --enable-k8s-endpoint-slice                                 Enables k8s EndpointSlice feature in Cilium if the k8s cluster supports it (default true)
       --enable-l2-pod-announcements                               Enable announcing Pod IPs with Gratuitous ARP
       --enable-monitor                                            Enable the monitor unix domain socket server (default true)
+      --endpoint-bpf-prog-watchdog-interval duration              Interval to trigger endpoint BPF programs load check watchdog (default 30s)
       --gops-port uint16                                          Port for gops server to listen on (default 9890)
   -h, --help                                                      help for hive
       --k8s-api-server string                                     Kubernetes API server URL
--- a/Documentation/cmdref/cilium-agent_hive_dot-graph.md
+++ b/Documentation/cmdref/cilium-agent_hive_dot-graph.md
@@ -38,6 +38,7 @@ cilium-agent hive dot-graph [flags]
       --enable-k8s-endpoint-slice                                 Enables k8s EndpointSlice feature in Cilium if the k8s cluster supports it (default true)
       --enable-l2-pod-announcements                               Enable announcing Pod IPs with Gratuitous ARP
       --enable-monitor                                            Enable the monitor unix domain socket server (default true)
+      --endpoint-bpf-prog-watchdog-interval duration              Interval to trigger endpoint BPF programs load check watchdog (default 30s)
       --gops-port uint16                                          Port for gops server to listen on (default 9890)
       --k8s-api-server string                                     Kubernetes API server URL
       --k8s-client-burst int                                      Burst value allowed for the K8s client


pkg/resiliency/endpoints/bpf_prog_reload_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/cilium/cilium/pkg/resiliency/endpoints"
+	"github.com/sirupsen/logrus/hooks/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/vishvananda/netlink"
+)
+
+func TestNeedsBPFProgReload(t *testing.T) {
+	uu := map[string]struct {
+		b endpoints.BPFProgBacker
+		e bool
```
thorn3r: nit: these are really short names, something a little longer might improve readability

daemon/cmd/daemon_main.go
```diff
@@ -1918,6 +1919,45 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
...
+	controller.NewManager().UpdateController(
+		epsBPFSync,
+		controller.ControllerParams{
+			Group: controller.NewGroup(epsBPFSync),
+			DoFunc: func(ctx context.Context) error {
+				var (
+					reload bool
+					eps    = d.endpointManager.GetEndpoints()
+				)
+				// !!BOZO!! Debug...
```
thorn3r: looks like some debugging leftover here :grin:

pkg/resiliency/endpoints/bpf_backer.go
```diff
@@ -0,0 +1,51 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package endpoints
```
thorn3r: nit: i think go best practices advocate for singular names rather than plural

daemon/cmd/daemon_main.go
```diff
@@ -1918,6 +1919,45 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
...
+					return nil
+				}
+
+				log.Debugf("Reloading endpoint BPF programs (%d)", len(eps))
+				_, err := d.TriggerReloadWithoutCompile(epsBPFSync)
+				return err
+			},
+			RunInterval: epsBPFSyncDuration,
+			Context:     d.ctx,
+		})
```
thorn3r: what do you think about putting this whole chunk into its own function, rather than directly in startDaemon?

pkg/resiliency/endpoints/bpf_backer.go
```diff
@@ -6,8 +6,9 @@ package endpoints
 import (
 	"fmt"
 
-	"github.com/cilium/cilium/pkg/endpoint"
 	"github.com/vishvananda/netlink"
+
+	"github.com/cilium/cilium/pkg/endpoint"
```
thorn3r: nit: squashing these changes into the the previous commit would clean this diff up a bit

thorn3r (changes requested):
functionally this looks good and a great improvement for resiliency :rocket:
left a few comments, mostly nits. lmk what you think!

daemon/cmd/daemon_main.go
derailed: Good catch!

daemon/cmd/daemon_main.go
derailed: Good point! I'll make it so.

daemon/cmd/daemon_main.go
```diff
@@ -1918,6 +1919,45 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
...
+		epsBPFSync,
+		controller.ControllerParams{
+			Group: controller.NewGroup(epsBPFSync),
+			DoFunc: func(ctx context.Context) error {
+				var (
+					reload bool
+					eps    = d.endpointManager.GetEndpoints()
+				)
+				for _, ep := range eps {
+					reload, err := resEP.NeedsBPFProgReload(log, resEP.NewEndpointBPFProgBacker(ep))
```
christarazi: Is it really necessary to have `NeedsBPFProgReload` be separate from the `endpoint` package? I like the idea of moving it to a helper, but I think the consequence of hooking this entire reconciliation loop from the daemon, instead of the endpointmanager directly, leaks implementation details throughout various layers.

I think each `Endpoint` can have a function to check the status of their BPF progs (similar to `NeedsBPFProgReload`) and the endpointmanager trigger the `epsBPFSync` controller. All the daemon needs to do is init the endpointmanager, which it already does.

This makes it so that not every layer in Cilium needs to be aware of the underlying layers. Meaning, the daemon doesn't need to have the implementation details of the responsibility of the endpointmanager, and the endpoint's BPF program logic doesn't need to be spread out to other packages (` pkg/resiliency/endpoint`).

christarazi (changes requested):
Looking good. One comment about the architecture of the code.

daemon/cmd/daemon_main.go
derailed: Good point! Tx Chris. I've debated a bit on this. Was hoping to leverage behaviors vs just hanging yet another method on endpoints. The trick here is the actual lift ie the bpf prog reload method is happening on the daemon and not the ep manager...

pkg/resiliency/endpoint/bpf_backer.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// GetLink retrieves the associated link.
+func (e *EndpointBPFProgBacker) GetLink() (netlink.Link, error) {
+	return netlink.LinkByName(e.ep.HostInterface())
+}
+
+// QdiscList retrieves associated qdisc attrs.
+func (e *EndpointBPFProgBacker) QdiscList(l netlink.Link) ([]netlink.Qdisc, error) {
+	return netlink.QdiscList(l)
```
joamaki: These are linux-specific and thus should live somewhere under `pkg/datapath/linux`. This should also be marked with `linux` build tag. I know it's not the case for a lot of code in cilium, but we should continue to be moving towards having side-effectful platform specific code in `pkg/datapath/linux`.

daemon/cmd/daemon_main.go
```diff
@@ -1918,6 +1919,45 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
...
+	checkEndpointBPFProgs(d)
+}
+
+// checkEndpointBPFProgs check to ensure endpoints associated BPF programs remain loaded.
+func checkEndpointBPFProgs(d *Daemon) {
+	const (
+		epsBPFSync         = "eps-bpf-sync"
+		epsBPFSyncDuration = 30 * time.Second
+	)
+	controller.NewManager().UpdateController(
```
joamaki, line 1932: While this way of doing this feature is good for backporting, I would still make this into a separate cell, both so that it's easy to test and then easy to remove once pkg/datapath/loader is a reconciler. 

e.g.
```
var daemonCell = ...
  cell.Invoke(registerEndpointBPFProgramWatchdog),
)

func registerEndpointBPFProgramWatchdog(daemonPromise promise.Promise[*Daemon], endpoints endpointmanager.EndpointsLookup) {
  controller.NewManager().UpdateController(...
     DoFunc: func(ctx context.Context) error {
       d, err := daemonPromise.Await(ctx)
       if err != nil { return err }

       for _, ep := range endpoints.GetEndpoints() { ... }
    })
}
```

Could also consider lifting the `TriggerReloadWithoutCompile` from `*Daemon`:
```
// cell.Provide(newTriggerReload)

type TriggerReloadFunc func(ctx context.Context, reason string) error

func newTriggerReload(daemonPromise promise.Promise[*Daemon]) TriggerReloadFunc {
  return func(reason string) error {
    d, err := daemonPromise.Await(ctx)
    if err != nil { return err }
    return d.TriggerReloadWithoutCompile(reason)
  }   
}
```

While it's a bit more boilerplate, it at least won't push us backwards in modularizing the daemon.

daemon/cmd/daemon_main.go
```diff
@@ -1918,6 +1919,45 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
 	if err != nil {
 		log.WithError(err).Error("Unable to store Viper's configuration")
 	}
+
+	checkEndpointBPFProgs(d)
+}
+
+// checkEndpointBPFProgs check to ensure endpoints associated BPF programs remain loaded.
+func checkEndpointBPFProgs(d *Daemon) {
```
joamaki: This is taking us backwards with regards to modularizing the daemon. This should be implemented as a standalone cell that depends on `Promise[*Daemon]` (or a bridged interface with `TriggerReloadWithCompile`, see `wrapAPIHandler` in api_handlers.go for what I mean) and `endpointmanager.EndpointsLookup`. It does make it harder/impossible to backport beyond v1.14 though, so not 100% sure whether we should take this in in this form and then refactor in follow-up after backports...

daemon/cmd/daemon_main.go
joamaki: This is never cleaned up when the daemon stops. This will cause issues in tests (e.g. test/controlplane) that start the agent multiple times, e.g. this controller ends up running multiple times in the background over multiple tests.

EDIT: Doh, thought I'd suggest using `(*Daemon).controllers` but that is not stopped properly either.  So either use that and add `RemoveAllAndWait` to `Close()`, or append a daemon cleanup step for the new manager so it won't leak in tests.

joamaki (changes requested):

pkg/resiliency/endpoint/bpf_backer.go
joamaki: Preferably this would either be part of `Loader` interface (`pkg/datapath/types/loader.go` and implemented by the loader next to `Compile*` methods. Then it'll naturally get refactored along with the loader. Or we just move the `netlink` stuff from here into `pkg/datapath/loader/netlink.go`. It doesn't make sense to add new packages that are closely related to existing functionality.

daemon/cmd/daemon_main.go
```diff
@@ -1918,6 +1919,45 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
...
+	const (
+		epsBPFSync         = "eps-bpf-sync"
+		epsBPFSyncDuration = 30 * time.Second
```
joamaki: Should we consider making this tunable and possible to disable it?

E.g. wondering if this for some unseen reason is too heavy to run every 30 seconds, or there's some bug in it. We might want to be able to disable it. A hidden option could work here. Though might be overkill...

pkg/resiliency/endpoint/types.go
```diff
@@ -0,0 +1,37 @@
...
+	"fmt"
+
+	"github.com/vishvananda/netlink"
+
+	"github.com/cilium/cilium/pkg/endpoint"
+)
+
+const (
+	qdiskClsact           = "clsact"
+	tcFilterParentIngress = 0xfffffff2
```
joamaki: What is this constant and where is it from? Can it change? Should't it live next to the code that sets up the tc filters so it doesn't accidentally get broken?

pkg/resiliency/endpoint/bpf_prog_reload.go
```diff
@@ -0,0 +1,50 @@
...
+package endpoint
+
+import (
+	"github.com/sirupsen/logrus"
+
+	"github.com/cilium/cilium/pkg/endpoint"
+)
+
+// NeedsBPFProgReload checks if endpoint tc filter/qdisc progs are missing and needs to be reloaded.
+func NeedsBPFProgReload(log logrus.FieldLogger, b BPFProgBacker) (bool, error) {
```
joamaki: This and the `BPFProgBacker` interface seems like overkill and again contributes to sprinkling a simple feature all over the codebase. That interface is also mixing two unrelated things: endpoint field access and the qdisc/filter stuff which is pretty confusing.

I would remove `pkg/resiliency/endpoint`, add the qdisc and filter helpers to `pkg/datapath/loader` or `pkg/datapath/linux` and then just do the straightforward thing in `daemon/cmd` to implement the watchdog. Making it self-contained into a single function as possible (with linux-specific code still in proper place) makes it easy to remove this once we don't need this stop-gap measure anymore.

pkg/resiliency/endpoint/bpf_backer.go
derailed: Thanks for the feedback Jussi! I've reworked the pr to line up with your expectations.

daemon/cmd/daemon_main.go
christarazi: For backports, we can come up with a separate implementation that's as close as possible to the version that lands in main, but I agree with this feedback of using the new mechanisms that enable modularization.

daemon/cmd/daemon_main.go
christarazi: Ah ok, in that case then, we can keep the hooking from the daemon, but the rest of the feedback is still applicable.

daemon/cmd/daemon_main.go
derailed: Great point! I'll update...

pkg/resiliency/endpoint/types.go
derailed: Good catch! I piggy back off cmd/cleanup which defined this const. But it looks like netlink already export them. Axing...

pkg/resiliency/endpoint/bpf_prog_reload.go
derailed: Right! the idea here was to leverage behavior vs type. Figured easier to test and also isolate the platform a bit.
So I parked it under `pkg/resilience` for the time being. I'll update...

daemon/cmd/daemon.go
```diff
@@ -1270,6 +1270,9 @@ func (d *Daemon) Close() {
 	}
 	identitymanager.RemoveAll()
 	d.cgroupManager.Close()
+
+	// Ensures all controllers are stopped!
+	d.controllers.RemoveAllAndWait()
```
christarazi: This is a separate fix altogether so it should be submitted separately in another PR.

daemon/cmd/daemon_main.go
derailed: I think having a configurable interval makes sense. However not sure about disabling it? Should resiliency be opted-in?

daemon/cmd/daemon_main.go
derailed: I've reworked the original proposal. Let see if that alleviate some of your concerns...

daemon/cmd/daemon.go
derailed: Not sure?? we introduce a new manager that would not properly stop when the daemon exits? 
Based on Jussi's comment above, I've gathered this could entail leaks for some of our tests?

daemon/cmd/daemon_main.go
derailed: Good point!

daemon/cmd/daemon.go
christarazi: The reality is that this code is not present in main, so adding another controller on top is not going to change the fact that this gap already exists. We might as well submit a separate fix to get it merged and backported, while this PR is being reviewed.

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/cilium/cilium/pkg/controller"
+	"github.com/cilium/cilium/pkg/datapath/loader"
+	"github.com/cilium/cilium/pkg/endpoint"
+)
+
+const epBPFProgWatchdog = "ep-bpf-prog-watchdog"
+
+// checkEndpointBPFProgs check to ensure endpoints associated BPF programs remain loaded.
+func checkEndpointBPFProgs(d *Daemon, rate time.Duration) {
```
christarazi: This function should be tied to `Daemon` anyway, right? It's already taking `Daemon` as a param.

christarazi: I'm also thinking about Jussi's comment of this code should probably just a Cell on its own that hook into the main daemon Cell. I believe that could allow the config options to be bundled with the Cell and not have to sprawl out all the pieces of this functionality all over the code base.

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,60 @@
...
+					err    error
+					eps    = d.endpointManager.GetEndpoints()
+				)
+				for _, ep := range eps {
+					if ep.GetState() != endpoint.StateReady {
+						continue
+					}
+					loaded, err = loader.DeviceHasTCProgramLoaded(ep.HostInterface())
+					if err != nil {
+						log.WithField("ep_host_interface", ep.HostInterface()).
```
christarazi: We have well-defined logfields for these things already, see `pkg/logging/logfields` for the appropriate one.

daemon/cmd/daemon.go
derailed: Tx Chris! makes sense

daemon/cmd/watchdogs.go
joamaki: Yep, perhaps it'd be a good exercise to do now. The way the code is now is something we can backport, but indeed would be nice to leave this as separate cell to keep encouraging moving into that style.

Here's roughly how it would look like:

```
type endpointBPFWatchdogConfig struct {
  ...
}
func (def endpointBPFWatchdogConfig) Flags(flags *pflag.Flagger) {
  ...
}

var endpointBPFWatchdogCell = cell.Module(
  "endpoint-bpf-program-watchdog",
  "Periodically checks that all endpoint interfaces have BPF program loaded onto them",

  cell.Config(endpointBPFWatchdogConfig),
  cell.Invoke(registerEndpointBPFWatchdog),
)

type endpointBPFWatchdogParams struct {
  cell.In

  Config endpointBPFWatchdogConfig
  DaemonPromise promise.Promise[*Daemon]
  Endpoints endpointmanager.EndpointsLookup
  JobRegistry job.Registry
  Logger logrus.FieldLogger
}

func registerEndpointBPFProgWatchdog(lc hive.Lifecycle, p endpointBPFWatchdogParams) {
  g := p.JobRegistry.NewGroup()
  g.Add(job.Timer(
    "endpoint-bpf-watchdog",
    p.Config.EndpointBPFWatchdogInterval,
    func(ctx context.Context) error {
      return checkEndpointBPFPrograms(ctx, p)
    }))
  lc.Append(g)
}

func checkEndpointBPFPrograms(ctx context.Context, p endpointBPFWatchdogParams]) error {
  eps := p.EndpointsLookup.GetEndpoints()
  for ... { ... }
  if loaded { return nil }

  p.Logger.Debugf(...)

  daemon, err := p.DaemonPromise.Await(ctx)
  if err != nil {
    return err
  }

  wg, err := daemon.TriggerReloadWithoutCompile(epBPFProgWatchdog)
  if err != nil {
    return err 
  }
  wg.Wait() // FIXME should respect 'ctx' cancellation.
}
```

Since this doesn't provide any API to other modules, it just uses `cell.Invoke` to create a job group and register it's start and stop hooks into the hive's lifecycle. `pkg/hive/job` is essentially `pkg/controller` with lifecycle integrated with Hive plus other features (OneShot, Observable, etc.). We did this as we wanted to be able to inspect not only timer jobs but all background jobs (e.g. one-shot "daemon" functions that are retried on error). It's still lacking a REST API for inspecting the jobs though! We'd like over time to merge `pkg/controller` and `pkg/hive/job` (by implementing `pkg/controller` on top of `job.Timer`), but it's not on top of our TODO list right now.

This could be improved further by moving more things out from `*Daemon`. Using `promise.Promise[*Daemon]` is a sign that something should be fixed (it's there just for transitioning out from `*Daemon`). For example here we could move the `Trigger*` functions out from `*Daemon` into its own API. I think everything but `mtu.Configuration` are already available. Alternative that doesn't change things as much is providing a "bridge" by exposing the trigger functions as their own API, but implementing it in terms of `promise.Promise[*Daemon]` (`cell.Provide(func(dp promise.Promise[*Daemon) Triggers { ... })`, `type Triggers interface { TriggerReloadWithoutCompile(ctx context.Context, reason string) error }`). This is also a way to avoid having to implement everything in `daemon/cmd`. If `Triggers` type is defined elsewhere, then `daemon/cmd` can implement it, but cells that use it can be elsewhere in the codebase. Over time we can then migrate functionality away from `daemon/cmd` into places where they make more sense.

derailed: Very cool. Thank you Jussi! I'll take a pass. I think there is currently an issue with package deps trying to move things out of pkg cmd since the job cell would refer to the daemon via the promise but the cell deps still needs to be specified on the daemon. Will push my initial crack and we can refine from there.

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		if !loaded {
+			break
+		}
+	}
+	if loaded {
+		return nil
+	}
+
+	log.Debugf("Reloading endpoint BPF programs for (%d) endpoints.", len(eps))
```
christarazi: I would make this a warning because this is unexpected.

christarazi: For `len(eps)`, I'm sure we have something like `logfields.Count` or similar.

pkg/option/config.go
```diff
@@ -2473,6 +2479,12 @@ type DaemonConfig struct {
 
 	// EnableK8sNetworkPolicy enables support for K8s NetworkPolicy.
 	EnableK8sNetworkPolicy bool
+
+	// EndpointBPFProgsWatchdogRate tracks endpoint bpf progs loaded watchdog rate.
+	EndpointBPFProgsWatchdogRate time.Duration
```
christarazi: I would change this to 
```suggestion
	EndpointBPFProgsWatchdogInterval time.Duration
```

to match other config flags with a similar role.

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,137 @@
...
+				epBPFProgWatchdog,
+				controller.ControllerParams{
+					Group: controller.NewGroup(epBPFProgWatchdog),
+					DoFunc: func(ctx context.Context) error {
+						d, err := p.DaemonPromise.Await(context.Background())
+						if err != nil {
+							return err
+						}
+						if d.ctx.Err() != nil {
+							mgr.RemoveAllAndWait()
```
christarazi: I don't think we want to do this because it will kill all of the controllers. This controller should only care about itself, otherwise we'll have this code and the daemon-stopping logic stepping on each other's toes.

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,137 @@
...
+					RunInterval: p.Config.EndpointBPFProgWatchdogRate,
+					Context:     ctx,
+				},
+			)
+
+			return nil
+		},
+		OnStop: func(hive.HookContext) error {
+			cancel()
+			mgr.RemoveAllAndWait()
```
christarazi, line 88: Same here regarding the remove all. I think we just want to remove this controller from the manager and cancel the context.

daemon/cmd/watchdogs.go
derailed: This should only impact this cell's controller and not the daemon's.

pkg/defaults/defaults.go
```diff
@@ -544,6 +544,12 @@ const (
 
 	// EnableK8sNetworkPolicy enables support for K8s NetworkPolicy.
 	EnableK8sNetworkPolicy = true
+
+	// EndpointBPFProgsWatchdogRate tracks bpf progs watchdog rate.
```
joamaki: These can be dropped now

pkg/option/config.go
joamaki: Changes in this file can be dropped now that the config is in a cell.

daemon/cmd/watchdogs.go
joamaki: @derailed is right, it's only affecting the manager that's created here.

But do not call `RemoveAllAndWait` from `DoFunc`. It should only be called from the stop hook.
And do not use `d.ctx`. This cell is separate from the daemon and thus should not use its context. And since this is depending on the daemon, it means it'll be stopped before daemon and thus `d.ctx` will never be cancelled before the stop hook of this cell is called, e.g. this path is never reached.

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,137 @@
...
+		mgr         = controller.NewManager()
+	)
+	p.Lifecycle.Append(hive.Hook{
+		OnStart: func(hive.HookContext) error {
+			mgr.UpdateController(
+				epBPFProgWatchdog,
+				controller.ControllerParams{
+					Group: controller.NewGroup(epBPFProgWatchdog),
+					DoFunc: func(ctx context.Context) error {
+						d, err := p.DaemonPromise.Await(context.Background())
```
joamaki: Use `ctx` here, don't make a new one.

joamaki (changes requested):

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,137 @@
...
+
+	Config        epBPFProgWatchdogConfig
+	Logger        logrus.FieldLogger
+	Lifecycle     hive.Lifecycle
+	DaemonPromise promise.Promise[*Daemon]
+}
+
+var (
+	// EndpointBPFrogWatchdogCell triggers a job to ensure device tc programs remain loaded.
+	EndpointBPFrogWatchdogCell = cell.Module(
```
joamaki: No need to export this variable.

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,137 @@
...
+			break
+		}
+	}
+	if loaded {
+		return nil
+	}
+
+	log.WithField(logfields.Count, len(eps)).Warn("Reloading endpoints BPF programs.")
+	wg, err := d.TriggerReloadWithoutCompile(epBPFProgWatchdog)
+	if err != nil {
```
joamaki, line 129: nit: 
```
wg, err := ...
if err == nil {
  wg.Wait()
}
return err
```

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,137 @@
...
+type epBPFProgWatchdogConfig struct {
+	EndpointBPFProgWatchdogInterval time.Duration
+	EndpointBPFProgWatchdogEnabled  bool
+}
+
+func (c epBPFProgWatchdogConfig) Flags(flags *pflag.FlagSet) {
+	flags.Bool("endpoint-bpf-prog-watchdog-enabled", c.EndpointBPFProgWatchdogEnabled,
+		"Enable endpoint BPF program watchdog")
+
+	flags.Duration("endpoint-bpf-prog-watchdog-rate", c.EndpointBPFProgWatchdogInterval,
```
joamaki: This should be `"endpoint-bpf-prog-watchdog-interval"`. CI currently fails due to the mismatch:
```
2023-09-15T00:27:07.709604887Z level=error msg="Invoke failed" 
"cmd.registerEndpointBPFProgWatchdog (watchdogs.go:62)" error="could not build arguments for 
function \"github.com/cilium/cilium/daemon/cmd\".registerEndpointBPFProgWatchdog (/go/src/github.com/cilium/cilium/daemon/cmd/watchdogs.go:62): failed to build cmd.epBPFProgWatchdogConfig: received non-nil error
 from function \"github.com/cilium/cilium/pkg/hive/cell\".(*config[...]).Apply.func2 (/go/src/github.com/cilium/cilium
/pkg/hive/cell/config.go:136): failed to unmarshal config struct cmd.epBPFProgWatchdogConfig: 2 error(s) 
decoding:\n\n* '' has invalid keys: endpoint-bpf-prog-watchdog-rate\n* '' has unset fields: EndpointBPFProgWatchdogInterval.\nHint: field 'FooBar' matches flag 'foo-bar', or use tag 
`mapstructure:\"flagname\"` to match field with flag" subsys=hive
```

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,137 @@
...
+
+const epBPFProgWatchdog = "ep-bpf-prog-watchdog"
+
+type epBPFProgWatchdogConfig struct {
+	EndpointBPFProgWatchdogInterval time.Duration
+	EndpointBPFProgWatchdogEnabled  bool
+}
+
+func (c epBPFProgWatchdogConfig) Flags(flags *pflag.FlagSet) {
+	flags.Bool("endpoint-bpf-prog-watchdog-enabled", c.EndpointBPFProgWatchdogEnabled,
```
joamaki: Having just the interval with `0` being disabled is probably enough. We have too many flags already...

daemon/cmd/watchdogs.go
joamaki: To test the agent hive instantiation: `go test ./daemon/cmd -test.run TestAgentCell -test.v`:
```
    cells_test.go:36: 
        	Error Trace:	/home/jussi/work/cilium2/daemon/cmd/cells_test.go:36
        	Error:      	Received unexpected error:
        	           	could not build arguments for function "github.com/cilium/cilium/daemon/cmd".registerEndpointBPFProgWatchdog
        	           		/home/jussi/work/cilium2/daemon/cmd/watchdogs.go:62:
        	           	failed to build cmd.epBPFProgWatchdogConfig:
        	           	received non-nil error from function "github.com/cilium/cilium/pkg/hive/cell".(*config[...]).Apply.func2
        	           		/home/jussi/work/cilium2/pkg/hive/cell/config.go:136:
        	           	failed to unmarshal config struct cmd.epBPFProgWatchdogConfig: 2 error(s) decoding:
        	           	
        	           	* '' has invalid keys: endpoint-bpf-prog-watchdog-rate
        	           	* '' has unset fields: EndpointBPFProgWatchdogInterval.
        	           	Hint: field 'FooBar' matches flag 'foo-bar', or use tag `mapstructure:"flag-name"` to match field with flag
        	Test:       	TestAgentCell
        	Messages:   	Populate()
--- FAIL: TestAgentCell (0.27s)
```

(This really should be somewhere in the docs.. at the very least big comment in `cells.go`... Hmm.)

joamaki: Actually even better is:
```
% go run ./daemon hive                                                                           1 ↵
level=info msg="Memory available for map entries (0.003% of 67377799168B): 168444497B" subsys=config
level=info msg="option bpf-ct-global-tcp-max set by dynamic sizing to 591033" subsys=config
level=info msg="option bpf-ct-global-any-max set by dynamic sizing to 295516" subsys=config
level=info msg="option bpf-nat-global-max set by dynamic sizing to 591033" subsys=config
level=info msg="option bpf-neigh-global-max set by dynamic sizing to 591033" subsys=config
level=info msg="option bpf-sock-rev-map-max set by dynamic sizing to 295516" subsys=config
level=warning msg="Running Cilium with \"kvstore\"=\"\" requires identity allocation via CRDs. Changing identity-allocation-mode to \"crd\"" subsys=config
level=error msg="Invoke failed" ="cmd.registerEndpointBPFProgWatchdog (watchdogs.go:62)" error="could not build arguments for function \"github.com/cilium/cilium/daemon/cmd\".registerEndpointBPFProgWatchdog (/home/jussi/work/cilium2/daemon/cmd/watchdogs.go:62): failed to build cmd.epBPFProgWatchdogConfig: received non-nil error from function \"github.com/cilium/cilium/pkg/hive/cell\".(*config[...]).Apply.func2 (/home/jussi/work/cilium2/pkg/hive/cell/config.go:136): failed to unmarshal config struct cmd.epBPFProgWatchdogConfig: 2 error(s) decoding:\n\n* '' has invalid keys: endpoint-bpf-prog-watchdog-rate\n* '' has unset fields: EndpointBPFProgWatchdogInterval.\nHint: field 'FooBar' matches flag 'foo-bar', or use tag `mapstructure:\"flag-name\"` to match field with flag" subsys=hive
level=fatal msg="Failed to populate object graph" error="could not build arguments for function \"github.com/cilium/cilium/daemon/cmd\".registerEndpointBPFProgWatchdog (/home/jussi/work/cilium2/daemon/cmd/watchdogs.go:62): failed to build cmd.epBPFProgWatchdogConfig: received non-nil error from function \"github.com/cilium/cilium/pkg/hive/cell\".(*config[...]).Apply.func2 (/home/jussi/work/cilium2/pkg/hive/cell/config.go:136): failed to unmarshal config struct cmd.epBPFProgWatchdogConfig: 2 error(s) decoding:\n\n* '' has invalid keys: endpoint-bpf-prog-watchdog-rate\n* '' has unset fields: EndpointBPFProgWatchdogInterval.\nHint: field 'FooBar' matches flag 'foo-bar', or use tag `mapstructure:\"flag-name\"` to match field with flag" subsys=hive
```

daemon/cmd/watchdogs.go
derailed: Tx Jussi! Good point.

pkg/defaults/defaults.go
derailed: Spring cleaning in full effect!

daemon/cmd/watchdogs.go
derailed: Good catch! Thank you.

daemon/cmd/watchdogs.go
derailed: Indeed!

daemon/cmd/watchdogs.go
derailed: Much better.

daemon/cmd/watchdogs.go
derailed: Nice! I did not know that. Thank you for the pointer!

daemon/cmd/watchdogs.go
derailed: Makes sense.

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,127 @@
...
+	Logger        logrus.FieldLogger
+	Lifecycle     hive.Lifecycle
+	DaemonPromise promise.Promise[*Daemon]
+}
+
+var (
+	// endpointBPFrogWatchdogCell triggers a job to ensure device tc programs remain loaded.
+	endpointBPFrogWatchdogCell = cell.Module(
+		epBPFProgWatchdog,
+		"Periodically checks that endpoints BPF programs remain loaded",
```
christarazi: Nit: 

```suggestion
		"Periodically checks that endpoints' BPF programs remain loaded",
```

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+		if !loaded {
+			break
+		}
+	}
+	if loaded {
+		return nil
+	}
+
+	log.WithField(logfields.Count, len(eps)).Warn("Reloading endpoints BPF programs.")
```
christarazi: Hm, I think this warning should be louder,

```suggestion
	log.WithField(logfields.Count, len(eps)).Warn("Detected and resolved unexpected endpoint BPF program removal. Consider investigating whether other software running on this machine is removing Cilium's endpoint BPF programs. If endpoint BPF programs are removed, the associated pods will lose connectivity and only reinstating the programs will restore connectivity.")
```

daemon/cmd/daemon_main.go
```diff
@@ -1918,6 +1919,7 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
 	if err != nil {
 		log.WithError(err).Error("Unable to store Viper's configuration")
 	}
+
```
christarazi: Unintended newline?

christarazi (approved):
LGTM, very clean! Just a few minor comments left.

daemon/cmd/watchdogs.go
derailed: Perfect!

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+const epBPFProgWatchdog = "ep-bpf-prog-watchdog"
+
+type epBPFProgWatchdogConfig struct {
+	EndpointBPFProgWatchdogInterval time.Duration
+}
+
+func (c epBPFProgWatchdogConfig) Flags(flags *pflag.FlagSet) {
+	flags.Duration("endpoint-bpf-prog-watchdog-interval", c.EndpointBPFProgWatchdogInterval,
```
tommyp1ckles, line 29: Just wondering if this is the pattern we want to follow for future such reconcilers (related to endpoint). Does it make sense to make this more general so we can capture further?

thorn3r (approved):
This looks excellent! Great work :)

pkg/datapath/loader/netlink.go
```diff
@@ -692,3 +694,33 @@ func renameDevice(from, to string) error {
...
+	if !found {
+		return false, nil
+	}
+
+	ff, err := netlink.FilterList(l, netlink.HANDLE_MIN_INGRESS)
+	if err != nil {
+		return false, fmt.Errorf("unable to fetch ingress filter list: %w", err)
+	}
+
+	return len(ff) > 0, nil
```
rgo3: If we just check that there is at least one program attached, we are not detecting if someone has removed our ep prog by replacing it with something else, which to me seems more likely to happen than just simple removal? Should we specifically check if cilium's ep progs are attached, e.g. the expected program name matches?

pkg/datapath/loader/netlink.go
```diff
@@ -692,3 +694,33 @@ func renameDevice(from, to string) error {
...
+		if d.Type() == qdiscClsact {
+			found = true
+			break
+		}
+	}
+	if !found {
+		return false, nil
+	}
+
+	ff, err := netlink.FilterList(l, netlink.HANDLE_MIN_INGRESS)
```
rgo3, line 723: Why only `INGRESS`? Looking at the respective loader code ([`reloadDatapath()`](https://github.com/cilium/cilium/blob/76aad59278e3d3e6b48a5e214fee1ae1a12a88ba/pkg/datapath/loader/loader.go#L396)) we also attach to `EGRESS` if the endpoint requires it.

pkg/datapath/loader/netlink.go
```diff
@@ -692,3 +694,33 @@ func renameDevice(from, to string) error {
 
 	return nil
 }
+
+// DeviceHasTCProgramLoaded checks whether a give device tc filter/qdisc progs are missing.
```
rgo3: Nit:
```suggestion
// DeviceHasTCProgramLoaded checks whether a given device has tc filter/qdisc progs attached.
```

rgo3 (changes requested):
Overall LGTM. Not sure if this has already been noted in previous review comments but I have some questions regarding when we should trigger a reload. Currently we only reload if there are no ep progs attached to tc INGRESS at all, but there are more failure cases (replacement) that could be covered with the proposed watchdog by checking if attached programs match the programs that cilium expects.

pkg/datapath/loader/netlink.go
derailed: Thank you for the review Robin! Initially I was checking both but after chatting with @christarazi he felt initially we only need to check the ingress side of the device with an option to open up egress since egress are managed by the daemon separately.

pkg/datapath/loader/netlink.go
derailed: Good point! Not sure what names we should check for as it may require additional context. 
It seems we are prefixing our prog names with `cil_*` is that what you were thinking? 
Also this feature will change when we leverage links to provide proper ownership for our bpf progs.

pkg/datapath/loader/netlink.go
christarazi: To clarify, if it's possible to check for both ingress and egress then we should do it as Robin suggests. I recall that when Fernand and I were trying out `tc` commands to see what the TC filters looked like, we only found ingress. We assumed that egress was hidden in the endpoint's netns and therefore wasn't visible to Cilium in the host netns. However, we could totally be off base and missing something here, so @rgo3 it would be great if you can chime in and clarify.

daemon/cmd/watchdogs.go
tommyp1ckles: (just wondering - not going to hold up this PR so feel free to dismiss this).

christarazi: To me, it's not yet clear if there is a need for a general pattern. It seems like it's best to consider it case by case and if we see a pattern emerge, then we can reevaluate.

pkg/datapath/loader/netlink.go
derailed: @rgo3 I've taken another pass to beef up the checks.

pkg/datapath/loader/netlink.go
derailed: @rgo3 Added checks for egress when present.

daemon/cmd/watchdogs.go
derailed: Agreed - use before reuse!

pkg/datapath/loader/netlink.go
rgo3: My question came from looking at `reloadDatapath()` as linked in the first comment. If `ep.RequireEgressProg()` is true we also attach an egress prog to the same interface as the ingress prog, however I don't know when this condition is true. Couldn't we leverage that function or it's returned value as a parameter to know when we need to check if an egress prog needs to be present on the interface of the endpoint?

pkg/datapath/loader/netlink.go
```diff
@@ -692,3 +695,62 @@ func renameDevice(from, to string) error {
 
 	return nil
 }
+
+// DeviceHasTCProgramLoaded checks whether a given device has tc filter/qdisc progs attached.
+func DeviceHasTCProgramLoaded(ifc string) (bool, error) {
```
rgo3: What if this function takes the direction as an argument and then the caller can basically do:
```
loaded, err = loader.DeviceHasTCProgramLoaded(ep.HostInterface(), INGRESS)
...
if ep.RequiresEgressProg {
    loaded, err = loader.DeviceHasTCProgramLoaded(ep.HostInterface(), EGRESS)
    ...
}
...
```
I think this would reduce the complexity of this function and make it easy to know when we need to check for egress progs. WDYT?

rgo3 (changes requested):

pkg/datapath/loader/netlink.go
```diff
@@ -692,3 +695,62 @@ func renameDevice(from, to string) error {
...
+	}
+	filtersCount = 0
+	for _, f := range ff {
+		if filter, ok := f.(*netlink.BpfFilter); ok {
+			if strings.HasPrefix(filter.Name, bpfProgPrefix) {
+				filtersCount++
+			}
+		}
+	}
+	if len(ff) > 0 && filtersCount == 0 {
```
asauber: To me, it also looks like a more reliable check here would be based on `ep.RequireEgressProg()` rather than the presence of egress filters.

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,132 @@
...
+			continue
+		}
+		loaded, err = loader.DeviceHasTCProgramLoaded(ep.HostInterface())
+		if err != nil {
+			log.WithField(logfields.Endpoint, ep.HostInterface()).
+				WithError(err).
+				Error("Unable to assert if endpoint BPF programs need to be reloaded")
+			return err
+		}
+		if !loaded {
```
asauber, line 114: A comment here along the lines of "If we detect a single unloaded program, then we break here and make use of the Daemon's existing functionality to reload all programs" would help with readability.

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,132 @@
...
+			break
+		}
+	}
+	if loaded {
+		return nil
+	}
+
+	log.WithField(logfields.Count, len(eps)).
+		Warn(
+			"Detected and resolved unexpected endpoint BPF program removal. " +
```
asauber: It seems strange to log "Detected and resolved", when we haven't yet completed the resolution (the function below could return an error). Perhaps simply say "detected" here, and log separately after the reload and/or add a log to reload that indicates that the reload completed.

asauber (changes requested):

daemon/cmd/watchdogs.go
christarazi: Good point. I like this idea because it avoids confusion in case we somehow fail to reload. Potentially some other software, probably the same one which removed Cilium BPF progs, could have locked down permissions in some way, locking out Cilium.

pkg/datapath/loader/netlink.go
derailed: I am not keen on the idea. I think leaving this up to the caller to decide is less than ideal. IMHO best to change the call to take in the ep and let it decide whether it should check for egress based on the ep nature. 
Would this make better sense?

pkg/datapath/loader/netlink.go
derailed: Great point! Thank you Robin. I'll update per your suggestions

daemon/cmd/watchdogs.go
derailed: Make sense! Tx Andrew! I'll update...

pkg/datapath/loader/netlink.go
derailed: Agreed!

joamaki (approved):

asauber (approved):

rgo3 (approved):

daemon/cmd/watchdogs.go
```diff
@@ -0,0 +1,136 @@
...
+		OnStop: func(hive.HookContext) error {
+			cancel()
+			mgr.RemoveAllAndWait()
+			return nil
+		},
+	},
+	)
+}
+
+func checkEndpointBPFPrograms(ctx context.Context, d *Daemon, p epBPFProgWatchdogParams) error {
```
christarazi: Minor: Shouldn't this function be

```suggestion
func (d *Daemon) checkEndpointBPFPrograms(ctx context.Context, d *Daemon, p epBPFProgWatchdogParams) error {
```

anyway since we pass `d` in?

derailed: Tx Chris! Could go either way on this... I think that makes sense given the rest of the package shape. I'll update

PR #27985, opened 2023-09-07
endpoint: Fix use of PolicyMapFullReconciliationInterval option

joamaki:
The last minute change in 867e41a610 to add an option for configuring the full endpoint policy sync forgot to actually change the use-site.

Fixes: 867e41a610 ("endpoint: Only perform full synchronization periodically")

### DIFF
--- a/pkg/endpoint/bpf.go
+++ b/pkg/endpoint/bpf.go
@@ -1345,12 +1345,6 @@ func (e *Endpoint) syncPolicyMapWithDump() error {
 	return err
 }
 
-const (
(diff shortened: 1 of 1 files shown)


joamaki:
Marking as `release-note/misc` since this fixes an issue in just merged PR that never made it into a release.

joamaki:
I validated that this option now correctly works by setting it to `1m` and running `cilium status --verbose` and validated that the controller indeed ran once per minute.

aanm (approved):

gandro:
Adding `backport/author` for v1.12 and v1.13 (I'll take care of v1.14)

PR #27997, opened 2023-09-07
ipam: let `allocator.Dump` return map of allocated IPs per pool

tklauser:
The first commit is preparatory refactoring in unit tests to simplify the second commit. The second commit contains the actual change:

Instead of encoding the pool name in the IP string, let `Allocator.Dump` return the pool information separately, i.e. a map of allocated IPs per pool. `(*IPAM).Dump` will keep returning a map of allocated IPs with their owner in order not to change the interface towards other packages.
    
Follow-up for commit afd8e4eef3b6 ("ipam: report IP owner of non-default pool IPs in multi-pool IPAM"), as suggested by @gandro in https://github.com/cilium/cilium/pull/27968#pullrequestreview-1613446046

### DIFF
--- a/pkg/ipam/allocator.go
+++ b/pkg/ipam/allocator.go
@@ -312,40 +312,46 @@ func (ipam *IPAM) ReleaseIP(ip net.IP, pool Pool) error {
 	return ipam.releaseIPLocked(ip, pool)
 }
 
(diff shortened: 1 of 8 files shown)


gandro (approved):
Awesome cleanup, thanks!

PR #28002, opened 2023-09-07
k8s: remove extensions/v1beta1 support

tklauser:
This API is no longer supported since k8s 1.16 [1] which is the minimum required k8s version for Cilium since version 1.10 [2], ref. commit 83b8f8b9d514 ("docs: bump minimal supported k8s version to 1.16").

[1] https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/
[2] https://docs.cilium.io/en/latest/network/kubernetes/compatibility/#k8scompatibility


youngnick (approved):
Nice.

PR #28003, opened 2023-09-07
labelsfilter: ignore StatefulSet-related labels by default for CID creation

tosi3k:
Beforehand, every single Pod owned by a StatefulSet object had its own `CiliumIdentity` because each such Pod had their own unique set of label values (more about these labels can be found [here](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-name-label) and [here](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-index-label)).

In large clusters with many StatefulSet-owned Pods it is problematic, especially when the churn of such objects is high, resulting in a high churn of such CIDs that doesn't scale well. In addition, it essentially limited the number of such Pods to the theoretical number of CIDs in a cluster which is 2^16 ~ 65k.

```release-note
Ignore StatefulSet-specific labels by default for CID creation. This includes the two following labels:
* statefulset.kubernetes.io/pod-name
* apps.kubernetes.io/pod-index
```

### DIFF
--- a/Documentation/operations/performance/scalability/identity-relevant-labels.rst
+++ b/Documentation/operations/performance/scalability/identity-relevant-labels.rst
@@ -21,19 +21,21 @@ application.
 By default, Cilium considers all labels to be relevant for identities, with the
 following exceptions:
 
-================================== ==============================================
-Label                               Description
----------------------------------- ----------------------------------------------
-``any:!io.kubernetes``             Ignore all ``io.kubernetes`` labels
-``any:!kubernetes\.io``            Ignore all other ``kubernetes\.io`` labels
-``any:!beta.kubernetes\.io``       Ignore all ``beta\.kubernetes\.io`` labels
-``any:!k8s\.io``                   Ignore all ``k8s\.io`` labels
-``any:!pod-template-generation``   Ignore all ``pod-template-generation`` labels
-``any:!pod-template-hash``         Ignore all ``pod-template-hash`` labels
-``any:!controller-revision-hash``  Ignore all ``controller-revision-hash`` labels
-``any:!annotation.*``              Ignore all ``annotation`` labels
-``any:!etcd_node``                 Ignore all ``etcd_node`` labels
-================================== ==============================================
+============================================= =====================================================
+Label                                         Description
+--------------------------------------------- -----------------------------------------------------
+``any:!io.kubernetes``                        Ignore all ``io.kubernetes`` labels
+``any:!kubernetes\.io``                       Ignore all other ``kubernetes.io`` labels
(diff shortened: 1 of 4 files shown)


alan-kut (approved):

Documentation/operations/performance/scalability/identity-relevant-labels.rst
```diff
@@ -21,19 +21,21 @@ application.
...
-``any:!controller-revision-hash``  Ignore all ``controller-revision-hash`` labels
-``any:!annotation.*``              Ignore all ``annotation`` labels
-``any:!etcd_node``                 Ignore all ``etcd_node`` labels
-================================== ==============================================
+============================================= =====================================================
+Label                                         Description
+--------------------------------------------- -----------------------------------------------------
+``any:!io.kubernetes``                        Ignore all ``io.kubernetes`` labels
+``any:!kubernetes\.io``                       Ignore all other ``kubernetes\.io`` labels
+``any:!statefulset\.kubernetes\.io/pod-name`` Ignore ``statefulset\.kubernetes\.io/pod-name`` label
```
ghost:
```suggestion
``any:!statefulset\.kubernetes\.io/pod-name`` Ignore ``statefulset.kubernetes.io/pod-name`` label
```
Formatting is already inconsistent in this table, so I wouldn't block on this suggestion really.

ghost (approved):
Docs ok, overall the change looks reasonable to me.

nathanjsweet (approved):

lambdanis:
@tosi3k CI is unhappy with the long commit message (https://github.com/cilium/cilium/actions/runs/6187912035/job/16803422512?pr=28003):
```
{[1/2] labelsfilter: ignore StatefulSet-specific labels by default for CID creation}

Error: ERROR:CUSTOM: Please avoid long commit subjects (max: 75, found: 76)
```

Could you please reword it?

tosi3k:
> @tosi3k CI is unhappy with the long commit message (https://github.com/cilium/cilium/actions/runs/6187912035/job/16803422512?pr=28003):
> 
> ```
> {[1/2] labelsfilter: ignore StatefulSet-specific labels by default for CID creation}
> 
> Error: ERROR:CUSTOM: Please avoid long commit subjects (max: 75, found: 76)
> ```
> 
> Could you please reword it?

Done; thanks for pointing this out.

joestringer:
Hi @tosi3k, it looks like this PR introduced upgrade impact for users upgrading to Cilium v1.15 because users may have policies that rely on some of the labels that are now ignored. This issue comment goes into detail about the issues behind this and some suggestions about how to improve: https://github.com/cilium/cilium/issues/32213#issuecomment-2081653111 .

PR #28005, opened 2023-09-07
Operator modular metrics

pippolo84:
Add a cell to provide global metrics registry and prometheus metrics server for the operator hive.

This makes possible to use `cell.Metric` to export metrics in the operator cells.

Since there are parts of the operator code not yet modularized, the legacy metrics variables and the global registry are kept exported, in order to be directly used by legacy code. This approach is inspired by the agent metrics cell.

Besides, this PR fixes the operator metrics namespace, setting it early at startup, so that all the code shared between other components (e.g: the agent) will report the correct `cilium_operator` namespace when updating the metrics.

Here a query that shows both the correct namespace for all the metrics and also all the metrics registered with `cell.Metric` (like the ones in the `operator/pkg/lbipam` cell) showing up:

```
$ curl localhost:9234/v1/metrics
[
    {
        "name": "cilium_operator_ces_queueing_delay_seconds"
    },
    {
        "name": "cilium_operator_ces_sync_errors_total"
    },
    {
        "name": "cilium_operator_lbipam_conflicting_pools_total"
    },
    {
        "name": "cilium_operator_lbipam_services_matching_total"
    },
    {
        "name": "cilium_operator_lbipam_services_unsatisfied_total"
    },
    {
        "name": "cilium_operator_number_of_ceps_per_ces"
    },
    {
        "name": "cilium_operator_process_cpu_seconds_total",
        "value": 1.03
    },
    {
        "name": "cilium_operator_process_max_fds",
        "value": 1048576
    },
    {
        "name": "cilium_operator_process_open_fds",
        "value": 12
    },
    {
        "name": "cilium_operator_process_resident_memory_bytes",
        "value": 95997952
    },
    {
        "name": "cilium_operator_process_start_time_seconds",
        "value": 1694106767.7
    },
    {
        "name": "cilium_operator_process_virtual_memory_bytes",
        "value": 1351573504
    },
    {
        "name": "cilium_operator_process_virtual_memory_max_bytes",
        "value": 18446744073709552000
    }
]
```

### DIFF
--- a/operator/metrics/cell.go
+++ b/operator/metrics/cell.go
@@ -1,0 +1,56 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package metrics
+
+import (
+	"github.com/spf13/pflag"
+
+	"github.com/cilium/cilium/pkg/hive/cell"
+)
+
+const (
+	// OperatorPrometheusServeAddr IP:Port on which to serve prometheus metrics
+	// (pass ":<port>" to bind on all interfaces).
+	OperatorPrometheusServeAddr = "operator-prometheus-serve-addr"
+)
+
+// Cell provides the modular metrics registry, metric HTTP server and
+// legacy metrics cell for the operator.
+var Cell = cell.Module(
+	"operator-metrics",
+	"Operator Metrics",
+
+	cell.Config(defaultConfig),
+	cell.Invoke(registerMetricsManager),
+
+	// Operator legacy metrics accessed via global variables
+	cell.Metric(newLegacyMetrics),
+)
+
+// Config contains the configuration for the operator-metrics cell.
+type Config struct {
+	OperatorPrometheusServeAddr string
+}
+
+var defaultConfig = Config{
+	// default server address for operator metrics
+	OperatorPrometheusServeAddr: ":9963",
+}
+
+func (def Config) Flags(flags *pflag.FlagSet) {
+	flags.String(OperatorPrometheusServeAddr, def.OperatorPrometheusServeAddr, "Address to serve Prometheus metrics")
+}
+
+// SharedConfig contains the configuration that is shared between
+// this module and others.
+// Metrics cell needs to know if GatewayAPI is enabled in order to use
+// the same Registry as controller-runtime and avoid to expose
+// multiple metrics endpoints or servers.
+type SharedConfig struct {
+	// EnableMetrics is set to true if operator metrics are enabled
+	EnableMetrics bool
+
+	// EnableGatewayAPI enables support of Gateway API
+	EnableGatewayAPI bool
+}
--- a/operator/metrics/legacy.go
+++ b/operator/metrics/legacy.go
@@ -1,0 +1,213 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+
+	"github.com/cilium/cilium/api/v1/operator/models"
+	"github.com/cilium/cilium/pkg/metrics"
+	"github.com/cilium/cilium/pkg/metrics/metric"
+)
+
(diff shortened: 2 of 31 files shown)


operator/metrics/cell.go
```diff
@@ -0,0 +1,59 @@
...
+var defaultConfig = Config{
+	// default server address for operator metrics
+	OperatorPrometheusServeAddr: ":9963",
+}
+
+func (def Config) Flags(flags *pflag.FlagSet) {
+	flags.String(OperatorPrometheusServeAddr, def.OperatorPrometheusServeAddr, "Address to serve Prometheus metrics")
+}
+
+type MetricsConfig struct {
```
joamaki: This can be removed?

operator/metrics/legacy.go
```diff
@@ -0,0 +1,212 @@
...
+	// It is replaced by CiliumEndpointSliceSyncTotal metric.
+	CiliumEndpointSliceSyncErrors = metrics.NoOpCounter
+
+	// CiliumEndpointSliceQueueDelay measures the time spent by CES's in the workqueue. This measures time difference between
+	// CES insert in the workqueue and removal from workqueue.
+	CiliumEndpointSliceQueueDelay = metrics.NoOpHistogram
+)
+
+type legacyMetrics struct {
+	IdentityGCSize                metric.Vec[metric.Gauge]
```
joamaki, line 87: Let's do a follow-up to remove `legacyMetrics`. Should be pretty doable to do all these after the CES PR lands?

joamaki (approved):

operator/metrics/cell.go
pippolo84: Yep, that's just a leftover from a previous iteration :sweat_smile:

operator/metrics/legacy.go
pippolo84: Definitely. Both the identity GC and the CES controller have been modularized, so we can switch to `cell.Metric` and remove all of this. :+1:

pippolo84:
Rebased to fix conflicts after merging of https://github.com/cilium/cilium/pull/26836

operator/metrics/legacy.go
```diff
@@ -0,0 +1,212 @@
...
+			Namespace: metrics.CiliumOperatorNamespace,
+			Name:      "ces_sync_errors_total",
+			Help:      "Number of CES sync errors",
+		}),
+
+		CiliumEndpointSliceSyncTotal: metric.NewCounterVec(metric.CounterOpts{
+			Namespace: metrics.CiliumOperatorNamespace,
+			Name:      "ces_sync_total",
+			Help:      "The number of completed CES syncs by outcome",
+		}, []string{"outcome"}),
```
derailed: Should we use the defined label aka LabelOutcome?

operator/metrics/legacy.go
```diff
@@ -0,0 +1,212 @@
...
+		return result, err
+	}
+
+	for _, val := range currentMetrics {
+
+		metricName := val.GetName()
+		metricType := val.GetType()
+
+		for _, metricLabel := range val.Metric {
+			labels := map[string]string{}
```
derailed: nit: prealloc the map? ie make(map[string]string, len(metricLabel.GetLabel()))

derailed (commented):
@pippolo84 Nice work Fabio! Just a couple comments...

pippolo84:
> @pippolo84 Nice work Fabio! Just a couple comments...

Thanks for the feedback. Fixed it!

christarazi (approved):

pippolo84:
Added a commit to remove the stale `enableMetrics` bool in identitygc (no need to check for metrics enabled flag anymore, since we switched to NoOp* types).

pippolo84:
Conformance Ingress failure tracked [here](https://github.com/cilium/cilium/issues/28169), rerunning

derailed (approved):
@pippolo84 LGTM. Thanks for the updates!

tommyp1ckles (approved):
IPAM changes look good

lmb:
I resolved a pending conversation to make this mergeable: https://github.com/cilium/cilium/pull/28005/files#r1322543478

PR #28007, opened 2023-09-07
k8s: Restrict configuring reserved:init policy via CNP

joestringer:
Typically if the policy target EndpointSelector does not include a
namespace, parsing will inject the namespace to ensure that only Pods
within the same namespace can be selected. Furthermore, for peer
EndpointSelectors there is a similar behaviour where ingress or egress
is allowed only to the current namespace unless the user specifies which
namespace the peer exists in. However, these defaults did not previously
apply to policies that select the reserved:init Identity, since this
Identity is not inherently namespaced. Automatically injecting the
namespace would cause the policy to no longer match this Identity.

The reserved:init Identity was introduced in order to allow a different
policy to apply to Pods as they start up, iff the full Identity of those
workloads cannot be determined when they are first connecting to the
Cilium network (CNI ADD). In order to support this special reserved:init
Identity, various exceptions were introduced into CiliumNetworkPolicy
parser that removed the automatic insertion of namespace into the
policies when there were label matches for the reserved:init namespace.

The awkward part about this abstraction early on was that there was no
way to directly associate the reserved:init Identity to any specific
Kubernetes namespace, and yet CiliumNetworkPolicy was always inherently
tied to a specific namespace. Since the introduction of the
reserved:init Identity, CiliumClusterwideNetworkPolicy was also
introduced, which is a much better fit for the ability to select
endpoints that are not inherently namespaced. This is because CCNP is
also not restricted to any specific namespace, it applies to Endpoints
in all namespaces. This patch proposes to even extend that policy to
apply to Endpoints that are not inside a namespace, such as Endpoints
with the `reserved:init` Identity.

This patch removes support for applying network policies for the
reserved:init Identity via CiliumNetworkPolicy in favour of
CiliumClusterwideNetworkPolicy. As a benefit, this allows us to simplify
the logic that applies the namespaces into the policies and reduce the
likelihood of misconfigurations.

### DIFF
--- a/Documentation/operations/upgrade.rst
+++ b/Documentation/operations/upgrade.rst
@@ -315,6 +315,11 @@ Annotations:
   Upgrading from Cilium 1.14.x or earlier to 1.15.y or later does not
   trigger this problem. Downgrading from Cilium 1.15.y or later to Cilium
   1.14.x or earlier may trigger this problem.
+* ``CiliumNetworkPolicy`` cannot match the ``reserved:init`` labels any more.
+  If you have ``CiliumNetworkPolicy`` resources that have a match for
+  labels ``reserved:init``, these policies must be converted to
+  ``CiliumClusterwideNetworkPolicy`` by changing the resource type for the
+  policy.
 
 .. _upgrade_cilium_cli_helm_mode:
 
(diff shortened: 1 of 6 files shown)


ghost (approved):
docs ok

aanm (approved):

PR #28011, opened 2023-09-08

jrfastab:
took some longer time. For example I ran ip x s in a pod about
received are dropped.

traffic with the new IPs.

XfrmInPolicy errors I've seen.


jrfastab:
hold as draft until main branch fix is merged.

PR #28014, opened 2023-09-08
Use image specified by Chart.yaml by default

jschwinger233:


jschwinger233:
Should open a PR from branch instead of fork, forget again...

PR #28016, opened 2023-09-08
CI: Let actions/cilium-config use Chart.yaml-specified image by default

jschwinger233:
Previously cilium-config action always generates helm-set flags for image settings, but in some cases we can just rely on Chart.yaml since we always set chart-dir. This helps when:

1. We want to install release version instead of ci version. Currently cilium-config action sets `cilium-ci` unconditionally.
2. We have complicated image tag requirements such as `v1.14.1-beta.1`.

### DIFF
--- a/.github/workflows/tests-ipsec-upgrade.yaml
+++ b/.github/workflows/tests-ipsec-upgrade.yaml
@@ -129,7 +129,6 @@ jobs:
         id: cilium-stable-config
         uses: ./.github/actions/cilium-config
         with:
-          image-tag: ${{ steps.vars.outputs.downgrade_version }}
           chart-dir: './cilium-${{ steps.vars.outputs.downgrade_version }}/install/kubernetes/cilium/'
           tunnel: ${{ matrix.tunnel }}
           endpoint-routes: ${{ matrix.endpoint-routes }}
--- a/.github/actions/cilium-config/action.yml
+++ b/.github/actions/cilium-config/action.yml
@@ -3,7 +3,7 @@ description: Derive Cilium installation config
 inputs:
   image-tag:
     description: 'SHA or tag'
-    required: true
+    required: false
   chart-dir:
     description: 'Path to Cilium charts directory'
     required: true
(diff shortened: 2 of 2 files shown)


brb (approved):
Thanks!

.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -129,7 +129,6 @@ jobs:
         id: cilium-stable-config
         uses: ./.github/actions/cilium-config
         with:
-          image-tag: ${{ steps.vars.outputs.downgrade_version }}
```
nebril, line 132: Won't this cause downgrade to `cilium/cilium-ci:latest`, which is AFAIU dev version of 1.15?

I am not sure if we want to test downgrade between PR version and current main instead of PR version and previous stable version.

nebril (changes requested):
Thanks for the PR! The action change looks good, but I left a comment in workflow file to make sure that this change does what we want.

.github/workflows/tests-ipsec-upgrade.yaml
jschwinger233: Without `image-tag`, cilium-config action gives installation flags without specific image repos and tags. Therefore, when `CILIUM_CLI_MODE=helm ./cilium-cli install` runs, it's `--chart-directory` which has a say in which image is using.

In this case, `chart-dir` is set to `./cilium-${{ steps.vars.outputs.downgrade_version }}/install/kubernetes/cilium/`, which will be checked out from v1.14 branch by this task: https://github.com/cilium/cilium/blob/cb6349bbce577484c3635816a9747e44369a1c3e/.github/workflows/tests-ipsec-upgrade.yaml#L173-L178

We can confirm by looking at the successful CI task:  https://github.com/cilium/cilium/actions/runs/6119666881/job/16610076122, the task `Downgrade Cilium to v1.14 & Test (3)` has logs demonstrating it was downgraded to v1.14.1

So I assume it causes downgrade to v1.14 as expected?

michi-covalent: yeah `downgrade_version` is set to `v1.14`

michi-covalent (approved):
😇

brlbil (approved):

.github/workflows/tests-ipsec-upgrade.yaml
nebril: I missed that, thanks for explaining!

nebril (approved):

PR #28020, opened 2023-09-08
egressgw: manager: test: mark helpers with c.Helper()

jibi:
in this way when a test fails we don't get the line number of the helper but rather the one of the caller


jschwinger233 (approved):
:+1:

PR #28025, opened 2023-09-08
Respond with ICMP reply for traffic to services without backends

### LINKED ISSUE #10002: Failures in [sig-network] Network should set TCP CLOSE_WAIT timeout and Services should be rejected when no endpoints exist
## Support

We are adding e2e tests to the [Kops](https://github.com/kubernetes/kops) project for most Kubernetes network plugins, Cilium included. Seems that some of the tests are failing and wanted to get some feedback here about them.
https://testgrid.k8s.io/sig-cluster-lifecycle-kops#kops-aws-cni-cilium

The tests that fail are:
```
[sig-network] Network should set TCP CLOSE_WAIT timeout
[sig-network] Services should be rejected when no endpoints exist
```

Would like to know if there are any settings to make these pass or if there is any plan to fix them. Thanks!

dylandreimerink:
So far we have been dropping packets meant for services which do not have endpoints/backends. This causes clients to needlessly wait for replies and to retry sending traffic. This PR adds the ability to send back an ICMP or ICMPv6 reply with Destination unreachable (type 3) + Host/Address unreachable (code 1) whenever this happens. 

This differs slightly from the kube-proxy behavior which would send back and type 3/code 3 port unreachable, given its the whole IP not a single port the address unreachable code made more sense.

This behavior is controllable via a new `--service-no-backend-response` flag, which defaults to `reject` so we match expected behavior by default. It can also be set to `drop` to preserve the existing behavior in case that was desired.

This new behavior works for both North/South traffic entering a node and East/West traffic responding to a request from a pod within the cluster.

Fixes: #10002

```release-note
Respond with ICMP reply for traffic to services without backends
```


dylandreimerink:
Closing this PR in favor of https://github.com/cilium/cilium/pull/28157

PR #28030, opened 2023-09-08
StateDB review follow-ups

joamaki:
Follow-ups to the original statedb v2.0 PR (https://github.com/cilium/cilium/pull/27160).

This addresses the original review comments, expands tests to improve coverage (81.8% => 86.1%) and
fixes few metrics related issues and adds tests for gauge metrics.


pkg/statedb/types.go
```diff
@@ -56,6 +56,19 @@ type Table[Obj any] interface {
...
+	//
+	// WARNING: Do not create a WriteTxn against this table and then call this
+	// with a ReadTxn. This will lead to a dead lock as in order for DeleteTracker()
+	// to be usable by readers it needs to perform a write transaction on behalf
+	// of the caller.
+	//
+	// If a suitable WriteTxn is provided, then that is used to add the tracker.
+	// In that case the tracker will only become active when the transaction is
+	// committed.
+	DeleteTracker(txn ReadTxn, trackerName string) (*DeleteTracker[Obj], error)
```
joamaki: Or we just keep it as `WriteTxn` and then `DeleteTracker` is the only usable method that someone without `RWTable` can use to "modify the table", e.g. readers who have `Table[Obj]` can do `WriteTxn(table)` and call `DeleteTracker` with it. Would get rid of this deadlock potential.

I think it's a reasonable trade-off to still expose `WriteTxn` against even read-only tables for everyone as it allows this DeleteTracker() stuff and it allows building smart writers on top of `RWTable[Obj]` while still allowing cross-table transactionality. The downside I see is that then arbitrary code can "lock" the table even when they can't modify it otherwise. Is this practical engineering or am I being reckless?

joamaki:
@giorio94 could you check that I've addressed your review comments from the original PR?

pkg/statedb/db_test.go
```diff
@@ -91,517 +89,504 @@ func testWithDB(t testing.TB, withTags bool, test func(db *DB, table Table[testO
 	require.NoError(t, h.Start(context.TODO()))
 	t.Cleanup(func() {
 		assert.NoError(t, h.Stop(context.TODO()))
+		logging.SetLogLevel(logrus.InfoLevel)
 	})
-	test(db, table)
+	return db, table
 }
 
 func TestDB_LowerBound_ByRevision(t *testing.T) {
```
bimmlerd, line 99: Could some of these be https://pkg.go.dev/testing#T.Parallel? 
(Was reminded about this due to the Cleanup thing, https://brandur.org/fragments/go-prefer-t-cleanup-with-parallel-subtests)

bimmlerd (approved):
LGTM

pkg/statedb/db_test.go
joamaki: Yes they can. I'll play with it.

joamaki: Done. `0.346s` to `0.210s`.

dylandreimerink (approved):
LGTM

giorio94:
> @giorio94 could you check that I've addressed your review comments from the original PR?

Looks good to me. I wonder if this one could still be worth being addressed: https://github.com/cilium/cilium/pull/27160#discussion_r1299927295

giorio94 (approved):

joamaki:
> > @giorio94 could you check that I've addressed your review comments from the original PR?
> 
> Looks good to me. I wonder if this one could still be worth being addressed: [#27160 (comment)](https://github.com/cilium/cilium/pull/27160#discussion_r1299927295)

Sorry missed that one. I improved the method comments to clarify what was meant by thread unsafe: https://github.com/cilium/cilium/pull/28030/commits/bce76ad0307804e1649ca0236f7026ac4331cbd8. The thread safety was about the returned `WriteTxn` or `ReadTxn` object and not about the ReadTxn() or WriteTxn() method.

PR #28035, opened 2023-09-08
BGP CP: Calls String() Afi/Safi Methods instead of Duplicative Funcs

danehans:
Removes duplicative `ToAPIAfi()` and `ToAPISafi()` functions in favor of using the Afi/Safi `String()` method.

PR #28036, opened 2023-09-08
BGP CP: API Helper Functions Cleanup

danehans:
- Moves functions next to types.
- Renames functions to ParseXXX
- Adds godocs.

### DIFF
--- a/pkg/bgpv1/api/conversions.go
+++ b/pkg/bgpv1/api/conversions.go
@@ -16,85 +16,18 @@ import (
 	"github.com/cilium/cilium/pkg/bgpv1/types"
 )
 
-func ToAgentAfi(s string) types.Afi {
-	switch s {
-	case "ipv4":
-		return types.AfiIPv4
-	case "ipv6":
-		return types.AfiIPv6
-	case "l2vpn":
-		return types.AfiL2VPN
-	case "ls":
-		return types.AfiLS
-	case "opaque":
-		return types.AfiOpaque
-	default:
-		return types.AfiUnknown
-	}
-}
-
-func ToAgentSafi(s string) types.Safi {
-	switch s {
-	case "unicast":
-		return types.SafiUnicast
-	case "multicast":
-		return types.SafiMulticast
-	case "mpls_label":
-		return types.SafiMplsLabel
-	case "encapsulation":
-		return types.SafiEncapsulation
-	case "vpls":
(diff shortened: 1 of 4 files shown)


asauber (approved):

rastislavs:
Thanks! in general I like this cleanup (especially considering it together with https://github.com/cilium/cilium/pull/28035).
Just one idea. As we are expanding `pkg/bgpv1/types/state.go` here, I think the name of this file no longer matches with what it is. That might also be the reason why some of the conversion funcs were duplicated. Not sure what should be a better name though ATM :)

danehans:
@rastislavs thanks for the review and feedback. I documented my ideas for reorganizing the gobgp pkg in [this](https://docs.google.com/document/d/1VLuvhaz3K879efymmqiCGzGMCedeD0KqdLCS5wT3Dkc/edit?usp=sharing) Google doc. PTAL when you have a moment and provide feedback. I'm happy to use this PR to implement the larger reorganization based on the outcome of the Google doc.

rastislavs:
> @rastislavs thanks for the review and feedback. I documented my ideas for reorganizing the gobgp pkg in [this](https://docs.google.com/document/d/1VLuvhaz3K879efymmqiCGzGMCedeD0KqdLCS5wT3Dkc/edit?usp=sharing) Google doc. PTAL when you have a moment and provide feedback. I'm happy to use this PR to implement the larger reorganization based on the outcome of the Google doc.

The changes in the doc are reasonable, but as those are constrained in gobgp package, I think it makes sense to open a PR for them separately.

pkg/bgpv1/api/conversions.go
```diff
@@ -16,85 +16,76 @@ import (
...
-		return types.AfiIPv4
-	case "ipv6":
-		return types.AfiIPv6
-	case "l2vpn":
-		return types.AfiL2VPN
-	case "ls":
-		return types.AfiLS
-	case "opaque":
-		return types.AfiOpaque
+func ToAPIAfi(afi types.Afi) string {
```
rastislavs: Why this adds back the code removed in https://github.com/cilium/cilium/pull/28035?

danehans: Thanks for catching this. It must have slipped back in during a rebase.

rastislavs (approved):
LGTM

PR #28044, opened 2023-09-08
CI Images: Don't push floating tags from feature branches

michi-covalent:
Floating tags like "latest" or "v1.14" should only be pushed from main and stable branches.

### DIFF
--- a/.github/workflows/build-images-ci.yaml
+++ b/.github/workflows/build-images-ci.yaml
@@ -148,7 +148,7 @@ jobs:
 
       # main branch pushes
       - name: CI Build ${{ matrix.name }}
-        if: ${{ github.event_name != 'pull_request_target' }}
+        if: ${{ github.event_name != 'pull_request_target' && !startsWith(github.ref_name, 'ft/') }}
         uses: docker/build-push-action@0565240e2d4ab88bba5387d719585280857ece09 # v5.0.0
         id: docker_build_ci_main
         with:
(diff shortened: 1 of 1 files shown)


michi-covalent:
i guess it does what we want, but it doesn't feel very elegant. requesting review from @nbusseneau to see if he has any suggestions 🚀 🙏 ✨

.github/workflows/build-images-ci.yaml
```diff
@@ -290,7 +290,7 @@ jobs:
 
       # PR updates
       - name: CI Build ${{ matrix.name }}
-        if: ${{ github.event_name == 'pull_request_target' }}
+        if: ${{ github.event_name == 'pull_request_target' || startsWith(github.ref_name, 'ft/') }}
```
nbusseneau: This one will do nothing because `ft/` is only for `push` events, right?

nbusseneau (approved):
Yeah I don't know... Technically we should split these steps into separate jobs and then have the `if` at the job level, instead of copy-pasting around on all steps. But at the same time these workflows are a mess and doing such a refactor is not trivial...

nbusseneau (changes requested):
Wrong button. Too used to just approving your PRs.

.github/workflows/build-images-ci.yaml
michi-covalent: basically what i want to say is: do this step if it's a pull request or it's a push to a feature branch 💭

michi-covalent: so maybe something like this?
```
        if: ${{ github.event_name == 'pull_request_target' || (github.event_name == 'push' && startsWith(github.ref_name, 'ft/')) }}

michi-covalent: alright updated the pr. could you take another look 🙏

nbusseneau: But I don't get it, isn't the push use case covered by the other steps above? 🤔

michi-covalent: no, because i added `&& !startsWith(github.ref_name, 'ft/')` to the condition.

basically what i'm trying to prevent is pushing `latest` tag from feature branches. it should only be pushed from `main` branch.

nbusseneau: Aaah now I get it.

nbusseneau (approved):

PR #28045, opened 2023-09-08
docs: Remove "by Default" suffix in cilium-agent metrics header

learnitall:
Some documentation on metrics for cilium-agent is below a header which reads "Exported Metrics by Default". This commit removes the "by Default" part of the header. This is needed because the tables shown below the header already have a 'Default' column which describes if the metric is exported by default.

```release-note
docs: Remove "by Default" suffix in cilium-agent metrics header
```


zacharysarah (approved):
LGTM

nathanjsweet (approved):

nathanjsweet:
Because this is a doc change the entire test suite is not needed.

PR #28046, opened 2023-09-08
Docs: Updates for Deprecation of CNI network-plugin Flag

### LINKED ISSUE #26531: Docs: Remove Deprecated network-plugin Kubelet Flag
From the [docs](https://docs.cilium.io/en/stable/network/kubernetes/requirements/):
>[CNI](https://docs.cilium.io/en/stable/glossary/#term-CNI) - Container Network Interface is the plugin layer used by Kubernetes to delegate networking configuration. CNI must be enabled in your Kubernetes cluster in order to install Cilium. This is done by passing --network-plugin=cni to kubelet on all nodes.

The `--network-plugin` flag was [removed](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/) in k8s v1.24 and should be removed or updated accordinly in the docs.

danehans:
- Updates kubelet command-line parameters in K8s requirements section.
- Updates install steps for deprecated CLI flag in minikube install section.
- Since the minikube install docs state `minikube ≥ v1.28.0`, notes specific to older minikube versions were removed.

Fixes: #26531

### DIFF
--- a/Documentation/gettingstarted/k8s-install-default.rst
+++ b/Documentation/gettingstarted/k8s-install-default.rst
@@ -137,20 +137,13 @@ to create a Kubernetes cluster locally or using a managed Kubernetes service:
 
        .. code-block:: shell-session
 
-          minikube start --network-plugin=cni --cni=false
+          minikube start --cni=cilium
 
        .. note::
 
-          From minikube v1.12.1+, cilium networking plugin can be enabled directly with
-          ``--cni=cilium`` parameter in ``minikube start`` command. However, this may not
-          install the latest version of cilium.
-
-          MacOS M1 users using a Minikube version < v1.28.0 with ``--cni=false`` will also need to run
-          ``minikube ssh -- sudo mount bpffs -t bpf /sys/fs/bpf`` in order to mount the BPF filesystem
-          ``bpffs`` to ``/sys/fs/bpf``.
-
-          It might be necessary to add ``--host-dns-resolver=false`` if using the Virtualbox provider,
-          otherwise DNS resolution may not work after Cilium installation.
+          - This may not install the latest version of cilium.
+          - It might be necessary to add ``--host-dns-resolver=false`` if using the Virtualbox provider,
+            otherwise DNS resolution may not work after Cilium installation.
 
     .. group-tab:: Rancher Desktop
 
--- a/Documentation/network/kubernetes/requirements.rst
+++ b/Documentation/network/kubernetes/requirements.rst
@@ -38,10 +38,10 @@ Enable CNI in Kubernetes
 ========================
 
 :term:`CNI` - Container Network Interface is the plugin layer used by Kubernetes to
-delegate networking configuration. CNI must be enabled in your Kubernetes
-cluster in order to install Cilium. This is done by passing
-``--network-plugin=cni`` to kubelet on all nodes. For more information, see
-the `Kubernetes CNI network-plugins documentation <https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/>`_.
+delegate networking configuration and is enabled by default in Kubernetes 1.24 and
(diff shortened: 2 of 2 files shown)


Documentation/gettingstarted/k8s-install-default.rst
```diff
@@ -137,20 +137,13 @@ to create a Kubernetes cluster locally or using a managed Kubernetes service:
 
        .. code-block:: shell-session
 
-          minikube start --network-plugin=cni --cni=false
+          minikube start --cni=cilium
 
        .. note::
 
-          From minikube v1.12.1+, cilium networking plugin can be enabled directly with
```
joamaki, line 144: Would it make sense to flip this so that we'd say in the note that for v1.12.0 and earlier the option is `--network-plugin-cni --cni=false` and Cilium is installed manually? Or is that already so ancient that it's not worth mentioning?

danehans: I can make this change if needed. IMHO it comes down to whether docs should be provided for an unsupported minikube version.

danehans: @joamaki do you have an opinion on my ^ comment?

learnitall (approved):
This looks great, thank you!

As a follow-up to this, it would be great if we could add information on how to install a specific version of Cilium into minikube using the manifest option for the `--cni` parameter. I think the workflow would involve rendering a manifest using `helm template` and then passing the file path of the manifest to `--cni`. @danehans if you have time and would like to tackle this as part of your PR then please feel free, otherwise I can just make a follow-up issue.

Documentation/gettingstarted/k8s-install-default.rst
learnitall: I think it makes sense to leave the old version out. Eventually, we'd have to remove the old instructions anyway. If users need to see the old instructions, they can go back in time to view the old installation instructions using the version selector in the docs.

learnitall: I think it makes sense to leave the old version out. Eventually, we'd have to remove the old instructions anyway. If users need to see the old instructions, they can go back in time to view the old installation instructions using the version selector in the docs.

danehans: ^ is exactly what I was thinking.

danehans:
> if you have time and would like to tackle this as part of your PR then please feel free, otherwise I can just make a follow-up issue.

@learnitall thanks for the review. I believe ^ is a separate issue from the one this PR is fixing. Feel free to create an issue and assign me.

learnitall:
Sounds good @danehans, will do!

Documentation/gettingstarted/k8s-install-default.rst
julianwiedmann: holding off the merge until this discussion is resolved, @joamaki please ping / re-label when you're good :)

danehans: @joamaki when you have a moment, do you mind commenting to unblock this PR?

danehans: @julianwiedmann do you have any suggestions on how to move this PR forward?

learnitall:
@joamaki, when you have a moment, do you mind taking another look so we can resolve the feedback you brought up?

danehans:
@learnitall @julianwiedmann is it possible to [dismiss the PR review](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review) from @joamaki to get this unblocked?

Documentation/gettingstarted/k8s-install-default.rst
joamaki: Makes sense to me!

joamaki (approved):

PR #28050, opened 2023-09-09
cni: use case-insensitive names for CNI chaining mode

asauber:
Allow case-insensitive names for CNI chaining mode

Upon startup, Cilium must write CNI config files for the configured chaining mode. For this purpose, it uses a package-scoped map with lowercase string keys to refer to each mode.

We have seen configurations in the wild which do not use all-lowercase names for these chaining modes. For example, in the reccomended configuration for OpenShift conformance testing we have:

    cni.chainingMode=portMap

This causes Cilium to fail on startup with:

    failed to render CNI configuration file: invalid CNI chaining mode:
    portMap

```release-note
Allow case-insensitive name for CNI chaining mode
```


ldelossa (approved):
Not a CNI expert, but code change and explanation LGTM :+1:

PR #28063, opened 2023-09-11
bpf: Delete obsolete do_netdev_encrypt_pools()

jschwinger233:
This PR deletes do_netdev_encrypt_pools() because:

1. It's no longer needed in 1.15. It was used to NAT packets IPsec-encrypted by old XFRM OUT state, which is no longer the case in 1.15.
2. It causes bug for IPv6, as this function only takes care of IPv4 and sometimes IPv6 packets get NAT-ed incorrectly.


brb (approved):
:pizza:

PR #28071, opened 2023-09-11
daemon, fqdn: Remove log "DNS request no matching endpoint" when endpoint is nil

doniacld:
This commit intends to remove the log when endpoint is nil. `notifyOnDNSMsg` is called with endpoint nil only before restoration starts meaning there is no endpoint created at that point, see `ServeDNS`. Callers are currently logging contextualized errors when something wrong happens. Client-side having already the error message does not need this additional log. Let’s just remove it.

```release-note
daemon, fqdn: Remove log "DNS request no matching endpoint" when endpoint is nil
```

### DIFF
--- a/daemon/cmd/fqdn.go
+++ b/daemon/cmd/fqdn.go
@@ -449,9 +449,12 @@ func (d *Daemon) lookupIPsBySecID(nid identity.NumericIdentity) []string {
 //   - Report a monitor error event and proxy metrics when the proxy sees an
 //     error, and when it can't process something in this function
 //   - Report the verdict in a monitor event and emit proxy metrics
-//   - Insert the DNS data into the cache when msg is a DNS response and we
-//     can lookup the endpoint related to it
+//   - Insert the DNS data into the cache when msg is a DNS response, and we
+//     can lookup the endpoint related to it.
 //
+// It may return dnsproxy.ErrDNSRequestNoEndpoint{} error if the endpoint is nil.
+// Note that the caller should log beforehand the contextualized error.
+
 // epIPPort and serverAddr should match the original request, where epAddr is
 // the source for egress (the only case current).
 // serverID is the destination server security identity at the time of the DNS event.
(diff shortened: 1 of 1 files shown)


daemon/cmd/fqdn.go
```diff
@@ -510,9 +510,15 @@ func (d *Daemon) notifyOnDNSMsg(lookupTime time.Time, ep *endpoint.Endpoint, epI
...
 		err := dnsproxy.ErrDNSRequestNoEndpoint{}
+
+		// Logging on debug mode and not an error. ep can currently be nil only in cases
+		// where the restoration did not happen, so there is no endpoint.
+		// Caller is currently logging the contextualised error. We keep this log for
+		// debug purposes only, but it has no value on client side.
 		log.WithFields(logrus.Fields{
 			logfields.L3n4Addr: epIPPort,
-		}).WithError(err).Error("cannot find matching endpoint")
+		}).Debugf("cannot find matching endpoint: %s", err)
```
gandro: Let's use `WithError` here

daemon/cmd/fqdn.go
```diff
@@ -510,9 +510,15 @@ func (d *Daemon) notifyOnDNSMsg(lookupTime time.Time, ep *endpoint.Endpoint, epI
 		// asserted via ep != nil here and msg.Response && msg.Rcode ==
 		// dns.RcodeSuccess below).
 		err := dnsproxy.ErrDNSRequestNoEndpoint{}
+
+		// Logging on debug mode and not an error. ep can currently be nil only in cases
```
gandro: `ep can currently be nil only in cases where the restoration did not happen, so there is no endpoint.`

This to me does not sound entirely correct. EP can be nil if
a) The caller passes in `ep=nil` due various errors that happened before the endpoint was determined (concurrency limit, invalid IP address in DNS packet etc)
b) The EP can be nil because it's legitimately from an endpoint that does not exist, either because the endpoint was just deleted, or because we received a bogus (e.g. spoofed) packet, or it could not be restored

In both cases the caller already logs the reason, so the rest of the comment is correct.

christarazi: I'm in favor of removing the logging in `notifyOnDNSMsg()` and relying on the caller in the relevant cases to log the reason for the endpoint being nil. There should be a code comment describing this so readers understand.

doniacld: Added a comment on `notifyOnDNSMsg()` function and removed the log.

gandro (approved):

pippolo84 (approved):
Thanks :100:

daemon/cmd/fqdn.go
```diff
@@ -449,9 +449,12 @@ func (d *Daemon) lookupIPsBySecID(nid identity.NumericIdentity) []string {
...
 //     error, and when it can't process something in this function
 //   - Report the verdict in a monitor event and emit proxy metrics
-//   - Insert the DNS data into the cache when msg is a DNS response and we
-//     can lookup the endpoint related to it
+//   - Insert the DNS data into the cache when msg is a DNS response, and we
+//     can lookup the endpoint related to it.
 //
+// It may return dnsproxy.ErrDNSRequestNoEndpoint{} error if the endpoint is nil.
+// Note that the caller should log beforehand the contextualized error.
+
```
christarazi, line 457: Nit: I think you need a `//` here to make sure the godoc is not separated.

PR #28090, opened 2023-09-12
bpf,ci: Remove coverage collection from BPF tests

dylandreimerink:
The test coverage collection currently gives more headaches than it is worth. Disabling it for now, until coverbee is improved.

```release-note
Remove coverage collection from BPF tests
```


lmb (approved):
Thanks!

brlbil (approved):

PR #28091, opened 2023-09-12
egressgateway: switch to Resource[T]

lmb:
egressgw: use Resource[CiliumNode]

egressgw: use Resource[CiliumEndpoint]

    Replace the custom watcher with Resource[CiliumEndpoint]. This also allows
    getting rid of the custom retry queue, since Resource will by default retry
    an event if Done is called with an error.


pkg/egressgateway/manager.go
```diff
@@ -559,32 +498,20 @@ func (manager *Manager) deleteEndpoint(id types.NamespacedName) {
...
-	delete(manager.pendingEndpointEvents, id)
-	manager.pendingEndpointEventsLock.Unlock()
-
-	manager.endpointEventsQueue.Add(id)
+	if event.Kind == resource.Upsert {
+		event.Done(manager.addEndpoint(endpoint))
+	} else {
+		manager.deleteEndpoint(id)
+		event.Done(nil)
+	}
```
jschwinger233, line 514: Do we need to add `manager.Lock(); defer manager.Unlock()` to protect these map insert/delete operations?

jschwinger233 (commented):
Expect for one potential race, the other parts LGTM.

pkg/egressgateway/manager.go
lmb: I just had a look, manager.addEndpoint and deleteEndpoint both already take the lock.

jschwinger233 (approved):

joamaki (approved):

PR #28122, opened 2023-09-12
ci: Run BPF lints on workflow definition changes

qmonnet:
In the past, we have merged automated updates to the coccicheck image version ([for example][0]) without realising they may break the workflow, because the workflow wouldn't run on the PR, which contained no BPF-related changes.

Let's make the workflow run when its definition file is updated, to make sure we can catch similar issues in the future.

[0]: https://github.com/cilium/cilium/pull/27947


qmonnet:
Hmm the file change was correctly detected, but the jobs were skipped anyway :thinking:

aanm (approved):

nebril (approved):

PR #28131, opened 2023-09-12
Refactor CiliumExecContext() Retry Logic

carnerito:
Previously, the CiliumExecContext() function would retry the cilium command up to 5 times (limitTimes) without verifying the success of previous executions. This pull request introduces a logic enhancement that validates the return code of each executed command and exits upon success.

Additionally, this change optimizes test execution time by reducing unnecessary retries. With a 200ms delay between retries, the overall test execution time is expected to improve by at least 200ms multiplied by the number of cilium commands and the number of cilium pods involved.

Local Ginkgo tests, specifically those focused on 'K8sDatapathBGPTests' 'K8sDatapathCustomCalls' and 'K8sDatapathLRPTests' have shown significant improvements:

Before:
`Ran 2 of 106 Specs in 233.348 seconds`

After:
`Ran 2 of 106 Specs in 168.563 seconds`

### DIFF
--- a/test/helpers/kubectl.go
+++ b/test/helpers/kubectl.go
@@ -2911,9 +2911,15 @@ func (kub *Kubectl) CiliumExecContext(ctx context.Context, pod string, cmd strin
 	for i := 0; i < limitTimes; i++ {
 		res = execute()
 		switch res.GetExitCode() {
+		case 0:
+			// Command succeeded. Return the result.
+			return res
 		case -1, 126:
-			// Retry.
+			// The preceding comments indicate that these return codes may occur frequently.
+			// To prevent excessive log entries in the default case, we catch these errors here
+			// and retry the command without generating additional log entries.
 		default:
+			// Command failed. Log failure and retry.
 			kub.Logger().Warningf("command terminated with exit code %d on try %d", res.GetExitCode(), i)
 		}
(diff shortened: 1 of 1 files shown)


christarazi (approved):
Nice find! Did you mean to flip the `before` and `after` around? :sweat_smile:

test/helpers/kubectl.go
```diff
@@ -2911,9 +2911,11 @@ func (kub *Kubectl) CiliumExecContext(ctx context.Context, pod string, cmd strin
 	for i := 0; i < limitTimes; i++ {
 		res = execute()
 		switch res.GetExitCode() {
-		case -1, 126:
-			// Retry.
```
tklauser: Previously we didn't log anything for these error codes. With this change we now log a warning in the `default` clause below. I'm not sure how frequent these errors are, but given the comment above they might be? Maybe we should keep the special case for these error codes to avoid spamming the logs two much.

tklauser (commented):
Nice find. One question re. logging on error codes -1 and 126 inline.

test/helpers/kubectl.go
carnerito: @tklauser I've returned the case that catches '-1' and '126' return codes and added explanation in the comments why this exists.

tklauser (approved):
Thanks!

PR #28133, opened 2023-09-13
bpf: lxc: fix source sec identity in hairpin trace notification

julianwiedmann:
CB_SRC_LABEL is cleared just a few lines above. So to get the source's actual security identity, we need to use our local variable.

Fixes: e2829a061a53 ("bpf: lxc: support Pod->Service->Pod hairpinning with endpoint routes")

```release-note
Fix the trace notification for hairpinned reply traffic, to indicate the correct security identity for the client.
```


qmonnet (approved):
Thanks!

PR #28134, opened 2023-09-13
bpf: overlay: clean up CB_SRC_LABEL handling in inter-cluster-SNAT path

julianwiedmann:
Align the logic with our standard pattern, where the entry-level function loads from CB_SRC_LABEL (+ clears it), and then passes the security identity to the lower-level functions.

Thus making it clear that handle_inter_cluster_revsnat() doesn't mutate the identity in any way.

No functional change intended.


YutaroHayakawa (approved):
LGTM 👍 Thanks!

PR #28140, opened 2023-09-13
StateDB: split write methods from Table into RWTable

joamaki:
- Splits Table[Obj] into Table[Obj] (reader) and RWTable[Obj] (reader-writer) and adds
`NewProtectedTableCell`. This allows modules to define protected tables where outside modules can only read from the table, or write to the table only in restricted/validated ways.
- Moves RWTable[*Device] and RWTable[*Route] under the scope of the devices-controller module to protect the tables. 

Open questions:
- Is `RWTable` a good name? What about `TableWriter`? Should `Table[Obj]` be `RTable[Obj]` or `TableReader[Obj]`? WDYT?
- WriteTxn still takes a TableMeta (from `Table[Obj]`). Reasoning is that we want to be able to construct a `WriteTxn` against multiple tables (for that cross-table atomic transaction sweetness) even when `RWTable` isn't directly available (but indirect smart writers are). An alternative I see to this is some sort of `WriteTxnBuilder` pattern to allow composing together the `WriteTxn` in a controlled way, e.g. a smart writer would take `WriteTxnBuilder` and then return `(WriterHandle, WriteTxnBuilder)`. This way one explicitly asks for access to the writer methods, but I'm not sure if this boilerplate brings a lot of value. The downside of allowing `WriteTxn` with `Table[Obj]` is that arbitrary code can lock the table without intention or ability to write to it, but I would expect this to be easily catched by code review.

### DIFF
--- a/pkg/datapath/tables/device.go
+++ b/pkg/datapath/tables/device.go
@@ -45,7 +45,7 @@ var (
 		},
 	}
 
-	DeviceTableCell = statedb.NewTableCell[*Device](
+	DeviceTableCell = statedb.NewProtectedTableCell[*Device](
 		"devices",
 		DeviceIDIndex,
 		DeviceNameIndex,
--- a/pkg/statedb/table.go
+++ b/pkg/statedb/table.go
@@ -17,7 +17,7 @@ func NewTable[Obj any](
 	tableName TableName,
 	primaryIndexer Indexer[Obj],
 	secondaryIndexers ...Indexer[Obj],
-) (Table[Obj], error) {
+) (RWTable[Obj], error) {
 	toAnyIndexer := func(idx Indexer[Obj]) anyIndexer {
 		return anyIndexer{
 			name: idx.indexName(),
@@ -212,3 +212,4 @@ func (t *genTable[Obj]) sortableMutex() lock.SortableMutex {
 }
 
 var _ Table[bool] = &genTable[bool]{}
+var _ RWTable[bool] = &genTable[bool]{}
--- a/pkg/statedb/table.go
+++ b/pkg/statedb/table.go
@@ -163,10 +163,18 @@ func (t *genTable[Obj]) Get(txn ReadTxn, q Query[Obj]) (Iterator[Obj], <-chan st
 	return &iterator[Obj]{iter}, watchCh
 }
 
-// Insert implements Table
 func (t *genTable[Obj]) Insert(txn WriteTxn, obj Obj) (oldObj Obj, hadOld bool, err error) {
 	var data any
-	data, hadOld, err = txn.getTxn().Insert(t, obj)
+	data, hadOld, err = txn.getTxn().Insert(t, Revision(0), obj)
+	if err == nil && hadOld {
+		oldObj = data.(Obj)
+	}
+	return
+}
+
+func (t *genTable[Obj]) CompareAndSwap(txn WriteTxn, rev Revision, obj Obj) (oldObj Obj, hadOld bool, err error) {
+	var data any
+	data, hadOld, err = txn.getTxn().Insert(t, rev, obj)
 	if err == nil && hadOld {
 		oldObj = data.(Obj)
 	}
@@ -175,7 +183,16 @@ func (t *genTable[Obj]) Insert(txn WriteTxn, obj Obj) (oldObj Obj, hadOld bool, 
 
 func (t *genTable[Obj]) Delete(txn WriteTxn, obj Obj) (oldObj Obj, hadOld bool, err error) {
 	var data any
-	data, hadOld, err = txn.getTxn().Delete(t, obj)
+	data, hadOld, err = txn.getTxn().Delete(t, Revision(0), obj)
+	if err == nil && hadOld {
+		oldObj = data.(Obj)
+	}
+	return
+}
+
+func (t *genTable[Obj]) CompareAndDelete(txn WriteTxn, rev Revision, obj Obj) (oldObj Obj, hadOld bool, err error) {
+	var data any
+	data, hadOld, err = txn.getTxn().Delete(t, rev, obj)
 	if err == nil && hadOld {
 		oldObj = data.(Obj)
 	}
@@ -186,7 +203,7 @@ func (t *genTable[Obj]) DeleteAll(txn WriteTxn) error {
 	iter, _ := t.All(txn)
 	itxn := txn.getTxn()
 	for obj, _, ok := iter.Next(); ok; obj, _, ok = iter.Next() {
-		_, _, err := itxn.Delete(t, obj)
+		_, _, err := itxn.Delete(t, Revision(0), obj)
 		if err != nil {
 			return err
 		}
--- a/pkg/statedb/txn.go
+++ b/pkg/statedb/txn.go
@@ -123,25 +123,20 @@ func (txn *txn) indexWriteTxn(name TableName, index IndexName) *iradix.Txn[objec
 	return indexTreeTxn
 }
 
-func (txn *txn) newRevision(tableName TableName) (Revision, error) {
-	table, ok := txn.modifiedTables[tableName]
-	if !ok {
-		return 0, tableError(tableName, ErrTableNotLockedForWriting)
-	}
-	table.revision++
-	return table.revision, nil
-}
-
-func (txn *txn) Insert(meta TableMeta, data any) (any, bool, error) {
+func (txn *txn) Insert(meta TableMeta, guardRevision Revision, data any) (any, bool, error) {
 	if txn.rootReadTxn == nil {
 		return nil, false, ErrTransactionClosed
 	}
 
+	// Look up table and allocate a new revision.
 	tableName := meta.Name()
-	revision, err := txn.newRevision(tableName)
-	if err != nil {
-		return nil, false, err
+	table, ok := txn.modifiedTables[tableName]
+	if !ok {
+		return nil, false, tableError(tableName, ErrTableNotLockedForWriting)
 	}
+	oldRevision := table.revision
+	table.revision++
+	revision := table.revision
 
 	obj := object{
 		revision: revision,
@@ -152,6 +147,25 @@ func (txn *txn) Insert(meta TableMeta, data any) (any, bool, error) {
 	idKey := meta.primaryIndexer().fromObject(obj).First()
 	idIndexTree := txn.indexWriteTxn(tableName, meta.primaryIndexer().name)
 	oldObj, oldExists := idIndexTree.Insert(idKey, obj)
+
+	// For CompareAndSwap() validate against the given guard revision
+	if guardRevision > 0 {
+		if !oldExists {
+			// CompareAndSwap requires the object to exist. Revert
+			// the insert.
+			idIndexTree.Delete(idKey)
+			table.revision = oldRevision
+			return nil, false, ErrObjectNotFound
+		}
+		if oldObj.revision != guardRevision {
+			// Revert the change. We're assuming here that it's rarer for CompareAndSwap() to
+			// fail and thus we're optimizing to have only one lookup in the common case
+			// (versus doing a Get() and then Insert()).
+			idIndexTree.Insert(idKey, oldObj)
+			table.revision = oldRevision
+			return oldObj, true, ErrRevisionNotEqual
+		}
+	}
 
 	// Update revision index
 	revIndexTree := txn.indexWriteTxn(tableName, RevisionIndex)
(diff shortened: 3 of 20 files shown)


pkg/datapath/tables/device.go
```diff
@@ -45,7 +46,7 @@ var (
 		},
 	}
 
-	DeviceTableCell = statedb.NewTableCell[*Device](
+	DeviceTableCell cell.Cell = statedb.NewProtectedTableCell[*Device](
```
giorio94:
```suggestion
	DeviceTableCell = statedb.NewProtectedTableCell[*Device](
```

Nit.

pkg/statedb/txn.go
```diff
@@ -153,6 +148,22 @@ func (txn *txn) Insert(meta TableMeta, data any) (any, bool, error) {
...
 	oldObj, oldExists := idIndexTree.Insert(idKey, obj)
 
+	// For UpdateIfEqual() validate against guard revision.
+	if guardRevision > 0 {
+		if !oldExists {
+			// UpdateIfEqual requires the object to exist.
+			return nil, false, ErrObjectNotFound
+		}
+		if oldObj.revision != guardRevision {
+			// Revert the change. We're assuming here that it's rarer for UpdateIfOlder() to
```
giorio94:
```suggestion
			// Revert the change. We're assuming here that it's rarer for UpdateIfExists() to
```

Nit. I guess.

pkg/statedb/txn.go
```diff
@@ -153,6 +148,22 @@ func (txn *txn) Insert(meta TableMeta, data any) (any, bool, error) {
 	idIndexTree := txn.indexWriteTxn(tableName, meta.primaryIndexer().name)
 	oldObj, oldExists := idIndexTree.Insert(idKey, obj)
 
+	// For UpdateIfEqual() validate against guard revision.
+	if guardRevision > 0 {
+		if !oldExists {
+			// UpdateIfEqual requires the object to exist.
+			return nil, false, ErrObjectNotFound
```
giorio94: Should the change be reverted also in this case, similarly to what is done below?

pkg/statedb/types.go
```diff
@@ -79,18 +79,33 @@ type RWTable[Obj any] interface {
...
 
+	// UpdateIfEqual updates the object if it exists and has a revision
+	// equal to the given revision.
+	//
+	// Possible errors:
+	// - ErrRevisionNotEqual: the object has mismatching revision
+	// - ErrObjectNotFound: object not found from the table
+	// - ErrTableNotLockedForWriting: table was not locked for writing
+	// - ErrTransactionClosed: the write transaction already committed or aborted
+	UpdateIfEqual(WriteTxn, Revision, Obj) (oldObj Obj, hadOld bool, err error)
```
giorio94: I wonder if something like `UpdateIfRevision` could be clearer as name for this API (and similarly for the delete case).

giorio94 (commented):
The overall logic looks good to me, just a couple of minor comments inline.

> Is RWTable a good name? What about TableWriter? Should Table[Obj] be RTable[Obj] or TableReader[Obj]? WDYT?

Personally, I like `TableReader` and `TableReadWriter`, but that's fairly verbose. `RWTable` and `RTable` also sound good to me (and seem more symmetric than just plain `Table`). The main issue with just `TableWriter` is that it actually also provides the read functionalities.

> WriteTxn still takes a TableMeta (from Table[Obj]). Reasoning is that we want to be able to construct a WriteTxn against multiple tables (for that cross-table atomic transaction sweetness) even when RWTable isn't directly available (but indirect smart writers are). An alternative I see to this is some sort of WriteTxnBuilder pattern to allow composing together the WriteTxn in a controlled way, e.g. a smart writer would take WriteTxnBuilder and then return (WriterHandle, WriteTxnBuilder). This way one explicitly asks for access to the writer methods, but I'm not sure if this boilerplate brings a lot of value. The downside of allowing WriteTxn with Table[Obj] is that arbitrary code can lock the table without intention or ability to write to it, but I would expect this to be easily catched by code review.

I'm personally fine with keeping the current structure for the moment, and perhaps re-evaluate if more of such usages start to pop up. As you mentioned it seems pretty easy to catch incorrect creations of write transactions on read-only tables.

pkg/statedb/table.go
```diff
@@ -194,13 +194,41 @@ func (t *genTable[Obj]) DeleteAll(txn WriteTxn) error {
...
+	// To allow DeleteTracker() to be used by readers, we need to deal with these cases:
+	//
+	// 1. txn is just a ReadTxn, no tables locked => new WriteTxn
+	// 2. txn is a WriteTxn, but it's not locking the target table => new WriteTxn
+	// 3. txn is a WriteTxn and it's locking the target table => reuse txn
+	//
+	// There's one unfortunate case still, where the given 'txn' is a ReadTxn, but the caller
+	// has created a WriteTxn that is still open and targets this table. This will lead to a
+	// deadlock and there isn't much we can do about that. DeleteTracker() in Table[Obj] documents
+	// this caviat.
```
bimmlerd: nit, spelling: caveat (also in commit msg)

pkg/statedb/types.go
```diff
@@ -56,6 +56,19 @@ type Table[Obj any] interface {
...
+	//
+	// WARNING: Do not create a WriteTxn against this table and then call this
+	// with a ReadTxn. This will lead to a dead lock as in order for DeleteTracker()
+	// to be usable by readers it needs to perform a write transaction on behalf
+	// of the caller.
+	//
+	// If a suitable WriteTxn is provided, then that is used to add the tracker.
+	// In that case the tracker will only become active when the transaction is
+	// committed.
+	DeleteTracker(txn ReadTxn, trackerName string) (*DeleteTracker[Obj], error)
```
bimmlerd: I think this is problematic.

Pedantically, saying this 'only' takes a `ReadTxn` is a lie, hence the potential for deadlock. AFAIU, the reason we would _want_ this nonetheless is the following situation:

Some module, for example the device controller, is the 'owner' of the devices table. Ideally, it's the only one being able to modify the table, all others should have read-only access. Thanks to this PR, we can now express this in hive, and depend on just the `Table[*Device]` in a consuming cell. If the consumer wants to use this `DeleteTracker`, though, and we disallow using a `ReadTxn` for the tracker, it's not possible.

But the primary point of having split read and write table access is to be able to reason more easily over how things can interact. If these types are 'lying' to us, this benefit goes away. An example of how this can go wrong:

```
Cell A has RWTable[t], Cell B has just Table[t]

Cell A opens WriteTxn(t)
Cell A invokes some API of Cell B (while holding the wtxn)
Cell B, in response, opens a ReadTxn on Table[t], and creates a DeleteTracker.
```

If I understand the issue correctly, this will deadlock. When attempting to understand what's going on, I'd not suspect implication of B, _since it cannot lock table t_ (as it holds a read-only reference).

I'm not sure how to resolve this. One idea I had was that maybe the delete tracker could work on its own table, but I'm not sure how feasible that is. Or maybe it can store its state outside of the table?

pkg/statedb/types.go
```diff
@@ -56,6 +56,19 @@ type Table[Obj any] interface {
 	// when anything in the table changes as more fine-grained notifications
 	// are not possible with a lower bound search.
 	LowerBound(ReadTxn, Query[Obj]) (iter Iterator[Obj], watch <-chan struct{})
+
+	// DeleteTracker creates a new delete tracker for the table
+	// starting from the given revision.
```
bimmlerd: is there a given revision?

pkg/statedb/table.go
```diff
@@ -163,10 +163,18 @@ func (t *genTable[Obj]) Get(txn ReadTxn, q Query[Obj]) (Iterator[Obj], <-chan st
...
 	var data any
-	data, hadOld, err = txn.getTxn().Insert(t, obj)
+	data, hadOld, err = txn.getTxn().Insert(t, Revision(0), obj)
+	if err == nil && hadOld {
+		oldObj = data.(Obj)
+	}
+	return
+}
+
+func (t *genTable[Obj]) UpdateIfEqual(txn WriteTxn, rev Revision, obj Obj) (oldObj Obj, hadOld bool, err error) {
```
bimmlerd: Is this `CompareAndSwap`? If so, can we call it the more traditional name?

bimmlerd (changes requested):
I'm afraid of the footgun in the DeleteTracker API.

pkg/statedb/types.go
joamaki: Agreed. I'll do the more obvious thing and keep it taking the `WriteTxn`, but have this method in `Table`. It'll essentially be the only "write" that can be done without `RWTable`.

pkg/statedb/table.go
joamaki: I contemplated that initially, but for some reason didn't go for it. Now `CompareAndSwap` and `CompareAndDelete` seem good, so I'll go with those.

pkg/statedb/txn.go
joamaki: Oopsie. Thanks a lot for catching this. Will expand the tests as well...

pkg/statedb/types.go
joamaki: Changed to `CompareAndSwap` and `CompareAndDelete`.

pkg/statedb/types.go
joamaki: Leftover from when it actually did just take a revision. Reworded.

pkg/statedb/table.go
joamaki: This is gone now.

giorio94 (approved):

pkg/statedb/table.go
joamaki: Or should we be really explicit and do `CompareRevisionAndSwap`?

pkg/statedb/types.go
```diff
@@ -93,7 +108,18 @@ type RWTable[Obj any] interface {
 	// iteration of updates and deletions (see (*DeleteTracker[Obj]).Process).
 	Delete(WriteTxn, Obj) (oldObj Obj, hadOld bool, err error)
 
-	// DeleteAll removes all objects in the table.
+	// CompareAndSwap compares the existing object's revision against the
+	// given revision and if equal it deletes the object.
+	//
+	// Possible errors:
+	// - ErrRevisionNotEqual: the object has mismatching revision
+	// - ErrObjectNotFound: object not found from the table
```
bimmlerd: Does it make sense for `ErrObjectNotFound` to be an error for `CompareAndDelete`? I get the use-case for `ErrRevisionNotEqual`, but in symmetry with the normal `Delete`, I'd have expected that `CompareAndDelete` suceeds when the object is gone. 

EDIT: I realise now that `Delete` also doesn't do this. Shouldn't Delete be an idempotent operation?

bimmlerd (commented):
Yep I like this much better. Confused about the Delete semantics currently, other than that looks good!

pkg/statedb/types.go
joamaki: Ah but I wanted it to be symmetric with `CompareAndSwap`, e.g. the `Compare` part of it failed with `ErrObjectNotFound`. But I can see it working the other way around with being symmetric to `Delete`.

joamaki: After bit of a think, let's make it symmetric with `Delete` and drop the error. Most use-cases likely are not interested in the `ErrObjectNotFound` in the first place and they can just check if an object was returned or not.

pkg/statedb/types.go
```diff
@@ -56,19 +57,49 @@ type Table[Obj any] interface {
...
+	CompareAndSwap(WriteTxn, Revision, Obj) (oldObj Obj, hadOld bool, err error)
+
 	// Delete an object from the table. Returns the object that was
-	// deleted if there was one. Error may be returned if the table
-	// is not locked for writing or if the write transaction has already
-	// been committed or aborted.
+	// deleted if there was one.
+	//
+	// Possible errors:
+	// - ErrObjectNotFound: object not found from the table
```
bimmlerd: This comment is wrong, `Delete` can't return this error.

pkg/statedb/types.go
```diff
@@ -77,13 +108,19 @@ type Table[Obj any] interface {
...
+	// found 'hadOld' will be false and 'err' nil.
+	//
+	// Possible errors:
+	// - ErrRevisionNotEqual: the object has mismatching revision
+	// - ErrTableNotLockedForWriting: table was not locked for writing
+	// - ErrTransactionClosed: the write transaction already committed or aborted
+	CompareAndDelete(WriteTxn, Revision, Obj) (oldObj Obj, hadOld bool, err error)
+
+	// DeleteAll removes all objects in the table. If the table is tracked
+	// for deletions, the objects are moved into the graveyard index.
```
bimmlerd: Possible errors probably the same as for usual `Delete` right?

bimmlerd (commented):
two nits, after that I'm happy!

pkg/statedb/types.go
joamaki: Fixed.

pkg/statedb/types.go
joamaki: Fixed.

bimmlerd (approved):

dylandreimerink (approved):
LGTM!

PR #28141, opened 2023-09-13
docs: Update IPsec key rotation command

jschwinger233:
The previous command relies on awk(1) which could behave differently on Linux and macOS. This patch uses `kubectl -o go-template` to remove dependency on awk(1), makes sure key rotation operation works properly on both platform.

### DIFF
--- a/Documentation/security/network/encryption-ipsec.rst
+++ b/Documentation/security/network/encryption-ipsec.rst
@@ -166,7 +166,7 @@ To replace cilium-ipsec-keys secret with a new key:
 
 .. code-block:: shell-session
 
-    KEYID=$(kubectl get secret -n kube-system cilium-ipsec-keys -o yaml | awk '/^\s*keys:/ {print $2}' | base64 -d | awk '{print $1}')
+    KEYID=$(kubectl get secret -n kube-system cilium-ipsec-keys -o go-template --template={{.data.keys}} | base64 -d | cut -c 1)
     if [[ $KEYID -ge 15 ]]; then KEYID=0; fi
     data=$(echo "{\"stringData\":{\"keys\":\"$((($KEYID+1))) "rfc4106\(gcm\(aes\)\)" $(echo $(dd if=/dev/urandom count=20 bs=1 2> /dev/null| xxd -p -c 64)) 128\"}}")
     kubectl patch secret -n kube-system cilium-ipsec-keys -p="${data}" -v=1


brb (approved):
Thanks! Another option is to use more canonical(?) `-o jsonpath=`.

Documentation/security/network/encryption-ipsec.rst
```diff
@@ -166,7 +166,7 @@ To replace cilium-ipsec-keys secret with a new key:
 
 .. code-block:: shell-session
 
-    KEYID=$(kubectl get secret -n kube-system cilium-ipsec-keys -o yaml | awk '/^\s*keys:/ {print $2}' | base64 -d | awk '{print $1}')
+    KEYID=$(kubectl get secret -n kube-system cilium-ipsec-keys -o go-template --template={{.data.keys}} | base64 -d | cut -c 1)
     if [[ $KEYID -ge 15 ]]; then KEYID=0; fi
     data=$(echo "{\"stringData\":{\"keys\":\"$((($KEYID+1))) "rfc4106\(gcm\(aes\)\)" $(echo $(dd if=/dev/urandom count=20 bs=1 2> /dev/null| xxd -p -c 64)) 128\"}}")
     kubectl patch secret -n kube-system cilium-ipsec-keys -p="${data}" -v=1
```
pchaigno, line 172: Maybe out-of-scope for this PR, but I'm wondering if we could add something in the list of commands to fail early if a command failed? Like some check on the `KEYID` or on `data` (length check?)?

pchaigno (approved):

Documentation/security/network/encryption-ipsec.rst
jschwinger233: Don't know if it's a good idea, but how about making it a `cilium` sub-command, such as `cilium encrypt key-rotation`?

margamanterola: I think a cilium encrypt subcommand to obtain the key information and another to do the complete key rotation would be quite valuable (but out of scope for this PR).

zacharysarah (approved):
LGTM for docs

Documentation/security/network/encryption-ipsec.rst
margamanterola: Filed https://github.com/cilium/cilium-cli/issues/1971 for this. Resolving this conversation.

PR #28146, opened 2023-09-13
metrics: add bpf_map_capacity metric which provides max size of maps

tommyp1ckles:
It is useful to have bpf map capacity metric, which is the defined bpf max size, for writing general Promtheus queries that can detect various issues with map usage (for example, providing an upper bound on FQDN IPs).

This is especially helpful in the face of configurable map sizes, where you cannot assume the default value is always used. 

bpf_map_capacity metrics are emitted per "group" of maps, this is any group of maps that share common purpose and attributes.
This avoids having excessive redundant cardinality due to similar map types each having their capacity emitted in a separate metric (ex. endpoint policy maps are all always the same size).

As well, maps that are the default map capacity size (i.e. 2^16) are not included in this metric.

A follow up to this change will be adding a bpf map size metric, which can be used in conjunction with this metric to specify exactly how much capacity there is left in a map.

bpf_map_capacity metrics are emitted


zacharysarah (approved):
LGTM for docs

pkg/bpf/map_linux.go
```diff
@@ -155,6 +159,13 @@ func (m *Map) Flags() uint32 {
 	return 0
 }
 
+func (m *Map) updateMetrics() {
```
rgo3, line 162: This and the call to `metrics.UpdateMapCapacity()` don't perform any update AFAIU, looks like something is missing to perform an actual operation?

nathanjsweet (approved):

pkg/bpf/map_linux.go
tommyp1ckles: Woops, was making some last minute refactors 😅 ... Fixed.

rgo3 (approved):

derailed (approved):
Nice work! LGTM.

markpash (approved):

PR #27983, opened 2023-09-07
daemon: Fix boostrap restore duration metric

PlatformLC:
The start timestamp update for 'restoreOldEndpoints' is missed for bootstrap stats calculation.

```release-note
Fix inaccurate calculation for bootstrap stats of restore
```

PR #28150, opened 2023-09-14
endpoint: correct stats for prepareBuild

PlatformLC:
Previous change relocated CT cleanup blocker, but left prepareBuild
stats calculated after that. Afterwards, some modifications also been
done for the code related to prebuild. So the stats related to prepareBuild
should be handled within runPreCompilationSteps.

Fixes: eaa486d787ef ("endpoint: Wait for CT cleanup to complete before BPF compilation")

```release-note
correct stats calculation for prepareBuild of endpoint_regeneration_time 
```

### DIFF
--- a/pkg/endpoint/bpf.go
+++ b/pkg/endpoint/bpf.go
@@ -543,8 +543,6 @@ func (e *Endpoint) regenerateBPF(regenContext *regenerationContext) (revnum uint
 	<-datapathRegenCtxt.ctCleaned
 	stats.waitingForCTClean.End(true)
 
-	stats.prepareBuild.End(true)
-
 	compilationExecuted, err = e.realizeBPFState(regenContext)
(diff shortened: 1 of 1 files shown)


aditighag (approved):
Thanks! I'm curious: how did you notice the bug? The commit you've linked is a bit old. Did you see incorrect stats?
nit: *leaved -> left  in the commit description

PlatformLC:
> Thanks! I'm curious: how did you notice the bug? The commit you've linked is a bit old. Did you see incorrect stats? nit: *leaved -> left in the commit description

Thanks for pointing out the typo! I just went through stats related implement and found these.
I think the *double*  `End`s have no functionality impact since the stats lib has considered such case, but we should clean it up.  Lack of `Start` has functionality impact, e.g. #27983

PR #28154, opened 2023-09-14
egressgw: make reconciliationEventsCount an atomic.Uint64

jibi:
as otherwise the reconciliation count logic in tests would be racy


lmb (approved):
Thanks :)

PR #28157, opened 2023-09-14
Respond with ICMP reply for traffic to services without backends

### LINKED ISSUE #10002: Failures in [sig-network] Network should set TCP CLOSE_WAIT timeout and Services should be rejected when no endpoints exist

dylandreimerink:
So far we have been dropping packets meant for services which do not have endpoints/backends. This causes clients to needlessly wait for replies and to retry sending traffic. This PR adds the ability to send back an ICMP or ICMPv6 reply with Destination unreachable (type 3) + Port unreachable (code 3) whenever this happens. 

This behavior is controllable via a new `--service-no-backend-response` flag, which defaults to `reject` so we match expected behavior by default. It can also be set to `drop` to preserve the existing behavior in case that was desired.

This new behavior works for both North/South traffic entering a node and East/West traffic responding to a request from a pod within the cluster.

Fixes: #10002

```release-note
Respond with ICMP reply for traffic to services without backends
```


lmb:
Meta comment without looking at the code yet: 1.8k lines is a lot. Can you reduce the size of this PR by pulling out some commits?

dylandreimerink:
Its 1.18k not 1.8k. 580 of which is tests. Can't really pull anything out that is meaningful in terms of LOC. We could do the IPv4 and IPv6 seperately 🤔

test/l4lb/test.sh
```diff
@@ -135,11 +135,11 @@ docker exec -t lb-node docker exec -t cilium-lb \
 
 # Do not stop on error
 set +e
-# Issue 10 requests to LB (with 500ms timeout) which are expected to timeout
+# Issue 10 requests to LB, we expect all to fail due to a Failed connection
 for i in $(seq 1 10); do
     curl -o /dev/null -m 0.5 "${LB_VIP}:80"
-    # code 28 - Operation timeout
-    if [ ! "$?" -eq 28 ]; then
+    # code 7 - Failed to connect ... : No route to host
```
lmb, line 141: Won't this break bisection between the commit where you change behaviour and this one?

bpf/complexity-tests/419/bpf_lxc/4.txt
```diff
@@ -24,3 +24,4 @@
 -DENABLE_IPV4_FRAGMENTS=1
 -DENCAP_IFINDEX=1
 -DTUNNEL_MODE=1
+-DSERVICE_NO_BACKEND_RESPONSE=1
```
lmb, line 27: What logic did you use to choose 0 vs 1? It'd be nice if you could document your thinking in the commit message.

bpf/lib/lb.h
```diff
@@ -1207,7 +1207,7 @@ lb4_extract_tuple(struct __ctx_buff *ctx, struct iphdr *ip4, int l3_off, int *l4
 			return ret;
 		return 0;
 	case IPPROTO_ICMP:
```
lmb, line 1207: Not related to this PR: does this mean we don't support forwarding ICMP to backends?

bpf/lib/nodeport.h
```diff
@@ -1322,6 +1322,14 @@ static __always_inline int nodeport_lb6(struct __ctx_buff *ctx,
 		ret = lb6_local(get_ct_map6(&tuple), ctx, l3_off, l4_off,
 				&key, &tuple, svc, &ct_state_new,
 				skip_l3_xlate, ext_err);
+
+#if SERVICE_NO_BACKEND_RESPONSE == 1
```
lmb: Why is does this compare against 1? Is there a comparison against 0 somewhere? From what I can tell the convention seems to be `if defined()` or similar instead of comparing the value.

bpf/lib/lb.h
```diff
@@ -1748,8 +1752,341 @@ lb4_ctx_restore_state(struct __ctx_buff *ctx, struct ct_state *state,
...
+		sample_len = ICMP_PACKET_MAX_SAMPLE_SIZE;
+	ctx_adjust_troom(ctx, sample_len + sizeof(struct ethhdr) - ctx_full_len(ctx));
+
+	data = ctx_data(ctx);
+	data_end = ctx_data_end(ctx);
+
+	/* Unrolled loop to calculate the checksum of the ICMP sample
+	 * Done manually because the compiler refuses with #pragma unroll
+	 */
+	csum = 0;
```
lmb: I'm really uneasy about this. Sticking it in the middle of the function makes it harder to test (is there a test somewhere?)

bpf/lib/lb.h
```diff
@@ -1748,8 +1752,341 @@ lb4_ctx_restore_state(struct __ctx_buff *ctx, struct ct_state *state,
...
+	ethhdr->h_proto = bpf_htons(ETH_P_IP);
+
+	/* Write reversed ip header, ready for egress */
+	ip4 = data + sizeof(struct ethhdr);
+	ip4->version = 4;
+	ip4->ihl = sizeof(struct iphdr) >> 2;
+	ip4->tos = tos;
+	ip4->tot_len = bpf_htons(sizeof(struct iphdr) + sizeof(struct icmphdr) +
+		       (__u16)sample_len);
+	ip4->id = (__be16)get_prandom_u32();
```
lmb: Controversial choice :D Can't we just set it to 0? The packet isn't fragmented after all. https://datatracker.ietf.org/doc/html/rfc6864

bpf/lib/lb.h
```diff
@@ -1748,8 +1752,341 @@ lb4_ctx_restore_state(struct __ctx_buff *ctx, struct ct_state *state,
...
+	ip4->protocol = IPPROTO_ICMP;
+	ip4->check = 0;
+	ip4->daddr = saddr;
+	ip4->saddr = daddr;
+	ip4->check = csum_fold(csum_diff(ip4, 0, ip4, sizeof(struct iphdr), 0));
+
+	/* Write reversed icmp header */
+	icmphdr = data + sizeof(struct ethhdr) + sizeof(struct iphdr);
+	icmphdr->type = ICMP_DEST_UNREACH;
+	icmphdr->code = ICMP_HOST_UNREACH;
```
lmb: Have you checked what the consequences of going between host unreachable and port unreachable are? As in, how do network stacks react to this? Do they cache it?

I would be more comfortable with sticking to what kube proxy does.

bpf/tests/tc_lxc_lb4_no_backend.c
```diff
@@ -0,0 +1,140 @@
...
+	assert(l3->ttl == 64);
+	assert(l3->protocol == IPPROTO_ICMP);
+
+	l4 = data + sizeof(__u32) + sizeof(struct ethhdr) + sizeof(struct iphdr);
+	if ((void *) l4 + sizeof(struct icmphdr) > data_end)
+		test_fatal("l4 header out of bounds");
+
+	assert(l4->type == ICMP_DEST_UNREACH);
+	assert(l4->code == ICMP_HOST_UNREACH);
+	assert(l4->checksum == bpf_htons(0x7992));
```
lmb: How did you come up with this magic number?

bpf/tests/tc_lxc_lb6_no_backend.c
```diff
@@ -0,0 +1,147 @@
...
+	assert(l3->version == 6);
+	assert(l3->nexthdr == IPPROTO_ICMPV6);
+
+	l4 = data + sizeof(__u32) + sizeof(struct ethhdr) + sizeof(struct ipv6hdr);
+	if ((void *) l4 + sizeof(struct icmp6hdr) > data_end)
+		test_fatal("l4 header out of bounds");
+
+	assert(l4->icmp6_type == ICMPV6_DEST_UNREACH);
+	assert(l4->icmp6_code == ICMPV6_ADDR_UNREACH);
+	assert(l4->icmp6_cksum == bpf_htons(0x7da9));
```
lmb: Dito.

lmb (dismissed):
I'd really like to see and end to end test for this (sorry :see_no_evil:) ICMP and checksumming is especially finicky, and I'd like to make sure that we really get the expected reaction from a real network stack. It would also be great if the checksumming was it's own helper that was tested with more than three packets.

At my previous job we bungled the same ICMP checksumming code and would generate an invalid checksum with a probability of 1/1^16 or so.

bpf/lib/lb.h
asauber: My quick thought here is that a Service K8s resource is defined as being associated with a port, so it would be important to NOT use host unreachable, as the IP is likely to be shared among multiple services and/or reused within some cache interval.

bpf/lib/lb.h
joestringer: Yeah, k8s services operate on a specific port. You can hence have many services on the same IP on different ports. Depending on the ICMP message type there may not be a port at all, so how do you pick the right service?

test/nat46x64/test.sh
```diff
@@ -151,6 +151,9 @@ for i in $(seq 1 10); do
     curl -s -o /dev/null "${LB_VIP}:80" || (echo "Failed $i"; exit -1)
 done
 
+# Give the agent some time to persist service update to the datapath
+sleep 5
```
joestringer: Please avoid sleeps in tests wherever possible. This is the cause of flakiness all over the place. Maybe query for presence of the service/backend instead on a regular period and fail out if it never shows up?

bpf/lib/lb.h
```diff
@@ -1748,8 +1752,341 @@ lb4_ctx_restore_state(struct __ctx_buff *ctx, struct ct_state *state,
...
+	#pragma unroll
+	for (i = 0; i < 20; i++)
+		csum += pseudo_header.words[i];
+
+	icmphdr->icmp6_cksum = csum_fold(csum);
+
+	/* Redirect ICMP to the interface we received it on. */
+	cilium_dbg_capture(ctx, DBG_CAPTURE_DELIVERY,
+			   ctx_get_ifindex(ctx));
+	return ctx_redirect(ctx, ctx_get_ifindex(ctx), 0);
```
joestringer, line 2035: How do we ratelimit ICMP? [RFC4443](https://datatracker.ietf.org/doc/html/rfc4443) 2.4.f requires implementations to limit how many ICMPv6 messages that a node generates.

bpf/lib/lb.h
```diff
@@ -1748,8 +1752,341 @@ lb4_ctx_restore_state(struct __ctx_buff *ctx, struct ct_state *state,
...
+	/* Fill pseudo header */
+	memcpy(&pseudo_header.fields.src_ip, &ip6->saddr, sizeof(struct in6_addr));
+	memcpy(&pseudo_header.fields.dst_ip, &ip6->daddr, sizeof(struct in6_addr));
+	pseudo_header.fields.top_level_length = bpf_htonl(sizeof(struct icmp6hdr) +
+						(__u32)sample_len);
+	__bpf_memzero(pseudo_header.fields.zero, sizeof(pseudo_header.fields.zero));
+	pseudo_header.fields.next_header = IPPROTO_ICMPV6;
+
+	#pragma unroll
+	for (i = 0; i < 20; i++)
```
joestringer: What's `20` here?

bpf/lib/lb.h
```diff
@@ -1748,8 +1752,341 @@ lb4_ctx_restore_state(struct __ctx_buff *ctx, struct ct_state *state,
...
+
+	body128(0)
+	body128(256)
+	body128(512)
+	body128(768)
+	body128(1024)
+
+csumend :
+
+	/* We need to insert a IPv4 and ICMP header before the original packet.
```
joestringer: Why IPv4 in the IPv6 function?

bpf/lib/lb.h
```diff
@@ -1748,8 +1752,341 @@ lb4_ctx_restore_state(struct __ctx_buff *ctx, struct ct_state *state,
...
+	 * Make that room.
+	 */
+
+#if __ctx_is == __ctx_xdp
+	xdp_adjust_head(ctx, 0 - (int)(sizeof(struct ipv6hdr) + sizeof(struct icmp6hdr)));
+#else
+	skb_adjust_room(ctx, sizeof(struct ipv6hdr) + sizeof(struct icmp6hdr), BPF_ADJ_ROOM_MAC, 0);
```
joestringer: No error checking?

Also curious why this isn't using `ctx_adjust_hroom`? Is there an import issue or something that prevents reuse of that helper?

joestringer (commented):
I got a bit curious about the checksum implementation and I got engrossed in digging around in some protocol details. Forgive the pass-by review :sweat_smile: . Not a comprehensive review by any means, I just noticed a few things. I note a few points where you've clearly been implementing to the RFC specs, but there was also some finer points of the RFCs not yet addressed as far as I could tell.

test/l4lb/test.sh
dylandreimerink: It will if you run CI while bisecting. I believe we expect the code to compile between every commit, not for the full CI suite to pass for every commit. That is what https://github.com/cilium/cilium/actions/runs/6197972695/job/16827550943 checks at least.

bpf/lib/nodeport.h
dylandreimerink: Yea, that is my bad. For a while it looked like we might need multiple fallback modes so that suck around, will update it.

bpf/lib/lb.h
dylandreimerink: Oh. Honestly never dove that deep into the RFCs. I have never seen 0 ids on PCAPs before so I figured setting wasn't optional :grin:. But it indeed seems like we can.

bpf/lib/lb.h
dylandreimerink: That is a good point, I don't know if this is cached somehow. Agreed, lets stick to port unreachable for now since we know that works for others.

bpf/tests/tc_lxc_lb4_no_backend.c
dylandreimerink: I dumped the context of the crafted packet with the DUMP_CTX option in the test runner, imported that into wireshark which then tells you if the checksum is correct and if not what it should be. I will add a comment to explain that.

test/nat46x64/test.sh
dylandreimerink: Yea, might be able to use the `cilium bpf ...` commands to check when the service is actually "ready"

bpf/lib/lb.h
dylandreimerink: Well we currently don't since I glossed over that part of the spec :fearful:. Any suggestions on how to add that? 

First instinct is to do so in BPF, add a map to track messages and just drop the incoming packet if we hit the quota. Though I suspect someone will insist on making such a mechanism generic and reusable to warrant the cost of introducing a new map.

Second thought is to re-use parts of the bandwidth manager. Which would mean that if we want this behavior to be the default (as it should be for k8s conformance) that bandwidth manager is always enabled. We might be able to just re-use the TC setup in the background so we can use it to rate-limit, without having to enable the whole feature. But it sounds ticky in any case.

I would urge a follow up PR for this works since it seems significant to add, even though that means we technically are not RFC compliant for now. I would argue that its of less importance in this code path anyway since we specifically deal with replies for services without backends, which is a very specific scenario as apposed to something like a echo reply.

bpf/lib/lb.h
dylandreimerink: Its the size of the words array `__u16 words[20];` in the pseudo header, but I think I can exchange that for `sizeof(pseudo_header.words)/sizeof(__u16)` to make it more clear

bpf/lib/lb.h
dylandreimerink: > No error checking?

Yea, I don't really have a recourse if these error, and I don't think it warrants creating a special error code. But I can always do an early return.

> Also curious why this isn't using ctx_adjust_hroom?

Well, initially I used the `bpf_skb_change_head` helper, but it doesn't for TC until v5.8, I only later switched. Though looking at the possibility now, I use the `BPF_ADJ_ROOM_MAC` mode which isn't implemented in our XDP implementation of the helper, the code currently asserts the `BPF_ADJ_ROOM_NET` mode. So we would have to add that as well (and the PR is already chunky).

test/nat46x64/test.sh
dylandreimerink: So long story short, checking with `cilium bpf ...` commands did not work. My new hack is to allow a few connections to fail initially until it starts working with delays so we don't have a "dumb" sleep just there.

bpf/lib/lb.h
joestringer: I'm not quite sure about the "map to track messages" approach, the standard just suggests a token bucket implementation. Bandwidth manager seems like it may have more in common.

As a general principle, I would be more comfortable with merging such functionality into a branch and potentially breaking up individual tasks into PRs to target that branch, then only merge the branch into main when the feature is complete / standards-compliant. I worry that it's too easy to say "well that's a big item so we'll do it later" and then later becomes ~forever, and Cilium ends up releasing a version that violates well-established standards. This is part of the spec for the feature that's being added in this PR.

dylandreimerink: > I worry that it's too easy to say "well that's a big item so we'll do it later" 

I personally don't mind putting it all into one PR, but I have been told that smaller PRs are easier to review hence the initial suggestion. :grin: 

> I would be more comfortable with merging such functionality into a branch and potentially breaking up individual tasks into PRs to target that branch

If we can merge this into a `feature/...` branch with reviewed PRs and then merge the whole branch into main as a whole that would be optimal in my opinion.

> I'm not quite sure about the "map to track messages" approach, the standard just suggests a token bucket implementation. Bandwidth manager seems like it may have more in common.

I meant implementing something like a token bucket algorithm with a map as storage. But I agree that reusing the bandwidth manager would be better than making something new at this point.

> [RFC4443](https://datatracker.ietf.org/doc/html/rfc4443) 2.4.f requires implementations to limit how many ICMPv6 messages

Motivation behind this PR is reports of users hitting this issue. Would it be acceptable to rip out the IPv6 path since its only the ICMPv6 spec that requires rate limiting? That way we can ship something in a reasonable amount of time, given I expect the bandwidth manager route to take significant time

I realize that is another way of saying "well that's a big item so we'll do it later", but its for a good cause :angel:

joestringer: I've created `feature/main/svc-icmp-response`, you should be able to submit PRs against that branch. The goal should be to ensure that merges into that branch match the expectations for merging into `main`. Periodically if necessary we can rebase it (I think you should have the ability to do this, but let me know if not). Once we think the feature is complete we can submit a PR to pull that branch into main.

> Motivation behind this PR is reports of users hitting this issue. Would it be acceptable to rip out the IPv6 path since its only the ICMPv6 spec that requires rate limiting? That way we can ship something in a reasonable amount of time, given I expect the bandwidth manager route to take significant time

This has been the behaviour for _years_. This is a brand new feature being submitted half way through the v1.15 development cycle. I don't see the urgency. I'll note that arguments like this are a common cause of tech debt and also a big reason why various features don't work with IPv6. By all means we can have a discussion with the relevant sig-datapath folks to set the expectations according to the shared community quality bar, but I don't really buy the argument why corners must be cut here.

.github/workflows/conformance-ingress.yaml
```diff
@@ -242,8 +242,25 @@ jobs:
       - name: Run Sanity check (external)
         timeout-minutes: 5
         run: |
+          set +e
```
asauber: Adding errexit here will cause this test to exit upon the curl failing, and not reach the `if [ $? -ne 0 ];` lines.

install/kubernetes/cilium/README.md
```diff
@@ -740,6 +740,7 @@ contributors across the globe, there is almost always someone available to help.
 | serviceAccounts.clustermeshcertgen | object | `{"annotations":{},"automount":true,"create":true,"name":"clustermesh-apiserver-generate-certs"}` | Clustermeshcertgen is used if clustermesh.apiserver.tls.auto.method=cronJob |
 | serviceAccounts.hubblecertgen | object | `{"annotations":{},"automount":true,"create":true,"name":"hubble-generate-certs"}` | Hubblecertgen is used if hubble.tls.auto.method=cronJob |
 | serviceAccounts.nodeinit.enabled | bool | `false` | Enabled is temporary until https://github.com/cilium/cilium-cli/issues/1396 is implemented. Cilium CLI doesn't create the SAs for node-init, thus the workaround. Helm is not affected by this issue. Name and automount can be configured, if enabled is set to true. Otherwise, they are ignored. Enabled can be removed once the issue is fixed. Cilium-nodeinit DS must also be fixed. |
+| serviceNoBackendResponse | string | `"reject"` | Configure what the response should be to traffic for a service without backends. Possible values:  - reject (default)  - drop |
```
asauber: A note here along the lines of "earlier kernels (< 5.10[?]) do not support the 'reject' mode", would be helpful. An upgrade note along these lines I think is also required.

.github/workflows/conformance-ingress.yaml
```diff
@@ -242,8 +242,25 @@ jobs:
...
           lb=$(kubectl get ingress basic-ingress -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
-          curl -s -v --connect-timeout 5 --max-time 20 --retry 3 --fail -- http://"$lb"
+          
+          # It may take some time after the service gets created for it to handle traffic.
+          # An un-ready service will return an ICMP so we retry a few times since the --retry
+          # flag only retries on connection timeouts.
+          curl -s -v --connect-timeout 5 --max-time 20 --fail -- http://"$lb"
+          if [ $? -ne 0 ]; then
+            sleep 5s
```
asauber: We want to avoid sleeps in tests, and this could be a loop with a counter rather than nested, but I think both are _okay_ here.

asauber (changes requested):
Thanks for switching to port unreachable.

A few more minor comments.

.github/workflows/conformance-ingress.yaml
joestringer: FWIW the danger is that if we have a bug that causes LB decisions to be flaky or inconsistent rather than moving from reliably not working to reliably working, then this style of testing will hide such flaky production behaviour.

joestringer:
Marking as draft, I think we should come up with a plan for this submission with @cilium/sig-datapath around how we develop this to spec of the related RFCs ([related thread above](https://github.com/cilium/cilium/pull/28157#discussion_r1353297596)).

.github/workflows/conformance-ingress.yaml
dylandreimerink: I believe this does the exact opposite, by default these scripts are run with `set -e` which causes an exit of the whole script on error, I had to add this line so I could use the `if [ $? -ne 0 ];` checks

bpf/lib/lb.h
```diff
@@ -1917,13 +1918,24 @@ int __tail_no_service_ipv6(struct __ctx_buff *ctx)
 	union macaddr dmac = {};
 	struct in6_addr saddr;
 	struct in6_addr daddr;
+	struct ratelimit_key rkey = {};
+	/* Rate limit to 100 ICMPv6 repies per second, burtable to 1000 responses/s */
```
joamaki: typos

kaworu (approved):
Hubble changes LGTM

viktor-kurchenko (approved):

sayboras (approved):
LGTM on helm changes

.github/workflows/conformance-ingress.yaml
```diff
@@ -243,7 +243,21 @@ jobs:
         timeout-minutes: 5
         run: |
           lb=$(kubectl get ingress basic-ingress -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
-          curl -s -v --connect-timeout 5 --max-time 20 --retry 3 --fail -- http://"$lb"
+
```
asauber: Instead of adding this logic, have you tried using cURL's `--retry-all-errors` flag? There is also `--retry-connrefused`.

It could simplify things here.

asauber (approved):
Another suggestion for the integration test. Not blocking.

Token bucket implementation looks clean. Thanks again again for port unreachable.

.github/workflows/conformance-ingress.yaml
dylandreimerink: Awesome, didn't know about this flag. Works according to CI, so going with this! Thx

joestringer:
I note that my previous specific feedback regarding ratelimiting in RFC 4443 was addressed with additional changes, but I haven't had a chance to review them yet. Just wondering @dylandreimerink whether you've been able to review the relevant RFCs and confirm that the implementation is conforming to those internet standards?.

dylandreimerink:
> Just wondering @dylandreimerink whether you've been able to review the relevant RFCs and confirm that the implementation is conforming to those internet standards?.

I have, this is the relevant section of RFC 4443 2.4(f)

>A recommended method for implementing the rate-limiting function
is a token bucket, limiting the average rate of transmission to
N, where N can be either packets/second or a fraction of the
attached link's bandwidth, but allowing up to B error messages to
be transmitted in a burst, as long as the long-term average is
not exceeded.
>
>Rate-limiting mechanisms that cannot cope with bursty traffic
(e.g., traceroute) are not recommended; for example, a simple
timer-based implementation, allowing an error message every T
milliseconds (even with low values for T), is not reasonable.
>
>The rate-limiting parameters SHOULD be configurable.  In the case
of a token-bucket implementation, the best defaults depend on
where the implementation is expected to be deployed (e.g., a
high-end router vs. an embedded host).  For example, in a
small/mid-size device, the possible defaults could be B=10,
N=10/s.

I implemented the exact algorithm described. The RFC did not specify which key should be used for the bucket. I took a look at several docs for routers which all seem to use the netdev as key, so that is what I did here as well.

The settings are configurable in the ratelimit implementation, however I hardcoded them at the current callsite. The RFC words it as `SHOULD` so this seems acceptable. Docs for the router implementations seem to default to 1% of netdev bandwidth. But adding a flag or passing netdev size to the datapath seemed overkill in this situation.

The current values are 100 requests per second average and burst up to 1000 requests per second. These seemed sane values for this very specific corner case of ICMPv6 reply.

joestringer:
Awesome. Did you review the rest of the RFC as well?

dylandreimerink:
>  Did you review the rest of the RFC (and corresponding IPv4 RFCs) as well?

I did, skipped parts that didn't seem relevant such as the formats of message types we are not dealing with or IANA number allocation procedures. Also RFC 792, again skipped unrelated message types and skimmed over RFC 2460, was mostly interested in 8.1 which is about checksum generation for ICMPv6 (and other high-level protocols).

Did I miss something obvious?

nathanjsweet (approved):

pkg/datapath/linux/probes/probes.go
```diff
@@ -484,6 +484,57 @@ func HaveOuterSourceIPSupport() (err error) {
...
+	prog, err := ebpf.NewProgram(progSpec)
+	if err != nil {
+		return err
+	}
+	defer prog.Close()
+
+	// This is a Eth + IPv4 + UDP + data packet. The helper relies on a valid packed being passed in
+	// since it wants to know offsets of the different layers.
+	pkt := []byte{
+		0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x0e, 0xf5, 0x16, 0x3d, 0x6b, 0xab, 0x08, 0x00, 0x45, 0x00,
```
ti-mo: Generate this using `gopacket` :pray:

pkg/datapath/linux/probes/probes.go
```diff
@@ -484,6 +487,78 @@ func HaveOuterSourceIPSupport() (err error) {
...
+		asm.FnSkbAdjustRoom.Call(),
+		asm.Return(),
+	}
+	prog, err := ebpf.NewProgram(progSpec)
+	if err != nil {
+		return err
+	}
+	defer prog.Close()
+
+	// This is a Eth + IPv4 + UDP + data packet. The helper relies on a valid packed being passed in
```
ti-mo: Nit (sorry)
```suggestion
	// This is a Eth + IPv4 + UDP + data packet. The helper relies on a valid packet being passed in
```

pkg/datapath/linux/probes/probes.go
```diff
@@ -484,6 +487,78 @@ func HaveOuterSourceIPSupport() (err error) {
...
+// to be expanded by 20 bytes. The helper checks the `mode` argument and will return -ENOSUPP if
+// the mode is unknown. Otherwise it should resize the SKB by 20 bytes and return 0.
+func HaveSKBAdjustRoomL2RoomMACSupport() (err error) {
+	defer func() {
+		if err != nil && !errors.Is(err, ebpf.ErrNotSupported) {
+			log.WithError(err).Fatal("failed to probe for bpf_skb_adjust_room L2 room MAC support")
+		}
+	}()
+
+	if err := rlimit.RemoveMemlock(); err != nil {
```
ti-mo: This can likely be removed, same with the one in `HaveOuterSourceIPSupport()` above.

ti-mo (approved):

pkg/datapath/linux/probes/probes.go
dylandreimerink: (sigh)

aojea:
lovely

aojea:
> This differs slightly from the kube-proxy behavior which would send back and type 3/code 3 port unreachable, given its the whole IP not a single port the address unreachable code made more sense

A funny thing is that it can be technically possible, that in the same Service, one Port has backends and other not, but  this is an artifact of the capability of using named ports for Services , but as Tim says in , that is a pretty esoteric configuration https://github.com/kubernetes/kubernetes/issues/24875#issuecomment-794596576 so I think that this is correct

dylandreimerink:
> > This differs slightly from the kube-proxy behavior which would send back and type 3/code 3 port unreachable, given its the whole IP not a single port the address unreachable code made more sense

> A funny thing is that it can be technically possible, that in the same Service, one Port has backends and other not, but this is an artifact of the capability of using named ports for Services , but as Tim says in , that is a pretty esoteric configuration https://github.com/kubernetes/kubernetes/issues/24875#issuecomment-794596576 so I think that this is correct

Ah, the description of the PR is out of date, we changed this after review to match exactly what kube-proxy does to avoid potential implementation differences of the ICMP code for clients. I will correct the description so it doesn't cause future confusion.

bpf/lib/ratelimit.h
```diff
@@ -0,0 +1,76 @@
...
+		if (ret < 0)
+			return false;
+		return true;
+	}
+
+	/* Topup the bucket if it has been at least more than 1 interval since we have done so */
+	since_last_topup = ktime_get_ns() - value->last_topup;
+	if (since_last_topup > settings->topup_interval_ns) {
+		/* Add tokens of every missed interval */
+		value->tokens += (since_last_topup / settings->topup_interval_ns) *
```
gentoo-root, line 59: Rounding here could skip intervals. For example, it this function is called at 0 s, 1.5 s and 3 s, it will add 1000 tokens at 0 s, another 1000 tokens at 1.5 s, and yet another 1000 tokens at 3 s = in total 3000 tokens. If, however, this function was called at 0 s, 1 s, 2 s and 3 s, it would add 1000 tokens at each call = in total 4000 tokens over the same time period.

bpf/lib/ratelimit.h
```diff
@@ -0,0 +1,76 @@
...
+	struct ratelimit_value new_value;
+	__u64 since_last_topup;
+	int ret;
+
+	/* Create a new bucket if we do not yet have one for the key */
+	value = map_lookup_elem(&RATELIMIT_MAP, key);
+	if (!value) {
+		new_value.last_topup = ktime_get_ns();
+		new_value.tokens = settings->tokens_per_topup - 1;
+		ret = map_update_elem(&RATELIMIT_MAP, key, &new_value, BPF_ANY);
```
gentoo-root, line 49: This lookup-and-update is racy if called from two CPUs. Do we care?

bpf/lib/ratelimit.h
```diff
@@ -0,0 +1,76 @@
...
+		return true;
+	}
+
+	/* Topup the bucket if it has been at least more than 1 interval since we have done so */
+	since_last_topup = ktime_get_ns() - value->last_topup;
+	if (since_last_topup > settings->topup_interval_ns) {
+		/* Add tokens of every missed interval */
+		value->tokens += (since_last_topup / settings->topup_interval_ns) *
+						  settings->tokens_per_topup;
+		value->last_topup = ktime_get_ns();
```
gentoo-root, line 61: We should reuse the ktime_get_ns() value fetched above, otherwise it's another source of inexactness, although only a tiny one.

gentoo-root (commented):
I'm too late for the party, but I've got some comments on the TBF implementation :sweat_smile:

bpf/lib/ratelimit.h
dylandreimerink: I considered this. It would mean that the ratelimit isn't 100% accurate, letting more traffic through than the limit. To fix that we would need to use atomics which are slow. I thought the performance is more important in this given situation. Perhaps we should add this to the comments in case other wonder.

bpf/lib/ratelimit.h
dylandreimerink: Right. So a better way to keep track would be:
```
long cur_time = ktime_get_ns();
[...]
long intervals = since_last_topup / settings->topup_interval_ns;
long remainder = since_last_topup % settings->topup_interval_ns;
value->last_topup = cur_time - remainder;
```

So a 1.5s, we set `last_topup` to `1s` instead of `1.5s`. So at the 3s mark we would add 2000 instead of 1000.

Is that correct?

gentoo-root: Looks correct to me.

bpf/lib/lb.h
```diff
@@ -1741,8 +1746,357 @@ lb4_ctx_restore_state(struct __ctx_buff *ctx, struct ct_state *state,
...
+
+	/* Add ICMP header checksum to sum of its body */
+	csum += csum_diff(icmphdr, 0, icmphdr, sizeof(struct icmphdr), 0);
+	icmphdr->checksum = csum_fold(csum);
+
+	/* Redirect ICMP to the interface we received it on. */
+	cilium_dbg_capture(ctx, DBG_CAPTURE_DELIVERY,
+			   ctx_get_ifindex(ctx));
+	return ctx_redirect(ctx, ctx_get_ifindex(ctx), 0);
+}
```
julianwiedmann, line 1864: I believe this needs an `edt_set_aggregate(ctx, 0)`, to prevent false-positives in `to-netdev`'s Bandwidth-Manager code.

julianwiedmann, line 1864: When this ICMP packet hits `to-netdev`, did you check how it interacts with the [SNAT engine](https://github.com/cilium/cilium/blob/2daca4c3935a43dac72044c7ca3e17ac57af2617/bpf/bpf_host.c#L1423)?

I would expect that it gets [dropped](https://github.com/cilium/cilium/blob/2daca4c3935a43dac72044c7ca3e17ac57af2617/bpf/lib/nat.h#L778), whenever the addressed service IP equals [IPV4_MASQUERADE](https://github.com/cilium/cilium/blob/2daca4c3935a43dac72044c7ca3e17ac57af2617/bpf/lib/nat.h#L526).

julianwiedmann (commented):
Two comments below - better late than never :).

bpf/lib/lb.h
dylandreimerink: That is a good point, I have not been able to check the interaction with the bandwidth manager. But I suspect you are right. Will have to do a followup.

dylandreimerink: > did you check how it interacts with the [SNAT engine](https://github.com/cilium/cilium/blob/2daca4c3935a43dac72044c7ca3e17ac57af2617/bpf/bpf_host.c#L1423)?

No, I did not. 

> I would expect that it gets [dropped](https://github.com/cilium/cilium/blob/2daca4c3935a43dac72044c7ca3e17ac57af2617/bpf/lib/nat.h#L778), whenever the addressed service IP equals [IPV4_MASQUERADE](https://github.com/cilium/cilium/blob/2daca4c3935a43dac72044c7ca3e17ac57af2617/bpf/lib/nat.h#L526).

Right, which would be in the case of a node port service?

julianwiedmann: > which would be in the case of a node port service?

Correct. That's a scenario we want to support, right?

bpf/bpf_lxc.c
```diff
@@ -108,6 +108,14 @@ static __always_inline int __per_packet_lb_svc_xlate_4(void *ctx, struct iphdr *
...
 				ETH_HLEN, l4_off, &key, &tuple, svc, &ct_state_new,
 				has_l4_header, false, &cluster_id, ext_err);
+
+#ifdef SERVICE_NO_BACKEND_RESPONSE
+		if (ret == DROP_NO_SERVICE) {
+			ep_tail_call(ctx, CILIUM_CALL_IPV4_NO_SERVICE);
+			return DROP_MISSED_TAIL_CALL;
+		}
+#endif
+
```
julianwiedmann, line 118: With ep-routes enabled, the ICMP packet should now pass through `to-container` on the way back into the pod. Would we thus require an ingress network policy change to allow this traffic?

This feels very similar to the topic of avoiding policy for service loopback replies ...

bpf/lib/lb.h
dylandreimerink: Yes, I think so. How does host traffic normally deal with this? Additionally, I see where have a marker to skip SNAT `ctx_snat_done_set(ctx);` would calling it before doing the redirect help?

It has been a while since I looked in depth at the SNAT path.

bpf/bpf_lxc.c
dylandreimerink: A CT entry should be created for the outgoing connection to the service, so when doing policy checking the ICMP reply should be flagged as return traffic an thus not subject to any ingress policy.

julianwiedmann: Unfortunately [lb4_local()](https://github.com/cilium/cilium/blob/4400169ba429babe013419b363d8c00e6872ff0f/bpf/lib/lb.h#L1579) currently doesn't create the RELATED entry (note the `NULL` for `map_related`). So I don't think there's any CT entry in place that would allow such ICMP traffic to pass through network policy enforcement.

bpf/lib/lb.h
julianwiedmann: > Yes, I think so. How does host traffic normally deal with this?

It gets dropped :). We only support a limited set of ICMP types. I hope we can extend this as needed.

> Additionally, I see where have a marker to skip SNAT `ctx_snat_done_set(ctx);` would calling it before doing the redirect help?

Yep, I had the same thought. I believe that would fit as work-around (and would even allow you to skip the HostFW in `to-netdev` ... that's another aspect we didn't consider yet in this PR). Long-term it would be best to teach the SNAT engine about this ICMP type.

bpf/bpf_lxc.c
dylandreimerink: Right, but isn't the CT entry created [here](https://github.com/cilium/cilium/blob/4400169ba429babe013419b363d8c00e6872ff0f/bpf/bpf_lxc.c#L959) before we get to the LB stage?

dylandreimerink: I see that depends on `ENABLE_PER_PACKET_LB` being enabled or not

julianwiedmann: > Right, but isn't the CT entry created [here](https://github.com/cilium/cilium/blob/4400169ba429babe013419b363d8c00e6872ff0f/bpf/bpf_lxc.c#L959) before we get to the LB stage?

Nope, that part is only reached *after* selecting the backend (this CT entry tracks the client -> backend connection).

bpf/lib/common.h
```diff
@@ -597,7 +599,7 @@ enum {
 #define DROP_INVALID_EXTHDR	-156
 #define DROP_FRAG_NOSUPPORT	-157
 #define DROP_NO_SERVICE		-158
-#define DROP_UNUSED8		-159 /* unused */
```
ti-mo, line 600: @dylandreimerink This drop reason wasn't added in `flow.proto` and `drop.go`.

Under normal circumstances, we shouldn't reuse any of these, since renaming a proto field/type causes a backwards-incompatible change. We're lucky that in this case, drop reason `159` is actually missing from the proto as well as from drop.go. :sweat_smile: 

```
    SERVICE_BACKEND_NOT_FOUND = 158;
    NO_TUNNEL_OR_ENCAPSULATION_ENDPOINT = 160;
```

In any case, I'm marking all unused ones as deprecated in https://github.com/cilium/cilium/pull/29482.

cc @rolinh

bpf/lib/lb.h
julianwiedmann: Copied to https://github.com/cilium/cilium/issues/33041.

bpf/bpf_lxc.c
julianwiedmann: Copied to https://github.com/cilium/cilium/issues/33041.

bpf/lib/lb.h
julianwiedmann: I'm playing around in https://github.com/cilium/cilium/pull/33046.

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -2276,6 +2276,13 @@ routingMode: ""
 # @default -- Port 8472 for VXLAN, Port 6081 for Geneve
 tunnelPort: 0
 
+# -- Configure what the response should be to traffic for a service without backends.
+# "reject" only works on kernels >= 5.10, on lower kernels we fallback to "drop".
+# Possible values:
+#  - reject (default)
```
julianwiedmann, line 2282: @dylandreimerink is the >= 5.10 requirement correct? We later say that the required BPF feature (`BPF_ADJ_ROOM_MAC` support in `bpf_skb_adjust_room()`) exists since kernel 5.2.

dylandreimerink: Yes, I believe you are correct. That should be 5.2, I am not sure how 5.10 ended up there.

julianwiedmann: ack, I glanced at a 5.4 sysdump and it looks all golden there. No error message from the probe.

I'll open a PR to just remove this on v1.16+, given all supported kernels should be providing the option.

julianwiedmann: https://github.com/cilium/cilium/pull/35734

PR #28162, opened 2023-09-14
build: fix usage of local golangci-lint installation

mhofstetter:
The make target `make lint` should prefer the locally installed `golangci-lint` if the installed version matches the expected one. This comes with the advantage of using the local cache.

It seems that there are differences in the version representation between the upstream binaries (including the `v` prefix) and the one installed via `go install ...` (without the prefix). The current implementation doesn't support the upstream binaries with the `v` prefix - it always falls back to executing the docker version.

This commit introduces support for both formats by stripping any given `v` prefix before comparison.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -414,7 +414,7 @@ govet: ## Run govet on Go source files in the repository.
 	$(QUIET) $(GO_VET) ./...
 
 golangci-lint: ## Run golangci-lint
-ifneq (,$(findstring $(GOLANGCILINT_WANT_VERSION),$(GOLANGCILINT_VERSION)))
+ifneq (,$(findstring $(GOLANGCILINT_WANT_VERSION:v%=%),$(GOLANGCILINT_VERSION)))
 	@$(ECHO_CHECK) golangci-lint $(GOLANGCI_LINT_ARGS)
 	$(QUIET) golangci-lint run $(GOLANGCI_LINT_ARGS)
 else
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -97,7 +97,7 @@ endif
 # renovate: datasource=docker depName=golangci/golangci-lint
 GOLANGCILINT_WANT_VERSION = v1.54.2
 GOLANGCILINT_IMAGE_SHA = sha256:2082f5379c48c46e447bc1b890512f3aa9339db5eeed1a483a34aae9476ba6ee
-GOLANGCILINT_VERSION = $(shell golangci-lint version 2>/dev/null)
+GOLANGCILINT_VERSION = $(shell golangci-lint version --format short 2>/dev/null)
 
 VERSION = $(shell cat $(dir $(lastword $(MAKEFILE_LIST)))/VERSION)
 VERSION_MAJOR = $(shell cat $(dir $(lastword $(MAKEFILE_LIST)))/VERSION | cut -d. -f1)


Makefile
```diff
@@ -414,7 +414,7 @@ govet: ## Run govet on Go source files in the repository.
 	$(QUIET) $(GO_VET) ./...
 
 golangci-lint: ## Run golangci-lint
-ifneq (,$(findstring $(GOLANGCILINT_WANT_VERSION),$(GOLANGCILINT_VERSION)))
+ifneq (,$(findstring $(GOLANGCILINT_VERSION),$(GOLANGCILINT_WANT_VERSION)))
```
tklauser: Another option would be to use `make`'s [substitution reference](https://www.gnu.org/software/make/manual/html_node/Substitution-Refs.html) to strip of the `v` from `$(GOLANGCILINT_WANT_VERSION)`:

```suggestion
ifneq (,$(findstring $(GOLANGCILINT_WANT_VERSION:v%=%),$(GOLANGCILINT_VERSION)))
```

(this is what we [use in the cilium-cli repo](https://github.com/cilium/cilium-cli/blob/a681b5e588e9542685070f00749fdaefc2737c57/Makefile#L86))

mhofstetter: thanks - good idea. Especially in combination with `golangci-lint version --format short` that would even allow for an equal comparison (instead of the `findstring`)) :thinking:

tklauser:
> (e.g. `golangci-lint has version 1.54.2 built with go1.21.0 from 411e0bbb on 2023-08-21T12:04:32Z` (that doesn't contain the version with a prefix `v`)

Interestingly, my locally installed `golangci-lint` does have the `v` prefix:

```% golangci-lint version
golangci-lint has version v1.54.2 built with go1.21.1 from (unknown, mod sum: "h1:oR9zxfWYxt7hFqk6+fw6Enr+E7F0SN2nqHhJYyIb0yo=") on (unknown)
```

I guess this is because I installed it using `go install`. The binaries provided by upstream don't have the `v` prefix. Presumably this is why we use `findstring` instead of an exact match.

mhofstetter:
> > (e.g. `golangci-lint has version 1.54.2 built with go1.21.0 from 411e0bbb on 2023-08-21T12:04:32Z` (that doesn't contain the version with a prefix `v`)
> 
> Interestingly, my locally installed `golangci-lint` does have the `v` prefix:
> 
> ```
> golangci-lint has version v1.54.2 built with go1.21.1 from (unknown, mod sum: "h1:oR9zxfWYxt7hFqk6+fw6Enr+E7F0SN2nqHhJYyIb0yo=") on (unknown)
> ```
> 
> I guess this is because I installed it using `go install`. The binaries provided by upstream don't have the `v` prefix. Presumably this is why we use `findstring` instead of an exact match.

Interesting find. I wondered why this is "suddenly" no longer working. I was pretty sure that this worked even after introducing automated Renovate upgrades. Currently i have it installed via their recommended "[binaries](https://golangci-lint.run/usage/install/#binaries)" approach.


What's the output of `golangci-lint version --format short` in your version? If this is more reliable - we should switch to this instead of the full version.

Makefile
mhofstetter: I switched to your proposed option - but kept the short format.

tklauser: Yeah, the short format is indeed safer (e.g. if the Go version included in the long format would match at some point :smile:). Will change to the short format in cilium-cli as well.

tklauser (approved):

Makefile
mhofstetter: great. thanks Tobias!

PR #28173, opened 2023-09-15
pkg/policy: index policies by k8s UID

aanm:
As k8s UID are unique, we can index the policies by their UId. This will decrease Cilium's boot-up time for clusters with a large number of network policies.

For a cluster with a lot of network policies that was taking ~4m20s to boot Cilium, with this change it will take ~20 seconds.

```release-note
Cut Cilium's initialization time for clusters with a large number of Kubernetes and Cilium Network Policies by 90%
```

### DIFF
--- a/pkg/policy/repository.go
+++ b/pkg/policy/repository.go
@@ -111,6 +111,9 @@ type Repository struct {
 	Mutex lock.RWMutex
 	rules ruleSlice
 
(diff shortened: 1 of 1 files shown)


christarazi (approved):
Nice!

nathanjsweet (approved):
I like it!

Issue #22133, opened 2022-11-14
CFP: Support omitting local/remote ASN with BGP CPlane


YutaroHayakawa:
## Cilium Feature Proposal

**Is your feature request related to a problem?**

Currently, users need to specify both local and remote ASN explicitly. Similar to the peer IP address problem described in #22132, this is automation unfriendly since users need to import this knowledge from an external registry.

**Describe the feature you'd like**

Support omitting local and remote ASN on `CiliumBGPPeeringPolicy`. We need to have a different approach for both of them. By combining it with BGP unnumbered support, it can dramatically reduce the configuration knob users need to care about (I guess with both support, users potentially can configure an entire cluster with a single CiliumBGPPeetingPolicy).

**(Optional) Describe your proposed solution**

1. Omitting local ASN

The potential idea to solve this is users can specify some ASN ranges, and Cilium assigns them to each virtual router. Or we can take some node-specific value and map it to the 32bit private ASN range (approx 26bits of free space).

2. Omitting remote ASN

Theoretically, it is possible to determine a remote peer's ASN from BGP OPEN message. In fact, FRR supports `remote-as external/internal` which only cares about the remote peer is internal or external peer. However, I'm not sure GoBGP supports it or not.

**Potential concern**
- We are currently using local ASN as an identifier of the virtual routers. Once we start to assign it automatically, it is hard for users to specify a virtual router. For example, we support overriding local router-id with annotation, but that annotation depends on the local ASN (https://docs.cilium.io/en/v1.12/gettingstarted/bgp-control-plane/#virtual-router-attributes).

bewing:
Note that IETF is performing work on BGP ASN autodetection/autoconfiguration:
https://datatracker.ietf.org/doc/draft-ietf-idr-bgp-autoconf-considerations/
https://datatracker.ietf.org/doc/draft-acee-idr-lldp-peer-discovery/
https://datatracker.ietf.org/doc/draft-ietf-lsvr-l3dl/

Issue #22132, opened 2022-11-14
CFP: BGP Unnumbered support for BGP CPlane


YutaroHayakawa:
## Cilium Feature Proposal

**Is your feature request related to a problem?**

Currently, users need to specify BGP peers with IP addresses. Thus, it needs explicit knowledge about the peer configuration. This is automation unfriendly since users need to import such knowledge from an external system like DC-wide IPAM.

**Describe the feature you'd like**

In modern DC networks, the technique called "BGP Unnumbered" is often used, which typically allows users to specify peers with network interface names like `eth0`. This is more automation-friendly than specifying a peer with an IP address because users don't need to know the peer's IP address and only know which interface it is connected to.

BGP Unnumbered consists of multiple techniques (there are several ways, but below I'll describe it with FRR's implementation)
1. BGP peering with IPv6 link-local address (users don't need to assign IP address for peering, this is why it is called "unnumbered").
2. Exchanging IPv4 prefix with IPv6 next hop ([RFC8960](https://datatracker.ietf.org/doc/rfc8950/))
3. Automatic neighbor discovery with IPv6 RA (users don't need to know the peer IP address, we can find it automatically with the IPv6 discovery mechanism).
4. Installing IPv4 prefix with IPv6 next hop to DPlane's FIB

**(Optional) Describe your proposed solution**

The configuration knob of `CiliumBGPPeeringPolicy` would look like the below. It may be worth having an annotation to override it per-node for the case nodes don't have a consistent device naming scheme.

```
apiVersion: "cilium.io/v2alpha1"
kind: CiliumBGPPeeringPolicy
metadata:
 name: 01-bgp-peering-policy
spec: # CiliumBGPPeeringPolicySpec
 nodeSelector:
   matchLabels:
     bgp-policy: a
 virtualRouters: # []CiliumBGPVirtualRouter
 - localASN: 64512
   exportPodCIDR: true
   neighbors: # []CiliumBGPNeighbor
    - peerInterface: 'eth0' <= new
      # peerAddress: 'fc00:f853:ccd:e793::50/128'
      peerASN: 64512
```

- For 1 and 2, GoBGP already supports it (including the configuration knob to specify peer with interface name). We just need to wire up it with Cilium's configuration knob. The only concern is in BGP unnumbered, users typically assign the node's IP address to the loopback address, keep physical devices unnumbered, and advertise the loopback interface's address with BGP. I don't know whether it works with the features such as KPR, which needs explicit IP address assignment to devices.
[...]

PR #28175, opened 2023-09-15
k8s: Introducing new BGP CRDs

harsimran-pabla:
Introduce new BGP CRDs to enable a more flexible way to configure BGP in the Cilium cluster.

- CiliumBGPClusterConfig
- CiliumBGPAdvertisement
- CiliumBGPPeerConfig
- CiliumBGPNodeConfig
- CiliumBGPNodeOverrideConfig

Please take a look at [CFP](https://docs.google.com/document/d/1mqmzlDeVKmNjsnDnnHTeoiW6JcItYMyDyuOM7SYoDi4/edit#heading=h.5mwz58y9syps) for more details. 

Note - The backend to consume these CRDs is not implemented. It is the first of many PRs to follow for complete implementation. 

```release-note
Introduce new BGP CRDs to provide a more flexible way to configure BGP in Cilium.
```


rastislavs:
CC @danehans - you may be interested in this.

harsimran-pabla:
I have identified minor updates to this PR, marking it draft till then.

danehans:
@rastislavs thanks for the cc. @harsimran-pabla I provided several comments throughout the [CFP](https://isogo.to/bgpv2-crds). I'd like to start the conversation there before providing specific feedback on this PR.

rastislavs (approved):
LGTM

danehans:
I see that [this comment](https://docs.google.com/document/d/1mqmzlDeVKmNjsnDnnHTeoiW6JcItYMyDyuOM7SYoDi4/edit?disco=AAAA4y5wpt4) in the design doc requested a review from @youngnick. IMHO it's a good idea to get API reviews from a wider audience before merging.

harsimran-pabla:
Hi @danehans, Yes sig-k8s group needs to review this change before merging.

youngnick:
I've commented on the CFP doc, but I agree that we need some changes to that doc before we move forward here.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_advert_types.go
```diff
@@ -0,0 +1,86 @@
...
+	// +kubebuilder:validation:MinItems=1
+	Advertisements []Advertisement `json:"advertisements"`
+}
+
+// Advertisement defines which routes Cilium should advertise to BGP peers. Optionally, additional attributes can be
+// set to the advertised routes.
+type Advertisement struct {
+	// AdvertisementType defines type of advertisement.
+	// - For "PodCIDR" the Selector matches k8s node resources (resulting into PodCIDR routes).
+	// - For "CiliumPodIPPool" the Selector matches CiliumPodIPPool custom resources (resulting into PoolCIDR routes).
```
danehans: matches advertised CIDRs of matching CiliumLoadBalancerIPPool custom resources.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
```diff
@@ -0,0 +1,120 @@
...
+	// Items is a list of CiliumBGPClusterConfig.
+	Items []CiliumBGPClusterConfig `json:"items"`
+}
+
+type CiliumBGPClusterConfigSpec struct {
+	// NodeSelector selects a group of nodes where this BGP Peering
+	// Policy applies.
+	//
+	// If empty / nil this policy applies to all nodes.
+	//
```
danehans: Consider using terminology other than "Peering Policy" as this may cause confusion with pkg/bgpv1 CiliumBGPPeeringPolicy.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
```diff
@@ -0,0 +1,120 @@
...
+	// +listMapKey=peerAddress
+	Peers []CiliumBGPPeer `json:"peers"`
+}
+
+type CiliumBGPPeer struct {
+	// PeerAddress is the IP address of the neighbor.
+	// Supports IPv4 and IPv6 addresses.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Format=ip
```
danehans: For backwards compatibility with CiliumBGPPeerinPolicy, the `ip` format will get converted to /32 `cidr` format, correct?

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
```diff
@@ -0,0 +1,120 @@
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=4294967295
+	PeerASN int64 `json:"peerASN"`
+
+	// PeerTemplateRef is a reference to a peer template resource which defines
+	// the configuration of this neighbor.
+	//
+	// +kubebuilder:validation:Optional
+	PeerTemplateRef *PeerTemplateReference `json:"peerTemplateRef,omitempty"`
```
danehans: xref my design doc [comment](https://docs.google.com/document/d/1mqmzlDeVKmNjsnDnnHTeoiW6JcItYMyDyuOM7SYoDi4/edit?disco=AAAA7i9ppiY), IMHO the concept of "Peer Groups" should be considered instead of CiliumBGPPeerTemplate to reduce the number of CRDs.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_override_types.go
```diff
@@ -0,0 +1,100 @@
...
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:categories={cilium,ciliumbgp},singular="ciliumbgpnodeconfigoverride",path="ciliumbgpnodeconfigoverrides",scope="Cluster",shortName={bgpnodeoverride}
+// +kubebuilder:printcolumn:JSONPath=".metadata.creationTimestamp",name="Age",type=date
+// +kubebuilder:storageversion
+
+// CiliumBGPNodeConfigOverride is used to overrides some of the BGP configurations which are node local.
+// Users can user this resource to override auto-generated BGP settings for the node.
+type CiliumBGPNodeConfigOverride struct {
```
danehans, line 19: Since CiliumBGPNodeConfigOverride and CiliumBGPClusterConfig are both cluster-scoped, what are the benefits of having a separate CRD for defining overrides instead of including a field in CiliumBGPClusterConfig? For example:

```yaml
apiVersion: cilium.io/v1alpha1
kind: CiliumBGPClusterConfig
metadata:
  name: example-config
spec:
  nodeSelector:
    matchLabels:
      rack: rack1
  bgpInstances:
  - localASN: 65001
    peerGroups:
    - name: example-group
      peers:
      - name: peer1
        address: 10.0.0.1
      - name: peer2
        address: 10.0.0.2
      - name: peer3
        address: 10.0.0.3
      peerASN: 65002
      eBGPMultihopTTL: 3
     ...
    groupOverrides:
    - peerName: peer3 
      eBGPMultihopTTL: 1
```

I feel like adding concepts such as peer groups and overrides to CiliumBGPClusterConfig provides a simple mental model for users while achieving the desired config flexibility. WDYT?

pkg/k8s/apis/cilium.io/v2alpha1/bgp_peer_types.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:categories={cilium,ciliumbgp},singular="ciliumbgppeertemplate",path="ciliumbgppeertemplates",scope="Cluster",shortName={bgppeer}
+// +kubebuilder:printcolumn:JSONPath=".metadata.creationTimestamp",name="Age",type=date
+// +kubebuilder:storageversion
+
+type CiliumBGPPeerTemplate struct {
```
danehans: As I mentioned in my other comment, IMHO we should consider adding `peerGroups` to the CiliumBGPClusterConfig CRD. A `peerGroup` would define BGP config that applies to all the peers in the group. For example:

```yaml
apiVersion: cilium.io/v1alpha1
kind: CiliumBGPClusterConfig
metadata:
  name: example-config
spec:
  nodeSelector:
    matchLabels:
      rack: rack1
  bgpInstances:
  - localASN: 65001
    peerGroups:
    - name: example-group
      peers:
      - name: peer1
        address: 10.0.0.1
      - name: peer2
        address: 10.0.0.2
      - name: peer3
        address: 10.0.0.3
      peerASN: 65002
      eBGPMultihopTTL: 3
     ...
```

WDYT?

danehans (changes requested):
@harsimran-pabla thank you for driving this effort and providing responses to my design doc comments. PTAL at my review feedback when you have a moment and let me know your thoughts.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_override_types.go
harsimran-pabla: Override is not changing CiliumBGPNodeConfig settings, but other auto-generated settings by the cilium agent like router-id and source IP for peering. This configuration is not set in CiliumBGPClusterConfig, so adding an override section to set per node config seems wrong. 

CiliumBGPClusterConfig applies to the group of nodes. Whereas, CiliumBGPNodeConfigOverride applies to a single node.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_peer_types.go
harsimran-pabla: Not sure how this is removing duplication of configuration among multiple CiliumBGPClusterConfig. 

```yaml
apiVersion: cilium.io/v1alpha1
kind: CiliumBGPClusterConfig
metadata:
  name: example-config
spec:
  nodeSelector:
    matchLabels:
      rack: rack1
  bgpInstances:
  - localASN: 65001
    peerGroups:
    - name: example-group
      peers:
      - name: peer1
        address: 10.0.0.1
      peerASN: 65002
      eBGPMultihopTTL: 3
     ... common config ...
```

For rack 2, we will have this configuration

```
apiVersion: cilium.io/v1alpha1
kind: CiliumBGPClusterConfig
metadata:
  name: example-config-2
spec:
  nodeSelector:
    matchLabels:
      rack: rack2
  bgpInstances:
  - localASN: 65001
    peerGroups:
    - name: example-group-2
      peers:
      - name: peer2
        address: 10.0.0.2
      peerASN: 65002
      eBGPMultihopTTL: 3
     ... common config ...
```

Node selector associates example-config to rack 1 and example-config-2 to rack 2. 

Common configuration is repeated between example-config and example-config-2. This common configuration can be lot of BGP knobs which is eliminated if we use peer template.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
harsimran-pabla: I am not exactly sure why /32 was there in the first place. It is an IP address, not an exact match prefix. 

I was thinking of doing this cleanup with this change.

harsimran-pabla: Yes, you're right. For backward compatibility, /32 from CiliumBGPPeeringPolicy will get converted to IP address.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_override_types.go
danehans: @harsimran-pabla thanks for the clarification on how CiliumBGPNodeConfigOverride works.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_peer_types.go
danehans: Thanks for the additional details. I was building a mental model around a single CiliumBGPClusterConfig. With that said, I believe the ^ use case could be accomplished by moving the node selection to the peer group level. For example:

```yaml
apiVersion: cilium.io/v1alpha1
kind: CiliumBGPClusterConfig
metadata:
  name: example-config
spec:
  bgpInstances:
  - localASN: 65001
    peerGroups:
    - name: rack1
      peers:
      - name: peer1
        address: 10.0.0.1
      - name: peer2
        address: 10.0.0.2
      - name: peer3
        address: 10.0.0.3
      peerASN: 65002
      eBGPMultihopTTL: 3
      nodeSelector:
        matchLabels:
          rack: rack1
    - name: rack2
      peers:
      - name: peer1
        address: 11.0.0.1
      - name: peer2
        address: 11.0.0.2
      - name: peer3
        address: 11.0.0.3
      peerASN: 65002
      eBGPMultihopTTL: 3
      nodeSelector:
        matchLabels:
          rack: rack2
     ...
```

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,174 @@
...
+	// LocalASN is the ASN of this virtual router.
+	// Supports extended 32bit ASNs
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=4294967295
+	LocalASN int64 `json:"localASN"`
+
+	// RouterID is the BGP router ID of this virtual router.
+	//
```
danehans, line 75: Since this is an optional field, consider adding `// When unspecified this default to....`.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_advert_types.go
harsimran-pabla: updated the definition.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
harsimran-pabla: General consensus in sig-bgp group is to keep peer config separate from cluster config. Pros and cons are discussed in the linked thread.

@danehans  Let me know if we can resolve this comment.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_peer_types.go
harsimran-pabla: @danehans  Let me know if we can resolve this comment. Let's discuss design in CFP, and whether we need to modify this approach or not.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
danehans: @harsimran-pabla you can consider this comment resolved. Thanks for the consideration and feedback on the "Peer Groups" concept.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_peer_types.go
danehans: Yes, this comment can be resolved and any additional discussion on this topic can take place in the design doc.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
```diff
@@ -0,0 +1,119 @@
...
+	// +listMapKey=peerAddress
+	Peers []CiliumBGPPeer `json:"peers"`
+}
+
+type CiliumBGPPeer struct {
+	// PeerAddress is the IP address of the neighbor.
+	// Supports IPv4 and IPv6 addresses.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Format=ip
```
danehans: I see `ipv4` and `ipv6` as supported formats in the [go-openapi toolkit](https://pkg.go.dev/github.com/go-openapi/strfmt) but not `ip`. According to `kubernetes/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1/types_jsonschema.go`, unknown formats are ignored. Where is this format documented?

harsimran-pabla: Ah you're right, good catch! I wonder if there is way for conditional check. Or we have to go regex route to validate v4/v6 address.

danehans: Unless we have separate IPv4 and IPv5 fields, we will need to use regex or validate within the controller.

danehans: `ip` format will not be supported until it's added to [JSON schema validation](https://json-schema.org/draft/2020-12/json-schema-validation#name-ip-addresses).

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
harsimran-pabla: Added comment that this configuration is derived from CiliumBGPNodeConfigOverride resource. In that resource, it is clarified if router-id is empty, it will be node's IP address.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
harsimran-pabla: Added regex pattern match.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
danehans: Great. You can consider this comment resolved.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_override_types.go
```diff
@@ -0,0 +1,100 @@
...
+	//
+	// +kubebuilder:validation:Required
+	LocalASN int64 `json:"localASN"`
+
+	// RouterID is BGP router id to use for this instance. It must be unique across all BGP instances.
+	// If not specified, the node's IP address will be used.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Format=ipv4
+	RouterID string `json:"routerID,omitempty"`
```
danehans: Should this type be a pointer for consistency with other optional types? I see this elsewhere, so do a find/replace through all API types if you agree with this comment.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,175 @@
...
+	BGPInstances []CiliumBGPNodeInstance `json:"bgpInstances"`
+}
+
+// CiliumBGPNodeInstance is a single BGP router instance configuration on the node.
+type CiliumBGPNodeInstance struct {
+	// LocalASN is the ASN of this virtual router.
+	// Supports extended 32bit ASNs
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=0
```
danehans: I know this is a carryover from `CiliumBGPVirtualRouter`, but why are `0` and `23456` acceptable values when [RFC 7607](https://www.rfc-editor.org/rfc/rfc7607.html) states that these numbers are reserved and shouldn't be used in BGP operations? If you agree with this concern, feel free to keep the validation as-is and create a tracker issue for further discussion.

Same comment for `peerASN`.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,175 @@
...
+	// +kubebuilder:validation:Maximum=4294967295
+	LocalASN int64 `json:"localASN"`
+
+	// RouterID is the BGP router ID of this virtual router.
+	// This configuration is derived from CiliumBGPNodeConfigOverride resource.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Format=ipv4
+	RouterID string `json:"routerID,omitempty"`
```
danehans: Do we anticipate the need to support v6 peering in the future?

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,175 @@
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +listType=map
+	// +listMapKey=peerAddress
+	Peers []CiliumBGPNodePeer `json:"peers"`
+}
+
+type CiliumBGPNodePeer struct {
+	// PeerAddress is the IP address of the neighbor.
+	// Supports IPv4 and IPv6 addresses.
```
danehans, line 106: If we support v4 and v6 peering, then shouldn't `routerID` support v4 and v6 addressing?

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// Uptime is the time since the last peering session was established.
+	Uptime string `json:"uptime,omitempty"`
+
+	// RoutesReceived is the number of routes received from this peer.
+	RoutesReceived int32 `json:"routesReceived,omitempty"`
+
+	// RoutesAdvertised is the number of routes advertised to this peer.
+	RoutesAdvertised int32 `json:"routesAdvertised,omitempty"`
+}
```
danehans: This status seems fairly large to start out with. Do you have any concerns about how this will scale?

pkg/k8s/apis/cilium.io/v2alpha1/bgp_peer_types.go
```diff
@@ -0,0 +1,148 @@
...
+	//
+	// +kubebuilder:validation:Optional
+	Families []CiliumBGPFamily `json:"families,omitempty"`
+}
+
+// CiliumBGPFamily represents a AFI/SAFI address family pair.
+type CiliumBGPFamily struct {
+	// Afi is the Address Family Identifier (AFI) of the family.
+	//
+	// +kubebuilder:validation:Enum=ipv4;ipv6;l2vpn;ls;opaque
```
danehans, line 98: I know this is a carryover but enums should be camel case: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants. If you agree, update throughout the new APIs.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_peer_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// +kubebuilder:validation:Required
+	Afi string `json:"afi,omitempty"`
```
danehans: Drop `omitempty` if this is a required field (same with `safi`).

pkg/k8s/apis/cilium.io/v2alpha1/bgp_peer_types.go
harsimran-pabla: This is a common structure used between old and new API, I do not want to modify it as it will break backward compatibility.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
harsimran-pabla: I don't have a concern about the size of the data, the scaling issue will come from the rate at which we update this information. We must throttle the rate of status updates in the agent so as not to flood the API server on BGP changes.

I think this information is a good starting point in the status field. More detailed information like routes, and policy information can be viewed via CLI.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
harsimran-pabla: Router ID of BGP is a 4 octet unsigned number, most commonly represented as an IPv4 address.
Ref: [RFC 4271](https://www.ietf.org/rfc/rfc4271.txt ) (BGP Identifier)

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
harsimran-pabla: We do support v6 peering, router-id does not impact peering setup.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_peer_types.go
danehans: One option is to create a new struct that is identical to CiliumBGPFamily but with camel-cased enums and is used by the new CRD. Not a bid deal, just a thought to align new CRDs with k8s API conventions as best as possible. Feel free to resolve this comment if you prefer sticking with the current enums.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
danehans: OK, feel free to resolve this comment.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
danehans: Ack, thanks for the reference.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
harsimran-pabla: hmm, I updated ASN to start from 1, but can't exclude 23456. That is something we will have to filter in the code.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_peer_types.go
harsimran-pabla: I am inclined to leave it as is. I understand camel case is standard practice, however, SAFI enums are getting a bit funky ( all the network acronyms when capitalized and underscore removed became unreadable ). 

Current mapping seems much more user-friendly.

danehans: OK, thanks for the feedback. You can consider this comment resolved.

danehans:
Should the design of these CRDs take https://github.com/cilium/cilium/issues/22132 and https://github.com/cilium/cilium/issues/22133 into account?

cc @YutaroHayakawa since you created both issues.

YutaroHayakawa:
Thanks for the heads up! I think the BGP Unnumbered can be implemented by extending `BGPPeer` by introducing `BGPPeer.PeerInterface` or something similar. However, I found an issue with supporting the AS autonumbering. As I wrote in the https://github.com/cilium/cilium/issues/22133, once we start to assign local ASN automatically, users cannot identify router instances because they don't know which instance got which ASN. So, I think we need to introduce some ID for instances other than ASN.

joamaki (approved):
The k8s API changes look technically good to me! Expecting BGP folks to review content.

danehans:
> I think the BGP Unnumbered can be implemented by extending BGPPeer by introducing BGPPeer.PeerInterface or something similar.

I agree, but [PeerAddress](https://github.com/cilium/cilium/pull/28175/files#diff-9e4cbf879956d8be7341fb3944f8c35d50f669d5f5a43a05c0f0465debb90581R81-R86) is a required field. To support https://github.com/cilium/cilium/issues/22132, PeerAddress should be optional with docs that include "this field is ignored when `peerInterface` is set."

> As I wrote in the https://github.com/cilium/cilium/issues/22133, once we start to assign local ASN automatically, users cannot identify router instances because they don't know which instance got which ASN. So, I think we need to introduce some ID for instances other than ASN.

The approach that many other projects take is to require a `name` field in the list and use this field as the map key.

harsimran-pabla:
@danehans 
> I agree, but [PeerAddress](https://github.com/cilium/cilium/pull/28175/files#diff-9e4cbf879956d8be7341fb3944f8c35d50f669d5f5a43a05c0f0465debb90581R81-R86) is a required field. To support #22132, PeerAddress should be optional with docs that include "this field is ignored when `peerInterface` is set."

Do you see an issue going from required to optional when we implement BGP unnumbered? With the current implementation, we do not support BGP unnumbered, so peer address is a required field. We can make it optional when the BGP unnumbered feature is implemented.

danehans:
> Do you see an issue going from required to optional when we implement BGP unnumbered?

This would be a backward incompatible change. Even though it's an alpha API, it's easier to start with `+optional` and a pointer type if we intend to implement https://github.com/cilium/cilium/issues/22132. A map key other than `peerAddress` is also required (see https://github.com/cilium/cilium/pull/28175#issuecomment-1755859561 for details). The same approach also applies to `peerAsn` to support https://github.com/cilium/cilium/issues/22133.

> We can make it optional when the BGP unnumbered feature is implemented.

If you believe the two referenced issues will be implemented in the near future, it's better to prepare the APIs now. The APIs are alpha so it's understandable if backward incompatible changes are made to support these issues. I don't have a strong opinion, I just want to make sure everyone is aware of the issues.

harsimran-pabla:
@danehans  please review the latest changes, now referencing both BGP instances and peers by name, which is a mandatory field and the rest of the fields are optional.

ldelossa (approved):

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,183 @@
...
+package v2alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:categories={cilium,ciliumbgp},singular="ciliumbgpnodeconfig",path="ciliumbgpnodeconfigs",scope="Cluster",shortName={bgpnode}
```
YutaroHayakawa: @youngnick Do you have any opinion about the `shortName` naming convention? You mentioned that the `Cilium` prefix in the full name helps avoid naming collision, but we don't have a cilium prefix here for short names. Currently, we don't have any rule for that, but once we expose this to the user, it's hard to change it later.

YutaroHayakawa (approved):
I have a small concern around the `shortName`, but since we don't expose this to users yet, we can solve it after merge. So, I'll approve.

youngnick:
>> Do you see an issue going from required to optional when we implement BGP unnumbered?

> This would be a backward incompatible change.

It's actually not! Because the value was required before, and then becomes no longer required, the previous value is _always_ valid under the new schema. It's going from optional to required that's a breaking change.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
youngnick: Yes, I think that maybe `cbgpnode` might be slightly better, although it will suck to type.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
```diff
@@ -0,0 +1,126 @@
...
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +listType=map
+	// +listMapKey=name
+	Peers []CiliumBGPPeer `json:"peers"`
+}
+
+type CiliumBGPPeer struct {
+	// Name is the name of the BGP peer.
```
danehans: nit: the godoc should state that name must be unique among the list of `peers`. This is implied by `listMapKey=name` but it's better to be explicit.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
```diff
@@ -0,0 +1,126 @@
...
+	PeerAddress *string `json:"peerAddress,omitempty"`
+
+	// PeerASN is the ASN of the peer BGP router.
+	// Supports extended 32bit ASNs.
+	//
+	// +kubebuilder:validation:Minimum=1
```
danehans, line 106: nit: `// +kubebuilder:validation:Optional`

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
```diff
@@ -0,0 +1,126 @@
...
+	// +kubebuilder:validation:Maximum=4294967295
+	PeerASN *int64 `json:"peerASN,omitempty"`
+
+	// PeerConfigRef is a reference to a peer configuration resource.
+	//
```
danehans: nit: not a bad idea to add godocs that state "when unspecified..." that describes the expected behavior when the field is unset.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
```diff
@@ -0,0 +1,126 @@
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default="cilium.io"
+	Group string `json:"group"`
+
+	// Kind is the kind of the peer config resource.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default="CiliumBGPPeerConfig"
+	Kind string `json:"kind"`
```
danehans: nit: if server-side defaulting is being used, then the fields can be optional. This will provide an improved UX for most use cases by only needing to supply a value for the `name` field.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// Name is the name of the peer config resource.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
```
danehans, line 137: nit: Since this API includes `group/kind`, the name should include but not be specific to `...peer config resource`. Also not a bad idea to include a few more details in the godocs and validation. For example:
```go
// Name refers to the name of a Kubernetes object (typically a CiliumBGPPeerConfig)
// and may have a variety of forms including RFC1123 subdomains, RFC 1123 labels,
// or RFC 1035 labels.
//
// +kubebuilder:validation:MinLength=1
// +kubebuilder:validation:MaxLength=253
```

__Note:__ I see this issue elsewhere. If you make the change here, please do the same throughout the APIs.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// Items is a list of CiliumBGPNodeConfig.
+	Items []CiliumBGPNodeConfig `json:"items"`
+}
+
+type CiliumBGPNodeSpec struct {
+	// BGPInstances is a list of BGP router instances on the node.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
```
danehans, line 49: nit: Not a bad idea to establish a max number of bgp instances and set `// +kubebuilder:validation:MaxItems`.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,183 @@
...
+type CiliumBGPNodeInstance struct {
+	// Name is the name of the BGP instance.
+	//
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// LocalASN is the ASN of this virtual router.
+	// Supports extended 32bit ASNs.
+	//
+	// +kubebuilder:validation:Minimum=1
```
danehans, line 69: nit: `// +kubebuilder:validation:Optional`

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,183 @@
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
+	// +kubebuilder:default=179
+	LocalPort *int32 `json:"localPort,omitempty"`
+
+	// Peers is a list of neighboring BGP peers for this virtual router
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
```
danehans: nit: Not a bad idea to establish a max number of bgp peers and set `// +kubebuilder:validation:MaxItems`.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,183 @@
...
+	PeerAddress *string `json:"peerAddress,omitempty"`
+
+	// PeerASN is the ASN of the peer BGP router.
+	// Supports extended 32bit ASNs
+	//
+	// +kubebuilder:validation:Required
```
danehans: If this field is a pointer and includes the `omitempty` tag, include `// +kubebuilder:validation:Optional` instead of `// +kubebuilder:validation:Required`.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,183 @@
...
+	// +listType=map
+	// +listMapKey=localASN
+	BGPInstances []CiliumBGPNodeInstanceStatus `json:"bgpInstances"`
+}
+
+type CiliumBGPNodeInstanceStatus struct {
+	// LocalASN is the ASN of this BGP instance.
+	//
+	// +kubebuilder:validation:Required
+	LocalASN int64 `json:"localASN"`
```
danehans: `spec.bgpInstances[].localASN` is now optional and `name` is now being used as the map key. You may want to do the same here. For example:
```go
type CiliumBGPNodeInstanceStatus struct {
	// Name refers to the name of a BGP router instance.
	//
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=253
	Name string `json:"name"`

	// LocalASN is the Autonomous System Number (ASN) of this
        // virtual BGP router. Supports extended 32-bit ASNs.
	//
	// +kubebuilder:validation:Minimum=1
	// +kubebuilder:validation:Maximum=4294967295
	LocalASN *int64 `json:"localASN,omitempty"`
...
}
```

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// PeerStatuses is the state of the BGP peers for this BGP instance.
+	//
+	// +listType=map
+	// +listMapKey=peerAddress
+	PeerStatuses []CiliumBGPNodePeerStatus `json:"peers"`
```
danehans: This is a required field. What if the bgp router instance has 0 peers? It seems like this field should be optional.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,183 @@
...
+type CiliumBGPNodePeerStatus struct {
+	// PeerAddress is the IP address of the neighbor.
+	//
+	// +kubebuilder:validation:Required
+	PeerAddress string `json:"peerAddress"`
+
+	// PeerASN is the ASN of the neighbor.
+	//
+	// +kubebuilder:validation:Required
+	PeerASN int64 `json:"peerASN"`
```
danehans: `peerAddress` and `peerAsn` are optional in spec and `name` is now being used as the map key. You may want to do the same here and include the same validation. For example:
```go
type CiliumBGPNodePeerStatus struct {
	// Name refers to the name of a BGP peer.
	//
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=253
	Name string `json:"name"`

	// PeerAddress is the IP address of the neighbor.
	//
// +kubebuilder:validation:Pattern=`((^\s*((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))\s*$)|(^\s*((([0-9A-Fa-f]{1,4}:){7}([0-9A-Fa-f]{1,4}|:))|(([0-9A-Fa-f]{1,4}:){6}(:[0-9A-Fa-f]{1,4}|((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){5}(((:[0-9A-Fa-f]{1,4}){1,2})|:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){4}(((:[0-9A-Fa-f]{1,4}){1,3})|((:[0-9A-Fa-f]{1,4})?:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){3}(((:[0-9A-Fa-f]{1,4}){1,4})|((:[0-9A-Fa-f]{1,4}){0,2}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){2}(((:[0-9A-Fa-f]{1,4}){1,5})|((:[0-9A-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){1}(((:[0-9A-Fa-f]{1,4}){1,6})|((:[0-9A-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(:(((:[0-9A-Fa-f]{1,4}){1,7})|((:[0-9A-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:)))(%.+)?\s*$))`
	PeerAddress *string `json:"peerAddress,omitempty"`

	// PeerASN is the Autonomous System Number (ASN) of the neighbor.
	//
	// +kubebuilder:validation:Minimum=1
	// +kubebuilder:validation:Maximum=4294967295
	PeerASN *int64 `json:"peerASN,omitempty"`
...
}
```

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,183 @@
...
+	Timers CiliumBGPTimersState `json:"timers,omitempty"`
+
+	// Uptime is the time since the last peering session was established.
+	Uptime string `json:"uptime,omitempty"`
+
+	// RoutesReceived is the number of routes received from this peer.
+	RoutesReceived int32 `json:"routesReceived,omitempty"`
+
+	// RoutesAdvertised is the number of routes advertised to this peer.
+	RoutesAdvertised int32 `json:"routesAdvertised,omitempty"`
```
danehans: These fields should be pointers and marked with `// +kubebuilder:validation:Optional` if they will be omitted from JSON serialization.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// CiliumBGPTimersState is the state of the negotiated BGP timers for a peer.
+type CiliumBGPTimersState struct {
+	// AppliedHoldTimeSeconds is the negotiated hold time for this peer.
+	AppliedHoldTimeSeconds int32 `json:"appliedHoldTimeSeconds,omitempty"`
+
+	// AppliedKeepaliveSeconds is the negotiated keepalive time for this peer.
+	AppliedKeepaliveSeconds int32 `json:"appliedKeepaliveSeconds,omitempty"`
```
danehans: These fields should be pointers and marked with `// +kubebuilder:validation:Optional` if they will be omitted from JSON serialization.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
```diff
@@ -0,0 +1,126 @@
...
+
+// CiliumBGPClusterConfig is the Schema for the CiliumBGPClusterConfig API
+type CiliumBGPClusterConfig struct {
+	// +deepequal-gen=false
+	metav1.TypeMeta `json:",inline"`
+	// +deepequal-gen=false
+	metav1.ObjectMeta `json:"metadata"`
+
+	// Spec defines the desired cluster configuration of the BGP control plane.
+	Spec CiliumBGPClusterConfigSpec `json:"spec,omitempty"`
```
danehans: `spec` should not be omitted from JSON serialization when it's a required field. Note that I see this issue with other APIs in this PR. If you change it here, do the same throughout the APIs.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_peer_types.go
```diff
@@ -0,0 +1,153 @@
...
+	// with this peer. If empty / nil, the graceful restart capability is disabled.
+	//
+	// +kubebuilder:validation:Optional
+	GracefulRestart *CiliumBGPNeighborGracefulRestart `json:"gracefulRestart,omitempty"`
+
+	// EBGPMultihopTTL controls the multi-hop feature for eBGP peers.
+	// Its value defines the Time To Live (TTL) value used in BGP
+	// packets sent to the peer.
+	// The value 1 implies that eBGP multi-hop feature is disabled
+	// (only a single hop is allowed).
```
danehans: `1` and unspecified mean the same behavior, e.g. "disabled". We should consider simplifying the API by simply saying "When unset, EBGP multihop is disabled. This field is ignored for iBGP neighbors.".

pkg/k8s/apis/cilium.io/v2alpha1/bgp_peer_types.go
```diff
@@ -0,0 +1,153 @@
...
+	// Advertisements selects group of BGP Advertisement(s) to advertise for this family.
+	//
+	// +kubebuilder:validation:Optional
+	Advertisements *slimv1.LabelSelector `json:"advertisements,omitempty"`
+}
+
+// CiliumBGPTransport defines the BGP transport parameters for the peer.
+type CiliumBGPTransport struct {
+	// LocalPort is the local port to be used for the BGP session.
+	//
```
danehans, line 122: nit: Although implied by `+kubebuilder:default=179`, it's not a bad idea to include a godoc that states the default, e.g. "// When unspecified, defaults to TCP port 179.". Note that this comment applies to all other optional fields that use this kubebuilder marker and the godocs do not explicitly state a default value.

danehans (approved):
@harsimran-pabla thanks for leading this effort and working through the API reviews. I'm marking `/approve` as most of my comments are nits that can be resolved in a follow-up PR.

danehans (approved):

pkg/k8s/apis/cilium.io/v2alpha1/bgp_cluster_types.go
youngnick: yeah, having a default and required doesn't really make sense - it should be either optional with a default or required, with the former being preferable for the reasons @danehans gives.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
harsimran-pabla: Set max to 16, not sure if there is a use case for more than a few BGP instances per node.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
harsimran-pabla: This is a bit tricky, as the number of peers can differ from deployment to deployment. I am leaning towards not setting any max on number of peers. 

Also, made this field optional.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
harsimran-pabla: It was carried over from CiliumBGPPeeringPolicy, I have made this field optional now.

kaworu (approved):
Helm changes LGTM.

harsimran-pabla:
> @harsimran-pabla thanks for leading this effort and working through the API reviews. I'm marking `/approve` as most of my comments are nits that can be resolved in a follow-up PR.

Thanks @danehans PTAL, I have fixed most of those comments.

rastislavs (approved):

nathanjsweet (approved):

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
YutaroHayakawa: @harsimran-pabla I think we should rename or add one more shortName. Especially, `bgppeer` is colliding with the MetalLB's `BGPPeer` CRD's name, so if users install MetalLB and Cilium at the same time, it may cause unexpected behavior.

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
youngnick: For MaxItems, it's way easier to lift that limit if required than to make it smaller later. 16 seems reasonable for a first guess to me.

pkg/k8s/apis/cilium.io/client/register.go
```diff
@@ -65,6 +65,21 @@ const (
...
+	BGPPCCRDName = k8sconstv2alpha1.BGPPCKindDefinition + "/" + k8sconstv2alpha1.CustomResourceDefinitionVersion
+
+	// BGPACRDName is the full name of the BGP Advertisement CRD.
+	BGPACRDName = k8sconstv2alpha1.BGPAKindDefinition + "/" + k8sconstv2alpha1.CustomResourceDefinitionVersion
+
+	// BGPNCCRDName is the full name of the BGP Agent CRD.
+	BGPNCCRDName = k8sconstv2alpha1.BGPNCKindDefinition + "/" + k8sconstv2alpha1.CustomResourceDefinitionVersion
+
+	// BGPNCOCRDName is the full name of the BGP Node Config Override CRD.
+	BGPNCOCRDName = k8sconstv2alpha1.BGPNCOKindDefinition + "/" + k8sconstv2alpha1.CustomResourceDefinitionVersion
```
youngnick: I know this is sticking to the existing convention, but having to count that (for example) it's _exactly_ three `C`s in `BGPCCCRDName` doesn't sound ideal.

I guess that IDEs will handle this, so this ones non-blocking feedback, especially since I can't think of anything better.

youngnick (approved):
The changes look good, but I'd really like to see some of the detail in the CFP end up in some documentation in the docs site before release. Can be in a follow-up, but this feature _badly_ needs some explanatory info for new users about how the objects fit together, and some examples of how to do a few different things (at the very least, the different types of Advertisement, but probably how to have different peer groups and so on too).

pkg/k8s/apis/cilium.io/v2alpha1/bgp_node_types.go
harsimran-pabla: @YutaroHayakawa I updated the short names to prefix them with 'c'.

harsimran-pabla:
> The changes look good, but I'd really like to see some of the detail in the CFP end up in some documentation in the docs site before release. Can be in a follow-up, but this feature _badly_ needs some explanatory info for new users about how the objects fit together, and some examples of how to do a few different things (at the very least, the different types of Advertisement, but probably how to have different peer groups and so on too).

@youngnick  Yes, I intend to re-write the BGP documentation, which clearly demonstrates configuration options and relationships between CRDs as well as some common configuration examples.

pkg/k8s/apis/cilium.io/client/register.go
harsimran-pabla: I agree, these short names were getting out of hand. I expanded it for better readability. 

Please have a look : 
https://github.com/cilium/cilium/compare/516204e8ac5b8c3d035b14797fc8b773d78f4f7b..06459b56e6e76781e8fc3259f57764d56f464f64

joestringer:
As triager, removing the ready-to-merge label due to failing tests and unresolved conversations. Please triage the failures and ensure they are not related to your PR, and work with the reviewers to resolve all outstanding feedback (or file issues for follow up if relevant).

harsimran-pabla:
[lbipam test failure](https://github.com/cilium/cilium/actions/runs/6535171242/job/17743995477): potentially  https://github.com/cilium/cilium/issues/28477

joestringer:
Ah I checked over the one remaining piece of feedback, it was mentioned to be non-blocking but you've also addressed it. Looks like this is good to merge, thanks for your work on this!

harsimran-pabla:
Thanks @joestringer, Yes I addressed Nick's comment. Was waiting for CI to rerun lbipam failure ( it passed so ready-to-merge label got readded ) 

Thanks for merging!

danehans:
@harsimran-pabla now that the APIs have landed, will you be creating issues to coordinate the implementation?

danehans:
@harsimran-pabla thanks for the feedback. As you suggest, it makes sense to have a CFP detail the implementation plans. I am also involved in another project but hope to have free cycles after KubeCon. I plan on reviewing/testing your operator code after KubeCon and will provide feedback.

PR #28180, opened 2023-09-15
.github: re-use common helm values from a single action

aanm:
A lot of our GH workflows use the same helm values to deploy Cilium.
Thus, it makes sense to re-use one file for that purpose use it as a GH
action composite.


qmonnet (approved):
Seems good to me, thanks!

joestringer (approved):
LGTM in principle

brlbil (approved):
LGTM

PR #28182, opened 2023-09-15
Enable k8s cache mutation detector in the CI

aanm:
This PR enables he k8s cache mutation detector in the CI and fixes the bugs already detected by it. Read on a per-commit basis.

### DIFF
--- a/.github/actions/helm-default/action.yaml
+++ b/.github/actions/helm-default/action.yaml
@@ -26,10 +26,15 @@ runs:
         else
           SHA="${{ github.sha }}"
         fi
-        echo sha=${SHA} >> $GITHUB_OUTPUT
 
         CILIUM_INSTALL_DEFAULTS="--chart-directory=${{ inputs.chart-dir }} \
-          --helm-set=image.repository=quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/cilium-ci \
+          --helm-set=debug.enabled=true \
+          --helm-set=bpf.monitorAggregation=none"
+
+        # only add SHA to the image tags if it was set
+        if [ -n "${SHA}" ]; then
+          echo sha=${SHA} >> $GITHUB_OUTPUT
+          CILIUM_INSTALL_DEFAULTS+=" --helm-set=image.repository=quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/cilium-ci \
           --helm-set=image.useDigest=false \
           --helm-set=image.tag=${SHA} \
           --helm-set=operator.image.repository=quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/operator \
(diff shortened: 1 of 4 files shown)


aanm:
The tests have passed, marking as ready and removing the "test commit"
image

squeed:
What happens if mutations are detected? Does it panic, or merely scream?

We sometimes ask end-users to enable `debug` in production to diagnose issues, so it better not panic :-).

aanm:
> What happens if mutations are detected? Does it panic, or merely scream?
> 
> We sometimes ask end-users to enable `debug` in production to diagnose issues, so it better not panic :-).

Yes, it panics that's why we shouldn't tell users to enable it :sweat_smile:

aanm:
@viktor-kurchenko @squeed can you take a look? Thank you

.github/actions/helm-default/action.yaml
```diff
@@ -29,7 +29,13 @@ runs:
         echo sha=${SHA} >> $GITHUB_OUTPUT
 
         CILIUM_INSTALL_DEFAULTS="--chart-directory=${{ inputs.chart-dir }} \
-          --helm-set=image.repository=quay.io/${{ env.QUAY_ORGANIZATION_DEV }}/cilium-ci \
+          --helm-set=debug.enabled=true \
+          --helm-set=bpf.monitorAggregation=none"
+
+        # only add SHA to the image tags if it was set
+        if [ -n "${SHA}" ]; then
+          echo sha=${SHA} >> $GITHUB_OUTPUT
```
viktor-kurchenko, line 36: Looks like this is redundant, isn't it?

viktor-kurchenko (approved):
LGTM.

.github/actions/helm-default/action.yaml
aanm: oh nice catch. I forgot to delete line 29

pkg/k8s/network_policy.go
```diff
@@ -61,41 +61,41 @@ func parseNetworkPolicyPeer(namespace string, peer *slim_networkingv1.NetworkPol
 	var retSel *api.EndpointSelector
 
 	if peer.NamespaceSelector != nil {
-		labelSelector := peer.NamespaceSelector
-		matchLabels := map[string]string{}
+		namespaceSelector := &slim_metav1.LabelSelector{
+			MatchLabels: map[string]slim_metav1.MatchLabelsValue{},
```
squeed: nit: since we know the size of this map, let's preallocate it in a `make(...)`

pkg/k8s/network_policy.go
```diff
@@ -254,6 +251,33 @@ func ParseNetworkPolicy(np *slim_networkingv1.NetworkPolicy) (api.Rules, error)
 	return api.Rules{rule}, nil
 }
 
+func parsePodSelector(podSelectorIn *slim_metav1.LabelSelector, namespace string) *slim_metav1.LabelSelector {
+	podSelector := &slim_metav1.LabelSelector{
+		MatchLabels: map[string]slim_metav1.MatchLabelsValue{},
```
squeed: maybe preallocate here too

squeed (approved):
Two optional suggestions, otherwise looks good!

PR #28183, opened 2023-09-15
bpf, ctmap: Implement map pressure metric for CT maps

### LINKED ISSUE #20069: CFP: Extend bpf_map_pressure metric to cover connection tracking, NAT maps

christarazi:
- go.mod: Bump github/cilium/ebpf for BatchCursor support
- maps/ctmap: Reduce code duplication to open CT maps
- bpf, ctmap: Implement map pressure metric for CT maps
- maps/ctmap: Add benchmark for batch lookups

---

Main commit for convenience:

> bpf, ctmap: Implement map pressure metric for CT maps

> This commit adds the ability to publish the CT map pressure via
> cilium_bpf_map_pressure metric. Only global CT maps are supported. Per
> endpoint CT maps (local maps) support for now is deferred due to rare
> use.
> 

> This commit adds the ability to publish the CT map pressure via
> cilium_bpf_map_pressure metric. Only global CT maps are supported. Per
> endpoint CT maps (local maps) support for now is deferred due to rare
> use.
> 
> It does this by counting the number of elements in the CT maps via batch
> map lookup, which is far more efficient than doing an element-by-element
> lookup. The counting is done via a controller that runs on a
> fixed-interval.
> 
> In order to prevent large memory allocations suddenly, the CT map dump
> is chunked into 4096 buffers. This effectively amortizes the number of
> needed syscalls to C, being the number of chunks, rather than N, being
> the number of elements in the map.
> 
> If elements are deleted / added during chunking, it does not affect hash
> maps, such as the CT map. Other map types which rely on
> `generic_map_lookup_batch()` will need to consider an implementation
> similar to `DumpReliablyWithCallback()`. See [1].
> 
> Benchmark results based on chunk size:
> 
> 1) 4096 (old) vs 2048 (new)
> 
> ```
> $ benchstat old.txt new.txt
> name                old time/op    new time/op    delta
> _MapBatchLookup-16     145ms ± 2%     148ms ± 6%     ~     (p=0.247 n=10+10)
> 
> name                old alloc/op   new alloc/op   delta
> _MapBatchLookup-16    37.0MB ± 0%    36.9MB ± 0%   -0.31%  (p=0.000 n=9+9)
> 
> name                old allocs/op  new allocs/op  delta
> _MapBatchLookup-16     3.20k ± 0%     6.40k ± 0%  +99.88%  (p=0.000 n=9+9)
> ```
> 
> 2) 4096 vs 8192
> 
> ```
> $ benchstat old.txt new.txt
> name                old time/op    new time/op    delta
> _MapBatchLookup-16     145ms ± 3%     145ms ± 3%     ~     (p=0.684 n=10+10)
> 
> name                old alloc/op   new alloc/op   delta
> _MapBatchLookup-16    37.0MB ± 0%    37.3MB ± 0%   +0.74%  (p=0.000 n=9+10)
> 
> name                old allocs/op  new allocs/op  delta
> _MapBatchLookup-16     3.20k ± 0%     1.60k ± 0%  -49.94%  (p=0.000 n=9+10)
> ```
> 
> In terms of overall performance overhead with a 30s interval for the
> controller, the difference in CPU usage for a full CT map is +8.5% in
> idle usage over a 24h period. Note, this percentage is not the same as
> the canonical CPU usage percentage that is observed in tools like `top`,
> for example. Constantly dumping the full CT map every 30s to calculate
> and update the pressure metric required an extra 0.0017 seconds per
> second according to the `container_cpu_usage_seconds_total` metric.
> 
> This was calculated by having a two node cluster where one node had a
> full CT map and the other "empty". The raw data for the above metric
> were:
>   - Full CT map node:    0.0207 seconds per second
>   - "Empty" CT map node: 0.0190 seconds per second

Fixes: https://github.com/cilium/cilium/issues/20069


pkg/maps/ctmap/ctmap.go
```diff
@@ -264,6 +268,46 @@ func (m *Map) DumpEntries() (string, error) {
 	return DoDumpEntries(m)
 }
 
+// DumpAll batch dumps the Map m and returns the count of the entries.
+func (m *Map) DumpAll() (int, error) {
+	// TODO: Preallocating like this can put quite a bit of pressure on Go GC
+	// presumably, could we do something about that?
+	mx := m.MaxEntries()
+	vout := make([]CtEntry, mx)
```
christarazi: @ti-mo I recall you were looking into batch lookups for CT map. Do you have any thoughts about the above TODO?

pkg/maps/ctmap/ctmap.go
```diff
@@ -831,3 +875,45 @@ func calculateInterval(prevInterval time.Duration, maxDeleteRatio float64) (inte
...
+
+				m.UpdatePressureMetricWithSize(int32(l))
+			}
+			return nil
+		},
+		// TODO: Figure out what the right interval.
+		RunInterval: 30 * time.Second,
+		// TODO: Figure out how to plumb context through or at least how to
+		// stop this controller from hive.
+		Context: ctx,
```
christarazi: I'm having trouble figuring out the hive semantics to plumb through either the `Daemon` context to the ctmap GC cell or to set up the stop hooks. Help appreciated, cc @cilium/sig-foundations.

pkg/maps/ctmap/gc/gc.go
```diff
@@ -37,7 +38,8 @@ type PerClusterCTMapsRetriever func() []*ctmap.Map
 type parameters struct {
 	cell.In
 
-	Logger logrus.FieldLogger
+	Context context.Context
```
bimmlerd: I don't think this is a good pattern in hive. The `cell.In` marker (see [docs](https://docs.cilium.io/en/stable/contributing/development/hive/#provide)) marks this struct as the dependencies to be injected by hive. Contrary to a logger or the k8s clientset (https://docs.cilium.io/en/stable/contributing/development/hive/#cell-showcase), we don't have a global context passed in by hive. Instead, use the Start and Stop hooks that hive provides, via https://docs.cilium.io/en/stable/contributing/development/hive/#lifecycle.

pkg/maps/ctmap/ctmap.go
joamaki: The `pkg/hive/job` package was added for exactly these type of situations. It doesn't yet have the visibility via `ciliumctl status` though, but we do want to bridge the gaps with `pkg/controller` and eventually migrate to `pkg/hive/job` as it plays much nicer with the lifecycles and supports other types of jobs.

Here's how this would look like with the `job` package:
```
  func New(..., lc hive.Lifecycle, jobs job.Registry) *GC {
    gc := &GC{...}
    // ...
    // Create a new group of jobs that are started and stopped together.
    group := jobs.NewGroup()

    // Add a timer job that calls calculatePressureMetrics every 30 seconds.
    group.Add(job.Timer(
      "ct-map-pressure", 
       gc.calculatePressureMetrics, // func(context.Context) error
       30 * time.Second, /* TODO: or rather cfg.CTMapPressureInterval? */
    ))

    // Append the job group to the lifecycle.
    lc.Append(group) 

    return gc
  }

  // -----
  // or if map pressure metrics is independent of GC:
  // (preferably to decouple when possible rather than build mega structs with lots of
  //  functionality)
  
  func registerMapPressureMetrics(lc hive.Lifecycle, jobs job.Registry, other Dependency) {
    // same as above...
    lc.Append(group)  
  }

  var Cell = cell.Module(
    ...   
    cell.Invoke(registerMapPressureMetrics),
  )
```

The job library takes care of the context and cancels it when the application is stopped.

If you'd like to do this with pkg/controller, then you'd need to manage the "module's" context yourself,
e.g. by having for example `func (*GC) Start(hive.HookContext) error` that sets `gc.ctx` and `gc.cancel` and then starts the controllers, and then `func (*GC) Stop(hive.HookContext) error` that calls `gc.cancel` and `RemoveAllAndWait`. In `New() *GC` you'd do `lifecycle.Append(gc)` to add the hooks.

The reason why hive doesn't have an application-wide context is that it's not enough by itself. We also need to wait for things to stop and cleanup (important for tests for example) and further it's often important to keep order when stopping things.

christarazi: Thanks for the context. I've opted for `pkg/controller`, but by using `cancel()` and `RemoveAllAndWait()` to stop and tear down. Credit to @pippolo84 for helping me out last week.

pkg/maps/ctmap/gc/gc.go
christarazi: :+1:, I've gotten ridden of it. See https://github.com/cilium/cilium/pull/28183#discussion_r1339416931.

pkg/maps/ctmap/ctmap.go
ti-mo: GC pressure isn't really a problem since CtEntry doesn't contain any pointers, making this only a single alloc.

The problem is memory consumption. Assuming the user has CT maps sized to 1 million entries (not unheard of with today's node sizes), a single dump could consume up to `CtEntry` (56) + `CtKey6Global` (38) * 1M = 94MB of memory. This will lead to a noticeable increase in steady-state memory occupation of the agent.

If we decide to go this route, we should choose smaller batch sizes to reduce peak memory consumption, but then we'll run into the same problem that `DumpReliablyWithCallback` aims to solve. We'll need to account for map entries coming and going during the iteration. I haven't tested how the (kernel-side) batch lookup implementation deals with this, we need to check if it's already 'reliable' by design.

[Some slides from LPC](https://lpc.events/event/4/contributions/488/attachments/232/408/LPC_2019_MAP_BATCHING.pdf) call this out explicitly (`Avoid the issue of current bpf_map_get_next_key() when the prev_key is not in the hashmap`), so I assume this was kept in mind when designing this.

pkg/maps/ctmap/ctmap.go
```diff
@@ -264,6 +268,46 @@ func (m *Map) DumpEntries() (string, error) {
 	return DoDumpEntries(m)
 }
 
+// DumpAll batch dumps the Map m and returns the count of the entries.
+func (m *Map) DumpAll() (int, error) {
```
ti-mo: I would name this `Count()`. The fact that it dumps internally doesn't matter to the caller.

pkg/bpf/map_linux.go
```diff
@@ -735,6 +735,12 @@ func (m *Map) Dump(hash map[string][]string) error {
 	return nil
 }
 
+// DumpBatchAll returns the count of elements in the map by dumping the map
+// using batch lookup.
+func (m *Map) DumpBatchAll(nextKeyOut, keysOut, valuesOut interface{}) (int, error) {
```
ti-mo: I'd expose `BatchLookup` verbatim here. `All` may be a bit redundant with `Dump`.

cilium/cmd/bpf_ct_list.go
```diff
@@ -34,13 +34,8 @@ var (
...
 				return
 			}
-			maps := getMaps(t, id)
-			ctMaps := make([]interface{}, len(maps))
-			for i, m := range maps {
-				ctMaps[i] = m
-			}
 			common.RequireRootPrivilege("cilium bpf ct list")
-			dumpCt(ctMaps, t)
+			dumpCt(mapsToInterfaceSlice(getMaps(t, id)), t)
```
ti-mo: Can't we fix up `dumpCt` a bit to make it typed? This empty interface business isn't ideal.

ti-mo (changes requested):

pkg/maps/ctmap/ctmap.go
christarazi: Good call out! I've implemented something similar to `DumpReliablyWithCallback` and wrote a unit test to confirm it.

pkg/maps/ctmap/ctmap.go
```diff
@@ -840,3 +931,54 @@ func calculateInterval(prevInterval time.Duration, maxDeleteRatio float64) (inte
...
+						scopedLog.Warn(msg)
+					}
+					continue
+				}
+				defer m.Close()
+
+				// The alternative is to use seq_ops from the kernel to dump
+				// the map. seq_file can be a buffer allocated from userspace
+				// which the kernel can write to. It's used to dump debug info
+				// mostly. But it requires the map to support BTF. No Cilium
```
squeed: Do we have an issue for BTF support? Should we?

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -33,6 +40,25 @@ func (k *CTMapPrivilegedTestSuite) SetUpSuite(c *C) {
...
+	m := newMap(MapNameTCP4Global+"_test", mapTypeIPv4TCPGlobal)
+	err := m.OpenOrCreate()
+	defer m.Map.Unpin()
+	assert.NoError(b, err)
+
+	_ = populateFakeDataCTMap4(m, option.CTMapEntriesGlobalTCPDefault)
+
+	b.ReportAllocs()
+	b.StopTimer()
+	b.ResetTimer()
```
squeed, line 52: nit: I _think_ you don't need the Stop / Start calls; ResetTimer() should be sufficient.

pkg/maps/ctmap/ctmap.go
```diff
@@ -238,6 +243,92 @@ func (m *Map) DumpEntries() (string, error) {
...
+	case !global && !v4:
+		return countBatch(m, nil, new(CtKey6))
+	}
+	return
+}
+
+func countBatch[T any](m *Map, prev interface{}, k *T) (count int, err error) {
+	mx := m.MaxEntries()
+	const chunk uint32 = 4096
+	chSize := int(mx / chunk)
```
squeed: Assuming this is floor division, will you have an off-by-one error here for the last chunk?

squeed (approved):
A few very minor things, but LGTM!

pkg/maps/ctmap/ctmap.go
squeed: Yes, this might be an issue:
```
	var mx uint32 = 15
	const chunk uint32 = 4096
	chSize := int(mx / chunk)
	fmt.Println(chSize)
```

prints "0"

christarazi: Added 

```go
mx := m.MaxEntries()
const chunk uint32 = 4096
chSize := int(mx / chunk)
if chSize == 0 {
	chSize = 1
}
```

pkg/maps/ctmap/ctmap.go
christarazi: Filed https://github.com/cilium/cilium/issues/29216. Empty now while I gather info, feel free to add if you have context already.

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -836,3 +860,111 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+	assert.NoError(t, err)
+	defer assert.NoError(t, m.Map.Unpin())
+
+	cache := populateFakeDataCTMap4(m, size)
+	initial := len(cache)
+
+	done := make(chan struct{})
+	toDelete := 4096 << 2
+	failedToDelete := 0
+	go func() {
```
nebril: This test smells a bit flaky to me.

There is no guarantee that `m.Delete` and `m.Count` will actually run in parallel. Granted, I don't know how else would we test this behaviour, but it seems like this test won't always actually test parallel execution.

I don't think that this is a blocking comment, but I would think whether this test is necessary to stay in the PR.

nebril (approved):

pkg/maps/ctmap/ctmap_privileged_test.go
christarazi: Good call out. I've added a channel to ensure that we don't start making assertions until the deletion code has started.

YutaroHayakawa (approved):
Sorry for the delay! LGTM 👍

cilium-dbg/cmd/bpf_ct_list.go
```diff
@@ -91,19 +86,23 @@ func parseArgs(args []string) (string, uint32, error) {
 	}
 }
 
-func getMaps(t string, id uint32) []*ctmap.Map {
+func getMaps(t string, id uint32) []ctmap.CtMap {
+	var m []*ctmap.Map
+	r := make([]ctmap.CtMap, 0)
```
ti-mo:
```suggestion
	var r []ctmap.CtMap
```

pkg/maps/ctmap/ctmap.go
```diff
@@ -238,6 +243,94 @@ func (m *Map) DumpEntries() (string, error) {
...
+		return countBatch(m, nil, new(CtKey4))
+	case !global && !v4:
+		return countBatch(m, nil, new(CtKey6))
+	}
+	return
+}
+
+func countBatch[T any](m *Map, prev interface{}, k *T) (count int, err error) {
+	mx := m.MaxEntries()
+	const chunk uint32 = 4096
```
ti-mo:
```suggestion
	const chunkSize uint32 = 4096
```

pkg/maps/ctmap/ctmap.go
```diff
@@ -238,6 +243,94 @@ func (m *Map) DumpEntries() (string, error) {
...
+	chSize := int(mx / chunk)
```
ti-mo:
```suggestion
	chunks := int(mx / chunk)
```

pkg/maps/ctmap/ctmap.go
ti-mo: Thanks for bringing it up, I've fleshed out https://github.com/cilium/cilium/issues/29216 and put it on our internal tracker.

> But it requires the map to support BTF.

I'm not sure if mapKV BTF is strictly required for map iterators, need to look into this a bit. A bigger limitation is that bpf iterators only landed in kernel 5.8. Batch lookups landed in 5.6, so we could've reasonably gone with iterators instead, they're just much more involved to set up.

I'm also not sure if iterators would be faster, they still require a `read()` for every entry, and it's the time spent in syscalls that really makes dumping large maps so expensive. With batch lookups, we trade memory for speed.

Not sure how seq_file is relevant here, the kernel version requirement is really the only thing I would mention in this comment, and that it could be migrated to use iterators or some other map occupancy kernel API if those ever become a thing..

> No Cilium map has BTF support as of November 2023.

This is not true, only agent-created maps lack BTF info.

pkg/maps/ctmap/ctmap.go
```diff
@@ -840,3 +933,54 @@ func calculateInterval(prevInterval time.Duration, maxDeleteRatio float64) (inte
...
+				defer m.Close()
+
+				// The alternative is to use seq_ops from the kernel to dump
+				// the map. seq_file can be a buffer allocated from userspace
+				// which the kernel can write to. It's used to dump debug info
+				// mostly. But it requires the map to support BTF. No Cilium
+				// map has BTF support as of November 2023.
+				l, err := m.Count()
+				switch {
+				case l > 0 && errors.Is(err, ebpf.ErrKeyNotExist):
```
ti-mo: Why is `ErrKeyNotExist` expected to be handled by callers of `Map.Count()`? Smells like a leaky abstraction to me.

pkg/maps/ctmap/ctmap.go
```diff
@@ -840,3 +933,54 @@ func calculateInterval(prevInterval time.Duration, maxDeleteRatio float64) (inte
...
+					// Kernel returns "not exist" when the batch lookup
+					// completes.
+				case errors.Is(err, ebpf.ErrNotSupported):
```
ti-mo: Taking into account my previous comment, this doesn't need to be a switch since there's an early return:
```go
if errors.Is(err, ebpf.ErrNotSupported) {
	// We don't have batch ops, so cancel context to kill this
	// controller.
	cancel(err)
	return err
}
if err != nil {
	errs = errors.Join(errs, fmt.Errorf("failed to dump CT map %v: %w", m.Name(), err))
}
```

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -33,6 +40,23 @@ func (k *CTMapPrivilegedTestSuite) SetUpSuite(c *C) {
 	c.Assert(err, IsNil)
 }
 
+func Benchmark_MapBatchLookup(b *testing.B) {
```
ti-mo:
```suggestion
func BenchmarkMapBatchLookup(b *testing.B) {
```

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -33,6 +40,23 @@ func (k *CTMapPrivilegedTestSuite) SetUpSuite(c *C) {
...
+	m := newMap(MapNameTCP4Global+"_test", mapTypeIPv4TCPGlobal)
+	err := m.OpenOrCreate()
+	defer m.Map.Unpin()
```
ti-mo: I'm planning on making pinning opt-in, could you call Unpin() here directly after OpenOrCreate? This acts as a reminder for me to remove all extraneous Unpin() calls from tests. Putting them behind a defer will make them fail silently. Thanks!

```suggestion
	assert.NoError(b, m.Map.Unpin())
```

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -836,3 +860,115 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+	assert.Equal(t, (initial-toDelete)+failedToDelete, batchCount)
+	assert.NoError(t, err)
+
+	var count int
+	assert.NoError(t, m.DumpWithCallback(func(_ bpf.MapKey, _ bpf.MapValue) { count++ }))
+	assert.Equal(t, count, batchCount)
+	assert.Equal(t, len(cache), batchCount)
+}
+
+func populateFakeDataCTMap4(m CtMap, size int) map[*CtKey4Global]struct{} {
```
ti-mo:
```suggestion
func populateFakeDataCTMap4(tb testing.TB, m CtMap, size int) map[*CtKey4Global]struct{} {
    tb.Helper()
```

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -836,3 +860,115 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+	}
+
+	cache := make(map[*CtKey4Global]struct{}, size)
+	for i := 0; i < size; i++ {
+		needGenerate := true
+		for needGenerate {
+			key := genKey()
+			if _, needGenerate = cache[key]; !needGenerate {
+				if err := m.Update(key, value); err != nil {
+					panic(err)
```
ti-mo:
```suggestion
					tb.Fatal(err)
```

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -836,3 +860,115 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(len(buf), Equals, 0)
 }
+
+// TestCount checks whether the CT map batch lookup dumps the count of the
+// entire map reliably, while the map is changing in the background.
+func (k *CTMapPrivilegedTestSuite) TestCount(c *C) {
```
ti-mo: No need to extend gocheck suites:
```suggestion
func TestCount(t *testing.T) {
    testutils.PrivilegedTest(t)
```

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -836,3 +860,115 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+	// test indeterministic and consequently cause flakes.
+	prev := option.Config.CTMapEntriesGlobalTCP
+	defer func() { option.Config.CTMapEntriesGlobalTCP = prev }()
+	option.Config.CTMapEntriesGlobalTCP = 4096 << 7
+	size := 4096 << 2
+
+	m := newMap(MapNameTCP4Global+"_test", mapTypeIPv4TCPGlobal)
+	err := m.OpenOrCreate()
+	assert.NoError(t, err)
+	defer assert.NoError(t, m.Map.Unpin())
```
ti-mo: See my other comment, this doesn't need to be deferred.

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -836,3 +860,115 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+	initial := len(cache)
+
+	start, done := make(chan struct{}), make(chan struct{})
+	toDelete := 4096 << 2
+	failedToDelete := 0
+	go func() {
+		c.Helper()
+
+		for i := 0; i < toDelete; i++ {
+			if i == 0 { // close the channel once at the first iteration.
```
ti-mo: Just run this right before entering the loop?

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -836,3 +860,115 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+func (k *CTMapPrivilegedTestSuite) TestCount(c *C) {
+	t := c.T
+
+	// Set the max size of the map explicitly so we can provide enough buffer
+	// for the LRU map to avoid eviction that makes the assertions within this
+	// test indeterministic and consequently cause flakes.
+	prev := option.Config.CTMapEntriesGlobalTCP
+	defer func() { option.Config.CTMapEntriesGlobalTCP = prev }()
+	option.Config.CTMapEntriesGlobalTCP = 4096 << 7
+	size := 4096 << 2
```
ti-mo: How was this value chosen? And why not write this as `2 << 14`, or simply 16384? If it's supposed to be a multiple of m.Count's batch size, use a multiplier (and a comment)..

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -836,3 +860,115 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+// entire map reliably, while the map is changing in the background.
+func (k *CTMapPrivilegedTestSuite) TestCount(c *C) {
+	t := c.T
+
+	// Set the max size of the map explicitly so we can provide enough buffer
+	// for the LRU map to avoid eviction that makes the assertions within this
+	// test indeterministic and consequently cause flakes.
+	prev := option.Config.CTMapEntriesGlobalTCP
+	defer func() { option.Config.CTMapEntriesGlobalTCP = prev }()
+	option.Config.CTMapEntriesGlobalTCP = 4096 << 7
```
ti-mo:
```suggestion
	option.Config.CTMapEntriesGlobalTCP = 524288
```

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -836,3 +860,115 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+	cache := populateFakeDataCTMap4(m, size)
+	initial := len(cache)
+
+	start, done := make(chan struct{}), make(chan struct{})
+	toDelete := 4096 << 2
+	failedToDelete := 0
+	go func() {
+		c.Helper()
+
+		for i := 0; i < toDelete; i++ {
```
ti-mo: This can just use `size`:
```suggestion
		for i := 0; i < size; i++ {
```

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -836,3 +860,115 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+			if i == 0 { // close the channel once at the first iteration.
+				close(start)
+			}
+			k := mapsexp.Keys(cache)[0]
```
ti-mo: I assume the intention is to pick a random key from `cache` in a loop? Why not use plain `range` over `cache`? Go map iteration is random-ish, and x/exp/maps does basically the same thing.. https://cs.opensource.google/go/x/exp/+/9a3e6036:maps/maps.go;l=10.

IMO we could even replace the outer for loop over `toDelete` with a `range cache`, that virtually amounts to the same thing and is much more readable.

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -836,3 +860,115 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+		RevNAT:           0,
+		TxFlagsSeen:      0x02,
+		RxFlagsSeen:      0x14,
+		SourceSecurityID: 40653,
+		LastTxReport:     15856,
+		LastRxReport:     15856,
+	}
+
+	cache := make(map[*CtKey4Global]struct{}, size)
+	for i := 0; i < size; i++ {
```
ti-mo: Instead of if/for statements 4 levels deep, why not:
```go
for i := 0; i < size; i++ {
  key := genKey()
  _, needGenerate = cache[key]
  if needGenerate {
    continue
  }
  ... m.Update() ...
}
```

Also, is the intention to populate the cache with n = `size` keys? Then we should call `genKey()` again if it generated a duplicate of an existing key. Right now, if `genKey()` returns a duplicate, there will be an 'empty' slot.

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -836,3 +860,115 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+	go func() {
+		c.Helper()
+
+		for i := 0; i < toDelete; i++ {
+			if i == 0 { // close the channel once at the first iteration.
+				close(start)
+			}
+			k := mapsexp.Keys(cache)[0]
+			if err := m.Delete(k); err != nil {
+				failedToDelete++
```
ti-mo: Failing deletes should never be ignored. Just make this fatal the test.

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -836,3 +860,115 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+				continue
+			}
+			delete(cache, k)
+		}
+		close(done)
+	}()
+
+	<-start
+out:
+	for {
```
ti-mo: What's the reason for calling Count() in a loop? Is the result ever expected to exceed `initial`?

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -836,3 +860,115 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+			break out
+		default:
+		}
+		batchCount, err := m.Count()
+		assert.LessOrEqual(t, batchCount, initial)
+		assert.NoError(t, err)
+	}
+
+	batchCount, err := m.Count()
+	assert.Equal(t, (initial-toDelete)+failedToDelete, batchCount)
```
ti-mo: This looks really contrived. What is this testing, exactly? Why not simply:
- count
- delete half of the map's entries
- count
- compare the results with `DumpWithCallback`?

pkg/maps/ctmap/ctmap_privileged_test.go
ti-mo: Even with synchronization added, this will always invite flakes. I'd suggest foregoing testing `Map.Count()` in favor of exercising `countBatch()` directly (it's only a thin wrapper anyways). When I saw `countBatch()` initially, I thought that was the reason it takes the `prev` and `k` parameters.

I suggest the following:

- populate the map
- call `countBatch(_, nil, _)`
- delete some key (the value assigned to prev?) that's expected to make the iteration reset
- call `countBatch(_, prev, _)`
- etc.

When testing like this, there is no need for adding concurrency to the mix. We can figure out a sequence of events that would make batch lookups reset. Then, recreate those circumstances to exercise the backtracking behaviour in `countBatch()`. This is why `DumpWithCallback()` and friends are not ideal, the caller lacks any sort of control over the dump's internal iterator.

ti-mo (dismissed):
I think this needs a bit of work, sorry. I can't comment on the core logic in `countBatch()` because I'm not yet aware if and how the kernel's batch lookup API deals with iterator resets if map entries are added or removed during iteration. The tests don't really demonstrate it either.

pkg/maps/ctmap/ctmap_privileged_test.go
christarazi: > Also, is the intention to populate the cache with n = size keys?

Yes, this is exactly why I had the code that I did and not the simplified version that you suggest because it doesn't take into account a duplicate returned from `genKey` without leaving an empty slot.

pkg/maps/ctmap/ctmap_privileged_test.go
christarazi: The reason for the apparent contriveness is explained in the godoc for the test:

```
// TestCount checks whether the CT map batch lookup dumps the count of the
// entire map reliably, while the map is changing in the background.
```

The idea is to test the map changing in the background while `Count` is called, to assert that it handles the same cases as `DumpReliablyWithCallback`.

pkg/maps/ctmap/ctmap_privileged_test.go
christarazi: No, it should never exceed `initial`. The reason for the loop is described in https://github.com/cilium/cilium/pull/28183#discussion_r1406761725.

pkg/maps/ctmap/ctmap.go
christarazi: Yep, this was a leftover from a previous iteration where when `Count` reached the end of the map, it returned whatever the underlying syscall returned. No longer the case now.

pkg/maps/ctmap/ctmap.go
christarazi: I've removed the comment because it's not really that relevant to the reader anymore.

Initially, I had it from a discussion with Lorenz where he proposed `seq_file` as an alternative to batch ops.

pkg/maps/ctmap/ctmap_privileged_test.go
christarazi: I spent the last day or so trying to simplify the test as you suggest and thereby the code as well. I learned (thanks to @aspsk) the batch lookup API does not accept map keys as input, i.e. `prev`. Rather, it's more like a cursor / handle type which must either be `nil` (to start the batching) or `next` which is returned by the batch API (`k` in this code). Anything else will trigger `ENOENT`. This means that we cannot use the batch API to return to us a portion of the map from an arbitrary point; it must start from the beginning. Therefore, `DumpReliablyWithCallback()`-like implementation and testing is not needed. Especially considering that the CT maps are hash maps, meaning they don't use `generic_map_lookup_batch()` under the hood, the [comment](https://github.com/cilium/cilium/pull/28183#discussion_r1347050889) pointing out the LPC slides about needing `DumpReliablyWithCallback()`-like functionality does not apply, according to @aspsk.

Ultimately, this is the reason why I exposed `Count()` without any params because the user should not need to set the params.

pkg/maps/ctmap/ctmap_privileged_test.go
christarazi: Left a comment.

pkg/maps/ctmap/ctmap_privileged_test.go
christarazi: See https://github.com/cilium/cilium/pull/28183#discussion_r1409956734.

pkg/maps/ctmap/ctmap_privileged_test.go
christarazi: Update: See https://github.com/cilium/cilium/pull/28183#discussion_r1409956734.

christarazi:
@lmb It's up to date with the current version of cilium/ebpf because I thought that it wasn't specifically necessary to wait for the new changes in https://github.com/cilium/ebpf/pull/1223. Ultimately, it's just an API improvement, but nothing fundamental to this PR, IMO.

pkg/maps/ctmap/ctmap.go
```diff
@@ -238,6 +243,88 @@ func (m *Map) DumpEntries() (string, error) {
...
+	case !global && v4:
+		return countBatch(m, nil, new(CtKey4))
+	case !global && !v4:
+		return countBatch(m, nil, new(CtKey6))
+	}
+	return
+}
+
+func countBatch[T any](m *Map, cursor interface{}, next *T) (count int, err error) {
+	mx := m.MaxEntries()
```
lmb: Based on our discussions both `prev` and `k` should be dropped here. If you don't want to bump the dep on `c/ebpf` we need to replicate the cursor allocation logic I'd say.

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -882,3 +906,100 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+
+	for _, k := range mapsexp.Keys(cache)[:size/4] {
+		k := k
+		if err := m.Delete(k); err != nil {
+			t.Fatal(err)
+		}
+		delete(cache, k)
+
+		batchCount, err := m.Count()
+		assert.Equal(t, len(cache), batchCount)
```
lmb, line 943: This assertion fails since `prevKey` is not nil here.

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -33,6 +40,23 @@ func (k *CTMapPrivilegedTestSuite) SetUpSuite(c *C) {
 	c.Assert(err, IsNil)
 }
 
+func BenchmarkMapBatchLookup(b *testing.B) {
```
lmb, line 43: I can't actually run this benchmark:

```
$ PRIVILEGED_TESTS=y go test -exec sudo -v -run XX -bench MapBatchLookup ./pkg/maps/ctmap/
goos: linux
goarch: amd64
pkg: github.com/cilium/cilium/pkg/maps/ctmap
cpu: 12th Gen Intel(R) Core(TM) i7-1260P
BenchmarkMapBatchLookup
    ctmap_privileged_test.go:55: 
        	Error Trace:	/home/lorenz/dev/cilium/pkg/maps/ctmap/ctmap_privileged_test.go:55
        	Error:      	Received unexpected error:
        	           	no space left on device
        	Test:       	BenchmarkMapBatchLookup
    ctmap_privileged_test.go:56: 
        	Error Trace:	/home/lorenz/dev/cilium/pkg/maps/ctmap/ctmap_privileged_test.go:56
        	Error:      	"2" is not greater than "262144"
        	Test:       	BenchmarkMapBatchLookup
--- FAIL: BenchmarkMapBatchLookup
```

pkg/maps/ctmap/ctmap_privileged_test.go
lmb: I think Timo forgot to update the loop condition: `for len(cache) < size`

pkg/maps/ctmap/ctmap_privileged_test.go
```diff
@@ -882,3 +906,100 @@ func (k *CTMapPrivilegedTestSuite) TestOrphanNatGC(c *C) {
...
+
+	cache := populateFakeDataCTMap4(t, m, size)
+	initial := len(cache)
+
+	batchCount, err := m.Count()
+	assert.Equal(t, initial, batchCount)
+	assert.NoError(t, err)
+
+	for _, k := range mapsexp.Keys(cache)[:size/4] {
+		k := k
```
lmb, line 936: Why not `m.Count()` here instead?

lmb (changes requested):

pkg/maps/ctmap/ctmap_privileged_test.go
christarazi: That's strange because I have no problems running it.

pkg/maps/ctmap/ctmap_privileged_test.go
christarazi: Ah, good point. Implemented.

pkg/maps/ctmap/ctmap_privileged_test.go
christarazi: Sorry, I'm not following. I don't see any failures in CI nor locally. What do you suggest to do here?

pkg/maps/ctmap/ctmap_privileged_test.go
christarazi: Could you elaborate? This code is iterating through the keys to delete.

pkg/maps/ctmap/ctmap.go
christarazi: I've bumped cilium/ebpf dependency to leverage the support for `BatchCursor`.

rolinh (approved):
vendor changes lgtm

pkg/maps/ctmap/ctmap_privileged_test.go
lmb: The comments refer to a previous version of your code. Not sure how GH screwed it up this time...

pkg/maps/ctmap/ctmap_privileged_test.go
lmb: Seems to work on top of the new code! That said the benchmark showed me that this is still allocating a ton, I think we might be doing something silly in the library. I'll have a look.

pkg/maps/ctmap/ctmap.go
```diff
@@ -238,6 +243,76 @@ func (m *Map) DumpEntries() (string, error) {
...
+	//   56.0
+	//   >>> (38*4096) / 1024 # CT IPv6 map key
+	//   152.0
+	//   >>> (56*4096) / 1024 # CT map value
+	//   224.0
+	kout := make([]T, chunkSize)
+	vout := make([]CtEntry, chunkSize)
+
+	var cursor ebpf.BatchCursor
+	for i := 0; i < chunks; i++ {
```
lmb: Why does this still take `chunks` into account? I think just checking for `ErrKeyNotExist` is enough.

lmb (changes requested):

pkg/maps/ctmap/ctmap_privileged_test.go
christarazi: I'm going to assume that there are no actionable steps that need to be done here, resolving.

pkg/maps/ctmap/ctmap_privileged_test.go
christarazi: Update: The API now uses the BatchCursor API from the ebpf lib.

lmb (approved):
LGTM. You might want to bump c/ebpf to include https://github.com/cilium/ebpf/commit/b89a4cb96b72d9866f80f0cbaaffd9538b1e7424

christarazi:
Bumped as suggested by Lorenz.

christarazi:
//test

PR #28184, opened 2023-09-15
Fix BPF Map Pressure for policy maps goes above 100%

tommyp1ckles:
endpoint: fix issue where bpf_map_pressure goes above 100% for policy

Endpoints that are being created/restored would have their
realizedPolicy and desiredPolicy to point at the same object in memory.

Under normal operations, this seems to not cause much issue since
everything that is intended to be in the desired policy state should
also end up in the realized policy state.

However, in the case where the bpf policy map fills up, the realized
policy map would appear to grow above its max size as the desired map
state would still be updated.

In such a case, this was noticeable with the bpf_map_pressure metric for
policy map types going above 100%.

This can be reproduced by setting bpf.policyMapMax=256 and creating a
Pod using fqdn policy and continuously performing DNS lookups from
within the Pod to domains that return many IPs (i.e. such as AWS
domains).
Eventually the policy map should fill up, continuing doing lookups
further will lead to unexpected bpf_map_pressure metrics for policy
maps.

**Note:** Submitting #28185 as a seperate PR to make backporting easier (this change will likely need to be backported, the former will not) as a series of changes to fix the map pressure metric for policy maps.

### DIFF
--- a/pkg/endpoint/endpoint.go
+++ b/pkg/endpoint/endpoint.go
@@ -528,6 +528,7 @@ func createEndpoint(owner regeneration.Owner, policyGetter policyRepoGetter, nam
 		status:           NewEndpointStatus(),
 		hasBPFProgram:    make(chan struct{}),
 		desiredPolicy:    policy.NewEndpointPolicy(policyGetter.GetPolicyRepository()),
+		realizedPolicy:   policy.NewEndpointPolicy(policyGetter.GetPolicyRepository()),
 		controllers:      controller.NewManager(),
 		regenFailedChan:  make(chan struct{}, 1),
 		allocator:        allocator,
@@ -540,8 +541,6 @@ func createEndpoint(owner regeneration.Owner, policyGetter policyRepoGetter, nam
 	ctx, cancel := context.WithCancel(context.Background())
 	ep.aliveCancel = cancel
 	ep.aliveCtx = ctx
-
-	ep.realizedPolicy = ep.desiredPolicy
 
 	ep.SetDefaultOpts(option.Config.Opts)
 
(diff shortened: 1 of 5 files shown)


pkg/endpoint/endpoint.go
```diff
@@ -516,7 +519,7 @@ func createEndpoint(owner regeneration.Owner, policyGetter policyRepoGetter, nam
 	ep.aliveCancel = cancel
 	ep.aliveCtx = ctx
 
-	ep.realizedPolicy = ep.desiredPolicy
+	ep.realizedPolicy = ep.desiredPolicy.DeepCopy()
```
christarazi: While I understand this is probably technically the most correct thing to do, a deep-copy is quite expensive, especially in the hot path. Could we explore a cheaper alternative to exceeding 100%? Maybe during the metrics-update, we check whether the current size exceeds the max, and then truncate?

My guess is that historically we have may have avoided deep copies due to the performance implications.

cc @jrajahalme @joestringer for background.

tommyp1ckles: I was a bit surprised to find this, but from what I can tell realizedPolicy and desiredPolicy always point at the same thing.  As a result, realizedPolicy is basically functioning the same as desiredPolicy - perhaps we could just get rid of realized policy?

But I'm missing some context here so I'd like to see what Joe or Jarno think.

tommyp1ckles:
Converting to draft for now, going to explore other options to DeepCopy'ing the entire map.

tommyp1ckles:
Out on PTO, will return to this once I'm back - setting it to draft for now.

nathanjsweet:
Is this unexpected? bpf_map_pressure is the total pressure applied to an existing BPF map. The number is effectively a measure of how big the given bpf map needs to be to accommodate the total desired entires. Capping the number to 100 destroys valuable information to the operator, who can use the number to resize the map to a larger size while they troubleshoot the issue.

aditighag (commented):
There are issues highlighted by the go linter. Do you want to put this into draft in the meantime?

thorn3r:
cleanup: removing backport labels and GH projects

PR #28185, opened 2023-09-16
endpointmanager: fix bpf policy pressure getting stuck.

tommyp1ckles:
Currently the policy map pressure metric only updates the map pressure metric when a new pressure value that is higher than the current one is set. This means that the metric can only ever go up, so when maps are shrunk (ex. such as after doing an cilium fqdn cache clean) the metric never goes down.

This changes the behaviour of the metric to maintain a map of map pressure values. When the trigger is invoked, it iterates all values and finds the max - updating the map_pressure gauge for policymaps to the max value. Endpoints that are shut down have their values removed.

CC: @christarazi 

**Note:** This is part of a two part fix for endpoint policy map pressure, for backport purposes the other change was split into a seperate PR: #28184

### DIFF
--- a/pkg/endpoint/endpoint.go
+++ b/pkg/endpoint/endpoint.go
@@ -404,8 +404,15 @@ func (e *Endpoint) InitEndpointScope(parent cell.Scope) {
 }
 
 func (e *Endpoint) Close() {
+	e.mutex.Lock()
+	defer e.mutex.Unlock()
+
 	if e.closeHealthReporter != nil {
 		e.closeHealthReporter()
+	}
+
+	if e.PolicyMapPressureUpdater != nil {
+		e.PolicyMapPressureUpdater.Remove(e.ID)
 	}
 }
 
--- a/pkg/endpointmanager/policymap_pressure.go
+++ b/pkg/endpointmanager/policymap_pressure.go
@@ -4,9 +4,10 @@
 package endpointmanager
 
 import (
-	"sync/atomic"
+	"golang.org/x/exp/maps"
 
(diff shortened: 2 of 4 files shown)


pkg/endpoint/endpoint.go
```diff
@@ -384,6 +384,13 @@ type Endpoint struct {
 	ciliumEndpointUID k8sTypes.UID
 }
 
+func (e *Endpoint) CloseMetrics() {
```
christarazi: Where is this called?

christarazi (changes requested):
Thanks for the fix! Just one comment.

pkg/endpointmanager/policymap_pressure.go
```diff
@@ -4,42 +4,56 @@
...
-	"sync/atomic"
 	"time"
 
 	"github.com/cilium/cilium/pkg/endpoint"
+	"github.com/cilium/cilium/pkg/lock"
 	"github.com/cilium/cilium/pkg/logging/logfields"
 	"github.com/cilium/cilium/pkg/maps/policymap"
 	"github.com/cilium/cilium/pkg/metrics"
 	"github.com/cilium/cilium/pkg/trigger"
+	"golang.org/x/exp/maps"
```
christarazi: Nit: This 3rd party import should be in the section before the Cilium imports

christarazi (approved):

christarazi:
Let's get this PR over the finish line @tommyp1ckles, it's basically there. Just some Go linter errors to fix up.

tommyp1ckles:
@christarazi woops yeah forgot about this one, let me get it ready

PR #28194, opened 2023-09-18
Add proxy protocol support for ingress controller

### LINKED ISSUE #21438: CFP: Support PROXY Protocol from external LoadBalancers
## Cilium Feature Proposal

Some Load Balancing solutions, such as Hetzner's LoadBalancer and AWS' Classic Load Balancer are using the [HAProxy Proxy Protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) to pass source IP information along. If you use those Load Balancers with a Cilium-enabled Kubernetes Cluster you currently can not whitelist certain IP ranges, because cilium only sees all connections coming from the Load Balancer. 

Cilium parses the HAProxy Headers and replace original source IP information with those in the header.

1. Add global options to:
    - Enable PROXY Protocol support
    - Set LoadBalancer Trusted IP range
2. Check if PROXY support is enabled and the original source IP comes from the trusted proxy IP range.
    - If not -> use original IP, do not parse anything
    - If both true > Try to parse the PROXY header, replace original source ip with parsed source ip. Fallback to original IP if parsing fails.

**More Information**
- https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt
- https://developers.cloudflare.com/spectrum/how-to/enable-proxy-protocol/#the-proxy-protocol-v2-header
- https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/
- https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/listener/proxy_protocol/v3/proxy_protocol.proto

zetaab:
Adds support for proxy protocol in listeners using https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/listener_filters/proxy_protocol

Fixes: #21438

```release-note
add Ingress controller proxy protocol support
```

### DIFF
--- a/install/kubernetes/cilium/values.yaml
+++ b/install/kubernetes/cilium/values.yaml
@@ -721,7 +721,7 @@ ingressController:
   # Incoming traffic to http listener will return 308 http error code with respective location in header.
   enforceHttps: true
 
-  # -- Enable proxy protocol for the listeners.
+  # -- Enable proxy protocol for all Ingress listeners. Note that _only_ Proxy protocol traffic will be accepted once this is enabled.
   enableProxyProtocol: false
 
   # -- IngressLBAnnotations are the annotation prefixes, which are used to filter annotations to propagate
--- a/install/kubernetes/cilium/values.yaml
+++ b/install/kubernetes/cilium/values.yaml
@@ -721,6 +721,9 @@ ingressController:
   # Incoming traffic to http listener will return 308 http error code with respective location in header.
   enforceHttps: true
 
+  # -- Enable proxy protocol for the listeners.
+  enableProxyProtocol: false
+
   # -- IngressLBAnnotations are the annotation prefixes, which are used to filter annotations to propagate
   # from Ingress to the Load Balancer service
   ingressLBAnnotationPrefixes: ['service.beta.kubernetes.io', 'service.kubernetes.io', 'cloud.google.com']
--- a/operator/option/config.go
+++ b/operator/option/config.go
@@ -249,7 +249,7 @@ const (
 	// Incoming traffic to http listener will return 308 http error code with respective location in header.
 	EnforceIngressHttps = "enforce-ingress-https"
 
-	// EnableIngressProxyProtocol enable proxy protocol for the listeners.
+	// EnableIngressProxyProtocol enable proxy protocol for all Ingress listeners. Note that _only_ Proxy protocol traffic will be accepted once this is enabled.
 	EnableIngressProxyProtocol = "enable-ingress-proxy-protocol"
 
 	// EnableIngressSecretsSync enables fan-in TLS secrets from multiple namespaces to singular namespace (specified
--- a/operator/option/config.go
+++ b/operator/option/config.go
@@ -249,6 +249,9 @@ const (
 	// Incoming traffic to http listener will return 308 http error code with respective location in header.
 	EnforceIngressHttps = "enforce-ingress-https"
 
+	// EnableIngressProxyProtocol enable proxy protocol for the listeners.
+	EnableIngressProxyProtocol = "enable-ingress-proxy-protocol"
+
 	// EnableIngressSecretsSync enables fan-in TLS secrets from multiple namespaces to singular namespace (specified
 	// by ingress-secrets-namespace flag
 	EnableIngressSecretsSync = "enable-ingress-secrets-sync"
@@ -507,6 +510,9 @@ type OperatorConfig struct {
 	// EnforceIngressHTTPS enforces https if required
 	EnforceIngressHTTPS bool
 
+	// EnableIngressProxyProtocol uses proxy protocol in listeners
+	EnableIngressProxyProtocol bool
+
 	// EnableIngressSecretsSync enables background TLS secret sync for Ingress
 	EnableIngressSecretsSync bool
 
@@ -595,6 +601,7 @@ func (c *OperatorConfig) Populate(vp *viper.Viper) {
 	c.EnableIngressController = vp.GetBool(EnableIngressController)
 	c.EnableGatewayAPI = vp.GetBool(EnableGatewayAPI)
 	c.EnforceIngressHTTPS = vp.GetBool(EnforceIngressHttps)
+	c.EnableIngressProxyProtocol = vp.GetBool(EnableIngressProxyProtocol)
 	c.IngressSecretsNamespace = vp.GetString(IngressSecretsNamespace)
 	c.GatewayAPISecretsNamespace = vp.GetString(GatewayAPISecretsNamespace)
 	c.ProxyIdleTimeoutSeconds = vp.GetInt(ProxyIdleTimeoutSeconds)
--- a/operator/pkg/model/translation/envoy_listener_test.go
+++ b/operator/pkg/model/translation/envoy_listener_test.go
@@ -4,6 +4,7 @@
 package translation
 
 import (
+	"slices"
 	"sort"
 	"testing"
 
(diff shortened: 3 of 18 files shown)


zetaab:
tests should fail still, I could not find way to run tests in my arm m1 mac.

install/kubernetes/cilium/values.yaml
```diff
@@ -712,6 +712,9 @@ ingressController:
   # Incoming traffic to http listener will return 308 http error code with respective location in header.
   enforceHttps: true
 
+  # -- Enable proxy protocol for the listeners.
+  enableProxyProtocol: false
```
kaworu: Please report this change to `install/kubernetes/cilium/values.yaml.tmpl` and then run

```
make -C install/kubernetes cilium/values.yaml && make -C Documentation update-helm-values
```

kaworu (changes requested):
Hi @zetaab :wave: and thank you for the PR!

Small comment on the way, but other than that Helm changes LGTM.

install/kubernetes/cilium/values.yaml
zetaab: @kaworu done

zetaab:
that commit does not exist in the PR anymore which did not have signed-off. The bot is out of sync. No idea how to refresh that and remove label

rolinh (approved):
Vendor changes lgtm.

lmb:
@zetaab can you please rebase the PR? There are conflicting changes to the vendor directory unfortunately.

Maybe you could also extend the commit message?

zetaab:
@lmb done

operator/pkg/model/translation/envoy_listener_test.go
```diff
@@ -113,4 +126,18 @@ func TestNewSNIListener(t *testing.T) {
...
+		listener := &envoy_config_listener.Listener{}
+		err = proto.Unmarshal(res.Value, listener)
+		require.Nil(t, err)
+
+		require.Equal(t, "dummy-name", listener.Name)
+		require.Len(t, listener.GetListenerFilters(), 2)
+		require.Len(t, listener.GetFilterChains(), 1)
+		require.Len(t, listener.GetFilterChains()[0].FilterChainMatch.ServerNames, 2)
+	})
+}
```
kaworu: [Travis-CI build failed](https://app.travis-ci.com/github/cilium/cilium/builds/266440058) with:

```
Unformatted Go source code:

./operator/pkg/model/translation/envoy_listener_test.go

diff ./operator/pkg/model/translation/envoy_listener_test.go.orig ./operator/pkg/model/translation/envoy_listener_test.go

--- ./operator/pkg/model/translation/envoy_listener_test.go.orig

+++ ./operator/pkg/model/translation/envoy_listener_test.go

@@ -140,4 +140,4 @@

 		require.Len(t, listener.GetFilterChains(), 1)

 		require.Len(t, listener.GetFilterChains()[0].FilterChainMatch.ServerNames, 2)

 	})

-}

\ No newline at end of file

+}
```

please add an EOL at the end of the file.

kaworu (changes requested):
Thanks for the update @zetaab, one more formatting request on my side.

operator/pkg/model/translation/envoy_listener_test.go
zetaab: @kaworu done

joamaki (approved):
Values.yaml and operator config changes LGTM.

kaworu (approved):
Helm changes LGTM.

operator/pkg/model/translation/translator_test.go
```diff
@@ -188,6 +188,33 @@ func TestSharedIngressTranslator_getServices(t *testing.T) {
...
+				},
+			},
+		},
+	})
+	require.Len(t, res, 1)
+	listener := &envoy_config_listener.Listener{}
+	err := proto.Unmarshal(res[0].GetValue(), listener)
+	require.NoError(t, err)
+
+	require.Len(t, listener.ListenerFilters, 2)
```
youngnick: I'm worried that this is a little fragile; it doesn't actually check that the extra filter is the correct one. Is it possible to add something to check that the second ListenerFilter is the ProxyProtocol one?

operator/pkg/model/translation/envoy_listener_test.go
```diff
@@ -113,4 +126,18 @@ func TestNewSNIListener(t *testing.T) {
...
+	t.Run("normal SNI listener with Proxy Protocol", func(t *testing.T) {
+		res, err := NewSNIListener("dummy-name", map[string][]string{"dummy-namespace/dummy-service:443": {"example.org", "example.com"}}, WithProxyProtocol())
+		require.Nil(t, err)
+
+		listener := &envoy_config_listener.Listener{}
+		err = proto.Unmarshal(res.Value, listener)
+		require.Nil(t, err)
+
+		require.Equal(t, "dummy-name", listener.Name)
+		require.Len(t, listener.GetListenerFilters(), 2)
```
youngnick: I know the test just above does this, but it would be good to check that the second filter is a Proxy Protocol filter here - that will protect us in case we add other filters in the future.

operator/option/config.go
```diff
@@ -249,6 +249,9 @@ const (
 	// Incoming traffic to http listener will return 308 http error code with respective location in header.
 	EnforceIngressHttps = "enforce-ingress-https"
 
+	// EnableIngressProxyProtocol enable proxy protocol for the listeners.
```
youngnick:
```suggestion
	// EnableIngressProxyProtocol enable proxy protocol for all Ingress listeners. Note that _only_ Proxy protocol traffic will be accepted once this is enabled.
```
It's _really_ important to ensure that we call out that once you enable this, only Proxy protocol traffic will be accepted.

operator/pkg/model/translation/envoy_listener_test.go
```diff
@@ -48,6 +48,19 @@ func TestNewHTTPListener(t *testing.T) {
...
+	t.Run("without TLS with Proxy Protocol", func(t *testing.T) {
+		res, err := NewHTTPListener("dummy-name", "dummy-secret-namespace", nil, WithProxyProtocol())
+		require.Nil(t, err)
+
+		listener := &envoy_config_listener.Listener{}
+		err = proto.Unmarshal(res.Value, listener)
+		require.Nil(t, err)
+
+		require.Equal(t, "dummy-name", listener.Name)
+		require.Len(t, listener.GetListenerFilters(), 2)
```
youngnick: Same comment as other places with this - could you please check that the second filter is the correct Proxy protocol filter?

youngnick (changes requested):
A couple of smallish things about testing and clarifying that proxy protocol is required for all Listeners once you turn this on. Aside from that, this is looking pretty good, although I'd like to see what @sayboras thinks as well.

operator/option/config.go
zetaab: @youngnick you are right. However, at least I would like to see similar behaviour that Traefik has (which means that you can define ip ranges from where it will accept proxy traffic. Basically Traefik is supporting both protocols: proxy and normal tcp at the same time). But that is something that is not supported by envoyproxy, so no can do in this repository.

youngnick:
Also, please make sure you're doing the DCO signoff as per the instructions - we won't be able to accept the PR without all the commits having that.

zetaab:
there are two commits and both have that?

youngnick:
Hmm, weird, sorry about that.

youngnick (approved):
Great work and thanks for hitting all my comments @zetaab! This looks great now.

youngnick:
Tests are still running as I write this, but looks like the Travis run failed because the listeners are in the opposite order, and also looks like there's some Helm changes that are needed as well.

Once the CI is green, this should be good to go.

zetaab:
@youngnick should be fixed now

zetaab:
oh I rebased the PR because there is label. I was thinking that its added/removed automatically by bot, but its not.

pippolo84 (approved):
Thanks!

meyskens (approved):
LGTM

zetaab:
so there are 6 approvers but nothing happens? Because of those 2 failing tests or? I have no idea what is wrong with those

joestringer:
I suspect some of the testing issues may have been unreliable tests, it seems like re-running them got them to pass. I'm not exactly sure why it took this long for the maintainer's little helper to set the label, but now that I look all of the reviews are in and the required CI jobs are passing so this is good to :ship: .

Thanks for your contribution @zetaab!

PR #28202, opened 2023-09-18
endpoint/id: use strings.IndexByte

tklauser:
Use `strings.IndexByte` instead of `strings.Index` to look for a single byte. This improves `BenchmarkSplitID` as follows:

```
name       old time/op    new time/op    delta
SplitID-8    62.5ns ±61%    41.1ns ± 5%  -34.14%  (p=0.000 n=10+9)
```


tklauser:
@tommyp1ckles gentle ping for review

tommyp1ckles (approved):

tommyp1ckles:
Looks good

PR #28203, opened 2023-09-18
dep: Replace deprecated github.com/golang/protobuf

sayboras:
This commit is to replace deprecated package github.com/golang/protobuf by google.golang.org/protobuf as mentioned in upstream README.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -45,7 +45,6 @@ require (
 	github.com/go-openapi/validate v0.22.1
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da
-	github.com/golang/protobuf v1.5.3
 	github.com/google/go-cmp v0.5.9
 	github.com/google/gopacket v1.1.19
 	github.com/google/gops v0.3.27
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -175,6 +174,7 @@ require (
 	github.com/gobuffalo/flect v1.0.2 // indirect
 	github.com/godbus/dbus/v5 v5.1.0 // indirect
 	github.com/golang-jwt/jwt/v4 v4.5.0 // indirect
+	github.com/golang/protobuf v1.5.3 // indirect
```
aspsk, line 177: Sorry, I am not very experienced in Go. How does version number work here, is it auto-generated? I see that the https://github.com/protocolbuffers/protobuf-go/blob/master/go.mod requires `v1.5.0`

aspsk (approved):
LGTM

rolinh (approved):

go.mod
rolinh: Go uses minimal version selection. See notably [this Go ref](https://go.dev/ref/mod#minimal-version-selection) and [this blog post](https://research.swtch.com/vgo-mvs) by Russ Cox.

PR #28204, opened 2023-09-18
Use default pool when specified pool has no corresponding ip family

hargrovee:
In dual-stack mode, it's necessary to allocate both IPv4 and IPv6 addresses for pods. However, custom IP pools may only focus on either IPv4 or IPv6, which can result in the custom IP pool being configured with only IPv4 or IPv6. When specifying an IP pool for a namespace or pod through annotations, this can cause pod creation to fail because the specified IP pool lacks the IPv4 or IPv6 family. This commit allows for the missing IPv4 or IPv6 to be obtained from the default pool when the specified IP pool lacks either of them. 
For example, in dual-stack mode, if a pod specifies an IP pool named “ipv4-pool" that contains only IPv4 addresses, its IPv4 address will come from "ipv4-pool",  and its IPv6 address will come from the corresponding IPv6 pod CIDRs in the default pool.

### DIFF
--- a/daemon/k8s/resources.go
+++ b/daemon/k8s/resources.go
@@ -62,6 +62,7 @@ var (
 			k8s.CiliumNetworkPolicyResource,
 			k8s.CiliumClusterwideNetworkPolicyResource,
 			k8s.CiliumCIDRGroupResource,
+			k8s.CiliumPodIPPoolResource,
 		),
 	)
 )
@@ -92,4 +93,5 @@ type Resources struct {
 	CiliumNetworkPolicies            resource.Resource[*cilium_api_v2.CiliumNetworkPolicy]
 	CiliumClusterwideNetworkPolicies resource.Resource[*cilium_api_v2.CiliumClusterwideNetworkPolicy]
 	CIDRGroups                       resource.Resource[*cilium_api_v2alpha1.CiliumCIDRGroup]
+	CiliumPodIPPool                  resource.Resource[*cilium_api_v2alpha1.CiliumPodIPPool]
 }
--- a/pkg/ipam/allocator.go
+++ b/pkg/ipam/allocator.go
@@ -30,12 +30,12 @@ var (
 	ErrIPv6Disabled = errors.New("IPv6 allocation disabled")
 )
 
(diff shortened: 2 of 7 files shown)


gandro (changes requested):
Thanks for the PR! I've discussed this proposal with @tklauser. We think this is a very valid use-case, but we think that the fall-back on the `default` pool might not always be wanted. What we propose instead is to introduce IP-family specific annotations, i.e. `ipam.cilium.io/ipv4-pool` and `ipam.cilium.io/ipv6-pool`.

For example, it could then work as follows:

```yaml
# Uses `mars` for IPv4, uses `jupiter` for IPv6
ipam.cilium.io/ipv4-pool=mars,
ipam.cilium.io/ipv6-pool=jupiter
```


```yaml
# Uses `mars` for IPv4, falls back to `default` for IPv6
ipam.cilium.io/ipv4-pool=mars,
```

```yaml
# Uses `mars` for IPv4, falls back to `custom-default` for IPv6
ipam.cilium.io/ipv4-pool=mars
ipam.cilium.io/ip-pool=custom-default
```

```yaml
# Uses `mars` for IPv4, uses `jupiter` for IPv6 (fallback is ignored)
ipam.cilium.io/ipv4-pool=mars
ipam.cilium.io/ipv6-pool=jupiter
ipam.cilium.io/ip-pool=custom-default
```


What do you think about that? I think it should still cover your use-case, but also allow for more sophisticated combinations.

hargrovee:
Thank you for your feedback. Introducing annotations for specific IP-Family pools is indeed a better solution. This can address the issues I encountered and cover a wider range of scenarios. It's a great approach!

hargrovee:
I have submitted a new PR https://github.com/cilium/cilium/pull/28244that can address the use cases here. This PR will be closed shortly.

gandro:
Sounds good, thanks! I moved this PR here to draft so it does not clutter notifications.

hargrovee:
OK, thanks！

hargrovee:
PR #28244 has coverd this case here,  the draft will  closed now.

PR #28217, opened 2023-09-18
Add hubble_relay_pool_peer_connection_status metric

### LINKED ISSUE #27890: CFP: Hubble Relay Peer Metrics
## Cilium Feature Proposal

I'd like to see Prometheus metrics for Hubble Relay related to peer availability, so platform operators can alert on a broken peer connection.

siwiutki:

      GitHub issue.

This change adds a new gauge metric to hubble-relay measuring the connectiion status to all peers. Metric keeps track of number of peers for each possible connectiion status

The current set of metrics is not enough to accurately measure the availability of hubble-relay. They measure the status of gRPC calls, but, for instance, in case all peers are unreachable when GetFlows is called, even though gRPC call will succeed and return "OK" status, the response will come with no flows gathered, rendering it useless. This new metric is introduced to cover such cases.

Fixes #27890

```release-note
Added hubble_relay_pool_peer_connection_status metric for measuring the connection status of all peers. Metric keeps track of number of peers for each possible connectiion status.
```

### DIFF
--- a/pkg/hubble/relay/pool/manager.go
+++ b/pkg/hubble/relay/pool/manager.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc/connectivity"
 
@@ -36,28 +37,31 @@ type PeerManager struct {
 	stop    chan struct{}
 	mu      lock.RWMutex
 	peers   map[string]*peer
+	metrics *PoolMetrics
 }
 
 // NewPeerManager creates a new manager that connects to a peer gRPC service to
 // manage peers and a connection to every peer's gRPC API.
-func NewPeerManager(options ...Option) (*PeerManager, error) {
+func NewPeerManager(registry prometheus.Registerer, options ...Option) (*PeerManager, error) {
 	opts := defaultOptions
 	for _, opt := range options {
 		if err := opt(&opts); err != nil {
 			return nil, fmt.Errorf("failed to apply option: %v", err)
 		}
 	}
+	metrics := NewPoolMetrics(registry)
 	return &PeerManager{
 		peers:   make(map[string]*peer),
 		offline: make(chan string, 100),
 		stop:    make(chan struct{}),
 		opts:    opts,
+		metrics: metrics,
 	}, nil
 }
 
 // Start starts the manager.
 func (m *PeerManager) Start() {
-	m.wg.Add(2)
+	m.wg.Add(3)
 	go func() {
 		defer m.wg.Done()
 		m.watchNotifications()
@@ -65,6 +69,10 @@ func (m *PeerManager) Start() {
 	go func() {
 		defer m.wg.Done()
 		m.manageConnections()
+	}()
+	go func() {
+		defer m.wg.Done()
+		m.reportConnectionStatus()
 	}()
 }
 
@@ -182,6 +190,34 @@ func (m *PeerManager) manageConnections() {
 				}
 			}
 			m.mu.RUnlock()
+		}
+	}
+}
+
+func (m *PeerManager) reportConnectionStatus() {
+	connTimer, connTimerDone := inctimer.New()
+	defer connTimerDone()
+	for {
+		select {
+		case <-m.stop:
+			return
+		case <-connTimer.After(m.opts.connStatusInterval):
+			m.mu.RLock()
+			connStates := make(map[connectivity.State]uint32)
+			var nilConnPeersNum uint32 = 0
+			for _, p := range m.peers {
+				p.mu.Lock()
+				if p.conn == nil {
+					nilConnPeersNum++
+					p.mu.Unlock()
+					continue
+				}
+				state := p.conn.GetState()
+				connStates[state] = connStates[state] + 1
+				p.mu.Unlock()
+			}
+			m.mu.RUnlock()
+			m.metrics.ObservePeerConnectionStatus(connStates, nilConnPeersNum)
 		}
 	}
 }
--- a/pkg/hubble/relay/pool/manager_test.go
+++ b/pkg/hubble/relay/pool/manager_test.go
@@ -7,13 +7,17 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"sort"
+	"strings"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"google.golang.org/grpc"
@@ -672,6 +676,7 @@ func TestPeerManager(t *testing.T) {
 			once = sync.Once{}
 
 			mgr, err := NewPeerManager(
+				prometheus.NewPedanticRegistry(),
 				WithPeerClientBuilder(tt.pcBuilder),
 				WithClientConnBuilder(tt.ccBuilder),
 				WithConnCheckInterval(1*time.Second),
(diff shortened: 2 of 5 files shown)


marqc:
Thanks for contributing. Such a metric sounds pretty useful to me.

If I understand correctly, on each call to any relay api method you are updating the `hubble_relay_observer_unavailable_nodes` gauge. This seems quite odd that it's labeled with the API method name, as node status is not related to the calling method. This is a state of peers list.

I can suggest a different approach. Instead of binding this to API calls, create a single timer goroutine that will periodically traverse peers list and report gauge metric of `hubble_relay_observer_nodes` with a state label reflecting possible nodes states https://github.com/cilium/cilium/blob/main/api/v1/relay/relay.proto#L22

Benefits:
- metric will reflect the current relay peers states even when there are no API calls
- metric will reflect all nodes and their states. not only unavailable nodes (better for troubleshooting peer discovery problems)

pkg/hubble/relay/observer/server.go
```diff
@@ -356,3 +375,44 @@ func (s *Server) ServerStatus(ctx context.Context, req *observerpb.ServerStatusR
...
+func (s *Server) peerConnectivityStatus() map[relaypb.NodeState]uint32 {
+	var numConnectedPeers, numUnavailablePeers uint32
+	peers := s.peers.List()
+	for _, p := range peers {
+		if !isAvailable(p.Conn) {
+			numUnavailablePeers++
+			s.opts.log.WithField("address", p.Address).Infof(
+				"No connection to peer %s, skipping", p.Name,
+			)
+			s.peers.ReportOffline(p.Name)
```
marqc: This looks like changing some state in the peers view. As this is only used in observability/metrics code it should not affect any state and work as read-only code.

pkg/hubble/relay/observer/server.go
```diff
@@ -356,3 +375,44 @@ func (s *Server) ServerStatus(ctx context.Context, req *observerpb.ServerStatusR
...
+	}()
+}
+
+func (s *Server) peerConnectivityStatus() map[relaypb.NodeState]uint32 {
+	var numConnectedPeers, numUnavailablePeers uint32
+	peers := s.peers.List()
+	for _, p := range peers {
+		if !isAvailable(p.Conn) {
+			numUnavailablePeers++
+			s.opts.log.WithField("address", p.Address).Infof(
```
marqc: nit: If you just report metrics with peers status, i would not log unavailable nodes here (or change log level to debug).

pkg/hubble/relay/observer/server.go
```diff
@@ -356,3 +375,44 @@ func (s *Server) ServerStatus(ctx context.Context, req *observerpb.ServerStatusR
...
+			}
+		}
+	}()
+}
+
+func (s *Server) peerConnectivityStatus() map[relaypb.NodeState]uint32 {
+	var numConnectedPeers, numUnavailablePeers uint32
+	peers := s.peers.List()
+	for _, p := range peers {
+		if !isAvailable(p.Conn) {
```
marqc: the `isAvailable` categorizes grpc connection (between relay and cilium-agent) states into 2 categories. The "Ready" and "Idle" states are  categorized as available. All other states are categorized as unavailable.

When debugging peer connectivity issues I believe it can be more useful to report the actual grpc connection state as defined in https://github.com/cilium/cilium/blob/main/vendor/google.golang.org/grpc/connectivity/connectivity.go#L31

pkg/hubble/relay/observer/server.go
```diff
@@ -28,6 +30,8 @@ import (
 // Reporting up to 10 unavailable nodes is probably reasonable.
 const numUnavailableNodesReportMax = 10
 
+const heartbeatInterval = time.Minute
```
marqc: nit: This seems to be quite a lightweight code path to execute (check and report status). I believe it makes sense to call this method more often, i.e. every 5 seconds.

marqc (commented):
Overall a good job. Thanks for implementing this. I have one suggestion to report real grpc peer connection status instead of pre-aggregated as available/non-available.

pkg/hubble/relay/observer/server.go
siwiutki: With heartbeat run every 5 seconds, this might actually be spamming logs very often and this info will be available through metrics anyway. I'm rather with you on removing this log instead of changing to debug.

marqc:
> Overall a good job. Thanks for implementing this. I have one suggestion to report real grpc peer connection status instead of pre-aggregated as available/non-available.

One more suggestion. This metric reports the connection status between relay and peers. It is implemented as part of observer/Server but actually the Server has different responsibilities as it is rather responsible for handling client requests. It fits better to be implemented inside PeerManager https://github.com/cilium/cilium/blob/main/pkg/hubble/relay/pool/manager.go#L59

pkg/hubble/relay/observer/server.go
siwiutki: I agree with your other comment to move this to PeerManager. Since in that component there is already a periodic check for offline peers, I think we can remove it from reporting status as well.

pkg/hubble/relay/observer/server.go
siwiutki: Sure.

pkg/hubble/relay/observer/server.go
siwiutki: Sure.

glrf:
I think this PR would also close: #27890 (I actually just started looking into that issue before I discovered this PR)

One input I have: It might be worth it to add the peer name and/or address to the metric. This would make the implementation slightly more complicated and would increase the cardinality of the metric, but I think for most cluster sizes this should be fine and it would allow cluster operators to easily see from the metric itself which peer is affected without having to dig through logs.

marqc:
@glrf 
> One input I have: It might be worth it to add the peer name and/or address to the metric. This would make the implementation slightly more complicated and would increase the cardinality of the metric, but I think for most cluster sizes this should be fine and it would allow cluster operators to easily see from the metric itself which peer is affected without having to dig through logs.

I believe that the key is in "most cluster sizes". For relatively small clusters that's not a problem, but it may become a problem with larger clusters with thousand nodes (or with dynamic node allocation/cluster autoscalers). This also changes the feature from metric that allows detecting if anything is/was wrong into the detailed status history of each peer. If there's a decision to add a node name/address into labels set then this should be configurable to enable/disable these labels to match the user's cluster configuration. For a general use case I would rather stay just with gauge reporting number of peers in a given connection state and then relate to logs for details on unavailable peers.

rolinh:
Thanks for your contribution @siwiutki ! It sounds like a very useful feature indeed.
I also share @marqc opinion that the logic should be implemented in the peer manager, which is the component that manages connections to peers rather than in the server.

glrf:
> I believe that the key is in "most cluster sizes". For relatively small clusters that's not a problem, but it may become a problem with larger clusters with thousand nodes (or with dynamic node allocation/cluster autoscalers). This also changes the feature from metric that allows detecting if anything is/was wrong into the detailed status history of each peer. If there's a decision to add a node name/address into labels set then this should be configurable to enable/disable these labels to match the user's cluster configuration. For a general use case I would rather stay just with gauge reporting number of peers in a given connection state and then relate to logs for details on unavailable peers.

Fair point, maybe adding peer information is too much, feel free to just ignore my input. :) 

My impression was that if the number of nodes in your cluster will cause high cardinality in your metrics, you'll need to be very careful what and how you scrape metrics anyways. But you're right, this metric should also work in these cases without having to jump through hoops. So an aggregated gauge is probably good enough.

siwiutki (commented):
I agree that this code fits better into the PeerManager code. Moved it there already.

Unit tests that I wrote for PeerManager utilise `time.Sleep` to test the actual code that exports this metric. I don't fully like it, but I don't see time being mocked anywhere in the current unit tests, so I didn't try to reinvent the wheel here. Timer used in PeerManager could be mocked and usually https://github.com/benbjohnson/clock library is a good pick for that.

pkg/hubble/relay/pool/manager.go
```diff
@@ -186,6 +194,41 @@ func (m *PeerManager) manageConnections() {
...
+				// if does not exists, num is 0 (zero value)
+				// which happens to be a perfect initial value
+				num, _ := connStates[state]
+				num++
+				connStates[state] = num
+				p.mu.Unlock()
+			}
+			m.mu.RUnlock()
+
+			m.mu.Lock()
```
marqc: nit: this lock is unnecessary, as the `ObservePeerConnectionStatus` only does the `Set` method on `prometheus GaugeVec`, and prometheus client is already thread-safe.

marqc (approved):
LGTM! Great job! Thanks for contributing!

pkg/hubble/relay/pool/manager.go
siwiutki: The lock was definitely needed, because `ObservePeerConnectionStatus` was not thread-safe - even though setting Prometheus gauge for a particular label was thread-safe, this function was calling `Set` multiple times and so having states {"IDLE": 3, "READY": 1} and {"IDLE": 1, "READY": 3} being observed at the same time, you could end up with states {"IDLE": 3, "READY": 3} being displayed through gauge metric.

Now though I just made the `ObservePeerConnectionStatus` thread-safe itself, so that the lock used in metrics won't block other PeerManager operations. Useful nit to that, so thanks.

marqc: I like your solution with having a separate mutex for the metric gauge setter. Thanks.

pkg/hubble/relay/pool/manager.go
```diff
@@ -186,6 +194,38 @@ func (m *PeerManager) manageConnections() {
...
+					nilConnPeersNum++
+					p.mu.Unlock()
+					continue
+				}
+				state := p.conn.GetState()
+				// if does not exists, num is 0 (zero value)
+				// which happens to be a perfect initial value
+				num, _ := connStates[state]
+				num++
+				connStates[state] = num
```
kaworu: nit: IMO it's fine to "silently" rely on zero values:
```suggestion
				connStates[state] = connStates[state] + 1
```

pkg/hubble/relay/pool/manager_test.go
```diff
@@ -704,6 +709,176 @@ func TestPeerManager(t *testing.T) {
...
+				WithConnStatusInterval(2 * time.Second),
+				// set interval large enough not to fire in 3 seconds sleep
+				WithConnCheckInterval(20 * time.Minute),
+			}
+
+			mgr, err := NewPeerManager(registry, options...)
+			assert.NoError(t, err)
+			mgr.Start()
+			// sleep so that metrics will be emitted at least once
+			time.Sleep(3 * time.Second)
```
kaworu: I'm a little worried to rely on sleeping here, as it is a common source of flaky tests. Is there a way to loop until the metric has changed once instead?

pkg/hubble/relay/pool/manager_test.go
```diff
@@ -704,6 +709,176 @@ func TestPeerManager(t *testing.T) {
...
+					}
+				}(),
+			},
+			wantStatus: map[string]uint32{
+				"READY":             2,
+				"CONNECTING":        1,
+				"IDLE":              2,
+				"TRANSIENT_FAILURE": 1,
+				"SHUTDOWN":          1,
+				"NIL_CONNECTION":    0,
```
glrf, line 817: nit: It might be a good idea to add a test case that actually returns a `NIL_CONNECTION`.  Currently we could just remove all of the `nilConnNum` counting logic and the tests will still pass.

glrf (approved):

kaworu (commented):
Hi @siwiutki :wave: and thanks for the PR!

Couple of comment, but overall LGTM. Agree with @glrf that testing `NIL_CONNECTION` would be an improvement.

pkg/hubble/relay/pool/manager_test.go
siwiutki: Sure, added a test case for that.

pkg/hubble/relay/pool/manager_test.go
siwiutki: Since you're using testify framework for tests, `assert.Eventually` is IMO a clean approach to what you're suggesting. And I assume that loop would need a timeout anyway, so it fits very well here. Changed the test to use this instead of that pesky sleep.

pkg/hubble/relay/pool/manager.go
siwiutki: Sure, applied your suggestion.

kaworu (approved):
Thanks @siwiutki!

kaworu:
@siwiutki can you please squash the two commits into one please?

siwiutki (commented):
The second commit was a merge commit to be in sync with current main branch. Synced it with rebase instead of merge, so should be fine with 1 commit now.

siwiutki:
Seems like one of the checks (Go Related Checks / Precheck) is failing due to something unrelated to this PR. @kaworu could you take a look?

kaworu:
[Integration Tests failed with](https://github.com/cilium/cilium/actions/runs/6408021405/job/17396095195)

```
Unable to find image 'cilium/docs-builder:latest' locally
docker: Error response from daemon: manifest for cilium/docs-builder:latest not found: manifest unknown: manifest unknown.
```

re-running as it seems unrelated to the patch.

rolinh (approved):
Awesome work, lgtm!

Issue #28167, opened 2023-09-14
Cilium v1.14.x - Deployment Failure of Helm Chart on fresh Cluster


seb-lafond:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Hi,

Everyday we rebuild/redeploy one of our Kubernetes/Cilium clusters to check for any regressions.   We use kubeadm for Kubernetes stack and Helm chart for Cilium.  All of this is managed through Ansible.  

As we use the cronjob method to generate clustermeshapi certificates, _clustermesh-apiserver-generate-certs_ job is deployed. With the v1.13.x chart, the job is deployed and **not completed** but chart is **deployed.** This is because the pod don't have tolerations to run on the master node.  

Since v1.14.x, the _clustermesh-apiserver-generate-certs_ job is deployed and not completed as usual but the Helm deployment fails. We observed that Helm hooks _post-install,post-upgrade_ have been added to the job. [https://github.com/cilium/cilium/pull/23102](https://github.com/cilium/cilium/pull/23102)
Now, the job need to be completed for chart deployment to succeed.

### RECAP
**v1.13.x using Ansible:**
- kubeadm init on first master
- Helm install of Cilium chart
- _clustermesh-apiserver-generate-certs_ job is deployed but not completed.  Job pod can't run on master node, but the result is a success as there are no Helm hooks post-install/post-upgrade
- kubeadm join on secondary masters and so on for worker nodes

**v1.14.x using Ansible:**
- kubeadm init on first master
- Helm install of Cilium chart
- _clustermesh-apiserver-generate-certs_ job is deployed but not completed.  Job pod can't run on master node. Result is a failure cause Helm hooks post-install/post-upgrade are present.  The job needs to be completed to have chart deployed successfully
-  Ansible task **failed** ...

**Workarounds:**
Remove taints on master node, deploy chart, re-add taints on master node
or
Remove "helm.sh/hook": post-install,post-upgrade" annotation on job
or
Add tolerations on job template _install/kubernetes/cilium/templates/clustermesh-apiserver/tls-cronjob/_job-spec.tpl_ and in Helm values

No issue when upgrading from v1.13.x to v1.14.x as the job runs normally on one of the worker nodes

### Cilium Version

cilium-cli: v0.15.7 compiled with go1.21.0 on linux/amd64
cilium image (default): v1.14.1
cilium image (stable): v1.14.1
cilium image (running): 1.14.1

### Kernel Version

[...]

PR #28221, opened 2023-09-19
Make tolerations configurable in clustermesh-apiserver certgen job

giorio94:
A certgen job can be used to generate the TLS certificates required both for hubble and clustermesh. Yet, the clustermesh one lacks the possibility of configuring additional tolerations, which are instead configurable in the hubble case. This is possibly useful for instance to allow running the job on control-plane nodes. Hence, let's fix this divergence.

Related: https://github.com/cilium/cilium/issues/28167

```release-note
Make tolerations configurable in clustermesh-apiserver certgen job
```


sayboras (approved):

gandro:
Should we backport this to v1.14? Considering it seems to be a blocker for some users (ref https://github.com/cilium/cilium/issues/28167)

giorio94:
> Should we backport this to v1.14? Considering it seems to be a blocker for some users (ref #28167)

Makes sense to me, marking for backport to v1.14.

PR #28224, opened 2023-09-19
 pkg/labels: print all leaf CIDRs, not just the last one. 

squeed:
When printing Labels (e.g. by `cilium identity list`), CIDR labels are de-expanded to only show the relevant ones. For example, the CIDR label `1.1.1.0/24` is under-the-hood expanded to `0.0.0.0/0, 0.0.0.0/1, ..., 1.1.0.0/23, 1.1.1.0/24`. Then, when the identity is printed for human consumption, the CIDR labels are de-expanded. However, this code did not anticipate a single identity having multiple "leaf" CIDR labels.

So, change the printing logic to output all interesting CIDRs. This adds a O(N^2) check, but given that N is never over ~100, it shouldn't be an issue.

Note that this output is only for human consumption; it is not persisted or used as an API.

### DIFF
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -7,7 +7,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"net"
+	"net/netip"
 	"slices"
 	"sort"
 	"strings"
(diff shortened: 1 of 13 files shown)


squeed:
Confirmed that this works as expected:

```
root@kind-control-plane:/home/cilium# cilium identity list
ID         LABELS
1          cidr:10.244.0.80/32
           cidr:172.19.0.2/32
           cidr:fc00:c111::2/128
           reserved:host
           reserved:kube-apiserver
```

nathanjsweet (approved):

pkg/labels/labels.go
```diff
@@ -143,26 +143,40 @@ type Labels map[string]Label
...
+		} else {
+			res = append(res, v.String())
 		}
-		res = append(res, v.String())
 	}
-	if cidr != "" {
-		res = append(res, cidr)
+
+	// Remove all prefixes that contain a smaller prefix
+	// Yes, this is O(n^2) but N is ~50
```
joestringer: For one identity. I guess on the commandline we could be printing 1000s of identities? That said as long as it's ideally <100ms even with a large identity list then it's probably in the right order of magnitude for proper interaction with the user. Note this will be hit during sysdump though.

(Would be nice to have a benchmark to evaluate this too..)

squeed: Good idea. This is actually relevant to more than just `ciliumctl`, as it is also used for generating log messages.

I'll do a quick benchmark.

squeed:
Rebased, refactored to use a real data structure. This improves performance over the existing, incorrect method as well as my previous naive attempt:

```
     │ GetPrintableModelOrig │      GetPrintableModelLeafCIDR      │       GetPrintableModelNaive        │
     │        sec/op         │   sec/op     vs base                │   sec/op     vs base                │
*-12            167.0µ ± 17%   108.8µ ± 3%  -34.84% (p=0.000 n=10)   285.4µ ± 5%  +70.93% (p=0.000 n=10)

     │ GetPrintableModelOrig │     GetPrintableModelLeafCIDR      │       GetPrintableModelNaive       │
     │       allocs/op       │ allocs/op   vs base                │ allocs/op   vs base                │
*-12             1440.0 ± 0%   557.0 ± 0%  -61.32% (p=0.000 n=10)   189.0 ± 0%  -86.88% (p=0.000 n=10)

```

This reduces time from 167µs to 108µs. Fewer allocs, too.

pkg/labels/labels.go
squeed: Wrote a benchmark; decided to make it faster. Good idea.

squeed:
Code golf: got it even faster :-)

```
                     │   orig.txt    │               new.txt               │
                     │    sec/op     │   sec/op     vs base                │
GetPrintableModel-12   166.98µ ± 17%   88.62µ ± 3%  -46.93% (p=0.000 n=10)

                     │  orig.txt   │              new.txt               │
                     │  allocs/op  │ allocs/op   vs base                │
GetPrintableModel-12   1440.0 ± 0%   401.0 ± 0%  -72.15% (p=0.000 n=10)
```

joestringer (approved):
Needs a rebase but otherwise LGTM. Nice one :rocket:

joestringer:
We might consider adding `affects/1.14` and other versions for tracking purposes.

julianwiedmann:
Is anyone still planning to tackle the backports for `v1.13` / `v1.14`, or can we downgrade them to `affects/` ?

julianwiedmann:
Downgrading the v1.13 label to `affects` ...

PR #28225, opened 2023-09-19
install/kubernetes: add the `cilium/values.yaml` target to `.PHONY`

nbusseneau:
This target should completely ignore the existing contents of `cilium/values.yaml` since we always generate from the template files.

### DIFF
--- a/install/kubernetes/Makefile
+++ b/install/kubernetes/Makefile
@@ -75,4 +75,4 @@ docs:
 check-docker-images: # Check whether docker images are available for the current version.
 	$(QUIET)$(ROOT_DIR)/contrib/release/check-docker-images.sh "$(CILIUM_VERSION)"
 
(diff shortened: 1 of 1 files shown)


sayboras (approved):

ldelossa:
Something odd is going on with tests here. Maybe this needs a rebase before we mark it ready to merge? 

@sayboras @nbusseneau

nbusseneau:
@ldelossa Rebased but still the same issue. I've done a bit of digging and the issue was introduced in `main` by 660f2f26f26cd43841180924e261a21692e18ef1, I think it unfortunately conflicted with the previous merge of 6f1e4961576beb6c492038704e0c3f91b05b5dd2.

doniacld (approved):
Thanks 🎉

nbusseneau:
This change does not need to run any CI. Marking `ready-to-merge`.

PR #28226, opened 2023-09-19
Fix missing NODE_ADD messages

AwesomePatrol:
When `local-router-ipv4` is set in `cilium-config` it causes all `CiliumNodes` to have `NodeCiliumInternalIP` set to the same value. When "upserting" it in the node manager it detects collision expecting to later send node_update message, but as the node is new it instead skips sending node_add message completely.

This avoids setting `dpUpdate = false` when collision is detected for IP addresses of `NodeCiliumInternalIP`.

Bug: b/284134355

```release-note
Fix missing NODE_ADD Hubble peer messages in some cases
```

### DIFF
--- a/pkg/node/manager/manager.go
+++ b/pkg/node/manager/manager.go
@@ -79,6 +79,7 @@ type Configuration interface {
 	TunnelingEnabled() bool
 	RemoteNodeIdentitiesEnabled() bool
 	NodeEncryptionEnabled() bool
+	IsLocalRouterIP(string) bool
 }
 
 var _ Notifier = (*manager)(nil)
@@ -500,12 +501,17 @@ func (m *manager) NodeUpdated(n nodeTypes.Node) {
 		// rest as it is the strongest source, i.e. only trigger datapath
 		// updates if the information we receive takes priority.
 		//
-		// The only exception are kube-apiserver entries. In that case,
-		// we still want to inform subscribers about changes in auxiliary
-		// data such as for example the health endpoint.
+		// There are two exceptions to the rules above:
+		// * kube-apiserver entries - in that case,
+		//   we still want to inform subscribers about changes in auxiliary
+		//   data such as for example the health endpoint.
+		// * CiliumInternal IP addresses that match configured local router IP.
+		//   In that case, we still want to inform subscribers about a new node
+		//   even when IP addresses may seem repeated across the nodes.
 		existing := m.ipcache.GetMetadataByPrefix(prefix).Source()
 		overwrite := source.AllowOverwrite(existing, n.Source)
-		if !overwrite && existing != source.KubeAPIServer {
+		if !overwrite && existing != source.KubeAPIServer &&
+			!(address.Type == addressing.NodeCiliumInternalIP && m.conf.IsLocalRouterIP(address.ToString())) {
 			dpUpdate = false
 		}
 
--- a/pkg/node/manager/manager_test.go
+++ b/pkg/node/manager/manager_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/cilium/cilium/pkg/datapath/fake"
 	datapath "github.com/cilium/cilium/pkg/datapath/types"
 	"github.com/cilium/cilium/pkg/hive/cell"
+	"github.com/cilium/cilium/pkg/identity/cache"
 	"github.com/cilium/cilium/pkg/inctimer"
 	"github.com/cilium/cilium/pkg/ipcache"
 	ipcacheTypes "github.com/cilium/cilium/pkg/ipcache/types"
@@ -28,6 +29,7 @@ import (
 	"github.com/cilium/cilium/pkg/node/addressing"
 	nodeTypes "github.com/cilium/cilium/pkg/node/types"
 	"github.com/cilium/cilium/pkg/source"
+	testidentity "github.com/cilium/cilium/pkg/testutils/identity"
 )
 
 func Test(t *testing.T) {
@@ -43,6 +45,9 @@ type configMock struct {
 	RemoteNodeIdentity bool
 	NodeEncryption     bool
 	Encryption         bool
+
+	LocalRouterIPv4 string
+	LocalRouterIPv6 string
 }
 
 func (c *configMock) TunnelingEnabled() bool {
@@ -59,6 +64,10 @@ func (c *configMock) NodeEncryptionEnabled() bool {
 
 func (c *configMock) EncryptionEnabled() bool {
 	return c.Encryption
+}
+
+func (c *configMock) IsLocalRouterIP(ip string) bool {
+	return ip != "" && (c.LocalRouterIPv4 == ip || c.LocalRouterIPv6 == ip)
 }
 
 type nodeEvent struct {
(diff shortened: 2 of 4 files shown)


sayboras:
:wave: 

Sorry for late reply, the explanation is reasonable to me. 

Can you help to correct or fix a couple of failures related to static check? Feel free to let us know if you need any help.

AwesomePatrol:
It seems that rebase fixed most issues as linter was complaining about code outside this change.

There is one issue remaining in conformance tests, but I am not sure if it is just a flake or is caused by this change:
```
Summarizing 2 Failures:
  [FAIL] [sig-node] Kubelet when scheduling an agnhost Pod with hostAliases [It] should write entries to /etc/hosts [NodeConformance] [Conformance]
  test/e2e/common/node/kubelet.go:162
  [FAIL] [sig-storage] Projected secret [It] should be consumable from pods in volume with mappings [NodeConformance] [Conformance]
  test/e2e/framework/pod/output/output.go:247
  ```

**Comment by christarazi** [comment]
I'm having a hard time following the description of the fix because I don't quite understand the buggy behavior that is supposedly broken.

Could you elaborate on 

> When "upserting" it in the node manager it detects collision expecting to later send node_update message, but as the node is new it instead skips sending node_add message completely.

and specifically, what you mean by "detects collision expecting to later send node_update message"?

The code is clearly trying to set `dpUpdate = false` in a more specific case, but it's not clear to me why this is needed and how that is relevant to a supposedly missing "node_add" event.

**Comment by AwesomePatrol** [comment]
I probably assumed that the test will be clear enough, but here is the bugged scenario:
1. `nodeA`'s `CiliumNode` is detected
2. In `NodeUpdated()`, `nodeA` is saved with `CiliumInternalIP` equal to the value of `local-router-ipv4 = 169.254.4.6`
3. `NODE_ADD` message is sent to all Hubble peers (as `dpUpdate = true`)
4. `nodeB`'s `CiliumNode` is detected
5. In `NodeUpdated()`, it tries to save `CiliumInternalIP` with the same value as `nodeB`, but there is already a value for that, so `overwrite = true`. It then sets to `dpUpdate = false`
6. `NODE_ADD` message is not sent (as `dpUpdate = false`)

I am not sure if skipping setting `dpUpdate = false` for `CiliumInternalIP` is the best approach. Maybe handling `dpUpdate` should be improved.

File: pkg/node/manager/manager.go [hunk 1]
```diff
@@ -464,7 +464,7 @@ func (m *manager) NodeUpdated(n nodeTypes.Node) {
 		// data such as for example the health endpoint.
 		existing := m.ipcache.GetMetadataByPrefix(prefix).Source()
 		overwrite := source.AllowOverwrite(existing, n.Source)
-		if !overwrite && existing != source.KubeAPIServer {
+		if !overwrite && existing != source.KubeAPIServer && address.Type != addressing.NodeCiliumInternalIP {
```
marqc: I believe that the problem's root is in resolving the `exisitng` state of the node. When using `local-router-ipv4` all nodes will have the same internal IP address assigned (i.e. "169.254.4.6"). When a new node is added the code here incorrectly assumes that the Node is already populated because it already has an entry for IP "169.254.4.6".

What you are proposing is to effectively exclude NodeCiliumInternalIP from making the decision if NodeAdded event should be populated. This change generally LGTM.

nit: it will be more clear and easier to read if you move this check to the end of the loop and instead modyfing this condition use fail-fast construction like

```
if address.Type == addressing.NodeCiliumInternalIp {
    continue
}
```

marqc (approved):

pkg/node/manager/manager.go
AwesomePatrol: I don't know how the rest of this loop works, so I wanted to keep behavior changes to the absolute minimum. It would be great to have somebody that is more familiar with these code parts suggest better place to do this.

marqc: Yeah, I'm not an expert on this as well. A comment from maintainers will be much appreciated here.

AwesomePatrol:
@sayboras, @christarazi, please review (1 line change + tests)

CI is green. We need your opinion on https://github.com/cilium/cilium/pull/28226#discussion_r1343917776

pkg/node/manager/manager.go
christarazi: I think I would agree with the code snippet shared in https://github.com/cilium/cilium/pull/28226/files#r1343917776. Additionally, I would suggest something like:

```go
if (option.Config.LocalRouterIPv4 == address || option.Config.LocalRouterIPv6 == address) && address.Type == addressing.NodeCiliumInternalIp {
    continue
}
```

so that the condition is never triggered unexpectedly.

christarazi (changes requested):
Overall logic makes sense to me. Just one suggestion to implement and we can move forward.

michi-covalent (approved):
✅ ack on cli-owned files

pkg/node/manager/manager.go
```diff
@@ -477,6 +478,10 @@ func (m *manager) NodeUpdated(n nodeTypes.Node) {
 			continue
 		}
 
+		if addr := address.ToString(); address.Type == addressing.NodeCiliumInternalIP && m.conf.IsLocalRouterIP(addr) {
+			continue
```
christarazi: Actually, I think we want 

```suggestion
			dpUpdate = false
```

here instead to not break other behavior like inserting the IP into the ipcache.

AwesomePatrol: Did you mean `dpUpdate = true`? Anyways, I think we have two options here with the lowest chance to break other stuff:
a) Skip setting `dpUpdate = false` when IP matches criteria [1]
b) Skip processing when IP matches criteria

Unfortunately I couldn't track down what happens if there is `ipcache` or not for `LocalRouterIP`.

[1] criteria are: is `NodeCiliumInternalIP`, and matches either `LocalRouterIP`

pkg/node/manager/manager_test.go
```diff
@@ -895,3 +904,97 @@ func TestNodeManagerEmitStatus(t *testing.T) {
...
+
+func (m *mockUpdater) UpdateIdentities(_, _ cache.IdentityCache, _ *sync.WaitGroup) {}
+
+type mockTriggerer struct{}
+
+func (m *mockTriggerer) UpdatePolicyMaps(ctx context.Context, wg *sync.WaitGroup) *sync.WaitGroup {
+	return wg
+}
+
+func (s *managerTestSuite) TestNodeWithSameInternalIP(c *check.C) {
```
christarazi, line 918: Could you also add `ipcacheExpect("upsert", "...")` for the local router IP as well in this test?

pkg/node/manager/manager.go
christarazi: > Did you mean dpUpdate = true?

Ah yes. This is getting confusing :/. We still want the ipcache entry for that IP.

pkg/node/manager/manager_test.go
AwesomePatrol: I can't add these checks as I need to run a real `ipcache` in these test. Otherwise there will be `upsert` events without any IP collision detection, so the bug won't trigger.

christarazi (dismissed):

pkg/node/manager/manager.go
```diff
@@ -505,7 +506,8 @@ func (m *manager) NodeUpdated(n nodeTypes.Node) {
 		// data such as for example the health endpoint.
 		existing := m.ipcache.GetMetadataByPrefix(prefix).Source()
 		overwrite := source.AllowOverwrite(existing, n.Source)
-		if !overwrite && existing != source.KubeAPIServer {
+		if !overwrite && existing != source.KubeAPIServer &&
+			!(address.Type == addressing.NodeCiliumInternalIP && m.conf.IsLocalRouterIP(address.ToString())) {
```
christarazi, line 514: Could you update the doc paragraph above with the new condition, explaining it?

christarazi (approved):

PR #28008, opened 2023-09-07
CI images: Define a variable for the floating tags

michi-covalent:
Define a variable for CI image floating tags instead of hardcoding it.

- For the main branch, the floating tag is "latest".
- For stable branches, the floating tag is the same as the branch name (e.g. "v1.14").

PR #28228, opened 2023-09-19
CI images: Define a variable for the floating tags

michi-covalent:
Define a variable for CI image floating tags instead of hardcoding it.

- For the main branch, the floating tag is "latest".
- For stable branches, the floating tag is the same as the branch name (e.g. "v1.14").

This commit was initially introduced in #28008, but it got reverted because it tried to push floating tags for feature branches and failed. Now that #28044 is merged, this workflow no longer pushes floating tags from feature branches. Therefore it is safe to use github.ref_name as a floating tag.

### DIFF
--- a/.github/workflows/build-images-ci.yaml
+++ b/.github/workflows/build-images-ci.yaml
@@ -95,6 +95,11 @@ jobs:
           else
             echo tag=${{ github.sha }} >> $GITHUB_OUTPUT
           fi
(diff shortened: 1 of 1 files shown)


michi-covalent:
tested and verified in my fork that this doesn't try to push a floating tag from feature branches: https://github.com/michi-covalent/cilium/actions/runs/6239250480/job/16936783178#step:20:8

nbusseneau (approved):

PR #28166, opened 2023-09-14
operator: Remove identity GC and CES controller legacy metrics

pippolo84:
Update Identity GC and CES controller metrics to use modular metrics support from pkg hive.
Existing tests have been updated accordingly.

~Depends on: https://github.com/cilium/cilium/pull/28005~

PR #28233, opened 2023-09-20
operator: Migrate Cilium Endpoint GC to hive

alan-kut:
operator: Migrate Cilium Endpoint GC to hive

```release-note
```


alan-kut:
@pippolo84 FYI

operator/endpointgc/cell.go
pippolo84, line 1: Nit: this file and `gc.go` have been added empty in the second commit. Can you please rebase it so they will be added only in third commit?

operator/endpointgc/cell.go
```diff
@@ -1 +1,43 @@
...
+
+	"github.com/cilium/cilium/pkg/hive/cell"
+)
+
+const (
+	// Interval is the interval in which cilium endpoints are
+	// attempted to be collected
+	Interval = "cilium-endpoint-gc-interval"
+)
+
```
pippolo84: This is unused, so we can remove it.

operator/endpointgc/gc.go
```diff
@@ -1 +1,266 @@
...
+			PropagationPolicy: &propagationPolicy,
+			// Set precondition to ensure we are only deleting CEPs owned by
+			// this agent.
+			Preconditions: &meta_v1.Preconditions{
+				UID: &cep.UID,
+			},
+		})
+	switch {
+	case err == nil:
+		metrics.EndpointGCObjects.WithLabelValues(metrics.LabelValueOutcomeSuccess).Inc()
```
pippolo84: We recently introduced support for modular metrics in the operator https://github.com/cilium/cilium/pull/28005. We can now use `cell.Metric` to declare cell-specific metrics instead of using the legacy (global) ones. Do you mind modularize the metrics too?
You can find an example of how to do so [here](https://github.com/cilium/cilium/pull/28166/commits/62ef3f46d776b1203f7be1b0ab12d146d55e3f62) (CES controller metrics modularization)

pippolo84 (commented):
This is nice, thanks! :100:
I've left a couple of nits inline and a suggestion about metrics, but overall LGTM.
I didn't review the tests yet, I aim to do that in a second pass.

operator/endpointgc/cell.go
alan-kut: Done

operator/endpointgc/cell.go
alan-kut: Removed, thanks.
Firstly I wanted to move the flag to the gc but it is also used a little bit in the root. I decided to use shard config but forgot to remove everything

operator/endpointgc/gc.go
alan-kut: Done

operator/api/metrics_test.go
pippolo84, line 1: You also need to remove the check on `cilium_operator_endpoint_gc_objects` in `TestMetricsHandlerWithMetrics`:

```diff
-       if err := testMetric(
-               metrics,
-               "cilium_operator_endpoint_gc_objects",
-               float64(1),
-               map[string]string{
-                       operatorMetrics.LabelOutcome: operatorMetrics.LabelValueOutcomeSuccess,
-               },
-       ); err != nil {
-               t.Fatalf("error while inspecting metric: %s", err)
-       }
```

otherwise the test fails:

```
--- FAIL: TestMetricsHandlerWithMetrics (0.45s)
    metrics_test.go:189: error while inspecting metric: "cilium_operator_endpoint_gc_objects" not found
```

operator/endpointgc/gc_test.go
```diff
@@ -0,0 +1,246 @@
...
+
+	// Create CiliumNodes
+	// - cilium node for cep6
+	fakeClient.CiliumV2().CiliumNodes().Create(context.Background(), createCiliumNode("node6"), meta_v1.CreateOptions{})
+}
+
+func waitForCEPs(t *testing.T, cepStore resource.Store[*cilium_v2.CiliumEndpoint], number int) {
+	if err := testutils.WaitUntil(func() bool {
+		return len(cepStore.List()) == number
+	}, waitTime*5); err != nil {
```
pippolo84: I suggest to increase this a bit, maybe to a total of 5 or 10 seconds. This is to reduce flakyness in CI, where the performance characteristic is so unpredictable. Also, this won't introduce any fixed delay, since the condition is still checked every 10ms.

operator/endpointgc/gc_test.go
```diff
@@ -0,0 +1,246 @@
...
+	waitForCEPs(t, cepStore, 8)
+	if err := hive.Stop(context.Background()); err != nil {
+		t.Fatalf("failed to stop: %s", err)
+	}
+}
+
+func prepareCiliumEndpointCRD(c *k8sClient.FakeClientset) error {
+	c.APIExtFakeClientset.PrependReactor("get", "*", func(action k8sTesting.Action) (handled bool, ret runtime.Object, err error) {
+		return true, nil, nil
+	})
```
pippolo84, line 140: Is this needed?

pippolo84 (changes requested):
Thanks for this next iteration! I think there is a fix to be done for an existing operator/api test.

operator/api/metrics_test.go
alan-kut: Fixed

operator/endpointgc/gc_test.go
alan-kut: Yes. It passes the:
```go
func (g *GC) checkForCiliumEndpointCRD(ctx hive.HookContext) bool {
	_, err := g.clientset.ApiextensionsV1().CustomResourceDefinitions().Get(
		ctx, k8sconstv2.CEPName, metav1.GetOptions{ResourceVersion: "0"},
	)
	if err == nil {
		return true
	}
        ...
        return false
}
```

operator/endpointgc/gc_test.go
alan-kut: Changed. I have a problem with a test for disabled endpoints gc.
```go
// wait for all CEPs created by prepareCiliumEndpoints
waitForCEPs(t, cepStore, 8)
// wait for potential GC
time.Sleep(500 * time.Millisecond)
// gc is disabled so no CEPs should be deleted
waitForCEPs(t, cepStore, 8)
```
It's hard to tell how long should we wait to make sure gc is really disabled.

But maybe it's fine to just remove this test.

pippolo84: > It's hard to tell how long should we wait to make sure gc is really disabled.

That's right.
In this case, I don't think we can do much to improve this. Maybe what we can do is explicitly setting the gc interval with a value much smaller than the observation interval:

```diff
diff --git a/operator/endpointgc/gc_test.go b/operator/endpointgc/gc_test.go
index d0e23d52e5..d3acc2a8ee 100644
--- a/operator/endpointgc/gc_test.go
+++ b/operator/endpointgc/gc_test.go
@@ -112,6 +112,7 @@ func TestRegisterControllerWithCRDDisabled(t *testing.T) {
                cell.Metric(NewMetrics),
                cell.Provide(func() SharedConfig {
                        return SharedConfig{
+                               Interval:                 100 * time.Millisecond,
                                DisableCiliumEndpointCRD: true,
                        }
                }),
```

Unfortunately, this is still time-dependent, but (hopefully) should be fine in most cases and makes the test intent a bit clearer.

pippolo84:
The conflicts with the main branch is due to this recently merged PR: https://github.com/cilium/cilium/pull/28166

@alan-kut I solved the conflicts removing the legacy metrics in `legacy.go` and fixing the test in `operator/api/metrics` to use a test-only metric. To do that, I had to change your last commit, here a preview: https://github.com/cilium/cilium/compare/main...pippolo84:cilium:pr/pippolo84/operator-endpointgc-metrics?expand=1

Is it ok for you if I force-push this branch with these changes?

alan-kut:
It's ok, go ahead

pippolo84:
> It's ok, go ahead

Done. Let me know your thoughts about this, and then I'd say we're good to go :+1:

alan-kut:
> > It's ok, go ahead
> 
> Done. Let me know your thoughts about [this](https://github.com/cilium/cilium/pull/28233#discussion_r1345410941), and then I'd say we're good to go 👍

Everything looks good to me

pippolo84:
Pushed again just to address [this comment](https://github.com/cilium/cilium/pull/28233#discussion_r1345410941) setting a low timeout.

pippolo84 (approved):
Awesome! Thanks for the contribution! :rocket:

pippolo84:
I think the Conformance Runtime failure is related: https://github.com/cilium/cilium/actions/runs/6418834513/job/17427580426#step:16:276

@alan-kut could you please take a look?

alan-kut:
Done. I removed the wait for all 8 CEPs to be in the Store. There may be a race in which the CEPs are deleted before we can check all of them were created.

alan-kut:
Integration Tests (ci-integration): github.com/cilium/cilium/operator/pkg/lbipam  test failed. It's a flake.

alan-kut:
The other tests failure logs don't show me any good reason why they failed.

PR #28239, opened 2023-09-21
Add option to pass api-rate-limit via Helm values

ungureanuvladvictor:
I was looking at tweaking this values when running Cilium in OpenShift, but realized there is no option directly for it.

I know we have the https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/values.yaml.tmpl#L203-L210 option and used that for now. Interested in your thoughts about having an explicit value for setting the api rate limiter settings.

Signed-off-by: Vlad Ungureanu <vladu@ungureanuvladvictor@gmail.com>

```release-note
Add option to pass api-rate-limit via Helm values
```

### DIFF
--- a/Documentation/helm-values.rst
+++ b/Documentation/helm-values.rst
@@ -40,6 +40,10 @@
      - Annotate k8s node upon initialization with Cilium's metadata.
      - bool
      - ``false``
+   * - :spelling:ignore:`apiRateLimit`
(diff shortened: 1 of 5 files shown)


michi-covalent (approved):

sayboras (approved):
LGTM :heavy_check_mark:

PR #28244, opened 2023-09-21
ipam/multipool: Introduce specific ip family annotations for specifying ip pools

hargrovee:
In dual-stack mode, it's necessary to allocate both IPv4 and IPv6 addresses for pods. However, custom IP pools may only focus on either IPv4 or IPv6, which can result in the custom IP pool being configured with only IPv4 or IPv6. When specifying an IP pool for a namespace or pod through annotations, this can cause pod creation to fail because the specified IP pool lacks the IPv4 or IPv6 family.

This PR introduces two IP-family specific annotations: ipam.cilium.io/ipv4-pool and ipam.cilium.io/ipv6-pool. When specifying pools for pods or namespaces, combining these annotations allows for handling more scenarios.

This PR covers the use cases mentioned here https://github.com/cilium/cilium/pull/28204#issue-1900776538 and has been implemented based on the suggestions here https://github.com/cilium/cilium/pull/28204#pullrequestreview-1631009351.

### DIFF
--- a/pkg/ipam/metadata/manager.go
+++ b/pkg/ipam/metadata/manager.go
@@ -12,6 +12,7 @@ import (
 	"github.com/cilium/cilium/daemon/k8s"
 	"github.com/cilium/cilium/pkg/annotation"
 	"github.com/cilium/cilium/pkg/hive"
+	"github.com/cilium/cilium/pkg/ipam"
 	ipamOption "github.com/cilium/cilium/pkg/ipam/option"
 	"github.com/cilium/cilium/pkg/k8s/resource"
 	slim_core_v1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/core/v1"
@@ -95,9 +96,32 @@ func splitK8sPodName(owner string) (namespace, name string, ok bool) {
 	return namespace, name, true
 }
 
-func (m *Manager) GetIPPoolForPod(owner string) (pool string, err error) {
+func determinePoolByAnnotations(annotations map[string]string, family ipam.Family) (pool string, ok bool) {
+	switch family {
+	case ipam.IPv4:
+		if annotations[annotation.IPAMIPv4PoolKey] != "" {
+			return annotations[annotation.IPAMIPv4PoolKey], true
+		} else if annotations[annotation.IPAMPoolKey] != "" {
+			return annotations[annotation.IPAMPoolKey], true
+		}
+	case ipam.IPv6:
+		if annotations[annotation.IPAMIPv6PoolKey] != "" {
+			return annotations[annotation.IPAMIPv6PoolKey], true
+		} else if annotations[annotation.IPAMPoolKey] != "" {
+			return annotations[annotation.IPAMPoolKey], true
+		}
+	}
+
+	return "", false
+}
+
+func (m *Manager) GetIPPoolForPod(owner string, family ipam.Family) (pool string, err error) {
 	if m.namespaceStore == nil || m.podStore == nil {
 		return "", &ManagerStoppedError{}
+	}
+
+	if family != ipam.IPv6 && family != ipam.IPv4 {
+		return "", fmt.Errorf("invalid IP family: %s", family)
 	}
 
 	namespace, name, ok := splitK8sPodName(owner)
@@ -116,8 +140,8 @@ func (m *Manager) GetIPPoolForPod(owner string) (pool string, err error) {
 		return "", fmt.Errorf("failed to lookup pod %q: %w", namespace+"/"+name, err)
 	} else if !ok {
 		return "", &ResourceNotFound{Resource: "Pod", Namespace: namespace, Name: name}
-	} else if ipPool, hasAnnotation := pod.Annotations[annotation.IPAMPoolKey]; hasAnnotation {
-		return ipPool, nil
+	} else if ippool, ok := determinePoolByAnnotations(pod.Annotations, family); ok {
+		return ippool, nil
 	}
 
 	// Check annotation on namespace
@@ -128,8 +152,8 @@ func (m *Manager) GetIPPoolForPod(owner string) (pool string, err error) {
 		return "", fmt.Errorf("failed to lookup namespace %q: %w", namespace, err)
 	} else if !ok {
 		return "", &ResourceNotFound{Resource: "Namespace", Name: namespace}
-	} else if ipPool, hasAnnotation := podNamespace.Annotations[annotation.IPAMPoolKey]; hasAnnotation {
-		return ipPool, nil
+	} else if ippool, ok := determinePoolByAnnotations(podNamespace.Annotations, family); ok {
+		return ippool, nil
 	}
 
 	// Fallback to default pool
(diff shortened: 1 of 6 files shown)


pkg/ipam/metadata/manager.go
```diff
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/cilium/cilium/pkg/ipam"
```
gandro: nit: This import here is sorted in the wrong order, this should be part of the local package group below.

pkg/ipam/metadata/manager.go
```diff
@@ -116,8 +118,25 @@ func (m *Manager) GetIPPoolForPod(owner string) (pool string, err error) {
 		return "", fmt.Errorf("failed to lookup pod %q: %w", namespace+"/"+name, err)
 	} else if !ok {
 		return "", &ResourceNotFound{Resource: "Pod", Namespace: namespace, Name: name}
-	} else if ipPool, hasAnnotation := pod.Annotations[annotation.IPAMPoolKey]; hasAnnotation {
-		return ipPool, nil
+	} else {
+		switch family {
```
gandro: nit: I think it might be worth splitting this out into a function that takes the familiy and the annotations as arguments, and returns `(pool string, ok bool)`

pkg/ipam/metadata/manager.go
```diff
@@ -95,7 +97,7 @@ func splitK8sPodName(owner string) (namespace, name string, ok bool) {
 	return namespace, name, true
 }
 
-func (m *Manager) GetIPPoolForPod(owner string) (pool string, err error) {
+func (m *Manager) GetIPPoolForPod(owner string, family ipam.Family) (pool string, err error) {
 	if m.namespaceStore == nil || m.podStore == nil {
 		return "", &ManagerStoppedError{}
 	}
```
gandro, line 121: I think we should add a check here to reject invalid IP families, as otherwise it will silently fall back on the default pool which hides this error.


```suggestion
func (m *Manager) GetIPPoolForPod(owner string, family ipam.Family) (pool string, err error) {
	if m.namespaceStore == nil || m.podStore == nil {
		return "", &ManagerStoppedError{}
	}
	
	if family != ipam.IPv6 && family != ipam.IPv4 {
		return "", fmt.Errorf("invalid IP family: %s", family)
	}
```

pkg/ipam/metadata/manager.go
```diff
@@ -116,8 +118,25 @@ func (m *Manager) GetIPPoolForPod(owner string) (pool string, err error) {
...
+				pod.Annotations[annotation.IPAMPoolKey]; hasAnnotation {
+				return ipPool, nil
+			}
+		case ipam.IPv6:
+			if ipv6Pool, hasIPv6Annotation :=
+				pod.Annotations[annotation.IPAMIPv6PoolKey]; hasIPv6Annotation {
+				return ipv6Pool, nil
+			} else if ipPool, hasAnnotation :=
+				pod.Annotations[annotation.IPAMPoolKey]; hasAnnotation {
+				return ipPool, err
```
gandro:
```suggestion
				return ipPool, nil
```

gandro (commented):
Awesome work, looks good overall. Thanks for taking care of this!

I've left some minor feedback about the code itself.

pkg/ipam/metadata/manager.go
hargrovee: Sure, thank you for pointing that out.

pkg/ipam/metadata/manager.go
hargrovee: OK, I will split them out into a function. Thanks!

pkg/ipam/metadata/manager.go
hargrovee: Alright, thank you for the guidance. I will add the checks.

pkg/ipam/metadata/manager.go
hargrovee: Thanks,  I will make the correction.

hargrovee:
Thanks! I have made the changes.

pkg/ipam/metadata/manager.go
```diff
@@ -95,11 +96,34 @@ func splitK8sPodName(owner string) (namespace, name string, ok bool) {
 	return namespace, name, true
 }
 
-func (m *Manager) GetIPPoolForPod(owner string) (pool string, err error) {
+func determinePoolByAnnotations(ipv4Annotation, ipv6Annotation, ipAnnotation string, family ipam.Family) (pool string, ok bool) {
```
gandro: nit: Could you pass in `annotations map[string]string` as the argument here, and then populate it either from `pod.Annotations` or `namespace.Annotations`?

gandro (approved):
Great, looks good! One remaining minor suggestion

pkg/ipam/metadata/manager.go
```diff
@@ -116,8 +140,17 @@ func (m *Manager) GetIPPoolForPod(owner string) (pool string, err error) {
...
-	} else if ipPool, hasAnnotation := pod.Annotations[annotation.IPAMPoolKey]; hasAnnotation {
-		return ipPool, nil
+	} else {
+		podAnnotations := map[string]string{
+			"ipv4Annotation": pod.Annotations[annotation.IPAMIPv4PoolKey],
+			"ipv6Annotation": pod.Annotations[annotation.IPAMIPv6PoolKey],
+			"ipAnnotation":   pod.Annotations[annotation.IPAMPoolKey],
+		}
+
+		ippool, ok := deternimePoolByAnnotations(podAnnotations, family)
```
gandro: I'm confused. Why can't you pass in

```suggestion
		ippool, ok := deternimePoolByAnnotations(pod.Annotations, family)
```

here?

hargrovee: Yes, I didn't notice that. Thank you.

pkg/ipam/metadata/manager.go
hargrovee: OK, I will change it. Thanks!

hargrovee:
Thank you. I made the changes. Please let me know if there are any other issues. Appreciate your guidance.

pkg/ipam/metadata/manager.go
```diff
@@ -95,11 +96,34 @@ func splitK8sPodName(owner string) (namespace, name string, ok bool) {
 	return namespace, name, true
 }
 
-func (m *Manager) GetIPPoolForPod(owner string) (pool string, err error) {
+func deternimePoolByAnnotations(annotations map[string]string, family ipam.Family) (pool string, ok bool) {
```
gandro: nit: determine is misspelled

gandro (commented):
One last suggestion, then it should be good to go!

pkg/ipam/metadata/manager.go
hargrovee: Yes, that's a typo. My mistake. Thank you.

hargrovee:
Thank you for your guidance and thoughtful suggestions. Much appreciated.

gandro (approved):
Thanks!

gandro:
@tommyp1ckles PTAL

aanm (approved):

PR #28247, opened 2023-09-21
ci: update k8s versions support for v1.13

nbusseneau:
Cilium 1.13 only supports K8s up to 1.26 as per
https://docs.cilium.io/en/v1.13/network/kubernetes/compatibility/#k8scompatibility

- We remove 1.27 testing on AKS as it shouldn't have been here in the first place as per our historical policy.
- We remove 1.23 testing on GKE since it's not supported anymore.

### DIFF
--- a/.github/actions/azure/k8s-versions.yaml
+++ b/.github/actions/azure/k8s-versions.yaml
@@ -7,7 +7,4 @@ include:
   - version: "1.26"
     location: westus
     index: 2
-  - version: "1.27"
-    location: eastasia
-    index: 3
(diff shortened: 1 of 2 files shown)


lmb (approved):

nbusseneau:
CI for affected workflows passed, marking ready to merge.

lmb:
I'm not super happy that I have to override branch protection to get this merged. If other CI runs are not needed we should make it so from the CI side, not rely on an individual's knowledge of CI.

PR #28253, opened 2023-09-22
Docs: Updates BGP CP for LB Class Support

### LINKED ISSUE #27056: Docs: Add LoadBalancerClass to BGP CP Service Announcements
Update the [service-announcements](https://docs.cilium.io/en/latest/network/bgp-control-plane/#service-announcements) docs for LoadBalancerClass support when https://github.com/cilium/cilium/pull/26905 merges.

danehans:
Adds [LB class](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class) support to BGP CP docs.

Fixes: #27056

```release-note
docs: Update BGP control plane documentation with regards to LB class support and service announcements
```

### DIFF
--- a/Documentation/network/bgp-control-plane.rst
+++ b/Documentation/network/bgp-control-plane.rst
@@ -347,7 +347,8 @@ Service announcements
 
 By default, virtual routers will not announce services. Virtual routers will announce
 the ingress IPs of any LoadBalancer services that matches the ``.serviceSelector``
-of the virtual router.
+of the virtual router and has `loadBalancerClass <https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class>`__
+unspecified or set to ``io.cilium/bgp-control-plane``.
 
 If you wish to announce ALL services within the cluster, a ``NotIn`` match expression
 with a dummy key and value can be used like:


danehans:
cc @rastislavs

harsimran-pabla (approved):

Documentation/network/bgp-control-plane.rst
```diff
@@ -364,6 +364,9 @@ with a dummy key and value can be used like:
          matchExpressions:
             - {key: somekey, operator: NotIn, values: ['never-used-value']}
 
+A service must have `loadBalancerClass <https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class>`__
+unspecified or set to "io.cilium/bgp-control-plane" to be announced by a policy.
```
qmonnet: Nit: I'd rather have pre-formatted text to mark that this is a specific string here,

```suggestion
unspecified or set to ``io.cilium/bgp-control-plane`` to be announced by a policy.
```

qmonnet (changes requested):
Thanks for the addition!
I've got a few comments.

* The beginning of the subsection says:
  > Virtual routers will announce the ingress IPs of any LoadBalancer services that matches the ``.serviceSelector`` of the virtual router.

  But then this is not correct, it should be `any ... without loadBalancerClass`? I'm a bit worried readers might read the beginning but skip the restriction you're mentioning. On the other hand, the subsection is small, so if users don't read all of it there's only so much we can do I guess?

* Should the paragraph briefly explain why service with other `loadBalancerClass`es are not announced?

Documentation/network/bgp-control-plane.rst
danehans: Commit 0c5f501 resolves this comment.

danehans:
@qmonnet thank you for reviewing my PR.

> I'm a bit worried readers might read the beginning but skip the restriction you're mentioning. 

I have repositioned and reworded the documentation to address this concern, PTAL.

> Should the paragraph briefly explain why service with other loadBalancerClasses are not announced?

IMHO the link to the k8s lb class doc provides the reader with the opportunity to learn more. The current approach provides a balance of directness with the opportunity for the reader to learn more through the link. I don't have a strong opinion on this issue so I'm happy to update if you and others feel differently.

qmonnet (approved):
OK, no strong opinion either, and I'm good with the current version. Thanks a lot!

PR #28262, opened 2023-09-25
gateway-api: Add conformance profile test

sayboras:
### Description

This is to support the conformance profile run, so that we can just easily capture the report and submit the same to upstream repo

https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/reports/v0.7.1/cilium-cilium.yaml

### Testing

Please find below the conformance report generated as part of this run

https://github.com/cilium/cilium/suites/16825211249/artifacts/960894970

```yaml
apiVersion: gateway.networking.k8s.io/v1alpha1
date: "2023-10-03T13:08:01Z"
gatewayAPIVersion: TODO
implementation:
  contact:
  - https://github.com/cilium/community/blob/main/roles/Maintainers.md
  organization: cilium
  project: cilium
  url: github.com/cilium/cilium
  version: main
kind: ConformanceReport
profiles:
- core:
    result: success
    statistics:
      Failed: 0
      Passed: 29
      Skipped: 0
    summary: ""
  extended:
    result: success
    statistics:
      Failed: 0
      Passed: 10
      Skipped: 0
    summary: ""
    supportedFeatures:
    - HTTPRouteMethodMatching
    - HTTPResponseHeaderModification
    - HTTPRoutePathRedirect
    - HTTPRouteQueryParamMatching
    - HTTPRouteSchemeRedirect
    - HTTPRouteRequestMirror
    - HTTPRouteHostRewrite
    - HTTPRoutePathRewrite
    - HTTPRoutePortRedirect
    - HTTPRouteRequestMultipleMirrors
  name: HTTP
- core:
    result: success
    statistics:
      Failed: 0
      Passed: 11
      Skipped: 0
    summary: ""
  name: TLS
```

### DIFF
--- a/.github/workflows/conformance-gateway-api.yaml
+++ b/.github/workflows/conformance-gateway-api.yaml
@@ -86,7 +86,11 @@ jobs:
       matrix:
         include:
         - crd-channel: experimental
+          conformance-profile: false
         - crd-channel: standard
+          conformance-profile: false
+        - crd-channel: experimental
+          conformance-profile: true
     steps:
       - name: Checkout context ref (trusted)
         uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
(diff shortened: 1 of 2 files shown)


meyskens:
Where would we get the report from exactly, am i missing that? I don't see the output of it, would be nice as an asset on the CI job

sayboras:
> Where would we get the report from exactly, am i missing that? I don't see the output of it, would be nice as an asset on the CI job

Good catch, I have added upload step to perform the same :thanks:

meyskens (approved):
LGTM

operator/pkg/gateway-api/experimental_conformance_test.go
```diff
@@ -0,0 +1,141 @@
...
+	conformanceProfiles  sets.Set[suite.ConformanceProfileName]
+	skipTests            []string
+)
+
+func TestExperimentalConformance(t *testing.T) {
+	testutils.GatewayAPIConformanceTest(t)
+
+	var err error
+	cfg, err = config.GetConfig()
+	if err != nil {
```
viktor-kurchenko, line 45: NIT: why don't we use `assert.NoError` for such checks?

.github/workflows/conformance-gateway-api.yaml
```diff
@@ -228,13 +232,39 @@ jobs:
...
           EOF
           
-          GATEWAY_API_CONFORMANCE_TESTS=1 go test \
-            -p 4 \
-            -v ./operator/pkg/gateway-api \
-            --gateway-class cilium \
-            --supported-features "${{ steps.vars.outputs.supported_features }}" \
-            -test.run "TestConformance" \
-            -test.skip "${{ steps.vars.outputs.skipped_tests }}"
+          TEST_RUN="TestConformance"
```
viktor-kurchenko: Why do we need `TEST_RUN`?

operator/pkg/gateway-api/experimental_conformance_test.go
sayboras: good point, this is to keep the upgrade path from upstream easier. The content of this test file is 99% coming from the upstream Gateway API.

.github/workflows/conformance-gateway-api.yaml
sayboras: ah good catch 🙇‍♂️

viktor-kurchenko (approved):
LGTM.

PR #28264, opened 2023-09-25
BPF Host Routing: do not skip fib lookups

### LINKED ISSUE #27343: Broken connectivity when using BFP masquerade, no tunnel and BPF Host Routing on EKS with 1 device
I tracked down this issue to a change that happened between v1.14.0-snapshot.1 and v1.14.0-snapshot.2

On an EKS cluster, packets that are supposed to go to the EKS kube-apiserver get silently dropped when this Helm config is applied:

```
bpf:
  masquerade: true
devices: eth0
eni:
  enabled: true
ipam:
  mode: eni
kubeProxyReplacement: strict
tunnel: disabled
```

Disabling BPF masquerading, using `eth+` on the devices or setting `vxlan` for the tunnel, all lead to a working communication with the kube-apiserver. KRP is unrelated, but it needs to be set to some value when using the 1.14.0 chart due to the default value changing to false.

Same Helm configuration, but as a command (for easier testing):
```
helm upgrade -n kube-system cilium cilium/cilium --version 1.14.0  \
  --set image.tag=v1.14.0-snapshot.2 --set image.useDigest=false \
  --set bpf.masquerade=true \
  --set devices=eth0 \
  --set tunnel=disabled \
  --set eni.enabled=true \
  --set ipam.mode=eni \
  --set kubeProxyReplacement=strict
```

When setting `devices: eth+`, I can see with `cilium monitor` that the packet is now being routed through `eth1`. 

With 1.13.5, the packet goes and comes back over `ifindex 0`.
```
Policy verdict log: flow 0x93b86448 local EP ID 3678, remote ID 16777218, proto 6, egress, action allow, match L3-L4, 10.4.201.117:40748 -> 10.4.1.206:443 tcp SYN
-> stack flow 0x93b86448 , identity 35566->16777218 state new ifindex 0 orig-ip 0.0.0.0: 10.4.201.117:40748 -> 10.4.1.206:443 tcp SYN
-> endpoint 3678 flow 0xe857e857 , identity 16777218->35566 state reply ifindex 0 orig-ip 10.4.1.206: 10.4.1.206:443 -> 10.4.201.117:40748 tcp SYN, ACK
-> stack flow 0x93b86448 , identity 35566->16777218 state established ifindex 0 orig-ip 0.0.0.0: 10.4.201.117:40748 -> 10.4.1.206:443 tcp ACK
```

With 1.14.0-snapshot.2 & eth0, it sends the packet on eth0 but the syn,ack never gets back:
```
Policy verdict log: flow 0xcf0da62e local EP ID 3678, remote ID 16777268, proto 6, egress, action allow, match L4-Only, 10.4.201.117:51344 -> 10.4.1.206:443 tcp SYN
-> network flow 0xcf0da62e , identity 35566->16777268 state new ifindex eth0 orig-ip 0.0.0.0: 10.4.201.117:51344 -> 10.4.1.206:443 tcp SYN
```

With 1.14.0-snapshot.2 & eth+, the packet gets sent on eth1 and we get the syn,ack back:
```
Policy verdict log: flow 0x2d05b7a4 local EP ID 3678, remote ID 16777217, proto 6, egress, action allow, match L4-Only, 10.4.201.117:35980 -> 10.4.1.206:443 tcp SYN
-> network flow 0x2d05b7a4 , identity 35566->16777217 state new ifindex eth1 orig-ip 0.0.0.0: 10.4.201.117:35980 -> 10.4.1.206:443 tcp SYN
-> endpoint 3678 flow 0xa6c8a6c8 , identity 16777217->35566 state reply ifindex lxc6bedd6da59c0 orig-ip 10.4.1.206: 10.4.1.206:443 -> 10.4.201.117:35980 tcp SYN, ACK
-> network flow 0x2d05b7a4 , identity 35566->16777217 state established ifindex eth1 orig-ip 0.0.0.0: 10.4.201.117:35980 -> 10.4.1.206:443 tcp ACK
```

When discussing this with Julian, he made me realize that https://github.com/cilium/cilium/pull/22006 (merged during the mentioned window) made changes around bpf host routing. So I tried switching to legacy host routing instead of BPF Host Routing, and connectivity was restored.

With 1.14.0-snapshot.2, `devices=eth0` and `bpf.hostLegacyRouting=true`. Again, syn and syn,ack on `ifindex 0`:
```
Policy verdict log: flow 0x2050dec6 local EP ID 3678, remote ID 16777268, proto 6, egress, action allow, match L4-Only, 10.4.201.117:34694 -> 10.4.1.206:443 tcp SYN
-> stack flow 0x2050dec6 , identity 35566->16777268 state new ifindex 0 orig-ip 0.0.0.0: 10.4.201.117:34694 -> 10.4.1.206:443 tcp SYN
-> endpoint 3678 flow 0xb495b495 , identity 16777268->35566 state reply ifindex 0 orig-ip 10.4.1.206: 10.4.1.206:443 -> 10.4.201.117:34694 tcp SYN, ACK
-> stack flow 0x2050dec6 , identity 35566->16777268 state established ifindex 0 orig-ip 0.0.0.0: 10.4.201.117:34694 -> 10.4.1.206:443 tcp ACK
```

I'm not sure if it means anything, but it's interesting that the Cilium identities for the kube-apiserver change depending on the configuration. With legacy host routing OR `devices: eth0`, it uses these two identities:
```
16777268   cidr:10.4.2.114/32
           reserved:kube-apiserver
           reserved:world
16777269   cidr:10.4.1.206/32
           reserved:kube-apiserver
           reserved:world
```

With BPF Host Routing AND devices: eth+, it uses these other two:
```
16777217   cidr:10.4.2.114/32
           reserved:kube-apiserver
           reserved:world
16777218   cidr:10.4.1.206/32
           reserved:kube-apiserver
           reserved:world
```

So, this particular combination of features doesn't work:
- EKS environment (`eni.enabled: true` + `ipam.mode: eni`)
- BPF Masquerade
- Direct Routing (`tunnel: disabled`)
- Only one interface (`devices: eth0`)
- BPF Host Routing (selected by default)

This is a change in behavior. The same configuration that worked correctly up until 1.14.0-snapshot.1 stopped working after 1.14.0-snapshot.2.

If this is now expected to be the case and people need to manually switch to legacy Host Routing in this case, we would need to at least document it. But ideally we should be able to do the right thing and not have packets silently dropped.

CC: @julianwiedmann, @aspsk

aspsk:
See the commit message.

Fixes: #27343

```release-note
Do not skip FIB lookup when running in BPF Host Routing when Endpoint Routes enabled
```

### DIFF
--- a/pkg/datapath/linux/config/config.go
+++ b/pkg/datapath/linux/config/config.go
@@ -1090,7 +1090,9 @@ func (h *HeaderfileWriter) writeTemplateConfig(fw *bufio.Writer, e datapath.Endp
 		}
 		fmt.Fprintf(fw, "#define DIRECT_ROUTING_DEV_IFINDEX %d\n", directRoutingIfIndex)
 		if len(option.Config.GetDevices()) == 1 {
-			fmt.Fprintf(fw, "#define ENABLE_SKIP_FIB 1\n")
+			if e.IsHost() || !option.Config.EnforceLXCFibLookup() {
+				fmt.Fprintf(fw, "#define ENABLE_SKIP_FIB 1\n")
+			}
 		}
 	}
 
--- a/pkg/option/config.go
+++ b/pkg/option/config.go
@@ -4433,3 +4433,13 @@ func parseBPFMapEventConfigs(confs BPFEventBufferConfigs, confMap map[string]str
 	}
 	return nil
 }
+
+func (d *DaemonConfig) EnforceLXCFibLookup() bool {
+	// See https://github.com/cilium/cilium/issues/27343 for the symptoms.
+	//
+	// We want to enforce FIB lookup if EndpointRoutes are enabled, because
+	// this was a config dependency change which caused different behaviour
+	// since v1.14.0-snapshot.2. We will remove this hack later, once we
+	// have auto-device detection on by default.
+	return d.EnableEndpointRoutes
+}


aspsk:
This patch removes the `ENABLE_SKIP_FIB` completely. Won't it be useful to have a flag to enable it for setups when only one interface is present? (cc: @julianwiedmann)

lmb (approved):
config change seems fine to me. No idea whether ENABLE_SKIP_FIB is great or not, but I'm all for reducing the number of code paths we have. Unless we have a strong indication that it's hurtful I'd say go for it!

julianwiedmann:
Ripping out `ENABLE_SKIP_FIB` completely feels like a *very* big hammer. And for instance also impacts unrelated areas like N/S LB forwarding to remote backends.

Is there no way to only exclude `ENABLE_SKIP_FIB` for the regressing config? Or even teach the problematic FIB lookup path (in bpf_lxc) to ignore `ENABLE_SKIP_FIB` ?

aspsk:
> Is there no way to only exclude `ENABLE_SKIP_FIB` for the regressing config?

The config can be described as follows: `no tunnel & there are interfaces not controlled by cilium`, so I am not sure how to automate this decision. We currently experience this only on EKS, as far as we know, but this is, of course, applicable to any config as defined above.

So, does this work if we only undef `ENABLE_SKIP_FIB` for lxc, and leave it as is for other callers?

aspsk:
> So, does this work if we only undef `ENABLE_SKIP_FIB` for lxc, and leave it as is for other callers?

For this specific case I can do a check like this (and only apply to lxc, not host so that nodeport is not affected):
```
option.Config.IPAM == ipamOption.IPAMENI
```
Then if we have more cases like this, we can add an option or extend the check. WDYT @julianwiedmann

aspsk:
I've posted a new version which only disables this on EKS. Tested that this actually works.

pkg/datapath/linux/config/config.go
```diff
@@ -1073,6 +1074,11 @@ func (h *HeaderfileWriter) WriteEndpointConfig(w io.Writer, e datapath.EndpointC
 	return h.writeTemplateConfig(fw, e)
 }
 
+func enforceFibLookup(e datapath.EndpointConfiguration) bool {
```
lmb: Could you put this on the type of `option.Config` instead? Then we wouldn't add another dependency on the global variable.

lmb (changes requested):

borkmann:
Would this situation be prevented if we had proper device detection at runtime, so that Cilium considers multiple devices when they get automatically added on EKS and therefore it needs to turn it off automatically when number grows from 1 to 2+? Cc @joamaki

joamaki:
> Would this situation be prevented if we had proper device detection at runtime, so that Cilium considers multiple devices when they get automatically added on EKS and therefore it needs to turn it off automatically when number grows from 1 to 2+? Cc @joamaki

Yep, I'd expect with runtime device detection enabled we wouldn't need this workaround (as then we'd reload if we get more devices at runtime). @bimmlerd and I are working on enabling it by default for v1.15. I would still go ahead with this fix (and backport it) and then remove it once it's no longer needed.

bimmlerd:
@aspsk this doesn't look like a backport to v1.14. Did you mean to add the `needs-backport/v1.14` label instead of `backport/v1.14`?

aspsk:
> Yep, I'd expect with runtime device detection enabled we wouldn't need this workaround

Yes, enabling the runtime device detection is the right way to do this, but the original report was that "this configuration worked, but not anymore" where the runtime detection is off.

Should we only apply this patch to `.14` and enable runtime detection in `.15` an later?

borkmann:
> > Yep, I'd expect with runtime device detection enabled we wouldn't need this workaround
> 
> Yes, enabling the runtime device detection is the right way to do this, but the original report was that "this configuration worked, but not anymore" where the runtime detection is off.
> 
> Should we only apply this patch to `.14` and enable runtime detection in `.15` an later?

I think that could be an option as long as we have a release blocker label to ensure the dev detection lands in time.

julianwiedmann:
> > So, does this work if we only undef `ENABLE_SKIP_FIB` for lxc, and leave it as is for other callers?
> 
> For this specific case I can do a check like this (and only apply to lxc, not host so that nodeport is not affected):
> 
> ```
> option.Config.IPAM == ipamOption.IPAMENI
> ```
> 
> Then if we have more cases like this, we can add an option or extend the check. WDYT @julianwiedmann

Sorry, was mostly afk last week.

Yep, that looks good on first glance :+1:. But instead of `ipamOption.IPAMENI`, shouldn't we check for `option.Config.EnableEndpointRoutes`  ?

julianwiedmann (approved):
Looks good, thank you! But could you please update the commit msg & release notes? :pray: 

@margamanterola maybe you want to confirm that this fixes the issue from https://github.com/cilium/cilium/issues/27343 ?

margamanterola:
For the release note, it should cover the symptoms/effects rather than the implementation. Is "Route packets to the correct interface when using BPF Host Routing on EKS" a good description of the outcome of this change?

I've destroyed the environment that I was using to test this, but I'll spin it up again and test this PR there.

lmb (approved):
Thanks :)

pkg/option/config.go
```diff
@@ -4391,3 +4391,13 @@ func parseBPFMapEventConfigs(confs BPFEventBufferConfigs, confMap map[string]str
 	}
 	return nil
 }
+
+func (d *DaemonConfig) EnforceLXCFibLookup() bool {
+	// See https://github.com/cilium/cilium/issues/27343 for the symptoms.
+	//
+	// We want to enforce FIB lookup if EndpointRoutes are enabled, because
+	// this was a config dependency change which caused different behaviour
+	// since v1.14.0-snapshot.2. We will remove this hack later, once we
```
derailed, line 4442: Wondering if we can tag these somehow so we can `grep` for code debt in the future?

derailed (approved):
@aspsk Nice work!

sylr:
Can this be merged ?

PR #28276, opened 2023-09-26
[DRAFT]: Generic reconciler

joamaki:
Builds on top of https://github.com/cilium/cilium/pull/28140.

This implements a "generic reconciler" which takes the desired state as a StateDB table
and a "reconciliation target" that defines the reconciliation operations. It performs 
fast incremental reconciliation of changes and periodic full reconciliation. Metrics
and health reporting are included.

`pkg/reconciler/example` has some rough early examples for sysctl and iptables. 

By wrapping the reconciliation into a reusable generic utility we gain:
- Single well-tested implementation for mission critical code
- Uniform metrics, health and status reporting
- Significant reduction in code duplication
- The "concept" of reconciler that can be understood and inspected from outside

Summary of the API:
```
package reconciler

// Reconcilable objects carry status
type Reconcilable[Obj comparable] interface {
  comparable
  GetStatus() Status
  WithStatus(Status) Obj
}

// Reconciliation target defines the operations for reconciling a single
// object.
type Target[Obj Reconcilable[Obj]] interface {
  Init(context.Context) error
  Update(context.Context, Obj) error
  Delete(context.Context, Obj) error
  Sync(context.Context, statedb.Iterator[Obj]) (outOfSync bool, err error)
}

type params[Obj Reconcilable[Obj]] struct {
  cell.In

  // ... other dependencies ...

  // Dependencies that parametrize the reconciler for the concrete
  // use case:
  Config
  Target Target[Obj]
  StatusIndex statedb.Index[Obj, StatusKind]
}

// Register creates a reconciler and adds it to the application lifecycle.
// "cell.Invoke(reconciler.Register[MyObject])".
func Register[Obj Reconcilable[Obj]](p params[Obj]) { ... }
  
```

Usage:
```
package example

// Define the object that presents the desired state.
type MyObject struct { ... }
func (m *MyObject) GetStatus() reconciler.Status { ... }
func (m *MyObject) WithStatus(s reconciler.Status) *MyObject { /* copy and set status */ }
var statusIndex = reconciler.NewStatusIndex[*MyObject]()
var idIndex = statedb.Index[*MyObject, UUID]{...}

// Define a safe API for managing the desired state. Alternatively if it's obvious
// the RWTable[*MyObject] could be directly exposed.
// (TODO: Not sure at all if we want to call these sort of things managers)
type MyObjectManager struct {...}
func (m *MyObjectManager) Set(...) error {}
func (m *MyObjectManager) Get(...) error {}
func (m *MyObjectManager) Wait(context.Context) error { 
  // see example/sysctl.go for how we can wait for reconciliation.
}
func newMyObjectManager(db *statedb.DB, table statedb.RWTable[*MyObject) *MyObjectManager { ... }

var Cell = cell.Module(
  "my-module", 
  "There are many like it, but this one is mine",

  // Create the table for the desired state. Keep write access protected to this module.
  statedb.NewProtectedTableCell[*MyObject]("my-objects", idIndex, statusIndex),

  // Provide a public API for managing the desired state to the rest of the application.
  cell.Provide(newMyObjectManager),

  // Provide the reconciler dependencies, but scoped to just this module.
  cell.ProvidePrivate(
    func() reconciler.Target[*MyObject] { return &myObjectTarget{} },
    func() reconciler.Config { ...},
    func() statedb.Index[*MyObject, reconciler.StatusKind] { return statusIndex },
  ),

  // Create and register the reconciler to the application lifecycle. 
  cell.Invoke(reconciler.Register[*MyObject]),
)

type myObjectTarget struct { ... }
func (r *myObjectTarget) Init(context.Context) error { ... }
func (r *myObjectTarget) Update(context.Context, Obj) error { ... }
func (r *myObjectTarget) Delete(context.Context, Obj) error { ... }
func (r *myObjectTarget) Sync(context.Context, statedb.Iterator[Obj]) (outOfSync bool, err error) { .. }
```

Health reporting:
```
Status{ModuleID: iptables, Level: Degraded, Since: 3.000499163s ago, Message: Incremental reconciliation failure, Err: oops iptables error}
```

Metrics: 
```
# HELP cilium_reconciler_full_duration_seconds Histogram over full reconciliation duration
# TYPE cilium_reconciler_full_duration_seconds histogram
cilium_reconciler_full_duration_seconds_bucket{module_id="sysctl",le="0.005"} 1
# ...
cilium_reconciler_full_duration_seconds_sum{module_id="sysctl"} 6.0364e-05
cilium_reconciler_full_duration_seconds_count{module_id="sysctl"} 1
# HELP cilium_reconciler_full_total Number of full reconciliations performed
# TYPE cilium_reconciler_full_total counter
cilium_reconciler_full_total{module_id="sysctl"} 1
# HELP cilium_reconciler_incremental_duration_seconds Histogram of per-operation duration during incremental reconciliation
# TYPE cilium_reconciler_incremental_duration_seconds histogram
cilium_reconciler_incremental_duration_seconds_bucket{module_id="iptables",le="0.005"} 4
# ...
cilium_reconciler_incremental_duration_seconds_sum{module_id="iptables"} 9.6464e-05
cilium_reconciler_incremental_duration_seconds_count{module_id="iptables"} 4
cilium_reconciler_incremental_duration_seconds_bucket{module_id="sysctl",le="0.005"} 14
# ...
cilium_reconciler_incremental_duration_seconds_sum{module_id="sysctl"} 0.0009456850000000001
cilium_reconciler_incremental_duration_seconds_count{module_id="sysctl"} 14
# HELP cilium_reconciler_incremental_errors_current The number of objects currently failing to be reconciled
# TYPE cilium_reconciler_incremental_errors_current gauge
cilium_reconciler_incremental_errors_current{module_id="iptables"} 1
cilium_reconciler_incremental_errors_current{module_id="sysctl"} 0
# HELP cilium_reconciler_incremental_errors_total Total number of errors encountered during incremental reconciliation
# TYPE cilium_reconciler_incremental_errors_total counter
cilium_reconciler_incremental_errors_total{module_id="iptables"} 4
cilium_reconciler_incremental_errors_total{module_id="sysctl"} 4
# HELP cilium_reconciler_incremental_total Number of incremental reconciliations performed
# TYPE cilium_reconciler_incremental_total counter
cilium_reconciler_incremental_total{module_id="iptables"} 4
cilium_reconciler_incremental_total{module_id="sysctl"} 15
```

### DIFF
--- a/pkg/reconciler/example/iptables.go
+++ b/pkg/reconciler/example/iptables.go
@@ -55,13 +55,13 @@ type iptablesTarget struct {
 }
 
 // Delete implements reconciler.Target
-func (*iptablesTarget) Delete(_ context.Context, rule *Rule) error {
+func (*iptablesTarget) Delete(_ context.Context, txn statedb.ReadTxn, rule *Rule) error {
 	fmt.Printf(">>> iptables %s\n", strings.Join(rule.ToArgs("-D"), " "))
 	return nil
 }
 
 // Sync implements reconciler.Target
-func (*iptablesTarget) Sync(_ context.Context, _ statedb.Iterator[*Rule]) (outOfSync bool, err error) {
+func (*iptablesTarget) Sync(_ context.Context, txn statedb.ReadTxn, _ statedb.Iterator[*Rule]) (outOfSync bool, err error) {
 	// TODO:
 	// - "iptables-save" and parse? Or "-C" on each rule?
 	// - Check that cilium chains exist
@@ -70,7 +70,7 @@ func (*iptablesTarget) Sync(_ context.Context, _ statedb.Iterator[*Rule]) (outOf
 }
 
 // Update implements reconciler.Target
-func (*iptablesTarget) Update(_ context.Context, rule *Rule) error {
+func (*iptablesTarget) Update(_ context.Context, txn statedb.ReadTxn, rule *Rule) error {
 	fmt.Printf(">>> iptables %s\n", strings.Join(rule.ToArgs("-A"), " "))
 	return errors.New("oops iptables error")
 }
--- a/pkg/reconciler/example/iptables.go
+++ b/pkg/reconciler/example/iptables.go
@@ -60,19 +60,14 @@ func (*iptablesTarget) Delete(_ context.Context, txn statedb.ReadTxn, rule *Rule
 	return nil
 }
 
-// Sync implements reconciler.Target
-func (*iptablesTarget) Sync(_ context.Context, txn statedb.ReadTxn, _ statedb.Iterator[*Rule]) (outOfSync bool, err error) {
-	// TODO:
-	// - "iptables-save" and parse? Or "-C" on each rule?
-	// - Check that cilium chains exist
-	// - Find unexpected rules in cilium chains?
-	return false, nil
+// Update implements reconciler.Target
+func (*iptablesTarget) Update(_ context.Context, txn statedb.ReadTxn, rule *Rule) (bool, error) {
+	fmt.Printf(">>> iptables %s\n", strings.Join(rule.ToArgs("-A"), " "))
+	return true, errors.New("oops iptables error")
 }
 
-// Update implements reconciler.Target
-func (*iptablesTarget) Update(_ context.Context, txn statedb.ReadTxn, rule *Rule) error {
-	fmt.Printf(">>> iptables %s\n", strings.Join(rule.ToArgs("-A"), " "))
-	return errors.New("oops iptables error")
+func (*iptablesTarget) Prune(context.Context, statedb.ReadTxn, statedb.Iterator[*Rule]) error {
+	return nil
 }
 
 func (*iptablesTarget) Init(context.Context) error {
--- a/pkg/reconciler/example/iptables.go
+++ b/pkg/reconciler/example/iptables.go
@@ -1,0 +1,297 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/netip"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/cilium/cilium/pkg/hive/cell"
+	"github.com/cilium/cilium/pkg/reconciler"
+	"github.com/cilium/cilium/pkg/statedb"
+	"github.com/cilium/cilium/pkg/statedb/index"
+)
+
+var (
+	RulePrimaryIndex = statedb.Index[*Rule, *Rule]{
+		Name: "primary",
+		FromObject: func(rule *Rule) index.KeySet {
+			return index.NewKeySet(rule.PrimaryKey())
+		},
+		FromKey: (*Rule).PrimaryKey,
+		Unique:  true,
+	}
+	RuleStatusIndex = reconciler.NewStatusIndex[*Rule]()
+)
+
+var IPTablesCell = cell.Module(
+	"iptables",
+	"IPTables desired state and reconciliation",
+
+	statedb.NewTableCell[*Rule]("iptables-rules", RulePrimaryIndex, RuleStatusIndex),
+
+	cell.ProvidePrivate(
+		func() reconciler.Target[*Rule] { return &iptablesTarget{} },
+		func() reconciler.Config {
+			return reconciler.Config{
+				FullReconcilationInterval: 10 * time.Second,
+				RetryBackoffMinDuration:   time.Second,
+				RetryBackoffMaxDuration:   10 * time.Second,
+			}
+		},
+		func() statedb.Index[*Rule, reconciler.StatusKind] { return RuleStatusIndex },
+	),
+	cell.Invoke(reconciler.Register[*Rule]),
+)
+
+//
+// iptables reconciliation target
+//
+
+type iptablesTarget struct {
+}
+
+// Delete implements reconciler.Target
+func (*iptablesTarget) Delete(_ context.Context, rule *Rule) error {
+	fmt.Printf(">>> iptables %s\n", strings.Join(rule.ToArgs("-D"), " "))
+	return nil
+}
+
+// Sync implements reconciler.Target
+func (*iptablesTarget) Sync(_ context.Context, _ statedb.Iterator[*Rule]) (outOfSync bool, err error) {
+	// TODO:
+	// - "iptables-save" and parse? Or "-C" on each rule?
+	// - Check that cilium chains exist
+	// - Find unexpected rules in cilium chains?
+	return false, nil
+}
+
+// Update implements reconciler.Target
+func (*iptablesTarget) Update(_ context.Context, rule *Rule) error {
+	fmt.Printf(">>> iptables %s\n", strings.Join(rule.ToArgs("-A"), " "))
+	return errors.New("oops iptables error")
+}
+
+func (*iptablesTarget) Init(context.Context) error {
+	// TODO:
+	// - Load kernel modules.
+	// - Probe for features.
+	// - Setup cilium chains.
+	return nil
+}
+
+var _ reconciler.Target[*Rule] = &iptablesTarget{}
+
+//
+// Rule
+//
+
+type Rule struct {
+	TableChain TableChain
+	IPv6       bool
+	Comment    string
+	Args       []ToArgs
+	Jump       Jump
+	Status     reconciler.Status
+}
+
+func (rule *Rule) GetStatus() reconciler.Status { return rule.Status }
+
+func (rule *Rule) WithStatus(new reconciler.Status) *Rule {
+	ruleCopy := *rule
+	ruleCopy.Status = new
+	return &ruleCopy
+}
+
+func (rule *Rule) PrimaryKey() index.Key {
+	// TODO: This is pretty expensive. Could the "Comment" make for a reasonable
+	// key?
+	return index.String(strings.Join(rule.ToArgs("X"), " "))
+}
+
+func (rule *Rule) ToArgs(action string) []string {
+	args := []string{
+		"-t", rule.TableChain.Table,
+		action, rule.TableChain.Chain,
+		"-m", "comment", "--comment", rule.Comment,
+	}
+	for _, a := range rule.Args {
+		args = append(args, a.ToArgs()...)
+	}
+	return append(args, rule.Jump.ToArgs()...)
+}
+
+//
+// Jumps
+//
+
+type Jump struct {
+	Target string
+	Args   []string
+}
+
+func (j Jump) ToArgs() (args []string) {
+	args = append(args, "-j", j.Target)
+	return append(args, j.Args...)
+}
+
+var (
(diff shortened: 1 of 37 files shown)


pkg/reconciler/example/sysctl.go
```diff
@@ -0,0 +1,142 @@
...
+
+var (
+	SysctlKeyIndex = statedb.Index[*SysctlSetting, string]{
+		Name: "key",
+		FromObject: func(s *SysctlSetting) index.KeySet {
+			return index.NewKeySet(index.String(s.Key))
+		},
+		FromKey: index.String,
+		Unique:  true,
+	}
```
aanm, line 98: Since this is an example, can we have comments for each field of this struct initialization for better understanding?

pkg/reconciler/example/sysctl.go
```diff
@@ -0,0 +1,142 @@
...
+	SysctlStatusIndex = reconciler.NewStatusIndex[*SysctlSetting]()
```
aanm, line 99: Can we also have a comment of why we need an "indexer"? Or why do we need this indexer?

pkg/reconciler/reconciler.go
```diff
@@ -0,0 +1,262 @@
...
+	ctx context.Context,
+	txn statedb.ReadTxn,
+	lastRev statedb.Revision,
+) (statedb.Revision, <-chan struct{}, error) {
+	start := time.Now()
+
+	// In order to not lock the table while doing potentially expensive operations,
+	// we collect the reconciliation statuses for the objects and then commit them
+	// afterwards. If the object has changed in the meanwhile the status update for
+	// the old object is skipped.
```
aanm, line 168: The reconciliation status is **per object** does this mean that if the commit fails it will fail for all objects or just for the object that has changed?

pkg/reconciler/reconciler.go
```diff
@@ -0,0 +1,262 @@
...
+		if status.Kind == StatusKindDone {
+			if len(errs) == 0 {
+				lastSuccessRev = rev
+			}
+			continue
+		}
+
+		var err error
+		if status.Delete {
+			err = r.Target.Delete(obj)
```
aanm: Please add a comment in this line stating that this is the business logic for "deletes". It was not obvious on a first sight.

pkg/reconciler/reconciler.go
```diff
@@ -0,0 +1,262 @@
...
+			if err == nil {
+				toBeDeleted[obj] = rev
+			} else {
+				updateResults[obj] = result{rev, StatusError(true, err)}
+			}
+		} else {
+			err = r.Target.Update(obj)
```
aanm: Please add a comment in this line stating that this is the business logic for "updates". It was not obvious on a first sight.

pkg/reconciler/reconciler.go
```diff
@@ -0,0 +1,262 @@
...
+
+	// lastSuccessRev is the highest revision that was successfully reconciled
+	// without any failures prior to it. This revision will be used in the next
+	// reconciliation round.
+	lastSuccessRev := statedb.Revision(lastRev)
+	var errs []error
+
+	// Iterate in revision order through new, changed or failed objects.
+	iter, watch := r.Table.LowerBound(txn, statedb.ByRevision[Obj](lastRev+1))
+	for obj, rev, ok := iter.Next(); ok; obj, rev, ok = iter.Next() {
```
aanm, line 180: Theoretically, each object has a different revision number?

pkg/reconciler/example/sysctl.go
```diff
@@ -0,0 +1,142 @@
...
+		oldValue, err = io.ReadAll(f)
+
+		if string(oldValue) != obj.Value {
+			outOfSync = true
+			f.Seek(0, 0)
+			f.Truncate(int64(len(obj.Value)))
+			f.WriteString(obj.Value)
+			f.Close()
+		}
+	}
```
aanm, line 151: two questions:
1. Why don't we store the "result" of each individual item in it? Similarly to what with do in the reconciliation with the updateResults?
2. Where are the "deletes"? If we are syncing the desired state, shouldn't we also support "deletes"?

pkg/reconciler/example/sysctl.go
```diff
@@ -0,0 +1,142 @@
...
+
+// Sync implements reconciler.Target
+func (t *sysctlTestTarget) Sync(iter statedb.Iterator[*SysctlSetting]) (outOfSync bool, err error) {
+	if err := maybeError(); err != nil {
+		return false, err
+	}
+
+	for obj, _, ok := iter.Next(); ok; obj, _, ok = iter.Next() {
+		var f *os.File
+		f, err = fakeProcFile(obj)
```
aanm, line 136: This is "reading" the current realized state?

pkg/reconciler/example/sysctl.go
```diff
@@ -0,0 +1,142 @@
...
+		if err != nil {
+			return
+		}
+
+		var oldValue []byte
+		oldValue, err = io.ReadAll(f)
+
+		if string(oldValue) != obj.Value {
```
aanm, line 144: Is this "comparing" the desired vs realized state? Shouldn't we have an interface, and use that interface, it so that all objects in the iterator implement?

pkg/reconciler/example/main.go
```diff
@@ -51,23 +58,33 @@ func modify(db *statedb.DB, t statedb.RWTable[*SysctlSetting]) {
...
+
+			// Wait until the new values are reconciled.
+			s.Wait(context.TODO())
+
 		}
 	}()
 }
 
+func modifyIPTables(db *statedb.DB, t statedb.RWTable[*Rule]) {
+	go func() {
```
aanm, line 95: why a seperate go routine?

pkg/reconciler/example/iptables.go
```diff
@@ -0,0 +1,297 @@
...
+	return nil
+}
+
+// Sync implements reconciler.Target
+func (*iptablesTarget) Sync(_ context.Context, _ statedb.Iterator[*Rule]) (outOfSync bool, err error) {
+	// TODO:
+	// - "iptables-save" and parse? Or "-C" on each rule?
+	// - Check that cilium chains exist
+	// - Find unexpected rules in cilium chains?
+	return false, nil
```
aanm: Ah, I was really curious to see this one :sweat_smile:

pkg/reconciler/example/iptables.go
```diff
@@ -0,0 +1,297 @@
...
+	// - Probe for features.
+	// - Setup cilium chains.
+	return nil
+}
+
+var _ reconciler.Target[*Rule] = &iptablesTarget{}
+
+//
+// Rule
+//
```
aanm, line 85: Do we need to define all types of "rules" "jump"? Can't the state be the full iptables command? For example, the desired state is `-A POSTROUTING -s 172.17.0.0/16 ! -o docker0 -j MASQUERADE` and the realized state is basically a execution of `iptables -A POSTROUTING -s 172.17.0.0/16 ! -o docker0 -j MASQUERADE`.

pkg/reconciler/example/route.go
```diff
@@ -0,0 +1,419 @@
...
+// Sync implements reconciler.Target
+func (t *routeTestTarget) Sync(ctx context.Context, txn statedb.ReadTxn, iter statedb.Iterator[*DesiredRoute]) (outOfSync bool, err error) {
+	// TODO: Might want to trigger sync when devices change their state.
+	// (e.g. go UP from DOWN).
+	fmt.Printf("Sync\n")
+
+	var errs []error
+
+	for desired, _, ok := iter.Next(); ok; desired, _, ok = iter.Next() {
+		actual, _, ok := t.routes.First(txn, tables.RouteIDIndex.Query(desired.RouteID()))
```
aanm: Why "First"?

pkg/reconciler/example/route.go
```diff
@@ -0,0 +1,419 @@
+package main
```
aanm: The routes reconciler seems to be the hardest to understand because I couldn't find the "reconciliation" logic to fetch the realized state and do the comparison between the desired and the realized state. The sysctl example was easier on this aspect.

aanm (changes requested):
I've reviewed on a per-commit basis.

For some of my comments please leave the reply in the source code as well since people that will use the code examples as a base won't have the same questions.

pkg/reconciler/example/route.go
joamaki: Because we’re only interested in the first match. StateDB supports both unique and non-unique indexes and doesn’t differentiate the query API for these.

pkg/reconciler/example/route.go
joamaki: Look again, it’s looking up the ”actual” route from the route table maintained by the devices controller (which pulls routes over netlink). This way it can efficiently reconcile a lot of routes without having to do individual syscalls to fetch routes one by one.

But indeed, it makes it harder to follow as it depends on this outside realized state table that’s populated elsewhere. The final version will need to document this well. Especially how it deals with acting on potentially stale ”realized state” data (that’s where full reconciliation is needed).

pkg/reconciler/example/iptables.go
joamaki: I think I shouldn’t have left this very unfinished example in as there’s too many open questions in it..

pkg/reconciler/example/sysctl.go
joamaki: Yes it’s comparing against realized state. I don’t understand what you mean about interfaces and iterators. BTW, refactored the full reconciliation logic fair bit, so maybe I already addressed this?

pkg/reconciler/reconciler.go
joamaki: Just for the object. CompareAndSwap does nothing if the revision doesn’t match

pkg/reconciler/reconciler.go
joamaki: Yes, every change to an object (including deletes if tracked with DeleteTracker) are assigned a new revision.

pkg/reconciler/example/sysctl.go
joamaki: 1) I don’t understand the question. What is result and what is ”it”?

2) we do support deletes, but it is not defined for sysctl/proc.

pkg/reconciler/example/main.go
joamaki: Because one can’t block in Invoke. I’ll make this example proper by using pkg/hive/job for these.

pkg/reconciler/example/sysctl.go
joamaki: Yes definitely! This is still a very messy draft so many things like that missing. I’ll make sure whatever the final ”canonical example” ends up being will include these.

pkg/reconciler/example/sysctl.go
joamaki: Yes! Also made me realize that the Index struct itself needs much better docs.

PR #28281, opened 2023-09-26
docs: Remove the duplicated envoy resource list

sayboras:
This is to reduce the maintenance effort as we are already referencing to cilium/proxy repo.

Suggested-by: Anna Kapuscinska <anna@isovalent.com>


learnitall (approved):
Great idea!

meyskens (approved):
Less maintenance work is instant approval

PR #28290, opened 2023-09-26
Add a new option to exclude unwanted k8s node labels from CiliumNode

hemanthmalla:
Updates to CiliumNode object can trigger expensive operations like `InjectLabels()` on the ipcache subsystem. Since updates to cilium node objects are sent to all nodes in the cluster, these unwanted updates can end up wasting quite some CPU cycles.

`exclude-node-label-patterns` allows for dropping unwanted labels added by other controllers from being propagated to CiliumNode CRD and data stored in kvstore under `cilium/state/nodes/v1/*`

### DIFF
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -188,6 +188,9 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	flags.Bool(option.AutoCreateCiliumNodeResource, defaults.AutoCreateCiliumNodeResource, "Automatically create CiliumNode resource for own node on startup")
 	option.BindEnv(vp, option.AutoCreateCiliumNodeResource)
 
+	flags.StringSlice(option.ExcludeNodeLabelPatterns, []string{}, "List of k8s node label regex patterns to be excluded from CiliumNode")
+	option.BindEnv(vp, option.ExcludeNodeLabelPatterns)
+
 	flags.String(option.BPFRoot, "", "Path to BPF filesystem")
 	option.BindEnv(vp, option.BPFRoot)
 
--- a/pkg/labelsfilter/filter.go
+++ b/pkg/labelsfilter/filter.go
@@ -338,3 +338,23 @@ func Filter(lbls labels.Labels) (identityLabels, informationLabels labels.Labels
 func FilterNodeLabels(lbls labels.Labels) (identityLabels, informationLabels labels.Labels) {
 	return validNodeLabelPrefixes.filterLabels(lbls)
 }
+
+func FilterLabelsByRegex(excludePatterns []*regexp.Regexp, labels map[string]string) map[string]string {
+	if len(excludePatterns) == 0 && labels != nil {
+		return labels
+	}
+	newLabels := make(map[string]string)
+	for k, v := range labels {
+		labelNeedsExclusion := false
+		for _, pattern := range excludePatterns {
+			if pattern.MatchString(k) {
+				labelNeedsExclusion = true
+				break
+			}
+		}
+		if !labelNeedsExclusion {
+			newLabels[k] = v
+		}
+	}
+	return newLabels
+}
--- a/pkg/labelsfilter/filter_test.go
+++ b/pkg/labelsfilter/filter_test.go
@@ -4,6 +4,8 @@
 package labelsfilter
 
 import (
+	"reflect"
+	"regexp"
 	"testing"
 
 	. "github.com/cilium/checkmate"
@@ -162,4 +164,68 @@ func (s *LabelsPrefCfgSuite) TestFilterLabelsDocExample(c *C) {
 	filtered, _ = dlpcfg.filterLabels(allLabels)
 	c.Assert(len(filtered), Equals, 6)
 	c.Assert(filtered, checker.DeepEquals, wanted)
+}
+
+func TestFilterLabelsByRegex(t *testing.T) {
+	type args struct {
+		excludePatterns []*regexp.Regexp
+		labels          map[string]string
+	}
+	tests := []struct {
+		name string
+		args args
+		want map[string]string
+	}{
+		{
+			name: "exclude_test",
+			args: args{
+				[]*regexp.Regexp{regexp.MustCompile("foobar.*")},
+				map[string]string{
+					"topology.kubernetes.io/region": "us-east-1",
+					"foobar.com":                    "unwanted-label",
+				},
+			},
+			want: map[string]string{
+				"topology.kubernetes.io/region": "us-east-1",
+			},
+		},
+		{
+			name: "multi_exclude_test",
+			args: args{
+				[]*regexp.Regexp{
+					regexp.MustCompile("foo.*"),
+					regexp.MustCompile("bar.*"),
+				},
+				map[string]string{
+					"topology.kubernetes.io/region": "us-east-1",
+					"foo.com":                       "unwanted-label",
+					"bar.com":                       "unwanted-label",
+				},
+			},
+			want: map[string]string{
+				"topology.kubernetes.io/region": "us-east-1",
+			},
+		},
+		{
+			name: "baseline_test",
+			args: args{
+				[]*regexp.Regexp{},
+				map[string]string{
+					"topology.kubernetes.io/region": "us-east-1",
+					"foobar.com":                    "unwanted-label",
+				},
+			},
+			want: map[string]string{
+				"topology.kubernetes.io/region": "us-east-1",
+				"foobar.com":                    "unwanted-label",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := FilterLabelsByRegex(tt.args.excludePatterns, tt.args.labels); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("FilterLabelsByRegex() = %v, want %v", got, tt.want)
+			}
+		})
+	}
 }
--- a/pkg/option/config.go
+++ b/pkg/option/config.go
@@ -14,6 +14,7 @@ import (
 	"net/netip"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"sort"
 	"strconv"
@@ -883,6 +884,10 @@ const (
 	// AutoCreateCiliumNodeResource enables automatic creation of a
 	// CiliumNode resource for the local node
 	AutoCreateCiliumNodeResource = "auto-create-cilium-node-resource"
+
+	// ExcludeNodeLabelPatterns allows for excluding unnecessary labels from being propagated from k8s node to cilium
+	// node object. This allows for avoiding unnecessary events being broadcast to all nodes in the cluster.
+	ExcludeNodeLabelPatterns = "exclude-node-label-patterns"
 
 	// IPv4NativeRoutingCIDR describes a v4 CIDR in which pod IPs are routable
 	IPv4NativeRoutingCIDR = "ipv4-native-routing-cidr"
@@ -2091,6 +2096,10 @@ type DaemonConfig struct {
 	// AutoCreateCiliumNodeResource enables automatic creation of a
 	// CiliumNode resource for the local node
 	AutoCreateCiliumNodeResource bool
+
+	// ExcludeNodeLabelPatterns allows for excluding unnecessary labels from being propagated from k8s node to cilium
+	// node object. This allows for avoiding unnecessary events being broadcast to all nodes in the cluster.
+	ExcludeNodeLabelPatterns []*regexp.Regexp
 
 	// IPv4NativeRoutingCIDR describes a CIDR in which pod IPs are routable
 	IPv4NativeRoutingCIDR *cidr.CIDR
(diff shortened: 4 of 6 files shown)


daemon/cmd/daemon_main.go
```diff
@@ -185,6 +185,9 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	flags.Bool(option.AutoCreateCiliumNodeResource, defaults.AutoCreateCiliumNodeResource, "Automatically create CiliumNode resource for own node on startup")
 	option.BindEnv(vp, option.AutoCreateCiliumNodeResource)
 
+	flags.StringSlice(option.ExcludeNodeLabelPatterns, []string{}, "List of k8s node label regex patterns to be excluded from CiliumNode")
+	option.BindEnv(vp, option.ExcludeNodeLabelPatterns)
```
giorio94, line 192: Nodes are additionally synchronized in the kvstore by the operator [here](https://github.com/cilium/cilium/blob/main/operator/cmd/cilium_node.go#L164-L167). My understanding is that we should additionally register this flag in the operator as well, to ensure consistency if the labels are filtered.

giorio94 (changes requested):
Overall changes look good to me. Just one comment inline to ensure consistency between the agent and the operator.

christarazi (commented):
I haven't reviewed the PR yet, but at a high-level, could we use `--labels` instead? https://docs.cilium.io/en/stable/gettingstarted/terminology/#id4

hemanthmalla:
@christarazi do you mean retain only identity relevant labels on CiliumNode ? Are we sure there aren't features elsewhere replying on specific labels ? I remember seeing custom labels / annotations being used in BGP features. But I don't know the specifics.

christarazi:
cc @cilium/sig-bgp ^

YutaroHayakawa:
Can I assume the change introduces a feature to move specific labels from the `.metadata.labels` of the `CiliumNode` CRD?

We rely on the labels of the `CiliumNode` to select the node to apply BGP policy but with a user-defined selector. So, users are responsible for not excluding the labels used for the selector. It applies to any resources selects node by labels, like `CiliumNetworkPolicy` (host firewall), `CiliumNodeConfig` and so on.

pkg/labelsfilter/filter.go
```diff
@@ -308,3 +308,20 @@ func (cfg *labelPrefixCfg) filterLabels(lbls labels.Labels) (identityLabels, inf
 func Filter(lbls labels.Labels) (identityLabels, informationLabels labels.Labels) {
 	return validLabelPrefixes.filterLabels(lbls)
 }
+
+func FilterLabelsByRegex(excludePatterns []*regexp.Regexp, labels map[string]string) map[string]string {
+	newLabels := make(map[string]string)
```
derailed, line 346: if excludePatterns is empty we should just return the labels directly.

pkg/labelsfilter/filter_test.go
```diff
@@ -162,3 +164,36 @@ func (s *LabelsPrefCfgSuite) TestFilterLabelsDocExample(c *C) {
 	c.Assert(len(filtered), Equals, 6)
 	c.Assert(filtered, checker.DeepEquals, wanted)
 }
+
+func TestFilterLabelsByRegex(t *testing.T) {
+	type args struct {
+		excludePatterns []*regexp.Regexp
+		labels          map[string]string
+	}
+	tests := []struct {
```
derailed, line 174: We could use a map here and leverage the use case name as the key.
Additionally we are only testing the happy path.

pkg/option/config.go
```diff
@@ -2500,6 +2509,7 @@ var (
 		AnnotateK8sNode:                 defaults.AnnotateK8sNode,
 		K8sServiceCacheSize:             defaults.K8sServiceCacheSize,
 		AutoCreateCiliumNodeResource:    defaults.AutoCreateCiliumNodeResource,
+		ExcludeNodeLabelPatterns:        []*regexp.Regexp{},
```
derailed: Could this just be the default i.e nil slice?

pkg/option/config.go
```diff
@@ -3625,7 +3635,19 @@ func (c *DaemonConfig) Populate(vp *viper.Viper) {
...
 	c.EnableK8sNetworkPolicy = vp.GetBool(EnableK8sNetworkPolicy)
+
 	c.PolicyCIDRMatchMode = vp.GetStringSlice(PolicyCIDRMatchMode)
+
+	// Parse node label patterns
+	nodeLabelPatterns := vp.GetStringSlice(ExcludeNodeLabelPatterns)
+	for _, pattern := range nodeLabelPatterns {
+		r, err := regexp.Compile(pattern)
+		if err != nil {
+			log.WithError(err).Errorf("Unable to compile regex pattern %s", pattern)
```
derailed: nit: Could we add a bit more context to this error ie which regex pattern are we referirng to?

derailed (commented):
@hemanthmalla Nice work! Just a few picks. 
Additionally I wonder if there are implications dropping labels that could be required downstream by other subsystems??

oblazek:
> I haven't reviewed the PR yet, but at a high-level, could we use `--labels` instead? https://docs.cilium.io/en/stable/gettingstarted/terminology/#id4

I am doing something similar in my PR https://github.com/cilium/cilium/pull/26924 (using `--node-labels`). Seems to me that we could really just include only specific labels (security relevant), instead of excluding a lot of unnecessary ones. Thoughts?

hemanthmalla:
This one slipped through cracks, opening it again since it is still relevant. Will rebase and address conflicts and comments shortly.

hemanthmalla:
> Seems to me that we could really just include only specific labels (security relevant), instead of excluding a lot of unnecessary ones

@oblazek we do have some subsystems relying on labels that aren't necessarily used for identity. See @YutaroHayakawa's comment above https://github.com/cilium/cilium/pull/28290#issuecomment-1738397992

> Additionally I wonder if there are implications dropping labels that could be required downstream by other subsystems??

@derailed that's the reason I wanted to make this an opt-in flag so that operators can consciously make that call.

daemon/cmd/daemon_main.go
hemanthmalla: This is addressed now.

pkg/labelsfilter/filter_test.go
hemanthmalla: Good catch. Added a new case for testing the baseline.
The array pattern seems to be a standard everywhere else though ? My tests follow the same pattern.

pkg/labelsfilter/filter.go
hemanthmalla: Updated code to address this.

pkg/option/config.go
hemanthmalla: Updated the log to include more details

pkg/option/config.go
hemanthmalla: Makes sense. Removed this line now.

pkg/node/types/node.go
```diff
@@ -65,12 +66,13 @@ func ParseCiliumNode(n *ciliumv2.CiliumNode) (node Node) {
 		Cluster:         option.Config.ClusterName,
 		ClusterID:       option.Config.ClusterID,
 		Source:          source.CustomResource,
-		Labels:          n.ObjectMeta.Labels,
 		Annotations:     n.ObjectMeta.Annotations,
 		NodeIdentity:    uint32(n.Spec.NodeIdentity),
 		WireguardPubKey: wireguardPubKey,
 	}
 
+	node.Labels = labelsfilter.FilterLabelsByRegex(option.Config.ExcludeNodeLabelPatterns, n.ObjectMeta.Labels)
```
giorio94: Upon a further analysis, this filter operation doesn't seem to be actually necessary to me. IIUC, your goal is to filter unnecessary labels in the `Node` to `CiliumNode` conversion phase (so that the `CiliumNode` only contains the relevant subset), while this function performs the opposite conversion. The same applies also to the operator flags, as the operator appears to do the synchronization based on `CiliumNodes` (sorry, I mistakenly assumed that synchronization logic was leveraging `Node` objects while writing my previous comment). Does that match your understanding?

pkg/nodediscovery/nodediscovery.go
```diff
@@ -171,6 +172,10 @@ func (n *NodeDiscovery) JoinCluster(nodeName string) error {
 		ln.Labels = maps.Clone(ln.Labels)
 		maps.Copy(ln.Labels, resp.Labels)
 
+		// Filter out ignored node labels to make sure old data from kvstore doesn't make it back to CiliumNode
+		// when exclude node label patterns are changed.
+		ln.Labels = labelsfilter.FilterLabelsByRegex(option.Config.ExcludeNodeLabelPatterns, ln.Labels)
+
```
giorio94: This is not required as this code path is executed only for external-workloads (i.e., agents which are executed outside of k8s), and they don't inherit the labels from a `Node` resource.

giorio94 (commented):
Thanks, just a couple of comments inline.

pkg/node/types/node.go
hemanthmalla: So, I also wanted to update data stored in kvstore under `cilium/state/nodes/v1/*`
IIRC from few months ago, I remember adding in more places to make sure exclude patterns are applied to kvstore as well. I can re-test this on a real cluster and get back to you on this.

giorio94: Thanks! The logic to update the CiliumNode resource and the equivalent kvstore representation got [recently refactored and uniformed](https://github.com/cilium/cilium/pull/29319), and they now both directly retrieve the source of truth from the [LocalNodeStore updates](https://github.com/cilium/cilium/blob/0554ca68117599dae110ff1e9c7d01731d534f2e/pkg/nodediscovery/nodediscovery.go#L233-L235), which sources them from the [corresponding k8s Node resource](https://github.com/cilium/cilium/blob/0554ca68117599dae110ff1e9c7d01731d534f2e/daemon/cmd/local_node_sync.go#L226-L230).

pkg/nodediscovery/nodediscovery.go
hemanthmalla: You're right, removing this.

pkg/node/types/node.go
christarazi: Given Marco's comment, should the filtration of the labels be in the "ingestion" layer here as soon as the object is received or should it be somewhere that's more clearly the "processing" layer? I think it would be useful for debugging if Cilium knew what the original event was and what the processed event is.

@giorio94, WDYT? Do you have thoughts on where this code should be?

giorio94: I'm personally in favor of keeping the filtering logic as part of the `ParseNode` function, both for consistency with the already existing filtering logic for annotations and to make sure that all consumers observe the same object version. Otherwise, we risk inconsistencies between let's say the propagation of the information into the CiliumNode and the corresponding kvstore representation, which already occurred in the past and caused subtle bouncing issues. I'd not be particularly concerned from the debugging perspective, given that this would be an opt in flag meant mainly for fine tuning, and it would perform a pretty straightforward filtering operation.

christarazi (approved):

derailed (approved):
@hemanthmalla Thank you for the updates!

pkg/node/types/node.go
hemanthmalla: @giorio94 updated the PR to invoke the filtering logic only on `ParseNode`. Removed the flag and doc updates from the operator as well.

pkg/labelsfilter/filter_test.go
```diff
@@ -163,3 +165,50 @@ func (s *LabelsPrefCfgSuite) TestFilterLabelsDocExample(c *C) {
...
```
giorio94, line 174: Nit. Could make sense to additionally cover the case of multiple filters.

pkg/option/config.go
```diff
@@ -3539,6 +3548,17 @@ func (c *DaemonConfig) Populate(vp *viper.Viper) {
...
 	c.EnableNodeSelectorLabels = vp.GetBool(EnableNodeSelectorLabels)
 	c.NodeLabels = vp.GetStringSlice(NodeLabels)
+
+	// Parse node label patterns
+	nodeLabelPatterns := vp.GetStringSlice(ExcludeNodeLabelPatterns)
+	for _, pattern := range nodeLabelPatterns {
+		r, err := regexp.Compile(pattern)
+		if err != nil {
+			log.WithError(err).Errorf("Unable to compile exclude node label regex pattern %s", pattern)
+			continue
```
giorio94, line 3558: I'm wondering whether we should bail out hard in case the regex cannot be compiled, to make sure that users can easily catch broken configurations.

giorio94 (approved):
Thanks, lgtm! I've just left a couple of final non-blocking comments inline.

pkg/option/config.go
hemanthmalla: Hmm. I feel like an error level log should provide enough signal to cluster operators ? Since this flag is effectively only an optimization to avoid unnecessary writes, this should be enough ?

pkg/labelsfilter/filter_test.go
hemanthmalla: ACK, will add one more test.

hemanthmalla: Done

giorio94 (approved):

PR #28294, opened 2023-09-26
gateway-api: Fix double slash when trying to strip path prefix

### LINKED ISSUE #28270: Gateway API HttpRoute cannot strip path prefix - double slash problem

### What happened?

I've previously created issue #27994 regarding this subject, but it was closed after PR that doesn't really address the issue so I thought I'd open new one to track double slash problem.
While using Gateway API i found out that given `PathPrefix: /echo`, `replacePrefixMatch: /` and request to `{someAddrs}/echo/test` leaves you with `{someAddrs}//test`, so double slash. That makes stripping prefix impossible, which is pretty handy feature (seems pretty fundamental for gateway).
This issue might be out of Cilium hands and could envoy issue. If I'm not mistaken it's envoy that actually does url manipulation and if that's the case, then it doesn't make sure that output url is valid. Still if Cilium accepted empty prefixes I think you could avoid double '/' by making  `replacePrefixMatch: ""` in URLRewrite.
I don't know cilium internals, so let me know if it's something that cilium can fix or should I create issue on envoy side.

### Cilium Version

1.15.0-pre.0

### Kernel Version

Linux 6.5

### Kubernetes Version

1.27

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

nxy7:
- Merge slashes in URI path
- Gateway to CEC translation fix

      GitHub issue.

Stripping path prefix leads to double slashes in URI path. To fix that I've enabled envoy `merge_slashes` option.

Fixes: #28270

```release-note
Fix: Gateway API double slash while stripping path prefix
```

### DIFF
--- a/operator/pkg/model/translation/envoy_virtual_host.go
+++ b/operator/pkg/model/translation/envoy_virtual_host.go
@@ -246,7 +246,7 @@ func envoyHTTPRoutes(httpRoutes []model.HTTPRoute, hostnames []string, hostNameS
 		if hRoutes[0].RequestRedirect != nil {
 			route.Action = getRouteRedirect(hRoutes[0].RequestRedirect, listenerPort)
 		} else {
-			route.Action = getRouteAction(backends, r.Rewrite, r.RequestMirrors)
+			route.Action = getRouteAction(&r, backends, r.Rewrite, r.RequestMirrors)
 		}
 		routes = append(routes, &route)
 		delete(matchBackendMap, r.GetMatchKey())
@@ -268,15 +268,23 @@ func hostRewriteMutation(rewrite *model.HTTPURLRewriteFilter) routeActionMutatio
 	}
 }
 
-func pathPrefixMutation(rewrite *model.HTTPURLRewriteFilter) routeActionMutation {
+func pathPrefixMutation(rewrite *model.HTTPURLRewriteFilter, httpRoute *model.HTTPRoute) routeActionMutation {
 	return func(route *envoy_config_route_v3.Route_Route) *envoy_config_route_v3.Route_Route {
-		if rewrite == nil || rewrite.Path == nil || len(rewrite.Path.Exact) != 0 || len(rewrite.Path.Regex) != 0 {
+		if rewrite == nil || rewrite.Path == nil || httpRoute == nil || len(rewrite.Path.Exact) != 0 || len(rewrite.Path.Regex) != 0 {
 			return route
 		}
-		if len(rewrite.Path.Prefix) == 0 {
-			// Refer to: https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPPathModifier
-			// ReplacePrefix is allowed to be empty.
-			route.Route.PrefixRewrite = "/"
+
+		// Refer to: https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPPathModifier
+		// ReplacePrefix is allowed to be empty.
+		if len(rewrite.Path.Prefix) == 0 || rewrite.Path.Prefix == "/" {
+
+			route.Route.RegexRewrite = &envoy_type_matcher_v3.RegexMatchAndSubstitute{
+				Pattern: &envoy_type_matcher_v3.RegexMatcher{
+					Regex: "^" + httpRoute.PathMatch.Prefix,
+				},
+				Substitution: "",
+			}
+
 		} else {
 			route.Route.PrefixRewrite = rewrite.Path.Prefix
 		}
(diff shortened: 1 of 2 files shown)


nxy7:
I've tested images built by this change and found out that while it can prevent more than two slashes, there's always at least 2 that remain (without this change you can have more than 2). This leads me to believe that one slash is added at cilium level (not envoy as I previously suspected). 
One thing that I think could fix that is slightly altering changes made at 70a19f4 (commit for the previous issue I've made), from `route.Route.PrefixRewrite = "/"` to `route.Route.PrefixRewrite = ""`, but it's possible that this additional slash is added in some other place.

nxy7:
Both `merge_slashes` and `normalize_paths` don't fix the issue, and setting envoy prefix to "" instead of "/" makes envoy ignore whole rewrite (so regardless of settings whole path is passed). I wonder if this PR should keep merge_slashes or normalize_path turned on.

nxy7:
Okay, so I found the root cause of the issue. It is indeed envoy problem. For whatever reason merging slashes only works between path segments (it seems like they are first merging slashes and only then doing path rewrites).
Also, we cannot pass empty rewritePrefix as that's ignored by envoy. The only solution to this issue seems to be using regexp rewrite, that can in fact accept empty prefix.
I've tested ebb5dce and it strips prefixes correctly. To do that I've had to pass `*model.HTTPRoute` to `getRouteAction` and `pathPrefixMutation` because regexp in `pathPrefixMutation` needs to be aware of  path that it's supposed to substitute.
What do you think about that solution @sayboras ?

lmb:
@sayboras can you take a look please?

lmb:
@nxy7 seems like your PR includes a merge of the main branch, please rebase your changes on top of main instead. Also try squashing into as few commits as possible please.

nxy7:
I've kind of made a mess with this git history (sorry for that, I'm not very good at using git), but it should be good now. I've reduced commits to 3 and rebased all changes on top of main.

vendor/sigs.k8s.io/gateway-api/conformance/tests/httproute-rewrite-path.go
```diff
@@ -59,6 +59,18 @@ var HTTPRouteRewritePath = suite.ConformanceTest{
 				Backend:   "infra-backend-v1",
 				Namespace: ns,
 			},
+			{
```
sayboras: is the change in vendor file intentional?

operator/pkg/model/translation/envoy_virtual_host.go
```diff
@@ -313,12 +324,12 @@ func requestMirrorMutation(mirror *model.HTTPRequestMirror) routeActionMutation
 	}
 }
 
-func getRouteAction(backends []model.Backend, rewrite *model.HTTPURLRewriteFilter, mirror *model.HTTPRequestMirror) *envoy_config_route_v3.Route_Route {
+func getRouteAction(backends []model.Backend, rewrite *model.HTTPURLRewriteFilter, mirror *model.HTTPRequestMirror, route *model.HTTPRoute) *envoy_config_route_v3.Route_Route {
```
sayboras: what do you think about having function signature as per below instead? 

```
funct func getRouteAction(route *model.HTTPRoute, backends []model.Backend) *envoy_config_route_v3.Route_Route {
```

operator/pkg/model/translation/envoy_virtual_host.go
```diff
@@ -268,15 +268,26 @@ func hostRewriteMutation(rewrite *model.HTTPURLRewriteFilter) routeActionMutatio
...
-			// Refer to: https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPPathModifier
-			// ReplacePrefix is allowed to be empty.
-			route.Route.PrefixRewrite = "/"
+
+		// Refer to: https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPPathModifier
+		// ReplacePrefix is allowed to be empty.
+		if len(rewrite.Path.Prefix) == 0 || rewrite.Path.Prefix == "/" {
+			if httpRoute == nil {
+				return route
+			}
```
sayboras: this could be merged with the first check.

sayboras (commented):
Thanks for your PR.

Please find below my comments. Also, can you help to merge all commits into one as we are using rebase and merge strategy to keep main history clean.

vendor/sigs.k8s.io/gateway-api/conformance/tests/httproute-rewrite-path.go
nxy7: Yes, it was intentional. I've assumed that this is part of e2e tests, that check conformance to gateway API (as changes in envoy_virtual_host_test check configuration but not behaviour). If I'm not mistaken these changes should check if prefix can be stripped correctly (according to k8s guidelines).

operator/pkg/model/translation/envoy_virtual_host.go
nxy7: Sure, I can rearrange arguments

vendor/sigs.k8s.io/gateway-api/conformance/tests/httproute-rewrite-path.go
sayboras: ah right, if it's the case, you should either upgrade gateway-api in go.mod, or submit the change to upstream gateway-api.

nxy7: Oh, I didn't realize that it's not part of cilium code. Okay, I'll delete my changes here and open PR on gateway-api side.

nxy7:
I've applied all your suggestions @sayboras, hope it's okay now.

sayboras (approved):
Thanks and LGTM :heavy_check_mark:

sayboras:
I tried to resolve the conflict and merge this in, however, it seems like I didn't have permission to push to your fork. 

Can you help to resolve the conflict? It should be quite straightforward like the below https://github.com/cilium/cilium/compare/main...sayboras:cilium:tam/resolve-conflict-28294

nxy7:
Done :-)

sayboras:
Thanks a lot for your help from analysing to sending the patch :1st_place_medal:

PR #28300, opened 2023-09-26
daemon: Deprecate EnableRemoteNodeIdentity

nathanjsweet:
Remote Node Identity is needed for various other features to work or work fully, including EgressGateway and Policies.

There is no benefit to having it turned off.

```release-note
daemon: The option "EnableRemoteNodeIdentity" is now deprecated and will be removed from the  v1.16 release.
```

### DIFF
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -280,6 +280,7 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	option.BindEnv(vp, option.IPv6MCastDevice)
 
 	flags.Bool(option.EnableRemoteNodeIdentity, defaults.EnableRemoteNodeIdentity, "Enable use of remote node identity")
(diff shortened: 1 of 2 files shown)


daemon/cmd/daemon_main.go
```diff
@@ -280,6 +280,7 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 	option.BindEnv(vp, option.IPv6MCastDevice)
 
 	flags.Bool(option.EnableRemoteNodeIdentity, defaults.EnableRemoteNodeIdentity, "Enable use of remote node identity")
+	flags.MarkDeprecated(option.EnableRemoteNodeIdentity, "This option will be removed in v1.16 and will be \"true\" by default.")
```
giorio94: I'd slightly rephrase the end of the message, as  to me  'true by default' would imply that this setting is still configurable. The message in the PR description looks clearer.

giorio94 (approved):
Looks good to me, just one comment inline. It seems that you also need to rerun the automatic generation of the cmdref files to make the CI happy.

PR #28307, opened 2023-09-27
docs: Use host port for serving docs

brb:
Previously, the sphinx-autobuild process running in a Docker reported:

    [I 230927 13:54:55 server:335] Serving on http://0.0.0.0:8000

This could have confused users thinking that the server is accessible through the tcp/8000 on the host. However, it was accessible https://0.0.0.0:9081.

Change the server's listen port to 9081 to avoid the confusion.


learnitall (approved):
Great catch, thank you. The port mismatch confused me when I first came across it.

PR #28308, opened 2023-09-27
loader: attach XDP programs using bpf_link

### LINKED ISSUE #27630: Attach XDP programs using bpf_link instead of netlink
Similar to https://github.com/cilium/cilium/issues/20739 where we implement cgroup attachments using bpf_link, the same can now be done for XDP using https://pkg.go.dev/github.com/cilium/ebpf/link#AttachXDP.

We can pin XDP links to `/sys/fs/bpf/cilium/xdp/<ifname>`. Cilium 1.14 and 1.13 need to be taught to remove a file at this location before attempting to attach XDP programs using netlink.

rgo3:
See individual commits.

The first two commits (prog removal and xdp bpf_link cleanup) will be submitted as a separate PR for backporting to 1.14 and 1.13 so these versions can handle downgrades from bpf_link-enabled Cilium agents. They were kept in this branch since future commits depended on them, and getting this broken up into smaller commits that make sense was already a challenge.

On a high level, this PR:
- attaches XDP progs using bpf_link
- adds per-device bpffs dirs, currently only for 'physical' devs, and only used for XDP links, at e.g. `/sys/fs/bpf/cilium/devices/enp5s0/links/cil_xdp_entry`
- adds a few utility functions to package bpf for working with bpffs and bpf_link pins
- refactors `attachProgram()` into `attachTCProgram()` and `attachXDPProgram()` (this was a minimal change)
- now uses ebpf-go's `link.XDPAttachFlags` as parameters instead of a simple u32
- adds netns-driven tests for attaching XDP progs to veths

This also puts us in a better position for adding tcx support during the next cycle.

Fixes: #27630

### DIFF
--- a/pkg/datapath/loader/netlink_test.go
+++ b/pkg/datapath/loader/netlink_test.go
@@ -35,6 +35,9 @@ func mustTCProgram(t *testing.T) *ebpf.Program {
 	if err != nil {
 		t.Skipf("tc programs not supported: %s", err)
 	}
+	t.Cleanup(func() {
+		p.Close()
+	})
 	return p
 }
 
@@ -50,6 +53,9 @@ func mustXDPProgram(t *testing.T) *ebpf.Program {
 	if err != nil {
 		t.Skipf("xdp programs not supported: %s", err)
 	}
+	t.Cleanup(func() {
+		p.Close()
+	})
 	return p
 }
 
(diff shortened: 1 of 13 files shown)


rgo3:
After the offline code walk with @ti-mo and @lmb last week I'll move this into ready for review.

I've left 2 TODO's open intentionally to be discussed during review, most important probably being the decision around the bpffs linkpath we choose to store pins.

I think that there is a possibility to backport this entire PR (the fix to `maybeUnloadObsoleteXDPPrograms` and the new xdp link code) as that would solve the removing links and their respective bpffs pins when downgrading. If we don't want to backport all of it we should discuss what needs to be backported instead.

pkg/datapath/loader/netlink_test.go
```diff
@@ -446,32 +446,36 @@ func TestAttachProgram(t *testing.T) {
...
-			err = netlink.LinkDel(veth)
-			require.NoError(t, err)
-
-			return nil
-		})
-
-	})
+	// the linkpath makes it a bit cumbersome to test this codepath of attachProgram now
+	// I think we can just remove it in favor of the new tests covering xdp attachements
+	// TODO(rgo3): remove or fix test
```
asauber: Given the tests added in this PR, my vote would be to remove this.

pkg/datapath/loader/xdp.go
```diff
@@ -5,18 +5,30 @@ package loader
...
 
+	"github.com/cilium/cilium/pkg/bpf"
 	"github.com/cilium/cilium/pkg/identity"
 	"github.com/cilium/cilium/pkg/mac"
 	"github.com/cilium/cilium/pkg/option"
 )
 
+// TODO: Clean up bpffs root logic and make this a var.
+func xdpLinkPath() string {
+	return filepath.Join(bpf.CiliumPath(), "loader", "links/xdp") // TODO(rgo3): not sure if we want that exact path, also relevant for cilium cleanup command
```
asauber: For now, these strings should at least be `const`s

cilium/cmd/cleanup.go
```diff
@@ -582,6 +583,10 @@ func removeXDPs(links []netlink.Link) error {
 		if err != nil {
 			return err
 		}
+		err = os.Remove(filepath.Join(bpf.CiliumPath(), "loader", "links/xdp", link.Attrs().Name))
```
asauber: Declared `const`s can be used here rather than string literals.

asauber (changes requested):
Reviewed for Config and CLI. Didn't review the attach/detach logic.

asauber (approved):

ti-mo:
@cilium/tophat Sorry for the noise, you got pinged because pkg/socketlb didn't have a codeowner. I assigned one now.

bimmlerd (approved):
CODEOWNERS change LGTM

ti-mo (approved):

squeed (approved):

dylandreimerink (approved):
LGTM!

julianwiedmann:
> The first two commits (prog removal and xdp bpf_link cleanup) will be submitted as a separate PR for backporting to 1.14 and 1.13 so these versions can handle downgrades from bpf_link-enabled Cilium agents.

@rgo3 If I see things correctly, this happened with https://github.com/cilium/cilium/pull/29105 and https://github.com/cilium/cilium/pull/29104. So the `needs-backport` for those two releases is actually complete?

rgo3:
Yes, we forgot to update the labels unfortunately :disappointed:

PR #28323, opened 2023-09-28
ci: Avoid using deprecated "tunnel" flag

gandro:
The tunnel option is deprecated and will be removed in Cilium v1.15. This commit fixes the remaining uses I have found where the CI explicitly set the old `tunnel` flag.

Note that the Cilium CLI also still sets the flag some times in our CI, this is addressed by cilium/cilium-cli#1993.

### DIFF
--- a/.github/workflows/conformance-clustermesh.yaml
+++ b/.github/workflows/conformance-clustermesh.yaml
@@ -211,7 +211,7 @@ jobs:
             --helm-set=clustermesh.useAPIServer=true \
             "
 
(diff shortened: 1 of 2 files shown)


gandro:
ClusterMesh CI failure looks legitimate. Will need to investigate. Putting back to draft. https://github.com/cilium/cilium/actions/runs/6339393376/job/17218425985

viktor-kurchenko (approved):
LGTM.

PR #28324, opened 2023-09-28
cmd: Disable local node routes when endpoint routes are enabled

gandro:
The command-line description of `enable-endpoint-routes` states: "Use per endpoint routes instead of routing via cilium_host". This description however has not been true: Even with per-endpoint routes enabled, Cilium would still install the local node route (i.e. "routing via cilium_host") unless explicitly disabled.

Not only is the local node route (which installs a single route based on the local pod CIDR) redundant to per-endpoint routes (which installs a route per endpoint), it also does not work on IPAM modes that do not have a local pod CIDR, such as ENI or Azure.

On Azure, GKE, and in the MultiPool CI, we already disabled the local node route explicitly when enabling per-endpoint routes, but in many other cases (ENI, as well as most of our CI (see
github/actions/cilium-config)) we forgot to disable local node routes explicitly when per-endpoint routes are enabled. Therefore, this commit improves UX by automatically disabling the local node route if per-endpoint routes are enabled. This approach was discussed at the Cilium Community meeting on June 7, 2023.

As an alternative to this change, we could also introduce a new tri-mode flag (i.e. "per-endpoint routes", "local node route", "none"), but such a change might introduce unnecessary churn if we decide to remove the local node routes further down the road.

### DIFF
--- a/.github/workflows/conformance-multi-pool.yaml
+++ b/.github/workflows/conformance-multi-pool.yaml
@@ -135,7 +135,6 @@ jobs:
             --helm-set=autoDirectNodeRoutes=true \
             --helm-set=routingMode=native \
             --helm-set=endpointRoutes.enabled=true \
-            --helm-set-string=extraConfig.enable-local-node-route=false \
             --helm-set=kubeProxyReplacement=strict \
             --helm-set=bpf.masquerade=true \
             --helm-set=bpf.hostLegacyRouting=true\
--- a/Documentation/network/concepts/routing.rst
+++ b/Documentation/network/concepts/routing.rst
@@ -296,8 +296,7 @@ The following configuration options must be set to run the datapath on GKE:
   * ``ipam: kubernetes``: Enable :ref:`k8s_hostscope` IPAM
   * ``routing-mode: native``: Enable native routing mode
   * ``enable-endpoint-routes: true``: Enable per-endpoint routing on the node
-  * ``enable-local-node-route: false``: Disable installation of the local node route
-
+    (automatically disables the local node route).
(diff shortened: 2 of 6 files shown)


christarazi (approved):

viktor-kurchenko (approved):

sayboras (approved):
Helm related changes lgttm :heavy_check_mark:

learnitall (changes requested):
Thank you for the detailed PR description. The doc changes you included look great, but there is one extra place that I believe still needs an update- similar to the multi-pool IPAM docs:

https://github.com/cilium/cilium/blob/bb6decf7c7e2e31da91cc94e7d2c050a2d937048/Documentation/network/concepts/routing.rst?plain=1#L288-L299

Can you please update this page and then re-request a review from me?

gandro:
@learnitall I've updated the highlighted section as follows:

```patch
diff --git a/Documentation/network/concepts/routing.rst b/Documentation/network/concepts/routing.rst
index 002407eaa24d..1e912248d477 100644
--- a/Documentation/network/concepts/routing.rst
+++ b/Documentation/network/concepts/routing.rst
@@ -296,7 +296,7 @@ The following configuration options must be set to run the datapath on GKE:
   * ``ipam: kubernetes``: Enable :ref:`k8s_hostscope` IPAM
   * ``routing-mode: native``: Enable native routing mode
   * ``enable-endpoint-routes: true``: Enable per-endpoint routing on the node
-  * ``enable-local-node-route: false``: Disable installation of the local node route
+   (this automatically disables the local node route).
 
 * ``ipv4-native-routing-cidr: x.x.x.x/y``: Set the CIDR in which native routing
   is supported.
```

learnitall (approved):

PR #28325, opened 2023-09-28
bpf: fix flakes when checking metrics map values.

### LINKED ISSUE #28280: CI: [v1.13] BPF unit/integration tests - TestBPF/ipsec_from_lxc_test.o
## CI failure

Hit on https://github.com/cilium/cilium/pull/28251: https://github.com/cilium/cilium/actions/runs/6309322047/job/17129032069?pr=28251
Succeeded when rerun a second time.

```
go test ./bpf_tests -exec sudo -bpf-test-path /home/runner/work/cilium/cilium/bpf/tests 
--- FAIL: TestBPF (3.02s)
    --- FAIL: TestBPF/ipsec_from_lxc_test.o (0.17s)
        bpf_test.go:212: Conntrack lookup 1/2: src=192.168.0.1:22330 dst=192.168.0.2:80
        bpf_test.go:212: Conntrack lookup 2/2: nexthdr=6 flags=1
        bpf_test.go:212: CT verdict: New, revnat=0
        bpf_test.go:212: Successfully mapped addr=192.168.0.2 to identity=233
        bpf_test.go:212: Conntrack create: proxy-port=0 revnat=0 src-identity=1048575 lb=0.0.0.0
        bpf_test.go:212: Unknown message type=197 arg1=74 arg2=65610
        --- FAIL: TestBPF/ipsec_from_lxc_test.o/01_ipv4_from_lxc_no_node_id (0.00s)
            bpf_test.go:421: assert failed at ipsec_from_lxc_test.c:142
        bpf_test.go:212: Conntrack lookup 1/2: src=192.168.0.1:22330 dst=192.168.0.2:80
        bpf_test.go:212: Conntrack lookup 2/2: nexthdr=6 flags=1
        bpf_test.go:212: CT entry found lifetime=296, revnat=0
        bpf_test.go:212: CT verdict: Established, revnat=0
        bpf_test.go:212: Successfully mapped addr=192.168.0.2 to identity=233
        bpf_test.go:212: Local container found ifindex 74 seclabel 18944
        bpf_test.go:212: Conntrack lookup 1/2: src=[::0:1]:22330 dst=[::0:2]:80
        bpf_test.go:212: Conntrack lookup 2/2: nexthdr=6 flags=1
        bpf_test.go:212: CT verdict: New, revnat=0
        bpf_test.go:212: Successfully mapped addr.p4=[::0:2] to identity=233
        bpf_test.go:212: Conntrack create: proxy-port=0 revnat=0 src-identity=1048575
        bpf_test.go:212: Encapsulating to node 94 (0x5e) from seclabel 65630
        bpf_test.go:212: Local container found ifindex 94 seclabel 24064
FAIL
FAIL	github.com/cilium/cilium/test/bpf_tests	3.064s
FAIL
make: *** [Makefile:88: run_bpf_tests] Error 1
make: Leaving directory '/home/runner/work/cilium/cilium/test'
Run 'make -C test run_bpf_tests' locally to investigate failures
```

tommyp1ckles:
The ipsec_from_lxc_test and tc_lxc_policy_drop test files both perform assertions on expected count values stored in the metric map. This results in test flakes as if the map put and map get are performed while running on different cpus then the assertions may fail.

This introduces a new helper macro in tests which iterates possible percpu hash values for a particular key until an expected count is found or we exceed the number of cpu slots.

Fixes: #28280

### DIFF
--- a/bpf/tests/common.h
+++ b/bpf/tests/common.h
@@ -64,6 +64,9 @@
 #define TEST_FAIL 2
 #define TEST_SKIP 3
 
+/* Max number of cpus to check when doing percpu hash assertions */
+#define NR_CPUS 128
+
 /* Use an array map with 1 key and a large value size as buffer to write results */
 /* into. */
 struct {
@@ -271,5 +274,22 @@ _Pragma("unroll")						\
 								\
 	return 0;						\
 }
+
+/* Asserts that the sum of per-cpu metrics map slots for a key equals count */
+#define assert_metrics_count(key, count) \
+({ \
+	struct metrics_value *__entry = NULL; \
+	__u64 sum = 0; \
+	/* Iterate until lookup encounters null when hitting cpu number */ \
+	/* Assumes at most 128 CPUS */ \
+	for (int i = 0; i < NR_CPUS; i++) { \
+		__entry = map_lookup_percpu_elem(&METRICS_MAP, &key, i); \
+		if (!__entry) { \
+			break; \
+		} \
+		sum += __entry->count; \
+	} \
+	assert(sum == count); \
+})
 
 #endif /* ____BPF_TEST_COMMON____ */
--- a/bpf/include/bpf/helpers.h
+++ b/bpf/include/bpf/helpers.h
@@ -35,6 +35,8 @@ static void *BPF_FUNC(map_lookup_elem, const void *map, const void *key);
 static int BPF_FUNC(map_update_elem, const void *map, const void *key,
 		    const void *value, __u32 flags);
 static int BPF_FUNC(map_delete_elem, const void *map, const void *key);
+static void *BPF_FUNC(map_lookup_percpu_elem, void *map, const void *key,
+				unsigned int cpu);
 
 /* Time access */
 static __u64 BPF_FUNC(ktime_get_ns);
--- a/bpf/include/linux/bpf.h
+++ b/bpf/include/linux/bpf.h
@@ -6,8 +6,8 @@
  * modify it under the terms of version 2 of the GNU General Public
  * License as published by the Free Software Foundation.
  */
-#ifndef __LINUX_BPF_H__
-#define __LINUX_BPF_H__
+#ifndef _UAPI__LINUX_BPF_H__
+#define _UAPI__LINUX_BPF_H__
 
 #include <linux/types.h>
 #include <linux/bpf_common.h>
@@ -20,6 +20,7 @@
 
 /* ld/ldx fields */
 #define BPF_DW		0x18	/* double word (64-bit) */
+#define BPF_MEMSX	0x80	/* load with sign extension */
 #define BPF_ATOMIC	0xc0	/* atomic memory ops - op type in immediate */
 #define BPF_XADD	0xc0	/* exclusive add - legacy name */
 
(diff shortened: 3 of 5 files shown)


bpf/tests/common.h
```diff
@@ -272,4 +272,27 @@ _Pragma("unroll")						\
...
+	struct metrics_value *__entry = NULL; \
+	bool found = false; \
+	/* Iterate until lookup encounters null when hitting cpu number */ \
+	/* Assumes at most 128 CPUS */ \
+	for (int i = 0; i < 128; i++) { \
+		__entry = map_lookup_percpu_elem(&METRICS_MAP, &key, i); \
+		if (!__entry) { \
+			break; \
+		} \
+		if (__entry->count == count) { \
```
jschwinger233: Maybe we should accumulate `__entry->count` and compare the `count` with summed up `__entry->count`?

Something like:
```suggestion
__u64 sum;
for (int i = 0; i < 128; i++) {
    __entry = map_lookup_percpu_elem(&METRICS_MAP, &key, i);
    if (!__entry) break;
    sum += __entry->count;
}
if (sum == count) found=true;
```

jschwinger233: I think it depends on the semantics of macro. The macro we create here is `assert_metrics_contains`, so your implementation above is accurate. My suggested code is more like `assert_metrics_count`.

jschwinger233 (commented):
Thanks Tom! Just a small question regarding macro semantic.

bpf/tests/common.h
```diff
@@ -272,4 +272,27 @@ _Pragma("unroll")						\
 	return 0;						\
 }
 
+static void *(*map_lookup_percpu_elem)(void *map, const void *key, unsigned int cpu) = (void *)195;
```
gentoo-root: Any reason not to use BPF_FUNC and put it in bpf/include/bpf/helpers.h?

bpf/tests/common.h
tommyp1ckles: This makes sense, made the change 👍

bpf/tests/common.h
```diff
@@ -272,4 +272,25 @@ _Pragma("unroll")						\
...
+
+/* Asserts that the per-cpu metrics map contains at least one entry that */
+/* contains a specified count given a metrics_key.*/
+#define assert_metrics_count(key, count) \
+({ \
+	struct metrics_value *__entry = NULL; \
+	__u64 sum = 0; \
+	/* Iterate until lookup encounters null when hitting cpu number */ \
+	/* Assumes at most 128 CPUS */ \
+	for (int i = 0; i < 128; i++) { \
```
gentoo-root: Let's maybe `#define NR_CPUS 128` somewhere in a header file? If we can't determine the real number of CPUs, it would still be better to define this constant with the standard name (as used in the kernel), than bury it deep in the code.

tommyp1ckles:
@gentoo-root sorry meant to hit the rerequest review button on grays review but you posted yours at that moment 😅

bpf/tests/common.h
tommyp1ckles: I'm using the cilium-builder docker image to build this, I get:

```
/home/tom/go/src/github.com/cilium/cilium/bpf/include/linux/../bpf/helpers.h:38:13: error: use of undeclared identifier 'BPF_FUNC_map_lookup_percpu_elem'
static void BPF_FUNC(map_lookup_percpu_elem, const void *map, const void *key, unsigned int cpu);
```

It seems like the version of `linux/bpf.h` we're using doesn't include that definition.

tommyp1ckles: @gentoo-root not sure the best course of action here, any advice?

gentoo-root: We can update bpf.h, we do that when needed:

https://github.com/cilium/cilium/commit/049a3a9da86e5221867013ba99a7bbb581b1b131
https://github.com/cilium/cilium/commit/271a96da32113b9c02465aca3e2587584164924a

tommyp1ckles: Added a commit to sync with latest upstream commit, there's a few changes in the diff - I assume those are all ok/compatible?

tommyp1ckles: @gentoo-root

gentoo-root: Yeah, should be good, as long as the tests don't start failing due to this.

gentoo-root (approved):

bpf/tests/common.h
```diff
@@ -272,4 +275,22 @@ _Pragma("unroll")						\
...
+	__u64 sum = 0; \
+	/* Iterate until lookup encounters null when hitting cpu number */ \
+	/* Assumes at most 128 CPUS */ \
+	for (int i = 0; i < NR_CPUS; i++) { \
+		__entry = map_lookup_percpu_elem(&METRICS_MAP, &key, i); \
+		if (!__entry) { \
+			break; \
+		} \
+		sum++; \
+		break; \
```
jschwinger233: This `break` limits `sum` from increasing beyond 1, is it intentional?

And maybe we should `sum += __entry->count` instead of `sum++`? Otherwise sum is always set to the number of CPUs.

jschwinger233 (changes requested):

bpf/tests/common.h
tommyp1ckles: Woops yup thanks 😅 , fixed

jschwinger233 (approved):
Thanks!

PR #19765, opened 2022-05-10
Restructure IPCache to handle metadata merging

joestringer:
Meta: https://github.com/cilium/cilium/issues/21142

IPCache has historically been based around a concept of associating IP prefixes with a certain _Identity_, where each caller independently determines the exact identity that a IP Prefix should have. Each of these callers has a corresponding `source.Source` which decides the priority between these different sources of Identity.

More recently however, features like #14724 have driven the need to combine multiple sources of information, such as associating both the `remote-node` labels (from k8s `CiliumNode` resources) and `kube-apiserver` labels (from k8s `Endpoints` / `EndpointSlices` resources). Having an exclusionary policy where only one source of information can be correct can have implications like causing conflicts between policies that match on information from each source. Up until now, we've resolved these conflicts somewhat manually by forcing the callers to handle the merging of this information. For most callers, they're independent enough that users would only encounter bugs in rare conditions of using two features together in an unexpected way. We have made some attempts in certain paths to resolve these conflicts where necessary, including the example above. However, over time this code becomes more and more :spaghetti: with each caller attempting to resolve conflicts with the other callers.

This PR reworks the metadata map inside the IPCache to handle multiple sources of information at once, keying them in the map first by IP prefix, then by a `{resourceType, resourceNamespace, resourceName}` tuple (`ResourceID`). When merging the information, a new `prefixInfo` structure provides convenient merging of the labels from all of the sources of information. In effect, this introduced one additional layer of indirection with a series of chained structures represented in the mermaid chart below:

```mermaid
flowchart
  subgraph resourceInfo
  labels.Labels
  source.Source
  end
  subgraph prefixInfo
  UA[ResourceID]-->LA[resourceInfo]
  UB[ResourceID]-->LB[resourceInfo]
  ...
  end
  subgraph identityMetadata
  IP_Prefix-->prefixInfo
  end
```

[...]

PR #28332, opened 2023-09-29
ipcache: fix flapping labels in SelectorCache when reserved:host identity has multiple IPs

### LINKED ISSUE #28259: "kube-apiserver" policy entity is broken on dual-stack clusters
### What happened?

A bug happened!

The mechanism that applies the `kube-apiserver` label can cause flaps when the IP in question belongs to the local node. This only applies to clusters where nodes have ipv6 addresses. (Note that whether or not Cilium is in dual-stack mode is irrelevant).

The root cause is somewhat complicated due to the number of special cases:

1. the ipcache "back-propagates" labels up to a global value only for the local identity.
2. the `default/kubernetes` service is always single-stack, even for dual-stack clusters
3. the node-manager doesn't understand the kube-apsierver identity; it is applied purely to IPs in the ipcache metadata layer.

So when there are multiple IPs for the `reserved:node` identity, but their set of labels is not uniform, it is last-write-wins. Whichever IP is most recently triggered will have those labels applied to the `reserved:node` identity.

### Cilium Version

v1.14 at least, perhaps older.

Edit: See https://github.com/cilium/cilium/issues/28259#issuecomment-1746489439.

squeed:
The `reserved:host` identity is special: the numeric identity is fixed and the set of labels is mutable. (The datapath requires this.) So, we need to determine all prefixes that have the `reserved:host` label and capture their labels. Then, we must aggregate *all* labels from all IPs and insert them as the `reserved:host` identity labels.

However, the code as written has a race condition whenever the local node has more than one IP address. This can happen when, for example vxlan or ipv6 is enabled. The basic sequence is this:

1. Insert IP A as `reserved:host` in to the ipcache. ID 1 now has labels `reserved:host`
2. Insert IP A as `reserved:kube-apiserver` in to the ipcache. ID 1 is updated with labels labels `reserved:host, reserved:kube-apsierver`
3. Insert IP B as `reserved:host` in to the ipcache. ID 1 is updated with labels `reserved:host`
And now policies are broken.

Likewise, we need to always update the SelectorCache; we cannot short-circuit if the ipcache already has that identity. Again, this is needed because the identity is mutable. So this bug can take another form:

1. Insert IP A as `reserved:host` in to the ipcache. Because IP A is not known to the ipcache, treat ID 1 as a new identity and update the selector cache
2. Insert IP A as `reserved:kube-apiserver`. Mutate the labels of ID 1. But, because IP A already has ID 1, short-circuit the update to the selector cache (if the Source is the same, which it _may_ be).
3. Now the selector cache has incorrect labels for ID 1.

Without this, when there are multiple IPs with the host label, the identity may flap and the SelectorCache may be missing updates.

Fixes: #28259
Fixes: b099ba614
Fixes: f5d95325c

### DIFF
--- a/pkg/identity/reserved.go
+++ b/pkg/identity/reserved.go
@@ -17,25 +17,30 @@ var (
 )
 
 // AddReservedIdentity adds the reserved numeric identity with the respective
-// label into the map of reserved identity cache.
-func AddReservedIdentity(ni NumericIdentity, lbl string) {
+// label into the map of reserved identity cache, and returns the resulting Identity.
+// This identity must not be mutated!
+func AddReservedIdentity(ni NumericIdentity, lbl string) *Identity {
 	identity := NewIdentity(ni, labels.Labels{lbl: labels.NewLabel(lbl, "", labels.LabelSourceReserved)})
 	cacheMU.Lock()
 	reservedIdentityCache[ni] = identity
 	cacheMU.Unlock()
+	return identity
 }
 
 // AddReservedIdentityWithLabels is the same as AddReservedIdentity but accepts
-// multiple labels.
-func AddReservedIdentityWithLabels(ni NumericIdentity, lbls labels.Labels) {
+// multiple labels. Returns the resulting Identity.
+// This identity must not be mutated!
+func AddReservedIdentityWithLabels(ni NumericIdentity, lbls labels.Labels) *Identity {
 	identity := NewIdentity(ni, lbls)
 	cacheMU.Lock()
 	reservedIdentityCache[ni] = identity
 	cacheMU.Unlock()
+	return identity
 }
 
 // LookupReservedIdentity looks up a reserved identity by its NumericIdentity
 // and returns it if found. Returns nil if not found.
+// This identity must not be mutated!
 func LookupReservedIdentity(ni NumericIdentity) *Identity {
 	cacheMU.RLock()
 	defer cacheMU.RUnlock()
@@ -43,7 +48,9 @@ func LookupReservedIdentity(ni NumericIdentity) *Identity {
 }
 
 func init() {
-	iterateReservedIdentityLabels(AddReservedIdentityWithLabels)
+	iterateReservedIdentityLabels(func(ni NumericIdentity, lbls labels.Labels) {
+		AddReservedIdentityWithLabels(ni, lbls)
+	})
 }
 
 // IterateReservedIdentities iterates over all reserved identities and
--- a/pkg/ipcache/metadata.go
+++ b/pkg/ipcache/metadata.go
@@ -81,12 +81,22 @@ type metadata struct {
 	// generate updates into the ipcache, policy engine and datapath.
 	queuedChangesMU lock.Mutex
 	queuedPrefixes  map[netip.Prefix]struct{}
+
+	// reservedHostLock protects the localHostLabels map. Holders must
+	// always take the metadata read lock first.
+	reservedHostLock lock.Mutex
+
+	// reservedHostLabels collects all labels that apply to the host identity.
+	// see updateLocalHostLabels() for more info.
+	reservedHostLabels map[netip.Prefix]labels.Labels
 }
 
 func newMetadata() *metadata {
 	return &metadata{
 		m:              make(map[netip.Prefix]PrefixInfo),
 		queuedPrefixes: make(map[netip.Prefix]struct{}),
+
+		reservedHostLabels: make(map[netip.Prefix]labels.Labels),
 	}
 }
 
@@ -206,14 +216,13 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
 		oldID, entryExists := ipc.LookupByIP(pstr)
 		oldTunnelIP, oldEncryptionKey := ipc.GetHostIPCache(pstr)
 		prefixInfo := ipc.metadata.getLocked(prefix)
+		var newID *identity.Identity
 		if prefixInfo == nil {
 			if !entryExists {
 				// Already deleted, no new metadata to associate
 				continue
 			} // else continue below to remove the old entry
 		} else {
-			var newID *identity.Identity
-
 			// Insert to propagate the updated set of labels after removal.
 			newID, _, err = ipc.resolveIdentity(ctx, prefix, prefixInfo, identity.InvalidIdentity)
 			if err != nil {
@@ -305,6 +314,25 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
 				entriesToDelete[prefix] = oldID
 			}
 		}
+
+		// The reserved:host identity is special: the numeric ID is fixed,
+		// and the set of labels is mutable. Thus, whenever it changes,
+		// we must always update the SelectorCache (normally, this is elided
+		// when no changes are present).
+		if newID != nil && newID.ID == identity.ReservedIdentityHost {
+			idsToAdd[newID.ID] = newID.Labels.LabelArray()
+		}
+
+		// Again, more reserved:host bookkeeping: if this prefix is no longer ID 1 (because
+		// it is being deleted or changing IDs), we need to recompute the labels
+		// for reserved:host and push that to the SelectorCache
+		if entryExists && oldID.ID == identity.ReservedIdentityHost &&
+			(newID == nil || newID.ID != identity.ReservedIdentityHost) {
+
+			i := ipc.updateReservedHostLabels(prefix, nil)
+			idsToAdd[i.ID] = i.Labels.LabelArray()
+		}
+
 	}
 	// Don't hold lock while calling UpdateIdentities, as it will otherwise run into a deadlock
 	ipc.metadata.RUnlock()
@@ -478,8 +506,12 @@ func (ipc *IPCache) resolveIdentity(ctx context.Context, prefix netip.Prefix, in
 		// for itself). For all other identities, we avoid modifying
 		// the labels at runtime and instead opt to allocate new
 		// identities below.
-		identity.AddReservedIdentityWithLabels(identity.ReservedIdentityHost, lbls)
-		return identity.LookupReservedIdentity(identity.ReservedIdentityHost), false, nil
+		//
+		// As an extra gotcha, we need need to merge all labels for all IPs
+		// that resolve to the reserved:host identity, otherwise we can
+		// flap identities labels depending on which prefix writes first. See GH-28259.
+		i := ipc.updateReservedHostLabels(prefix, lbls)
+		return i, false, nil
 	}
 
 	// If no other labels are associated with this IP, we assume that it's
@@ -509,6 +541,36 @@ func (ipc *IPCache) resolveIdentity(ctx context.Context, prefix netip.Prefix, in
 		id.CIDRLabel = labels.NewLabelsFromModel([]string{labels.LabelSourceCIDR + ":" + prefix.String()})
 	}
 	return id, isNew, err
+}
+
+// updateReservedHostLabels adds or removes labels that apply to the local host.
+// The `reserved:host` identity is special: the numeric identity is fixed
+// and the set of labels is mutable. (The datapath requires this.) So,
+// we need to determine all prefixes that have the `reserved:host` label and
+// capture their labels. Then, we must aggregate *all* labels from all prefixes and
+// update the labels that correspond to the `reserved:host` identity.
+//
+// This could be termed a meta-ipcache. The ipcache metadata layer aggregates
+// an arbitrary set of resources and labels to a prefix. Here, we are aggregating an arbitrary
+// set of prefixes and labels to an identity.
+func (ipc *IPCache) updateReservedHostLabels(prefix netip.Prefix, lbls labels.Labels) *identity.Identity {
+	ipc.metadata.reservedHostLock.Lock()
+	defer ipc.metadata.reservedHostLock.Unlock()
+	if lbls == nil {
+		delete(ipc.metadata.reservedHostLabels, prefix)
+	} else {
+		ipc.metadata.reservedHostLabels[prefix] = lbls
+	}
+
+	// aggregate all labels and update static identity
+	newLabels := labels.NewFrom(labels.LabelHost)
+	for _, l := range ipc.metadata.reservedHostLabels {
+		newLabels.MergeLabels(l)
+	}
+
+	log.WithField(logfields.Labels, newLabels).Debug("Merged labels for reserved:host identity")
+
+	return identity.AddReservedIdentityWithLabels(identity.ReservedIdentityHost, newLabels)
 }
 
 // RemoveLabelsExcluded removes the given labels from all IPs inside the IDMD
--- a/pkg/ipcache/metadata_test.go
+++ b/pkg/ipcache/metadata_test.go
@@ -236,6 +236,88 @@ func TestInjectLabels(t *testing.T) {
 	option.Config.EnableIPv6 = true
 }
 
+// Test that when multiple IPs have the `resolved:host` label, we correctly
+// aggregate all labels *and* update the selector cache correctly.
+// This reproduces GH-28259.
+func TestUpdateLocalNode(t *testing.T) {
+	cancel := setupTest(t)
+	defer cancel()
+
+	ctx := context.Background()
+
+	bothLabels := labels.Labels{}
+	bothLabels.MergeLabels(labels.LabelHost)
+	bothLabels.MergeLabels(labels.LabelKubeAPIServer)
+
+	selectorCacheHas := func(lbls labels.Labels) {
+		t.Helper()
+		id := PolicyHandler.identities[identity.ReservedIdentityHost]
+		assert.NotNil(t, id)
+		assert.Equal(t, lbls.LabelArray(), id)
+	}
+
+	injectLabels := func(ip netip.Prefix) {
+		t.Helper()
+		remaining, err := IPIdentityCache.InjectLabels(ctx, []netip.Prefix{ip})
+		assert.NoError(t, err)
+		assert.Len(t, remaining, 0)
+	}
+
+	idIs := func(ip netip.Prefix, id identity.NumericIdentity) {
+		t.Helper()
+		assert.Equal(t, IPIdentityCache.ipToIdentityCache[ip.String()].ID, id)
+	}
+
+	// Mark .4 as local host
+	IPIdentityCache.metadata.upsertLocked(inClusterPrefix, source.Local, "node-uid", labels.LabelHost)
+	injectLabels(inClusterPrefix)
+	idIs(inClusterPrefix, identity.ReservedIdentityHost)
+	selectorCacheHas(labels.LabelHost)
+
+	// Mark .4 as kube-apiserver
+	// Note that in the actual code, we use `source.KubeAPIServer`. However,
+	// we use the same source in test case to try and ferret out more bugs.
+	IPIdentityCache.metadata.upsertLocked(inClusterPrefix, source.Local, "kube-uid", labels.LabelKubeAPIServer)
+	injectLabels(inClusterPrefix)
+	idIs(inClusterPrefix, identity.ReservedIdentityHost)
+	selectorCacheHas(bothLabels)
+
+	// Mark .5 as local host
+	IPIdentityCache.metadata.upsertLocked(inClusterPrefix2, source.Local, "node-uid", labels.LabelHost)
+	injectLabels(inClusterPrefix2)
+	idIs(inClusterPrefix, identity.ReservedIdentityHost)
+	idIs(inClusterPrefix2, identity.ReservedIdentityHost)
+	selectorCacheHas(bothLabels)
+
+	// remove kube-apiserver from .4
+	IPIdentityCache.metadata.remove(inClusterPrefix, "kube-uid", labels.LabelKubeAPIServer)
+	injectLabels(inClusterPrefix)
+	idIs(inClusterPrefix, identity.ReservedIdentityHost)
+	idIs(inClusterPrefix2, identity.ReservedIdentityHost)
+	selectorCacheHas(labels.LabelHost)
+
+	// add kube-apiserver back to .4
+	IPIdentityCache.metadata.upsertLocked(inClusterPrefix, source.Local, "kube-uid", labels.LabelKubeAPIServer)
+	injectLabels(inClusterPrefix)
+	idIs(inClusterPrefix, identity.ReservedIdentityHost)
+	idIs(inClusterPrefix2, identity.ReservedIdentityHost)
+	selectorCacheHas(bothLabels)
+
+	// remove host from .4
+	IPIdentityCache.metadata.remove(inClusterPrefix, "node-uid", labels.LabelHost)
+	injectLabels(inClusterPrefix)
+
+	// Verify that .4 now has just kube-apiserver and CIDRs
+	idIs(inClusterPrefix, identity.IdentityScopeLocal) // the first CIDR identity
+	id := PolicyHandler.identities[identity.IdentityScopeLocal]
+	assert.True(t, id.Has("reserved.kube-apiserver"))
+	assert.True(t, id.Has("cidr."+inClusterPrefix.String()))
+
+	// verify that id 1 is now just reserved:host
+	idIs(inClusterPrefix2, identity.ReservedIdentityHost)
+	selectorCacheHas(labels.LabelHost)
+}
+
 // TestInjectExisting tests "upgrading" an existing identity to the apiserver.
 // This is a common occurrence on startup - and this tests ensures we don't
 // regress the known issue in GH-24502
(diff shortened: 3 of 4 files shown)


nathanjsweet:
Additional CI for "pod-to-node-policycidr" test: https://github.com/cilium/cilium/pull/28334.

christarazi:
> Without this, when there are multiple IPs with the host label, the identity may flap and the SelectorCache may be missing updates.

I'm having trouble understanding this and without understanding, it's difficult to review the PR. Could you go into more detail and also provide an example?

nathanjsweet:
[This test](https://github.com/cilium/cilium/actions/runs/6353246315/job/17322728953) keeps failing, but when I run it locally, it passes. I'm going to rerun it a few more times, because the error is simply one of the test pods not coming up.

pkg/ipcache/metadata.go
```diff
@@ -81,12 +81,21 @@ type metadata struct {
 	// generate updates into the ipcache, policy engine and datapath.
 	queuedChangesMU lock.Mutex
 	queuedPrefixes  map[netip.Prefix]struct{}
+
+	// reservedHostLock protects the localHostLabels map
```
schlosna:
```suggestion
	// reservedHostLock protects the reservedHostLabels map
```

pkg/ipcache/metadata.go
```diff
@@ -511,6 +538,33 @@ func (ipc *IPCache) resolveIdentity(ctx context.Context, prefix netip.Prefix, in
...
+		ipc.metadata.reservedHostLabels[prefix] = lbls
+	}
+
+	// aggregate all labels and update static identity
+	newLabels := labels.Labels{}
+	for _, l := range ipc.metadata.reservedHostLabels {
+		newLabels.MergeLabels(l)
+	}
+
+	log.WithField(logfields.Labels, newLabels).Debug("Merged labels for reserved:host identity")
```
schlosna: Should this be pulled out to a separate function to allow releasing the lock to reduce critical section and possible contention before adding and looking up identity?

pkg/ipcache/metadata.go
```diff
@@ -511,6 +538,33 @@ func (ipc *IPCache) resolveIdentity(ctx context.Context, prefix netip.Prefix, in
...
+
+	identity.AddReservedIdentityWithLabels(identity.ReservedIdentityHost, newLabels)
+	return identity.LookupReservedIdentity(identity.ReservedIdentityHost)
+}
```
schlosna: more specifically, I was wondering if something like the following might be beneficial to avoid lock contention on both the `IPCache.metadata.reservedHostLock` and also update `AddReservedIdentityWithLabels` to return the identity added to the cache to avoid an additional cache read lock and access.

```suggestion
func (ipc *IPCache) updateReservedHostLabels(prefix netip.Prefix, lbls labels.Labels) *identity.Identity {
	newLabels := ipc.mergeReservedHostLabels(prefix, lbls)
	log.WithField(logfields.Labels, newLabels).Debug("Merged labels for reserved:host identity")
	return identity.AddReservedIdentityWithLabels(identity.ReservedIdentityHost, newLabels)
}

func (ipc *IPCache) mergeReservedHostLabels(prefix netip.Prefix, lbls labels.Labels) labels.Labels {
	ipc.metadata.reservedHostLock.Lock()
	defer ipc.metadata.reservedHostLock.Unlock()
	if lbls == nil {
		delete(ipc.metadata.reservedHostLabels, prefix)
	} else {
		ipc.metadata.reservedHostLabels[prefix] = lbls
	}

	// aggregate all labels and update static identity
	newLabels := labels.Labels{}
	for _, l := range ipc.metadata.reservedHostLabels {
		newLabels.MergeLabels(l)
	}

	return newLabels
}
```

`pkg/identity/reserved.go`:
```
func AddReservedIdentityWithLabels(ni NumericIdentity, lbls labels.Labels) *Identity {
	identity := NewIdentity(ni, lbls)
	cacheMU.Lock()
	reservedIdentityCache[ni] = identity
	cacheMU.Unlock()
	return identity
}

func init() {
	iterateReservedIdentityLabels(func(ni NumericIdentity, lbls labels.Labels) {
		AddReservedIdentityWithLabels(ni, lbls)
	})
}
```

joestringer (approved):
Core change LGTM from a high level, I walked through the logic along with @christarazi . He may have some other minor feedback.

pkg/ipcache/metadata.go
squeed: I'd considered doing just that (but hadn't to keep the PR small). But yes, I think it's a good idea.

squeed: Yeah, I had considered something like this as well, but left it out to make a small PR. But you've convinced me it's worth it. I'll remove the extra lock.

squeed: However, we can't drop `reservedHostLock` until after we update the identity store -- otherwise we might have an A-B-A problem when there are parallel updates. (both InjectLabels and AllocateCIDRs call this code, and they only have a read lock on the metadata cache).

squeed:
@schlosna updated based on feedback; PTAL.

pkg/ipcache/metadata.go
squeed: It shouldn't; see below. (resolving this thread)

schlosna (approved):
Thanks for addressing my comments, this looks good to me.

The commit check is unhappy with a duplicated word `labels` in the commit message and references to commits b099ba614 & f5d95325c that were part of https://github.com/cilium/cilium/pull/19765

> {[1/1] ipcache: aggregate labels from all IPs with local host identity}
>   Running on ae661971b3e961b64b9f7fc28a6e162eb72ed58e
>   Warning: WARNING:REPEATED_WORD: Possible repeated word: 'labels'
>   #[20](https://github.com/cilium/cilium/actions/runs/6404376766/job/17384819951?pr=28332#step:3:37): 
>     with labels labels `reserved:host, reserved:kube-apsierver`
>   
>   Warning: WARNING:UNKNOWN_COMMIT_ID: Unknown commit id 'b099ba614', maybe rebased or not pulled?
>   #[41](https://github.com/cilium/cilium/actions/runs/6404376766/job/17384819951?pr=28332#step:3:42): 
>   Fixes: b099ba614
>   
>   Warning: WARNING:UNKNOWN_COMMIT_ID: Unknown commit id 'f5d95325c', maybe rebased or not pulled?
>   #[42](https://github.com/cilium/cilium/actions/runs/6404376766/job/17384819951?pr=28332#step:3:43): 
>   Fixes: f5d95325c

pkg/identity/reserved.go
```diff
@@ -18,20 +18,22 @@ var (
...
 	cacheMU.Lock()
 	reservedIdentityCache[ni] = identity
 	cacheMU.Unlock()
+	return identity
 }
 
 // AddReservedIdentityWithLabels is the same as AddReservedIdentity but accepts
 // multiple labels.
-func AddReservedIdentityWithLabels(ni NumericIdentity, lbls labels.Labels) {
+func AddReservedIdentityWithLabels(ni NumericIdentity, lbls labels.Labels) *Identity {
```
christarazi: Let's update the godocs to mention that the added identity is now returned and more importantly, should **not** be modified as it's a live reference.

pkg/ipcache/metadata.go
```diff
@@ -511,6 +538,32 @@ func (ipc *IPCache) resolveIdentity(ctx context.Context, prefix netip.Prefix, in
 	return id, isNew, err
 }
 
+// updateReservedHostLabels adds or removes labels that apply to the local host.
+// The `reserved:host` identity is special: the numeric identity is fixed
+// and the set of labels is mutable. (The datapath requires this.) So,
+// we need to determine all prefixes that have the `reserved:host` label and
+// capture their labels. Then, we must aggregate *all* labels from all IPs and
+// insert them as the `reserved:host` identity labels.
```
christarazi: Just clarifying a bit more.

```suggestion
// capture their labels. Then, we must aggregate *all* labels from all prefixes and
// update the labels that correspond to the `reserved:host` identity.
```

christarazi: Additionally, I think it would be useful to call out is that this function is doing another merging of metadata of which there are two now: one in this function and the other is the main ipcache metadata consolidator. The main difference from the latter is that we have to consider other prefixes, whereas the latter is per prefix.

pkg/ipcache/metadata.go
```diff
@@ -511,6 +538,32 @@ func (ipc *IPCache) resolveIdentity(ctx context.Context, prefix netip.Prefix, in
...
+	ipc.metadata.reservedHostLock.Lock()
+	defer ipc.metadata.reservedHostLock.Unlock()
+	if lbls == nil {
+		delete(ipc.metadata.reservedHostLabels, prefix)
+	} else {
+		ipc.metadata.reservedHostLabels[prefix] = lbls
+	}
+
+	// aggregate all labels and update static identity
+	newLabels := labels.Labels{}
```
christarazi: Nit: Can this be

```suggestion
	newLabels := labels.NewFrom(labels.LabelHost)
```

?

pkg/ipcache/metadata_test.go
```diff
@@ -236,6 +236,63 @@ func TestInjectLabels(t *testing.T) {
...
+	bothLabels := labels.Labels{}
+	bothLabels.MergeLabels(labels.LabelHost)
+	bothLabels.MergeLabels(labels.LabelKubeAPIServer)
+
+	selectorCacheHas := func(lbls labels.Labels) {
+		t.Helper()
+		id := PolicyHandler.identities[hostid]
+		assert.NotNil(t, id)
+
+		assert.True(t, id.Equals(lbls.LabelArray()), "labels should equal: %v -- %v", id, lbls.LabelArray())
```
christarazi: When running this test locally, I struggled to understand which value that's outputted is expected vs obtained. When I modified this to 

```suggestion
		assert.Equal(t, lbls.LabelArray(), id, "labels should equal")
```

it was much easier to read.

pkg/ipcache/metadata_test.go
```diff
@@ -236,6 +236,63 @@ func TestInjectLabels(t *testing.T) {
...
+	// Mark .4 as local host
+	IPIdentityCache.metadata.upsertLocked(inClusterPrefix, source.Local, "node-uid", labels.LabelHost)
+	remaining, err := IPIdentityCache.InjectLabels(ctx, []netip.Prefix{inClusterPrefix})
+	assert.NoError(t, err)
+	assert.Len(t, remaining, 0)
+	assert.Equal(t, IPIdentityCache.ipToIdentityCache["10.0.0.4/32"].ID, hostid)
+	selectorCacheHas(labels.LabelHost)
+
+	// Mark .4 as kube-apiserver
+	IPIdentityCache.metadata.upsertLocked(inClusterPrefix, source.Local, "kube-uid", labels.LabelKubeAPIServer)
```
christarazi, line 280: This should actually be

```suggestion
	IPIdentityCache.metadata.upsertLocked(inClusterPrefix, source.KubeAPIServer, "kube-uid", labels.LabelKubeAPIServer)
```

but I don't think there's a functional impact here.

pkg/ipcache/metadata_test.go
```diff
@@ -236,6 +236,63 @@ func TestInjectLabels(t *testing.T) {
...
+	assert.Equal(t, IPIdentityCache.ipToIdentityCache["10.0.0.5/32"].ID, hostid)
+	selectorCacheHas(bothLabels)
+
+	// remove kube-apiserver
+	IPIdentityCache.metadata.remove(inClusterPrefix, "kube-uid", overrideIdentity(false), labels.LabelKubeAPIServer)
+	remaining, err = IPIdentityCache.InjectLabels(ctx, []netip.Prefix{inClusterPrefix})
+	assert.NoError(t, err)
+	assert.Len(t, remaining, 0)
+	assert.Equal(t, IPIdentityCache.ipToIdentityCache["10.0.0.4/32"].ID, hostid)
+
```
christarazi, line 315: Nit: extra newline?

pkg/ipcache/metadata.go
```diff
@@ -81,12 +81,21 @@ type metadata struct {
 	// generate updates into the ipcache, policy engine and datapath.
 	queuedChangesMU lock.Mutex
 	queuedPrefixes  map[netip.Prefix]struct{}
+
+	// reservedHostLock protects the localHostLabels map
+	reservedHostLock lock.Mutex
```
christarazi: I think it's worth calling out here that we require this lock to be taken only within the `metadata` **read** lock.

pkg/ipcache/metadata.go
```diff
@@ -211,6 +220,14 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
...
 				continue
 			} // else continue below to remove the old entry
+
+			// Special case: if we are removing an IP from the reserved:host identity,
+			// we need to remove this prefix's labels from the local identity and
+			// force an update to the selector cache.
+			if oldID.ID == identity.ReservedIdentityHost {
+				i := ipc.updateReservedHostLabels(prefix, nil)
+				idsToAdd[i.ID] = i.Labels.LabelArray()
+			}
```
christarazi: I'm wondering if it's easier on the reader to have all of the cases for "deletion" under the `releaseIdentity` section on line 292. That way when the reader is considering all of the deletion cases, there's just one part of the function to look at, instead of overhead of scrolling up and down to evaluate different conditions, etc.

pkg/ipcache/metadata.go
```diff
@@ -239,6 +256,12 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
 				break
 			}
 
+			// Special case: the host identity is mutable; we need to always update the SelectorCache.
+			// If the labels are unchanged, the SelectorCache will short-circuit.
+			if newID.ID == identity.ReservedIdentityHost {
+				idsToAdd[newID.ID] = newID.Labels.LabelArray()
+			}
+
```
christarazi: Similar sentiment as my previous comment, but a bit further. Same spirit as keeping the loop core logic more readable. 

This line here is updating the **selectorcache**. The logic inside `resolveIdentity` is updating the **host identity labels**. I propose we keep that general structure.

However, it seems this check highlighted here was added presumably before the condition that could trigger jumping to `releaseIdentity`. To me this suggests there could be a helper function that returns bools based on whether the selectorcache should be updated or if we need to `releaseIdentity`.

That would involve refactoring the few lines of code below starting with `if oldID.ID == newID.ID && prefixInfo.Source() == oldID.Source &&` into a separate function and folding in `if newID.ID == identity.ReservedIdentityHost {` somewhere in that function.

I've took a stab at this below so feel free to apply and amend if needed. Naming could be improved as per usual :), unit tests are still passing for me so I think I've got it right:

<details><summary>Diff</summary>
<p>

```diff
diff --git a/pkg/ipcache/metadata.go b/pkg/ipcache/metadata.go
index 6874be5951..43955fc7f1 100644
--- a/pkg/ipcache/metadata.go
+++ b/pkg/ipcache/metadata.go
@@ -232,8 +232,8 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
 			var newID *identity.Identity
 
 			// Insert to propagate the updated set of labels after removal.
-			newID, _, err = ipc.resolveIdentity(ctx, prefix, prefixInfo, identity.InvalidIdentity)
-			if err != nil {
+			newID, isNew, err2 := ipc.resolveIdentity(ctx, prefix, prefixInfo, identity.InvalidIdentity)
+			if err2 != nil {
 				// NOTE: This may fail during a 2nd or later
 				// iteration of the loop. To handle this, break
 				// the loop here and continue executing the set
@@ -252,26 +252,18 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
 					"Failed to allocate new identity while handling change in labels associated with a prefix.",
 				)
 				remainingPrefixes = modifiedPrefixes[i:]
-				err = fmt.Errorf("failed to allocate new identity during label injection: %w", err)
+				err = fmt.Errorf("failed to allocate new identity during label injection: %w", err2)
 				break
 			}
 
-			// Special case: the host identity is mutable; we need to always update the SelectorCache.
-			// If the labels are unchanged, the SelectorCache will short-circuit.
-			if newID.ID == identity.ReservedIdentityHost {
+			selectorCacheUpdate, releaseID := determineIdentityUpdate(oldID, newID, isNew, prefixInfo, oldTunnelIP, oldEncryptionKey)
+			if selectorCacheUpdate {
 				idsToAdd[newID.ID] = newID.Labels.LabelArray()
 			}
-
-			// We can safely skip the ipcache upsert if the entry matches with
-			// the entry in the metadata cache exactly.
-			// Note that checking ID alone is insufficient, see GH-24502.
-			if oldID.ID == newID.ID && prefixInfo.Source() == oldID.Source &&
-				oldTunnelIP.Equal(prefixInfo.TunnelPeer().IP()) &&
-				oldEncryptionKey == prefixInfo.EncryptKey().Uint8() {
+			if releaseID {
 				goto releaseIdentity
 			}
 
-			idsToAdd[newID.ID] = newID.Labels.LabelArray()
 			entriesToReplace[prefix] = ipcacheEntry{
 				identity: Identity{
 					ID:                  newID.ID,
@@ -288,6 +280,7 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
 				// identity and replace it with this new identity.
 				force: entryExists && prefixInfo.Source() != oldID.Source && oldID.ID != newID.ID,
 			}
+
 		}
 	releaseIdentity:
 		if entryExists {
@@ -406,6 +399,37 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
 	return remainingPrefixes, err
 }
 
+func determineIdentityUpdate(
+	oldID Identity, newID *identity.Identity,
+	isNewID bool,
+	prefixInfo PrefixInfo,
+	oldTunnelIP net.IP,
+	oldEncryptionKey uint8,
+) (selectorCacheUpdate, releaseIdentity bool) {
+	// Always update the SelectorCache if the identity is new.
+	selectorCacheUpdate = isNewID
+
+	// Special case: the host identity is mutable; we need to always update the
+	// SelectorCache, even though isNewID might be false.
+	//
+	// If the labels are unchanged, the SelectorCache will short-circuit
+	// anyway.
+	if newID.ID == identity.ReservedIdentityHost {
+		selectorCacheUpdate = true
+	}
+
+	// We can safely skip the ipcache upsert if the entry matches with
+	// the entry in the metadata cache exactly.
+	// Note that checking ID alone is insufficient, see GH-24502.
+	if oldID.ID == newID.ID && prefixInfo.Source() == oldID.Source &&
+		oldTunnelIP.Equal(prefixInfo.TunnelPeer().IP()) &&
+		oldEncryptionKey == prefixInfo.EncryptKey().Uint8() {
+		releaseIdentity = true
+	}
+
+	return
+}
+
 // UpdatePolicyMaps pushes updates for the specified identities into the policy
 // engine and ensures that they are propagated into the underlying datapaths.
 func (ipc *IPCache) UpdatePolicyMaps(ctx context.Context, addedIdentities, deletedIdentities map[identity.NumericIdentity]labels.LabelArray) {
```

</p>
</details>

christarazi (commented):
The core logic LGTM and I have some suggestions that IMO improve the code readability and hopefully maintainability. See the comments below.

Stepping back, initially I was a bit confused as it looked to me that this PR was needed to fix https://github.com/cilium/cilium/pull/27464, but the [linked issue](https://github.com/cilium/cilium/issues/28259) was something different. In the end, what it looks like is that this PR is needed to fix both cases, https://github.com/cilium/cilium/pull/27464 and the [linked issue](https://github.com/cilium/cilium/issues/28259), but this was not clear to me.

To understand the root cause of https://github.com/cilium/cilium/issues/28259, I git bisected down to https://github.com/cilium/cilium/pull/19765/commits/b099ba614e7090ea6a19e8dbadfe0db07eb83298 and https://github.com/cilium/cilium/pull/19765/commits/f5d95325c10762d3cd563be4a71a87bbdfd701ee, which shows a regression going back to v1.13.0-rc1. https://github.com/cilium/cilium/issues/28259 stated that the issue exists on main as well as v1.14, but it looks like this fix will need to be backported to v1.13 as well. Additionally, it turns out that there's no CI coverage for this case, so it's great you found & fixed it (:tada:) and we'll need to fill this gap in coverage. From an offline sync, @nathanjsweet offered to put up a PR for that.

The summary of my root cause analysis is in the regression we had removed the special handling of the host identity labels, and this PR basically adds that back. 

Policy, identity, ipcache, etc, are all code areas with logic that can sometimes be extremely subtle. A one line change could cause a significant regression, so understanding the full root cause is crucial in evaluating a change. For future changes, understanding which change introduced the regression is important context to provide in commit msgs / descriptions so we can determine which versions are affected. Additionally, it will greatly ease code reviews of particularly curly areas of Cilium like ipcache & policy-related code.

pkg/ipcache/metadata.go
squeed: I have to check the mutation rules, but probably, yes.

pkg/ipcache/metadata_test.go
squeed: Keeping the sources the same makes the bug much more likely. So it is a good test-case for the specific bug, even if it doesn't mirror current usage exactly.

pkg/ipcache/metadata.go
squeed: I'll see if the blocks can be correctly joined without affecting logic.

pkg/ipcache/metadata.go
squeed: it's not just SelectorCacheUpdate; we elide the entire ipcache update when the comparison is true.

Given how far this needs to be backported, I'd prefer a minimal diff.

pkg/ipcache/metadata.go
squeed: While thinking through this, I realized there was a potential bug when an IP lost reserved:node status. So the logic needed to be refactored to the "bottom" anyways.

pkg/ipcache/metadata.go
christarazi: > it's not just SelectorCacheUpdate; we elide the entire ipcache update when the comparison is true.

Do you mean the `if oldID.ID == newID.ID && prefixInfo.Source() == oldID.Source &&` comparisons? AFAICT, if `releaseIdentity` is true, then it will also skip the ipcache update, and the selectorcache update always happens if it needs to.

> Given how far this needs to be backported, I'd prefer a minimal diff.

Fair enough. I'm not opposed to keeping this minimal and submitting the refactor in a future `main` PR.

pkg/ipcache/metadata_test.go
squeed: Somehow I thought that label arrays didn't have ordering guarantees, but that was wrong, and we can use `assert.Equal`

pkg/ipcache/metadata.go
squeed: It could, though that just does a MergeLabels under the hood.

pkg/ipcache/metadata.go
squeed: OK, check out the latest revision.

pkg/ipcache/metadata.go
squeed: OK, I added some wording to that effect.

pkg/ipcache/metadata_test.go
christarazi: > Keeping the sources the same makes the bug much more likely

The only cases of labels attaching to the host identity is either kube-apiserver or CIDR labels in the case of node CIDR policy. AFAICT in node CIDR policy case, CIDR labels upserted as metadata into the ipcache from the nodemanager with the same `Local` source because the node being handled in the local node.

In this test case, we are testing the association of the kube-apiserver label which always comes with the `KubeAPIServer` source. If we were testing the association of CIDR labels, I'd agree with keeping `Local`. It's technically incorrect to use `Local` with kube-apiserver. Because `source` only plays a factor for the _actual_ ipcache entry itself, not the identity, it has no functional impact in this test, so in theory it does not make the bug more likely, AFAIU. The unit test is deterministic.

Anyway, I think this is a very minor thing because there's no functional impact. I was more calling it out from the perspective of a future reader potentially getting confused and thinking there's a bug in the test.

christarazi (approved):

pkg/ipcache/metadata.go
christarazi: Out of curiosity, how did you discover the potential bug?

squeed: > Out of curiosity, how did you discover the potential bug?

I was thinking about more possible transitions (after our in-person conversation) and I realized I had missed a case. I also added this to the unit test.

As a hobby project, it would be fun to try and fuzz the ipcache in general. I'm thinking of a way to invent a sequence of operations and invariants.

pkg/ipcache/metadata_test.go
squeed: Added a comment to this effect -- that the source is normally different, but it's the same in the test case to maximize the chance of triggering the bug.

squeed:
FYI @joestringer I was chatting with @christarazi, then I realized a few hours later that I had missed a case (albeit unlikely): if an IP loses `reserved:host` but keeps, say, `reserved:kube-apiserver` or `cidr:xxxx`

pkg/ipcache/metadata.go
christarazi: Sounds like a potentially nice-to-have item that could go in https://github.com/cilium/cilium/issues/21142.

PR #28341, opened 2023-09-30
docs: Remove bare URLs from Flow gRPC API Reference

kimstacy:
The **[Flow gRPC API Reference](https://docs.cilium.io/en/stable/_api/v1/flow/README/)** contains bare URLs to Cilium's source code on GitHub and the W3C Trace Context specification. This changes the non-tabled bare URLs to Markdown links for an improved reading experience. Following this change, users can click on the links and directly access a page instead of having to copy and paste URLs. This also adds punctuation in several places for consistency.

I tested these changes locally.

### DIFF
--- a/api/v1/flow/README.md
+++ b/api/v1/flow/README.md
@@ -126,7 +126,7 @@ CiliumEventType from which the flow originated.
 <a name="flow-DNS"></a>
 
 ### DNS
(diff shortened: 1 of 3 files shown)


kimstacy:
Closing with intent to re-submit in the future.

PR #28343, opened 2023-10-01
Check for cilium.sock in /healthz endpoint

### LINKED ISSUE #27436: Check for cilium.sock in /healthz endpoint

### What happened?

The cilium health-server relies on cilium.sock to query the cilium-agent daemon for status information to report health. If this file is deleted, the daemon's `/healthz` endpoint will still return 200, but tools making use of the health-server such as the agent's embedded cilium CLI will not be able to return any status information. The health-server will also be unable to check the health of other nodes in the cluster, and cilium-agent will produce warning logs as follows:

```
level=error msg="unable to get cluster nodes" error="unable to get nodes' cluster: Get \"http://localhost/v1/cluster/nodes\": dial unix /var/run/cilium/cilium.sock: connect: no such file or directory" subsys=health-server
```

### Cilium Version

v1.14.0

### Kernel Version

5.15.0-1033-gke

### Kubernetes Version

v1.26.5-gke.1400

### Sysdump

_No response_

### Relevant log output

```shell
level=error msg="unable to get cluster nodes" error="unable to get nodes' cluster: Get \"http://localhost/v1/cluster/nodes\": dial unix /var/run/cilium/cilium.sock: connect: no such file or directory" subsys=health-server
```

```
root@xxx:/home/cilium# cilium status
Get "http://localhost/v1/healthz": dial unix /var/run/cilium/cilium.sock: connect: no such file or directory
Is the agent running?
```

chaunceyjiang:
Fixes #27436

      GitHub issue.

Fixes: #issue-number

```release-note
Check for cilium.sock in /healthz endpoint
```

### DIFF
--- a/pkg/health/server/health.go
+++ b/pkg/health/server/health.go
@@ -4,6 +4,9 @@
 package server
 
 import (
+	"fmt"
+	"os"
+
 	"github.com/go-openapi/runtime/middleware"
 
 	healthModels "github.com/cilium/cilium/api/v1/health/models"
@@ -11,6 +14,7 @@ import (
 	ciliumModels "github.com/cilium/cilium/api/v1/models"
 	"github.com/cilium/cilium/pkg/api"
 	"github.com/cilium/cilium/pkg/client"
+	"github.com/cilium/cilium/pkg/option"
 )
 
 type getHealthz struct {
@@ -33,7 +37,9 @@ func (h *getHealthz) getCiliumStatus() (*ciliumModels.StatusResponse, error) {
 // Handle handles GET requests for /healthz .
 func (h *getHealthz) Handle(params GetHealthzParams) middleware.Responder {
 	log.Debug("Handling request for /healthz")
-
+	if _, err := os.Stat(option.Config.SocketPath); os.IsNotExist(err) {
+		return api.Error(GetHealthzFailedCode, fmt.Errorf("missing socket: %s does not exist", option.Config.SocketPath))
+	}
 	ciliumStatus, err := h.getCiliumStatus()
 	if err != nil {
 		return api.Error(GetHealthzFailedCode, err)


chaunceyjiang:
before:
image

chaunceyjiang:
After:

image

pkg/health/server/health.go
```diff
@@ -33,7 +37,9 @@ func (h *getHealthz) getCiliumStatus() (*ciliumModels.StatusResponse, error) {
 // Handle handles GET requests for /healthz .
 func (h *getHealthz) Handle(params GetHealthzParams) middleware.Responder {
 	log.Debug("Handling request for /healthz")
-
+	if _, err := os.Stat(option.Config.SocketPath); os.IsNotExist(err) {
+		return api.Error(GetHealthzFailedCode, fmt.Errorf("%s\nIs the agent running?", option.Config.SocketPath))
+	}
```
joestringer, line 42: Shouldn't this be implemented client-side? If the `GET /healthz` request is already being handled then the CLI has successfully connected to the server already.

chaunceyjiang: > The cilium health-server relies on cilium.sock to query the cilium-agent daemon for status information to report health. If this file is deleted, the daemon's /healthz endpoint will still return 200

To be honest, I'm not sure whether it should be implemented on the client-side or server-side.

@thorn3r What do you think?

chaunceyjiang: I have a question and would like to ask you about it.

> Shouldn't this be implemented client-side? If the GET /healthz request is already being handled then the CLI has successfully connected to the server already.



I didn’t quite understand what you meant. Are you suggesting implementing it on the server-side? Does server-side refer to this?

https://github.com/cilium/cilium/blob/main/daemon/cmd/status.go#L734

@joestringer

joestringer: I am suggesting to implement client-side. If this is implemented server-side, then that means that the client must first connect to the server to find out if this socket exists. But then if this is actually important for the connection to be successful then connecting to the server will also fail. So then there's no point in implementing an improvement on the server side because the client cannot connect in order for the new logic to work.

joestringer: Related, I'm actually confused by your example above. How is the "before:" case successful if the socket is missing? Maybe that command doesn't actually require this socket in order to work?

joestringer: Maybe @thorn3r has more context here on the right solution.

thorn3r: I think server-side is the way to go here. The idea is to update /healthz to report an error when the socket is missing so that kubelet probes can detect this and restart the agent.

@joestringer the embedded `cilium` tool in the agent uses the socket to query the health-server and would fail, whereas `cilium-health` just hits the rest endpoint. So as it stand currently, with the socket missing, `cilium-health` would report healthy but `cilium status` would fail.

pkg/health/server/health.go
```diff
@@ -33,7 +37,9 @@ func (h *getHealthz) getCiliumStatus() (*ciliumModels.StatusResponse, error) {
 // Handle handles GET requests for /healthz .
 func (h *getHealthz) Handle(params GetHealthzParams) middleware.Responder {
 	log.Debug("Handling request for /healthz")
-
+	if _, err := os.Stat(option.Config.SocketPath); os.IsNotExist(err) {
+		return api.Error(GetHealthzFailedCode, fmt.Errorf("%s\nIs the agent running?", option.Config.SocketPath))
```
thorn3r:
```suggestion
		return api.Error(GetHealthzFailedCode, fmt.Errorf("missing socket: %s does not exist", option.Config.SocketPath))
```
I see that you followed the pattern of the CLI, which I think makes sense there, but here I think it would be better to return a single-line error that just indicates what the problem is (the CLI adds  "Is the agent running?" as a helpful message, but in this context we are running within the agent).

thorn3r (changes requested):
thanks for taking on this issue! I think the changes here look good, aside from just changing the wording of the error a bit.

pkg/health/server/health.go
chaunceyjiang: @thorn3r Updated. Thanks~

pkg/health/server/health.go
chaunceyjiang: > The idea is to update /healthz to report an error when the socket is missing so that kubelet probes can detect this and restart the agent.

Yes, that's exactly my idea.

thorn3r (approved):

chaunceyjiang:
Hi, @joestringer Please take a look~ Thanks~.

PR #28352, opened 2023-10-02
Split mapstate keys into allow and deny

bimmlerd:
The first commit converts the MapState type from a newtype an interface (mostly Nate's work, rebased on newer `main`)

The second commit then splits the map in the struct into two - to track allows and denies seperately. 

This allows for an optimization in the third commit: `denyPreferredInserts` for the allow case, now doesn't have to loop through all other allows (which we don't care about).

Commit msgs for convenience:

`policy: Make MapState an interface`
    
    `MapState` has many methods, some of which are public.
    In order to keep state to accomplish creating a cohesive
    map state that can keep state besides what will be put into
    an ebpf map, `MapState` can no longer be a type declaration
    for a golang map, but must be a struct that keeps state.
    
    The endpoint package must also be refactored to use `MapState`
    as an interface rather than a instantiated type of a map.

`mapstate: split allows and denies`
    
    Now that mapState is a struct, we can split the tracking of allows and
    denies. This commit should not introduce functional changes, but
    prepares us for an optimization in a later commit.
   
`mapstate: optimize denyPreferredInsert`
    
    Since the denyPreferredInsert logic cleanly splits on whether the to be
    inserted entry is a deny entry, we can avoid iterating over some
    entries: If we are inserting a deny, we are mostly interested in
    iteration over existing allows, and vice versa.
    
    In the case where we have a FQDN policy for a hostname which resolves to
    many IPs (for example with a fqdn policy covering Amazon's S3), this
    becomes a hot function (without performing a lot of useful work). The S3
    workload is expected to be heavily skewed towards allow entries for the
    many CIDR identities allocated for S3. When we insert such an entry, we
    only need to iterate deny entries, of which there are expected to be
    much fewer. Now that the tracking is split, we can employ this fact and,
    instead of iterating all entries, iterate only the deny entries.

### DIFF
--- a/pkg/policy/mapstate.go
+++ b/pkg/policy/mapstate.go
@@ -72,7 +72,8 @@ type MapState interface {
 
 // mapState is a state of a policy map.
 type mapState struct {
-	keys map[Key]MapStateEntry
+	allows map[Key]MapStateEntry
+	denies map[Key]MapStateEntry
 }
 
 type Identities interface {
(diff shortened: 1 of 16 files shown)


pkg/policy/mapstate.go
```diff
@@ -220,32 +244,176 @@ func getNets(identities Identities, ident uint32) []*net.IPNet {
...
+		return v, ok
+	}
+}
+
+// Insert the Key and matcthing MapStateEntry into the
+// MapState
+func (ms *mapState) Insert(k Key, v MapStateEntry) {
+	isDeny := v.IsDeny
+	if ms != nil && ((isDeny && ms.denies != nil) || (!isDeny && ms.allows != nil)) {
+		if isDeny {
```
joamaki: We probably want to remove the key from the other map here.
Why do we check for `denies != nil` and `allows != nil`?

pkg/policy/mapstate.go
```diff
@@ -220,32 +244,176 @@ func getNets(identities Identities, ident uint32) []*net.IPNet {
...
+	}
+	return m
+}
+
+// Get the MapStateEntry that matches the Key.
+func (ms *mapState) Get(k Key) (MapStateEntry, bool) {
+	if ms == nil || (ms.allows == nil && ms.denies == nil) {
+		return MapStateEntry{}, false
+	}
+	if ms.denies == nil {
```
joamaki: When would these be nil? I would just simplify this down to first checking denies and then checking allows.

pkg/policy/resolve.go
```diff
@@ -133,13 +135,35 @@ func (p *selectorPolicy) DistillPolicy(policyOwner PolicyOwner, isHost bool) *En
...
+
+// SetPolicyMap sets the policy map state as the interface
+// MapState. If the main argument is nil, then this method
+// will initialize a new MapState object for the caller.
+func (p *EndpointPolicy) SetPolicyMap(ms MapState) error {
+	if ms == nil {
+		p.policyMapState = newMapState(nil)
+		return nil
+	}
+	v, ok := ms.(*mapState)
```
joamaki: Couldn't we just keep `policyMapState` as `MapState` instead of `*mapState`?

pkg/policy/mapstate.go
```diff
@@ -220,32 +244,176 @@ func getNets(identities Identities, ident uint32) []*net.IPNet {
...
+			ms.denies[k] = v
+		} else {
+			ms.allows[k] = v
+		}
+	}
+}
+
+// Delete removes the Key an related MapStateEntry.
+func (ms *mapState) Delete(k Key) {
+	if ms != nil {
```
joamaki: Why do we check for nil here and every other method (except `Len` for some reason)? Do we rely on being able to use `*mapState` with nil value somewhere?

pkg/policy/mapstate.go
```diff
@@ -220,32 +244,176 @@ func getNets(identities Identities, ident uint32) []*net.IPNet {
...
+func (ms *mapState) addDependentOnEntry(owner Key, e MapStateEntry, dependent Key, changes ChangeState) {
 	if _, exists := e.dependents[dependent]; !exists {
 		if changes.Old != nil {
-			changes.Old[owner] = e
+			changes.Old.Insert(owner, e)
 		}
 		e.AddDependent(dependent)
-		keys[owner] = e
+		if e.IsDeny {
+			ms.denies[owner] = e
```
joamaki, line 372: We probably should delete `owner` from the other map to make sure we don't end up 
with duplicate keys.

aditighag (approved):
Nice refactor! 🚀 `pkg/endpoint` changes look good.

joamaki (approved):

nathanjsweet (approved):

jrajahalme:
@bimmlerd Changed v1.12 backport label to needs-backport/1.12 to be able to complete the release process. Looks like we can't have any pending backports when doing the release.

bimmlerd:
removing backport-pending labels here, as the project has decided against backporting this change -- it's too invasive for extensive backporting.

PR #28354, opened 2023-10-02
Improve initial waiting for node information with informers

### LINKED ISSUE #26833: Use informer for getting node information instead of periodic polling during agent initialization
## Current implementation
Right now, the agent waits for node information with exponential backoff - https://github.com/cilium/cilium/blob/e55a7ea7148fbc135011f8a3896726bca0a16018/pkg/k8s/init.go#L35C1-L41C3
which makes periodic calls to Kubernetes API: https://github.com/cilium/cilium/blob/12ea4bbfe14ca5b00f5124b5782c97a7d9fe20f4/pkg/k8s/watchers/cilium_node.go#L169C2-L172C22

This results in:
A) Unnecessary load on Kubernetes API
B) Can delay node initialization by up to 120s, for example when the agent waits for podCIDR allocation done by the operator

## Desired implementation

Use informer instead so we don't make unnecessary calls do Kubernetes API and also react to updates almost instantly.

marseel:
Before, we were doing periodic calls to k8s control plane with exponential backoff. Now, we use informers to get updates about Node/CiliumNode, which allows us to react much faster after cilium-operator assigns CIDR to node.

Fixes: https://github.com/cilium/cilium/issues/26833

```release-note
Improve cilium-agent bootstrap time when using cluster-pool ipam. 
```


daemon/k8s/resources.go
```diff
@@ -96,3 +104,10 @@ type Resources struct {
 	CiliumClusterwideNetworkPolicies resource.Resource[*cilium_api_v2.CiliumClusterwideNetworkPolicy]
 	CIDRGroups                       resource.Resource[*cilium_api_v2alpha1.CiliumCIDRGroup]
 }
+
```
derailed, line 107: nit: Comment?

daemon/k8s/init.go
```diff
@@ -154,12 +129,11 @@ func WaitForNodeInformation(ctx context.Context, k8sGetter k8sGetter) error {
...
 		useNodeCIDR(n)
-
-		restoreRouterHostIPs(n)
+		restoreRouterHostIPs(n, log)
 	} else {
 		// if node resource could not be received, fail if
 		// PodCIDR requirement has been requested
-		if option.Config.K8sRequireIPv4PodCIDR || option.Config.K8sRequireIPv6PodCIDR {
+		if requireIPv4CIDR || requireIPv6CIDR {
 			log.Fatal("Unable to derive PodCIDR via Node or CiliumNode resource, giving up")
```
derailed: I understand this line is not part of this PR but seems a bit odd to fatal here while the function returns an error? Might be better to propagate an error and let the call site decide how to best handle it?

derailed (commented):
@marseel Tx Marcel. Nice work! Just a couple comments...

daemon/k8s/resources.go
marseel: Done

daemon/k8s/init.go
marseel: Changed. 
Both callers to this function should handle returned error properly.

youngnick (approved):

PR #28355, opened 2023-10-02
[v1.14] GHA: Add clustermesh upgrade and downgrade tests

giorio94:
This PR introduces a new GitHub actions workflow which validates the Cilium upgrade and downgrade paths when clustermesh is enabled, ensuring that one minor version skew is correctly supported.
    
More specifically, the test initially deploys a mixed version cluster mesh, with one cluster running Cilium from the tip of the previous stable branch, and the other the tip of the current stable branch. At this point, the first cluster is upgraded to the tip of the current stable branch, enabling kvstoremesh at the same time. Finally, the cluster is downgraded again to the tip of the previous stable branch, disabling kvstoremesh.  

We do not enable the connectivity disruption checks in this workflow, because Cilium versions older than 1.14 are currently affected by known issues causing possible brief connectivity distruption of cross-cluster traffic when the agents get restarted.  

To reduce the total amount of time required by this workflow, only a limited subset of tests is enabled, while the full suite is run in the conformance-clustermesh workflow.

This is a stripped-down version of https://github.com/cilium/cilium/pull/27232, without the connectivity disruption checks.

### DIFF
--- a/.github/ariane-config.yaml
+++ b/.github/ariane-config.yaml
@@ -14,6 +14,7 @@ triggers:
     - conformance-ingress.yaml
     - conformance-runtime.yaml
     - integration-test.yaml
+    - tests-clustermesh-upgrade.yaml
     - tests-datapath-verifier.yaml
     - tests-l4lb.yaml
   /ci-aks:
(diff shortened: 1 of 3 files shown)


ti-mo (approved):
Rubber-stamping for Top Hat.

brlbil (approved):

giorio94:
Reviews are in, marking as ready to merge, given that it touches only GHA workflows.

giorio94:
The workflow seems pretty stable, let's mark it for backport to earlier branches.

PR #28359, opened 2023-10-02
Updated script location

ASHspace:
The path/location for docker script was not correct. So it was updated with this change request.

Fixes: Wrong Path/Location


ASHspace:
wrong version

PR #28369, opened 2023-10-03
Hive obj output improvements

bimmlerd:
A few improvements to the hive obj graph. Mostly work by Jussi I'm shepherding through while we hack on other stuff.


pkg/hive/internal/reflect.go
```diff
@@ -30,7 +31,25 @@ func FuncNameAndLocation(fn any) string {
...
+		return fmt.Sprintf("%s (%s:%d)", name, usefulPathSegment(file), line)
 	}
 	return name
 }
+
+// Purely a heuristic.
+var commonRoots = map[string]struct{}{
+	"pkg": {},
+	"cmd": {},
+}
```
giorio94, line 43: What about also adding `api`?

giorio94 (approved):
As far as I can tell, the PR looks good to me, just one nit inline. Unclear to me why I was pulled in as reviewer though, as I'm not part of sig-foundations :thinking:

dylandreimerink (approved):

pkg/hive/internal/reflect.go
dylandreimerink: I don't think we have that many `api/*` objects in hive at the moment, do we? We can always add this later on.

PR #28373, opened 2023-10-03
workload struct added and solved hubble metrics workload missing problem

sujoshua:
rebased, original pr #27974

**Which issue(s) this PR fixes:**
issue #25676

**Does this PR introduce a user-facing change?**
```
ACTION REQUIRED: to fix hubble metrics workload missing problem, user must update ciliumEndpoint CRD to get fixed this issue 
```


chancez (approved):

api/v1/openapi.yaml
```diff
@@ -3715,6 +3718,23 @@ definitions:
...
+      +deepequal-gen=true
+      +k8s:deepcopy-gen=true
+    properties:
+      name:
+        description: workload's name
+        type: string
+      namespace:
+        description: workload's namespace
+        type: string
+      kind:
```
squeed, line 3821: suggestion: a bare `kind` is insufficient; this should also have a `group` field.

sujoshua: hubble's workload struct is 
https://github.com/cilium/cilium/blob/a0899b9aeb6f14fe438c76fd49396b51fb921733/api/v1/flow/flow.pb.go#L2070
```
type Workload struct {
    state         protoimpl.MessageState
    sizeCache     protoimpl.SizeCache
    unknownFields protoimpl.UnknownFields
    Name          string `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`
    Kind          string `protobuf:"bytes,2,opt,name=kind,proto3" json:"kind,omitempty"`
}
```
Since `Group` filed originally is not added here, I didn't consider `Group` filed there either.
So do we need to change this api to add `Group` filed？

jrajahalme (approved):

pkg/ipcache/ipcache.go
```diff
@@ -809,6 +811,15 @@ func (m *K8sMetadata) Equal(o *K8sMetadata) bool {
 			return false
 		}
 	}
+	if m.Workload != nil {
```
nathanjsweet: `if m.Workload.Equals(o.Workload)` is sufficient. The `Equals` receiver already checks for `nil`.

pkg/types/workload.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+// Workload Kubernetes pod’s workload info (workloads are: Deployment, Statefulset, Daemonset, ReplicationController, CronJob, Job, DeploymentConfig (OpenShift), etc).
+type Workload struct {
+	Name      string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,3,opt,name=namespace"`
+	Kind      string `json:"kind,omitempty" protobuf:"bytes,1,opt,name=kind"`
+}
+
+// Equal compare two workload is equal
```
nathanjsweet:
```suggestion
// Equals compare two workload is equal
```

pkg/types/workload.go
```diff
@@ -0,0 +1,125 @@
...
+func (w *Workload) Equal(o *Workload) bool {
```
nathanjsweet:
```suggestion
func (w *Workload) Equals(o *Workload) bool {
```

pkg/types/workload.go
```diff
@@ -0,0 +1,125 @@
...
+	if w == o {
+		return true
+	}
+	if (w == nil && o != nil) || (w != nil && o == nil) {
```
nathanjsweet:
```suggestion
	if w == nil || o == nil {
```

pkg/types/workload_test.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			workload, err := GetWorkloadDataFromPod(tt.pod)
+			if tt.expectOK != (err == nil) {
+				t.Fatalf("expected ok=%t, got ok=%t, return err shows %s", tt.expectOK, err == nil, err.Error())
+			}
+			if err == nil {
+				if !reflect.DeepEqual(workload, tt.wantWorkload) {
```
nathanjsweet: This is a good spot to use your equality method. Which would be good to test as well.

nathanjsweet (changes requested):
Sorry for the late review. Overall, this looks good. Small changes.

pkg/types/workload.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// default to the owner kind/name
+	workload := &Workload{
+		Name:      controllerRef.Name,
+		Namespace: pod.Namespace,
+		Kind:      controllerRef.Kind,
+	}
+
+	// heuristic for deployment detection
+	if workload.Kind == "ReplicaSet" && pod.Labels["pod-template-hash"] != "" && strings.HasSuffix(controllerRef.Name, pod.Labels["pod-template-hash"]) {
```
derailed, line 97: Might be a good to ease off `if/else if` and just return when we find a match.

pkg/types/workload_test.go
```diff
@@ -0,0 +1,356 @@
...
+func TestDeploymentMetadata(t *testing.T) {
+	controller := true
+	tests := []struct {
+		name         string
+		pod          *slim_corev1.Pod
+		wantWorkload *Workload
+		expectOK     bool
+	}{
+		{
+			name: "deployment-name-deploy",
```
derailed, line 94: Since we are naming the use cases why not leverage a map vs a slice?

pkg/endpoint/api.go
```diff
@@ -338,6 +338,20 @@ func (e *Endpoint) getNamedPortsModel() (np models.NamedPorts) {
 	return np
 }
 
+// getWorkloadModel returns the endpoint's Workload object.
+func (e *Endpoint) getWorkloadModel() *models.Workload {
+	w := e.Workload.Load()
```
derailed, line 343: What happens if the workload is nil?

pkg/endpoint/endpoint.go
```diff
@@ -1232,6 +1236,16 @@ func (e *Endpoint) GetPod() *slim_corev1.Pod {
 	return e.pod.Load()
 }
 
+// SetWorkload sets the workload related to this endpoint.
+func (e *Endpoint) SetWorkload(workload *types.Workload) {
+	e.Workload.Store(workload)
```
derailed, line 1293: Same here is workload is nil?

pkg/endpoint/endpoint.go
```diff
@@ -1232,6 +1236,16 @@ func (e *Endpoint) GetPod() *slim_corev1.Pod {
...
+}
+
+// GetWorkload retrieves the workload related to this endpoint
+func (e *Endpoint) GetWorkload() *types.Workload {
+	return e.Workload.Load()
```
derailed, line 1298: And here per above

pkg/hubble/parser/seven/parser.go
```diff
@@ -215,11 +214,9 @@ func (p *Parser) computeResponseTime(r *accesslog.LogRecord, timestamp time.Time
...
 func (p *Parser) updateEndpointWorkloads(ip netip.Addr, endpoint *flowpb.Endpoint) {
 	if ep, ok := p.endpointGetter.GetEndpointInfo(ip); ok {
-		if pod := ep.GetPod(); pod != nil {
-			workload, workloadTypeMeta, ok := utils.GetWorkloadMetaFromPod(pod)
-			if ok {
-				endpoint.Workloads = []*flowpb.Workload{{Kind: workloadTypeMeta.Kind, Name: workload.Name}}
-			}
+		if workload := ep.GetWorkload(); workload != nil {
+			endpoint.Workloads = []*flowpb.Workload{{Kind: workload.Kind, Name: workload.Name}}
+			return
```
derailed: Do we need the explicit return?

pkg/k8s/watchers/cilium_endpoint.go
```diff
@@ -185,10 +185,12 @@ func (k *K8sWatcher) endpointUpdated(oldEndpoint, endpoint *types.CiliumEndpoint
 		scopedLog.Debug("Endpoint is not well-known; skipping ipcache upsert")
 		return
 	}
+	workload := ciliumTypes.GetWorkloadFromModel(endpoint.Workload)
```
derailed, line 189: Do we need to log here if the workload is nil? Since we are logging when trying to retrieve workload from pod later on and logging if no workload can be located.

derailed (commented):
@sujoshua Nice work! Just a few comments...

api/v1/openapi.yaml
squeed: Aha, I didn't realize that. Hmm. Bummer. I wonder if any of the hubble contributors can chime in here.

chancez: I don't think users will use `group` in the metrics very much to be honest, so I think it's fine. That said, it might still be a good idea to add the `Group` into the endpoint.

aanm:
@sujoshua slight ping :)

aanm:
I've marked the Pull Request as draft. Once you have finished up working on it feel free to click on the "Ready for review" button.

image

Thank you

pkg/endpoint/api.go
sujoshua: Is ok if workload is nil:
1. Firstly i comment in `workload` struct definition that this filed may be nil to notice follwers,.
2. Secondly we cannot always get workload, because sometime theres is lack of ownerReference in pod metadata.
3. Thirdly, I set omitempty tag in every filed about workload, so when it comes  to encode/decode, encoder/decoder will handle this correctly!

And thank you for your advice, but due to my mistake(this pr is my master branch, I have to discard all changes to sync with current upstream to handle merging thing), I have to open another pr, sorry for that.

pkg/endpoint/endpoint.go
sujoshua: same as above

pkg/endpoint/endpoint.go
sujoshua: same as above

sujoshua:
Sorry for my late responce, recenly I have been busy with organizing a summit.

And appologize for a closed pr again. Cause this pr is my master branch, I have to discard all changes to sync with upstream.(New to git, I will never do this again :(

Once I resolve all conversations there, I'll open another pr. This won't take too long !

pkg/endpoint/api.go
sujoshua: Thanks to @gentoo-root , I finally try to get git.
slight ping, Is this ok?

api/v1/openapi.yaml
sujoshua: I'll try to do work on this later. Maybe commit in another new pr ?

squeed (approved):

gentoo-root (approved):

derailed (approved):
@sujoshua Thank you for the correction!

nathanjsweet:
@sujoshua Please sign-off on all of your commits. This is close to being merged.

nathanjsweet:
It looks like there are [some legitimate issues](https://github.com/cilium/cilium/actions/runs/6893429741/job/18765899299?pr=28373) in CI.

lmb:
Converting to draft until CI is sorted.

PR #28382, opened 2023-10-03
Metrics associated with a deleted node should not be reported

derailed:

      GitHub issue.

When a node is deleted from a cluster, metrics associated with that node are still being exported to prometheus. 
Short of restarting the agent, we want to dynamically delete these metrics when a node is removed from the cluster.

This PR ensures node_connectivity_status and node_connectivity_latency no longer report metrics for nodes that are no longer present on the cluster.

Fixes: #issue-number

```release-note
Cilium now properly deletes stale (deleted) nodes from the node_connectivity_status and node_connectivity_latency_seconds metrics, reducing metric cardinality.
```

### DIFF
--- a/pkg/metrics/interfaces.go
+++ b/pkg/metrics/interfaces.go
@@ -24,13 +24,14 @@ var (
 	NoOpMetric    prometheus.Metric    = &mockMetric{}
 	NoOpCollector prometheus.Collector = &collector{}
 
-	NoOpCounter     metricpkg.Counter                 = &counter{NoOpMetric, NoOpCollector}
-	NoOpCounterVec  metricpkg.Vec[metricpkg.Counter]  = &counterVec{NoOpCollector}
-	NoOpObserver    metricpkg.Observer                = &observer{}
-	NoOpHistogram   metricpkg.Histogram               = &histogram{NoOpCollector}
-	NoOpObserverVec metricpkg.Vec[metricpkg.Observer] = &observerVec{NoOpCollector}
-	NoOpGauge       metricpkg.Gauge                   = &gauge{NoOpMetric, NoOpCollector}
-	NoOpGaugeVec    metricpkg.Vec[metricpkg.Gauge]    = &gaugeVec{NoOpCollector}
+	NoOpCounter           metricpkg.Counter                       = &counter{NoOpMetric, NoOpCollector}
+	NoOpCounterVec        metricpkg.Vec[metricpkg.Counter]        = &counterVec{NoOpCollector}
+	NoOpObserver          metricpkg.Observer                      = &observer{}
+	NoOpHistogram         metricpkg.Histogram                     = &histogram{NoOpCollector}
+	NoOpObserverVec       metricpkg.Vec[metricpkg.Observer]       = &observerVec{NoOpCollector}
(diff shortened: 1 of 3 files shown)


pkg/metrics/interfaces.go
```diff
@@ -156,6 +157,24 @@ func (g *gauge) Opts() metricpkg.Opts { return metricpkg.Opts{} }
 
 // GaugeVec
 
+type gaugeDeletableVec struct {
```
chancez, line 160: Is it possible to define all this on `metricpkg.Vec` so it doesn't need to be implemented for every metric type individually?

derailed: Right! I've debated on this a bit. I feel it makes sense to distinguish metrics that could be deleted vs not as a marker. Most metrics should not fall under this constraint, so I think it make sense to introduce these specific types so devs are aware that the metrics categories could be potentially support deletions??

chancez: Im not sure that makes sense when all underlying metrics with labels can be deleted in prometheus/client-golang. Whether or not we want to distinguish is a different question, but in practice, the underlying API supports it, so I think we should just surface/forward that.

derailed: I had looked into this but noticed we already discriminate between Vec and DeletableVec. Tho prom/vec does support delete labels operations, we found it necessary to further qualify in our implementation, so figured best to follow that lead...

chancez: Yeah, I'm not familiar with the current cilium metrics subsystem so I'll defer to you and others on this.

christarazi: I think I'm leaning towards not having a specific type. While there is some precedent for having different implementations, my guess was that there was an explicit need. In this case, I don't see an explicit need. I don't think developers will all of sudden start introducing metric deletion logic for arbitrary metrics so it feels like overengineering to protect against something that we don't know is a problem. I'd bet that most developers didn't know you could even do this to begin with (like me) :smile:.

christarazi:
Just a minor nit on the release note

> When a node is deleted from a cluster, we must ensure prometheus metrics associated with deleted node are no longer reported. Notably: node_connectivity_status and node_connectivity_latency_seconds.

Typically we frame release notes as impact, so something like

> Cilium now properly deletes stale (deleted) nodes from the node_connectivity_status and node_connectivity_latency_seconds metrics, reducing metric cardinality.

pkg/metrics/interfaces.go
derailed: I understand the concerns, but think tho a 're-alignement' is in order it would be a bit more intrusive to implement given the current distinctions between Vec and DeletableVec . Hence figured this is the straighter shot for this PR's scope that would allow us to back port the fix.

christarazi (approved):
AFAIU, there's already a differentiation for metrics that can be deleted, so this PR is just following the pattern. IMO, we can merge this now to fix the ongoing problems with the node connectivity metrics, and then followup with a refactor.

ldelossa (approved):

chancez (approved):

squeed:
Checkpatch is complaining that the commit subject line is too long. Overridden.

mikejennings:
did this make it into the newest release?

PR #28389, opened 2023-10-04
workflows: cilium-config: parametrize egressgw helm values

jibi:


pchaigno (approved):
Shouldn't this be owned by ci-structure instead of contributing?

jibi:
> Shouldn't this be owned by ci-structure instead of contributing?

sounds like, I'll fill a PR for CODEOWNERS :+1:

PR #28391, opened 2023-10-04
bpf: lxc: remove stale ENABLE_IDENTITY_MARK ifdefs

julianwiedmann:
4de20fc2027a ("bpf: Pass security ID via skb->cb from lxc to host") switched these parts from using set_identity_mark() to set_identity_meta(). Thus we no longer have to respect the opt-in that controls whether the skb mark may be used to store the identity.

### DIFF
--- a/bpf/bpf_lxc.c
+++ b/bpf/bpf_lxc.c
@@ -663,9 +663,7 @@ pass_to_stack:
 		ret = set_ipsec_encrypt_mark(ctx, encrypt_key, tunnel_endpoint);
 		if (unlikely(ret != CTX_ACT_OK))
 			return ret;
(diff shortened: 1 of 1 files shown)


julianwiedmann:
I'm undecided on whether this is a bug fix vs enhancement. Opinions? :)

gentoo-root (approved):
>I'm undecided on whether this is a bug fix vs enhancement. Opinions? :)

Looks more like an enhancement to me.

julianwiedmann:
> > I'm undecided on whether this is a bug fix vs enhancement. Opinions? :)
> 
> Looks more like an enhancement to me.

Given that it most likely didn't work before either? Yeah that's fair, thank you! We can always backport later if needed.

PR #28393, opened 2023-10-04
devices: Remove logging and report reason in device struct

joamaki:
With runtime device detection we may see a lot of churn in devices, addresses and routes which can lead to a lot of log spam from the devices controller.

Since the reason why a certain device is not selected for use in the datapath is important when debugging we don't want to just remove logging and call it a day, so instead this patch moves the information into a field in the Device struct that can be then inspected with "ciliumctl statedb dump" and which is included in sysdumps:

```
  {"Index":7,"MTU":1500,"Name":"br0","HardwareAddr":"86:ab:d8:9b:f9:33",
   "Flags":19,"Addrs":[{"Addr":"192.168.5.1","Scope":0}],"RawFlags":69699,
   "Type":"bridge","MasterIndex":0,
   "Selected":false,
   "NotSelectedReason":"bridge-like device, use --devices to override"}
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Also included is a fix to the "HardwareAddr" marshalling which now shows up as expected in JSON instead of as: "HardwareAddr":"vpVG/hUg".

### DIFF
--- a/pkg/datapath/linux/devices_controller.go
+++ b/pkg/datapath/linux/devices_controller.go
@@ -19,6 +19,7 @@ import (
 	"github.com/vishvananda/netns"
 	"golang.org/x/exp/slices"
 	"golang.org/x/sys/unix"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/asm"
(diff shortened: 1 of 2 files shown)


borkmann (approved):
awesome, thanks! 🚀

bimmlerd (approved):
:rocket:

PR #28395, opened 2023-10-04
Modularize bandwidth manager

bimmlerd:
Upstreaming some modularization changes while working on runtime device detection, these are in preparation of removing `option.Config.GetDevices` and replacing it with a devices table, which requires depending subsystems to be in hive.

In theory, this should not contain functional changes, but in practice there's some questions around the order of operations, specifically in the commit modularizing the bandwidth manager.


dylandreimerink:
Taking over this work, I moved to using a branch in the cilium repo so this can be worked on collectively if needed, new PR here: #28619

Issue #26739, opened 2023-07-10
Drop due to missed tail call when doing downgrade from main to v1.13.4


brb:
Caught in https://github.com/cilium/cilium/actions/runs/5506830267/jobs/10036019281.

```
cilium-bugtool-cilium-7n6nf-20230710-105913/cmd/cilium-metrics-list.md
124:cilium_drop_bytes_total                               direction="INGRESS" reason="Missed tail call"                                                                                                                                                                                                               322.000000
137:cilium_drop_count_total                               direction="INGRESS" reason="Missed tail call"                                                                                                                                                                                                               1.000000
```

```
1737:2023-07-10T10:45:49.823655924Z level=warning msg="Map type mismatch for BPF map" file-path=/sys/fs/bpf/tc/globals/cilium_policy_02097 new=Hash old="Longest prefix match trie" subsys=bpf
1738:2023-07-10T10:45:49.823663489Z level=warning msg="Key-size mismatch for BPF map" file-path=/sys/fs/bpf/tc/globals/cilium_policy_02097 new=8 old=12 subsys=bpf
1739:2023-07-10T10:45:49.823666855Z level=warning msg="Removing map to allow for property upgrade (expect map data loss)" file-path=/sys/fs/bpf/tc/globals/cilium_policy_02097 subsys=bpf
1812:2023-07-10T10:45:49.842068276Z level=warning msg="Map type mismatch for BPF map" file-path=/sys/fs/bpf/tc/globals/cilium_policy_00603 new=Hash old="Longest prefix match trie" subsys=bpf
1813:2023-07-10T10:45:49.842071232Z level=warning msg="Key-size mismatch for BPF map" file-path=/sys/fs/bpf/tc/globals/cilium_policy_00603 new=8 old=12 subsys=bpf
1814:2023-07-10T10:45:49.842073967Z level=warning msg="Removing map to allow for property upgrade (expect map data loss)" file-path=/sys/fs/bpf/tc/globals/cilium_policy_00603 subsys=bpf
1823:2023-07-10T10:45:49.842117819Z level=warning msg="Map type mismatch for BPF map" file-path=/sys/fs/bpf/tc/globals/cilium_policy_00361 new=Hash old="Longest prefix match trie" subsys=bpf
1824:2023-07-10T10:45:49.842121826Z level=warning msg="Key-size mismatch for BPF map" file-path=/sys/fs/bpf/tc/globals/cilium_policy_00361 new=8 old=12 subsys=bpf
[...]

PR #28397, opened 2023-10-04
datapath: Pin CT tail call buffers

### LINKED ISSUE #28088: CI: test-conn-disrupt-client failed due to interrupted traffic during upgrade/downgrade
## CI failure

Failed job: https://github.com/cilium/cilium/actions/runs/6154793242/job/16700697606

```
 [=] Test [no-interrupted-connections]

  [-] Scenario [no-interrupted-connections/no-interrupted-connections]
  🟥 Pod test-conn-disrupt-client-cc557fcf9-8sjnb flow was interrupted (restart count does not match 2 != 3)
```

[cilium-sysdump-2-final.zip](https://github.com/cilium/cilium/files/12583812/cilium-sysdump-2-final.zip)

This issue is different from [Drop due to missed tail call when doing downgrade from main to v1.13.4](https://github.com/cilium/cilium/issues/26739) because there is no `Missed tail call` in cilium-bugtool/cmd/cilium-metrics-list.md.

This issue is not an IPsec related issue because:
1. This failed check didn't enable IPsec, although the job name is `IPsec upgrade`. The job given above was triggered from https://github.com/cilium/cilium/pull/28086 which disabled IPsec.
2. XFRM counters didn't increase.

brb:
This commit pins the per-endpoint CT tail call buffer maps, so that they can be preserved during an endpoint regeneration (e.g., when cilium-agent restarts).

Previously, the buffers were recreated during the regeneration. The problem with this is that the TAIL_CT_LOOKUP* stores a CT lookup value in the buffer map, then does the tail call to a next program which retrieves the value from the map. As we don't atomically swap the whole prog map, it is possible that the next program is a new program with the new buffer map. So in this case, the CT lookup value cannot be restored, and the packet is dropped. This behavior was observed in the test
failures \[1\].

The proper fix is to always atomically swap the whole tail call (aka the prog map), and it is documented in \[2\]. Until \[2\] has been fixed, let's pin the buffer maps.

\[1\]: https://github.com/cilium/cilium/issues/28088
\[2\]: https://github.com/cilium/cilium/issues/20691

Fix #28088

### DIFF
--- a/bpf/bpf_host.c
+++ b/bpf/bpf_host.c
@@ -133,6 +133,7 @@ struct {
 	__type(key, __u32);
 	__type(value, struct ct_buffer6);
 	__uint(max_entries, 1);
+	__uint(pinning, LIBBPF_PIN_BY_NAME);
 } CT_TAIL_CALL_BUFFER6 __section_maps_btf;
 
 static __always_inline int
@@ -519,6 +520,7 @@ struct {
 	__type(key, __u32);
 	__type(value, struct ct_buffer4);
 	__uint(max_entries, 1);
+	__uint(pinning, LIBBPF_PIN_BY_NAME);
 } CT_TAIL_CALL_BUFFER4 __section_maps_btf;
 
 static __always_inline int
--- a/bpf/bpf_lxc.c
+++ b/bpf/bpf_lxc.c
@@ -360,6 +360,7 @@ struct {
 	__type(key, __u32);
 	__type(value, struct ct_buffer6);
 	__uint(max_entries, 1);
+	__uint(pinning, LIBBPF_PIN_BY_NAME);
 } CT_TAIL_CALL_BUFFER6 __section_maps_btf;
 
 /* Handle egress IPv6 traffic from a container after service translation has been done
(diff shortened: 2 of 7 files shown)


jschwinger233:
Is it possible this also fixes https://github.com/cilium/cilium/issues/26739

brb:
> Is it possible this also fixes https://github.com/cilium/cilium/issues/26739

Aye, good point! Let's drop the missed tail calls drop exclusion from the tests, and the future will show us.

pchaigno:
How could pinning the buffer map fix a missed tail call drop happening with the tail call map?

jschwinger233:
Paul is right. I was thinking the "root cause" (old bpf prog may tailcall new bpf prog) could also cause missed tailcall issue. This PR doesn't address the "old tailcalls new" issue, so it shouldn't fix that.

brb:
The upgrade failures in `ci-ipsec-upgrade` are expected until this PR is backported and released in the stable branches.

brb:
Putting back to draft - the host netdev substitutions need to be fixed too as pointer by Julian and Paul offline.

julianwiedmann:
Can we close this one, now that https://github.com/cilium/cilium/issues/20691 is merged? Or is there any aspect here that wasn't covered by https://github.com/cilium/cilium/issues/20691?

ti-mo:
This was an alternative proposal to #20691, closing.

thorn3r:
removing the `needs-backport` labels to clean up release tooling

PR #28400, opened 2023-10-04
ipsec: New Prometheus metrics for XFRM configs

pchaigno:
Add new Prometheus metrics to report the number of keys, XFRM IN/OUT states, and XFRM IN/OUT/FWD policies. See commits for details.


zacharysarah (approved):
@pchaigno Good work.

pkg/common/ipsec/utils_test.go
ti-mo, line 1: Why add new tests using `checkmate`? We should make these `testing.T` instead.

ti-mo (changes requested):

pkg/common/ipsec/utils_test.go
pchaigno: Because I didn't know about the switch.

ti-mo (approved):

pkg/common/ipsec/utils_test.go
```diff
@@ -27,6 +25,23 @@ func getXfrmState(src string, dst string, spi int, key string, mark uint32) netl
 			Key:    k,
 			ICVLen: 64,
 		},
+		Mark: &netlink.XfrmMark{
```
gentoo-root, line 28: Nit: looks like it could be part of the first commit.

gentoo-root (approved):
Looks good to me.

pkg/common/ipsec/utils_test.go
pchaigno: It's not needed until this third commit, which is why I didn't include it in the first. If we were to revert the third and fourth commits later, I'd want this to be removed as well.

gentoo-root: > If we were to revert the third and fourth commits later,

There would remain an unused parameter which could be misleading.

pchaigno: Yep, exactly.

pchaigno:
@squeed Note we don't receive notifications when labels are changed. Might be worth leaving a comment in that case :slightly_smiling_face:

nathanjsweet (approved):

joestringer:
Why was the `ready-to-merge` label manually set when this is a metrics PR and the @cilium/metrics codeowner hasn't approved the PR?

pchaigno:
@joestringer Because the author is tired of tracking down reviewers.

chancez (approved):
It seems like these should be separate metrics rather than one metric with a bunch of different labels, but this seems to be the already established pattern for these metrics, so I'm not going to block on that, but I would encourage considering an approach using separate metrics. 

Reasoning: usually you use labels for different dimensions of a given metric, and you can aggregate over the dimensions in a meaningful way. Given that each of these labels represents different "statistics", you can't really meaningfully aggregate them, which is why I suggest that they are broken apart into separate metrics instead.

Again, non-blocking because your following existing conventions, but it might be good to create a follow up to revisit these xfrm metrics as a whole.

pkg/datapath/linux/ipsec/xfrm_collector.go
```diff
@@ -101,5 +124,19 @@ func (x *xfrmCollector) Collect(ch chan<- prometheus.Metric) {
 	ch <- prometheus.MustNewConstMetric(x.xfrmErrorDesc, prometheus.GaugeValue, float64(stats.XfrmOutPolDead), labelErrorTypeOutbound, labelErrorPolicyDead)
 	ch <- prometheus.MustNewConstMetric(x.xfrmErrorDesc, prometheus.GaugeValue, float64(stats.XfrmOutPolError), labelErrorTypeOutbound, labelErrorPolicy)
 	ch <- prometheus.MustNewConstMetric(x.xfrmErrorDesc, prometheus.GaugeValue, float64(stats.XfrmOutStateInvalid), labelErrorTypeOutbound, labelErrorStateInvalid)
+}
 
+func (x *xfrmCollector) collectConfigStats(ch chan<- prometheus.Metric) {
+	states, err := x.xfrmStateFunc()
+	if err != nil {
+		log.WithError(err).Error("Error while getting XFRM states")
```
tommyp1ckles: nit: would it better to return the error here and log at the collect level with some context of what the outcome of this error is.

pkg/datapath/linux/ipsec/xfrm_collector.go
```diff
@@ -134,6 +138,20 @@ func (x *xfrmCollector) collectConfigStats(ch chan<- prometheus.Metric) {
...
 	nbKeys := ipsec.CountUniqueIPsecKeys(states)
 	ch <- prometheus.MustNewConstMetric(x.xfrmConfigStatDesc, prometheus.GaugeValue, float64(nbKeys), labelStatKeys)
+
+	nbStatesIn, nbStatesOut := ipsec.CountXfrmStatesByDir(states)
+	ch <- prometheus.MustNewConstMetric(x.xfrmConfigStatDesc, prometheus.GaugeValue, float64(nbStatesIn), labelStatStateIn)
+	ch <- prometheus.MustNewConstMetric(x.xfrmConfigStatDesc, prometheus.GaugeValue, float64(nbStatesOut), labelStatStateOut)
+
+	policies, err := x.xfrmPolicyFunc()
+	if err != nil {
+		log.WithError(err).Error("Error while getting XFRM policies")
```
tommyp1ckles: Same question here, would be nice to know that this only affect metrics

tommyp1ckles (approved):
Changes look good, sorry for the delay - was on PTO.

pchaigno:
> Reasoning: usually you use labels for different dimensions of a given metric, and you can aggregate over the dimensions in a meaningful way.

Makes sense. I've split the metrics for XFRM states and policies.

Issue #28415, opened 2023-10-05
CI: Runtime Test (privileged) - Test_NeighborGracefulRestart - `did not receive expected peering state ["ESTABLISHED"]`


gandro:
## CI failure

```
--- FAIL: Test_NeighborGracefulRestart (60.05s)
    --- FAIL: Test_NeighborGracefulRestart/add_neighbor_with_defaults (59.49s)
        neighbor_test.go:309: 
            	Error Trace:	/host/pkg/bgpv1/test/neighbor_test.go:309
            	            				/root/sdk/go1.21.1/src/runtime/asm_amd64.s:1650
            	Error:      	"[]" should have 1 item(s), but has 0
            	Test:       	Test_NeighborGracefulRestart/add_neighbor_with_defaults
        neighbor_test.go:325: 
            	Error Trace:	/host/pkg/bgpv1/test/neighbor_test.go:325
            	Error:      	Condition never satisfied
            	Test:       	Test_NeighborGracefulRestart/add_neighbor_with_defaults
    --- FAIL: Test_NeighborGracefulRestart/update_graceful_restart_with_defaults (0.00s)
        neighbor_test.go:298: 
            	Error Trace:	/host/pkg/bgpv1/test/neighbor_test.go:298
            	Error:      	Received unexpected error:
            	            	did not receive expected peering state ["ESTABLISHED"], context deadline exceeded
            	Test:       	Test_NeighborGracefulRestart/update_graceful_restart_with_defaults
    --- FAIL: Test_NeighborGracefulRestart/update_graceful_restart,_restart_time (0.00s)
        neighbor_test.go:298: 
            	Error Trace:	/host/pkg/bgpv1/test/neighbor_test.go:298
            	Error:      	Received unexpected error:
            	            	did not receive expected peering state ["ESTABLISHED"], context deadline exceeded
            	Test:       	Test_NeighborGracefulRestart/update_graceful_restart,_restart_time
    --- FAIL: Test_NeighborGracefulRestart/disable_graceful_restart (0.00s)
        neighbor_test.go:298: 
            	Error Trace:	/host/pkg/bgpv1/test/neighbor_test.go:298
            	Error:      	Received unexpected error:
            	            	did not receive expected peering state ["ESTABLISHED"], context deadline exceeded
            	Test:       	Test_NeighborGracefulRestart/disable_graceful_restart
FAIL
```

[...]

PR #28420, opened 2023-10-05
bgpv1,ci: Fix BGP component tests reusing the same VirtualRouter config

rastislavs:
As some of the BGP component tests mutate the `CiliumBGPVirtualRouter` config, we need to make sure that mutated config is not used by the subsequent tests. That was the case before this fix, as multiple tests are using the same `baseBGPPolicy` variable. Now each test creates a deep copy of the virtual routers  from the base config, which are fine to mutate within the test.

This change also ensures that `fixture.config.policy` always contains the most up-to-date policy, as it is used to update the policy in some tests. This could lead to incorrect updates of the `CiliumBGPVirtualRouter` config.

This can potentially fix the flake described in issue https://github.com/cilium/cilium/issues/28415, although I'm not sure, as I can not reproduce that issue locally. I have found this problem when developing a new BGP component test where the symptoms were similar.

### DIFF
--- a/pkg/bgpv1/test/adverts_test.go
+++ b/pkg/bgpv1/test/adverts_test.go
@@ -142,7 +142,7 @@ func Test_PodCIDRAdvert(t *testing.T) {
 	defer testDone()
 
 	// setup topology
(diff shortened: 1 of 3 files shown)


harsimran-pabla (approved):
Thank you!!

PR #28425, opened 2023-10-05
Print full labelset for all identities in 'cilium ip list' output

joestringer:
The host identity can have multiple labels that are not hardcoded directly from
the number, so we should just always fetch the full labelset from the agent for
all identities to ensure that the CLI prints what the agent knows. No need for
special cases for reserved identities.

Technically this bug goes way back and impacts any version since the kube-apiserver
entity policy feature has existed, but it's a minor bug so I'm only proposing this
for 1.14 at this time.

### DIFF
--- a/cilium-dbg/cmd/ip_list.go
+++ b/cilium-dbg/cmd/ip_list.go
@@ -91,11 +91,13 @@ func printEntry(w *tabwriter.Writer, entry *models.IPListEntry) {
 			params := ipApi.NewGetIdentityIDParams().WithID(identity).WithTimeout(api.ClientTimeout)
 			id, err := client.Policy.GetIdentityID(params)
 			if err != nil {
-				Fatalf("Cannot get identity for given ID %s: %s\n", id, err)
-			}
-			lbls := labels.NewLabelsFromModel(id.Payload.Labels)
-			for _, lbl := range lbls {
-				identities = append(identities, lbl.String())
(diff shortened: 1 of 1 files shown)


christarazi (approved):
:ship:

squeed (approved):

cilium/cmd/ip_list.go
```diff
@@ -80,27 +90,14 @@ func printEntry(w *tabwriter.Writer, entry *models.IPListEntry) {
...
-			lbls := labels.NewLabelsFromModel(id.Payload.Labels)
-			for _, lbl := range lbls {
-				identities = append(identities, lbl.String())
-			}
-		}
+		labels = append(labels, getLabels(ni)...)
 	}
 	first := true
-	for _, identity := range identities {
+	for _, identity := range labels {
```
derailed: nit: Since we've renamed identities-labels perhaps we should use label for the loop var for clarity?

derailed (approved):
@joestringer Nice work!

joestringer:
All required tests passing and reviews are in. I'll fix up the minor nit from Fernand and repush. Should be good to merge after that.

cilium/cmd/ip_list.go
joestringer: Fixed, good pointer :)

squeed:
All required tests are passing, button is green. Merging.

squeed:
Oh wait, I'll let @joestringer do the honors.

PR #28432, opened 2023-10-06
gha: explicit branch and trigger in ariane-scheduled workflow

giorio94:
Let's use explicit bash variables to construct the trigger phrase, instead of directly using the matrix value.


nbusseneau (approved):

giorio94:
Reviews are in. Marking as ready to merge, as full CI is not required.

PR #28438, opened 2023-10-06
Extend BPF unit tests for IPsec

jschwinger233:
Existing IPsec BPF test only covers tunnel=vxlan, so this PR covers combination of tunnel={vxlan,disabled} and endpointRoutes={disabled,enabled}. 

Fixes: https://github.com/cilium/cilium/pull/25699

### DIFF
--- a/bpf/tests/ipsec_from_host_generic.h
+++ b/bpf/tests/ipsec_from_host_generic.h
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/* SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause) */
 /* Copyright Authors of Cilium */
 
 #include "common.h"
@@ -13,11 +13,9 @@
 #define ENABLE_IPV4
 #define ENABLE_IPV6
 #define ENABLE_IPSEC
-#define TUNNEL_MODE
-#define HAVE_ENCAP
-#define ENCAP_IFINDEX 4
 #define SECCTX_FROM_IPCACHE 1
 
+#define ENCAP_IFINDEX 4
 #define skb_set_tunnel_key mock_skb_set_tunnel_key
 #define ctx_redirect mock_ctx_redirect
 
@@ -135,10 +133,13 @@ int ipv4_ipsec_from_host_check(__maybe_unused const struct __ctx_buff *ctx)
 		test_fatal("status code out of bounds");
 
 	status_code = data;
-	assert(*status_code == CTX_ACT_REDIRECT);
+	assert(*status_code == EXPECTED_STATUS_CODE);
 
 	assert(ctx->mark == 0);
+
+#ifdef CHECK_CB_ENCRYPT_IDENTITY
 	assert(ctx_load_meta(ctx, CB_ENCRYPT_IDENTITY) == 0);
+#endif
 
 	l2 = data + sizeof(*status_code);
 
@@ -253,10 +254,13 @@ int ipv6_ipsec_from_host_check(__maybe_unused const struct __ctx_buff *ctx)
 		test_fatal("status code out of bounds");
 
 	status_code = data;
-	assert(*status_code == CTX_ACT_REDIRECT);
+	assert(*status_code == EXPECTED_STATUS_CODE);
+
+#ifdef CHECK_CB_ENCRYPT_IDENTITY
+	assert(ctx_load_meta(ctx, CB_ENCRYPT_IDENTITY) == 0);
+#endif
 
 	assert(ctx->mark == 0);
-	assert(ctx_load_meta(ctx, CB_ENCRYPT_IDENTITY) == 0);
 
 	l2 = data + sizeof(*status_code);
 
--- a/bpf/tests/ipsec_from_host_tunnel.c
+++ b/bpf/tests/ipsec_from_host_tunnel.c
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/* Copyright Authors of Cilium */
+
+#define TUNNEL_MODE
+#define HAVE_ENCAP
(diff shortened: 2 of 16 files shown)


bpf/tests/ipsec_from_lxc_tunnel.c
```diff
@@ -0,0 +1,15 @@
...
+#include "common.h"
+#include <bpf/ctx/skb.h>
+#include "pktgen.h"
+#define ROUTER_IP
+#include "config_replacement.h"
+#undef ROUTER_IP
+
+#define TUNNEL_MODE
+#define HAVE_ENCAP
+#define ENABLE_ROUTING
```
pchaigno, line 6: I notice we have `ENABLE_ROUTING` here but not in `bpf/tests/ipsec_from_lxc_nativerouting.c`. Is that expected?

bpf/tests/ipsec_from_host_tunnel_endpointroutes.c
```diff
@@ -0,0 +1,46 @@
...
+}
+
+int mock_ctx_redirect(const struct __sk_buff *ctx __maybe_unused, int ifindex, __u32 flags)
+{
+	if (ifindex != ENCAP_IFINDEX)
+		return -1;
+	if (flags != 0)
+		return -2;
+	return CTX_ACT_REDIRECT;
+}
```
pchaigno: AFAIU, this is duplicated between `bpf/tests/ipsec_from_host_tunnel.c` and here. What would be the downside of keeping those two mocked helpers in the `_generic.c` file?

bpf/tests/ipsec_from_host_tunnel.c
```diff
@@ -0,0 +1,46 @@
...
+#include <bpf/ctx/skb.h>
+#include "pktgen.h"
+#define ROUTER_IP
+#include "config_replacement.h"
+#undef ROUTER_IP
+
+#define TUNNEL_MODE
+#define HAVE_ENCAP
+#define ENCAP_IFINDEX 4
+#define ENABLE_ROUTING
```
pchaigno, line 6: Same note here regarding `ENABLE_ROUTING`.

pchaigno (commented):
Nice commit history!

Minor comments only from me.

bpf/tests/ipsec_from_lxc_tunnel.c
jschwinger233: Yes, but I handled it wrong. Let me rectify.

In short, `ENABLE_ROUTING` means the endpoint requires BPF routing instead of stack routing. When endpointRoutes is enabled, the macro doesn't exist.

In detail, config.RequireRouting = false if option.Config.EnableEndpointRoutes:
https://github.com/cilium/cilium/blob/28a3cb7077aec3b03e54b873765cb764b0d46fa0/pkg/endpoint/endpoint.go#L965-L978
Then cilium skips `#define ENABLE_ROUTING 1`:
https://github.com/cilium/cilium/blob/28a3cb7077aec3b03e54b873765cb764b0d46fa0/pkg/datapath/linux/config/config.go#L1081-L1082

Back to my test cases, only ipsec_from_*_endpointroutes.c should NOT have this macro, while ipsec_from_nativerouting.c should have it, but I messed it up.

I'll also double check other macro definition differences among combinations.

bpf/tests/ipsec_from_lxc_generic.h
```diff
@@ -1,20 +1,11 @@
-// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/* SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause) */
```
markpash, line 1: The comment formatting has changed here and isn't consistent with other files.

bpf/tests/ipsec_from_host_generic.h
```diff
@@ -1,23 +1,14 @@
-// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/* SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause) */
```
markpash, line 1: Same here again with comment format change.

bpf/tests/ipsec_from_overlay_generic.h
```diff
@@ -1,19 +1,6 @@
-// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/* SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause) */
```
markpash, line 1: here too.

markpash (approved):

brb (approved):

bpf/tests/ipsec_from_overlay_generic.h
jschwinger233: They were intentionally changed, otherwise CI checkpatch (`make -C bpf checkpatch`) will complain:

```
ERROR:C99_COMMENTS: do not use C99 // comments
#7: FILE: bpf/tests/ipsec_from_host_generic.h:1:
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
```

bpf/tests/ipsec_from_lxc_tunnel.c
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/* Copyright Authors of Cilium */
+
+#include "common.h"
+#include <bpf/ctx/skb.h>
+#include "pktgen.h"
```
pchaigno: Why are these three includes not part of the `_generic.h`?

pchaigno (approved):

bpf/tests/ipsec_from_lxc_tunnel.c
jschwinger233: Just moved them to _generic.h

PR #28439, opened 2023-10-06
ci-e2e: add job testing node cidr feature

squeed:
Add a job that sets policy-cidr-match-mode=nodes.


squeed:
Hmm, CI doesn't seem to be picking up my new job. Does this need to be a branch on cilium/cilium?

PR #28441, opened 2023-10-06
bpf: encap: catch mis-scoped security identity during encapsulation

julianwiedmann:
The security identity is a __u32 value, but transporting it via the VNI field in the VXLAN / Geneve header allows for only 24 bits.

So when inserting the source's security identity into the tunnel header, make sure that we are not trying to transport a local-scoped identity across the tunnel.

### DIFF
--- a/bpf/lib/encap.h
+++ b/bpf/lib/encap.h
@@ -16,6 +16,8 @@
 
 #include "high_scale_ipcache.h"
 
+#define ENCAP_MAX_VNI	0x00ffffff
+
 #ifdef HAVE_ENCAP
 static __always_inline int
 __encap_with_nodeid(struct __ctx_buff *ctx, __u32 src_ip, __be16 src_port,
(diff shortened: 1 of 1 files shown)


bpf/lib/encap.h
```diff
@@ -32,6 +34,9 @@ __encap_with_nodeid(struct __ctx_buff *ctx, __u32 src_ip, __be16 src_port,
 	if (seclabel == HOST_ID)
 		seclabel = LOCAL_NODE_ID;
 
+	if (seclabel > ENCAP_MAX_VNI)
+		return DROP_INVALID_IDENTITY;
```
christarazi, line 38: What do we do if we see these drops? Does it signal that the `seclabel` (security identity) of the workload is somehow incorrect?

julianwiedmann: I would consider it as indication of a code bug.

Either the agent selecting a bad identity for the workload. Or the datapath trying to forward an non-eligible identity - I've caught myself wanting to forward a source identity in the N/S loadbalancer path, but not considering CIDR identities.

PR #28449, opened 2023-10-06
docs: remove annotations-based l7 visibility

networkop:
L7-based annotations have had little adoption and are causing a [lot of bugs](https://github.com/cilium/cilium/issues?q=is%3Aissue+proxy-visibility). This change is to move them to the background in the docs and bring forward the more standard way of enabling L7 visibility (CNP)

### DIFF
--- a/Documentation/observability/visibility.rst
+++ b/Documentation/observability/visibility.rst
@@ -12,63 +12,78 @@ Layer 7 Protocol Visibility
 
 .. note::
 
-    This feature requires enabling L7 Proxy support. Without it, the visibility annotation is ignored.
+    This feature requires enabling L7 Proxy support.
 
-While :ref:`monitor` provides introspection into datapath state, by default it
-will only provide visibility into L3/L4 packet events. If :ref:`l7_policy` are
-configured, one can get visibility into L7 protocols, but this requires the full
-policy for each selected endpoint to be written. To get more visibility into the
-application without configuring a full policy, Cilium provides a means of
-prescribing visibility via `annotations <https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/>`_
-when running in tandem with Kubernetes.
+While :ref:`monitor` provides introspection into datapath state, by default, it
+will only provide visibility into L3/L4 packet events. If you want L7
+protocol visibility, you can use L7 Cilium Network Policies (see :ref:`l7_policy`).
 
-Visibility information is represented by a comma-separated list of tuples in
-the annotation:
 
-``<{Traffic Direction}/{L4 Port}/{L4 Protocol}/{L7 Protocol}>``
+.. note::
 
-For example:
+    Historically, it had been possible to enable L7 visibility using Pod
+    annotations (``policy.cilium.io/proxy-visibility``). This method is
+    no longer supported and we recommend users to switch to L7 policies instead.
+
+To enable visibility for L7 traffic, create a ``CiliumNetworkPolicy`` that specifies
+L7 rules. Traffic flows matching a L7 rule in a ``CiliumNetworkPolicy`` will become
+visible to Cilium and, thus, can be exposed to the end user. It's important to 
+remember that L7 network policies not only enables visibility but also restrict 
+what traffic is allowed to flow in and out of a Pod.
+
+
+The following example enables visibility for DNS (TCP/UDP/53) and HTTP
+(ports TCP/80 and TCP/8080) traffic within the ``default`` namespace by
+specifying two L7 rules -- one for DNS and one for HTTP. It also restricts
+egress communication and drops anything that is not matched. L7 matching
+conditions on the rules have been omitted or wildcarded, which will
+permit all requests that match the L4 section of each rule:
+
+
+.. code-block:: yaml
+
+      apiVersion: "cilium.io/v2"
+      kind: CiliumNetworkPolicy
+      metadata:
+        name: "l7-visibility"
+      spec:
+        endpointSelector:
+          matchLabels:
+            "k8s:io.kubernetes.pod.namespace": default
+        egress:
+        - toPorts:
+          - ports:
+            - port: "53"
+              protocol: ANY
+            rules:
+              dns:
+              - matchPattern: "*"
+        - toEndpoints:
+          - matchLabels:
+              "k8s:io.kubernetes.pod.namespace": default
+          toPorts:
+          - ports:
+            - port: "80"
+              protocol: TCP
+            - port: "8080"
+              protocol: TCP
+            rules:
+              http: [{}]
+
+Based on the above policy, Cilium will pick up all TCP/UDP/53, TCP/80 and TCP/8080 
+egress traffic from Pods in the ``default`` namespace and redirect it to the 
+proxy (see :ref:`proxy_injection`) such that the output of ``cilium monitor`` or 
+``hubble observe`` shows the L7 flow details. 
+Below is the example of running ``hubble observe -f -t l7 -o compact`` command:
 
 ::
 
-  <Egress/53/UDP/DNS>,<Egress/80/TCP/HTTP>
+    default/testapp-5b9cc645cb-4slbs:45240 (ID:26450) -> kube-system/coredns-787d4945fb-bdmdq:53 (ID:9313) dns-request proxy FORWARDED (DNS Query web.default.svc.cluster.local. A)
+    default/testapp-5b9cc645cb-4slbs:45240 (ID:26450) <- kube-system/coredns-787d4945fb-bdmdq:53 (ID:9313) dns-response proxy FORWARDED (DNS Answer "10.96.118.37" TTL: 30 (Proxy web.default.svc.cluster.local. A))
+    default/testapp-5b9cc645cb-4slbs:33044 (ID:26450) -> default/echo-594485b8dc-fp57l:8080 (ID:32531) http-request FORWARDED (HTTP/1.1 GET http://web/)
+    default/testapp-5b9cc645cb-4slbs:33044 (ID:26450) <- default/echo-594485b8dc-fp57l:8080 (ID:32531) http-response FORWARDED (HTTP/1.1 200 4ms (GET http://web/))
 
 
-To do this, you can provide the annotation in your Kubernetes YAMLs, or via the
-command line, e.g.:
-
-.. code-block:: shell-session
-
-    kubectl annotate pod foo -n bar policy.cilium.io/proxy-visibility="<Egress/53/UDP/DNS>,<Egress/80/TCP/HTTP>"
-
-Cilium will pick up that pods have received these annotations, and will
-transparently redirect traffic to the proxy such that the output of
-``cilium-dbg monitor`` shows traffic being redirected to the proxy, e.g.:
-
-::
-
-    -> Request http from 1474 ([k8s:id=app2 k8s:io.kubernetes.pod.namespace=default k8s:appSecond=true k8s:io.cilium.k8s.policy.cluster=default k8s:io.cilium.k8s.policy.serviceaccount=app2-account k8s:zgroup=testapp]) to 244 ([k8s:io.cilium.k8s.policy.cluster=default k8s:io.cilium.k8s.policy.serviceaccount=app1-account k8s:io.kubernetes.pod.namespace=default k8s:zgroup=testapp k8s:id=app1]), identity 30162->42462, verdict Forwarded GET http://app1-service/ => 0
-    -> Response http to 1474 ([k8s:zgroup=testapp k8s:id=app2 k8s:io.kubernetes.pod.namespace=default k8s:appSecond=true k8s:io.cilium.k8s.policy.cluster=default k8s:io.cilium.k8s.policy.serviceaccount=app2-account]) from 244 ([k8s:io.cilium.k8s.policy.serviceaccount=app1-account k8s:io.kubernetes.pod.namespace=default k8s:zgroup=testapp k8s:id=app1 k8s:io.cilium.k8s.policy.cluster=default]), identity 30162->42462, verdict Forwarded GET http://app1-service/ => 200
-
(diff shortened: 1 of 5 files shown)


sayboras (approved):
Thanks ✅

nebril (approved):

Documentation/observability/visibility.rst
```diff
@@ -12,63 +12,73 @@ Layer 7 Protocol Visibility
...
-
-    kubectl annotate pod foo -n bar policy.cilium.io/proxy-visibility="<Egress/53/UDP/DNS>,<Egress/80/TCP/HTTP>"
+.. note::
 
-Cilium will pick up that pods have received these annotations, and will
-transparently redirect traffic to the proxy such that the output of
-``cilium monitor`` shows traffic being redirected to the proxy, e.g.:
+    Historically, it had been possible to enable L7 visibility using Pod
+    annotations (``policy.cilium.io/proxy-visibility``). This method is
+    no longer supported and we recommend users to switch to L7 policies instead.
```
learnitall, line 26: I'm not sure how I feel about adding a note like this. Would you be up for removing it? It might add some maintenance overhead, as we'd likely have to come back in a release or two and remove it anyway.

learnitall: This kind of note might make more sense for release notes or the upgrade guide

Documentation/observability/visibility.rst
```diff
@@ -12,63 +12,73 @@ Layer 7 Protocol Visibility
...
+the L7 flow details:
 
 ::
 
-    -> Request http from 1474 ([k8s:id=app2 k8s:io.kubernetes.pod.namespace=default k8s:appSecond=true k8s:io.cilium.k8s.policy.cluster=default k8s:io.cilium.k8s.policy.serviceaccount=app2-account k8s:zgroup=testapp]) to 244 ([k8s:io.cilium.k8s.policy.cluster=default k8s:io.cilium.k8s.policy.serviceaccount=app1-account k8s:io.kubernetes.pod.namespace=default k8s:zgroup=testapp k8s:id=app1]), identity 30162->42462, verdict Forwarded GET http://app1-service/ => 0
-    -> Response http to 1474 ([k8s:zgroup=testapp k8s:id=app2 k8s:io.kubernetes.pod.namespace=default k8s:appSecond=true k8s:io.cilium.k8s.policy.cluster=default k8s:io.cilium.k8s.policy.serviceaccount=app2-account]) from 244 ([k8s:io.cilium.k8s.policy.serviceaccount=app1-account k8s:io.kubernetes.pod.namespace=default k8s:zgroup=testapp k8s:id=app1 k8s:io.cilium.k8s.policy.cluster=default]), identity 30162->42462, verdict Forwarded GET http://app1-service/ => 200
+    default/testapp-5b9cc645cb-4slbs:45240 (ID:26450) -> kube-system/coredns-787d4945fb-bdmdq:53 (ID:9313) dns-request proxy FORWARDED (DNS Query web.default.svc.cluster.local. A)
```
learnitall, line 81: Can you add the command that was used to generate this output? I think it makes sense to be specific here, since I'm not sure if this is from `hubble observe` or `cilium monitor`

Documentation/observability/visibility.rst
```diff
@@ -12,63 +12,73 @@ Layer 7 Protocol Visibility
...
+              http: [{}]
+        - toEntities:
+          - all
+
+The final rule allowing all egress traffic (``toEntities: [ all ]``) is
+optional and can be removed or replaced with a more specific set of rules.
+
+Based on the above policy, Cilium will pick up all UDP/53, TCP/80 and TCP/8080 
+egress traffic from Pods in the ``default`` namespace and redirect it to the 
+proxy such that the output of ``cilium monitor`` or ``hubble observe`` shows
```
learnitall: Do we have documentation on the proxy? If so, can you link it here? This would help users who have no idea what the proxy is.

Documentation/observability/visibility.rst
```diff
@@ -12,63 +12,73 @@ Layer 7 Protocol Visibility
...
-``cilium monitor`` shows traffic being redirected to the proxy, e.g.:
+    Historically, it had been possible to enable L7 visibility using Pod
+    annotations (``policy.cilium.io/proxy-visibility``). This method is
+    no longer supported and we recommend users to switch to L7 policies instead.
+
+The following example demonstrates how to enable DNS and HTTP visibility for all
+traffic within the ``default`` namespace. We do that by creating a ``CilimNetworkPolicy`` 
+resource containing two L7 rules -- one for DNS and one for HTTP traffic. 
+If all you're interested in is visibility, you can omit or wildcard
+the L7 matching conditions as it is shown below:
```
learnitall: Can you adjust this description to be more specific to the ports that the policy mentions? A user might be confused reading this because you state that the policy will 'enable ... HTTP visibility for all traffic...' when it will only enable visibility for HTTP traffic flowing on ports 80 and 8080.

Also, can you rephrase this paragraph to refer to the reader as 'you' instead of as 'we'?

Documentation/security/policy/language.rst
```diff
@@ -827,7 +827,7 @@ latter rule will have no effect.
 .. note:: Layer 7 rules are not currently supported in `HostPolicies`, i.e.,
           policies that use :ref:`NodeSelector`.
 
-.. note:: Layer 7 policies --and pod annotations-- result in traffic being
+.. note:: Layer 7 policies result in traffic being
```
learnitall: Small suggestion, but not required: `Layer 7 policies will proxy traffic through an Envoy...`

Documentation/observability/visibility.rst
```diff
@@ -12,63 +12,73 @@ Layer 7 Protocol Visibility
...
-While :ref:`monitor` provides introspection into datapath state, by default it
-will only provide visibility into L3/L4 packet events. If :ref:`l7_policy` are
-configured, one can get visibility into L7 protocols, but this requires the full
-policy for each selected endpoint to be written. To get more visibility into the
-application without configuring a full policy, Cilium provides a means of
-prescribing visibility via `annotations <https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/>`_
-when running in tandem with Kubernetes.
+While :ref:`monitor` provides introspection into data path state, by default, it
+will only provide visibility into L3/L4 packet events. If you want to get L7
+protocol visibility you can use L7 Cilium Network Policies (see :ref:`l7_policy`).
```
learnitall: Small suggestions on wording and spelling:

```suggestion
While :ref:`monitor` provides introspection into datapath state, by default, it
will only provide visibility into L3/L4 packet events. If you want L7
protocol visibility, you can use L7 Cilium Network Policies (see :ref:`l7_policy`).
```

learnitall (changes requested):
Looks great! I just have a couple of suggestions. Also, can you add a note about this in the upgrade guide: https://github.com/cilium/cilium/blob/main/Documentation/operations/upgrade.rst#115-upgrade-notes?

Documentation/observability/visibility.rst
networkop: I feel like it makes sense to keep this note here at least for the 1.15 release. So that it's clear when we've decided to deprecate it (without going through release notes).  But it's not a strong opinion, so I'm happy to remove it.

Documentation/observability/visibility.rst
```diff
@@ -12,63 +12,74 @@ Layer 7 Protocol Visibility
...
+              http: [{}]
+        - toEntities:
+          - all
+
+The final rule allowing all egress traffic (``toEntities: [ all ]``) is
+optional and can be removed or replaced with a more specific set of rules.
+
+Based on the above policy, Cilium will pick up all UDP/53, TCP/80 and TCP/8080 
+egress traffic from Pods in the ``default`` namespace and redirect it to the 
+proxy (see :ref:`proxy_injection`) such that the output of ``cilium monitor`` or ``hubble observe`` shows
```
learnitall: Could you wrap line 71 so it matches the line length of the rest of the page?

Documentation/observability/visibility.rst
```diff
@@ -12,63 +12,74 @@ Layer 7 Protocol Visibility
...
-``cilium monitor`` shows traffic being redirected to the proxy, e.g.:
+    Historically, it had been possible to enable L7 visibility using Pod
+    annotations (``policy.cilium.io/proxy-visibility``). This method is
+    no longer supported and we recommend users to switch to L7 policies instead.
+
+The following example demonstrates how to enable visibility for all DNS (UDP/53)
+and HTTP (ports TCP/80 and TCP/8080) traffic within the ``default`` namespace. 
+You do that by creating a ``CilimNetworkPolicy`` resource containing two L7 rules 
+-- one for DNS and one for HTTP traffic. If all you're interested in is visibility, 
+you can omit or wildcard the L7 matching conditions as it is shown below:
```
learnitall: I have a suggestion for rewording this a little to pad it out. I think it also makes sense to make it more clear that when we omit or wildcard these rules, we are explicitly permitting all traffic that matches the ports in those rules.

```suggestion
To enable visibility for L7 traffic, create a ``CiliumNetworkPolicy`` that specifies
L7 rules. Traffic matching a L7 rule in a ``CiliumNetworkPolicy`` will become
visible. The following example enables visibility for DNS (UDP/53) and HTTP
(ports TCP/80 and TCP/8080) traffic within the ``default`` namespace by
specifying two L7 rules -- one for DNS and one for HTTP. L7 matching
conditions on the rules have been omitted or wildcarded, which will
permit all requests that match the L4 section of each rule:
```

Documentation/observability/visibility.rst
```diff
@@ -12,63 +12,74 @@ Layer 7 Protocol Visibility
...
+        name: "l7-visibility"
+      spec:
+        endpointSelector:
+          matchLabels:
+            "k8s:io.kubernetes.pod.namespace": default
+        egress:
+        - toPorts:
+          - ports:
+            - port: "53"
+              protocol: ANY
```
learnitall, line 57: Does this need to be UDP instead of ANY?

```suggestion
              protocol: UDP
```

Documentation/observability/visibility.rst
```diff
@@ -12,63 +12,74 @@ Layer 7 Protocol Visibility
...
+              protocol: TCP
+            - port: "8080"
+              protocol: TCP
+            rules:
+              http: [{}]
+        - toEntities:
+          - all
+
+The final rule allowing all egress traffic (``toEntities: [ all ]``) is
+optional and can be removed or replaced with a more specific set of rules.
```
learnitall: Can we remove this part? I don't think it fits here since we are just talking about L7 visibility, rather than constructing a policy in general.

Documentation/observability/visibility.rst
learnitall: Ah I see what you're saying, that's a good point. Let's keep it in.

learnitall (changes requested):
Changes look great, thanks! I have a couple of suggestions for you, and then we'll be good to go.

Documentation/observability/visibility.rst
networkop: I've modified the description to reflect UDP and TCP protocols since DNS proxy listens on both,

learnitall (approved):
LGTM! Thank you!

joestringer:
@networkop this fell out of date, please rebase. @cilium/proxy please review

sayboras:
> cilium/proxy please review

:+1:, my review can be counted from proxy team. I had a chat with Michael offline before as well.

sayboras (approved):

Documentation/observability/visibility.rst
```diff
@@ -12,63 +12,72 @@ Layer 7 Protocol Visibility
...
+.. note::
 
-Cilium will pick up that pods have received these annotations, and will
-transparently redirect traffic to the proxy such that the output of
-``cilium monitor`` shows traffic being redirected to the proxy, e.g.:
+    Historically, it had been possible to enable L7 visibility using Pod
+    annotations (``policy.cilium.io/proxy-visibility``). This method is
+    no longer supported and we recommend users to switch to L7 policies instead.
+
+To enable visibility for L7 traffic, create a ``CiliumNetworkPolicy`` that specifies
```
joestringer, line 28: I think it's critical to make it crystal clear to users that this is not just a visibility policy, this is allowing traffic matching the statements below to flow. That was previously the difference between visibility policy vs this; visibility did not change the actual policy behaviour of network flows.

networkop: this is why I had `toEntities: [ all ]`  at the end of the policy in the beginning -- to make the example behave more like a visibility policy. @joestringer do you think it's worth adding that catch-all rule back or just stress the point that this is a network policy that will drop traffic in the text?
see [this comment](https://github.com/cilium/cilium/pull/28449#discussion_r1353069344) for more context

joestringer: I think my feedback is less about the policy and more about having a dedicated paragraph  of test to frame the solution. My thought is that when users scan through the doc, they should see that the position of the project is "we don't have a dedicated visibility-only solution, the way to solve this problem is with enforcement network policy and hence you need to do this with awareness of the security impacts".

When I scanned through, I saw "if you want visibility, do this:" without a clear statement that describes that this has security implications.

networkop: yep, makes sense. I've added a mention of this to a couple of places in the text

joestringer: LGTM thanks

dylandreimerink:
@networkop Given its docs we don't need to run the whole test suite, is this "ready-to-merge"?

joestringer:
@dylandreimerink CI should detect that it's only docs and skip the other tests. This way we can just run CI on every PR, and not debate about ready-to-merge from CI perspective. If this is triggering full CI then we can iterate on the detection, I would consider that a CI bug. I'm trying to get us to a point where we don't try to make judgement calls about whether to skip CI or not, because that often becomes a point of failure in the process where we let buggy logic into the tree.

PR #28452, opened 2023-10-06
fix: preserve ownerreferences when updating ingresses with a shared load balancer

bittermandel:
When the IngressController reconciles a Ingress object which is configured to use a Shared LoadBalancer, each field is copied over from a DeepCopy of the original Ingress and the Status.LoadBalancer field is set to the values of the Shared LoadBalancer.

During this process, the OwnerReferences field is discarded. This causes issues when using other operators, such as Cert Manager, which relies on OwnerReferences during its reconcile process.

This fix makes sure the update process copies over the OwnerReferences field as well. 

PS: I remade the PR into a fresh one as I wasn't sure how to save the mess that was #28379.

```release-note
fix: Preserve OwnerReferences when updating Ingresses with Load Balancer in shared mode
```


bittermandel:
After playing around with this, this solution is not enough to resolve the issue.
Here's my findings:

1. This only happens when using a Shared load balancer. updateIngressStatus only runs on shared ones: https://github.com/bittermandel/cilium/blob/4bba5d3a5ded22b55b11014206e21f95cd3ae5db/operator/pkg/ingress/ingress.go#L446

2. The reason why OwnerReference was not copied over is due to using the `slim` version of Ingress. That versions Meta field does not include the `uuid` field nor `apiVersion`.  This is the response from `networkingv1.Ingress.UpdateStatus`: 
```│ level=warning msg="Failed to update ingress status" error="Ingress.extensions \"cm-acme-http-solver-n2ck9\" is invalid: [metadata.ownerReferences.apiVersion: Invalid value: \"\": version must not be empty, metadata. │
│ ownerReferences.uid: Invalid value: \"\": uid must not be empty]" subsys=ingress-controller```
Here you can find the slim version of the TypeMeta type: https://github.com/bittermandel/cilium/blob/4bba5d3a5ded22b55b11014206e21f95cd3ae5db/pkg/k8s/slim/k8s/apis/meta/v1/types.go#L207

Adding uuid and apiVersion fields to the slim TypeMeta would certainly fix that issue, but I fear that it would have implications outside of this particular PR.

I am not sure how to progress from here. Is there any history associated with removing/not including the UUID and apiVersion field?

**Comment by bittermandel** [comment]
With the changes from the past commit, I was able to get Cert Manager working with a shared cilium LB without any issues 🎉

**Review by meyskens — APPROVED** [review]
For the ingress changes LGTM. For the changes to the slim-API i think they should be fine however would love a second opinion from @tommyp1ckles to be sure

**Comment by bittermandel** [comment]
Is there anything I can help with to get this PR through? 🙂 We currently have this patched into our clusters as we're not able to utilize Cert Manager without it.

**Review by squeed — APPROVED** [review]

**Comment by squeed** [comment]
@bittermandel looks good. Would you mind adding a quick release note in the PR description? Thanks!

File: pkg/k8s/slim/k8s/apis/meta/v1/types.go [hunk 1]
```diff
@@ -205,12 +205,17 @@ const (
...
+	APIVersion string `json:"apiVersion" protobuf:"bytes,5,opt,name=apiVersion"`
 	// Kind of the referent.
 	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
 	Kind string `json:"kind" protobuf:"bytes,1,opt,name=kind"`
 	// Name of the referent.
 	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names#names
 	Name string `json:"name" protobuf:"bytes,3,opt,name=name"`
+	// UID of the referent.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names#uids
+	UID types.UID `json:"uid" protobuf:"bytes,4,opt,name=uid,casttype=k8s.io/apimachinery/pkg/types.UID"`
```
danehans, line 218: Changes are consistent with k8s v1 [OwnerReference](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#OwnerReference).

danehans (approved):

pkg/k8s/slim/k8s/apis/meta/v1/types.go
```diff
@@ -205,12 +205,17 @@ const (
 // be cluster-scoped, so there is no namespace field.
 // +structType=atomic
 type OwnerReference struct {
+	// API version of the referent.
```
tommyp1ckles, line 208: referent -> reference?

pkg/k8s/slim/k8s/apis/meta/v1/types.go
```diff
@@ -205,12 +205,17 @@ const (
...
+	APIVersion string `json:"apiVersion" protobuf:"bytes,5,opt,name=apiVersion"`
 	// Kind of the referent.
 	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
 	Kind string `json:"kind" protobuf:"bytes,1,opt,name=kind"`
 	// Name of the referent.
 	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names#names
 	Name string `json:"name" protobuf:"bytes,3,opt,name=name"`
+	// UID of the referent.
```
tommyp1ckles, line 216: referent -> reference?

pkg/k8s/slim/k8s/apis/meta/v1/types.go
bittermandel: These are copied from the official docs: https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#OwnerReference

pkg/k8s/slim/k8s/apis/meta/v1/types.go
danehans: Referent definition:
>the thing that a word or phrase denotes or stands for.
“the Morning Star” and “the Evening Star” have the same referent (the planet Venus)"

The use of `referent` is also consistent with the k8s v1 [OwnerReference](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#OwnerReference).

tommyp1ckles (approved):

PR #21895, opened 2022-10-26
fqdn/dnsproxy: Rewrite dnsproxy benchmark

odinuge:
This rewrites the Benchmark_perEPAllow_setPortRulesForID benchmark to test a "real world" scenario in terms of allocations, cpu usage and heap memory. Looong regexes take a lot of space, and the previous version of this benchmark wasn't too representative.

In general this test will now simulate an agent that has 60 running pods at any given time, and they keep recycling until we have created pod number 10000 (nEPs). They all have distinct rules, with some exceptions. Every everyNIsEqual EP has the same dns rules, and every everyNHasWildcard has a wildcard rule.

This is very useful since we don't need to rely on some CNP dump, and it behaves like a real world agent with a "limited" number of pods running at any given time. It would also help benchmark new changes, since the pprof dump now only contain the relevant data, and not ~mostly yaml+json parsing like the Benchmark_perEPAllow_setPortRulesForID_large. Also, previously it looked like this benchmark had two distinct selectors, but since they were null, in the end it only had one.

The new "B(HeapInUse)/op" value is a rough calculation of how much the heap size grows from the start of the benchmark to the end. This is a very good proxy for how much memory the dnsproxy will consume during runtime.
```
$ go test -v -bench='^Benchmark_perEPAllow_setPortRulesForID$' -run="^$" -benchtime 1x -memprofile memprofile.prof ./pkg/fqdn/dnsproxy

cache size of 1024:
Benchmark_perEPAllow_setPortRulesForID-7               1        91863112900 ns/op       1786241024 B(HeapInUse)/op      78642436208 B/op        167454992 allocs/op

cache size of 128:
Benchmark_perEPAllow_setPortRulesForID-7               1        88338408700 ns/op        224722944 B(HeapInUse)/op      78642493144 B/op        167456155 allocs/op

cache size of 1:
Benchmark_perEPAllow_setPortRulesForID-7               1        79318500100 ns/op        105177088 B(HeapInUse)/op      86872021272 B/op        180312236 allocs/op

cache size of 0 (disabled):
Benchmark_perEPAllow_setPortRulesForID-7               1        144261660700 ns/op       209502208 B(HeapInUse)/op      169280546296 B/op       309321759 allocs/op

```

Signed-off-by: Odin Ugedal <ougedal@palantir.com>

Sampl pprof snapshot;
Screen Shot 2022-10-26 at 8 54 03 PM

Please ensure your pull request adheres to the following guidelines:

[...]

PR #28465, opened 2023-10-09
labels/cidr: Memoize labels for already seen prefixes

pippolo84:
GetCIDRLabels turns a CIDR into a set of labels representing the cidr itself and all broader CIDRS which include the specified CIDR in them.

In case of ToFQDNs policies matching many names that maps to many IPs with very short TTLs, this function can easily become a hot path.

To improve performance, the PR adds a LRU cache to memoize intermediate results.

See individual commit messages for further details.

### DIFF
--- a/pkg/labels/cidr/cidr.go
+++ b/pkg/labels/cidr/cidr.go
@@ -10,7 +10,10 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/hashicorp/golang-lru/v2/simplelru"
+
 	"github.com/cilium/cilium/pkg/labels"
+	"github.com/cilium/cilium/pkg/lock"
 	"github.com/cilium/cilium/pkg/option"
 )
 
@@ -82,6 +85,11 @@ func IPStringToLabel(ip string) (labels.Label, error) {
 //
 // The identity reserved:world is always added as it includes any CIDR.
 func GetCIDRLabels(prefix netip.Prefix) labels.Labels {
+	once.Do(func() {
+		// simplelru.NewLRU fails only when given a negative size, so we can skip the error check
+		cidrLabelsCache, _ = simplelru.NewLRU[netip.Prefix, []labels.Label](cidrLabelsCacheMaxSize, nil)
+	})
+
 	addr := prefix.Addr()
 	ones := prefix.Bits()
 	lbls := make(labels.Labels, 1 /* this CIDR */ +ones /* the prefixes */ +1 /*world label*/)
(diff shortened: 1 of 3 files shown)


nathanjsweet (approved):

pippolo84:
Integration Tests hit https://github.com/cilium/cilium/issues/27080

go.mod
```diff
@@ -34,6 +34,7 @@ require (
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/docker/docker v24.0.6+incompatible
 	github.com/docker/libnetwork v0.8.0-dev.2.0.20210525090646-64b7a4574d14
+	github.com/dustin/go-humanize v1.0.1
```
rolinh: Is it worth adding a new dependency merely for two log statements in a test?

pippolo84: Changed it with a handmade conversion function :+1:

pkg/labels/cidr/cidr.go
```diff
@@ -18,7 +22,7 @@ import (
 //
 // For IPv6 addresses, it converts ":" into "-" as EndpointSelectors don't
 // support colons inside the name section of a label.
-func maskedIPToLabelString(ip netip.Addr, prefix int) string {
+func maskedIPToLabel(ip netip.Addr, prefix int) labels.Label {
 	ipStr := ip.String()
 	ipNoColons := strings.Replace(ipStr, ":", "-", -1)
```
squeed, line 27: I wonder if we can skip another allocation here by doing the character replacement in-place? (Strings are immutable but `[]byte` isn't... I wonder if the compiler is clever enough here).

nebril (approved):

rolinh (approved):

pkg/labels/cidr/cidr.go
pippolo84: Good point! I think I found a way to avoid the allocation. I'm gonna open a follow up PR for that.

pippolo84: https://github.com/cilium/cilium/pull/28647

odinuge:
Hi!

It does seem like this introduces some strange behavior. My colleague @sjdot discovered that this change results in excessive numbers of false positive policy drops, for legit traffic that should not be dropped.


By adding this test

```diff
diff --git a/pkg/labels/cidr/cidr_test.go b/pkg/labels/cidr/cidr_test.go
index 8ee0dbd605..9a6dd722c4 100644
--- a/pkg/labels/cidr/cidr_test.go
+++ b/pkg/labels/cidr/cidr_test.go
@@ -4,6 +4,7 @@
 package cidr

 import (
+       "fmt"
        "net/netip"
        "testing"

@@ -309,6 +310,18 @@ func BenchmarkLabels_SortedListCIDRIDs(b *testing.B) {
        }
 }

+func TestThisThing(t *testing.T) {
+       for _, ip := range []string{
+               "10.0.0.0/1",
+               "10.0.0.0/3",
+       } {
+               t.Run(ip, func(b *testing.T) {
+                       lbls := GetCIDRLabels(netip.MustParsePrefix(ip))
+                       fmt.Printf("%q: %+v\n", netip.MustParsePrefix(ip), lbls.LabelArray())
+               })
+       }
+}
+
 func BenchmarkIPStringToLabel(b *testing.B) {
        for _, ip := range []string{
                "0.0.0.0/0",
  ```
  , we get the following output;
  
  ```
  === RUN   TestThisThing
=== RUN   TestThisThing/10.0.0.0/1
"10.0.0.0/1": [cidr:0.0.0.0/0 cidr:0.0.0.0/1 reserved:world-ipv4]
=== RUN   TestThisThing/10.0.0.0/3
"10.0.0.0/3": [cidr:0.0.0.0/0 cidr:0.0.0.0/1 reserved:world-ipv4]
--- PASS: TestThisThing (0.00s)
    --- PASS: TestThisThing/10.0.0.0/1 (0.00s)
    --- PASS: TestThisThing/10.0.0.0/3 (0.00s)
PASS
```

When disabling the cache, we instead get this that make more sense.

```
=== RUN   TestThisThing
=== RUN   TestThisThing/10.0.0.0/1
"10.0.0.0/1": [cidr:0.0.0.0/0 cidr:0.0.0.0/1 reserved:world-ipv4]
=== RUN   TestThisThing/10.0.0.0/3
"10.0.0.0/3": [cidr:0.0.0.0/0 cidr:0.0.0.0/1 cidr:0.0.0.0/2 cidr:0.0.0.0/3 reserved:world-ipv4]
--- PASS: TestThisThing (0.00s)
    --- PASS: TestThisThing/10.0.0.0/1 (0.00s)
    --- PASS: TestThisThing/10.0.0.0/3 (0.00s)
PASS
```

So it overall seems like there is some issue in the way things are cached here, or is this expected..? @pippolo84 @joamaki

pippolo84:
Hi @odinuge , thank you for the bringing this up. :pray: 
I confirm this was not expected. We are working on a fix, together with better unit testing.

pkg/labels/cidr/cidr_test.go
```diff
@@ -309,6 +320,124 @@ func BenchmarkLabels_SortedListCIDRIDs(b *testing.B) {
...
+		}
+	}
+	b.StopTimer()
+
+	runtime.GC()
+	runtime.GC()
+	runtime.ReadMemStats(&m2)
+
+	usage := m2.HeapAlloc - m1.HeapAlloc
+	b.Logf("Memoization map heap usage: %.2f KiB", float64(usage)/1024)
```
odinuge, line 396: unless you were aware, you can also plumb this into the benchmark result to make it easier to use things like benchstat to compare it; https://github.com/cilium/cilium/pull/21895/files#diff-a5084c37ba77e5fe28cbc8d15ffe1e5c312e2daf79c8a363ec52df758818d26dR1091

pippolo84: Thanks for the heads up, but I think that in this case it is better to switch to a simple test instead of a benchmark. That's because we want to measure the memory usage when the cache is full and to do that with the benchmark we would need to either:

- pass `benchtime=1x` to complete only a single pass
- reset the cache at each pass as part of the code to benchmark (but this resulted in longer benchmark run, probably for the time needed to stabilize the time measure)

PR #28473, opened 2023-10-09
v1.14: avoid relying on golang.org/exp/slices.SortFunc

rolinh:
The signature of `slices.SortFunc` has changed recently, breaking builds for projects that vendor Cilium 1.14 and a newer version of `golang.org/x/exp` with the signature change.

Due to the way dependencies work, it's unfortunately not always possible to downgrade `golang.org/x/exp` to a previous version.

The patch is based on the patch that has been applied to the main branch to allow upgrading `golang.org/x/exp` but without actually bumping the dependency.

ref: 036f1dc0eeefc407492c8230831a90156ecb9b6b

### DIFF
--- a/pkg/hive/cell/health.go
+++ b/pkg/hive/cell/health.go
@@ -6,13 +6,13 @@ package cell
 import (
 	"context"
 	"fmt"
+	"sort"
 	"sync/atomic"
 	"time"
 
 	"github.com/cilium/cilium/pkg/lock"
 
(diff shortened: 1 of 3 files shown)


aanm (approved):

christarazi (approved):

qmonnet:
/test-backport-1.14

squeed (approved):

PR #27888, opened 2023-09-01
Merge clustermesh-apiserver and kvstoremesh into a single image

giorio94:
Reporting here the message from the main commit for convenience:

```
Merge the kvstoremesh logic into the clustermesh-apiserver as a separate
subcommand, so that we can get rid of one image and reduce the overall
time to bootstrap the clustermesh-apiserver deployment. The additional
boilerplate to build the kvstoremesh images is dropped.

Size of the images before this change:
$ make docker-clustermesh-apiserver-image
$ docker inspect -f '{{ .Size }}' quay.io/cilium/clustermesh-apiserver:latest | \
         numfmt --to=si --suffix=B --format="%.2f"
---> 72.09MB
$ make docker-kvstoremesh-image
$ docker inspect -f '{{ .Size }}' quay.io/cilium/kvstoremesh:latest | \
         numfmt --to=si --suffix=B --format="%.2f"
---> 64.62MB

Size of the image after this change:
$ make docker-clustermesh-apiserver-image
$ docker inspect -f '{{ .Size }}' quay.io/cilium/clustermesh-apiserver:latest | \
         numfmt --to=si --suffix=B --format="%.2f"
---> 72.26MB
```

~~Depends on https://github.com/cilium/cilium/pull/27823, https://github.com/cilium/cilium/pull/27817, https://github.com/cilium/cilium/pull/27873~~

To be discussed in the next community meeting: should we rename the clustermesh-apiserver pod/image to something more expressive which better highlights the new structure (the container is planned to be extended to include the etcd binary, as well as the current init-container logic)? This would require a fairly large amount of CI changes (some of which will likely only pop-up during the next release), as well as make backporting of CI changes more tricky in the future.  

```release-note
Merge clustermesh-apiserver and kvstoremesh into a single image
```

PR #28478, opened 2023-10-10
clustermesh-apiserver: extract external workloads in a separate cell

giorio94:
Refactor the initialization of the external workloads logic into a separate cell. A few of the existing parameters are dropped, as they were never explicitly configured, and more in detail:

* --identity-allocation-mode: the clustermesh-apiserver is expected to be used only when Cilium is configured in CRD mode, while kvstore mode is not supported; the same also applies to the current implementation of external workloads;
* --allocator-list-timeout: it has been hard-coded to the default value for the sake of simplicity;
* --enable-well-known-identities: not applicable to the external workloads from the clustermesh-apiserver perspective, and thus not necessary.

For https://github.com/cilium/cilium/pull/27888.

### DIFF
--- a/clustermesh-apiserver/vmmanager.go
+++ b/clustermesh-apiserver/vmmanager.go
@@ -6,21 +6,22 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net"
 	"path"
 	"sort"
 
-	k8sv1 "k8s.io/api/core/v1"
+	"github.com/spf13/pflag"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/client-go/tools/cache"
 
 	"github.com/cilium/cilium/api/v1/models"
 	"github.com/cilium/cilium/pkg/cidr"
 	cmtypes "github.com/cilium/cilium/pkg/clustermesh/types"
+	"github.com/cilium/cilium/pkg/defaults"
+	"github.com/cilium/cilium/pkg/hive"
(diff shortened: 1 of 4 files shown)


thorn3r (approved):
excellent work

clustermesh-apiserver/vmmanager.go
```diff
@@ -29,74 +30,107 @@ import (
...
+) *VMManager {
+	if !cfg.EnableExternalWorkloads {
+		return nil
 	}
-	m.identityAllocator = identityCache.NewCachingIdentityAllocator(m)
 
-	if option.Config.EnableWellKnownIdentities {
-		identity.InitWellKnownIdentities(option.Config, cinfo)
+	// External workloads require CRD allocation mode
+	option.Config.IdentityAllocationMode = option.IdentityAllocationModeCRD
```
tommyp1ckles, line 81: Does this draw from the regular cilium-config configmap, if so should we terminate if we're using an external kvstore in case Cilium is using that (instead of silently reconfiguring)?

tommyp1ckles (approved):
k8s changes look good, just one question..

joestringer:
(Awaiting unresolved conversation before merging)

clustermesh-apiserver/vmmanager.go
giorio94: The clustermesh-apiserver pod doesn't mount the cilium-config configmap, which means that the IdentityAllocationMode was never set to a value different from the default one also in the current version. From this point of view, no regression is introduced by this PR, which just removes that (unused) parameter and hard-codes the default value for the sake of simplicity.

More in general, the entire clustermesh-apiserver (not only the external workloads support) is not compatible with Cilium running in kvstore mode, and I agree that we should error out in this case to make the incompatibility more explicit (e.g., introducing a validation at the helm level). I'd postpone that as a follow-up item though, being a separate QoL enhancement unrelated from this PR.

giorio94: Resolving the conversation so that this PR can be merged.

tommyp1ckles:
@joestringer my question is resolved

PR #28480, opened 2023-10-10
clustermesh-apiserver/kvstoremesh: unify metrics cell

giorio94:
Reuse the same metrics cell for both the clustermesh-apiserver and kvstoremesh, to reduce code duplication and prepare to merge the two components into a single binary.

For https://github.com/cilium/cilium/pull/27888.


thorn3r (approved):
looks great!

PR #28484, opened 2023-10-10
bgpv1,ci: Add Test_AdvertisedPathAttributes into BGP component tests

rastislavs:
Adds a new BGP component test which tests the `AdvertisedPathAttributes` options of the `CiliumBGPPeeringPolicy` CRD. As iBGP peering is needed to test Local Preference, it modifies the test framework to allow it.

### DIFF
--- a/pkg/bgpv1/test/adverts_test.go
+++ b/pkg/bgpv1/test/adverts_test.go
@@ -5,6 +5,9 @@ package test
 
 import (
 	"context"
+	"reflect"
+	"strconv"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/bgpv1/test/adverts_test.go
```diff
@@ -622,3 +627,188 @@ func Test_LBEgressAdvertisement(t *testing.T) {
...
+					}),
+				},
+			},
+		},
+	}
+
+	testCtx, testDone := context.WithTimeout(context.Background(), maxTestDuration)
+	defer testDone()
+
+	// setup topology - iBGP (ASN == ciliumASN)
```
harsimran-pabla, line 712: Why use iBGP for this test? Are some of these communities non-transitive between AS?

rastislavs: To test the Local Preference Attribute. That is not advertised to eBGP peers.

harsimran-pabla (approved):

ldelossa (approved):

PR #28489, opened 2023-10-10
bgpv1,ci: Do not use asserts in Eventually() test conditions

rastislavs:
As `assert.Equal` / `assert.ElementsMatch` fail the test when evaluated as false (even though not immediately, but after the test finishes), we should replace these conditions with non-assert comparison logic in the conditions used for `require.Eventually()` in the BGP component tests.

### DIFF
--- a/pkg/bgpv1/test/neighbor_test.go
+++ b/pkg/bgpv1/test/neighbor_test.go
@@ -5,6 +5,8 @@ package test
 
 import (
 	"context"
+	"reflect"
+	"sort"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/bgpv1/test/neighbor_test.go
```diff
@@ -173,7 +174,17 @@ func Test_NeighborAddDel(t *testing.T) {
 						holdTimeSeconds: peer.AppliedHoldTimeSeconds,
 					})
 				}
-				return assert.ElementsMatch(t, step.expectedPeerStates, runningState, step.description)
+				sort.Slice(step.expectedPeerStates, func(i, j int) bool {
```
harsimran-pabla: Consider moving this check to separate method, so this check can be reused by other tests.

IsPeerStateEqual(expected, running) bool

rastislavs: good idea, done

YutaroHayakawa (approved):

Issue #6839, opened 2019-01-29
operator restarts kube-dns repeatedly


tgraf:
When the pod doesn't restart quick enough, multiple restarts are issued:

```
level=info msg="Cilium Operator " subsys=cilium-operator
level=info msg="Connecting to etcd server..." config=/var/lib/etcd-config/etcd.config endpoints="[https://cilium-etcd-client.cilium.svc:2379]" subsys=kvstore
level=info msg="Waiting for k8s api-server to be ready..." subsys=k8s
level=info msg="Connected to k8s api-server" ipAddr="https://10.47.240.1:443" subsys=k8s
level=info msg="Restarting unmanaged kube-dns pod kube-system/kube-dns-7549f99fcc-4sk78 started 8m26.731288672s ago" subsys=cilium-operator
level=info msg="Restarting unmanaged kube-dns pod kube-system/kube-dns-7549f99fcc-6mprc started 8m18.737163036s ago" subsys=cilium-operator
level=info msg="Restarting unmanaged kube-dns pod kube-system/kube-dns-7549f99fcc-4sk78 started 8m41.755593612s ago" subsys=cilium-operator
level=info msg="Restarting unmanaged kube-dns pod kube-system/kube-dns-7549f99fcc-6mprc started 8m33.763022725s ago" subsys=cilium-operator
level=info msg="Restarting unmanaged kube-dns pod kube-system/kube-dns-7549f99fcc-4sk78 started 8m56.781309619s ago" subsys=cilium-operator
level=info msg="Restarting unmanaged kube-dns pod kube-system/kube-dns-7549f99fcc-6mprc started 8m48.788871881s ago" subsys=cilium-operator
```

tjamet:
Hi,
I am unsure I am suffering the same problem but I am observing that my kube-dns pod is regularly restarted:
```
level=info msg="Restarting unmanaged kube-dns pod kube-system/coredns-69cbb76ff8-pvxjj started 44.755530054s ago" subsys=cilium-operator
level=debug msg="Found unmanaged kube-dns pod kube-system/coredns-69cbb76ff8-thfp4" subsys=cilium-operator
level=info msg="Restarting unmanaged kube-dns pod kube-system/coredns-69cbb76ff8-thfp4 started 44.783591045s ago" subsys=cilium-operator
level=debug msg="Found unmanaged kube-dns pod kube-system/coredns-69cbb76ff8-9nn8m" subsys=cilium-operator
[...]

PR #28507, opened 2023-10-10
operator: remove lastPodRestart in unmanaged pod controller

wedaly:
operator: remove lastPodRestart in unmanaged pod controller

The restart-unmanaged-pod controller logic was:

1. Find unmanaged pod to delete.
2. Lookup the last restart time by the pod's ID (namespace + name).
3. If last restart time is less than 5 minutes ago, skip.
4. Otherwise, delete the pod.
5. If the deletion was successful, set lastPodRestart[podID] = time.Now()

However, step 2 would never find pod ID in the lastRestartTime map, because podIDs are only added to the map after a successful deletion (step 5). Any new pods created after the deletion would be assigned a unique, different, pod ID. Therefore, step 3 would never be executed.

Since the controller behaves the same without checking lastPodRestart, and I don't believe there is a way to reliably identify which pod replaced the deleted pod, remove the lastPodRestart map.

### DIFF
--- a/operator/cmd/k8s_pod_controller.go
+++ b/operator/cmd/k8s_pod_controller.go
@@ -21,13 +21,10 @@ import (
 )
 
 const (
-	minimalPodRestartInterval = 5 * time.Minute
-	unmanagedPodMinimalAge    = 30 * time.Second
+	unmanagedPodMinimalAge = 30 * time.Second
 )
 
 var (
-	lastPodRestart = map[string]time.Time{}
-
 	restartUnmanagedPodsControllerGroup = controller.NewGroup("restart-unmanaged-pods")
 )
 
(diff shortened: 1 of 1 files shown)


joamaki:
Why would the `podID` be different after restart? It's the `namespace/name` after all and not e.g. the `types.ID`?

squeed:
There is one case in which pod names are constant, even across deletion: StatefulSets. I agree, though, the logic is convoluted to say the least.

joamaki:
Ok that makes sense. Has the naming behavior changed at some point in Kubernetes? Looking at the original issue (https://github.com/cilium/cilium/issues/6839) it shows that it was added to fix multiple restarts of kube-dns pods and there the name didn't change. I would be hesitant to removing this code until we're sure we don't just reintroduce #6839 back again.

wedaly:
> Ok that makes sense. Has the naming behavior changed at some point in Kubernetes? Looking at the original issue (#6839) it shows that it was added to fix multiple restarts of kube-dns pods and there the name didn't change. I would be hesitant to removing this code until we're sure we don't just reintroduce #6839 back again.

hmm, looking at that issue I'm now less confident in this change

Reading through https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination I suspect cilium-operator might end up sending multiple delete requests between when apiserver sets `DeletionTimestamp` on the pod and when kubelet finishes cleaning it up.

Given that the current code isn't causing any problems and I'm not 100% sure removing it is safe, I'll close this PR. Thanks for the review and comments!

PR #28512, opened 2023-10-11
hive: ModuleID and FullModuleID, use full ID in module health

joamaki:
First patch adds ModuleID and FullModuleID types:

```
    
      cell.Module(
        "outer",
        "Outer",
    
        cell.Module(
          "inner",
          "Inner",
    
          cell.Invoke(id cell.ModuleID, fullID cell.FullModuleID) {
            // id == "inner"
            // fullID == "outer.inner"
          },
        ),
      )
```

Second switches health reporter to use full module id:

```
      Modules Health:
        Module                            Status   Message                      Last Updated
        agent.controlplane.node-manager   OK       Node validation successful             4s
```

Final patch removes the unnecessary module exclusion from health REST API handler and
instead drops the unused HealthReporter dependency from daemon.

### DIFF
--- a/daemon/cmd/vitals.go
+++ b/daemon/cmd/vitals.go
@@ -17,7 +17,7 @@ import (
 // moduleExcludes tracks modules that should be excluded from the health report.
 var moduleExcludes = map[string]struct{}{
 	// daemon currently does not rollup status and hence reports default health aka unknown.
-	"daemon": {},
+	"agent.daemon": {},
 }
 
 type getHealth struct {
@@ -39,7 +39,7 @@ func (d *Daemon) getHealthReport() models.ModulesHealth {
 	mm := d.healthProvider.All()
 	rr := make([]*models.ModuleHealth, 0, len(mm))
 	for _, m := range mm {
-		if _, ok := moduleExcludes[m.ModuleID]; ok {
+		if _, ok := moduleExcludes[m.FullModuleID.String()]; ok {
 			continue
 		}
 		rr = append(rr, toModuleHealth(m))
@@ -52,7 +52,7 @@ func (d *Daemon) getHealthReport() models.ModulesHealth {
 
 func toModuleHealth(m cell.Status) *models.ModuleHealth {
 	return &models.ModuleHealth{
-		ModuleID:    m.ModuleID,
+		ModuleID:    m.FullModuleID.String(),
 		Message:     m.Message,
 		Level:       string(m.Level),
 		LastOk:      toAgeHuman(m.LastOK),
--- a/daemon/cmd/vitals.go
+++ b/daemon/cmd/vitals.go
@@ -14,12 +14,6 @@ import (
 	"github.com/cilium/cilium/pkg/hive/cell"
 )
 
-// moduleExcludes tracks modules that should be excluded from the health report.
-var moduleExcludes = map[string]struct{}{
-	// daemon currently does not rollup status and hence reports default health aka unknown.
-	"agent.daemon": {},
-}
-
 type getHealth struct {
 	daemon *Daemon
 }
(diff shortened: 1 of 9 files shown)


pkg/hive/cell/module.go
```diff
@@ -26,6 +26,13 @@ func Module(id, title string, cells ...Cell) Cell {
 	return &module{id, title, cells}
 }
 
+// ModuleID is the module identifier. Provided in the module's scoped.
```
rgo3: Nit:
```suggestion
// ModuleID is the module identifier. Provided in the module's scope.
```

rgo3 (approved):

daemon/cmd/vitals.go
```diff
@@ -52,7 +43,7 @@ func (d *Daemon) getHealthReport() models.ModulesHealth {
 
 func toModuleHealth(m cell.Status) *models.ModuleHealth {
 	return &models.ModuleHealth{
-		ModuleID:    m.ModuleID,
+		ModuleID:    string(m.ModuleID),
```
derailed: nit: This is going to be common usage and since we now have a type why not make it a Stringer?

pkg/hive/cell/module.go
```diff
@@ -26,6 +26,13 @@ func Module(id, title string, cells ...Cell) Cell {
...
+type ModuleID string
+
+// FullModuleID is the fully qualified module identifier, e.g. the
+// concat of nested module ids, e.g. "agent.controlplane.endpoint-manager".
```
derailed, line 35: Wondering if using '/' vs "." could help us here as we could leverage filepath conveniences to assert root module, subpath or expressions to either derive health or others or issue db queries...

pkg/hive/cell/module.go
```diff
@@ -56,16 +63,31 @@ func (m *module) logger(log logrus.FieldLogger) logrus.FieldLogger {
...
 }
 
-func (m *module) moduleScopedStatusReporter(p Health) HealthReporter {
-	return p.forModule(m.id)
+func (m *module) moduleID() ModuleID {
+	return ModuleID(m.id)
+}
+
+func (m *module) fullModuleID(parent FullModuleID) FullModuleID {
+	if parent == "" {
```
derailed: We should hang ctors on the type and let the type encapsulate how this is represented internally vs exposing the semantic at the call site?

pkg/hive/hive.go
```diff
@@ -146,21 +146,23 @@ func (h *Hive) Viper() *viper.Viper {
...
-			Lifecycle:   h.lifecycle,
-			Logger:      log,
-			Shutdowner:  h,
-			InvokerList: h,
+			Flags:             h.flags,
+			Lifecycle:         h.lifecycle,
+			Logger:            log,
+			Shutdowner:        h,
+			InvokerList:       h,
+			EmptyFullModuleID: cell.FullModuleID(""),
```
derailed: nit: Perhaps define a const BlankModuleID if deemed useful?

derailed (commented):
@joamaki Very nice cleanup and concept!! Just a few minor comments

daemon/cmd/vitals.go
joamaki: Hmm now that you mentioned this, I think I'll go back to what I had before with `type FullModuleID []string` and with a `String()`, and then use `String()` as map keys here.

pkg/hive/cell/module.go
joamaki: I don't want to use "/" as then it might make someone think that there's some correspondence to file paths or Go package names. Also "." is the classic module separator and looks nicer.

pkg/hive/cell/module.go
joamaki: Yep, will clean up.

pkg/hive/hive.go
joamaki: Changed `FullModuleID` to be `[]string` underneath, so now it's just `nil`.

derailed (approved):

tommyp1ckles (approved):

thorn3r (approved):

PR #24882, opened 2023-04-14
datapath: Remove 2005 route table for IPv6

jschwinger233:
This reverts 3ed62d51d5ac27bbe2c60372850d4ec78c17bb6e for IPv6 part only, as issue #21954 has been resolved by #24208.

Another PR #24807  removes 2005 route table for IPv4.

Signed-off-by: Zhichuan Liang <gray.isovalent.com>

```release-note
Fix broken IPv6 connectivity from outside to NodePort service when L7 ingress policy applied by removing PROXY_RT route table.
```

PR #24807, opened 2023-04-11
datapath: Remove 2005 route table for IPv4

jschwinger233:
This reverts 3ed62d51d5ac27bbe2c60372850d4ec78c17bb6e for IPv4 part only, as issue #21954 has been resolved by #24208.

Another PR #24882 removes 2005 route table for IPv6.

Signed-off-by: Zhichuan Liang <gray.isovalent.com>

```release-note
Fix broken IPv4 connectivity from outside to NodePort service when using L7 ingress policy, by removing PROXY_RT route table.
```

Issue #21954, opened 2022-11-01
Reply from pod to outside is dropped when L7 ingress policy is used


brb:
This issue is about an L7 ingress policy problem a when a pod is reached directly from outside client / via a NodePort BPF service.

Let's consider the following L7 netpol:
```
apiVersion: "cilium.io/v2"
kind: CiliumNetworkPolicy
metadata:
  name: foobar
spec:
  description: "Allow to GET on echo from outside"
  endpointSelector:
    matchLabels:
      kind: echo
  ingress:
  - fromEntities:
    - "world"
    toPorts:
    - ports:
      - port: "8080"
        protocol: TCP
      rules:
        http:
        - method: "GET"
          path: "/$"
```

When the netpol is applied, accessing the `echo` pod from outside the cluster fails with:
```
xx drop (Stale or unroutable IP) flow 0xf863e34e to endpoint 0, file bpf_host.c line 665, , identity 28143->unknown: 10.0.1.49:80 -> 192.168.34.1:32884 tcp SYN, ACK
```
The drop is triggered by https://github.com/cilium/cilium/blob/v1.12.3/bpf/bpf_host.c#L626.

What happens is that the L7 proxy sends the SYN-ACK which gets handled by `bpf_host @ cilium_host`, and then dropped. See the  `pwru` output (ifindex=9 is `cilium_host`):

```
  SKB    CPU          PROCESS                     FUNC
0xffff9a859048e800      7        [<empty>]             ip_local_out netns=4026531992 mark=0xa00 ifindex=0 proto=0 mtu=0 len=60 10.0.1.49:80->192.168.34.1:32884(tcp)
0xffff9a859048e800      7        [<empty>]           __ip_local_out netns=4026531992 mark=0xa00 ifindex=0 proto=0 mtu=0 len=60 10.0.1.49:80->192.168.34.1:32884(tcp)
0xffff9a859048e800      7        [<empty>]             nf_hook_slow netns=4026531992 mark=0xa00 ifindex=0 proto=8 mtu=0 len=60 10.0.1.49:80->192.168.34.1:32884(tcp)
0xffff9a859048e800      7        [<empty>]                ip_output netns=4026531992 mark=0xa00 ifindex=0 proto=8 mtu=0 len=60 10.0.1.49:80->192.168.34.1:32884(tcp)
[...]

PR #28537, opened 2023-10-11
Ensure IPsec packets go through the right path when applying L7 policies

jschwinger233:
This PR changes the packet path when IPsec is enabled and L7 policy is applied.

```release-note
```

### DIFF
--- a/bpf/bpf_host.c
+++ b/bpf/bpf_host.c
@@ -231,6 +231,8 @@ handle_ipv6_cont(struct __ctx_buff *ctx, __u32 secctx, const bool from_host,
 	int l3_off = ETH_HLEN;
 	struct remote_endpoint_info *info = NULL;
 	struct endpoint_info *ep;
+	__u8 __maybe_unused encrypt_key;
+	bool __maybe_unused from_proxy;
 	int ret;
 
 	if (!revalidate_data(ctx, &data, &data_end, &ip6))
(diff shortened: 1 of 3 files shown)


bpf/bpf_host.c
```diff
@@ -335,11 +337,20 @@ handle_ipv6_cont(struct __ctx_buff *ctx, __u32 secctx, const bool from_host,
 	dst = (union v6addr *) &ip6->daddr;
 	info = ipcache_lookup6(&IPCACHE_MAP, dst, V6_CACHE_KEY_LEN, 0);
 
+	encrypt_key = 0;
+#if defined(ENABLE_IPSEC)
+	//from_proxy = (ctx->mark & MARK_MAGIC_HOST_MASK) == MARK_MAGIC_PROXY_INGRESS;
+	from_proxy = true;
```
jschwinger233, line 343: We should encrypt from_host packets only if they are from proxy and marked with `0xa00`. But at this time, both skb->mark and skb->cb has been cleared, I couldn't find a way to check mark.

Possible solutions:
a. find another position where we can get mark (`magic` variable). 
b. find a way to pass mark or some variable to this position.

Options B seems to be easier, because at this position we can leverage existing functions and `info` from ipcache.

@pchaigno Do you have any idea?

jschwinger233:
Apart from the outstanding issue left above, this PR reverts #24807 and #24882, brings back https://github.com/cilium/cilium/issues/21954. I need to find a way to fix that resurged bug.

jschwinger233:
/test-backport-1.14

jschwinger233:
Cannot reproduce https://github.com/cilium/cilium/issues/21954. Looking closer, I found external traffic (with external src ip) was SNAT-ed before reaching proxy, so proxy didn't send back a reply to external IP, so expected packet drop at https://github.com/cilium/cilium/blob/v1.12.3/bpf/bpf_host.c#L626 didn't happen.

(Don't know if the bug was fixed by some other patches, or I just failed to reproduce it due to wrong steps.

PR #28547, opened 2023-10-11
makefile: fix 'fast' targets for cilium-dbg

aanm:
Fixes: 4deff3cd65c5 ("cilium: Rename local CLI to cilium-dbg")


joestringer (approved):
Sorry about that, thanks for the fix :)

PR #28292, opened 2023-09-26
contrib: Add ContainerLab-based BGP CPlane development environment

YutaroHayakawa:
Please see the commit message for more details.

```release-note
contrib: Add ContainerLab-based BGP CPlane development environment
```

PR #28548, opened 2023-10-11
Fix kind targets

chancez:
Fixes a few issues brought up in slack: https://cilium.slack.com/archives/C2B917YHE/p1697062058979519

- Fix `kind-ready` when dealing with clusters that are not named `kind`, fixing the `fast` variant of the kind targets.
- Do not assume a cluster is named `kind` in the non-fast kind targets, instead pick the first KIND cluster, unless overriden
- Fix kind fast targets for MacOS by setting `GOOS=linux`, since binaries need to be built for Linux to run within KIND (and agent only builds for GOOS=linux).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -563,7 +563,7 @@ kind-install-cilium-clustermesh: kind-clustermesh-ready ## Install a local Ciliu
 	$(CILIUM_CLI) clustermesh status --context kind-clustermesh1 --wait
 	$(CILIUM_CLI) clustermesh status --context kind-clustermesh2 --wait
 
-KIND_CLUSTER_NAME ?= kind
+KIND_CLUSTER_NAME ?= $(shell kind get clusters -q | head -n1)
 
 .PHONY: kind-ready
 kind-ready:
--- a/Makefile
+++ b/Makefile
@@ -621,19 +621,19 @@ kind-install-cilium-fast: kind-ready ## Install a local Cilium version into the 
 
 .PHONY: build-cli
 build-cli: ## Build cilium cli binary
-	$(QUIET)$(MAKE) -C cilium-dbg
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -622,19 +621,19 @@ kind-install-cilium-fast: kind-ready ## Install a local Cilium version into the
 
 .PHONY: build-cli
 build-cli: ## Build cilium cli binary
-	$(QUIET)$(MAKE) -C cilium-dbg
+	$(QUIET)$(MAKE) -C cilium-dbg GOOS=linux
```
joestringer, line 624: nit: I wonder whether it's worth doing a `GOOS ?= linux` definition somewhere like `Makefile.defs` so that this can be easily overridden as well. I have from time to time passed GOOS=darwin to try some things out, though I recognize that Cilium does not fully compile like that currently.

chancez: Yeah I considered it, I wasn't sure about if I should do that or not. 

A global variable definition would change every invocation of our builds, whereas this is scoped to just the targets being used by the `kind-image-*-fast` targets, and as far as I can tell, these `build-*` targets are only used by the `kind-image-*-fast` targets. 

For example, if we blindly set `GOOS=linux` then running the tests will fail for non-Linux users, since it'll build the test binary for Linux, which cannot be executed on non-Linux.

joestringer: Ah good point, hmm. Alright I can't think of a better solution for now. We can always iterate on this in future. :+1:

joestringer (approved):

YutaroHayakawa (approved):
I confirmed my original goal in #28292 is still accomplished. LGTM 👍

aanm (approved):
Thank you!

PR #28554, opened 2023-10-12
[DRAFT] Proof-of-concept gRPC client for StateDB

joamaki:
Explore how we could expose remote access to StateDB that mirrors it's normal usage. Useful for `cilium-dbg` to allow simplified implementation of state inspection without specialized REST API handlers.

```
         table, err := statedb.NewRemoteTable[tables.Backend]("backends", "localhost:8456")
         if err != nil {
                 panic(err)
         }

         // Query using an index
         iter, err := table.Get(context.TODO(), tables.BackendPortIndex.Query(1234))
         // ...
 
         // Watch all changes to the object ("kubectl get --follow")
         iter, err := table.Watch(context.TODO())
         if err != nil {
                 panic(err)
         }
         for obj, _, ok := iter.Next(); ok; obj, _, ok = iter.Next() {
                 fmt.Printf("%+v\n", obj)
         } 
```

To test in action:
```
  cd pkg/statedb/example
  go run .
  # another terminal
  cd pkg/statedb/example/client
  go run .
```


joamaki:
Will split this PR soon into:
- Modular gRPC API server (pkg/api/server.go and associated changes to api/v1 and daemon/cmd)
- StateDB grpc server and client implementation
- "cilium-dbg statedb devices" command (perhaps just "cilium-dbg devices"?)

PR #28577, opened 2023-10-12
Documentation: Consistently use --set for cilium install

michi-covalent:
- Replace --helm-set with --set since --helm-set is deprecated.
- Don't use --set-string unless it's necessary.
- Use envoyConfig.enabled to set enable-envoy-config to true in cilium-config ConfigMap instead of using extraConfig.

Ref: https://github.com/cilium/cilium-cli/issues/2030


meyskens (approved):
LGTM

nebril (approved):

PR #28593, opened 2023-10-13
egressgw: replace custom endpoint store with `resource.Store[CiliumEndpoint]`

rastislavs:
Use `resource.Store` for `CiliumEndpoints` instead of EGW Manager's custom endpoint cache. This removes some complexity of the custom cache, and resolves potential race conditions, e.g. when events for the same `NamespacedName` (e.g. statfulset pod) come in less common order upon pod restarts.


pkg/egressgateway/manager.go
```diff
@@ -263,6 +259,10 @@ func newEgressGatewayManager(p Params) (*Manager, error) {
 			if probes.HaveLargeInstructionLimit() != nil {
 				return fmt.Errorf("egress gateway needs kernel 5.2 or newer")
 			}
+			manager.endpointStore, err = manager.endpoints.Store(hc)
```
lmb: I think this will block until the store is initialized, which is probably not what we want?

lmb (commented):
I think you've hit a footgun in `Resource` in that using both the even stream and the Store at the same time is not allowed. cc @joamaki for confirmation or denial.

rastislavs:
@lmb 
> I think you've hit a footgun in `Resource` in that using both the even stream and the Store at the same time is not allowed.

Oh really? I am pretty sure we are using both `Store` and `Events` in BGP CP at the moment (see for example in the [diffStore](https://github.com/cilium/cilium/blob/aa10d9400ee980c141a52a1b4c5fbbfaae6dbff5/pkg/bgpv1/manager/diffstore.go)).

pkg/egressgateway/manager.go
joamaki: Yep this will block the starting and potentially will fail the start if this takes too long (the default start timeout is 5 minutes). Move this into `processEvents`.

rastislavs: You are right, blocks until it has been synchronized. I saw `Store` being called in the Start hooks on some other places, but that does not mean that it is correct :) Would it look better if I moved it to`manager.processEvents`?

rastislavs: > Move this into processEvents

makes sense, thanks!

joamaki:
> I think you've hit a footgun in Resource in that using both the even stream and the Store at the same time is not allowed. cc @joamaki for confirmation or denial.

It is allowed, it's just discouraged as then you really need to reason about what's the relation between an event and how the store is updated. It's much safer to only use the event stream and maintain a consistent state from that, like how the original code does it. The way this PR seems to use it seems safe: the event stream is only used to trigger reconciliation, but I'm not entirely sure it's worth it due to the unwritten assumptions on ordering of Events() in regards to Store(), the risks in messing up triggering and the inefficiencies of calling `List()` often against a store of potentially 10s of thousands of CiliumEndpoint objects.

Incidentally, the rational for this PR is what I think is the biggest short-coming of `Resource[T]`, in that you need to maintain your own projection of the state reconstructed from the event stream and why I think we should explore replacing `Resource[T]` with statedb where the `Table[*CiliumEndpoint]` would give you both access to state (with rich indexing) and the triggers when the state changes, but that's still few months away in the future.

pkg/egressgateway/manager.go
joamaki: Those uses of `Store()` in a `Start` hook are definitely dubious. Looks like bgpv1 does it for `CiliumBGPPeeringPolicy`, which is slightly safer in the sense that there's only few of these objects so the synchronization is going to be fast, but it is still better to avoid such usage in order to keep the startup process fast.

rastislavs: > it is still better to avoid such usage in order to keep the startup process fast

good to know, thanks @joamaki

pkg/egressgateway/policy.go
```diff
@@ -55,32 +62,37 @@ type PolicyConfig struct {
...
-	config.matchedEndpoints = make(map[endpointID]*endpointMetadata)
+// updateMatchedEndpoints update the policy's cache of matched endpoint IDs
+func (config *PolicyConfig) updateMatchedEndpoints(epStore resource.Store[*k8sTypes.CiliumEndpoint], idAllocator identityCache.IdentityAllocator) {
+	config.matchedEndpoints = nil
 
-	for _, endpoint := range epDataStore {
-		if config.matchesEndpointLabels(endpoint) {
-			config.matchedEndpoints[endpoint.id] = endpoint
+	for _, endpoint := range epStore.List() {
+		epLabels, err := getEndpointLabels(idAllocator, endpoint)
```
joamaki, line 89: This seems dangerously inefficient. We're iterating potentially very often over all endpoints (there may be thousands of them) and doing point-wise lookups to the identity allocator, a core piece of the agent. Increased lock contention there will slow everything down.

pkg/egressgateway/manager.go
```diff
@@ -116,16 +113,16 @@ type Manager struct {
 	// endpoints allows reading endpoint CRD from k8s.
 	endpoints resource.Resource[*k8sTypes.CiliumEndpoint]
 
+	// endpointStore is a read-only store of CiliumEndpoint resources
+	endpointStore resource.Store[*k8sTypes.CiliumEndpoint]
```
joamaki, line 117: A meta comment: Why are we changing the handling of `CiliumEndpoint` at all? It's wrong to even use `CiliumEndpoint` here as it doesn't work for kvstore mode in which the whole use of the `CiliumEndpoint` CRD should be disabled! The whole point of the kvstore mode is to make Cilium more scalable by disseminating the data in an efficient form via etcd.

Instead the use of it here should be removed completely and replaced with ipcache subscription.

joamaki (changes requested):

rastislavs:
OK, based on @joamaki comments it seems that this change would bring more issues than benefits, so closing it. It seems that the eventual solution should be different - as mentioned in @joamaki 's comments.

Thanks a lot for all the information here, it is very helpful.

rastislavs:
Re-opening after some internal discussion, as the better solution will probably take a bit more time - it seems we can start with this change and iterate on it later.

pkg/egressgateway/policy.go
rastislavs: CC @jibi 
Thanks to 1s `EgressGatewayReconciliationTriggerInterval` maybe this is manageable for the time being until a better solution is implemented?

pkg/egressgateway/manager.go
```diff
@@ -295,25 +291,16 @@ func (manager *Manager) eventBitmapIsSet(events ...eventType) bool {
...
-		return nil, fmt.Errorf("identity %d not found", securityIdentity)
-	}
-	return identity.Labels, nil
-}
-
 // processEvents spawns a goroutine that waits for the agent to
 // sync with k8s and then runs the first reconciliation.
 func (manager *Manager) processEvents(ctx context.Context) {
+	var err error
+	manager.endpointStore, err = manager.endpoints.Store(ctx)
```
jibi, line 298: can we move this inside `newEgressGatewayManager()` to have all the initialization logic in a single place?

pkg/egressgateway/policy.go
jibi: I initially missed the fact that we are fetching _all_ endpoints' labels each time we run the reconciliation :disappointed:

I don't know if the 1s ratelimit will be enough or not (as this also depends on the number of endpoints in the cluster), I'll have to measure some numbers.

And I guess having a local cache for labels (that gets updated only once per endpoint event in `handleEndpointEvent()`) defeats the purpose of the new `epStore` as the 2 can go out of sync, e.g. we can iterate over an endpoint in `epStore` for which we didn't receive yet the event and so labels are not yet updated?

pkg/egressgateway/manager.go
joamaki: You cannot do such a blocking operation from a constructor (otherwise tools like “cilium-agent hive” and others won’t work). It has to be done from the start hook or later.

But I agree this is a bit annoying and a better design for Store() would perhaps block on the operations and not when asking the handle.  Or having a separate `Store().Synced() <-chan struct{}` method.

rastislavs:
@jibi 
> And I guess having a local cache for labels (that gets updated only once per endpoint event in handleEndpointEvent()) defeats the purpose of the new epStore as the 2 can go out of sync

yeah, I would say so.

One more option that comes to my mind is using the labels from `CiliumEndpoint.status.identity.labels` instead of fetching the labels from the identity cache. They are in different format (with [label-source](https://docs.cilium.io/en/latest/gettingstarted/terminology/#label-source)) but that should not be that big issue? Would there be any disadvantages with that approach?

joamaki:
> @jibi
> 
> > And I guess having a local cache for labels (that gets updated only once per endpoint event in handleEndpointEvent()) defeats the purpose of the new epStore as the 2 can go out of sync
> 
> yeah, I would say so.
> 
> One more option that comes to my mind is using the labels from `CiliumEndpoint.status.identity.labels` instead of fetching the labels from the identity cache. They are in different format (with [label-source](https://docs.cilium.io/en/latest/gettingstarted/terminology/#label-source)) but that should not be that big issue. Would there be any disadvantages with that approach?

Hey, that's a great idea! The only difference that I can think of between using this and the identity allocator is slightly increased delay, but on the other hand, since we're anyway reacting to the changes to `CiliumEndpoint` and the fact that pod labels are fairly stable, I don't see that as an issue (`CiliumEndpoint` sync runs every 10 seconds). This definitely would be more consistent as this way we there's way less possible race conditions where we might end up using stale data.

rastislavs:
As there were some concerns about changing the current behavior (mainly introducing a delay upon label change until it is reflected on `CiliumEndpoint`) it seems that introducing this as intermediary solution is not a good idea, if we aim to resolve it in a different way soon.

PR #28594, opened 2023-10-13
Extend cilium scale-test to export results and gather additional data

marseel:
More details in commit messages.


.github/workflows/scale-test-gce.yaml
```diff
@@ -200,27 +204,43 @@ jobs:
...
             --state ${{ secrets.GCP_PERF_KOPS_STATE_STORE }} \
             --yes
+
+      - name: Export results and sysdump to GS bucket
+        if: ${{ always() && steps.run-cl2.outcome != 'skipped' }}
+        run: |
+          # BUILD_ID needs to be an integer
+          BUILD_ID="$(date +%s)"
+          EXPORT_DIR="${{ secrets.GCP_PERF_RESULTS_BUCKET }}/logs/scale-test-${{ github.ref_name }}/${BUILD_ID}/artifacts/"
+          gsutil -m cp -R cilium-sysdump-final.zip ./perf-tests/clusterloader2/report/* $EXPORT_DIR
```
learnitall: Can we include the repository details in the export dir path? That way we can separate results that may have been run from a fork, so as not to taint the data:

```
EXPORT_DIR="${{ secrets.GCP_PERF_RESULTS_BUCKET }}/logs/scale-test-${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.ref_name }}/${BUILD_ID}/artifacts/"
```

Can we also add the following in the artifact directory somewhere? I was thinking it would be useful to have some reference to:

* Cluster name
* Commit SHA
* URL to the run on GHA

This extra info would be useful for debugging.

.github/workflows/scale-test-gce.yaml
```diff
@@ -140,6 +143,7 @@ jobs:
             --control-plane-count 1 \
             --control-plane-size n2-standard-4 \
             --node-count 1 \
+            --node-size e2-standard-8 \
```
learnitall, line 145: Do you mind adding a comment explaining why we are using n2 for cp and e2 for worker nodes? A quick Google search explains it, but it'd be nice to have it in here too.

learnitall (changes requested):
I have two suggestions, but otherwise LGTM! Super exciting!

.github/workflows/scale-test-gce.yaml
marseel: done.

.github/workflows/scale-test-gce.yaml
marseel: I wouldn't add a repository - other repositories shouldn't be allowed to push results here, they can specify their own bucket instead. 
Good call with commit sha and URL, it definitely makes sense to add it. I've also added CL2 output too.

marseel:
Successful run: https://github.com/cilium/cilium/actions/runs/6532058299/job/17734518103
Results available in `gs://cilium-scale-results/logs/scale-test-marseel-100-node-scale-test-cl2/1697451716/`

.github/workflows/scale-test-gce.yaml
learnitall: Do you mind adding a comment in the workflow explaining that folks need to use their own bucket if running from a fork? Just to make it more explicit?

marseel: But it's also true for other variables like `GCP_PERF_SA_KEY`, `GCP_PERF_PROJECT_ID`, or `GCP_PERF_KOPS_STATE_STORE` as we don't want to interfere with ci scale tests and this is expected also when running any other CI in a fork I think?

learnitall (approved):

nbusseneau (approved):

marseel:
Marking as ready to merge as all reviews are done and it modifies new workflow only, so no need to run full tests.

PR #28595, opened 2023-10-13
hubble/relay: Remove ReportOffline and refactor PeerManager

glrf:

      GitHub issue.

This commit removes the `ReportOffline` method as the gRPC library is already able to notice broken connections and will try to reconnect on its own and uses exponential backoff to try again. While the `ReportOffline` feature might have speed up re-connection time in some cases, it potentially leaked goroutines, tried to reconnect far too often ignoring any backoff, and can result in unexpected interactions with the gRPC library.

The goroutine leak could be triggered by making more than one request every 5 seconds, when at least one peer is inaccessible. We spawned a goroutine for every request that will try and reconnect to the peer. However, we only closed a goroutine whenever a dial times out, which is every 5 second. So if the rate of request is higher than one request every 5 second, we slowly build up goroutines.

At the same time this change fixes two different issues:

* Whenever the relay reconnected to the peer service, it leaked all exiting connections, as it didn't close exiting connections when adding peers with the same name. We fix this by treating `update` and `add` notifications the same.
* There was a minor data race as `List` did not lock the peer when copying the connection (which might be updated by `connect` or `disconnect`)

### DIFF
--- a/pkg/hubble/relay/observer/observer.go
+++ b/pkg/hubble/relay/observer/observer.go
@@ -25,11 +25,9 @@ func isAvailable(conn poolTypes.ClientConn) bool {
 	if conn == nil {
 		return false
 	}
-	switch conn.GetState() {
-	case connectivity.Ready, connectivity.Idle:
-		return true
-	}
-	return false
+	state := conn.GetState()
+	return state != connectivity.TransientFailure &&
+		state != connectivity.Shutdown
 }
 
 func retrieveFlowsFromPeer(
--- a/pkg/hubble/relay/observer/server.go
+++ b/pkg/hubble/relay/observer/server.go
@@ -35,28 +35,14 @@ type PeerLister interface {
 	List() []poolTypes.Peer
 }
 
-// PeerReporter is the interface that wraps the ReportOffline method.
-type PeerReporter interface {
-	// ReportOffline allows the caller to report a peer as being offline. The
-	// peer is identified by its name.
-	ReportOffline(name string)
-}
-
-// PeerListReporter is the interface that groups the List and ReportOffline
-// methods.
-type PeerListReporter interface {
-	PeerLister
-	PeerReporter
-}
-
 // Server implements the observerpb.ObserverServer interface.
 type Server struct {
 	opts  options
-	peers PeerListReporter
+	peers PeerLister
 }
 
 // NewServer creates a new Server.
-func NewServer(peers PeerListReporter, options ...Option) (*Server, error) {
+func NewServer(peers PeerLister, options ...Option) (*Server, error) {
 	opts := defaultOptions
 	for _, opt := range options {
 		if err := opt(&opts); err != nil {
@@ -98,7 +84,6 @@ func (s *Server) GetFlows(req *observerpb.GetFlowsRequest, stream observerpb.Obs
 			s.opts.log.WithField("address", p.Address).Infof(
 				"No connection to peer %s, skipping", p.Name,
 			)
-			s.peers.ReportOffline(p.Name)
 			unavailableNodes = append(unavailableNodes, p.Name)
 			continue
 		}
@@ -201,7 +186,6 @@ func (s *Server) GetNodes(ctx context.Context, req *observerpb.GetNodesRequest) 
 			s.opts.log.WithField("address", p.Address).Infof(
 				"No connection to peer %s, skipping", p.Name,
 			)
-			s.peers.ReportOffline(p.Name)
 			continue
 		}
 		n.State = relaypb.NodeState_NODE_CONNECTED
@@ -248,7 +232,6 @@ func (s *Server) GetNamespaces(ctx context.Context, req *observerpb.GetNamespace
 			s.opts.log.WithField("address", p.Address).Infof(
 				"No connection to peer %s, skipping", p.Name,
 			)
-			s.peers.ReportOffline(p.Name)
 			continue
 		}
 
@@ -302,7 +285,6 @@ func (s *Server) ServerStatus(ctx context.Context, req *observerpb.ServerStatusR
 			s.opts.log.WithField("address", p.Address).Infof(
 				"No connection to peer %s, skipping", p.Name,
 			)
-			s.peers.ReportOffline(p.Name)
 			if len(unavailableNodes) < numUnavailableNodesReportMax {
 				unavailableNodes = append(unavailableNodes, p.Name)
 			}
--- a/pkg/hubble/relay/observer/server.go
+++ b/pkg/hubble/relay/observer/server.go
@@ -19,6 +19,7 @@ import (
 	"github.com/cilium/cilium/pkg/hubble/build"
 	"github.com/cilium/cilium/pkg/hubble/observer"
 	poolTypes "github.com/cilium/cilium/pkg/hubble/relay/pool/types"
+	"github.com/cilium/cilium/pkg/lock"
 )
 
 // numUnavailableNodesReportMax represents the maximum number of unavailable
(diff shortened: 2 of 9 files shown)


pkg/hubble/relay/observer/observer.go
```diff
@@ -25,11 +25,9 @@ func isAvailable(conn poolTypes.ClientConn) bool {
...
 		return false
 	}
-	switch conn.GetState() {
-	case connectivity.Ready, connectivity.Idle:
-		return true
-	}
-	return false
+	state := conn.GetState()
+	return state != connectivity.TransientFailure &&
+		state != connectivity.Shutdown
```
kaworu, line 30: If I understand correctly: when the state is `connectivity.Connecting` we return `false` before this patch and `true` after this patch. What is the rational behind this change?

pkg/hubble/relay/observer/server.go
```diff
@@ -318,6 +302,13 @@ func (s *Server) ServerStatus(ctx context.Context, req *observerpb.ServerStatusR
 					"error": err,
 					"peer":  p,
 				}).Warning("Failed to retrieve server status")
+				mu.Lock()
+				numConnectedNodes--
+				numUnavailableNodes++
+				if len(unavailableNodes) < numUnavailableNodesReportMax {
+					unavailableNodes = append(unavailableNodes, p.Name)
+				}
+				mu.Unlock()
```
kaworu: if `mu` is protecting `numConnectedNodes`, `numUnavailableNodes`, and `unavailableNodes`, then we should lock it when accessing them from the "main" goroutine in the `for` loop above right? (lines 286, 290-292, and 295).

pkg/hubble/relay/observer/server.go
```diff
@@ -333,13 +324,6 @@ func (s *Server) ServerStatus(ctx context.Context, req *observerpb.ServerStatusR
 	}()
 	resp := &observerpb.ServerStatusResponse{
 		Version: build.RelayVersion.String(),
-		NumConnectedNodes: &wrapperspb.UInt32Value{
-			Value: numConnectedNodes,
-		},
-		NumUnavailableNodes: &wrapperspb.UInt32Value{
-			Value: numUnavailableNodes,
-		},
-		UnavailableNodes: unavailableNodes,
```
kaworu, line 342: Why are theses lines moved?

pkg/hubble/relay/pool/manager_test.go
```diff
@@ -709,6 +710,115 @@ func TestPeerManager(t *testing.T) {
...
+		not: &peerpb.ChangeNotification{
+			Name:    "foo",
+			Address: "192.0.1.2",
+			Type:    peerpb.ChangeNotificationType_PEER_ADDED,
+		},
+	}
+
+	assert.Eventually(t, func() bool {
+		peers := mgr.List()
+		return peers[0].Address.String() == "192.0.1.2:4244"
```
kaworu: just to be less reliant on the current implementation
```suggestion
		return len(peers) == 1 && peers[0].Address.String() == "192.0.1.2:4244"
```

pkg/hubble/relay/pool/manager_test.go
```diff
@@ -709,6 +710,115 @@ func TestPeerManager(t *testing.T) {
...
+	}, time.Second, 10*time.Millisecond)
+
+	assert.Eventually(t, func() bool {
+		return ccDialCount.Load() == 2
+	}, time.Minute, 10*time.Millisecond, "reconnect to  new address")
+	assert.Eventuallyf(t, func() bool {
```
kaworu: nit
```suggestion
	assert.Eventually(t, func() bool {
```

kaworu (commented):
Thanks for the patch @glrf, some questions but overall LGTM. Would be great it the patch could be split into more commits to ease review :pray:

kaworu:
Pulling in @rolinh and @gandro for review to get more eyes on the patch.

pkg/hubble/relay/observer/observer.go
glrf: Because `connectivity.Connecting` IMO shouldn't be treated as unavailable. If we're actively trying to connect I think we should try to make the request, which will block until we either successfully reconnected or failed.

But I'm also oke with leaving it as it was.

pkg/hubble/relay/observer/server.go
glrf: Good catch! Yes I we should

pkg/hubble/relay/observer/server.go
glrf: Because at that point the requesting goroutines are still updating these values. They're only ready once the `statueses` channel is closed. They are replaced by L342-L350

glrf:
> Would be great it the patch could be split into more commits to ease review 🙏

You're right, the commit grew a bit too much. I'll try to spit it into more reviewable chunks

glrf:
Split the commit into 4 separate commits. I hope this makes it easier to review

pkg/hubble/relay/pool/manager.go
```diff
@@ -236,20 +236,40 @@ func (m *PeerManager) List() []poolTypes.Peer {
 	return peers
 }
 
-func (m *PeerManager) add(hp *peerTypes.Peer) {
+func (m *PeerManager) put(hp *peerTypes.Peer) {
```
gandro: nit, informal (i.e. non-blocking): `put` is absolutely fine, but I wanted to point out that most other code in the Cilium codebase uses the name `upsert` for such functions

pkg/hubble/relay/pool/manager.go
```diff
@@ -236,20 +236,40 @@ func (m *PeerManager) List() []poolTypes.Peer {
...
+	}
+	m.peers[hp.Name] = &peer{Peer: *hp}
 	m.mu.Unlock()
 	select {
 	case <-m.stop:
 	case m.updated <- hp.Name:
 	}
 }
 
+func peerEqual(a, b peerTypes.Peer) bool {
```
gandro: I'm a bit worried about future changes to the peer type invalidating this function. Would it make sense to move this into `peerTypes`, and maybe even make it a method of the `Peer` struct, such that it is closer to the definition of its fields?

gandro (approved):
Awesome work, some really nice catches! I've left some very minor feedback, feel free to ignore. Thanks!

pkg/hubble/relay/pool/manager.go
glrf: I personally find `upsert` a bit fuggly so I usually use `put`, but if `upsert` is more common let's rename it to `upsert`.

pkg/hubble/relay/pool/manager.go
glrf: Good point, adding an `Equal` method makes more sense.

pkg/hubble/relay/observer/server.go
```diff
@@ -276,18 +277,21 @@ func (s *Server) ServerStatus(ctx context.Context, req *observerpb.ServerStatusR
...
 			s.opts.log.WithField("address", p.Address).Infof(
 				"No connection to peer %s, skipping", p.Name,
 			)
 			if len(unavailableNodes) < numUnavailableNodesReportMax {
 				unavailableNodes = append(unavailableNodes, p.Name)
 			}
+			mu.Unlock()
 			continue
 		}
 		numConnectedNodes++
```
kaworu: I think we need to hold `mu` when incrementing this one as well.

Overall, maybe we should only keep track of `numUnavailableNodes` and compute `numConnectedNodes` as `len(peers) - numUnavailableNodes` at the end. Both the math and locking would be a tad simpler, what do you think?

pkg/hubble/relay/observer/server.go
kaworu: Ah right I overlooked that `g.Wait()` is called in another goroutine, thanks!

kaworu (changes requested):

pkg/hubble/relay/observer/server.go
glrf: Yes, you're absolutely right. And computing `numConnectedNodes` instead of counting it makes a lot of sense too. Thanks!

rolinh (approved):
More reliable, more tests, less code, less leak. Can we have more of that, please?

kaworu (approved):

PR #28598, opened 2023-10-14
ingress: Propagate labels from Ingress to LB Service

### LINKED ISSUE #28576: CFP: Propagate labels from dedicated Ingress to LB service
## Cilium Feature Proposal
        
Dedicated ingress controller mode: Cilium adds only single fixed label `cilium.io/ingress: "true"` to automated LB service however it would be highly advantageous to extend support for propagating user defined labels from the Ingress to the load balancer service. These propagated labels could serve as [LB IPAM Service Selectors](https://docs.cilium.io/en/stable/network/lb-ipam/#service-selectors), enabling the allocation of IP addresses from various IP pools.

Ultimate goal/use case here is to be able to use different IP pools for cilium's dedicated ingress controller services and/or be able to control service announcement with BGP CP based on service matching labels.

* Currently cilium operator can propagate certain annotations from Ingress to LB service, similar logic could be implemented to propagate labels.
* Alternative would be to implement annotation based service selector or pool allocation feature under LB IPAM.

log1cb0mb:
This PR adds the labels from Ingress to Load Balancer Service.

Fixes: #28576

```release-note
Propagate prefixed labels from Ingress resource to LB service
```

### DIFF
--- a/install/kubernetes/cilium/values.yaml.tmpl
+++ b/install/kubernetes/cilium/values.yaml.tmpl
@@ -721,8 +721,7 @@ ingressController:
   # -- Enable proxy protocol for all Ingress listeners. Note that _only_ Proxy protocol traffic will be accepted once this is enabled.
   enableProxyProtocol: false
 
-  # -- IngressLBAnnotations are the annotation prefixes, which are used to filter annotations to propagate
-  # from Ingress to the Load Balancer service
+  # -- IngressLBAnnotations are the annotation and label prefixes, which are used to filter annotations and/or labels to propagate from Ingress to the Load Balancer service
   ingressLBAnnotationPrefixes: ['service.beta.kubernetes.io', 'service.kubernetes.io', 'cloud.google.com']
 
   # -- Default secret namespace for ingresses without .spec.tls[].secretName set.
--- a/operator/pkg/ingress/ingress.go
+++ b/operator/pkg/ingress/ingress.go
@@ -629,8 +629,8 @@ func (ic *Controller) regenerate(ing *slim_networkingv1.Ingress, forceShared boo
 		"loadbalancer": loadbalancerMode,
 	}).Debug("Generated model for ingress")
 	cec, svc, ep, err := translator.Translate(m)
-	// Propagate Ingress annotation if required. This is applicable only for dedicated LB mode.
-	// For shared LB mode, the service annotation is defined in other higher level (e.g. helm).
+	// Propagate Ingress annotation and label if required. This is applicable only for dedicated LB mode.
+	// For shared LB mode, the service annotation and label are defined in other higher level (e.g. helm).
 	if svc != nil {
 		for key, value := range ing.GetAnnotations() {
 			for _, prefix := range ic.lbAnnotationPrefixes {
@@ -639,6 +639,17 @@ func (ic *Controller) regenerate(ing *slim_networkingv1.Ingress, forceShared boo
 						svc.Annotations = make(map[string]string)
 					}
 					svc.Annotations[key] = value
+				}
+			}
+		}
+		// Same lbAnnotationPrefixes config option is used for label propagation
+		for key, value := range ing.GetLabels() {
+			for _, prefix := range ic.lbAnnotationPrefixes {
+				if strings.HasPrefix(key, prefix) {
+					if svc.Labels == nil {
+						svc.Labels = make(map[string]string)
+					}
+					svc.Labels[key] = value
 				}
 			}
 		}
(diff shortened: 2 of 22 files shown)


operator/pkg/ingress/ingress.go
```diff
@@ -610,8 +610,8 @@ func (ic *Controller) regenerate(ing *slim_networkingv1.Ingress, forceShared boo
 		"model":        m,
 	}).Debug("Generated model for ingress")
 	cec, svc, ep, err := translator.Translate(m)
-	// Propagate Ingress annotation if required. This is applicable only for dedicated LB mode.
-	// For shared LB mode, the service annotation is defined in other higher level (e.g. helm).
+	// Propagate Ingress annotation and label if required. This is applicable only for dedicated LB mode.
+	// For shared LB mode, the service annotation and label are defined in other higher level (e.g. helm).
```
youngnick, line 633:
```suggestion
	// Propagate Ingress annotations and labels if required. This is applicable only for dedicated LB mode.
	// For shared LB mode, the service annotations and labels are defined in other higher level (e.g. helm).
```

youngnick (changes requested):
Thanks for your contribution @log1cb0mb.

I have complicated feelings about this one. In general, I'm not in favor of copying labels or annotations around - I think it's a pretty dirty hack to fit config that should really be stored in a structured field somewhere into labels. But, because we don't have any way to set the settings we want to on the generated Service object, we're kinda stuck with it.

Additionally, we've tried to make this a little more specific with annotations - do you think that we could use a similar mechanism for labels and only copy labels that start with certain strings? I haven't looked at the labels in use for configuring things on Service for some time.

If we agree we want to blindly copy _everything_, though, the code change itself is fine, but I'd like to see a couple of extra things please?

Firstly, can you update the docs page in `Documentation/network/servicemesh/ingress.rst` with some information about how all labels will now be copied in addition to annotations?

Secondly, if you could have a look at the tests and see if you can make a test that checks that everything works as expected on the generated Service, that would be _awesome_. That's a non-blocking request though, so if you would like to merge this first, one of us can either work with you or pick it up to add it later.

So, tl;dr, three things:
- Can you talk a little about copying _everything_? Would it be okay to use the same logic as the annotations and only copy some labels?
- Please add some docs in the page I mentioned.
- Extra task if you can, but non-blocking, see if you can add some tests?

Thanks again for the contribution, @log1cb0mb, really great to see and well done!

operator/pkg/ingress/ingress.go
log1cb0mb: How about copying same annotations with set prefixes as labels to LB service as well? Would that be acceptable? 
OR should we essentially duplicate the exact annotation logic for labels?

youngnick: I think duplicating the logic is better; annotations and labels are technically distinct, even though a lot of folks use them interchangeably.

youngnick: We should reuse the prefix config setting for both annotations and labels though.

log1cb0mb: @youngnick is it too much of a duplicate now?

squeed:
@log1cb0mb can you add a release note entry in the PR description? We do this for most instances of new functionality. It's just a block comment that looks like this:
````
```release-note
Adds the ability to disable tribble replication after 00:00:00Z.
```
````

File: install/kubernetes/cilium/values.yaml.tmpl [hunk 2]
```diff
@@ -722,6 +722,10 @@ ingressController:
   # from Ingress to the Load Balancer service
   ingressLBAnnotationPrefixes: ['service.beta.kubernetes.io', 'service.kubernetes.io', 'cloud.google.com']
 
+  # -- IngressLBLabels are the label prefixes, which are used to filter labels to propagate
+  # from Ingress to the Load Balancer service
```
squeed: suggestion: add "Labels matching any of these prefixes will be copied to the corresponding Service."

squeed (approved):
Code changes look good for me. A few small suggestions, but LGTM.

I did a quick check to make sure we don't elide updates if only the labels have changed (e.g. by just comparing `.Spec`. We're good there.

operator/pkg/ingress/ingress.go
youngnick: hmm, yes, it doesn't feel like we'll need separate configurability for labels vs annotations to me. I'd probably rather see a single labels+annotations prefixes config setting that applies to both.

log1cb0mb: Just so I understand, do you mean the user facing config to be same parameter that applies to both annotation and label?

If so, do you think we can run into a case where user expects to have different prefixes set for both?

youngnick: Yes, I think that the prefix config should use the same parameter between annotation and labels.

I also think that it's unlikely that the user will want that, and if people do ask for it, we can bring it back if necessary. But having separate config feels like YAGNI (You Ain't Gonna Need It) to me.

youngnick (approved):
Thanks @log1cb0mb !

meyskens (approved):
changes to ingress re LGTM

joamaki (approved):

log1cb0mb:
Is it possible to add backport-1.14 label to this?

giorio94:
Closed and reopened to see if travis gets triggered, as it was stuck

giorio94:
Reviews are in, and checks are green. Marking as ready to merge.

PR #28606, opened 2023-10-16
gateway-api: fix empty URI when removing path prefix

dddddai:
https://github.com/cilium/cilium/blob/2f8bd31d490b9439ea21d5ffdff1b9902611de24/operator/pkg/model/translation/envoy_virtual_host.go#L282-L287

The current implementation does not handle the case where the entire path is removed. In this case we must set the URI to `/`. Please see https://www.rfc-editor.org/rfc/rfc9112.html#section-3.2.1-2.
> If the target URI's path component is empty, the client __MUST__ send "/" as the path within the origin-form of request-target.

If I have a `RegexRewrite` like:
```yaml
regexRewrite:
  pattern:
    regex: ^/foo
  substitution: ""
```
I will get 502 when accessing `/foo`

### DIFF
--- a/operator/pkg/model/ingestion/gateway.go
+++ b/operator/pkg/model/ingestion/gateway.go
@@ -4,6 +4,8 @@
 package ingestion
 
 import (
+	"strings"
+
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


sayboras (approved):
The changes are verified locally with a couple of strip prefix cases. 

I think we can take another route with merge slash option, but it might require more testing.

PR #28608, opened 2023-10-16
bpf: overlay: clean up extraction of source identity

julianwiedmann:
This PR removes some clunkiness in `from-overlay`'s approach to handling the source identity, in particular when combined with `ENABLE_NODEPORT`.

Ideally we would also lift the ipcache lookup into earlier stages of the `from-overlay` program, but that's a future item.


YutaroHayakawa (approved):
LGTM 👍

brb (approved):
Nice!

PR #28610, opened 2023-10-16
helm: Correct spire labels indentation

### LINKED ISSUE #28599: Helm chart: broken spire setup

### What happened?

```
authentication:
  enabled: true
  mutual:
    spire:
      enabled: true
      install:
        enabled: true
        agent:
          labels:
            k8s-app: cilium-spire-agent
        server:
          labels:
            k8s-app: cilium-spire-server
```

Setting `labels` as above will break the helm install with:

```
Error: YAML parse error on cilium/templates/spire/agent/daemonset.yaml: error converting YAML to JSON: yaml: line 8: mapping values are not allowed in this context
```

### Cilium Version

1.14.2

### Kernel Version

-

### Kubernetes Version

-

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

Not setting labels will allow helm install to happen.

### Code of Conduct

sayboras:
Fixes: #28599

```release-note
fix: Correct spire labels identation in helm chart
```

### DIFF
--- a/install/kubernetes/cilium/templates/spire/agent/daemonset.yaml
+++ b/install/kubernetes/cilium/templates/spire/agent/daemonset.yaml
@@ -15,8 +15,8 @@ metadata:
   {{- end }}
   labels:
     app: spire-agent
-  {{- with .Values.authentication.mutual.spire.install.server.labels }}
-    {{- toYaml . | nindent 8 }}
+  {{- with .Values.authentication.mutual.spire.install.agent.labels }}
+    {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
   selector:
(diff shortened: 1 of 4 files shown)


meyskens (approved):
Good catch! LGTM

squeed (approved):

joestringer:
@sayboras Could you add a release note to the PR description? Glancing over the changes it looks like it's fixing the labels for SPIRE deployments, so it may be helpful to better highlight that in the release note.

sayboras:
> Could you add a release note to the PR description? Glancing over the changes it looks like it's fixing the labels for SPIRE deployments, so it may be helpful to better highlight that in the release note.

Thanks and done :+1:

joestringer:
@sayboras ah, what I meant is that for a user "fixing the indentation" doesn't declare what the side effects are. If this is a cosmetic change then there's no need to backport and it's not a bugfix. If this ensures that when the user configures the feature with setting X or Y, it actually works, then we should document the user impact that way so that users can understand exactly which configurations are affected and decide if the bug is related to their problem or not. And in that case yeah it makes sense to backport this to 1.14.

PR #28611, opened 2023-10-16
datapath: Introduce fake datapath cell

joamaki:
As more of the datapath code is modularized into cells, the fakes needed to be enumerated in big integration tests (daemon/cmd/daemon_test.go and test/controlplane) are starting to become too numerous. To combat this, add fake.Cell as a counterpart for the real datapath cell.

The gc.Enabler is for now left in there due to import cycle.

This PR is preparation for modularization of further components under pkg/datapath and pkg/maps.

### DIFF
--- a/daemon/cmd/daemon_test.go
+++ b/daemon/cmd/daemon_test.go
@@ -19,7 +19,6 @@ import (
 	fakecni "github.com/cilium/cilium/daemon/cmd/cni/fake"
 	"github.com/cilium/cilium/pkg/controller"
 	fakeDatapath "github.com/cilium/cilium/pkg/datapath/fake"
-	"github.com/cilium/cilium/pkg/datapath/tables"
 	datapath "github.com/cilium/cilium/pkg/datapath/types"
 	"github.com/cilium/cilium/pkg/endpoint"
 	"github.com/cilium/cilium/pkg/envoy"
@@ -33,12 +32,7 @@ import (
 	"github.com/cilium/cilium/pkg/kvstore/store"
 	"github.com/cilium/cilium/pkg/labelsfilter"
 	"github.com/cilium/cilium/pkg/lock"
-	"github.com/cilium/cilium/pkg/maps/authmap"
-	fakeauthmap "github.com/cilium/cilium/pkg/maps/authmap/fake"
 	ctmapgc "github.com/cilium/cilium/pkg/maps/ctmap/gc"
-	"github.com/cilium/cilium/pkg/maps/egressmap"
-	"github.com/cilium/cilium/pkg/maps/signalmap"
-	fakesignalmap "github.com/cilium/cilium/pkg/maps/signalmap/fake"
 	"github.com/cilium/cilium/pkg/metrics"
 	monitorAgent "github.com/cilium/cilium/pkg/monitor/agent"
 	monitorAPI "github.com/cilium/cilium/pkg/monitor/api"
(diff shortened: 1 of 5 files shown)


joamaki:
@giorio94 thoughts on what to do about the `gc.Enabler`? The import cycle happens due to ctmap/gc importing `pkg/endpoint` and then `manager_test.go` importing `pkg/datapath/fake`  (and test being in the `manager` package and not `manager_test`, tricky to change though). I was thinking of just adding a `types` package under `pkg/maps/ctmap`...

daemon/cmd/daemon_test.go
```diff
@@ -158,13 +151,8 @@ func (ds *DaemonSuite) SetUpTest(c *C) {
 				cs.Disable()
 				return cs
 			},
-			func() datapath.Datapath { return fakeDatapath.NewDatapath() },
-			func(dp datapath.Datapath) datapath.NodeIDHandler { return dp.NodeIDs() },
 			func() *option.DaemonConfig { return option.Config },
-			func() cnicell.CNIConfigManager { return &fakecni.FakeCNIConfigManager{} },
```
markpash, line 164: I'm confused why this (`CNIConfigManager`) can disappear here.

daemon/cmd/daemon_test.go
```diff
@@ -158,13 +151,8 @@ func (ds *DaemonSuite) SetUpTest(c *C) {
...
-			func() signalmap.Map { return fakesignalmap.NewFakeSignalMap([][]byte{}, time.Second) },
-			func() authmap.Map { return fakeauthmap.NewFakeAuthMap() },
-			func() egressmap.PolicyMap { return nil },
+			fakeDatapath.Cell,
```
markpash: If `fakeDatapath` provides the `tables` cell then can we remove `tables.Cell` below?

markpash: In `suite/agent.go` the `fakeDatapath.Cell` is placed directly in the hive definition. Here it appears nested in the `cell.Provide(...)`. I'm not great on my hive-fu so I'm not sure if it makes a difference. If it has to be this way, I'm curious to know why.

giorio94:
> @giorio94 thoughts on what to do about the `gc.Enabler`? The import cycle happens due to ctmap/gc importing `pkg/endpoint` and then `manager_test.go` importing `pkg/datapath/fake` (and test being in the `manager` package and not `manager_test`, tricky to change though). I was thinking of just adding a `types` package under `pkg/maps/ctmap`...

I personally don't have any really good ideas here. The `gc.Enabler` hack is currently required because the endpoints restoration logic has not been modularized yet, and the first GC pass is synchronous (not sure if that is really necessary though). I agree that we could either have a types package where to move the interface definition or just leave it where it is now (alternatively it could be made optional, but that seems more risky to me).

daemon/cmd/daemon_test.go
joamaki: Oops, that was a mistake. I originally was putting it into the fake datapath cell without thinking and then realized it really doesn't belong there, but looks like I had removed it from here. Will fix.

daemon/cmd/daemon_test.go
joamaki: Good catch, yes.

joamaki: This most definitely doesn't work. I was sloppy with the changes in this file. Will fix this and the CNIConfigManager one and make sure to integration test locally.

joamaki:
> > @giorio94 thoughts on what to do about the `gc.Enabler`? The import cycle happens due to ctmap/gc importing `pkg/endpoint` and then `manager_test.go` importing `pkg/datapath/fake` (and test being in the `manager` package and not `manager_test`, tricky to change though). I was thinking of just adding a `types` package under `pkg/maps/ctmap`...
> 
> I personally don't have any really good ideas here. The `gc.Enabler` hack is currently required because the endpoints restoration logic has not been modularized yet, and the first GC pass is synchronous (not sure if that is really necessary though). I agree that we could either have a types package where to move the interface definition or just leave it where it is now (alternatively it could be made optional, but that seems more risky to me).

Yep. Let's leave it where it is for now and then clean up once the endpoint restoration logic is lifted out.

pkg/datapath/fake/cells.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// Cell provides a fake version of the datapath cell.
+//
+// Used in integration tests such as test/controlplane.
+var Cell = cell.Module(
+	"fake-datapath",
+	"Fake Datapath",
+
+	cell.Provide(
```
mhofstetter, line 27: i wonder whether it would be worth to further extract a "fake-maps" cell.

but i think as long as the fake-datapath is small it seems ok as is.

mhofstetter (approved):
thanks for this @joamaki :+1:

nebril (approved):
tests LGTM

markpash (approved):

pkg/datapath/fake/cells.go
joamaki: Yeah, let's do that as soon as it becomes a pain.

PR #28619, opened 2023-10-16
Modularize bandwidth manager

dylandreimerink:
Upstreaming some modularization changes while working on runtime device detection, these are in preparation of removing option.Config.GetDevices and replacing it with a devices table, which requires depending subsystems to be in hive.

In theory, this should not contain functional changes, but in practice there's some questions around the order of operations, specifically in the commit modularizing the bandwidth manager.

```release-note
Modularized the bandwidth manager
```


pkg/datapath/linux/bandwidth/cell.go
```diff
@@ -0,0 +1,70 @@
...
+
+var Cell = cell.Module(
+	"bandwidth-manager",
+	"Linux Bandwidth Manager for EDT-based pacing",
+
+	cell.Config(Config{false, false}),
+	cell.Provide(newBandwidthManager),
+)
+
+type ManagerInterface interface {
```
joamaki: How about `s/Manager/manager/`, `s/ManagerInterface/Manager/`?

pkg/datapath/types/datapath.go
```diff
@@ -32,4 +32,7 @@ type Datapath interface {
 	LBMap() LBMap
 
 	Procfs() string
+
+	// Delete the endpoint entry in the throttle map
+	DeleteEndpointBandwidthLimit(epID uint16) error
```
joamaki: Could this instead be part of `bandwidth.Manager` interface? And to ease the transition, we could have `BandwithManager() bandwidth.Manager` in the `Datapath` interface.

test/controlplane/suite/agent.go
```diff
@@ -67,6 +69,7 @@ func (h *agentHandle) setupCiliumAgentHive(clientset k8sClient.Clientset, extraC
 			func() *option.DaemonConfig { return option.Config },
 			func() cnicell.CNIConfigManager { return &fakecni.FakeCNIConfigManager{} },
 			func() gc.Enabler { return gc.NewFake() },
+			func() bandwidth.ManagerInterface { return &fake.BandwidthManager{} },
```
joamaki: This should now be in `fake.Cell`

daemon/cmd/daemon_test.go
```diff
@@ -155,6 +156,7 @@ func (ds *DaemonSuite) SetUpTest(c *C) {
 			func() *option.DaemonConfig { return option.Config },
 			func() cnicell.CNIConfigManager { return &fakecni.FakeCNIConfigManager{} },
 			func() ctmapgc.Enabler { return ctmapgc.NewFake() },
+			func() bandwidth.ManagerInterface { return &bandwidth.Manager{} },
```
joamaki: This should be in `fake.Cell`

joamaki (approved):

asauber (approved):
Approving for CLI, there were no changes to the flags. I'll leave questions about init order to the datapath teams.

pkg/datapath/linux/bandwidth/cell_darwin.go
```diff
@@ -0,0 +1,17 @@
...
+
+//go:build darwin
+
+package bandwidth
+
+// NOTE: We need to mock out the whole bandwidth package on Darwin, because importing
+//       the bwmap package will cause an import of pkg/ebpf and pkg/bpf which don't
+//       build on darwin.
+
+type Manager interface {
```
joamaki: Perhaps better to add e.g. `types.go` where the `Manager` interface lives so these don't need to be kept in sync?

pkg/datapath/linux/config/config.go
```diff
@@ -77,20 +74,24 @@ var (
 // HeaderfileWriter is a wrapper type which implements datapath.ConfigWriter.
 // It manages writing of configuration of datapath program headerfiles.
 type HeaderfileWriter struct {
+	log                logrus.FieldLogger
 	nodeExtraDefines   dpdef.Map
 	nodeExtraDefineFns []dpdef.Fn
```
lmb, line 70: Why not use an extra define fn instead of adding the bandwith manager to the headerfile writer directly?

lmb (changes requested):

daemon/cmd/status.go
```diff
@@ -165,15 +165,15 @@ func (d *Daemon) getIPV4BigTCPStatus() *models.IPV4BigTCP {
...
 	}
 
-	if !option.Config.EnableBandwidthManager {
+	if !d.bwManager.Enabled() {
 		return s
 	}
 
 	s.CongestionControl = models.BandwidthManagerCongestionControlCubic
-	if option.Config.EnableBBR {
+	if d.bwManager.BBREnabled() {
```
aditighag, line 186: Are we moving all `option.Config.*` to their respective modules? Can you share more context for these changes?

Documentation/cmdref/cilium-agent_hive.md
```diff
@@ -25,6 +25,8 @@ cilium-agent hive [flags]
       --controller-group-metrics strings                          List of controller group names for which to to enable metrics. Accepts 'all' and 'none'. The set of controller group names available is not guaranteed to be stable between Cilium versions.
       --egress-gateway-policy-map-max int                         Maximum number of entries in egress gateway policy map (default 16384)
       --egress-gateway-reconciliation-trigger-interval duration   Time between triggers of egress gateway state reconciliations (default 1s)
+      --enable-bandwidth-manager                                  Enable BPF bandwidth manager
+      --enable-bbr                                                Enable BBR for the bandwidth manager
```
tommyp1ckles, line 30: on its own this is a pretty ambiguous flag name, would it be better if did something like `--enable-bandwidth-manager-bbr=xxx`?

pkg/datapath/linux/bandwidth/bandwidth.go
```diff
@@ -39,44 +62,123 @@ func GetBytesPerSec(bandwidth string) (uint64, error) {
...
 		// - https://lpc.events/event/11/contributions/953/
 		// - https://lore.kernel.org/bpf/20220302195519.3479274-1-kafai@fb.com/
 		if probes.HaveProgramHelper(ebpf.SchedCLS, asm.FnSkbSetTstamp) != nil {
-			log.Fatalf("Cannot enable --%s, needs kernel 5.18 or newer.",
-				option.EnableBBR)
+			return fmt.Errorf("cannot enable --%s, needs kernel 5.18 or newer", EnableBBR)
+		}
+	}
+
+	// Going via host stack will orphan skb->sk, so we do need BPF host
```
tommyp1ckles, line 111: > Going via host stack will orphan skb->sk

How come? (not a review point - just a question)

pkg/datapath/linux/bandwidth/bandwidth.go
```diff
@@ -39,44 +62,123 @@ func GetBytesPerSec(bandwidth string) (uint64, error) {
...
+	// routing for it to work properly.
+	if m.params.Config.EnableBBR && m.params.DaemonConfig.EnableHostLegacyRouting {
+		return fmt.Errorf("BPF bandwidth manager's BBR setup requires BPF host routing.")
```
tommyp1ckles, line 114: Why do we error out in this case instead of disabling bw management feature similar to other probe cases?

pkg/datapath/linux/bandwidth/bandwidth.go
```diff
@@ -39,44 +62,123 @@ func GetBytesPerSec(bandwidth string) (uint64, error) {
...
+			return fmt.Errorf("cannot replace root Qdisc to %s on device %s: %w", qdisc.QdiscType, device, err)
+		}
+		qdisc = &netlink.GenericQdisc{
+			QdiscAttrs: netlink.QdiscAttrs{
+				LinkIndex: link.Attrs().Index,
+				Parent:    netlink.HANDLE_ROOT,
+			},
+			QdiscType: "mq",
+		}
+		which := "mq with fq leafs"
```
tommyp1ckles: super annoying nit: leafs -> leaves

pkg/datapath/linux/bandwidth/cell.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (*manager) Stop(hive.HookContext) error {
+	return nil
+}
+
+type bandwidthManagerParams struct {
+	cell.In
+
+	Log          logrus.FieldLogger
```
tommyp1ckles, line 66: Why do we need to pass the logger around this way?

pkg/datapath/linux/bandwidth/cell.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func newBandwidthManager(lc hive.Lifecycle, p bandwidthManagerParams) Manager {
+	m := &manager{params: p}
+	lc.Append(m)
+	return m
+}
+
+func (m *manager) Start(hive.HookContext) error {
+	err := m.probe()
```
tommyp1ckles, line 49: Failure to init bw manager could be a good use case for the modular health checker 😉

pkg/endpoint/endpoint.go
```diff
@@ -1635,7 +1635,7 @@ type MetadataResolverCB func(ns, podName string) (pod *slim_corev1.Pod, _ []slim
 //
 // This assumes that after the initial successful resolution, other mechanisms
 // will handle updates (such as pkg/k8s/watchers informers).
-func (e *Endpoint) RunMetadataResolver(resolveMetadata MetadataResolverCB) {
+func (e *Endpoint) RunMetadataResolver(bwm bandwidth.Manager, resolveMetadata MetadataResolverCB) {
```
tommyp1ckles, line 1645: Can we just add the bw manager interface as a member variable of *Endpoint instead of passing it around like this (or do we expect this to be passed different instances of bw manager?)?

tommyp1ckles:
K8s changes look good 🚀 ... just some random comments/questions.

Documentation/cmdref/cilium-agent_hive.md
dylandreimerink: That would be nice, but this flag isn't new. It appears that way because now that we provide it via hive it gets added to these docs, but this flag is already in use, so we can't just change it.

pkg/datapath/linux/bandwidth/bandwidth.go
dylandreimerink: This comment is taken from the [original commit](https://github.com/cilium/cilium/commit/fdbcbc076fc5bf7e0bf57f47ae0bc2aa58153225) adding BBR, so I don't know

pkg/datapath/linux/bandwidth/cell.go
dylandreimerink: When the logger is provided via hive it automatically adds context to its fields such as the current module. It is part of the spirit of moving from dependencies on global variables to dependency injection.

daemon/cmd/status.go
dylandreimerink: We do not have hard rules about this, in general yes. However, we do not yet have solutions for all parts of `option.Config`. The situation currently is that it contains both settings and runtime config/information which we want to separate from each other. 

In this case there are reasons why bandwidth manager can't be enabled even if it has been requested to. Instead of modifying the `EnableBandwidthManager` config to reflect this we provide the runtime info via these methods now. That way there is a clear distinction between requested state and actual state.

pkg/endpoint/endpoint.go
dylandreimerink: > Can we just add the bw manager interface as a member variable of *Endpoint

That is harder than you make it sound. The current change is 4 callsites to `RunMetadataResolver`. 

Endpoints are not yet modularized so we gave to plumb the bandwidth manager to 5 callsites of `createEndpoint` which in turn is called via 42 other callsites, a portion of which will likely also need to get the bandwidth manager. So we are looking at a likely a 50 to 150 LOC change to do so :smiling_face_with_tear:

tommyp1ckles: Makes sense, that doesn't seem worth the trouble - thanks for the explanation.

Perhaps we can add a TODO for whomever modularizes endpoints/epm so we can clean up this function signature later on?  IMO In the long term, adding dependencies into the endpoint API seems like it will bloat things up a bit.

daemon/cmd/status.go
aditighag: Thanks for the context. Please add this information to the commit description if not already done.

aditighag (approved):
Minor comment about code comments. 
Approved for cilium/contributing.

tommyp1ckles (approved):
Just a few suggestions left over from my review, but relevant changes lgtm.

lmb (approved):

lmb:
Thanks for switching to the extra defines thingie.

dylandreimerink:
> Thanks for switching to the extra defines thingie.

Would have been a bit dumb to do all that work to introduce it and then not use it :grin:

pkg/datapath/linux/bandwidth/bandwidth.go
marseel: Bumping, I would like to understand that too. It seems a bit sketchy.

daemon/cmd/daemon.go
```diff
@@ -275,8 +278,6 @@ func (d *Daemon) init() error {
 	}
 
 	if !option.Config.DryMode {
-		bandwidth.InitBandwidthManager()
```
marseel, line 277: It seems like `DryMode` is not going to work with new implementation as `init()` will always be called and will create maps etc (or maybe I just missed something)?

pkg/datapath/linux/bandwidth/bandwidth.go
dylandreimerink: > Why do we error out in this case instead of disabling bw management feature

This is a case of conflicting configuration both `EnableBBR` and `EnableHostLegacyRouting` are flags. And when we encounter incompatible configuration its standard practice in Cilium to crash the agent to notify the user of their configuration. (take `initKubeProxyReplacementOptions` for example)

>  similar to other probe cases?

Perhaps this is an odd corner case where these values are both flags and dynamically set. That is a less then ideal design decision. But since its both, I feel like we should assume the user specified this combination of configuration on purpose.

> I would like to understand that too

I don't know why this incompatibility exists at the moment, if that is your question then we should involve Daniel

marseel: ok, I was more thinking about case `kernelGood` where we log warning and return nil, but for  `EnableBBR` which requires newer kernel, we return error - this seemed more inconsistent.

daemon/cmd/daemon.go
dylandreimerink: The map is only created when it is used. You are right that the `init` would run regardless of the dry mode. I changes the cell so we only register the start hook when not in dry mode. That effectively does the same as the old version.

pkg/datapath/linux/bandwidth/bandwidth.go
dylandreimerink: This behavior is inconsistent in the whole agent. The code as is, is just moved but we have not changed the semantics, all of the decisions on when to crash, fallback, and report were already there. While I do agree that this isn't ideal, I also don't see a quick fix, it really depends on who you talk to. Some people want to crash other never, so we first need to figure out a policy and then apply that consistently. But I frankly don't think that should be part of this PR. So with your permission I would like to make an issue to discuss this further and then resolve the conversation here.

pkg/datapath/linux/bandwidth/cell.go
dylandreimerink: Right, but currently the `probe` func expects that an returned error is fatal, lets group this with the other discussion. https://github.com/cilium/cilium/pull/28619#discussion_r1381140603

pkg/datapath/linux/bandwidth/bandwidth.go
dylandreimerink: https://github.com/cilium/cilium/issues/29169

PR #28620, opened 2023-10-16
ipcache: fix releasing node CIDRs after restoration

squeed:
This fixes a bug where node CIDR identities were not cleaned up after restoration. Specifically, the set of labels used for allocation vs. the set of labels released may not be the same.

This is made more awkward by the fact that we don't actually *know* the set of labels that were allocated -- only the CIDRs. This means we have to go and recreate them. So, we guess. However, now that CIDR identities can be either global *or* remote-node, we need to update our guess.

So, try and release CIDR identiites as usual, but if the node-cidr feature is enabled, and the first guess is incorrect, also try in the node-cidr style.

Fixes: 02406f2c


joestringer (approved):
Good find. Makes sense :+1:

joestringer:
(removing Jarno since he was autoassigned and I reviewed for same codeowner, will leave other review requests in place as they were explicit)

christarazi (approved):

nathanjsweet (approved):

PR #28621, opened 2023-10-16
Some small fixes to make kind-fast

squeed:
Two fixes:

1. Stop unnecessarily deleting agent pods. There was a bug where *all* agent pods were deleted for all nodes, rather than once per rollout
2. Stop unnecessarily pulling the base image. Images with the `:latest` tag have an ImagePullPolicy defaulted to `Always`, which is a waste of bandwidth and time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -660,9 +660,8 @@ kind-image-fast-agent: kind-ready build-cli build-agent ## Build cilium cli and 
 			docker exec -ti $${node_name} rm -f "${dst}/cilium-agent"; \
 			docker cp "./daemon/cilium-agent" $${node_name}:"${dst}"; \
 			docker exec -ti $${node_name} chmod +x "${dst}/cilium-agent"; \
-			\
-			kubectl --context=kind-$${cluster_name} delete pods -n kube-system -l k8s-app=cilium --force; \
 		done; \
+		kubectl --context=kind-$${cluster_name} delete pods -n kube-system -l k8s-app=cilium --force; \
 	done
 
 .PHONY: kind-image-fast-operator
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -655,9 +655,8 @@ kind-image-fast-agent: kind-ready build-cli build-agent ## Build cilium cli and
 			docker exec -ti $${node_name} rm -f "${dst}/cilium-agent"; \
 			docker cp "./daemon/cilium-agent" $${node_name}:"${dst}"; \
 			docker exec -ti $${node_name} chmod +x "${dst}/cilium-agent"; \
-			\
-			kubectl --context=kind-$${cluster_name} delete pods -n kube-system -l k8s-app=cilium --force; \
 		done; \
+		kubectl --context=kind-$${cluster_name} rollout restart -n kube-system daemonset -l app.kubernetes.io/name=cilium-agent; \
```
aanm: I agree with the node vs cluster deletion but I'm not convinced with the "rollout restart". The reason behind delete pods is that we want to deploy Cilium as fast as possible. The rollout sounds like nice way to deploy the new Cilium version but it's slower.

aanm (changes requested):

contrib/testing/kind-common.yaml
```diff
@@ -4,6 +4,10 @@ operator:
   nodeSelector:
     kubernetes.io/os: linux
     kubernetes.io/hostname: kind-worker
+  image:
+    pullPolicy: IfNotPresent
```
aditighag: Isn't `Always` a better option in dev workflow? Typically, developers would iterate over changes, and expect images to be re-loaded.

Makefile
squeed: Yeah, fair enough. Unless we patch the DaemonSet to have a larger surge this will add time. I'll set it back.

contrib/testing/kind-common.yaml
squeed: Normally yes, but the `-fast` workflow does a `docker cp` of the generated binaries, so there's no point in re-pulling the image. It won't have the desired changes anyways.

Makefile
squeed: Fixed.

contrib/testing/kind-common.yaml
aanm: Why not make this part of `contrib/testing/kind-fast.yaml` instead? This way we won't be changing the behavior of "non-fast" dev workflows

aanm (approved):

contrib/testing/kind-common.yaml
squeed: :facepalm: yes, of course. `make kind-install-cilium` already sets `pullPolicy: Never`.. which I added.

squeed: Done.

jibi (approved):

squeed:
Marking as ready-to-merge; all approvals are in and this is some development tooling.

PR #28635, opened 2023-10-17
WIP - Cilium CLI tests

giorio94:


.github/workflows/conformance-e2e.yaml
```diff
@@ -53,7 +53,7 @@ concurrency:
 env:
   # renovate: datasource=github-releases depName=cilium/cilium-cli
   cilium_cli_version: v0.15.10
```
brb: Unfortunately, you need to `s/v0.15.10//`, as it takes precedence over the `cli_ci_version` - https://github.com/cilium/cilium-cli/blob/main/action.yaml#L64.

giorio94: Ah, I thought the precedence was the other way round. Fixed now

giorio94: Thanks!

PR #28638, opened 2023-10-17
fixed cilium-operator delete CEC cilium-ingress when other ingressclass resources are created

### LINKED ISSUE #28289: cilium-operator delete CEC cilium-ingress when other ingressclass resources are created

### What happened?

Having cilium up and running using its ingress feature all works fine. We have several ingress objects serving different services. 
We are also running other ingress controller listening for ingress resources with a specific ingressclass.

When we create an ingress resources with an ingressclass does not belong to cilium, the `ciliumEnvoyConfig` resource in `kube-system` called `cilium-ingress` is deleted making all ingress routing by cilium fail.

This CEC resource is recreated only if the `cilium-operator` pod is restarted 

In addition, we have also noticed that the `cilium-ingress` CEC is deleted and created every time cilium-operator is restarted. Could it be safer mantain a reconcilation loop with this object and just update it in case it already exists instead delete? Furthermore being a managed resource, I would expect it being recreated in case of accidental deletion or manipulation.

### Cilium Version

cilium image (running): v1.14.1

### Kernel Version

5.15.128

### Kubernetes Version

v1.27.4-eks-2d98532

### Sysdump

_No response_

### Relevant log output

```shell
level=debug msg="Handling ingress added event" ingress=test k8sNamespace=default subsys=ingress-controller

level=debug msg="Generated model for ingress" forcedShared=false ingress=test k8sNamespace=default model="REDACTED" subsys=ingress-controller

level=debug msg="Translated resources for ingress" ciliumEnvoyConfig=REDACTED endpoint=nil ingress=test k8sNamespace=default service=nil subsys=ingress-controller

level=debug msg="Deleted object which was no longer tracked" obj="&{{ } {cilium-ingress  kube-system  [....] subsys=ingress-controller

level=debug msg="Skipping ingress as it is not the cilium class or default" ingress=test k8sNamespace=default subsys=ingress-controller
```

### Anything else?

example ingress triggering the issue 
```
kind: Ingress
metadata:
  creationTimestamp: null
  annotations:
    kubernetes.io/ingress.class: core
  labels:
    app: example
  name: example-ingress
  namespace: default
spec:
  rules:
  - host: example.com
    http:
      paths:
      - backend:
          service:
            name: serviceA
            port:
              number: 443
        path: /test
        pathType: ImplementationSpecific
```

### Code of Conduct

chaunceyjiang:
Fixes: #28289

When the `ingress controller` is in `shared` mode,  we create an ingress resources with an ingressclass does not belong to cilium, the ciliumEnvoyConfig resource in kube-system called cilium-ingress is deleted making all ingress routing by cilium fail.

```release-note
fixed cilium-operator delete CEC cilium-ingress when other ingressclass resources are created
```

### DIFF
--- a/operator/pkg/ingress/ingress.go
+++ b/operator/pkg/ingress/ingress.go
@@ -654,6 +654,10 @@ func (ic *Controller) isEffectiveLoadbalancerModeDedicated(ing *slim_networkingv
 }
 
 func (ic *Controller) garbageCollectOwnedResources(ing *slim_networkingv1.Ingress) error {
+	// When the Ingress is in shared mode, shared resources cannot be deleted.
+	if !ic.isEffectiveLoadbalancerModeDedicated(ing) {
+		return nil
+	}
 	cec, svc, ep, err := ic.regenerate(ing, false)
 	if err != nil {
 		return err


mhofstetter:
Thanks a lot for the contribution @chaunceyjiang

The check `BPF Checks / Check Patch` is failing with the following error message.

`Please avoid long commit subjects (max: 75, found: 93)`

Would be nice if you could shorten the commit message subject (first line) of your commit

chaunceyjiang:
@mhofstetter Thank you for your reminder.

operator/pkg/ingress/ingress.go
```diff
@@ -659,7 +659,8 @@ func (ic *Controller) garbageCollectOwnedResources(ing *slim_networkingv1.Ingres
 		return err
 	}
 
-	if cec != nil {
+	if cec != nil && ic.isEffectiveLoadbalancerModeDedicated(ing) {
```
mhofstetter: Actually, we could even skip garbage collection completely in case of a shared `Ingress`. This way it becomes obvious that no shared resource is ever garbage collected in these cases. This is also the case for the `Service` & `Endpoints` - in these cases the shared translator only returns `nil` (https://github.com/cilium/cilium/blob/main/operator/pkg/model/translation/translator.go#L90)

Therefore, i would suggest to refactor this to directly skip garbage collection if the `Ingress` is shared.

```
func (ic *Controller) garbageCollectOwnedResources(ing *slim_networkingv1.Ingress) error {
  // When the Ingress is in shared mode, shared resources cannot be deleted.
  if !ic.isEffectiveLoadbalancerModeDedicated(ing) {
    return nil
  }
  ...
```

@sayboras @chaunceyjiang WDYT?

chaunceyjiang: +1

operator/pkg/ingress/ingress.go
```diff
@@ -654,6 +654,10 @@ func (ic *Controller) isEffectiveLoadbalancerModeDedicated(ing *slim_networkingv
 }
 
 func (ic *Controller) garbageCollectOwnedResources(ing *slim_networkingv1.Ingress) error {
+	// When the Ingress is in shared mode, shared resources cannot be deleted.
+	if ic.isEffectiveLoadbalancerModeDedicated(ing) {
```
mhofstetter: sorry, i accidentally proposed the wrong solution - the condition should be negated with `!`. it needs to match all shared Ingress resources.

chaunceyjiang: Oh, Thank you for reminder.

operator/pkg/ingress/ingress.go
sayboras: ah it makes sense with Marco's explanation :+1:

mhofstetter (approved):
LGTM. thanks for considering my suggestion

PR #28641, opened 2023-10-17
gateway-api: watch ownerreference to enable stricter reconcilation

mhofstetter:
Currently, manually deleted Services, Endpoints & CiliumEnvoyConfigs that belong to a Gateway resource aren't immediately reconciled and recreated.

This commit sets the ownerreference (controller) on these resources and adds an owning watch in the respective controller. This triggers an immediate re-creation of the deleted resources.

### DIFF
--- a/operator/pkg/gateway-api/gateway.go
+++ b/operator/pkg/gateway-api/gateway.go
@@ -22,6 +22,7 @@ import (
 	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	"github.com/cilium/cilium/operator/pkg/gateway-api/helpers"
+	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
 	"github.com/cilium/cilium/pkg/logging/logfields"
 )
 
@@ -78,6 +79,10 @@ func (r *gatewayReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		// Watch related namespace in allowed namespaces
 		Watches(&corev1.Namespace{},
 			r.enqueueRequestForAllowedNamespace()).
+		// Watch created and owned resources
+		Owns(&ciliumv2.CiliumEnvoyConfig{}).
+		Owns(&corev1.Service{}).
+		Owns(&corev1.Endpoints{}).
 		Complete(r)
 }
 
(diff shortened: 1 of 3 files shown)


operator/pkg/model/translation/gateway-api/translator.go
```diff
@@ -63,12 +63,14 @@ func (t *translator) Translate(m *model.Model) (*ciliumv2.CiliumEnvoyConfig, *co
...
 
 	// Set the owner reference to the CEC object.
+	isController := true
 	cec.OwnerReferences = []metav1.OwnerReference{
 		{
 			APIVersion: gatewayv1beta1.GroupVersion.String(),
 			Kind:       source.Kind,
 			Name:       source.Name,
 			UID:        types.UID(source.UID),
+			Controller: &isController,
```
sayboras: nit: there is a small utility for getting addr

```suggestion
			Controller: model.AddressOf(true),
```

operator/pkg/gateway-api/gateway.go
```diff
@@ -78,6 +79,10 @@ func (r *gatewayReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		// Watch related namespace in allowed namespaces
 		Watches(&corev1.Namespace{},
 			r.enqueueRequestForAllowedNamespace()).
+		// Watch created and owned resources
+		Owns(&ciliumv2.CiliumEnvoyConfig{}).
```
sayboras, line 83: are you planning to add predicate to ignore object if it has some annotation (e.g. ignore.cilium.io: true), mainly for the case that some manual workaround/modification in case of any bug/issue? Probably for the parent resource as well.

sayboras (approved):
few comments as per below, but not a blocker.

operator/pkg/model/translation/gateway-api/translator.go
mhofstetter: ah nice, was looking for this. wanted to avoid adding an additional utility :) will fix this

operator/pkg/gateway-api/gateway.go
mhofstetter: it's not planned for the time being. in these cases it might already work to delete the ownerreference completely

operator/pkg/model/translation/gateway-api/translator.go
mhofstetter: changed to use `model.AddressOf(...)`

operator/pkg/gateway-api/gateway.go
joestringer: Does this comment thread require follow-up? Consider filing an issue if so.

mhofstetter: IMO it doesn't need a follow-up. I assume that other controllers also have / will introduce stricter reconciliation in the context of increasing the resilience. I'd expect raised issues ones there's an explicit need for this.

PR #28642, opened 2023-10-17
ipsec: Improve logging

pchaigno:
See commits for details.

```release-note
Fix IPsec error logs to always have all information needed to identify the XFRM configuration on which the error happened.
```

### DIFF
--- a/pkg/datapath/linux/ipsec/ipsec_linux.go
+++ b/pkg/datapath/linux/ipsec/ipsec_linux.go
@@ -200,9 +200,11 @@ func xfrmStateReplace(new *netlink.XfrmState) error {
 	}
 
 	scopedLog := log.WithFields(logrus.Fields{
-		logfields.SPI:           new.Spi,
-		logfields.SourceIP:      new.Src,
-		logfields.DestinationIP: new.Dst,
+		logfields.SPI:              new.Spi,
+		logfields.SourceIP:         new.Src,
+		logfields.DestinationIP:    new.Dst,
+		logfields.TrafficDirection: getDirFromXfrmMark(new.Mark),
+		logfields.NodeID:           getNodeIDFromXfrmMark(new.Mark),
 	})
 
 	// Check if the XFRM state already exists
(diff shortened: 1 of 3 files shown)


pkg/datapath/linux/ipsec/ipsec_linux.go
```diff
@@ -67,6 +67,14 @@ const (
 	oldXFRMOutPolicyPriority = 50
 )
 
+type dir string
```
asauber, line 70: Given the common meaning, `dir` I think would be better named `direction`.

pkg/datapath/linux/node.go
```diff
@@ -623,13 +624,13 @@ func (n *linuxNodeHandler) encryptNode(newNode *nodeTypes.Node) error {
 			ipsecIPv4Wildcard := &net.IPNet{IP: wildcardIP, Mask: net.IPv4Mask(0, 0, 0, 0)}
 			errs = errors.Join(errs, n.replaceNodeIPSecInRoute(ipsecLocal))
 			spi, err = ipsec.UpsertIPsecEndpoint(ipsecLocal, ipsecIPv4Wildcard, internalIPv4, wildcardIP, 0, ipsec.IPSecDirIn, false)
-			errs = errors.Join(errs, upsertIPsecLog(err, "EncryptNode local IPv4", ipsecLocal, ipsecIPv4Wildcard, spi))
+			errs = errors.Join(errs, upsertIPsecLog(err, "EncryptNode local IPv4", ipsecLocal, ipsecIPv4Wildcard, spi, 0))
```
asauber, line 627: A const declaration for this `0` would be more readable (similar elsewhere in this file).

asauber (dismissed):

pkg/datapath/linux/ipsec/ipsec_linux.go
pchaigno: It took me a while to see what you were talking about :sweat_smile: You mean "directory", right? I guess that probably wouldn't be top of mind for someone working on the datapath. I'll change for extra clarity if other changes are requested.

pkg/datapath/linux/node.go
pchaigno: Looks nice, but we don't have this as a common practice across the codebase. It's also somewhat orthogonal to this set of changes (you may have noticed that this zero is the same zero that's on the line above). So I'm not against it, but won't push just for that.

brb (approved):

PR #28643, opened 2023-10-17
bigtcp: Modularize and use the devices table

joamaki:
As part of the runtime device reconfiguration work, modularize bigtcp implementation and
switch use of `option.Config.GetDevices()` over to a query on `Table[*Device]`.

The  reconciliation on device changes is not yet fully implemented (WIP in https://github.com/joamaki/cilium/tree/pr/joamaki/bigtcp-reconciliation) as it requires also reconciling the GSO/GRO on all lxc and cilium_ devices.

The changes in this PR were manually tested with `test/bigtcp` in a kind cluster.


pkg/datapath/linux/bigtcp/bigtcp.go
```diff
@@ -154,149 +210,157 @@ func probeTSOMaxSize() int {
...
-	if err != nil {
-		if haveIPv4 {
-			bigTCPConfig.groIPv4MaxSize = defaultGROMaxSize
-			bigTCPConfig.gsoIPv4MaxSize = defaultGSOMaxSize
-		}
-		if haveIPv6 {
-			bigTCPConfig.groIPv6MaxSize = defaultGROMaxSize
-			bigTCPConfig.gsoIPv6MaxSize = defaultGSOMaxSize
-		}
-		for _, device := range modifiedDevices {
```
joamaki, line 281: @NikAleksandrov @borkmann hey, what I've done in this PR is to switch BIG TCP to reconcile on any changes to the native devices that Cilium is using. In addition to that it's now retrying any failures to set the GRO/GSO, but with a fairly long maximum retry interval. With that, I decided to drop the rollback code completely.

However, I'm having some doubts about whether this is the right behavior. What happens if we persistently fail to set the high GRO/GSO on one of the devices? The retry interval will rise all the way to 30 minutes to avoid log spam, but I wonder if mismatching settings in itself could lead to problems?

joamaki: We'll also follow up with a very similar PR to do these same changes for Bandwidth Manager, so in order to save some time, do you see similar issues there? Anything we should keep in mind? cc @dylandreimerink.

borkmann: > @NikAleksandrov @borkmann hey, what I've done in this PR is to switch BIG TCP to reconcile on any changes to the native devices that Cilium is using. In addition to that it's now retrying any failures to set the GRO/GSO, but with a fairly long maximum retry interval. With that, I decided to drop the rollback code completely.
> 
> However, I'm having some doubts about whether this is the right behavior. What happens if we persistently fail to set the high GRO/GSO on one of the devices? The retry interval will rise all the way to 30 minutes to avoid log spam, but I wonder if mismatching settings in itself could lead to problems?

I haven't tested with HW NIC in terms of what happens if the phys NICs mismatch that of the veth one. Right now it's all aligned, and ideally it should be all aligned. If for some weird reason the limit becomes lower on one of the NICs (I presume this would only happen if a new NIC is dynamically added which would have lower limit), then it would probably make sense to align it also for all NICs in hostns (phys ones, cilium_*, lxc*).

Similarly, if a user turns on the feature where it was off before, the reconciler could detect this and attempt to align the higher limit to all NICs in hostns (phys ones, cilium_*, lxc*), and vice versa when it was on and it's turned off now.

One thing that is important is the ordering of settings, so first IPv6, then IPv4 as the other way around it would override the IPv4 one.

joamaki: Thanks! Seems like I had missed the veth gro/gso setup in `pkg/datapath/connector/veth.go`. So the reconciler will need to not only deal with native devices, but all lxc and `cilium_*` devices. 

As we're trying to solve the runtime device detection in stages, with the first one just being switching uses of `GetDevices()` to the `Table[*Device]` I'm going to split the reconciliation commits into its own PR and iterate on that there later on.

dylandreimerink (approved):

pippolo84 (approved):
LGTM :100:

asauber (approved):
Config changes LGTM. Love to see more modular flags.

PR #28614, opened 2023-10-16
chore: Use slices package from Go std lib

pippolo84:
Go 1.21 promoted the slices package from the experimental repository ("golang.org/x/exp/slices") to the standard library. The commit updates the import paths in the entire codebase.

See https://go.dev/doc/go1.21#slices for more info about slices package in Go 1.21

PR #28649, opened 2023-10-17
labels: Use slices.Sort instead of sort.Strings

pippolo84:
As suggested by package "sort" documentation (https://pkg.go.dev/sort#Strings), `slices.Sort` should be preferred to `sort.Strings` since it runs faster.

Comparison between `sort.Strings` ("old") and `slices.Sort` ("new"):

```
name                        old time/op    new time/op    delta
pkg:github.com/cilium/cilium/pkg/labels goos:linux goarch:amd64
Labels_SortedList-8            787ns ± 4%     736ns ±12%   -6.42%  (p=0.014 n=7+8)
pkg:github.com/cilium/cilium/pkg/labels/cidr goos:linux goarch:amd64
Labels_SortedListCIDRIDs-8    5.22µs ± 8%    4.67µs ± 4%  -10.63%  (p=0.000 n=10+10)

name                        old alloc/op   new alloc/op   delta
pkg:github.com/cilium/cilium/pkg/labels goos:linux goarch:amd64
Labels_SortedList-8             360B ± 0%      336B ± 0%   -6.67%  (p=0.000 n=10+10)
pkg:github.com/cilium/cilium/pkg/labels/cidr goos:linux goarch:amd64
Labels_SortedListCIDRIDs-8    1.62kB ± 0%    1.60kB ± 0%   -1.48%  (p=0.000 n=10+10)

name                        old allocs/op  new allocs/op  delta
pkg:github.com/cilium/cilium/pkg/labels goos:linux goarch:amd64
Labels_SortedList-8             3.00 ± 0%      2.00 ± 0%  -33.33%  (p=0.000 n=10+10)
pkg:github.com/cilium/cilium/pkg/labels/cidr goos:linux goarch:amd64
Labels_SortedListCIDRIDs-8      3.00 ± 0%      2.00 ± 0%  -33.33%  (p=0.000 n=10+10)
```

### DIFF
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"slices"
 	"sort"
 	"strings"
 )
@@ -512,7 +513,7 @@ func (l Labels) SortedList() []byte {
 	for k := range l {
 		keys = append(keys, k)
 	}
-	sort.Strings(keys)
+	slices.Sort(keys)
 
 	// Labels can have arbitrary size. However, when many CIDR identities are in
 	// the system, for example due to a FQDN policy matching S3, CIDR labels


nathanjsweet (approved):

pkg/labels/labels.go
```diff
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"slices"
 	"sort"
```
schlosna, line 12: Is it worth replacing the other usage on line 169 in this file? Would a broader followup PR (or set of PRs) to replace `sort.Strings` and similar functions with `slices.Sort` be amenable?

https://github.com/cilium/cilium/blob/b5c7721274a97a37cf30291b5dc71182e3374cdb/pkg/labels/labels.go#L169

pippolo84: Sure! I was focusing on `SortedList` performance and thus I applied the change just there, but it would be nice to update the entire codebase. Are you up to open a PR for that?

schlosna: Sure, I can try to put together a PR for some broader sorting & slices cleanup (e.g. migrating off `golang.org/x/exp/slices` where possible).

pippolo84: Regarding the migration `golang.org/x/exp/slices` -> `slices` I've already opened https://github.com/cilium/cilium/pull/28614.
But sounds good for replacing sorting calls, thanks!

PR #28654, opened 2023-10-17
 gateway-api: Support GRPCRoute resource

### LINKED ISSUE #21928: CFP: Support GRPCRoute in Cilium Gateway API
## Cilium Feature Proposal
        
The first cut for Gateway API was done in https://github.com/cilium/cilium/pull/21749, however, we have a couple of follow-up tasks to support more features from Gateway API upstream.

This is to track the work to support [GRPCRoute](https://gateway-api.sigs.k8s.io/api-types/grpcroute/).

## Goals

Based on the progress and code re-usability introduced in [TLSRoutes](https://github.com/cilium/cilium/issues/21927) we should build a similar implementation of GRPCRoute. This route should re-use logic from both TLS and HTTP routes to route gRPC. 

To implement it we should re-use the Envoy logic as an GRPCRoute can be attached to a HTTP(S) Gateway, we can therefore re-use that for TLS termination. Preferably it should use gRPC specific routing in Envoy where possible. 

## Proposed Solution

TBD

## Tasks

- [ ] Add required reconcile functions to validate GRPCRoute
  - [ ] Validate against hostname re-use between HTTP and gRPC
- [ ] Add a translator module for GRPCRoutes rules to Envoy rules
  - [ ] Consider a gRPC Web filter https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/grpc_web_filter & https://github.com/grpc/grpc-web/blob/master/net/grpc/gateway/examples/echo/envoy.yaml
- [ ] Handle edge cases according to spec (TBD)
- [ ] Validate required CRDs will be installed
- [ ] Pass conformance tests
- [ ] Awesome documentation!

sayboras:
### Description

Fixes: https://github.com/cilium/cilium/issues/21928

```release-note
gateway-api: Support GRPCRoute resource
```

### DIFF
--- a/.github/workflows/conformance-gateway-api.yaml
+++ b/.github/workflows/conformance-gateway-api.yaml
@@ -218,15 +218,6 @@ jobs:
       - name: Run Gateway API conformance test
         timeout-minutes: 30
         run: |
-          cat << EOF | kubectl apply -f -
-          apiVersion: gateway.networking.k8s.io/v1
-          kind: GatewayClass
-          metadata:
-            name: cilium
-          spec:
-            controllerName: io.cilium/gateway-controller
-          EOF
-          
           if [ ${{ matrix.conformance-profile }} == "true" ]; then
             GATEWAY_API_CONFORMANCE_TESTS=1 go test \
               -p 4 \
--- a/.github/workflows/conformance-gateway-api.yaml
+++ b/.github/workflows/conformance-gateway-api.yaml
@@ -169,12 +169,14 @@ jobs:
           kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${{ env.gateway_api_version }}/config/crd/${{ matrix.crd-channel }}/gateway.networking.k8s.io_referencegrants.yaml
           ## TLSRoute is only available in experimental channel in v0.7.0
           kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${{ env.gateway_api_version }}/config/crd/experimental/gateway.networking.k8s.io_tlsroutes.yaml
+          kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${{ env.gateway_api_version }}/config/crd/experimental/gateway.networking.k8s.io_grpcroutes.yaml
 
           # To make sure that Gateway API CRs are available
           kubectl wait --for condition=Established crd/gatewayclasses.gateway.networking.k8s.io --timeout=${{ env.timeout }}
           kubectl wait --for condition=Established crd/gateways.gateway.networking.k8s.io --timeout=${{ env.timeout }}
           kubectl wait --for condition=Established crd/httproutes.gateway.networking.k8s.io --timeout=${{ env.timeout }}
           kubectl wait --for condition=Established crd/tlsroutes.gateway.networking.k8s.io --timeout=${{ env.timeout }}
+          kubectl wait --for condition=Established crd/grpcroutes.gateway.networking.k8s.io --timeout=${{ env.timeout }}
           kubectl wait --for condition=Established crd/referencegrants.gateway.networking.k8s.io --timeout=${{ env.timeout }}
 
       - name: Install Cilium
--- a/.github/workflows/conformance-gateway-api.yaml
+++ b/.github/workflows/conformance-gateway-api.yaml
@@ -215,6 +215,24 @@ jobs:
             --values metallb_values.yaml \
             --wait
 
+      - name: Run simple Gateway API GRPCRoute test (temporary till upstream conformance tests)
+        timeout-minutes: 10
+        run: |
+          kubectl apply -f examples/kubernetes/gateway/grpc-route.yaml
+          # Install grpcurl binary
+          go install github.com/fullstorydev/grpcurl/cmd/grpcurl@latest
+          
+          # Wait for the deployment
+          kubectl wait --for=condition=Available --all deployment --timeout=${{ env.timeout }}
+          lb=$(kubectl get services cilium-gateway-grpc -o json | jq '.status.loadBalancer.ingress[0].ip' | jq -r .)
+          grpcurl -plaintext -authority=my-grpc-service.foo.com $lb:80 yages.Echo/Ping
+          curl -s -v --fail $lb/yages.Echo/Ping \
+            -XPOST \
+            -H 'Host: my-grpc-service.foo.com' \
+            -H 'Content-Type: application/grpc-web-text' \
+            -H 'Accept: application/grpc-web-text' \
+            -d'AAAAAAA='
+
       - name: Run Gateway API conformance test
         timeout-minutes: 30
         run: |
--- a/operator/pkg/gateway-api/gateway_reconcile.go
+++ b/operator/pkg/gateway-api/gateway_reconcile.go
@@ -84,6 +84,12 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return fail(err)
 	}
 
+	grpcRouteList := &gatewayv1alpha2.GRPCRouteList{}
+	if err := r.Client.List(ctx, grpcRouteList); err != nil {
+		scopedLog.WithError(err).Error("Unable to list GRPCRoutes")
+		return fail(err)
+	}
+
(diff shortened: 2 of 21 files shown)


qmonnet (approved):
I came for the docs, and the docs looked good.

nbusseneau (approved):
Trivial ack for ci-structure.

squeed (approved):

operator/pkg/gateway-api/grpcroute.go
```diff
@@ -0,0 +1,169 @@
...
+					)
+				}
+			}
+			return backendServices
+		},
+	); err != nil {
+		return err
+	}
+
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &gatewayv1alpha2.GRPCRoute{}, gatewayIndex,
```
squeed: This line is is some proper arcane incantation. Can you please add a comment :-)?

operator/pkg/gateway-api/grpcroute_reconcile.go
```diff
@@ -0,0 +1,132 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	gatewayv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+	"github.com/cilium/cilium/operator/pkg/gateway-api/routechecks"
+	"github.com/cilium/cilium/pkg/logging/logfields"
+)
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
```
squeed, line 23: Please add a comment giving a basic overview, especially the inputs and outputs.

.github/workflows/conformance-gateway-api.yaml
```diff
@@ -169,6 +169,7 @@ jobs:
           kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${{ env.gateway_api_version }}/config/crd/${{ matrix.crd-channel }}/gateway.networking.k8s.io_referencegrants.yaml
           ## TLSRoute is only available in experimental channel in v0.7.0
           kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${{ env.gateway_api_version }}/config/crd/experimental/gateway.networking.k8s.io_tlsroutes.yaml
+          kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${{ env.gateway_api_version }}/config/crd/experimental/gateway.networking.k8s.io_grpcroutes.yaml
 
           # To make sure that Gateway API CRs are available
```
mhofstetter, line 174: shouldn't we add a check for the `GrpcRoute`?

operator/pkg/gateway-api/grpcroute.go
```diff
@@ -0,0 +1,169 @@
...
+		func(rawObj client.Object) []string {
+			hr := rawObj.(*gatewayv1alpha2.GRPCRoute)
```
mhofstetter: nit: `hr` -> `gr` (or just `route`)

operator/pkg/gateway-api/grpcroute.go
```diff
@@ -0,0 +1,169 @@
...
+	client.Client
+	Scheme *runtime.Scheme
+
+	Model *internalModel
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *grpcRouteReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &gatewayv1alpha2.GRPCRoute{}, backendServiceIndex,
+		func(rawObj client.Object) []string {
```
mhofstetter: might be worth to define the index functions as normal functions (and reference them here) instead of anonymous functions. should be easier to read and document them.

joamaki (approved):
Approve for `install/kubernetes` changes.

operator/pkg/gateway-api/grpcroute.go
```diff
@@ -0,0 +1,169 @@
...
+			hr, ok := rawObj.(*gatewayv1alpha2.GRPCRoute)
```
mhofstetter: nit: `hr` -> `gr` (or just `route`)

operator/pkg/gateway-api/gateway_reconcile.go
```diff
@@ -226,6 +233,16 @@ func (r *gatewayReconciler) filterHTTPRoutesByGateway(ctx context.Context, gw *g
 	return filtered
 }
 
+func (r *gatewayReconciler) filterGRPCRoutesByGateway(ctx context.Context, gw *gatewayv1beta1.Gateway, routes []gatewayv1alpha2.GRPCRoute) []gatewayv1alpha2.GRPCRoute {
```
mhofstetter: maybe out of scope for this PR: we should start thinking about a way to get rid of this boilerplate for every route-type.

.github/workflows/conformance-gateway-api.yaml
```diff
@@ -169,6 +169,7 @@ jobs:
           kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${{ env.gateway_api_version }}/config/crd/${{ matrix.crd-channel }}/gateway.networking.k8s.io_referencegrants.yaml
           ## TLSRoute is only available in experimental channel in v0.7.0
           kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${{ env.gateway_api_version }}/config/crd/experimental/gateway.networking.k8s.io_tlsroutes.yaml
+          kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/${{ env.gateway_api_version }}/config/crd/experimental/gateway.networking.k8s.io_grpcroutes.yaml
```
mhofstetter, line 172: do we have to add the grpcRoute to the list of supported features to include this in the gateway-api conformance tests?

https://github.com/cilium/cilium/blob/7add21c7e44386cb5d9f72954bf3b911e4ce242a/.github/workflows/conformance-gateway-api.yaml#L121

edit: or is this the gRPC Route not yet covered by the conformance tests or included in the http route? :thinking: https://gateway-api.sigs.k8s.io/geps/gep-917/#conformance-profiles

mhofstetter (commented):
nice work @sayboras :partying_face: 

i took a first look and added some comments inline. I'm not sure whether the gateway-api conformance tests are already enabled without having added the grpcRoute feature to the list of supported features. Might be worth to activate that first and have some feedback from the conformance tests too :smile: (comment: https://github.com/cilium/cilium/pull/28654#discussion_r1365137956)

.github/workflows/conformance-gateway-api.yaml
sayboras: ah good catch :100:

operator/pkg/gateway-api/gateway_reconcile.go
sayboras: yeah, we might need to wrap the types, similar to what we have with input check. 

Will do this as a follow up.

.github/workflows/conformance-gateway-api.yaml
sayboras: yup, there is no conformance test right now, however, the translation logic is well covered in HTTP conformance tests, which gives us high enough confidence :pray:

mhofstetter: But i assume the HTTP conformance tests don't make use of the CRD `GRPCRoute`? if not we should at least introduce such a test (e.g. in the connectivity tests) as a follow up. WDYT?

sayboras: yup, makes sense to add some basic grpcroute test instead of doing it manually :+1:

sayboras: let me do it in this PR as well, marking this draft to add grpc test

sayboras: done, one simple grpc echo route, and curl with grpc web steps are added.

mhofstetter: nice, thanks!

mhofstetter (approved):
LGTM, thanks again!

Issue #27405, opened 2023-08-09
GC interval auto calculation is wrong when woken up by the map full signals


Jiang1155:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

During operation, we sometimes saw CT table is about full and not cleaned up for about an hour. Clients see some http errors.

After some investigation, I noticed that there are 15 "Conntrack garbage collector interval recalculated" messages printed out in within one minute. As a result, the newInterval became larger and larger. Starting from 2 min to 12h.  And then there was no GC messages for about an hour later. During that time, the system was in fact under a heavy load, and the GC should be invoked more frequently. 

I think one reason for this issue is because of this line:
`case <-ctTimer.After(ctmap.GetInterval(maxDeleteRatio)):`

https://github.com/cilium/cilium/blob/main/pkg/maps/ctmap/gc/gc.go#L137

When the gc is woken up by signal, the select will be done and then go to next for loop iteration. In the next iteration, the GetInterval will be called again even if it has not expired last time. As a result, the time interval will become longer and longer which contradicts the intention to run gc when in heavy load. The GetInterval() has a wrong assumption that it will be only called after timer expires.

To fix it, one idea is to pass an additional argument to GetInterval() to let it know if it is woken by signal or not. If yes, no need to calculate interval again ( or make interval even smaller?)

E.g
```
           select {
			case x := <-wakeup:
				triggeredBySignal = true
				ipv4 = false
				ipv6 = false
				if x == signal.SignalProtoV4 {
					ipv4 = true
				} else if x == signal.SignalProtoV6 {
					ipv6 = true
				}
				
				// Drain current queue since we just woke up anyway.
				for len(wakeup) > 0 {
					x := <-wakeup
					if x == signal.SignalProtoV4 {
						ipv4 = true
					} else if x == signal.SignalProtoV6 {
						ipv6 = true
					}
				}
+                               wakeupBySig = true
			case <-ctTimer.After(ctmap.GetInterval(mapType, maxDeleteRatio, wakeupBySig)):
				ipv4 = ipv4Orig
				ipv6 = ipv6Orig
			}
```

 

### Cilium Version

Found issue on 1.11, but same code exists in 1.13

### Kernel Version

5.10

### Kubernetes Version

1.21

### Relevant log output

```shell
[...]

PR #28657, opened 2023-10-17
maps: Consider actual passed time for GC interval calculation

gentoo-root:
When GetInterval calculates the new GC interval, it uses the result of the previous calculation as a pivot point. However, if GC was triggered by a signal, smaller time interval has passed, therefore, expectations on the delete ratio should be lower. Adjust the delete ratio proportionally to avoid increasing the interval uncontrollably when multiple signals arrive over a short period of time.

Ref: #27405

```release-note
Fix GC interval calculation by taking into account the actual time passed between GC runs.
```

### DIFF
--- a/pkg/maps/ctmap/gc/gc.go
+++ b/pkg/maps/ctmap/gc/gc.go
@@ -101,6 +101,7 @@ func (gc *GC) Enable(restoredEndpoints []*endpoint.Endpoint) {
 		ipv6 := gc.ipv6
 		triggeredBySignal := false
 		ctTimer, ctTimerDone := inctimer.New()
+		var gcPrev time.Time
 		defer ctTimerDone()
 		for {
 			var (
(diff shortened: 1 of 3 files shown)


pkg/maps/ctmap/gc/gc.go
```diff
@@ -88,6 +88,7 @@ func (gc *GC) Enable(restoredEndpoints []*endpoint.Endpoint) {
 		ipv6 := gc.ipv6
 		triggeredBySignal := false
 		ctTimer, ctTimerDone := inctimer.New()
+		var gcPrev time.Time
```
joamaki, line 104: Where are we assigning to this?

gentoo-root: Thanks! I missed it.

joamaki (approved):

PR #28659, opened 2023-10-17
don't remove neighbor link state file if migrateOnly

liuyuan10:
NodeCleanNeighbors is also called when enable-l2-neigh-discovery with migrateOnly==false. In that case, all existing arp entries can still be removed because neighLastPingByNextHop is still empty (https://github.com/cilium/cilium/blob/main/pkg/datapath/linux/node.go#L2013). The neighbor link state file then is removed https://github.com/cilium/cilium/blob/main/pkg/datapath/linux/node.go#L2113

The file is only resotred again when next time
NodeConfigurationChanged() is called.

This can break arp entry cleanup when next time cilium is started without enable-l2-neigh-discovery

### DIFF
--- a/pkg/datapath/linux/node.go
+++ b/pkg/datapath/linux/node.go
@@ -2109,7 +2109,7 @@ func (n *linuxNodeHandler) NodeCleanNeighbors(migrateOnly bool) {
 	// up all neighbors.
 	successClean := true
 	defer func() {
-		if successClean {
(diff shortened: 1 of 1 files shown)


borkmann (approved):

squeed:
IPSec review no longer needed (in fact, it was removed from this file in a subsequent PR). All required tests are passing. Marking as ready-to-merge.

PR #25493, opened 2023-05-16
Use cli-based Helm install for `tests-smoke` conformance workflow

bleggett:
Move `conformance-smoke` workflow to helm install mode, in an effort to knock one of the line items off of: #25156

This is a small change for the most part since these tests already used Helm, just not via`cilium-cli` proper.

`act` isn't cooperating so I'm leaning on CI for troubleshooting, I expect it to fail.

PR #28661, opened 2023-10-17
Correctly use cli installer action in ipv4/6 smoke

bleggett:
Fix for regression in smoke test script introduced by https://github.com/cilium/cilium/pull/25493

See https://github.com/cilium/cilium/pull/25493/files#r1362783422

### DIFF
--- a/.github/workflows/tests-smoke-ipv6.yaml
+++ b/.github/workflows/tests-smoke-ipv6.yaml
@@ -130,12 +130,10 @@ jobs:
           echo cilium_install_defaults=${CILIUM_INSTALL_DEFAULTS} >> $GITHUB_OUTPUT
 
       - name: Install Cilium CLI
-        run: |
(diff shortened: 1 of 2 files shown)


joestringer (changes requested):
Thanks! Looks like `.github/workflows/tests-smoke.yaml` could do with the same treatment.

joestringer (commented):
Changes LGTM, could you rebase to main / drop the merge commit?

joestringer:
^ Context, we try to keep Cilium git log linear for easier bisectability, and keep the commit messages simple and well-written. The original commit message is good from that perspective, but the second one lacks context. Perhaps squash them all together during the rebase and just reword the commit to drop the IPv6 part?

joestringer (approved):
LGTM

nbusseneau (approved):

nbusseneau:
Both affected workflows passed, no need to run the rest of the CI on this one, marking ready to merge.

PR #28663, opened 2023-10-17
operator: Fix logic used to sync Cilium's IngressClass on startup

learnitall:

      GitHub issue.
- [X] Are you a user of Cilium? Please add yourself to the [Users doc](https://github.com/cilium/cilium/blob/main/USERS.md)

Please see the commit for more details. The TL;DR here is that the Ingress Controller in the Cilium Operator will always assume that Cilium is the non-default IngressClass on startup. This is because the Ingress Controller processes events from its Ingress resource Informer before processing events from its IngressClass Informer (which is a part of the `ingressClassManager`).

This bug may cause two full reconciliations of Ingress resources on the startup of the operator: one after the Ingress resource Informer is synced, and one after the ingress controller learns Cilium should act as the default IngressClass in the cluster.

```release-note
Fix incorrect logic used by the Ingress Controller to sync Cilium's IngressClass on startup.
```

### DIFF
--- a/operator/k8s/resource_ctors.go
+++ b/operator/k8s/resource_ctors.go
@@ -16,6 +16,7 @@ import (
 	cilium_api_v2alpha1 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2alpha1"
 	"github.com/cilium/cilium/pkg/k8s/client"
 	"github.com/cilium/cilium/pkg/k8s/resource"
+	slim_networkingv1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/networking/v1"
 	"github.com/cilium/cilium/pkg/k8s/utils"
 )
 
@@ -56,4 +57,14 @@ func CiliumEndpointSliceResource(lc hive.Lifecycle, cs client.Clientset, opts ..
 		opts...,
 	)
 	return resource.New[*cilium_api_v2alpha1.CiliumEndpointSlice](lc, lw, resource.WithMetric("CiliumEndpointSlice")), nil
+}
+
+func IngressClassResource(lc hive.Lifecycle, cs client.Clientset, opts ...func(*metav1.ListOptions)) (resource.Resource[*slim_networkingv1.IngressClass], error) {
+	if !cs.IsEnabled() {
+		return nil, nil
+	}
+	lw := utils.ListerWatcherWithModifiers(
+		utils.ListerWatcherFromTyped[*slim_networkingv1.IngressClassList](cs.Slim().NetworkingV1().IngressClasses()), opts...,
+	)
+	return resource.New[*slim_networkingv1.IngressClass](lc, lw, resource.WithMetric("IngressClass")), nil
 }
--- a/operator/pkg/ingress/ingress.go
+++ b/operator/pkg/ingress/ingress.go
@@ -6,7 +6,6 @@ package ingress
 import (
 	"context"
 	"fmt"
-	"strconv"
 	"strings"
 
 	"github.com/sirupsen/logrus"
@@ -26,6 +25,7 @@ import (
 	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
 	k8sClient "github.com/cilium/cilium/pkg/k8s/client"
 	"github.com/cilium/cilium/pkg/k8s/informer"
+	"github.com/cilium/cilium/pkg/k8s/resource"
 	slim_corev1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/core/v1"
 	slim_networkingv1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/networking/v1"
 	"github.com/cilium/cilium/pkg/k8s/utils"
@@ -92,15 +92,20 @@ type Controller struct {
 	sharedLBServiceName     string
 	ciliumNamespace         string
 	defaultLoadbalancerMode string
-	isDefaultIngressClass   bool
-	defaultSecretNamespace  string
-	defaultSecretName       string
+
+	defaultSecretNamespace string
+	defaultSecretName      string
 
 	sharedLBStatus *slim_corev1.LoadBalancerStatus
 }
 
 // NewController returns a controller for ingress objects having ingressClassName as cilium
-func NewController(clientset k8sClient.Clientset, options ...Option) (*Controller, error) {
+func NewController(
+	ctx context.Context,
+	clientset k8sClient.Clientset,
+	ingressClasses resource.Resource[*slim_networkingv1.IngressClass],
+	options ...Option,
+) (*Controller, error) {
 	opts := DefaultIngressOptions
 	for _, opt := range options {
 		if err := opt(&opts); err != nil {
@@ -157,10 +162,7 @@ func NewController(clientset k8sClient.Clientset, options ...Option) (*Controlle
 		nil,
 	)
 
-	ingressClassManager, err := newIngressClassManager(clientset, ic.queue, opts.MaxRetries)
-	if err != nil {
-		return nil, err
-	}
+	ingressClassManager := newIngressClassManager(ctx, ic.queue, ingressClasses)
 	ic.ingressClassManager = ingressClassManager
 
 	serviceManager, err := newServiceManager(clientset, ic.queue, opts.MaxRetries)
@@ -195,19 +197,28 @@ func NewController(clientset k8sClient.Clientset, options ...Option) (*Controlle
 }
 
 // Run kicks off the controlled loop
-func (ic *Controller) Run() {
+func (ic *Controller) Run(ctx context.Context) error {
 	defer ic.queue.ShutDown()
-	go ic.ingressInformer.Run(wait.NeverStop)
-	if !cache.WaitForCacheSync(wait.NeverStop, ic.ingressInformer.HasSynced) {
-		return
+
+	go ic.ingressClassManager.Run(ctx)
+
+	// This should only return an error if the context is canceled.
+	if err := ic.ingressClassManager.WaitForSync(ctx); err != nil {
+		return err
 	}
 
-	go ic.ingressClassManager.Run()
+	go ic.ingressInformer.Run(wait.NeverStop)
+	if !cache.WaitForCacheSync(wait.NeverStop, ic.ingressInformer.HasSynced) {
+		return fmt.Errorf("unable to wait for Ingress cache sync")
+	}
+
 	go ic.serviceManager.Run()
 	go ic.secretManager.Run()
 
 	for ic.processEvent() {
 	}
+
+	return nil
 }
 
 func (ic *Controller) processEvent() bool {
@@ -246,7 +257,7 @@ func hasEmptyIngressClass(ingress *slim_networkingv1.Ingress) bool {
 func (ic *Controller) isCiliumIngressEntry(ingress *slim_networkingv1.Ingress) bool {
 	className := getIngressClassName(ingress)
 
-	if (className == nil || *className == "") && ic.isDefaultIngressClass {
+	if (className == nil || *className == "") && ic.ingressClassManager.IsDefault() {
 		return true
 	}
 
@@ -354,40 +365,30 @@ func (ic *Controller) handleIngressServiceUpdatedEvent(ingressServiceUpdated ing
 }
 
 func (ic *Controller) handleCiliumIngressClassUpdatedEvent(event ciliumIngressClassUpdatedEvent) error {
-	log.Debugf("Cilium IngressClass updated")
-	previousValue := ic.isDefaultIngressClass
-	if val, ok := event.ingressClass.GetAnnotations()[slim_networkingv1.AnnotationIsDefaultIngressClass]; ok {
-		isDefault, err := strconv.ParseBool(val)
-		if err != nil {
-			log.WithError(err).Warnf("Failed to parse annotation value for %q", slim_networkingv1.AnnotationIsDefaultIngressClass)
-			return err
-		}
-		ic.isDefaultIngressClass = isDefault
-	} else {
-		// if the annotation is not set we are not the default ingress class
-		ic.isDefaultIngressClass = false
+	if !event.changed {
+		return nil
 	}
 
-	if previousValue != ic.isDefaultIngressClass {
-		log.Debugf("Cilium IngressClass default value changed, re-syncing ingresses")
-		// ensure that all ingresses are in the correct state
-		for _, k := range ic.ingressStore.ListKeys() {
-			ing, err := ic.getByKey(k)
-			if err != nil {
+	log.WithField(CiliumIngressClassIsDefault, event.isDefault).Info(
+		"Cilium IngressClass default value changed, re-syncing ingresses",
+	)
+	// ensure that all ingresses are in the correct state
+	for _, k := range ic.ingressStore.ListKeys() {
+		ing, err := ic.getByKey(k)
+		if err != nil {
+			return err
+		}
+
+		if ic.isCiliumIngressEntry(ing) {
+			// make sure that the ingress is in the correct state
+			if err := ic.ensureResources(ing, false); err != nil {
 				return err
 			}
-
-			if ic.isCiliumIngressEntry(ing) {
-				// make sure that the ingress is in the correct state
-				if err := ic.ensureResources(ing, false); err != nil {
-					return err
-				}
-			} else if hasEmptyIngressClass(ing) && !ic.isDefaultIngressClass {
-				// if we are no longer the default ingress class, we need to clean up
-				// the resources that we created for the ingress
-				if err := ic.deleteResources(ing); err != nil {
-					return err
-				}
+		} else if hasEmptyIngressClass(ing) && !event.isDefault {
+			// if we are no longer the default ingress class, we need to clean up
+			// the resources that we created for the ingress
+			if err := ic.deleteResources(ing); err != nil {
+				return err
 			}
 		}
 	}
@@ -396,28 +397,27 @@ func (ic *Controller) handleCiliumIngressClassUpdatedEvent(event ciliumIngressCl
 }
 
 func (ic *Controller) handleCiliumIngressClassDeletedEvent(event ciliumIngressClassDeletedEvent) error {
-	log.Debug("Cilium IngressClass deleted")
+	if !event.wasDefault {
+		return nil
+	}
 
-	if ic.isDefaultIngressClass {
-		// if we were the default ingress class, we need to clean up all ingresses
-		for _, k := range ic.ingressStore.ListKeys() {
-			ing, err := ic.getByKey(k)
-			if err != nil {
-				return err
-			}
-
-			if hasEmptyIngressClass(ing) {
-				// if we are no longer the default ingress class, we need to clean up
-				// the resources that we created for the ingress
-				if err := ic.deleteResources(ing); err != nil {
-					return err
-				}
-			}
+	log.Debug("Cilium IngressClass deleted, performing cleanup")
+	// if we were the default ingress class, we need to clean up all ingresses
+	for _, k := range ic.ingressStore.ListKeys() {
+		ing, err := ic.getByKey(k)
+		if err != nil {
+			return err
 		}
 
-		// disable the default ingress class behavior
-		ic.isDefaultIngressClass = false
+		if hasEmptyIngressClass(ing) {
+			// if we are no longer the default ingress class, we need to clean up
+			// the resources that we created for the ingress
+			if err := ic.deleteResources(ing); err != nil {
+				return err
+			}
+		}
 	}
+
 	return nil
 }
 
@@ -506,10 +506,10 @@ func (ic *Controller) handleEvent(event interface{}) error {
 		log.WithField(logfields.ServiceKey, ev.ingressService.Name).WithField(logfields.K8sNamespace, ev.ingressService.Namespace).Debug("Handling ingress service updated event")
 		err = ic.handleIngressServiceUpdatedEvent(ev)
 	case ciliumIngressClassUpdatedEvent:
-		log.WithField(logfields.IngressClass, ev.ingressClass.Name).Debug("Handling cilium ingress class updated event")
+		log.Debug("Handling cilium ingress class updated event")
 		err = ic.handleCiliumIngressClassUpdatedEvent(ev)
 	case ciliumIngressClassDeletedEvent:
-		log.WithField(logfields.IngressClass, ev.ingressClass.Name).Debug("Handling cilium ingress class deleted event")
+		log.Debug("Handling cilium ingress class deleted event")
 		err = ic.handleCiliumIngressClassDeletedEvent(ev)
 	default:
 		err = fmt.Errorf("received an unknown event: %t", ev)
--- a/operator/pkg/ingress/ingress_class.go
+++ b/operator/pkg/ingress/ingress_class.go
@@ -4,172 +4,196 @@
 package ingress
 
 import (
-	"fmt"
+	"context"
+	"strconv"
+	"sync/atomic"
 
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 
-	"github.com/cilium/cilium/pkg/k8s"
-	k8sClient "github.com/cilium/cilium/pkg/k8s/client"
-	"github.com/cilium/cilium/pkg/k8s/informer"
+	"github.com/cilium/cilium/pkg/k8s/resource"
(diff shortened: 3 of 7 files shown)


operator/pkg/ingress/ingress.go
```diff
@@ -157,7 +161,7 @@ func NewController(clientset k8sClient.Clientset, options ...Option) (*Controlle
 		nil,
 	)
 
-	ingressClassManager, err := newIngressClassManager(clientset, ic.queue, opts.MaxRetries)
+	ingressClassManager, err := newIngressClassManager(context.Background(), ic.queue, ingressClasses)
```
pippolo84: I suggest to plumb `legacy.ctx` here from `root.go`.

operator/pkg/ingress/ingress.go
```diff
@@ -197,12 +201,16 @@ func NewController(clientset k8sClient.Clientset, options ...Option) (*Controlle
 // Run kicks off the controlled loop
 func (ic *Controller) Run() {
 	defer ic.queue.ShutDown()
+
+	ctx := context.Background()
+
+	go ic.ingressClassManager.Run(ctx)
```
pippolo84: Same here for the context

operator/pkg/ingress/ingress_class.go
```diff
@@ -4,172 +4,201 @@
...
+
+func (i *ingressClassManager) processEvents(
+	ctx context.Context,
+	syncHandler ingressClassSyncedHandler,
+	upsertHandler, deleteHandler ingressClassModifiedHandler,
+	initialSync bool,
+) error {
+	select {
+	case event := <-i.events:
+		log.WithField("event", event).Warn("got event")
```
pippolo84: I'd probably use a debug log level here, or just remove this entirely

operator/pkg/ingress/ingress_class.go
```diff
@@ -4,172 +4,201 @@
...
+}
+
+func (i *ingressClassManager) IsDefault() bool {
+	return i.isDefaultIngressClass.Load()
+}
+
+func (i *ingressClassManager) processEvents(
+	ctx context.Context,
+	syncHandler ingressClassSyncedHandler,
+	upsertHandler, deleteHandler ingressClassModifiedHandler,
```
pippolo84: `processEvents` is always called with the same callbacks as sync/upsert/delete handlers, so what is the point of parameterizing the handlers here? Since we are always using the `ingressClassManager` methods for that, we should call them directly.

operator/pkg/ingress/ingress_class.go
```diff
@@ -4,172 +4,201 @@
...
+	initialSync bool,
+) error {
+	select {
```
pippolo84: I think a `for` loop is missing here, otherwise we are gonna handle the first event only.

operator/pkg/ingress/ingress_class.go
```diff
@@ -4,172 +4,201 @@
...
 // Run kicks off the control loop
-func (i *ingressClassManager) Run() {
-	for i.processEvent() {
+func (i *ingressClassManager) Run(ctx context.Context) error {
+	// We should always be synced here, but just in case.
+	if !i.synced.Load() {
+		err := i.sync(ctx)
+		if err != nil {
+			return err
+		}
```
pippolo84: My understanding is that, in the IngressController, we want to avoid handling Ingress events before syncing the IngressClass resource, otherwise we could end up in the scenario described in the commit message.

So, why not change the ingress controller Run so that:

- IngressClassManager Run loop starts first
- Ingress controller waits for IngressClass to be synced
- Ingress controller starts the ingressInformer

This should avoid the need for calling `sync` during IngressClassManager creation and repeat it here, too. Am I missing something?

pippolo84 (changes requested):
Thanks for the PR!
I've done a first pass (skipped the tests, for now) and left a question to better understand the reason behind the changes.

operator/pkg/ingress/ingress_class.go
learnitall: Ah nice catch, I added this in for testing and accidentally left it in.

operator/pkg/ingress/ingress.go
learnitall: Good idea :+1:

operator/pkg/ingress/ingress_class.go
learnitall: Good point!

operator/pkg/ingress/ingress.go
learnitall: To do the plumbing I think we'd need to adjust `NewController` and `Controller.Run` to take a context as a parameter. This might be confusing since only the ingress class manager would actually use the context, and otherwise it would be ignored. Do you think we should also adjust the Ingress Controller and its sub-components to use the context? Or is OK to leave it for now?

operator/pkg/ingress/ingress_class.go
learnitall: I don't think you're missing anything, I just think that the way I implemented this wasn't intuitive. In order for the Ingress controller to wait for IngressClass to be synced, we need to have the ingressClassManager's `sync` method be called. If `sync` returns an error, it's important that the error isn't dismissed, which is why I didn't want to call it in the ingress controller's `Run` method. That's why I put it in `newIngressClassManager`. I added a second potential call to `sync` in the ingressClassManger's `Run` method to make it clear that `sync` must be called before `Run`.

I made the `sync` method public and put a call to it in the ingress controller's constructor. I think this should clear things up. This way we have the following:

* Operator calls ingress.NewController
* ingressClassManager is constructed
* Inside ingress.NewController we call ingressClassManager.Sync
* If an error occurs, return it from ingress.NewController so it hits: 
https://github.com/cilium/cilium/blob/83cafbb9723fcde0093a412dc798ecb58f33067f/operator/cmd/root.go#L728-L731
* Call ingress controller's Run method which then calls the ingressClassManager's Run method (before the ingress informer is started).

WDYT?

operator/pkg/ingress/ingress_class.go
learnitall: Ah good catch. I fixed up the tests too, as I'm surprised they didn't catch this.

operator/pkg/ingress/ingress.go
pippolo84: Yep, this isn't perfect, but I still think it is worth to improve at least the ingress class manager. We can then plumb the context into the other pieces of the controller as soon as we move to the new Resource[T] framework for all the other resources. But I'd leave this for a follow up PR.

operator/pkg/ingress/ingress_class.go
pippolo84: If stopping the operator is a requirement in case `Sync` errors out, I think this is surely better. :+1: 

But if we could just log the error in case of a failed sync (not dismissing it), then we can avoid starting two separate goroutines to execute the same events loop and move all in the `Run` controller method.
Using two goroutines forces us to write thread-safe code despite the events channel and it'll be easier to add racy code in case of changes.

Also, the only possible error returned by `Sync` seems to be related to context cancellation, that is triggered only during operator shutdown.

operator/pkg/ingress/ingress_class.go
```diff
@@ -4,172 +4,181 @@
...
-	for i.processEvent() {
+func (i *ingressClassManager) IsDefault() bool {
+	return i.isDefaultIngressClass.Load()
+}
+
+func (i *ingressClassManager) processEvents(ctx context.Context, initialSync bool) error {
+	for {
+		select {
+		case event := <-i.events:
+			if event.Kind == resource.Sync {
```
mhofstetter: why handling the kind `sync` in it's own if statement instead of including in the existing switch-statement? i think it would increase the readability of this function.

operator/pkg/ingress/ingress_class.go
```diff
@@ -4,172 +4,181 @@
...
+			// learn if cilium is the default IngressClass and expose this
+			// information to the ingress controller via the IsDefault method.
+
+			switch event.Kind {
+			case resource.Upsert:
+				i.handleUpsertEvent(event, !initialSync)
+			case resource.Delete:
+				i.handleDeleteEvent(event, !initialSync)
+			default:
+				if event.Done != nil {
```
mhofstetter: i don't think that this `nil` check is necessary & common. The function should be set on all events.

operator/pkg/ingress/ingress_class.go
```diff
@@ -4,172 +4,181 @@
...
+func (i *ingressClassManager) IsDefault() bool {
+	return i.isDefaultIngressClass.Load()
+}
+
+func (i *ingressClassManager) processEvents(ctx context.Context, initialSync bool) error {
+	for {
+		select {
+		case event := <-i.events:
+			if event.Kind == resource.Sync {
+				i.handleSyncEvent(event)
```
mhofstetter: i wouldn't delegate the responsibility to mark the event as "done" into the handle functions. 

It's [mandatory](https://github.com/cilium/cilium/blob/main/pkg/k8s/resource/event.go#L28-L30) to call the function in all cases anyway. Therefore i would let the handle functions return an error and call `Done` in `processEvent`.

just introduce an error variable `err` and propagate this by the handle functions and pass this to `event.Done(err)` after the switch statement. This way it's obvious and ensured that `Done` is called in all cases.

operator/pkg/ingress/ingress_class.go
```diff
@@ -4,172 +4,181 @@
...
+				// If we are only supposed to do an initial
+				// sync, then exit when we received a sync event.
+				if initialSync {
+					return nil
+				}
+
+				break
+			}
+
+			if event.Object != nil && event.Object.GetName() != ciliumIngressClassName {
```
mhofstetter: if this check is the reason for not including the sync kind in the switch statement i would just move this "responsibility check" into the `handleUpsertEvent` & `handleDeleteEvent` functions or their case block.

operator/pkg/ingress/ingress_class.go
```diff
@@ -4,172 +4,181 @@
...
 type ciliumIngressClassDeletedEvent struct {
-	ingressClass *slim_networkingv1.IngressClass
+	wasDefault bool
 }
 
-func newIngressClassManager(clientset k8sClient.Clientset, ingressQueue workqueue.RateLimitingInterface, maxRetries int) (*ingressClassManager, error) {
+type ingressClassManager struct {
+	isDefaultIngressClass atomic.Bool
+	synced                atomic.Bool
+	ingressClasses        resource.Resource[*slim_networkingv1.IngressClass]
```
mhofstetter: looks like the field `ingressClasses` isn't used and can be removed

operator/pkg/ingress/ingress_class.go
```diff
@@ -4,172 +4,181 @@
...
+	queue                 workqueue.RateLimitingInterface
+	events                <-chan resource.Event[*slim_networkingv1.IngressClass]
```
mhofstetter: nit: i would name this field `ingressClassEvents`

operator/k8s/resource_ctors.go
```diff
@@ -57,3 +58,13 @@ func CiliumEndpointSliceResource(lc hive.Lifecycle, cs client.Clientset, opts ..
 	)
 	return resource.New[*cilium_api_v2alpha1.CiliumEndpointSlice](lc, lw, resource.WithMetric("CiliumEndpointSlice")), nil
 }
+
+func IngressClassResource(lc hive.Lifecycle, cs client.Clientset, opts ...func(*metav1.ListOptions)) (resource.Resource[*slim_networkingv1.IngressClass], error) {
+	if !cs.IsEnabled() {
```
mhofstetter, line 63: do we need an additional check that this resource is only initialized if ingress controller functionality is enabled?

mhofstetter (commented):
Thanks a lot for this @learnitall 

Some initial questions & proposals.

operator/pkg/ingress/ingress_class.go
learnitall: There must be some other bug going on then. I agree with you that this isn't common, but if I take this nil check out, I get a panic in my tests. I added a log in the default case of the switch statement to show the event struct and it looks empty:

```
level=info msg=Invoked duration="15.411µs" function="ingress.Test_ingressClassHandleEvent.func11.2 (pkg/ingress/ingress_class_test.go:426)" subsys=hive
level=info msg=Starting subsys=hive
level=info msg="Start hook executed" duration="1.702µs" function="*resource.resource[*v1.IngressClass].Start" subsys=hive
level=info msg="IngressClass resources synced" cilium-is-default=false subsys=ingress-controller
level=info msg=Stopping subsys=hive
level=warning msg="got event in default case" event="{  nil <nil>}" subsys=ingress-controller
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x248bf11]

goroutine 63 [running]:
github.com/cilium/cilium/operator/pkg/ingress.(*ingressClassManager).processEvents(0xc00033de60, {0x3115718, 0xc000571900}, 0x0)
	/home/rydrew/Documents/cilium/operator/pkg/ingress/ingress_class.go:93 +0x3d1
github.com/cilium/cilium/operator/pkg/ingress.(*ingressClassManager).Run(0x0?, {0x3115718, 0xc000571900})
	/home/rydrew/Documents/cilium/operator/pkg/ingress/ingress_class.go:126 +0x72
created by github.com/cilium/cilium/operator/pkg/ingress.Test_ingressClassHandleEvent.func11 in goroutine 53
	/home/rydrew/Documents/cilium/operator/pkg/ingress/ingress_class_test.go:446 +0x81e
FAIL	github.com/cilium/cilium/operator/pkg/ingress	0.127s
FAIL
```

operator/pkg/ingress/ingress_class.go
learnitall: I don't think I understand. The `Sync` method is not called in a separate goroutine. It's called in the main goroutine of the operator and is supposed to block the ingress controller until IngressClass resources are synced. After IngressClass resources are synced, the `Sync` method returns. When the ingress controller calls the ingress class manager's `Run` method, there will only be one goroutine processing events from the IngressClass events channel.

If we put everything in `Run`, then it makes it more difficult to have the ingress controller block until IngressClasses are synced. In order to avoid race conditions, IngressClasses must be synced before the Ingress informer is started. Putting everything in `Run` won't guarantee that, because it would be possible for the ingress controller to start processing Ingress resources before the ingress class manager has finished syncing.

operator/pkg/ingress/ingress_class.go
learnitall: That makes sense, will do.

operator/pkg/ingress/ingress.go
learnitall: Sounds like a plan!

operator/pkg/ingress/ingress_class.go
learnitall: I just realized I might have misunderstood which `Run` method we were talking about. Anyways, I tried to make some adjustments to simplify the code and address your concerns. I renamed the `Sync` method to `WaitForSync`. It uses the `WaitForCacheSync` function from `k8s.io/client-go/tools/cache` to wait for a Sync event to be received. Now the `Run` method in the ingress controller looks like this:

```go
func (ic *Controller) Run(ctx context.Context) error {
	defer ic.queue.ShutDown()

	go ic.ingressClassManager.Run(ctx)

	// This should only return an error if the context is cancelled.
	if err := ic.ingressClassManager.WaitForSync(ctx); err != nil {
		return err
	}

	go ic.ingressInformer.Run(wait.NeverStop)
	if !cache.WaitForCacheSync(wait.NeverStop, ic.ingressInformer.HasSynced) {
		return fmt.Errorf("unable to wait for Ingress cache sync")
	}

	go ic.serviceManager.Run()
	go ic.secretManager.Run()

	for ic.processEvent() {
	}

	return nil
}
```

operator/pkg/ingress/ingress_class.go
learnitall: Done :+1:

operator/pkg/ingress/ingress_class.go
learnitall: Good catch!

operator/pkg/ingress/ingress_class.go
learnitall: Ah good idea, will do!

operator/k8s/resource_ctors.go
learnitall: We could, but I don't know how helpful it would be? If ingress controller functionality isn't enabled, then the resource won't be marked as needed. I think that will prevent any goroutines from starting in the case that ingress functionality is disabled.

learnitall:
@pippolo84 @mhofstetter I made some changes to address your feedback. Let me know what you think!

operator/pkg/ingress/ingress_class.go
```diff
@@ -4,172 +4,199 @@
...
-		i.queue.Forget(event)
-	}
-	return true
+// Run kicks off the the main control loop for the ingressClassManager.
+// The ingressClassManager will start processing IngressClass events and signaling updates by
+// sending events on the queue that was provided during construction.
+func (i *ingressClassManager) Run(ctx context.Context) error {
+	log.Debug("Starting ingressClassManager")
+
+	return i.processEvents(ctx)
```
pippolo84: Nit: you can remove `processEvents` and move its body here, since it is not called elsewhere.

operator/pkg/ingress/ingress_class_test.go
```diff
@@ -4,96 +4,558 @@
...
+	testCancel()
+	i.synced.Store(false)
+
+	err := i.WaitForSync(testCtx)
+	if err == nil {
+		t.Fatalf("unexpected nil error while doing forced block sync: %s", err)
+	}
+
+	if err != nil && !errors.Is(err, context.Canceled) {
+		t.Fatalf("unexpected error whil doing forced block sync, wanted context.Canceled: %s", err)
```
pippolo84: Typo: `whil` -> `while`

operator/pkg/ingress/ingress_class_test.go
```diff
@@ -4,96 +4,558 @@
...
+	}
+
+	h.Stop(ctx)
```
pippolo84: nit: I suggest to check the error returned by `Stop`

operator/pkg/ingress/ingress_class_test.go
```diff
@@ -4,96 +4,558 @@
...
+			Kind:   resource.Upsert,
+			Object: nonCiliumIngressClass,
+			Done:   func(_ error) {},
+		},
+	)
+	if err != nil {
+		t.Fatalf("unexpected error handling upsert event for non-cilium IngressClass: %s", err)
+	}
+
+	assert.Equal(t, 0, queue.Len())
```
pippolo84, line 129: nit: `hive.Stop` is missing here

operator/pkg/ingress/ingress_class_test.go
```diff
@@ -4,96 +4,558 @@
...
+		}
+
+		if c.doDelete {
+			ingressClassesClient.Create(ctx, c.class, v1.CreateOptions{})
+		}
+
+		handleEventExpectation(t, c.expectedRestoreEvent, queue)
+		assert.Equal(t, c.defaultAtStart, i.IsDefault())
+
+		h.Stop(ctx)
```
pippolo84: nit: check the error returned by `Stop`

operator/pkg/ingress/ingress_class.go
pippolo84: I think this is happening after `hive.Stop` is called: the `ingressClassEvents` channel is closed and the `select` in `(i *ingressClassManager) processEvents` might unblock on the first case with a bogus event. This is due to the fact that, even if the `ctx` is cancelled and the other select case can run too, `select` chooses pseudo-randomly which path to take.

With this:

```diff
@@ -128,9 +128,12 @@ func (i *ingressClassManager) processEvents(ctx context.Context) error {
 
        for {
                select {
-               case event := <-i.ingressClassEvents:
-                       err = nil
+               case event, ok := <-i.ingressClassEvents:
+                       if !ok {
+                               return nil
+                       }
 
+                       err = nil
                        switch event.Kind {
                        case resource.Sync:
                                err = i.handleSyncEvent()
@@ -140,9 +143,7 @@ func (i *ingressClassManager) processEvents(ctx context.Context) error {
                                err = i.handleDeleteEvent(event)
                        }
 
-                       if event.Done != nil {
-                               event.Done(err)
-                       }
+                       event.Done(err)
                case <-ctx.Done():
                        return ctx.Err()
                }
```

the tests run without any panic

pippolo84 (commented):
Great! I like the way you changed the `Run` method in the ingress controller, I think this is effective and simpler!

I've left some nits about the test, but nothing blocking. Only thing left to address is the check on the `ingressClassEvents` channel.

operator/pkg/ingress/ingress_class.go
mhofstetter: :point_up: yep, we must take the closed channel into account

operator/k8s/resource_ctors.go
mhofstetter: leaving it as is is fine for me :+1: just wanted to mention it

operator/pkg/ingress/ingress_class.go
```diff
@@ -4,172 +4,199 @@
...
+// handleUpsertEvent handles an Upsert event for an IngressClass resource.
+// If the IngressClass has a name that is not the Cilium IngressClass name, it will be ignored.
+// If an error is returned, then the IngressClass contained in the event should be retried at a later
+// point in time.
+func (i *ingressClassManager) handleUpsertEvent(event resource.Event[*slim_networkingv1.IngressClass]) error {
+	log.WithField(logfields.IngressClass, event.Object).Warn("Handling IngressClass Upsert event")
 
-	if event.ingressClass.GetName() == ciliumIngressClassName {
-		i.ingressQueue.Add(ciliumIngressClassDeletedEvent(event))
+	if event.Object == nil || (event.Object != nil && !isIngressClassCilium(event.Object)) {
```
mhofstetter: no need to check `event.Object` being not nil in the second part of the condition.

-> `if event.Object == nil || !isIngressClassCilium(event.Object) {`

operator/pkg/ingress/ingress_class.go
```diff
@@ -4,172 +4,199 @@
...
-	if ic.GetName() == ciliumIngressClassName {
-		i.ingressQueue.Add(ciliumIngressClassUpdatedEvent{ingressClass: ic})
+// handleDeleteEvent handles a Delete event for an IngressClass resource.
+// If the IngressClass has a name that is not the Cilium IngressClass name, it will be ignored.
+// If an error is returned, then the IngressClass contained in the event should be retried at a later
+// point in time.
+func (i *ingressClassManager) handleDeleteEvent(event resource.Event[*slim_networkingv1.IngressClass]) error {
+	log.WithField(logfields.IngressClass, event.Object).Debug("Handling IngressClass Delete event")
+
+	if event.Object == nil || (event.Object != nil && !isIngressClassCilium(event.Object)) {
```
mhofstetter: same here: no need to check `event.Object` being not nil in the second part of the condition.

-> `if event.Object == nil || !isIngressClassCilium(event.Object) {`

operator/pkg/ingress/ingress_class_test.go
```diff
@@ -4,96 +4,558 @@
...
+	if err := i.WaitForSync(ctx); err != nil {
+		t.Fatalf("unexpected error while doing second sync: %s", err)
+	}
+
+	// Try exiting a sync using context
+	testCtx, testCancel := context.WithCancel(ctx)
+	testCancel()
+	i.synced.Store(false)
+
+	err := i.WaitForSync(testCtx)
```
mhofstetter: nit: combining the related if statements

```
if err := i.WaitForSync(testCtx); err == nil {
...
} else if !errors.Is(err, context.Canceled) {
...
```

operator/pkg/ingress/ingress_class.go
learnitall: Ah right ok, that makes sense. I'll make the change!

operator/pkg/ingress/ingress_class.go
learnitall: Good point!

operator/pkg/ingress/ingress_class_test.go
learnitall: Nice catch!

operator/pkg/ingress/ingress_class.go
learnitall: Ah yeah good point.

operator/pkg/ingress/ingress_class_test.go
learnitall: Oh sweet, I didn't know that you could reference variables defined in an if statement in other branches.

pippolo84 (approved):
Great, thanks! :100:

mhofstetter (approved):
LGTM, thanks again!

PR #28665, opened 2023-10-18
bpf: let set_identity_mark() also set MARK_MAGIC_IDENTITY

julianwiedmann:
Setting MARK_MAGIC_IDENTITY is really a property of the identity encoding, hide this detail from the users.

No functional change.


YutaroHayakawa (approved):
LGTM

PR #28666, opened 2023-10-18
Best effort for xdp

### LINKED ISSUE #24768: XDP: Failing to attach XDP program to a tagged VLAN interface

poblahblahblah:

- [?] All code is covered by unit and/or runtime tests where feasible.
      GitHub issue.
- [?] Are you a user of Cilium? Please add yourself to the [Users doc](https://github.com/cilium/cilium/blob/main/USERS.md)

This allows a user to specify a new mode, `best-effort` when enabling XDP. We need this ability since we are running physical hosts where the primary NICs do support XDP, but have additional `vlan` devices that _do not_ have XDP support in the driver.

More background can be found in #24768, but to summarize we have 1 physical device (ens785np0) per host and due to network design we need an additional `vlan` type interface (`vlan3`). Regrettably the `vlan` driver in the kernel does not support XDP. When enabling XDP without this change Cilium enters a crashloop with this error:

```
level=fatal msg="Failed to compile XDP program" error="program cil_xdp_entry: attaching XDP program to interface vlan3: operation not supported" subsys=datapath-loader
```

With this change applied and by setting `loadBalancer.acceleration` to `best-effort` in our values.yaml file we get the following log output with XDP enabled only on devices that support it:

```
level=info msg="attaching XDP program to interface vlan3 failed. Ignoring" subsys=datapath-loader
```

```
$ sudo bpftool net
xdp:
ens785np0(2) driver id 28652

tc:
ens785np0(2) clsact/ingress cil_from_netdev-ens785np0 id 28699
ens785np0(2) clsact/egress cil_to_netdev-ens785np0 id 28721
vlan3(4) clsact/ingress cil_from_netdev-vlan3 id 28731
vlan3(4) clsact/egress cil_to_netdev-vlan3 id 28739
cilium_net(5) clsact/ingress cil_to_host-cilium_net id 28692
cilium_host(6) clsact/ingress cil_to_host-cilium_host id 28674
cilium_host(6) clsact/egress cil_from_host-cilium_host id 28686
cilium_vxlan(7) clsact/ingress cil_from_overlay-cilium_vxlan id 28663
cilium_vxlan(7) clsact/egress cil_to_overlay-cilium_vxlan id 28664
lxc_health(11) clsact/ingress cil_from_container-lxc_health id 28708

flow_dissector:
```

I initially created https://github.com/cilium/cilium/pull/25870 and it was indicated in [this comment thread](https://github.com/cilium/cilium/pull/25870#discussion_r1313247066) that swallowing these errors if set to "best-effort" would be an acceptable solution.

Any and all feedback would be very welcome!

Fixes: #24768

```release-note
Adds "best-effort" mode for XDP to skip interfaces without driver support
```

### DIFF
--- a/Documentation/network/kubernetes/kubeproxy-free.rst
+++ b/Documentation/network/kubernetes/kubeproxy-free.rst
@@ -676,8 +676,10 @@ In case of a multi-device environment, where Cilium's device auto-detection sele
 more than a single device to expose NodePort or a user specifies multiple devices
 with ``devices``, the XDP acceleration is enabled on all devices. This means that
 each underlying device's driver must have native XDP support on all Cilium managed
-nodes. In addition, for performance reasons we recommend kernel >= 5.5 for
-the multi-device XDP acceleration.
+nodes. If you have an environment where some devices support XDP but others do not
+you can have XDP enabled on the supported devices by setting
+``loadBalancer.acceleration`` to ``best-effort``. In addition, for performance
+reasons we recommend kernel >= 5.5 for the multi-device XDP acceleration.
 
 A list of drivers supporting XDP can be found in :ref:`the XDP documentation<xdp_drivers>`.
 
--- a/install/kubernetes/cilium/values.yaml
+++ b/install/kubernetes/cilium/values.yaml
@@ -1837,8 +1837,11 @@ loadBalancer:
   # mode: snat
 
   # -- acceleration is the option to accelerate service handling via XDP
-  # e.g. native, disabled
-  # acceleration: disabled
+  # Applicable values can be: disabled (do not use XDP), native (XDP BPF
+  # program is run directly out of the networking driver's early receive
+  # path), or best-effort (use native mode XDP acceleration on devices
+  # that support it).
+  acceleration: disabled
 
   # -- dsrDispatch configures whether IP option or IPIP encapsulation is
   # used to pass a service IP and port to remote backend
--- a/install/kubernetes/cilium/values.yaml.tmpl
+++ b/install/kubernetes/cilium/values.yaml.tmpl
@@ -1834,8 +1834,11 @@ loadBalancer:
   # mode: snat
 
   # -- acceleration is the option to accelerate service handling via XDP
-  # e.g. native, disabled
-  # acceleration: disabled
+  # Applicable values can be: disabled (do not use XDP), native (XDP BPF
+  # program is run directly out of the networking driver's early receive
+  # path), or best-effort (use native mode XDP acceleration on devices
+  # that support it).
+  acceleration: disabled
 
   # -- dsrDispatch configures whether IP option or IPIP encapsulation is
   # used to pass a service IP and port to remote backend
--- a/pkg/datapath/loader/link.go
+++ b/pkg/datapath/loader/link.go
@@ -11,7 +11,7 @@ import (
 
 func SetXDPMode(mode string) error {
 	switch mode {
-	case option.XDPModeNative:
+	case option.XDPModeNative, option.XDPModeBestEffort:
 		if option.Config.XDPMode == option.XDPModeLinkNone ||
 			option.Config.XDPMode == option.XDPModeLinkDriver {
 			option.Config.XDPMode = option.XDPModeLinkDriver
--- a/pkg/datapath/loader/netlink.go
+++ b/pkg/datapath/loader/netlink.go
@@ -226,7 +226,7 @@ func attachTCProgram(link netlink.Link, prog *ebpf.Program, progName string, qdi
 	}
 
 	if err := netlink.FilterReplace(filter); err != nil {
-		return fmt.Errorf("replacing tc filter: %w", err)
+		return fmt.Errorf("replacing tc filter for interface %s: %w", link.Attrs().Name, err)
 	}
 
 	return nil
--- a/pkg/datapath/loader/xdp.go
+++ b/pkg/datapath/loader/xdp.go
@@ -27,7 +27,7 @@ import (
 
 func xdpModeToFlag(xdpMode string) link.XDPAttachFlags {
 	switch xdpMode {
-	case option.XDPModeNative, option.XDPModeLinkDriver:
+	case option.XDPModeNative, option.XDPModeLinkDriver, option.XDPModeBestEffort:
 		return link.XDPDriverMode
 	case option.XDPModeGeneric, option.XDPModeLinkGeneric:
 		return link.XDPGenericMode
--- a/pkg/option/config.go
+++ b/pkg/option/config.go
@@ -260,7 +260,7 @@ const (
 	NodePortAlg = "node-port-algorithm"
 
 	// NodePortAcceleration indicates whether NodePort should be accelerated
-	// via XDP ("none", "generic" or "native")
+	// via XDP ("none", "generic", "native", or "best-effort")
 	NodePortAcceleration = "node-port-acceleration"
 
 	// Alias to NodePortMode
@@ -874,6 +874,9 @@ const (
 	// XDPModeNative for loading progs with XDPModeLinkDriver
 	XDPModeNative = "native"
 
+	// XDPModeBestEffort for loading progs with XDPModeLinkDriver
+	XDPModeBestEffort = "best-effort"
+
 	// XDPModeGeneric for loading progs with XDPModeLinkGeneric
 	XDPModeGeneric = "testing-only"
 
@@ -1344,6 +1347,9 @@ const (
 
 	// NodePortAccelerationNative means we accelerate NodePort via native XDP in the driver (preferred)
 	NodePortAccelerationNative = XDPModeNative
+
+	// NodePortAccelerationBestEffort means we accelerate NodePort via native XDP in the driver (preferred), but will skip devices without driver support
+	NodePortAccelerationBestEffort = XDPModeBestEffort
 
 	// KubeProxyReplacementPartial specifies to enable only selected kube-proxy
 	// replacement features (might panic)
@@ -2047,7 +2053,7 @@ type DaemonConfig struct {
 	MaglevHashSeed string
 
 	// NodePortAcceleration indicates whether NodePort should be accelerated
-	// via XDP ("none", "generic" or "native")
+	// via XDP ("none", "generic", "native", or "best-effort")
 	NodePortAcceleration string
 
 	// NodePortBindProtection rejects bind requests to NodePort service ports
--- a/Documentation/helm-values.rst
+++ b/Documentation/helm-values.rst
@@ -2191,7 +2191,11 @@
    * - :spelling:ignore:`loadBalancer`
      - Configure service load balancing
      - object
-     - ``{"l7":{"algorithm":"round_robin","backend":"disabled","ports":[]}}``
+     - ``{"acceleration":"disabled","l7":{"algorithm":"round_robin","backend":"disabled","ports":[]}}``
+   * - :spelling:ignore:`loadBalancer.acceleration`
+     - acceleration is the option to accelerate service handling via XDP Applicable values can be: disabled (do not use XDP), native (XDP BPF program is run directly out of the networking driver's early receive path), or best-effort (use native mode XDP acceleration on devices that support it).
+     - string
+     - ``"disabled"``
    * - :spelling:ignore:`loadBalancer.l7`
      - L7 LoadBalancer
      - object
--- a/api/v1/models/kube_proxy_replacement.go
+++ b/api/v1/models/kube_proxy_replacement.go
@@ -1134,7 +1134,7 @@ func (m *KubeProxyReplacementFeaturesNat46X64Service) UnmarshalBinary(b []byte) 
 type KubeProxyReplacementFeaturesNodePort struct {
 
 	// acceleration
-	// Enum: [None Native Generic]
+	// Enum: [None Native Generic Best-Effort]
 	Acceleration string `json:"acceleration,omitempty"`
 
 	// algorithm
@@ -1184,7 +1184,7 @@ var kubeProxyReplacementFeaturesNodePortTypeAccelerationPropEnum []interface{}
 
 func init() {
 	var res []string
-	if err := json.Unmarshal([]byte(`["None","Native","Generic"]`), &res); err != nil {
+	if err := json.Unmarshal([]byte(`["None","Native","Generic","Best-Effort"]`), &res); err != nil {
 		panic(err)
 	}
 	for _, v := range res {
@@ -1202,6 +1202,9 @@ const (
 
 	// KubeProxyReplacementFeaturesNodePortAccelerationGeneric captures enum value "Generic"
 	KubeProxyReplacementFeaturesNodePortAccelerationGeneric string = "Generic"
+
+	// KubeProxyReplacementFeaturesNodePortAccelerationBestDashEffort captures enum value "Best-Effort"
+	KubeProxyReplacementFeaturesNodePortAccelerationBestDashEffort string = "Best-Effort"
 )
 
 // prop value enum
--- a/api/v1/openapi.yaml
+++ b/api/v1/openapi.yaml
@@ -2387,6 +2387,7 @@ definitions:
                 - None
                 - Native
                 - Generic
+                - Best-Effort
               portMin:
                 type: integer
               portMax:
--- a/api/v1/server/embedded_spec.go
+++ b/api/v1/server/embedded_spec.go
@@ -3699,7 +3699,8 @@ func init() {
                   "enum": [
                     "None",
                     "Native",
-                    "Generic"
+                    "Generic",
+                    "Best-Effort"
                   ]
                 },
                 "algorithm": {
@@ -9382,7 +9383,8 @@ func init() {
                   "enum": [
                     "None",
                     "Native",
-                    "Generic"
+                    "Generic",
+                    "Best-Effort"
                   ]
                 },
                 "algorithm": {
@@ -9564,7 +9566,8 @@ func init() {
               "enum": [
                 "None",
                 "Native",
-                "Generic"
+                "Generic",
+                "Best-Effort"
               ]
             },
             "algorithm": {
@@ -9733,7 +9736,8 @@ func init() {
           "enum": [
             "None",
             "Native",
-            "Generic"
+            "Generic",
+            "Best-Effort"
           ]
         },
         "algorithm": {
--- a/daemon/cmd/kube_proxy_replacement.go
+++ b/daemon/cmd/kube_proxy_replacement.go
@@ -170,7 +170,8 @@ func initKubeProxyReplacementOptions() error {
 
 		if option.Config.NodePortAcceleration != option.NodePortAccelerationDisabled &&
 			option.Config.NodePortAcceleration != option.NodePortAccelerationGeneric &&
-			option.Config.NodePortAcceleration != option.NodePortAccelerationNative {
+			option.Config.NodePortAcceleration != option.NodePortAccelerationNative &&
+			option.Config.NodePortAcceleration != option.NodePortAccelerationBestEffort {
 			return fmt.Errorf("Invalid value for --%s: %s", option.NodePortAcceleration, option.Config.NodePortAcceleration)
 		}
 
--- a/install/kubernetes/cilium/README.md
+++ b/install/kubernetes/cilium/README.md
@@ -597,7 +597,8 @@ contributors across the globe, there is almost always someone available to help.
 | l7Proxy | bool | `true` | Enable Layer 7 network policy. |
 | livenessProbe.failureThreshold | int | `10` | failure threshold of liveness probe |
 | livenessProbe.periodSeconds | int | `30` | interval between checks of the liveness probe |
-| loadBalancer | object | `{"l7":{"algorithm":"round_robin","backend":"disabled","ports":[]}}` | Configure service load balancing |
+| loadBalancer | object | `{"acceleration":"disabled","l7":{"algorithm":"round_robin","backend":"disabled","ports":[]}}` | Configure service load balancing |
+| loadBalancer.acceleration | string | `"disabled"` | acceleration is the option to accelerate service handling via XDP Applicable values can be: disabled (do not use XDP), native (XDP BPF program is run directly out of the networking driver's early receive path), or best-effort (use native mode XDP acceleration on devices that support it). |
 | loadBalancer.l7 | object | `{"algorithm":"round_robin","backend":"disabled","ports":[]}` | L7 LoadBalancer |
 | loadBalancer.l7.algorithm | string | `"round_robin"` | Default LB algorithm The default LB algorithm to be used for services, which can be overridden by the service annotation (e.g. service.cilium.io/lb-l7-algorithm) Applicable values: round_robin, least_request, random |
 | loadBalancer.l7.backend | string | `"disabled"` | Enable L7 service load balancing via envoy proxy. The request to a k8s service, which has specific annotation e.g. service.cilium.io/lb-l7, will be forwarded to the local backend proxy to be load balanced to the service endpoints. Please refer to docs for supported annotations for more configuration.  Applicable values:   - envoy: Enable L7 load balancing via envoy proxy. This will automatically set enable-envoy-config as well.   - disabled: Disable L7 load balancing by way of service annotation. |
--- a/pkg/datapath/loader/base.go
+++ b/pkg/datapath/loader/base.go
@@ -248,8 +248,13 @@ func (l *Loader) reinitializeXDPLocked(ctx context.Context, extraCArgs []string)
 		if dev == wgTypes.IfaceName {
 			continue
 		}
+
 		if err := compileAndLoadXDPProg(ctx, dev, option.Config.XDPMode, extraCArgs); err != nil {
-			return err
+			if option.Config.NodePortAcceleration == option.XDPModeBestEffort {
+				log.WithError(err).WithField(logfields.Device, dev).Info("Failed to attach XDP program, ignoring due to best-effort mode")
+			} else {
+				return fmt.Errorf("attaching XDP program to interface %s: %w", dev, err)
+			}
 		}
 	}
 	return nil


pkg/option/config.go
```diff
@@ -881,6 +881,9 @@ const (
 	// XDPModeNative for loading progs with XDPModeLinkDriver
 	XDPModeNative = "native"
 
+	// XDPModeBestEffort for loading progs with XDPModeLinkDriver
+	XDPModeBestEffort = "best-effort"
```
mhofstetter, line 878: The additional value `best-effort` should be documented in the helm chart too. Maybe it's also worth to add the already missing value `generic`.

https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/values.yaml.tmpl#L1774-L1776

Please follow the instructions in the Cilium docs: https://docs.cilium.io/en/latest/contributing/development/dev_setup/#making-changes-to-the-helm-chart

mhofstetter, line 878: The additional value `best-effort` needs to be reflected in the Cilium CLI model too. It's used by the command `cilium status --verbose`.

https://github.com/cilium/cilium/blob/main/api/v1/models/kube_proxy_replacement.go#L1137-L1138
&
https://github.com/cilium/cilium/blob/main/api/v1/models/kube_proxy_replacement.go#L1195-L1205

mhofstetter (changes requested):
Thanks for the contribution @poblahblahblah

Added some config-related comments.

@borkmann Adding you as reviewer because it seems that you already have the context from the issue and the previous PR

pkg/datapath/loader/link.go
```diff
@@ -27,6 +27,14 @@ func SetXDPMode(mode string) error {
 			return fmt.Errorf("XDP Mode conflict: current mode is %s, trying to set conflicting %s",
 				option.Config.XDPMode, option.XDPModeLinkGeneric)
 		}
+	case option.XDPModeBestEffort:
```
rgo3: Could this case be combined with `case option.XDPModeNative` to avoid the code duplication?

pkg/datapath/loader/xdp.go
```diff
@@ -21,6 +21,8 @@ func xdpModeToFlag(xdpMode string) uint32 {
 	switch xdpMode {
 	case option.XDPModeNative:
 		return nl.XDP_FLAGS_DRV_MODE
+	case option.XDPModeBestEffort:
```
rgo3: nit: we could combine some of cases on the switch here as well as a drive-by.

pkg/datapath/loader/netlink.go
```diff
@@ -170,7 +170,11 @@ func attachProgram(link netlink.Link, prog *ebpf.Program, progName string, qdisc
 		// Omitting XDP_FLAGS_UPDATE_IF_NOEXIST equals running 'ip' with -force,
 		// and will clobber any existing XDP attachment to the interface.
 		if err := netlink.LinkSetXdpFdWithFlags(link, prog.FD(), int(xdpFlags)); err != nil {
-			return fmt.Errorf("attaching XDP program to interface %s: %w", link.Attrs().Name, err)
+			if option.Config.NodePortAcceleration == option.XDPModeBestEffort {
+				log.Infof("attaching XDP program to interface %s failed. Ignoring", link.Attrs().Name)
```
rgo3: Should we include the error in the log message to display a reason for the failure for debugging purposes?

rgo3 (changes requested):
Thanks, just a couple nits.

pkg/option/config.go
poblahblahblah: No matter what I did I could not figure out how to get those files to be updated. Here's what I ran:

```
make generate-api
```

However, having these updated does not appear to be necessary. Here's the output from `cilium status --verbose`:

```
KubeProxyReplacement Details:
  Status:                 Strict
  Socket LB:              Enabled
  Socket LB Tracing:      Enabled
  Socket LB Coverage:     Full
  Devices:                ens785np0 10.121.160.29 (Direct Routing)
  Mode:                   SNAT
  Backend Selection:      Random
  Session Affinity:       Enabled
  Graceful Termination:   Enabled
  NAT46/64 Support:       Disabled
  XDP Acceleration:       Best-Effort
  Services:
  - ClusterIP:      Enabled
  - NodePort:       Enabled (Range: 30000-32767)
  - LoadBalancer:   Enabled
  - externalIPs:    Enabled
  - HostPort:       Enabled
```

poblahblahblah:
I'll squash commits into a single commit once this is looking like it will be ready for merge.

pkg/option/config.go
mhofstetter: You should be able to update the [OpenAPI model](https://github.com/cilium/cilium/blob/main/api/v1/openapi.yaml#L2335) and re-generate with `make generate-api`.

At least for the sake of completeness

pkg/datapath/loader/netlink.go
```diff
@@ -170,7 +170,11 @@ func attachProgram(link netlink.Link, prog *ebpf.Program, progName string, qdisc
 		// Omitting XDP_FLAGS_UPDATE_IF_NOEXIST equals running 'ip' with -force,
 		// and will clobber any existing XDP attachment to the interface.
 		if err := netlink.LinkSetXdpFdWithFlags(link, prog.FD(), int(xdpFlags)); err != nil {
-			return fmt.Errorf("attaching XDP program to interface %s: %w", link.Attrs().Name, err)
+			if option.Config.NodePortAcceleration == option.XDPModeBestEffort {
```
squeed: I don't like swallowing the error here; this function is not specific to NodePorts. Rather, this config check should be specific to the code path that attaches the NodePort XDP program

squeed (changes requested):

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1772,7 +1772,7 @@ loadBalancer:
   # mode: snat
 
   # -- acceleration is the option to accelerate service handling via XDP
-  # e.g. native, disabled
+  # e.g. native, disabled, generic, best-effort
```
squeed: As @mhofstetter, it would be nice to have a word or two explaining what these different options are (or, a link to the documentation). Helm chart docs, sadly, don't well with paragraph breaks / formatting because of some silly transformations we do :crying_cat_face:. So, if documenting this gets hard, it should be a section in the XDP documentation.

Sorry, I know this is a bit more work than you bargained for.

squeed:
A few small changes, but looks basically good. Thanks!

pkg/datapath/loader/netlink.go
rgo3: While not entirely specific to NodePort code path it could live in `reinitializeXDPLocked()`. I think the original PR #25870 placed its logic there as well. Plus there is already logic to skip certain devices in that function.

install/kubernetes/cilium/values.yaml.tmpl
poblahblahblah: No worries - I will take care of this either way

pkg/datapath/loader/netlink.go
poblahblahblah: Yeah, this section felt like a pretty good fit when I was going through everything. Happy to adjust if there are strong opinions about it.

pkg/option/config.go
poblahblahblah: That is very helpful, thank you @mhofstetter

install/kubernetes/cilium/values.yaml.tmpl
poblahblahblah: Also updated the regular documentation (Documentation/bpf/progtypes.rst) in commit 305dfa694c82987ab69733e9b1e199f672b44350

mhofstetter (approved):
LGTM for the agent bits.

pkg/datapath/loader/netlink.go
squeed: I'll let @cilium/loader make the final decision.

squeed (approved):

pkg/datapath/loader/netlink.go
```diff
@@ -170,7 +170,11 @@ func attachProgram(link netlink.Link, prog *ebpf.Program, progName string, qdisc
...
+				log.Infof("attaching XDP program to interface %s failed. Ignoring. %w", link.Attrs().Name, err)
```
joamaki: nit: `log.WithError(err).WithField(logfields.Device, link.Attrs().Name).Info("Attaching ... ")`. This way the device name and error are available in the structured logging as separate fields.

pkg/datapath/loader/netlink.go
joamaki: A good rule of thumb to apply is to avoid use of `option.Config` at deeper layers since that makes testing fairly annoying since now in tests you'll need to mutate the global `option.Config` and then restore it. 

One possibility here is to just pass the `option.Config.XDPMode` that `reinitializeXDPLocked` grabs all the way down to `attachProgram` and handle the "BestEffort" option there and then convert to XDP flags. This way this low-level function isn't coupled to the `option.Config` global. E.g. just move the "xdpModeToFlags" from `replaceDatapath` to `attachProgram`.

Also talking about tests, could you look into adding a test case for this into `netlink_test.go`? `TestAttachProgram` already has a "XDP" sub-test which would be easy to copy + adapt.

pkg/datapath/loader/xdp.go
```diff
@@ -19,13 +19,9 @@ import (
 
 func xdpModeToFlag(xdpMode string) uint32 {
 	switch xdpMode {
-	case option.XDPModeNative:
+	case option.XDPModeNative, option.XDPModeLinkDriver, option.XDPModeBestEffort:
```
joamaki, line 30: Hmm, now `XDPModeBestEffort` in a way implies `XDPModeLinkDriver`. Perhaps it'd be better to reflect that in the name? `XDPModeLinkDriverBestEffort`or `XDPModeTryLinkDriver` come to mind, but they're a bit of a mouthful.

Documentation/bpf/progtypes.rst
```diff
@@ -323,6 +323,13 @@ talked about XDP this mode is typically implied.
   production environment either the native or offloaded mode is better
   suited and the recommended way to run XDP.
 
+..
+
+* **Best Effort**
```
joamaki: This is generic, non-cilium specific, documentation about BPF and thus this doesn't belong here.

pkg/datapath/loader/netlink.go
joamaki: Also I wonder if skipping is the right thing to do here. We can also try the generic mode before skipping entirely. OTOH, quite likely we'd prefer the tc filter program as a fallback to generic XDP... Perhaps someone from sig-loader can chime in on whether it'd make sense or not.

joamaki (changes requested):

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1772,7 +1772,17 @@ loadBalancer:
...
+  # disabled - do not use XDP
+  # native - XDP BPF program is run directly out of the networking driver’s
+  #   early receive path.
+  # generic - For drivers not implementing native or offloaded XDP yet, the
+  #   kernel provides an option for generic XDP which does not require any
+  #   driver changes since run at a much later point out of the networking
+  #   stack. This setting is primarily targeted at developers who want to
+  #   write and test programs against the kernel’s XDP API, and will not
+  #   operate at the performance rate of the native or offloaded modes.
+  # best-effort - Use native where possible, but ignore devices without XDP
```
joamaki: "ignore devices" can be a bit misleading. Cilium will still handle traffic to/from this device using the program attached to a TC filter. Perhaps something like "Use native mode XDP acceleration on devices that support it".

pkg/datapath/loader/netlink.go
poblahblahblah: @joamaki From the documentation it doesn't sound like Generic is recommended outside of testing scenarios.

Taken from https://docs.cilium.io/en/stable/bpf/progtypes/

```
For drivers not implementing native or offloaded XDP yet, the kernel provides an option for generic XDP which does not require any driver changes since run at a much later point out of the networking stack. This setting is primarily targeted at developers who want to write and test programs against the kernel’s XDP API, and will not operate at the performance rate of the native or offloaded modes. For XDP usage in a production environment either the native or offloaded mode is better suited and the recommended way to run XDP.
```

As this PR is currently written it will fall back to tc if XDP "native" fails.

pkg/datapath/loader/xdp.go
poblahblahblah: Happy to update this to whatever consensus the Cilium project can reach!

Documentation/bpf/progtypes.rst
poblahblahblah: Will remove.

ti-mo:
Note there's also https://github.com/cilium/cilium/pull/28308 in-flight that needs to work with this. Ideally the xdp_link stuff should go in first.

pkg/datapath/loader/xdp.go
qmonnet: For what it's worth I like `XDPModeTryLinkDriver`, it's not that long.

poblahblahblah: Does anyone have a strong opinion on this?

Keep in mind that BestEffort is what was originally suggested by @borkmann so this wasn't just me pulling something out of thin air!

pkg/datapath/loader/netlink.go
poblahblahblah: Does anyone else want to weigh in on this?

poblahblahblah:
@joamaki 

Responding to a comment from above that I can't respond to inline

```
A good rule of thumb to apply is to avoid use of option.Config at deeper layers since that makes testing fairly annoying since now in tests you'll need to mutate the global option.Config and then restore it.

One possibility here is to just pass the option.Config.XDPMode that reinitializeXDPLocked grabs all the way down to attachProgram and handle the "BestEffort" option there and then convert to XDP flags. This way this low-level function isn't coupled to the option.Config global. E.g. just move the "xdpModeToFlags" from replaceDatapath to attachProgram.

Also talking about tests, could you look into adding a test case for this into netlink_test.go? TestAttachProgram already has a "XDP" sub-test which would be easy to copy + adapt.
```

I went with using option.Config here because option.Config was already being called when defining the bpf filter on line 194. Is this a change you still want to see made?

Happy to attempt a test regardless, but do let me know your thoughts on the above question

pkg/datapath/loader/netlink.go
borkmann: > Also I wonder if skipping is the right thing to do here. We can also try the generic mode before skipping entirely. OTOH, quite likely we'd prefer the tc filter program as a fallback to generic XDP... Perhaps someone from sig-loader can chime in on whether it'd make sense or not.

Generic mode should not be used in production, it will bypass GRO engine. Fallback to tc is significantly better.

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1772,7 +1772,17 @@ loadBalancer:
...
-  # e.g. native, disabled
+  # disabled - do not use XDP
+  # native - XDP BPF program is run directly out of the networking driver’s
+  #   early receive path.
+  # generic - For drivers not implementing native or offloaded XDP yet, the
+  #   kernel provides an option for generic XDP which does not require any
+  #   driver changes since run at a much later point out of the networking
+  #   stack. This setting is primarily targeted at developers who want to
+  #   write and test programs against the kernel’s XDP API, and will not
+  #   operate at the performance rate of the native or offloaded modes.
```
borkmann: small nit: Lets leave out 'generic' as imho nobody should use it. It also won't work with 'generic' the undocumented string would be 'testing-only'.

Documentation/bpf/progtypes.rst
```diff
@@ -323,6 +323,13 @@ talked about XDP this mode is typically implied.
...
+
+  Attempt to use Native XDP on all devices, but ignore devices without XDP
+  driver support in the kernel.
```
borkmann: small nit: Please remove this doc extension in progtypes, because this refers to kernel types, not Cilium.

Please extend https://github.com/cilium/cilium/blob/main/Documentation/network/kubernetes/kubeproxy-free.rst#loadbalancer--nodeport-xdp-acceleration instead given there the option 'best-effort' should be mentioned.

install/kubernetes/cilium/values.yaml
```diff
@@ -1775,7 +1775,17 @@ loadBalancer:
...
-  # e.g. native, disabled
+  # disabled - do not use XDP
+  # native - XDP BPF program is run directly out of the networking driver’s
+  #   early receive path.
+  # generic - For drivers not implementing native or offloaded XDP yet, the
+  #   kernel provides an option for generic XDP which does not require any
+  #   driver changes since run at a much later point out of the networking
+  #   stack. This setting is primarily targeted at developers who want to
+  #   write and test programs against the kernel’s XDP API, and will not
+  #   operate at the performance rate of the native or offloaded modes.
```
borkmann: (ditto wrt removing generic on the auto-generated yaml)

pkg/datapath/loader/xdp.go
borkmann: small nit: could you pls not squash the switch/case, but just add option.XDPModeBestEffort with fall through. 

@joamaki isn't the option.XDPModeLink* from the old-style XDP code where we had the prefilter?

borkmann:
Hi Patrick, pls also squash your commits into a single one or multiple but logically split rather than the PR feedback diffs, so it better helps to review. Thanks a lot for pushing this fwd!

pkg/datapath/loader/xdp.go
poblahblahblah: @borkmann done

poblahblahblah:
@borkmann I think I've taken care of your feedback! Thank you for the review!

joamaki (approved):

joamaki:
> @joamaki
> 
> Responding to a comment from above that I can't respond to inline
> 
> ```
> A good rule of thumb to apply is to avoid use of option.Config at deeper layers since that makes testing fairly annoying since now in tests you'll need to mutate the global option.Config and then restore it.
> 
> One possibility here is to just pass the option.Config.XDPMode that reinitializeXDPLocked grabs all the way down to attachProgram and handle the "BestEffort" option there and then convert to XDP flags. This way this low-level function isn't coupled to the option.Config global. E.g. just move the "xdpModeToFlags" from replaceDatapath to attachProgram.
> 
> Also talking about tests, could you look into adding a test case for this into netlink_test.go? TestAttachProgram already has a "XDP" sub-test which would be easy to copy + adapt.
> ```
> 
> I went with using option.Config here because option.Config was already being called when defining the bpf filter on line 194. Is this a change you still want to see made?
> 
> Happy to attempt a test regardless, but do let me know your thoughts on the above question

Yep, I'd be still happy to see this change and the test as they shouldn't be too much effort. I'm however OK with this so marking approved from my side (heading on vacation today).

Documentation/network/kubernetes/kubeproxy-free.rst
```diff
@@ -675,7 +675,9 @@ In case of a multi-device environment, where Cilium's device auto-detection sele
 more than a single device to expose NodePort or a user specifies multiple devices
 with ``devices``, the XDP acceleration is enabled on all devices. This means that
 each underlying device's driver must have native XDP support on all Cilium managed
-nodes. In addition, for performance reasons we recommend kernel >= 5.5 for
+nodes. If you have an environment where some devices support XDP but others do not
+You can have XDP enabled on the supported devices by setting loadBalancer.mode to 
+"best-effort". In addition, for performance reasons we recommend kernel >= 5.5 for
```
qmonnet: Phrasing/formatting nitpicks:

```suggestion
nodes. If you have an environment where some devices support XDP but others do not,
you can enable XDP on the supported devices by setting ``loadBalancer.mode`` to 
``best-effort``. For performance reasons we recommend kernel >= 5.5 for
```

Documentation/network/kubernetes/kubeproxy-free.rst
```diff
@@ -675,7 +675,9 @@ In case of a multi-device environment, where Cilium's device auto-detection sele
 more than a single device to expose NodePort or a user specifies multiple devices
 with ``devices``, the XDP acceleration is enabled on all devices. This means that
 each underlying device's driver must have native XDP support on all Cilium managed
-nodes. In addition, for performance reasons we recommend kernel >= 5.5 for
+nodes. If you have an environment where some devices support XDP but others do not
+You can have XDP enabled on the supported devices by setting loadBalancer.mode to
```
qmonnet: Shouldn't this be `loadBalancer.acceleration` rather than `loadBalancer.mode`??

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1772,7 +1772,11 @@ loadBalancer:
   # mode: snat
 
   # -- acceleration is the option to accelerate service handling via XDP
-  # e.g. native, disabled
+  # disabled - do not use XDP
+  # native - XDP BPF program is run directly out of the networking driver’s
+  #   early receive path.
+  # best-effort - Use native mode XDP acceleration on devices that support
+  #   it.
```
qmonnet: I see you updated both `.../values.yaml` and `.../values.yaml.tmpl`, but you also need to run `make -C install/kubernetes` to propagate the change to `install/kubernetes/cilium/README.md`.

When you do so, you should observe that the formatting you're using here (the list) translates rather badly in the `README.md`, where the comment lines are joined together. Can you please improve it? For example:

```suggestion
  # Applicable values can be: disabled (do not use XDP), native (XDP BPF
  # program is run directly out of the networking driver's early receive path),
  # or best-effort (use native mode XDP acceleration on devices that support it).
```

install/kubernetes/cilium/values.yaml.tmpl
```diff
@@ -1772,7 +1772,11 @@ loadBalancer:
   # mode: snat
 
   # -- acceleration is the option to accelerate service handling via XDP
-  # e.g. native, disabled
+  # disabled - do not use XDP
+  # native - XDP BPF program is run directly out of the networking driver’s
```
qmonnet: Please avoid curly apostrophes, they're a pain for grepping.

qmonnet (changes requested):

Documentation/network/kubernetes/kubeproxy-free.rst
qmonnet: This should also address the complaint in the CI workflow for documentation.

Documentation/network/kubernetes/kubeproxy-free.rst
poblahblahblah: Yes, good catch.

install/kubernetes/cilium/values.yaml.tmpl
poblahblahblah: Oh weird. Not sure how that snuck in.

install/kubernetes/cilium/values.yaml.tmpl
poblahblahblah: I thought I had run `make -C install/kubernetes` in order to generate values.yaml but I didn't see it making any changes to install/kubernetes/cilium/README.md. I'll do this again and dig a little bit.

poblahblahblah: Ok - figured out the issue here.

Since `acceleration` was cancelled out it was never showing up in the helm chart's README.md file since it wasn't registering as being a value that could be set.

For now I have uncommented it in `values.yaml.tmpl` setting it to `disabled` as the default and ran `make -C install/kubernetes`.

it's now documented in the README.

Does this look good @qmonnet?

Documentation/network/kubernetes/kubeproxy-free.rst
```diff
@@ -675,8 +675,10 @@ In case of a multi-device environment, where Cilium's device auto-detection sele
 more than a single device to expose NodePort or a user specifies multiple devices
 with ``devices``, the XDP acceleration is enabled on all devices. This means that
 each underlying device's driver must have native XDP support on all Cilium managed
-nodes. In addition, for performance reasons we recommend kernel >= 5.5 for
-the multi-device XDP acceleration.
+nodes. If you have an environment where some devices support XDP but others do not
+you can have XDP enabled on the supported devices by setting
+``loadBalancer.acceleration`` to   ``best-effort``. In addition, for performance
```
qmonnet:
```suggestion
``loadBalancer.acceleration`` to ``best-effort``. In addition, for performance
```

install/kubernetes/cilium/values.yaml.tmpl
qmonnet: OK, sorry I didn't realise the `acceleration` was simply left out. I didn't mean you had to uncomment it, but that's probably fine. And yes, the reformatted comment looks better to me, thank you!

install/kubernetes/cilium/values.yaml.tmpl
qmonnet: Not sure in this particular case, but I sometimes see them sneaking in when someone copy-paste from a Google doc or some other source that turned straight apostrophes/quotes to curly ones automatically.

pkg/datapath/loader/netlink.go
ti-mo: > Generic mode should not be used in production, it will bypass GRO engine. Fallback to tc is significantly better.

Agreed. Also, adding a retry loop will make the pending changes in #28308 a bit more painful to integrate. Making best-effort mode result in 'optional' XDP progs will result in the least friction overall.

#28308 lifts the `LinkSetXdpFdWithFlags()` here into a new `attachXDPProgram()` function and renames `attachProgram()` to `attachTCProgram()`, so the if statement here will have to move to the call site in `replaceDatapath()`. That all looks :+1:.

qmonnet (changes requested):
Looks better, thanks! But now I don't understand where the new files `install/kubernetes/cilium/Chart.yaml-e`, `...-ee`, and `...-eee` come from. Were they committed by mistake?

install/kubernetes/cilium/values.yaml.tmpl
qmonnet: Ah, if you keep the uncommented `acceleration`, then you need to run `make -C Documentation update-helm-values` as well and commit the changes. This is what the integration tests complain about.

install/kubernetes/cilium/values.yaml.tmpl
poblahblahblah: Those were generated by `make -C install/kubernetes` and yes, they were committed by mistake!

install/kubernetes/cilium/values.yaml.tmpl
poblahblahblah: Ok - I think we are good here.

poblahblahblah: Also, thank you for all of the help with this!

ti-mo (approved):

ti-mo:
@poblahblahblah I've merged https://github.com/cilium/cilium/pull/28308 in the meantime, so I gave this a rebase and moved the error check to `reinitializeXDPLocked()`. I'll kick off tests as well.

poblahblahblah:
I'm not super familiar with the failures we're seeing. Can/should these be run again?

@qmonnet you've been very generous with your time and I appreciate that. It looks like there's another helm failure. Do you have any insight into this?

nathanjsweet (approved):
Approval for API.

qmonnet:
> It looks like there's another helm failure. Do you have any insight into this?

Just as the workflow output says :slightly_smiling_face::

```
HINT: to fix this, run 'make -C Documentation update-helm-values'
```

So please try running this command and committing the changes in `Documentation/helm-reference.rst` (This is a different command from `make -C install/kubernetes`).

poblahblahblah:
> Just as the workflow output says 🙂

Can't believe I missed that. Apologies!

rgo3 (approved):

qmonnet (approved):
Thank you!

qmonnet:
Jussi, Daniel, Nate and I are all part of sig-datapath, so I think we can skip Louis' review.

borkmann (approved):

poblahblahblah:
Thank you for the help, everyone!

Great experience all around!

julianwiedmann:
:wave: should the new `bpf-lb-acceleration` value ("best-effort") also be mentioned in the agent documentation?
https://docs.cilium.io/en/v1.15/cmdref/cilium-agent/#options

>       --bpf-lb-acceleration string                                BPF load balancing acceleration via XDP ("native", "disabled") (default "disabled")

PR #28673, opened 2023-10-18
ipcache: move CIDR restoration to asynchronous APIs

squeed:
This PR contains a key insight: we don't need to insert restored CIDRs directly in to the ipcache. Rather, we just need to ensure their identities are not used by any other prefixes.

This PR accomplishes this by adding a new identity reservation mechanism in to the local allocator. It then plumbs up the ability to request a reserved (i.e. restored) identity when inserting in to the ipcache. The primary benefit of this approach is that we don't have to try and back-fill labels based on a dump of the ipcache. Rather, we just reserve already-existing numeric identities and let all the usual ipcache writers do their thing. When the initial k8s sync is finished and ipcache writing is allowed, then all the restored identities *should* have the exact same labels they had before agent restart - and all without any fancy logic!

This PR has one significant immediate benefit: the `reserved:kube-apiserver` label no longer flaps on restart :tada: .

The previous flow:
1. Dump the ipcache, listing (prefix -> id) pairs.
2. Upsert that prefix -> id directly in the ipcache, guessing at the set of labels `(cidr:xxxx, ... reserved:world)`. At this point, a numeric ID that previously had the kube-apiserver label has now lost it.
3. Perform a k8s sync, adding the `reserved:kube-apiserver` label back to the IP.
4. Do an `ipcache.UpsertLabels()`, which causes the kube-apiserver cidr to allocate a new numeric ID. This is because the set of labels allocated for the cidr didn't include the kube-apiserver label.

The new flow:
1. Dump the ipcache, reserving all IDs from allocation
2. Upsert the prefix in to the ipcache metadata layer, tagging it as having a restored identity.
3. Perform a k8s sync, adding in all relevant metadata / labels
4. Do an `ipcache.UpsertLabels()`, which will allocate an identity with the same set of labels as before, *and* use the previous numeric ID for that prefix.

As a bonus, we can remove the special hack for re-creating node-cidr labels. That was added in `resolveIdentity()` but it really didn't belong there.

If we decide this is the right way forward, outstanding items are:
- [ ] Add test that ensures identities remain constant after restarts
- [ ] Clean up code comments
- [ ] Test extensively with FQDN policies

### DIFF
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -28,7 +28,6 @@ import (
 	"github.com/cilium/cilium/pkg/bandwidth"
 	"github.com/cilium/cilium/pkg/bgp/speaker"
 	bgpv1 "github.com/cilium/cilium/pkg/bgpv1/agent"
-	"github.com/cilium/cilium/pkg/bpf"
 	"github.com/cilium/cilium/pkg/cgroups/manager"
 	"github.com/cilium/cilium/pkg/clustermesh"
 	"github.com/cilium/cilium/pkg/controller"
@@ -57,7 +56,6 @@ import (
 	ipamMetadata "github.com/cilium/cilium/pkg/ipam/metadata"
 	ipamOption "github.com/cilium/cilium/pkg/ipam/option"
 	"github.com/cilium/cilium/pkg/ipcache"
-	ipcachetypes "github.com/cilium/cilium/pkg/ipcache/types"
 	"github.com/cilium/cilium/pkg/k8s"
 	k8sConst "github.com/cilium/cilium/pkg/k8s/apis/cilium.io"
 	k8sClient "github.com/cilium/cilium/pkg/k8s/client"
@@ -69,7 +67,6 @@ import (
 	"github.com/cilium/cilium/pkg/logging"
 	"github.com/cilium/cilium/pkg/logging/logfields"
 	"github.com/cilium/cilium/pkg/maps/ctmap"
-	ipcachemap "github.com/cilium/cilium/pkg/maps/ipcache"
 	"github.com/cilium/cilium/pkg/maps/lbmap"
 	"github.com/cilium/cilium/pkg/maps/policymap"
 	"github.com/cilium/cilium/pkg/metrics"
@@ -89,7 +86,6 @@ import (
 	"github.com/cilium/cilium/pkg/resiliency"
 	"github.com/cilium/cilium/pkg/service"
 	serviceStore "github.com/cilium/cilium/pkg/service/store"
-	"github.com/cilium/cilium/pkg/source"
 	"github.com/cilium/cilium/pkg/status"
 	"github.com/cilium/cilium/pkg/time"
 	"github.com/cilium/cilium/pkg/trigger"
@@ -198,7 +194,7 @@ type Daemon struct {
 	bgpControlPlaneController *bgpv1.Controller
 
 	// CIDRs for which identities were restored during bootstrap
-	restoredCIDRs []netip.Prefix
+	restoredCIDRs map[netip.Prefix]identity.NumericIdentity
 
 	// Controllers owned by the daemon
 	controllers *controller.Manager
@@ -506,66 +502,19 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 		return nil, nil, fmt.Errorf("error while initializing BPF pcap recorder: %w", err)
 	}
 
-	// Collect old CIDR identities
-	var oldNIDs []identity.NumericIdentity
-	var oldIngressIPs []netip.Prefix
+	// Collect CIDR identities from the "old" bpf ipcache and restore them
+	// in to the metadata layer.
 	if option.Config.RestoreState && !option.Config.DryMode {
-		if err := ipcachemap.IPCacheMap().DumpWithCallback(func(key bpf.MapKey, value bpf.MapValue) {
-			k := key.(*ipcachemap.Key)
-			v := value.(*ipcachemap.RemoteEndpointInfo)
-			nid := identity.NumericIdentity(v.SecurityIdentity)
-			if scope := nid.Scope(); scope == identity.IdentityScopeLocal ||
-				(scope == identity.IdentityScopeRemoteNode && option.Config.PolicyCIDRMatchesNodes()) {
-				d.restoredCIDRs = append(d.restoredCIDRs, k.Prefix())
-				oldNIDs = append(oldNIDs, nid)
-			} else if nid == identity.ReservedIdentityIngress && v.TunnelEndpoint.IsZero() {
-				prefix := k.Prefix()
-				oldIngressIPs = append(oldIngressIPs, prefix)
-				addr := prefix.Addr()
-				if addr.Is4() {
-					node.SetIngressIPv4(addr.AsSlice())
-				} else {
-					node.SetIngressIPv6(addr.AsSlice())
-				}
-			}
-		}); err != nil && !os.IsNotExist(err) {
-			log.WithError(err).Debug("Error dumping ipcache")
+		// this *must* be called before initMaps(), which will "hide"
+		// the "old" ipcache.
+		err := d.restoreIPCache()
+		if err != nil {
+			log.WithError(err).Warn("Failed to restore existing identities from the previous ipcache. This may cause policy interruptions during restart.")
 		}
-		// DumpWithCallback() leaves the ipcache map open, must close before opened for
-		// parallel mode in Daemon.initMaps()
-		ipcachemap.IPCacheMap().Close()
 	}
 
 	if err := d.initPolicy(); err != nil {
 		return nil, nil, fmt.Errorf("error while initializing policy subsystem: %w", err)
-	}
-
-	// Preallocate IDs for old CIDRs. This must be done before any Identity allocations are
-	// possible so that the old IDs are still available. That is why we do this ASAP after the
-	// new (userspace) ipcache is created above.
-	//
-	// CIDRs were dumped from the old ipcache, they are re-allocated here, hopefully with the
-	// same numeric IDs as before, but the restored identities are to be upsterted to the new
-	// (datapath) ipcache after it has been initialized below. This is accomplished by passing
-	// 'restoredCIDRidentities' to AllocateCIDRs() and then calling
-	// UpsertGeneratedIdentities(restoredCIDRidentities) after initMaps() below.
-	restoredCIDRidentities := make(map[netip.Prefix]*identity.Identity)
-	if len(d.restoredCIDRs) > 0 {
-		log.Infof("Restoring %d old CIDR identities", len(d.restoredCIDRs))
-		_, err = d.ipcache.AllocateCIDRs(d.restoredCIDRs, oldNIDs, restoredCIDRidentities)
-		if err != nil {
-			log.WithError(err).Error("Error allocating old CIDR identities")
-		}
-		// Log a warning for the first CIDR identity than could not be restored with the
-		// same numeric identity as before the restart. This can only happen if we have
-		// re-introduced bugs into this agent bootstrap order, so we want to surface this.
-		for i, prefix := range d.restoredCIDRs {
-			id, exists := restoredCIDRidentities[prefix]
-			if !exists || id.ID != oldNIDs[i] {
-				log.WithField(logfields.Identity, oldNIDs[i]).Warn("Could not restore all CIDR identities")
-				break
-			}
-		}
 	}
 
 	d.endpointManager = params.EndpointManager
@@ -656,22 +605,6 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 	if err != nil {
 		log.WithError(err).Error("error while opening/creating BPF maps")
 		return nil, nil, fmt.Errorf("error while opening/creating BPF maps: %w", err)
-	}
-	// Upsert restored CIDRs after the new ipcache has been opened above
-	if len(restoredCIDRidentities) > 0 {
-		d.ipcache.UpsertGeneratedIdentities(restoredCIDRidentities, nil)
-	}
-	// Upsert restored local Ingress IPs
-	for _, ingressIP := range oldIngressIPs {
-		d.ipcache.UpsertLabels(
-			ingressIP,
-			labels.LabelIngress,
-			source.Restored,
-			ipcachetypes.NewResourceID(ipcachetypes.ResourceKindDaemon, "", "ingress"),
-		)
-	}
-	if len(oldIngressIPs) > 0 {
-		log.WithField(logfields.Ingress, oldIngressIPs).Info("Restored ingress IPs")
 	}
 
 	// Read the service IDs of existing services from the BPF map and
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -1858,21 +1858,10 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
 		ms.CollectStaleMapGarbage()
 		ms.RemoveDisabledMaps()
 
-		if len(d.restoredCIDRs) > 0 {
-			// Release restored CIDR identities after a grace period (default 10
-			// minutes).  Any identities actually in use will still exist after
-			// this.
-			//
-			// This grace period is needed when running on an external workload
-			// where policy synchronization is not done via k8s. Also in k8s
-			// case it is prudent to allow concurrent endpoint regenerations to
-			// (re-)allocate the restored identities before we release them.
-			time.Sleep(option.Config.IdentityRestoreGracePeriod)
-			log.Debugf("Releasing reference counts for %d restored CIDR identities", len(d.restoredCIDRs))
-			d.ipcache.ReleaseCIDRIdentitiesByCIDR(d.restoredCIDRs)
-			// release the memory held by restored CIDRs
-			d.restoredCIDRs = nil
-		}
+		// Sleep for the --identity-restore-grace-period (default 10 minutes), allowing
+		// the normal allocation processes to finish, before releasing restored resources.
+		time.Sleep(option.Config.IdentityRestoreGracePeriod)
+		d.releaseRestoredCIDRs()
 	}()
 	d.endpointManager.Subscribe(d)
 	// Add the endpoint manager unsubscribe as the last step in cleanup
--- a/daemon/cmd/ipcache.go
+++ b/daemon/cmd/ipcache.go
@@ -4,15 +4,33 @@
 package cmd
 
 import (
+	"fmt"
 	"net"
+	"net/netip"
+	"os"
 
 	"github.com/go-openapi/runtime/middleware"
 
 	"github.com/cilium/cilium/api/v1/models"
 	. "github.com/cilium/cilium/api/v1/server/restapi/policy"
 	"github.com/cilium/cilium/pkg/api"
+	"github.com/cilium/cilium/pkg/bpf"
 	cmtypes "github.com/cilium/cilium/pkg/clustermesh/types"
+	"github.com/cilium/cilium/pkg/identity"
 	"github.com/cilium/cilium/pkg/ipcache"
+	ipcachetypes "github.com/cilium/cilium/pkg/ipcache/types"
+	"github.com/cilium/cilium/pkg/labels"
+	"github.com/cilium/cilium/pkg/logging/logfields"
+	ipcachemap "github.com/cilium/cilium/pkg/maps/ipcache"
+	"github.com/cilium/cilium/pkg/node"
+	"github.com/cilium/cilium/pkg/option"
+	"github.com/cilium/cilium/pkg/source"
+)
+
+var (
+	// the ipcache resource id (just a constant string) used for restored CIDRs
+	restoredCIDRResource = ipcachetypes.NewResourceID(ipcachetypes.ResourceKindDaemon, "", "restored")
+	ingressResource      = ipcachetypes.NewResourceID(ipcachetypes.ResourceKindDaemon, "", "ingress")
 )
 
 func getIPHandler(d *Daemon, params GetIPParams) middleware.Responder {
@@ -86,4 +104,146 @@ func containsSubnet(outer, inner net.IPNet) bool {
 	innerOnes, innerBits := inner.Mask.Size()
 
 	return outerBits == innerBits && outerOnes <= innerOnes && outer.Contains(inner.IP)
+}
+
+// restoreIPCache dumps the existing (old) bpf ipcache, adding relevant information
+// back in to the ipcache / identity allocator.
+//
+// The goal of this logic is to ensure, as much as possible, that local identities (i.e. CIDR)
+// get the same numeric identity upon agent restart.
+//
+// For all local (cidr / remote-node) identities found, this adds a placeholder entry in the
+// ipcache metadata layer requesting the previous numeric identity. When the agent initializes
+// and the ipcache finally recreates the bpf map, this placeholder entry ensures the prefixes
+// exist and have the same identity as before.
+//
+// After a grace period, the placeholder metadata is removed and any prefixes not referenced
+// by other subsystems will be deallocated, see releaseRestoredCIDRs().
+// (Aside: prefix references mostly come from network policies, either directly through CIDR
+// selectors or via ToFQDN rules.)
+//
+// For ingress IPs, it will add those to the ipcache and configure the local node
+// accordingly.
+//
+// This *must* be called before initMaps(), which will hide the "old" ipcache.
+func (d *Daemon) restoreIPCache() error {
+	ingressIPs := make([]netip.Prefix, 0, 2)
+	// need to preserve this so we can remove it later.
+	d.restoredCIDRs = map[netip.Prefix]identity.NumericIdentity{}
+
+	// Dump the bpf ipcache, recording any prefixes with local or ingress
+	// numeric identities.
+	err := ipcachemap.IPCacheMap().DumpWithCallback(func(key bpf.MapKey, value bpf.MapValue) {
+		k := key.(*ipcachemap.Key)
+		v := value.(*ipcachemap.RemoteEndpointInfo)
+		nid := identity.NumericIdentity(v.SecurityIdentity)
+
+		if isLocalIdentity(nid) {
+			d.restoredCIDRs[k.Prefix()] = nid
+		} else if nid == identity.ReservedIdentityIngress && v.TunnelEndpoint.IsZero() {
+			ingressIPs = append(ingressIPs, k.Prefix())
+		}
+	})
+	// dumpwithcallback() leaves the ipcache map open, must close before opened for
+	// parallel mode in daemon.initmaps()
+	ipcachemap.IPCacheMap().Close()
+
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return fmt.Errorf("error dumping ipcache: %w", err)
+	}
+
+	// Now that the map is dumped,
+	// - upsert relevant metadata in to the ipcache
+	// - withhold all existing CIDR identities
+	// - add Ingress IPs to local Node.
+	metaUpdates := make([]ipcache.MU, 0, len(ingressIPs)+len(d.restoredCIDRs))
+	nidsToWithhold := make([]identity.NumericIdentity, 0, len(d.restoredCIDRs))
+
+	for prefix, nid := range d.restoredCIDRs {
+		nidsToWithhold = append(nidsToWithhold, nid)
+		metaUpdates = append(metaUpdates, ipcache.MU{
+			Prefix:   prefix,
+			Source:   source.Restored,
+			Resource: restoredCIDRResource,
+			Metadata: []ipcache.IPMetadata{ipcachetypes.RequestedIdentity(nid)},
+		})
+	}
+	for _, prefix := range ingressIPs {
+		metaUpdates = append(metaUpdates, ipcache.MU{
+			Prefix:   prefix,
+			Source:   source.Restored,
+			Resource: ingressResource,
+			Metadata: []ipcache.IPMetadata{labels.LabelIngress},
+		})
+
+		// Set any restored ingress IPs back on the LocalNode object
+		d.nodeLocalStore.Update(func(n *node.LocalNode) {
+			addr := prefix.Addr()
+			if addr.Is4() {
+				n.IPv4IngressIP = addr.AsSlice()
+			} else {
+				n.IPv6IngressIP = addr.AsSlice()
+			}
+		})
+	}
+	if len(ingressIPs) > 0 {
+		log.WithField(logfields.Ingress, ingressIPs).Info("Restored ingress IPs")
+	}
+
+	// Insert the batched changes in to the ipcache.
+	// Even though the ipcache map hasn't been initialized yet, this is
+	// safe to do so, because the ipcache's apply controller is currently
+	// paused.
+	d.ipcache.IdentityAllocator.WithholdLocalIdentities(nidsToWithhold)
+	d.ipcache.UpsertMetadataBatch(metaUpdates...)
+
+	return nil
+}
+
+// releaseRestoredCIDRS removes the placeholder metadata that was inserted
+// in to the ipcache when local identities were restored.
+// Any identities actually in use will still exist after this.
+//
+// This should be called after a grace period (default 10 minutes, set
+// by --identity-restore-grace-period).
+// This grace period is needed when running on an external workload
+// where policy synchronization is not done via k8s. Also in k8s
+// case it is prudent to allow concurrent endpoint regenerations to
+// (re-)allocate the restored identities before we release them.
+//
+// Any CIDRs still in use after the grace period will have other sources
+// of metadata in the ipcache, and thus will remain. CIDRs for which
+// restoration was the only source of metadata will be deallocated.
+func (d *Daemon) releaseRestoredCIDRs() {
+	defer func() {
+		// release the memory held by restored CIDRs
+		d.restoredCIDRs = nil
+	}()
+	if len(d.restoredCIDRs) == 0 {
+		return
+	}
+
+	log.WithField(logfields.Count, len(d.restoredCIDRs)).Info("Removing identity reservations for restored CIDR identities")
+	updates := make([]ipcache.MU, 0, len(d.restoredCIDRs))
+	nids := make([]identity.NumericIdentity, 0, len(d.restoredCIDRs))
+	for prefix, nid := range d.restoredCIDRs {
+		nids = append(nids, nid)
+		updates = append(updates, ipcache.MU{
+			Prefix:   prefix,
+			Resource: restoredCIDRResource,
+			Metadata: []ipcache.IPMetadata{ipcachetypes.RequestedIdentity(0)},
+		})
+	}
+
+	d.ipcache.RemoveMetadataBatch(updates...)
+	d.ipcache.IdentityAllocator.UnwithholdLocalIdentities(nids)
+}
+
+func isLocalIdentity(nid identity.NumericIdentity) bool {
+	scope := nid.Scope()
+	return scope == identity.IdentityScopeLocal ||
+		(scope == identity.IdentityScopeRemoteNode && option.Config.PolicyCIDRMatchesNodes())
 }
--- a/pkg/identity/cache/allocator.go
+++ b/pkg/identity/cache/allocator.go
@@ -135,6 +135,15 @@ type IdentityAllocator interface {
 	// ReleaseCIDRIdentitiesByID() is a wrapper for ReleaseSlice() that
 	// also handles ipcache entries.
 	ReleaseCIDRIdentitiesByID(context.Context, []identity.NumericIdentity)
+
+	// WithholdLocalIdentities holds a set of numeric identities out of the local
+	// allocation pool(s). Once withheld, a numeric identity can only be used
+	// when explicitly requested via AllocateIdentity(..., oldNID).
+	WithholdLocalIdentities(nids []identity.NumericIdentity)
+
+	// UnwithholdLocalIdentities removes numeric identities from the withheld set,
+	// freeing them for general allocation.
+	UnwithholdLocalIdentities(nids []identity.NumericIdentity)
 }
 
 // InitIdentityAllocator creates the global identity allocator. Only the first
@@ -392,6 +401,25 @@ func (m *CachingIdentityAllocator) AllocateIdentity(ctx context.Context, lbls la
 	}
 
 	return identity.NewIdentity(identity.NumericIdentity(idp), lbls), isNew, nil
+}
+
+func (m *CachingIdentityAllocator) WithholdLocalIdentities(nids []identity.NumericIdentity) {
+	log.WithField(logfields.Identity, nids).Debug("Withholding numeric identities for later restoration")
+
+	// The allocators will return any identities that are not in-scope.
+	nids = m.localIdentities.withhold(nids)
+	nids = m.localNodeIdentities.withhold(nids)
+	if len(nids) > 0 {
+		log.WithField(logfields.Identity, nids).Error("Attempt to restore invalid numeric identities.")
+	}
+}
+
+func (m *CachingIdentityAllocator) UnwithholdLocalIdentities(nids []identity.NumericIdentity) {
+	log.WithField(logfields.Identity, nids).Debug("Unwithholding numeric identities")
+
+	// The allocators will ignore any identities that are not in-scope.
+	m.localIdentities.unwithhold(nids)
+	m.localNodeIdentities.unwithhold(nids)
 }
 
 // Release is the reverse operation of AllocateIdentity() and releases the
--- a/pkg/identity/cache/local.go
+++ b/pkg/identity/cache/local.go
@@ -13,6 +13,7 @@ import (
 	"github.com/cilium/cilium/pkg/kvstore"
 	"github.com/cilium/cilium/pkg/labels"
 	"github.com/cilium/cilium/pkg/lock"
+	"github.com/cilium/cilium/pkg/logging/logfields"
 )
 
 type localIdentityCache struct {
@@ -24,6 +25,15 @@ type localIdentityCache struct {
 	minID               identity.NumericIdentity
 	maxID               identity.NumericIdentity
 	events              allocator.AllocatorEventSendChan
+
+	// withheldIdentities is a set of identities that should be considered unavailable for allocation,
+	// but not yet allocated.
+	// They are used during agent restart, where local identities are restored to prevent unnecessary
+	// ID flapping on restart.
+	//
+	// If an old nID is passed to lookupOrCreate(), then it is allowed to use a withhend entry here. Otherwise
+	// it must allocate a new ID not in this set.
+	withheldIdentities map[identity.NumericIdentity]struct{}
 }
 
 func newLocalIdentityCache(scope, minID, maxID identity.NumericIdentity, events allocator.AllocatorEventSendChan) *localIdentityCache {
@@ -35,6 +45,7 @@ func newLocalIdentityCache(scope, minID, maxID identity.NumericIdentity, events 
 		minID:               minID,
 		maxID:               maxID,
 		events:              events,
+		withheldIdentities:  map[identity.NumericIdentity]struct{}{},
 	}
 }
 
@@ -57,18 +68,32 @@ func (l *localIdentityCache) getNextFreeNumericIdentity(idCandidate identity.Num
 			// let nextNumericIdentity be, allocated identities will be skipped anyway
 			log.Debugf("Reallocated restored local identity: %d", idCandidate)
 			return idCandidate, nil
+		} else {
+			log.WithField(logfields.Identity, idCandidate).Debug("Requested local identity not available to allocate")
 		}
 	}
 	firstID := l.nextNumericIdentity
 	for {
 		idCandidate = l.nextNumericIdentity | l.scope
-		if _, taken := l.identitiesByID[idCandidate]; !taken {
+		_, taken := l.identitiesByID[idCandidate]
+		_, withheld := l.withheldIdentities[idCandidate]
+		if !taken && !withheld {
 			l.bumpNextNumericIdentity()
 			return idCandidate, nil
 		}
 
 		l.bumpNextNumericIdentity()
 		if l.nextNumericIdentity == firstID {
+			// Desperation: no local identities left (unlikely). If there are withheld
+			// but not-taken identities, claim one of them.
+			for withheldID := range l.withheldIdentities {
+				if _, taken := l.identitiesByID[withheldID]; !taken {
+					delete(l.withheldIdentities, withheldID)
+					log.WithField(logfields.Identity, withheldID).Warn("Local identity allocator full; claiming first withheld identity. This may cause momentary policy drops")
+					return withheldID, nil
+				}
+			}
+
 			return 0, fmt.Errorf("out of local identity space")
 		}
 	}
@@ -153,6 +178,40 @@ func (l *localIdentityCache) release(id *identity.Identity) bool {
 	}
 
 	return false
+}
+
+// withhold marks the nids as unavailable. Any out-of-scope identities are returned.
+func (l *localIdentityCache) withhold(nids []identity.NumericIdentity) []identity.NumericIdentity {
+	if len(nids) == 0 {
+		return nil
+	}
+
+	unused := make([]identity.NumericIdentity, 0, len(nids))
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
+	for _, nid := range nids {
+		if nid.Scope() != l.scope {
+			unused = append(unused, nid)
+			continue
+		}
+		l.withheldIdentities[nid] = struct{}{}
+	}
+
+	return unused
+}
+
+func (l *localIdentityCache) unwithhold(nids []identity.NumericIdentity) {
+	if len(nids) == 0 {
+		return
+	}
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
+	for _, nid := range nids {
+		if nid.Scope() != l.scope {
+			continue
+		}
+		delete(l.withheldIdentities, nid)
+	}
 }
 
 // lookup searches for a local identity matching the given labels and returns
--- a/pkg/ipcache/cidr.go
+++ b/pkg/ipcache/cidr.go
@@ -27,16 +27,12 @@ import (
 // identities are placed in 'newlyAllocatedIdentities' and it is the caller's
 // responsibility to upsert them into ipcache by calling UpsertGeneratedIdentities().
 //
-// Previously used numeric identities for the given prefixes may be passed in as the
-// 'oldNIDs' parameter; nil slice must be passed if no previous numeric identities exist.
-// Previously used NID is allocated if still available. Non-availability is not an error.
-//
 // Upon success, the caller must also arrange for the resulting identities to
 // be released via a subsequent call to ReleaseCIDRIdentitiesByCIDR().
 //
 // Deprecated: Prefer UpsertLabels() instead.
 func (ipc *IPCache) AllocateCIDRs(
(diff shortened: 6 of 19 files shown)


squeed:
I had an idea for how we might do this, and it only took a few hours to sketch out and implement. Looking for feedback, especially from @joestringer @jrajahalme

squeed:
Not that this is ready to go, but CI is green :-)

daemon/cmd/daemon.go
```diff
@@ -693,10 +676,13 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
...
 		return nil, nil, fmt.Errorf("error while opening/creating BPF maps: %w", err)
 	}
-	// Upsert restored CIDRs after the new ipcache has been opened above
-	if len(restoredCIDRidentities) > 0 {
-		d.ipcache.UpsertGeneratedIdentities(restoredCIDRidentities, nil)
+
+	// Upsert restored CIDRs and their identities in to the metadata layer
+	// Note that the identities themselves were reserved in the local allocator above
+	for prefix, nid := range d.restoredCIDRs {
+		d.ipcache.UpsertMetadata(prefix, source.Restored, restoredCIDRResource, ipcachetypes.RestoredIdentity(nid))
```
joestringer: nit: rather than `restoredCIDRResource` , we may want to do something like `ipcachetypes.NewResourceID(ipcachetypes.ResourceKindDaemon, "", "restored")`

^ idea being to try to (a) reuse the same symbols for easier grep and (b) uniquely identify callsites. Usually `ResourceID` is about the k8s resource, but for in-daemon resources I think we can abuse the last field.

(I noticed this will collide with the following calls)

```
$ git grep ResourceKindDaemon
daemon/cmd/daemon.go:                   ipcachetypes.NewResourceID(ipcachetypes.ResourceKindDaemon, "", ""),
daemon/cmd/datapath.go: daemonResourceID := ipcachetypes.NewResourceID(ipcachetypes.ResourceKindDaemon, "", "")
pkg/ipcache/types/types.go:     ResourceKindDaemon   = ResourceKind("daemon")
```

At some point these should be exposed through API / sysdump so we can properly validate how ipcache entries are formed.

joestringer: Actually this isn't a nit. If we have multiple codepaths using the same resource ID then they can delete entries that were inserted by a completely different codepath.

Fix: https://github.com/cilium/cilium/pull/28770

daemon/cmd/daemon_main.go
```diff
@@ -1855,7 +1856,9 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
 			// (re-)allocate the restored identities before we release them.
 			time.Sleep(option.Config.IdentityRestoreGracePeriod)
 			log.Debugf("Releasing reference counts for %d restored CIDR identities", len(d.restoredCIDRs))
-			d.ipcache.ReleaseCIDRIdentitiesByCIDR(d.restoredCIDRs)
+			for cidr := range d.restoredCIDRs {
+				d.ipcache.RemoveMetadata(cidr, restoredCIDRResource, ipcachetypes.RestoredIdentity(0))
+			}
```
joestringer: nit, Ah we probably should figure out better batch APIs for this. Could be hundreds/thousands of lock/edit/unlock and triggering of next label injection. I think the behaviour should be OK in terms of ratelimits and so on but it'd be nice to make all the changes at once.

pkg/ipcache/metadata.go
```diff
@@ -224,7 +224,7 @@ func (ipc *IPCache) InjectLabels(ctx context.Context, modifiedPrefixes []netip.P
 			} // else continue below to remove the old entry
 		} else {
 			// Insert to propagate the updated set of labels after removal.
-			newID, _, err = ipc.resolveIdentity(ctx, prefix, prefixInfo, identity.InvalidIdentity)
+			newID, _, err = ipc.resolveIdentity(ctx, prefix, prefixInfo, identity.NumericIdentity(prefixInfo.RestoredIdentity()))
```
joestringer: :eyes: how does this behave if there's another update for this prefix after restore but before the restored identity is removed from the ipcache? I presume this will fill in a numeric identity that is already allocated to a (potentially different) set of labels.

joestringer: `getNextFreeNumericIdentity()` only allocates it if it wasn't already taken :ok_hand:

pkg/identity/cache/local.go
```diff
@@ -62,7 +72,9 @@ func (l *localIdentityCache) getNextFreeNumericIdentity(idCandidate identity.Num
 	firstID := l.nextNumericIdentity
 	for {
 		idCandidate = l.nextNumericIdentity | l.scope
-		if _, taken := l.identitiesByID[idCandidate]; !taken {
+		_, taken := l.identitiesByID[idCandidate]
+		_, reserved := l.reservedIdentities[idCandidate]
```
joestringer: How does the callpath from ipcache determine that it is the legitimate one allocating the identity with this reserved numeric identity? I don't see any unreserve call or special path where that caller passes down the context to skip the reserve check.

Result is this will always ignore the reserved identity and avoid allocating it.

joestringer (commented):
Seems like a reasonable approach in general, though I suspect in the details the current draft isn't quite working as intended due to the lack of unreserve before allocation, see feedback below.

I might suggest splitting the identity allocation logic & tests into a dedicated commit for easier subsequent review & to keep small commits, but that's as much personal preference as anything.

pkg/ipcache/metadata.go
joestringer: Hmm see next feedback below, I think this last parameter is being ignored currently due to the reservation and lack of unreserve call.

joestringer: Maybe to clarify, my concern here is that we should not reallocate the same number to a different set of labels (or at least not without being super careful about flushing/updating all other users of the numeric identity)

daemon/cmd/daemon.go
joestringer: I guess technically right now it's fine because the data types for `aux` are not the same, but I feel like it's reasonable to split these out further by resourceID.

daemon/cmd/daemon_main.go
squeed: Sure, we should add batched `UpdateMetadataMulti()` and `RemoveMetadataMulti()` methods

pkg/identity/cache/local.go
squeed: If `idCandidate` is non-zero, then it is allowed to request a reserved identity. Otherwise, the next free non-reserved identity is selected.

pkg/ipcache/metadata.go
squeed: **edit**: this comment is wrong, see below.

For sure. Fortunately, we already test this case extensively, because right now any cidr identities with the `reserved:kube-apiserver` label *always* flap on every restart.

The implicit ideal is that - once the k8s caches are synced - the set of labels for a given prefix should not change after agent restart. This is probably not a 100% airtight assumption, but it seems pretty reasonable. Unless we do snapshotting (and maybe we should!), there is no way for us to reconstruct the set of labels for a given prefix. We only know its identity.

So, we reserve the prefix -> identity mapping and hope the set of labels is stable after a restart.

squeed: Wait, that previous comment didn't quite address what you were getting at.

> Maybe to clarify, my concern here is that we should not reallocate the same number to a different set of labels (or at least not without being super careful about flushing/updating all other users of the numeric identity).

That's a very valid concern, and there's an easy fix for this too: we just need to explicitly Unreserve numeric IDs after the reservation grace period is over. The assumption, then, is that we have reached a quiescent state after the reservation grace period is over, and can safely "free" the pool of reserved identities back in to use.

pkg/identity/cache/local.go
joestringer: Oh, I misunderstood the intent of these changes. I wasn't familiar with this code, saw the changes to `idCandidate`, saw  this was the parameter to the function, and assumed that this path is about returning the function argument. I missed that the `idCandidate` variable is reused in this loop to iterate through the number space. Basically this code is implementing the protection of the reserved identity, not trying to return the parameter the caller gives.

Makes sense :+1: , I retract my concern here.

squeed:
@joestringer I updated the PR based on your feedback. No big changes, just some small cleanups. The biggest change is explicitly managing reservation lifetime.

I also renamed the metadata-bits from RestoredIdentity to RequestedIdentity to better match semantics.

daemon/cmd/daemon_main.go
squeed: @joestringer take a look, tell me what you think of the API. It's... fine.

pkg/ipcache/ipcache.go
```diff
@@ -460,6 +468,31 @@ func (ipc *IPCache) UpsertMetadata(prefix netip.Prefix, src source.Source, resou
 	ipc.TriggerLabelInjection()
 }
 
+// UpsertMetadataBatch applies updates to multiple prefixes in a single transaction,
+// reducing potential lock contention.
+func (ipc *IPCache) UpsertMetadataBatch(updates []MetadataUpdate) {
```
christarazi: Hm, we could convert `[]MetadataUpdate` to be 

```
func (ipc *IPCache) UpsertMetadataBatch(updates ...MetadataUpdate) {
```

and that way in the future, we could have just `UpsertMetadata()` with a variadic arg and all the call sites would convert over to using `MetadataUpdate`. There would be no need to differentiate between a scalar call or a batch call because the variadic function can accept one or more args. In the latter case (via `len(update)`) the logic can detect a "batch" case and do the right thing.

I recognize that converting all the call sites to `MetadataUpdate` is likely out of scope for this PR, but making the API compatible with that change in this PR seems like a nice step.

squeed: Done I also renamed the type to `MU`. Maybe that's OK.

pkg/ipcache/ipcache.go
```diff
@@ -488,9 +526,9 @@ func (ipc *IPCache) UpsertPrefixes(prefixes []netip.Prefix, src source.Source, r
 	ipc.metadata.Lock()
 	for _, p := range prefixes {
 		ipc.metadata.upsertLocked(p, src, resource, labels.GetCIDRLabels(p))
-		ipc.metadata.enqueuePrefixUpdates(p)
 	}
 	ipc.metadata.Unlock()
+	ipc.metadata.enqueuePrefixUpdates(prefixes...)
```
christarazi, line 530: I just noticed that there's an lock ordering change here as well. Previously `enqueuePrefixUpdates` was called within the metadata lock, and now it's being done outside. It's worth checking whether this won't have unintended side effects.

squeed: I stared at the wall for a while and couldn't think of any. The prefix update loop has an entirely independent lifecycle, so any possible interleaving of upsert and updates is still possible. In other words, async gonna async.

That is to say, `(upsertMetadata(A), enqueue(a), upsertMetadata(b), enqueue(b))` can result in `(inject(a), inject(b))` **or** `inject(a, b)` depending on the status of the ipcache injection controller.

I'm pretty sure the original code was unnecessary. All other functions save `enqueuePrefixUpdates()` for the end.

(note: edited to address that this was an upsert, not remove, call)

daemon/cmd/ipcache.go
```diff
@@ -87,3 +104,126 @@ func containsSubnet(outer, inner net.IPNet) bool {
...
+			Source:   source.Restored,
+			Resource: restoredCIDRResource,
+			Metadata: []ipcache.IPMetadata{ipcachetypes.RequestedIdentity(nid)},
+		})
+	}
+	for _, prefix := range ingressIPs {
+		metaUpdates = append(metaUpdates, ipcache.MU{
+			Prefix:   prefix,
+			Source:   source.Restored,
+			Resource: ipcachetypes.NewResourceID(ipcachetypes.ResourceKindDaemon, "", "ingress"),
```
christarazi: We can extract this out like it was done for `restoredCIDRResource`

daemon/cmd/ipcache.go
```diff
@@ -87,3 +104,126 @@ func containsSubnet(outer, inner net.IPNet) bool {
...
+	// Even though the ipcache map hasn't been initialized yet, this is
+	// safe to do so, because the ipcache's apply controller is currently
+	// paused.
+	d.ipcache.IdentityAllocator.ReserveIdentities(nidsToReserve)
+	d.ipcache.UpsertMetadataBatch(metaUpdates...)
+
+	return nil
+}
+
+// releaseRestoredCIDRS removes the "stub" metadata that was inserted
```
christarazi: It might help to clarify what that "stub" metadata is, i.e. `RequestedIdentity`. It makes it easier for readers that don't have as much ipcache knowledge to know what it's referring to.

daemon/cmd/ipcache.go
```diff
@@ -87,3 +104,126 @@ func containsSubnet(outer, inner net.IPNet) bool {
...
+// in to the ipcache when local identities were restored.
+// Any identities actually in use will still exist after this.
+//
+// This first waits a grace period (default 10 minutes).
```
christarazi: I think it also makes sense to just make an explicit reference to `IdentityRestoreGracePeriod` here so that the docs don't go stale if we ever change the 10m default

pkg/ipcache/ipcache.go
joestringer: I noticed and pondered this earlier too, and came to the same conclusions as Casey. Given that `TriggerLabelInjection()` and the dequeue/enqueue functions for the prefix updates don't interact with the metadata lock, I couldn't identify any meaningful impact to ordering.

Strictly speaking it's plausible for some of these updates to be handled by `InjectLabels()` before the prefix is enqueued, iff the prefix was enqueued previously and `InjectLabels()` is starting to process the updates immediately after the `ipc.metadata.Unlock()` above. The result would be that the next round for running `InjectLabels()` would end up handling this prefix and doing nothing for it. I believe this should be idempotent. If we see an issue there then we could iterate further on the locking pattern or handling logic, but I think we can count that as beyond the scope of this PR.

pkg/ipcache/ipcache.go
```diff
@@ -457,24 +458,61 @@ func (ipc *IPCache) DumpToListener(listener IPIdentityMappingListener) {
...
+	}
 	ipc.metadata.Unlock()
-	ipc.metadata.enqueuePrefixUpdates(prefix)
+	ipc.metadata.enqueuePrefixUpdates(prefixes...)
 	ipc.TriggerLabelInjection()
 }
 
+// RemoveMetadata removes metadata associated with a specific resource from the
+// supplied prefix. Individual metadata types must be supplied for removal, but the
+// data need not match.
```
joestringer, line 493: I guess the data need not match, but the data *type* must match because it hits `resourceInfo.unmerge()`.

pkg/identity/cache/local.go
```diff
@@ -57,18 +68,32 @@ func (l *localIdentityCache) getNextFreeNumericIdentity(idCandidate identity.Num
...
+			// Desperation: no local identities left (unlikely). If there are reserved
+			// but not-taken identities, claim one of them.
+			for reservedID := range l.reservedIdentities {
+				if _, taken := l.identitiesByID[reservedID]; !taken {
+					delete(l.reservedIdentities, reservedID)
+					log.WithField(logfields.Identity, reservedID).Warn("Local identity allocator full; claiming first reserved identity. This may cause momentary policy drops")
+					return reservedID, nil
+				}
+			}
+
```
joestringer: Do we need this case?

I'm a little concerned that in this case we're in a scenario of "no resources left" and the tradeoff may end up be between "impact existing traffic" or "impact new traffic", and this extra logic might end up tipping us over into "impact existing traffic" territory. The warning is useful to understand we're in this situation, but I struggle a little bit to decide whether it's a clear win to have this particular behaviour in that scenario.

I'd be interested in any insights you might have around the background thinking to this  logic.

joestringer: Maybe to expand some perspectives for discussions' sake:

- I can understand that if for some reason users of this code hold their reservations infinitely, this provides an escape hatch. However, that is a bug in the caller APIs, so we should be testing & validating usage of the API rather than solving it here (?)
- Consider a case where the set of identities to reserve is exceedingly high. During startup it would first reserve many identities. Over time k8s resources will get synced and may also attempt to allocate identities. In an ideal situation I think that the restore logic should restore the identities, trigger the usage of them, then remove all reservations, and ideally afterwards all other k8s resource sync logic would then start to allocate identities from the remaining space. This would have the smallest impact on the changes to identities during restore, and hence the lowest likelihood of connection impact. On the other hand it also assumes that whatever other allocation logic that exists would be able to retry in order to eventually allocate an identity from the actual available remaining allocation space. Let's say that this code continues to exist, it could mean that some of the restored identities will transition their identities unnecessarily, and in the worst case in ways that may not be able to be recovered from.

In some sense when you're at/near the limit maybe things are borked anyway, but it's one thing to say that "there are no more identities"; it's another thing to say "there are no more identities and by the way Cilium decided to arbitrarily rearrange which peers have identities now".

pkg/ipcache/cidr.go
```diff
@@ -36,7 +36,7 @@ import (
 //
 // Deprecated: Prefer UpsertLabels() instead.
 func (ipc *IPCache) AllocateCIDRs(
-	prefixes []netip.Prefix, oldNIDs []identity.NumericIdentity, newlyAllocatedIdentities map[netip.Prefix]*identity.Identity,
+	prefixes []netip.Prefix, newlyAllocatedIdentities map[netip.Prefix]*identity.Identity,
```
joestringer, line 35: Please update the function doc regarding `oldNIDs`.

daemon/cmd/daemon.go
```diff
@@ -524,67 +520,19 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
...
-				if addr.Is4() {
-					node.SetIngressIPv4(addr.AsSlice())
-				} else {
-					node.SetIngressIPv6(addr.AsSlice())
-				}
-			}
-		}); err != nil && !os.IsNotExist(err) {
-			log.WithError(err).Debug("Error dumping ipcache")
+		// this *must* becalled before initMaps(), which will clobber
+		// the "old" ipcache.
```
joestringer: nit on the space, and I think it's a little misleading to describe it as "clobbering"; the ambiguity of that wording could imply that the content of ipcache may be impacted, not just the pointer. `initMaps()` will change the open userspace reference to the ipcache so that it no longer refers to the instance that was used by the previous datapath + userspace and instead is a brand new fresh one to be used by the "new" instance of the datapath (which is about to be loaded).  Otherwise it's a good callout, this is a critical aspect which... unfortunately is only protected by this comment at this point. Maybe at some point we'll encode this constraint more reliably into the startup logic.

```suggestion
		// this *must* be called before initMaps(), which will hide
		// the "old" ipcache.
```

daemon/cmd/daemon_main.go
```diff
@@ -1850,22 +1850,7 @@ func startDaemon(d *Daemon, restoredEndpoints *endpointRestoreState, cleaner *da
...
-			// where policy synchronization is not done via k8s. Also in k8s
-			// case it is prudent to allow concurrent endpoint regenerations to
-			// (re-)allocate the restored identities before we release them.
-			time.Sleep(option.Config.IdentityRestoreGracePeriod)
-			log.Debugf("Releasing reference counts for %d restored CIDR identities", len(d.restoredCIDRs))
-			d.ipcache.ReleaseCIDRIdentitiesByCIDR(d.restoredCIDRs)
-			// release the memory held by restored CIDRs
-			d.restoredCIDRs = nil
-		}
+		go d.releaseRestoredCIDRs() // called from goroutine as it sleeps
```
joestringer: This is already in a goroutine from line 1842, I don't see the need to spawn an additional one here. If we want to make the sleep more obvious, I would keep it in this function and leave `releaseRestoredCIDRs` to just the business logic of actually releasing the CIDRs. But that's more a matter of preference, I don't feel strongly either way about that.
```suggestion
		d.releaseRestoredCIDRs()
```

daemon/cmd/daemon.go
joestringer: Similar "clobber" nit on `restoreIPCache()`.

daemon/cmd/ipcache.go
```diff
@@ -87,3 +104,126 @@ func containsSubnet(outer, inner net.IPNet) bool {
...
+	d.restoredCIDRs = map[netip.Prefix]identity.NumericIdentity{}
+
+	// Dump the bpf ipcache, recording any prefixes with local or ingress
+	// numeric identities.
+	err := ipcachemap.IPCacheMap().DumpWithCallback(func(key bpf.MapKey, value bpf.MapValue) {
+		k := key.(*ipcachemap.Key)
+		v := value.(*ipcachemap.RemoteEndpointInfo)
+		nid := identity.NumericIdentity(v.SecurityIdentity)
+
+		if isLocalIdentity(nid) {
```
joestringer, line 141: I welcome the refactoring here, but I think it would be wise to split the refactor vs. the functional change into separate commits. At a glance I can see that this conditional check is dropping the other conditionals in the existing logic in the main branch, which makes me nervous that this commit may be breaking other features while doing the refactor + logical change. The simpler the commits are, the lower the likelihood we'll miss something in development & review around how these changes are being applied.

daemon/cmd/ipcache.go
```diff
@@ -87,3 +104,126 @@ func containsSubnet(outer, inner net.IPNet) bool {
...
+	for _, prefix := range ingressIPs {
+		metaUpdates = append(metaUpdates, ipcache.MU{
+			Prefix:   prefix,
+			Source:   source.Restored,
+			Resource: ipcachetypes.NewResourceID(ipcachetypes.ResourceKindDaemon, "", "ingress"),
+			Metadata: []ipcache.IPMetadata{labels.LabelIngress},
+		})
+
+		// Set any restored ingress IPs back on the LocalNode object
+		d.nodeLocalStore.Update(func(n *node.LocalNode) {
```
joestringer, line 183: Mildly curious, I don't see this logic in the current tree. Is this a bug? Why is this logic now needed? Or did it exist somewhere else and I missed it?

daemon/cmd/ipcache.go
```diff
@@ -87,3 +104,126 @@ func containsSubnet(outer, inner net.IPNet) bool {
...
+		// release the memory held by restored CIDRs
+		d.restoredCIDRs = nil
+	}()
+	if len(d.restoredCIDRs) == 0 {
+		return
+	}
+
+	time.Sleep(option.Config.IdentityRestoreGracePeriod)
+
+	log.Infof("Releasing metadata for %d restored CIDR identities", len(d.restoredCIDRs))
```
joestringer: If we're making this an Info level log, please use structured logging.

(Does this message make any sense to users, do we need to inform them? I guess this is the kind of thing that helps as a smoking gun if someone notices issues 10m after startup, but users may be confused what it means to release metadata... is that a good thing? bad thing? Why am I being told?)

joestringer: If we intend this to be more user-facing, we could rephrase it in terms of "releasing preallocated memory for identity restore" or something, at least then it's a bit less ambiguous.

pkg/ipcache/cidr.go
joestringer: (nice cleanup by the way, I always felt a little weird about this parameter to this function)

joestringer (changes requested):
I went back through, logically it seems sound to me. I spaced out a little bit going through the main `restoreIPCache()` logic patch due to the combination of refactor + logical change, so that could do with a revisit. I did notice some discrepancies there that need to be addressed, and ideally that should be split into two patches. Other than that it's mostly minor nits around function comments and logs, and one unnecessary goroutine. The corner case handling of identity allocation from the reserved range when under identity pressure wasn't entirely clear to me either. We can discuss further in the threads below.

daemon/cmd/ipcache.go
joestringer: I guess arguably if reviewers can pick up on these discrepancies then the patch is not so difficult to follow that it *needs* splitting out, so if it's too much of a pain then let me know and I can take a fresh look at this patch.

squeed: In this case, the logic is identical. I just thought the condition was almost impossible to read, so I split it out to a function (that will surely be inlined by the compiler anyways).

I also expect this predicate to gain complexity as more features are added (even though I don't have any in mind right now).

daemon/cmd/ipcache.go
squeed: It was [here](https://github.com/cilium/cilium/pull/28673/files#diff-2ed5415f522d0b72066d62527fb40dbea9bf745f45d5e2d47f7715680f165fcaL543). AIUI we're trying to migrate off of the top-level local node mutators, so I did that.

daemon/cmd/ipcache.go
squeed: Right, we've had enough bugs over the years with this logic that I wanted to log *something* by default. I'll think of a better message. Something like "Removing identity reservations for restored CIDR identities".

pkg/identity/cache/local.go
squeed: I thought about this for a while, and decided that allowing allocations to make progress was the better choice of the two. The reason is, basically:
- If the ipcache can't make progress, then we have less of a chance of releasing identities to free up room
- If local allocation is failing, then we're potentially causing policy drops in the case of FQDN policies anyways
- Identity reservation is best-effort :-)
- The most likely scenario where we hit this would be a bug with identity restoration, in which case a policy flap is already happening. In any case, a flap is better than a livelock

That is an interesting point about shortening the grace period -- we probably can just wait for a full kvstore / k8s sync and be done after that. That lessens the time where identity pressure would be exacerbated by reservations.

FWIW, once #28652 or something similar is merged, there should be no policy drops if identities do flap on restart.

pkg/ipcache/ipcache.go
squeed: Yup, I tried to capture this in the example.

daemon/cmd/daemon.go
squeed: the clobber's children have no... bpf maps?  (point taken)

daemon/cmd/daemon_main.go
squeed: OK, fair enough. I'll just add a comment of ("no code after this line, it sleeps for an arbitrary time")

daemon/cmd/ipcache.go
squeed: Also, @christarazi yelled at me for doing too many commits, so I tried to scale back a bit :-)

pkg/identity/cache/allocator.go
```diff
@@ -135,6 +135,15 @@ type IdentityAllocator interface {
 	// ReleaseCIDRIdentitiesByID() is a wrapper for ReleaseSlice() that
 	// also handles ipcache entries.
 	ReleaseCIDRIdentitiesByID(context.Context, []identity.NumericIdentity)
+
+	// ReserveIdentities holds a set of numeric identities out of the local
+	// allocation pool(s). Once reserved, a numeric identity can only be used
+	// when explicitly requested via AllocateIdentity(..., oldNID).
+	ReserveIdentities(nids []identity.NumericIdentity)
```
gandro: nit (non-blocking): Should we call this `ReserveLocalIdentities`? Since the identities being past in must be local and users of this interface might not read the documentation close enough.

pkg/identity/cache/allocator.go
```diff
@@ -394,6 +403,22 @@ func (m *CachingIdentityAllocator) AllocateIdentity(ctx context.Context, lbls la
 	return identity.NewIdentity(identity.NumericIdentity(idp), lbls), isNew, nil
 }
 
+func (m *CachingIdentityAllocator) ReserveIdentities(nids []identity.NumericIdentity) {
+	log.WithField(logfields.Identity, nids).Debug("Reserving numeric identities for later restoration")
+
+	// The allocators will ignore any identities that are not in-scope.
```
gandro: nit (non-blocking): How hard would it be do detect out of scope (i.e. in particular also global) identities and log an error? That would make CI fail if something attempted to reserve something that cannot be reserved.

daemon/cmd/ipcache.go
```diff
@@ -87,3 +104,128 @@ func containsSubnet(outer, inner net.IPNet) bool {
...
+		} else if nid == identity.ReservedIdentityIngress && v.TunnelEndpoint.IsZero() {
+			ingressIPs = append(ingressIPs, k.Prefix())
+		}
+	})
+	if err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("error dumping ipcache: %w", err)
+	}
+	// DumpWithCallback() leaves the ipcache map open, must close before opened for
+	// parallel mode in Daemon.initMaps()
+	ipcachemap.IPCacheMap().Close()
```
gandro: potential bug: The old version of the code did log on error, while the new code does an early return if `DumpWithCallback` failed.  And given that the caller continues on error, I think this violates the comment here which says that close is important.

I think we want to close this irrespective of the `err`, i.e. move to before the `if` statement (or perhaps `defer` it)

daemon/cmd/ipcache.go
gandro: +1 on the cleanup.

(I'm not a fan of how ingress IP restoration is cluttered all over the place (i.e. the IP stored here still needs to be allocated from IPAM later on), but alas, that's a refactor for another time)

daemon/cmd/daemon.go
```diff
@@ -674,22 +622,6 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
...
-	for _, ingressIP := range oldIngressIPs {
-		d.ipcache.UpsertLabels(
-			ingressIP,
-			labels.LabelIngress,
-			source.Restored,
-			ipcachetypes.NewResourceID(ipcachetypes.ResourceKindDaemon, "", "ingress"),
-		)
-	}
-	if len(oldIngressIPs) > 0 {
-		log.WithField(logfields.Ingress, oldIngressIPs).Info("Restored ingress IPs")
```
gandro, line 674: This log line is lost in the refactor

gandro (changes requested):
Looks like an overall improvement to me, awesome!

I'm not a high-confidence reviewer when it comes to identity restoration, since that's not code I've interacted with before, but the IPCache change itself look mostly good to me. One potential bug, and a couple very nitpicky nits that I don't feel strongly about.

pkg/identity/cache/allocator.go
squeed: I'd initially shied away from that, but yes, I think we probably should.

pkg/identity/cache/allocator.go
squeed: A previous version of the PR did that, but it was removed for efficient batching. I'll see about adding *something*.

daemon/cmd/ipcache.go
squeed: Oh, very good catch.

pkg/identity/cache/allocator.go
squeed: Did it.

daemon/cmd/ipcache.go
joestringer: Ah okay, it was just a second-level refactor that I missed during the first pass.

joestringer: I didn't realize I was giving conflicting guidance so sorry about that :) I think it's a good question to grapple with and establish a shared understanding as a community though. Looking at the stats:

```
 daemon/cmd/daemon.go         |  80 +++----------------------------------
 daemon/cmd/daemon_main.go    |  17 +-------
 daemon/cmd/ipcache.go        | 140 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 pkg/ipcache/metadata.go      |  10 -----
 pkg/ipcache/metadata_test.go |  69 +++++++++++++++++---------------
 5 files changed, 184 insertions(+), 132 deletions(-)
```

That's over 300 lines of change in a single patch, which when combining refactors + logical changes can be difficult to keep in your head during review. That's the background to my request. That said, if we don't have agreement amongst reviewers in general then it's unfair for me to push harder on my point here. We can bring this discussion up in general to try to figure out where the line is. Personally I almost exclusively review commit-by-commit and ensure each change is logically separate and does not include sizeable logical and cosmetic changes in the same patch, and I use the patch-by-patch steps as a way to stage my review to reason through the whole thing. Above around 200 lines or so I find that the time to grok a single patch goes from ~single digit minutes to ~double digit minutes as I scan through the diffs in the patch side-by-side.

Anyways, I digress, I will take a fresh look at this patch now I'm more familiar with it.

gandro (approved):

pkg/identity/cache/allocator.go
gandro: The "Did it" refers to to the other comment, right? https://github.com/cilium/cilium/pull/28673#discussion_r1380162865

I don't see any logging in the current version, but as mentioned before, that's not a blocker for me

squeed: Got too excited, forgot to push.

daemon/cmd/ipcache.go
christarazi: Just chiming in, I don't recall giving commit feedback on this PR and I don't see it as a comment here either :sweat_smile:. The only thing I see regarding commit feedback is https://github.com/cilium/cilium/pull/28673#pullrequestreview-1695783530. If I had given some confusing feedback, I apologize. 

To be clear, I believe having many small and isolated commits as possible makes the reviewer's job easier and thus makes code reviews much more efficient and effective. Keeping refactor commits separate from functional impacting commits is always easier to review and IMO a good practice to evangelize in our community standards. I believe these are standards that we've aligned on in the community AFAICT and I am happy continue contributing in clarifying those and discussing it with the wider community as well.

Back to this PR, I would agree with Joe that splitting out a refactoring commit from the functional change is easier to review. I leave it to you Casey to handle that as you think is best.

daemon/cmd/ipcache.go
joestringer: I think "Removing reservations" sounds a bit less alarming than "releasing metadata", so :+1: . I guess time will tell if it continues to be confusing.

pkg/identity/cache/allocator.go
squeed: It's down on line 413 -- any failed reservations will surface an error message.

daemon/cmd/ipcache.go
squeed: @christarazi  It's all good :p, it was in a different PR where I was... perhaps a bit excessive with the refactoring commits.

In this case, it wasn't really possible to have a noop refactoring commit. There was logic in a few places within `newDaemon()`, and the need for it to be carefully interleaved only went away when switching to the asynchronous APIs. So I couldn't lift to a single function cleanly -- it would have been a fair bit of throwaway work.

daemon/cmd/ipcache.go
squeed: I even meant to do that but it slipped my mind.

daemon/cmd/ipcache.go
squeed: Oh yeah, good point. "stub" isn't even that useful a term, since it's valid metadata like anything else. Perhaps "placeholder" is a better term. I'll write some more words in the docblock

pkg/identity/cache/local.go
joestringer: Hmm. So we're debating the case of behaviour when allocating (16M - N reserved) local identities right? When I phrase it that way it seems like such a corner case that we're unlikely to hit this anyway, so it's barely worth debating. Probably my key principle for such a rare case would be just to make it fail as simply as possible, no smarts, just make the behaviour obvious. To be fair, the log statement is good enough to check that box.

With high N I agree that it's useful to minimize the risk of some livelock type scenario where the total identity space is big enough but somehow we can't allocate because N gets out of control. I suppose my argument was thinking more like "I wouldn't want to debug a system that ignores the reservations and hence rearranges the prefix <-> identity mappings upon restart." That said, I'm happy to defer to you on this one. Between the rarity, the warning log and the test cases, I think it's good enough.

daemon/cmd/ipcache.go
joestringer: Oh I see, it's just directly calling the Update function openly instead of using the helper. There's no functional change here right?

daemon/cmd/daemon.go
```diff
@@ -523,68 +519,20 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 		return nil, nil, fmt.Errorf("error while initializing BPF pcap recorder: %w", err)
 	}
 
-	// Collect old CIDR identities
-	var oldNIDs []identity.NumericIdentity
-	var oldIngressIPs []netip.Prefix
+	// Collect old CIDR identities from the
```
joestringer: Collect them from where?

daemon/cmd/ipcache.go
```diff
@@ -87,3 +105,144 @@ func containsSubnet(outer, inner net.IPNet) bool {
...
+			Resource: restoredCIDRResource,
+			Metadata: []ipcache.IPMetadata{ipcachetypes.RequestedIdentity(nid)},
+		})
+	}
+	for _, prefix := range ingressIPs {
+		metaUpdates = append(metaUpdates, ipcache.MU{
+			Prefix:   prefix,
+			Source:   source.Restored,
+			Resource: ingressResource,
+			Metadata: []ipcache.IPMetadata{labels.LabelIngress},
```
joestringer, line 179: :thinking: I guess we never clean up these restored ipcache metadata updates for Ingress in the way that we would for the other CIDRs. But then... these are just statically allocated once during runtime and used for the rest of the agent life so I guess this is fine, as long as we never want to have some "reinitialize ingress" logic?

I don't think we need to make any change, just an observation. Also this is the same as the existing code :+1:

daemon/cmd/ipcache.go
```diff
@@ -87,3 +105,144 @@ func containsSubnet(outer, inner net.IPNet) bool {
...
+// This should be called after a grace period (default 10 minutes, set
+// by --identity-restore-grace-period).
+// This grace period is needed when running on an external workload
+// where policy synchronization is not done via k8s. Also in k8s
+// case it is prudent to allow concurrent endpoint regenerations to
+// (re-)allocate the restored identities before we release them.
+//
+// Any CIDRs still in use after the grace period will have other sources
+// of metadata in the ipcache, and thus will remain. CIDRs for which
+// restoration was the only source of metadata will be deallocated.
```
joestringer, line 219: Point of mild interest: A bunch of these may actually have been previously allocated by the FQDN subsystem during the previous agent run, and this now reallocates them... then also separately I guess the FQDN restore logic will do its own repeated allocation of these prefixes (?) in order to ensure this property for those particular IPs.

I'm actually not entirely sure this is true since I ran into issues with this when I previously attempted to solve this problem, but at the same time I don't see how this commit is impacting that behaviour at all. So if we have a CIDR identity restore issue with the FQDN logic, we can treat that as a whole separate problem, no need to try to tie it into this PR.

daemon/cmd/ipcache.go
```diff
@@ -87,3 +105,144 @@ func containsSubnet(outer, inner net.IPNet) bool {
...
+		nids = append(nids, nid)
+		updates = append(updates, ipcache.MU{
+			Prefix:   prefix,
+			Resource: restoredCIDRResource,
+			Metadata: []ipcache.IPMetadata{ipcachetypes.RequestedIdentity(0)},
+		})
+	}
+
+	d.ipcache.RemoveMetadataBatch(updates...)
+	d.ipcache.UnreserveLocalIdentities(nids)
```
joestringer: nit: The way I see the reserve / unreserve being tied to the metadata batch, it makes me wonder whether the ipcache should just make the reservation when inserting metadata with this type. One reason I thought of this is that it seems like the whole reason we allocate these `nids` on the stack is in order to have a slightly-differently-typed argument and I wondered whether these two could share the same argument, then wondered if they could just be the same call and simplify usage from `daemon`. 

But then I recognize that lower down we probably need to create the slice and do the datatype swap anyway, so we're probably just shuffling the logic around slightly if we did something like this. We could always iterate and tidy and improve if we think there's a better way on this, so no need to address at this time. Feel free to resolve this comment without action.

pkg/ipcache/metadata.go
```diff
@@ -455,16 +455,6 @@ func (ipc *IPCache) resolveIdentity(ctx context.Context, prefix netip.Prefix, in
...
 
-	// If we are restoring an identity in the remote-node identity scope, we need to
-	// unconditionally add remote-node and cidr labels back. We do not need to check
-	// if CIDR selection for nodes is enabled here, as we explicitly check further down
-	// when we remove CIDR labels for identities that should not have them.
-	if restoredIdentity.Scope() == identity.IdentityScopeRemoteNode {
-		lbls.MergeLabels(labels.LabelRemoteNode)
-		cidrLabels := labels.GetCIDRLabels(prefix)
-		lbls.MergeLabels(cidrLabels)
-	}
```
joestringer, line 466: :tada: nice to clean up this piece now that it's tied implicitly into the prefix insert + the choice of restored identity (for the remote node label).

joestringer (approved):
Looks good, just one comment at the top about a half-finished func doc. The rest were just observations while reviewing but I think that's all in line with others' understanding so no changes should be necessary from the other pieces of feedback.

pkg/identity/cache/allocator.go
christarazi, line 1: Passing thought: I wonder if the usage of the term "reserved" in this context is a bit overloaded / confusing. We already have a notion of "reserved identities" such as the ones that range from 1-128, i.e. world, host, remote-node, etc. What do you think of the term "claimed" instead? I don't feel super strong about it, but I do think it's a bit confusing to have two notions of "reserved identities".

daemon/cmd/ipcache.go
christarazi: The corresponding 

```
d.ipcache.IdentityAllocator.ReserveLocalIdentities(nidsToReserve)
```

is called on the `IdentityAllocator`, in the highlighted line it is acting directly on the ipcache. Is that asymmetry intended?

christarazi (commented):
LGTM, just a few questions left for me.

daemon/cmd/ipcache.go
squeed: Correct, behavior is identical. The helper can't go away yet, but this removes one more user.

daemon/cmd/ipcache.go
squeed: Expanded the docblock a bit.

daemon/cmd/daemon.go
squeed: There needs to be a facepalm reaction.

daemon/cmd/ipcache.go
squeed: Yeah, I kept the original logic. It did seem odd to me too.

daemon/cmd/ipcache.go
squeed: Right, the FQDN subsystem will unconditionally do allocations. However, those allocations are still metadata-aware, so this all should work.

daemon/cmd/ipcache.go
squeed: I did think about that; it felt funny to touch the identity allocator in the "entrypoint" methods. The precedent is that the id allocator is only touched in `resolveIdentity()`. Not that we can't override precedent, mind.

I *hope* nothing else uses the identity reservation system :-).

pkg/ipcache/metadata.go
squeed: yeah it felt good to get rid of this bit.

daemon/cmd/ipcache.go
squeed: Specifically, we do the ipcache restoration strictly before the fqdn proxy is started, so the placeholder metadata will already exist and the requested numeric identity should be respected.

pkg/identity/cache/allocator.go
squeed: Ah hm. How about "preorder" :-p. Though, I'm struggling to come up with a better term. The identities are not in use, but they're not free for general use. Like a, well, reserved table at a popular restaurant. This is where @gandro's suggestion to name it "ReserveLocalIdentities()" makes sense. I suppose we could use "withhold" if that sits well with you.

Aside: the code also is sometimes ambiguous between identity-as-label-set and identity-as-32-bit-number. It would be nice to tighten up that verbiage somehow.

daemon/cmd/ipcache.go
squeed: No, definitely not intended. How did that ever work!? (there are some strange interfaces afoot that I don't entirely understand).

daemon/cmd/ipcache.go
squeed: Renamed this from "stub" to "placeholder" and tried to give some more context in the documentation.

squeed:
Rebased, added a ginkgo test to ensure that numeric identities are not shuffled on agent restart.

daemon/cmd/ipcache.go
```diff
@@ -87,3 +105,144 @@ func containsSubnet(outer, inner net.IPNet) bool {
...
+// get the same numeric identity upon agent restart.
+//
+// For all local (cidr / remote-node) identities found, this adds a placeholder entry in the
+// ipcache metadata layer requesting the previous numeric identity. When the agent initializes
+// and the ipcache finally recreates the bpf map, this placeholder entry ensures the prefixes
+// exist and have the same identity as before.
+//
+// After a grace period, the placeholder metadata is removed and any no-longer-referenced
+// prefixes will be deallocated, see releaseRestoredCIDRs(). (Aside: prefixes are typically
+// referenced by ToCIDR rules in a network policy, but there are other sources as well.)
```
squeed: note to self: add obvious missing source: FQDNs.

test/k8s/manifests/cnp-to-cidr-oneoneoneone.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: "cilium.io/v2"
+kind: CiliumNetworkPolicy
+metadata:
+  name: "to-cidr-oneoneoneone"
+spec:
+  endpointSelector:
+    matchLabels:
+      {}
```
squeed: note to self: move this to a label selector that does not select any pods :facepalm:

pkg/identity/cache/allocator.go
christarazi: I'll leave it up to you. I'm ok with withhold / withheld. :+1:

squeed:
I was stress-testing this PR with FQDN proxy and found a nasty bug where prefixes were incorrectly removed from the BPF ipcache while still in use with FQDN. The problem fqdn proxy does not do an `UpsertGeneratedIdentities()` if an identity is not new, and thus the source of a prefix is not "upgraded".

Luckily, it was an easy fix: since the set of all referenced identities is passed to `UpsertGeneratedIdentities()`, we can overwrite `createdFromMetadata` retroactively and prevent deletion.

pkg/identity/cache/allocator.go
squeed: Moved to "withheld"

daemon/cmd/ipcache.go
squeed: An update: This didn't work. I had to fix it. See [comment](https://github.com/cilium/cilium/pull/28673#issuecomment-1794792763)

squeed:
re-requested @gandro's review, as he added the `createdFromMetadata` field.

pkg/ipcache/cidr.go
```diff
@@ -151,17 +144,28 @@ func (ipc *IPCache) UpsertGeneratedIdentities(newlyAllocatedIdentities map[netip
...
-		toUpsert[prefix] = id
+		if existing.createdFromMetadata {
+			// the createdFromMetadata field is used to tell the ipcache that it is safe to delete
+			// a prefix when all entries are removed from the metadata layer. However, as this is the
+			// "old-style" API, we need to tell InjectLabels(): hands off!
+			//
+			// This upsert tells the ipcache that the prefix is now in the domain of an older user
+			// and thus should not be deleted by clearing createdFromMetadata
+			toUpsert[prefix] = id
+		}
```
gandro: Outch! Good catch.

But yeah, it should be safe - we basically downgrade the identity from being managed by `InjectLabels` to become manually managed by the caller. Before this PR, all such prefixes were created using the legacy API, so they had `createdFromMetadata = false` from the start. Now, we create the identities via the new API, then call this function here to "downgrade" them to be owned by the old API, and then we call `d.ipcache.RemoveMetadataBatch` for all prefixes that we have downgraded, making them exclusively owned by the old API caller. It works, but it's not pretty.

One more indicator that moving the CIDR/FQDN identities to the new IPCache API cannot happen soon enough.

gandro (approved):

pkg/ipcache/cidr.go
squeed: Working on it :-)

test/helpers/kubectl.go
```diff
@@ -743,6 +743,34 @@ func (kub *Kubectl) GetCiliumHostEndpointState(ciliumPod string) (string, error)
 	return strings.TrimSpace(res.Stdout()), nil
 }
 
+// GetCiliumIdentityForIP returns the numeric identity for a given IP address
+// according to a node's BPF ipcache.
+func (kub *Kubectl) GetCiliumIdentityForIP(ciliumPod, ip string) (int, error) {
+	cmd := fmt.Sprintf("cilium-dbg bpf ipcache get %s", ip)
```
christarazi, line 749: Seeing tests like this suggests that we need to consider reinvesting again into controlplane unit tests (`test/controlplane`) at some point in the near future. This is a good addition in the meantime though.

christarazi (commented):
> I was stress-testing this PR with FQDN proxy and found a nasty bug where prefixes were incorrectly removed from the BPF ipcache while still in use with FQDN. The problem fqdn proxy does not do an UpsertGeneratedIdentities() if an identity is not new, and thus the source of a prefix is not "upgraded".

> 
> Luckily, it was an easy fix: since the set of all referenced identities is passed to UpsertGeneratedIdentities(), we can overwrite createdFromMetadata retroactively and prevent deletion.
> 
> I've added a new commit and test case for this scenario.

Nice find. A couple of questions:

1. What does the stress testing look like? I think it would be valuable to upstream it at some point so when we have an infrastructure for stress testing in the future, we have the FQDN proxy covered.
2. Does this bug exist in `main` or was it something that arose from what this PR is solving? If it's the former, then I would suggest a separate PR so that it can be backported separately from this PR. Additionally, do we have an idea on how we missed it before?

squeed:
>     1. What does the stress testing look like? I think it would be valuable to upstream it at some point so when we have an infrastructure for stress testing in the future, we have the FQDN proxy covered.

Nothing exciting, just ensuring that fqdn policies still work while spamming restarts. We have a test case that covers this already, it just needs to be updated to wait for the restore grace period to stop.

> 
>     2. Does this bug exist in `main` or was it something that arose from what this PR is solving? If it's the former, then I would suggest a separate PR so that it can be backported separately from this PR. Additionally, do we have an idea on how we missed it before?

I don't _think_ so. The specific interaction that's problematic is 
1. UpsertMetadata
2. AllocateCIDRs & UpsertGeneratedIdentities
3. RemoveMetadata

Moving CIDR restoration over to the new API is what triggered this. That said, I'm sure there's some permutation that I'm not thinking of, so it can't hurt to split out.

squeed:
Test failures look interesting. One of them is a bug in my new test, the other is ensuring fqdn works after restart. The latter is failing, which is concerning. Works On My Machine, of course.

squeed:
Found another missing case where we were stepping on toes, this time when allocating identities after a restart. Phew. That should fix it.

squeed:
CI was green except a broken Travis. Rebased on main.

pkg/ipcache/ipcache.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/cilium/cilium/pkg/identity"
 	"github.com/cilium/cilium/pkg/identity/cache"
 	ipcacheTypes "github.com/cilium/cilium/pkg/ipcache/types"
+	ipcachetypes "github.com/cilium/cilium/pkg/ipcache/types"
```
christarazi: We can just use one of these import paths

christarazi (approved):
Just one nit to fix up.

pkg/ipcache/ipcache.go
squeed: :facepalm:

squeed:
All approvals in, only one non-required CI job failing, MLH has added ready-to-merge. Merging.

PR #28676, opened 2023-10-18
install: Remove AWS-CONNMARK-CHAIN iptable rules

nebril:
We were missing some AWS-specific iptable rules that need to be deleted when running in ENI mode.

related: #25804 

```release-note
Remove AWS-CONNMARK-CHAIN iptable rules when running in ENI mode.
```


squeed (approved):

joestringer:
Do we need to backport this, at least to v1.14?

PR #28677, opened 2023-10-18
hive/jobs: Add health reporting to jobs

dylandreimerink:
Jobs now report their health to the modular health reporter. Job groups now require a `cell.Scope` during creation which is provided by a `cell.Module` wrapping the cell in which the job group is created.

When jobs are added to the job group, each job will get a health reporter under the given scope, so directly under the module scope.

One shot jobs are marked OK (running) once started unless they return an error. The callback signature has changed and one shot jobs now are provided the `cell.HealthReporter` so long running jobs can report their health without exiting.

Timer jobs are initially marked OK (Primed). After each invocation the job reports OK ({time spent}) unless an error was returned.

Observer jobs are also initially marked OK (Primed). They report OK ({time spent}) [{events processed}] after a event, but at most once every ten seconds to avoid spamming the health reporter.

To facilitate the pattern where jobs are spawned dynamically such as for each selected service in the case of L2 announcements, scoped groups were added. A scoped group can be created with `Group.Scoped(name)`, any jobs added to this scoped group will be added to a sub-scope with the given name.

Example output:
```
# cilium status --verbose
[...]
Modules Health:
  Module                    Status   Message   Last Updated
  agent.controlplane.auth   OK              3m28s
  
   observer-job-auth request-authentication OK Primed        1h0m0.000164074s ago (x1)
   observer-job-auth gc-identity-events     OK Primed        1h0m0.000162725s ago (x1)
   timer-job-auth gc-cleanup                OK OK (31.919µs) 1h0m0.000137548s ago (x1)
  agent.controlplane.daemon             Unknown   No status reported yet            n/a
  agent.controlplane.endpoint-manager   OK                  3s
  
   cilium-endpoint-470 (kube-system/coredns-5d78c9869d-2jrz6)
    cep-k8s-sync        OK sync-to-k8s-ciliumendpoint (470) 1h3m20.874172654s ago (x382)
    datapath-regenerate OK Endpoint regeneration successful 1h3m20.826398573s ago (x2)
    policy map sync     OK sync-policymap-470               1h3m16.678519928s ago (x5)
   cilium-endpoint-1263 (kube-system/coredns-5d78c9869d-mxnfp)
               cep-k8s-sync        OK          sync-to-k8s-ciliumendpoint (1263) 1h3m20.87392989s ago  (x382)
               datapath-regenerate OK          Endpoint regeneration successful  1h3m20.822672251s ago (x2)
               policy map sync     OK          sync-policymap-1263               1h3m16.670332097s ago (x5)
   endpoint-gc OK                  endpoint-gc 1h3m25.058578593s ago             (x13)
   cilium-endpoint-3145 (/)
    policy map sync     OK sync-policymap-3145              1h3m20.730445585s ago (x5)
    datapath-regenerate OK Endpoint regeneration successful 1h3m22.972805065s ago (x4)
   cilium-endpoint-302 (/)
    policy map sync     OK sync-policymap-302               1h3m16.676547083s ago (x5)
    datapath-regenerate OK Endpoint regeneration successful 1h3m21.916903513s ago (x2)
   cilium-endpoint-1188 (local-path-storage/local-path-provisioner-6bc4bddd6b-c9gfj)
    cep-k8s-sync        OK sync-to-k8s-ciliumendpoint (1188) 1h3m20.87430288s ago  (x382)
    datapath-regenerate OK Endpoint regeneration successful  1h3m20.863074528s ago (x2)
    policy map sync     OK sync-policymap-1188               1h3m16.648360935s ago (x5)
  agent.controlplane.l2-announcer   OK            28s
  
   leader-election
                                   job-leader-election/kube-system/kube-dns    OK              Running              59m58.136220986s ago (x1)
                                   job-leader-election/default/kubernetes      OK              Running              59m58.136271731s ago (x1)
                                   job-leader-election/kube-system/hubble-peer OK              Running              59m58.136250023s ago (x1)
   job-l2-announcer run            OK                                          Running         1h3m0.002508311s ago (x1)
   timer-job-l2-announcer lease-gc OK                                          OK (1.930675ms) 1h3m0.002484125s ago (x1)
  agent.controlplane.node-manager   OK            56s
  
   background-sync OK Node validation successful 1h2m31.411019841s ago (x58)
  agent.datapath.agent-liveness-updater   OK             0s
  
   timer-job-agent-liveness-updater OK OK (31.58µs) 1h3m26.000438231s ago (x1)
  agent.datapath.l2-responder   OK            63m
  
   job-l2-responder-reconciler OK Running 136.135µs ago (x1)
```

```release-note
Jobs now report health
```

### DIFF
--- a/pkg/auth/cell.go
+++ b/pkg/auth/cell.go
@@ -75,6 +75,7 @@ type authManagerParams struct {
 	Logger      logrus.FieldLogger
 	Lifecycle   hive.Lifecycle
 	JobRegistry job.Registry
+	Scope       cell.Scope
 
 	Config       config
 	AuthMap      authmap.Map
@@ -119,6 +120,7 @@ func registerAuthManager(params authManagerParams) (*AuthManager, error) {
 	})
 
 	jobGroup := params.JobRegistry.NewGroup(
+		params.Scope,
 		job.WithLogger(params.Logger),
 		job.WithPprofLabels(pprof.Labels("cell", "auth")),
 	)
--- a/pkg/hive/job/job.go
+++ b/pkg/hive/job/job.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"runtime/pprof"
+	"strconv"
 	"sync"
 	"time"
 
@@ -37,7 +38,10 @@ var Cell = cell.Module(
 // A Registry creates Groups, it maintains references to these groups for the purposes of collecting information
 // centralized like metrics.
 type Registry interface {
-	NewGroup(opts ...groupOpt) Group
+	// NewGroup creates a new group of jobs which can be started and stopped together as part of the cells lifecycle.
+	// The provided scope is used to report health status of the jobs. A `cell.Scope` can be obtained via injection
+	// an object with the correct scope is provided by the closest `cell.Module`.
+	NewGroup(scope cell.Scope, opts ...groupOpt) Group
 }
 
 type registry struct {
@@ -64,7 +68,7 @@ func newRegistry(
 
 // NewGroup creates a new Group with the given `opts` options, which allows you to customize the behavior for the
 // group as a whole. For example by allowing you to add pprof labels to the group or by customizing the logger.
-func (c *registry) NewGroup(opts ...groupOpt) Group {
+func (c *registry) NewGroup(scope cell.Scope, opts ...groupOpt) Group {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
@@ -80,6 +84,7 @@ func (c *registry) NewGroup(opts ...groupOpt) Group {
 	g := &group{
 		options: options,
 		wg:      &sync.WaitGroup{},
+		scope:   scope,
 	}
 
 	c.groups = append(c.groups, g)
@@ -93,13 +98,15 @@ func (c *registry) NewGroup(opts ...groupOpt) Group {
 // is bound to the lifecycle of the cell.
 type Group interface {
 	Add(...Job)
+	// Scoped creates a scroped group, jobs added to this scoped group will appear as a sub scope in the health reporter
+	Scoped(name string) ScopedGroup
 	hive.HookInterface
 }
 
 // Job in an interface that describes a unit of work which can be added to a Group. This interface contains unexported
 // methods and thus can only be implemented by functions in this package such as OneShot, Timer, or Observer.
 type Job interface {
-	start(ctx context.Context, wg *sync.WaitGroup, options options)
+	start(ctx context.Context, wg *sync.WaitGroup, scope cell.Scope, options options)
 }
 
 type group struct {
@@ -111,6 +118,8 @@ type group struct {
 	ctx        context.Context
 	cancel     context.CancelFunc
 	queuedJobs []Job
+
+	scope cell.Scope
 }
 
 type options struct {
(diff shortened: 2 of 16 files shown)


pkg/auth/cell.go
```diff
@@ -119,6 +120,7 @@ func registerAuthManager(params authManagerParams) (*AuthManager, error) {
 	})
 
 	jobGroup := params.JobRegistry.NewGroup(
```
mhofstetter, line 122: From a pure usage perspective it would be nicer to get rid of the injecting the additional explicit `cell.Scope` - at least for the "default" case where one don't want to set an explicit scope. This should help in keeping the API smaller and easier to understand.

Would it be possible that the job registry itself would already have the correct scope and provide the possibility to retrieve a scoped job - either by injecting a `job.ScopedRegistry` or providing a function like `NewScopedGroup`?

pkg/hive/job/job.go
```diff
@@ -37,7 +38,7 @@ var Cell = cell.Module(
 // A Registry creates Groups, it maintains references to these groups for the purposes of collecting information
 // centralized like metrics.
 type Registry interface {
-	NewGroup(opts ...groupOpt) Group
+	NewGroup(scope cell.Scope, opts ...groupOpt) Group
```
mhofstetter, line 44: needs to be updated in the documentation: https://docs.cilium.io/en/latest/contributing/development/hive/#job-groups

mhofstetter, line 44: maybe out of scope for this PR: it would be great to have an example of the job framework in the hive examples: https://github.com/cilium/cilium/tree/main/pkg/hive/example

mhofstetter (commented):
nice addition - thanks @dylandreimerink 

some comments inline

pkg/auth/cell.go
dylandreimerink: We considered this but its difficult. `cell.Scope` does this trick internally where the object you get always contains the name of the current module and with the logger where we already set fields depending on the module name, but the same doesn't work for `job`. That is because `job` has its own package which also imports `cell`, so when we add the logic for this to the module we get an import loop. So we would have to move job into the `cell` package to make that work. That isn't unthinkable, but decided not to do so now.

mhofstetter: Unfortunate. Thanks for the info!

I think it's OK as long as we document this well enough. (see my other two comments).

And it might be worth to enrich the comment of the function `NewGroup` that the `cell.Scope` can get retrieved via injection.

pkg/hive/job/job.go
dylandreimerink: I moved the example we had in the docs to the example directory of hive, that way its also easier to keep up to date when refactoring changes such as these.

dylandreimerink: Thanks, I always forget about the docs, they have also been updated.

mhofstetter (approved):

pkg/auth/cell.go
joamaki: I do like that `job` is in a separate package, just for naming (e.g. `job.Timer` and so on) so I wouldn't fully move it into the `cell` package. 

We could consider seeing how it'd look like if we'd move the required data type definitions from `job` into `cell` (e.g. `job.Registry` becomes `cell.JobRegistry`, with type alias in `job`), but let's try that out separately.

learnitall (approved):
Docs changes look good!

joamaki (approved):

aditighag (approved):

aditighag:
I was added as a reviewer for cilium/contributing, so only approving for that code owner. Re-adding other review groups.

markpash (approved):

joamaki:
I'd say we've got enough approvals now, merging.

PR #28682, opened 2023-10-18
docs: Clarify BPF Map Pressure Metric

nathanjsweet:
BPF Map Pressure does not just measure "usage",
as the current iteration of the docs suggests. It
actually measures the required map size against
the configured/allocated map size. This metric
is useful at values >= 1.0 as it conveys the scope
of the problem to users of the metric.


qmonnet (approved):
Looks good, thanks

nebril (approved):

PR #28704, opened 2023-10-19
k8s: Invalidate Policies that Support "EndPort"

nathanjsweet:
Cilium does not currently support port ranges in
network policies.

```release-note
policy: Cilium will not process or enforce network policies with port ranges or Kubernetes network policies that use "EndPort".
```

### DIFF
--- a/pkg/k8s/network_policy.go
+++ b/pkg/k8s/network_policy.go
@@ -251,6 +251,26 @@ func ParseNetworkPolicy(np *slim_networkingv1.NetworkPolicy) (api.Rules, error) 
 	return api.Rules{rule}, nil
 }
 
+// NetworkPolicyHasEndPort returns true if the network policy has an
+// EndPort.
+func NetworkPolicyHasEndPort(np *slim_networkingv1.NetworkPolicy) bool {
+	for _, iRule := range np.Spec.Ingress {
+		for _, port := range iRule.Ports {
+			if port.EndPort != nil && *port.EndPort > 0 {
+				return true
+			}
+		}
+	}
+	for _, eRule := range np.Spec.Egress {
+		for _, port := range eRule.Ports {
+			if port.EndPort != nil && *port.EndPort > 0 {
+				return true
+			}
+		}
+	}
+	return false
+}
+
 func parsePodSelector(podSelectorIn *slim_metav1.LabelSelector, namespace string) *slim_metav1.LabelSelector {
 	podSelector := &slim_metav1.LabelSelector{
 		MatchLabels: make(map[string]slim_metav1.MatchLabelsValue, len(podSelectorIn.MatchLabels)),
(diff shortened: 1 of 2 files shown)


joestringer:
I think the subtle question here is exactly what behaviour we want on upgrade. Currently the policy with endport will put endpoints into default deny and therefore lock down the endpoints. If we reject the policy completely, that could mean the policy is ineffective and opens up the endpoints completely if that's the only policy. It may be safer to accept but log, given this is long-standing behaviour (?)

pkg/k8s/network_policy.go
```diff
@@ -264,9 +272,12 @@ func ipBlockToCIDRRule(block *slim_networkingv1.IPBlock) api.CIDRRule {
 }
 
 // parsePorts converts list of K8s NetworkPolicyPorts to Cilium PortRules.
-func parsePorts(ports []slim_networkingv1.NetworkPolicyPort) []api.PortRule {
+func parsePorts(ports []slim_networkingv1.NetworkPolicyPort) ([]api.PortRule, error) {
 	portRules := []api.PortRule{}
 	for _, port := range ports {
+		if port.EndPort != nil {
+			return portRules, ErrEndPortUnsupported
```
joamaki: The changes themselves look good, but they raised a few general questions regarding how we're handling NetworkPolicy. Not 100% sure they're something this PR needs to tackle...

Returning this error from here will cause this network policy parsing to be retried forever so the logs will be spammed with "Error while parsing k8s kubernetes NetworkPolicy".

Problem of course isn't really here, but in `K8sWatcher.addK8sNetworkPolicyV1`. It shouldn't be returning an error on parse failures since we don't want those to be forever retried.

Can we do something better than just log and forget here though? Would it make sense to set the NetworkPolicy's status to report that Cilium does not support this? Would a metric gauge be helpful to track how many policies are invalid?

joamaki (dismissed):

pkg/k8s/network_policy.go
joestringer: This has long been a pain point in validating network policies, and I agree it feels like we should be able to have a better solution. So far this is pretty much what we do (though I think usually it's in the `sanitize()` functions, not out here in the parse function).

Unfortunately updating networkpolicy status from arbitrary agents can introduce a scale issue by hammering the kube-apiserver from thousands of agents at once. We could consider moving some of this sort of validation into the operator, which I believe does not currently monitor for network policies. For a lot of the CNP validation we have been pushing it server-side into kube-apiserver... I'm not sure whether we could do that from Cilium for KNP as well, I think the validation is tied into resource registration. If we could change this for KNP, that would be the ideal solution --- prevent the user from even inserting the policy with these statements.

I like the metric idea.

joestringer (approved):
I think this makes sense, it's hard for me to argue for something better for the short term and the existing stable releases. I know the support for this feature is making some progress soon so hopefully we won't need a better solution to communicating this problem in the end. :+1:

joamaki (approved):

PR #28709, opened 2023-10-20
helm: add hubble UI support for GKE dataplane v2 

### LINKED ISSUE #28708: Helm chart does not support Hubble UI lon GKE Dataplane v2

### What happened?

After deploying a GKE cluster with [Dataplane v2](https://cloud.google.com/kubernetes-engine/docs/concepts/dataplane-v2) enabled and observability configured I am able to see the cilium CNI and hubble relay pods running.

When following the [documentation](https://docs.cilium.io/en/latest/gettingstarted/hubble/) to install helm standalone with the following values

```yaml
agent: false
operator:
  enabled: false
cni:
  install: false
hubble:
  enabled: false
  relay:
    enabled: false
    tls:
      server:
        enabled: true
  ui:
    enabled: true
    standalone:
      enabled: true
      tls:
        certsVolume:
          projected:
            defaultMode: 0400
            sources:
              - secret:
                  name: hubble-relay-client-certs
                  items:
                    - key: tls.crt
                      path: client.crt
                    - key: tls.key
                      path: client.key
                    - key: ca.crt
                      path: hubble-relay-ca.crt
```

Hubble UI is deployed into the cluster but when trying to access the service no flows are shown and the pod logs show the following error

```
hubble authentication handshake failed: tls: failed to verify certificate: x509: certificate is valid for *.kube-system.svc.cluster.local, not ui.hubble-relay.cilium.io
```

From the GCP configuration [docs](https://cloud.google.com/kubernetes-engine/docs/how-to/configure-dpv2-observability) the following environment variable needs to be set

```yaml
            - name: TLS_RELAY_SERVER_NAME
              value: relay.kube-system.svc.cluster.local
```

However this is hardcoded in the [helm chart](https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/templates/hubble-ui/deployment.yaml#L104-L105) to ` ui.hubble-relay.cilium.io`

### Cilium Version

- gke.gcr.io/cilium/hubble-relay:v1.12.10-gke.11
- gke.gcr.io/cilium/cilium:v1.12.10-gke.11

### Kernel Version

[GKE with Dataplane v2](https://cloud.google.com/kubernetes-engine/docs/concepts/dataplane-v2)

### Kubernetes Version

v1.27.3-gke.100

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

As a workaround manually editing the deployment to set 

```yaml
            - name: TLS_RELAY_SERVER_NAME
              value: relay.kube-system.svc.cluster.local
```

resolves the issue

### Code of Conduct

dwalker-sabiogroup:
To support GKE dataplane v2 the TLS_RELAY_SERVER_NAME backend environment variable needs to match the configure google managed DNS names (*.kube-system.svc.cluster.local)

This change retains the default behaviour but allows the environment variable to be configurable

      GitHub issue.

Fixes: https://github.com/cilium/cilium/issues/28708

### DIFF
--- a/install/kubernetes/cilium/templates/hubble-ui/deployment.yaml
+++ b/install/kubernetes/cilium/templates/hubble-ui/deployment.yaml
@@ -102,7 +102,7 @@ spec:
         - name: TLS_TO_RELAY_ENABLED
           value: "true"
         - name: TLS_RELAY_SERVER_NAME
-          value: ui.hubble-relay.cilium.io
+          value: {{ .Values.hubble.relay.tls.server.relayName }}
         - name: TLS_RELAY_CA_CERT_FILES
           value: /var/lib/hubble-ui/certs/hubble-relay-ca.crt
         - name: TLS_RELAY_CLIENT_CERT_FILE
--- a/install/kubernetes/cilium/values.yaml
+++ b/install/kubernetes/cilium/values.yaml
@@ -1331,6 +1331,12 @@ hubble:
         extraDnsNames: []
         # -- extra IP addresses added to certificate when its auto gen
         extraIpAddresses: []
+        # DNS name used by the backend to connect to the relay
+        # This is a simple workaround as the relay certificates are currently hardcoded to 
+        # *.hubble-relay.cilium.io 
+        # See https://github.com/cilium/cilium/pull/28709#discussion_r1371792546
+        # For GKE Dataplane V2 this should be set to relay.kube-system.svc.cluster.local
+        relayName: "ui.hubble-relay.cilium.io"
 
     # -- Dial timeout to connect to the local hubble instance to receive peer information (e.g. "30s").
     dialTimeout: ~
--- a/Documentation/helm-values.rst
+++ b/Documentation/helm-values.rst
@@ -1695,7 +1695,7 @@
    * - :spelling:ignore:`hubble.relay.tls`
      - TLS configuration for Hubble Relay
      - object
-     - ``{"client":{"cert":"","key":""},"server":{"cert":"","enabled":false,"extraDnsNames":[],"extraIpAddresses":[],"key":"","mtls":false}}``
+     - ``{"client":{"cert":"","key":""},"server":{"cert":"","enabled":false,"extraDnsNames":[],"extraIpAddresses":[],"key":"","mtls":false,"relayName":"ui.hubble-relay.cilium.io"}}``
    * - :spelling:ignore:`hubble.relay.tls.client`
      - base64 encoded PEM values for the hubble-relay client certificate and private key This keypair is presented to Hubble server instances for mTLS authentication and is required when hubble.tls.enabled is true. These values need to be set manually if hubble.tls.auto.enabled is false.
      - object
@@ -1703,7 +1703,7 @@
    * - :spelling:ignore:`hubble.relay.tls.server`
      - base64 encoded PEM values for the hubble-relay server certificate and private key
      - object
-     - ``{"cert":"","enabled":false,"extraDnsNames":[],"extraIpAddresses":[],"key":"","mtls":false}``
+     - ``{"cert":"","enabled":false,"extraDnsNames":[],"extraIpAddresses":[],"key":"","mtls":false,"relayName":"ui.hubble-relay.cilium.io"}``
    * - :spelling:ignore:`hubble.relay.tls.server.extraDnsNames`
      - extra DNS names added to certificate when its auto gen
      - list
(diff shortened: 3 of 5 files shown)


install/kubernetes/cilium/values.yaml
```diff
@@ -1286,6 +1286,9 @@ hubble:
         extraDnsNames: []
         # -- extra IP addresses added to certificate when its auto gen
         extraIpAddresses: []
+        # DNS name used by the backend to connect to the relay
+        # For GKE Dataplane V2 this should be set to relay.kube-system.svc.cluster.local
+        relayName: "ui.hubble-relay.cilium.io"
```
gandro, line 1339: Elsewhere, we called this `clusterDomain`, e.g. here:
https://github.com/cilium/cilium/blob/fd490f6bdef23ef0501687237b6ea721c5ff82e9/install/kubernetes/cilium/values.yaml.tmpl#L1056

It is then used together with the release namespace to build the service name as follows:
https://github.com/cilium/cilium/blob/692fe68d08090d6f8e9a890c3fcd9db475e31012/install/kubernetes/cilium/templates/hubble-relay/configmap.yaml#L15

I think it would make sense to keep the two options as similar as possible

gandro (changes requested):
Thanks! I've left some inline feedback.

Please also make sure to apply any changes to the values to the `values.yaml.tmpl`, as our `values.yaml` is auto-generated. You will have to run `make -C install/kubernetes` and `make -C Documentation` after adding new Helm values, to also update the docs.

install/kubernetes/cilium/values.yaml
dwalker-sabiogroup: @gandro - I have run the `make` commands provided and aligned the key name as suggested which I think is what you meant in the comment above but if not please let me know and I can amend.

install/kubernetes/cilium/templates/hubble-ui/deployment.yaml
```diff
@@ -102,7 +102,7 @@ spec:
         - name: TLS_TO_RELAY_ENABLED
           value: "true"
         - name: TLS_RELAY_SERVER_NAME
-          value: ui.hubble-relay.cilium.io
+          value: {{ .Values.hubble.relay.tls.server.clusterDomain }}
```
gandro: Almost. We should only use the clusterDomain for the cluster domain part. The rest should be derived differently:

```suggestion
          value: relay.{{ .Release.Namespace }}.svc.{{ Values.hubble.relay.tls.server.clusterDomain }}
```

I'm not sure if the default TLS name (`ui.hubble-relay.cilium.io`) is even correct. I'll ping some Hubble folks regarding this.

gandro (commented):
Thanks! This is looking much better already, but the clusterDomain should only be used for the cluster domain.

Also, I believe you need to run `make -C Documentation helm-values.rst` as well

install/kubernetes/cilium/templates/hubble-ui/deployment.yaml
gandro: Okay, so it seems our Relay server certificates are currently hard-coded to use `*.hubble-relay.cilium.io`:
https://github.com/cilium/cilium/blob/44698dfecdd65788c621c705581cbf92ec46e3fd/install/kubernetes/cilium/templates/hubble/tls-helm/relay-server-secret.yaml#L3

That makes it impossible to use the `clusterDomain`, because we will want to have backwards-compatibility with existing certs. On the other hand, I'm also not a  fan of introducing a value for GKE datapath v2 that cannot be used with Cilium OSS.

For now, I'm fine accepting a patch which basically is the same as your original PR: A simple Helm value that only works on GKE datpath v2. It should be clear that this is a workaround and might be removed in future versions once we have a better solution.

Long-term however, I think a proper solution would be to fix up our all our certificate templates (`tls-cronjob`, `tls-helm`, `tls-certmanager`) to assemble the CommonName of the certificates from a `clusterDomain` and namespace, with some option to override it too.

dwalker-sabiogroup: @gandro - I have reverted the changes back to a simple key/value that can be customised and run the relevant make commands to generate the docs.  Let me know if there is anything else that should be changed.

gandro (commented):
Looks good, thanks for you patience on this!

Could you please squash all commits into one (don't forget add the sign-off), then this should be good to go. Thanks!

install/kubernetes/cilium/templates/hubble-ui/deployment.yaml
chancez: We could pass this to `tpl` so you can use template vars.


The template:
```
        - name: TLS_RELAY_SERVER_NAME
          value: {{ .Values.hubble.relay.tls.server.clusterDomain | tpl }}
```

The override:

eg:
```
clusterDomain: "relay.{{ .Release.Namespace }}.svc.cluster.local"
```

> Long-term however, I think a proper solution would be to fix up our all our certificate templates (tls-cronjob, tls-helm, tls-certmanager) to assemble the CommonName of the certificates from a clusterDomain and namespace, with some option to override it too.

I agree, but it'll take a motivated contributor to fix it all, as it's got a lot of gotchas and compatibility considerations to take into account.

michi-covalent (approved):

dwalker-sabiogroup:
@gandro - Commits squashed - should be good but let me know if not.

gandro (approved):
Great, thank you!

chancez (approved):

joamaki (approved):

michi-covalent:
looks like the gateway tests are failing because the branch is missing `examples/kubernetes/gateway/grpc-route.yaml` which got added recently. we probably need to rebase 👀

PR #28710, opened 2023-10-20
bpf: ensure inner-map flags are updated on map create

ldelossa:
Fixes an issue experienced when creating nested eBPFs map.

In this commit the OpenOrCreate function is updated to also set flags appropriately for the inner map.

The failure to do this may lead to a meta data check in the kernel failing when inserting an inner map created with the OpenOrCreate function into an outer map.

This may occur when you define a nested map with an inner map spec which did not take the conditional flag setting, which occurs in OpenOrCreate, into account.

```release-note
bpf: fixes an issue where inserting inner maps into an outer may fail with EINVAL due to flags mismatch 
```


ti-mo (approved):
Thanks!

PR #28711, opened 2023-10-20
datapath: alignchecker: allow to extend toCheck and toCheckSizes

jibi:
make toCheck and toCheckSizes global variables and add a couple of setters (registerToCheck and registerToCheckSizes) so that other implementations can extend these maps for example from init()


ldelossa (approved):
LGTM

Issue #24481, opened 2023-03-21
NodePort Services do not work with new Node addresses


aojea:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

A Service NodePort type exposes the Pods backends via a NodePort on the node addresses, forwarding any traffic from the frontend given by NodeAddresses:NodePort.

The key here is the concept of Node Addresses, after a quick skim to the code, it seems that cilium only consider the Node Addresses that are associated to the Service (ExternalIP, LoadBalancerIP) or the ones associated to the node object Node.Status.InternalIP or ExternalIP, but there are situations that a node can get new IPs and those are not reflected in Kubernetes. Advanced networking kubernetes users use to have more complex setups with multiple NICs, and they need to receive all the NodePort traffic on all the nics, or. just some subset of them.

Kube-proxy has a flag that allow users to define the range of addresses they want to specify to use for NodePorts, it defaults to all local address, and everytime that a new address is added to the host, is considered for the NodePort if it matches the range

```
--nodeport-addresses strings
  | A string slice of values which specify the addresses to use for NodePorts. Values may be valid IP blocks (e.g. 1.2.3.0/24, 1.2.3.4/32). The default empty string slice ([]) means to use all local addresses. This parameter is ignored if a config file is specified by --config.
```

I expect cilium to behave the same as kube-proxy, so new addresses on the nodes can be used to receive NodePort traffic, and also have the capacity to define which ranges of the local addresses should be allowed.

### How to repro

You can repro with GKE: create a NodePort service and manually create a GCE loadbalancer.
GCE load balancers add the LB frontend IP to the `loopback` interface of all the nodes, and then route the traffic to that IP through the nodes.
 The observation is that all the healthchecks directed to the GCE LB IP and the NodePort are dropped, , if my intuition is correct, it seems this new IP is not considered by cilium, hence it will not forward the traffic to it.

However, it is really easy to repo with Kind

1. Create a kind cluster with cilium
```
$ KUBEPROXY_MODE=none contrib/scripts/kind.sh
$ make kind-image
$ cilium install --wait --chart-directory=install/kubernetes/cilium \
            --helm-set=image.useDigest=false \
            --helm-set=clustermesh.apiserver.image.useDigest=false \
            --helm-set-string=kubeProxyReplacement=strict \
            --helm-set=sessionAffinity=true \
            --rollback=false \
            --config monitor-aggregation=none \
            --version=
```
3. Create a nodeport service with working backends
```
[...]

PR #28712, opened 2023-10-20
[DRAFT]: NodeAddress table and refactored NodeAddressing

joamaki:
Add `Table[NodeAddress]` which derives from the low-level `Table[*Device]` to pick which IP addresses are considered host addresses and used for NodePort and BPF masquerading.

Refactor NodeAddressing to look up IP addresses from the `Table[NodeAddress]` and K8s related information from LocalNodeStore.

To add support for multiple NodePort IP addresses per network device, this PR adds the flag `--nodeport-addresses` that
allows user to specify which CIDRs the NodePort frontends can be picked from. This allows us to solve e.g. #24481.


joamaki:
Ended up implementing the `Table[NodeAddress]`. This is essentially the same information as was stored in `ipv*NodePortAddrs`.

Redid `LocalAddresses` by extracting global addresses from all devices as the original code did with `netlink.AddrList`. It'd probably be a good idea to think whether it'd be possible to merge `NodePortAddresses` and `LocalAddresses` to simplify things overall, but hard to judge at this point what the impact would be, so this PR keeps things semantically the same.

joamaki:
Closing as this is being pulled into multiple individual PRs.

PR #28713, opened 2023-10-20
Modularize kernel modules manager into its own cell

pippolo84:
Modularize the kernel modules manager into its own cell.

Instead of exporting an Init() function, the manager stores the list of loaded modules once at startup.

The tests are updated accordingly and since we are at it, migrate them from checkmate (the temporary wrapper for gopkg.in/check.v1) to the standard Go testing framework.

This is a first step needed to modularize the iptables manager.


YutaroHayakawa (approved):
LGTM 👍

joamaki (changes requested):
Oops, reviewed this in another PR that depended on this. See comments there: https://github.com/cilium/cilium/pull/28746

pippolo84:
@joamaki I've addressed the feedback regarding modules manager here. PTAL.

pkg/datapath/cells.go
```diff
@@ -52,6 +53,10 @@ var Cell = cell.Module(
 	// The monitor agent, which multicasts cilium and agent events to its subscribers.
 	monitorAgent.Cell,
 
+	// The modules manager to search and load kernel modules.
+	// Must appear before the datapath so that its start hook is registered before the iptables one.
```
joamaki: This comment about ordering is not true. Since `newDatapath` depends on the modules manager it will be started before iptables. Also since `modules.Cell` does not have an invoke, it doesn't matter where it is here since the order only matters for invokes.

pkg/datapath/linux/modules/modules.go
```diff
@@ -50,3 +45,11 @@ func (m *ModulesManager) FindOrLoadModules(expectedNames ...string) error {
 	}
 	return nil
 }
+
+func newManager(lc hive.Lifecycle) *Manager {
+	mm := Manager{}
+
+	lc.Append(hive.Hook{OnStart: mm.Start})
```
joamaki: Hmm, I'm not sure what this does. `mm` is `Manager` and not `*Manager`, so there might be a copy happening here. Please do `m := &Manager{}` instead. And please call it `m` instead of `mm` to match method receivers.

pkg/datapath/linux/modules/modules_test.go
```diff
@@ -0,0 +1,23 @@
...
+package modules
+
+import (
+	"context"
+	"testing"
+
+	"github.com/cilium/cilium/pkg/hive"
+)
+
+func TestInit(t *testing.T) {
```
joamaki: This test doesn't seem to do anything useful. `Cell` only has a single `Provide` for an object that is never used by the test, thus it's never constructed. I would just delete this.

joamaki (changes requested):

pkg/datapath/cells.go
pippolo84: Got confused with `Invoke` and `OnStart`, I'm gonna fix this :+1:

pkg/datapath/linux/modules/modules.go
pippolo84: Good catch.

pkg/datapath/cells.go
joamaki: Thanks! :smiley:

joamaki (approved):

pippolo84:
ci-clustermesh failure tracked [here](https://github.com/cilium/cilium/issues/27762#issuecomment-1781172003), rerunning

dylandreimerink (approved):

pippolo84:
@aditighag ping :pray:

aditighag (approved):

Issue #27955, opened 2023-09-05
CFP: mac spoofing protection


shen-cloud:
## Cilium Feature Proposal
        
Thanks for taking time to make a feature proposal for Cilium! If you have usage questions, please try the [slack channel](http://slack.cilium.io/) and see the [FAQ](https://github.com/cilium/cilium/issues?utf8=%E2%9C%93&q=is:issue+label:kind/question+) first.

**Is your proposed feature related to a problem?**

In current implementation, there is no mac spoofing protection. A malicious pod in a Kubernetes cluster can use the ARP protocol to spoof the IP address of another pod on the same node. This causes other pods on the node to talk to the attacker's pod instead of the legitimate pod. With the proposal, any traffic that exits from the pod should match the source mac address and arp responses should match the assigned IP and mac address to protect from mac spoofing.

**Describe the feature you'd like**

1. source mac of traffic from the container should match pod-side veth mac
2. arp response from the container should match pod-side veth IP and mac

**(Optional) Describe your proposed solution**

1. add feature flag for mac spoof check 
2. expose pod-side veth mac to bpf code through macro like NODE_MAC
3. check source mac address for the traffic in from-container
4. check arp response in from-container

For longer proposals, you are welcome to link to an external doc (e.g. a Google doc). We have a [Cilium Feature Proposal template](https://docs.google.com/document/d/1vtE82JExQHw8_-pX2Uhq5acN1BMPxNlS6cMQUezRTWg/edit) to help you structure your proposal - if you would like to use it, please make a copy and ensure it's publicly visible, and then add the link here.

Once the CFP is close to being finalized, please add it as a PR to the [design-cfps](https://github.com/cilium/design-cfps) repo for final approval.

shen-cloud:
@aojea

youngnick:
Thanks for logging this request @shen-cloud.

I don't think it will be actionable without some more information though - currently the justification section is not complete.

> In current implementation, there is no mac spoofing protection. It adds security concerns when ENABLE_ARP_PASSTHROUGH is present since ARP will be

If you could please update this with some more detail, including explaining more _why_ you feel that this creates security concerns, and how your proposed approach will manage those concerns, then we may be able to proceed.

shen-cloud:
Updated the justification section. Please let me know if you need more clarifications.

youngnick:
[...]

PR #28715, opened 2023-10-20
CFP: mac spoofing protection #27955

shen-cloud:
1. add feature flag for mac spoof check
2. expose pod-side veth mac to bpf code with new macro LXC_MAC
3. check source mac address for the traffic in from-container
4. check arp response in from-container
5. introduce new drop reason DROP_INVALID_SRC_MAC

### DIFF
--- a/bpf/bpf_lxc.c
+++ b/bpf/bpf_lxc.c
@@ -1272,6 +1272,9 @@ static __always_inline int __tail_handle_ipv4(struct __ctx_buff *ctx,
 		return DROP_FRAG_NOSUPPORT;
 #endif
 
+	if (unlikely(!is_valid_lxc_src_mac(ctx)))
+		return DROP_INVALID_SMAC;
+
 	if (unlikely(!is_valid_lxc_src_ipv4(ip4)))
 		return DROP_INVALID_SIP;
 
@@ -1308,6 +1311,9 @@ int tail_handle_arp(struct __ctx_buff *ctx)
 	union macaddr smac;
 	__be32 sip;
 	__be32 tip;
+
+	if (arp_validate_mac_spoof(ctx) < 0)
+		return DROP_INVALID_SMAC;
 
 	/* Pass any unknown ARP requests to the Linux stack */
 	if (!arp_validate(ctx, &mac, &smac, &sip, &tip))
@@ -1370,7 +1376,7 @@ int cil_from_container(struct __ctx_buff *ctx)
 		break;
 #ifdef ENABLE_ARP_PASSTHROUGH
 	case bpf_htons(ETH_P_ARP):
-		ret = CTX_ACT_OK;
+		ret = arp_validate_mac_spoof(ctx);
 		break;
 #elif defined(ENABLE_ARP_RESPONDER)
 	case bpf_htons(ETH_P_ARP):
--- a/pkg/testutils/endpoint.go
+++ b/pkg/testutils/endpoint.go
@@ -24,6 +24,7 @@ type TestEndpoint struct {
 	Identity *identity.Identity
 	Opts     *option.IntOptions
 	MAC      mac.MAC
+	LXCMAC   mac.MAC
 	IPv6     netip.Addr
 	isHost   bool
 }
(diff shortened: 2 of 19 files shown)


joestringer:
Hi, thanks for the PR! Can you talk a bit about why this is a useful feature?

I was under the understanding that all L2 addresses are replaced by Cilium, so regardless of ARP spoofing it doesn't matter because Cilium will substitute the MAC corresponding to the endpoint:

https://github.com/cilium/cilium/blob/58b67dc281544ff463d4c92c0fe5444feebf731b/bpf/lib/l3.h#L63

akhilles:
@shen-cloud can you link the CFP issue in the PR description: https://github.com/cilium/cilium/issues/27955?

shen-cloud:
> Hi, thanks for the PR! Can you talk a bit about why this is a useful feature?
> 
> I was under the understanding that all L2 addresses are replaced by Cilium, so regardless of ARP spoofing it doesn't matter because Cilium will substitute the MAC corresponding to the endpoint:
> 
> https://github.com/cilium/cilium/blob/58b67dc281544ff463d4c92c0fe5444feebf731b/bpf/lib/l3.h#L63

The smac rewrite happens when `ENABLE_ROUTING` is defined. If the endpoint BPF Routing is not enabled, the feature would be useful to protect from mac spoof. It's also useful to validate ARP response in `ENABLE_ARP_PASSTHROUGH` mode. Added the original CFP link in the description of the PR.

pkg/option/config.go
```diff
@@ -3230,6 +3238,7 @@ func (c *DaemonConfig) Populate(vp *viper.Viper) {
 	c.EnablePMTUDiscovery = vp.GetBool(EnablePMTUDiscovery)
 	c.IPv6NAT46x64CIDR = defaults.IPv6NAT46x64CIDR
 	c.IPAMCiliumNodeUpdateRate = vp.GetDuration(IPAMCiliumNodeUpdateRate)
+	c.EnableMacSpoofCheck = vp.GetBool(EnableMacSpoofCheck)
```
akhilles: Do we need a config flag? Can this be enabled by default instead? Maybe follow the same pattern as SIP verification: add a field to `DatapathConfiguration` called `DisableSmacVerification` that's set to `false` by default.

pkg/testutils/endpoint.go
```diff
@@ -64,6 +66,7 @@ func (e *TestEndpoint) GetIdentity() identity.NumericIdentity       { return e.I
 func (e *TestEndpoint) GetIdentityLocked() identity.NumericIdentity { return e.Identity.ID }
 func (e *TestEndpoint) GetSecurityIdentity() *identity.Identity     { return e.Identity }
 func (e *TestEndpoint) GetNodeMAC() mac.MAC                         { return e.MAC }
+func (e *TestEndpoint) LXCMac()() mac.MAC                           { return e.LXCMAC }
```
akhilles:
```suggestion
func (e *TestEndpoint) LXCMac() mac.MAC                             { return e.LXCMAC }
```

bpf/bpf_lxc.c
```diff
@@ -1272,6 +1272,11 @@ static __always_inline int __tail_handle_ipv4(struct __ctx_buff *ctx,
 		return DROP_FRAG_NOSUPPORT;
 #endif
 
+#ifdef ENABLE_MAC_SPOOF_CHECK
```
akhilles: Use `DISABLE_SMAC_VERIFICATION` instead? to be consistent with `DISABLE_SIP_VERIFICATION`.

joestringer: As a general rule, using `ENABLE_FOO` is much easier to reason about from a developer perspective than `DISABLE_FOO`.

pkg/option/config.go
shen-cloud: given the nature of this security feature, I think turn it on by default with an option to disable works for me. But tagging @joestringer to weight in on this decision.

shen-cloud:
@joestringer could you please trigger the test on this patch and review when you get a chance?

NOBLES5E:
Is this pull request still under review?

joestringer:
@NOBLES5E not currently. If @shen-cloud or you or anyone wishes to pick this back up, rebase against the latest main branch and repost the PR, then we could proceed with further review. My guess given the previous activity is that the PR was not passing the tests but I can't tell for sure.

PR #28717, opened 2023-10-22
bpf: ipv4: refactor L4 port extraction for fragmented packets

julianwiedmann:
Untangle the fragmentation helper from `conntrack.h`, and allow callers to pass a validated L3 header (inspired by https://github.com/cilium/cilium/pull/25340#discussion_r1342580917).

### DIFF
--- a/bpf/lib/ipv4.h
+++ b/bpf/lib/ipv4.h
@@ -7,6 +7,7 @@
 #include <linux/ip.h>
 
 #include "dbg.h"
+#include "l4.h"
 #include "metrics.h"
 
 #define IPV4_SADDR_OFF		offsetof(struct iphdr, saddr)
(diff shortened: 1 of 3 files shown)


bpf/lib/ipv4.h
```diff
@@ -181,4 +182,31 @@ ipv4_handle_fragmentation(struct __ctx_buff *ctx,
 }
 #endif
 
+static __always_inline int
+ipv4_load_l4_ports(struct __ctx_buff *ctx, struct iphdr *ip4 __maybe_unused,
+		   int l4_off, enum ct_dir dir __maybe_unused,
+		   __be16 *ports, bool *has_l4_header __maybe_unused)
+{
+#ifdef ENABLE_IPV4_FRAGMENTS
+	if (!ip4) {
```
gentoo-root, line 191: This one looks a bit odd, I would prefer it to be responsibility of the caller to pass a non-NULL ip4, but on the other hand some callers don't have ip4 ready, and doing revalidate_data would be inefficient if fragments are disabled; wrapping it with `#ifdef` exposes implementation details. Trying to pass ip4 from up above seems to be a long chain. So I guess your version is the cleanest anyway. LGTM.

gentoo-root (approved):

bpf/lib/ipv4.h
julianwiedmann: Heh, I had the exact same thinking. I've been slowly chipping away at all the hidden `revalidate_data()` calls, so I'm confident that we'll get this into a cleaner shape eventually.

PR #28723, opened 2023-10-23
bpf: improve trace events in SNAT path

julianwiedmann:
```release-note
Enhance trace events from the outbound SNAT path, to report the pre-SNAT IP address and the interface index of the egress interface.
```

### DIFF
--- a/bpf/lib/nodeport.h
+++ b/bpf/lib/nodeport.h
@@ -174,6 +174,7 @@ nodeport_has_nat_conflict_ipv6(const struct ipv6hdr *ip6 __maybe_unused,
 }
 
 static __always_inline int nodeport_snat_fwd_ipv6(struct __ctx_buff *ctx,
+						  union v6addr *saddr,
 						  struct trace_ctx *trace,
 						  __s8 *ext_err)
 {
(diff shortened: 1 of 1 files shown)


kaworu:
@julianwiedmann should we add a note in `tail_handle_snat_fwd_ipv6` to also populate `saddr` and use `send_trace_notify6` when egress gateway IPv6 support is implemented? If yes, maybe [here](https://github.com/cilium/cilium/blob/e041100c51bc1e2a8598989390f4fa8985d8864d/bpf/lib/nodeport.h#L1541-L1548)?

julianwiedmann:
> @julianwiedmann should we add a note in `tail_handle_snat_fwd_ipv6` to also populate `saddr` and use `send_trace_notify6` when egress gateway IPv6 support is implemented? If yes, maybe [here](https://github.com/cilium/cilium/blob/e041100c51bc1e2a8598989390f4fa8985d8864d/bpf/lib/nodeport.h#L1541-L1548)?

I'll need to refresh my memory, but pretty sure that we can wire up the IPv6 path even today :).

We'll populate the `orig-saddr` for all kinds of SNAT operations, not just EgressGW. So for instance plain IPv6 Masquerading in BPF, where we would report the pod's IP in the trace event.

julianwiedmann:
@kaworu I rebased and added the IPv6 handling. Also making sure that the `orig_addr` is only set when we actually SNAT the packet.

Would you have a moment to double-check this still works as expected?

jibi (approved):

sayboras:
:wave: 

Seems like we might need some extra commits from https://github.com/cilium/cilium/pull/26674 for 1.14 backport.

PR #28732, opened 2023-10-23
daemon,pkg/service: Use hive cell infra for pkg/service

rastislavs:
This PR uses hive cell infra for wiring the `pkg/service` in the Daemon. This allows for easier injection of the `pkg/service` dependency in other modules in the future. 

Also defines the `ServiceManager` interface for the `pkg/service.Service`, to allow for easier mocking of this dependency.

As it was necessary to avoid an import cycle, this also moves `endpointsynchronizer` from `pkg/k8s/watchers` to `pkg/endpointmanager` so that `endpointmanager` does not need to import the top-level `pkg/k8s/watchers`.

### DIFF
--- a/pkg/service/cell.go
+++ b/pkg/service/cell.go
@@ -1,0 +1,31 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package service
+
+import (
+	"github.com/cilium/cilium/pkg/datapath/types"
+	"github.com/cilium/cilium/pkg/hive/cell"
+	monitorAgent "github.com/cilium/cilium/pkg/monitor/agent"
+	"github.com/cilium/cilium/pkg/proxy"
+)
+
+// Cell provides access to the Service Manager.
+var Cell = cell.Module(
+	"service-manager",
+	"Service Manager",
+
+	cell.Provide(newServiceManager),
+)
+
+type serviceManagerParams struct {
+	cell.In
+
+	L7Proxy      *proxy.Proxy
+	Datapath     types.Datapath
+	MonitorAgent monitorAgent.Agent
+}
+
+func newServiceManager(params serviceManagerParams) ServiceManager {
+	return NewService(params.MonitorAgent, params.L7Proxy, params.Datapath.LBMap())
+}
--- a/pkg/service/service.go
+++ b/pkg/service/service.go
@@ -30,6 +30,7 @@ import (
 	"github.com/cilium/cilium/pkg/logging/logfields"
 	"github.com/cilium/cilium/pkg/maps/lbmap"
 	"github.com/cilium/cilium/pkg/metrics"
+	monitorAgent "github.com/cilium/cilium/pkg/monitor/agent"
 	monitorAPI "github.com/cilium/cilium/pkg/monitor/api"
 	"github.com/cilium/cilium/pkg/node"
 	nodeTypes "github.com/cilium/cilium/pkg/node/types"
@@ -71,11 +72,6 @@ func (e *ErrLocalRedirectServiceExists) Is(target error) bool {
 type healthServer interface {
 	UpsertService(svcID lb.ID, svcNS, svcName string, localEndpoints int, port uint16)
 	DeleteService(svcID lb.ID)
-}
-
-// monitorNotify is used to send update notifications to the monitor
-type monitorNotify interface {
-	SendNotification(msg monitorAPI.AgentNotifyMessage) error
 }
 
 // envoyCache is used to sync Envoy resources to Envoy proxy
@@ -248,9 +244,9 @@ type Service struct {
 	// not for loadbalancing decisions.
 	backendByHash map[string]*lb.Backend
 
-	healthServer  healthServer
-	monitorNotify monitorNotify
-	envoyCache    envoyCache
+	healthServer healthServer
+	monitorAgent monitorAgent.Agent
+	envoyCache   envoyCache
 
 	lbmap         datapathTypes.LBMap
 	lastUpdatedTs atomic.Value
@@ -261,7 +257,7 @@ type Service struct {
 }
 
 // NewService creates a new instance of the service handler.
-func NewService(monitorNotify monitorNotify, envoyCache envoyCache, lbmap datapathTypes.LBMap) *Service {
+func NewService(monitorAgent monitorAgent.Agent, envoyCache envoyCache, lbmap datapathTypes.LBMap) *Service {
 
 	var localHealthServer healthServer
 	if option.Config.EnableHealthCheckNodePort {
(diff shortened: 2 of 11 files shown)


pkg/service/service.go
```diff
@@ -2104,3 +2100,11 @@ func (s *Service) SyncServicesOnDeviceChange(nodeAddressing types.NodeAddressing
 		}
 	}
 }
+
+// sendNotification sends an agent notification to the monitor
+func (s *Service) sendNotification(notification monitorAPI.AgentNotifyMessage) error {
+	if option.Config.DryMode {
```
joamaki: Not sure if need `DryMode` here. The monitor agent cell has an optional dependency to the events map and thus works fine in non-privileged integration tests.

joamaki:
Since `pkg/service/api` imports `pkg/service` I don't quite understand how this avoids import cycles?
We've also usually used `api` package for things related to the UNIX socket REST API and used `types` package for interfaces and types. 

WDYT about moving `pkg/service/api/cell.go` to `pkg/service/cell.go`, renaming `api` to `types` and keeping the interface type there so it can be used in tests without pulling in all the dependencies of `pkg/service`?

pkg/service/api/cell.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"github.com/cilium/cilium/pkg/datapath/types"
+	"github.com/cilium/cilium/pkg/hive/cell"
+	monitorAgent "github.com/cilium/cilium/pkg/monitor/agent"
+	"github.com/cilium/cilium/pkg/proxy"
+	"github.com/cilium/cilium/pkg/service"
+)
+
+// ServiceManagerCell provides access to the datapath Service Manager.
+var ServiceManagerCell = cell.Module(
```
joamaki: I would just call this `Cell` and use it as `service.Cell` (if you follow my other comment and move it into pkg/service).

joamaki:
> @joamaki
> 
> > Since `pkg/service/api` imports `pkg/service` I don't quite understand how this avoids import cycles? We've also usually used `api` package for things related to the UNIX socket REST API and used `types` package for interfaces and types.
> > WDYT about moving `pkg/service/api/cell.go` to `pkg/service/cell.go`, renaming `api` to `types` and keeping the interface type there so it can be used in tests without pulling in all the dependencies of `pkg/service`?
> 
> Yeah I have tried exactly that at first, but this is what I am getting:
> 
> ```
> 2.577 package github.com/cilium/cilium/daemon
> 2.577 	imports github.com/cilium/cilium/daemon/cmd
> 2.577 	imports github.com/cilium/cilium/pkg/auth
> 2.577 	imports github.com/cilium/cilium/pkg/endpointmanager
> 2.577 	imports github.com/cilium/cilium/pkg/k8s/watchers
> 2.577 	imports github.com/cilium/cilium/pkg/redirectpolicy
> 2.577 	imports github.com/cilium/cilium/pkg/service
> 2.577 	imports github.com/cilium/cilium/pkg/proxy
> 2.577 	imports github.com/cilium/cilium/pkg/proxy/logger/endpoint
> 2.577 	imports github.com/cilium/cilium/pkg/endpointmanager: import cycle not allowed
> 2.584 make[1]: Leaving directory '/go/src/github.com/cilium/cilium/daemon'
> ```

Right, `*proxy.Proxy` causes the cycle and it was worked around earlier by defining an interface in pkg/service that was a subset of what it provided. Looking at the imports, I think probably silliest import here is the `pkg/endpointmanager` importing `pkg/k8s/watchers` as `watchers` is a very top-level thing that's supposed to be calling downstream to other components. Not sure what I was thinking :smile:. So that's probably what I would fix first and see if it's enough to fix the problem. Since the `EndpointSynchronizer` is not dependent on endpoint manager nor on `K8sWatcher`, I would probably just move its construction into its own cell (perhaps just a cell.Provide in the `endpointmanager.Cell`) and move the code under `pkg/endpointmanager` (or `pkg/endpointmanager/synchronizer`?).

rastislavs:
@joamaki 
> Since the EndpointSynchronizer is not dependent on endpoint manager nor on K8sWatcher, I would probably just move its construction into its own Provide (perhaps part of endpointmanager.Cell) and move the code from pkg/k8s/watchers to pkg/endpointmanager

Thanks a lot for the suggestion, this resolved it.

joamaki (approved):

pkg/service/cell.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/cilium/cilium/pkg/datapath/types"
+	"github.com/cilium/cilium/pkg/hive/cell"
+	monitorAgent "github.com/cilium/cilium/pkg/monitor/agent"
+	"github.com/cilium/cilium/pkg/proxy"
+)
+
+// Cell provides access to the datapath Service Manager.
+var Cell = cell.Module(
+	"datapath-service-manager",
```
aditighag: nit: Just `service manager` (`load-balancer-service-manager` might've been better, but there is a load-balancer service type in k8s so it may create confusion)

pkg/service/types.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"time"
+
+	"github.com/cilium/cilium/pkg/datapath/types"
+	"github.com/cilium/cilium/pkg/k8s"
+	lb "github.com/cilium/cilium/pkg/loadbalancer"
+	"github.com/cilium/cilium/pkg/lock"
+)
+
+// ServiceManager is a datapath service handler. Its main responsibility is to reflect
```
aditighag: Suggestion: `ServiceManager` provides an interface for service related operations, and it is implemented by ...

pkg/service/cell.go
```diff
@@ -0,0 +1,31 @@
...
+	"Datapath Service Manager",
+
+	cell.Provide(newServiceManager),
```
aditighag, line 18: Is there a shutdown handler you can register that's reciprocal to `cell.Provide(newServiceManager)`?

aditighag (changes requested):
>daemon,pkg/service: Use hive cell infra for pkg/service.Service

Thanks! Much needed refactoring change that'll simplify `pkg/service` lifecycle management and dependencies. 

(I didn't review the first commit, but I would suggest adding "Suggested-by: Jussi..." tag to the commit description per OSS guidelines.)

pkg/service/cell.go
rastislavs: I don't think so - why would we need it?

What is possible though is to implement [lifecycle hooks](https://docs.cilium.io/en/latest/contributing/development/hive/#lifecycle) (`Start` and `Stop`) for the Service if it is needed. In the current code I have not found a need for them, but they can be added at any time.

pkg/service/cell.go
rastislavs: thx, renamed

pkg/service/types.go
rastislavs: thx, reworded

aditighag (approved):
🚢

aditighag (approved):
Hmm... It looks like GH didn't recognize Jussi and my reviews for the code owners after you moved the PR out of draft. 
I''ll try approving the changes again in case that does the trick.

joamaki (approved):

PR #28740, opened 2023-10-23
Always migrate cilium_calls_* during ELF load

### LINKED ISSUE #20691: Always migrate tail call maps

ti-mo:
Note: this is obviously a hack. There's a long-standing issue with Cilium's
loader infrastructure where program arrays like cilium_calls_* are reused if
the map's properties in the ELF are compatible with the pinned version present
on the system. This has been the case since the early iproute2 days, and
ebpf-go has the same behaviour by default. This is unsound, but doesn't always
cause issues in practice.

However, as Cilium's eBPF code base grew, along with its contributor base, and
increasing code churn and the need to backport fixes, this method started
falling apart. Pinned tail call maps are actively used for handling traffic,
and should be considered fixed, as they are an integral part of the program's
control flow. When another BPF object needs to be attached to a kernel hook,
it's incorrect to replace entries in an existing program array used by a prior
version of the BPF object, since logic in some tail calls may have been enabled
or disabled in the new one, or may have been removed altogether.

A prog array's contents cannot be replaced atomically, meaning that while the
prog array is being repopulated by a new ELF, a packet may be processed by a
chain of programs that are partially old and partially new. This is bad.

This patch fixes the specific case of cilium_calls_*, since it's the only
prog array across all ELFs that consistently acts as the central tail call map.
The main motivation for this approach is backportability as far back as 1.12,
since we're seeing increased connectivity interruptions (leading to test flakes)
during up/downgrades and restarts in that branch. This is caused by other
backports creating code churn, as well as the stark difference in the contents
of tail calls between major Cilium versions.

This code will be replaced by a 'commit' system that does away with the re-
pinning behaviour and keeps new map fd's in memory; only committing them to
bpffs when the necessary endpoint prog(s) were successfully replaced.

Fixes: https://github.com/cilium/cilium/issues/20691

```release-note
Always replace the cilium_call_* tail call map during upgrade/restart to avoid "Missed tail call" errors
```

### DIFF
--- a/pkg/datapath/loader/netlink.go
+++ b/pkg/datapath/loader/netlink.go
@@ -71,7 +71,7 @@ type progDefinition struct {
 // For example, this is the case with from-netdev and to-netdev. If eth0:to-netdev
 // gets its program and maps replaced and unpinned, its eth0:from-netdev counterpart
 // will miss tail calls (and drop packets) until it has been replaced as well.
-func replaceDatapath(ctx context.Context, ifName, objPath string, progs []progDefinition, xdpMode string) (func(), error) {
+func replaceDatapath(ctx context.Context, ifName, objPath string, progs []progDefinition, xdpMode string) (_ func(), err error) {
 	// Avoid unnecessarily loading a prog.
 	if err := ctx.Err(); err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


margamanterola:
I have no opinions on the code, but regarding the release notes, remember that the readers of the release notes are our users, that don't necessarily understand the inner workings of Cilium. Particularly since this is tagged "bug", it's important to let users know what kind of bug is getting fixed and when they would be affected by it.

I suggest something like:
Avoid "Missed tail call" errors on upgrades and downgrades by replacing the cilium_call_* tail call map instead of modifying the existing one.

ti-mo:
> I suggest something like: Avoid "Missed tail call" errors on upgrades and downgrades by replacing the cilium_call_* tail call map instead of modifying the existing one.

Good idea, thanks :+1:

pkg/datapath/loader/netlink.go
```diff
@@ -98,6 +98,33 @@ func replaceDatapath(ctx context.Context, ifName, objPath string, progs []progDe
...
+
+		if err := bpf.RepinMap(bpf.TCGlobalsPath(), key, ms); err != nil {
+			return nil, fmt.Errorf("repinning map %s: %w", key, err)
+		}
+
+		defer func() {
+			revert := false
+			// This captures named return variable err.
+			if err != nil {
+				revert = true
```
rgo3, line 116: I'm not entirely sure if that is possible but what happens if we need to revert the unconditional migration on a cilium_calls_* map but the map has been recreated and pinned through `bpf.LoadCollection` already? AFAIU `bpf.FinalizeMap()` will fail and we log an error but is there any danger when this happens?

ti-mo: That's the change I wanted to make while you were reviewing today; `FinalizeMap()` now does an `os.Remove(dest)` in the revert path so the rollback can't reasonably fail.

rgo3 (approved):
Thanks again for the code walk :)

pchaigno:
Discussed offline with Timo. We won't backport to v1.12 as it's too big of a backport and thus risky. Instead, we'll skip this specific error in v1.12 CI (and v1.13 downgrade tests).

PR #20090, opened 2022-06-06
k8s: allow setting multiple k8s API server addresses

fristonio:
See commit message for detailed description

Introduce a new command line parameter(`--k8s-api-server-urls`) and helm option(`k8s.apiServerURLs`) to specify multiple k8s API server addresses for the client to use.

Fixes: #19038 

```release-note
Allow setting of multiple k8s API server URLs for client
```

PR #28741, opened 2023-10-23
[WIP] feat: support k8s multiple apiservers

bittermandel:
PSA: This is a Draft PR to keep track of progress and get early comments.
 
This PR is a continuation of https://github.com/cilium/cilium/pull/20090/.

With the upstream changes since that PR was worked on, this PR had to do things slightly different.

Short explanation:

The config used for the Kubernetes client is wrapped in a Roundtripper that rotates the Host based on a list passed through the config field `k8s-api-server-urls`. While establishing the initial connection, it rotates through the APIServerURLs one by one until it manages to establish connection.

Features missing:
* Always rotate URL if the client fails to connect for X time, not only during startUp.
* When initial connection has been established, replace the apiserver urls passed through config with the endpoints of the `kubernetes` service.

### DIFF
--- a/pkg/k8s/client/cell.go
+++ b/pkg/k8s/client/cell.go
@@ -6,8 +6,10 @@ package client
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"net"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -41,6 +43,7 @@ import (
 	slim_clientset "github.com/cilium/cilium/pkg/k8s/slim/k8s/client/clientset/versioned"
 	slim_fake "github.com/cilium/cilium/pkg/k8s/slim/k8s/client/clientset/versioned/fake"
 	k8sversion "github.com/cilium/cilium/pkg/k8s/version"
+	"github.com/cilium/cilium/pkg/lock"
 	"github.com/cilium/cilium/pkg/logging"
 	"github.com/cilium/cilium/pkg/logging/logfields"
 	"github.com/cilium/cilium/pkg/version"
@@ -109,6 +112,9 @@ type compositeClientset struct {
 	log           logrus.FieldLogger
 	closeAllConns func()
 	restConfig    *rest.Config
+	rt            *rotatingHttpRoundTripper
+
+	lock.RWMutex
 }
 
 func newClientset(lc hive.Lifecycle, log logrus.FieldLogger, cfg Config) (Clientset, error) {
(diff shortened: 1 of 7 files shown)


pkg/k8s/client/cell.go
```diff
@@ -109,29 +112,69 @@ type compositeClientset struct {
 	log           logrus.FieldLogger
 	closeAllConns func()
 	restConfig    *rest.Config
+	rt            *rotatingHttpRoundTripper
+
+	lock.RWMutex
```
bittermandel, line 117: I was not able to put this lock in `cell.Config` due to Cell assuming that all struct fields are passed by pflags. Not sure if there is a better place for it than here.

pkg/k8s/client/cell.go
```diff
@@ -109,29 +112,69 @@ type compositeClientset struct {
...
+	} else {
+		apiServerURL, err := url.Parse(restConfig.Host)
+		if err != nil {
+			return nil, err
+		}
+		client.rt.apiServerURL = apiServerURL
+		log.Infof("Setting APIServerURL to %s", apiServerURL)
+	}
+
+	restConfig.Wrap(client.WrapRoundTripper)
```
bittermandel, line 181: This makes rotatingHttpRoundTripper.Roundtrip run before the pre-configured one. This enables us to change the receiving Host dynamically.

squeed:
Sweet.

What is the best-practice for kube-proxy and kubelet here? AFAICT, kubelet still doesn't support multiple api servers; is there a plan for that to change as well?

brb:
> What is the best-practice for kube-proxy and kubelet here

* kube-proxy - use Cilium's kube-proxy replacement :smile: 
* kubelet - how does apiserver URL gets specified to it? I guess it doesn't use ClusterIP of the `kubernetes` service due to the :chicken: :egg:  problem?

brb:
@bittermandel  bump ^^

bittermandel:
Sorry, I took a bit of a detour!

After some considerations, I am not sure this feature makes sense since the Kubelet has to be configured with a single Kubeconfig cluster as well. Even kubeadm join gets its apiserver endpoint from the `kube-public/cluster-info` config map, which is also a single Kubeconfig cluster.

To achieve resilience and reliable connectivity, we ended up relying on kube-vip for the kubelet as well as Cilium. Until the Kubelet can also support multiple apiservers, I don't see how this feature can be deployed and relied upon fully. See the very long and old discussion here: https://github.com/kubernetes/kubernetes/issues/18174.

TLDR: When using kubeProxyReplacement, we will have to rely on the same address that is used by the Kubelet. The Kubelet  does not support multiple apiservers, so that will have to be solved in tangent.

PR #28744, opened 2023-10-23
Improve CLI user experience for "kubectl get ciliumendpoints"

### LINKED ISSUE #27459: CFP: Recreate CNP CRDs based on feature enablement
Currently, Cilium Operator installs all CNP CRD in the cluster upon initialization. Since this CRD is defined with kubemarkers, its configuration is hardcoded [here](https://github.com/cilium/cilium/blob/31fa619d8fea7ff8228c24779f726424fb1b08d5/pkg/k8s/apis/cilium.io/v2/types.go#L19-L31). This causes some issues as pointed out in https://github.com/cilium/cilium/issues/17880 and can be confusing for users.

To prevent these issues, we could perform some conditional checking while we are constructing the CRD in [here](https://github.com/cilium/cilium/blob/4076314941da6ef1ed3728d2b0328996a31d4d73/pkg/k8s/apis/cilium.io/client/register.go#L315-L338) and define the CEP column views based on the features that are enabled. For example, if IPv6 is not enabled then we shouldn't display the IPv6 column, or, if we start Cilium agent with `--endpoint-status=policy` then we should show the column for policy enforcement status otherwise it should be omitted.

Iiqbal2000:
Fix the empty columns by adding priority fields to columns that may be empty.

Fixes: #27459 

```release-note
Hide empty columns by default in "kubectl get ciliumendpoints" output
```

### DIFF
--- a/pkg/k8s/apis/cilium.io/v2/types.go
+++ b/pkg/k8s/apis/cilium.io/v2/types.go
@@ -21,11 +21,10 @@ import (
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +k8s:openapi-gen=false
 // +kubebuilder:resource:categories={cilium},singular="ciliumendpoint",path="ciliumendpoints",scope="Namespaced",shortName={cep,ciliumep}
-// +kubebuilder:printcolumn:JSONPath=".status.id",description="Cilium endpoint id",name="Endpoint ID",type=integer
-// +kubebuilder:printcolumn:JSONPath=".status.identity.id",description="Cilium identity id",name="Identity ID",type=integer
-// +kubebuilder:printcolumn:JSONPath=".status.policy.ingress.state",description="Ingress enforcement in the endpoint",name="Ingress Enforcement",type=string
-// +kubebuilder:printcolumn:JSONPath=".status.policy.egress.state",description="Egress enforcement in the endpoint",name="Egress Enforcement",type=string
-// +kubebuilder:printcolumn:JSONPath=".status.visibility-policy-status",description="Status of visibility policy in the endpoint",name="Visibility Policy",type=string
+// +kubebuilder:printcolumn:JSONPath=".status.identity.id",description="Security Identity",name="Security Identity",type=integer
+// +kubebuilder:printcolumn:JSONPath=".status.policy.ingress.state",description="Ingress enforcement in the endpoint",name="Ingress Enforcement",type=string,priority=1
+// +kubebuilder:printcolumn:JSONPath=".status.policy.egress.state",description="Egress enforcement in the endpoint",name="Egress Enforcement",type=string,priority=1
+// +kubebuilder:printcolumn:JSONPath=".status.visibility-policy-status",description="Status of visibility policy in the endpoint",name="Visibility Policy",type=string,priority=1
 // +kubebuilder:printcolumn:JSONPath=".status.state",description="Endpoint current state",name="Endpoint State",type=string
 // +kubebuilder:printcolumn:JSONPath=".status.networking.addressing[0].ipv4",description="Endpoint IPv4 address",name="IPv4",type=string
 // +kubebuilder:printcolumn:JSONPath=".status.networking.addressing[0].ipv6",description="Endpoint IPv6 address",name="IPv6",type=string
--- a/pkg/k8s/apis/cilium.io/client/crds/v2/ciliumendpoints.yaml
+++ b/pkg/k8s/apis/cilium.io/client/crds/v2/ciliumendpoints.yaml
@@ -21,25 +21,24 @@ spec:
   scope: Namespaced
   versions:
   - additionalPrinterColumns:
-    - description: Cilium endpoint id
-      jsonPath: .status.id
-      name: Endpoint ID
-      type: integer
-    - description: Cilium identity id
+    - description: Security Identity
       jsonPath: .status.identity.id
-      name: Identity ID
+      name: Security Identity
       type: integer
     - description: Ingress enforcement in the endpoint
       jsonPath: .status.policy.ingress.state
       name: Ingress Enforcement
+      priority: 1
       type: string
     - description: Egress enforcement in the endpoint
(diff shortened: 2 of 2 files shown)


pkg/k8s/apis/cilium.io/v2/types.go
```diff
@@ -22,9 +22,9 @@ import (
 // +kubebuilder:resource:categories={cilium},singular="ciliumendpoint",path="ciliumendpoints",scope="Namespaced",shortName={cep,ciliumep}
 // +kubebuilder:printcolumn:JSONPath=".status.id",description="Cilium endpoint id",name="Endpoint ID",type=integer
```
joestringer: I might suggest dropping the EndpointID as well. This particular line is different for every node and not globally that useful.

```suggestion
```

(Please keep the `Cilium identity id`, though that should really say "Security Identity")

joestringer (changes requested):
In terms of hiding columns, I think this makes sense. I'll leave the review of how this becomes reality to Andre; this is all magic to me, and I don't understand how changing a comment makes the code different :)

joestringer:
Ah CI gives the next steps, see the failed job for `Go related checks`.

pkg/k8s/apis/cilium.io/v2/types.go
Iiqbal2000: Should I change `Cilium identity id` to `Security Identity`?

Iiqbal2000:
> Ah CI gives the next steps, see the failed job for `Go related checks`.

It seems I missed a step

pkg/k8s/apis/cilium.io/v2/types.go
joestringer: Sure, I think that would be clearer given the terminology we publish on the docs.

Iiqbal2000: So it would look like this:

```
// +kubebuilder:printcolumn:JSONPath=".status.identity.id",description="Security Identity",name="Identity ID",type=integer
```

joestringer: Does that show up in the command output? I would have thought it would be the `name` field.

aanm (commented):
@Iiqbal2000 please run `make generate-k8s-api && make manifests' and submit your changes

pkg/k8s/apis/cilium.io/v2/types.go
Iiqbal2000: I think I misunderstood. If it is shown as a column name in the output, you are right it should be inserted into the name field, it would be:
```
// +kubebuilder:printcolumn:JSONPath=".status.identity.id",description="Cilium identity id",name="Security Identity",type=integer
```

joestringer: I don't know how this works, you'll have to try it out and see.

aanm:
I've marked the Pull Request as draft. Once you have finished up working on it feel free to click on the "Ready for review" button.

image

Thank you

Iiqbal2000:
> I've marked the Pull Request as draft. Once you have finished up working on it feel free to click on the "Ready for review" button.
> 
> image
> 
> Thank you

Thanks for the note! I'm a bit swamped lately, I'll get back to the PR and ping you once I've hit 'Ready for review'

pkg/k8s/apis/cilium.io/v2/types.go
Iiqbal2000: Ok @joestringer, I'll get on that soon

Iiqbal2000: Hi @joestringer, I have followed your feedback, and it will results:

```
NAMESPACE            NAME                                      SECURITY IDENTITY ID   ENDPOINT STATE   IPV4           IPV6
kube-system          coredns-5d78c9869d-629d7                  1867                   ready            10.244.1.226   fd00:10:244:1::4f15
kube-system          coredns-5d78c9869d-pvvb5                  1867                   ready            10.244.1.38    fd00:10:244:1::49fb
local-path-storage   local-path-provisioner-6bc4bddd6b-f8lqm   4756                   ready            10.244.1.63    fd00:10:244:1::a1ad
```

What do you think?

joestringer: Looks good to me. One nit, `SECURITY IDENTITY ID` is a bit redundant, it can just say `SECURITY IDENTITY`.

Iiqbal2000: That's great, I will change it

joestringer:
I rebased & squashed the commits to tidy it up and to address one of the CI failures. Will trigger full testsuite after the smoke tests go through.

Iiqbal2000:
Thank you @joestringer 
I forgot to rebase it. is there any action I need to take?

joestringer:
No action is needed at this time, just looking for the CI tests to complete. I don't expect that this change should impact CI at all, hopefully CI doesn't rely on the textual output of this command.

joestringer (approved):
Change LGTM. Assuming it passes CI this should be good to merge.

youngnick:
Looks like the external workloads test failed, I can't see any errors aside from not being able to find a specific pod in the connectivity tests, I'll rerun.

aanm (approved):

joestringer:
Thanks for your contribution @Iiqbal2000 !

Kimjonghyeon:
@aanm 
I'm curious about something.
Why were the items below removed?
There doesn't seem to be any clear reason mentioned in the issue.
https://github.com/cilium/cilium/issues/27459

```
    - description: Cilium endpoint id
      jsonPath: .status.id
      name: Endpoint ID
      type: integer
```
```
// +kubebuilder:printcolumn:JSONPath=".status.id",description="Cilium endpoint id",name="Endpoint ID",type=integer
```

joestringer:
@Kimjonghyeon Endpoint ID is only unique within the local node, so it's not globally meaningful in the cluster. Highlighting it in centralized clusterwide output creates confusion, so it's better not to highlight it by default.

PR #28745, opened 2023-10-23
services: don't wait for clustermesh to delete local stale backends

giorio94:
fe4dda76dd6a ("services: prevent temporary connectivity loss on agent restart") modified the handling of restored backends to prevent possibly causing temporary connectivity disruption on agent restart if a service is either associated with multiple endpointslices (e.g., it has more than 100 backends, or is dual stack) or has backends spanning across multiple clusters (i.e., it is a global service). At a high level, we now keep a list of restored backends, which continue being merged with the ones we received an update for, until the bootstrap phase completes. At that point, we trigger an update for each service still associated with stale backends, so that they can be removed.

One drawback associated with this approach, though, is that when clustermesh is enabled we currently wait for full synchronization from all remote clusters before triggering the removal of stale backends, regardless of whether the given service is global (i.e., possibly includes also remote backends) or not. One specific example in which such behavior is problematic relates to the clustermesh-apiserver. Indeed, if it gets restarted at the same time of the agents (e.g., during an upgrade), the associated service might end up including both the address of the previous pod (which is now stale) and that of the new one, which is correct. When kvstoremesh is enabled, local agents connect to it through that service. In this case, there's a circular dependency: the agent may pick the stale backend, and the connection to etcd fails, which in turn prevents the synchronization from being started, and eventually complete to trigger the removal of the stale backend. Although this dependency eventually resolves as a different backend is picked by the service load-balancing algorithm, unnecessary delay is introduced (the same could also happen for remote agents connecting through a NodePort if KPR is enabled).

To remove this dependency, let's perform a two-pass cleanup of stale backends: the first one as soon as we synchronize with Kubernetes, targeting non-global services only; the second triggered by full clustermesh synchronization, covering all remaining ones. Hence, non-global services can be fully functional also before the completion of the full clustermesh synchronization. It is worth mentioning that this fix applies to all non-global services, which can now converge faster also in case of large clustermeshes.

Marking for backport to v1.14, as the mentioned issue was observed in CI, and caused increased flakiness.

```release-note
Improve deletion of stale backends associated with non-global services, without waiting for full Cluster Mesh synchronization 
```

### DIFF
--- a/daemon/cmd/state.go
+++ b/daemon/cmd/state.go
@@ -14,10 +14,12 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/cilium/cilium/pkg/controller"
 	"github.com/cilium/cilium/pkg/endpoint"
 	"github.com/cilium/cilium/pkg/ipam"
+	"github.com/cilium/cilium/pkg/k8s"
 	slim_corev1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/core/v1"
 	"github.com/cilium/cilium/pkg/k8s/watchers/resources"
 	"github.com/cilium/cilium/pkg/labels"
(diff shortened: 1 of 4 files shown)


daemon/cmd/state.go
```diff
@@ -452,32 +452,44 @@ func (d *Daemon) initRestore(restoredEndpoints *endpointRestoreState, endpointsR
...
 			if d.clientset.IsEnabled() {
+				// Configure the controller which removes any leftover Kubernetes
+				// services that may have been deleted while Cilium was not
+				// running. Once this controller succeeds, because it has no
+				// RunInterval specified, it will not run again unless updated
+				// elsewhere. This means that if, for instance, a user manually
+				// adds a service via the CLI into the BPF maps, that it will
+				// not be cleaned up by the daemon until it restarts.
+				syncServices := func(localOnly bool) {
+					controller.NewManager().UpdateController(
```
joamaki: Hey, let's use `d.controllers` here so that this will also be stopped via `Close()` and thus tests won't leave a controller running in the background.

pkg/k8s/service_cache.go
```diff
@@ -274,6 +274,16 @@ func (s *ServiceCache) DeleteService(k8sSvc *slim_corev1.Service, swg *lock.Stop
 	}
 }
 
+// IsServiceLocal returns whether the given service is known, and all
+// backends belong to the local cluster.
+func (s *ServiceCache) IsServiceLocal(svcID ServiceID) bool {
+	s.mutex.RLock()
```
joamaki, line 283: Doesn't this lead to a dead lock?

1. `SyncWithK8sFinished` does `Service.Lock()`
2. `ServiceCache` gets a new service and sends it to `Events` channel
3. `watcher.K8sServiceHandler` reads an event from the `ServiceCache` channel and calls `svcManager.UpsertService`. Blocks acquiring `Service.Lock()`
4. `ServiceCache` gets another service and sends it to `Events` channel (repeat here until buffer is full)
5. `ServiceCache` gets another service, takes the `ServiceCache` write lock, and under the lock tries to send it to `Events` channel and blocks.
6. `ServiceManager` calls `IsServiceLocal` and tries to acquire `ServiceCache` read lock.

We now have a deadlock with two goroutines:
- `ServiceManager.SyncWithK8sFinished`: `ServiceManager` is write locked and it tries to acquire `ServiceCache` read lock.
- `ServiceCache.UpdateService`: `ServiceCache` is write locked and it tries to indirectly lock `ServiceManager` due to a blocking send to a full channel.

Could you try to avoid this cyclic locking issue by passing the `IsServiceLocal` property down to `ServiceManager` as part of `lb.SVC`? Or alternatively you could pass a map of these properties to `SyncWithK8sFinished`.

joamaki (changes requested):

pkg/k8s/service_cache.go
joamaki: ~~Or we can probably fix this by releasing the `ServiceCache` lock before the send to the `Events` channel.~~
EDIT: Bad idea, ServiceCache methods can be called concurrently, this may cause reordering of the events.

daemon/cmd/state.go
giorio94: Sure!

pkg/k8s/service_cache.go
giorio94: Hmm, you are definitely right about the possible deadlock. Let me rework it to prevent this possibility.

giorio94: The list of local services is now retrieved in advance, as you suggested, so that we don't need to hold both locks at the same time while checking whether a given service is local or not.

giorio94:
Rebased onto main to pick the necessary CI changes to successfully run the tests.

joamaki (approved):
LGTM thanks!

jrajahalme (approved):

giorio94:
Reviews from all code owners are in, tests are green except for a non-required one (and known to be flaky). Marking as ready to merge.

giorio94:
Not sure why MLH removed the ready to merge label. Adding back again.

Issue #24990, opened 2023-04-19
CI: ClusterMesh: timeouts in client-egress-l7


pchaigno:
I've seen a few variations of this, all with L7 on egress and clustermesh:
```
  ❌ client-egress-l7-named-port/pod-to-pod/curl-1: cilium-test/client2-bc59f56d5-w5w9c (10.242.1.182) -> cilium-test/echo-other-node-5c58b8cf8b-g4lwr (10.244.1.197:8080)
  ❌ client-egress-l7/pod-to-world/http-to-google.com-0: cilium-test/client2-bc59f56d5-bzr5n (10.242.1.229) -> google.com-http (google.com:80)
  ❌ client-egress-l7/pod-to-world/http-to-google.com-0: cilium-test/client2-bc59f56d5-qkj62 (10.242.1.156) -> google.com-http (google.com:80)
```

Link: https://github.com/cilium/cilium/actions/runs/4722694005/jobs/8377638774
Sysdump: 
[cilium-sysdumps-5.zip](https://github.com/cilium/cilium/files/11278001/cilium-sysdumps-5.zip)

Link: https://github.com/cilium/cilium/actions/runs/4719364276/jobs/8370097019
Sysdump: 
[cilium-sysdumps-1.zip](https://github.com/cilium/cilium/files/11278002/cilium-sysdumps-1.zip)

Link: https://github.com/cilium/cilium/actions/runs/4713940509/jobs/8359970494
Sysdump: 
[cilium-sysdumps-2.zip](https://github.com/cilium/cilium/files/11278003/cilium-sysdumps-2.zip)

michi-covalent:
https://github.com/cilium/cilium/actions/runs/4827305584/jobs/8599940047

this one is with `client-egress-l7-named-port` but i'm guessing it's the same issue:
```
❌ 1/32 tests failed (1/224 actions), 1 tests skipped, 1 scenarios skipped:
Test [client-egress-l7-named-port]:
  ❌ client-egress-l7-named-port/pod-to-world/http-to-google.com-0: cilium-test/client2-76f4d7c5bc-k75nf (10.242.1.24) -> google.com-http (google.com:80)
```

[cilium-sysdumps-7.zip](https://github.com/cilium/cilium/files/11357337/cilium-sysdumps-7.zip)

qmonnet:
[Hit](https://github.com/cilium/cilium/actions/runs/5355075400) on https://github.com/cilium/cilium/pull/26205

```
[...]

PR #28746, opened 2023-10-23
Modularize iptables manager

pippolo84:
Modularize the iptables configuration manager in a cell.
Direct dependency to the DaemonConfig global object is avoided through a SharedConfig privately provided to the cell. This config gets the values from all the referenced DaemonConfig helpers at startup, in order to be used by the new cell.

Also, cells depending on iptables (like the node manager one) now takes a reference to it through dependency injection instead of directly calling side effectful exported functions.

~Depends on https://github.com/cilium/cilium/pull/28713~

### DIFF
--- a/pkg/datapath/cells.go
+++ b/pkg/datapath/cells.go
@@ -57,6 +57,9 @@ var Cell = cell.Module(
 	// The modules manager to search and load kernel modules.
 	modules.Cell,
 
+	// Manages Cilium-specific iptables rules.
+	iptables.Cell,
+
 	cell.Provide(
 		newWireguardAgent,
 		newDatapath,
@@ -131,20 +134,7 @@ func newDatapath(params datapathParams) types.Datapath {
 		ProcFs:     option.Config.ProcFs,
 	}
 
-	iptablesManager := &iptables.Manager{}
-
-	params.LC.Append(hive.Hook{
-		OnStart: func(hive.HookContext) error {
-			// FIXME enableIPForwarding should not live here
-			if err := enableIPForwarding(); err != nil {
-				log.Fatalf("enabling IP forwarding via sysctl failed: %s", err)
-			}
-
-			iptablesManager.Init(params.ModulesManager)
-			return nil
-		}})
-
-	datapath := linuxdatapath.NewDatapath(datapathConfig, iptablesManager, params.WgAgent, params.NodeMap, params.ConfigWriter)
+	datapath := linuxdatapath.NewDatapath(datapathConfig, params.IptablesManager, params.WgAgent, params.NodeMap, params.ConfigWriter)
 
 	params.LC.Append(hive.Hook{
 		OnStart: func(hive.HookContext) error {
@@ -174,6 +164,8 @@ type datapathParams struct {
 	DeviceManager *linuxdatapath.DeviceManager
 
 	ModulesManager *modules.Manager
+
+	IptablesManager *iptables.Manager
 
 	ConfigWriter types.ConfigWriter
 }
--- a/pkg/datapath/cells.go
+++ b/pkg/datapath/cells.go
@@ -131,7 +131,7 @@ func newDatapath(params datapathParams) types.Datapath {
 		ProcFs:     option.Config.ProcFs,
 	}
 
-	iptablesManager := &iptables.IptablesManager{}
+	iptablesManager := &iptables.Manager{}
 
 	params.LC.Append(hive.Hook{
 		OnStart: func(hive.HookContext) error {
--- a/pkg/datapath/iptables/cell.go
+++ b/pkg/datapath/iptables/cell.go
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package iptables
+
+import (
+	"github.com/spf13/pflag"
+
+	"github.com/cilium/cilium/pkg/hive/cell"
+	"github.com/cilium/cilium/pkg/option"
+	"github.com/cilium/cilium/pkg/time"
+)
+
+var Cell = cell.Module(
+	"iptables",
+	"Handle iptables-related configuration for Cilium",
+
+	cell.Config(defaultConfig),
+	cell.ProvidePrivate(func(
+		cfg *option.DaemonConfig,
+	) SharedConfig {
+		return SharedConfig{
+			EnableIPv4:             cfg.EnableIPv4,
+			EnableIPv6:             cfg.EnableIPv6,
+			RoutingMode:            cfg.RoutingMode,
+			EnableXTSocketFallback: cfg.EnableXTSocketFallback,
+			EnableBPFTProxy:        cfg.EnableBPFTProxy,
+		}
+	}),
+	cell.Provide(newIptablesManager),
+)
+
+type Config struct {
+	// IPTablesLockTimeout defines the "-w" iptables option when the
+	// iptables CLI is directly invoked from the Cilium agent.
+	IPTablesLockTimeout time.Duration
+}
+
+var defaultConfig = Config{
+	IPTablesLockTimeout: 5 * time.Second,
+}
+
+func (def Config) Flags(flags *pflag.FlagSet) {
+	flags.Duration("iptables-lock-timeout", def.IPTablesLockTimeout, "Time to pass to each iptables invocation to wait for xtables lock acquisition")
+}
+
+type SharedConfig struct {
+	EnableIPv4             bool
+	EnableIPv6             bool
+	RoutingMode            string
+	EnableXTSocketFallback bool
+	EnableBPFTProxy        bool
+}
--- a/pkg/datapath/iptables/cell.go
+++ b/pkg/datapath/iptables/cell.go
@@ -6,6 +6,7 @@ package iptables
 import (
 	"github.com/spf13/pflag"
 
+	"github.com/cilium/cilium/pkg/cidr"
 	"github.com/cilium/cilium/pkg/hive/cell"
 	"github.com/cilium/cilium/pkg/option"
 	"github.com/cilium/cilium/pkg/time"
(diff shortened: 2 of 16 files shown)


pkg/modules/cell.go
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package modules
+
+import "github.com/cilium/cilium/pkg/hive/cell"
+
+var Cell = cell.Module(
+	"modules-manager",
+	"modules manager to search and load kernel modules",
```
joamaki: No need to repeat the module id in the title, e.g. you could just have "Loads kernel modules required by Cilium".
This package also should be relocated to `pkg/datapath/linux/modules`.

pkg/modules/modules.go
```diff
@@ -50,3 +40,20 @@ func (m *ModulesManager) FindOrLoadModules(expectedNames ...string) error {
 	}
 	return nil
 }
+
+func newManager(lc hive.Lifecycle) *Manager {
+	mm := Manager{}
+
+	lc.Append(hive.Hook{
+		OnStart: func(_ hive.HookContext) error {
+			modulesList, err := listModules()
```
joamaki: bit of code golf:
```
func (hive.HookContext) (err error) {
  m.modulesList, err = listModules()
  return
}
```
or maybe even better would be just:

```
func (m *Manager) Start(hive.HookContext) (err error) {
  m.moduleList, err = listModules()
  return
}

func newManager(...) ... {
  ...
  lc.Append(hive.Hook{OnStart: m.Start})
}
```

pkg/modules/cell.go
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package modules
+
+import "github.com/cilium/cilium/pkg/hive/cell"
+
+var Cell = cell.Module(
+	"modules-manager",
```
joamaki: Maybe `kernel-module-loader` would be more informative?

pkg/datapath/iptables/cell.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+const (
+	// IPTablesLockTimeout defines the "-w" iptables option when the
+	// iptables CLI is directly invoked from the Cilium agent.
+	IPTablesLockTimeout = "iptables-lock-timeout"
+)
+
+var Cell = cell.Module(
+	"iptables-manager",
```
joamaki: I would just call this `iptables` as `-manager` doesn't really add anything.

pkg/datapath/iptables/cell.go
```diff
@@ -0,0 +1,46 @@
...
+	"iptables manager to handle iptables-related configuration for Cilium",
```
joamaki: No need to repeat module id in title

pkg/datapath/cells.go
```diff
@@ -57,6 +57,20 @@ var Cell = cell.Module(
 	// Must appear before the datapath so that its start hook is registered before the iptables one.
 	modules.Cell,
 
+	// iptables configuration manager.
+	iptables.Cell,
+	cell.Provide(func(
```
joamaki: I would move this into `iptables.Cell` and keep things fairly tight here.

pkg/datapath/iptables/iptables.go
```diff
@@ -267,46 +275,73 @@ type Manager struct {
...
+
+	p.Lifecycle.Append(iptMgr)
+
+	return iptMgr
+}
+
+// Start initializes the iptables manager and checks for iptables kernel modules availability.
+func (m *Manager) Start(ctx hive.HookContext) error {
+	if err := enableIPForwarding(); err != nil {
+		log.Fatalf("enabling IP forwarding via sysctl failed: %s", err)
```
joamaki: Don't fatal, but return an error instead.

pkg/datapath/iptables/iptables.go
```diff
@@ -267,46 +275,73 @@ type Manager struct {
...
-	if err := modulesManager.FindOrLoadModules(
-		"ip6_tables", "ip6table_mangle", "ip6table_raw", "ip6table_filter"); err != nil {
-		if option.Config.EnableIPv6 {
-			log.WithError(err).Fatal(
+
+	if err := m.modulesMgr.FindOrLoadModules(
+		"ip6_tables", "ip6table_mangle", "ip6table_raw", "ip6table_filter",
+	); err != nil {
+		if m.sharedCfg.EnableIPv6 {
+			return errors.New(
```
joamaki: The error is now dropped, I would wrap it instead with fmt.Errorf and `%w`.

joamaki (changes requested):

pkg/modules/cell.go
pippolo84: Fixed in https://github.com/cilium/cilium/pull/28713 as well as for the other comments regarding modules manager.

pippolo84:
I've addressed all the feedbacks (the ones regarding modules manager have been pushed in the specific PR as well) but I'm planning to rework this quite a lot, so still not ready for review yet.

pkg/datapath/iptables/cell.go
```diff
@@ -0,0 +1,95 @@
...
+	// iptables CLI is directly invoked from the Cilium agent.
+	IPTablesRandomFully = "iptables-random-fully"
+)
+
+var Cell = cell.Module(
+	"iptables",
+	"Handle iptables-related configuration for Cilium",
+
+	cell.Config(defaultConfig),
+	cell.Provide(func(
```
dylandreimerink: This is essentially just a layer around the daemonConfig, I don't see the need to expose this as a separate type in hive. I understand it makes sense internally for the purposes of mocking though

dylandreimerink (changes requested):

pkg/datapath/iptables/cell.go
pippolo84: Yep, main reason to do that is to avoid depending on the big global config object `DaemonConfig`. But it's better to provide it only privately. I'll fix this, thanks :+1:

pkg/datapath/iptables/cell.go
```diff
@@ -0,0 +1,94 @@
...
+	IPTablesRandomFully        bool
+}
+
+var defaultConfig = Config{
+	IPTablesLockTimeout: 5 * time.Second,
+}
+
+func (def Config) Flags(flags *pflag.FlagSet) {
+	flags.Duration(IPTablesLockTimeout, def.IPTablesLockTimeout, "Time to pass to each iptables invocation to wait for xtables lock acquisition")
+	flags.StringSlice(DisableIptablesFeederRules, def.DisableIptablesFeederRules, "Chains to ignore when installing feeder rules.")
```
mhofstetter: changing this to a Hive config no longer parses the passed string as space-separated string. Instead we have to adapt the Helm Chart to pass a comma-separated string -> https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/templates/cilium-configmap.yaml#L926

pkg/datapath/iptables/cell.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"github.com/spf13/pflag"
+
+	"github.com/cilium/cilium/pkg/cidr"
+	"github.com/cilium/cilium/pkg/hive/cell"
+	"github.com/cilium/cilium/pkg/option"
+	"github.com/cilium/cilium/pkg/time"
+)
+
+const (
```
mhofstetter: i'd prefer having these const values inlined. It's still possible to keep the comments above the corresponding flag definitions in `Flags(...)`

pkg/datapath/iptables/iptables.go
```diff
@@ -267,46 +274,74 @@ type Manager struct {
...
+
+	p.Lifecycle.Append(iptMgr)
+
+	return iptMgr
+}
+
+// Start initializes the iptables manager and checks for iptables kernel modules availability.
+func (m *Manager) Start(ctx hive.HookContext) error {
+	if err := enableIPForwarding(); err != nil {
+		return fmt.Errorf("enabling IP forwarding via sysctl failed: %w", err)
```
dylandreimerink: If we fail to set the sysctl values, it does not necessity mean that Cilium will not function (the correct values may be set but we don't have write access), so I would convert this to a warning like the check below.

dylandreimerink (approved):
LGTM. Just one nit, but given its copied code will not block on that

pkg/datapath/cells.go
```diff
@@ -57,6 +57,9 @@ var Cell = cell.Module(
 	// The modules manager to search and load kernel modules.
 	modules.Cell,
 
+	// iptables configuration manager.
```
joamaki: Maybe better as something like "Manages Cilium-specific iptables rules"?

pkg/datapath/iptables/cell.go
joamaki: Yep likewise. I would move the doc comments on to the struct fields.

joamaki (approved):

pkg/datapath/iptables/cell.go
pippolo84: I've put this on hold to wait for the final decision about how to approach the StringSlice issue: https://github.com/cilium/cilium/pull/28919#issuecomment-1790405427

pippolo84: I'll take care of this in a follow-up PR as soon as we reach consensus.

mhofstetter (approved):
Thanks - LGTM!

pippolo84:
Force-pushed to fix controlplane tests: `... missing type: *iptables.Manager ...`

pippolo84:
Conformance Cluster Mesh failure tracked [here](https://github.com/cilium/cilium/issues/24990#issuecomment-1801895280), rerunning.

PR #28747, opened 2023-10-23
report endpoint ID on endpoint BPF program

aanm:
Presenting only the endpoint interface name into the logs might not be enough to understand which endpoint has the problem. Thus, we should print the endpoint ID well.

Fixes: d93659c058bf ("Resiliency: Checks endpoints BPF programs remain loaded")


christarazi (approved):
Have you seen this logic actually triggering? It should be extremely rare, so I'm asking because it might be finding false positives.

Issue #20960, opened 2022-08-18
CFP: Support TLS (HTTPS) backends and TLS passthrough on Cilium Ingress


eyanulis:
## Cilium Feature Proposal

**Is your feature request related to a problem?**

Follow-up from Slack thread found [here](https://cilium.slack.com/archives/C53TG4J4R/p1660718164559439). In short, while Cilium's `Ingress` controller has full support for TLS _frontends_, it does not currently appear to have support for TLS _backends_.

There exist some Kubernetes services that will serve HTTPS connections from the pod itself. An example of this is the Ceph Dashboard from Rook, which (if configured) will generate and serve a self-signed certificate on the dashboard port.

Attempting to point Cilium's `Ingress` controller at such a backend will not work. The Ingress config is as follows:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    cert-manager.io/cluster-issuer: lng1-letsencrypt-clusterissuer
    external-dns.alpha.kubernetes.io/hostname: ceph-dash-lng1.k8s.yanulis.net
  name: rook-ceph-dashboard
  namespace: rook-ceph
spec:
  ingressClassName: cilium
  rules:
  - host: ceph-dash-lng1.k8s.yanulis.net
    http:
      paths:
      - backend:
          service:
            name: rook-ceph-mgr-dashboard
            port:
              name: https-dashboard
        path: /
        pathType: Prefix
  tls:
  - hosts:
    - ceph-dash-lng1.k8s.yanulis.net
    secretName: ceph-dash-ingress-tls
```

(Note that this Ingress _is_ annotated such that cert-manager will issue a certificate for the _frontend_ connection (client to Envoy), but that is not relevant here - this discussion is regarding the connection between Envoy and the _backend_.)

As-is, with the backend pod serving TLS with a self-signed certificate, attempts to reach this `Ingress` from a client will fail with an HTTP 503 and the following error:
```
upstream connect error or disconnect/reset before headers. reset reason: connection termination
```

After some digging, it has become clear that Cilium's `Ingress` controller simply does not configure Envoy for TLS upstreams. This can be made to work by manually updating the relevant `CiliumEnvoyConfig` and changing the `Cluster` configuration from this (original):
```
 - '@type': type.googleapis.com/envoy.config.cluster.v3.Cluster
    connectTimeout: 5s
    name: rook-ceph/rook-ceph-mgr-dashboard:https-dashboard
    outlierDetection:
[...]

PR #28751, opened 2023-10-24
Add ssl-passthrough implementation support

youngnick:
This PR adds a new `ingress.cilium.io/ssl-passthrough` annotation for Ingress objects, and support for handling them, along with some tests to cover this, and associated docs.

Updates #20960

```release-note
Added new `ingress.cilium.io/ssl-passthrough` annotation for Ingress objects
```

### DIFF
--- a/Documentation/network/servicemesh/ingress.rst
+++ b/Documentation/network/servicemesh/ingress.rst
@@ -107,6 +107,21 @@ Supported Ingress Annotations
      - | Enable websocket passthrough support.
        | Applicable values are ``enabled`` and ``disabled``.
      - ``disabled``
+   * - ``ingress.cilium.io/ssl-passthrough``
+     - | Enable SSL Passthrough mode for this Ingress.
+       | Applicable values are ``enabled`` and ``disabled``,
+       | although boolean-style values will also be
+       | accepted.
+       | Note that some conditions apply to SSL
+       | Passthrough Ingresses, due to how
+       | TLS Passthrough works:
+       | * A ``host`` field must be set in the Ingress
+       | * Default backends are ignored
+       | * Rules with paths other than ``/`` are ignored
+       | If all the rules in an Ingress are ignored for
+       | these reasons, no Envoy config will be generated
+       | and the Ingress will have no effect.
+     - ``disabled``
 
 Additionally, cloud-provider specific annotations for the LoadBalancer service
 are supported.
--- a/Documentation/network/servicemesh/ingress.rst
+++ b/Documentation/network/servicemesh/ingress.rst
@@ -107,12 +107,13 @@ Supported Ingress Annotations
      - | Enable websocket passthrough support.
        | Applicable values are ``enabled`` and ``disabled``.
      - ``disabled``
-   * - ``ingress.cilium.io/ssl-passthrough``
-     - | Enable SSL Passthrough mode for this Ingress.
+   * - ``ingress.cilium.io/tls-passthrough``
+     - | Enable TLS Passthrough mode for this Ingress.
        | Applicable values are ``enabled`` and ``disabled``,
        | although boolean-style values will also be
        | accepted.
-       | Note that some conditions apply to SSL
+       |
+       | Note that some conditions apply to TLS
        | Passthrough Ingresses, due to how
        | TLS Passthrough works:
        | * A ``host`` field must be set in the Ingress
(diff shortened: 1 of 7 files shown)


Documentation/network/servicemesh/ingress.rst
```diff
@@ -107,6 +107,17 @@ Supported Ingress Annotations
      - | Enable websocket passthrough support.
        | Applicable values are ``enabled`` and ``disabled``.
      - ``disabled``
+   * - ``ingress.cilium.io/ssl-passthrough``
```
meyskens: I would like to discuss the name, as 4 lines further you call it `TLS Passthrough` would this be a more correct name?

youngnick: TLS Passthrough is more correct, but the expected annotation that people will be converting from is `ssl-passthrough`. We could change the name, I suppose?

meyskens: They would already come from `some.ingress.com/ssl-passthrough` to `ingress.cilium.io` in a regex driven replace :) 
`ingress.cilium.io/tls-passthrough` sounds good and links to our TLSRoute

Documentation/network/servicemesh/ingress.rst
```diff
@@ -107,6 +107,18 @@ Supported Ingress Annotations
...
+     - | Enable SSL Passthrough mode for this Ingress.
+       | Applicable values are ``enabled`` and ``disabled``, although boolean-style
+       | values will also be accepted.
+       | Note that some conditions apply to SSL Passthrough Ingresses, due to how
+       | TLS Passthrough works:
+       | * A ``host`` field must be set in the Ingress
+       | * Default backends are ignored
+       | * Rules with paths other than ``/`` are ignored
+       | If all the rules in an Ingress are ignored for these reasons, no Envoy config
+       | will be generated and the Ingress will have no effect.
```
qmonnet: I'm getting a weird formatting issue with the rendering of the table with your change, compare the current table:

<details><summary>Screenshot 1</summary>

table1
</details>

with the one after your change:

<details><summary>Screenshot 2</summary>

table2
</details>

There are some earlier lines that are no longer wrapped and that overflow into the third column. This is probably some bug in how Sphinx renders tables, because my understanding is that these lines should not be wrapped in the first place :shrug:.

Anyway, I don't have a clean fix, so please would you mind wrapping your own description _more_, which seems to keep the earlier lines as they are?

```suggestion
     - | Enable SSL Passthrough mode for this Ingress.
       | Applicable values are ``enabled`` and ``disabled``,
       | although boolean-style values will also be
       | accepted.
       | Note that some conditions apply to SSL
       | Passthrough Ingresses, due to how
       | TLS Passthrough works:
       | * A ``host`` field must be set in the Ingress
       | * Default backends are ignored
       | * Rules with paths other than ``/`` are ignored
       | If all the rules in an Ingress are ignored for
       | these reasons, no Envoy config will be generated
       | and the Ingress will have no effect.
```

qmonnet (changes requested):
Thanks! I've one request regarding formatting, please see below.

I'd also recommend folding the whitespace fixes from last commit into the relevant, earlier commits.

Documentation/network/servicemesh/ingress.rst
youngnick: Done, thanks!

Documentation/network/servicemesh/ingress.rst
youngnick: Yeah, okay, I'll change this tomorrow.

youngnick:
Fixed a couple more things, I think this is reviewable again now.

qmonnet (approved):
Looks good, thanks!

dylandreimerink:
@meyskens are you happy this this PR? Asking since this PR is mainly cilium/sig-servicemesh focused and we have no official approval

operator/pkg/ingress/annotations/annotations.go
```diff
@@ -16,6 +16,7 @@ const (
 	ServiceTypeAnnotation      = annotation.IngressPrefix + "/service-type"
 	InsecureNodePortAnnotation = annotation.IngressPrefix + "/insecure-node-port"
 	SecureNodePortAnnotation   = annotation.IngressPrefix + "/secure-node-port"
+	SSLPassthroughAnnotation   = annotation.IngressPrefix + "/tls-passthrough"
```
meyskens: Any reason we still call this `SSLPassthroughAnnotation` while the value is TLS?

youngnick: Oops, renamed.

meyskens (approved):
LGTM

PR #28757, opened 2023-10-24
gateway-api: Bump the version to v1.0.0-rc1

sayboras:
Most of the changes are related to API versions and go dependency, also some of below extra changes:

- Add default Listener ResolvedRef condition
- Update go.mod to current upstream main commit to include fixes for v1.0.0-rc1 conformance tests


meyskens (approved):
LGTM :) 
Not a lot has changes just a lot of changes...

gandro (approved):
Is there any upgrade impact to this, e.g. do users need to bump the API version of their custom resources too? If so, we probably should mention that in the docs.

sayboras:
> Is there any upgrade impact to this, e.g. do users need to bump the API version of their custom resources too? If so, we probably should mention that in the docs.

Good point, the minimum requirement is to have users install v1 CRDs, then kube api server will be able to publish events for both v1beta1 and v1 resources. I will update the docs separately 👍.

nebril (approved):

aanm (approved):
As pointed out by other reviewers this will require documentation changes to mention that users need to upgrade the CRD to v1.

I would suggest to add a note in the upgrade guide leaving a node specifically for gateway API users that v1beta1 will NOT work unless the v1 CRD is deployed first.

PR #28760, opened 2023-10-24
Modularize scale-test

marseel:
Make scale-test more modular to allow for adding more scale-test scenarios.
Also, increase the frequency of node-throughput test as it only creates two VMs.

### DIFF
--- a/.github/actions/scale-tests/create-cluster/action.yaml
+++ b/.github/actions/scale-tests/create-cluster/action.yaml
@@ -1,0 +1,85 @@
+name: 'Create kops cluster'
+description: "Workflow that creates kops cluster"
+inputs:
+  cluster_name:
+    description: "Name of the cluster to create"
+    required: true
+  control_plane_size:
+    description: "Control plane node size"
+    required: true
+  control_plane_count:
+    description: "Number of control plane nodes"
+    required: true
+  node_size:
+    description: "Node size"
+    required: true
+  node_count:
+    description: "Number of nodes"
+    required: true
+  kops_state:
+    description: "Bucket where to store kops state"
+    required: true
+  project_id:
+    description: "GCP Project in which to create cluster"
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Check if we should build the Cilium CLI from source
+      shell: bash
+      run: |
+        ./kops create cluster \
+          ${{ inputs.cluster_name }} \
+          --state ${{ inputs.kops_state }} \
+          --zones us-west1-a \
+          --project=${{ inputs.project_id }} \
+          --control-plane-count ${{ inputs.control_plane_count }} \
+          --control-plane-size ${{ inputs.control_plane_size }} \
+          --node-count ${{ inputs.node_count }} \
+          --node-size ${{ inputs.node_size }} \
+          --networking cni \
+          --yes
+
+    - name: Dump cluster config
+      shell: bash
+      run: |
+        ./kops get clusters \
+          ${{ inputs.cluster_name }} \
+          --state ${{ inputs.kops_state }} \
+          -o yaml
+
+    - name: Ensure cluster state matches config
+      shell: bash
+      run: |
+        ./kops update cluster \
+          ${{ inputs.cluster_name }} \
+          --state ${{ inputs.kops_state }} \
+          --yes
+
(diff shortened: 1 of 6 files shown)


marseel:
Successful run: https://github.com/cilium/cilium/actions/runs/6627684165/job/18003120946

.github/actions/scale-tests/create-cluster/action.yaml
```diff
@@ -0,0 +1,85 @@
...
+  kops_state:
+    description: "Bucket where to store kops state"
+    required: true
+  project_id:
+    description: "GCP Project in which to create cluster"
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Check if we should build the Cilium CLI from source
```
learnitall, line 28: I think the name for this step is incorrect?

.github/actions/scale-tests/export-results/action.yaml
```diff
@@ -0,0 +1,28 @@
...
+  other_files:
+    description: "Additional files to export"
+    required: true
+  test_name:
+    description: "Name of the test"
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Delete cluster
```
learnitall, line 19: I think the name for this step is incorrect here too.

.github/actions/scale-tests/install-kops/action.yaml
```diff
@@ -0,0 +1,17 @@
+name: 'Install kops'
+description: "Workflow that installs kops cli"
+inputs:
+  release_version:
+    description: "version of kops"
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Delete cluster
```
learnitall, line 10: This step name seems off also.

.github/actions/scale-tests/validate-cluster/action.yaml
```diff
@@ -0,0 +1,20 @@
+name: 'Validate cluster'
+description: "Workflow that wait for cluster to be in a ready state"
```
learnitall, line 2:
```suggestion
description: "Workflow that waits for cluster to be in a ready state"
```

.github/workflows/scale-test-gce.yaml
```diff
@@ -36,8 +36,8 @@ env:
   go_version: 1.21.3
   # Adding k8s.local to the end makes kops happy-
   # has stricter DNS naming requirements.
-  cluster_name: ${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}-${{ github.run_attempt }}.k8s.local
-  cl2_config: ${{ github.workspace }}/.github/actions/scale-test-gce/cl2-config.yaml
+  test_name: node-throughput
+  cluster_name: ${{ github.run_id }}-${{ github.run_attempt }}.k8s.local
```
learnitall, line 40: Can you switch these two lines? I think the `cluster_name` should be directly underneath the comment about k8s.local.

.github/actions/scale-tests/validate-cluster/action.yaml
```diff
@@ -0,0 +1,20 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Validate cluster
+      shell: bash
+      run: |
+        ./kops validate cluster \
+          ${{ inputs.cluster_name }} \
+          --state ${{ inputs.kops_state }} \
+          --wait 20m \
```
learnitall, line 19: Nit: this timeout value is hard coded, but a user could specify a longer timeout when consuming the workflow through the GHA option `timeout-minutes`.

.github/workflows/scale-test-gce.yaml
```diff
@@ -1,8 +1,8 @@
-name: Scale Test GCE
+name: Node throughput test
```
learnitall, line 1: Maybe it would make sense to also rename the file?

learnitall (changes requested):
After looking through this, I think it might be a good idea if we just make one large action that can be consumed, rather than splitting out each step into its own action. For instance, if we were to add another workflow that runs a different scale test, we would essentially copy and paste the node-throughput test and just edit the CL2 config and number of nodes. I'm not sure what the use case would be for consuming the smaller actions on their own. WDYT?

Besides that, I have a couple of small nits.

PR #28765, opened 2023-10-24
Improve Hubble Relay Kubernetes Readiness/Liveness check

### LINKED ISSUE #23542: Improve Hubble Relay Kubernetes Readiness/Liveness check
At the moment, Hubble Relay's Kubernetes readiness and liveness check are simply checking if the TCP port is open:

https://github.com/cilium/cilium/blob/f62b617059407f437757458fc4318064076c32e7/install/kubernetes/cilium/templates/hubble-relay/deployment.yaml#L66-L71

This kind of check is not particularly useful in determining if Hubble is ready. Instead, the user is required to manually inspect Relay's logs for common errors, such as: No access to peer service, mTLS authentication failures, pod2host connectivity broken and no connection to Hubble Observers.

We should introduce a more meaningful health-check which performs some basic checks, for example:

  - Have we received cluster information from the peer service?
  - Are we connected to a Hubble Observer on at least one node?
  - (Bonus): Do the connected Hubble Observers contain any flows in their ring buffer?

@kaworu mentioned that this health check could be exposed via the gRPC endpoint, which an upcoming Kubernetes version seems to support. This means we do not have to create a custom HTTP server just for this. @kaworu feel free to provide more details.

For older versions of Kubernetes, we could provide a `hubble-relay healthz` sub-command and use a [`command`](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command) based readiness/liveness probe.

glrf:

      GitHub issue.

This commit improves the exiting Hubble Relay Readiness and liveness checks by using the gRPC Health Checking Protocol.

To do that we add the [grpc_health_probe](https://github.com/grpc-ecosystem/grpc-health-probe) binary to the hubble relay container image and use `exec probes`. We can't use the built-in gRPC health checking as it doesn't yet support TLS and because we still support k8s versions before v1.23.

We also update the the existing gRPC health server to check connectivity to the peer service and hubble backends. The health server is only  serving, if the peer service and at least one hubble observe service is available.

This PR also contains a relatively unrelated fix for graceful shutdown of hubble relay. It simplified the tests, but I can also split it into two PRs if requested

closes #23542

### DIFF
--- a/install/kubernetes/cilium/templates/hubble-relay/deployment.yaml
+++ b/install/kubernetes/cilium/templates/hubble-relay/deployment.yaml
@@ -71,11 +71,26 @@ spec:
               protocol: TCP
             {{- end }}
           readinessProbe:
-            tcpSocket:
-              port: grpc
+            {{- include "hubble-relay.probe" . | nindent 12 }}
+            {{- if semverCompare "<1.20-0" .Capabilities.KubeVersion.Version }}
+            # Starting from Kubernetes 1.20, we are using startupProbe instead
+            # of this field.
+            initialDelaySeconds: 5
+            {{- end }}
           livenessProbe:
-            tcpSocket:
-              port: grpc
+            {{- include "hubble-relay.probe" . | nindent 12 }}
+            {{- if semverCompare "<1.20-0" .Capabilities.KubeVersion.Version }}
+            # Starting from Kubernetes 1.20, we are using startupProbe instead
+            # of this field.
+            initialDelaySeconds: 60
+            {{- end }}
+          {{- if semverCompare ">=1.20-0" .Capabilities.KubeVersion.Version }}
+          startupProbe:
+            # give the relay one minute to start up
+            {{- include "hubble-relay.probe" . | nindent 12 }}
+            failureThreshold: 20
+            periodSeconds: 3
+          {{- end }}
           {{- with .Values.hubble.relay.extraEnv }}
           env:
             {{- toYaml . | trim | nindent 12 }}
@@ -163,4 +178,18 @@ spec:
                   path: server.key
           {{- end }}
       {{- end }}
+{{- end }}
+
+{{- define "hubble-relay.probe" }}
+{{- /* This distinction can be removed once we drop support for k8s 1.23 */}}
+{{- if semverCompare ">=1.24-0" .Capabilities.KubeVersion.Version -}}
+grpc:
+  port: 4222
+{{- else }}
+exec:
+  command:
+  - grpc_health_probe
+  - -addr=localhost:4222
+{{- end }}
+timeoutSeconds: 3
 {{- end }}
--- a/pkg/hubble/relay/defaults/defaults.go
+++ b/pkg/hubble/relay/defaults/defaults.go
@@ -17,6 +17,8 @@ const (
 	// DialTimeout is the timeout that is used when establishing a new
 	// connection.
 	DialTimeout = 5 * time.Second
+	// HealthCheckInterval is the time interval between health checks.
+	HealthCheckInterval = 5 * time.Second
 	// GopsPort is the default port for gops to listen on.
 	GopsPort = 9893
 	// PprofAddress is the default port for pprof to listen on.
@@ -45,4 +47,8 @@ var (
 	// ListenAddress is the address on which the Hubble Relay server listens
 	// for incoming gRPC requests.
 	ListenAddress = fmt.Sprintf(":%d", hubbledefaults.RelayPort)
+
+	// HealthListenAddress is the address on which the Hubble Relay gRPC health
+	// server listens on
+	HealthListenAddress = fmt.Sprintf(":%d", 4222)
 )
--- a/pkg/hubble/relay/pool/manager.go
+++ b/pkg/hubble/relay/pool/manager.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"sync/atomic"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
@@ -31,13 +32,19 @@ type peer struct {
 // PeerManager manages a pool of peers (Peer) and associated gRPC connections.
 // Peers and peer change notifications are obtained from a peer gRPC service.
 type PeerManager struct {
-	opts    options
-	updated chan string
-	wg      sync.WaitGroup
-	stop    chan struct{}
-	mu      lock.RWMutex
-	peers   map[string]*peer
-	metrics *PoolMetrics
+	opts                 options
+	updated              chan string
+	wg                   sync.WaitGroup
+	stop                 chan struct{}
+	peerServiceConnected atomic.Bool
+	mu                   lock.RWMutex
+	peers                map[string]*peer
+	metrics              *PoolMetrics
+}
+
+type Status struct {
+	PeerServiceConnected bool
+	AvailablePeers       int
 }
 
 // NewPeerManager creates a new manager that connects to a peer gRPC service to
@@ -51,11 +58,12 @@ func NewPeerManager(registry prometheus.Registerer, options ...Option) (*PeerMan
 	}
 	metrics := NewPoolMetrics(registry)
 	return &PeerManager{
-		peers:   make(map[string]*peer),
-		updated: make(chan string, 100),
-		stop:    make(chan struct{}),
-		opts:    opts,
-		metrics: metrics,
+		peers:                make(map[string]*peer),
+		updated:              make(chan string, 100),
+		stop:                 make(chan struct{}),
+		opts:                 opts,
+		metrics:              metrics,
+		peerServiceConnected: atomic.Bool{},
 	}, nil
 }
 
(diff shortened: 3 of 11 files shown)


install/kubernetes/cilium/templates/hubble-relay/deployment.yaml
```diff
@@ -71,11 +71,11 @@ spec:
               protocol: TCP
             {{- end }}
           readinessProbe:
-            tcpSocket:
-              port: grpc
+            exec:
```
gandro: Can we use the gRPC check if the user is deploying a newer version of K8s and leave a note that this can be removed once we drop support for K8s 1.23? You can check with the K8s version via `.Capabilities.KubeVersion.Version`, we do this elsewhere in the Helm charts too.

Afaik, Cilium moved away from `exec` specifically because it is much more expensive than a network-based probe

install/kubernetes/cilium/templates/hubble-relay/deployment.yaml
```diff
@@ -164,3 +164,14 @@ spec:
...
       {{- end }}
 {{- end }}
+
+{{- define "hubble-relay.probecmd" -}}
+command:
+- grpc_health_probe
+- -addr=localhost:{{ .Values.hubble.relay.listenPort }}
+{{- if and .Values.hubble.tls.enabled  .Values.hubble.relay.tls.server.enabled }}
+- -tls
+# We don't know the server name in every case. Not verifying the cert for health checks should be fine
```
gandro: We should also check for `Values.hubble.relay.tls.server.mtls` - in that case the probe probably won't work at all, because there is no client cert for the probe.

It might be overkill, but could we expose the health service in a unprotected localhost socket to avoid issues with TLS?

gandro (commented):
Awesome work, glad to see this implemented! Two minor comments around the Kubernetes manifest

install/kubernetes/cilium/templates/hubble-relay/deployment.yaml
tklauser: > Afaik, Cilium moved away from `exec` specifically because it is much more expensive than a network-based probe

Yep, we moved away from `exec` based liveness/readiness probes mainly because of the resource overhead.

glrf: The issue is that the inbuilt gRPC check doesn't support TLS. But I could use the gRPC check if TLS is disabled and if we're running a recent enough Kubernetes version, and fall back to exec otherwise?

install/kubernetes/cilium/templates/hubble-relay/deployment.yaml
glrf: Thanks! I totally missed `mtls`.

Adding another local gRPC server just for the health service does seem a bit overkill, but I don't have a better solution. We could just fall back to checking the TCP socket in that case, but that's also not a great solution..

gandro: Having a localhost socket would also allow us to use the [built-in Kubernetes health checking](https://kubernetes.io/blog/2018/10/01/health-checking-grpc-servers-on-kubernetes/) in the future, so we would kill two birds with one stone. But it all depends on how much more complex it would make the code I think.

glrf: Ah right, we'd completely avoid all the TLS issues for built-in health checks. That convinced me. I'll see if I can expose the health service on a local socket.

glrf: Turns out only listening on localhost doesn't work as the health probe doesn't come from localhost (duh..). But I don't think exposing the health server unencrypted is a security risk, so I added another gRPC server that always runs unencrypted. That means for newer kubernetes versions we can always use the built-in gRPC health probes.

pkg/hubble/relay/defaults/defaults.go
```diff
@@ -45,4 +47,8 @@ var (
 	// ListenAddress is the address on which the Hubble Relay server listens
 	// for incoming gRPC requests.
 	ListenAddress = fmt.Sprintf(":%d", hubbledefaults.RelayPort)
+
+	// HealthListenAddress is the address on which the Hubble Relay gRPC health
+	// server listens on
+	HealthListenAddress = fmt.Sprintf(":%d", 4222)
```
gandro, line 53: Shouldn't this be `localhost:4222`?

install/kubernetes/cilium/templates/hubble-relay/deployment.yaml
```diff
@@ -71,11 +71,9 @@ spec:
               protocol: TCP
             {{- end }}
           readinessProbe:
-            tcpSocket:
-              port: grpc
+            {{- include "hubble-relay.probe" . | nindent 12 }}
```
gandro, line 74: I know the old code didn't do it, but we should strongly consider adding `initialDelaySeconds` and/or a `startupProbe`, especially now where the probe indicates actual readiness and might take a few seconds to connect to the peer service and the first peer.

It might also be worth setting `timeoutSeconds: 3`, that's what other ciilum components (e.g. agent, operator) do.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes

gandro (commented):
Nice! Glad to see that the local server simplifies the probe. I've left some additional feedback

pkg/hubble/relay/defaults/defaults.go
glrf: I thought so too, but no because the built-in gRPC probes don't originate from the container itself (which makes sense in hindsight). So we can't just listen on localhost.

That's not ideal, but I also don't really see an issue with answering health probes in plain text.

install/kubernetes/cilium/templates/hubble-relay/deployment.yaml
gandro: > Turns out only listening on localhost doesn't work as the health probe doesn't come from localhost (duh..).

That's interesting. We are using a localhost *HTTP* health probe both for agent and operator. There, you can configure the incoming interface via the `host` field: https://github.com/cilium/cilium/blob/33573bd0e7e797e25d81ab183e742cb348729d95/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml#L141

I wonder why the gRPC probe (at least according to the docs) does not support that :confused:

gandro (approved):
Thanks a lot! I hope the gRPC probe will support accessing from localhost in the future, so we can switch it to listening on localhost only. But for now, this seems like the best approach

pkg/hubble/relay/pool/manager.go
```diff
@@ -279,6 +290,32 @@ func (m *PeerManager) ReportOffline(name string) {
 	}()
 }
 
+// Status provides the status of the manager
+func (m *PeerManager) Status() Status {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+	availablePeers := 0
+	for _, peer := range m.peers {
+		peer.mu.Lock()
```
joamaki, line 258: To make the lock handling more obvious this could be simplified to:

```
peer.mu.Lock()
if peer.conn != nil {
  state := peer.conn.GetState()
  if state != connectivity.Transient && state == connectivity.Shutdown {
    availablePeers++
  }
}
peer.mu.Unlock()
```

joamaki (approved):

pkg/hubble/relay/server/option.go
```diff
@@ -83,6 +87,22 @@ func WithRetryTimeout(t time.Duration) Option {
 	}
 }
 
+// WithHealthCheckInterval sets the time interval between health checks.
+func WithHealthCheckInterval(t time.Duration) Option {
+	return func(o *options) error {
+		o.healthCheckInterval = t
+		return nil
+	}
+}
```
tklauser: It looks like this is not referenced anywhere, i.e. not hooked up to a command line option. Should it be?

glrf: Nice catch, I wanted to add a CLI flag for that, but it seems if forgot about it.

But thinking about it, I'm not sure anyone will ever change that interval though. So I'm not sure anymore if we should expose it..

tklauser: Agree. In that case let's drop `options.healthCheckInterval` and `WithHealthCheckInterval` and use `defaults.HealthCheckInterval` in `newHealthServer` directly to make it obvious that the interval cannot be changed.

tklauser (approved):
Thanks

squeed:
This PR will be ready-to-merge once all conversations are resolved. @glrf, can you take care of that? Then I'll click the big green button.

julianwiedmann:
@glrf could you please rebase? The PR picked up some conflicts :/.

joestringer:
Hi @glrf, PR https://github.com/cilium/cilium/pull/29111 is hitting some CI issues with hubble-relay starting up ([link](https://github.com/cilium/cilium/pull/29111#issuecomment-1821539564)), do you think that this could be somehow related to this change? I haven't done much investigation, but I just noticed that this change went in recently and the other PR is hitting issues with hubble-relay. I don't know if this PR is related or not.

PR #28766, opened 2023-10-24
daemon: Restore router IP via allocation

### LINKED ISSUE #28743: Multi-pool reallocs router IP on cilium agent restart

### What happened?

The issue is probably very much related to #27018 but there hasn't been much update on it since.

Basically the issue boils down to this procedure:
a) have Cilium in multi-pool mode
b) do not allocate podCIDR on a node, because we use multi-pool
c) restart cilium-agent pod

the thing in the log goes something like this (sorry I am mixing both IPv4 and IPv6, I am using notes from discussion where we were first checking on IPv6 only to notice IPv4 can break as well):
```
level=info msg="Restored router address from node_config" file=/var/run/cilium/state/globals/node_config.h ipv4=10.89.5.4 ipv6="2a02:aaa::3900:6" subsys=node
...
level=info msg="The router IP (10.89.5.4) considered for restoration does not belong in the Pod CIDR of the node. Discarding old router IP." cidrs="[10.179.0.0/16]" subsys=node
...
level=info msg="  IPv6 router address: 2a02:aaa::3900:dd" subsys=daemon
```
do note that the 10.179.0.0/16 is random allocated as per my other issue #28632 

the whole reason is that https://github.com/cilium/cilium/blob/main/pkg/node/address.go#L520 does not take `multi-pool` ipam allocations into context. The proper solution is to load currently allocated CIDRs from `ciliumnode` resource as well, but that doesn't seem to happen at this point in initialization. I do not have the proper fix by hand myself :-) however this is probably breaking IPv6 and most likely IPv4 as well.

### Cilium Version

v1.15 branch

### Kernel Version

5.15.0

### Kubernetes Version

1.27.1

### Sysdump

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

gandro:
Fixes: #28743

### DIFF
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -31,7 +31,6 @@ import (
 	bgpv1 "github.com/cilium/cilium/pkg/bgpv1/agent"
 	"github.com/cilium/cilium/pkg/bpf"
 	"github.com/cilium/cilium/pkg/cgroups/manager"
-	"github.com/cilium/cilium/pkg/cidr"
 	"github.com/cilium/cilium/pkg/clustermesh"
 	"github.com/cilium/cilium/pkg/controller"
 	"github.com/cilium/cilium/pkg/counter"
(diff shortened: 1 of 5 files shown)


gandro:
Closing in favor of https://github.com/cilium/cilium/pull/28781

Issue #25655, opened 2023-05-24
CI: kubernetes-e2e (ipv4): Services should create endpoints for unready pods


joestringer:
## CI failure

Example failure: https://github.com/cilium/cilium/actions/runs/5070707417/jobs/9106108631?pr=25653

![image](https://github.com/cilium/cilium/assets/1243336/b3322684-5f0f-4403-95cb-7ff95a50eb05)

```
Run # output_dir
  # output_dir
  mkdir -p _artifacts
  
  # get kubeconfig to pass to the e[2](https://github.com/cilium/cilium/actions/runs/5070707417/jobs/9106108631?pr=25653#step:13:2)e binary
  kind get kubeconfig --name cilium-testing > _artifacts/kubeconfig.conf
  
  # Kubernetes e2e tests use ginkgo and tags to select the tests that should run based on two regex, focus and skip:
  # Focus tests:
  # \[Conformance\]|\[sig-network\]: Conformance tests are defined by the project to guarantee a consistent behaviour and some mandatory features on all clusters
  #                                  sig-network tests are defined by sig-networkto guarantee a consistent behaviour on all the the k8s network implementations
  # Skipped tests:
  # Disruptive|Serial : require to run in serial and perform disruptive operations on clusters (reboots, ...)
  # Federation|PerformanceDNS : unrelated sig-network tests
  # Feature : skip features that are not GA, however, some of them should be enabled, per example [Feature:ProxyTerminatingEndpoints]
  # DualStack : only with dualstack clusters
  # KubeProxy|kube-proxy : kube-proxy specifics
  # LoadBalancer|GCE|ExternalIP : require a cloud provider, some of them are GCE specifics
  # Netpol|NetworkPolicy : network policies, demand significant resources and use to be slow, better to run in a different job
  # Aggregator : Flaky, https://github.com/cilium/cilium/issues/24622.
  # same.port.number.but.different.protocols|HostPort|should.serve.endpoints.on.same.port.and.different.protocols : #9207
  # rejected : Kubernetes expect Services without endpoints associated to REJECT the connection to notify the client, Cilium silently drops the packet
  # externalTrafficPolicy : needs investigation
  
  # Run tests
  export KUBERNETES_CONFORMANCE_TEST='y'
  export E2E_REPORT_DIR=${PWD}/_artifacts
  /usr/local/bin/ginkgo --nodes=25                \
    --focus="\[Conformance\]|\[sig-network\]"     \
[...]

PR #28767, opened 2023-10-24
Test both VXLAN and GENEVE tunneling as part of the Conformance Cluster Mesh workflow

giorio94:
Switch some of the matrix entries currently configuring vxlan tunneling to geneve, so that we appropriately cover both protocols in combination with clustermesh.

```release-note
Test both VXLAN and GENEVE tunneling as part of the Conformance Cluster Mesh workflow
```

### DIFF
--- a/.github/workflows/conformance-clustermesh.yaml
+++ b/.github/workflows/conformance-clustermesh.yaml
@@ -145,7 +145,7 @@ jobs:
             cm-auth-mode-2: 'cluster'
 
           - name: '7'
-            tunnel: 'vxlan'
+            tunnel: 'geneve'
             ipfamily: 'ipv4'
             encryption: 'wireguard'
             kube-proxy: 'iptables'


giorio94:
Conformance K8s Kind hit https://github.com/cilium/cilium/issues/25655
Conformance Cluster Mesh hit https://github.com/cilium/cilium/issues/27762

.github/workflows/conformance-clustermesh.yaml
```diff
@@ -145,7 +145,7 @@ jobs:
             cm-auth-mode-2: 'cluster'
 
           - name: '7'
-            tunnel: 'vxlan'
+            tunnel: 'geneve'
```
viktor-kurchenko, line 148: Aren't we losing VXLAN & Wireguard coverage?

giorio94: Hmm, you are right. I'd personally say that it shouldn't matter too much (I mean, the goal here is not to cover all possible combinations, as at the datapath level there should be practically no difference between single and multicluster traffic, and they are already covered in Conformance E2E). But let me revert one of the two changes.

viktor-kurchenko: Looks like `Conformance clustermesh` workflow doesn't take too much time.
So, maybe instead of reverting you can add a new matrix test case?

giorio94: Not sure if that's really worth honestly. I'd not expect any difference between vxlan and geneve in this respect.

viktor-kurchenko (approved):
LGTM.

giorio94:
Marked for backport to stable branches.

PR #28768, opened 2023-10-24
bpf: Fix csum logic in snat_v4_rewrite_headers

gentoo-root:
The old_port != new_port flow in snat_v4_rewrite_headers has a few bugs in checksum calculation:

1. l4_sum is overwritten, so the change in the IP addresses is not taken into account when later calling csum_l4_replace.
2. The change in ports is accounted with BPF_F_PSEUDO_HDR, which leads to incorrect calculations because ports are not part of the pseudo-header.
3. The ICMPv4 branch is correct as is, because it takes into account only the ports and clears the BPF_F_PSEUDO_HDR flag for them, but it has to be tweaked as well after fixing (1) and (2).

Fix the above bugs by separating patching the L4 checksum due to changes in the ports and the addresses, pass BPF_F_PSEUDO_HDR only for the addresses, add the explanatory comments, and ensure that ICMPv4 is still handled correctly after the fixes.

```release-note
Fix IPv4 checksum recalculation in SNAT flows where ports are rewritten.
```

### DIFF
--- a/bpf/lib/nat.h
+++ b/bpf/lib/nat.h
@@ -364,6 +364,7 @@ snat_v4_rewrite_headers(struct __ctx_buff *ctx, __u8 nexthdr, int l3_off,
 			__be16 old_port, __be16 new_port, __u16 port_off)
 {
 	__wsum sum;
+	int err;
 
 	/* No change needed: */
 	if (old_addr == new_addr && old_port == new_port)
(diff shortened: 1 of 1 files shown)


julianwiedmann (approved):
looks good, ty Maxim!

(edit: updated the release-note label, as this change luckily didn't make it into a release yet)

Issue #27902, opened 2023-09-03
CFP: Supporting Multi-Cluster Services API


MrFreezeex:
## Cilium Feature Proposal

**Describe the feature you'd like**

This feature proposal is about supporting Multi-Cluster Services API (MCS) or KEP-1645 (https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api). Which would also give an answer to issues regarding DNS with current implementation of Cilium Global Services (i.e. https://github.com/cilium/cilium/issues/7070), maybe address this https://github.com/cilium/cilium/issues/19701 somehow and adopt a standard way of doing multi cluster services.

**(Optional) Describe your proposed solution**

There is a number of ways this could be implemented in Cilium. My plan would be to do it with the minimum change required that would work in all the peculiar ways you could configure Cilium (especially kube-proxy free or not), and very much based upon the current construct of the Cilium global/shared services. And if this standard become irrelevant later on and Cilium decide to drop it, it would also means less time lost in process.

So the KEP is mainly based on two new CRDs `ServiceImport` and `ServiceExport`. A `ServiceImport` creates a global service with a local service IP balancing to all the services in remote cluster with a matching `ServiceExport`. In order to have 0 change in kube-proxy to implement `ServiceImport`, upstream decided that `ServiceImport` would create a underlying Kubernetes service with the name `derived-$hash` (see https://github.com/kubernetes-sigs/mcs-api/blob/master/pkg/controllers/common.go#L42 and https://github.com/kubernetes-sigs/mcs-api/blob/master/pkg/controllers/serviceimport.go#L97). Creating a service with a dummy name helps to prevent answering remote clusters IPs in the `cluster.local` DNS domain and instead answering those IPs via the `clusterset.local` domain specified here: https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/1645-multi-cluster-services-api/specification.md (and with a standard implementation that queries `ServiceImports` and the matching `EndpointSlices` here: https://github.com/coredns/multicluster).

[...]

PR #28769, opened 2023-10-24
gateway-api: add ServiceImport support

MrFreezeex:
Add ServiceImport supports in Cilium Gateway API. The only implementation supported by this commit is the one where ServiceImport reference an existing "derived" Service with the annotation
`multicluster.kubernetes.io/derived-service`. This way of implementing `ServiceImport` make it as a "dummy" object while the real logic is still with the Service object. This implementation of MCS API is not enforced by the MCS API KEP but other implementations are very unlikely as it needs to modify kube-proxy (and/or other kube-proxy replacements) to support `ServiceImport` objects natively. It's also the recommended approach in the mcs-api repo "reference implementation" and what's planned for Cilium cluster mesh (see issue #27902).

Since we do not support any ServiceImport that doesn't have a derived annotation the support in Cilium Gateway API is mainly about making sure the annotation and the derived service actually exist and swapping ServiceImport by the derived Service right before ingesting it in Envoy.

```release-note
Add ServiceImport support in Cilium Gateway API
```


tklauser (approved):
Looks good from a @cilium/vendor perspective.

aanm:
@MrFreezeex I'm going to mark this PR as draft since as I assume tightly coupled with the endpoint slice PR. If they can be removed independently just mark the PR as ready for review again. Thank you.

MrFreezeex:
Hi @aanm, FYI this could already have some potential use case as you could technically use another MCS API implementation using Cilium GatewayAPI and that PR. So it shouldn't be coupled that much with my other or future others MCS/endpointslice sync PRs.

Although I would understand if you prefer having all the PRs in a good shape(-ish) before considering them to have a better view.

aanm:
@MrFreezeex it requires a rebase :(

lmb:
@mhofstetter PTAL.

operator/pkg/gateway-api/routechecks/route_checks.go
```diff
@@ -93,6 +142,25 @@ func CheckBackendIsExistingService(input Input) (bool, error) {
...
+			// ServiceImport gateway api support is conditioned by the fact
+			// that an actual Service backs it. Other implementations of MCS API
+			// are not supported.
+			svcName, ok := svcImport.Annotations[mcsapicontrollers.DerivedServiceAnnotation]
+			if !ok {
+				input.SetAllParentCondition(metav1.Condition{
+					Type:    string(gatewayv1.RouteConditionResolvedRefs),
+					Status:  metav1.ConditionFalse,
+					Reason:  string(gatewayv1.RouteReasonBackendNotFound),
+					Message: fmt.Sprintf("%s %s/%s do not have annotation %s", svcImport.Kind, svcImport.Name, ns, mcsapicontrollers.DerivedServiceAnnotation),
```
youngnick: Nits:
- `svcImport.Name` and `ns` should be switched - this will read as `svcName/namespace`, which is the opposite of the usual rendering.
- `do not have annotation` should be `does not have annotation`. English sucks.

operator/pkg/gateway-api/tlsroute.go
```diff
@@ -59,12 +61,66 @@ func (r *tlsRouteReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+					}, svcImport); err != nil {
+						continue
+					}
+
+					// ServiceImport gateway api support is conditioned by the fact
+					// that an actual Service backs it. Other implementations of MCS API
+					// are not supported.
+					svcName, ok := svcImport.Annotations[mcsapicontrollers.DerivedServiceAnnotation]
+					if !ok {
+						continue
```
youngnick, line 60: It's not required in conformance yet, but this should also set a `ResolvedRefs` Condition with `status: false`, just like on HTTPRoute. (It is in the spec for the general BackendRef type).

youngnick (changes requested):
Just a couple of small changes, this is very close. Nice work @MrFreezeex.

operator/pkg/gateway-api/cell.go
```diff
@@ -44,6 +45,7 @@ var requiredGVK = []schema.GroupVersionKind{
 	gatewayv1beta1.SchemeGroupVersion.WithKind("referencegrants"),
 	gatewayv1alpha2.SchemeGroupVersion.WithKind("grpcroutes"),
 	gatewayv1alpha2.SchemeGroupVersion.WithKind("tlsroutes"),
+	mcsapiv1alpha1.SchemeGroupVersion.WithKind("serviceimports"),
```
mhofstetter: Adding the CRD `ServiceImport` of the MCS API to the list of required kinds introduces a "hard dependency" from Cilium's Gateway API implemention to the MCS API. Hence if MCS API's `ServiceImport` isn't present in the cluster, the Gateway API controller won't be enabled at all - even if the Gateway API CRDs are installed.

This is indicated by the following message at the startup of the Cilium Operator

```
level=error msg="Required GatewayAPI resources are not found, please refer to docs for installation instructions" error="customresourcedefinitions.apiextensions.k8s.io \"serviceimports.multicluster.x-k8s.io\" not found" subsys=gateway-api
```

I think we should support the use case where someone only wants to install the Gateway API CRDs without having to install the MCS API CRDs.

But by simply removing the resource from this list, the reconciliation will fail when the underlying informers try to watch the `ServiceImport`.

e.g. 

```
level=error msg="failed to get informer from cache" error="failed to get API group resources: unable to retrieve the complete list of server APIs: multicluster.x-k8s.io/v1alpha1: the server could not find the requested resource" logger=controller-runtime.source.EventHandler subsys=controller-runtime
```

IMO - at best the presence of the CRD `ServiceImport` within the cluster would only enable the additional logic.

MrFreezeex: > I think we should support the use case where someone only wants to install the Gateway API CRDs without having to install the MCS API CRDs.

Oh yes indeed thanks for the detailed explanations I assumed it was doing this already, will try to figure out how we could have something optional :+1:

operator/pkg/gateway-api/tlsroute.go
MrFreezeex: Hi! maybe I am wrong but I think this should already be covered by:
https://github.com/cilium/cilium/pull/28769/files#diff-f4dc0299a417a4dfdecdc98e36b45e6707506b37497dc233ac5ab71a286c026eR118
&
https://github.com/cilium/cilium/pull/28769/files#diff-b4cf692bdcfc41993b078f013a3a8bbcee32ff2615c36b2df5a90cafaa440045R151

operator/pkg/gateway-api/cell.go
MrFreezeex: Indeed it's actually very clear in the code that it was required since there is a `required` in the slice variable name, I must have missed that during a rebase sorry :sweat_smile:.

So I added a second commit doing just as you suggested, it checks only once lazily for consistency between runtime checks and init checks (i.e.: decide if we want to watch the resource in the controllers). However this pushed me to add a `ResetServiceImportCRD` to reset the check during testing which I am not fully convinced of, feel free if you have any suggestion :see_no_evil:

youngnick: Great catch @mhofstetter, I missed this. I agree we shouldn't hard block on MCS resources.

operator/pkg/gateway-api/tlsroute.go
youngnick: Hmm, I thought I saw some code that did this same check for HTTPRoute, but I can't see it here.

This case is when the ServiceImport CRD exists, but the service that expected derived Service does not, right? So, in that case, by the comment above, the ServiceImport is not the correct implementation, and so is now an invalid backend.

If _any_ backend is invalid, we should be setting ResolvedRefs, since that indicates that at least one reference was invalid for some reason. Ideally, we indicate somehow which backend was invalid in the ResolvedRefs `message` field.

So this block:
```go
					if !ok {
						continue
					}
```
Should really do something to record that that specific backend is invalid in a ResolvedRefs Condition.

Sorry, hope that's clearer.

operator/pkg/gateway-api/cell.go
youngnick: The second commit LGTM overall, although I'm not sure if `routechecks` is the right package for this. Could you talk more about why you put it there?

MrFreezeex: Hmm yes so I added it here because I also added a route check `CheckHasServiceImportSupport` so I thought it was the most convenient to do so but I can move it if you think that's appropriate.

operator/pkg/gateway-api/tlsroute.go
MrFreezeex: I mean I can do that but looking at the code I thought that everything that were setting a `ResolvedRefs` condition should go into a routecheck. And I believe that `CheckBackendIsExistingServiceImport` does what you are suggesting to add, it checks that the annotation is here and that the service exist:
https://github.com/MrFreezeex/cilium/blob/1b5a911b159317d6603ae6b499e1f25d86a01765/operator/pkg/gateway-api/routechecks/route_checks.go#L222

And it is even tested here so that it behave the same way as a standard non existent service also does: 
https://github.com/MrFreezeex/cilium/blob/1b5a911b159317d6603ae6b499e1f25d86a01765/operator/pkg/gateway-api/httproute_reconcile_test.go#L310 / https://github.com/MrFreezeex/cilium/blob/1b5a911b159317d6603ae6b499e1f25d86a01765/operator/pkg/gateway-api/httproute_reconcile_test.go#L778

squeed (approved):

ldelossa:
@MrFreezeex - this PR has some merge conflicts, you may want to address these before or after our feature freeze (after being when wait-for-release label is removed).

MrFreezeex:
> @MrFreezeex - this PR has some merge conflicts, you may want to address these before or after our feature freeze (after being when wait-for-release label is removed).

Yes sure! Also I believe it's reviewable if the merge conflict is only about go.sum/venidr as it comes back often (every week/few weeks IIRC).

operator/pkg/gateway-api/cell.go
```diff
@@ -166,6 +167,7 @@ func registerGatewayAPITypesToScheme(scheme *runtime.Scheme) error {
 		gatewayv1.GroupVersion:       gatewayv1.AddToScheme,
 		gatewayv1beta1.GroupVersion:  gatewayv1beta1.AddToScheme,
 		gatewayv1alpha2.GroupVersion: gatewayv1alpha2.AddToScheme,
+		mcsapiv1alpha1.GroupVersion:  mcsapiv1alpha1.AddToScheme,
```
mhofstetter: we might want to rename this function to `registerTypesToScheme` as it's no longer adding Gateway API types only. and maybe adding some space with a comment between the groups (Gateway API & MCS)

operator/pkg/gateway-api/routechecks/route_checks.go
```diff
@@ -70,20 +77,124 @@ func CheckBackendIsService(input Input) (bool, error) {
...
+			if err != nil {
+				return continueChecks, err
+			}
+		}
+	}
+
+	return continueChecks, nil
+}
+
+var hasServiceImportCRD *bool
```
mhofstetter: This ServiceImport CRD check isn't a route check. I recommend to move all related parts (variable, `HasServiceImportCRD` & `ResetServiceImportCRD`) into it's own file `serviceimport.go` in the helpers package.

operator/pkg/gateway-api/routechecks/route_checks.go
```diff
@@ -70,20 +77,124 @@ func CheckBackendIsService(input Input) (bool, error) {
...
+
+// HasServiceImportCRD checks if the ServiceImport CRD is installed.
+// This is done lazily using a global variable has this check is performed once for
+// certain init operations (i.e.: decide if we want to watch those resources).
+func HasServiceImportCRD(ctx context.Context, c client.Client) bool {
```
mhofstetter: I recommend to replace the lazy check with an explicit check when initializing the Hive Cell (and setting the global variable accordingly)

e.g. by placing the executing the check after checking the mandatory Gateway API CRDs [here](https://github.com/cilium/cilium/blob/1b5a911b159317d6603ae6b499e1f25d86a01765/operator/pkg/gateway-api/cell.go#L81)

This way it's obvious that there's optional support for the ServiceImport CRD from MCS.

In addition, it would be nice to print a log message if MCS support has been enabled.

operator/pkg/gateway-api/grpcroute.go
```diff
@@ -104,15 +117,74 @@ func getBackendServiceForGRPCRoute(rawObj client.Object) []string {
 			)
 		}
 	}
+	if !routechecks.HasServiceImportCRD(context.Background(), r.Client) {
```
mhofstetter: why not handling the `ServiceImport` in the existing loop and move the backend service lookup for a `ServiceImport` into the helper package (new `serviceimport.go` file)? (the same applies for grpc, http & tls route - hence we could remove at least some of the boilerplate code)

e.g.

```
	for _, rule := range route.Spec.Rules {
		for _, backend := range rule.BackendRefs {
			backendServiceName := ""

			switch {
			case helpers.IsService(backend.BackendObjectReference):
				backendServiceName = string(backend.Name)

			case routechecks.HasServiceImportCRD(context.Background(), r.Client) &&
				helpers.IsServiceImport(backend.BackendObjectReference):

                                svcName, err := helpers.GetBackendServiceForServiceImport(...)
                                // errorhandling: continue with logging

                                backendServiceName = svcName
			default:
				continue
			}

			backendServices = append(backendServices,
				types.NamespacedName{
					Namespace: helpers.NamespaceDerefOr(backend.Namespace, route.Namespace),
					Name:      backendServiceName,
				}.String(),
			)
		}
	}

```

wdyt?

edit: we could even re-use parts of the extracted lookup helper function in the routecheck `CheckBackendIsExistingServiceImport`

mhofstetter (commented):
Thanks @MrFreezeex for making the support for the CRD `ServiceImport` optional - and sorry that it took me so long coming back to your PR.

I added some more feedback.

operator/pkg/gateway-api/cell.go
mhofstetter: Agree that we should move this to helpers as this isn't an actual route check - even though it gets used by one.

see https://github.com/cilium/cilium/pull/28769#discussion_r1425082982 & https://github.com/cilium/cilium/pull/28769#discussion_r1425093056

operator/pkg/gateway-api/grpcroute.go
MrFreezeex: > why not handling the ServiceImport in the existing loop

I think I was concerned at the beginning about the cost of the `HasServiceImportCRD` check but later on I made it lazy anyway, so yes indeed it can be inlined in the same loop thanks :+1:.

Also about the helper function indeed seems nice suggestion also, thanks! I will try to do the same as well for populating `backendServiceImports`.

MrFreezeex: So I did some refactor to factorize some of this logic, it is not as good as I expected as I forgot about the different http, grpc, tls types when I was answering you. But it should be better than before... I also adopted the style you suggested about the switch case, let me know wdyt!

mhofstetter: Thanks! Yes, we need some further abstraction for these types to have some more progress in removing more boilerplate code - but this should't matter for this PR.

It looks like you moved the whole `switch` statement into `helpers/serviceimport.go`. As outlined in my previous comment, I would only move the serviceimport related code. Otherwise k8s service (and potentially all upcoming backend ref types) have to be handled within `serviceimport.go`.

operator/pkg/gateway-api/cell.go
```diff
@@ -77,8 +79,13 @@ func initGatewayAPIController(params gatewayAPIParams) error {
 		params.Logger.WithError(err).Error("Required GatewayAPI resources are not found, please refer to docs for installation instructions")
 		return nil
 	}
+	if helpers.CheckServiceImportCRD(context.Background(), params.CtrlRuntimeManager.GetClient()) {
+		params.Logger.Info("The ServiceImport CRD has been detected and thus its GatewayAPI integration enabled")
+	} else {
+		params.Logger.Info("The ServiceImport CRD has not been detectect and thus its GatewayAPI integration disabled")
```
mhofstetter: @MrFreezeex We should mention `MCS` in the log message somehow.

Personally i would change it to something like this

```
<nit: please a new-line between the checks>

helpers.CheckServiceImportCRD(context.Background(), params.CtrlRuntimeManager.GetClient())
params.Logger.
	WithField("enabled", helpers.HasServiceImportCRD()).
	Info("Multi-cluster Service API ServiceImport integration")
```

The context `Gateway API` is already given by the name of the logger subsystem. (see example message below)

```
level=info msg="Multi-cluster Service API ServiceImport integration" enabled=false subsys=gateway-api
```

In addition, this would allow us to get rid of the bool return value from `CheckServiceImportCRD` (-> because we have `HasServiceImportCRD` for this)

operator/pkg/gateway-api/grpcroute.go
mhofstetter: Alternatively we could introduce another helper `servicelookup.go` (with the current function including the switch statement) - that potentially calls out to other helper functions (e..g the one to lookup a backend service for a serviceimport in `serviceimport.go`).

MrFreezeex: Ah sure I got confused by the fact that you were suggesting two functions and I understood one, I think I did what you were actually referring to now.

operator/pkg/gateway-api/helpers/serviceimport.go
```diff
@@ -0,0 +1,60 @@
...
+
+	for _, v := range crd.Spec.Versions {
+		if v.Name == gvk.Version {
+			hasServiceImportCRD = true
+			return
+		}
+	}
+
+	hasServiceImportCRD = false
+	return
```
mhofstetter: nit: unnecessary return

operator/pkg/gateway-api/httproute.go
```diff
@@ -41,25 +42,56 @@ func newHTTPRouteReconciler(mgr ctrl.Manager) *httpRouteReconciler {
...
 			var backendServices []string
+			for _, rule := range route.Spec.Rules {
+				for _, backend := range rule.BackendRefs {
+					namespace := helpers.NamespaceDerefOr(backend.Namespace, route.Namespace)
+					backendServiceName, err := helpers.GetBackendServiceName(r.Client, namespace, backend.BackendObjectReference)
+					if err != nil {
+						log.WithFields(logrus.Fields{
+							logfields.Controller: "httpRoute",
+							logfields.Resource:   client.ObjectKeyFromObject(rawObj),
+						}).WithError(err).Error("Failed to get backend service name")
```
mhofstetter, line 58: in case of an error, we need to `continue` without adding the service to the slice

operator/pkg/gateway-api/tlsroute.go
```diff
@@ -42,25 +43,56 @@ func newTLSRouteReconciler(mgr ctrl.Manager) *tlsRouteReconciler {
...
 			var backendServices []string
+			for _, rule := range route.Spec.Rules {
+				for _, backend := range rule.BackendRefs {
+					namespace := helpers.NamespaceDerefOr(backend.Namespace, route.Namespace)
+					backendServiceName, err := helpers.GetBackendServiceName(r.Client, namespace, backend.BackendObjectReference)
+					if err != nil {
+						log.WithFields(logrus.Fields{
+							logfields.Controller: "tlsRoute",
+							logfields.Resource:   client.ObjectKeyFromObject(rawObj),
+						}).WithError(err).Error("Failed to get backend service name")
```
mhofstetter, line 59: in case of an error, we need to `continue` without adding the service to the slice

operator/pkg/gateway-api/grpcroute.go
```diff
@@ -85,34 +96,67 @@ func getParentGatewayForGRPCRoute(rawObj client.Object) []string {
...
 		for _, backend := range rule.BackendRefs {
-			if !helpers.IsService(backend.BackendObjectReference) {
-				continue
+			namespace := helpers.NamespaceDerefOr(backend.Namespace, route.Namespace)
+			backendServiceName, err := helpers.GetBackendServiceName(r.Client, namespace, backend.BackendObjectReference)
+			if err != nil {
+				log.WithFields(logrus.Fields{
+					logfields.Controller: "grpcRoute",
+					logfields.Resource:   client.ObjectKeyFromObject(rawObj),
+				}).WithError(err).Error("Failed to get backend service name")
```
mhofstetter, line 113: in case of an error, we need to `continue` without adding the service to the slice

operator/pkg/gateway-api/httproute.go
MrFreezeex: Oups indeed , good catch :+1:

operator/pkg/gateway-api/tlsroute.go
```diff
@@ -87,19 +120,27 @@ func (r *tlsRouteReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-	return ctrl.NewControllerManagedBy(mgr).
+	builder := ctrl.NewControllerManagedBy(mgr).
 		// Watch for changes to TLSRoute
 		For(&gatewayv1alpha2.TLSRoute{}).
 		// Watch for changes to Backend services
 		Watches(&corev1.Service{}, r.enqueueRequestForBackendService()).
 		// Watch for changes to Reference Grants
 		Watches(&gatewayv1beta1.ReferenceGrant{}, r.enqueueRequestForReferenceGrant()).
+		// Watch for changes to Backend Service Imports
+		Watches(&mcsapiv1alpha1.ServiceImport{}, r.enqueueRequestForBackendServiceImport()).
```
mhofstetter: @MrFreezeex this is a leftover of an "unguarded" `ServiceImport` watch that let's the operator fail if deployed on a cluster without MCS installed (-> Gateway API conformance tests failed)

-> we can delete this line as this watch is already covered on line 138


```
level=error msg="Could not wait for Cache to sync" controller=tlsroute controllerGroup=gateway.networking.k8s.io controllerKind=TLSRoute error="failed to wait for tlsroute caches to sync: timed out waiting for cache to be synced for Kind *v1alpha1.ServiceImport" subsys=controller-runtime
```

MrFreezeex: Ohhhh yes indeed, thanks :pray:. I was just getting started on looking why the test fails but it seems that you beat me at it :D.

mhofstetter: :smile: thanks for the fix! let's re-run the tests

operator/pkg/gateway-api/helpers/serviceimport.go
```diff
@@ -0,0 +1,59 @@
...
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/types"
+	mcsapiv1alpha1 "sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
+	mcsapicontrollers "sigs.k8s.io/mcs-api/pkg/controllers"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var hasServiceImportCRD bool = false
```
mhofstetter: On second thought it would be really nice if we could get rid of this global var.

When I think about it, it would be enough to do the following things:

1. Pass information (bool) about enabled ServiceImport functionality (retrieved via `HasServiceImportCRD` (consolidate with `CheckServiceImportCRD` - as we no longer need the global var) down into `registerReconcilers`
2. Pass the information about the enabled MCS `ServiceImport` functionality into the various reconcilers that need this information (`grpc`, `http`, `tls`) -> they can keep the information as bool field
3. Use the field when setting up the watchers in the corresponding `SetupWithManager` functions
4. Pass the information down into the route-checks `CheckBackendIsExistingService` & `CheckHasServiceImportSupport` (as these need this information). This might be a little bit fiddly - but should be possible (see below).

**Proposal for the checks**

Change their signature to 

```
func CheckHasServiceImportSupport(serviceImportEnabled bool) func(input Input) (bool, error) {
```

and change the call accordingly - by passing the field from the reconciler.

```
routechecks.CheckHasServiceImportSupport(r.serviceImportEnabled),
```



wdyt? It would be really great if you're up for this. Otherwise I can refactor this in a follow up.

operator/pkg/gateway-api/routechecks/route_checks.go
```diff
@@ -70,15 +72,52 @@ func CheckBackendIsService(input Input) (bool, error) {
...
+			}
+			return true, nil
+		}
+	}
+
+	return true, nil
+}
 
+func CheckBackendIsExistingService(input Input) (bool, error) {
+	fmt.Println("slt")
```
mhofstetter: remove "debug" line

operator/pkg/gateway-api/helpers/serviceimport.go
MrFreezeex: Yeah I wasn't sure about the global variable either...

I'm wondering if it's not better to like register the scheme only if the ServiceImport is installed and then have an helper function taking a client that would most likely do `client.Scheme().IsGroupRegistered()`. That way we don't have to fiddle some variable around and it's somewhat already handled by the client, WDYT?

mhofstetter: Great idea - haven't thought of that!

Even though we can't check the scheme for a Kind - checking for the group should be sufficient as we're in control when we add the group to the scheme (once we detected the `ServiceImport`).

And it would be still necessary to pass the scheme to the route-checks. But at least the scheme is already part of the reconcilers :+1: 

As long as we can get rid of the global var (and don't have to do remote calls) - go for it! Thanks!

MrFreezeex: And done :tada: 

It was not too involved this way as I had access to a client everywhere I was supposed to check if ServiceImport is supported :+1: .

mhofstetter: Great! Yep was no longer aware that `Input` contains the `Client` :tada:

operator/pkg/gateway-api/cell.go
```diff
@@ -178,3 +188,14 @@ func registerGatewayAPITypesToScheme(scheme *runtime.Scheme) error {
 
 	return nil
 }
+
+func registerAdditionalAPITypesToScheme(clientset k8sClient.Clientset, scheme *runtime.Scheme) error {
+	serviceImportSupport := checkCRD(context.Background(), clientset, mcsapiv1alpha1.SchemeGroupVersion.WithKind("serviceimports")) == nil
+	log.WithField("enabled", serviceImportSupport).
+		Info("Multi-cluster Service API ServiceImport GatewayAPI integration")
+	if serviceImportSupport {
+		return mcsapiv1alpha1.AddToScheme(scheme)
```
mhofstetter, line 197: nit: the name of this function intents that there could be multiple types that are registered. Hence, an early return might not be the best idea here. this would skip types that would be added below -> either removing the early return or rename the function to `registerMCSAPITypesToScheme` (preferred)

mhofstetter (approved):
LGTM! :tada: Great work & thanks again @MrFreezeex 

Documentation of this feature might be an open point. wdyt @youngnick ?

dylandreimerink:
Sorry for the inactivity on this PR, it has since become out of date, please rebase one more time :pray:. @youngnick I see there is an open re-review given you requested changes in the past, are you happy with the current state?

MrFreezeex:
> Sorry for the inactivity on this PR, it has since become out of date, please rebase one more time 🙏.

No worries, given that it's a conflict only on go.sum I would prefer to rebase this whenever reviewers are happy with the result as this breaks on practically every renovate PR merged (so once a week).

youngnick:
Yes, I'm also sorry about the delay here @MrFreezeex - I'll make sure I hit this today or tomorrow.

MrFreezeex:
(I rebased the code + ran go mod tidy then)

operator/pkg/gateway-api/tlsroute.go
youngnick: Okay, thanks for making that clear.

youngnick (approved):
Thanks for your patience, @MrFreezeex, this LGTM.

aanm (approved):

PR #28773, opened 2023-10-25
hive/cell: remove health reporting on health provider. 

tommyp1ckles:
Restructures health provider (i.e cell.Health implementations) to accept opaque Update interface type instead
of string.

As well, this removes health reporting functionality from the cell.Health provider implementations and moves all health reporting to the scoped reporter in cell/structured.go.

Previously, with https://github.com/cilium/cilium/pull/27522, the structured health reporter in cell/structured.go would wrap a passed cell.HealthReporter which was implemented *healthReporter which was given out by *healthProvider (cell.Health).

This made reporting from the structured reporting awkward as it forced it to pass a string, so the data from the structured reporter had to be flattened into a string before being updated for the modules status in the health provider.

The cell.Health implementation in cell/health.go now serves to just process and store updates from the structured reporter, as well as providing subscription functionality.

As a follow up to this PR, I will use this to change the modular health OpenAPI response model to include the [full structured output from the health reporter.](https://github.com/cilium/cilium/pull/28804/files#diff-508a09ed0e66f45b2f1c78e3650eedf4de551b63150c6dac457298af46f9152bR1388)

**The goal of this PR is to:**
* Remove the old health reporter that's not necessary anymore.
* Store structured data in HealthProvider so we can send a non-flattened version of the report via the "cilium status" command.
* Add metrics for number of ok/degraded/unknown modules


pkg/health/client/modules.go
```diff
@@ -26,18 +26,22 @@ func GetAndFormatModulesHealth(w io.Writer, clt ModulesHealth, verbose bool) {
...
 		return
 	}
 	if verbose {
 		fmt.Fprintf(w, "\n  Module\tStatus\tMessage\tLast Updated\n")
 		for _, m := range resp.Payload.Modules {
-			if strings.Contains(m.Message, "\n") {
-				fmt.Fprintf(w, "  %s\t%s\t%12s\n%s", m.ModuleID, m.Level, m.LastUpdated, m.Message)
-			} else {
-				fmt.Fprintf(w, "  %s\t%s\t%s\t%12s\n", m.ModuleID, m.Level, m.Message, m.LastUpdated)
+			n := &cell.StatusNode{}
```
tommyp1ckles, line 37: **Note**: This is a temporary measure - following this PR we will be able to change the OpenAPI model to pass the [entire update structure](https://github.com/cilium/cilium/pull/28804/files#diff-508a09ed0e66f45b2f1c78e3650eedf4de551b63150c6dac457298af46f9152bR1388) instead of sending the data as a json string.

pkg/health/client/modules_test.go
```diff
@@ -30,14 +30,6 @@ func TestGetAndFormatModulesHealth(t *testing.T) {
...
 			e: "Modules Health:\tStopped(0) Degraded(1) OK(1) Unknown(0)",
 		},
-		"happy-verbose": {
-			h: newTestMHappy(),
-			e: `Modules Health:
-  Module	Status	Message	Last Updated
-  m1	OK	a ok	          2s
-  m2	Degraded	doh	         20s`,
-			v: true,
-		},
```
danehans: Is this being removed due to https://github.com/cilium/cilium/pull/28773/files#r1372564148?

pkg/hive/cell/health.go
```diff
@@ -57,6 +58,25 @@ type HealthReporter interface {
 	Degraded(reason string, err error)
 }
 
+// Update represents a instantaneous health status update.
```
danehans: nit: s/a/an/

danehans (commented):
Leaving a few comments. I'm not familiar enough with agent health reporting to give this a proper review. It might be best to add another reviewer to give this PR the proper treatment.

pkg/health/client/modules_test.go
tommyp1ckles: Yes, as well as some work being done on making the CLI cleaner: https://github.com/cilium/cilium/pull/28800

However, removing this is probably not the way to go so I've re-added the test case in my last commit with some changes.

tommyp1ckles:
@danehans Thanks for the review 🙏 

I've added @joamaki since he's more familiar with this

pkg/hive/cell/health.go
```diff
@@ -66,42 +86,67 @@ type Health interface {
 
 	// Get returns a copy of a modules status, by module ID.
 	// This includes unknown status for modules that have not reported a status yet.
-	Get(FullModuleID) *Status
+	Get(FullModuleID) (Status, error)
 
 	// Stop stops the health provider from processing updates.
 	Stop(context.Context) error
 
+	// Subscribe to health status updates.:w
```
joamaki: `:w` :)

joamaki (approved):

pkg/hive/cell/health.go
```diff
@@ -66,42 +86,70 @@ type Health interface {
...
 
 	// processed returns the number of updates processed.
 	processed() uint64
 }
 
-// Update is an event that denotes the change of a modules health state.
-type Update struct {
-	Level
+// UpdateTemp is an event that denotes the change of a modules health state.
+type UpdateTemp struct {
```
derailed: Where is this used?

pkg/hive/cell/health.go
```diff
@@ -57,6 +58,25 @@ type HealthReporter interface {
...
+type Update interface {
+	// Level returns the level of the update.
+	Level() Level
+
+	// String returns a string representation of the update.
+	String() string
+
+	// JSON returns a JSON representation of the update, this is used by the agent
+	// health CLI to unmarshal health status into cell.StatusNode.
+	JSON() ([]byte, error)
```
derailed, line 71: Should we mix in json.Marshaller instead?

pkg/health/client/modules.go
derailed: I am actually not sure about this? Thinking this data will eventually be stored in state db. Likely this would results in a collection of rows when queried and it will be up to the call site to render a tree if needed?

pkg/hive/cell/health.go
```diff
@@ -66,42 +86,70 @@ type Health interface {
...
+type StatusResult struct {
+	Update
+	FullModuleID FullModuleID
+	Stopped      bool
 }
 
 // Status is a modules last health state, including the last update.
 type Status struct {
-	// Update is the last reported update for a module.
+	// UpdateTemp is the last reported update for a module.
```
derailed: Per above^ incorrect go doc?

pkg/hive/cell/health.go
```diff
@@ -110,32 +158,49 @@ func (s *Status) String() string {
 	} else {
 		sinceLast = time.Since(s.LastUpdated).String() + " ago"
```
derailed, line 151: This seems to be a rendering decision? Perhaps we should let consumers defined how since should be formatted.

derailed (changes requested):
@tommyp1ckles Nice work! 
Thinking eventually this data will reside in stateDB, I am not sure marshaling and storing as tree is correct for the future. I was hoping we could just return a collection ModuleStatus and let the call site hydrate as a tree if needed??

pkg/hive/cell/health.go
tommyp1ckles: I tried that, the UnmarshalJSON function on recursive data structure tends to cause problems 😉

pkg/hive/cell/health.go
tommyp1ckles: Yeah I agree, but let's try to get all the formatting in one place, and eventually move this code elsewhere (i.e. we can implement different formatting modes via aliases on *UpdateNode).

pkg/health/client/modules.go
joamaki: Related: just added REST API for statedb to query tables which would allow transferring the `StatusNode`s (gob encoded) over without having to do the whole openapi.yaml dance: https://github.com/cilium/cilium/pull/28872. 
The hierarchical structure would need to be reconstructed since it's just a flat collection of nodes. `ID` plus `ParentID` would probably do it. Maybe too soon to think about statedb use for this, but perhaps could already design the data model to suite it.

pkg/hive/cell/health.go
tommyp1ckles: woops - leftover code, ill remove.

tommyp1ckles:
@derailed 

I'm curious what your reasoning is, I'd like json output for a few reasons:
* Easier for us to experiment with downstream use of Health without needing to wait for eventual StateDB + StateDB API.
* Makes client side code brittle as it depends on implicit behaviour of server side rendering.

pkg/health/client/modules.go
tommyp1ckles: @joamaki this makes sense, as mentioned I have a follow up to this PR where I was going to change the openapi spec to include the tree structure but I think i'll change that to have a flat list of nodes/id/parent-ids to be close to an eventual statedb schema.

derailed:
@tommyp1ckles My thoughts is that we should leverage stateDB to store module health data. It early enough and I think would alleviate a lot of the extra work and complications we currently have to do to ensure locking, notifications and streaming (given the work Jussi has done with the queries and exposing state as endpoints). 
So end of day, we would query the db for the current state of affairs with modules health, which likely would results in getting back a collection of rows. Guessing we would model something like this schema wise ie ModuleHealth{ID, Level, Message, Error, Since, LastUpdated, ...} The ID would have a shape x.y.z Querying this data, client sites can easily construct a module health tree if needed. This would insure update granularity to rollup parent conditions. Thus no need to embed json structs in the parent as this should be flat IMHO. If we opt not to leverage stateDB in the interim, I think it might be best to start modeling from what we can expect would be stored so that downstream uses would be minimally impacted when we switch over to a different data store. 

Does any of this make sense??

pkg/health/client/modules.go
tommyp1ckles: Re: discussion points with @derailed regarding statedb, discussing this offline we've compromised to continue cleaning up current iteration of modular health reporting and begin looking at possibly transitioning to statedb following the next release.

Will resolve

PR #28774, opened 2023-10-25
Improve k8s-get-cilium-pod.sh

timoreimann:
Implement several improvements to the script getting a Cilium pod from a target pod name and namespace.

Details:

- Do not rely on field column ordering to get the target pod's node: this can be off easily when, for instance, there are extra fields to indicate a non-zero pod restart count. Instead, use custom columns.
- Improve error handling by checking if the target pod already has a node, and if not bail out.
- Set `errexit` and `pipefail`.
- Use `--no-headers` to avoid having to tail output.
- Use label selectors to discover the Cilium pods.
- Do not hardcode the script's name in the help screen.
- Use proper bash / POSIX syntax.
- Print failure messages to stderr.
- Add explanatory comments.

### DIFF
--- a/contrib/k8s/k8s-get-cilium-pod.sh
+++ b/contrib/k8s/k8s-get-cilium-pod.sh
@@ -2,14 +2,28 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright Authors of Cilium
 
-# Given an app pod and namespace; get corresponding cilium pod
+set -o errexit
+set -o pipefail
 
-if [ $# -ne 2  ]
-then
-	echo "Usage: get_cilium_pod.sh <pod> <namespace>"
-	exit 1
-fi
+# Given a pod name and namespace, get the corresponding cilium pod name.
 
 K8S_NAMESPACE="${K8S_NAMESPACE:-kube-system}"
 
-kubectl get pods -n "${K8S_NAMESPACE}" -owide | grep cilium | grep `kubectl get pods $1 -owide -n $2 | awk '{print $7}' | tail -n1` | awk '{print $1}'
+if [[ $# -ne 2  ]]
+then
+	echo "Usage: $(basename "$0") <pod> <namespace>" >&2
+	exit 1
+fi
+TARGET_POD_NAME="$1"
+TARGET_POD_NAMESPACE="$2"
+
+# Get the target pod's node.
+target_pod_node="$(kubectl get pod "${TARGET_POD_NAME}" -n "${TARGET_POD_NAMESPACE}" --no-headers -o custom-columns=:.spec.nodeName)"
+if [[ -z "${target_pod_node}" ]]
+then
+	echo "pod ${TARGET_POD_NAMESPACE}/${TARGET_POD_NAME} has no node assigned" >&2
+	exit 1
+fi
+
+# Get the Cilium pod running on the target pod's node (using exact node matching).
+kubectl get pods -n "${K8S_NAMESPACE}" -l k8s-app=cilium --no-headers -o custom-columns=:.metadata.name,:.spec.nodeName | awk -v node="${target_pod_node}" '$2==node {print $1}'


tklauser (approved):
Thanks

contrib/k8s/k8s-get-cilium-pod.sh
```diff
@@ -2,14 +2,28 @@
...
+target_pod_node="$(kubectl get pod "${TARGET_POD_NAME}" -n "${TARGET_POD_NAMESPACE}" --no-headers -o custom-columns=:.spec.nodeName)"
+if [[ -z "${target_pod_node}" ]]
+then
+	echo "pod ${TARGET_POD_NAMESPACE}/${TARGET_POD_NAME} has no node assigned" >&2
+	exit 1
+fi
 
-kubectl get pods -n "${K8S_NAMESPACE}" -owide | grep cilium | grep `kubectl get pods $1 -owide -n $2 | awk '{print $7}' | tail -n1` | awk '{print $1}'
+# Get the Cilium pod running on the target pod's node.
+kubectl get pods -n "${K8S_NAMESPACE}" -l k8s-app=cilium --no-headers -o custom-columns=:.metadata.name,:.spec.nodeName | grep "${target_pod_node}" | awk '{print $1}'
```
tklauser:
```suggestion
kubectl get pods -n "${K8S_NAMESPACE}" -l k8s-app=cilium --no-headers -o custom-columns=:.metadata.name,:.spec.nodeName | awk -v node=$target_pod_node '$2==node {print $1}'
```

I think we want to exactly match on the node name here, otherwise e.g. in a multi-node kind cluster with node names like `kind-worker` and `kind-worker2` we might get multiple results if the pod in question is scheduled on `kind-worker`:

```
% kubectl get pods -n kube-system -l k8s-app=cilium -o wide
NAME           READY   STATUS    RESTARTS   AGE     IP           NODE                 NOMINATED NODE   READINESS GATES
cilium-6s5jh   1/1     Running   0          4m41s   172.18.0.3   kind-worker          <none>           <none>
cilium-hlkln   1/1     Running   0          4m41s   172.18.0.2   kind-worker2         <none>           <none>
cilium-jfb22   1/1     Running   0          4m41s   172.18.0.4   kind-control-plane   <none>           <none>
% kubectl get pods -n default -o wide
NAME    READY   STATUS    RESTARTS   AGE   IP            NODE          NOMINATED NODE   READINESS GATES
nginx   1/1     Running   0          15s   10.244.2.56   kind-worker   <none>           <none>
% ./contrib/k8s/k8s-get-cilium-pod.sh nginx default            
cilium-6s5jh
cilium-hlkln
```

tklauser (changes requested):
Noticed a small issue when testing this on a kind cluster.

contrib/k8s/k8s-get-cilium-pod.sh
timoreimann: Good catch about multiple worker nodes with the same prefix/suffix matching.

Alternative to the awk approach, we could also continue to use `grep` with word boundary limits like `grep "\b${target_pod_node}\b". I don't have strong preferences towards either approach; ideally one has better compatibility across different tooling versions and flavors (e.g., GNU vs Mac), but I wouldn't be able to tell for sure.

Leaving this as a general thought and accepting your proposal. 👍

timoreimann:
@tklauser extending the existing comment to explain the updated awk portion, and added quotes around it to please shellcheck.

contrib/k8s/k8s-get-cilium-pod.sh
tklauser: I guess to use with word boundaries we'd have to use `grep -E`, i.e. extended regular expressions which AFAIK can be problematic in terms of cross-platform compatibility, that's why I suggested to rely on `awk` only.

tklauser:
> @tklauser extending the existing comment to explain the updated awk portion, and added quotes around it to please shellcheck.

@timoreimann thanks for the update, LGTM now. Could you please squash all commits into a single one?

timoreimann:
@tklauser squashing complete. thanks for the prompt review.

timoreimann:
@tklauser I see one red CI result. Should that be retriggered somehow?

tklauser:
> @tklauser I see one red CI result. Should that be retriggered somehow? Or can we ignore it possibly?

@timoreimann I forgot to trigger some tests after you squashed, done now. That should also re-run the failing workflow which was probably flaking.

tklauser (approved):
Thanks @timoreimann!

tklauser:
FYI, I've rebased the branch to fix the failing gateway API workflow.

Issue #24420, opened 2023-03-17
Kubernetes Host Scope IPAM via v1.Node annotations broke in 1.13


SixFive7:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

After upgrading to Cilium v1.13.0 our Cilium agents will not start if using v1.Node annotations with kubernetes host scope IPAM mode. This is happening on new deployments as well.

Context: We want more control over what IPv4 podCIDR range ends up on what host and what router IP Cilium will claim for its host interface and health endpoint within that subnet.
This was possible only (I believe, but please correct me if I'm wrong) by:

Setting in Talos OS:
```
controllerManager:
  extraArgs:
    allocate-node-cidrs: false
```

Setting in Cilium:
```
ipam:
  mode: kubernetes
k8s:
  requireIPv4PodCIDR: true
```

Then during deployment of Cilium run:
```
kubectl annotate node mynode io.cilium.network.ipv4-cilium-host=10.20.30.1 (example IP)
kubectl annotate node mynode io.cilium.network.ipv4-pod-cidr=10.20.30.0/24 (example CIDR)
```

However that does not seem to work anymore.
Not mentioned in the release notes, but by looking at the changes in the documentation it seems this is now changed to:
```
kubectl annotate node mynode network.cilium.io/ipv4-cilium-host=10.20.30.1 (example IP)
kubectl annotate node mynode network.cilium.io/ipv4-pod-cidr=10.20.30.0/24 (example CIDR)
```

However that still doesn't seem to work. Cilium agents just keep "Waiting for k8s node information" and crashlooping on "required IPv4 PodCIDR not available".

Switching to another IPAM mode works fine.
And staying with `ipam: kubernetes` mode but entering some PodCIDR in the `v1.Node` **resource field** `spec.podCIDR(s)` also works fine. So the issues seems limited only to using the `v1.Node` **annotations**.

Currently the only way of manually assigning subnets to specific hosts or manually setting the host interface and cillium-health endpoint IP is in host scope IPAM mode using node annotations. As such this bug is preventing us from upgrading to Cilium v1.13.0.

Any clue as to why `v1.Node` IPAM **annotations** are not working anymore but everything else is?

### Cilium Version

v1.13.0 

### Kernel Version

v6.1.12

### Kubernetes Version

v1.26.1

### Relevant log output

```shell
level=info msg="Waiting until all Cilium CRDs are available" subsys=k8s
[...]

PR #28781, opened 2023-10-25
daemon: Simplify `cilium_host` IP restoration

### LINKED ISSUE #28743: Multi-pool reallocs router IP on cilium agent restart

gandro:
Before this PR, the `cilium_host` (aka router) IP restoration was done in two steps:

The first step collected previous IPs from the filesystem and Kubernetes and attempted to guess which one might still be valid in the new IPAM configuration. This validation was done using either the Pod CIDR, the VPC CIDR, or the native routing CIDR.

In a second step, the IP was then actually re-allocated via the IPAM subsystem. That second step could however still fail - just because an IP is part of the VPC CIDR or native routing CIDR does not mean it actually still assigned to the node.

This PR attempts to simplify the logic: Instead of trying to guess if the restore IP(s) could be allocated using IPAM, we just try to allocate them. If the allocation fails, we still have a fall back logic, and the old IP address(es) are still removed from the `cilium_host` interface. This way, the old CIDR check becomes obsolete, as we now use the IPAM subsystem as the source of truth if an IP can be restored or not.

This also fixes a bug where Multi-Pool IPAM always reallocated the router IP, because it did not implement the CIDR check in the first step.

**Review per commit.**

Fixes: #28743

### DIFF
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -1013,11 +1013,11 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 	// Fetch the router (`cilium_host`) IPs in case they were set a priori from
 	// the Kubernetes or CiliumNode resource in the K8s subsystem from call
 	// k8s.WaitForNodeInformation(). These will be used later after starting
-	// IPAM initialization to finish off the `cilium_host` IP restoration (part
-	// 2/2).
-	router4FromK8s, router6FromK8s := node.GetInternalIPv4Router(), node.GetIPv6Router()
+	// IPAM initialization to finish off the `cilium_host` IP restoration.
+	var restoredRouterIPs restoredIPs
+	restoredRouterIPs.IPv4FromK8s, restoredRouterIPs.IPv6FromK8s = node.GetInternalIPv4Router(), node.GetIPv6Router()
 	// Fetch the router IPs from the filesystem in case they were set a priori
-	router4FromFS, router6FromFS := node.ExtractCiliumHostIPFromFS()
+	restoredRouterIPs.IPv4FromFS, restoredRouterIPs.IPv6FromFS = node.ExtractCiliumHostIPFromFS()
 
 	// Configure IPAM without using the configuration yet.
 	d.configureIPAM()
@@ -1052,19 +1052,6 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 
 	// Start IPAM
 	d.startIPAM(params.Resources.LocalCiliumNode)
-	// After the IPAM is started, in particular IPAM modes (CRD, ENI, Alibaba)
-	// which use the VPC CIDR as the pod CIDR, we must attempt restoring the
-	// router IPs from the K8s resources if we weren't able to restore them
-	// from the fs. We must do this after IPAM because we must wait until the
-	// K8s resources have been synced. Part 2/2 of restoration.
-	var restoredRouterIPv4, restoredRouterIPv6 net.IP
-	if option.Config.EnableIPv4 {
-		restoredRouterIPv4 = d.restoreCiliumHostIPs(false, router4FromK8s, router4FromFS)
-	}
-	if option.Config.EnableIPv6 {
-		restoredRouterIPv6 = d.restoreCiliumHostIPs(true, router6FromK8s, router6FromFS)
-	}
-	removeOldCiliumHostIPs(ctx, restoredRouterIPv4, restoredRouterIPv6)
 
 	bootstrapStats.restore.Start()
 	// restore endpoints before any IPs are allocated to avoid eventual IP
@@ -1076,7 +1063,9 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 	}
 	bootstrapStats.restore.End(true)
 
-	if err := d.allocateIPs(); err != nil { // will log errors/fatal internally
+	// We must do this after IPAM because we must wait until the
+	// K8s resources have been synced.
+	if err := d.allocateIPs(ctx, restoredRouterIPs); err != nil { // will log errors/fatal internally
 		return nil, nil, err
 	}
 
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -296,7 +296,7 @@ func (d *Daemon) init() error {
 // that the most up-to-date information has been retrieved. At this point, the
 // daemon is aware of all the necessary information to restore the appropriate
 // IP.
-func (d *Daemon) restoreCiliumHostIPs(ipv6 bool, fromK8s net.IP) error {
+func (d *Daemon) restoreCiliumHostIPs(ipv6 bool, fromK8s net.IP) (restoredIP net.IP) {
 	var (
 		cidrs  []*cidr.CIDR
 		fromFS net.IP
@@ -326,8 +326,7 @@ func (d *Daemon) restoreCiliumHostIPs(ipv6 bool, fromK8s net.IP) error {
 		fromFS = node.GetInternalIPv4Router()
 	}
 
-	restoredIP := node.RestoreHostIPs(ipv6, fromK8s, fromFS, cidrs)
-	return removeOldRouterState(ipv6, restoredIP)
+	return node.RestoreHostIPs(ipv6, fromK8s, fromFS, cidrs)
 }
 
 // removeOldRouterState will try to ensure that the only IP assigned to the
@@ -1037,13 +1036,21 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 	// router IPs from the K8s resources if we weren't able to restore them
 	// from the fs. We must do this after IPAM because we must wait until the
 	// K8s resources have been synced. Part 2/2 of restoration.
+	var restoredRouterIPv4, restoredRouterIPv6 net.IP
+	if option.Config.EnableIPv4 {
+		restoredRouterIPv4 = d.restoreCiliumHostIPs(false, router4FromK8s)
+	}
+	if option.Config.EnableIPv6 {
+		restoredRouterIPv6 = d.restoreCiliumHostIPs(true, router6FromK8s)
+	}
+
 	gcHostIPsFn := func(ctx context.Context, retries int) (done bool, err error) {
 		var errs error
-		if option.Config.EnableIPv4 {
-			errs = errors.Join(errs, d.restoreCiliumHostIPs(false, router4FromK8s))
+		if restoredRouterIPv4 != nil {
+			errs = errors.Join(errs, removeOldRouterState(false, restoredRouterIPv4))
 		}
-		if option.Config.EnableIPv6 {
-			errs = errors.Join(errs, d.restoreCiliumHostIPs(true, router6FromK8s))
+		if restoredRouterIPv6 != nil {
+			errs = errors.Join(errs, removeOldRouterState(true, restoredRouterIPv6))
 		}
 		if resiliency.IsRetryable(errs) {
 			log.WithField(logfields.Attempt, retries).WithError(errs).Warnf("Failed to remove old router IPs from cilium_host.")
--- a/daemon/cmd/daemon.go
+++ b/daemon/cmd/daemon.go
@@ -373,6 +373,28 @@ func removeOldRouterState(ipv6 bool, restoredIP net.IP) error {
 	return err
 }
 
+// removeOldCiliumHostIPs calls removeOldRouterState() for both IPv4 and IPv6
+// in a retry loop.
+func removeOldCiliumHostIPs(ctx context.Context, restoredRouterIPv4, restoredRouterIPv6 net.IP) {
+	gcHostIPsFn := func(ctx context.Context, retries int) (done bool, err error) {
+		var errs error
+		if option.Config.EnableIPv4 {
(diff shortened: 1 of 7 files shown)


gandro:
Two open questions:

- This fixes a bug in multi-pool IPAM. Should we backport this to v1.14, or is it too risky? I can also workaround the bug in v1.14 by using a fake CIDR for multi-pool IPAM
- Does restoration from K8s even work in non-clusterpool mode? We fetch the [k8s `Node` object](https://github.com/cilium/cilium/blob/9638c9d41ab162b396b99ef0ede2089bd0021046/daemon/k8s/init.go#L55) (not `CiliumNode`) and it no longer seems to contain an annotation with the router IP? Might be related to https://github.com/cilium/cilium/issues/24420

gandro:
CI passed, but there is a typo in one of the commit messages. Repushing.

gandro:
GKE and external workload failures look legitimate. Will look into it.

pkg/node/address.go
```diff
@@ -695,10 +674,14 @@ func getCiliumHostIPsFromFile(nodeConfig string) (ipv4GW, ipv6Router net.IP) {
 	return ipv4GW, ipv6Router
 }
 
-// getCiliumHostIPs returns the Cilium IPv4 gateway and router IPv6 address from
+// ExtractCiliumHostIPfromFS returns the Cilium IPv4 gateway and router IPv6 address from
 // the node_config.h file if is present; or by deriving it from
 // defaults.HostDevice interface, on which only the IPv4 is possible to derive.
-func getCiliumHostIPs() (ipv4GW, ipv6Router net.IP) {
+func ExtractCiliumHostIPfromFS() (ipv4GW, ipv6Router net.IP) {
```
christarazi: Nit: `ExtractCiliumHostIPFromFS` (capital "From")

gandro:
> GKE and external workload failures look legitimate. Will look into it.

On closer look, the failures actually seem unrelated, and the more likely explanation is a minor GKE outage yesterday:
- [conformance-gke](https://github.com/cilium/cilium/actions/runs/6642236893) 3 of the 4 jobs timed out. This is similar to a run on [main](https://github.com/cilium/cilium/actions/runs/6642236893) from around the same timeframe. In particular, they all timed out during the `health` test due to it being unable to exec into the cilium pods. The only [job which didn't time out](https://github.com/cilium/cilium/actions/runs/6642236893/job/18046581416#step:22:1163) hit similar issues communicating with the kubeapiserver, i.e. we see multiple occurrences of communication failures between the CI test driver and the GKE apiserver.
- [ci-external-workloads](https://github.com/cilium/cilium/actions/runs/6642236822/job/18046551151) also seemed to have issues of the form `error getting cilium-agent logs for kube-system/cilium-6f7qt: pods "cilium-6f7qt" not found`. Notably, that workflow also runs on GKE, so it will suffer from GKE infrastructure outages too.

I've checked the sysdump of the failed jobs and did not see anything related to this PR. In fact, none of these GKE-based workflows did restart the Cilium pods before they failed, so no router restoration was performed. It should also be noted that the same pipelines were [previously green](https://github.com/cilium/cilium/pull/28781#issuecomment-1779427423) on a very similar version of the branch.

---

- [ci-upgrade-ipsec](https://github.com/cilium/cilium/actions/runs/6642237211) is a non-required workflow that is currently known to be flakey, but I still looked into it since it is one of the few workflows where we actually restart the Cilium pods. However, both of the failed jobs (1/3 was green) failed during *downgrade*, i.e. the failed pods were running Cilium 1.14.3 which does not have these changes. Since the code here does not change how we store the router IP on disk, it should not affect downgrade either. Indeed, we can see that both downgraded cilium instances were able to restore the router IP without issue.

---

Considering the above, I'm fairy confident the failures are unrelated. I'll restart CI.

gandro:
> /ci-ipsec-upgrade

The last IPsec failiure looks a bit suspicious, as it is running the code from `main`. In particular that it restored the old router IP, but then still finds old state in `cilium_host` to cleanup. I might have to a few more debug log statements to get to the cause of this.

daemon/cmd/daemon.go
```diff
@@ -1058,13 +1057,21 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
...
+	}
+	if option.Config.EnableIPv6 {
+		restoredRouterIPv6 = d.restoreCiliumHostIPs(true, router6FromK8s)
+	}
+
 	gcHostIPsFn := func(ctx context.Context, retries int) (done bool, err error) {
 		var errs error
-		if option.Config.EnableIPv4 {
-			errs = errors.Join(errs, d.restoreCiliumHostIPs(false, router4FromK8s))
+		if restoredRouterIPv4 != nil {
```
tommyp1ckles: In the previous case:

```go
	restoredIP := node.RestoreHostIPs(ipv6, fromK8s, fromFS, cidrs)
	return removeOldRouterState(ipv6, restoredIP)
```

I believe removeOldRouterState would always be attempted for the respective ip family, even if the restoredIP was nil.
Should this also always attempt remove old router state even if there is no restored IP, looking at the removeOldRouterState code that it was designed to handle this [case](https://github.com/cilium/cilium/blob/main/daemon/cmd/daemon.go#L375)

daemon/cmd/daemon.go
```diff
@@ -313,7 +313,7 @@ func createPrefixLengthCounter() *counter.PrefixLengthCounter {
 // then it attempts to clear all IPs from the interface.
 func removeOldRouterState(ipv6 bool, restoredIP net.IP) error {
 	l, err := netlink.LinkByName(defaults.HostDevice)
-	if errors.As(err, &netlink.LinkNotFoundError{}) && restoredIP == nil {
```
tommyp1ckles, line 338: Per my last comment, wont this only ever be called when restoredIP != nil?

daemon/cmd/daemon.go
gandro: The code you highlighted is from an intermediate commit - in the final version `removeOldRouterState` should be always called with the real (not the restored) router IP:
https://github.com/cilium/cilium/blob/26308e2550cc533fc14fb800047a3d47ee6f0945/daemon/cmd/ipam.go#L486-L487

The final version should (at least in theory) always clean up the interface, unless the IP family is disabled. That latter condition is equivalent to the old version of the code ([see here, where restoreCiliumHostIPs later calls into removeOldRouterState](https://github.com/cilium/cilium/blob/5ec730f5e22a7af45f16d7cc593e755ed8265dd2/daemon/cmd/daemon.go#L1042-L1047)). 

I can try to fix up the intermediate commit, but it shouldn't change the final outcome.

daemon/cmd/daemon.go
gandro: With this PR, yes. But this was not the case before this PR. Before this PR, it could happen that `restoredIP` did not match the pod CIDR or native routing CIDR and thus was set to `nil` and we removed all IPs from the `cilium_host` device.

In the new version, we call this _after_ we re-allocated the `restoredIP`, so we know it's valid and not nil. And we call unconditionally  both for restoration and the first time Cilium is started, so the `cilium_host` device might not exist yet.

gandro:
Rebased due to conflicts again. Waiting on reviews.

ci-ipsec-upgrade failures seem to be unrelated, I've investigated multiple sysdumps and did not see anything that points towards router IP restoration not working. The workflow is rather unstable on `main` too.

tommyp1ckles:
> Does restoration from K8s even work?

Probably something worth bringing up at the next community meeting 🤔

tommyp1ckles (approved):

tommyp1ckles:
Changes look good on my end, probably worth discussing the k8s restore question (maybe we can just officially get rid of this?).

gandro:
Many CI failures due to cloudflare outage. Rebasing and restarting.

daemon/cmd/ipam.go
```diff
@@ -192,38 +197,82 @@ func coalesceCIDRs(rCIDRs []string) (result []string) {
...
 		if err != nil {
-			log.Warn("Router IP could not be re-allocated. Need to re-allocate. This will cause brief network disruption")
-
-			// The restored router IP is not part of the allocation range.
-			// This indicates that the allocation range has changed.
-			deleteHostDevice()
+			log.WithError(err).Warnf(
+				"Unable to restore router IP %s from filesystem",
+				fromFS,
+			)
```
christarazi: Let's move this to a structured log for the IP address rather than a dynamically formatted string

daemon/cmd/ipam.go
```diff
@@ -192,38 +197,82 @@ func coalesceCIDRs(rCIDRs []string) (result []string) {
...
+	// If we were not able to restore the IP from the filesystem, try to use
+	// the IP from the Kubernetes resource.
+	if result == nil && fromK8s != nil {
+		result, err = alloc.AllocateIPWithoutSyncUpstream(fromK8s, "router", ipam.PoolDefault())
+		if err != nil {
+			log.WithError(err).Warnf(
+				"Unable to restore router IP %s from kubernetes",
+				fromK8s,
+			)
+			result = nil
```
christarazi: Same here re: structured logging

daemon/cmd/daemon.go
```diff
@@ -291,47 +290,12 @@ func (d *Daemon) init() error {
...
-
 // removeOldRouterState will try to ensure that the only IP assigned to the
 // `cilium_host` interface is the given restored IP. If the given IP is nil,
 // then it attempts to clear all IPs from the interface.
 func removeOldRouterState(ipv6 bool, restoredIP net.IP) error {
 	l, err := netlink.LinkByName(defaults.HostDevice)
-	if errors.As(err, &netlink.LinkNotFoundError{}) && restoredIP == nil {
+	if errors.As(err, &netlink.LinkNotFoundError{}) {
 		// There's no old state remove as the host device doesn't exist and
 		// there's no restored IP anyway.
```
christarazi: The comment needs to be updated as well.

christarazi (approved):
LGTM, thanks for this, much better looking code now!

A couple of nits below. 

> Does restoration from K8s even work? We fetch the [k8s Node object](https://github.com/cilium/cilium/blob/9638c9d41ab162b396b99ef0ede2089bd0021046/daemon/k8s/init.go#L55) (not CiliumNode) and it no longer seems to contain an annotation with the router IP? And even with clusterpool, we call UpdateCiliumNodeResource before WaitForNodeInformation, which [simply clears out the old router IP](https://github.com/cilium/cilium/blob/e30e46a5b43945900cace8e7a73fd571d0365db0/pkg/nodediscovery/nodediscovery.go#L438) (and [ignores](https://github.com/cilium/cilium/blob/e30e46a5b43945900cace8e7a73fd571d0365db0/pkg/nodediscovery/nodediscovery.go#L490) the router IP from the Node resource if it was present, but it anyway doesn't seem to be). At least with Kind, I have not been able to get it to work both before and after this PR

I'm _fairly_ certain that this was not the case when the K8s restoration code was added. This restoration logic relies on the annotation to be set. So if somewhere else in Cilium stopped setting the annotation, then that is what needs to be investigated. If we no longer set the annotation at all, then we should probably remove the K8s restoration code because we then have no mechanism to restore state from K8s.

daemon/cmd/daemon.go
gandro: Good catch!

gandro:
Thanks for the reviews! I believe I've address the feedback!

> I'm fairly certain that this was not the case when the K8s restoration code was added. This restoration logic relies on the annotation to be set. So if somewhere else in Cilium stopped setting the annotation, then that is what needs to be investigated. If we no longer set the annotation at all, then we should probably remove the K8s restoration code because we then have no mechanism to restore state from K8s.

I'll create an issue. I think the annotation is indeed no longer set for at least one release, and this has been reported before, though in that case for the pod CIDR: https://github.com/cilium/cilium/issues/24420

IMO, we should attempt to restore the router IP from the local CiliumNode object, rather than adding annotations to the K8s Node object. But that's a separate discussion.

PR #28783, opened 2023-10-25
pkg/proxy/logger: switch to netip.Addr

tklauser:
Modernize the package by switching to the `netip.Addr` type provided [since Go 1.18](https://tip.golang.org/doc/go1.18#netip).

Towards #24246

### DIFF
--- a/pkg/proxy/logger/logger.go
+++ b/pkg/proxy/logger/logger.go
@@ -4,8 +4,7 @@
 package logger
 
 import (
-	"net"
-	"strconv"
+	"net/netip"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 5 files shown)


pkg/proxy/logger/logger.go
```diff
@@ -136,8 +137,11 @@ func (logTags) Addressing(i AddressingInfo) LogTag {
 	return func(lr *LogRecord) {
 		ipstr, port, err := net.SplitHostPort(i.SrcIPPort)
```
danehans: [MustParseAddrPort()](https://pkg.go.dev/net/netip#MustParseAddrPort) can be used instead of SplitHostPort() and drop the net pkg.

danehans (changes requested):

pkg/proxy/logger/logger.go
tklauser: Good call, thanks. I'd switch to [`ParseAddrPort`](https://pkg.go.dev/net/netip#ParseAddrPort) instead though. We want to avoid `panic`ing in the unlikely case of a malformed `i.SrcIPPort`.

danehans (approved):

tklauser:
@jrajahalme ping for review

jrajahalme (approved):
Thanks!

PR #28785, opened 2023-10-25
operator: introduce gateway api cell

mhofstetter:
This commit moves the Gateway API controller registration from the legacyOnLeader cell to its own cell.

In addition, the config properties `enable-gateway-api-secrets-sync` & `gateway-api-secrets-namespace` are moved from the global operator config to a dedicate gatewayApiConfig struct.

### DIFF
--- a/operator/pkg/gateway-api/cell.go
+++ b/operator/pkg/gateway-api/cell.go
@@ -1,0 +1,66 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package gateway_api
+
+import (
+	"fmt"
+
+	"github.com/bombsimon/logrusr/v4"
+	"github.com/spf13/pflag"
+	ctrlRuntime "sigs.k8s.io/controller-runtime"
+
+	operatorOption "github.com/cilium/cilium/operator/option"
+	"github.com/cilium/cilium/pkg/hive"
+	"github.com/cilium/cilium/pkg/hive/cell"
+	"github.com/cilium/cilium/pkg/logging"
+)
+
+// Cell manages the Gateway API related controllers.
+var Cell = cell.Module(
+	"gateway-api",
+	"Manages the Gateway API controllers",
+
+	cell.Config(gatewayApiConfig{
+		EnableGatewayAPISecretsSync: true,
+		GatewayAPISecretsNamespace:  "cilium-secrets",
+	}),
+	cell.Invoke(registerController),
+)
+
+type gatewayApiConfig struct {
+	EnableGatewayAPISecretsSync bool
+	GatewayAPISecretsNamespace  string
+}
+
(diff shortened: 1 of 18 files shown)


operator/pkg/gateway-api/cell.go
```diff
@@ -0,0 +1,64 @@
...
+
+func registerController(lc hive.Lifecycle, config gatewayApiConfig) error {
+	if operatorOption.Config.EnableGatewayAPI {
+		// Setting global logger for controller-runtime
+		ctrlRuntime.SetLogger(logrusr.New(logging.DefaultLogger, logrusr.WithName("controller-runtime")))
+
+		gatewayController, err := NewController(
+			config.EnableGatewayAPISecretsSync,
+			config.GatewayAPISecretsNamespace,
+			operatorOption.Config.ProxyIdleTimeoutSeconds,
```
sayboras: this global flag (as well as operatorOption.Config.EnableGatewayAPI) will be loaded before the cell I assume ?

sayboras (approved):
LGTM :heavy_check_mark:

operator/pkg/gateway-api/cell.go
mhofstetter: yes these flags are loaded by then.

operator/pkg/gateway-api/cell.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create gateway controller: %w", err)
+	}
+
+	lc.Append(hive.Hook{
+		OnStart: func(_ hive.HookContext) error {
+			go gatewayController.Run()
+			return nil
+		},
+	})
```
pippolo84, line 63: One possible drawback of using a naked goroutine is to have possible false positives when using this cell in tests with goleak. Alternatively, we can rely on a workerpool with a single worker (to be started and stopped in `OnStart` and `OnStop`, respectively) or even better on a `OneShot` Job registered in a group, that also gives the ability to report its status to the health reporter.
But I think it is better to leave this for a follow up PR (we can also do the same for the Ingress cell).

pippolo84 (approved):
Great, thanks!
I've left a suggestion for a follow up PR inline.

operator/pkg/gateway-api/cell.go
mhofstetter: yep, the intention of this PR was more to lift & shift it into its own cell. but agree that there's room for improvement. might come up with a follow up. thanks for the pointer.

mhofstetter:
removed review request from @meyskens as @sayboras already covered `cilium/sig-servicemesh`

mhofstetter:
rebased to main as gateway api tests were failing -> https://github.com/cilium/cilium/pull/28654 added an additional gRPC test (file was missing)

PR #28788, opened 2023-10-25
Fix CIDR labels computation

pippolo84:
The previous version of the GetCIDRLabels implementation was computing the set of labels for a CIDR incorrectly in case of a cache hit: https://github.com/cilium/cilium/pull/28465#issuecomment-1777517198

Also, netip.PrefixFrom(...) does not mask the internally stored address, thus lowering the cache hit ratio even if two different CIDRs, used as keys in the LRU cache, are equal in terms of "masked address". (e.g: "1.1.1.1/16" and "1.1.0.0/16").
So, netip.Addr.Prefix(...) is used instead.

After the fix, the performance are roughly equal (but with an increased chance of having a cache hit). To keep the maximum heap usage reasonably low, the cache size is now halved.

Also, the tests are refactored to increase the coverage (checking the labels computation both in case of cache miss and in case of a cache hit).

Finally, the only remaining test using the deprecated checker package is migrated to the testing package from the standard Go library.

Read each commit message for further details and performance comparison.

cc @odinuge @sjdot

### DIFF
--- a/pkg/labels/cidr.go
+++ b/pkg/labels/cidr.go
@@ -133,7 +133,7 @@ var (
 	mu lock.Mutex
 )
 
-const cidrLabelsCacheMaxSize = 16384
+const cidrLabelsCacheMaxSize = 8192
 
 func addWorldLabel(addr netip.Addr, lbls Labels) {
 	switch {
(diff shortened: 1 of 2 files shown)


joamaki (approved):

odinuge:
Thanks @pippolo84! This passes my little simple test [case](https://github.com/cilium/cilium/pull/28465#issuecomment-1777517198)! 🚀

nathanjsweet (approved):

PR #28790, opened 2023-10-25
bugtool: Collect XFRM error counters twice

### LINKED ISSUE #16538: Collect `/proc/net/xfrm_stat` twice in bugtool
The `/proc/net/xfrm_stat` file is useful for IPsec incidents to understand if errors are happening in the node. The absolute values don't tell us much however and we often need to ask users for a second bugtool report, to know if the counters are increasing.

Instead, we should probably collect this file twice, with a short interval in between the two collections.

pchaigno:
This pull request changes the bugtool report to collect the XFRM error counters (i.e., `/proc/net/xfrm_stat`) twice instead of only once. We will collect at the beginning and end of the bugtool collection. In that way, there will be around 5-6 seconds between the two collections and we may see if any counter is currently increasing.

    $ diff cilium-bugtool-cilium-7d54p-20231025-115151/cmd/cat*--proc-net-xfrm_stat.md
    5c5
    < XfrmInStateProtoError   	4
    ---
    > XfrmInStateProtoError   	6

In this example, we can easily see that the `XfrmInStateProtoError` is increasing. That suggests a key rotation issue is currently ongoing (cf. IPsec troubleshooting docs).

I tried other approaches to collect over a longer timespan. That may allow us to see slower increases. They all end up being more complex or messier (we'd need to collect at beginning and end of the sysdump instead). In the end, considering this is already a fallback plan for when customers don't collect Prometheus metrics, I think the current, simpler approach is good enough.

Fixes: https://github.com/cilium/cilium/issues/16538.

### DIFF
--- a/bugtool/cmd/configuration.go
+++ b/bugtool/cmd/configuration.go
@@ -81,6 +81,10 @@ func defaultCommands(confDir string, cmdDir string, k8sPods []string) []string {
 	var commands []string
 	// Not expecting all of the commands to be available
 	commands = []string{
+		// We want to collect this twice: at the very beginning and at the
+		// very end of the bugtool collection, to see if the counters are
+		// increasing.
+		"cat /proc/net/xfrm_stat",
 		// Host and misc
 		"ps auxfw",
 		"hostname",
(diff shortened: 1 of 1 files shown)


bugtool/cmd/configuration.go
```diff
@@ -214,6 +218,11 @@ func defaultCommands(confDir string, cmdDir string, k8sPods []string) []string {
 		commands = append(commands, tcCommands...)
 	}
 
+	// We want to collect this twice: at the very beginning and at the
+	// very end of the bugtool collection, to see if the counters are
+	// increasing.
+	commands = append(commands, "cat -u /proc/net/xfrm_stat")
```
tklauser, line 228: Out of curiosity, why are we using `cat -u` here and not in the other commands?

tklauser (approved):
LGTM, one question inline to help my own understanding.

bugtool/cmd/configuration.go
pchaigno: Ouch, that's a good question which I should be documenting in the comment :disappointed: 

The output from all these commands is written into bugtool files whose name is generated. More specifically, the filenames are generated from the commands themselves. So for example this command's output will be stored in `cat--u-proc-net-xfrm_stat.md` IIRC. Hence, we need some difference in the command to end up with two difference filenames. I picked `-u` because that has no effect.

tklauser: Thanks a lot for elaborating and documenting it in a comment!

PR #28791, opened 2023-10-25
makefile: fix 'make kind' for mac

f1ko:
'sed -z' is only available for GNU sed and not the pre-installed version on MacOS. Changing to 'awk' as it accomplishes the same desired output and the specified command works on all popular awk distributions.


tklauser (approved):

tklauser:
@f1ko FYI, I've rebased this PR against latest `main` to fix the Gateway API workflow failure.

PR #28792, opened 2023-10-25
daemon: Remove auto-generated allocation range

gandro:
Before this commit, Cilium used to always auto-generate an endpoint allocation CIDR based on the local node IP. This CIDR however was never actually used for endpoint IP allocation. Instead, the subnet endpoint IPs are allocated from depends on the used IPAM mode:

- kubernetes: The allocation CIDR is the podCIDR of the K8s `Node` resource
- clusterpool: The allocation CIDR is the podCIDR of the `CiliumNode` resource
- ENI/Azure/AlibabaCloud: There is no CIDR from which local IPs are allocated. Instead endpoint IPs are allocated from one or more shared subnets.
- MultiPool: IPs are based from different CIDRs based on the workload.
- Delegated: Cilium-Agent does not allocate any IPs. The router IP is configured externally.

Even with the auto-generated CIDR removed, users can still provide their own allocation CIDR using the `ipv4/6-range` option.


joamaki (approved):
Yay thanks!

PR #28794, opened 2023-10-25
operator: introduce Ingress cell

mhofstetter:
This commit moves the Ingress controller registration from the legacyOnLeader cell into its own cell.

In addition, the corresponding config properties have been moved into its own config struct `ingressConfig` where possible.

### DIFF
--- a/operator/pkg/ingress/cell.go
+++ b/operator/pkg/ingress/cell.go
@@ -1,0 +1,101 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package ingress
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spf13/pflag"
+
+	operatorK8s "github.com/cilium/cilium/operator/k8s"
+	operatorOption "github.com/cilium/cilium/operator/option"
+	"github.com/cilium/cilium/pkg/hive"
+	"github.com/cilium/cilium/pkg/hive/cell"
+	k8sClient "github.com/cilium/cilium/pkg/k8s/client"
+)
+
+// Cell manages the Kubernetes Ingress related controllers.
+var Cell = cell.Module(
+	"ingress",
+	"Manages the Kubernetes Ingress controllers",
+
+	cell.Config(ingressConfig{
+		EnableIngressController:     false,
+		EnforceIngressHTTPS:         true,
+		EnableIngressProxyProtocol:  false,
+		EnableIngressSecretsSync:    true,
+		IngressSecretsNamespace:     "cilium-secrets",
+		IngressLBAnnotationPrefixes: []string{"service.beta.kubernetes.io", "service.kubernetes.io", "cloud.google.com"},
+		IngressSharedLBServiceName:  "cilium-ingress",
+		IngressDefaultLBMode:        "dedicated",
+	}),
+	cell.Invoke(registerController),
+)
+
+type ingressConfig struct {
+	EnableIngressController       bool
+	EnforceIngressHTTPS           bool
+	EnableIngressProxyProtocol    bool
+	EnableIngressSecretsSync      bool
+	IngressSecretsNamespace       string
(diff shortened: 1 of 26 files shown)


mhofstetter:
Converting to draft: looks like some stuff during controller creation already depends on the CRDs being present. But they are added during the lifecycle hooks.

joamaki:
> Converting to draft: Smoke tests are failing. looks like some stuff (informers are already started) during controller creation already depends on the CRDs being present. But they are added during the lifecycle hooks that are executed later.

Can we make the dependency on the CRD creation explicit? E.g. there's the `RegisterCRDsCell`, but it's currently just registering a start hook and not providing anything we could depend on. It could perhaps work better if it would have e.g. `CRDRegistration` object and instead of group value you could do `CRDRegistration.Register(some-crd-metadata)`? This way `CRDRegistration.Start` would always be called before anything that called `Register`. cc @pippolo84

mhofstetter:
> > Converting to draft: Smoke tests are failing. looks like some stuff (informers are already started) during controller creation already depends on the CRDs being present. But they are added during the lifecycle hooks that are executed later.
> 
> Can we make the dependency on the CRD creation explicit? E.g. there's the `RegisterCRDsCell`, but it's currently just registering a start hook and not providing anything we could depend on. It could perhaps work better if it would have e.g. `CRDRegistration` object and instead of group value you could do `CRDRegistration.Register(some-crd-metadata)` from an Invoke? This way `CRDRegistration.Start` would always be called before anything that called `Register`. cc @pippolo84

@joamaki thanks for the input. agree that this would be way better than the implicit dependency due to the order of registration in the root cell. If it's OK i'd like to handle this in a separate PR after the initial PRs for Ingress (this PR) and Gateway API (https://github.com/cilium/cilium/pull/28785) are merged.

mhofstetter:
> Converting to draft: Smoke tests are failing. looks like some stuff (informers are already started) during controller creation already depends on the CRDs being present. But they are added during the lifecycle hooks that are executed later.

Moved the start of the informers from `ingress.NewController` (called during cell initialization) into `ingressController.Run` (called by cell onCreate lifecycle hook)

operator/pkg/ingress/cell.go
```diff
@@ -0,0 +1,99 @@
...
+	flags.String("ingress-secrets-namespace", r.IngressSecretsNamespace, "Namespace having tls secrets used by Ingress and CEC.")
+	flags.StringSlice("ingress-lb-annotation-prefixes", r.IngressLBAnnotationPrefixes, "Annotations which are needed to propagate from Ingress to the Load Balancer.")
+	flags.String("ingress-shared-lb-service-name", r.IngressSharedLBServiceName, "Name of shared LB service name for Ingress.")
+	flags.String("ingress-default-lb-mode", r.IngressDefaultLBMode, "Default loadbalancer mode for Ingress. Applicable values: dedicated, shared")
+	flags.String("ingress-default-secret-namespace", r.IngressDefaultSecretNamespace, "Default secret namespace for Ingress.")
+	flags.String("ingress-default-secret-name", r.IngressDefaultSecretName, "Default secret name for Ingress.")
+}
+
+func registerController(lc hive.Lifecycle, clientset k8sClient.Clientset, resources operatorK8s.Resources, config ingressConfig) error {
+	if config.EnableIngressController {
```
pippolo84: Nit: maybe just a matter of personal style, but I think early return can improve readability here

```suggestion
	if !config.EnableIngressController {
		return nil
	}
	...
```

pippolo84 (approved):
LGTM, just a non-blocking nit left inline :100:

operator/pkg/ingress/cell.go
mhofstetter: agree -> changed

mhofstetter:

meyskens (approved):
LGTM! Very happy with the cell improvements

PR #28800, opened 2023-10-25
Update cilium status command module health output

derailed:

      GitHub issue.

To date, we render modules health information as tabular data. As modules health instrumentation grows this rendering makes it hard for users to grok the output. This PR proposes changing the rendering of modular health using a tree like structure to give users a better line of sight to gain an understanding of the agent modules health state.

Fixes: #issue-number

```release-note
Changed cilium status CLI output to render the modules health section as a tree structure vs tabular data.
```

### DIFF
--- a/pkg/health/client/modules.go
+++ b/pkg/health/client/modules.go
@@ -7,9 +7,19 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"strings"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/duration"
 
 	"github.com/cilium/cilium/api/v1/client/daemon"
 	"github.com/cilium/cilium/pkg/hive/cell"
+)
+
+const (
+	noPod    = "(/)"
+	rootNode = "agent"
+	noErr    = "<nil>"
 )
 
 // ModulesHealth represent hive modules health API.
@@ -32,17 +42,16 @@ func GetAndFormatModulesHealth(w io.Writer, clt ModulesHealth, verbose bool) {
 		return
 	}
 	if verbose {
-		fmt.Fprintln(w)
+		r := newRoot(rootNode)
 		for _, m := range resp.Payload.Modules {
-			n := &cell.StatusNode{}
-			if err := json.Unmarshal([]byte(m.Message), n); err != nil {
-				panic(err)
-			}
 			if m.Level == string(cell.StatusUnknown) {
 				continue
 			}
-			fmt.Fprintf(w, "%s", n.StringIndent(2))
+			if err := buildTree(r, m.Message); err != nil {
+				fmt.Fprintf(w, "Modules Health rendering failed: %s\n", err)
+			}
 		}
+		fmt.Fprintln(w, "\n"+r.String())
 		return
 	}
 	tally := make(map[cell.Level]int, 4)
@@ -59,4 +68,52 @@ func GetAndFormatModulesHealth(w io.Writer, clt ModulesHealth, verbose bool) {
 		cell.StatusUnknown,
 		tally[cell.StatusUnknown],
 	)
+}
+
+func buildTree(n *node, raw string) error {
+	var sn cell.StatusNode
+	if err := json.Unmarshal([]byte(raw), &sn); err != nil {
+		return err
+	}
+	build(n, &sn)
+	return nil
+}
+
+func ensurePath(n *node, pp []string) *node {
+	current := n
+	for _, p := range pp {
+		if v := current.find(p); v != nil {
+			current = v
+			continue
+		}
+		current = current.addBranch(strings.Replace(p, noPod, "", 1))
+	}
+
+	return current
+}
+
+func build(n *node, sn *cell.StatusNode) {
+	meta := fmt.Sprintf("[%s] %s", strings.ToUpper(string(sn.LastLevel)), sn.Message)
+	if sn.Error != "" {
+		meta += " -- " + sn.Error
+	}
+	meta += fmt.Sprintf(" (%s, x%d)", ToAgeHuman(sn.UpdateTimestamp), sn.Count)
+	pp := strings.Split(sn.Name, ".")
+	current := ensurePath(n, pp)
+	if len(sn.SubStatuses) == 0 {
+		current.meta = meta
+		return
+	}
+	for _, s := range sn.SubStatuses {
+		build(current, s)
+	}
+}
+
+// ToAgeHuman converts time to duration.
+func ToAgeHuman(t time.Time) string {
+	if t.IsZero() {
+		return "n/a"
+	}
+
+	return duration.HumanDuration(time.Since(t))
 }
--- a/pkg/health/client/modules_test.go
+++ b/pkg/health/client/modules_test.go
@@ -5,6 +5,7 @@ package client_test
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"testing"
@@ -34,9 +35,15 @@ func TestGetAndFormatModulesHealth(t *testing.T) {
 		},
 		"happy-verbose": {
 			h: newTestMHappy(),
-			e: "Modules Health:\n" +
-				"  m1 OK status nominal 2s ago (x0)\n" +
-				"  m2 Degraded doh 20s ago (x0)",
+			e: `Modules Health:
+agent
+├── m1                                                      [OK] status nominal (2s, x0)
+└── a
+    └── b
+        └── c
+            ├── fred                                        [OK] yo (20s, x1)
+            │   └── blee                                    [OK] doh (20s, x1)
+            └── dork                                        [DEGRADED] bozo -- BOOM! (20s, x1)`,
 			v: true,
 		},
 	}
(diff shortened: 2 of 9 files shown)


pkg/health/client/modules.go
```diff
@@ -6,12 +6,20 @@ package client
 import (
 	"fmt"
 	"io"
+	"regexp"
 	"strings"
 
+	"github.com/opentracing/opentracing-go/log"
```
tklauser: Any reason to use the opentracing log package for logging instead of `github.com/sirupsen/logrus` like most other cilium packages? AFAICT it's only used for the log line in `buildTree` below.

derailed: Good catch! Thank you Tobias.

ti-mo:
@derailed Could you update the release note in the PR description?

pkg/health/client/modules.go
```diff
@@ -56,3 +64,49 @@ func GetAndFormatModulesHealth(w io.Writer, clt ModulesHealth, verbose bool) {
...
+		if len(ff) != 7 {
+			continue
+		}
+
+		meta := fmt.Sprintf("[%s] %s", strings.ToUpper(ff[2]), ff[5])
+		if ff[6] != noErr {
+			meta += fmt.Sprintf(" -- %s", ff[6])
+		}
+		meta += fmt.Sprintf(" (%s, x%s)", ff[3], ff[4])
+		addNode(t, mid+"."+ff[1], meta)
```
pippolo84: The column indexes feel a bit like magic numbers, what about adding a comment with an example of the expected format and maybe some constants in place of the numbers to improve readability?

pkg/health/client/modules_test.go
pippolo84, line 1: It seems that `buildTree` is not covered by the current test, might it be worth adding another one?

pkg/health/client/tree.go
```diff
@@ -0,0 +1,219 @@
...
+	meta interface{}
+	tree interface {
+		addNode(v val) tree
+		addNodeWithMeta(v val, m meta) tree
+		addBranch(v val) tree
+		addBranchWithMeta(v val, meta meta) tree
+		find(val val) tree
+		String() string
+		asBytes() []byte
+	}
```
pippolo84: I don't see much value in defining these interfaces.
This tree implementation seems something specialized for the health reporter, so I'd say we can just use `string` for `val` and `meta` and simply remove the `tree` interface. This should also simplify the code a bit (e.g.: no need to use `reflect.DeepEqual` in `find` and no need for the unused `addNode`).
WDYT?

pkg/hive/cell/structured.go
```diff
@@ -415,30 +415,39 @@ func (s *StatusNode) allOk() bool {
...
 		return ""
 	}
-	buf := bytes.NewBuffer(nil)
-	w := tabwriter.NewWriter(buf, 0, 0, 1, ' ', 0)
-	s.writeTo(w, ident)
-	w.Flush()
-	return buf.String()
+	w := bytes.NewBuffer(nil)
+	s.writeTo(w, s.Name)
+	return w.String()
```
pippolo84: Nit: `strings.Builder` should be slightly more efficient and it satisfies `io.Writer` too:

```suggestion
	var sb strings.Builder
	s.writeTo(&sb, s.Name)
	return sb.String()
```

pippolo84 (commented):
Great work! I'd left some questions inline.

pkg/health/client/tree.go
derailed: Good point. Thank you Fabio! Was thinking this could be useful somewhere else down the line but who knows?? I'll dumb it down.

pkg/health/client/modules.go
derailed: Thinking this will be short lived. Will update in the meantime...

pkg/health/client/tree.go
```diff
@@ -0,0 +1,207 @@
...
+	leafMaxWidth            = 40
+	link         decoration = "│"
+	mid          decoration = "├──"
+	end          decoration = "└──"
+)
+
+type (
+	decoration string
+
+	// val  interface{}
```
tommyp1ckles: Remove comments code

pkg/health/client/modules.go
```diff
@@ -56,3 +64,55 @@ func GetAndFormatModulesHealth(w io.Writer, clt ModulesHealth, verbose bool) {
...
+		}
+		if parent != nil {
+			parent = parent.addBranch(name)
+		} else {
+			parent = p.addBranch(name)
+		}
+	}
+}
+
+func buildTree(t *node, mid, s string) {
```
tommyp1ckles: FYI: With [#28773](https://github.com/cilium/cilium/pull/28773) I'm changing the way the module health report is sent - instead of sending a formatted string a serialized version of the tree will be sent.  I think this is the way we want to do this going forward instead of trying to parse around raw formatted text output.

I'm a bit concerned about this approach since it seems brittle (i.e. it depends on an implicit formatting of text generated by cell), so changes there will need to reflected by changes here trying to parse out the text.

Alongside the changes in https://github.com/cilium/cilium/pull/28773, could we put all this logic in one place?

The CLI text rendering code being in pkg/hive/cell is admittedly a bit awkward - perhaps we could do something like:

```go
type node struct {
    cell.StatusNode
    ...
}

func buildTree(t *node) { // note: module-id etc are now embedded in *node struct.  

}
```
So we can reuse some of that code and have all the code dealing rendering text live in the CLI code rather than in the hive/cell internals.

Wdyt?

derailed: @tommyp1ckles Thanks! 

I was trying to operate with what was there since that pr was not on the radar as of yet. 
I'll take a peek... 

I am actually not sure we want to send out a json version of the tree??
End of day, this data will likely be managed in statedb and queried from there. 
Hence my feel is we might be best served to keep sending out an array of module health reports.

pkg/health/client/modules.go
```diff
@@ -60,3 +68,52 @@ func GetAndFormatModulesHealth(w io.Writer, clt ModulesHealth, verbose bool) {
...
+	if err := json.Unmarshal([]byte(raw), &sn); err != nil {
+		return err
+	}
+	build(n, &sn)
+	return nil
+}
+
+func ensurePath(n *node, pp []string) *node {
+	current := n
+	for i := 0; i < len(pp); i++ {
```
tommyp1ckles: nit: could be `i, _ := range pp { ... }`

pkg/health/client/modules.go
```diff
@@ -60,3 +68,52 @@ func GetAndFormatModulesHealth(w io.Writer, clt ModulesHealth, verbose bool) {
...
+		if v := current.find(pp[i]); v != nil {
+			current = v
+			continue
+		}
+		name := strings.Replace(pp[i], "(/)", "", 1)
```
tommyp1ckles: there already appears to be a constant for this (`noPod`)

pkg/health/client/tree.go
```diff
@@ -0,0 +1,193 @@
...
+
+func (n *node) asBytes() []byte {
+	buf := new(bytes.Buffer)
+	level := 0
+	var levelsEnded []int
+	if n.parent == nil {
+		if n.meta != "" {
+			buf.WriteString(fmt.Sprintf("%v %v", n.val, n.meta))
+		} else {
+			buf.WriteString(fmt.Sprintf("%v", n.val))
```
tommyp1ckles: nit: could be cleaner with `fmt.Fprintf`

pkg/health/client/modules.go
```diff
@@ -32,17 +42,16 @@ func GetAndFormatModulesHealth(w io.Writer, clt ModulesHealth, verbose bool) {
...
-			n := &cell.StatusNode{}
-			if err := json.Unmarshal([]byte(m.Message), n); err != nil {
-				panic(err)
-			}
 			if m.Level == string(cell.StatusUnknown) {
 				continue
 			}
-			fmt.Fprintf(w, "%s", n.StringIndent(2))
+			if err := buildTree(r, m.Message); err != nil {
+				fmt.Fprintf(w, "Modules Health rendering failed: %s\n", err)
```
tommyp1ckles, line 51: Should we do `os.Exit(1)` here?

pkg/health/client/modules_test.go
```diff
@@ -83,13 +81,21 @@ func (m *testMHappy) GetHealth(params *daemon.GetHealthParams, opts ...daemon.Cl
 					LastUpdated: "2s",
 				},
 				{
-					ModuleID:    "m2",
-					Level:       string(cell.StatusDegraded),
-					Message:     fmt.Sprintf("{\"name\": \"m2\", \"level\":\"Degraded\",\"message\":\"doh\", \"timestamp\": %q}", t2),
+					ModuleID: "a.b.c",
+					Level:    string(cell.StatusDegraded),
+					Message: fmt.Sprintf(`
+					{"level": "OK", "name": "a.b.c", "count": 1, "sub_statuses": [
```
tommyp1ckles: Could we just use `*UpdateStatus` and then marshal to make it more clear where this json comes from for someone maintaining this?

pkg/health/client/modules.go
```diff
@@ -60,3 +69,52 @@ func GetAndFormatModulesHealth(w io.Writer, clt ModulesHealth, verbose bool) {
...
+			continue
+		}
+		name := strings.Replace(pp[i], "(/)", "", 1)
+		current = current.addBranch(name)
+	}
+
+	return current
+}
+
+func build(n *node, sn *cell.StatusNode) {
```
tommyp1ckles, line 95: With this, I think the only use of `cell.(*UpdateNode).String()` family of functions is for a debug log in hive/health.go.

We should probably just get rid of that code to clean things up (doesn't have to be part of this PR - but lets make it a point to circle back and clean that up).

tommyp1ckles (commented):
Output looks great! - thanks working around the last set of changes to the reporter.

Just a few comments

pkg/health/client/modules.go
derailed: We may not be end of line here? imho this is ok.

pkg/health/client/modules_test.go
derailed: Good point!

pkg/health/client/modules.go
derailed: Right I was wondering about that too. Don't think this is within the scope of this PR tho
Additionally tho the output has been enhance our messaging is lame. ie OK OK or worse no messages!. This need TLC and I think this should be our next priority!

pkg/health/client/modules.go
tommyp1ckles: Makes sense, this failing would certainly be unexpected but we can try to keep plowing forward.

pkg/health/client/modules.go
tommyp1ckles: Yeah right now the modular health updates are making debug logs way too noisy.  With that one we should just update the subreporter node that changed instead of formatting out the whole tree.

tommyp1ckles (approved):
lgmt

pkg/health/client/modules_test.go
```diff
@@ -34,9 +35,7 @@ func TestGetAndFormatModulesHealth(t *testing.T) {
 		},
 		"happy-verbose": {
 			h: newTestMHappy(),
-			e: "Modules Health:\n" +
-				"  m1 OK status nominal 2s ago (x0)\n" +
-				"  m2 Degraded doh 20s ago (x0)",
+			e: "Modules Health:\nagent\n├── m1                                                      [OK] status nominal (2s, x0)\n└── a\n    └── b\n        └── c\n            ├── fred                                        [OK] yo (20s, x1)\n            │   └── blee                                    [OK] doh (20s, x1)\n            └── dork                                        [DEGRADED] bozo -- BOOM! (20s, x1)",
```
pippolo84: Nit: using backticks here can help improve the readability of the expected output:

```go
			e: `Modules Health:
agent
├── m1                                                      [OK] status nominal (2s, x0)
└── a
    └── b
        └── c
            ├── fred                                        [OK] yo (20s, x1)
            │   └── blee                                    [OK] doh (20s, x1)
            └── dork                                        [DEGRADED] bozo -- BOOM! (20s, x1)`,
```

pkg/health/client/tree.go
```diff
@@ -0,0 +1,192 @@
...
+		if l == level {
+			return true
+		}
+	}
+
+	return false
+}
+
+func dumpVal(level int, node *node) string {
+	lines := strings.Split(fmt.Sprintf("%v", node.val), "\n")
```
pippolo84: Now that `val` is a string this can be simplified:

```go
lines := strings.Split(node.val, "\n")
```

There are also some other places where `%s` can be used instead of `%v`

pippolo84 (approved):
LGTM, thanks! :rocket: 
A couple of non-blocking nits left inline.

pkg/health/client/tree.go
derailed: Great catch! Tx Fabio I'll update.

jrajahalme (approved):

tklauser (approved):

PR #28806, opened 2023-10-26
Add lbipam support for shared ips

### LINKED ISSUE #21776: lb-ipam: IP sharing between services
PR #21764 introduces the initial version of the LB-IPAM feature. This version contains the basics but is by no means feature complete.
This issue tracks the proposal to add a feature similar[ MetalLB's IP sharing feature](https://metallb.universe.tf/usage/#ip-address-sharing).

MetalLB allows users to add the `metallb.universe.tf/allow-shared-ip` annotation to services. MetalLB will attempt to assign the same IP to all services with the same "sharing key" which is the value of the annotation. MetalLB doesn't guarantee this and will only assign the same IP if a few conditions are met:
* They both have the same sharing key.
* They request the use of different ports (e.g. tcp/80 for one and tcp/443 for the other).
  *  As long as #9207 is unresolved we should not allow the merging of services with overlapping ports regardless of protocol
* They both use the Cluster external traffic policy, or they both point to the exact same set of pods (i.e. the pod selectors are identical).

The stated use case is as follows:
> There are two main reasons to colocate services in this fashion: to work around a Kubernetes limitation, and to work with limited IP addresses.

The mentioned Kubernetes limitation is the [inability to have mixed protocol services](https://github.com/kubernetes/kubernetes/issues/23880). But this seems to have been addressed in recent versions.

usiegl00:
This disables the checks that prevent the sharing of ips between services.

Fixes: #21776

### DIFF
--- a/Documentation/network/lb-ipam.rst
+++ b/Documentation/network/lb-ipam.rst
@@ -429,3 +429,49 @@ for the network and broadcast addresses respectively.
     $ kubectl -n example get svc                
     NAME           TYPE           CLUSTER-IP     EXTERNAL-IP               PORT(S)          AGE
     service-blue   LoadBalancer   10.96.26.105   20.0.10.100,20.0.10.200   1234:30363/TCP   43s
+
+Sharing Keys
+------------
+
+Services can share the same IP or set of IPs with other services. This is done by setting the ``io.cilium/lb-ipam-sharing-key`` annotation on the service.
+Services that have the same sharing key annotation will share the same IP or set of IPs. The sharing key is a string that can be any value.
+
+.. code-block:: yaml
+
+  apiVersion: v1
+  kind: Service
+  metadata:
+    name: service-blue
+    namespace: example
+    labels:
+      color: blue
+    annotations:
+      "io.cilium/lb-ipam-sharing-key": "1234"
+  spec:
+    type: LoadBalancer
+    ports:
+    - port: 1234
+  ---
+  apiVersion: v1
+  kind: Service
+  metadata:
+    name: service-red
+    namespace: example
+    labels:
+      color: red
+    annotations:
+      "io.cilium/lb-ipam-sharing-key": "1234"
+  spec:
+    type: LoadBalancer
+    ports:
+    - port: 2345
+
+.. code-block:: shell-session
+
+  $ kubeclt -n example get svc
+  NAME           TYPE           CLUSTER-IP     EXTERNAL-IP               PORT(S)          AGE
+  service-blue   LoadBalancer   10.96.26.105   20.0.10.100               1234:30363/TCP   43s
+  service-red    LoadBalancer   10.96.26.106   20.0.10.100               2345:30131/TCP   43s
+
+As long as the services do not have conflicting ports, they will be allocated the same IP. If the services have conflicting ports, they will be allocated different IPs, which will be added to the set of IPs belonging to the sharing key.
+If a service has a sharing key and also requests a specific IP, the service will be allocated the requested IP and it will be added to the set of IPs belonging to that sharing key.
--- a/operator/pkg/lbipam/lbipam.go
+++ b/operator/pkg/lbipam/lbipam.go
@@ -48,6 +48,7 @@ const (
 
 	// The annotation LB IPAM will look for when searching for requested IPs
 	ciliumSvcLBIPSAnnotation = "io.cilium/lb-ipam-ips"
+	ciliumSvcLBISKAnnotation = "io.cilium/lb-ipam-sharing-key"
 
 	// The string used in the FieldManager field on update options
 	ciliumFieldManager = "cilium-operator-lb-ipam"
@@ -396,26 +397,23 @@ func (ipam *LBIPAM) satisfyAndUpdateCounts(ctx context.Context) error {
 // on if the service requests are satisfied or not.
 func (ipam *LBIPAM) handleUpsertService(ctx context.Context, svc *slim_core_v1.Service) error {
 	key := resource.NewKey(svc)
-	sv, found, _ := ipam.serviceStore.GetService(key)
-	if !found {
-		sv = &ServiceView{
-			Key: key,
-		}
-	}
 
 	// Ignore services which are not meant for us
 	if !ipam.isResponsibleForSVC(svc) {
+		sv, found, _ := ipam.serviceStore.GetService(key)
 		if !found {
+			// We were not responsible for this service before, so nothing to do
 			return nil
 		}
 
-		// Release allocations
-		for _, alloc := range sv.AllocatedIPs {
-			alloc.Origin.alloc.Free(alloc.IP)
-		}
-		ipam.serviceStore.Delete(sv.Key)
+		// we were responsible before, but not anymore
 
-		// Remove all ingress IPs
+		// Release allocations and other references as if the service was deleted
+		if err := ipam.svcOnDelete(ctx, key, svc); err != nil {
+			return fmt.Errorf("svcOnDelete: %w", err)
+		}
+
+		// Remove all ingress IPs and conditions, cleaning up the service for reuse by another controller
 		sv.Status.LoadBalancer.Ingress = nil
 		slim_meta.RemoveStatusCondition(&sv.Status.Conditions, ciliumSvcRequestSatisfiedCondition)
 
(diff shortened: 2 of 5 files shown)


dylandreimerink:
Hi @usiegl00, thank you for working on this. However, this work does not check for the conditions mentioned in #21776 which we need in order to have feature parity with MetalLB.

My suggestion is to change the allocator in the `LBRange` to store a list of `ServiceView`s and to make an index on the `rangesStore` that maps from sharing key to `LBRange + IP`. 

The algorithm would go something like:
```
When we check for an unsatisfied IP:
* We check if the service has a sharing key annotation(cilium version). 
  * If it does, we can see if it exists in the `rangeStore` via the index. 
     * If it exists, we go to the `LBRange` and get the list of `ServiceViews`.
     * Check if the ports and external traffic policy of the current service is compatible with the existing `ServiceViews`
       * if it is, add the service view to the list, and satisfy the IP
* If any of the above fail we allocate a new IP
* Once we allocated a new IP, and we have a sharing key, check if the sharing key is present in the index (we can reuse the result from earlier)
  * If no sharing key exists, add our newly allocated IP to the index
  
When a service is deleted or updated such that it no longer requests the IP:
* remove only the` ServiceView` for which the IP is no longer in use
  * If this was the last `ServiceView`, free the IP from the allocator
* If there are still entries left
  * Scan for other `ServiceViews` with the same sharing key but that are not sharing the key (another index might be wise)
    * Check if they still conflict
       * if they don't conflict, free their current IP and add to the list of `ServiceViews`
       
When a service is updated such that the ports or external traffic policy changes:
* Check if the `ServiceView` has a sharing key
  * If it does, check if its part of the allocation indicated by the sharing key index of the range store
    * If it is not, check if the `ServiceView` is still conflicting, if not, free its allocation and add to the shared IP
```

That way we nicely handle all scenarios. I know its a lot more work, but such are the requirements of the feature. These behaviors also need to be tested.

dylandreimerink:
> handle a service with a sharing key that requests multiple ips? Could rangesStore have an index that maps to a slice of ServiceViewIPs instead of just one?

That is a good point, I think it should. In dual stack mode we expect a service to get an IPv4 and IPv6 address which will both have to be in different ranges. 

This also made be think about how to handle requests for specific IPs when dealing with shared IPs :thinking:. Currently if two service request the same IP we will error on the second request. But if they have a sharing key set and are compatible then we might want to allow that. That would also mean that we could have multiple IPv4 or IPv6 IPs that are shared under a given sharing key which perhaps complicates things.

> Should the documentation changes go in this pr or a separate one?

We typically like doc changes to be in the same PR if possible.

dylandreimerink:
Awesome! I have triggered additional tests which need authorization to run, they revealed some things still. Also, like the bot has been complaining about, the Cilium project requires authors to add signoffs to their commits. See 
 https://docs.cilium.io/en/stable/contributing/development/contributing_guide/#developer-s-certificate-of-origin for details.
 
If you feel like the PR is complete you can mark it ready for review, at that pointer one or more reviewers are assigned. I will also do an additional review. Once all reviews are in and the full CI is green we can merge.

operator/pkg/lbipam/service_store.go
```diff
@@ -77,6 +83,28 @@ type ServiceView struct {
...
+func (sv *ServiceView) isCompatible(osv *ServiceView) bool {
+	// They request the use of different ports (e.g. tcp/80 for one and tcp/443 for the other).
+	for _, port1 := range sv.Ports {
+		for _, port2 := range osv.Ports {
+			if port1 == port2 {
+				return false
+			}
+		}
+	}
+	// They both use the Cluster external traffic policy, or they both point to the exact same set of pods (i.e. the pod selectors are identical).
```
dylandreimerink: > or they both point to the exact same set of pods (i.e. the pod selectors are identical).

There is a little bit more to this check than currently implemented. There are a few missing things:
* Services are namespaced, so you also have to make sure the services are in the same namespace in addition to the selectors.
* Services can have `selectors` but there are also other ways to associate endpoints/pods with services, namely with `Endpoints`, `EndpointSlices` and their Cilium specific variants. 

Implementing support for the separate endpoint objects is hard, as in a lot of work to get it right, so I suggest you to not support those.

So enforce that service with sharing-keys have to have a `Selector` and document the limitation, preferably even add a github issue for yourself or someone else to add this in a followup PR.

operator/pkg/lbipam/service_store.go
```diff
@@ -77,6 +83,28 @@ type ServiceView struct {
...
 
+func (sv *ServiceView) isCompatible(osv *ServiceView) bool {
+	// They request the use of different ports (e.g. tcp/80 for one and tcp/443 for the other).
+	for _, port1 := range sv.Ports {
+		for _, port2 := range osv.Ports {
+			if port1 == port2 {
+				return false
+			}
+		}
+	}
```
dylandreimerink: `.Ports` is of type `[]slim_core_v1.ServicePort`, so `port1` and `port2` are not port numbers, they are structures including `.Name`, `.Protocol`, `.Port`, and `.NodePort`. The current check would allow overlapping ports as long as the name is different or overlapping ports if they have different protocols which we currently don't support. 

There is also the matter of `.NodePort`. Its description says that it can be used for LoadBalancer services:
> The port on each node on which this service is exposed when type is NodePort or LoadBalancer.

But I am not sure if we need to check it here, given that node ports globally can't overlap (so I would expect the API server to have already rejected a sevice with the same nodeport as another).
```suggestion
	for _, port1 := range sv.Ports {
		for _, port2 := range osv.Ports {
			if port1.Port == port2.Port {
				return false
			}
		}
	}
```

operator/pkg/lbipam/lbipam.go
```diff
@@ -604,22 +615,25 @@ func (ipam *LBIPAM) stripOrImportIngresses(sv *ServiceView) (statusModified bool
 				continue
 			}
 
-			err = lbRange.alloc.Alloc(ip, true)
+			serviceViews := make([]*ServiceView, 0, 1)
+			serviceViews = append(serviceViews, sv)
```
dylandreimerink: Small nit, but you can write this faster

```suggestion
			serviceViews := []*ServiceView{sv}
```

operator/pkg/lbipam/lbipam.go
```diff
@@ -710,6 +755,44 @@ func (ipam *LBIPAM) satisfyServices(ctx context.Context) error {
...
+			}
+			// if it is, add the service view to the list, and satisfy the IP
+			if compatible {
+				sv.AllocatedIPs = append(sv.AllocatedIPs, *serviceViewIP)
+				serviceViews = append(serviceViews, sv)
+				lbRange.alloc.Update(serviceViewIP.IP, serviceViews)
+				break
+			}
+		}
+	}
```
dylandreimerink: Not sure this is right. We are not taking the requests of the service into account. This logic finds the first IP within the IP sharing group and assigns it. So if the group only has an IPv6 and the service requests only an IPv4 IP it will get the IPv6 address anyway.

Or if a service requests a very specific IPv4 address, it will get whatever IP from the sharing group and then potentially also the IP it actually requested.

I think the sharing key logic should be more integrated with the logic to give the service what it wants which is:
* A list of specific IPs
* an arbitrary IPv4
* an arbitrary IPv6
* an arbitrary IPv4 and an arbitrary IPv4

These requests are not in addition to the sharing key, they should be combined. 

In the case of a list of specific IPs an additional check must be added that the existing service with that IP and the new one has the same key and is compatible.

In the case of the non specific requests we should first attempt to satisfy the requests from the shared group, and if not possible fallback to the normal process.

dylandreimerink (changes requested):
Thank you so much for the work you have put into this PR! Overall really nice, though I do think I spotted some things that need to be addressed

dylandreimerink (changes requested):
@usiegl00 thanks again for making the adjustments. I have been playing these changes a bit to test things out and found some bugs. While debugging I concluded that the code had become hard to read due to the functions becoming to big so I took it upon myself to do a bit of restructuring. I figured that would be easier than to communicate all of my nit picks. I forked your branch and added 3 commits atop with tests for the bug, the bugfix and my structural changes https://github.com/dylandreimerink/cilium/tree/feature/lbipam-shared-ips

I also noticed that the history of this PR is getting a bit messy, we are not at 9 commits, not including these fixes. My suggestion is that you merge these commits into you own branch, then squash all code related commits into one commit (including my additions)  and let the docs commit be a separate commit.

dylandreimerink (approved):
The LB-IPAM logic looks good to me! :tada:

Documentation/network/lb-ipam.rst
```diff
@@ -429,3 +429,49 @@ for the network and broadcast addresses respectively.
     $ kubectl -n example get svc                
     NAME           TYPE           CLUSTER-IP     EXTERNAL-IP               PORT(S)          AGE
     service-blue   LoadBalancer   10.96.26.105   20.0.10.100,20.0.10.200   1234:30363/TCP   43s
+
+Sharing Keys
+------------
+
+Services can share the same ip or set of ips with other services. This is done by setting the ``io.cilium/lb-ipam-sharing-key`` annotation on the service.
```
ghost:
```suggestion
Services can share the same IP or set of IPs with other services. This is done by setting the ``io.cilium/lb-ipam-sharing-key`` annotation on the service.
```
Please change `ip` to `IP` here and in all other places. A bit of a nit, but it's required for consistency and for the build to pass.

ghost (changes requested):

Documentation/network/lb-ipam.rst
usiegl00: Thanks! I've updated the pr.

ghost (approved):
docs alright

PR #28813, opened 2023-10-26
bpf: fix missing ipv6 ct entry for SNATed pod traffic

### LINKED ISSUE #26886: IPv6 packets falsely dropped; conntrack issue

### What happened?

1. Set up cilium on a dual stack cluster
2. Enable firewall (`CiliumClusterwideNetworkPolicy`) with rules allowing http or ping access to external sources, but no explicit ingress rule
3. Watch dropped packages with the command I provided

On outgoing traffic the identity of the pod is resolved correctly. But on incoming traffic/the response to the outgoing traffic, the packets are dropped because the target identity can not be figured out.
If `::/0` is a explicitly allowed in the firewall, the traffic comes through though.
This seems like a problem with the conntrack table where the lookup of the conntrack table would be done after the traffic is let through, which it isn't because of the firewall setup. Therefore the packets are dropped, despite a conntrack entry existing.
I tried to find a hint in the code, but I am not familiar enough with this code base to give any specific information on that.

### Cilium Version

Client: 1.14.0-snapshot.4 6c8db759 2023-06-16T12:17:20-07:00 go version go1.20.5 linux/amd64
Daemon: 1.14.0-snapshot.4 6c8db759 2023-06-16T12:17:20-07:00 go version go1.20.5 linux/amd64

### Kernel Version

5.15.0-76-generic #83-Ubuntu SMP Thu Jun 15 19:16:32 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

### Kubernetes Version

Server Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.0", GitCommit:"1b4df30b3cdfeaba6024e81e559a6cd09a089d65", GitTreeState:"clean", BuildDate:"2023-04-11T17:04:24Z", GoVersion:"go1.20.3", Compiler:"gc", Platform:"linux/amd64"}

### Sysdump

_No response_

### Relevant log output

```shell
# root@server:/home/cilium# cilium monitor --type drop | grep 2a06:98c1:3120::3
Listening for events on x CPUs with x of shared memory
Press Ctrl-C to quit
level=info msg="Initializing dissection cache..." subsys=monitor
xx drop (Policy denied) flow 0xea08f9f8 to endpoint 0, ifindex 5, file bpf_host.c:379, , identity world->unknown: [2a06:98c1:3120::3]:80 -> [xx01:xx8:0:axx1::2exx:xxxx]:51824 tcp SYN, ACK
xx drop (Policy denied) flow 0xea08f9f8 to endpoint 0, ifindex 5, file bpf_host.c:379, , identity world->unknown: [2a06:98c1:3120::3]:80 -> [xx01:xx8:0:axx1::2exx:xxxx]:51824 tcp SYN, ACK
xx drop (Policy denied) flow 0xea08f9f8 to endpoint 0, ifindex 5, file bpf_host.c:379, , identity world->unknown: [2a06:98c1:3120::3]:80 -> [xx01:xx8:0:axx1::2exx:xxxx]:51824 tcp SYN, ACK
xx drop (Policy denied) flow 0xea08f9f8 to endpoint 0, ifindex 5, file bpf_host.c:379, , identity world->unknown: [2a06:98c1:3120::3]:80 -> [xx01:xx8:0:axx1::2exx:xxxx]:51824 tcp SYN, ACK
xx drop (Policy denied) flow 0xea08f9f8 to endpoint 0, ifindex 5, file bpf_host.c:379, , identity world->unknown: [2a06:98c1:3120::3]:80 -> [xx01:xx8:0:axx1::2exx:xxxx]:51824 tcp SYN, ACK
xx drop (Policy denied) flow 0xea08f9f8 to endpoint 0, ifindex 5, file bpf_host.c:379, , identity world->unknown: [2a06:98c1:3120::3]:80 -> [xx01:xx8:0:axx1::2exx:xxxx]:51824 tcp SYN, ACK

# root@server:/home/cilium# cilium monitor --from 408 
Listening for events on x CPUs with x of shared memory
Press Ctrl-C to quit
level=info msg="Initializing dissection cache..." subsys=monitor
-> stack flow 0xb26c28d7 , identity 3625->world state new ifindex 0 orig-ip 0.0.0.0: [fdxx:ffxx:2xx1:xx::exxe]:59284 -> [2a06:98c1:3120::3]:80 tcp SYN
-> stack flow 0x31790a5b , identity 3625->world state established ifindex 0 orig-ip 0.0.0.0: [fdxx:ffxx:2xx1:xx::exxe]:59284 -> [2a06:98c1:3120::3]:80 tcp SYN
-> stack flow 0x571bc5f7 , identity 3625->world state established ifindex 0 orig-ip 0.0.0.0: [fdxx:ffxx:2xx1:xx::exxe]:59284 -> [2a06:98c1:3120::3]:80 tcp SYN
-> stack flow 0x5a1e7b5f , identity 3625->world state established ifindex 0 orig-ip 0.0.0.0: [fdxx:ffxx:2xx1:xx::exxe]:59284 -> [2a06:98c1:3120::3]:80 tcp SYN
```

### Anything else?

`bpf_host.c:379` from the logs: https://github.com/cilium/cilium/blob/v1.14.0-snapshot.4/bpf/bpf_host.c#L378 I think

### Code of Conduct

oblazek:
When combining HostFW with kube-proxy based masquerading
host traffic will have HOST_ID for both - packet mark based
src_id and ip cache based src_id.
While SNATed pod traffic will only have HOST_ID from the
ipcache (as it doesn't pass through the iptables rules that
set the mark).
Therefore for SNATed traffic we get HOST_ID only from
ipcache and not from packet mark it means we are handling
SNATed traffic from pod that needs to have a conntrack entry
so that return traffic doesn't go through host firewall and
is skipped.

This commit fixes the issue and adds a missing ct entry for
SNATed traffic.

cc @julianwiedmann 

Fixes: https://github.com/cilium/cilium/issues/26886

related discussion: https://cilium.slack.com/archives/CDKG8NNHK/p1698155364279479

```release-note
bpf: Fix drop of IPv6 reply traffic when 1) pod-originating connection is SNATed by iptables, and 2) Host Firewall is enabled.
```

### DIFF
--- a/bpf/bpf_host.c
+++ b/bpf/bpf_host.c
@@ -91,9 +91,10 @@ static __always_inline bool identity_from_ipcache_ok(void)
 #ifdef ENABLE_IPV6
 static __always_inline __u32
 resolve_srcid_ipv6(struct __ctx_buff *ctx, struct ipv6hdr *ip6,
-		   __u32 srcid_from_proxy, const bool from_host)
+		   __u32 srcid_from_ipcache, __u32 *sec_identity,
+		   const bool from_host)
 {
-	__u32 src_id = WORLD_IPV6_ID, srcid_from_ipcache = srcid_from_proxy;
+	__u32 src_id = WORLD_IPV6_ID;
 	struct remote_endpoint_info *info = NULL;
 	union v6addr *src;
 
@@ -102,7 +103,8 @@ resolve_srcid_ipv6(struct __ctx_buff *ctx, struct ipv6hdr *ip6,
 		src = (union v6addr *) &ip6->saddr;
 		info = lookup_ip6_remote_endpoint(src, 0);
 		if (info) {
-			if (info->sec_identity) {
+			*sec_identity = info->sec_identity;
+			if (*sec_identity) {
 				/* When SNAT is enabled on traffic ingressing
 				 * into Cilium, all traffic from the world will
 				 * have a source IP of the host. It will only
@@ -111,8 +113,8 @@ resolve_srcid_ipv6(struct __ctx_buff *ctx, struct ipv6hdr *ip6,
 				 * the host. So we can ignore the ipcache if it
 				 * reports the source as HOST_ID.
 				 */
-				if (info->sec_identity != HOST_ID)
-					srcid_from_ipcache = info->sec_identity;
+				if (*sec_identity != HOST_ID)
+					srcid_from_ipcache = *sec_identity;
 			}
 		}
 		cilium_dbg(ctx, info ? DBG_IP_ID_MAP_SUCCEED6 : DBG_IP_ID_MAP_FAILED6,
@@ -135,6 +137,7 @@ struct {
 
 static __always_inline int
 handle_ipv6(struct __ctx_buff *ctx, __u32 secctx __maybe_unused,
+	    __u32 ipcache_srcid __maybe_unused,
 	    const bool from_host __maybe_unused,
 	    __s8 *ext_err __maybe_unused)
 {
@@ -143,6 +146,7 @@ handle_ipv6(struct __ctx_buff *ctx, __u32 secctx __maybe_unused,
 	bool need_hostfw = false;
 	__u8 nexthdr;
 	int hdrlen;
+	bool is_host_id = false;
 #endif /* ENABLE_HOST_FIREWALL */
 	void *data, *data_end;
 	struct ipv6hdr *ip6;
@@ -184,10 +188,11 @@ handle_ipv6(struct __ctx_buff *ctx, __u32 secctx __maybe_unused,
 
 #ifdef ENABLE_HOST_FIREWALL
 	if (from_host) {
-		if (ipv6_host_policy_egress_lookup(ctx, secctx, ip6, &ct_buffer)) {
+		if (ipv6_host_policy_egress_lookup(ctx, secctx, ipcache_srcid, ip6, &ct_buffer)) {
 			if (unlikely(ct_buffer.ret < 0))
 				return ct_buffer.ret;
 			need_hostfw = true;
+			is_host_id = secctx == HOST_ID;
 		}
 	} else if (!ctx_skip_host_fw(ctx)) {
 		/* Verifier workaround: R5 invalid mem access 'scalar'. */
@@ -209,7 +214,8 @@ handle_ipv6(struct __ctx_buff *ctx, __u32 secctx __maybe_unused,
 
 skip_host_firewall:
 	ctx_store_meta(ctx, CB_FROM_HOST,
-		       (need_hostfw ? FROM_HOST_FLAG_NEED_HOSTFW : 0));
+		       (need_hostfw ? FROM_HOST_FLAG_NEED_HOSTFW : 0) |
+		       (is_host_id ? FROM_HOST_FLAG_HOST_ID : 0));
 #endif /* ENABLE_HOST_FIREWALL */
 
 	return CTX_ACT_OK;
@@ -251,11 +257,15 @@ handle_ipv6_cont(struct __ctx_buff *ctx, __u32 secctx, const bool from_host,
 			/* The map value is zeroed so the map update didn't happen somehow. */
 			return DROP_INVALID_TC_BUFFER;
 
-		if (from_host)
-			ret = __ipv6_host_policy_egress(ctx, ip6, ct_buffer, &trace, ext_err);
-		else
+		if (from_host) {
+			bool is_host_id = from_host_raw & FROM_HOST_FLAG_HOST_ID;
+
+			ret = __ipv6_host_policy_egress(ctx, is_host_id, ip6, ct_buffer, &trace,
+							ext_err);
+		} else {
 			ret = __ipv6_host_policy_ingress(ctx, ip6, ct_buffer, &remote_id, &trace,
 							 ext_err);
+		}
 		if (IS_ERR(ret))
 			return ret;
 	}
@@ -390,7 +400,7 @@ int tail_handle_ipv6_cont_from_netdev(struct __ctx_buff *ctx)
 }
 
 static __always_inline int
-tail_handle_ipv6(struct __ctx_buff *ctx, const bool from_host)
+tail_handle_ipv6(struct __ctx_buff *ctx, __u32 ipcache_srcid, const bool from_host)
 {
 	__u32 src_sec_identity = ctx_load_meta(ctx, CB_SRC_LABEL);
 	int ret;
@@ -398,7 +408,7 @@ tail_handle_ipv6(struct __ctx_buff *ctx, const bool from_host)
 
 	ctx_store_meta(ctx, CB_SRC_LABEL, 0);
 
-	ret = handle_ipv6(ctx, src_sec_identity, from_host, &ext_err);
+	ret = handle_ipv6(ctx, src_sec_identity, ipcache_srcid, from_host, &ext_err);
 
 	/* TC_ACT_REDIRECT is not an error, but it means we should stop here. */
 	if (ret == CTX_ACT_OK) {
@@ -424,13 +434,20 @@ tail_handle_ipv6(struct __ctx_buff *ctx, const bool from_host)
 __section_tail(CILIUM_MAP_CALLS, CILIUM_CALL_IPV6_FROM_HOST)
 int tail_handle_ipv6_from_host(struct __ctx_buff *ctx __maybe_unused)
 {
-	return tail_handle_ipv6(ctx, true);
+	__u32 ipcache_srcid = 0;
+
+#if defined(ENABLE_HOST_FIREWALL) && !defined(ENABLE_MASQUERADE_IPV6)
+	ipcache_srcid = ctx_load_meta(ctx, CB_IPCACHE_SRC_LABEL);
+	ctx_store_meta(ctx, CB_IPCACHE_SRC_LABEL, 0);
+#endif /* defined(ENABLE_HOST_FIREWALL) && !defined(ENABLE_MASQUERADE_IPV6) */
+
+	return tail_handle_ipv6(ctx, ipcache_srcid, true);
 }
 
 __section_tail(CILIUM_MAP_CALLS, CILIUM_CALL_IPV6_FROM_NETDEV)
 int tail_handle_ipv6_from_netdev(struct __ctx_buff *ctx)
 {
-	return tail_handle_ipv6(ctx, false);
+	return tail_handle_ipv6(ctx, 0, false);
 }
 
 # ifdef ENABLE_HOST_FIREWALL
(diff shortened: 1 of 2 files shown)


julianwiedmann:
Hi @oblazek , ty for looking into this! Will have a closer look ASAP.

But just to set some context for review, as the description doesn't capture my current understanding.
With `ENABLE_HOST_FIREWALL` set, we need to handle two cases in the egress path:
- actual host traffic. Here we want to apply egress policy.
- SNATed pod traffic (when not using BPF-masquerading). Here we *don't* apply egress policy. But we still create a CT entry, so that replies can naturally bypass ingress policy.

We differentiate between the two cases by comparing the mark-derived identity with the ipcache-derived identity. Host traffic will have `HOST_ID` for both, while SNATed pod traffic will only have `HOST_ID` from the ipcache (as it doesn't pass through the iptables rule that sets the mark).

Throwing `--install-iptables-rules=false` into this picture means that legitimate host traffic will not be marked either, and thus is mis-identified as "SNATed pod traffic". And so while we create a CT entry for it (and the reply traffic flow works), we don't actually apply egress policy to it when we should.

Do you agree with that assessment, or is there still something missing from the picture?

oblazek:
> Throwing `--install-iptables-rules=false` into this picture means that legitimate host traffic will not be marked either, and thus is **mis-identified** as "SNATed pod traffic". And so while we create a CT entry for it (and the reply traffic flow works), we don't actually apply egress policy to it when we should.
> 
> Do you agree with that assessment, or is there still something missing from the picture?

You are totally right, if iptables are not present, there is no mark and `return __ipv6_host_policy_egress(ctx, src_id == HOST_ID,...`  can't be relied on. So at least there should be an logical OR:
`src_id == HOST_ID || ipcache_srcid == HOST_ID` so that egress policy is correctly enforced.

bpf/bpf_host.c
```diff
@@ -424,13 +435,20 @@ tail_handle_ipv6(struct __ctx_buff *ctx, const bool from_host)
...
 {
-	return tail_handle_ipv6(ctx, true);
+	__u32 ipcache_srcid = 0;
+
+#if defined(ENABLE_HOST_FIREWALL) && !defined(ENABLE_MASQUERADE_IPV6)
+	ipcache_srcid = ctx_load_meta(ctx, CB_IPCACHE_SRC_LABEL);
+	ctx_store_meta(ctx, CB_IPCACHE_SRC_LABEL, 0);
+#endif /* defined(ENABLE_HOST_FIREWALL) && !defined(ENABLE_MASQUERADE_IPV6) */
+
+	return tail_handle_ipv6(ctx, ipcache_srcid, true);
```
julianwiedmann, line 444: This is still missing the corresponding write to `CB_IPCACHE_SRC_LABEL` in `do_netdev()`, correct?

bpf/lib/host_firewall.h
```diff
@@ -394,7 +437,8 @@ ipv4_host_policy_egress(struct __ctx_buff *ctx, __u32 src_id,
 	if (ct_buffer.ret < 0)
 		return ct_buffer.ret;
 
-	return __ipv4_host_policy_egress(ctx, src_id == HOST_ID, ip4, &ct_buffer, trace, ext_err);
+	return __ipv4_host_policy_egress(ctx, src_id == HOST_ID || ipcache_srcid == HOST_ID,
+					ip4, &ct_buffer, trace, ext_err);
```
julianwiedmann: Let's please tackle the two problems with separate patches / PRs. Understanding the IPv6 whitelisting change is difficult enough, this subtle IPv4 change only makes things more complicated. And it introduces a severe regression for current users (they now suddenly need host-level IPv4 egress policies for their SNATed pod traffic).

So let's focus this first patch on the missing IPv6 allow-listing, when combining HostFW with kube-proxy based masquerading. Could you update the PR / patch description accordingly? Happy to help with tackling `--install-iptables-rules=false` once this is sorted.

julianwiedmann:
> > Throwing `--install-iptables-rules=false` into this picture means that legitimate host traffic will not be marked either, and thus is **mis-identified** as "SNATed pod traffic". And so while we create a CT entry for it (and the reply traffic flow works), we don't actually apply egress policy to it when we should.
> > Do you agree with that assessment, or is there still something missing from the picture?
> 
> You are totally right, if iptables are not present, there is no mark and `return __ipv6_host_policy_egress(ctx, src_id == HOST_ID,...` can't be relied on. So at least there should be a logical OR: `src_id == HOST_ID || ipcache_srcid == HOST_ID` so that egress policy is correctly enforced.
> 
> I believe the same should be done for ipv4

agree, but this needs to be a conditional change (only when `--install-iptables-rules=false` is set).

julianwiedmann (changes requested):
See comments inline.

bpf/bpf_host.c
oblazek: agree

bpf/lib/host_firewall.h
oblazek: Understood, will make the changes.

oblazek:
updated, now the PR should only contain ipv6 whitelisting related changes.. will do the rest in the followup PR

bpf/bpf_host.c
```diff
@@ -481,7 +488,7 @@ resolve_srcid_ipv4(struct __ctx_buff *ctx, struct iphdr *ip4,
 	/* Packets from the proxy will already have a real identity. */
 	if (identity_is_reserved(srcid_from_ipcache)) {
 		info = lookup_ip4_remote_endpoint(ip4->saddr, 0);
-		if (info != NULL) {
+		if (info) {
```
julianwiedmann: In case that you need to rebase once more, please drop this cleanup :).

bpf/lib/host_firewall.h
```diff
@@ -23,8 +52,14 @@ ipv6_host_policy_egress_lookup(struct __ctx_buff *ctx, __u32 src_sec_identity,
...
-	/* Only enforce host policies for packets from host IPs. */
-	if (src_sec_identity != HOST_ID)
+	/* Further action is needed in two cases:
+	 * 1. Packets from host IPs: need to enforce host policies.
+	 * 2. SNATed packets from pods: need to create a CT entry to skip
+	 *    applying host policies to reply packets
+	 *    (see ipv6_whitelist_snated_egress_connections).
+	 */
+	if (src_sec_identity != HOST_ID &&
+	    (is_defined(ENABLE_MASQUERADE_IPV6)))
```
julianwiedmann:
```suggestion
	if (src_sec_identity != HOST_ID &&
	    (is_defined(ENABLE_MASQUERADE_IPV6)  || ipcache_srcid != HOST_ID))
```

bpf/bpf_host.c
julianwiedmann: Hmm now the `CB_IPCACHE_SRC_LABEL` part is completely gone ... that's unexpected? Pretty sure we need it.

julianwiedmann (changes requested):
Ty @oblazek ! There's one part where we need to check if the mark-based identity and ipcache-based identity match up (see below). But it's missing the ipcache-based identity ... could you have a look into passing that through, along what IPv4 does?

oblazek:
> Hmm now the CB_IPCACHE_SRC_LABEL part is com...

yeah I must have missed that when working on different branches :(

bpf/lib/host_firewall.h
oblazek: done

bpf/bpf_host.c
oblazek: done

bpf/lib/host_firewall.h
```diff
@@ -15,16 +15,51 @@
...
-	/* Only enforce host policies for packets from host IPs. */
-	if (src_sec_identity != HOST_ID)
+	/* Further action is needed in two cases:
+	 * 1. Packets from host IPs: need to enforce host policies.
+	 * 2. SNATed packets from pods: need to create a CT entry to skip
+	 *    applying host policies to reply packets
+	 *    (see ipv6_whitelist_snated_egress_connections).
+	 */
+	if (src_sec_identity != HOST_ID &&
+	    (is_defined(ENABLE_MASQUERADE_IPV6) || ipcache_srcid != HOST_ID))
```
YutaroHayakawa, line 62: I might be confused, but should this be `!is_defined(ENABLE_MASQUERADE_IPV6)`? We want to catch the case that masquerading is disabled in this case right?

bpf/bpf_host.c
```diff
@@ -91,7 +91,8 @@ static __always_inline bool identity_from_ipcache_ok(void)
 #ifdef ENABLE_IPV6
 static __always_inline __u32
 resolve_srcid_ipv6(struct __ctx_buff *ctx, struct ipv6hdr *ip6,
-		   __u32 srcid_from_proxy, const bool from_host)
+		   __u32 srcid_from_proxy, __u32 *sec_identity,
```
YutaroHayakawa:
```suggestion
		   __u32 srcid_from_ipcache, __u32 *sec_identity,
```

and delete below assignment.

bpf/bpf_host.c
```diff
@@ -439,7 +456,7 @@ handle_to_netdev_ipv6(struct __ctx_buff *ctx, struct trace_ctx *trace, __s8 *ext
 {
 	void *data, *data_end;
 	struct ipv6hdr *ip6;
-	__u32 src_id = 0;
+	__u32 src_id = 0, ipcache_srcid = 0;
```
YutaroHayakawa:
```suggestion
	__u32 srcid = 0, ipcache_srcid = 0;
```

for consistency.

YutaroHayakawa (changes requested):
I made an initial look. The logic to set `CB_IPCACHE_SRC_LABEL` is still missing, but otherwise, the logic looks correct to me. I found some small naming inconsistency and one potential bug.

bpf/lib/host_firewall.h
oblazek: I don't think so, if we are talking about "kube-proxy based masq" (which means that pod traffic is SNATed anyway, this function gets HOST_IP in the packet, but there is no packet mark) --> `if (1 && (0 || 0))` (src_sec_identity is not HOST_ID, is_defined(ENABLE_MASQUERADE_IPV6) is false, ipcache_srcid is HOST_ID) => false, which means the code returns true which we want so that `ipv6_whitelist_snated_egress_connections` can be run afterwards

bpf/bpf_host.c
oblazek: :+1:

bpf/lib/host_firewall.h
YutaroHayakawa: Ah, never mind. I was wrong. `is_defined(ENABLE_MASQUERADE_IPV6) == true` means "BPF-based IPv6 masquerading is enabled", so when `is_defined(ENABLE_MASQUERADE_IPV6) == true`, the packet is not masqueraded at this point.

YutaroHayakawa (approved):
Now LGTM 👍

julianwiedmann (approved):
ty!

Issue #14882, opened 2021-02-08
Upgrade/downgrade test for patch releases


pchaigno:
On master, K8sUpdates tests the transition from master to the development branch of our latest stable release and back. For example on current master, [it tests](https://github.com/cilium/cilium/blob/265d6775c2e0eb4e9d1dfef3ebeaa19a9b976e5f/test/k8sT/Updates.go#L124-L129): `master -> v1.9 -> master`. On [the v1.9 development branch](https://github.com/cilium/cilium/tree/v1.9), K8sUpdates [tests](https://github.com/cilium/cilium/blob/v1.9/test/k8sT/Updates.go#L113-L119) the transition `v1.9 -> v1.8 -> v1.9`.

We should also test transitions between patch versions. Our latest v1.9.x release is v1.9.4, so on v1.9, K8sUpdates should test the transition `v1.9 -> v1.9.4 -> v1.9`. When we release a new patch version, we would update the test accordingly to always test the transition to/from the latest patch release.

On master, the test for patch releases would be disabled as we usually don't have corresponding patch releases.

This issue is a matter of:
- [ ] Running K8sUpdates locally by following [the documentation](https://docs.cilium.io/en/latest/contributing/testing/e2e/#running-kubernetes-tests).
- [ ] Adding a new `It` block in `K8sUpdates.go` to test the above transition.
- [ ] Validating by running the test again.

The new test can be developed on master with the transition `v1.9 -> v1.9.x -> v1.9` (on whatever the latest stable minor release is at the time of development). We will then disable it on master and enable it only in backports.

ti-mo:
@brb is tackling this in ci-e2e: https://github.com/cilium/cilium/pull/25790. Set this issue to close when merged.

qmonnet:
{Up/down}grade from patch release was merged to `main`, the 3 related backport PRs are pending the last reviews/CI checks.

qmonnet:
Tests have been merged on all branches (#28815, #28876, #29003, #29005). Happy to close this as completed.

PR #28815, opened 2023-10-26
Test upgrade/downgrade to patch release for IPsec

qmonnet:
- contrib/scripts: Support patch releases in print-downgrade-version.sh
- ci/ipsec: Add upgrade/downgrade tests for patch releases
- ci/ipsec: Skip upgrade/downgrade test to patch release on main branch

Related to
- #14882

Corresponding backport PRs:
- https://github.com/cilium/cilium/pull/28876
- https://github.com/cilium/cilium/pull/29003
- https://github.com/cilium/cilium/pull/29005

### DIFF
--- a/.github/workflows/tests-ipsec-upgrade.yaml
+++ b/.github/workflows/tests-ipsec-upgrade.yaml
@@ -75,8 +75,11 @@ jobs:
       fail-fast: false
       max-parallel: 16
       matrix:
+        config: ['5.4', '5.10', 'bpf-next']
+        mode: ['minor', 'patch']
         include:
-          - name: '1'
+          # Define three config sets
+          - config: '5.4'
             # renovate: datasource=docker depName=quay.io/lvh-images/kind
             kernel: '5.4-20231026.065108'
             kube-proxy: 'iptables'
@@ -84,7 +87,7 @@ jobs:
             tunnel: 'disabled'
             encryption: 'ipsec'
 
-          - name: '2'
+          - config: '5.10'
             # renovate: datasource=docker depName=quay.io/lvh-images/kind
             kernel: '5.10-20231026.065108'
             kube-proxy: 'iptables'
@@ -93,7 +96,7 @@ jobs:
             encryption: 'ipsec'
             endpoint-routes: 'true'
 
-          - name: '3'
+          - config: 'bpf-next'
             # renovate: datasource=docker depName=quay.io/lvh-images/kind
             kernel: 'bpf-next-20231030.012704'
             kube-proxy: 'iptables'
@@ -101,6 +104,31 @@ jobs:
             tunnel: 'vxlan'
             encryption: 'ipsec'
             endpoint-routes: 'true'
+
+          # Add names to matrix combinations of {config, mode}
+          - config: '5.4'
+            mode: 'minor'
+            name: '1'
+
+          - config: '5.10'
+            mode: 'minor'
+            name: '2'
+
+          - config: 'bpf-next'
+            mode: 'minor'
+            name: '3'
+
+          - config: '5.4'
+            mode: 'patch'
+            name: '4'
+
+          - config: '5.10'
+            mode: 'patch'
+            name: '5'
+
+          - config: 'bpf-next'
+            mode: 'patch'
+            name: '6'
 
     timeout-minutes: 60
     steps:
@@ -122,14 +150,32 @@ jobs:
             SHA="${{ github.sha }}"
           fi
           echo sha=${SHA} >> $GITHUB_OUTPUT
-          CILIUM_DOWNGRADE_VERSION=$(contrib/scripts/print-downgrade-version.sh)
+          if [ "${{ matrix.mode }}" = "minor" ]; then
+            CILIUM_DOWNGRADE_VERSION=$(contrib/scripts/print-downgrade-version.sh)
+            IMAGE_TAG=${CILIUM_DOWNGRADE_VERSION}
+          else
+            # Upgrade from / downgrade to patch release.
+            # In some cases we expect to fail to get the version number, do not
+            # fail the workflow in such case. This includes:
+            # - on main branch where we don't have preceeding patch releases
+            # - on stable branches on top of release preparation commits, where
+            #   we bump the patch version number to 90 and we can't easily
+            #   derive the number of the previous patch release version from
+            #   that.
+            CILIUM_DOWNGRADE_VERSION=$(contrib/scripts/print-downgrade-version.sh patch || true)
+            # Pass an empty tag to the cilium-config action to fall back to the
+            # default release image, without crafting an image path with the
+            # "-ci" suffix
+            IMAGE_TAG=''
+          fi
           echo downgrade_version=${CILIUM_DOWNGRADE_VERSION} >> $GITHUB_OUTPUT
+          echo image_tag=${IMAGE_TAG} >> $GITHUB_OUTPUT
 
       - name: Derive stable Cilium installation config
         id: cilium-stable-config
         uses: ./.github/actions/cilium-config
         with:
-          image-tag: ${{ steps.vars.outputs.downgrade_version }}
+          image-tag: ${{ steps.vars.outputs.image_tag }}
           chart-dir: './cilium-${{ steps.vars.outputs.downgrade_version }}/install/kubernetes/cilium/'
           tunnel: ${{ matrix.tunnel }}
           endpoint-routes: ${{ matrix.endpoint-routes }}
--- a/.github/workflows/tests-ipsec-upgrade.yaml
+++ b/.github/workflows/tests-ipsec-upgrade.yaml
@@ -173,6 +173,7 @@ jobs:
 
       - name: Derive stable Cilium installation config
         id: cilium-stable-config
+        if: ${{ steps.vars.outputs.downgrade_version != '' }}
         uses: ./.github/actions/cilium-config
         with:
           image-tag: ${{ steps.vars.outputs.image_tag }}
@@ -192,6 +193,7 @@ jobs:
 
       - name: Derive newest Cilium installation config
         id: cilium-newest-config
+        if: ${{ steps.vars.outputs.downgrade_version != '' }}
         uses: ./.github/actions/cilium-config
         with:
           image-tag: ${{ steps.vars.outputs.sha }}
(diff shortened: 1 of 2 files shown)


contrib/scripts/print-downgrade-version.sh
```diff
@@ -1,15 +1,38 @@
...
 else
-  echo "ERROR: failed to parse version '$VERSION'"
+  >&2 echo "ERROR: failed to parse version '$VERSION'"
   exit 1
 fi
+
+if [[ ${1-} == "patch" ]] ; then
+    # If user passed "patch" as first argument, print the latest patch version
+    case ${patch} in
+        0|90)
```
pchaigno, line 23: (Not to address in this PR. Just trying to start that discussion.)

Should we try to test up/downgrades from/to release candidates as well? Would the last release candidate be the version prior to the `.0` here?

.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -221,6 +267,7 @@ jobs:
             mkdir -p cilium-junits
 
       - name: Wait for images to be available
+        if: ${{ matrix.mode != 'patch' }}
```
pchaigno: These are the images for the PR under test, aren't they? Why don't we need those images when testing the patch up/downgrade path?

pchaigno (commented):
Nice commit history and description! :heart:

contrib/scripts/print-downgrade-version.sh
qmonnet: > Should we try to test up/downgrades from/to release candidates as well?

I haven't worked with IPsec as much as you have, and I don't have the same hindsight about it, so... you tell me? I suppose that yes, we could add upgrades/downgrades with previews next, to test the latest changes on `main`.

> Would the last release candidate be the version prior to the `.0` here?

No, with the current scheme we always get `.0` on `main`. I don't think we have a way to find previews with the current script, we would probably need to download the Git history in the workflow and look at tags.

.github/workflows/tests-ipsec-upgrade.yaml
qmonnet: Sounds right, I got confused and thought they were the ones for the patch releases. The tests still passed, likely because the image is usually ready in time, which comforted me in that belief... I'll revert this bit. Thanks!

qmonnet: I'll propagate the fix to the backport PRs, too.

contrib/scripts/print-downgrade-version.sh
pchaigno: Nothing IPsec specific here but yes, there could be IPsec changes in release candidates as well. Knowing about regressions before the .0 would probably be useful.

My main concern is that we'd grow the script more complex and an error in the script can cause some tests to be incorrectly skipped today.

pchaigno (approved):

contrib/scripts/print-downgrade-version.sh
qmonnet: > My main concern is that we'd grow the script more complex and an error in the script can cause some tests to be incorrectly skipped today.

Sounds like a legitimate concern, we could probably come up with a different script for that case though, given that we won't be able to infer the release candidate commits from the `VERSION` file. We'd need some additional conditions in the GitHub workflow to run a different script for {up,down}grades with previews. Including a way to properly figure out we're on the `main` branch (we don't want to find and test the latest preview on stable branches).

I can take a look after this PR, if this is something we want.

brb (approved):
Thanks!

julianwiedmann:
Sorry, I need to ask because I've been tearing my hair out :slightly_smiling_face:.

The commit description says

```
Currently, we test upgrades and downgrades for IPsec against the
previous stable branch, for example:

    - On main branch (v1.15-dev):
            main (PR HEAD) -> v1.14 (branch tip) -> main (PR HEAD)

    - On older stable branches:
            v1.14 (PR HEAD) -> v1.13 (branch tip) -> v1.14 (PR HEAD)
```

My understanding was that the tested path is an **upgrade** (say, when [running](https://github.com/cilium/cilium/actions/runs/6892430659/job/18749701531) on a v1.13 PR):
- install `v1.12` tip
- upgrade to `v1.13` PR HEAD
- downgrade to `v1.12` tip

This would be the **opposite** order from what you describe. Is that just a mixup?

qmonnet:
@julianwiedmann Good thing you asked, this commit description is wrong indeed. I wrote it the other way initially, but apparently I got confused at some point and changed to the actual, mistaken description - then forgot to check it a final time before pushing :thinking:. Anyway, I just fixed it.

contrib/scripts/print-downgrade-version.sh
pchaigno: Let's postpone for now. There are higher priority tasks around CI and IPsec.

julianwiedmann (approved):
> @julianwiedmann Good thing you asked, this commit description is wrong indeed. I wrote it the other way initially, but apparently I got confused at some point and changed to the actual, mistaken description - then forgot to check it a final time before pushing 🤔. Anyway, I just fixed it.

Thank you!

pchaigno (approved):
Re-approving to catch the new cilium/ipsec review request.

.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -227,6 +280,7 @@ jobs:
           done
 
       - name: Install Cilium ${{ steps.vars.outputs.downgrade_version }} (${{ matrix.name }})
+        if: ${{ steps.vars.outputs.downgrade_version != '' }}
```
pchaigno, line 283: @qmonnet It seems this line is missing in the `Wait for images to be available` step above. Not intentional I guess?

qmonnet: Right, this is a mistake. Resulting from the fact my first version skipped waiting for the images entirely for upgrades from patch versions, as we discussed [earlier](https://github.com/cilium/cilium/pull/28815#discussion_r1401011688). I'll send a fix, thanks!

qmonnet: https://github.com/cilium/cilium/pull/29793

.github/workflows/tests-ipsec-upgrade.yaml
```diff
@@ -122,14 +150,33 @@ jobs:
...
+          if [ "${{ matrix.mode }}" = "minor" ]; then
+            CILIUM_DOWNGRADE_VERSION=$(contrib/scripts/print-downgrade-version.sh)
+            IMAGE_TAG=${CILIUM_DOWNGRADE_VERSION}
+          else
+            # Upgrade from / downgrade to patch release.
+            # In some cases we expect to fail to get the version number, do not
+            # fail the workflow in such case. This includes:
+            # - on main branch where we don't have preceeding patch releases
+            # - on stable branches on top of release preparation commits, where
+            #   we bump the patch version number to 90 and we can't easily
```
joestringer, line 162: We don't bump the version number to 90 any more, but we use vX.Y-dev instead. There's still no dedicated patch release though so I guess this doesn't change anything.

joestringer: I didn't think we set the version in stable branch release prep commits at all :thinking:

qmonnet: > We don't bump the version number to 90 any more, but we use vX.Y-dev instead. There's still no dedicated patch release though so I guess this doesn't change anything.

Correct, in this case this is the same thing as having `0` as the patch release for the script.

> I didn't think we set the version in stable branch release prep commits at all 🤔

Double-checking now, it seems you're correct here as well. I thought I saw it on stable branches, but I suppose this was just the successive preparation commits added to `main` at the time, in preparation for the new minor releases. It's no in use after we've forked a new stable branch, as far as I can see.

joestringer: What's the consequence of this? Does it mean that we should just drop these lines that incorrectly document the behaviour with stable branches, or is there also an impact on other parts of the workflow?

qmonnet: There is no impact on other parts of the workflow. The script fails to retrieve the previous patch version when the VERSION file ends with 0 or 90.

- Ending with 90 does not in fact occur on stable branches, so that's just one fewer case to worry about. It also no longer occurs on main branch given that we're using another scheme; so it no longer occurs at all. We could remove the mentions to this case, and even remove this case from `print-dowgrade-version.sh`.
- What we have instead is the `x.y.0-dev` scheme, which for the script is equivalent to `x.y.0` and treated the same as `x.y.90` for finding the previous patch version: we fail to get it:

    ```console
    $ VERSION=1.15.0-dev ./contrib/scripts/print-downgrade-version.sh           
    v1.14
    $ VERSION=1.15.0-dev ./contrib/scripts/print-downgrade-version.sh patch
    ERROR: failed to deduce patch release previous to version '1.15.0-dev'
    ```

- The case where the patch release found in `VERSION` is 0 remains unchanged, this part of the comment is till valid; in this case we fail to retrieve the previous patch version, because we're most likely on `main` and we don't have a previous patch version to find.

So everything related to the version ending in `.90` can probably be removed, but in the meantime this is just dead code. The only case it would have an impact is if we reach `x.y.90` for real, in which case we'd skip the upgrade/dowgrade to/from `x.y.89`. Sounds like a minor concern at this time.

I can do a follow-up to clean up these bits but don't want to commit to it for this week.

joestringer: Awesome, thanks for the analysis. It would be great if you could follow up, but there's no rush.

Incidentally, there are some potentially relevant bash snippets around identifying the tags in a target branch here:

https://github.com/cilium/cilium/blob/3d3f69ce3386d836ea45219239f0d69daea7af7a/Documentation/check-crd-compat-table.sh#L44

If it makes sense, we could factor some of these out for reuse.

qmonnet: OK, setting a reminder to myself for the new year

qmonnet: Looking again at `cilium/Documentation/check-crd-compat-table.sh`, I prefer to not consolidate the two scripts. The reason is that one retrieves the tags from Git, meaning that the full Git history must be available; whereas `contrib/scripts/print-downgrade-version.sh` tries to simply deduce the previous number from the value in `VERSION`, which allows CI workflows to run it on shallow clones only.

qmonnet: https://github.com/cilium/cilium/pull/30413

PR #28817, opened 2023-10-26
Srv6 encap relocation

ldelossa:
This change moves SRv6 h.encap to bpf_lxc rather then bpf_host. 

This ensures that the packet is encapsulated before it enters the host networking stack.

Ensuring the encapsulated packet hits the network stack resolves issues with IPv4Masq compatibility. 

Additionally, this change now ensures the source address of the encapsulated packet is consistent, see commit messages for more details. 

This changes has the added benefit of removing a double FIB lookup on an SRv6 encapsulated egress packet. 

```release-note
srv6: modify h.encap location in the datapath to avoid incompatibility with IPv4Masq
```

### DIFF
--- a/bpf/bpf_host.c
+++ b/bpf/bpf_host.c
@@ -1142,96 +1142,6 @@ handle_netdev(struct __ctx_buff *ctx, const bool from_host)
 	return do_netdev(ctx, proto, from_host);
 }
 
-#ifdef ENABLE_SRV6
-static __always_inline int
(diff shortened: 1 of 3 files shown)


ldelossa:
' Cilium IPsec upgrade (ci-ipsec-upgrade) ' failure is unrelated to this change.

YutaroHayakawa (approved):
LGTM 👍

PR #28818, opened 2023-10-26
Config option to set the default IP Pool

### LINKED ISSUE #27131: CFP: Config option to customize the default IP Pool when using MultiPool
## Cilium Feature Proposal
        
With the new MultiPool IPAM mode, I want to be able to set the default IP Pools a node uses, so I can have nodes in different datacenters using IP ranges within the over-arching IP summary for that DC.

For example:
 - DC1 - 10.0.0.0/16
    - Node 1 - 10.0.0.0/24
    - Node 2 - 10.0.1.0/24
 - DC2 - 10.1.0.0/16
    - Node 3 - 10.1.0.0/24
    - Node 4 - 10.1.1.0/24

I'd need to be able to apply different default IP pools to different nodes based on Node label.

Pod / Namespace specified IPPools should naturally take priority.

As per the [conversation in Slack](https://cilium.slack.com/archives/C1MATJ5U5/p1690535827793529), we already have a Const `PoolDefault` which is used [here](https://github.com/cilium/cilium/blob/d18cf579816ab8a68e341994fc036aa9efba7ac2/pkg/ipam/ipam.go#L233) as the fallback if the pod/namespace doesn't specify a specific pool.

The proposed solution is to change that `const` to a `var` and set it via a new `default-ip-pool (DefaultIPPool)` option in the Daemon Configuration, which would have a default value of `default` (to match current behaviour).

This can then be set at a cluster level, but more importantly for this issue, at a node level using `CiliumNodeConfig`. The above use case could be solved by creating a `CiliumNodeConfig` for each DC, use a label selector like `topology.kubernetes.io/zone == DCX`, and set the new `DefaultIPPool` option to the `IPPool` for `DCX`.

A future enhancement would be to allow the selection of a DefaultIPPool using a labelselector, to avoid hard coding in configs, but that would be a next step / in-addition.

I will log a PR with these changes once I get an initial go ahead here. Please do let me know what I should name this new flag.

Once the CFP is close to being finalized, please add it as a PR to the [design-cfps](https://github.com/cilium/design-cfps) repo for final approval.

chaunceyjiang:
Config option to customize the default IP Pool when using MultiPool

Fixes: #27131

Fixes: #issue-number

```release-note
 Config option to customize the default IP Pool when using MultiPool
```

### DIFF
--- a/daemon/cmd/daemon_main.go
+++ b/daemon/cmd/daemon_main.go
@@ -297,9 +297,13 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 
 	flags.Var(option.NewNamedMapOptions(option.IPAMMultiPoolPreAllocation, &option.Config.IPAMMultiPoolPreAllocation, nil),
 		option.IPAMMultiPoolPreAllocation,
-		fmt.Sprintf("Defines the minimum number of IPs a node should pre-allocate from each pool (default %s)", defaults.IPAMMultiPoolPreAllocation))
-	vp.SetDefault(option.IPAMMultiPoolPreAllocation, defaults.IPAMMultiPoolPreAllocation)
+		fmt.Sprintf("Defines the minimum number of IPs a node should pre-allocate from each pool (default %s=8)", defaults.IPAMDefaultIPPool))
+	vp.SetDefault(option.IPAMMultiPoolPreAllocation, "")
 	option.BindEnv(vp, option.IPAMMultiPoolPreAllocation)
+
+	flags.String(option.IPAMDefaultIPPool, defaults.IPAMDefaultIPPool, "Name of the default IP Pool when using multi-pool")
+	vp.SetDefault(option.IPAMDefaultIPPool, defaults.IPAMDefaultIPPool)
+	option.BindEnv(vp, option.IPAMDefaultIPPool)
 
 	flags.StringSlice(option.ExcludeLocalAddress, []string{}, "Exclude CIDR from being recognized as local address")
 	option.BindEnv(vp, option.ExcludeLocalAddress)
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -380,9 +380,8 @@ const (
 	// CiliumNode.Spec.IPAM.PreAllocate if no value is set
 	IPAMPreAllocation = 8
 
-	// IPAMMultiPoolPreAllocation is the default value for multi-pool IPAM
-	// pre-allocations
-	IPAMMultiPoolPreAllocation = "default=8"
+	// IPAMDefaultIPPool is the default value for the multi-pool default pool name.
+	IPAMDefaultIPPool = "default"
 
 	// ENIFirstInterfaceIndex is the default value for
 	// CiliumNode.Spec.ENI.FirstInterfaceIndex if no value is set.
--- a/pkg/ipam/ipam.go
+++ b/pkg/ipam/ipam.go
@@ -15,6 +15,7 @@ import (
 	"github.com/cilium/cilium/pkg/k8s/client"
 	"github.com/cilium/cilium/pkg/logging"
 	"github.com/cilium/cilium/pkg/logging/logfields"
+	"github.com/cilium/cilium/pkg/option"
 )
 
 var (
@@ -210,7 +211,12 @@ func (ipam *IPAM) isIPExcluded(ip net.IP, pool Pool) (string, bool) {
 // PoolOrDefault returns the default pool if no pool is specified.
 func PoolOrDefault(pool string) Pool {
 	if pool == "" {
-		return PoolDefault
+		return PoolDefault()
 	}
 	return Pool(pool)
+}
+
+// PoolDefault returns the default pool
+func PoolDefault() Pool {
+	return Pool(option.Config.IPAMDefaultIPPool)
 }
--- a/pkg/ipam/types.go
+++ b/pkg/ipam/types.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/cilium/cilium/pkg/cidr"
 	"github.com/cilium/cilium/pkg/datapath/types"
-	ipamOption "github.com/cilium/cilium/pkg/ipam/option"
 	"github.com/cilium/cilium/pkg/lock"
 )
 
@@ -144,7 +143,3 @@ type Pool string
 func (p Pool) String() string {
 	return string(p)
 }
-
-const (
-	PoolDefault Pool = ipamOption.PoolDefault
-)
--- a/pkg/option/config.go
+++ b/pkg/option/config.go
@@ -878,6 +878,9 @@ const (
 	// IPAMMultiPoolPreAllocation defines the pre-allocation value for each IPAM pool
 	IPAMMultiPoolPreAllocation = "ipam-multi-pool-pre-allocation"
 
+	// IPAMDefaultIPPool defines the default IP Pool when using multi-pool
+	IPAMDefaultIPPool = "ipam-default-ip-pool"
+
 	// XDPModeNative for loading progs with XDPModeLinkDriver
 	XDPModeNative = "native"
 
@@ -2162,7 +2165,8 @@ type DaemonConfig struct {
 
 	// IPAMMultiPoolPreAllocation defines the pre-allocation value for each IPAM pool
 	IPAMMultiPoolPreAllocation map[string]string
-
+	// IPAMDefaultIPPool the default IP Pool when using multi-pool
+	IPAMDefaultIPPool string
 	// AutoCreateCiliumNodeResource enables automatic creation of a
 	// CiliumNode resource for the local node
 	AutoCreateCiliumNodeResource bool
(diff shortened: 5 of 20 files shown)


pkg/ipam/types.go
```diff
@@ -145,6 +145,6 @@ func (p Pool) String() string {
 	return string(p)
 }
 
-const (
-	PoolDefault Pool = ipamOption.PoolDefault
+var (
+	PoolDefault = Pool(option.Config.IPAMDefaultIPPool)
```
tklauser: I don't think this will work. This will be initialized at package load time, i.e. before flags are parsed and `option.Config.*` is populated. So `PoolDefault` will always contain the default value of `option.Config.IPAMDefaultIPPool`.

tklauser (changes requested):

tklauser:
Requested a review from @gandro given he is currently working on addressing some tech-debt in the multi-pool IPAM code and this might conflict with his work.

tommyp1ckles:
Nice improvement, changes make sense to me.

@gandro Should this have some sort of validation that the pools in question exist?

pkg/ipam/types.go
chaunceyjiang: Updated. @tklauser

pkg/ipam/ipam.go
```diff
@@ -210,7 +211,12 @@ func (ipam *IPAM) isIPExcluded(ip net.IP, pool Pool) (string, bool) {
...
 	if pool == "" {
-		return PoolDefault
+		return PoolDefault()
 	}
 	return Pool(pool)
 }
+
+// PoolDefault returns the default pool
+func PoolDefault() Pool {
+	return Pool(option.Config.IPAMDefaultIPPool)
```
tommyp1ckles, line 221: What happens if a default pool is specified that doesn't exist?

chaunceyjiang: You will receive the following information, and the Pod cannot be created.
```
unable to allocate from pool test (family ipv4): pool not (yet) available
```

tommyp1ckles: Presumably that would happen on the first allocation attempt - in this case I wonder if it would be preferable to validate this early (@gandro)?

gandro: I have not yet tested it with this PR, but basically cilium-agent should not be come ready if the default pool (or more precise, any pool from which we want to pre-allocate) does not exist. This is similar to how it works with e.g. a misconfigured ENI mode (e.g. the local node requests pod IPs from a ENI subnet that does not exist)


https://github.com/cilium/cilium/blob/d6b1637880da13ab76c8465851c08c630e93f10c/pkg/ipam/multipool.go#L328-L330

pkg/defaults/defaults.go
```diff
@@ -384,6 +384,9 @@ const (
 	// pre-allocations
 	IPAMMultiPoolPreAllocation = "default=8"
```
gandro: Hm, so this is something I did not consider - it seems that we're still using `default` for the `ipam-multi-pool-pre-allocation` flag, which means that if a user only sets `ipam-default-ip-pool`, the local node will still attempt to pre-allocate from the `default` pool.

I would propose the following:
1. Let's set the default value of `ipam-multi-pool-pre-allocation` to the empty string [here](https://github.com/cilium/cilium/blob/c0c422d7389d78e3617b56a4efc40aee329c4833/daemon/cmd/daemon_main.go#L301)
2. In `option.Config.Populate()`, after we've parsed the flag, let's add a check that if `c.IPAMMultiPoolPreAllocation` is an empty map that we set it to `${c.IPAMDefaultIPPool}=8 [here](https://github.com/cilium/cilium/blob/d6b1637880da13ab76c8465851c08c630e93f10c/pkg/option/config.go#L3555)

That way, the default pool will be pre-allocated even if it's not named "default"

gandro (commented):
This is awesome work! Looks excellent to me overall. I've left some feedback regarding the pre-allocation

pkg/defaults/defaults.go
chaunceyjiang: Got it.

gandro: This can be removed now

```suggestion
```

gandro (commented):
Nice, thanks for the update! One last minor nit

pkg/defaults/defaults.go
chaunceyjiang: Updated.

daemon/cmd/daemon_main.go
```diff
@@ -297,10 +297,14 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 
 	flags.Var(option.NewNamedMapOptions(option.IPAMMultiPoolPreAllocation, &option.Config.IPAMMultiPoolPreAllocation, nil),
 		option.IPAMMultiPoolPreAllocation,
-		fmt.Sprintf("Defines the minimum number of IPs a node should pre-allocate from each pool (default %s)", defaults.IPAMMultiPoolPreAllocation))
-	vp.SetDefault(option.IPAMMultiPoolPreAllocation, defaults.IPAMMultiPoolPreAllocation)
+		fmt.Sprintf("Defines the minimum number of IPs a node should pre-allocate from each pool (default %s)", defaults.IPAMDefaultIPPool))
```
gandro: This needs to be:

```suggestion
		fmt.Sprintf("Defines the minimum number of IPs a node should pre-allocate from each pool (default %s)", ""))
```

Also, please don't forget to run `make -C Documentation update-cmdref` afterwards

chaunceyjiang: Don't we need to display the default value? Currently, `ipam-multi-pool-pre-allocation` still has a default value.

gandro: Oh, right I misread the code. However, we should display the =8 value still, i.e.
```suggestion
		fmt.Sprintf("Defines the minimum number of IPs a node should pre-allocate from each pool (default %s=8)", defaults.IPAMDefaultIPPool))
```

chaunceyjiang:
> Error: buildx failed with: ERROR: failed to solve: docker.io/library/alpine:3.18.4@sha256:eece025e432126ce23f223450a0326fbebde39cdf496a85d8c016293fc851978: failed to do request: Head "https://registry-1.docker.io/v2/library/alpine/manifests/sha256:eece025e432126ce23f223450a0326fbebde39cdf496a85d8c016293fc851978": net/http: TLS handshake timeout

This is a GitHub error.

chaunceyjiang:
It seems GitHub is having troubles. Let me close and re-open the PR, that should re-trigger the workflows.

tommyp1ckles (approved):

tommyp1ckles:
> This is a GitHub error.

Test image build may have failed causing a bunch of cascading test failures. Iet's retry.

daemon/cmd/daemon_main.go
```diff
@@ -297,10 +297,14 @@ func InitGlobalFlags(cmd *cobra.Command, vp *viper.Viper) {
 
 	flags.Var(option.NewNamedMapOptions(option.IPAMMultiPoolPreAllocation, &option.Config.IPAMMultiPoolPreAllocation, nil),
 		option.IPAMMultiPoolPreAllocation,
-		fmt.Sprintf("Defines the minimum number of IPs a node should pre-allocate from each pool (default %s)", defaults.IPAMMultiPoolPreAllocation))
-	vp.SetDefault(option.IPAMMultiPoolPreAllocation, defaults.IPAMMultiPoolPreAllocation)
+		fmt.Sprintf("Defines the minimum number of IPs a node should pre-allocate from each pool (default %s=8)", defaults.IPAMDefaultIPPool))
+	vp.SetDefault(option.IPAMMultiPoolPreAllocation, "")
 	option.BindEnv(vp, option.IPAMMultiPoolPreAllocation)
 
+	flags.String(option.IPAMDefaultIPPool, defaults.IPAMDefaultIPPool, "Name of the default IP Pool when using MultiPool")
```
tklauser: nit:
```suggestion
	flags.String(option.IPAMDefaultIPPool, defaults.IPAMDefaultIPPool, "Name of the default IP Pool when using multi-pool")
```

gandro:
- [ci-ipsec-e2e](https://github.com/cilium/cilium/actions/runs/6701825294/job/18209852541) failed with `no-missed-tail-calls`which I _think_ is a variant of known CI flake #28088 - in any case, it should be unrelated, restaging

pkg/defaults/defaults.go
```diff
@@ -380,9 +380,8 @@ const (
 	// CiliumNode.Spec.IPAM.PreAllocate if no value is set
 	IPAMPreAllocation = 8
 
-	// IPAMMultiPoolPreAllocation is the default value for multi-pool IPAM
-	// pre-allocations
-	IPAMMultiPoolPreAllocation = "default=8"
+	// IPAMDefaultIPPool is the default value for multi-pool names.
```
tklauser: nit:

```suggestion
	// IPAMDefaultIPPool is the default value for the multi-pool default pool name.
```

pkg/option/config.go
```diff
@@ -878,6 +878,9 @@ const (
 	// IPAMMultiPoolPreAllocation defines the pre-allocation value for each IPAM pool
 	IPAMMultiPoolPreAllocation = "ipam-multi-pool-pre-allocation"
 
+	// IPAMDefaultIPPool defines the default IP Pool when using MultiPool
```
tklauser:
```suggestion
	// IPAMDefaultIPPool defines the default IP Pool when using multi-pool
```

pkg/option/config.go
```diff
@@ -2162,7 +2165,8 @@ type DaemonConfig struct {
 
 	// IPAMMultiPoolPreAllocation defines the pre-allocation value for each IPAM pool
 	IPAMMultiPoolPreAllocation map[string]string
-
+	// IPAMDefaultIPPool the default IP Pool when using MultiPool
```
tklauser:
```suggestion
	// IPAMDefaultIPPool the default IP Pool when using multi-pool
```

tklauser (commented):
LGTM, only some small spelling nits.

tklauser (approved):
Thanks

Issue #28632, opened 2023-10-17
Cilium allocates random allocation CIDR on startup in pure multi-pool mode


CallMeFoxie:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

It seems that Cilium allocates a random IP on startup with `Using autogenerated IPv4 allocation range` message. We do not specify any CIDR locally as we use multi-pool mode.

This generated IP is now also announced via BGP, eventually going to hit our actual ranges and make a havoc on the network.

It would be nice if I could disable this (for me) random IP.

Thanks
Ashley

### Cilium Version

1.15.0 branch

### Kernel Version

5.15

### Kubernetes Version

1.27.1

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

CallMeFoxie:
Meaning I get something like
```
# cilium bgp routes advertised ipv4 unicast peer 10.255.20.111    
VRouter      Prefix           NextHop         Age        Attrs
4294967294   10.187.0.0/16    10.244.21.187   16h49m3s   [{Origin: i} {AsPath: } {Nexthop: 10.244.21.187} {LocalPref: 100}]   
4294967294   10.89.4.144/28   10.244.21.187   16h49m3s   [{Origin: i} {AsPath: } {Nexthop: 10.244.21.187} {LocalPref: 100}]   
4294967294   10.89.4.176/28   10.244.21.187   16h49m3s   [{Origin: i} {AsPath: } {Nexthop: 10.244.21.187} {LocalPref: 100}] 
```
where (obviously) the one is not something I want and very close to actual subnets...

CallMeFoxie:
For the record we use this argument
```
level=info msg="  --auto-create-cilium-pod-ip-pools='default=ipv4-cidrs:10.89.4.0/24;ipv4-mask-size:28;ipv6-cidrs:2a02:598:64:8a00::2080:0/105;ipv6-mask-size:120'" subsys=daemon
```

CallMeFoxie:
It seems that setting `exportPodCIDR` to `false` stops it at least from being announced on BGP, that's better than nothing :)

gandro:
This is related to https://github.com/cilium/cilium/issues/27018 

As a potential mitigation, we could teach the BGP code to ignore the local alloc CIDR in any IPAM mode other than `kubernetes` or `cluster-pool`, as all other IPAM modes (multi-pool, eni, etc) do not have a per-node podCIDR.

[...]

PR #28819, opened 2023-10-26
Docs: Adds CiliumPodIPPool Special Purpose Selectors

### LINKED ISSUE #28335: Docs: Add PodIPPoolSelector Special Selector to BGP CP
When https://github.com/cilium/cilium/pull/28314 merges, two special selectors will be introduced and should be documented.

danehans:
Adds usage of CiliumPodIPPool special purpose selectors to the BGP Control Plane documentation.

Fixes: #28335

### DIFF
--- a/Documentation/network/bgp-control-plane.rst
+++ b/Documentation/network/bgp-control-plane.rst
@@ -525,6 +525,16 @@ with a dummy key and value can be used like:
          matchExpressions:
             - {key: somekey, operator: NotIn, values: ['never-used-value']}
 
+There are two special purpose selector fields that match CiliumPodIPPools based on ``name`` and/or
+``namespace`` metadata instead of labels:
+
+=============================== ===================
+Selector                        Field
+------------------------------- -------------------
+io.cilium.podippool.namespace   ``.meta.namespace``
+io.cilium.podippool.name        ``.meta.name``
+=============================== ===================
+
 For additional details regarding CiliumPodIPPools, see the :ref:`ipam_crd_multi_pool` section.
 
 CLI
--- a/Documentation/spelling_wordlist.txt
+++ b/Documentation/spelling_wordlist.txt
@@ -534,6 +534,7 @@ pluggable
 pmdabcc
 png
 podCIDR
+podippool
 policyCIDRMatchMode
 policymap
 poller


harsimran-pabla (approved):
LGTM, I did some testing with multi-pool IPAM, one unrelated comment regarding CiliumPodIPPool advertisement. 

- If users are configuring CiliumPodIPPools, the way to advertise default pod CIDR is not via `exportPodCIDR` bool flag, but instead `default` CiliumPodIPPool must be advertised. It might be worth mentioning in the docs.

danehans:
@harsimran-pabla thanks for the review and feedback. Can you clarify your statement b/c that sounds like a bug? I would expect `exportPodCIDR` to be required to advertise `ciliumnode.spec.ipam.podCIDRs[]`. If you describe is expected behavior, then the API docs should be updated.

Thoughts @YutaroHayakawa @rastislavs

qmonnet (approved):
Looks all good from docs-structure's side.

Documentation/network/bgp-control-plane.rst
```diff
@@ -525,6 +525,16 @@ with a dummy key and value can be used like:
...
 
+There are two special purpose selector fields that match CiliumPodIPPools based on ``name`` and/or
+``namespace`` metadata instead of labels:
+
+=============================== ===================
+Selector                        Field
+------------------------------- -------------------
+io.cilium.podippool.namespace   ``.meta.namespace``
+io.cilium.podippool.name        ``.meta.name``
+=============================== ===================
```
qmonnet, line 536: Spell checker complains about these - let's quote them to make them literals.

```suggestion
=================================== ===================
Selector                            Field
----------------------------------- -------------------
``io.cilium.podippool.namespace``   ``.meta.namespace``
``io.cilium.podippool.name``        ``.meta.name``
=================================== ===================
```

qmonnet (changes requested):
(Requesting change per the above, to address CI's report.)

Documentation/network/bgp-control-plane.rst
danehans: @qmonnet thanks for the review. ^ changes are inconsistent with [existing docs](https://docs.cilium.io/en/latest/network/bgp-control-plane/#service-announcements). Instead of changing the existing docs to resolve the CI spellchecker, I added `podippool` to `Documentation/spelling_wordlist.txt`.

harsimran-pabla:
> @harsimran-pabla thanks for the review and feedback. Can you clarify your statement b/c that sounds like a bug? I would expect `exportPodCIDR` to be required for advertising `ciliumnode.spec.ipam.podCIDRs[]`. If you describe is expected behavior, then the API docs should be updated.
> 
> Thoughts @YutaroHayakawa @rastislavs

I was referencing to https://github.com/cilium/cilium/issues/28632#issuecomment-1776434166

I think @gandro  is also considering setting `ciliumnode.spec.ipam.podCIDRs[]` to be empty when using multi-pool ipam mode. 

Having an incorrect value in `ciliumnode.spec.ipam.podCIDRs[]` is a bug, however, we do need to document what is the correct way to advertise the default pool when multi-pool ipam mode is used. 

I don't want to derail this PR, we can open another issue to document default pool advertisement with BGP/mulit-pool ipam.

Documentation/network/bgp-control-plane.rst
qmonnet: I'd rather have the existing docs using inline literals too, it would make more sense than adding them as valid words in my opinion. But OK, works for me.

qmonnet (approved):

qmonnet:
Is the change relevant to any previous version among the supported branches (1.12/1.13/1.14)? If yes, can you please label the PR for backport accordingly?

danehans:
A back port is not required for this change since the documented feature was added post 1.15.

PR #28828, opened 2023-10-27
Modularization of loader

dylandreimerink:
This PR is the first in an effort to "clean up" the loader. This work un-exports all internal functions and types and exposes only the externally used functions via the Loader interface. This interface is implemented by a loader struct and is available as cell. This will allow us to make the loader dependent on other functionality and be depended upon, thus going from implicit dependencies to explicit dependencies.

```release-note
Modularization of loader
```


jschwinger233 (approved):
Makes sense overall.

daemon/cmd/daemon.go
```diff
@@ -1075,8 +1075,16 @@ func newDaemon(ctx context.Context, cleaner *daemonCleanup, params *daemonParams
 		return nil, nil, err
 	}
 
-	if err := loader.RestoreTemplates(option.Config.StateDir); err != nil {
-		log.WithError(err).Error("Unable to restore previous BPF templates")
+	// Simplest implementation: Just garbage-collect everything.
+	// In future we should make this smarter.
```
lmb, line 1020: I disagree: let's keep it dumb :laughing:

daemon/cmd/kube_proxy_replacement.go
```diff
@@ -448,8 +447,25 @@ func finishKubeProxyReplacementInit() error {
 	}
 
 	if option.Config.NodePortAcceleration != option.NodePortAccelerationDisabled {
-		if err := loader.SetXDPMode(option.Config.NodePortAcceleration); err != nil {
-			return fmt.Errorf("Cannot set NodePort acceleration: %w", err)
+		switch option.Config.NodePortAcceleration {
+		case option.XDPModeNative:
+			if option.Config.XDPMode == option.XDPModeLinkNone ||
```
lmb, line 446: This might read easier as a switch as well?

pkg/datapath/loader/hash.go
```diff
@@ -14,10 +14,10 @@ import (
 )
 
 var (
-	// DatapathSHA256 is set during build to the SHA across all datapath BPF
+	// datapathSHA256 is set during build to the SHA across all datapath BPF
 	// code. See the definition of CILIUM_DATAPATH_SHA256 in Makefile.defs for
 	// details.
-	DatapathSHA256 string
+	datapathSHA256 string
```
lmb, line 20: Doesn't the makefile need updating if we change the name?

pkg/datapath/types/loader.go
```diff
@@ -25,6 +27,9 @@ type Loader interface {
 	Reinitialize(ctx context.Context, o BaseProgramOwner, tunnelConfig tunnel.Config, deviceMTU int, iptMgr IptablesManager, p Proxy) error
 	HostDatapathInitialized() <-chan struct{}
 	DeviceHasTCProgramLoaded(hostInterface string, checkEgress bool) (bool, error)
+	ELFSubstitutions(ep Endpoint) (map[string]uint64, map[string]string)
```
lmb, line 32: Could this type just be an alias for `loader.Loader`?

pkg/datapath/loader/loader.go
```diff
@@ -70,7 +70,12 @@ type loader struct {
 	// templateCache is the cache of pre-compiled datapaths.
 	templateCache *objectCache
 
-	ipsecMu lock.Mutex // guards reinitializeIPSec
+	ipsecMu         lock.Mutex    // guards reinitializeIPSec
+	compilationLock *lock.RWMutex // guards compiler usage
```
lmb: Why not without the pointer type for consistency with ipsecMu?

lmb (changes requested):
Why does `loader: Make remaining global functions methods of the loader object` make sense? Seems like the functions don't actually end up accessing state from the loader object. Maybe some more explanation in the commit message would help.

aditighag (approved):
Looks good for my codeowners.

daemon/cmd/daemon.go
dylandreimerink: Not my comment, I fully agree! Lets not make things smarter than they have to be.

pkg/datapath/loader/hash.go
dylandreimerink: Good catch!

pkg/datapath/types/loader.go
dylandreimerink: No we can't, this type is duplicated to avoid an import cycle. One day, we will be able to remove it, but not yet.

Makefile.defs
```diff
@@ -138,7 +138,7 @@ endif
 # ROOT_DIR can be either `../` or absolute path, each of these need to be stripped
 BPF_SRCFILES := $(filter-out $(BPF_SRCFILES_IGNORE),$(subst ../,,$(subst $(ROOT_DIR)/,,$(BPF_SRCFILES))))
 CILIUM_DATAPATH_SHA256=$(shell cd $(ROOT_DIR); cat $(BPF_SRCFILES) | sha256sum | awk '{print $$1}')
-GO_BUILD_LDFLAGS += -X "github.com/cilium/cilium/pkg/datapath/loader.DatapathSHA256=$(CILIUM_DATAPATH_SHA256)"
+GO_BUILD_LDFLAGS += -X "github.com/cilium/cilium/pkg/datapath/loader.datapathSHA256=$(CILIUM_DATAPATH_SHA256)"
```
lmb, line 144: Is it possible to add a unit test which asserts that this isn't empty or similar?

lmb (approved):

dylandreimerink:
This PR has been replaced with #29823, same commits, but based on feature/devices-reconciliation-v1.16 + cherry-picked commits from this PR.

PR #28833, opened 2023-10-27
example/connectivity-check: fix port conflict, capture termination log

### LINKED ISSUE #28832: CI: random flakes due to NodePort / HostNetwork pod conflicts
In some of the Ginkgo tests, we spin up some HostNetwork pods. We also configure those pods to listen on a port in the NodePort range, for example, [here](https://github.com/cilium/cilium/blob/4fd2030660024d6eb6b1829b8d3308b7994619f8/examples/kubernetes/connectivity-check/echo-servers.cue#L75). So, those pods can randomly fail to start when they can't do a `listen()` on that port.

Here's an [example](https://github.com/cilium/cilium/actions/runs/6655070430/job/18084639710) of a failing test. Digging in to the dumps, and you see the following message from the pod that failed to start:

```
2023-10-26T14:04:01.053717793Z Some error occurred Error: listen EADDRINUSE: address already in use :::31002
2023-10-26T14:04:01.053728493Z     at Server.setupListenHandle [as _listen2] (node:net:1380:16)
2023-10-26T14:04:01.053733181Z     at listenInCluster (node:net:1428:12)
2023-10-26T14:04:01.053736988Z     at Server.listen (node:net:1516:7)
2023-10-26T14:04:01.053741106Z     at Function.listen (/usr/local/lib/node_modules/json-server/node_modules/express/lib/application.js:635:24)
2023-10-26T14:04:01.053744482Z     at /usr/local/lib/node_modules/json-server/lib/cli/run.js:135:20 {
2023-10-26T14:04:01.053747949Z   code: 'EADDRINUSE',
2023-10-26T14:04:01.053752237Z   errno: -98,
2023-10-26T14:04:01.053755954Z   syscall: 'listen',
2023-10-26T14:04:01.053759691Z   address: '::',
2023-10-26T14:04:01.053763749Z   port: 31002
2023-10-26T14:04:01.053766985Z }
```

Looking at the iptables output, I see that that port is also assigned to a NodePort service:
```
[0:0] -A KUBE-NODEPORTS -s 127.0.0.0/8 -p tcp -m comment --comment "default/test-nodeport-local-k8s2:http" -m tcp --dport 31002 -j KUBE-MARK-MASQ
```

Given that the NodePort range is 30000-32767, and the ephemeral port range starts at 32768, we should avoid those ranges entirely.

squeed:
It was discovered that these containers listen in the NodePort range by default, opening them up to random flakes if there's a NodePort service that happens to be there.

So, move the listening port out of that range.

Also, add a reasonable TerminationMessagePolicy so we can capture error status on exit.

Fixes: #28832

### DIFF
--- a/examples/kubernetes/connectivity-check/connectivity-check-hostport.yaml
+++ b/examples/kubernetes/connectivity-check/connectivity-check-hostport.yaml
@@ -25,6 +25,7 @@ spec:
         - containerPort: 8080
         image: quay.io/cilium/json-mock:v1.3.2@sha256:bc6c46c74efadb135bc996c2467cece6989302371ef4e3f068361460abaf39be
         imagePullPolicy: IfNotPresent
+        terminationMessagePolicy: FallbackToLogsOnError
         readinessProbe:
           timeoutSeconds: 7
           exec:
@@ -82,6 +83,7 @@ spec:
           hostPort: 40000
         image: quay.io/cilium/json-mock:v1.3.2@sha256:bc6c46c74efadb135bc996c2467cece6989302371ef4e3f068361460abaf39be
         imagePullPolicy: IfNotPresent
+        terminationMessagePolicy: FallbackToLogsOnError
         readinessProbe:
           timeoutSeconds: 7
           exec:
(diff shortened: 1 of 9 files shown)


tommyp1ckles (approved):
nice thanks!

PR #28834, opened 2023-10-27
Miscellaneous improvements to the etcd client

giorio94:
Please review commit by commit, and refer to the corresponding descriptions for additional information.

### DIFF
--- a/pkg/kvstore/etcd.go
+++ b/pkg/kvstore/etcd.go
@@ -15,7 +15,6 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/blang/semver/v4"
 	"github.com/sirupsen/logrus"
 	"go.etcd.io/etcd/api/v3/mvccpb"
 	v3rpcErrors "go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
@@ -41,7 +40,6 @@ import (
 	ciliumratemetrics "github.com/cilium/cilium/pkg/rate/metrics"
 	"github.com/cilium/cilium/pkg/spanstat"
 	"github.com/cilium/cilium/pkg/time"
-	"github.com/cilium/cilium/pkg/versioncheck"
 )
 
 const (
@@ -64,8 +62,6 @@ const (
 	// by ListAndWatch operations. A 0 value equals to no limit.
 	EtcdListLimitOption = "etcd.limit"
 
-	minRequiredVersionStr = ">=3.1.0"
-
 	etcdLockSessionRenewNamePrefix = "kvstore-etcd-lock-session-renew"
 
 	// etcdMaxKeysPerLease is the maximum number of keys that can be attached to a lease
@@ -87,11 +83,6 @@ type etcdModule struct {
 	config *client.Config
 }
 
-// versionCheckTimeout is the time we wait trying to verify the version
-// of an etcd endpoint. The timeout can be encountered on network
-// connectivity problems.
-const versionCheckTimeout = 30 * time.Second
-
 var (
 	// statusCheckTimeout is the timeout when performing status checks with
 	// all etcd endpoints
@@ -100,8 +91,6 @@ var (
 	// initialConnectionTimeout  is the timeout for the initial connection to
 	// the etcd server
 	initialConnectionTimeout = 15 * time.Minute
-
-	minRequiredVersion = versioncheck.MustCompile(minRequiredVersionStr)
 
 	// etcdDummyAddress can be overwritten from test invokers using ldflags
 	etcdDummyAddress = "http://127.0.0.1:4002"
@@ -618,10 +607,6 @@ func (e *etcdClient) renewLockSession(ctx context.Context) error {
 
 	e.logger.WithField(fieldSession, newSession).Debug("Renewing etcd lock session")
 
-	if err := e.checkMinVersion(ctx, versionCheckTimeout); err != nil {
-		return err
-	}
-
 	return nil
 }
 
@@ -750,10 +735,6 @@ func connectEtcdClient(ctx context.Context, config *client.Config, cfgPath strin
 
 			ec.logger.Info("Initial etcd session established")
 
-			if err = ec.checkMinVersion(ctx, versionCheckTimeout); err != nil {
-				return fmt.Errorf("unable to validate etcd version: %s", err)
-			}
-
 			return nil
 		}()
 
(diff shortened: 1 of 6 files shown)


giorio94:
Last commit introduced an additional check to prevent a possible source of flakiness.

giorio94:
Rebased onto main to fix conflict

tklauser (approved):
LGTM, but it's been a while since I last touched that part of the code base. I've requested a review from @marseel per https://github.com/cilium/cilium/pull/28834#issuecomment-1782950554

giorio94:
Rebased onto main to fix conflict

giorio94:
Rebased onto main to see if failures reproduce again (mainly conformance Gateway API, which seems totally unrelated).

pkg/kvstore/etcd.go
```diff
@@ -822,59 +803,13 @@ func makeSessionName(sessionPrefix string, opts *ExtraOptions) string {
...
 	e.RWMutex.RLock()
 	err = e.sessionErr
 	e.RWMutex.RUnlock()
 	return
 }
 
-// checkMinVersion checks the minimal version running on etcd cluster.  This
-// function should be run whenever the etcd client is connected for the first
-// time and whenever the session is renewed.
-func (e *etcdClient) checkMinVersion(ctx context.Context, timeout time.Duration) error {
```
marseel, line 849: It definitely makes sense to remove version validation here. Maybe we could follow-up in cilium-cli to get Etcd version and print some warning in case it's old/unsupported version? While it doesn't matter for clustermesh too much, for kvstore mode it could make sense. If it's fine with you, maybe let's create follow-up issue in cilium-cli?

pkg/kvstore/etcd_lease.go
```diff
@@ -69,12 +63,26 @@ func newEtcdLeaseManager(cl etcdLeaseClient, ttl time.Duration, limit uint32, ex
...
 func (elm *etcdLeaseManager) GetLeaseID(ctx context.Context, key string) (client.LeaseID, error) {
+	session, err := elm.GetSession(ctx, key)
+	if err != nil {
+		return client.NoLease, err
+	}
+
+	return session.Lease(), nil
+}
+
+// GetLeaseID returns a session, and associates it to the given key. It leverages
```
marseel: nit: s/GetLeaseID/GetSession/

pkg/kvstore/etcd_lease.go
```diff
@@ -199,57 +209,62 @@ func (elm *etcdLeaseManager) Wait() {
 	elm.wg.Wait()
 }
 
-func (elm *etcdLeaseManager) newLease(ctx context.Context) (c client.LeaseID, cancelFn context.CancelFunc, err error) {
+func (elm *etcdLeaseManager) newLease(ctx context.Context) (session *concurrency.Session, err error) {
```
marseel: nit: newLease/newSession

pkg/kvstore/etcd.go
```diff
@@ -359,11 +352,13 @@ type etcdClient struct {
...
 	lock.RWMutex
 
-	sessionErr        error
-	lockSession       *concurrency.Session
-	lockSessionCancel context.CancelFunc
+	sessionErr error
 
+	// leaseManager manages the lazy acquisition of etcd leases for generic purposes
 	leaseManager *etcdLeaseManager
+	// lockLeaseManager manages the lazy acquisition of etcd leases for locking
```
marseel: I would probably remove `lazy` from the comments here as it's implementation detail, not important in current context imho.

marseel (commented):
Looks good, I will go over this PR one more time to make sure I checked all edge cases.

pkg/kvstore/etcd.go
giorio94: :thinking: Are you thinking about checking that at installation time or when doing `cilium status`? The main issue I see is that it is tricky to ensure that the CLI is always able to access the etcd members, independently of whether they are outside or inside the cluster. And I'm not 100% sure whether the effort is really worth considering that Cilium is currently compatible with any supported version of etcd.

marseel: I was thinking about doing that in `cilium status`. For now, just creating issue in cilium-cli to capture that we might want to introduce this check. Even though we support all versions, I think it would be nice to suggest upgrades etc in case users are running Etcd older than 3.4, or 3.5 with [known data inconsistency issues](https://groups.google.com/a/kubernetes.io/g/dev/c/B7gJs88XtQc/m/rSgNOzV2BwAJ) (versions 3.5.0-3.5.2)

pkg/kvstore/etcd_lease.go
```diff
@@ -94,7 +102,7 @@ func (elm *etcdLeaseManager) GetLeaseID(ctx context.Context, key string) (client
 			elm.keys[key] = elm.current
 			elm.mu.Unlock()
 
-			return elm.current, nil
+			return info.session, nil
```
marseel, line 123: We should check here if sessions in not Orphaned I think.
in https://github.com/cilium/cilium/pull/28834/files#diff-85b5ad3b5a6b7004aafad37cdbb264e395fd491231f4c8cae8ce88554e7d1145R193
we can Orphan session, but it potentially can still be in `elm.leases` as we release lock after that and remove it later on in `waitForExpiration`

marseel: I was thinking about something like this, to make it less likely to happen:
```
select {
case <-info.session.Done():
    continue
default:
}
```

giorio94: The main issue is that we cannot completely get rid of the uncertainty period, due to the delay between server-side expiration and client-side detection (with consequent removal of the session from the list of known ones). While additional checks could help slightly reducing that window, they also make the logic more complex (as they would break the assumption that sessions get GCed only by the `waitForExpiration` function). For this reason, the idea here is to essentially rely on client-side retries in the unlikely case that the session is expired. 

As discussed offline, I've extended the comments of the GetLease/GetSession functions for additional clarity about that assumption: https://github.com/cilium/cilium/pull/28834/commits/48518d6781c67198f62fbd0db094c820309bf9cf

pkg/kvstore/etcd.go
giorio94: https://github.com/cilium/cilium-cli/issues/2107

giorio94:
Rebased onto main to make CI happy

marseel (approved):
Looks good, thanks!

PR #28840, opened 2023-10-27
pkg/k8s: do not perform an early return on ipcache errors

aanm:
Although ipcache might return an error, we should not perform an early
return in the event handling of the pod event because some other
components that are unrelated to ipcache might depend on this specific
event per perform correctly. Instead we will log the ipcache error and
let the execution of the method to continue.

The pod might, for example, be in a "Pending" state which has its
status IP addresses not set but might be able to change labels in the
meanwhile which Cilium should react upon those changes which are not
related with ipcache.

```release-note
Improved event handling for pod events by removing an unnecessary early return, allowing unrelated components to execute correctly, while enhancing ipcache error logging.
```

### DIFF
--- a/pkg/k8s/watchers/pod.go
+++ b/pkg/k8s/watchers/pod.go
@@ -334,11 +334,11 @@ func (k *K8sWatcher) updateK8sPodV1(oldK8sPod, newK8sPod *slim_corev1.Pod) error
 
 	oldPodIPs := k8sUtils.ValidIPs(oldK8sPod.Status)
 	newPodIPs := k8sUtils.ValidIPs(newK8sPod.Status)
-	err = k.updatePodHostData(oldK8sPod, newK8sPod, oldPodIPs, newPodIPs)
-
-	if err != nil {
-		logger.WithError(err).Warning("Unable to update ipcache map entry on pod update")
-		return err
+	if len(oldPodIPs) != 0 || len(newPodIPs) != 0 {
+		err = k.updatePodHostData(oldK8sPod, newK8sPod, oldPodIPs, newPodIPs)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


joestringer (approved):

pkg/k8s/watchers/pod.go
```diff
@@ -338,7 +338,6 @@ func (k *K8sWatcher) updateK8sPodV1(oldK8sPod, newK8sPod *slim_corev1.Pod) error
 
 	if err != nil {
 		logger.WithError(err).Warning("Unable to update ipcache map entry on pod update")
-		return err
```
joestringer: I think that ipcache should receive this info and eventually reconcile the correct state corresponding to this new information.

What concerns me a little is that ipcache might be ignoring the state and then other parts of Cilium calculate with this in mind, and that somehow leads to out-of-sync issues.

I'm also not the biggest fan of having a bare warning. What will users do about it? They can certainly report it to the developers. I guess it's better to let them know that there was some sort of state reconciliation issue here, but it'd be even better to just not have the problem in the first place.

joamaki: It's hard to say what this change does given the many different errors, from parse errors to service manager errors (lbmap BPF errors might bubble up here?). Not doing an early return leaves things in a weird half-broken state, but also not continuing has it's issues as it seems like the error could also be minor (say one of the named ports has a minor typo), or transient (say lbmap is full temporarily). The fact that there's also no retrying here may then cause permanently broken state for this pod.

So it does seem to me it is indeed safer to not do an early return, however there's clearly some major issues in handling edge conditions that need to be resolved. Let's have a discussion on this this week.

joamaki (approved):

